PR #656, opened 2019-04-12
Rename 2020-04-12-mercury.md to 2019-04-12-mercury.md

antonio-ramadas:
Let's not get ahead of ourselves. We're still in 2019 😄


olafurpg:
Thank you for this contribution! I’m afraid this change causes the current URL to become a 404, maybe we can change it sometime in the future but since this post was published and shared today I think it’s best to keep the 2020 date.

PR #671, opened 2019-04-15
Update stable version in issue template

gabro:


gabro:
The failed Travis build is on the `push` event, because I initially pushed this branch on the main repo by mistake, and I canceled the build right away to avoid noise.

olafurpg:
Thanks!

Issue #667, opened 2019-04-15
Tests on Appveyor CI are consistently failing


olafurpg:
The Appveyor CI consistently fails now with 1-2 random tests (usually cyclicng between the same group of ~5 tests). We should either fix/disable the flaky tests or switch to Azure so that our CI status checks become reliably green again.

olafurpg:
CI has been fairly stable after #676, optimistically closing this for now.

olafurpg:
Reopening since tests are failing on Appveyor again `tests.ReferenceSlowSuite.edit-distance`

olafurpg:
Tests are also a bit flaky on Azure, I feel like we need to improve the robustness of BSP connection establishment.

olafurpg:
Appveyor has been quite reliable lately, closing.

PR #676, opened 2019-04-15
Disable flaky test on Appveyor.

olafurpg:
This test has repeatedly failed on Appveyor, let's see if disabling it
helps stabilize the CI. Towards #667


olafurpg:
That's the first green build on Appveyor in a while, let's hope this PR did the trick, although I suspect we still have a few more flaky tests on Windows.

PR #683, opened 2019-04-19
fix incorrect date for v0.5.0 release blog post

strobe:
It seems like original *.md file incorrectly named with 2020 instead 2019.


olafurpg:
Thank you for this contribution! This is a duplicate of #656, changing the date would result in the post URL to become a 404

PR #686, opened 2019-04-20
Handle document symbols for nested declarations

gabro:
Before this PR, we would only consider document symbols defined as children of classes, objects, or traits.

This caused interesting symbols not to appear in the editor outline.

For example, consider this (quite common) pattern:

```scala
trait A {
  def m(): Unit
}
object A {
  def create: A = new A {
    def m(): Unit = ()
  }
}
```

Prior to this PR, we would show a document symbol for the *declaration* of `m`, but not for its *definition*.

This PR adds document symbols defined in defs, vals, vars, and `new` definitions.

Here's a fairly exhaustive example.

### Before
image

### After
image

A few things to notice about the anonymous `new` definitions:

- if the template node has >= 1 inits (e.g. `new Foo with Bar {}`) we show `new Foo with Bar` (see the `create` and `createWithLogging` examples in the screenshot above)

- if the template node has 0 inits (e.g. `new {}`) we show `new (anonymous)` (see the `val constants` example in the screenshot above)

- if the template node has no statements e.g. `new Foo {}` we skip it (see the `createDefault` example in the screenshot above)

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
@@ -93,7 +93,17 @@ class DocumentSymbolProvider(trees: Trees) {
         case t: Pkg =>
           addChild(t.ref.syntax, SymbolKind.Package, t.pos, t.ref.pos, "")
           newOwner()
-        case _: Source | _: Template =>
+        case t: Term.NewAnonymous =>
+          val (name, selection) = t.templ.inits match {
+            case Nil => ("(anonymous)", t.pos)
+            case inits =>
+              (inits.map(_.tpe.syntax).mkString(" with "), inits.head.pos)
+          }
+          if (t.templ.stats.nonEmpty) {
+            addChild(s"new $name", SymbolKind.Interface, t.pos, selection, "")
+            newOwner()
+          } else continue()
+        case _: Source | _: Template | _: Term.Block =>
           continue()
         case t: Defn.Class =>
           addChild(
@@ -139,6 +149,7 @@ class DocumentSymbolProvider(trees: Trees) {
             t.name.pos,
             t.decltpe.fold("")(_.syntax)
           )
+          newOwner()
         case t: Decl.Def =>
           addChild(
             t.name.value,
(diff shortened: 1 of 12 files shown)


metals/src/main/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
```diff
@@ -93,7 +93,17 @@ class DocumentSymbolProvider(trees: Trees) {
         case t: Pkg =>
           addChild(t.ref.syntax, SymbolKind.Package, t.pos, t.ref.pos, "")
           newOwner()
-        case _: Source | _: Template =>
+        case t: Term.NewAnonymous =>
+          val (name, selection) = t.templ.inits match {
+            case Nil => ("(anonymous)", t.pos)
+            case inits =>
+              (inits.map(_.tpe.syntax).mkString(" with "), inits.head.pos)
```
gabro: I'm definitely not sure about that `inits.heads.pos`

metals/src/main/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
```diff
@@ -93,7 +93,17 @@ class DocumentSymbolProvider(trees: Trees) {
...
+          }
+          if (t.templ.stats.nonEmpty) {
+            addChild(s"new $name", SymbolKind.Interface, t.pos, selection, "")
+            newOwner()
+          } else continue()
+        case _: Source | _: Template | _: Term.Block =>
```
gabro: `Term.Block` is needed in order to find `b` in the following example

```scala
def a = {
  def b = ()
  ()
}
```

metals/src/main/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
olafurpg: You can do `inits.headOption.fold(t.pos)(_.pos)` to sleep good at night. Feel free to merge!

gabro: yeah, the doubt is mostly about the position rather than the code itself.

I'll merge anyway, we can always fix it later.

PR #688, opened 2019-04-21
Update a couple .lines to .linesIterator for java11

er1c:


olafurpg:
Thank you for this contribution! Please note that there is more work left to add Java 11 support, see https://github.com/scalameta/metals-feature-requests/issues/19 for updated.

Issue #653, opened 2019-04-12
Add automatic build import for Gradle Scala projects


tgodzik:
Currently, you need go through several manual installation steps to use Metals with Gradle project. It would be nice if Metals provided the same user experience as with sbt:
> - When user opens new Gradle project: "Detected new Gradle workspace. Would you like to import the build? Yes / No / Don't ask again"
> - When user changes an existing Gradle project: "Would you like to import changes? Yes / No / Don't ask again"

Some challenges:
- bloop gradle plugin doesn't download source dependencies - should be fixed maybe using the eclipse plugin in the gradle bloop plugin itself
- detecting when gradle had actual changes to it - there might be multiple files in the project that define the build - most likely just *.gradle and *.gradle.kts - we could create a list when running gradle and put it into .metals directory

tgodzik:
This has just been merged and works pretty well.

PR #694, opened 2019-04-24
Add gradle support to imports in metals and refactor build tool support.

tgodzik:
Previously, we could only use sbt to work with metals, but now gradle builds are also supported via gradle bloop plugin.

This is work on https://github.com/scalameta/metals/issues/653


metals/src/main/scala/scala/meta/internal/builds/BuildTool.scala
```diff
@@ -0,0 +1,40 @@
...
+      userConfig: () => UserConfiguration
+  ): List[String]
+
+  def digest(workspace: AbsolutePath): Option[String]
+
+  def version: String
+
+  def minimumVersion: String
+
+  def isCompatibleVersion: Boolean = {
```
olafurpg: nit: can we move this into static methods on an object?

metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
```diff
@@ -0,0 +1,141 @@
...
+       |          }
+       |          dependencies {
+       |              scalaCompilerPlugin 'org.scalameta:semanticdb-scalac_' + scalaVersion + ':4.1.4'
+       |          }
+       |          String semanticDb = configurations.scalaCompilerPlugin.asPath.split(':').find{it.contains('semanticdb')}
+       |          if (!semanticDb) {
+       |              throw new RuntimeException("SemanticDB plugin not found!")
+       |          }
+       |          tasks.withType(ScalaCompile) {
+       |              def params = [
```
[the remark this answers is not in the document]
tgodzik: Yes, it's only added explicitly when providing init script, which is in a tmp directory.

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -0,0 +1,252 @@
...
+  protected def digestGeneralJvm(
+      file: AbsolutePath,
+      digest: MessageDigest
+  ): Boolean = {
+    try {
+      Source
+        .fromFile(file.toFile)
+        .getLines()
+        .mkString("\n")
+        .replaceAll("""/\*([^*]|[\r\n]|(\*+([^*/]|[\r\n])))*\*+/""", "") // match anything between /**/ comments and remove
```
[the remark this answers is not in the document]
tgodzik: Yep, that's the regex to remove /**/ comments. Aside of normal tokenization I don't think there is a simpler way to do it. :P

metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
```diff
@@ -54,31 +54,41 @@ final class BuildTools(
...
+    if (isGradle) buf += "Gradle"
+    if (isMaven) buf += "Maven"
+    if (isPants) buf += "Pants"
+    if (isBazel) buf += "Bazel"
     buf.result()
   }
   def isEmpty: Boolean = {
     all.isEmpty
   }
+  def get(): Option[BuildTool] = {
```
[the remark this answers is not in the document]
tgodzik: 'BuildTools.loadSupported()' seems sensible to me, `BuildTools.loadSupportedBuildTool()` on the other hand is a bit exesive.

tgodzik:
> What is the expected behavior for Java-only Gradle builds? Would it be possible to automatically detect Gradle builds that don't use the Scala plugin?

So if we don't have a scala library then nothing will happen when importing, but it might pop up with an option to import the build - that should not be a problem since we can dismiss it. 
I don't think we can automatically detect whether project has scala - gradle needs to run the configuration for that. Even src/scala might not exist since gradle is highly configurable - altough I will think about what is the best way to do it here.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -36,7 +36,7 @@ class Messages(icons: Icons) {
       new MessageActionItem("Import changes")
     def params: ShowMessageRequestParams = {
       val params = new ShowMessageRequestParams()
-      params.setMessage("sbt build needs to be re-imported")
```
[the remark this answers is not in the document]
tgodzik: It's used in places where we don't have info about workspace - this could be possible, but I think it's more work than it's worth

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -51,6 +52,17 @@ object UserConfiguration {
         |like using environment variables.
         |""".stripMargin
     ),
+    UserConfigurationOption(
+      "gradle-script",
+      """empty string `""`.""",
+      "/usr/local/bin/gradle",
+      "gradle script",
+      """Optional absolute path to a `gradle` executable to use for running `gradle bloopInstall`.
+        |By default, Metals uses `java -jar sbt-launch.jar` with an embedded launcher while respecting
```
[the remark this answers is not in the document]
tgodzik: I was sure I changed that :octopus:

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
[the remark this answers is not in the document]
olafurpg: What about passing in the build tool name as a string? I think it would be helpful to know from the message what kind of build was detected,some workspaces might have both a gradle and sbt build (although I suspect it's uncommon).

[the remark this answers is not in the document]
tgodzik: Changed to string and reworked the logic behind the fake LSP client - I only check for what the message contains.

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -0,0 +1,251 @@
...
+)
+
+object Digest {
+
+  /**
+   * Bump up this version if parameters outside of the sbt sources themselves require
+   * re-running `bloopInstall`. For example a SemanticDB or Bloop version upgrade.
+   */
+  val version: String = "v4"
+  sealed abstract class Status(val value: Int)
```
marek1840: why not something like:
```scala
final class Status private(val value: Int){
  def ...
}

object Status{
  val Requested = new Status(0)
  val Started = new Status(1)
  ...
}
```

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  def digestDirectory(
+      path: AbsolutePath,
+      digest: MessageDigest
+  ): Boolean = {
+    if (!path.isDirectory) true
+    else {
+      var isSuccess = true
+      Files.list(path.toNIO).forEach { file =>
```
marek1840: maybe `reduce` or even some kind of `forAll` would be simpler/more readable?

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -0,0 +1,251 @@
...
+    }
+    true
+  }
+
+  def digestFile(
+      path: AbsolutePath,
+      digest: MessageDigest
+  ): Boolean = {
+    val ext = PathIO.extension(path.toNIO)
+    val isScala = ext match {
```
marek1840: would `ext == "sbt" || ext == "scala"` be more readable?
or even `Set("scala", "sbt").contains(ext)` -  since there already is a precedence of using sets with `isGradle` ;)

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -0,0 +1,251 @@
...
+      case "sbt" | "scala" => true
+      case _ => false
+    }
+    // we can have both gradle and gradle.kts and build plugins can be written in any of three languages
+    val isGradle =
+      Set("gradle", "groovy", "gradle.kts", "java", "kts").exists(
+        path.toString().endsWith(_)
```
marek1840: why ` path.toString().endsWith(_)` when there is an `ext` ?

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -0,0 +1,251 @@
...
+      )
+    if (isScala && path.isFile) {
```
marek1840: what do you think about a new line here? to clearly separate the "set up" stage from the "act" one ;)

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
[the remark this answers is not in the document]
marek1840: I hope it is **thoroughly** tested ;P

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -0,0 +1,251 @@
...
+    }
+    buildSrcDigest && Digest.digestDirectory(workspace, digest) && digestSubProjects(
+      workspace,
+      digest
+    )
+  }
+
+  def digestBuildSrc(path: AbsolutePath, digest: MessageDigest): Boolean = {
+    var isSuccessful = true
+    Files.walk(path.toNIO).forEach { file =>
```
marek1840: Second time - maybe it could be extracted  as some kind of `def fileIterator(path): Iterator[AbsolutePath]` and then we could use `fileIterator(path).forAll(file => digestFile(file, digest))`

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -0,0 +1,251 @@
...
+      workspace: AbsolutePath,
+      digest: MessageDigest
+  ): Boolean = {
+    val (subprojects, dirs) = Files
+      .list(workspace.toNIO)
+      .filter(AbsolutePath(_).isDirectory)
+      .collect(Collectors.toList[Path])
+      .asScala
+      .partition { file =>
+        AbsolutePath(file).isDirectory && Files
```
marek1840: isn't line `221` already checking if the file is a directory?

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -0,0 +1,251 @@
...
+          .list(file)
+          .anyMatch { path =>
+            val stringPath = path.toString
+            stringPath.endsWith(".gradle") || stringPath.endsWith("gradle.kts")
+          }
+      }
+    /*
+     If a dir contains a gradle file we need to treat is as a workspace
+     */
+    val isSuccessful = subprojects.foldLeft(true) {
```
marek1840: `subprojects.forall` ?

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -0,0 +1,251 @@
...
+    isSuccessful
+  }
+
+  def digestSubProjects(
+      workspace: AbsolutePath,
+      digest: MessageDigest
+  ): Boolean = {
+    val (subprojects, dirs) = Files
+      .list(workspace.toNIO)
+      .filter(AbsolutePath(_).isDirectory)
```
marek1840: `AbsolutePath(file).isDirectory` -> `Files.isDirectory(file)`. No need to create an object unnecessarily ;)

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -0,0 +1,251 @@
...
+        success && digestWorkspace(
+          AbsolutePath(file),
+          digest
+        )
+    }
+
+    /*
+     If it's a dir we need to keep searching since gradle can have non trivial workspace layouts
+     */
+    dirs.foldLeft(isSuccessful) {
```
marek1840: `forall`

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -960,52 +959,46 @@ class MetalsLanguageServer(
...
-        if (!buildTools.isAutoConnectable) {
-          warnings.noBuildTool()
-        }
-        Future.successful(BuildChange.None)
-      case Some(sbt) =>
-        if (!isCompatibleSbtVersion(sbt.version)) {
+
+    buildTools.loadSupported match {
+      case Some(buildTool) =>
+        if (!BuildTool.isCompatibleVersion(
```
marek1840: I find it easier to reason about an `if` when the condition is not negated

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
tgodzik: ext doesn't work with `.gradle.kts`

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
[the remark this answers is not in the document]
tgodzik: It is tested in the Suites :D

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
tgodzik: This is basically the same and lacks ability to sensibly pattern match. I would leave it as is.

jvican:
@tgodzik Great work here. Have you tried the integration with real-world Gradle builds? I think it would be useful to give https://github.com/salesforce/TransmogrifAI a try with Metals and see how it behaves. It should be a good test on the bloop side to guarantee exporting the build from Gradle works out-of-the-box.

tgodzik:
> @tgodzik Great work here! Have you tried the integration with real-world Gradle builds? I think it would be useful to give https://github.com/salesforce/TransmogrifAI a try with Metals and see how it behaves. It should be a good test on the bloop side to guarantee exporting the build from Gradle works out-of-the-box.

I trying it and I think I overdid something with the digests. Will check what's going on. Thanks for the suggestion!

tests/slow/src/test/scala/tests/GradleSlowSuite.scala
```diff
@@ -0,0 +1,330 @@
...
+        // only 3 projects since no empty test targets are created for gradle
+        CheckDoctor.multipleMisconfiguredProjects(3)
+      )
+      _ <- Future.sequence(
+        ('a' to 'f')
+          .map(project => s"$project/src/main/scala/a/A.scala")
+          .map(file => server.didOpen(file))
+      )
+      _ = assertNoDiff(client.workspaceDiagnostics, "")
+      _ = assertNoDiff(
```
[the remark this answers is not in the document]
tgodzik: fixed it , just test if all compiles and works

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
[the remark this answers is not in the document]
tgodzik: this actually is a bit more fragile than suspected, removed it for now and will add a test suite

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -54,7 +54,7 @@ class Messages(icons: Icons) {
     def params: ShowMessageRequestParams = {
       val params = new ShowMessageRequestParams()
       params.setMessage(
-        "New sbt workspace detected, would you like to import the build?"
```
olafurpg: Can we keep the build tool name in this message as well?

Also, when running "Import build" for Gradle workspaces the `metals/slowTask` message is "sbt bloopInstall"

Screenshot 2019-05-02 at 19 27 47

tgodzik: Will change that, lost that one message

tgodzik:
> is semanticdb-scalac only enabled in Scala projects? If so, the Metals doctor might be buggy and report warnings in Java-only projects

Yes, it's not enabled in java projects, should we fix how the doctor handles it here?

> Compiling the clients project causes cryptic compile errors in Bloop that have no attached position

Custom `processMessages` task is needed unfortunately.

olafurpg:
Once the `New workspace` and `sbt bloopInstall` message have been fixed then I believe this PR is ready to go!

> Yes, it's not enabled in java projects, should we fix how the doctor handles it here?

Looking at the JSON logs it looks difficult to fix on our side since Bloop returns `scalacOptions` for the Java-only module `clients` as well as `languageIds=[scala,java]`. This problem does not happen in sbt-metals since we `addCompilerPlugin(semanticdb-scalac)` even in Java-only projects (which ends up being ignored).

```json
{
        "id": {
          "uri": "file:/Users/olafurpg/dev/kafka/clients/?id\u003dclients"
        },
        "displayName": "clients",
        "baseDirectory": "file:///Users/olafurpg/dev/kafka/clients/",
        "tags": [
          "library"
        ],
        "languageIds": [
          "scala",
          "java"
        ],
        "dependencies": [],
        "capabilities": {
          "canCompile": true,
          "canTest": true,
          "canRun": true
        },
        "data": {
          "scalaOrganization": "org.scala-lang",
          "scalaVersion": "2.12.8",
          "scalaBinaryVersion": "2.12",
          "platform": 1,
          "jars": [
            "file:///Users/olafurpg/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.12.8/scala-library-2.12.8.jar",
            "file:///Users/olafurpg/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-reflect/2.12.7/scala-reflect-2.12.7.jar",
            "file:///Users/olafurpg/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-compiler/2.12.7/scala-compiler-2.12.7.jar",
            "file:///Users/olafurpg/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/modules/scala-xml_2.12/1.1.0/scala-xml_2.12-1.1.0.jar",
            "file:///Users/olafurpg/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/jline/jline/2.14.4/jline-2.14.4.jar"
          ]
        }
      }
```
Either way, this can be fixed separately. This PR is already a huge improvement with the new abstractions over the build tool. Great job @tgodzik

PR #706, opened 2019-04-30
Miscellaneous improvements

olafurpg:
- bump local publish version to 0.6.0-SNAPSHOT
- upgrade lsp4j dependency
- make "connect to build server" more robust to errors
- drop support for deprecated Scala versions
- avoid log messages with noisy stack traces when running full test
suite.
- fix potential flakiness issues in ReferenceSlowSuite


olafurpg:
I'm hoping some of these changes may improve the flak test failures we're seeing both on appveyor and azure.

PR #712, opened 2019-05-07
Add keyword completion

markus1189:
Adds keyword completion 

especially nice:
  - complete `lazy val`, `sealed trait`, `sealed abstract class`

Greetings from the OpenSource Spree at flatMap(Oslo) 2019!


mtags/src/main/scala/scala/meta/internal/pc/Keywords.scala
```diff
@@ -0,0 +1,139 @@
...
+    List("case class", "class", "trait", "object")
+
+  private[this] def definitionModifiers =
+    List("final", "private", "protected", "sealed trait",
+      "sealed abstract class", "abstract")
+
+  // Categories
+
+  private[this] def expressionKeywords: List[String] =
+    List("super", "this", "if", "for", "while", "do", "true", "false", "new",
```
tgodzik: If we have try then maybe `catch` too? And we could consider apply as being useful to include

mtags/src/main/scala/scala/meta/internal/pc/Keywords.scala
```diff
@@ -0,0 +1,139 @@
...
+
+  private[this] def getIdentifierName(
+      enclosing: List[Tree],
+      pos: Position
+  ): Option[Name] = {
+    enclosing match {
+      case Ident(name) :: _ => Some(name)
+      case PackageDef(ref, _) :: _ if !ref.pos.includes(pos) =>
+        var i = pos.point
+
```
tgodzik: Could we drop the additional newlines?

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -43,7 +43,8 @@ object CompletionMatchSuite extends BaseCompletionSuite {
       |case object Weekend extends Weekday
       |@@
       |""".stripMargin,
-    ""
+    "",
+    filter = _ => false
```
tgodzik: Is this used any other way than `_ => false` or `_=> true` ? It could just be a flag then

mtags/src/main/scala/scala/meta/internal/pc/Keywords.scala
```diff
@@ -0,0 +1,139 @@
...
+  private[this] def blockKeywords: List[String] =
+    List("implicit") ++ expressionKeywords ++ blockDefinitions
+
+  private[this] def templateKeywords: List[String] =
+    List("implicit", "final") ++ expressionKeywords ++ blockDefinitions ++ toplevelDefinitions ++ definitionModifiers
+
+  private[this] def toplevelKeywords: List[String] =
+    List("package", "import") ++ toplevelDefinitions ++ definitionModifiers
+
+  private[this] def defKeywords: List[String] = List("return")
```
gabro: also, `throw`?

mtags/src/main/scala/scala/meta/internal/pc/Keywords.scala
```diff
@@ -0,0 +1,139 @@
...
+      case _ => false
+    }
+
+  private[this] def isBlock(enclosing: List[Tree]): Boolean =
+    enclosing match {
+      case Ident(_) :: Block(_, _) :: _ => true
+      case _ => false
+    }
+
+  private[this] def addTrailingSpace(keyword: String): Boolean =
```
gabro:
```suggestion
  private[this] def needsTrailingSpace(keyword: String): Boolean =
```

mtags/src/main/scala/scala/meta/internal/pc/Keywords.scala
olafurpg: Context-sensitive keywords like `catch`, `finally`, `yield` and `extends` can be left for a separate PR.

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
olafurpg: The `filter: String => Boolean` parameter is identical to the `checkEditLine` method.

mtags/src/main/scala/scala/meta/internal/pc/Keywords.scala
markus1189: done

mtags/src/main/scala/scala/meta/internal/pc/Keywords.scala
markus1189: good catch :) added as an expression keyword

mtags/src/main/scala/scala/meta/internal/pc/Keywords.scala
markus1189: I agree that this sounds better ;)

mtags/src/main/scala/scala/meta/internal/pc/Keywords.scala
markus1189: apply is not a keyword though, so I would not include it here

olafurpg:
I haven’t forgotten about this pr! I have a wip local branch that refractors the code a bit and adds additional keyword completions in different corner cases. It won’t be ready until next week however.

mtags/src/main/scala/scala/meta/internal/pc/Keyword.scala
```diff
@@ -0,0 +1,101 @@
...
+    Keyword("object", isTemplate = true, isPackage = true, isDefinition = true),
+    Keyword("package", isPackage = true),
+    Keyword("import", isBlock = true, isTemplate = true, isPackage = true),
+    Keyword("final", isTemplate = true, isPackage = true),
+    Keyword("private", isTemplate = true, isPackage = true),
+    Keyword("protected", isTemplate = true, isPackage = true),
+    Keyword("abstract class", isTemplate = true, isPackage = true),
+    Keyword("sealed trait", isTemplate = true, isPackage = true),
+    Keyword("sealed abstract class", isTemplate = true, isPackage = true),
+    Keyword("sealed class", isTemplate = true, isPackage = true),
```
olafurpg: @markus1189 FYI, I added `sealed class` since it's strictly speaking valid although less common

markus1189:
@olafurpg And thank you as well for the session, it was great fun!

PR #720, opened 2019-05-15
WIP: implement textDocument/codeLens - run 

marek1840:
Previously, metals did not provide any code lens for running the main classes.

Now, it places them right above the `object` containing/inheriting a main method


metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -39,9 +39,16 @@ object ClientCommands {
...
+    "run",
+    "runs code"
+  )
+
   def all: List[Command] = List(
     RunDoctor,
     ToggleLogs,
-    FocusDiagnostics
+    FocusDiagnostics,
+    RunCode
```
olafurpg: This list is used to auto-generate the documentation on the website. If users are not expected to trigger this command manually then we can leave it out of the list. 

A better name for the variable instead of `all` might be `allUserFacingCommandThatShouldBeDocumentedOnTheWebsite` 😅

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -39,9 +39,16 @@ object ClientCommands {
        |""".stripMargin
   )
 
+  val RunCode = Command(
+    "metals-scala-run",
+    "run",
```
olafurpg: It would be good to document the expected arguments here.

metals/src/main/scala/scala/meta/internal/metals/PostCompileCache.scala
```diff
@@ -0,0 +1,60 @@
+package scala.meta.internal.metals
+
+import java.util.Collections.singletonList
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicBoolean
+import ch.epfl.scala.{bsp4j => b}
+import scala.meta.internal.metals.MetalsEnrichments._
+
+final class PostCompileCache(buildServer: () => Option[BuildServerConnection]) {
```
olafurpg: Would `MainClasses` be an appropriate name for this index?

metals/src/main/scala/scala/meta/internal/metals/PostCompileCache.scala
```diff
@@ -0,0 +1,60 @@
...
+  val mainClasses = new ConcurrentHashMap[String, b.ScalaMainClass]()
+
+  def isInitialized: Boolean = initialized.get()
+
+  def initialize(target: b.BuildTargetIdentifier): CompletableFuture[Void] =
+    initialize(singletonList(target))
+
+  def initialize(
+      targets: java.util.List[b.BuildTargetIdentifier]
+  ): CompletableFuture[Void] = {
```
olafurpg: We normally use `scala.concurrent.Future[Unit]` for our internal APIs and convert `.asJava` at the end before returning to lsp4j

tests/unit/src/main/scala/tests/CodeLensesTextEdits.scala
```diff
@@ -0,0 +1,23 @@
+package tests
+
+import java.util
+import org.eclipse.lsp4j.TextEdit
+import org.eclipse.{lsp4j => l}
+import scala.collection.JavaConverters._
+
+object CodeLensesTextEdits {
+  def apply(ranges: util.List[l.CodeLens]): List[l.TextEdit] = {
```
olafurpg: Let's use `Seq[l.CodeLens]` and force the call-site to use `.asScala` if necessary. By convention, I try to use Scala collections/futures in our internal APIs

metals/src/main/scala/scala/meta/internal/metals/RunCodeLensProvider.scala
```diff
@@ -0,0 +1,47 @@
+package scala.meta.internal.metals
+import java.util
+import java.util.Collections._
+import org.eclipse.{lsp4j => l}
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.TokenEditDistance.fromBuffer
+import scala.meta.internal.mtags.Semanticdbs
+import scala.meta.io.AbsolutePath
+
+final class RunCodeLensProvider(
```
olafurpg: s/RunCodeLensProvider/CodeLensProvider/

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -508,4 +508,8 @@ object MetalsEnrichments
       trailingTokens.find(predicate)
   }
 
+  implicit class XtensionMetalsCommand(command: Command) {
```
olafurpg: This method can be defined on `Command` directly, we mostly use extension methods for 3rd party types

olafurpg: It could be that we have a few `toLSP` extension methods on mtags types from the time mtags didn't depend on lsp4j.

metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
```diff
@@ -27,7 +27,8 @@ final class ForwardingMetalsBuildClient(
     config: MetalsServerConfig,
     statusBar: StatusBar,
     time: Time,
-    didCompile: CompileReport => Unit
+    didCompile: CompileReport => Unit,
+    cache: PostCompileCache
```
olafurpg: `mainClasses: MainClasses`

metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
```diff
@@ -131,6 +132,9 @@ final class ForwardingMetalsBuildClient(
           if (!compilation.isNoOp) {
             scribe.info(s"time: compiled $name in ${compilation.timer}")
           }
+          if (isSuccess && (!cache.isInitialized || !compilation.isNoOp)) {
+            cache.initialize(target)
```
olafurpg: `mainClasses.onDidChange(target)`

tests/unit/src/test/scala/tests/CodeLensesSlowSuite.scala
```diff
@@ -0,0 +1,41 @@
...
+           |}
+           |
+           |/a/src/main/scala/Main.scala
+           |object Main {
+           |  def main(args: Array[String]): Unit = {}
+           |}""".stripMargin
+      )
+      _ <- verifyCodeLenses(
+        "a/src/main/scala/Main.scala",
+        """<<metals-scala-run>>
```
olafurpg: Nice!

metals/src/main/scala/scala/meta/internal/metals/PostCompileCache.scala
marek1840: the test classes will be collected here as well, so it would have to be renamed to something more general anyway.
Could also be something like `JvmEntryPoints` or something like that - what do you think?

metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
marek1840: `entryPoints: JvmEntryPoints` ?

tests/unit/src/test/scala/tests/CodeLensesSlowSuite.scala
marek1840: but not ideal :(
it could also cover the UI name and command arguments like in:
`<<name shown on the UI | command(arg1, arg2, arg3)>>`
e.g. `<<run|metals-scala-run(file://foo/Main.scala,Main)>>`

but it would be much more cluttered (not to mention, that it would be ugly when expecting a `uri` argument.

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
marek1840: I will add a comment here :)

on the other hand, users are not expected to trigger this command but must implement it to be able to run/debug code. Where can that be expresses?

olafurpg: The contract of code lens is that the client executes this command when the lens is selected so we shouldn't need to document this functionality.

metals/src/main/scala/scala/meta/internal/metals/PostCompileCache.scala
olafurpg: `BuildTargetClasses`? It would represent the main/test classes given a build target

tests/unit/src/test/scala/tests/CodeLensesSlowSuite.scala
olafurpg: URIs can be pretty-printed to relative files. I think it might be best to to display the text that appears in the editor and ignore the command ID and arguments (since they're implementation details). We care most about how the code lens appears in the UI, we should test the functionality works as expected separately.

tests/unit/src/test/scala/tests/CodeLensesSlowSuite.scala
```diff
@@ -0,0 +1,41 @@
...
+           |{
+           |  "a": { }
+           |}
+           |
+           |/a/src/main/scala/Main.scala
+           |object Main {
+           |  def main(args: Array[String]): Unit = {}
+           |}""".stripMargin
+      )
+      _ <- verifyCodeLenses(
```
olafurpg: by convention, we use "assert" as a prefix for functions that compare two strings: `assertCodeLens`. This would be consistent with for example `assertCompletion`

metals/src/main/scala/scala/meta/internal/metals/PostCompileCache.scala
marek1840: given that name I would expect it to contain any kind of class from a build target. Maybe `BuildTargetEntryPoints` ?

@tgodzik what do you think? This class will store every runnable main/test class per build target.

metals/src/main/scala/scala/meta/internal/metals/PostCompileCache.scala
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Collections.singletonList
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicBoolean
+import ch.epfl.scala.{bsp4j => b}
+import scala.meta.internal.metals.MetalsEnrichments._
+
+final class PostCompileCache(buildServer: () => Option[BuildServerConnection]) {
+  private val initialized = new AtomicBoolean(false)
+  val mainClasses = new ConcurrentHashMap[String, b.ScalaMainClass]()
```
marek1840: make it per buildTarget
e.g. 
```scala
private class EntryPoints{
  val mainClasses = new ConcurrentHashMap[String, b.ScalaMainClass]()
}

val entryPoints = new ConcurrentHashMap[BuildTargetIdentifier, EntryPoints]()
```

metals/src/main/scala/scala/meta/internal/metals/PostCompileCache.scala
olafurpg: There would be methods `BuildTargetClasses.{mainClasses,testClasses}` to clarify what kinds of classes the index contains.

metals/src/main/scala/scala/meta/internal/metals/PostCompileCache.scala
olafurpg: What are the keys in this map? We BTW normally use `TrieMap` by default for indexes and only `ConcurrentHashMap` if there's a specific need for it.

metals/src/main/scala/scala/meta/internal/metals/PostCompileCache.scala
marek1840: consider me convinced ;)

metals/src/main/scala/scala/meta/internal/metals/PostCompileCache.scala
marek1840: there are three threads using those structures:
1. one actually fetching main classes
2. one listening for bsp messages
3. one asking for main classes to create code lenses

metals/src/main/scala/scala/meta/internal/metals/PostCompileCache.scala
```diff
@@ -0,0 +1,60 @@
...
+      case Some(connection) =>
+        val parameters = new b.ScalaMainClassesParams(targets)
+        val task = connection
+          .mainClasses(parameters)
+          .thenAccept(initializeMainClasses)
+
+        initialized.set(true)
+        task
+      case None =>
+        CompletableFuture.completedFuture(null)
```
tgodzik: Can't we just do `CompletableFuture.completedFuture(())` instead of null?

marek1840: it was `Void`, so nope :(
but anyway, it is fixed now - changed to `Unit` ;)

marek1840:
Closed, since it will be merged alongside the #721 anyway.

PR #721, opened 2019-05-16
Support running main methods

### LINKED ISSUE #652: Add "run" button above main functions
Currently, there is no way to run the code from within the text editor.
Our goal is to introduce a `run` button above the main method, which would run the program (possibly with arguments) and show its output to the console.

Challenges:
a) How to cancel running program?
It looks like the editor does not have any knowledge about the running process, so the cancelation must be handled by the server.
I propose introducing a server command `stop ${task-description}` which would force the server to cancel a task. It could be achieved by notifying the client about running tasks (e.g. registering additional commands, like `stop running program from /path/foo.scala`, `stop running program from /path/bar.scala`)

b) By default, we cannot access the standard input of our program (it is piped to the `Output` tab, which -as the name suggests - does not allow any input). We can either follow the path of `vscode-code-runner` and run the program from the terminal (server just sends the commands, which must then be executed) or find some other way.

marek1840:
Previously, it was impossible to start a main method from the IDE.

Now:
1. Metals provide code lenses with `run` action
2. Metals handles running main methods using the debug adapter protocol

run

closes #652


marek1840:
vscode support is in: https://github.com/scalameta/metals-vscode/pull/106

metals/src/main/scala/scala/meta/internal/debug/DebugSession.scala
```diff
@@ -0,0 +1,63 @@
...
+final class DebugSession(
+    process: Process,
+    val exitCode: Future[Int]
+) {
+  def whenTerminated[A](f: Int => A)(implicit ec: ExecutionContext): Future[A] =
+    exitCode.map(f)
+}
+
+final class DebugSessionFactory(client: IDebugProtocolClient) {
+  def createCommand(params: LaunchParameters): Array[String] = {
```
tgodzik: Does this need to be public? Maybe let's do it consistent like `createWorkingDir`

metals/src/main/scala/scala/meta/internal/debug/DebugSession.scala
```diff
@@ -0,0 +1,63 @@
...
+    val workingDir = createWorkingDir(params)
+    val listener = new Listener()
+
+    val process = Process(command, workingDir).run(listener)
+
+    val exitCode = Future(process.exitValue())
+    new DebugSession(process, exitCode)
+  }
+
+  private def createWorkingDir(params: LaunchParameters): File = {
```
tgodzik: Could we avoid all the `null`? - do option and then `orNull` if really needed

What happens if the directory is null anyway?

metals/src/main/scala/scala/meta/internal/debug/JvmDebugAdapter.scala
```diff
@@ -0,0 +1,53 @@
...
+    debug.Capabilities
+  ] = {
+    val capabilities = new debug.Capabilities()
+    CompletableFuture.completedFuture(capabilities)
+  }
+
+  @JsonRequest
+  def launch(
+      params: LaunchParameters
+  ): CompletableFuture[Void] = {
```
tgodzik: Unit?

metals/src/main/scala/scala/meta/internal/debug/JvmDebugAdapter.scala
```diff
@@ -0,0 +1,53 @@
...
+    val session = sessionFactory.create(params)
+
+    session.whenTerminated { exitCode =>
+      client.terminated(new debug.TerminatedEventArguments)
+      client.exited(ExitedEventArguments(exitCode))
+    }
+
+    CompletableFuture.completedFuture(null)
```
tgodzik: `()` instead maybe?

metals/src/main/scala/scala/meta/internal/debug/JvmDebugServer.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.eclipse.lsp4j.debug.services.IDebugProtocolClient
+import org.eclipse.lsp4j.jsonrpc.Launcher
+import org.eclipse.lsp4j.jsonrpc.debug.DebugLauncher
+import scala.concurrent.ExecutionContextExecutorService
+import scala.concurrent.Future
+
+object JvmDebugServer {
+  def launch(
+      adapter: DebugAdapter
+  )(implicit ec: ExecutionContextExecutorService): java.lang.Integer = {
```
tgodzik: `java.lang.Integer` -> `Int` ?

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -39,6 +39,15 @@ object ClientCommands {
        |""".stripMargin
   )
 
+  val RunCode = Command(
+    "metals-scala-run",
+    "run",
```
tgodzik: Maybe something nicer? Could we do a unicode character if unicode is enabled? something like `➡ Run main` 

And `runs code` seems a little lackluster

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -508,4 +513,12 @@ object MetalsEnrichments
       trailingTokens.find(predicate)
   }
 
+  final implicit class XtensionScalaOption[A](option: Option[A]) {
+    def getOrThrow(e: => Exception): A = {
```
tgodzik: IS that used anywhere?

tests/unit/src/main/scala/tests/debug/Sources.scala
```diff
@@ -0,0 +1,27 @@
...
+      |import java.nio.file.Paths
+      |
+      |object PrintHelloWorld {
+      |  def main(args: Array[String]): Unit = {
+      |    println("Hello, World!")
+      |  }
+      |}"""
+  )
+
+  val CreateFileInWorkingDir: MainClass =
```
tgodzik: Can't we use the same thing like in SbtSlowSuite? `FileLayout` I think it is

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
tgodzik: or `▶ run`

metals/src/main/scala/scala/meta/internal/debug/DebugSession.scala
marek1840: good catch! I think we should create a temporary directory if one is not present. otherwise it would create junk in metals working directory anytime someone does not specify `cwd`

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
marek1840: > And runs code seems a little lackluster
yup, would "Runs main method" be better? I do not like unnecessary words, as you can see ;P

> Could we do a unicode character (...) ?
Up to a debate (@olafurpg ?) I do not have an opinion here

kubukoz: I'd go with `Run`, using unicode seems like cheating and might not play well if Code adds an icon itself, I guess (I don't know much about the protocol so idk if that sort of change could happen, but it might be worth considering :) )

olafurpg: I agree with @kubukoz, "Run" is good.

tests/unit/src/main/scala/tests/debug/MetalsBaseDebugSuite.scala
```diff
@@ -0,0 +1,92 @@
...
+    client = new DebugProtocolClient
+    server = DebugProtocolServer(
+      workspace,
+      metalsServer.compilationScheduler,
+      metalsServer.buildTargets
+    )
+    server.setClient(client)
+  }
+
+  protected final def testDebug(name: String)(
```
olafurpg: Although it seems appealing to abstract over the for comprehension boilerplate in `testAsync`, I am concerned that the current implementation of `testDebug` won't be able to cover more advanced user workflows like restart/cancel. I think it would be better if debug tests were  regular slow tests except with additional `debugServer` and `debugClient` helpers in scope.

In the completion test suite, the `check` method has a lot of parameters with default values like `filter: String => Boolean = _ => true`. However, the difference between completions `check` and `testDebug` is that
-  completions are a single synchronous request->response method
- run/debug can be multiple steps with bi-directional asynchronous communication between the client and server

metals/src/main/scala/scala/meta/internal/debug/DebugSession.scala
```diff
@@ -0,0 +1,68 @@
...
+      .resolve("java")
+      .toString
+
+    val classpath = params.classpath.mkString(File.pathSeparator)
+    Array(java, "-cp", classpath, params.mainClass)
+  }
+
+  private def createWorkingDir(params: LaunchParameters): File = {
+    Option(params.cwd)
+      .map(URI.create)
```
olafurpg: nit: for consistency with the rest of the codebase, try to use pattern matching on `Option` over long chains of `map` and `filter`

```scala
Option(params.cwd).map(_.toAbsolutePath) match {
  case Some(cwd) if cwd.isDirectory => ...
  case _ => ...
}
```

metals/src/main/scala/scala/meta/internal/debug/protocol/OutputEventArguments.scala
```diff
@@ -0,0 +1,22 @@
+package scala.meta.internal.debug.protocol
+
+import org.eclipse.lsp4j.debug
+import org.eclipse.lsp4j.debug.OutputEventArgumentsCategory._
+
+object OutputEventArguments {
```
olafurpg: Can we remove this class altogether and inline the `stdout` and `stderr` methods? They are only used once and the name `OutputEventArguments` conflicts with the lsp4j name.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugParametersAdapter.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import ch.epfl.scala.{bsp4j => b}
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.meta.internal.debug.{protocol => jvm}
+import scala.meta.internal.metals.BuildTargets
+import scala.meta.internal.metals.CompilationScheduler
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.debug.protocol.LaunchParameters
+import scala.meta.internal.metals.debug.{protocol => metals}
```
olafurpg: Let's avoid renaming packages into `metals`. By convention in this codebase, we there are two namespaces with conflicting names like `scala.meta.Position` and `org.eclipse.lsp4j.Position` then we use single character renames

```scala
import scala.{meta => m}
// m.Position
// l.Position
```

metals/src/main/scala/scala/meta/internal/debug/protocol/LaunchParameters.scala
```diff
@@ -0,0 +1,7 @@
+package scala.meta.internal.debug.protocol
+
+final case class LaunchParameters(
```
olafurpg: Can we please remove the nested `debug.protocol` package? By convention in this codebase, we use a flat structure of packages where each package has a functional role

```
package scala.meta.internal.metals // application code for language server
package scala.meta.internal.mtags  // library code for Scala and Java source indexing
package scala.meta.internal.pc     // library code for using presentation compiler
package scala.meta.internal.builds // library code for dealing with sbt/Gradle/Bloop builds
package scala.meta.internal.debug  // library code for dealing with Debug Adapter Protocol
```

When creating a new class, it should either have a unique name in an existing package or go into a new toplevel `internal.$NAME` package. The reason I dislike deeper nesting within those packages is that it makes it easy to end up with conflicting names that are difficult to distinguish, for example `LaunchParameters` in this PR

Screenshot 2019-05-20 at 13 12 45

I admit, that the `internal.metals` package has too much code at the moment and would benefit from breaking it into smaller packages. I think @tgodzik already did a great job of extracting the build related code into `internal.builds` and I think we could do the same for other related functionality like indexing/search.

metals/src/main/scala/scala/meta/internal/metals/CompilationScheduler.scala
```diff
@@ -0,0 +1,108 @@
...
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.io.AbsolutePath
+
+final class CompilationScheduler(
+    buildTargets: BuildTargets,
+    classes: BuildTargetClasses,
+    workspace: () => AbsolutePath,
+    buildServer: () => Option[BuildServerConnection]
+)(implicit ec: ExecutionContext) {
+  private val batch = new BatchedFunction[Request, b.CompileResult](compile)
```
olafurpg: As discussed offline, there should be two separate compile queues

- regular compiles for `didSave`
- cascade compiles for miscellaneous use-cases (import build, find references, manual invocation)

metals/src/main/scala/scala/meta/internal/debug/DebugSession.scala
```diff
@@ -0,0 +1,68 @@
...
+
+final class DebugSessionFactory(client: IDebugProtocolClient) {
+  def create(
+      params: LaunchParameters
+  )(implicit ec: ExecutionContext): DebugSession = {
+    val command = createCommand(params)
+    val workingDir = createWorkingDir(params)
+    val listener = new Listener()
+
+    val process = Process(command, workingDir).run(listener)
```
olafurpg: What do you think about using `NuProcess` here? I've found `java.lang.ProcessBuilder` and `scala.sys.Process` to be inadequate for a many reasons.

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -0,0 +1,82 @@
...
+import ch.epfl.scala.bsp4j.ScalaMainClass
+import ch.epfl.scala.{bsp4j => b}
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.meta.internal.metals.MetalsEnrichments._
+
+final class BuildTargetClasses(
+    buildServer: () => Option[BuildServerConnection]
+)(implicit val ec: ExecutionContext) {
+  private val buildTargets =
```
olafurpg: `val index = TrieMap.empty[b.BuildTargetIdentifier, Classes]`

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -0,0 +1,82 @@
...
+    for {
+      item <- result.getItems.asScala
+      target = item.getTarget
+      aClass <- item.getClasses.asScala
+      objectSymbol = createObjectSymbol(aClass.getClassName)
+    } classesOf(target).main.put(objectSymbol, aClass)
+  }
+
+  final class Classes {
+    val initialized = new AtomicReference[Boolean](false)
```
olafurpg: `isInitialized`

metals/src/main/scala/scala/meta/internal/debug/DebugSession.scala
```diff
@@ -0,0 +1,68 @@
...
+import java.nio.file.Paths
+import org.eclipse.lsp4j.debug.services.IDebugProtocolClient
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.meta.internal.debug.protocol.LaunchParameters
+import scala.meta.internal.debug.protocol.OutputEventArguments._
+import scala.sys.process.Process
+import scala.sys.process.ProcessLogger
+
+final class DebugSession(
```
olafurpg: Can we replace this class with `Future[Int]`?

metals/src/main/scala/scala/meta/internal/debug/DebugSession.scala
```diff
@@ -0,0 +1,68 @@
...
+    val listener = new Listener()
+
+    val process = Process(command, workingDir).run(listener)
+
+    val exitCode = Future(process.exitValue())
+    new DebugSession(process, exitCode)
+  }
+
+  private def createCommand(params: LaunchParameters): Array[String] = {
+    val java = Paths
```
olafurpg: Let's reuse `scala.meta.internal.metals.JavaBinary(None)` and optionally pass around `javaHome: Option[AbsolutePath]` from `MetalsLanguageServer` to remove `None`.

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -0,0 +1,82 @@
...
+        Future.successful(())
+    }
+  }
+
+  private def classesOf(target: b.BuildTargetIdentifier): Classes =
+    buildTargets.computeIfAbsent(target, _ => new Classes)
+
+  private def initializeMainClasses(result: b.ScalaMainClassesResult): Unit = {
+    def createObjectSymbol(className: String): String = {
+      val isRootPackage = !className.contains(".")
```
olafurpg: > `isInsideEmptyPackage`

There's a difference between the root package and empty package

- `_empty_` is the enclosing package for classes that are not defined inside a package
- `_root_` is the owner of all packages (including `_empty_`).

To work with SemanticDB symbols we normally use the following helpers
```scala
val sym = "Toplevel."
Symbols.Global(EmptyPackage, sym) // "_empty_/Toplevel."
```

metals/src/main/scala/scala/meta/internal/metals/debug/MetalsDebugAdapter.scala
```diff
@@ -0,0 +1,54 @@
...
+import org.eclipse.lsp4j.debug.InitializeRequestArguments
+import org.eclipse.lsp4j.debug.services.IDebugProtocolClient
+import org.eclipse.lsp4j.jsonrpc.services.JsonRequest
+import scala.concurrent.ExecutionContext
+import scala.meta.internal.debug.DebugAdapter
+import scala.meta.internal.debug.JvmDebugAdapter
+import scala.meta.internal.metals.BuildTargets
+import scala.meta.internal.metals.CompilationScheduler
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.debug.{protocol => metals}
```
olafurpg: ditto

metals/src/main/scala/scala/meta/internal/debug/ClientDelegate.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.eclipse.lsp4j.debug.LoadedSourceEventArguments
+import org.eclipse.lsp4j.debug.ModuleEventArguments
+import org.eclipse.lsp4j.debug.OutputEventArguments
+import org.eclipse.lsp4j.debug.ProcessEventArguments
+import org.eclipse.lsp4j.debug.StoppedEventArguments
+import org.eclipse.lsp4j.debug.TerminatedEventArguments
+import org.eclipse.lsp4j.debug.ThreadEventArguments
+import org.eclipse.lsp4j.debug.services.IDebugProtocolClient
+
+private[debug] final class ClientDelegate extends IDebugProtocolClient {
```
olafurpg: > `DelegatingDebugClient`

For consistency with `DelegatingLanguageClient`

metals/src/main/scala/scala/meta/internal/metals/debug/MetalsDebugAdapter.scala
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Metals receive requests without the full knowledge - e.g. without classpath.
```
olafurpg: When writing docstrings for classes like this, try to explain the motivation for the existence of this class instead of implementation specifics.

> MetalsDebugAdapter is similar to MetalsLanguageServer except it implements the Debug Adapter Protocol instead of Language Server Protocol.

metals/src/main/scala/scala/meta/internal/debug/DebugAdapter.scala
```diff
@@ -0,0 +1,6 @@
+package scala.meta.internal.debug
+import org.eclipse.lsp4j.debug.services.IDebugProtocolClient
+
+trait DebugAdapter {
```
olafurpg: Is this trait necessary? It seems we replace it with `MetalsDebugAdapter` whereever `DebugAdapter` is used as a parameter type.

metals/src/main/scala/scala/meta/internal/debug/JvmDebugAdapter.scala
```diff
@@ -0,0 +1,53 @@
...
+import scala.concurrent.ExecutionContext
+import scala.meta.internal.debug.protocol.ExitedEventArguments
+import scala.meta.internal.debug.protocol.LaunchParameters
+
+final class JvmDebugAdapter(implicit val ex: ExecutionContext)
+    extends DebugAdapter {
+  private val client = new ClientDelegate
+  private val sessionFactory = new DebugSessionFactory(client)
+
+  @JsonRequest
```
olafurpg: Are these annotations necessary on `JvmDebugAdapter`? I think it's only needed for `MetalsDebugServer`.

metals/src/main/scala/scala/meta/internal/metals/debug/MetalsDebugAdapter.scala
```diff
@@ -0,0 +1,54 @@
...
+ * Here, we provide such information
+ */
+class MetalsDebugAdapter(
+    underlying: JvmDebugAdapter,
+    parametersAdapter: DebugParametersAdapter
+)(implicit ec: ExecutionContext)
+    extends DebugAdapter {
+
+  @JsonRequest
```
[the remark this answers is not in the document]
olafurpg: I don't feel strongly about this, I just realized this doesn't seem to be necessary for DAP since all the requests are simple identifiers. For LSP the endpoints are named `textDocument/definition` so that's why we pass in the string name in `MetalsLanguageServer`. Feel free to do it only if you think it helps with consistency for our usage of `@JsonRequest`

tests/slow/src/test/scala/tests/debug/MetalsDebugSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+    act = server => {
+      server.launch("a/src/main/scala/PrintHelloWorld.scala")
+    },
+    assert = client => {
+      assertEquals(client.output.stderr, "")
+      assertEquals(client.output.stdout, "Hello, World!")
+    }
+  )
+
+  testDebug("working-dir")(
```
olafurpg: Can we add a test case for the situation when the project has a compile error? It should be clear to the user that running doesn't work because there's a compile error. Currently, an internal exception is thrown

Screenshot 2019-05-20 at 17 19 35

olafurpg: I think we could ask the user what to do via `window/showMessageRequest`

> This project can't be run because it has compile errors: Run anyways | Show errors | Cancel

Selecting `Run anyways` can use the old classpath and `Show errors` can trigger the `"metals-diagnostics-focus"` client command.

metals/src/main/scala/scala/meta/internal/debug/DebugSession.scala
marek1840: Oh, I did not notice it was in scope. But for now let's leave it as it is (even though `NuProcess` is much better to use) - I will revisit this idea if my approach of using a build server to run processes fails, because if it succeeds then there will be no need for using processes directly anyway

olafurpg: I think it would be good to try to reach the final design (BSP run vs. NuProcess) in this PR before it's merged.

olafurpg: Did you consider using the `runInTerminal` request in DAP? https://microsoft.github.io/debug-adapter-protocol/specification#Reverse_Requests_RunInTerminal

metals/src/main/scala/scala/meta/internal/debug/JvmDebugAdapter.scala
marek1840: If we want to maintain clear separation of concerns then yes, it is needed.
Otherwise `internal.eval` would have to be aware of the internals of`internal.metals`

metals/src/main/scala/scala/meta/internal/debug/DebugAdapter.scala
marek1840: If we want to maintain clear separation of concerns then yes, it is needed.
Otherwise `internal.eval` would have to be aware of the content of `nternal.metals`

metals/src/main/scala/scala/meta/internal/debug/DebugSession.scala
marek1840: > I think it would be good to try to reach the final design (BSP run vs. NuProcess) in this PR before it's merged.

We would depend on at least `BSP` v. 2.0.0-M4. `bloop` is using 2.0.0-M1, therefore either we postpone this feature or use `NuProcess`

marek1840: Changed implementation to use a `NuProcess`

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
marek1840: I will use `isInsideDefaultPackage`. When I hear "it is inside an empty package" then I think: How can it be an empty package if it has something inside.
Or maybe `isInsideUnnamedPackage`? :)

olafurpg: The “empty package” is defined in the Scala language spec https://www.scala-lang.org/files/archive/spec/2.12/09-top-level-definitions.html#packagings

In the Java language spec the same package is called “unnamed”.

We are dealing with SemanticDB symbols here where we use the “empty package” for both languages https://scalameta.org/docs/semanticdb/specification.html#symbol-1

olafurpg:
Should we close this PR? As discussed offline, we're gonna try to move the debug adapter server into Bloop so that we faithfully reproduce the build execution for run/test, avoiding the need to roll our own `Process("java ...")` logic.

marek1840:
Closing, since most of the logic here will be moved over to `bloop`. We will revisit the issue once that gets done once that gets done

PR #722, opened 2019-05-16
Adds maven integration to metals

tgodzik:
Previously metals didn't allow to import maven projects automatically and now it will automatically ask user to do so.

Adds maven integration to metals, currently doesn't add the semanticDB plugin due to limitations in maven bloop plugin. Further work will be done on bloop side.


metals/src/main/scala/scala/meta/internal/builds/MavenBuildTool.scala
```diff
@@ -0,0 +1,63 @@
...
+      workspace: AbsolutePath,
+      userConfig: () => UserConfiguration,
+      config: MetalsServerConfig
+  ): List[String] = {
+    // TODO this should have stable maven version
+    // import scala.meta.internal.metals.BuildInfo
+    // val command = s"ch.epfl.scala:maven-bloop_2.10:${BuildInfo.bloopVersion}:bloopInstall"
+    val command =
+      List(
+        s"ch.epfl.scala:maven-bloop_2.10:1.2.5+418-110de48b:bloopInstall",
```
tgodzik: We need new bloop release to make it work.

metals/src/main/resources/db/migration/V1__Create_tables.sql
```diff
@@ -8,7 +8,7 @@ create table dependency_source(
 
 -- The relationship between what library dependency sources under .metals/readonly/**
 -- map to which build targets.
-create table sbt_digest(
+create table digest(
```
marek1840: Shouldn't there be a new script altering table name?

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -239,3 +263,19 @@ object GradleDigest extends Digestable {
...
   }
 }
+
+object MavenDigest extends Digestable {
+  override protected def digestWorkspace(
+      workspace: AbsolutePath,
+      digest: MessageDigest
+  ): Boolean = {
+    Digest.digestFile(workspace.resolve("pom.xml"), digest)
+    Files.walk(workspace.toNIO).iterator().asScala.forall { file =>
```
marek1840: Java stream should have something like 'forall' defined

tests/slow/src/test/scala/tests/MavenSlowSuite.scala
```diff
@@ -0,0 +1,201 @@
...
+import java.nio.file.Paths
+import scala.collection.JavaConverters._
+import scala.meta.internal.metals.ServerCommands
+
+object MavenSlowSuite extends BaseImportSuite("maven-import") {
+
+  val defaultPom = Files
+    .readAllLines(
+      Paths.get(
+        this.getClass().getClassLoader().getResource("test-pom.xml").toURI()
```
marek1840: Extract to variable, so it is more readable

metals/src/main/scala/scala/meta/internal/builds/MavenBuildTool.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  def digest(workspace: AbsolutePath): Option[String] =
+    MavenDigest.current(workspace)
+
+  def minimumVersion: String = "3.5.2"
+
+  def version: String = "3.6.1"
+
+  def isMavenRelatedPath(workspace: AbsolutePath, path: AbsolutePath) = {
+    path.toNIO.startsWith(workspace.toNIO) && path.toNIO
```
marek1840: path.filename (from Xtension)

metals/src/main/resources/db/migration/V1__Create_tables.sql
tgodzik: I don't think we have a way to do it, but with a new name it shouldn't cause any issues I think. @olafurpg ?

marek1840: I do not know what mechanism is used underneath, but migration scripts in my experience should almost never change. It may cause issues for clients who have already have them applied (different script checksum etc.)

Anyway, let me try to understand how this part actually works

marek1840: Yup, I believe you should create a file `migration/V4_rename_digest.sql` with:
```sql
alter table sbt_digest rename to digest;
```
(for reference https://www.h2database.com/html/commands.html#alter_table_rename)

tgodzik: Thanks, I will do the rename in a separate commit - it's not strictly related to maven

tests/slow/src/test/resources/test-pom.xml
```diff
@@ -0,0 +1,57 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.virtuslab</groupId>
+  <artifactId>maven-test-repo</artifactId>
+  <version>1.0-SNAPSHOT</version>
+  <name>${project.artifactId}</name>
+  <description>My wonderfull scala app</description>
```
olafurpg: wonderful

metals/src/main/scala/scala/meta/internal/builds/MavenBuildTool.scala
olafurpg: What is the workaround for people to try out this PR?

metals/src/main/scala/scala/meta/internal/builds/MavenBuildTool.scala
```diff
@@ -0,0 +1,63 @@
+package scala.meta.internal.builds
+
+import scala.meta.internal.metals.{MetalsServerConfig, UserConfiguration}
+import scala.meta.io.AbsolutePath
+import scala.tools.nsc.Properties
+import scala.meta.internal.metals.MetalsEnrichments._
+
+object MavenBuildTool extends BuildTool {
+
+  private lazy val embeddedMavenLauncher: AbsolutePath = {
```
olafurpg: What happens if we turn `MavenBuildTool` into a `class` instead of `object`? Same for `GradleBuildTool`. I'm a concerned about doing I/O on lazy vals for global objects.

tests/slow/src/test/scala/tests/MavenSlowSuite.scala
```diff
@@ -0,0 +1,200 @@
...
+import java.nio.file.Paths
+import scala.collection.JavaConverters._
+import scala.meta.internal.metals.ServerCommands
+
+object MavenSlowSuite extends BaseImportSuite("maven-import") {
+
+  val defaultPomPath = Paths.get(
+    this.getClass().getClassLoader().getResource("test-pom.xml").toURI()
+  )
+  val defaultPom = Files
```
olafurpg: FYI, we have a `AbsolutePath.readText` helper to read paths into strings. In this particular case, the idiomatic way to read the file from resources is

https://github.com/scalameta/metals/blob/f28c35ac855ed58bd5bf30fcea70e8497c49e697/metals-bench/src/main/scala/bench/SourceCompletion.scala#L44-L50

No need for `getResource("...").toURI().toPath.readText`

tests/slow/src/test/scala/tests/MavenSlowSuite.scala
```diff
@@ -0,0 +1,200 @@
...
+            |$defaultPom
+            |""".stripMargin
+      )
+      _ = assertNoDiff(
+        client.workspaceMessageRequests,
+        List(
+          ImportBuild.params("maven").getMessage,
+          bloopInstallProgress("maven").message,
+          // we don't have semanticDB plugin yet
+          CheckDoctor.allProjectsMisconfigured
```
olafurpg: 👍 Hopefully this will be automatically handled by Bloop in the future

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -90,16 +90,40 @@ object Digest {
...
       true
     }
   }
 
+  def digestXml(
+      file: AbsolutePath,
+      digest: MessageDigest
+  ): Boolean = {
+    try {
+      Files
```
olafurpg: FYI, not sure we want to do it, but one option is to parse the file with `scala.xml.XML.loadFile` and digest the parsed XML

metals/src/main/scala/scala/meta/internal/builds/MavenBuildTool.scala
tgodzik: Publish newest bloop locally and then replace this version with that bloop version.

tgodzik: Alternatively you can use 1.2.5 version, but it will not have sources.

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
tgodzik: Added, I think I was just a bit lazy previously and forgot about it.

tests/slow/src/test/scala/tests/MavenSlowSuite.scala
tgodzik: Fixed

metals/src/main/scala/scala/meta/internal/builds/MavenBuildTool.scala
tgodzik: Right, turned into classes, that wasn't an issue.

tests/slow/src/test/resources/test-pom.xml
tgodzik: fixed

tgodzik:
This is blocked on release of bloop.

olafurpg:
Is this ready to merge now that Bloop v1.3.0-RC1 is out?

tgodzik:
> Is this ready to merge now that Bloop v1.3.0-RC1 is out?

Should be good, anything else that I might need to fix here?

metals/src/main/scala/scala/meta/internal/builds/MavenBuildTool.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  def digest(workspace: AbsolutePath): Option[String] =
+    MavenDigest.current(workspace)
+
+  def minimumVersion: String = "3.5.2"
+
+  def version: String = "3.6.1"
+
+  override def toString(): String = "maven"
```
olafurpg:
```suggestion
  override def toString(): String = "Maven"
```
?

tgodzik: I think I need to rethink the tool name vs executable name in all of the build tools - I will create a follow up PR after mill is merged

olafurpg: The message "New maven workspace detected" would read nicer with an uppercase `M`. Same for Mill.

PR #727, opened 2019-05-21
Fix newlines in process output.

tgodzik:
Previously we would wrongly log process messages with newlines in wrong places, now we make sure that they are at exactly the same places as in the process output.

This came up both with gradle and maven integrations.


metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
```diff
@@ -243,9 +244,14 @@ object BloopInstall {
         fn: String => Unit
     ): Unit = {
       if (!closed) {
-        val text = toPlainString(buffer).trim
-        if (text.nonEmpty) {
-          fn(text)
+        val text = toPlainString(buffer)
+        output = output + text
+        while (output.contains('\n')) {
```
marek1840: `output.indexOf('\n') > 0` ? no need to follow up `contains` with the `indexOf`. Asa a result we can also drop the inner `if` :)

marek1840: or maybe something like:
```
val lines = output.splitBy(char == '\n') // if no '\n' then lines == Array(output)
lines.dropRight(1).forEach(fn)
output = lines.last  // last entry is everything after the last '\n'
```

metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
```diff
@@ -243,9 +244,14 @@ object BloopInstall {
         fn: String => Unit
     ): Unit = {
       if (!closed) {
-        val text = toPlainString(buffer).trim
-        if (text.nonEmpty) {
-          fn(text)
+        val text = toPlainString(buffer)
+        output = output + text
```
marek1840: isn't it équiva and then Java lent of appending inside a loop? Here, we are not the one controlling how often will that happen, so (even though Java 8 and then Java 9 greatly increased the performance of appending strings) I am not sure just concatenating strings here is the best approach.

metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
```diff
@@ -214,6 +214,7 @@ object BloopInstall {
   private class ProcessHandler() extends NuAbstractProcessHandler {
     var response: Option[CompletableFuture[_]] = None
     val completeProcess = Promise[BloopInstallResult]()
+    var output = ""
```
marek1840: It looks like both of the input channels are using the same "queue"? Is there no contention between them? Does the `onStdout` and `onStderr` happen in the exact same order as in the process? Are they handled by the same thread underneath (i.e. the Process class in scala library there is a thread for each of them so they might trigger those methods in any order they want). If there are two threads handling `stderr` and `stdout` then we either need some means of synchronization on the `log` method or even better two queues. Otherwise those would  probably get intertwined. 
But if everything here is safe, then please add a comment ;)

metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
olafurpg: Is it possible to use `String.linesIterator.foreach(...)` in this situation?

tgodzik: the problems is that `"asbcd\n".linesIterator.size == 1` and I need to differentiate between "asbcd\n" and "asbcd" situation. I think this while is a more straightforward solution

metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
tgodzik: Using StringBuilder after sugestion

metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
tgodzik: Separated the different outputs - thanks! Haven't noticed it.

metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
tgodzik: also if I do it this way I will always have everything written properly till the end since build tools always end with \n

olafurpg: Here's an implementation that doesn't use while, vars or string concat

```scala
@   def forallLines(text: String, fn: String => Unit): Unit = {
      def loop(start: Int): Unit = {
        val nl = text.indexOf('\n', start)
        if (nl < 0) {
          fn(text.substring(start))
        } else {
          fn(text.substring(start, nl))
          loop(nl + 1) // hop over newline character
        }
      }
      loop(0)
    }
defined function forallLines

@ forallLines("a\nb\n", line => pprint.log(line))
ammonite.$sess.cmd9.res9:1 line: "a"
ammonite.$sess.cmd9.res9:1 line: "b"
ammonite.$sess.cmd9.res9:1 line: ""
```

olafurpg: I'm concerned this doesn't handle `\r\n` however, which `String.linesIterator` does. I would recommend using `linesIterator` instead and if the trailing endline is important then you can add a special case for it
```scala
output.linesIterator.foreach(fn)
if (output.endsWith("\n")) fn("")
```

tgodzik: That var is a buffer between invocations though, that's why it's important. Maven send a single line in 3 parts. I did switch it to StringBuilder which should be better for now. As for lineIterator I will check what will happen with `\r\n`.

olafurpg: Is it necessary to do `fn(line)` on each individual line? What happens if we call `scribe.info` on messages with multiple newlines?

tgodzik: Ok, added using lineIter and I also check if we should include the last line depending if it ends with \n

metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
```diff
@@ -214,6 +214,8 @@ object BloopInstall {
   private class ProcessHandler() extends NuAbstractProcessHandler {
     var response: Option[CompletableFuture[_]] = None
     val completeProcess = Promise[BloopInstallResult]()
+    val standardOutput = new StringBuilder
+    val errorOutput = new StringBuilder
 
     override def onStart(nuProcess: NuProcess): Unit = {
```
olafurpg: Should we flush the buffer on exit?

metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
```diff
@@ -232,20 +234,28 @@ object BloopInstall {
...
+        val text = toPlainString(buffer)
+        output.append(text)
+        val lines = output.linesIterator.toList
+        if (output.endsWith("\n")){
+          lines.foreach(fn)
+          output.clear()
+        } else {
+          lines.take(lines.size - 1)
+          val last = lines.last
+          output.clear()
```
olafurpg: What happens to the non-last lines? It looks like they get removed here 🤔

tgodzik: my bad, forgot foreach - that would come up in maven

metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
tgodzik: Which buffer? We just use the logger and pretty sure it has autoflush set.

olafurpg: What happens in the following scenario?

- stdout prints `"Hello"`, with no newline
- process exits

I expect `"Hello"` to get printed out on exit.

tgodzik: But that is not going to happen, we only run build tools using this class.  

I can account for scenario, but it seems really hypothetical.

tgodzik: Ok, added additional logging on exit.

olafurpg:
@tgodzik I took the liberty to implement a new `ProcessOutput` helper to manage the process output. Please let me know what you think.

metals/src/main/scala/scala/meta/internal/metals/ProcessOutput.scala
```diff
@@ -0,0 +1,74 @@
...
+import fansi.ErrorMode
+
+/**
+ * Handles streaming console output from a system process with potential ANSI codes.
+ *
+ * @param onLine The callback handler when the process has printed a single line.
+ *               Guaranteed to have no newline \n characters.
+ */
+class ProcessOutput(onLine: String => Unit) {
+  private var buffer = new StringBuilder()
```
marek1840: val and then `buffer.clear()`?

metals/src/main/scala/scala/meta/internal/metals/ProcessOutput.scala
```diff
@@ -0,0 +1,74 @@
...
+import java.nio.charset.StandardCharsets
+import fansi.ErrorMode
+
+/**
+ * Handles streaming console output from a system process with potential ANSI codes.
+ *
+ * @param onLine The callback handler when the process has printed a single line.
+ *               Guaranteed to have no newline \n characters.
+ */
+class ProcessOutput(onLine: String => Unit) {
```
marek1840: De facto a `LineListener`. Let's keep this abstraction simple - it does not know anything about the source of those lines, nor should it. Therefore we can drop the `Process` from the name.
Also, this name would play nicely with the name `onLine` callback.

metals/src/main/scala/scala/meta/internal/metals/ProcessOutput.scala
```diff
@@ -0,0 +1,74 @@
...
+  private var buffer = new StringBuilder()
+
+  /** The process has exited, clears out buffered output. */
+  def onProcessExit(): Unit = {
```
marek1840: flush

metals/src/main/scala/scala/meta/internal/metals/ProcessOutput.scala
```diff
@@ -0,0 +1,74 @@
...
+    if (buffer.length() > 0) {
+      onLine(buffer.toString())
+      reset()
+    }
+  }
+
+  def onByteOutput(bytes: ByteBuffer): this.type = {
```
marek1840: `appendBytes`

metals/src/main/scala/scala/meta/internal/metals/ProcessOutput.scala
```diff
@@ -0,0 +1,74 @@
...
+    onPlainOutput(toPlainString(bytes))
+  }
+
+  def onStringOutput(text: String): this.type = {
```
marek1840: `appendString`

tests/unit/src/test/scala/tests/ProcessOutputSuite.scala
```diff
@@ -0,0 +1,57 @@
+package tests
+
+import scala.meta.internal.metals.ProcessOutput
+import scala.collection.mutable
+
+object ProcessOutputSuite extends BaseSuite {
+  def check(
+      name: String,
+      fn: ProcessOutput => Unit,
```
marek1840: `act` is a well established name for those and it also has some meaning ;)

tests/unit/src/test/scala/tests/ProcessOutputSuite.scala
```diff
@@ -0,0 +1,57 @@
...
+      val output = new ProcessOutput(line => buf += line)
+      fn(output)
+      output.onProcessExit()
+      assertEquals(buf.toList, expected)
+    }
+  }
+
+  check(
+    "flush",
+    _.onStringOutput("a")
```
marek1840: IMO chaining makes it look worse than:
```
foo => {
  foo.onStringOutput("a")
  foo.onStringOutput("b")
}
```

metals/src/main/scala/scala/meta/internal/metals/ProcessOutput.scala
olafurpg: I tried finding a `clear()` method on `java.lang.StringBuilder` but couldn't find any.

tests/unit/src/test/scala/tests/ProcessOutputSuite.scala
olafurpg: Done.

metals/src/main/scala/scala/meta/internal/metals/ProcessOutput.scala
```diff
@@ -0,0 +1,74 @@
...
+      } else {
+        onLine(lineSubstring(text, start, newline))
+        loop(newline + 1)
+      }
+    }
+    loop(0)
+    this
+  }
+
+  private def lineSubstring(text: String, from: Int, to: Int): String = {
```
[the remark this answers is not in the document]
tgodzik: Instead we get  two methods and a recursion which is a bit more complex to reason about.

tests/unit/src/test/scala/tests/ProcessOutputSuite.scala
olafurpg: Good point 👍

tgodzik:
Overall, I am not sure if recursion is the way to go here, especially that we need three different methods now and that always makes it harder to reason.

However, as I just intended this as a small fix I don't really care that much with how we achieve this.

metals/src/main/scala/scala/meta/internal/metals/ProcessOutput.scala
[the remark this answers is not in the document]
olafurpg: The recursive implementation may not be as readable but it achieves several performance improvements

- less copying, we perform the minimal number of `substring` operations (one per line)
- fewer redundant operations, the recursive loop iterates over each character at most once

This is fairly performance sensitive code since we're feeding the standard output of a build process (downloading dependencies, ...) to the LSP client.

metals/src/main/scala/scala/meta/internal/metals/ProcessOutput.scala
tgodzik: It's in scala.collection.mutable.StringBuilder

tgodzik:
Thanks for taking a look at it! This looks like a pretty resilient solution. 

We still could go with Scala's StringBuilder and that would allow us to have:
```scala
private val buffer = new StringBuilder()
```
but honestly if you think it's OK now then let's merge it.

tests/unit/src/test/scala/tests/LineListenerSuite.scala
```diff
@@ -0,0 +1,63 @@
+package tests
+
+import scala.meta.internal.metals.LineListener
+import scala.collection.mutable
+
+object LineListenerSuite extends BaseSuite {
```
marek1840: Couldn't hurt to check how it behaves with unicode now, when we operate on bytes.

metals/src/main/scala/scala/meta/internal/metals/LineListener.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  private def onCharacter(char: Char): Unit = {
+    if (char == '\n') {
+      flush()
+    } else {
+      buffer.append(char)
+    }
+  }
+
+  case class State(apply: Int => State)
```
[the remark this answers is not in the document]
olafurpg: I've refactored the state machine into a separate file and given the states more readable names. I personally find the new approach easier to maintain and evolve since we're conceptually doing a simpler task, we filter out a few bytes from the process output and emit lines as soon as we find a `\n`. The previous implementation had multiple layers of conversions that were difficult to follow and extend without adding yet another full iteration on the character output stream.

metals/src/main/scala/scala/meta/internal/metals/LineListener.scala
```diff
@@ -0,0 +1,117 @@
...
+        'L' | 'M' | 'N' | 'O' | 'P' | 'Q' | 'R' | 'S' | 'T' | 'U' | 'V' | 'W' |
+        'X' | 'Y' | 'Z' | '[' | '\\' | ']' | '^' | '_' | '`' | 'a' | 'b' | 'c' |
+        'd' | 'e' | 'f' | 'g' | 'h' | 'i' | 'j' | 'k' | 'l' | 'm' | 'n' | 'o' |
+        'p' | 'q' | 'r' | 's' | 't' | 'u' | 'v' | 'w' | 'x' | 'y' | 'z' | '{' |
+        '|' | '}' | '~' =>
+      T // end of ANSI escape
+    case _ => F
+  }
+
+  private var stack = mutable.ListBuffer.empty[Char]
```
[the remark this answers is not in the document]
olafurpg: The stack and the state variables are now next to the string builder buffer.

metals/src/main/scala/scala/meta/internal/metals/LineListener.scala
```diff
@@ -0,0 +1,121 @@
...
+  val I2: State = State {
+    case '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9' => I2
+    case ';' => I1
+    case '@' | 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'G' | 'H' | 'I' | 'J' | 'K' |
+        'L' | 'M' | 'N' | 'O' | 'P' | 'Q' | 'R' | 'S' | 'T' | 'U' | 'V' | 'W' |
+        'X' | 'Y' | 'Z' | '[' | '\\' | ']' | '^' | '_' | '`' | 'a' | 'b' | 'c' |
+        'd' | 'e' | 'f' | 'g' | 'h' | 'i' | 'j' | 'k' | 'l' | 'm' | 'n' | 'o' |
+        'p' | 'q' | 'r' | 's' | 't' | 'u' | 'v' | 'w' | 'x' | 'y' | 'z' | '{' |
+        '|' | '}' | '~' =>
+      T // end of ANSI escape
```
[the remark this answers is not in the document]
olafurpg: They're handled by `Start`, the new names should make it clearer

```scala
  val Start: AnsiStateMachine = AnsiStateMachine {
    case 27 => Ansi1
    case 13 => CarriageReturn // \r
    case _ => Print
  }
```

metals/src/main/scala/scala/meta/internal/metals/LineListener.scala
```diff
@@ -0,0 +1,121 @@
...
+            onCharacter(i)
+          }
+          onCharacter(ch)
+          resetState()
+        case T =>
+          resetState()
+        case L =>
+          flush()
+          resetState()
+        case s =>
```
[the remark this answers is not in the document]
olafurpg: I've renamed it to `other`

tests/unit/src/test/scala/tests/LineListenerSuite.scala
olafurpg: The state machine operates on characters and does not nothing until it hits the first character with int value 27. I've added a new test case to demonstrate that unicode characters are unaffected.

tests/unit/src/test/scala/tests/LineListenerSuite.scala
```diff
@@ -0,0 +1,70 @@
...
+  check(
+    "windows", { out =>
+      out.appendString("a\r\nb")
+    },
+    List("a", "b")
+  )
+
+  check(
+    "emoji", { out =>
+      out.appendString("🇮🇸🇵🇱")
```
marek1840: +1

metals/src/main/scala/scala/meta/internal/ansi/AnsiStateMachine.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  val CarriageReturn: AnsiStateMachine = AnsiStateMachine {
+    case 10 => LineFeed // drop \r from \r\n
+    case _ => Print
+  }
+  val LineFeed: AnsiStateMachine = AnsiStateMachine(_ => LineFeed)
+  val Print: AnsiStateMachine = AnsiStateMachine(_ => Print)
+  val Discard: AnsiStateMachine = AnsiStateMachine(_ => Discard)
+  val Ansi1: AnsiStateMachine = AnsiStateMachine {
```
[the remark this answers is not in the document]
olafurpg: Good point, I did my best to come up with helpful names here and also added a note on why we have this low-level implementation for processing build logs.

marek1840:
I believe we have spent long enough on this issue, so if the functionality works - if we are all more or less happy then let's just merge it and hope we won't have to spend too much time in the future interacting with this state machine ;)

PR #728, opened 2019-05-23
Implement textDocument/codeLens to run main function

marek1840:
Previously, metals did not provide any code lens for running the main classes.

Now, we have the whole infrastructure for providing those code lenses set up. We will just have to enable them once we can actually run the main classes.

When this is merged, #652 should be updated to reflect the need of enabling this feature

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
@@ -1,10 +1,5 @@
 package scala.meta.internal.metals
 
-import ch.epfl.scala.bsp4j.BuildClientCapabilities
-import ch.epfl.scala.bsp4j.CompileParams
-import ch.epfl.scala.bsp4j.CompileResult
-import ch.epfl.scala.bsp4j.InitializeBuildParams
-import ch.epfl.scala.bsp4j.InitializeBuildResult
 import java.io.InputStream
 import java.io.OutputStream
 import java.util.Collections
@@ -12,13 +7,20 @@ import java.util.concurrent.CompletableFuture
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
 import java.util.concurrent.atomic.AtomicBoolean
+import ch.epfl.scala.bsp4j.BuildClientCapabilities
+import ch.epfl.scala.bsp4j.CompileParams
+import ch.epfl.scala.bsp4j.CompileResult
+import ch.epfl.scala.bsp4j.InitializeBuildParams
+import ch.epfl.scala.bsp4j.InitializeBuildResult
+import ch.epfl.scala.bsp4j.ScalaMainClassesParams
+import ch.epfl.scala.bsp4j.ScalaMainClassesResult
 import org.eclipse.lsp4j.jsonrpc.Launcher
 import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContextExecutorService
 import scala.concurrent.Future
+import scala.meta.internal.pc.InterruptException
 import scala.meta.io.AbsolutePath
 import scala.util.Try
-import scala.meta.internal.pc.InterruptException
 
 /**
  * An actively running and initialized BSP connection.
@@ -67,6 +69,15 @@ case class BuildServerConnection(
 
   def compile(params: CompileParams): CompletableFuture[CompileResult] = {
     register(server.buildTargetCompile(params))
+  }
+
+  def mainClasses(
+      params: ScalaMainClassesParams
+  ): CompletableFuture[ScalaMainClassesResult] = {
+    // TODO use server.buildTargetScalaMainClasses when bloop releases version supporting mainClasses
+    CompletableFuture.completedFuture(
+      new ScalaMainClassesResult(Collections.emptyList())
+    )
   }
 
   private val cancelled = new AtomicBoolean(false)
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
@@ -1,0 +1,75 @@
+package scala.meta.internal.metals
+
+import ch.epfl.scala.{bsp4j => b}
+import scala.collection.concurrent.TrieMap
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.semanticdb.Scala.Descriptor
+import scala.meta.internal.semanticdb.Scala.Symbols
+
+/**
+ * In-memory index of main class symbols grouped by their enclosing build target
+ */
+final class BuildTargetClasses(
+    buildServer: () => Option[BuildServerConnection]
+)(implicit val ec: ExecutionContext) {
+  private val index = new TrieMap[b.BuildTargetIdentifier, Classes]()
+
+  def main(target: b.BuildTargetIdentifier): TrieMap[String, b.ScalaMainClass] =
+    classesOf(target).main
+
+  val onCompiled: BatchedFunction[b.BuildTargetIdentifier, Unit] =
+    BatchedFunction.fromFuture(fetchMainClassesFor)
+
+  private def fetchMainClassesFor(
+      targets: Seq[b.BuildTargetIdentifier]
+  ): Future[Unit] = {
+    targets.foreach { target =>
+      classesOf(target).clear()
+    }
+
+    buildServer() match {
+      case Some(connection) =>
+        val parameters = new b.ScalaMainClassesParams(targets.asJava)
+        val task = for {
+          result <- connection.mainClasses(parameters).asScala
+          _ = cacheMainClasses(result)
+        } yield ()
+
+        task
+      case None =>
+        Future.successful(())
+    }
+  }
+
+  private def classesOf(target: b.BuildTargetIdentifier): Classes =
+    index.getOrElseUpdate(target, new Classes)
+
+  private def cacheMainClasses(result: b.ScalaMainClassesResult): Unit = {
+    def createObjectSymbol(className: String): String = {
+      val symbol = className.replaceAll("\\.", "/")
+      val isInsideEmptyPackage = !className.contains(".")
+      if (isInsideEmptyPackage) {
+        Symbols.Global(Symbols.EmptyPackage, Descriptor.Term(symbol))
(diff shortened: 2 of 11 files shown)


metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -69,6 +71,15 @@ case class BuildServerConnection(
     register(server.buildTargetCompile(params))
   }
 
+  def mainClasses(
+      params: ScalaMainClassesParams
+  ): CompletableFuture[ScalaMainClassesResult] = {
+    // TODO use server.buildTargetScalaMainClasses when bloop realises version supporting mainClasses
```
olafurpg: > releases

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -0,0 +1,79 @@
...
+
+import java.util.concurrent.atomic.AtomicReference
+import ch.epfl.scala.bsp4j.ScalaMainClass
+import ch.epfl.scala.{bsp4j => b}
+import scala.collection.concurrent.TrieMap
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.meta.internal.metals.MetalsEnrichments._
+
+final class BuildTargetClasses(
```
olafurpg: Docstring proposal

> Manages in-memory index of main/test class symbols grouped by their enclosing build targets.

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -0,0 +1,79 @@
...
+    buildServer: () => Option[BuildServerConnection]
+)(implicit val ec: ExecutionContext) {
+  private val index = new TrieMap[b.BuildTargetIdentifier, Classes]()
+
+  def isInitialized(target: b.BuildTargetIdentifier): Boolean =
```
olafurpg: Where is `isInitialized` used?

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    buildServer() match {
+      case Some(connection) =>
+        val parameters = new b.ScalaMainClassesParams(targets.asJava)
+        val task = for {
+          mainClasses <- connection.mainClasses(parameters).asScala
+          _ = initializeMainClasses(mainClasses)
+        } yield
+          for (target <- targets) {
```
olafurpg: nit: Please use `targets.foreach { target => ... }` instead. This codebase 

- uses `.foreach` for basic iterations
- uses for comprehensions for non-trivial `map/filter/flatMap` chains
- never uses the `for (a <- b)` because it's not very flexible

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  private def classesOf(target: b.BuildTargetIdentifier): Classes =
+    index.getOrElseUpdate(target, new Classes)
+
+  private def initializeMainClasses(result: b.ScalaMainClassesResult): Unit = {
+    def createObjectSymbol(className: String): String = {
+      val symbol = className.replaceAll("\\.", "/") + "."
+      val isInsideEmptyPackage = !className.contains(".")
+      if (isInsideEmptyPackage) {
+        "_empty_/" + symbol
```
olafurpg: Let's use the SemanticDB symbol constructors
```scala
Symbols.Global(EmptyPackage, symbol)
```

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -0,0 +1,79 @@
...
+      item <- result.getItems.asScala
+      target = item.getTarget
+      aClass <- item.getClasses.asScala
+      objectSymbol = createObjectSymbol(aClass.getClassName)
+    } classesOf(target).main.put(objectSymbol, aClass)
+  }
+
+  final class Classes {
+    val isInitialized = new AtomicReference[Boolean](false)
+    val main = new TrieMap[String, ScalaMainClass]()
```
olafurpg: nit: `TrieMap.empty[...]`

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -39,6 +39,15 @@ object ClientCommands {
        |""".stripMargin
   )
 
+  val RunCode = Command(
```
olafurpg: `RunMain`?

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -39,6 +39,15 @@ object ClientCommands {
...
+    "metals-scala-run",
```
olafurpg: `metals-main-run`?

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -39,6 +39,15 @@ object ClientCommands {
...
+    "run",
+    "Runs main method",
+    """|`string`, Build target identifier URI
```
olafurpg:
```
Array of strings with length 2 where
- first array element is a build target identifier URI
- second array element is the name of the class containing the main method 
Example: `["mybuild://core-test", "com.app.Main"]`
```

metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  // we are maintaining a separate queue for cascade compilation since those must happen ASAP
+  private val compilationBatch = new CompilationBatch
+  private val cascadeBatch = new CascadeCompilationBatch
+
+  private val isCompiling = TrieMap.empty[b.BuildTargetIdentifier, Boolean]
+  private var lastCompile: collection.Set[b.BuildTargetIdentifier] = Set.empty
+
+  def currentlyCompiling: Iterable[b.BuildTargetIdentifier] = isCompiling.keys
+  def currentlyCompiling(buildTarget: b.BuildTargetIdentifier): Boolean =
```
olafurpg: `isCurrentlyCompiling`

metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
```diff
@@ -0,0 +1,121 @@
...
+    isCompiling.contains(buildTarget)
+
+  def previouslyCompiled: Iterable[b.BuildTargetIdentifier] = lastCompile
+  def previouslyCompiled(buildTarget: b.BuildTargetIdentifier): Boolean =
```
olafurpg: `isPreviouslyCompiled`

metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  def cascadeCompileFiles(paths: Seq[AbsolutePath]): Future[b.CompileResult] =
+    cascadeBatch.enqueue(paths)
+
+  def cancel(): Unit = {
+    compilationBatch.cancel()
+    cascadeBatch.cancel()
+  }
+
+  private abstract class Batch extends Cancelable {
```
olafurpg: Can we try to remove this class? This is a heavy abstraction IMO for a simple `Seq[BuildTargetIdentifier] => Seq[BuildTargetIdentifier]` function. 

I would prefer to keep
```scala
private val compile = new BatchedFunction[b.BuildTargetIdentifier, b.CompileResult](compile)
private val cascade = new BatchedFunction[b.BuildTargetIdentifier, b.CompileResult](compile)
```
and then expand the targets at the call-site
```scala
 def compileFiles(paths: Seq[AbsolutePath]): Future[b.CompileResult] =
    compile(expand(paths))

  def cascadeCompileFiles(paths: Seq[AbsolutePath]): Future[b.CompileResult] =
    cascade(expand(paths).flatMap(inverseDependencies).distinct)
```
Benefits of this approach:
- no need to come up with slightly different but new names for existing abstactions: `Batch` instead of `BatchedFunction`, `Batch.enqueue` instead of `BatchedFunction.apply`
- uses no abstract methods like `Batch.resolve`, which are tricky to reason about in my experience

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
marek1840: then the compiler spits warnings at me (it wants type annotation and I don't)

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
marek1840: Is there something like `Symbol{ def apply(aClass: Class[_]) }`?

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
olafurpg: Aah, I had forgotten about that one in IntelliJ 😅 It's not really the "compiler" though, just a custom IntelliJ inspection that you can disable if you want :p 

Anyways, fair enough, we might want to use `new` in other places as well in that case... I generally like `.empty[T]` constructors because they read nicely and work consistently for options and collections.

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
olafurpg: No, because it's not always the case that the symbol we're processing is available on the classpath. Also, many symbols have no corresponding `Class[_]`, for example vals.

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -39,6 +39,17 @@ object ClientCommands {
        |""".stripMargin
   )
 
+  val RunMain = Command(
+    "metals-main-run",
+    "run",
```
olafurpg: `Run` instead of `run`?

olafurpg: This can be changed down the road once the Bloop release is out.

PR #729, opened 2019-05-24
Miscellaneous polish fixes

olafurpg:
- clear diagnostics on file delete events
- don't crash completions when editing the last character in a file
- don't display "$(sync)" for "Downloading Bloop" message

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -669,7 +669,11 @@ class MetalsLanguageServer(
       event: DirectoryChangeEvent
   ): CompletableFuture[Unit] = {
     val path = AbsolutePath(event.path())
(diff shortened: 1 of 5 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -669,7 +669,11 @@ class MetalsLanguageServer(
       event: DirectoryChangeEvent
   ): CompletableFuture[Unit] = {
     val path = AbsolutePath(event.path())
-    if (!savedFiles.isRecentlyActive(path) && path.isScalaOrJava) {
+    val isScalaOrJava = path.isScalaOrJava
+    if (isScalaOrJava && event.eventType() == EventType.DELETE) {
+      diagnostics.didDelete(path)
```
olafurpg: We already have tests to see that diagnostics are cleared for deleted files after a `didSave` notification. I manually verified that this works as soon as the file is deleted without a `didSave` notification. However, I don't think we can write a non-flaky test case to stress this situation since file watchers are quite unreliable between operating systems.

PR #731, opened 2019-05-28
Ensure build server shuts down before starting new build connection.

olafurpg:
Previously, we risked connecting to a new build server before completing
the shutdown for the old build connection resulting in the following
sequence of requests:

```
metals ---shutdown----> bloop (old connection)
metals ---initialize--> bloop (new connection)
metals <--shutdown----- bloop (old connection)
```

Now, Metals waits for the `shutdown` response from the old connection
before sending `initialize` for the new connection.

This bug was introduced by a recent refactoring of how we shut down
the build server connection. This commit adds a new test case to
prevent similar regressions in the future.

### DIFF
--- a/tests/unit/src/test/scala/tests/BillSlowSuite.scala
+++ b/tests/unit/src/test/scala/tests/BillSlowSuite.scala
@@ -2,15 +2,17 @@ package tests
 
 import bill._
 import scala.concurrent.Future
-import scala.meta.internal.metals.Messages
 import scala.meta.io.AbsolutePath
+import scala.meta.internal.metals.Messages
+import scala.meta.internal.metals.ServerCommands
+import scala.meta.internal.metals.Directories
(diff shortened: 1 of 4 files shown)


tests/unit/src/test/scala/tests/BillSlowSuite.scala
```diff
@@ -45,6 +47,46 @@ object BillSlowSuite extends BaseSlowSuite("bill") {
...
+          .resolve(Directories.log)
+          .readText
+          .linesIterator
+          .filter(_.startsWith("trace:"))
+          .mkString("\n")
+        assertNoDiff(
+          logs,
+          // Assert that "Connect to build server" waits for the shutdown
+          // response from the build server before sending "initialize".
+          """|trace: initialize
```
olafurpg: I manually verified that this test failed on the current master.

PR #735, opened 2019-05-29
Enforce scala version for embedded bloop server

tgodzik:
Previously we could download a wrong version of scala-compiler as a transitive dependency, now we make sure we have the right one.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
@@ -98,7 +98,9 @@ final class Embedded(
 }
 
 object Embedded {
-  def downloadSettings(dependency: Dependency): Settings =
+  def downloadSettings(
+      dependency: Dependency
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -98,7 +98,10 @@ final class Embedded(
 }
 
 object Embedded {
-  def downloadSettings(dependency: Dependency): Settings =
+  def downloadSettings(
+      dependency: Dependency,
+      forceDependencies: List[Dependency] = Nil
```
olafurpg: nit: is this parameter necessary? We can use `withForceVersions` at the call-site instead

tgodzik: Ok, added it there. Sorry for the many force pushes - somehow I wasn't formatting although I was sure I was.

PR #737, opened 2019-05-31
Add automatic Mill import to metals

tgodzik:
Previously mill builds would have to be imported manually via bloop and now we add automatic mill import to metals whenever the workspace is detected.

Also, moved build tools digest calculation to separate files.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
@@ -1,0 +1,64 @@
+package scala.meta.internal.builds
+import scala.meta.internal.metals.{MetalsServerConfig, UserConfiguration}
+import scala.meta.io.AbsolutePath
+import scala.util.Properties
+import java.nio.file.Files
+import java.nio.file.Path
+
+case class MillBuildTool() extends BuildTool {
+
+  private val predefScript =
+    s"import $$ivy.`com.lihaoyi::mill-contrib-bloop:$version`".getBytes()
+
+  private val predefScriptName = "predef.sc"
+
+  private lazy val predefScriptPath: Path = {
+    Files.write(tempDir.resolve(predefScriptName), predefScript)
+  }
+
+  private lazy val embeddedMillWrapper: AbsolutePath = {
+    val millWrapper =
+      if (Properties.isWin) "millw.bat"
+      else "millw"
+    val out = BuildTool.copyFromResource(tempDir, millWrapper)
+    out.toFile.setExecutable(true)
+    AbsolutePath(out)
+  }
+
+  override def redirectErrorOutput: Boolean = true
+
+  override def args(
+      workspace: AbsolutePath,
+      userConfig: () => UserConfiguration,
+      config: MetalsServerConfig
+  ): List[String] = {
+    val cmd = List(
+      "--predef",
+      predefScriptPath.toString,
+      "mill.contrib.Bloop/install"
+    )
+    userConfig().millScript match {
+      case Some(script) =>
+        script :: cmd
+      case None =>
+        embeddedMillWrapper.toString() :: "--mill-version" :: version :: cmd
+    }
+  }
+
+  override def digest(workspace: AbsolutePath): Option[String] =
+    MillDigest.current(workspace)
+
+  override def minimumVersion: String = "0.4.0"
+
+  override def version: String = "0.4.0"
+
+  override def toString(): String = "mill"
+
+}
+
+object MillBuildTool {
+  def isMillRelatedPath(workspace: AbsolutePath, path: AbsolutePath) = {
+    val filename = path.toNIO.getFileName.toString
+    filename.endsWith(".sc")
+  }
+}
--- a/metals/src/main/scala/scala/meta/internal/builds/MillDigest.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/MillDigest.scala
@@ -1,0 +1,96 @@
+package scala.meta.internal.builds
+import scala.util.control.NonFatal
+import scala.meta.tokens.Token
+import java.nio.file.Paths
+import scala.meta.io.AbsolutePath
+import scala.collection.mutable
+import java.security.MessageDigest
+import scala.meta.internal.mtags.MtagsEnrichments._
+
+object MillDigest extends Digestable {
+  override protected def digestWorkspace(
+      workspace: AbsolutePath,
+      digest: MessageDigest
+  ): Boolean = {
+    def analyzeBuildScript(file: AbsolutePath): Boolean = {
+      val imported = findImportedScripts(file)
+      imported.forall(script => analyzeBuildScript(script)) && Digest
+        .digestFile(file, digest)
+    }
+    val buildFile = workspace.resolve("build.sc")
+    analyzeBuildScript(buildFile)
+  }
+
+  private case class ImportLinesAcc(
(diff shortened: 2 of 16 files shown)


metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
```diff
@@ -0,0 +1,65 @@
+package scala.meta.internal.builds
+import scala.meta.internal.metals.{MetalsServerConfig, UserConfiguration}
+import scala.meta.io.AbsolutePath
+import scala.util.Properties
+import java.nio.file.Files
+import java.nio.file.Path
+
+class MillBuildTool extends BuildTool {
```
gabro: Minor, but I've noticed that SbtBuildTool is a case class while MillBuildTool and GradleBuildTool are not. Is this a choice or is it accidental?

tgodzik: accidental, will fix. Thanks!

metals/src/main/scala/scala/meta/internal/builds/MillDigest.scala
```diff
@@ -0,0 +1,92 @@
+package scala.meta.internal.builds
+import scala.util.control.NonFatal
+import scala.meta.tokens.Token
+import java.nio.file.Paths
+import scala.meta.io.AbsolutePath
+import scala.collection.mutable
+import java.security.MessageDigest
+import scala.meta.internal.mtags.MtagsEnrichments._
+
+object MillDigest extends Digestable {
```
tgodzik: So this doesn't cover all corner cases yet, I think I might need to rethink how to make it simpler.

If anyone got any ideas let me know, but I think I might need to filter out whitespaces and then try matching.

tgodzik: Should be better now.

metals/src/main/scala/scala/meta/internal/builds/MillDigest.scala
```diff
@@ -0,0 +1,98 @@
...
+      val input = file.toInput
+      val tokens = input.tokenize.get.tokens
+      val acc = ImportLinesAcc()
+      tokens.toList.foreach {
+        case token =>
+          token match {
+            // recognize line with import statement
+            case _: Token.KwImport =>
+              acc.hadImport = true
+              acc
```
marek1840: remove

metals/src/main/scala/scala/meta/internal/builds/MillDigest.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  // find all `import $file.path`
+  private def findImportedScripts(
+      file: AbsolutePath
+  ): List[AbsolutePath] = {
+    try {
+      val input = file.toInput
+      val tokens = input.tokenize.get.tokens
+      val acc = ImportLinesAcc()
+      tokens.toList.foreach {
```
olafurpg: is the `.toList` needed?

tests/unit/src/main/scala/tests/TestingClient.scala
```diff
@@ -169,7 +169,9 @@ final class TestingClient(workspace: AbsolutePath, buffers: Buffers)
     def isSameMessage(
         createParams: String => ShowMessageRequestParams
     ): Boolean = {
-      createParams("gradle") == params || createParams("sbt") == params
+      Set("gradle", "sbt", "maven", "mill")
+        .map(createParams)
+        .exists(_ == params)
```
olafurpg:
```suggestion
        .contains(params)
```

tests/slow/src/test/scala/tests/MillSlowSuite.scala
```diff
@@ -0,0 +1,239 @@
...
+        List(
+          ImportBuild.params("mill").getMessage,
+          bloopInstallProgress("mill").message
+        ).mkString("\n")
+      )
+      _ = assertStatus(_.isInstalled)
+    } yield ()
+  }
+
+  testAsync("fatal-warnings") {
```
olafurpg: This test can be removed since we don't modify the build settings, it's failing right now.

metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  override def digest(workspace: AbsolutePath): Option[String] =
+    MillDigest.current(workspace)
+
+  override def minimumVersion: String = "0.4.0"
+
+  override def version: String = "0.4.0"
+
+  override def toString(): String = "mill"
```
olafurpg:
```suggestion
  override def toString(): String = "Mill"
```
?

metals/src/main/scala/scala/meta/internal/builds/MillDigest.scala
```diff
@@ -0,0 +1,98 @@
...
+              acc.current = ""
+              acc.skip = false
+              if (token.isInstanceOf[Token.RightBrace]) {
+                acc.insideBraces = false
+                acc.prefix = ""
+              }
+              if (!acc.insideBraces) {
+                acc.hadFile = false
+              }
+              if (token.isInstanceOf[Token.LF] || token
```
olafurpg: nit; please try to manually break after `||` and `&&`.

olafurpg: scalafmt doesn't do it for you.

metals/src/main/scala/scala/meta/internal/builds/MillDigest.scala
```diff
@@ -0,0 +1,98 @@
...
+  // find all `import $file.path`
+  private def findImportedScripts(
+      file: AbsolutePath
+  ): List[AbsolutePath] = {
+    try {
+      val input = file.toInput
+      val tokens = input.tokenize.get.tokens
+      val acc = ImportLinesAcc()
+      tokens.toList.foreach {
+        case token =>
```
olafurpg: nit: no need for `case`

metals/src/main/scala/scala/meta/internal/builds/MillDigest.scala
```diff
@@ -0,0 +1,98 @@
...
+          token match {
```
olafurpg: I'm not a huge fan of the implementation details in this mini-parser. I have a feeling it can be simplified a bit, but if it works as it is then I'm happy 😊

metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
tgodzik: Will fix that up along with maven

metals/src/main/scala/scala/meta/internal/builds/MillDigest.scala
tgodzik: Not needed - will remove

metals/src/main/scala/scala/meta/internal/builds/MillDigest.scala
tgodzik: I think I am fine with the implementation now, but if you have any suggestions let me know. Marek wrote something trying to prove that it can be simpler, but turns out it is of a similar complexity.

olafurpg:
Travis CI is green even if the GitHub status says it's pending. Feel free to merge.

Issue #672, opened 2019-04-15
Codebase cleanup


olafurpg:
With all the recent development I think it's time to do a few code cleanups

- ~~rename "SlowSuite" into "LspSuite" to better reflect their purpose: test interaction between editor and server via LSP. These tests were historically very slow to run (hence the name) but thanks to `QuickBuild` these tests can run modestly fast now (1-3 seconds)~~
- ~~split `Completions.scala` into smaller files, it's at 1.6k lines now with a lot of unrelated functionality grouped in one place~~
- split `MetalsLanguageServer.scala` into smaller files, it contains several utility functions that don't need to be next to the LSP request/response methods
- ~~refactor the mess of null vars in `MetalsLanguageServer` so we get compile errors for forward references~~
- split `TestingServer.scala` into smaller files
- split cross tests into multiple packages: tests.completion, tests.definition, ...
- ~~add a `prePR` command to run scalafmt+scalafix before a PR, currently you need to run the steps separately (`scalafixAll` in sbt shell and `./bin/scalafmt --diff` from the terminal)~~
- new packages: `async`, `editdistance`, `eval`
- remove unused global methods
- move helper methods that are only used for tests into test modules (for example `encloses` in #718)

EDIT [@tgodzik] : 
- make gradle import use gradle wrapper jar instead of script

tgodzik:
> add a prePR command to run scalafmt+scalafix before a PR, currently you need to run the steps separately (scalafixAll in sbt shell and ./bin/scalafmt --diff from the terminal)

Do you think we could just install a git hook that runs /bin/scalafmt before commit? Though it might be annoying if someone commits a lot.

gabro:
I had the same idea, but I think running sbt in a git hook is not going to be good ux, since it takes very long to run

olafurpg:
@tgodzik I think these steps are too slow to work as a git commit or push hook. I personally commit and push way more often than I open a pull request. What we could do is create a PR description template that reminds you to run `prePR`

tgodzik:
We could only run scalafmt on changed files, which would be much faster, with something like:

`git diff HEAD..HEAD^1 --name-only | ./bin/scalafmt`

Not a bash expert though.

olafurpg:
@tgodzik `./bin/scalafmt --diff` does that for you, it only formats files that have changed from master

tgodzik:
[...]

PR #738, opened 2019-06-02
Setup git hooks for scalafmt and scalafix.

olafurpg:
Previously, contributors needed to manually run scalafmt and scalafix
when opening pull requests in order to make the CI green. Now, we
automatically install git hooks so that scalafmt and scalafix are always
run before pushing a commit.

To speed up the `sbt scalafixAll` command, we use the experimental
`-client` flag which delegates the command to a running sbt server
instance if one exists. In my local experiments, this makes the pre-push
hooks run in ~5 seconds if the project is already compiled.

Related to #672

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -96,22 +96,21 @@ inThisBuild(
 )
 
 onLoad.in(Global) ~= { old =>
-  import java.nio.file._
-  val sh =
-    if (scala.util.Properties.isWin) "#!C:/Program\\ Files/Git/usr/bin/sh.exe"
-    else "#!/bin/sh"
-  val prePush = Paths.get(".git", "hooks", "pre-push")
-  Files.createDirectories(prePush.getParent)
-  Files.write(
-    prePush,
-    s"""$sh
-       |set -eux
-       |sbt -client 'all compile:scalafix test:scalafix'
-       |bin/scalafmt --diff
-       |git diff --exit-code
-       |""".stripMargin.getBytes()
-  )
-  prePush.toFile.setExecutable(true)
+  if (!scala.util.Properties.isWin) {
+    import java.nio.file._
+    val prePush = Paths.get(".git", "hooks", "pre-push")
+    Files.createDirectories(prePush.getParent)
+    Files.write(
+      prePush,
+      """#!/bin/sh
+        |set -eux
+        |sbt -client 'all compile:scalafix test:scalafix'
+        |bin/scalafmt --diff
+        |git diff --exit-code
+        |""".stripMargin.getBytes()
+    )
+    prePush.toFile.setExecutable(true)
+  }
   old
 }
 cancelable.in(Global) := true
--- a/build.sbt
+++ b/build.sbt
@@ -95,6 +95,25 @@ inThisBuild(
   )
 )
 
+onLoad.in(Global) ~= { old =>
+  import java.nio.file._
+  val sh =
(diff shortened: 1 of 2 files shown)


olafurpg:
If this turns out to be too annoying (slow git push) then we can revert later. I have never properly used git commit hooks before and I'm keen to give them a try.

build.sbt
```diff
@@ -95,6 +95,14 @@ inThisBuild(
   )
 )
 
+onLoad.in(Global) ~= { old =>
+  import scala.sys.process._
+  require(
+    "yarn install".! == 0,
+    "command failed: yarn install. Is yarn installed? https://yarnpkg.com/en/docs/install"
```
gabro: Since yarn is not exactly standard in Scala builds, maybe it's worth adding some context here in the error message?

gabro: Also, we're not really using anything node-specific here.
What about replacing husky with a simpler script that just installs the git hooks we need on load?

olafurpg:
I updated the `onLoad` hook to create the pre-push hook directly without yarn.

olafurpg:
Can someone with a Windows computer maybe verify this PR?

tgodzik:
> Can someone with a Windows computer maybe verify this PR?

I can do it when I am back at home. Got a desktop with windows.

tgodzik:
It doesn't seem to be so rosy on Windows unfortunately
Firstly, runinng the hook ruins the logs:
`[0m[[0m[0minfo[0m] [0m[0mentering *experimental* thin client - BEEP WHIRR[0m`

Running the same command on cmd doesn't cause any issues. That's not a big issue though.

Another issue is that -client option doesn't seem to work and just hangs with sbt server printing:
```
Exception in thread "sbt-networkchannel-0" java.io.IOException: GetOverlappedResult() failed for read operation: 109
        at org.scalasbt.ipcsocket.Win32NamedPipeSocket$Win32NamedPipeSocketInputStream.read(Win32NamedPipeSocket.java:161)
        at java.io.InputStream.read(InputStream.java:101)
        at sbt.internal.server.NetworkChannel$$anon$2.run(NetworkChannel.scala:147)
```

Also scalafmt seems not to work on windows:
```
Exception in thread "main" java.util.regex.PatternSyntaxException: Illegal/unsupported escape sequence near index 33
(test-workspace|metals-bench\src\main\resources|tests\unit\src\test\resources|sbt-metals\src\sbt-test)
```
:/

build.sbt
```diff
@@ -95,6 +95,25 @@ inThisBuild(
...
+  val sh =
+    if (scala.util.Properties.isWin) "#!C:/Program\\ Files/Git/usr/bin/sh.exe"
+    else "#!/bin/sh"
+  val prePush = Paths.get(".git", "hooks", "pre-push")
+  Files.createDirectories(prePush.getParent)
+  Files.write(
+    prePush,
+    s"""$sh
+       |set -eux
+       |sbt -client 'all compile:scalafix test:scalafix'
```
tgodzik: -client doesn't not work on windows

build.sbt
```diff
@@ -95,6 +95,25 @@ inThisBuild(
   )
 )
 
+onLoad.in(Global) ~= { old =>
+  import java.nio.file._
+  val sh =
+    if (scala.util.Properties.isWin) "#!C:/Program\\ Files/Git/usr/bin/sh.exe"
```
tgodzik: Maybe use the condition to not write the hook on windows for now.

olafurpg: Sounds good, done.

PR #739, opened 2019-06-03
Factor out a ClasspathLoader.getURLs(classLoader) helper for java9+ compatability, update usages

er1c:
I got metals mostly working on java11/updated bloop, here is one of the changes I had to update.  I'll try and split out changes as separate changes/PRs

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -369,7 +369,6 @@ class MetalsLanguageServer(
   ): CompletableFuture[InitializeResult] = {
     timed("initialize")(Future {
       setupJna()
-      warnUnsupportedJavaVersion()
       initializeParams = Option(params)
       updateWorkspaceDirectory(params)
       val capabilities = new ServerCapabilities()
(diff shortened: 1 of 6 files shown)


olafurpg:
Thank you for this contribution! I tried running the tests and got the following error 🤔 
```
> unit/testOnly -- tests.DocumentHighlightSlowSuite.single
...
[E] Caused by: java.lang.NoClassDefFoundError: javax/tools/ToolProvider
[E] 	at sbt.internal.inc.javac.JavaCompiler$.local(JavaCompiler.scala:68)
[E] 	at sbt.internal.inc.javac.JavaTools$.directOrFork(JavaCompiler.scala:55)
[E] 	at sbt.internal.inc.ZincUtil$.compilers(ZincUtil.scala:94)
[E] 	at bloop.CompilerCache.newCompilers(CompilerCache.scala:27)
```
Any thoughts on how to fix that error?

chikei:
Changing [parent](https://github.com/scalameta/metals/blob/b88b840a1b1bdb2243b5343f90d01c04228b7c47/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala#L185) classloader from `null` to `ClassLoader.getPlatformClassLoader` should fix this.

This is from the JDK9 migration guide:
> The changes in JDK 9 may impact code that creates class loaders with null (that is, the bootstrap class loader) as the parent class loader and assumes that all platform classes are visible to the parent. Such code may need to be changed to use the platform class loader as the parent (see ClassLoader.getPlatformClassLoader).

er1c:
I have the same error on tests, I will try @chikei suggestion and resubmit

er1c:
Updated to: https://github.com/scalameta/metals/pull/765

PR #741, opened 2019-06-05
Update emacs docs as scala is supported by lsp-mode directly now

kurnevsky:


JesusMtnez:
:+1: LGTM

JesusMtnez:
Thank you @kurnevsky !

PR #750, opened 2019-06-11
Added clarification to the sublime documentation

ggreiner:
### Changes

I found that I needed to open a Scala file before LSP would prompt me to "Import Changes". I added a clarifying sentence to the Sublime documentation to explain this.

### Testing

I ran the website locally and verified changes were present.


olafurpg:
Ignoring flaky test failure on Travis. I suspect it started getting flaky after #737 because the Travis machines have little memory and we're testing all the build tools (sbt, maven, gradle and mill) in the same job 🤔

PR #751, opened 2019-06-11
Avoid inifinite loop when shortening types, fixes #749.

### LINKED ISSUE #749: Infinite loop doing completions
**Describe the bug**

Triggering a completion results in the Metals process getting stuck and using a lot of CPU.

**To Reproduce**
Trigger a completion at this line here https://github.com/megri/metals-bug-example/blob/c752debe3ba0cab3ac3bd6ae9175a684c0574cc6/js/src/main/scala/com/megri/App.scala#L11

Big thanks to @megri for minimizing the reproduction 🙏 

**Expected behavior**

The Metals process should return normal completions instead of getting stuck.

**Stack trace**

The stack trace leads to a call to `Symbols.overrides` in the compiler.
```
$ jps -l | grep scala.meta.metals.Main | awk '{ print $1 }' | xargs -I % jstack %
...
at scala.reflect.internal.Symbols$SymbolContextApiImpl.overrides(Symbols.scala:208)
    at scala.meta.internal.pc.MetalsGlobal.loop$1(MetalsGlobal.scala:192)
...
```

**Installation:**
- Metals version: v0.5.0

**Search terms**

infinite loop, forever, slow, frozen, hot

olafurpg:
Previously, triggering completions on a type that had recursive type
aliases would make Metals get stuck in an infinite loop. Now, we prevent
infinite loops by adding a cache to what types we have tried to shorten.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/MetalsGlobal.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/MetalsGlobal.scala
@@ -167,95 +167,103 @@ class MetalsGlobal(
    * making sure to not convert two different symbols into same short name.
    */
   def shortType(longType: Type, history: ShortenedNames): Type = {
-    def loop(tpe: Type, name: Option[ShortName]): Type = tpe match {
-      case TypeRef(pre, sym, args) =>
-        val ownerSymbol = pre.termSymbol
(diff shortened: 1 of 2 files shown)


gabro:
tip for the review: the diff reads much nicer disabling the whitespaces

PR #763, opened 2019-06-13
Add 2.13 support and drop deprecated versions

### LINKED ISSUE #761: Scala 2.13 support
Metals currently doesn't support 2.13 builds. Here are the thing that need to happen to add 2.13 support:

- [ ] publish `semanticdb-scalac` for 2.13 (https://github.com/scalameta/scalameta/issues/1873)
- [ ] fix whatever breaks with the PC integration
- [ ] add tests for 2.13
- [ ] update docs

I'm sure I'm missing something, but let's add it to this post as we find it out.

gabro:
Closes #761  

TODO:

- [x] figure out the `Traversable` -> `Iterable` migration
- [x] deprecate 2.11.12 and 2.12.7
- [x] figure out the `ControlThrowable` thing
- [x] make tests pass


build.sbt
```diff
@@ -137,10 +139,8 @@ lazy val V = new {
   // List of supported Scala versions in SemanticDB. Needs to be manually updated
   // for every SemanticDB upgrade.
   def supportedScalaVersions =
-    Seq("2.12.8", "2.12.7", "2.11.12") ++ deprecatedScalaVersions
-  def deprecatedScalaVersions = Seq[String](
-    "2.12.6", "2.12.5", "2.12.4", "2.11.11", "2.11.10", "2.11.9"
-  )
+    Seq("2.13.0-RC1", "2.12.8", "2.12.7", "2.11.12")
+  def deprecatedScalaVersions = Seq[String]()
```
gabro: This is completely arbitrary, but I think it might be a good time to drop the deprecated Scala versions. More concretely, we need Scalameta 4.1.11, which is not published for the deprecated versions anymore.

build.sbt
```diff
@@ -174,7 +174,7 @@ lazy val mtags = project
       if (isCI) Nil
       // NOTE(olafur) pprint is indispensable for me while developing, I can't
       // use println anymore for debugging because pprint.log is 100 times better.
-      else List("com.lihaoyi" %% "pprint" % "0.5.3")
+      else Nil // List("com.lihaoyi" %% "pprint" % "0.5.4")
```
gabro: To bring back as soon as we test against 2.13.0 final.

mtags/src/main/scala/scala/meta/internal/metals/ControlCancellationException.scala
```diff
@@ -8,7 +8,7 @@ import scala.util.control.ControlThrowable
  */
 class ControlCancellationException(
     cause: CancellationException = new CancellationException()
-) extends CancellationException(cause.getMessage)
-    with ControlThrowable {
+) extends CancellationException(cause.getMessage) {
+  // with ControlThrowable { // TODO(gabro)
```
gabro: I don't know enough about how this is used to understand how to fix it.

`ControlThrowable` has been made an abstract class in 2.13, so we can't mix it in this way anymore. See a relevant discussion at https://github.com/scala/bug/issues/11494

mtags/src/main/scala/scala/meta/internal/metals/TrigramSubstrings.scala
```diff
@@ -5,7 +5,10 @@ import java.lang.StringBuilder
 /**
  * Returns all possible substrings of length 3 for the given string.
  */
-class TrigramSubstrings(string: String) extends Traversable[String] {
+class TrigramSubstrings(string: String) extends Iterable[String] {
+  // TODO(gabro)
+  override def iterator: Iterator[String] = ???
```
gabro: I've literally never in my life implemented an `Iterator`, but this shouldn't be hard I presume.

mtags/src/main/scala/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
```diff
@@ -11,6 +11,7 @@ import scala.meta.internal.semanticdb.Scala._
 import scala.meta.internal.{semanticdb => s}
 import scala.meta.io.AbsolutePath
 import scala.util.control.NonFatal
+import scala.meta.internal.mtags.ListFiles
```
gabro: I'm not sure what was happening here, but it seemed that this `ListFiles` was being shadowed by `scala.meta.internal.io.ListFiles`

mtags/src/main/scala/scala/meta/internal/mtags/ListFiles.scala
```diff
@@ -1,40 +1,15 @@
...
-            )
-          } catch {
-            case _: NoSuchFileException =>
-              () // error is reported by the JDK
-          }
-        }
-      }
+  def apply(root: AbsolutePath): Iterable[AbsolutePath] =
+    new Iterable[AbsolutePath] {
+      override def iterator: Iterator[AbsolutePath] =
```
olafurpg: An iterator is not safe to use when walking file trees since we need to properly close the streams provided by NIO. It's better to return a `collection.Seq` like `ArrayBuffer`. It might be worth considering using Haoyi's geny here eventually 🤔

mtags/src/main/scala/scala/meta/internal/metals/TrigramSubstrings.scala
```diff
@@ -43,7 +46,10 @@ object TrigramSubstrings {
   def uppercased(
       query: String,
       maxCount: Int = 250
-  ): Traversable[String] = new Traversable[String] {
+  ): Iterable[String] = new Iterable[String] {
```
olafurpg: Same here, `collection.Seq` implemented by `ArrayBuffer` is fine

tests/slow/src/test/scala/tests/SbtSlowSuite.scala
```diff
@@ -180,7 +180,7 @@ object SbtSlowSuite extends BaseImportSuite("sbt-import") {
       _ = assertStatus(!_.isInstalled)
       _ = client.messageRequests.clear()
       _ <- server.didSave("build.sbt") { _ =>
-        """scalaVersion := "2.12.7" """
+        """scalaVersion := "2.12.88888888" """
```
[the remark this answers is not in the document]
gabro: Haha, I guess I typed `8i8` in vim 😅

mtags/src/main/scala/scala/meta/internal/metals/TrigramSubstrings.scala
godenji: @gabro  see this scala 2.13 [write up](https://docs.scala-lang.org/overviews/core/custom-collections.html) by Odersky et al.
```
trait CustomIterable { self =>
  def iterator: Iterator[Foo] = new AbstractIterator[Foo] {
    private var current = 0
    def hasNext = current < self.length
    def next(): Foo = {
      val elem = self(current)
      current += 1
      elem
    }
  }

  def children: List[Foo] // implement in extended classes
  def apply(i: Int) = children(i)
  def length = children.size
}
```

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -89,19 +88,15 @@ abstract class BasePCSuite extends BaseSuite {
...
-          List(
-            new Dependency(
-              "org.scala-lang",
-              "scala-library",
-              BuildInfoVersions.scala212
-            )
-          )
-        )
-    )
+    val sources = Fetch
```
gabro: @olafurpg this should do it.

gabro:
Ok, we finally got to the point where tests are failing due to legitimate reasons! 🎉 

At first glance, most failures are due to the changes in the collection library, which cause the results for completions, hover and friends to be different since 2.12.

I'm not sure of what's the best approach here. Split the 2.13 tests? Change the examples in cross tests so that they're consistent across multiple versions?

olafurpg:
> I'm not sure of what's the best approach here. Split the 2.13 tests? Change the examples in cross tests so that they're consistent across multiple versions?

I would go with `compat = Map("2.13" -> ...)`. It requires a lot of thought to design tests against scala-library APIs that don't change between compiler versions.

gabro:
> I would go with compat = Map("2.13" -> ...)

Not sure what this means. I'm moving the offending suites in dedicated directories, e.g.

```
tests/cross/src/test/scala-2.12/tests/pc/SignatureHelpSuite.scala
tests/cross/src/test/scala-2.13/tests/pc/SignatureHelpSuite.scala
```

Is that what you mean>

olafurpg:
See https://github.com/scalameta/metals/blob/4acd4659473321b897ce4059215656dbce4353d5/tests/cross/src/test/scala/tests/pc/SignatureHelpSuite.scala#L87-L91

olafurpg:
It shouldn't be necessary to move files to scala specific directories.

gabro:
Ahh, got it. Makes sense

gabro:
Update: I've fixed a bazilion tests in the `cross` project that only required trivial changes due to obvious differences in 2.13.

We're left with a few (<10 IIRC) cross tests that fail for non-obvious reasons, i.e. it should be worth to look into them as they may be bugs.

metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
```diff
@@ -60,7 +60,7 @@ case class GradleBuildTool() extends BuildTool {
        |              scalaCompilerPlugin
        |          }
        |          dependencies {
-       |              scalaCompilerPlugin 'org.scalameta:semanticdb-scalac_' + scalaLib.version + ':4.1.4'
+       |              scalaCompilerPlugin 'org.scalameta:semanticdb-scalac_' + scalaLib.version + ':4.1.12'
```
olafurpg: This version should come from `BuildInfo.scalameta`

build.sbt
```diff
@@ -303,17 +339,29 @@ lazy val mtest = project
...
+    scalacOptions --= crossSetting(
+      scalaVersion.value,
+      if211 = List("-Ywarn-unused:imports")
     ),
     buildInfoPackage := "tests",
     buildInfoObject := "BuildInfoVersions",
     buildInfoKeys := Seq[BuildInfoKey](
       "scala211" -> V.scala211,
-      "scala212" -> V.scala212
+      "scala212" -> V.scala212,
```
olafurpg: Missing `scala213`

docs/build-tools/sbt.md
```diff
@@ -117,6 +117,6 @@ Try to remove any usage of `sbt-metals` in your build.
 
 ### Module not found: org.scalameta#semanticdb-scalac_2.12.4
 
-Metals works only with Scala 2.12.7 and 2.11.12 at the moment, down to the exact
+Metals works only with Scala 2.13.0, 2.12.8, 2.12.7, and 2.11.12 at the moment, down to the exact
```
olafurpg: We can remove this FAQ entry since we only add the compiler plugin for supported Scala versions. We already day on the editor installation pages what exact scala versions are supported (which is generated from build.sbt settings).

metals-docs/src/main/scala/docs/RequirementsModifier.scala
```diff
@@ -20,15 +20,16 @@ class RequirementsModifier extends StringModifier {
...
+       |**OpenJDK or Oracle Java 8**. Eclipse OpenJ9 and Java 11 are not
+       |supported, please make sure the `JAVA_HOME` environment variable
        |points to valid Java 8 installation.
        |
        |**macOS, Linux or Windows**. Metals is developed on macOS and every PR is
        |tested on Ubuntu+Windows.
        |
-       |**Scala 2.12 and 2.11**. Metals works only with Scala versions $supportedScalaVersions.
-       |Note that 2.10.x and 2.13.x are not supported.
+       |**Scala 2.13 and 2.12**. Metals supports these Scala versions $supportedScalaVersions.
```
olafurpg: Let's keep `Scala 2.13, 2.12 and 2.11` and add instead "Note that 2.11.x support is deprecated, it's recommended to upgrade to 2.12.".

mtags/src/main/scala/scala/meta/internal/metals/TrigramSubstrings.scala
```diff
@@ -1,34 +1,40 @@
...
-          arr(1) = string.charAt(j)
-          arr(2) = string.charAt(k)
-          f(new String(arr))
-          k += 1
+object TrigramSubstrings {
+
+  /**
+   * Returns all possible substrings of length 3 for the given string.
+   */
+  def apply(string: String): ArrayBuffer[String] = {
```
olafurpg: Hmm, I'm wondering if we should do instead 

```scala
object TrigramSubstrings {
  def foreach(string: String, fn: String => Unit): Unit = ...
  def seq(string: String): Seq[String] = {
    val buf = ArrayBuffer.empty[String]
    foreach(string, trigram => buf += trigram)
    buf.result()
  }
}
```

It looks like we don't need the intermediary buffer in `Fuzzy`. This is a **very** hot path.

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -89,19 +87,21 @@ abstract class BasePCSuite extends BaseSuite {
...
-              "scala-library",
-              BuildInfoVersions.scala212
-            )
-          )
+    val sources = Fetch()
+      .addClassifiers(Classifier.sources)
+      .addDependencies(
+        Dependency(
+          mod"org.scala-lang:scala-library",
+          // NOTE(gabro): we should ideally just use BuildoInfoVersions.scalaVersion
```
olafurpg: :+1:

metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
tgodzik: Ups, we should probably do `${BuildInfo.semanticdbVersion}` - I should have done it from the start. Can you change it here?

tests/slow/src/test/scala/tests/feature/DefinitionCrossSlowSuite.scala
```diff
@@ -32,4 +32,33 @@ object DefinitionCrossSlowSuite
       _ = assertNoDiff(client.workspaceDiagnostics, "")
     } yield ()
   }
+
+  testAsync("2.13") {
```
olafurpg: Can we abstract over this test so that it's clearer we test the same functionality on 2.11 and 2.13?

metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
gabro: Sure, I was about to yesterday, but I wasn't sure. Thanks for confirming.

gabro:
Thanks for the reviews! I'll work on this a bit more later today and hopefully get to a green build 💪

build.sbt
gabro: I'll add it for consistency, but it's not used anywhere.

build.sbt
```diff
@@ -1,5 +1,28 @@
...
+    scalaVersion: String,
+    if211: List[A],
+    otherwise: List[A] = Nil
+): List[A] =
+  CrossVersion.partialVersion(scalaVersion) match {
+    case Some((2, 11)) => if211
+    case _ => otherwise
+  }
+
+def crossSetting[A](
```
gabro: memo, this is probably unused

gabro:
Travis is green 🚀
image

I'm not sure about this failure in the `MillDigestSuite` (it fails on both appveyor and azure):

```
2019-06-18T16:44:28.3126526Z   tests.TestFailedException: obtained=<9F74416ABE12B64362A9CCCA3BFC54E6> == expected=<9F74416ABE12B64362A9CCCA3BFC54E6> (hint: The digests should not be equal)
```

I'll take a look

gabro:
Ok, fixed. It was an error I've introduced by mistake.

gabro:
Tried this locally with a simple Scala 2.13 project and everything seems to work 🙌

mtags/src/main/scala/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
```diff
@@ -5,7 +5,7 @@ import java.nio.charset.StandardCharsets
 import scala.collection.concurrent.TrieMap
 import scala.meta.inputs.Input
 import scala.meta.internal.io.PathIO
-import scala.meta.internal.io._
+import scala.meta.internal.io.{ListFiles => _, _}
```
tgodzik: Why do we do `ListFiles => _` here? I don't see any change conencted to that.

gabro: Great question, I don't know 😅
On Scala 2.13 the `ListFiles` from `scala.meta.internal.io` takes precedence over our `ListFiles` so it doesn't compile.

tgodzik: Ach alright, I didn't know we had multiples of that. Thanks!

PR #764, opened 2019-06-13
Include mill version from .mill-version file

tgodzik:


.travis.yml
```diff
@@ -3,7 +3,7 @@ jdk:
 os:
   - linux
 language: scala
-
+dist: trusty
```
marek1840: what does it do?

metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
```diff
@@ -37,11 +37,23 @@ case class MillBuildTool() extends BuildTool {
       predefScriptPath.toString,
       "mill.contrib.Bloop/install"
     )
+    import scala.collection.JavaConverters._
+    val millVersionPath = workspace.resolve(".mill-version")
+    val millVersion = if (millVersionPath.isFile) {
+      Files
+        .readAllLines(millVersionPath.toNIO)
```
marek1840: does this file contain anything except the one-line version? If no, then you can use `Files.readAllBytes`

.travis.yml
tgodzik: specifies distribution to be ubuntu trusty - I think there is an issue when the distro is newer and it doesn't have jdk 8.

metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
tgodzik: readAllBytes doesn't give me Strings so this is a bit more useful.

.travis.yml
tgodzik: @olafurpg does this make sense? I had issues with jdk8 before adding this, but I have no experience with travis.

olafurpg: LGTM

metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
olafurpg: There is FileIO.slurp from our internal io libraries.

PR #766, opened 2019-06-14
Fix concurrent modification

chikei:
in `def textDocument`, `textDocumentCache.computeIfAbsent` calls `compile` which modifies `textDocumentCache`, java 8 fails to detect this but java 11 detects correctly.


tgodzik:
Really cool! Thanks for finding this!

PR #765, opened 2019-06-13
Factor out a ClasspathLoader.getURLs(classLoader) helper for java9+ compatability, update usages

er1c:
This updates https://github.com/scalameta/metals/pull/739, tests still fail on java11, but no longer from  `Caused by: java.lang.NoClassDefFoundError: javax/tools/ToolProvider`

```
sbt:default-40f014> unit/testOnly -- tests.DocumentHighlightSlowSuite.single
[info] Compiling 1 Scala source to /c/Users/Eric Peters/IdeaProjects/metals/metals/target/scala-2.12/classes ...
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.protobuf.UnsafeUtil (file:/c/Users/Eric%20Peters/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scalameta/semanticdb-scalac_2.12.8/4.1.9/semanticdb-scalac_2.12.8-4.1.9.jar) to field java.nio.Buffer.address
WARNING: Please consider reporting this to the maintainers of com.google.protobuf.UnsafeUtil
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
[info] Done compiling.
------------ Running Tests tests.DocumentHighlightSlowSuite.single ------------
------------ Running Tests tests.DocumentHighlightSlowSuite.single ------------
INFO  tests.TestingServer#initialize
INFO  time: generated quick build in 1.3s
INFO  logging to file /c/Users/Eric Peters/IdeaProjects/metals/tests/unit/target/e2e/documentHighlight/single/.metals/metals.log
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.protobuf.UnsafeUtil (file:/c/Users/Eric%20Peters/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scalameta/semanticdb-scalac_2.12.8/4.1.9/semanticdb-scalac_2.12.8-4.1.9.jar) to field java.nio.Buffer.address
WARNING: Please consider reporting this to the maintainers of com.google.protobuf.UnsafeUtil
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
INFO  time: initialize in 0.28s
INFO  running embedded 'bloop bsp --protocol local --socket /tmp/bsp13549952426963183349/-9uxbm22vm0b.socket'
[...]

PR #767, opened 2019-06-14
Loads JDK jars by scalameta which is the same for java8 and also

chikei:
handle java9


chikei:
seems #765 included this, close for now

Issue #755, opened 2019-05-04
Fallback to "show usages" from "Goto definition" if reference is definition itself.


rossoha:
**Is your feature request related to a problem? Please describe.**
It's not related to a problem. 

**Describe the solution you'd like**
Simply feature is a "fallback" for "go to" in cases when you click root definition itself. Instead of ignoring this request (cause it's a definition but not a reference), it would be nice to show a list of usages/references ("peek references" in VSCode is an example).

**Describe alternatives you've considered**
no, so far

**Additional context**
![](https://www.jetbrains.com/help/img/idea/2019.1/show_usages.png)

**Search terms**:
 "find usages", "show usages", "peek references"

olafurpg:
Thank you for reporting! @tanishiking, @tgodzik and me looked into implementing this feature at the Spree in Lausanne today and we should be able to support this feature in the near future :)

tgodzik:
Fixed in https://github.com/scalameta/metals/pull/775

PR #775, opened 2019-06-18
Fallback to "show usages" from "Goto definition" if the symbol represents a definition itself

tanishiking:
This PR adds a feature that fallbacks to "show usages" from "goto definition" if the symbol at the given text document position represents a definition itself, which is proposed in https://github.com/scalameta/metals/issues/755.

Now, Metals returns the reference locations instead of definition location for `textDocument/definition` if the symbol at the given text document position represents the definition itself.

## VSCode example
In this recording, I used only "go to definition".

Peek 2019-06-17 18-41

<details>
<summary>Here's the pieces of JSON RPC logs while recording</summary>

### When `textDocument/definition` returns `definitonOrReferences`'s reference locations'

```
{
  "jsonrpc": "2.0",
  "id": 51,
  "method": "textDocument/definition",
  "params": {
    "textDocument": {
      "uri": "file:///home/tanishiking/dev/src/github.com/tanishiking/metals/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala"
    },
    "position": {
      "line": 1430,
      "character": 6
    }
  }
}
{
  "jsonrpc": "2.0",
  "id": 51,
  "result": [
    {
      "uri": "file:///home/tanishiking/dev/src/github.com/tanishiking/metals/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala",
      "range": {
        "start": {
          "line": 742,
          "character": 6
        },
        "end": {
          "line": 742,
          "character": 28
        }
      }
    },
    {
      "uri": "file:///home/tanishiking/dev/src/github.com/tanishiking/metals/tests/unit/src/main/scala/tests/TestingServer.scala",
      "range": {
        "start": {
          "line": 582,
          "character": 30
        },
        "end": {
          "line": 582,
          "character": 52
        }
      }
    }
  ]
}
```

### When `textDocument/definition` returns the location of its definition

```
{
  "jsonrpc": "2.0",
  "id": 49,
  "method": "textDocument/definition",
  "params": {
    "textDocument": {
      "uri": "file:///home/tanishiking/dev/src/github.com/tanishiking/metals/tests/unit/src/main/scala/tests/TestingServer.scala"
    },
    "position": {
      "line": 582,
      "character": 33
    }
  }
}
{
  "jsonrpc": "2.0",
  "id": 49,
  "result": [
    {
      "uri": "file:///home/tanishiking/dev/src/github.com/tanishiking/metals/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala",
      "range": {
        "start": {
          "line": 1430,
          "character": 6
        },
        "end": {
          "line": 1430,
          "character": 28
        }
      }
    }
  ]
}
```

</details>

## nvim + coc.nvim example

Peek 2019-06-19 01-50

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -740,7 +740,7 @@ class MetalsLanguageServer(
       position: TextDocumentPositionParams
   ): CompletableFuture[util.List[Location]] =
     CancelTokens.future { token =>
-      definitionResult(position, token).map(_.locations)
+      definitionOrReferences(position, token).map(_.locations)
     }
 
   @JsonRequest("textDocument/typeDefinition")
@@ -820,21 +820,7 @@ class MetalsLanguageServer(
       params: ReferenceParams
   ): CompletableFuture[util.List[Location]] =
     CancelTokens { _ =>
-      val timer = new Timer(time)
-      val result = referencesResult(params)
-      if (config.statistics.isReferences) {
-        if (result.symbol.isEmpty) {
-          scribe.info(s"time: found 0 references in $timer")
-        } else {
-          scribe.info(
-            s"time: found ${result.locations.length} references to symbol '${result.symbol}' in $timer"
-          )
-        }
-      }
-      if (result.symbol.nonEmpty) {
-        compileAndLookForNewReferences(params, result)
-      }
-      result.locations.asJava
+      referencesResult(params).locations.asJava
     }
 
   // Triggers a cascade compilation and tries to find new references to a given symbol.
@@ -870,7 +856,7 @@ class MetalsLanguageServer(
       newParams match {
         case None =>
         case Some(p) =>
-          val newResult = referencesResult(p)
+          val newResult = referencesProvider.references(p)
           val diff = newResult.locations.length - result.locations.length
           val isSameSymbol = newResult.symbol == result.symbol
           if (isSameSymbol && diff > 0) {
@@ -884,8 +870,23 @@ class MetalsLanguageServer(
       }
     }
   }
-  def referencesResult(params: ReferenceParams): ReferencesResult =
-    referencesProvider.references(params)
+  def referencesResult(params: ReferenceParams): ReferencesResult = {
+    val timer = new Timer(time)
+    val result = referencesProvider.references(params)
+    if (config.statistics.isReferences) {
+      if (result.symbol.isEmpty) {
+        scribe.info(s"time: found 0 references in $timer")
+      } else {
+        scribe.info(
+          s"time: found ${result.locations.length} references to symbol '${result.symbol}' in $timer"
+        )
+      }
+    }
+    if (result.symbol.nonEmpty) {
+      compileAndLookForNewReferences(params, result)
+    }
+    result
+  }
   @JsonRequest("textDocument/completion")
   def completion(params: CompletionParams): CompletableFuture[CompletionList] =
     CancelTokens.future { token =>
@@ -1417,6 +1418,65 @@ class MetalsLanguageServer(
       slowConnectToBuildServer(forceImport = false)
     } else {
       Future.successful(BuildChange.None)
+    }
+  }
+
+  /**
+   * Returns the the definition location or reference locations of a symbol
+   * at a given text document position.
+   * If the symbol represents the definition itself, this method returns
+   * the reference locations, otherwise this returns definition location.
+   * https://github.com/scalameta/metals/issues/755
+   */
+  def definitionOrReferences(
+      position: TextDocumentPositionParams,
+      token: CancelToken = EmptyCancelToken
+  ): Future[DefinitionResult] = {
+    val source = position.getTextDocument.getUri.toAbsolutePath
+    if (source.toLanguage.isScala) {
+      (for {
+        doc <- semanticdbs.textDocument(source).documentIncludingStale
+        positionOccurrence = definitionProvider.positionOccurrence(
+          source,
+          position,
+          doc
+        )
+        occ <- positionOccurrence.occurrence
+      } yield occ) match {
+        case Some(occ) =>
+          if (occ.role.isDefinition) {
+            val referecneContext = new ReferenceContext(false)
+            val refParams = new ReferenceParams(referecneContext)
+            refParams.setTextDocument(position.getTextDocument())
+            refParams.setPosition(position.getPosition())
+            val result = referencesResult(refParams)
+            if (result.locations.isEmpty) {
+              // Fallback again to the original behavior that returns
+              // the definition location itself if no reference locations found,
+              // for avoiding the confusing messages like "No definition found ..."
+              definitionResult(position, token)
+            } else {
+              Future.successful(
+                DefinitionResult(
+                  locations = result.locations.asJava,
+                  symbol = result.symbol,
+                  definition = None,
+                  semanticdb = None
+                )
+              )
+            }
+          } else {
+            definitionResult(position, token)
+          }
+        case None =>
+          warnings.noSemanticdb(source)
+          // Even if it failed to retrieve the symbol occurrence from semanticdb,
+          // try to find its definitions from presentation compiler.
+          definitionResult(position, token)
+      }
+    } else {
+      // Ignore non-scala files.
+      Future.successful(DefinitionResult.empty)
     }
   }
 
--- a/tests/unit/src/test/scala/tests/DefinitionSlowSuite.scala
+++ b/tests/unit/src/test/scala/tests/DefinitionSlowSuite.scala
@@ -74,8 +74,8 @@ object DefinitionSlowSuite extends BaseSlowSuite("definition") {
            |package a
            |import java.util.concurrent.Future/*Future.java*/ // unused
            |import scala.util.Failure/*Try.scala*/ // unused
-           |object Main/*L3*/ extends App/*App.scala*/ {
-           |  val message/*L4*/ = Message/*Message.java:1*/.message/*Message.java:2*/
+           |object Main/*MainSuite.scala:6*/ extends App/*App.scala*/ {
+           |  val message/*;L6;MainSuite.scala:6*/ = Message/*Message.java:1*/.message/*Message.java:2*/
            |  new java.io.PrintStream/*PrintStream.java*/(new java.io.ByteArrayOutputStream/*ByteArrayOutputStream.java*/())
            |  println/*Predef.scala*/(message/*L4*/)
            |}
@@ -86,7 +86,7 @@ object DefinitionSlowSuite extends BaseSlowSuite("definition") {
            |import org.scalatest.FunSuite/*FunSuite.scala*/
            |object MainSuite/*L4*/ extends FunSuite/*FunSuite.scala*/ {
            |  test/*FunSuiteLike.scala*/("a") {
-           |    val condition/*L6*/ = Main/*Main.scala:3*/.message/*Main.scala:4*/.contains/*String.java*/("Hello")
+           |    val condition/*L7*/ = Main/*Main.scala:3*/.message/*Main.scala:4*/.contains/*String.java*/("Hello")
            |    assert/*Assertions.scala*/(condition/*L6*/)
            |  }
            |}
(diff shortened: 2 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -820,21 +820,7 @@ class MetalsLanguageServer(
...
-          scribe.info(
-            s"time: found ${result.locations.length} references to symbol '${result.symbol}' in $timer"
-          )
-        }
-      }
-      if (result.symbol.nonEmpty) {
-        compileAndLookForNewReferences(params, result)
-      }
-      result.locations.asJava
+      referencesResult(params).locations.asJava
```
tanishiking: Just moved the implementation into `referencesResult` that returns `ReferencesResult` instead of `CompletableFuture[util.List[Location]]`.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1420,6 +1421,58 @@ class MetalsLanguageServer(
...
+        positionOccurrence = definitionProvider.positionOccurrence(
+          source,
+          position,
+          doc
+        )
+        occ <- positionOccurrence.occurrence
+      } yield occ) match {
+        case Some(occ) =>
+          if (occ.role.isDefinition) {
+            val referecneContext = new ReferenceContext(false)
```
gabro: typo
```suggestion
            val referenceContext = new ReferenceContext(false)
```

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1420,6 +1421,58 @@ class MetalsLanguageServer(
...
+            val refParams = new ReferenceParams(referecneContext)
```
gabro:
```suggestion
            val refParams = new ReferenceParams(referenceContext)
```

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1420,6 +1421,58 @@ class MetalsLanguageServer(
...
+            refParams.setTextDocument(position.getTextDocument())
+            refParams.setPosition(position.getPosition())
+            val result = referencesResult(refParams)
+            Future.successful(
+              DefinitionResult(
+                locations = result.locations.asJava,
+                symbol = result.symbol,
+                definition = None,
+                semanticdb = None
+              )
```
tanishiking: Returns `symbol` in addition to the `locations` because `TestingServer` requires symbol information.
Wrapping those by `DefinitionResult` for structuring the return value, but what I actually want is something like [`Definitionresult | ReferencesResult`](https://dotty.epfl.ch/docs/reference/new-types/union-types.html) haha.

tests/unit/src/test/scala/tests/DefinitionSlowSuite.scala
```diff
@@ -84,9 +84,9 @@ object DefinitionSlowSuite extends BaseSlowSuite("definition") {
            |import java.util.concurrent.Future/*Future.java*/ // unused
            |import scala.util.Failure/*Try.scala*/ // unused
            |import org.scalatest.FunSuite/*FunSuite.scala*/
-           |object MainSuite/*L4*/ extends FunSuite/*FunSuite.scala*/ {
+           |object MainSuite/*<no symbol>*/ extends FunSuite/*FunSuite.scala*/ {
```
gabro: This change means that we were previously returning the symbol itself, whereas now we don't return any symbol. I'm wondering whether we should always fall back to the symbol itself if there are no references.

I'm not feeling strongly about this, just wanted to discuss it, since it's a minor change from the previous behavior.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1420,6 +1421,58 @@ class MetalsLanguageServer(
...
+            )
+          } else {
+            definitionResult(position, token)
+          }
+        case None =>
+          warnings.noSemanticdb(source)
+          // Even if it failed to retrieve the symbol occurrence from semanticdb,
+          // try to find its definitions from presentation compiler.
+          definitionResult(position, token)
```
tanishiking: Call `definitionResult` even if it failed to retrieve the symbol information from semanticdb because `DefinitionProvider#definition` [fallback to finding definitions from a presentation compiler](https://github.com/scalameta/metals/blob/50912543e057d958e4a336a6028f74ebca71271e/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala#L62) if no definition locations found from semanticdb.

tests/unit/src/test/scala/tests/DefinitionSlowSuite.scala
tanishiking: Now, the symbol that represents a definition shows the locations of its references, instead of their locations themselves.
In this case, `MainSuite` shows `<no symbol>` because `MainSuite` represents a definition, and no one refers to `MainSuite`.

tests/unit/src/test/scala/tests/DefinitionSlowSuite.scala
```diff
@@ -123,9 +123,9 @@ object DefinitionSlowSuite extends BaseSlowSuite("definition") {
...
            |import scala.util.Failure/*Try.scala*/ // unused
            |import org.scalatest.FunSuite/*FunSuite.scala*/
-           |object MainSuite/*L6*/ extends FunSuite/*FunSuite.scala*/ {
+           |object MainSuite/*<no symbol>*/ extends FunSuite/*FunSuite.scala*/ {
            |  test/*FunSuiteLike.scala*/(testName/*<no symbol>*/) {
-           |    val condition/*L8*/ = Main/*Main.scala:5*/.message/*<no symbol>*/.contains/*String.java*/("Hello")
+           |    val condition/*L9*/ = Main/*Main.scala:5*/.message/*<no symbol>*/.contains/*String.java*/("Hello")
            |    assert/*Assertions.scala*/(condition/*L8*/)
            |  }
            |}
```
tanishiking: I tweaked this test case for fixing some that are related to the change, but this test case is still failing (even in `master` branch)...

```
=======
=> Diff
=======
---∙
+++∙
@@ -2,3 +2,3 @@
 package a
-import java.util.concurrent.Future/*<no symbol>*/ // unused
+import java.util.concurrent.Future/*Future.java*/ // unused
 import scala.util.Failure/*Try.scala*/ // unused
@@ -6,3 +6,3 @@
   val message/*L4*/ = Message/*Message.java:1*/.message/*Message.java:2*/
-  new java.io.PrintStream/*<no symbol>*/(new java.io.ByteArrayOutputStream/*<no symbol>*/())
+  new java.io.PrintStream/*PrintStream.java*/(new java.io.ByteArrayOutputStream/*ByteArrayOutputStream.java*/())
   println/*Predef.scala*/(message/*L4*/)
@@ -11,3 +11,3 @@
 package a
-import java.util.concurrent.Future/*<no symbol>*/ // unused
+import java.util.concurrent.Future/*Future.java*/ // unused
 import scala.util.Failure/*Try.scala*/ // unused
@@ -16,3 +16,3 @@
   test/*FunSuiteLike.scala*/("a") {
-    val condition/*L6*/ = Main/*Main.scala:3*/.message/*Main.scala:4*/.contains/*<no symbol>*/("Hello")
+    val condition/*L6*/ = Main/*Main.scala:3*/.message/*Main.scala:4*/.contains/*String.java*/("Hello")
     assert/*Assertions.scala*/(condition/*L6*/)
  tests.DiffAssertions$.colored(DiffAssertions.scala:103)
  tests.BaseSuite.assertNoDiff(BaseSuite.scala:102)
  tests.DefinitionSlowSuite$.$anonfun$new$5(DefinitionSlowSuite.scala:71)
  scala.util.Success.$anonfun$map$1(Try.scala:255)
  scala.util.Success.map(Try.scala:213)
  scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
  scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
  scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
  scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
  java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
  java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
  java.lang.Thread.run(Thread.java:748)
```

I'm not sure why we cannot resolve the location of java resources :thinking:

tanishiking:
Thank you for looking at this @gabro !

> In #755 I was worried about how clients dealt with N results from textDocument/definition, have you perhaps tried this with clients other than VSCode?

Not yet..., and I'm going to try it on other editors :)
I think I can test it easily on `vim + coc.nvim`, but I've never setup other text editors than vim and VSCode, so it may take some time.

tests/unit/src/test/scala/tests/DefinitionSlowSuite.scala
olafurpg: I am inclined to fall back to the original behaviour in this case and return the definition result. The reason is to avoid confusing messages like "No definition found for MainSuite"

Screenshot 2019-06-18 at 15 00 59

tests/unit/src/test/scala/tests/DefinitionSlowSuite.scala
tanishiking: Ah, this was just the problem on my local computer, `sudo apt install openjdk-8-source` fixed this :)

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tanishiking: That's right! But this time, let me leave the return value's type `Future[DefinitionResult]` because nesting `Either` inside `Future` makes things complicated and feels like to use something like `EitherT` ...

tests/unit/src/test/scala/tests/DefinitionSlowSuite.scala
```diff
@@ -134,6 +134,63 @@ object DefinitionSlowSuite extends BaseSlowSuite("definition") {
...
+      )
+      _ <- server.didOpen("a/src/main/scala/a/A.scala")
+      _ <- server.didOpen("b/src/main/scala/a/B.scala")
+      _ = assertNoDiff(client.workspaceDiagnostics, "")
+      _ = assertNoDiff(
+        server.workspaceDefinitions,
+        """|/a/src/main/scala/a/A.scala
+           |package a
+           |object A/*B.scala:3*/ {
+           |  val name/*;B.scala:3;L4*/ = "John"
```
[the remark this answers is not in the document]
tanishiking: sorted https://github.com/scalameta/metals/pull/775/files#diff-7b39581a6c581b96e5939753620ed527R604

tests/unit/src/test/scala/tests/DefinitionSlowSuite.scala
tanishiking: Sounds reasonable! done https://github.com/scalameta/metals/pull/775/files#diff-f8c05eebbf12c9c21a7d568f09b500eaR1457

tests/unit/src/test/scala/tests/DefinitionSlowSuite.scala
```diff
@@ -217,6 +274,9 @@ object DefinitionSlowSuite extends BaseSlowSuite("definition") {
            |.metals/readonly/cats/Contravariant.scala:10:45: information: <error> takes no type parameters, expected: one
            |  def compose[G[_]: Contravariant]: Functor[λ[α => F[G[α]]]] =
            |                                            ^
+           |.metals/readonly/cats/Contravariant.scala:13:15: information: not found: value Contravariant
+           |      val G = Contravariant[G]
+           |              ^^^^^^^^^^^^^
```
tanishiking: Hmm, I'm not sure why this change effects like this ...

tanishiking:
Now, I fixed things based on the review and tests passed :)

- I left a comment on the test for `annotation`
  - https://github.com/scalameta/metals/pull/775/files#r294970446
- Test manually on `nvim + coc.nvim`
  - I'm going to test on other editors in this week https://github.com/scalameta/metals/pull/775#issuecomment-503053512

tests/unit/src/test/scala/tests/DefinitionSlowSuite.scala
olafurpg: Our docs should probably mention this for ubuntu users 🤔

tests/unit/src/test/scala/tests/DefinitionSlowSuite.scala
olafurpg: Hmm, I'm not sure either. The only thing I can think of is that we're not calling this line here in some cases https://github.com/scalameta/metals/blob/4cfc3a51daeb457a8f1e26bea4196104de517380/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1501

Either way, the change looks benign so let's keep the diff.

Issue #760, opened 2019-06-13
NoSuchElementException on lsp-ui-imenu


iravid:
**Describe the bug**
Getting the following exception when calling `lsp-ui-imenu` in Emacs on `ZIO.scala` in https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/ZIO.scala:

```
SEVERE: Internal error: java.util.NoSuchElementException
java.util.concurrent.CompletionException: java.util.NoSuchElementException
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:308)
	at java.util.concurrent.CompletableFuture.uniAccept(CompletableFuture.java:647)
	at java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:632)
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:474)
	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1977)
	at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1(CancelTokens.scala:38)
	at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1$adapted(CancelTokens.scala:36)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.util.NoSuchElementException
	at java.util.ArrayList$Itr.next(ArrayList.java:862)
	at scala.collection.convert.Wrappers$JIteratorWrapper.next(Wrappers.scala:44)
	at scala.collection.IterableLike.head(IterableLike.scala:109)
	at scala.collection.IterableLike.head$(IterableLike.scala:108)
	at scala.collection.AbstractIterable.head(Iterable.scala:56)
	at scala.collection.TraversableLike.last(TraversableLike.scala:434)
	at scala.collection.TraversableLike.last$(TraversableLike.scala:433)
	at scala.collection.AbstractTraversable.last(Traversable.scala:108)
[...]

PR #776, opened 2019-06-18
Fix an issue when `val _ = ()` is the only thing in the block

tgodzik:
Fixes https://github.com/scalameta/metals/issues/760.

The issue is that `val _ =` was not added since it has no symbol and the body was actually empty.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
@@ -82,7 +82,9 @@ class DocumentSymbolProvider(trees: Trees) {
     override def apply(tree: Tree): Unit = {
       def continue(withNewOwner: Boolean = false): Unit = {
         val oldRoot = owner
-        if (withNewOwner) owner = owner.getChildren.asScala.last
+        val children = owner.getChildren.asScala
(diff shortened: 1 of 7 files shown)


metals/src/main/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
```diff
@@ -160,6 +162,7 @@ class DocumentSymbolProvider(trees: Trees) {
           )
           newOwner()
         case t: Defn.Val =>
+          scribe.warn(t.pats.toString())
```
olafurpg: Remove

tests/input/src/main/scala/example/Miscellaneous.scala
```diff
@@ -4,6 +4,10 @@ class Miscellaneous {
   // backtick identifier
   val `a b` = 42
 
+  // block with only wildcard value
+  def apply(): Unit = {
+    val _ = 42
```
olafurpg: Nice job tracking down the issue! I encountered it myself while browsing the fs2 sources.

tgodzik:
Funnily enough, this is a legit code:
```
val _ = "avb"
println(`_`)
```
and will print avb.

tgodzik:
The failing test is not related.

Issue #753, opened 2019-06-11
Auto-import doesn't work for nested objects


gabro:
**Describe the bug**
When completing a term enclosed in a top-level, Metals prefixes the term with the top-level but it doesn't import the top-level itself.

```scala
// a/src/main/a/A.scala
package a

object A {
 object Nested
}

// b/src/main/b/B.scala
object B {
  Nest@@
}
```

Obtained:

```scala
// b/src/main/b/B.scala
package b

object B {
  A.Nested
}
```

Expected:

```scala
// b/src/main/b/B.scala
package b
import a.A.Nested

object B {
  Nested
}
```

or 

```scala
// b/src/main/b/B.scala
package b
import a.A

object B {
  A.Nested
}
```

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.6.1

**Search terms**
autoimport, import, autocompletion, nested

tgodzik:
Fixed in https://github.com/scalameta/metals/pull/778

PR #778, opened 2019-06-18
Fix importing nested objects

tgodzik:
Working on https://github.com/scalameta/metals/issues/753

`TypeRef(NoPrefix, sym, Nil)` cuases us to include parent object, which i turn breaks auto importing. Checking if I didn't break anything in the process.

@gabro Did you mean something like this?

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/MetalsGlobal.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/MetalsGlobal.scala
@@ -236,12 +236,8 @@ class MetalsGlobal(
             }
           }
         case ThisType(sym) =>
-          if (sym.hasPackageFlag) {
-            if (history.tryShortenName(name)) NoPrefix
-            else new PrettyType(history.fullname(sym))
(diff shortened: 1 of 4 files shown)


tgodzik:
Seems it breaks:

```
  check(
    "implicit-class",
    """
      |object A {
      |  implicit class XtensionMethod(a: Int) {
      |    def increment = a + 1
      |  }
      |  Xtension@@
      |}""".stripMargin,
    """|XtensionMethod(a: Int): A.XtensionMethod
       |""".stripMargin
  )
```
A.XtensionMethod -> XtensionMethod

But that is expected here. 

@olafurpg @gabro Do we want to go the other way and import the parent object in such cases?
That would be the second option Gabriele suggested:
```
// b/src/main/b/B.scala
package b
import a.A

object B {
  A.Nested
}
```

gabro:
I don't have a particular preference, but I think the first option would be closer to what happens with packages.

tgodzik:
It seems that current option is also what happens in Intellij.

Issue #783, opened 2019-06-19
auto-import inserts entire path on multiline imports


vito-c:
I'm using max os, coc-nvim and running sbt in the shell to compile the source code.

I have this in my sbt plugins:
```
addSbtPlugin("org.scalameta" % "sbt-metals" % "0.6.2-SNAPSHOT")
addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "1.3.2+0-e473d887+20190614-0041")
```
my source code is `2.11.12` and using sbt version `0.13.17`

let's say I have:
```
import some.foo.path.{
  ClassB,
  ClassC,
  ClassD
}
```
then I do `val foo = new AnotherClass()`
I ended up with this:
```
import some.foo.path.{
  import some.foo.other.diffrent.path.AnotherClass
  ClassB,
  ClassC,
  ClassD
}
```
I would have expted
```
import some.foo.other.diffrent.path.AnotherClass
import some.foo.path.{
  ClassB,
  ClassC,
  ClassD
}
```

PR #785, opened 2019-06-19
Make sure we add an autoimport in the correct line

tgodzik:
Previously when using imports with braces in the last import position auto imports would be added to a wrong place.

Now in case of braces we search for the closing brace and add the import afterwards.

Fixes https://github.com/scalameta/metals/issues/783

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/AutoImports.scala
@@ -54,7 +54,7 @@ trait AutoImports { this: MetalsGlobal =>
               .getOrElse(pkg.pid)
             Some(
               new AutoImportPosition(
-                pos.source.lineToOffset(lastImport.pos.line),
+                pos.source.lineToOffset(lastImport.pos.focusEnd.line),
                 text
               )
             )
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -48,13 +59,14 @@ trait AutoImports { this: MetalsGlobal =>
           case Some(pkg)
               if pkg.symbol != rootMirror.EmptyPackage ||
                 pkg.stats.headOption.exists(_.isInstanceOf[Import]) =>
-            val lastImport = pkg.stats
-              .takeWhile(_.isInstanceOf[Import])
+            val lastImportLine = pkg.stats
+              .collect { case imp: Import => imp }
```
olafurpg: This changes the behavior in the following case
```
package a

import b.c

class A { import Foo@@ }

import d.e
```

We want to use `takeWhile` so that we stop at `import b.c` instead of `import d.e`

tgodzik:
Actually `focusEnd` was really enough - I had an issue with the test that I didn't realise and it didn't seem to work then.

PR #789, opened 2019-06-21
Use 2.13 dialect for syntax errors

### LINKED ISSUE #787: Literal types not accepted as identifiers
**Describe the bug**
When using string literals as types, as is supported by Scala 2.13, Metals provides an error from scalameta that `identifier expected but string constant found`. With integer literals the same happens, with the message `identifier expected but integer constant found`.

The code compiles fine.

**To Reproduce**
1. Add the line `val o: Option["name"] = ???` to any valid location in a 2.13 project.

**Expected behavior**
No errors in the IDE.

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.6.1+90-cde40f08-SNAPSHOT

gabro:
Fixes #787 

This PR switches to the 2.13 Scalameta dialect for parsing buffers.

It's important to notice that we always the 2.13 dialect even for 2.12 or 2.11 builds, which means that in this case:

```scala
val x: Option["lit"] = Some("lit")
```

we won't be reporting an error as you type for 2.12 / 2.11 builds.

The error is still reported after compilation (i.e. on file save) so this should be an acceptable compromise.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
+++ b/mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
@@ -38,7 +38,7 @@ class ScalaToplevelMtags(
     val input: Input.VirtualFile,
     includeInnerClasses: Boolean
 ) extends MtagsIndexer {
-  private val scanner = new LegacyScanner(input, dialects.Scala212)
+  private val scanner = new LegacyScanner(input, dialects.Scala213)
   scanner.reader.nextChar()
   def isDone: Boolean = scanner.curr.token == EOF
   def isNewline: Boolean =
--- a/tests/unit/src/test/scala/tests/SyntaxErrorSlowSuite.scala
+++ b/tests/unit/src/test/scala/tests/SyntaxErrorSlowSuite.scala
@@ -312,4 +312,22 @@ object SyntaxErrorSlowSuite extends BaseSlowSuite("syntax-error") {
     )
   )
 
+  testAsync("literal-types") {
+    cleanWorkspace()
+    for {
+      _ <- server.initialize(
+        s"""
+           |/metals.json
+           |{"a": { "scalaVersion": "2.13.0" }}
+           |/a/src/main/scala/A.scala
+           |object A {
+           |  val x: Option["literal"] = Some("literal")
+           |}
+           |""".stripMargin
+      )
+      _ <- server.didOpen("a/src/main/scala/A.scala")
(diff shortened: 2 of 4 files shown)


mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
```diff
@@ -38,7 +38,7 @@ class ScalaToplevelMtags(
     val input: Input.VirtualFile,
     includeInnerClasses: Boolean
 ) extends MtagsIndexer {
-  private val scanner = new LegacyScanner(input, dialects.Scala212)
```
gabro: This is not strictly related to the issue at hand, but it felt consistent to update it

tests/unit/src/test/scala/tests/SyntaxErrorSlowSuite.scala
```diff
@@ -312,4 +312,26 @@ object SyntaxErrorSlowSuite extends BaseSlowSuite("syntax-error") {
...
+  testAsync("literal-types") {
+    cleanWorkspace()
+    for {
+      _ <- server.initialize(
+        s"""
+           |/metals.json
+           |{"a": {}}
+           |/project/build.properties
+           |sbt.version=1.2.6
+           |/build.sbt
```
olafurpg: Let's remove all build-related files, we only need a standalone file to test this change.

Side note, tests with sbt builds should live in the `slow` project.

gabro: 👍 I've added those info for debugging, then I forgot to remove them.

gabro: ah, the standalone file reports an error (from scalac, not scalameta) probably because "rambo mode" uses 2.12. Should we change it to use 2.13 instead? This is not strictly related to this issue, again, but it might make sense

olafurpg: Can you elaborate? The standalone compiler doesn't report diagnostics.

The "rambo mode" can stay on 2.12 for now, we can change it to 2.13 when the metals server upgrades to 2.13.

gabro: Sorry, I got confused. :D Tests should be ok now.

gabro:
Ci failure on Azure seems unrelated

olafurpg:
Yeah azure has been flaky recently.

PR #792, opened 2019-06-21
Make classpath indexing linear instead of quadratic.

olafurpg:
This change unblocks JDK 11 support since previously we didn't index the
JDK 11 classpath due to bad indexing performance.

Previously, we indexed the classpath using a quadratic iteration on all
possible prefixes of classfile names. For example, given the classfile
`SimpleFileVisitor.class` we entered the following strings into a set
```scala
S
Si
Sim
Simp
Simpl
Simple
F
Fi
Fil
File
...
```

After this commit, we avoid the quadratic iteration by keeping a
streaming hash of the string prefixes so adding a new string prefix only
requires hashing a single character.

Benchmarks show that the new classpath indexer performs ~2x better on a 235mb classpath.
```diff
> bench/jmh:run -i 10 -wi 10 -f1 -t1 .*ClasspathIndexingBench
[info] Benchmark                   Mode  Cnt     Score    Error  Units
-[info] ClasspathIndexingBench.run    ss   10  1809.068 ± 61.461  ms/op # JDK 8
[info] Benchmark                   Mode  Cnt    Score    Error  Units
+[info] ClasspathIndexingBench.run    ss   10  919.237  ± 42.827  ms/op # JDK 8
+[info] ClasspathIndexingBench.run    ss   10  1316.451 ± 22.595  ms/op # JDK 11
```


olafurpg:
Although this PR should fix the last remaining issue on Java 11 that I'm aware of, I would like to keep the Java 11 support experimental for now since to officially support Java 11 we should run the full CI pipeline on Java 11 as well. I'm not keen about the idea of running the full CI pipeline on Java 11 and Java 8 so I'm leaning towards running it by default on Java 11 and then maybe a daily cron job to test on Java 8. It will require a bit more thinking than I have time for right now.

mtags/src/main/scala/scala/meta/internal/metals/ExactCharSequence.scala
```diff
@@ -15,11 +15,11 @@ case class ExactCharSequence(value: CharSequence) extends CharSequence {
...
       12345.toChar
     } else {
-      value.charAt(index + 1)
+      value.charAt(index - 1)
     }
   }
 
   override def subSequence(start: Int, end: Int): CharSequence = {
-    ExactCharSequence(value.subSequence(start + 1, end + 1))
+    ???
```
tgodzik: Should this be left at just `???` ?

olafurpg: Yes, this method should not be called.

olafurpg: I can replace this with a `throw new UnsupportedOperationException`

olafurpg:
I didn't have time yet to clean up the last commit. The implementation is still quite dirty, will try to clean up later.

olafurpg:
The implementation should be a bit cleaner now and the CI seems to only be failing on flaky tests.

I feel like the CI started getting extra flaky recently and I'm wondering if it's related to #775 :thinking: It's mostly definition suites that are failing now.

Either way, I believe this PR is ready to go!

mtags/src/main/scala/scala/meta/internal/metals/WorkspaceSymbolQuery.scala
```diff
@@ -53,20 +54,24 @@ object WorkspaceSymbolQuery {
...
 
   object AlternativeQuery {
     def apply(query: String): AlternativeQuery = {
-      AlternativeQuery(
-        query,
-        Fuzzy.bloomFilterQueryStrings(query).toArray
-      )
+      val hasher = new StringBloomFilter(0)
+      val queries = Fuzzy.bloomFilterQueryStrings(query).toArray
+      val digest = MessageDigest.getInstance("MD5")
```
tgodzik: Why do we need this digest?

Issue #791, opened 2019-06-20
Disable dot as a commit character when writing imports


jad-hamza:
**Is your feature request related to a problem? Please describe.**

I have an sbt project where some Scala files are unmanaged by sbt. In these files, I can write imports that are unknown to sbt, and in that case metals might replace them with imports in the standard library. For instance, when I write `import Environment.`, the moment when I enter the dot, the whole line is replaced with: `import java.rmi.activation.ActivationGroupDesc.CommandEnvironment.`

**Describe the solution you'd like**

I would like to have an option to disable dot as a commit character.

jad-hamza:
That was quick, thanks!

olafurpg:
Having coded quite intensively with Metals recently on the latest stable version (not since this issue was fixed), I've hit on other situations where the commit character is also problematic:

- when writing a renamed prefix like `l.Range` then I may type "l." and then insert then follow up with `import org.eclipse.{ls4p => l}`. With commit characters, the "l." completes to an unrelated item.
- when the project does not compile, the compiler symbol table quickly goes off-sync and doesn't pick up new symbols that are defined since the last compile error. In this case it's really annoying when you have  meticulously typed the full name of a new symbol and then "."  select a different identifier.

PR #793, opened 2019-06-21
Remove dot as a commit character

tgodzik:
Fixes https://github.com/scalameta/metals/issues/791

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
@@ -200,23 +200,8 @@ class CompletionProvider(
         if (r.sym.isDeprecated) {
           item.setDeprecated(true)
         }
-        // Commit character
-        r match {
-          case t: TextEditMember if t.commitCharacter.isDefined =>
-            item.setCommitCharacters(t.commitCharacter.toList.asJava)
-          case _ =>
-            if (r.sym.isMethod && !isNullary(r.sym)) {
-              // NOTE(olafur) don't use `(` as a commit character for methods because it conflicts with
-              // the `($0)` snippet behavior resulting in a redundant unit literal: `println(())`.
-              // The ideal solution would be to not use the `($0)` snippet when the commit character `(` is used,
-              // however language servers can't distinguish what commit character is used. It works as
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -211,10 +211,8 @@ class CompletionProvider(
               // The ideal solution would be to not use the `($0)` snippet when the commit character `(` is used,
               // however language servers can't distinguish what commit character is used. It works as
               // expected in IntelliJ.
-            } else if (r.sym.isError) {
-              () // No commit character
             } else {
-              item.setCommitCharacters(List(".").asJava)
+              () // No commit character
```
tgodzik: @olafurpg do we want to replace it or just altogether remove?

mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -211,10 +211,8 @@ class CompletionProvider(
               // The ideal solution would be to not use the `($0)` snippet when the commit character `(` is used,
               // however language servers can't distinguish what commit character is used. It works as
               // expected in IntelliJ.
-            } else if (r.sym.isError) {
-              () // No commit character
             } else {
-              item.setCommitCharacters(List(".").asJava)
```
olafurpg: The whole pattern match can be removed, we can add a comment instead

```scala
// NOTE: We intentionally don't set the commit character because there are valid scenarios where
// the user wants to type a dot '.' character without selecting a completion item.
```

mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
olafurpg: altogether remove the commit character pattern match

tgodzik: Alright, removed it.

olafurpg:
Ignoring flaky failure 
```
tests.DefinitionSlowSuite.definition-fallback-to-show-usages 
```
This test case is pretty new, I am wondering if the test needs to be refactored 🤔

PR #794, opened 2019-06-22
Migrate Java converters to 2.13

gabro:
This PR migrates the deprecated `scala.collection.JavaConverters` to the newer `scala.jdk.CollectionConverters`, using `scala-collection-compat` for cross-compatibility.

This removes a bunch of deprecations warnings when cross-building for Scala 2.13.

Note: we don't technically need to do this for the `metals` main module, since it doesn't cross-compile, but I figured it doesn't harm and it's one fewer thing to do when we'll migrate the codebase to 2.13. The `scala-collection-compat` dep can dropped from the `metals` module when this happens.

This is still WIP because `scala-collection-compat` 2.1.0 artifacts are not on Maven Central yet (should be there soon)

### DIFF
--- a/metals-bench/src/main/scala/bench/CompletionBench.scala
+++ b/metals-bench/src/main/scala/bench/CompletionBench.scala
@@ -1,6 +1,6 @@
 package bench
 
-import scala.collection.JavaConverters._
+import scala.meta.internal.jdk.CollectionConverters._
 import java.nio.file.Path
 import java.util.concurrent.TimeUnit
 import org.eclipse.lsp4j.CompletionList
--- a/metals-docs/src/main/scala/docs/Snapshot.scala
+++ b/metals-docs/src/main/scala/docs/Snapshot.scala
@@ -3,7 +3,7 @@ package docs
 import java.time._
 import java.time.format.DateTimeFormatter
 import org.jsoup.Jsoup
-import scala.collection.JavaConverters._
+import scala.meta.internal.jdk.CollectionConverters._
 import scala.meta.internal.metals.BuildInfo
 import scala.util.control.NonFatal
 import scala.util.Try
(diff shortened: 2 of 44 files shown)


gabro:
I've changed my mind about `scala-collection-compat` for two reasons:

- I couldn't get it to work (it might be something silly, but I don't think it's worth spending more time on it)
- Our cross-compatibility need boils down to 4 lines of code, which is easier to keep in the project, rather than adding a dependency

gabro:
mm, the cross suite seems to be failing, I'll take a look.
Meanwhile, it seems that the scala-collection-compat release was broken (see conversation on gitter https://gitter.im/scala/contributors?at=5d0fd339cedb297a81a83b5), although I stand by my decision above to not use the dep

gabro:
Tests should be ok now (there was an accidental change in the offending test). Waiting for the CI.

tgodzik:
The failing test is most likely not related - I think we can merges safely.

gabro:
Seems unrelated, but I can't restart to verify. Maybe @olafurpg can?

tgodzik:
> Seems unrelated, but I can't restart to verify. Maybe @olafurpg can?

Restarted it ;)

mtags/src/main/scala-2.12/scala/jdk/CollectionConverters.scala
```diff
@@ -0,0 +1,5 @@
+package scala.jdk
```
olafurpg: Let's define this in our own `scala.meta.internal.jdk` package to prevent conflicts with scala-compat

olafurpg: Context: mtags is published as a standalone library. I don't think anybody is using it as a library besides Metals but we should nevertheless try to be good citizen

gabro: makes sense 👍

PR #806, opened 2019-06-28
Use Scalafmt 2.0.0 when setting the version in .scalafmt.conf

### LINKED ISSUE #801: Update default Scalafmt version to 2.0.0
Scalafmt 2.0.0 can format 2.13-specific features, so we should default to it when adding the version to `.scalafmt.conf` files (or creating them from scratch).

It should be a matter of changing https://github.com/scalameta/metals/blob/4851df4499d864cdf016d19747072e4139844dae/build.sbt#L148 and adding a test.

gabro:
Closes #801

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -146,6 +146,7 @@ lazy val V = new {
   val sbtBloop = bloop
   val gradleBloop = bloop
   val scalafmt = "2.0.0"
+  val scalafmtDynamic = "2.0.0-RC4" // Not v2.0.0 because it has a regression https://github.com/scalameta/scalafmt/issues/1442
   // List of supported Scala versions in SemanticDB. Needs to be manually updated
   // for every SemanticDB upgrade.
   def supportedScalaVersions =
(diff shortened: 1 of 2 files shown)


gabro:
The Azure failure looks legit, but I won't be at a computer until next weekend.
It anyone wants to pick this up feel free, otherwise I'll get back to it when I'm back!

olafurpg:
The test failure is legit and it's caused by a recent regression in scalafmt-dynamic https://github.com/scalameta/scalafmt/issues/1442 We can work around it for now by using an older version of scalafmt-dynamic but still recommend 2.0.0

gabro:
@tgodzik the failure in https://dev.azure.com/olafurpg/Metals/_build/results?buildId=649 looks spurious, are you able to restart it?

tgodzik:
I think we can merge, but we need to figure out how to fix the Hover tests - they seem to failing pretty consistently.

PR #810, opened 2019-06-30
Don't infer build target for non-readonly files.

olafurpg:
A last minute change in the Tree View PR to fix a navigation bug when
browsing library dependenency sources from the tree view resulted in a
performance regression when editing standalone Scala files.
Previously, we tried to infer the build target using an expensive
classloading technique on every completion/hover/signatureHelp request
in standalone files. Now, the expensive technique is only used for files
in the `.metals/readonly` directory.


tgodzik:
Appveyor is being flaky - unrelated to this PR

Issue #813, opened 2019-07-01
Autocomplete with source info is broken by interleaved comments


ivanopagano:
**Describe the bug**
When cascading multiple method calls on separate lines autocompletion gives detailed information on the available methods from semantic-db (I guess).
Yet if the lines are interleaved with a comment (as in the snippet below) only "local" completion is working

```scala
(1 to 10).toList
  .map(_ + 1) //interleaved comment here <--
  .sum
```

**To Reproduce**
1. Create any code snippet like the one above
2. Try autocomplete (Ctrl+Space) with and without the inlined comment
3. Look at the different behaviour

**Expected behavior**
The autocompletion should give full detailed information available in both cases

**Screenshots**
Working normally
![working](https://user-images.githubusercontent.com/1067461/60420754-bdc47e80-9be8-11e9-8653-bfe124cd1216.png)

Not working due to the line-comment
![not-working-with-comment](https://user-images.githubusercontent.com/1067461/60420751-bd2be800-9be8-11e9-926a-82ed19d8544b.png)

**Installation:**
- Operating system: macOS Mojave 10.14.5 [15-inch mpb 2018 2.6G i7, 32GB Ram]
- Editor: Visual Studio Code Version 1.35.1 (1.35.1)
- Metals version: 0.7.0

**Additional context**

**Search terms**
inline comment, auto-completion, line comment

tgodzik:
Thanks for reporting! Confirmed it's indeed a bug, I will take a look and try to fix that as soon as possible.

tgodzik:
A workaround merged into master and a fix into scala : https://github.com/scala/scala/pull/8193

PR #814, opened 2019-07-01
Fix issue with missing completions after comment and newline.

tgodzik:
Previously, completions after a comment and newline would yield no results.

Now we remove the comment from the name that PC gives us and this causes us to get all completions.

This is connected to a presentation compiler bug that would include comment as a legal identifier. We will try to fix that inside the compiler - the implementation here might not be most efficient, but it only happens in case of comments.

Fixes https://github.com/scalameta/metals/issues/813

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
@@ -369,9 +369,28 @@ class CompletionProvider(
       val isTypeMember = kind == CompletionListKind.Type
       params.checkCanceled()
       val matchingResults = completions.matchingResults { entered => name =>
-        if (isTypeMember) CompletionFuzzy.matchesSubCharacters(entered, name)
-        else CompletionFuzzy.matches(entered, name)
+        /** NOTE(tgodzik): presentation compiler bug https://github.com/scala/scala/pull/8193
+         *  should be removed once we drop support for 2.12.8 and 2.13.0
+         *  in case we have a comment inline presentation compiler will see it as the name
+         *  CompletionIssueSuite.issue-813 for more details
+         * $div$div = '//'
+         * $u000A = '\n'
+         * $u0020 = ' '
+         * $u002E = '.'
+         */
+        val enteredFixed =
+          if (entered.startsWith("$div$div")) { // start of a comment: "//"
+            entered.toString
+              .replaceAll("\\$div\\$div.+\\$u000A", "") // entire one-line comment
+              .replaceAll("(\\$u0020|\\$u002E)", "") // any space or dot
+          } else {
+            entered.toString
+          }
+        if (isTypeMember)
+          CompletionFuzzy.matchesSubCharacters(enteredFixed, name)
+        else CompletionFuzzy.matches(enteredFixed, name)
       }
+
       val latestParentTrees = getLastVisitedParentTrees(pos)
       val completion =
         completionPosition(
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -369,9 +369,22 @@ class CompletionProvider(
       val isTypeMember = kind == CompletionListKind.Type
       params.checkCanceled()
       val matchingResults = completions.matchingResults { entered => name =>
-        if (isTypeMember) CompletionFuzzy.matchesSubCharacters(entered, name)
-        else CompletionFuzzy.matches(entered, name)
+        /** @tgodzik presentation compiler bug
```
gabro: 2 things:
- can you link to the bug?
- can you use the `NOTE(tgodizk): ` notation for consistency with the rest of the codebase?

tgodzik: No one reported the bug yet since ScalaIDE uses a different method for matching names and I don't think there is anybody else that uses PC as much. I plan to fix it though, so don't want to report it, just create a PR.

I will add the `NOTE`

tgodzik: This bug: https://github.com/scala/scala/pull/8193 uff...

tgodzik: Let us see if I manage to get this merged - then we can just fix when upgrading to `2.12.9` and `2.13.1`

mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -369,9 +369,22 @@ class CompletionProvider(
       val isTypeMember = kind == CompletionListKind.Type
       params.checkCanceled()
       val matchingResults = completions.matchingResults { entered => name =>
-        if (isTypeMember) CompletionFuzzy.matchesSubCharacters(entered, name)
-        else CompletionFuzzy.matches(entered, name)
+        /** NOTE(tgodzik): presentation compiler bug
+         *  in case we have a comment inline presentation compiler will see it as the name
+         *  CompletionIssueSuite.issue-813 for more details
+         */
+        val enteredFixed = if (entered.startsWith("$div$div")) {
```
marek1840: Could you add comments to make those "$div$div" more readable? like:
- `// start of a comment: "//"`
- `// entire one-line comment`

tgodzik:
I also increased memory on the Appveyor windows tests since there should be at least 6 gb available on those machines. Let's see if it helps, some further issues might need us to profile them more thoroughly.

gabro:
Nice!

tests/cross/src/test/scala/tests/pc/CompletionIssueSuite.scala
```diff
@@ -155,6 +155,22 @@ object CompletionIssueSuite extends BaseCompletionSuite {
     topLines = Some(2)
   )
 
+  check(
+    "issue-813-multi",
+    """|package a
+       |
+       |object Main {
+       |  Array(1, 1,10)
+       |  .map(_ + 1) /* comment breaks completions */
```
gabro: Are completions supposed to be working when the comment is truly multiline?

tgodzik: I think they should. The code compiles, so we should provide completions. Though, it would be weird to write a code like that for sure :O

tgodzik:
There seem to be more issues with OOO errors - I will try to investigate on a separate PR when I have time

tests/cross/src/test/scala/tests/pc/CompletionIssueSuite.scala
ivanopagano: I guess you're allowed to break the cascade of code calls with multiline comments and continue with a `.nextMethod(...)` on a new line, after all the comments should be stripped by the compiler sooner or later...
It's a weird practice tho, breaking the flow of calls like that, I wouldn't encourage it...

tgodzik:
Appveyor issues unrelated - we will try to work on those in the near future

Issue #819, opened 2019-07-08
Variables in docs are not replaced


Jasper-M:
https://scalameta.org/metals/docs/build-tools/maven.html contains some variables that should have been replaced by their values, such as `@@SCALAMETA_VERSION@@`, but the generated page shows `@SCALAMETA_VERSION@@` (notice the lonely `@`) instead.

PR #820, opened 2019-07-08
Fix versions in docs - wrong number of @ was used

tgodzik:
Fixes https://github.com/scalameta/metals/issues/819


tgodzik:
Checked it locally and it works. I previously wrongly assumed that we needed two @ and later didn't check.

PR #821, opened 2019-07-09
Setting h2.bindAddress to avoid exposing open ports

blast-hardcheese:
Not a great style, but the properties API isn't super great.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Tables.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Tables.scala
@@ -90,6 +90,10 @@ final class Tables(
       else ""
     val dbfile = workspace.resolve(".metals").resolve("metals")
     Files.createDirectories(dbfile.toNIO.getParent)
+    System.setProperty(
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/Tables.scala
```diff
@@ -90,6 +90,7 @@ final class Tables(
       else ""
     val dbfile = workspace.resolve(".metals").resolve("metals")
     Files.createDirectories(dbfile.toNIO.getParent)
+    System.setProperty("h2.bindAddress", System.getProperty("h2.bindAddress", "127.0.0.1"))
```
tgodzik: We need to run scalafmt on the code - can be run using `,/bin/scalafmt`

blast-hardcheese:
Not sure what's currently upset, it looks to be failing for reasons outside of just this change?

tgodzik:
The failure is not related, thanks for fixing the formatting issue!

PR #822, opened 2019-07-09
Timestamps should use currentTimeMillis 

tgodzik:
`nanoTime` is not actually time in nanos from 1970, but is used to measure difference. So all timestamps would be quite random since they can repeat themselves. We can't use them as timestamps for sure.

Also add pom.xml to be watched, since I must have forgot about it previously.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Time.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Time.scala
@@ -1,17 +1,17 @@
 package scala.meta.internal.metals
 
-import java.util.concurrent.TimeUnit
-
 /**
- * Wrapper around `System.nanoTime` to allow easier testing.
+ * Wrapper around `System.nanoTime` and `System.currentTimeMillis`
+ * to allow easier testing.
  */
 trait Time {
   def nanos(): Long
-  final def millis(): Long = TimeUnit.NANOSECONDS.toMillis(nanos())
+  def currentMillis(): Long
 }
 
 object Time {
   object system extends Time {
     def nanos(): Long = System.nanoTime()
+    def currentMillis(): Long = System.currentTimeMillis()
   }
 }
--- a/metals/src/main/scala/scala/meta/internal/builds/Digests.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/Digests.scala
@@ -15,7 +15,7 @@ final class Digests(conn: () => Connection, time: Time) {
     conn().update(
       s"insert into sbt_digest values (?, ?, ?);"
     ) { stmt =>
-      val timestamp = new Timestamp(time.millis())
+      val timestamp = new Timestamp(time.currentMillis())
       stmt.setString(1, md5Digest)
       stmt.setByte(2, status.value.toByte)
       stmt.setTimestamp(3, timestamp)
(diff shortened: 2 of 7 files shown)


metals/src/main/scala/scala/meta/internal/metals/Time.scala
```diff
@@ -1,13 +1,11 @@
...
 
-import java.util.concurrent.TimeUnit
-
 /**
  * Wrapper around `System.nanoTime` to allow easier testing.
  */
 trait Time {
   def nanos(): Long
-  final def millis(): Long = TimeUnit.NANOSECONDS.toMillis(nanos())
+  final def millis(): Long = System.currentTimeMillis()
```
gabro: Beware: this is not related to the nanos method anymore, which I think it was the intention

ivanopagano: You should check how the `Time` trait is used for testing, since that seems to be the original intent.
As @gabro points out, the idea might be that `millis()` is coupled to `nanos()` for testing purposes, and the `Time` is just a commodity util over nanotimes...

If that's the case, I'd suggest renaming `millis()` to something more explicit like `asMillis()` and document that.
Or even better, create an additional `millisTimestamp()` that doesn't relates to `nanos()` in any way but explicitly states it's getting some form of system timestamp...
On top of that you could rename `nanos` -> `monotonicNanos`, `millis` -> `monotonicMillis`?

Just wild ideas

tgodzik: Millis were always used as a timestamp and nanos for timing, which is correct. The only thing is that there is no actual relation between nanoTime and current time.

Fixed it in the tests.

marek1840: To be honest, I don't see the benefit of using nanos here.
There is one place where nanos are used directly: to calculate the [StatusBar priority](https://github.com/scalameta/metals/blob/master/metals/src/main/scala/scala/meta/internal/metals/StatusBar.scala#L174) needed for sorting the statuses. I would argue that using milliseconds here should not be a noticeable change.

Therefore I would like to propose dropping the support for nanoseconds.

marek1840: Of course there is another solution - never ever using this method for creating timestamps, but then that is an additional complexity

metals/src/main/scala/scala/meta/internal/metals/Time.scala
```diff
@@ -1,17 +1,17 @@
...
  */
 trait Time {
   def nanos(): Long
-  final def millis(): Long = TimeUnit.NANOSECONDS.toMillis(nanos())
+  def millis(): Long
 }
 
 object Time {
   object system extends Time {
     def nanos(): Long = System.nanoTime()
```
marek1840: We can also include the "absolute time base" for proper calculation of the elapsed nanoseconds:
```suggestion
    private val absoluteTimeBaseNanos: Long = {
      val now = Instant.now()
      val seconds = TimeUnit.SECONDS.toNanos(now.getEpochSecond)
      val nanos = now.getNano
      seconds + nanos
    }

    private val relativeTimeBaseNanos = System.nanoTime()

    def nanos(): Long = {
      absoluteTimeBaseNanos + elapsed
    }

    private def elapsed: Long = System.nanoTime() - relativeTimeBaseNanos
```


Note, that this does not handle the time zones at all. Probably it is safe to assume the time of the server as the valid one. At least for now

metals/src/main/scala/scala/meta/internal/metals/Time.scala
tgodzik: I think we can leave nanos for StatusBar priority. I will rename `millis` to `currentMillis` to indicate the difference.

metals/src/main/scala/scala/meta/internal/metals/Time.scala
tgodzik: I think it might be too much complexity for now.

Issue #774, opened 2019-06-17
When using string interpolation in multiline strings we don't add `s` at the start


tgodzik:
**Describe the bug**
If we have a multiline string without string interpolation and we do a completion on a value inside that string, `s` will not be added to the start of the string.

**To Reproduce**

```scala
object Main{
  val abc = "info"
  """
  This is just $ab@
  """
```

**Expected behavior**
`s` will be added at the start of the string.

**Installation:**
- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.6.1

**Additional context**
Started work on this here: https://github.com/tgodzik/metals/tree/fix-multiline-edits

**Search terms**
multiline string completion

PR #826, opened 2019-07-12
Fix multiline edits

tgodzik:
Fix multiline edits by switching from long single edit to multiple edits in the document.
Long edits spanning multiple lines are not supported by some of the LSP clients.

Fixes https://github.com/scalameta/metals/issues/774

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -742,39 +742,34 @@ trait Completions { this: MetalsGlobal =>
         interpolator: InterpolationSplice,
         text: String
     ) extends CompletionPosition {
+
       val offset = if (lit.pos.focusEnd.line == pos.line) CURSOR.length else 0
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -742,49 +742,45 @@ trait Completions { this: MetalsGlobal =>
         interpolator: InterpolationSplice,
         text: String
     ) extends CompletionPosition {
+
       val offset = if (lit.pos.focusEnd.line == pos.line) CURSOR.length else 0
-      val litpos = lit.pos.withEnd(lit.pos.end - offset)
+      val litpos = pos.withStart(pos.start - interpolator.name.size)
```
marek1840: can we use `literalPosition` or at least `literalPos` ?

tgodzik: Fixed

tgodzik:
Failing tests unrelated - TreeView one is failing everywhere - will investigate.

PR #827, opened 2019-07-12
Add automatic value completions on named arguments

tgodzik:
I talked around in the office and someone said it would be useful to have something that autofills named parameters. After tinkering a bit with possibilities I come up with:

current-auto

What does anyone think? I was pretty easy to implement, but not sure about the ideal functionality.  I am not sure about default arguments.

All suggestions are welcome.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -935,28 +935,17 @@ trait Completions { this: MetalsGlobal =>
         isParamName(member.sym.name.toString().trim())
       }
 
-      private def defaultLiteral(paramType: Type) =
-        paramType.dealiasWiden match {
-          case TypeRef(ThisType(pkg), cls, _) =>
-            val pkgName = pkg.name.toString
-            val clsName = cls.name.toString()
-            Arg.defaultValues.getOrElse(s"$pkgName.$clsName", "???")
-          case _ => "???"
-        }
-
       private def matchingTypesInScope(
           paramType: Type
       ): List[String] = {
         completions match {
           case CompletionResult.ScopeMembers(positionDelta, results, name) =>
-            val allResults = results
+            results
               .filter(
                 mem =>
                   mem.sym.tpe.toLongString == paramType.toLongString && mem.sym.isTerm
               )
               .map(_.sym.name.toString().trim())
-            if (allResults.size == 0) List(defaultLiteral(paramType))
-            else allResults
           case _ =>
             Nil
         }
@@ -966,8 +955,10 @@ trait Completions { this: MetalsGlobal =>
         val matchingType = matchingTypesInScope(param.tpe)
         if (matchingType.size == 1) {
           s":${matchingType.head}"
-        } else {
+        } else if (matchingType.size > 1) {
           s"|${matchingType.mkString(",")}|"
+        } else {
+          ":???"
         }
       }
 
@@ -1013,21 +1004,6 @@ trait Completions { this: MetalsGlobal =>
       override def contribute: List[Member] = {
         params.map(param => new NamedArgMember(param)) ::: findPossibleDefaults() ::: fillAllFields()
       }
-    }
-
-    object Arg {
-      val defaultValues = Map(
-        "scala.Int" -> "0",
-        "scala.Long" -> "0l",
-        "scala.Short" -> "0",
-        "scala.Byte" -> "0",
-        "scala.Double" -> "0.0",
-        "scala.Float" -> "0.0f",
-        "scala.Boolean" -> "false",
-        "lang.String" -> "\"\"",
-        "scala.Char" -> "''",
-        "scala.Option" -> "None"
-      )
     }
 
     /**
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -243,6 +243,12 @@ trait Completions { this: MetalsGlobal =>
         }
     }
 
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -889,16 +889,14 @@ trait Completions { this: MetalsGlobal =>
...
-          .filter(param => param.name.startsWith(prefix))
-          .toList
+      val prefix = ident.name.toString.stripSuffix(CURSOR)
+      lazy val allParams: List[Symbol] = {
+        baseParams.iterator.filterNot { param =>
+          isNamed(param.name) ||
+          param.name.containsChar('$') // exclude synthetic parameters
+        }.toList
       }
+      val params = allParams.filter(param => param.name.startsWith(prefix))
```
marek1840: if this is a `val` then is there any gain from `allParams` being `lazy` ?

tgodzik: Fixed it

tgodzik:
Fixed it a bit to show up as an additional option after normal named arguments or in empty completion:

peek2

tgodzik:
Also thinking of default values, some ideas:
1. Do a default value ("" for string, false for Boolean etc.)
2. Add `???` and a concrete value if the name is the same or is the only thing in scope

marek1840:
> 1. Do a default value ("" for string, false for Boolean etc.)
> 2. Add `???` and a concrete value if the name is the same or is the only thing in scope

I think that picking the variable "by name" from current scope is an awesome step in making the user experience much better. 

As to the defaults, I think we would benefit from using the `???` for every parameter without matching variable in scope (same name and assignable type). The most important stuff: developers would be forced to provide a value, which would prevent various bugs (forgetting to change false to true, changing the `1` to `i`, that sort of stuff ;) ).

tgodzik:
Added snippets to navigate over autofilled arguments and fixed detection of same members:

snippets

olafurpg:
Thank you @tgodzik for taking the initiative to implement this cool idea!

I am concerned about false positives. While this feature is useful in the scenarios where the guessed values are correct, the benefits of this feature may be outweighed by the frequent situation where the guessed values are incorrect and the user needs to manually edit the guessed values.

I am also concerned that users will blindly use guessed values like `isAdult = false` resulting in buggy programs.

At the price of a few additional keystrokes, the current behavior where users incrementally add one argument at a time offers a more flexible workflow where users don't have to spend mental energy wondering whether the "autofill" completion will work correctly at each call-site.

It's not clear to me what the expected behavior is in some scenarios:
- should arguments be completed for parameters with default values? I would probably say we don't autocomplete parameters that already have default values.
- should `false` or `???` be a default value for `Boolean` parameters? I would lean towards `???`
- should `None` or `???` be a default value for `Option[T]` parameters? I would towards `???`.
- should we support custom empty values for other `Option`-like data structures? I would lean towards not supporting custom empty value.

I'm curious to know what other people expect the behavior to be like.

Do you think implementing the "Use named argument for current and subsequent arguments" inspection in IntelliJ could reduce the need for this feature?

Screenshot 2019-08-25 at 12 55 13

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -206,8 +221,82 @@ object CompletionArgSuite extends BaseCompletionSuite {
...
+        |  def foo(argument : Int) : Int = argument
+        |  val number = 1
+        |  val number2 = 2
+        |  val number4 = 4
+        |  val number8 = 8
+        |  foo(ar@@)
+        |}
+        |""".stripMargin,
+    """|argument = : Int
+       |argument = number : Int
```
olafurpg: I am concerned we're trying to be too smart in this particular case. I think it's fine to complete only one thing at a time, for example there may be variables in scope that start with `arg*`.

tgodzik: We first check for values with the name, then for named arguments, named arguments with values and the autocomplete is last.

I think that's the best option and is compatible with the current behaviour. Wouldn't cause users to need to scroll more then now to use the current capabilities.

tgodzik:
> I am concerned about false positives. While this feature is useful in the scenarios where the guessed values are correct, the benefits of this feature may be outweighed by the frequent situation where the guessed values are incorrect and the user needs to manually edit the guessed values.
> 
> I am also concerned that users will blindly use guessed values like `isAdult = false` resulting in buggy programs.
> 
> At the price of a few additional keystrokes, the current behavior where users incrementally add one argument at a time offers a more flexible workflow where users don't have to spend mental energy wondering whether the "autofill" completion will work correctly at each call-site.

I think we could drop the default literals and replace them with `???` this would make it clear that Metals don't want to guess the values. In all other scenarios we don't really guess, but just use the only available scope information, so the user would still can choose if there is more than one applicable value.

> * should arguments be completed for parameters with default values? I would probably say we don't autocomplete parameters that already have default values.

I think we should most likely use the default value, but not sure if we have access to it. So we could skip all those that have default values if there is no applicable value in scope.

> * should `false` or `???` be a default value for `Boolean` parameters? I would lean towards `???`
> * should `None` or `???` be a default value for `Option[T]` parameters? I would towards `???`.
> * should we support custom empty values for other `Option`-like data structures? I would lean towards not supporting custom empty value.

We might change that all to ???, since all literal values would be guessing. I agree

> Do you think implementing the "Use named argument for current and subsequent arguments" inspection in IntelliJ could reduce the need for this feature?
> 

I don't think we need that since the option to autofil will be last, so people should know what they are doing if scrolling down the list.

marek1840:
with the `???` as a fallback value I think, this is completed.

PR #828, opened 2019-07-14
Install use-package if not already installed

Compro-Prasad:
Updated installation section for `lsp-mode` and `eglot`.
`(require 'package)` is needed else it throws error.


tgodzik:
Merged it, since this is just a change in the docs.

PR #830, opened 2019-07-16
Add ReloadDoctor to the list of all commands.

kurnevsky:


tgodzik:
Thanks for the contribution! :tada: 

The build failures are unrelated, I think we can merge. Does anyone know why it was not in the list of all commands? From what I see it was only invoke while indexing the workspace normally.

kurnevsky:
I noticed it because it's not listed in the documentation in the client commands section.

tgodzik:
> I noticed it because it's not listed in the documentation in the client commands section.

I am just wondering if it was done on purpose :smile: What documentation do you mean?

kurnevsky:
https://scalameta.org/metals/docs/editors/new-editor.html#metals-client-commands

It's automatically generated from the code.

PR #836, opened 2019-07-22
Fix Mercury blog post date by adding an unlisted redirect

gabro:
This uses a new feature of Docusaurus 1.12.0, contributed by @olafurpg 🎉, which allows marking blog posts as unlisted.

With this, we can:

- republish the faulty blog post with the correct date
- make the old post point to the new one and make it unlisted

This way old links will continue to work.

**NOTE**
During development, unlisted blog posts are still visibile, so we will still see the "redirect" blog post when doing `yarn start`.

To test this change, I did:

```
yarn build
npx http-server build
```

image

This is what the old link will point to. Notice the absence of this particular entry in the sidebar.


gabro:
@tgodzik I've installed https://github.com/apps/merge-when-green on the Metals repo to try it out. Let's see whether it works :D

tgodzik:
> @tgodzik I've installed https://github.com/apps/merge-when-green on the Metals repo to try it out. Let's see whether it works :D

We should probably turn on the setting requiring at least on approval.

gabro:
> Also I think one test is currently broken :O

I'll take care of it soon! It seems a minor node issue

gabro:
> We should probably turn on the setting requiring at least on approval.

The merge is not entirely automatic, it still requires adding the label. I was thinking of adding the label after a successful review (like I did here).

The thing we definitely need to configure is to wait for all tests (including AppVeyor and Azure)

gabro:
AppVeyor failure looks spurious. Merging.

PR #837, opened 2019-07-22
Add exact libraries that are expected and fix the test

tgodzik:
The test seems to have a random number of libraries shown, which is pretty weird.

I added the exact current libraries to see next time what changes.


gabro:
> The test seems to have a random number of libraries shown

 what you mean by random?

tgodzik:
> > The test seems to have a random number of libraries shown
> 
> what you mean by random?

It was 22 at some point and now it's 30 without any changes to actual dependencies.

Issue #841, opened 2019-07-25
Incorrect mill bloop plugin version


uosis:
**Describe the bug**
Predef script generated [here](https://github.com/scalameta/metals/blob/05d0e46a339d2bde7a046dd453f1ae4cad7c573b/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala#L11) uses hardcoded version of mill instead of using .mill-version. This results in automatic import failing to generate correct bloop files.

**To Reproduce**
Try to import mill 0.5.0 project with scala 2.13.

**Expected behavior**
Correct mill version provided in predef script.

Note that mill executable is invoked with a correct version. It's just the script that's incorrect. 

**Search terms**
mill bloop version

tgodzik:
Thanks for reporting! Fixing it in https://github.com/scalameta/metals/pull/842/files

PR #842, opened 2019-07-25
Fix issue with wrong mill version in predef script and update millw scripts

tgodzik:
Fixes https://github.com/scalameta/metals/issues/841


tgodzik:
Test failure unrelated.

PR #850, opened 2019-07-30
Do not report issues in Java only projects

tgodzik:
Related to https://github.com/scalacenter/bloop/issues/880

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
@@ -243,7 +243,6 @@ final class Doctor(
       val scala = target.info.asScalaBuildTarget
       val scalaVersion =
         scala.fold("<unknown>")(_.getScalaVersion)
-      val isSemanticdbEnabled = target.isSemanticdbEnabled
       val definition: String =
         if (ScalaVersions.isSupportedScalaVersion(scalaVersion)) {
           Icons.unicode.check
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -254,13 +253,18 @@ final class Doctor(
...
-        if (isSemanticdbEnabled) {
-          Icons.unicode.check
-        } else {
+        if (isSemanticdbNeeded) {
           Icons.unicode.alert
+        } else {
+          Icons.unicode.check
         }
       val center = "style='text-align: center'"
+      val fixRecommendation =
```
marek1840: Why is it called **fix**Recommendation?

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -269,7 +273,9 @@ final class Doctor(
           .element("td", center)(_.text(completions))
           .element("td", center)(_.text(references))
           .element("td")(
-            _.raw(recommendation(scalaVersion, isSemanticdbEnabled, target))
+            _.raw(
```
marek1840: Can it be put in a single line?

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: Recommendation about the fix

tgodzik: Changed to recommenedFix, ok?

Issue #808, opened 2019-06-29
sbt-scalafix breaks Metals Import when semanticdb versions mismatch


megri:
**Describe the bug**
The VS code task "build import" is very sensitive to semanticdb-versions. If the extension sbt-scalafix is enabled on a project, and sbt-scalafix's required version of semanticdb doesn't exactly match whatever version of semanticdb metals add to the bloop configs, the import will fail.

**To Reproduce**
1. Add sbt-scalafix to project/plugins.sbt: `addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.5")`
2. Enable scalafixSemanticdb on project: `addCompilerPlugin(scalafixSemanticdb)`
3. Run Metals Import (using version 1.5.1 of the VS Code extension in my case)
4. Watch import logs

**Expected behavior**
Importing a project where scalafix is used should work.

Since metals adds its desired version of semanticdb to all projects it imports while generating bloop configs, this may well clash with the version required by scalafixSemanticdb. I'm not sure how to work around this; perhaps metals could be more sensitive in detecting if there are other versions of semanticdb already installed and use those if they are compatible.

**Error log**
```
[warn] Multiple dependencies with the same organization/name but different versions. To avoid conflict, pick one version:
[warn]  * org.scalameta:semanticdb-scalac_2.12.8:(4.1.9, 4.2.0)
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
…
[error] Missing required plugin: semanticdb
[error] bad option: -P:semanticdb:synthetics:on
[error] bad option: -P:semanticdb:failures:warning
[error] bad option: -P:semanticdb:sourceroot:/Users/martin/dev/future-memories/sortapp-es
[error] four errors found
```

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.5.1

**Additional context**

**Search terms**
semanticdb metals build import

tgodzik:
Thanks for reporting! We are actually working on simplifying the way we add semanticdb plugin and that should fix the issue. Will let you know when we get it working.

PR #852, opened 2019-08-05
Improve SemanticDB integration using Bloop's newest features

tgodzik:
Previously, adding semanticDB plugin would require a lot of fragile configuration in Gradle and sbt.

Thanks to the change in Bloop, we now are able to define what semanticDB plugin we need when initializing and Bloop takes care of finding it and applying to scalac options.

Fixes https://github.com/scalameta/metals/issues/808


metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
```diff
@@ -21,6 +21,9 @@ case class GradleBuildTool() extends BuildTool {
        |initscript {
        |    repositories {
        |        mavenCentral()
+       |        
+       |        // TODO - Remove
```
tgodzik: Will need to update Bloop version and remove additional repos once Bloop is released

tgodzik:
I wonder if we should cleanup `project/metals.sbt` file after running the command, any opinions?

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -51,31 +51,40 @@ case class SbtBuildTool(version: String) extends BuildTool {
...
+  ): Unit = {
     val bytes = SbtBuildTool
-      .globalMetalsSbt(config.bloopSbtVersion)
+      .sbtPlugin(config.bloopSbtVersion)
       .getBytes(StandardCharsets.UTF_8)
-    val destination = plugins.resolve("metals.sbt")
-    if (destination.isFile && destination.readAllBytes.sameElements(bytes)) {
+    val projectDir = workspace.resolve("project")
+    projectDir.toFile.mkdir()
+    val metalsPluginfile = projectDir.resolve("metals.sbt")
```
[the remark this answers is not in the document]
tgodzik: The sbt targets are not yet supported, so this will only be needed once we actually add sbt support

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -51,31 +51,40 @@ case class SbtBuildTool(version: String) extends BuildTool {
...
-    val plugins =
-      if (version.startsWith("0.13")) SbtBuildTool.pluginsDirectory("0.13")
-      else SbtBuildTool.pluginsDirectory("1.0")
-    plugins.createDirectories()
+  private def writeSbtMetalsPlugin(
+      workspace: AbsolutePath,
+      config: MetalsServerConfig
+  ): Unit = {
     val bytes = SbtBuildTool
-      .globalMetalsSbt(config.bloopSbtVersion)
```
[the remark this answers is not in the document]
tgodzik: Great idea! Thanks!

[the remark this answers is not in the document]
tgodzik: Added!

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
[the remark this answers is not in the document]
tgodzik: Additionally we can add to metals.sbt `bloopExportJarClassifiers := Some(Set(\"source\"))` if we add plugin to project/project. Should be easy to do when working on sbt support

docs/build-tools/gradle.md
```diff
@@ -94,4 +94,4 @@ plugin in Bloop config.
 This is much more complex than in case of the automatic installation, so it is
 recommended to only do manual installation when having problems with the
 automatic one. You can also always try to reach us on the Metals gitter channel
```
marek1840: link would be super-nice
```suggestion
automatic one. You can also always try to reach us on the [Metals gitter channel](url)
```

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -51,31 +51,56 @@ case class SbtBuildTool(version: String) extends BuildTool {
           sbtArgs
         ).flatten
     }
-    writeSbtMetalsPlugin(config)
+    removeOldPlugin()
```
marek1840: why are we removing the old plugin? Shouldn't we just stop using it and not touch the home directory unless absolutely necessary?
If there is a reason for doing this, please add it as a comment

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -51,31 +51,56 @@ case class SbtBuildTool(version: String) extends BuildTool {
...
     val bytes = SbtBuildTool
-      .globalMetalsSbt(config.bloopSbtVersion)
+      .sbtPlugin(config.bloopSbtVersion)
       .getBytes(StandardCharsets.UTF_8)
-    val destination = plugins.resolve("metals.sbt")
-    if (destination.isFile && destination.readAllBytes.sameElements(bytes)) {
+    val projectDir = workspace.resolve("project")
+    projectDir.toFile.mkdir()
+    val metalsPluginfile = projectDir.resolve("metals.sbt")
+    if (metalsPluginfile.isFile && metalsPluginfile.readAllBytes.sameElements(
```
marek1840: could you extract the condition to a variable? alternatively you could extract it to an extension method hasContent or something like that.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -130,13 +133,33 @@ object BuildServerConnection {
     )
   }
 
+  final case class BloopExtraBuildParams(
+      semanticdbVersion: String,
+      supportedScalaVersions: Seq[String]
```
marek1840: make it a java.List and you probably won't need a `toJson` method - and it will match the other structures in the bsp/lsp protocol.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
gabro: I think the reason is this https://github.com/scalameta/metals/pull/852#discussion_r310684012

build.sbt
```diff
@@ -142,9 +142,9 @@ lazy val V = new {
   val scalameta = "4.2.0"
   val semanticdb = scalameta
   val bsp = "2.0.0-M4"
-  val bloop = "1.3.2"
-  val sbtBloop = bloop
-  val gradleBloop = bloop
+  val bloop = "1.3.2+125-2d6bf327"
+  val sbtBloop = "1.3.2"
+  val gradleBloop = "1.3.2"
```
gabro: reminder to re-instate the `bloop` shared variable before merging

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
gabro: Maybe something more telling like:

```suggestion
    removeLegacyGlobalPlugin()
```

(PS: I haven't followed closely, but what's the technical reason for switching to a local plugin?)

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -51,31 +51,56 @@ case class SbtBuildTool(version: String) extends BuildTool {
...
+    }
     val plugins =
-      if (version.startsWith("0.13")) SbtBuildTool.pluginsDirectory("0.13")
-      else SbtBuildTool.pluginsDirectory("1.0")
-    plugins.createDirectories()
+      if (version.startsWith("0.13")) pluginsDirectory("0.13")
+      else pluginsDirectory("1.0")
+
+    val metalsFile = plugins.resolve("metals.sbt")
+    if (metalsFile.exists) metalsFile.delete()
```
gabro: alternatively:

```suggestion
    Files.deleteIfExists(metalsFile.toNIO)
```

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -51,31 +51,56 @@ case class SbtBuildTool(version: String) extends BuildTool {
...
-    val destination = plugins.resolve("metals.sbt")
-    if (destination.isFile && destination.readAllBytes.sameElements(bytes)) {
+    val projectDir = workspace.resolve("project")
+    projectDir.toFile.mkdir()
+    val metalsPluginfile = projectDir.resolve("metals.sbt")
+    if (metalsPluginfile.isFile && metalsPluginfile.readAllBytes.sameElements(
+        bytes
+      )) {
       // Do nothing if the file is unchanged. If we write to the file unconditionally
       // we risk triggering sbt re-compilation of global plugins that slows down
```
gabro: not sure this is true anymore. At the very least, the message should be updated to talk about local plugins

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -51,31 +51,56 @@ case class SbtBuildTool(version: String) extends BuildTool {
...
       // build import greatly. If somebody validates it doesn't affect load times
       // then feel free to remove this guard.
       ()
     } else {
-      Files.write(destination.toNIO, bytes)
+      Files.write(metalsPluginfile.toNIO, bytes)
+    }
+    val gitignore = workspace.resolve(".gitignore")
```
gabro: maybe extract the gitignore-related stuff into a method?

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -101,53 +126,19 @@ object SbtBuildTool {
...
+    val resolvers = if (bloopSbtVersion.contains("+")) {
+      """resolvers += Resolver.bintrayRepo("scalacenter", "releases")"""
+    } else {
+      ""
+    }
     s"""|// DO NOT EDIT! This file is auto-generated.
-        |// By default, this file does not do anything.
-        |// If the environment variable METALS_ENABLED has the value 'true',
-        |// then this file enables sbt-metals and sbt-bloop.
+        |// This file enables sbt-bloop to created bloop config files.
```
gabro:
```suggestion
        |// This file enables sbt-bloop to create bloop config files.
```

docs/build-tools/gradle.md
```diff
@@ -93,5 +93,5 @@ plugin in Bloop config.
 
 This is much more complex than in case of the automatic installation, so it is
 recommended to only do manual installation when having problems with the
-automatic one. You can also always try to reach us on the Metals gitter channel
+automatic one. You can also always try to reach us on the [Metals gitter channel](url)
```
gabro: typo? This should be `https://gitter.im/scalameta/metals`

tgodzik: Right, I just commited Marek's changes without looking exactly.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
tgodzik: I will add a comment.

tgodzik: It doesn't add the plugin for every workspace, which is useful to reduce some amount of overhead and is not encouraged by the sbt docs. Additionally, I don't think we would be able to work with sbt support, because global plugins don't add plugins to meta builds.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
tgodzik: Fixed

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
tgodzik: added it to a method and also added some tests for it

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
tgodzik: removed the whole comment, it's pretty obvious why we don't write if it already exists.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
tgodzik: Good idea, changed.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
tgodzik: Fixed the name :)

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
tgodzik: Done, thanks for the idea!

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
tgodzik: Extracted to a variable.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
tgodzik: Added

tgodzik:
Had to tidy up a bit and squashed the commits, but I applied all the suggestions and added some more tests for `.gitignore` modification

metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
```diff
@@ -16,72 +16,28 @@ case class GradleBuildTool() extends BuildTool {
       .map(s => scala.meta.Lit.String(s).syntax)
       .mkString("[", ",", "]")
 
+  private def additionalRepos =
+    if (BuildInfo.gradleBloopVersion.contains("+"))
```
marek1840: can you extract the condition into the variable?

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -101,53 +127,19 @@ object SbtBuildTool {
...
-           |  } else {
-           |    List()
-           |  }
-           |}
-           |""".stripMargin
-      } else {
-        ""
-      }
+  private def sbtPlugin(bloopSbtVersion: String): String = {
+    val resolvers = if (bloopSbtVersion.contains("+")) {
```
gabro:
```suggestion
    val isSnapshot = bloopSbtVersion.contains("+")
    val resolvers = if (isSnapshot) {
```

tgodzik: Fixed

metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
tgodzik: Fixed

tgodzik:
Fixed some last things, but we are waiting for the newest Bloop release - @jvican wants to make sure that is all stable. When it's released I will replace the version to use the same bloop version everwhere.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -51,31 +51,57 @@ case class SbtBuildTool(version: String) extends BuildTool {
...
+      if (version.startsWith("0.13")) pluginsDirectory("0.13")
+      else pluginsDirectory("1.0")
+
+    val metalsFile = plugins.resolve("metals.sbt")
+    Files.deleteIfExists(metalsFile.toNIO)
+  }
+
+  private def gitignoreMetals(workspace: AbsolutePath) = {
+    val gitignore = workspace.resolve(".gitignore")
+    val gitIgnoreContents = "project/metals.sbt"
```
olafurpg: What do you think about adding the following entries as well?
```
.metals/
.bloop/
```
These entries should ideally be in the workspace .gitignore instead of the global .gitignore because VS Code only uses the workspace gitignore in "Open file".

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -101,53 +127,20 @@ object SbtBuildTool {
...
-           |  } else {
-           |    List()
-           |  }
-           |}
-           |""".stripMargin
-      } else {
-        ""
-      }
+  private def sbtPlugin(bloopSbtVersion: String): String = {
+    val isSnapshotVersion = bloopSbtVersion.contains("+")
```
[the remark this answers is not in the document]
tgodzik: I don't think bloop ever releases with `version-SNAPSHOT`, but I can add it here, just don't think it will change much.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
tgodzik: I was thinking about it, but decided not to do in this PR, but I can for sure do it here.

olafurpg:
Bloop v1.3.3 is out now unblocking this PR 🎉

olafurpg:
🎉

gabro:
😍

Baccata:
🥇

PR #853, opened 2019-08-05
Report an issue only if SemanticDB file doesn't really exist.

tgodzik:
Previously we would report that SemanticDB file does not exist in case when symbol was correctly not found (Ctrl on an empty place), now we check if SemanticDB really exists.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1548,8 +1548,10 @@ class MetalsLanguageServer(
   ): Future[DefinitionResult] = {
     val source = position.getTextDocument.getUri.toAbsolutePath
     if (source.toLanguage.isScala) {
+      val semanticDBDoc =
+        semanticdbs.textDocument(source).documentIncludingStale
       (for {
-        doc <- semanticdbs.textDocument(source).documentIncludingStale
+        doc <- semanticDBDoc
         positionOccurrence = definitionProvider.positionOccurrence(
           source,
           position,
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1583,7 +1585,9 @@ class MetalsLanguageServer(
             definitionResult(position, token)
           }
         case None =>
-          warnings.noSemanticdb(source)
+          if (semanticDBDoc.isEmpty) {
```
marek1840: or we could do:
```scala
val semanticDBDoc = semanticdbs.textDocument(source).documentIncludingStale
if(semanticDBDoc.isEmpty){
  warnings.noSemanticdb(source)
} else {
  val occurrence = for { ...} yield occ
  occurrence match {
    case None =>
      // ignore
    case Some(occ) =>
      (...)
  }
}
```
would probably make the logic simpler (no need to remember about semanticDBDoc being emoty or not when working on occurrences).

or even something like that:

```scala

val semanticDBDoc = semanticdbs.textDocument(source).documentIncludingStale 
semanticDBDoc match {
  case None => warnings.noSemanticdb(source)
  case Some(doc) =>
    val occurrence = definitionProvider.positionOccurrence(source, position, doc).occurrence 
    occurrence.foreach { occ =>
        (...)
    }
```

tgodzik: I'd rather not change too much now.

tgodzik:
Failures do not look related. merging.

Issue #770, opened 2019-06-16
Handle non-directory sources


olafurpg:
**Describe the bug**

Metals reports a "no build target" warning when working in a build with targets that contain individual file sources.

Concretely, the BSP request `buildTarget/sources` can return both file and directory sources (https://github.com/scalacenter/bsp/blob/master/docs/bsp.md#build-target-sources-request) and it looks like Metals only supports directory sources (not file sources).

**Expected behavior**

Metals should map the individual file to the build target it contains.

**Additional context**

This problem came up while @jvican and @cosmicexplorer where working on adding a Pants integration for Bloop.

At a quick glance, I don't think there are any fundamental assumptions in the Metals codebase that require sources to be directories. Most of the changes to fix this issue will be in the `BuildTargets` class. Another part of the Metals codebase that needs to be updated is this condition here https://github.com/scalameta/metals/blob/4acd4659473321b897ce4059215656dbce4353d5/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1172

PR #855, opened 2019-08-06
Add support for non-directory sources

tgodzik:
Fixes https://github.com/scalameta/metals/issues/770

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
@@ -49,25 +49,25 @@ final class FileWatcher(
   }
 
   def restart(): Unit = {
-    val directoriesToWatch = new util.ArrayList[Path]()
+    val sourcesItemsToWatch = new util.ArrayList[Path]()
     val createdSourceDirectories = new util.ArrayList[AbsolutePath]()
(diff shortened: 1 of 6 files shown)


metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
```diff
@@ -49,25 +49,26 @@ final class FileWatcher(
...
-      if (!dir.isDirectory) {
-        dir.createDirectories()
-        if (isSourceDirectory) createdSourceDirectories.add(dir)
+    def watch(path: AbsolutePath, isSourceItem: Boolean): Unit = {
+      if (!path.isDirectory && !path.isFile) {
+        path.createDirectories()
+        if (isSourceItem) createdSourceDirectories.add(path)
       }
-      directoriesToWatch.add(dir.toNIO)
+      if (path.isFile) sourcesItemsToWatch.add(path.toNIO.getParent())
```
marek1840: let's hope this won't become a bottleneck

tgodzik:
Found an issue in the code - need to rework tests a bit.

metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
tgodzik: Added a separate file watcher that watches parent dir and filters on the known files.

PR #857, opened 2019-08-07
Add support for non-directory sources

### LINKED ISSUE #770: Handle non-directory sources

tgodzik:
Fixes #770

Reopened the PR here. I fixed the suggestions:
- added a file watcher which only reports on the specific file
- if a source file item does not exist we then create the parent dir and watch it then delete it afterwards as with the source directories
- renamed isSourceItem to isSource

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -286,7 +286,7 @@ final class BuildTargets() {
       .getOrElse(Iterable.empty)
   }
 
-  def inverseSourceDirectory(source: AbsolutePath): Option[AbsolutePath] =
+  def inverseSourceItem(source: AbsolutePath): Option[AbsolutePath] =
     sourceItems.find(dir => source.toNIO.startsWith(dir.toNIO))
 
   def isInverseDependency(
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -91,11 +91,11 @@ final class BuildTargets() {
   }
 
   def addSourceItem(
-      sourceItemPath: AbsolutePath,
+      sourceItem: AbsolutePath,
       buildTarget: BuildTargetIdentifier
   ): Unit = {
     val queue = sourceItemsToBuildTarget.getOrElseUpdate(
-      sourceItemPath,
+      sourceItem,
       new ConcurrentLinkedQueue()
     )
     queue.add(buildTarget)
@@ -275,19 +275,19 @@ final class BuildTargets() {
   }
 
   def sourceBuildTargets(
-      source: AbsolutePath
+      sourceItem: AbsolutePath
   ): Iterable[BuildTargetIdentifier] = {
     sourceItemsToBuildTarget
       .collectFirst {
-        case (sourceItem, buildTargets)
-            if source.toNIO.startsWith(sourceItem.toNIO) =>
+        case (source, buildTargets)
+            if sourceItem.toNIO.startsWith(source.toNIO) =>
           buildTargets.asScala
       }
       .getOrElse(Iterable.empty)
   }
 
   def inverseSourceItem(source: AbsolutePath): Option[AbsolutePath] =
-    sourceItems.find(dir => source.toNIO.startsWith(dir.toNIO))
+    sourceItems.find(item => source.toNIO.startsWith(item.toNIO))
 
   def isInverseDependency(
       query: BuildTargetIdentifier,
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -31,7 +31,7 @@ final class BuildTargets() {
     workspace = newWorkspace
   }
   private var tables: Option[Tables] = None
-  private val sourceDirectoriesToBuildTarget =
+  private val sourceItemsToBuildTarget =
     TrieMap.empty[AbsolutePath, ConcurrentLinkedQueue[BuildTargetIdentifier]]
   private val buildTargetInfo =
     TrieMap.empty[BuildTargetIdentifier, BuildTarget]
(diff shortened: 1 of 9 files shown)


metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -31,7 +31,7 @@ final class BuildTargets() {
     workspace = newWorkspace
   }
   private var tables: Option[Tables] = None
-  private val sourceDirectoriesToBuildTarget =
+  private val sourceItemsToBuildTarget =
```
marek1840: I was thinking about renaming the `sourceItems` to just `sources` whenever we find it ;)

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1277,15 +1277,20 @@ class MetalsLanguageServer(
 
   private def indexWorkspaceSources(): Unit = {
     for {
-      (sourceDirectory, targets) <- buildTargets.sourceDirectoriesToBuildTargets
-      if sourceDirectory.isDirectory
-      source <- ListFiles(sourceDirectory)
+      (sourceItem, targets) <- buildTargets.sourceItemsToBuildTargets
+      sources = if (sourceItem.isDirectory) ListFiles(sourceItem)
```
marek1840: can we make `ListFiles(source)` return a one element list if the source is a file?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1277,15 +1277,20 @@ class MetalsLanguageServer(
...
+      else List(sourceItem)
+      source <- sources
       if source.isScalaOrJava
     } {
       targets.asScala.foreach { target =>
         buildTargets.linkSourceFile(target, source)
       }
-      indexSourceFile(source, Some(sourceDirectory))
+      val sourceDirectory =
```
marek1840: What do you think about:

```scala
if(sourceItem.isDirectory) {
  indexSourceFile(source, Some(sourceItem))
} else if (source.exists) { //here should be a comment what does it mean when a source exists but a sourceItem is not a directory
  indexSourceFile(source, None)
}
```

same amount of lines but at the same time more expressive

metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
```diff
@@ -37,62 +37,103 @@ final class FileWatcher(
...
-  class Listener extends DirectoryChangeListener {
+  class DirectoryListener extends DirectoryChangeListener {
     override def onEvent(event: DirectoryChangeEvent): Unit = {
       if (!Files.isDirectory(event.path())) {
         didChangeWatchedFiles(event)
       }
     }
   }
 
+  class FileListener(files: Set[Path]) extends DirectoryChangeListener {
```
marek1840:
```suggestion
  class FileListener(watched: Set[Path]) extends DirectoryChangeListener {
```
this way, we can have `(...) && watched(event.path())` instead of `(...) && files(event.path())`

metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
```diff
@@ -37,62 +37,103 @@ final class FileWatcher(
...
-    activeWatcher.foreach(_.close())
+    directoryExecutor.shutdown()
+    fileExecutor.shutdown()
+    activeDirectoryWatcher.foreach(_.close())
+    activeFileWatcher.foreach(_.close())
   }
 
   def restart(): Unit = {
-    val directoriesToWatch = new util.ArrayList[Path]()
+    val sourcesDirectoriesToWatch = new util.ArrayList[Path]()
```
marek1840: sourceDirectories
```suggestion
    val sourceDirectoriesToWatch = new util.ArrayList[Path]()
```

metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
```diff
@@ -37,62 +37,103 @@ final class FileWatcher(
...
+    val sourcesFilesToWatch = new util.ArrayList[Path]()
```
marek1840:
```suggestion
    val sourceFilesToWatch = new util.ArrayList[Path]()
```

metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
```diff
@@ -37,62 +37,103 @@ final class FileWatcher(
...
     val createdSourceDirectories = new util.ArrayList[AbsolutePath]()
-    def watch(dir: AbsolutePath, isSourceDirectory: Boolean): Unit = {
-      if (!dir.isDirectory) {
-        dir.createDirectories()
-        if (isSourceDirectory) createdSourceDirectories.add(dir)
+    def watch(path: AbsolutePath, isSource: Boolean): Unit = {
+      if (!path.isDirectory && !path.isFile) {
```
marek1840: please, add a comment that we create those paths just to get a notification they don't exist

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: Done

metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
tgodzik: added

metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
tgodzik: changed

marek1840:
I have a stupid idea.
What about:
```scala
final class ExternalSourceWatcher(emit: FileChangedEvent => Unit) {
  private val files = mutable.Map.empty[Path, State]
  private val watcher = new Thread(() => {
    Thread.sleep(250)
    files.foreach { case (file, previousState) =>
      val currentState = ???
      if(currentState != previousState){
        val event = ???
        emit(event)
      }
    }
  })
}
```

that way we can avoid potentially a lot of traffic when listening for a one file in a big directory.

tgodzik:
We would lose a lot of time to check on files even if nothing was happening, it would add more problems than just having the events from the file system I think. And using files as sources is pretty rare currently. If it proves to be a problem we can experiment on some real life projects.

tgodzik:
Currently it's failing on Windows (a couple of tests have GC issues), which we need to take a look at and the other is TreeViewSlowSuite which seems to change dependencies despite not changing them :/ Some versions of the dependencies change.

PR #863, opened 2019-08-13
Update scalafmt to 2.0.1

tanishiking:
I just released scalafmt v2.0.1 that contains the [change](https://github.com/scalameta/scalafmt/pull/1467) which fixes https://github.com/scalameta/scalafmt/issues/1442
Some tests are failing, but they seem to have nothing to do with this change.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -145,8 +145,7 @@ lazy val V = new {
   val bloop = "1.3.2"
   val sbtBloop = bloop
   val gradleBloop = bloop
-  val scalafmt = "2.0.0"
-  val scalafmtDynamic = "2.0.0-RC4" // Not v2.0.0 because it has a regression https://github.com/scalameta/scalafmt/issues/1442
+  val scalafmt = "2.0.1"
   // List of supported Scala versions in SemanticDB. Needs to be manually updated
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -145,8 +145,8 @@ lazy val V = new {
   val bloop = "1.3.2"
   val sbtBloop = bloop
   val gradleBloop = bloop
-  val scalafmt = "2.0.0"
-  val scalafmtDynamic = "2.0.0-RC4" // Not v2.0.0 because it has a regression https://github.com/scalameta/scalafmt/issues/1442
+  val scalafmt = "2.0.1"
+  val scalafmtDynamic = "2.0.1"
```
gabro: let's remove this variable and only use scalafmt (it was like this before the regression)

tanishiking: fixed https://github.com/scalameta/metals/pull/863/commits/1f7379a8db70cf8568303bf110b4a899d587a55b 👍

gabro:
The two failures seem unrelated. @tgodzik the Azure one seems deterministic: it's resolving a slightly different version of the artifacts in the TVP tests (a new version of the deps came out, it looks like)

PR #864, opened 2019-08-13
Fix TreeViewSlowSuite tests

gabro:
This should fix the current failures, but it's definitely not permanent, given the nature of the fix.

Any idea on how to make these more stable, @tgodzik / @olafurpg?


tgodzik:
> This should fix the current failures, but it's definitely not permanent, given the nature of the fix.
> 
> Any idea on how to make these more stable, @tgodzik / @olafurpg?

It seems transitive dependencies change and I am not entirely sure if this test can be stable at all.

We could check for a subset of dependencies and ignore the ones that are changing, but that doesn't look perfect either.

Issue #859, opened 2019-06-21
Insert pipe character | when editing multiline strings with stringMargin


olafurpg:
**Is your feature request related to a problem? Please describe.**

When editing multiline strings with `.stripMargin` it's currently necessary to manually type the `|` character and indent it.

**Describe the solution you'd like**

It would be nice if Metals inserted the `|` character at the correct indentation as we type, the same way it works in IntelliJ.

![2019-06-21 20 15 30](https://user-images.githubusercontent.com/1408093/59943043-97bd1280-9461-11e9-8d22-3ddd5089c046.gif)

**Describe alternatives you've considered**

In the PR https://github.com/scalameta/metals-vscode/pull/121, we added basic support for similar functionality in the VS Code extension. This solution has several limitations however:

- it triggers even when editing outside multiline strings

![2019-06-21 20 05 30](https://user-images.githubusercontent.com/1408093/59942646-8b848580-9460-11e9-94e8-4d14f709f7f4.gif)

- it doesn't auto-insert `|` for copy-pasted multiline code

![2019-06-21 20 11 00](https://user-images.githubusercontent.com/1408093/59942717-b5d64300-9460-11e9-87af-0eb3f59f1f59.gif)

- it only works for VS Code

**Additional context**

In LSP there is a `textDocument/formattingOnType` request that allows us to produce the same behaviour that allows us to implement this analysis on top of tokens and make it work for all editors.

marek1840:
partially fixed by https://github.com/scalameta/metals/pull/865

What is still missing is inserting "|" when copy-pasting inside a multiline strings

PR #865, opened 2019-08-13
Insert pipe character when editing multiline strings

Karola766:
Partially solves #859. Doesn't work for copypasted strings yet, though.

The behaviour isn't identical to that demonstrated in gifs under issue #859. For now, before making a TextEdit, Provider checks whether there was a pipe in the last line, even, if it's the first line of the string. It doesn't yet add .stripMargin after the end of the string, either. So we might want to expand on that.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -848,7 +848,7 @@ class MetalsLanguageServer(
         token
       )
     }
-
+  /*in order to use onTypeFormatting in vscode, you'll have to set editor.formatOnType = true in settings*/
   @JsonRequest("textDocument/onTypeFormatting")
   def onTypeFormatting(
       params: DocumentOnTypeFormattingParams
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -153,6 +153,7 @@ class MetalsLanguageServer(
   private var definitionProvider: DefinitionProvider = _
   private var documentHighlightProvider: DocumentHighlightProvider = _
   private var formattingProvider: FormattingProvider = _
+  private var onTypeFormattingProvider: OnTypeFormattingProvider = _
   private var initializeParams: Option[InitializeParams] = None
   private var referencesProvider: ReferenceProvider = _
   private var workspaceSymbols: WorkspaceSymbolProvider = _
@@ -317,6 +318,10 @@ class MetalsLanguageServer(
           Nil
       }
     )
+    onTypeFormattingProvider = new OnTypeFormattingProvider(
+      semanticdbs,
+      buffers
+    )
     referencesProvider = new ReferenceProvider(
       workspace,
       semanticdbs,
@@ -409,6 +414,9 @@ class MetalsLanguageServer(
       capabilities.setHoverProvider(true)
       capabilities.setReferencesProvider(true)
       capabilities.setDocumentHighlightProvider(true)
+      capabilities.setDocumentOnTypeFormattingProvider(
+        new DocumentOnTypeFormattingOptions("\n")
+      )
       capabilities.setSignatureHelpProvider(
         new SignatureHelpOptions(List("(", "[").asJava)
       )
@@ -838,6 +846,16 @@ class MetalsLanguageServer(
       formattingProvider.format(
         params.getTextDocument.getUri.toAbsolutePath,
         token
+      )
+    }
+
+  @JsonRequest("textDocument/onTypeFormatting")
+  def onTypeFormatting(
+      params: DocumentOnTypeFormattingParams
+  ): CompletableFuture[util.List[TextEdit]] =
+    CancelTokens.future { _ =>
+      onTypeFormattingProvider.format(
+        params
       )
     }
 
--- a/metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
@@ -1,0 +1,75 @@
+package scala.meta.internal.metals
+
+import org.eclipse.lsp4j.{DocumentOnTypeFormattingParams, Range, TextEdit}
+
+import scala.meta.internal.mtags.Semanticdbs
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.meta.inputs.Input
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.tokens.Token
+import scala.meta.tokens.Token.Constant
+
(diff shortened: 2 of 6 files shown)


metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
```diff
@@ -0,0 +1,71 @@
...
+    val bit = toInput.text.substring(0, pos.start)
+    val lastNewline = bit.lastIndexOf("\n")
+    val lastPipe = bit.lastIndexOf("|")
+    val length = toInput.text
+      .substring(lastNewline, lastPipe)
+      .length - 1
+    space * length
+  }
+
+  private def longString(text: String, token: Token): Boolean = {
```
marek1840: those private defs should be below the one method uses them (or even defined within this method)

marek1840: `isLongString` sound a bit nicer nicer when read inside an if: 
`if((...) and isLongString(foo)`

marek1840: maybe `isMultilineString` ? one-line string can also be very long ;)

metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
```diff
@@ -0,0 +1,71 @@
...
+    text.substring(token.start, token.start + 3).equals(quot)
+  }
+
+  private def inToken(pos: meta.Position, token: Token): Boolean = {
+    pos.start >= token.start && pos.end <= token.end
+  }
+
+  private def pipeInScope(pos: meta.Position, text: String): Boolean = {
+    val lastNew =
```
marek1840: how about `lastNewline` ?

metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
```diff
@@ -0,0 +1,71 @@
...
+      params: DocumentOnTypeFormattingParams
+  ): Future[Option[TextEdit]] = {
+    val source = params.getTextDocument.getUri.toAbsolutePath
+    if (source.exists) {
+      val range = new Range(params.getPosition, params.getPosition)
+      val pos = params.getPosition.toMeta(source.toInput)
+      val sourceText = source.toInput.text
+      val tokens = source.toInput.tokenize.toOption
+      var edit: Option[TextEdit] = None
+      tokens.get.foreach {
```
marek1840: why `get` ? what if there are no tokens? maybe `getOrElse(Nil)` would be better?
or something like this?
```scala
if(pipeInScope(pos, sourceText){
  val tokenized = source.toInput.tokenize.toOption
  tokenized.flatMap{ tokens => 
    tokens.collectFirst{ 
      case token: Constant.String if inToken(pos, token) && longString(sourceText, token) => 
        new TextEdit(range, indent(source.toInput, pos) + "|")
      }
  }
} else {
  None
}
```

metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
```diff
@@ -0,0 +1,71 @@
...
+          if (inToken(pos, token) && longString(sourceText, token) && pipeInScope(
+              pos,
+              sourceText
+            ))
+            edit = Some(new TextEdit(range, indent(source.toInput, pos) + "|"))
+        case _ => None
+      }
+      Future.successful(edit)
+    } else {
+      Future.failed(new Exception("There is no source"))
```
marek1840: Why failure instead of a `None` ?

tgodzik:
Checking this in VSCode - it seems doesn't yet work. We need to change some things.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -841,6 +849,16 @@ class MetalsLanguageServer(
       )
     }
 
+  @JsonRequest("textDocument/onTypeFormatting")
```
marek1840: does it work in vscode?  The [specification](https://microsoft.github.io/language-server-protocol/specification#textDocument_onTypeFormatting) expects a `List`, not an `Option[TextEdit]`.

metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
```diff
@@ -0,0 +1,73 @@
...
+    text
+      .substring(lastNewline + 1, pos.start)
+      .contains("|")
+  }
+
+  def format(
+      params: DocumentOnTypeFormattingParams
+  ): Future[Option[TextEdit]] = {
+    val source = params.getTextDocument.getUri.toAbsolutePath
+    var edit: Option[TextEdit] = None
```
marek1840: we can remove this `var` if we add some else blocks:
```scala
else Future.successful(None)
```

metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
```diff
@@ -0,0 +1,73 @@
...
+        Input.VirtualFile(source.toString(), sourceText)
+      )
+      if (pipeInScope(pos, sourceText)) {
+        val tokens =
+          Input.VirtualFile(source.toString(), sourceText).tokenize.toOption
+        tokens.flatMap { tokens =>
+          tokens.collectFirst {
+            case token: Constant.String
+                if inToken(pos, token) && isLongString(sourceText, token) =>
+              edit = Some(new TextEdit(range, indent(sourceText, pos) + "|"))
```
marek1840: Let's just return the `new TextEdit(...)` here. as a result, the `tokens.flatMap` on line 62 will be returning this edit.
```scala
if (pipeInScope(pos, sourceText)) {
  val tokens = Input.VirtualFile(...)
  tokens.flatMap { tokens => tokens.collectFirst{ case (...) if (...) => new TextEdit(...) } }
} else Future.successfull(None)
```
or we could do:
```scala
tokens match {
  case None => 
    Future.successful(None)
  case tokens => 
    tokens.collectFirst{ case (...) if (...) => new TextEdit(...) }
}
```

whatever you find more readable

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
Karola766: it doesn't, and you're right, I haven't noticed. But I've changed my ontypeformatting product to a list and the server still doesn't seem to be responding properly, so it's probably something deeper, too.

metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
Karola766: ? how can we return Future.succesful(None) when we want to return Future[TextEdit]? wouldn't a None be more fitting for Option[TextEdit]?

PR #867, opened 2019-08-15
Scala 2.12.9 support

### LINKED ISSUE #856: Add support for Scala 2.12.9
After upgrading to scala 2.12.9 and re-importing the build Navigation is disabled and Metals Doctor reports:

```
Upgrade to Scala 2.12.8 and run 'Build import' to enable code navigation.
```

gabro:
Closes #856 

This PR does three things:

- Update Scalameta to v4.2.1 consequently adding support for Scala 2.12.9
- ~Use Scala 2.12.9 on Metals itself~ reverted, we don't care about it for now
- Update slow tests to use 2.12.9, except for Mill, see below

The last two points are not strictly necessary, but since we're at it...

Hopefully, the CI gods will assist us 🙏


build.sbt
```diff
@@ -149,7 +149,7 @@ lazy val V = new {
   // List of supported Scala versions in SemanticDB. Needs to be manually updated
   // for every SemanticDB upgrade.
   def supportedScalaVersions =
-    Seq(scala213, scala212) ++ deprecatedScalaVersions
+    Seq(scala213, scala212, "2.12.8") ++ deprecatedScalaVersions
```
gabro: At some point we will deprecate 2.12.8, maybe when 2.12.10 is out?

tgodzik: I think that's a good idea. 2.12.10 should be released soon since they found some issues there.

tgodzik:
> Note: Unresolved dependencies path:
> INFO  [warn] 		org.scalameta:semanticdb-scalac_2.12.7:4.2.1 (MetalsPlugin.scala#L61)
> INFO  [warn] 		  +- default:cancel_2.12:1.1

We should most likely remove support for 2.12.7 and change the tests using that.

tgodzik:
Looks like a similar issue to the tests failing on 2.13. However I did confirm that the completions work without any issue when running on the generated workspaces in `target` directory.

better-monadic-for  2.12.8
```scala
package a {
  object Main extends scala.AnyRef {
    def <init>() = {
      super.<init>();
      ()
    };
    Option("").foreach(((x$1: String) => x$1 match {
      case (x @ _) => {
        implicit <synthetic> <artifact> val x$implicit$1: String = x;
        implicitly[String].toCharArr_CURSOR_
      }
    }))
  }
}
```
better-monadic-for  2.12.9
```scala
package a {
  object Main extends scala.AnyRef {
    def <init>() = {
      super.<init>();
      ()
    };
    Option("").withFilter(((check$ifrefutable$1: String) => check$ifrefutable$1: @scala.unchecked match {
  case <implicit0: error>((<x: error> @ (_: String))) => true
  case _ => false
})).<foreach: error>(((<x$1: error>: <error>) => <x$1: error>: @scala.unchecked match {
      case <implicit0: error>((<x: error> @ (_: String))) => implicitly[String].<toCharArr_CURSOR_: error>
    }))
  }
}
```

tgodzik:
I think the test are broken, but it works overall. Best that someone else checks that it works with the plugins and we ignore the tests. After I am back I can start looking both at these tests and the broken 2.13 ones.

tests/unit/src/test/scala/tests/CompletionSlowSuite.scala
```diff
@@ -100,36 +100,40 @@ object CompletionSlowSuite extends BaseCompletionSlowSuite("completion") {
...
-        """|def baz[F[_], A]: F[A] = ???
-           |baz[Either[Int, ?], String].fold@@
-           |""".stripMargin,
-        "fold[C](fa: Int => C, fb: String => C): C"
-      )
-    } yield ()
-  )
+  // FIXME(gabro): the tests don't pass with 2.12.9, although the plugins seem to work fine when
+  // tested manually
+  if (BuildInfo.scalaVersion != "2.12.9") {
```
gabro: FYI @tgodzik / @olafurpg

gabro:
Mill tests are failing due to a strange error (I tried to reproduce it locally and it seems that `mill.contrib.Bloop/install` fails due to a missing `.bloop` directory)

Cross tests are failing due to changes in the Scaladoc in 2.12.7, which again is strange.

I'll investigate later.

gabro:
> Cross tests are failing due to changes in the Scaladoc in 2.12.7, which again is strange.

Ok, it was an issue with the indexing of `scala-library` in tests, which I have fixed in https://github.com/scalameta/metals/pull/867/commits/230d194444c1456e1bde61149c61264c44fb2da4

The only remaining issue should be the Mill test suite

gabro:
Ok, only the mill suite is failing, working on it!

gabro:
Aha, the Mill build is failing due to https://github.com/lihaoyi/mill/pull/678.

I propose we merge by reverting the Mill tests to use 2.12.8 and updating them once a new Mill version is out.

gabro:
The AppVeyor failure seems spurious. I think we can merge this

sarahgerweck:
Any chance of a 0.7.1 release build that uses this?
I can't really push out Scala 2.12.9 until this is in place. 🙍

gabro:
@sarahgerweck that’s the plan, but we’re still ironing out some issues with 2.12.9 support.
Hopefully by next week, but don’t take my word for it ^^

sarahgerweck:
Thanks for the insight @gabro. 👍

PR #883, opened 2019-08-28
Fix VS code docs to include Java 11

tgodzik:


docs/editors/vscode.md
```diff
@@ -44,8 +44,8 @@ variables.
 
 The VS Code plugin uses by default the `JAVA_HOME` environment variable (via
 [`find-java-home`](https://www.npmjs.com/package/find-java-home)) to locate the
-`java` executable. Metals only works with Java 8 so this executable cannot point
-to another version such as Java 11.
+`java` executable. Metals only works with Java 8 or 11 so this executable cannot point
```
marek1840:
```suggestion
`java` executable. Metals works with Java 8 or 11 so this executable cannot point
```

tgodzik:
> This requires changing also the code that selects which java to use

Looking into that.

tgodzik:
I changed the wording in the section, it seems that the library is not finding Java 11 for me, not sure why and I don't want to spend more time on this for now.

Also, the library doesn't really use `JAVA_HOME`, it just returns it if the variable exists, otherwise uses `which` or registry etc.

gabro:
As you noted, the library just finds whatever is installed in the system. If you specify `11` as the target it won't find it? I assumed we could look for 8 or 11 and just run whatever version we find first.

tgodzik:
> As you noted, the library just finds whatever is installed in the system. If you specify `11` as the target it won't find it? I assumed we could look for 8 or 11 and just run whatever version we find first.

Not sure why but it's not finding java 11 :/ `which javac` shows a proper path, but the library doesn't find anything.

tgodzik:
I will leave this as is right not and try to fix the java issue at some point.

PR #887, opened 2019-08-30
Bump Metals Scala version to 2.12.9

tgodzik:
Should fix issues we were having with crashing PC

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -136,7 +136,7 @@ commands += Command.command("save-expect") { s =>
 lazy val V = new {
   val scala210 = "2.10.7"
   val scala211 = "2.11.12"
-  val scala212 = "2.12.8"
+  val scala212 = "2.12.9"
   val scala213 = "2.13.0"
   val scalameta = "4.2.2"
   val semanticdb = scalameta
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -148,7 +148,7 @@ lazy val V = new {
   // List of supported Scala versions in SemanticDB. Needs to be manually updated
   // for every SemanticDB upgrade.
   def supportedScalaVersions =
-    Seq(scala213, "2.12.9", scala212) ++ deprecatedScalaVersions
+    Seq(scala213, "2.12.8", scala212) ++ deprecatedScalaVersions
```
olafurpg: Let's change this to scala212

build.sbt
```diff
@@ -148,7 +148,7 @@ lazy val V = new {
...
   def deprecatedScalaVersions = Seq("2.12.7", scala211)
```
olafurpg:
```suggestion
  def deprecatedScalaVersions = Seq("2.12.7", "2.12.8", scala211)
```

build.sbt
olafurpg: Nevermind, I didn't see it was already there 😅

build.sbt
olafurpg: Nevermind, not needed.

olafurpg:
I have a WIP branch to fix the CI failure of `TreeViewSlowSuite`.

PR #889, opened 2019-09-01
Avoid file leaks when digesting Gradle builds

olafurpg:
Should fix a bug reported on Gitter https://gitter.im/scalameta/metals?at=5d68316b2f10da7cb357a62e

Would be good to include this fix in the upcoming release

cc/ @several27

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/GradleDigest.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/GradleDigest.scala
@@ -1,10 +1,12 @@
 package scala.meta.internal.builds
+
 import scala.meta.io.AbsolutePath
+import scala.meta.internal.mtags.ListFiles
+import scala.meta.internal.jdk.CollectionConverters._
 import java.security.MessageDigest
 import java.nio.file.Files
 import java.util.stream.Collectors
 import java.nio.file.Path
-import scala.meta.internal.jdk.CollectionConverters._
 
 object GradleDigest extends Digestable {
   override protected def digestWorkspace(
@@ -24,9 +26,10 @@ object GradleDigest extends Digestable {
   }
 
   def digestBuildSrc(path: AbsolutePath, digest: MessageDigest): Boolean = {
-    Files.walk(path.toNIO).allMatch { file =>
-      Digest.digestFile(AbsolutePath(file), digest)
+    ListFiles.foreach(path) { file =>
+      Digest.digestFile(file, digest)
     }
+    true
   }
 
   def digestSubProjects(
--- a/metals/src/main/scala/scala/meta/internal/builds/GradleDigest.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/GradleDigest.scala
@@ -24,7 +24,7 @@ object GradleDigest extends Digestable {
   }
 
   def digestBuildSrc(path: AbsolutePath, digest: MessageDigest): Boolean = {
-    Files.walk(path.toNIO).iterator().asScala.forall { file =>
+    Files.walk(path.toNIO).allMatch { file =>
       Digest.digestFile(AbsolutePath(file), digest)
     }
   }
--- a/metals/src/main/scala/scala/meta/internal/builds/GradleDigest.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/GradleDigest.scala
@@ -1,7 +1,7 @@
 package scala.meta.internal.builds
 
 import scala.meta.io.AbsolutePath
-import scala.meta.internal.mtags.ListFiles
+import scala.meta.internal.mtags.WalkFiles
 import scala.meta.internal.jdk.CollectionConverters._
 import java.security.MessageDigest
 import java.nio.file.Files
@@ -26,7 +26,7 @@ object GradleDigest extends Digestable {
   }
 
   def digestBuildSrc(path: AbsolutePath, digest: MessageDigest): Boolean = {
-    ListFiles.foreach(path) { file =>
+    WalkFiles.foreach(path) { file =>
       Digest.digestFile(file, digest)
     }
     true
--- a/metals/src/main/scala/scala/meta/internal/builds/MavenDigest.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/MavenDigest.scala
@@ -1,7 +1,9 @@
 package scala.meta.internal.builds
-import java.nio.file.Files
+
 import java.security.MessageDigest
 import scala.meta.io.AbsolutePath
+import scala.meta.internal.mtags.ListFiles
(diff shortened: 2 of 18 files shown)


metals/src/main/scala/scala/meta/internal/builds/GradleDigest.scala
```diff
@@ -24,7 +24,7 @@ object GradleDigest extends Digestable {
   }
 
   def digestBuildSrc(path: AbsolutePath, digest: MessageDigest): Boolean = {
-    Files.walk(path.toNIO).iterator().asScala.forall { file =>
+    Files.walk(path.toNIO).allMatch { file =>
```
tgodzik: Both `.walk` and `list` mention:
```
     * <p> The returned stream encapsulates one or more {@link DirectoryStream}s.
     * If timely disposal of file system resources is required, the
     * {@code try}-with-resources construct should be used to ensure that the
     * stream's {@link Stream#close close} method is invoked after the stream
     * operations are completed.  Operating on a closed stream will result in an
     * {@link java.lang.IllegalStateException}.
```

I am wondering if the issue is that we don't call `close()` on those methods. `digestBuildSrc` is not used very often, since that is used to define plugins in-build, so I suspect this is not the issue.

tgodzik: Although, I might be completely wrong here.

olafurpg:
Flaky `tests.DefinitionSlowSuite` failure in CI

mtags/src/main/scala/scala/meta/internal/mtags/ListFiles.scala
```diff
@@ -1,50 +1,21 @@
...
-import java.nio.file.SimpleFileVisitor
-import java.nio.file.attribute.BasicFileAttributes
 import scala.meta.io.AbsolutePath
-import scala.collection.mutable.ArrayBuffer
+import scala.collection.mutable
 
 object ListFiles {
-  def foreach(root: AbsolutePath)(fn: AbsolutePath => Unit): Unit = {
-    ListFiles(root).foreach(fn)
+  def apply(root: AbsolutePath): mutable.ArrayBuffer[AbsolutePath] = {
```
marek1840: this should be in an extension method for `AbsolutePath`, like `list` or `children`

mtags/src/main/scala/scala/meta/internal/mtags/WalkFiles.scala
```diff
@@ -0,0 +1,41 @@
...
+          }
+        )
+      } catch {
+        case _: NoSuchFileException =>
+          () // error is reported by the JDK
+      }
+    }
+  }
+
+  def apply(root: AbsolutePath): ArrayBuffer[AbsolutePath] = {
```
marek1840: this should be an extension method for AbsolutePath, like `walk` or `listRecursively` or `descendants` or event `childrenRecursively`

mtags/src/main/scala/scala/meta/internal/mtags/ListFiles.scala
```diff
@@ -1,50 +1,21 @@
...
-      }
-    }
-
-    if (root.isFile) {
-      ArrayBuffer(root)
-    } else {
-      val buf = ArrayBuffer.empty[AbsolutePath]
-      runForeach(file => { buf += file })
-      buf
+  def foreach(root: AbsolutePath)(fn: AbsolutePath => Unit): Unit = {
```
tgodzik: It would be useful to have the couple of methods to use like `map` or `forall` or `exist`. Shouldn't be difficult to do.

mtags/src/main/scala/scala/meta/internal/mtags/ListFiles.scala
```diff
@@ -1,50 +1,21 @@
...
 import java.nio.file.Files
-import java.nio.file.NoSuchFileException
-import java.nio.file.Path
-import java.nio.file.SimpleFileVisitor
-import java.nio.file.attribute.BasicFileAttributes
 import scala.meta.io.AbsolutePath
-import scala.collection.mutable.ArrayBuffer
+import scala.collection.mutable
 
 object ListFiles {
```
marek1840: The name can be misleading, as it lists both files and directories

mtags/src/main/scala/scala/meta/internal/mtags/WalkFiles.scala
```diff
@@ -0,0 +1,41 @@
...
+import java.nio.file.FileVisitResult
+import java.nio.file.Files
+import java.nio.file.NoSuchFileException
+import java.nio.file.Path
+import java.nio.file.SimpleFileVisitor
+import java.nio.file.attribute.BasicFileAttributes
+import scala.meta.io.AbsolutePath
+import scala.collection.mutable.ArrayBuffer
+
+object WalkFiles {
```
tgodzik: Maybe add to extension method? It's really hidden in here, if someone doesn't know about this then they will use something else.

metals/src/main/scala/scala/meta/internal/builds/GradleDigest.scala
```diff
@@ -24,9 +26,10 @@ object GradleDigest extends Digestable {
   }
 
   def digestBuildSrc(path: AbsolutePath, digest: MessageDigest): Boolean = {
-    Files.walk(path.toNIO).iterator().asScala.forall { file =>
-      Digest.digestFile(AbsolutePath(file), digest)
+    WalkFiles.foreach(path) { file =>
```
tgodzik: It would be nice to have the forall method in ListFiles, returning `true` doesn't look good.

metals/src/main/scala/scala/meta/internal/builds/MavenDigest.scala
```diff
@@ -1,20 +1,21 @@
...
       workspace: AbsolutePath,
       digest: MessageDigest
   ): Boolean = {
     Digest.digestFile(workspace.resolve("pom.xml"), digest)
-    Files.walk(workspace.toNIO).allMatch { file =>
-      if (file.getFileName.toString == "pom.xml") {
-        Digest.digestFile(AbsolutePath(file), digest)
-      } else {
-        true
+    WalkFiles.foreach(workspace) { file =>
```
tgodzik: Also here it would be nicer to just have forall.

tgodzik:
> Flaky `tests.DefinitionSlowSuite` failure in CI

I think it didn't get fixed unfortunately in your PR :/

tgodzik:
Let's merge this as is, but I will add a suggestion to the issue about code cleanup.

marek1840:
As it turns out, this doesn't fix the core issue, which is that the resources are not being closed inside the `GradleDigest::digestSubProjects` method. @tgodzik is working on a follow up.

PR #891, opened 2019-09-02
Replace foreach workaround with geny foreach.

olafurpg:
Previously, we implemented a `forall` combinator using a mutable
variable `success` and a `ListFiles.foreach` helper method. Now, we use
the library geny to implement a `JFiles.list(path): Generator[Path]`
method, which exposes a `forall` combinator that safely closes the returned
`DirectoryStream`.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/mtags/JFiles.scala
+++ b/mtags/src/main/scala/scala/meta/internal/mtags/JFiles.scala
@@ -1,0 +1,63 @@
+package scala.meta.internal.mtags
+
+import geny.Generator
+import scala.meta.io.AbsolutePath
+import java.nio.file.Files
+import java.nio.file.SimpleFileVisitor
+import java.nio.file.Path
+import java.nio.file.FileVisitResult
+import java.nio.file.attribute.BasicFileAttributes
+
+object JFiles {
+  def walk(path: AbsolutePath): Generator[AbsolutePath] =
+    new Generator[AbsolutePath] {
+      def generate(
+          handleItem: AbsolutePath => Generator.Action
+      ): Generator.Action = {
+        var result: Generator.Action = Generator.Continue
+        def toResult(action: Generator.Action): FileVisitResult = action match {
+          case Generator.Continue => FileVisitResult.CONTINUE
+          case Generator.End =>
+            result = Generator.End
+            FileVisitResult.TERMINATE
+        }
+        Files.walkFileTree(
+          path.toNIO,
+          new SimpleFileVisitor[Path] {
+            override def preVisitDirectory(
+                dir: Path,
+                attrs: BasicFileAttributes
+            ): FileVisitResult = {
+              toResult(handleItem(AbsolutePath(dir)))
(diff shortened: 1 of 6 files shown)


tgodzik:
This seems a bit more complicated than what we need now. Just using try with finally will solve our issues and we will not be needing to add a whole new library.

Is this really needed? What issue does it solve that we cannot on our own?

marek1840:
> This seems a bit more complicated than what we need now. Just using try with finally will solve our issues

@tgodzik is right in that it looks complex and if one is not accustomed to the monix-observable style might spend some time to understand the code. I think, the priority now is to get the release done, so using the simplest solution (try/finally) is the way to go. This PR looks promising in that it could solve more complex problems in the future, so let's revisit the idea once we are not in a rush to release a new version.

mtags/src/main/scala/scala/meta/internal/mtags/JFiles.scala
```diff
@@ -0,0 +1,63 @@
...
+import scala.meta.io.AbsolutePath
+import java.nio.file.Files
+import java.nio.file.SimpleFileVisitor
+import java.nio.file.Path
+import java.nio.file.FileVisitResult
+import java.nio.file.attribute.BasicFileAttributes
+
+object JFiles {
+  def walk(path: AbsolutePath): Generator[AbsolutePath] =
+    new Generator[AbsolutePath] {
```
marek1840: why not use `Files.walk`?
```scala
val stream = Files.walk(path.toNIO)
Generator.selfClosing(stream.iterator.asScala, () => stream.close())
```

mtags/src/main/scala/scala/meta/internal/mtags/JFiles.scala
```diff
@@ -0,0 +1,63 @@
...
+        result
+      }
+    }
+
+  def list(path: AbsolutePath): Generator[AbsolutePath] =
+    new Generator[AbsolutePath] {
+      def generate(
+          handleItem: AbsolutePath => Generator.Action
+      ): Generator.Action = {
+        val ds = Files.newDirectoryStream(path.toNIO)
```
marek1840: Why a custom generator?
```scala
val stream = Files.newDirectoryStream(path.toNIO)
Generator.selfClosing(stream.iterator.asScala, () => stream.close())
```

mtags/src/main/scala/scala/meta/internal/mtags/JFiles.scala
tgodzik: Wow, this would be much nicer.

marek1840:
Also, methods from `JFiles` should be put in an extension for `AbsolutePath` with names telling us whether this method traverses nested directories or not (maybe even a single method `def list(recursively: Boolean = false): Generator[AbsolutePath]`  ? )

olafurpg:
I don't mind closing. The code in this PR was adapted from the https://github.com/lihaoyi/os-lib/ implementation.

PR #894, opened 2019-09-03
Update website overview and add information about Eclipse

tgodzik:


docs/editors/overview.md
```diff
@@ -154,20 +169,14 @@ from source.
 **✅**: it is possible to import a build such as an sbt project directly from
 the editor.
 
-**Requires browser**: importing a build requires additional steps in a web
-browser using a localhost server. It is not possible to import a build within
-the editor.
+**Manual import**: importing a build requires running `bloopInstall` in one of
```
olafurpg: Does eclipse LSP support executing commands? If so, running "Import build" should do the trick.

The "Requires browser" option also works for Eclipse as long as the `-Dmetals.http=true` is enabled. I believe this is a better workflow than manually running "sbt bloopInstall" because Metals "import build" automatically enables downloading of dependency sources + SemanticDB.

docs/editors/eclipse.md
```diff
@@ -0,0 +1,27 @@
+---
+id: eclipse
+title: Eclipse
+---
+
+Metals works with Eclipse thanks to the
+[`metals-eclipse`](https://github.com/scalameta/metals-eclipse) plugin.
+
+![Eclipse demo](https://i.imgur.com/jrmjVf0.gif)
```
olafurpg: GIF tip: I recommend trying to minimize the screen size and minimize the length. In this particular case, it might be enough to demonstrate a single completion + single diagnostic. Heavy GIFs make the website slower to load. Our landing page + overview pages are already too fat 😅

docs/editors/eclipse.md
```diff
@@ -0,0 +1,27 @@
+---
+id: eclipse
+title: Eclipse
+---
+
+Metals works with Eclipse thanks to the
+[`metals-eclipse`](https://github.com/scalameta/metals-eclipse) plugin.
```
olafurpg: Should we add a note that this is a new integration that is under development?

docs/editors/overview.md
tgodzik: I haven't found any way to run commands, so probably running from localhost makes sense. I will revert and add the option to the plugin.

docs/editors/eclipse.md
tgodzik: Sure, makes sense.

docs/editors/eclipse.md
tgodzik: I will try to minimize that.

docs/editors/overview.md
mundacho: What do you mean by running commands? Do you mean, like running metals from eclipse?

olafurpg: The LSP `workspace/executeCommand` request. Here's the list of commands that Metals supports http://scalameta.org/metals/docs/editors/new-editor.html#metals-server-commands

mundacho: ok, thank you, I'll check that out

docs/editors/eclipse.md
tgodzik: Should be better :)

tgodzik:
@mundacho We should most likely release the latest version with `metals.http=true` before merging this. Do you need help setting up the Travis job?

mundacho:
hi, I'll do the travis job this afternoon, there should be no problem I think. do you have some particular place where I should do this? Or do I just use my personal travis-ci account (I just created one)

tgodzik:
> hi, I'll do the travis job this afternoon, there should be no problem I think. do you have some particular place where I should do this? Or do I just use my personal travis-ci account (I just created one)

Best to do it on the Scalameta one. Maybe @olafurpg activate it for https://travis-ci.org/scalameta/metals-eclipse ?

gabro:
> Best to do it on the Scalameta one. Maybe @olafurpg activate it for [travis-ci.org/scalameta/metals-eclipse](https://travis-ci.org/scalameta/metals-eclipse) ?

done 👍

mundacho:
The travis job is setup, I updated the readme file to document the release process.

PR #901, opened 2019-09-05
Fix more possible presentation compiler classpath issues

tgodzik:
Found two issues with how we create presentation compiler:
- we don't check if semanticDB version in scalac options is a correct one
- when downloading dependencies for PC we don't make sure correct Scala compiler version is downloaded

This should fix both.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
@@ -99,7 +99,8 @@ final class Embedded(
 
 object Embedded {
   def downloadSettings(
-      dependency: Dependency
+      dependency: Dependency,
+      scalaVersion: String
   ): Settings =
     new coursiersmall.Settings()
       .withTtl(Some(Duration.Inf))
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -120,23 +140,19 @@ object Embedded {
       s"mtags_${ScalaVersions.dropVendorSuffix(info.getScalaVersion)}",
       BuildInfo.metalsVersion
     )
-    val needsFullClasspath = !scalac.isSemanticdbEnabled
+    val semanticdbJars = scalac.getOptions.asScala.collect {
+      case opt
+          if opt.startsWith("-Xplugin:") &&
+            opt.contains("semanticdb-scalac") &&
+            opt.contains(BuildInfo.scalametaVersion) =>
```
gabro:
```suggestion
            opt.contains(BuildInfo.semanticdbVersion) =>
```

although they are the same right now, I feel it's more appropriate to use the more specific one

tgodzik: Good catch! Thanks!

Issue #893, opened 2019-09-03
When inserting "|" automatically into multiline strings check for `stripMargin`


MateuszKubuszok:
**Describe the bug**

As mentioned in https://github.com/scalameta/metals/issues/870 when multiline string is inserted, then `|` is added. It is generally OK unless:

 * we use some string interpolation other than `s""`, e.g. `fr""` from Dooby (there `|` could break somthing)
 * there is no `.stripMargin` after the string

In both of these cases this feature might accidentally introduce a bug into code.

**To Reproduce**

1. copy-paste multiline string
2. check the output

**Expected behavior**

Probably something like _check if there is no-interpolation/interpolations known to be safe to work with `|` and `.stripMargin`_ and/or _add `.stripMargin` if neccessary_.

**Installation:**
- I guess `master` of metals?

**Additional context**
This issue was created to not block https://github.com/scalameta/metals/issues/870 while also preserving information that some improvement could be made.

**Search terms**
string interpolation, strip margin, pipe character

tgodzik:
Thanks for reporting! I think we could just verify that `.stripMargin` is at the end. Is that enough for you?

Also, is the current behaviour adding `|` for you when copy pasting or do you refer to the future feature?

MateuszKubuszok:
I think both would be valid improvements: verifying presence of `.stripMargin` during writing as well as during copy-pasting.

PR #904, opened 2019-09-06
Check for stripMargin when adding |

tgodzik:
Fixes https://github.com/scalameta/metals/issues/893

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
@@ -9,6 +9,7 @@ import scala.meta.internal.mtags.Semanticdbs
 import scala.meta.tokens.Token
 import scala.meta.tokens.Token.Constant
 import scala.meta.tokens.Tokens
+import scala.meta.tokens.Token.Interpolation
 
 /*in order to use onTypeFormatting in vscode,
 you'll have to set editor.formatOnType = true in settings*/
@@ -17,19 +18,17 @@ final class OnTypeFormattingProvider(
     buffer: Buffers
 )(implicit ec: ExecutionContext) {
 
-  private val tripleQuote = """\u0022\u0022\u0022"""
+  private val quote = '"'
   private val space = " "
   private val stripMargin = "stripMargin"
 
-  private def isFinishedByStripMargin(
+  private def hasStripMarginSuffix(
       stringTokenIndex: Int,
       tokens: Tokens
   ): Boolean = {
     var methodIndex = stringTokenIndex + 1
-
     while (tokens(methodIndex).isWhiteSpaceOrComment ||
       tokens(methodIndex).isInstanceOf[Token.Dot]) methodIndex += 1
-
     tokens(methodIndex) match {
       case token: Token.Ident if token.value == stripMargin =>
         true
@@ -48,7 +47,10 @@ final class OnTypeFormattingProvider(
   }
 
   private def isMultilineString(text: String, token: Token) = {
-    text.substring(token.start, token.start + 3).equals(tripleQuote)
+    val start = token.start
+    text(start) == quote &&
+    text(start + 1) == quote &&
+    text(start + 2) == quote
   }
 
   private def inToken(pos: meta.Position, token: Token): Boolean = {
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
```diff
@@ -8,17 +8,35 @@ import scala.meta.internal.metals.MetalsEnrichments._
...
+  private val space = " "
+  private val stripMargin = "stripMargin"
+
+  private def isFinishedByStripMargin(
+      stringTokenIndex: Int,
+      tokens: Tokens
+  ): Boolean = {
+    var methodIndex = stringTokenIndex + 1
+
+    while (tokens(methodIndex).isWhiteSpaceOrComment ||
```
olafurpg: Does this handle interpolated strings?

```
s"""
  |Hello $world
""".stripMargin
```

tgodzik: Yes, however I don't think there is a test case for it. I will add that.

gabro: I just tried the latest snapshot and the `|` addition didn't work for me when using an interpolator. Does this PR fixes it?

tgodzik: Och, I think there might be an issue actually coming to think of that 🙁 Will add test cases and fix it here.

metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
```diff
@@ -8,17 +8,35 @@ import scala.meta.internal.metals.MetalsEnrichments._
...
     buffer: Buffers
 )(implicit ec: ExecutionContext) {
 
-  val tripleQuote = """\u0022\u0022\u0022"""
-  val space = " "
+  private val tripleQuote = """\u0022\u0022\u0022"""
+  private val space = " "
+  private val stripMargin = "stripMargin"
+
+  private def isFinishedByStripMargin(
```
marek1840: `hasStripMarginSuffix` ? `finished` sound a bit... final? ;)

metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
```diff
@@ -60,16 +78,23 @@ final class OnTypeFormattingProvider(
...
         val tokens =
           Input.VirtualFile(source.toString(), sourceText).tokenize.toOption
-        tokens.flatMap { tokens =>
-          tokens.collectFirst {
-            case token: Constant.String
-                if inToken(pos, token) &&
-                  isMultilineString(sourceText, token) =>
-              new TextEdit(range, indent(sourceText, pos) + "|")
+        tokens.flatMap { tokens: Tokens =>
+          val tokenIndex = tokens.indexWhere {
```
marek1840: Just a though, no need to act on it.

it would be nice to have the tokens as something akin to the iterator:
```scala
tokens.skipUntil{
 case token: Constant.String => (...)
 case _ =>  false
}
```

the `isFinishedByStripMargin` implementation would be much nicer

tgodzik: This would imply that Tokens are mutable underneath or I would need to create a mutable structure over it :/ I prefer to keep any mutable state local and explicit.

metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
tgodzik: fixed

metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
tgodzik: Ok, now I fixed it. I just assumed we tested it.

metals/src/main/scala/scala/meta/internal/metals/OnTypeFormattingProvider.scala
tgodzik: And I needed to change it even more to make it work with interpolated strings

PR #905, opened 2019-09-06
Update sbt to 1.3.0

gabro:
Let's give this a spin.

I've also taken the chance to upgrade the `recommendedVersion` in `SbtBuildTool`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -165,7 +165,11 @@ lazy val interfaces = project
     libraryDependencies ++= List(
       V.lsp4j
     ),
-    crossVersion := CrossVersion.disabled
+    crossVersion := CrossVersion.disabled,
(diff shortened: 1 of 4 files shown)


gabro:
The error we see in the CI is caused by the `@implNote` tag in the Javadoc of `PresentationCompiler.java`.

Repro with

```scala
sbt
> interfaces / Compile / doc
```

Removing the offending tag fixes the issue, but I would like to understand this a bit better first.

gabro:
Ok, apparently `@implNote` (together with `@implSpec` and `@apiNote`) is not recognized by default in Javadoc (despite them being used frequently in Java 8 own docs), so it has to be enabled with the `-tag` flag.

I'm not sure why this was working at all before, but at least the failure makes sense now. My last commit should fix it.

PR #907, opened 2019-09-09
update `isCompiled` status of the build target regardless of the compilation result

marek1840:
Previously, build target was removed from the `isCompiled` set only when the compilation finished successfully.

Now, we clear this set after every compilation.
Also, the `lastCompile` semantics remains unchanged: it contains only the build targets which were previously compiled successfully.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
@@ -1,11 +1,13 @@
 package scala.meta.internal.metals
+
 import ch.epfl.scala.{bsp4j => b}
+
 import scala.collection.concurrent.TrieMap
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.io.AbsolutePath
-import ch.epfl.scala.bsp4j.BuildTargetIdentifier
 
 final class Compilations(
     buildTargets: BuildTargets,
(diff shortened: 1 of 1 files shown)


olafurpg:
Can you elaborate on the motivation for the change?

tgodzik:
> Can you elaborate on the motivation for the change?

Seems we were not clearing `isCompiling` status when we finished compilation with an error.

marek1840:
> Can you elaborate on the motivation for the change?

Basically, I like when the names reflect the reality. In this case, we could either rename the `isCompiling` to `isCompilingOrFailed` or change the logic to reflect the name. Unless you know any reason to keep the logic, I would like to merge this change.

metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
```diff
@@ -88,19 +90,17 @@ final class Compilations(
...
     val params = new b.CompileParams(targets.asJava)
     targets.foreach(target => isCompiling(target) = true)
-    val compilation = connection.compile(params)
-    val task = for {
-      result <- compilation.asScala
-      _ <- {
-        lastCompile = isCompiling.keySet
+    val compilation = connection
+      .compile(params)
+      .whenComplete((_, error) => {
```
olafurpg: Let's try to use Scala's `Future` operations over `CompletableFuture`. It's easy to mess up with the execution context in the Java API in my experience and it's not idiomatic to for example guard against `error == null`

PR #908, opened 2019-09-09
Add snippets only for supported clients

tgodzik:
Previously, we would use snippet syntax for all clients and now we make sure that the client has that capability.

I'm thinking we could separate the compiler config since it might be dependent on server capabilities and we would be able to create it after the initialize request. I don't think it needs to be in the server config? What do you think @olafurpg ?

Either way we could merge this as a fix before the release and separate the configuration while refactoring MetalsLanguageServer.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -345,10 +345,21 @@ class MetalsLanguageServer(
       interactiveSemanticdbs.toFileOnDisk
     )
     foldingRangeProvider = FoldingRangeProvider(trees, buffers, params)
+
+    val snippetsSupported = params
+      .getCapabilities()
+      .getTextDocument()
+      .getCompletion()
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -345,10 +345,21 @@ class MetalsLanguageServer(
       interactiveSemanticdbs.toFileOnDisk
     )
     foldingRangeProvider = FoldingRangeProvider(trees, buffers, params)
+
+    val snippetsSupported = params
+      .getCapabilities()
+      .getTextDocument()
+      .getCompletion()
```
olafurpg: These  methods can return null.

tgodzik:
> I believe we also use the snippet syntax in “override def” completions.

Actually, it looks like the snippets are already used in a lot of places, so I think snippets are supported in most editors. Otherwise we would already get some complaints.  We can reopen this and merge if the users complain.

PR #913, opened 2019-09-11
Add Scala 2.12.10 support and use 2.12.10 on Metals

gabro:


build.sbt
```diff
@@ -149,7 +149,7 @@ lazy val V = new {
   // List of supported Scala versions in SemanticDB. Needs to be manually updated
   // for every SemanticDB upgrade.
   def supportedScalaVersions =
-    Seq(scala213, "2.12.8", scala212) ++ deprecatedScalaVersions
+    Seq(scala213, "2.12.8", "2.12.9", scala212) ++ deprecatedScalaVersions
   def deprecatedScalaVersions = Seq("2.12.7", scala211)
```
tgodzik: Let's maybe drop 2.12.7 and make 2.12.8 deprecated?

tgodzik: We can deprecate it with the next version, since it's working now.

PR #914, opened 2019-09-11
Revert "Update sbt to 1.3.0"

tgodzik:
Reverts scalameta/metals#905

We cannot publish to 0.13.17 in 1.3.0 :/

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -165,11 +165,7 @@ lazy val interfaces = project
     libraryDependencies ++= List(
       V.lsp4j
     ),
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -165,11 +165,7 @@ lazy val interfaces = project
     libraryDependencies ++= List(
       V.lsp4j
     ),
-    crossVersion := CrossVersion.disabled,
-    javacOptions in (Compile / doc) ++= List(
```
gabro: I would keep this, it's harmless and it's actually more correct

PR #923, opened 2019-09-13
Add debug proxy

marek1840:


metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
```diff
@@ -1,50 +1,114 @@
...
     semanticdbs: Semanticdbs
-) {
-  def findLenses(path: AbsolutePath): util.List[l.CodeLens] = {
-    buildTargets.inverseSources(path) match {
-      case Some(buildTarget) if classes.main(buildTarget).nonEmpty =>
-        findLenses(path, buildTarget).asJava
-      case _ =>
-        emptyList[l.CodeLens]()
-    }
+)(implicit ec: ExecutionContext) {
```
tgodzik: Can we use explicit `ExecutionContext` in parameters?

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
```diff
@@ -1,50 +1,114 @@
...
       case _ =>
         Nil
     }
   }
 }
+
+object CodeLensProvider {
+  import JsonParser._
+  sealed trait CommandFactory[A] {
+    protected def names: List[String]
```
tgodzik: `names` seems to always be a single parameter.

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
```diff
@@ -1,50 +1,114 @@
...
+        } yield new l.CodeLens(range, command, null)
       case _ =>
         Nil
     }
   }
 }
+
+object CodeLensProvider {
+  import JsonParser._
+  sealed trait CommandFactory[A] {
```
tgodzik: I don't see any need for creating the whole class hierarchy that is immediately discarded - it's only used to create something else.

Could we just have 3 methods? `commands` and `testCommands`/`mainCommands` using them ?

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
```diff
@@ -1,50 +1,114 @@
...
+        dataKind,
+        data(arg).toJson
+      )
+
+      names.map { name =>
+        new l.Command(name, StartDebugSession.id, singletonList(params))
+      }
+    }
+
+    protected def data(data: A): AnyRef
```
tgodzik: This gives us no type safety at all, it can turn anything into anything. The method is not really useful, data could easily be a parameter to a method.

metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
```diff
@@ -1,19 +1,16 @@
...
+
+import ch.epfl.scala.bsp4j._
+import ch.epfl.scala.{bsp4j => b}
+import com.google.gson.JsonObject
 import org.eclipse.lsp4j.jsonrpc.services.JsonNotification
 import org.eclipse.{lsp4j => l}
+
 import scala.collection.concurrent.TrieMap
-import scala.concurrent.Promise
+import scala.concurrent.{ExecutionContext, Promise}
```
tgodzik: Can we keep the structure of imports - turn off import formatting in Intellij?

metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
```diff
@@ -24,12 +21,15 @@ final class ForwardingMetalsBuildClient(
...
+    buildTargetClasses: BuildTargetClasses,
     config: MetalsServerConfig,
     statusBar: StatusBar,
     time: Time,
     didCompile: CompileReport => Unit,
-    treeViewProvider: () => TreeViewProvider
-) extends MetalsBuildClient
+    treeViewProvider: () => TreeViewProvider,
+    isCurrentlyOpened: b.BuildTargetIdentifier => Boolean
+)(implicit ec: ExecutionContext)
```
tgodzik: Can we make the execution context explicit here too?

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -51,17 +53,36 @@ object ClientCommands {
...
 
-  val RunMain = Command(
-    "metals-main-run",
-    "run",
-    "Runs main method",
-    """|Array of strings of length 2 where
-       |- first array element is a build target identifier URI
-       |- second array element is the name of the class containing the main method 
-       |Example: `["mybuild://workspace/foo/?id=foo", "com.app.Main"]`
+  val StartDebugSession = Command(
```
tgodzik: Is there any support in LSP for that or it will always need to be done via command?

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -51,17 +53,36 @@ object ClientCommands {
...
+        |      className: "com.foo.App"
+        |   }
+        |}```
     """.stripMargin
   )
 
+  val RefreshModel = Command(
+    "metals-model-refresh",
+    "Refresh model",
+    "Notifies the client that the model has been updated " +
```
tgodzik: Use multiline string?

metals/src/main/scala/scala/meta/internal/metals/JsonParser.scala
```diff
@@ -0,0 +1,35 @@
+package scala.meta.internal.metals
+import com.google.gson.{Gson, JsonElement, JsonObject}
```
tgodzik: Let's keep separate line imports, otherwise it's a bit of a mess in the project. Let's keep it consistent

metals/src/main/scala/scala/meta/internal/metals/JsonParser.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import scala.reflect.{ClassTag, classTag}
+import scala.util.{Failure, Success, Try}
+
+sealed trait JsonParser {
```
tgodzik: Does this need to be a trait? It seems to be only extended here.

metals/src/main/scala/scala/meta/internal/metals/JsonParser.scala
```diff
@@ -0,0 +1,35 @@
...
+  implicit class Parsable(data: Any) {
```
tgodzik: Adding a method on `Any` seems ify (it will be available on everything :fearful: ). Can we make it on a `JsonElement`? We can then add it to `MetalsEnrichments`. It will involve an additional step of matching, but I think it's cleaner.

metals/src/main/scala/scala/meta/internal/metals/JsonParser.scala
```diff
@@ -0,0 +1,35 @@
...
+          Try(JsonParser.gson.fromJson(json, targetType))
+        case value: A =>
+          Success(value)
+        case _ =>
+          Failure(new IllegalArgumentException(s"Cannot convert [$data]"))
+      }
+    }
+  }
+
+  implicit class Serializable(data: Any) {
```
tgodzik: Let's maybe do it for AnyRef.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1087,10 +1106,60 @@ class MetalsLanguageServer(
             )
           )
         }.asJavaObject
+      case ServerCommands.StartDebugAdapter() =>
+        val args = params.getArguments.asScala
+        args match {
+          case Seq(param) =>
```
tgodzik: We can match `Seq(params: JsonElement)` and then as[b.DebugSessionParams] will work without having to use implicit class on Any

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -147,6 +151,23 @@ object MetalsEnrichments
           throw e
       }
     }
+
+    def withTimeout(length: Int, unit: TimeUnit)(
+        implicit ec: ExecutionContext
+    ): Future[A] = {
+      Future(Await.result(future, FiniteDuration(length, unit)))
```
tgodzik: `Try`? Future seems here like a way to pack an error.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1087,10 +1106,60 @@ class MetalsLanguageServer(
             )
           )
         }.asJavaObject
+      case ServerCommands.StartDebugAdapter() =>
```
tgodzik: Any chance to move it to an outside provider? There is already a lot of code inside the server, let's try not to add more.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -56,30 +32,18 @@ import scala.concurrent.Future
...
-import scala.meta.internal.metals.WindowStateDidChangeParams
-import scala.meta.internal.metals.Directories
 import scala.meta.internal.metals.MetalsEnrichments._
-import scala.meta.internal.metals.MetalsLanguageServer
-import scala.meta.internal.metals.MetalsServerConfig
 import scala.meta.internal.metals.PositionSyntax._
-import scala.meta.internal.metals.ProgressTicks
-import scala.meta.internal.metals.Time
-import scala.meta.internal.metals.UserConfiguration
+import scala.meta.internal.metals._
```
tgodzik: Let's not use wildcard imports, they often produce unexpected results and might make compilation slover

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1087,10 +1106,60 @@ class MetalsLanguageServer(
...
+                      val address =
+                        new InetSocketAddress(uri.getHost, uri.getPort)
+                      val timeout = TimeUnit.SECONDS.toMillis(10).toInt
+                      socket.connect(address, timeout)
+
+                      socket
+                    })
+                }
+
+                val server = DebugServer.create(awaitClient, connectToServer)
```
tgodzik: From what I see both `awaitClient` and `connectToServer` both timeout - could we just merge the futures and await for them in DebugServer?

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -283,15 +248,26 @@ final class TestingServer(
...
+  def startDebugging(
+      a: String,
+      kind: String,
+      parameter: AnyRef
+  ): Future[TestDebugger] = {
+    val targets = List(new b.BuildTargetIdentifier(buildTarget(a)))
+    val params = new DebugSessionParams(targets.asJava, kind, parameter.toJson)
+
+    executeCommand(ServerCommands.StartDebugAdapter.id, params)
+      .map(_.asInstanceOf[URI])
```
tgodzik: maybe `collect` with match on URI?  Let's at least merge the two maps

tests/unit/src/main/scala/tests/debug/RemoteConnection.scala
```diff
@@ -0,0 +1,47 @@
...
+  final def listening: Future[Unit] = connection.future
+
+  override final def cancel(): Unit = {
+    connection.cancel()
+  }
+}
+
+object RemoteConnection {
+  def start(launcher: Launcher[_], socket: Socket)(
+      implicit executor: ExecutionContext
```
tgodzik: What is the ExecutionContext used here? It's not the default one, right?

tests/unit/src/main/scala/tests/debug/RemoteConnection.scala
```diff
@@ -0,0 +1,47 @@
...
+  ): CancelableFuture[Unit] = {
+    val cancelable = launcher.startListening()
+    val future = Future(cancelable.get()).andThen { case _ => socket.close() }
+    new CancelableFuture(future.ignoreValue, () => cancelable.cancel(true))
+  }
+
+  def builder[A: ClassTag](socket: Socket, service: Any)(
+      implicit executor: ExecutorService
```
tgodzik: I don't think we should be passing ExecutorService as an implicit too - don't think it's needed.

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
olafurpg: What do you think about implementing this class like this instead?

```suggestion
case class CommandFactory[A](names: List[String], kind: String, arg: A => AnyRef) {
  def commands(..) = ???
}
val mainCommandFactory = CommandFactory[b.ScalaMainClass](List("run"), b.DebugSessionParamsDataKind.SCALA_MAIN_CLASS, arg => arg)
val testCommandFactory = CommandFactory[String](List("run"), b.DebugSessionParamsDataKind.SCALA_TEST_CLASS, arg => List(arg).asJava)
```

I personally would find such an  implementation easier to read and extend compared to the current trait design.

tests/unit/src/main/scala/tests/debug/ClientProxy.scala
```diff
@@ -0,0 +1,50 @@
+package tests.debug
+
+import org.eclipse.lsp4j.debug._
+import org.eclipse.lsp4j.debug.services.IDebugProtocolClient
+
+trait ClientProxy extends IDebugProtocolClient {
```
tgodzik: Why do we need this class? We were talking about just forwarding messages without any actual knowledge of the protocol.

tests/unit/src/main/scala/tests/debug/TestDebugger.scala
```diff
@@ -0,0 +1,129 @@
...
+      outputs.clear()
+    }
+  }
+
+  def disconnect: Future[Unit] = {
+    server.disconnect(new DisconnectArguments).asScala.ignoreValue
+  }
+
+  /**
+   * Not waiting for exited because TODO
```
tgodzik: TODO?

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
tgodzik: To explain, I think it would be useful to give explicit context to Providers so that we know what each of them are using. Will be easier to figure out if we need to optimize.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -37,6 +40,7 @@ import scala.meta.internal.builds.BuildTool
 import scala.meta.internal.builds.BuildTools
 import scala.meta.internal.io.FileIO
 import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.debug.DebugServer
```
tgodzik: Where is DebugServer coming from? Can't find it in the current PR

tests/unit/src/main/scala/tests/debug/ClientProxy.scala
tgodzik: Nvm, it's for tests. I am blind officially.

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
olafurpg: The convention in the codebase so far has been to pass around the `ExecutionContext` implicitly, is there a reason we should do it differently for `CodeLensProvider`?

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
olafurpg: See https://github.com/scalameta/metals/pull/923#discussion_r325610141 for a refactoring proposal.

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
olafurpg: Multiline strings impose a line break, which might not be desirable in this case here. I think these descriptions are always formatted as markdown however, so the line breaks won't affect the formatting of the output as long. Empty lines separate paragraphs however.

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
tgodzik: I'm fine with implicit, I just thought it might help to have explicit ones for the providers. But should not be important currently.

olafurpg: It's not possible to explicitly pass the `ExecutionContext` in for comprehensions
```scala
for {
  a <- foo()
} yield a + 2
```

tgodzik: I just really meant this for constructors like here.

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
marek1840: There is no support there. from the vscode point of view, we have to provide it a port for the adapter to connect. How we do it is completely up to us.

metals/src/main/scala/scala/meta/internal/metals/JsonParser.scala
marek1840: That would break on {de}serializing ints but it is not an important case right now, so we can do it.

Issue #879, opened 2019-08-26
No document symbols inside try/catch/finally blocks


gabro:
**Describe the bug**

Given this code

```scala
val bar = "bar"
try {
  val foo = "foo"
} finally {}
```

**To Reproduce**
Essentially this https://youtu.be/kc2jrTEs5ug?t=755

**Expected behavior**
I would expect `foo` to be present in document symbols, but it's not.

**Installation:**
- Operating system: macOS/Windows/Linux
- Editor: Visual Studio Code/Atom/Vim/Sublime/Emacs
- Metals version: v0.7.0

**Search terms**
document symbol outline try catch

PR #930, opened 2019-09-16
Make sure we continue searching for symbols in try.

tgodzik:
Fixes https://github.com/scalameta/metals/issues/879

image

Edit: Removed the additional block, it could be problematic to present in some clients

image

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
@@ -38,7 +38,9 @@ class DocumentSymbolProvider(trees: Trees) {
     )
     def symbols(tree: Tree): util.List[DocumentSymbol] = {
       apply(tree)
-      owner.getChildren
+      val a = owner.getChildren
+      a.asScala.foreach(a => scribe.warn(a.toString()))
+      a
     }
 
     def addChild(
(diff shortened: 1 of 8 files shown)


metals/src/main/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
```diff
@@ -105,7 +105,7 @@ class DocumentSymbolProvider(trees: Trees) {
             addChild(s"new $name", SymbolKind.Interface, t.pos, selection, "")
             newOwner()
           } else continue()
-        case _: Source | _: Template | _: Term.Block =>
+        case _: Source | _: Template | _: Term.Block | _: Term.Try =>
```
gabro: I gave this some thought and I was wondering on whether we should introduce a new owner, similarly to how we do for `new Foo {}`, since it introduces a different scope.

Or maybe this is fine, I don't know.

I implemented this myself the other day and I got stuck on this point 😂

tgodzik: I was also thinking about it and I think we should try with new owner.

tgodzik: Ok, I changed it around and also added finally and catch. @gabro what do you think?

PR #932, opened 2019-09-17
Refactor autofill option

tgodzik:
- alternatives now start with ???
- it only shows up when using named parameters or is explicitly called
- doesn't auto-fill default arguments
After suggestion from @olafurpg  also:
- add subtypes as valid values

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -941,11 +941,12 @@ trait Completions { this: MetalsGlobal =>
         completions match {
           case CompletionResult.ScopeMembers(positionDelta, results, name) =>
             results
-              .filter(
-                mem =>
-                  mem.sym.tpe.toLongString == paramType.toLongString && mem.sym.isTerm
-              )
-              .map(_.sym.name.toString().trim())
+              .collect {
+                case mem if mem.sym.tpe <:< paramType && mem.sym.isTerm =>
+                  mem.sym.name.toString().trim()
+              }
+              // None and Nil are always in scope
+              .filter(name => name != "Nil" && name != "None")
           case _ =>
             Nil
         }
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -956,24 +956,29 @@ trait Completions { this: MetalsGlobal =>
...
         } else {
           ":???"
         }
       }
 
       private def fillAllFields(): List[TextEditMember] = {
-        if (allParams.size > 1) {
+        val suffix = "autofill"
+        val shouldShow = allParams.exists(_.toString().startsWith(prefix))
+        val isExpilicitelyCalled = suffix.startsWith(prefix)
```
olafurpg:
```suggestion
        val isExplicitlyCalled = suffix.startsWith(prefix)
```

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -956,24 +956,29 @@ trait Completions { this: MetalsGlobal =>
...
+
+        if ((shouldShow || isExpilicitelyCalled) && allParams.size > 1) {
```
olafurpg:
```suggestion
        if ((shouldShow || isExpilicitelyCalled) && allParams.count(!_.hasDefault) > 1) {
```
For example, `pprint.log` has only one non-default parameter

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -956,24 +956,29 @@ trait Completions { this: MetalsGlobal =>
...
           val editText = allParams.zipWithIndex
-            .map {
-              case (param, index) =>
+            .collect {
+              case (param, index) if !param.hasDefault =>
                 s"${param.name} = $${${index + 1}${findDefaultValue(param)}}"
             }
             .mkString(", ")
           val edit = new l.TextEdit(editRange, editText)
+          val filter = if (isExpilicitelyCalled) suffix else prefix + suffix
```
olafurpg: When is filter not `"autofill"`?

tgodzik: It's always `autofill`, at least part of it. This is because I wanted it to be the last option and not show up before any legit completions. The suffix in that case is not that important, it just needs to be something so that it goes later in the list.

olafurpg: The sort text determines the order while the filter text determines what completion match what the user types. I'm struggling to understand the logic behind `if (isExpilicitelyCalled) suffix else prefix + suffix`, what is the difference between that and `"autofill"`?

tgodzik: Ach right,  I was wrong, this was done to show autofill when parameters are completed.
Nothing actually to do with sorting, sorry!

olafurpg: I would personally be in favor of not presenting autofill completions unless the user has typed a prefix of `"autofill"` or the empty string.

tgodzik: Will leave it also when writing parameter names for now. I think it can be much more useful this way, otherwise nobody will find it. If that proves to be problematic, I will remove it myself.

PR #934, opened 2019-09-19
Add goto implementation

tgodzik:
This adds the functionality for `goto implementation` which is done by:
- moving handling of semanticDB file changes out of ReferencesProvider
- creating and additional class that handles extracting semantic data on changes and delegates them to other providers that handle concrete indexes
- adding a new ImplementationProvider with new index for class hierarchies


metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,87 @@
...
+    val result = semanticdbs.textDocument(source)
+    for {
+      doc <- result.documentIncludingStale.toList
+      positionOccurrence = definitionProvider.positionOccurrence(
+        source,
+        params,
+        doc
+      )
+      occ <- positionOccurrence.occurrence.toList
+      defn <- findImplementation(occ.symbol)
```
marek1840: `defn` -> `impl` ?

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,87 @@
...
+      range <- defn.symbol.range
+      revised <- positionOccurrence.distance.toRevised(range.toLSP)
+      path = workspace.toNIO.resolve(Paths.get(defn.uri))
+      uri = path.toUri.toString
+    } yield new Location(uri, revised)
+  }
+
+  private def findImplementation(symbol: String): Set[ClassLocation] = {
+    def findAllImpl(symbol: String): Set[ClassLocation] = {
+      implementations.get(symbol) match {
```
marek1840: maybe we could use `getOrDefault(symbol, Set.empty)` to simplify the logic? 

```scala
implementations.getOrDefault(symbol, Set.empty).flatMap { implementation =>
  implementation ++ findAllImpl(implementation.symbol.symbol)
}

File: metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala [hunk 3]
```diff
@@ -0,0 +1,87 @@
...
+        case None => Set.empty
+        case Some(set) =>
+          set ++ set
+            .flatMap(
+              loc => findAllImpl(loc.symbol.symbol)
```
marek1840: can a loc.symbol.symbol be equal to `symbol` the parameter? If yes, then that would be an infinite loop

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,87 @@
...
+          case ClassSignature(typeParameters, parents, self, declarations) =>
+            parents.collect {
+              case TypeRef(_, symbol, _) =>
+                doc.occurrences
+                  .find(
+                    occ =>
+                      occ.symbol == thisSymbol.symbol && occ.role.isDefinition
+                  )
+                  .foreach { occ =>
+                    val children =
```
marek1840: making the set in `implementations` mutable would simplify the logic as long as those sets remain "in scope" of this provider.

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,87 @@
...
+                    )
+                  }
+            }
+          case _ =>
+        }
+      }
+    }
+  }
+
+  case class ClassLocation(symbol: SymbolOccurrence, uri: String)
```
marek1840: Just my purism kicks in:
```suggestion
  private case class ClassLocation(symbol: SymbolOccurrence, uri: String)
```

metals/src/main/scala/scala/meta/internal/metals/SemanticDBIndexer.scala
```diff
@@ -0,0 +1,64 @@
...
+        stream.forEach(file => if (file.toFile().isFile()) onChange(file))
+      } finally {
+        stream.close()
+      }
+    }
+  }
+
+  def onChange(file: Path): Unit = {
+    if (file.isSemanticdb) {
+      val td = TextDocuments.parseFrom(Files.readAllBytes(file))
```
marek1840: `doc` instead of `td`?

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -612,7 +612,33 @@ final class TestingServer(
       (text, params) <- offsetParams(filename, query, root)
       highlights <- server.documentHighlights(params).asScala
     } yield {
-      TestHighlights.renderAsString(text, highlights.asScala.toList)
+      TestRanges.renderHighlightsAsString(text, highlights.asScala.toList)
+    }
+  }
+
+  def assertImplementation(
```
marek1840: I know we have a convention of calling those names `assertFoo` but I would like to propose that at one point we rename those `assertFoo` to `verifyFoo` because after reading `assert implementation` I want to ask `assert that implementation *what*?` Let's make the act of verification explicit in the name.

tests/unit/src/test/scala/tests/ImplementationSuite.scala
```diff
@@ -0,0 +1,64 @@
+package tests
+
+object ImplementationSuite extends BaseSlowSuite("implementation") {
+
+  check(
+    "basic",
+    """|package a
+       |trait Livin@@gObject
```
marek1840: `LivingBeing` ? :P

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: We only get a loop if class hierachies create a loop which I don't think is possible.

tgodzik: Anyway loc.symbol can't be equal to symbol since their relation is `extends` and `A extends A` is not possible

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: I don't think this will change much, one line less and I prefer using immutable sets there. Will refactor the logic there though.

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,87 @@
...
+import scala.meta.internal.semanticdb.ClassSignature
+import scala.meta.internal.semanticdb.TypeRef
+import scala.collection.concurrent.TrieMap
+
+final class ImplementationProvider(
+    semanticdbs: Semanticdbs,
+    workspace: AbsolutePath,
+    definitionProvider: DefinitionProvider
+) {
+  private val implementations: TrieMap[String, Set[ClassLocation]] =
```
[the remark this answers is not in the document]
tgodzik: Switched to concurrentHashMap in case we ever run indexing multithreaded.

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: A bit different, but now it takes up less code.

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: Changed to ConcurrentHashMap and switched to `compute` I will leace the immutable set.

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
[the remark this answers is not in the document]
marek1840: isn't a TrieMap (from `scala.collection.concurrent`) already thread-safe?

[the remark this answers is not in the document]
olafurpg: `TrieMap` is already thread safe.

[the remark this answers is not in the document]
tgodzik: It doesn't have the compute method though, which I am using here.

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,95 @@
...
+      impl <- findImplementation(occ.symbol)
+      range <- impl.symbol.range
+      revised <- positionOccurrence.distance.toRevised(range.toLSP)
+      path = workspace.toNIO.resolve(Paths.get(impl.uri))
+      uri = path.toUri.toString
+    } yield new Location(uri, revised)
+  }
+
+  def onChange(docs: TextDocuments): Unit = {
+    docs.documents.foreach { doc =>
```
olafurpg: nit: I personally think for comprehensions make it easier to read what's going on
```suggestion
for {
  doc <- docs.documents.iterator
  thisSymbol <- doc.symbols.iterator
  occ <- doc.occurrences.iterator.find(occ => occ.symbol == thisSymbol.symbol && occ.role.isDefinition)
} {
  addFromSignature(thisSymbol.signature, occ, doc)
}
```

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,95 @@
...
+    findAllImpl(symbol)
+  }
+
+  private def addFromSignature(
+      signature: Signature,
+      occ: SymbolOccurrence,
+      doc: TextDocument
+  ): Unit = {
+    signature match {
+      case ClassSignature(
```
marek1840: maybe we could use 
```scala
case sig: ClassSignature => 
  (...)
```

and use field accessors instead?

olafurpg: Nit: I generally avoid binding variable names that are unused, I find this easier to read
```suggestion
      case c: ClassSignature => c.parents.collect { ... }
```

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -612,7 +612,33 @@ final class TestingServer(
...
+        "obtained",
+        "expected"
+      )
+    }
+  }
+
+  def implementation(filename: String, query: String) = {
+    for {
+      (text, params) <- offsetParams(filename, query, workspace)
+      implemtees <- server.implementation(params).asScala
```
marek1840: why not implementations?

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -612,7 +612,33 @@ final class TestingServer(
       (text, params) <- offsetParams(filename, query, root)
       highlights <- server.documentHighlights(params).asScala
     } yield {
-      TestHighlights.renderAsString(text, highlights.asScala.toList)
+      TestRanges.renderHighlightsAsString(text, highlights.asScala.toList)
+    }
+  }
+
+  def verifyImplementation(
```
olafurpg: For consistency with other methods
```suggestion
  def assertImplementation(
```

tests/unit/src/test/scala/tests/ImplementationSuite.scala
```diff
@@ -0,0 +1,75 @@
...
+       |    class <<Dog>> extends Animal
+       |    class <<Cat>> extends Animal
+       |    class Unrelated extends LivingBeing
+       |  }
+       |}
+       |""".stripMargin
+  )
+
+  // TODO needs scalameta update
+  // check(
```
olafurpg: 👍

metals/src/main/scala/scala/meta/internal/metals/SemanticDBIndexer.scala
```diff
@@ -0,0 +1,66 @@
+package scala.meta.internal.metals
+
+import java.nio.file.Path
+import MetalsEnrichments._
+import java.nio.file.Files
+import ch.epfl.scala.bsp4j.ScalacOptionsResult
+import scala.meta.internal.semanticdb.TextDocuments
+
+class SemanticDBIndexer(
```
olafurpg: nit:
```suggestion
class SemanticdbIndexer(
```

In written documentation the correct capitalization is `SemanticDB` but for better or worse, we use `SemanticdbFoo` in code since I didn't like how `SemanticDBFoo` reads

tests/unit/src/main/scala/tests/TestingServer.scala
tgodzik: I liked the word I made up :P #zabawopsuj

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,95 @@
...
+            implementations.compute(symbol, { (_, set) =>
+              if (set == null) Set(loc)
+              else set + loc
+            })
+        }
+      case _ =>
+    }
+  }
+
+  private case class ClassLocation(symbol: SymbolOccurrence, uri: String)
```
olafurpg:
```suggestion
  private case class ClassLocation(symbol: SymbolOccurrence, path: java.nio.file.Path)
```

Wherever possible I prefer nio `Path` because they have a small memory footprint.

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,90 @@
...
+      doc <- result.documentIncludingStale.toList
+      positionOccurrence = definitionProvider.positionOccurrence(
+        source,
+        params,
+        doc
+      )
+      occ <- positionOccurrence.occurrence.toList
+      impl <- findImplementation(occ.symbol)
+      range <- impl.symbol.range
+      revised <- positionOccurrence.distance.toRevised(range.toLSP)
```
olafurpg: I think you need to use a custom `positionOccurrence` for the file where the `impl` symbol is defined. The test cases don't seem to stress the situation when there are multiple source files. It would also be good to have a test case with multiple projects, although I don't think the current implementation has issues with that it will become more relevant as we add `implementation` for methods.

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,90 @@
...
+    val result = semanticdbs.textDocument(source)
+    for {
+      doc <- result.documentIncludingStale.toList
+      positionOccurrence = definitionProvider.positionOccurrence(
+        source,
+        params,
+        doc
+      )
+      occ <- positionOccurrence.occurrence.toList
+      impl <- findImplementation(occ.symbol)
```
olafurpg: The `impl.uri` file may no longer exist since the index doesn't clean up stale data.

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,90 @@
...
+import scala.meta.internal.semanticdb.TextDocument
+import java.util.concurrent.ConcurrentHashMap
+
+final class ImplementationProvider(
+    semanticdbs: Semanticdbs,
+    workspace: AbsolutePath,
+    definitionProvider: DefinitionProvider
+) {
+  private val implementations =
+    new ConcurrentHashMap[String, Set[ClassLocation]]
```
olafurpg: Would it be possible to use an index that is shaped something like this `Map[Path, ???]`? I think that would make it easier to incrementally update the index and remove stale data as files get deleted. It's OK if querying takes a performance hit, I don't querying will be a bottleneck.

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: Right, I started looking at that and will for sure need to add test cases with multiple files. Right now I don't think it's really needed, but I can add that nonetheless.

tgodzik: Actually, I was all wrong :O After adding the multiple files suite, I found the issue you were talking about

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: Fixed

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: Done

tests/unit/src/main/scala/tests/TestingServer.scala
tgodzik: Changed to assert

metals/src/main/scala/scala/meta/internal/metals/SemanticDBIndexer.scala
tgodzik: Changed

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: Changed to Path

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: Added path as the key in the main map, it serves as a kind of namespace now. Thanks!

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: Fixed and tested.

tgodzik:
@olafurpg @marek1840 Thanks for the reviews! Managed to find a lot of issues thanks to them.

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,121 @@
...
+      range <- impl.symbol.range
+      revised <- distance.toRevised(range.toLSP)
+      uri = impl.file.toUri.toString
+    } yield new Location(uri, revised)
+  }
+
+  private def findImplementation(symbol: String): Set[ClassLocation] = {
+    def findAllImpl(symbol: String): Set[ClassLocation] = {
+      val directImpl = for {
+        (path, symbols) <- implementationsInPath.asScala
```
marek1840: is `path` used anywhere?

tests/unit/src/main/scala/tests/TestRanges.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  def renderLocationsAsString(
+      sourceFiles: Map[String, String],
+      locations: List[Location]
+  ): Map[String, String] = {
+    val resolved = for {
+      (file, code) <- sourceFiles.toSeq
+      validLocations = locations.filter(_.getUri().contains(file))
+    } yield
+      file -> validLocations.foldLeft(code) { (base, highlight) =>
```
marek1840: `highlight` -> `location` ?

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
```diff
@@ -0,0 +1,67 @@
...
+        onChangeDirectory(root)
+      case None =>
+    }
+  }
+
+  def onChangeDirectory(dir: Path): Unit = {
+    if (Files.isDirectory(dir)) {
+      val stream = Files.walk(dir)
+      try {
+        stream.forEach(path => if (path.toFile().isFile()) onChange(path))
```
marek1840: we could hide the `isFile` check inside `onChange`:
```scala
def onChange(path: Path) = {
  if(!Files.isFile(path)) ()
  else (...)
}
```

or at least we could use `Files.isFile(path: Path)` instead of `path.toFile.isFile()`

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  def onChange(file: Path): Unit = {
+    if (file.isSemanticdb) {
+      val doc = TextDocuments.parseFrom(Files.readAllBytes(file))
+      referenceProvider.onChange(doc, file)
+      implementationProvider.onChange(doc, file)
+    } else {
+      scribe.warn(s"not semanticdb file: $file")
```
marek1840:
```suggestion
      scribe.warn(s"not a semanticdb file: $file")
```

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   * Handle EventType.OVERFLOW, meaning we lost file events for a given path.
+   *
+   * We walk up the file tree to the parent `META-INF/semanticdb` parent directory
+   * and re-index all of its `*.semanticdb` children.
+   */
+  def onOverflow(path: Path): Unit = {
+    path.semanticdbRoot match {
```
marek1840:
```scala  
path.semanticdbRoot.foreach(onChangeDirectory(_))
```

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
```diff
@@ -0,0 +1,67 @@
...
+import scala.meta.internal.semanticdb.TextDocuments
+
+class SemanticdbIndexer(
+    referenceProvider: ReferenceProvider,
+    implementationProvider: ImplementationProvider
+) {
+
+  def onScalacOptions(scalacOptions: ScalacOptionsResult): Unit = {
+    for {
+      item <- scalacOptions.getItems.asScala
```
marek1840:
```scala
scalacOptions.getItems.foreach { item =>
  val targetroot = item.targetroot
  onChangeDirectory(targetroot.resolve(Directories.semanticdb).toNIO)
}

File: metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala [hunk 26]
```diff
@@ -0,0 +1,121 @@
...
+    implementationsInPath.clear()
+  }
+
+  def onDelete(path: Path): Unit = {
+    implementationsInPath.remove(path)
+  }
+
+  def onChange(docs: TextDocuments, path: Path): Unit = {
+    implementationsInPath.compute(
+      path, { (_, _) =>
```
marek1840: could you extract the body to make it more readable?

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,121 @@
...
+      else TokenEditDistance.fromBuffer(fileSource, doc.text, buffer)
+      impl <- locations
+      range <- impl.symbol.range
+      revised <- distance.toRevised(range.toLSP)
+      uri = impl.file.toUri.toString
+    } yield new Location(uri, revised)
+  }
+
+  private def findImplementation(symbol: String): Set[ClassLocation] = {
+    def findAllImpl(symbol: String): Set[ClassLocation] = {
```
marek1840: do we even need the `findAllImpl` ? Could we move its body to `findImplementation` ?

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,121 @@
...
+      currentDoc <- findSemanticdb(source)
+      positionOccurrence = definitionProvider.positionOccurrence(
+        source,
+        params,
+        currentDoc
+      )
+      occ <- positionOccurrence.occurrence.toList
+      (file, locations) <- findImplementation(occ.symbol).groupBy(_.file)
+      fileSource = AbsolutePath(file)
+      doc <- findSemanticdb(fileSource)
```
marek1840: it is only used in a false branch of the next `if` statement. Maybe those could be merged?

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,121 @@
...
+
+        allParents.groupBy(_._1).map {
+          case (symbol, locations) =>
+            symbol -> locations.map(_._2).toSet
+        }
+      }
+    )
+  }
+
+  private def addFromSignature(
```
marek1840: it returns `Option[List[(Symbol, Location)]]` right? We could easily make it a `List[(Symbol, Location)]` which would simplify `onChange` a bit:
```scala
          parents <- addFromSignature(thisSymbol.signature, occ, doc).toList
          parent <- parents
        } yield parent
```

would become 
```scala
          parent <- addFromSignature(thisSymbol.signature, occ, doc).toList
        } yield parent
```

marek1840: rename function to `collectParentsOf` or something?

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
tgodzik: It's basically the same thing, I will leave it but switch {} to ()

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: Actually, this is an optimization that most likely will not be needed. I will remove the `if else`, it clouds the logic somewhat

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: No, replaced by `_`

tests/unit/src/main/scala/tests/TestRanges.scala
tgodzik: Fixed

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
tgodzik: Changed to `!Files.isDirectory(path)` since it shows better the intent, and moved to `onChange`

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
tgodzik: done

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
tgodzik: changed

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: Moved it to another method

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: Not needed, removed and left one func.

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: Switched to Seq and it simplified the code.

tgodzik: Renamed

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  def onChange(docs: TextDocuments, path: Path): Unit = {
+    implementationsInPath.compute(
+      path, { (_, _) =>
+        computeInheritance(docs)
+      }
+    )
+  }
+
+  private def computeInheritance(docs: TextDocuments) = {
```
olafurpg: Can you please add type annotations to methods, it would make it easier to review the code on GitHub

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,120 @@
...
+      signature: Signature,
+      occ: SymbolOccurrence,
+      doc: TextDocument
+  ) = {
+    val filePath = workspace.toNIO.resolve(Paths.get(doc.uri))
+    val loc = ClassLocation(occ, filePath)
+    signature match {
+      case classSig: ClassSignature =>
+        val allLocations = classSig.parents.collect {
+          case TypeRef(_, symbol, _) =>
```
olafurpg:
```suggestion
          case t: TypeRef => t.symbol
```

tests/unit/src/test/scala/tests/ImplementationSuite.scala
```diff
@@ -0,0 +1,107 @@
...
+         |trait Livin@@gBeing
+         |abstract class <<Animal>> extends LivingBeing
+         |class <<Dog>> extends Animal
+         |class <<Cat>> extends Animal
+         |""".stripMargin
+  )
+
+  check(
+    "advanced",
+    "src/main/scala/a/LivingBeing.scala" ->
```
olafurpg: I would prefer to keep the multiline string syntax for declaring a file layout
```scala
"""|
   |/src/main/scala/a/LivingBeing.scala
   |object foo
   |/src/main/scala/a/Bar.scala
   |object bar
   |""".stripMargin
```

The large number of `"""`, `->`, commas and other syntax makes it harder to read and evolve the file structure.

tgodzik: Fixed it :)

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
tgodzik: Done

metals/src/main/scala/scala/meta/internal/metals/ImplementationProvider.scala
```diff
@@ -0,0 +1,120 @@
...
+        computeInheritance(docs)
+      }
+    )
+  }
+
+  private def computeInheritance(docs: TextDocuments) = {
+    val allParents = for {
+      doc <- docs.documents
+      thisSymbol <- doc.symbols
+      occ <- doc.occurrences
```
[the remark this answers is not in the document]
tgodzik: Changed to calculate occurrence only when being asked about the implementation

tgodzik:
Need to add support for non-workspace classes still - working on that.

tgodzik:
Alright, I managed to fix the issue reported by @olafurpg and added some tests. Then added some more tests and now I am out of ideas for what might not work anymore. Let me know if anyone can think of more corner cases in the tests.

Tried to tidy the code as much as possible, let me know if something is not clear.
I would appreciate every possible feedback.

olafurpg:
I hit on one case in Akka where I expected results but got "no implementation found"

Screenshot 2019-10-03 at 12 22 30

tgodzik:
> I hit on one case in Akka where I expected results but got "no implementation found"
> 
> Screenshot 2019-10-03 at 12 22 30

Checking!

metals/src/main/scala/scala/meta/internal/implementation/InheritanceContext.scala
```diff
@@ -0,0 +1,50 @@
+package scala.meta.internal.implementation
+
+import scala.meta.internal.semanticdb.SymbolInformation
+import java.nio.file.Path
+import scala.collection.mutable
+
+case class InheritanceContext(
```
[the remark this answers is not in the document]
tgodzik: I think I prefer the Context.

metals/src/main/scala/scala/meta/internal/implementation/ClassLocation.scala
```diff
@@ -0,0 +1,111 @@
...
+) {
+
+  def asSeenFromMap: Map[String, String] = asSeenFrom match {
+    case Some(v) => v
+    case None => Map.empty
+  }
+
+  def translateAsSeenFrom(other: ClassLocation): ClassLocation = {
+    val newASF = other.asSeenFrom match {
+      case None => other.asSeenFrom
```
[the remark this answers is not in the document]
tgodzik: Done!

tgodzik:
> > I hit on one case in Akka where I expected results but got "no implementation found"
> > Screenshot 2019-10-03 at 12 22 30
> 
> Checking!

That's fixed, problem was when symbol was not found in GlobalSymbolTable

jvican:
Congratulations on getting this one merged, I'm excited to try this out, looks fantastic 👍

PR #938, opened 2019-09-23
replace java streams with generators

marek1840:
Previously, we created a wrapper for file walking and children listing to automatically close the underlying java stream. It lacked the flexibility and ease of use of standard collections.
Now, we use the geny library to close the stream automatically. It also provides a subset of scala collection methods (forall, map, etc.)

This is based on a [previous work](https://github.com/scalameta/metals/pull/891) of @olafurpg


mtags/src/main/scala/scala/meta/internal/mtags/MtagsEnrichments.scala
```diff
@@ -359,4 +363,31 @@ trait MtagsEnrichments {
...
+    def parent: AbsolutePath = {
+      AbsolutePath(path.toNIO.getParent)
+    }
+
+    def exists: Boolean = {
+      Files.exists(path.toNIO)
+    }
+
+    def walk: Generator[AbsolutePath] = {
+      if (exists) Files.walk(path.toNIO).asScala.map(AbsolutePath(_))
```
olafurpg: Looks like there's an implicit conversion going on here and the stream won't be closed. I would use the `Geny.selfClosing` constructor to ensure that it's properly closed.

marek1840: streams don't have an implicit conversion with `asScala` - that is [my own addition](https://github.com/scalameta/metals/pull/938/files/392c3f84dd8bc06a82310a7f14b0358bfd1c9465#diff-ef3c89457c7ae83653bcdf6c20438e96R369). It tries to mimick the scala conversion to avoid any surprises on the callee site (like: "what was the name of that special method for streams?") - I want it to be as seamless as possible.

marek1840:
> I'm not sure I like the `AbsolutePath.{walk,list}` extension methods. I feel like it would be better with a `JFiles.walk(AbsolutePath)` method on an object.

I think it fits. there already is a method `parent` in `AbsolutePath`, so why just make it easier for walking the tree in just a one direction? I think there is no gain in having separate object for those special cases if they are so similar to what is already available in `AbsolutePath`

mtags/src/main/scala/scala/meta/internal/mtags/MtagsEnrichments.scala
```diff
@@ -359,4 +363,32 @@ trait MtagsEnrichments {
...
+  implicit class XtensionAbsolutePath(path: AbsolutePath) {
+    def parent: AbsolutePath = {
+      AbsolutePath(path.toNIO.getParent)
+    }
+
+    def exists: Boolean = {
+      Files.exists(path.toNIO)
+    }
+
+    def iterate: Generator[AbsolutePath] = {
```
marek1840: I am not sure about the name. While `walk` is sometimes used in this context, I feel it does not carry the information of what is actually done. I was thinking about something with the `hierarchy` in the name, like `traverseHierarchy`, `toTree`, maybe even the closest-to-original `walkTree` ? What do you think?

marek1840: Hmm, how about `list` and `listRecursive`?

mtags/src/main/scala/scala/meta/internal/mtags/MtagsEnrichments.scala
olafurpg: Aha, I'm sorry I didn't see that one. Looks great 👍

PR #940, opened 2019-09-23
Experiment with GitHub Actions

olafurpg:
GitHub Actions (https://github.com/features/actions) offer a promising alternative to Travis CI. This PR sets up automatic CI releases for this build via GitHub Actions so that we can disable the Travis, Appveyor and Azure CIs.


olafurpg:
It looks like it's not yet possible to re-run individual job failures that are caused by flakyness https://github.community/t5/GitHub-Actions/quot-This-check-suite-has-disabled-re-running-individual-check/m-p/31696

olafurpg:
@tgodzik do these error ring a bell?
```
INFO  running '/tmp/metals1573090658375376626/mvnw ch.epfl.scala:maven-bloop_2.10:1.3.2:bloopInstall -DdownloadSources=true'
ERROR /tmp/metals1573090658375376626/mvnw: 186: /tmp/metals1573090658375376626/mvnw: cannot open ./.mvn/wrapper/maven-wrapper.properties: No such file
ERROR ./.mvn/wrapper/maven-wrapper.jar: No such file or directory
ERROR Error: Could not find or load main class org.apache.maven.wrapper.MavenWrapperMain
```
From https://github.com/scalameta/metals/pull/940/checks?check_run_id=233844637

olafurpg:
Looks like we should try to split the `unit/test` job into multiple jobs. It's currently the bottleneck on the Ubuntu job and it's also hitting on OOM errors in Windows.

.github/workflows/ci.yml
```diff
@@ -1,17 +1,77 @@
...
+      - uses: actions/checkout@v1
+      - uses: olafurpg/setup-scala@v2
+      - name: Run command
+        run: csbt 'slow/testOnly -- tests.${{ matrix.tool }}'
+  PresentationCompiler:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        scala:
+          - "2.13.0"
```
gabro: Also, `2.13.11

mtags/src/main/scala/scala/meta/internal/metals/PackageIndex.scala
```diff
@@ -191,7 +191,7 @@ object PackageIndex {
           Classpath(v).entries
       }.toList
       entry <- entries
-      if entry.isFile
+      if entry.isFile && !entry.toNIO.endsWith("jfr.jar")
```
gabro: what's `jfr.jar`? Maybe this is worth a comment?

.github/workflows/ci.yml
olafurpg: We'll need to add a sbt command to auto-generate this matrix to ensure it's in sync with the build. I don't expect to merge this PR for now because GitHub Actions don't support re-running individual checks yet. I'm just trying to get a feel for how much we can push it 😄

mtags/src/main/scala/scala/meta/internal/metals/PackageIndex.scala
olafurpg: Good point, added (but not pushed yet).

.github/workflows/ci.yml
olafurpg: We will also need to ensure that the `unit/test` cross-matrix covers all the packages under `tests`

gabro: See also https://twitter.com/milessabin/status/1176058458262704128

build.sbt
```diff
@@ -102,7 +103,55 @@ inThisBuild(
...
+                baseDirectory.in(ThisBuild).value /
+                  "tests" / "unit" / "src" / "test" / "scala" / "tests",
+                partitionCount = Some(3)
+              )
+            )
+          ),
+          "Mtags" -> githubActionsJob(
+            "Run tests",
+            "csbt ++${{  matrix.scala }} cross/test",
+            githubActionsMatrix("scala", V.supportedScalaVersions)
```
olafurpg: @milessabin One option for syncing the CI with `crossScalaVersions` is to automatically generate `.travis.yml` from build.sbt. In this PR we're using GitHub Actions but the same approach should work for Travis CI.

milessabin: @olafurpg you must be pulling my leg ... @djspiewak told me this is "fundamentally impossible"! ;-)

But seriously ... thanks for the pointer :-)

olafurpg: It doesn't even seem necessary to print YAML, you can print JSON as long as the filename stays `travis.yml`

djspiewak: > In this PR we're using GitHub Actions but the same approach should work for Travis CI.

It doesn't work with travisci unless you commit the results. If Github Actions has multi-staged builds without a need to hard-commit results, then it would be capable of this, but Travis' architecture doesn't support that.

milessabin: > It doesn't work with travisci unless you commit the results.

When did I ever say I didn't want to commit the results!?!

olafurpg:
I'm gonna close this PR for now as an interesting experiment.

* I think it's a good idea to generate the build matrix from build.sbt.
* I think it's a good idea to split the codebase into smaller packages so that we can more easily run tests in parallel
* I think it's best to wait until we switch the CI to GitHub Actions, primarily the feature that is missing IMO is the ability to re-run an individual job that failed for some reason (flaky download, etc). Lack of caching doesn't seem to be a blocker.

There are too many unrelated changes in this PR mixed together for me to feel comfortable merging it. I think it's best that we consider switching by incrementally enabling parts of the build.

PR #942, opened 2019-09-24
Add support for running and testing via Debug Adapter Protocol

marek1840:
Previously there was no way to launch a main class or test suite from within vscode. Now, we provide code lenses for those. 
Using this lenses starts a debug adapter within `bloop` and creates a proxy in metals.

run-and-test

This is a continuation of #923. It turns out github didn't handle a rebase very well


tests/slow/src/test/scala/tests/sbt/SbtSlowSuite.scala
```diff
@@ -245,22 +245,17 @@ object SbtSlowSuite extends BaseImportSuite("sbt-import") {
...
-        assertNoDiff(
-          client.workspaceClientCommands,
-          List(
-            ClientCommands.ReloadDoctor.id,
-            ClientCommands.RunDoctor.id
-          ).mkString("\n")
-        )
+        val expected = ClientCommands.ReloadDoctor.id :: ClientCommands.RunDoctor.id :: Nil
+        val actual = client.workspaceClientCommands
+        assert(actual.startsWith(expected))
```
olafurpg: What is the motivation for this change? The ScalaTest error message inside asserts for large lists is quite difficult to read. The output format from `assertNoDiff` is easier to debug in my experience (although it may take some time to get used to the output)

marek1840: because now the `client.workspaceClientCommands` contain also ClientCommands.RefreshModel, so the `noDiff` ` actually won't do.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1099,10 +1113,32 @@ class MetalsLanguageServer(
             )
           )
         }.asJavaObject
+      case ServerCommands.StartDebugAdapter() =>
+        val args = params.getArguments.asScala
+        args match {
+          case Seq(param: JsonElement) =>
+            param.as[b.DebugSessionParams] match {
+              case Failure(exception) =>
```
tgodzik: Can we use map and match only on end result

tests/unit/src/test/scala/tests/DebugProtocolSuite.scala
```diff
@@ -0,0 +1,104 @@
...
+  private val mainFile = "a/src/main/scala/a/Main.scala"
+
+  testAsync("start") {
+    for {
+      _ <- server.initialize(
+        s"""/metals.json
+           |{
+           |  "a": {}
+           |}
+           |/$mainFile
```
tgodzik: Can we replace `mainFile` withe actual file, it would be a bit clearer and all code in one place.

tests/unit/src/test/scala/tests/CodeLensesSlowSuite.scala
```diff
@@ -96,12 +93,48 @@ object CodeLensesSlowSuite extends BaseSlowSuite("codeLenses") {
...
+    } yield ()
+  }
+
   private def assertCodeLenses(
       filename: String,
       expected: String
   ): Future[Unit] =
     for {
       _ <- server.didOpen(filename)
+      _ <- server.didSave(filename)(x => x) // first compilation could have not yet persisted analysis
```
tgodzik: Maybe we should report it as a bug in Bloop?

marek1840:
The tests fail because bloop is unable to launch the debuggee:
>ERROR Error looking up function 'Java_java_lang_ProcessImpl_forkAndExec': 
/usr/lib/jvm/zulu-8-azure-amd64/jre/lib/amd64/libjava.so: undefined symbol: Java_java_lang_ProcessImpl_forkAndExec

which are likely due to [those lines](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/linux/LinuxProcess.java#L93-L94)

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -1,75 +1,91 @@
...
+    } classesOf(target).testSuites.put(objectSymbol, className)
+  }
+
+  private def createObjectSymbol(className: String): String = {
+    val symbol = className.replaceAll("\\.", "/")
+    val isInsideEmptyPackage = !className.contains(".")
+    if (isInsideEmptyPackage) {
+      Symbols.Global(Symbols.EmptyPackage, Descriptor.Term(symbol))
+    } else {
+      symbol + "."
```
olafurpg: This conversion will fail for non-ascii class names such as `HælsingSuite` or `::`. Although, I don't think these will be common, I believe the following change could make it work even in those cases

```suggestion
      import scala.reflect.NameTransformer
      symbol.split(".").foldLeft(Symbols.RootPackage) {
        case (owner, name) => Symbols.Global(owner, Descriptor.Term(NameTransformer.decode(name))
```

Either way, it would be good to have a test case to show that it does or doesn't work with non-ascii class names.

metals/src/main/scala/scala/meta/internal/metals/Cancelable.scala
```diff
@@ -21,6 +21,11 @@ object Cancelable {
   def cancelEach[T](iterable: Iterable[T])(fn: T => Unit): Unit = {
     cancelAll(iterable.map(elem => Cancelable(() => fn(elem))))
   }
+
+  def cancelAll(cancelables: Cancelable*): Unit = {
```
olafurpg: nit: can we remove this overload in favor of explicitly wrapping call-site arguments in a `List(..)`? I generally try to avoid varargs where possible.

metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
```diff
@@ -139,6 +141,11 @@ final class ForwardingMetalsBuildClient(
             scribe.info(s"time: compiled $name in ${compilation.timer}")
           }
           if (isSuccess) {
+            buildTargetClasses
+              .rebuildIndex(target)
+              .filter(_ => isCurrentlyFocused(target))
```
olafurpg: nit: `filter` should generally be avoided on `Future` values because it makes the `Future` fail, it's better to do this instead

```scala
rebuildIndex().foreach { _ =>
  if (foo()) { doBar() }
}
```

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -1,75 +1,91 @@
 package scala.meta.internal.metals
 
 import ch.epfl.scala.{bsp4j => b}
+
```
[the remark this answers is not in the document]
marek1840: yup, it is already done - the file was just not modified after setting the config in intellij, so it was not reformatted :(

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
marek1840: `HælsingSuite` worked out of the box, but indeed the `::` was failing. Thanks!

olafurpg:
The Maven CI is frequently failing with
```
INFO  [ERROR] Plugin ch.epfl.scala:maven-bloop_2.10:1.3.2+252-b4d82fc9 or one of its dependencies could not be resolved: Failure to find ch.epfl.scala:maven-bloop_2.10:jar:1.3.2+252-b4d82fc9 in https://maven-central.storage-download.googleapis.com/repos/central/data/ was cached in the local repository, resolution will not be reattempted until the update interval of google-maven-central has elapsed or updates are forced -> [Help 1]
```
@tgodzik Any idea what might be causing this error?

tgodzik:
> The Maven CI is frequently failing with
> 
> ```
> INFO  [ERROR] Plugin ch.epfl.scala:maven-bloop_2.10:1.3.2+252-b4d82fc9 or one of its dependencies could not be resolved: Failure to find ch.epfl.scala:maven-bloop_2.10:jar:1.3.2+252-b4d82fc9 in https://maven-central.storage-download.googleapis.com/repos/central/data/ was cached in the local repository, resolution will not be reattempted until the update interval of google-maven-central has elapsed or updates are forced -> [Help 1]
> ```
> 
> @tgodzik Any idea what might be causing this error?

The bloopMaven plugin's snapshot is not published to maven central, so this will not work until we have a full release.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProxy.scala
```diff
@@ -14,13 +14,13 @@ import scala.meta.internal.metals.debug.DebugProxy._
 private[debug] final class DebugProxy(
     client: RemoteEndpoint,
     server: RemoteEndpoint
-)(
-    implicit ec: ExecutionContext
-) {
+)(implicit ec: ExecutionContext) {
   private val exitStatus = Promise[ExitStatus]()
   @volatile private var outputTerminated = false
+  @volatile private var cancelled = false
```
olafurpg: Can we use `AtomicBoolean` instead?

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -74,9 +71,20 @@ case class BuildServerConnection(
...
+    register(server.buildTargetScalaMainClasses(params))
+  }
+
+  def testClasses(
+      params: ScalaTestClassesParams
+  ): CompletableFuture[ScalaTestClassesResult] = {
+    register(server.buildTargetScalaTestClasses(params))
+  }
+
+  def startDebugSession(params: DebugSessionParams): CompletableFuture[URI] = {
```
olafurpg: Would it make sense to add a guard here to make sure we only send this request to an up-to-date Bloop server that supports debugging? I wonder if we should introduce a debug capability to BSP during the `build/initialize` handshake 🤔

marek1840: What do you mean by "up to date Bloop server"?
I think it would be much simpler to just let the bsp server tell us by responding"I don't support debugging because: A or B or C" (which already happens, for example when bloop runs on jre and not on jdk). The cost is relatively low, because running is a seldom triggered action. 

Of course, in principle you are right, the debug capability should also be included in the capabilities of the server but in this case the gain would be minuscule

olafurpg: I see, thanks for the explanation. For now this is fine. It would be nice to introduce a "debug capability" in BSP so that we only send debug requests to build servers that support it. Currently, Metals even ignores the "compile capability".

marek1840:
> Would it be possible to somehow get this information from the build tool whether a test suite is an object or a class?

Bloop is creating the list of test frameworks using `isModule` flag, so it could be propagated.
Currently, we use:
```scala
trait ScalaTestClassesItem {
  /** The build target that contains the test classes. */
  def target: BuildTargetIdentifier

  /** The fully qualified names of the test classes in this target */
  def classes: List[String]
}
```

we could use:
```scala
trait ScalaTestClassesItem {
  /** The build target that contains the test classes. */
  def target: BuildTargetIdentifier

  /** The fully qualified names of the test classes in this target */
  def classes: List[String]
  def modules: List[String]
}
```
or something like that?

marek1840:
Alternatively, we could put the code lenses on both classes and objects without distinguishing them. Even if as a temporary solution only.

olafurpg:
> Alternatively, we could put the code lenses on both classes and objects without distinguishing them. Even if as a temporary solution only.

This is an acceptable workaround to unblock this PR. We can follow up by opening an issue in either BSP or Bloop to discuss how to extract the `object` test classes

jvican:
@marek1840 Does restarting a debug session send a compile request to Bloop?

marek1840:
> @marek1840 Does restarting a debug session send a compile request to Bloop?

Metals start completely new debug session when receiving a restart request.This, in turn, [compiles projects](https://github.com/scalacenter/bloop/blob/master/frontend/src/main/scala/bloop/bsp/BloopBspServices.scala#L511).

jvican:
Sounds good, excited to try this out :+1:

PR #944, opened 2019-09-24
Replace mvn and mvn.cmd script with jar invocation

tgodzik:
Previously we would use a wrapper script for running mvn commands and now we only use the wrapper jar


olafurpg:
Awesome! Thank you :pray:

Issue #878, opened 2019-08-25
Re-enable Windows CI for pull requests


olafurpg:
I just disabled Appveyor CI for pull requests because it's repeatedly failing with out of memory errors. It's a bad contributing experience when the CI fails for valid pull requests. Appveyor will continue to run against merged pull requests, making it possible to still manually check once in a while if master works on Windows.

We should try out the upcoming GitHub CI once it's out to see if provides more powerful machines to run our test suites.

PR #947, opened 2019-09-26
Replace Azure pipelines with GitHub Actions

olafurpg:
On GitHub Actions we have OpenJDK 8 installed, which doesn't have issues with the nuprocess library that we're using. The tests in https://github.com/scalameta/metals/pull/942#issuecomment-535468094 are failing because Azure uses Zulu JDK which doesn't support the same JNA APIs that are used by nuprocess.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,17 +1,11 @@
 name: CI
-
-on: [push]
-
+on:
+  push:
+  pull_request:
 jobs:
   build:
-
     runs-on: ubuntu-latest
-    
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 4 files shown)


olafurpg:
Yeah, it would be lovely to re-enable the windows tests.

tgodzik:
> Yeah, it would be lovely to re-enable the windows tests.

Somehow I thought these were the windows tests :O But could we also run with Github Actions on windows?

olafurpg:
We haven't been running windows tests since https://github.com/scalameta/metals/issues/878 because Appveyor was too unreliable. GitHub Actions supports Windows so it's worth a shot to try again.

PR #948, opened 2019-09-27
Prioritize build targets with supported Scala versions.

olafurpg:
Previously, if two build targets in the repo were eligible for a source
file, Metals would sometimes prioritize a target with an unsupported
Scala version over a target with a supported Scala version. This
resulted in completions/hover/parameterHints not to work while editing
cross-built sources.

This PR introduces a change to the logic where we select a build target
for a given source file. Now we sort the elibible candidates by two
criteria: the Scala version and the platform (JVM/JS/Native). This
change takes the first step towards
https://github.com/scalameta/metals-feature-requests/issues/13, which is
about making this logic configurable.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -7,6 +7,7 @@ import ch.epfl.scala.bsp4j.BuildTarget
 import ch.epfl.scala.bsp4j.BuildTargetIdentifier
 import ch.epfl.scala.bsp4j.ScalacOptionsItem
 import ch.epfl.scala.bsp4j.ScalacOptionsResult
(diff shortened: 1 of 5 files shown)


metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -188,11 +193,25 @@ final class BuildTargets() {
...
+        .flatMap(_.dependencySources.getBuildTarget(source))
+        .orElse(inferBuildTarget(source))
+    } else {
+      Some(buildTargets.maxBy { t =>
+        var score = 1
+
+        val isSupportedScalaVersion = scalaInfo(t).exists(
+          t => ScalaVersions.isSupportedScalaVersion(t.getScalaVersion())
+        )
+        if (isSupportedScalaVersion) score <<= 2
```
jvican: Nice👌

PR #949, opened 2019-09-27
Don't propose completions for null or nothing

tgodzik:
Found it being an issue after adding subtyping.


build.sbt
```diff
@@ -1,4 +1,4 @@
-def localSnapshotVersion = "0.7.6-SNAPSHOT"
+def localSnapshotVersion = "0.7.7-SNAPSHOT"
```
olafurpg: I was just gonna open a PR to do this ❤️

PR #950, opened 2019-09-27
Don't file watch inside jar files.

olafurpg:
The Scala compiler supports a feature where it compiles the output
classfiles directly to `*.jar` files. This feature is used in build
tools like Pants and Bazel.

Previously, Metals would crash on build targets that used this feature
of the compiler. Now, we don't do anything even if it means missing file
watching notifications that are needed for the "find symbol references"
index.

Thank you @illicitonion for reporting this issue!


olafurpg:
We can file watch the jar itself, and reindex all the SemanticDB files on every change. We can do that in a separate PR, this change is just to fix the current crash.

olafurpg:
Ignoring independent flaky tests in both GH and Travis.

PR #957, opened 2019-10-03
Warn if bloop version is older than 1.3.3

tgodzik:


metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -287,6 +287,27 @@ class Messages(icons: Icons) {
...
+        version: String,
+        displayName: String
+    ): Option[MessageParams] = {
+      val minimumBloopVersion = "1.3.3"
+      if (!BuildTool.isCompatibleVersion(
+          minimumBloopVersion,
+          version
+        ) && displayName.contains("bloop")) {
+        val message =
+          s"""|Bloop server version is too old, some functionality might not work if you reimport the build.
```
olafurpg:
```suggestion
          s"""|The installed Bloop server version is $oldVersion while Metals requires at least Bloop version $requiredMinimumVersion, some functionality might not work. To fix this problem, ...
```

gabro:
Maybe we can add this to the doctor view?

tgodzik:
> Maybe we can add this to the doctor view?

It's now reported as a recommendation is semanticDB options are not detected. Should work better with anything that is not reimported or snapshot 1.3.2 bloop versions.

PR #959, opened 2019-10-04
goto-type-definition

sswistun-vl:
Added goto-type-definition


metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -154,6 +155,26 @@ class Compilers(
         .asScala
     }.getOrElse(Future.successful(new CompletionList()))
 
+  def getLocation(
```
tgodzik: Let's maybe keep to the current naming convention: `getLocation` -> `typeDefinition`

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
```diff
@@ -57,6 +52,11 @@
      */
     public abstract CompletableFuture<Optional<Hover>> hover(OffsetParams params);
 
+    /**
+     * Returns location of the expression's type definition at the given position.
+     */
+    public abstract CompletableFuture<Optional<Location>> getLocation(OffsetParams params);
```
tgodzik: `getLocation` -> `typeDefinition`

mtags/src/main/scala/scala/meta/internal/pc/HoverProvider.scala
```diff
@@ -163,6 +211,7 @@ class HoverProvider(val compiler: MetalsGlobal, params: OffsetParams) {
       pos: Position,
       range: Position
   ): Option[Hover] = {
+    tpe.typeSymbol
```
tgodzik: Remove?

mtags/src/main/scala/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -110,6 +111,19 @@ case class ScalaPresentationCompiler(
       new SignatureHelpProvider(global).signatureHelp(params)
     }
 
+  def getLocation(
```
tgodzik: getLocation -> typeDefinition

mtags/src/main/scala/scala/meta/internal/pc/HoverProvider.scala
```diff
@@ -11,6 +12,53 @@ import scala.meta.internal.jdk.CollectionConverters._
 class HoverProvider(val compiler: MetalsGlobal, params: OffsetParams) {
   import compiler._
 
+  def getTree: Option[Tree] = {
```
tgodzik: This could be used in the hover function too now.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -901,6 +901,36 @@ class MetalsLanguageServer(
...
+  def typeDefinition(
+      params: TextDocumentPositionParams,
+      token: CancelToken
+  ): Future[java.util.List[Location]] = {
+    compilers
+      .getLocation(params, token, interactiveSemanticdbs)
+      .flatMap {
+        case Some(loc) =>
+          Future(List(loc).asJava)
+        case None =>
```
tgodzik: We can remove the None branch now since we do everything in the presentation compile.

sswistun-vl: Sometimes presentation compiler failed to correctly resolve type so I use definitionOrReferences as fallback method.

tgodzik: We can't use `definitionOrReferences` since that can give us references when we are at definition. So at value definition it will find all references to it and that will produce some random behaviour. 

If we cannot resolve type and we are not at the definition that can also be done in the presentation compiler.

tgodzik:
Let's also fix the commit list. It would be best to just have commits that are relevant to the PR, so most likely we need to do an another branch:

```
git checkout master
git checkout -b new-branch-name
git cherry-pick 4bcf8cb2e001886f8b1dbe683d39a6c730d681dc
```

And whenever you need master do it using `git rebase master`.
I will close this one and you can reopen after adding some tests and creating a new branch.

PR #962, opened 2019-10-05
Fall back to mtags position occurrences

### LINKED ISSUE #961: Go to implementation does not work in presence of macro annotations
**Describe the bug**
When trying to find implementations of a trait that has macro annotations, Metals won't return any result.

Removing the macro annotation fixes the issue.

**To Reproduce**

Given a sealed family such as:

```scala
sealed trait MyError
object MyError {
  case object Error1 extends MyError
  case object Error2 extends MyError
}
```

trying to find the implementations of `MyError` works correctly.

Given the same family, but with a macro annotation, it does not work:

```scala
import io.circe.generic.JsonCodec

@JsonCodec sealed trait MyError
object MyError {
  case object Error1 extends MyError
  case object Error2 extends MyError
}
```

**Expected behavior**
Go to implementation works in both cases

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.7.6+71-570c5065-SNAPSHOT

**Search terms**
go to implementation, macro annotations, sealed trait, adt

gabro:
Fixes #961

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
@@ -26,6 +26,7 @@ import scala.meta.internal.semanticdb.TypeSignature
 import scala.collection.mutable
 import scala.meta.internal.symtab.GlobalSymbolTable
 import scala.util.control.NonFatal
+import scala.meta.internal.mtags.Mtags
 
 final class ImplementationProvider(
     semanticdbs: Semanticdbs,
@@ -90,7 +91,13 @@ final class ImplementationProvider(
         params,
         currentDocument
       )
-      symbolOccurrence <- positionOccurrence.occurrence.toIterable
+      symbolOccurrence <- {
+        lazy val mtagsOccurrence = Mtags
+          .allToplevels(source.toInput)
+          .occurrences
+          .find(_.encloses(params.getPosition))
+        positionOccurrence.occurrence.orElse(mtagsOccurrence).toIterable
+      }
     } yield {
       // 1. Search locally for symbol
       // 2. Search inside workspace
--- a/metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
@@ -196,7 +196,7 @@ final class ImplementationProvider(
         classContext,
         implReal
       )
-      implOccurence <- findDefOccurence(implDocument, implSymbol)
+      implOccurence <- findDefOccurence(implDocument, implSymbol, source)
       range <- implOccurence.range
       revised <- distance.toRevised(range.toLSP)
     } yield new Location(file.toUri.toString, revised)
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -88,10 +88,20 @@ final class DefinitionProvider(
...
+      occurrence <- snapshot.occurrences
+        .find(
+          _.encloses(queryPosition, includeLastCharacter)
+        )
+        .orElse(
+          Mtags
+            .allToplevels(source.toInput)
+            .occurrences
+            .find(
+              _.encloses(queryPosition, includeLastCharacter)
```
olafurpg: This position doesn’t need to be from edit distance since we parse top levels from the latest text contents of the file.

gabro:
As discussed offline, I've moved the fallback to `ImplementationProvider` only, so that it's not in the hot path for definitions.

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -90,7 +91,13 @@ final class ImplementationProvider(
         params,
         currentDocument
       )
-      symbolOccurrence <- positionOccurrence.occurrence.toIterable
+      symbolOccurrence <- {
```
tgodzik: Later when we look for the implementation we load semanticDB for the file and find occurence in `textDocument.occurences`, can the same situation happen there? If the annotation is on the implementing class?

gabro: ah, I think so. The occurrence of the sealed trait is not a "definition" anymore in that case (it's a "reference" IIRC). Can you think of a test case for stressing it?

tgodzik: That could be a modified test case in this PR I think:
```
    """|/a/src/main/scala/a/Main.scala
       |package a
       |import io.circe.generic.JsonCodec
       |trait Be@@ing
       |@JsonCodec sealed trait <<Animal>> extends Being
       |object Animal {
       |  case object <<Dog>> extends Animal
       |  case object <<Cat>> extends Animal
       |}
       |""".stripMargin
```

this most likely will not find `Animal`

gabro: you're right, I've added the (now failed) test, so that we can work on this tomorrow.

gabro: ok, I've pushed a fix for that case too. There's indeed a lot of repetion and I hope we can come up with a more harmonized strategy for finding a definition of a symbol in the next days.

tgodzik: Thanks @gabro !

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -90,7 +91,13 @@ final class ImplementationProvider(
...
+        lazy val mtagsOccurrence = Mtags
```
olafurpg:
```suggestion
        def mtagsOccurrence = Mtags
```

PR #970, opened 2019-10-07
Update Bloop to 1.3.4

tgodzik:


tgodzik:
I also remembered that we can now upgrade to 1.3.0 and hoping that can fix the tests :confused:

tests/slow/src/test/scala/tests/gradle/GradleSlowSuite.scala
```diff
@@ -227,79 +227,6 @@ object GradleSlowSuite extends BaseImportSuite("gradle-import") {
         |""".stripMargin
   }
 
-  testAsync("different-scala") {
```
tgodzik: We have the same test in sbt so I don't think it's super needed.

tests/slow/src/test/scala/tests/sbt/SbtSlowSuite.scala
```diff
@@ -200,19 +200,15 @@ object SbtSlowSuite extends BaseImportSuite("sbt-import") {
       _ <- server.initialize(
         s"""
            |/project/build.properties
-           |sbt.version=1.2.6
+           |sbt.version=1.3.0
            |/build.sbt
            |scalaVersion := "2.12.10"
-           |lazy val a = project.settings(scalaVersion := "2.12.4")
```
tgodzik: removed 2.12.4 since it seemed to get stuck on it - checking what happens if we remove it

tgodzik:
There is a deadlock currently in 1.3.4, we'll need to upgrade a next version.

PR #973, opened 2019-10-08
Tone down "No SemanticDB" warning

olafurpg:
Previously, we reported a warning "No SemanticDB" in the status bar when
Metals couldn't find a SemanticDB file for a given source file. Now, we
only log the warning without updating the status bar because this
problem is quite harmless in most situations.

It's difficult to say what the root cause of this problem, my theory is
that it happens in some of the following cases:

- new Bloop BSP connection with an empty class directory
- git checkout to a new branch where new files got added and the file
  hasn't been compiled yet.

Either way, it's not the end of the world when the SemanticDB file
doesn't exist. Most functionality including completions, type at
point,parameter hints and goto definition should continue to work.

cc/ @bishabosha


bishabosha:
LGTM

olafurpg:
Ignoring unrelated CI failure, see https://github.com/scalameta/metals/issues/975

PR #978, opened 2019-10-10
Implement textDocument/typeDefinition

sswistun-vl:
Before, the textDocument/typeDefinition wasn't supported. Now, Metals provides type definitions for variables, method invocations etc.


metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -154,6 +155,26 @@ class Compilers(
         .asScala
     }.getOrElse(Future.successful(new CompletionList()))
 
+  def typeDefinition(
+      params: TextDocumentPositionParams,
+      token: CancelToken,
+      interactiveSemanticdbs: InteractiveSemanticdbs
+  ): Future[Option[Location]] =
+    withPC(params, Some(interactiveSemanticdbs)) { (pc, pos) =>
+      pc.getLocation(
```
tgodzik: `getLocation` -> `typeDefinition`

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -154,6 +155,26 @@ class Compilers(
...
+          CompilerOffsetParams(
+            pos.input.syntax,
+            pos.input.text,
+            pos.start,
+            token
+          )
+        )
+        .asScala
```
tgodzik: Can we move those inside the method and return scala types from it.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -666,6 +666,45 @@ final class TestingServer(
...
+  }
+
+  def typeDefinition(
+      filename: String,
+      query: String
+  ): Future[List[l.Location]] = {
+
+    for {
+      (_, params) <- offsetParams(filename, query, workspace)
+      typeDefinitions <- server
```
tgodzik: I would change it to `definitions` really, because it can be mixed up with the method name.

tests/unit/src/test/scala/tests/TypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,146 @@
...
+import scala.concurrent.duration.Duration
+import scala.meta.internal.metals.MetalsServerConfig
+import scala.meta.internal.metals.StatisticsConfig
+
+object TypeDefinitionLspSuite
+    extends BaseLspSuite("typeDefinition")
+    with TestHovers {
+  override def serverConfig: MetalsServerConfig =
+    super.serverConfig.copy(
+      statistics = new StatisticsConfig("diagnostics")
```
tgodzik: Is this needed?

tests/unit/src/test/scala/tests/TypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,146 @@
...
+
+import org.eclipse.{lsp4j => l}
+import scala.concurrent.Future
+import scala.concurrent.duration.Duration
+import scala.meta.internal.metals.MetalsServerConfig
+import scala.meta.internal.metals.StatisticsConfig
+
+object TypeDefinitionLspSuite
+    extends BaseLspSuite("typeDefinition")
+    with TestHovers {
```
tgodzik: Is `TestHovers` needed?

tests/unit/src/test/scala/tests/TypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,146 @@
...
+          |}""".stripMargin,
+        locationFromString(".metals/readonly/scala/Int.scala/\\25:21/\\25:24")
+      )
+    } yield ()
+  }
+
+  testAsync("basic") {
+    cleanWorkspace()
+    for {
+      _ <- server.initialize(
```
tgodzik: We can move initialize to separate helper method like in ImplementationProvider.

tests/unit/src/test/scala/tests/TypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,146 @@
...
+      _ <- server.didOpen("a/src/main/scala/a/Main.scala")
+      _ = server.workspaceDefinitions
+      _ <- server.assertTypeDefinition(
+        "a/src/main/scala/a/Main.scala",
+        """
+          |package a
+          |class Main(i: Int) {}
+          |
+          |object Main extends App {
+          |  def foo(mn: Main): Unit = {
```
tgodzik: Let's add a test with cursor on a invocation `foo` - it should go to the method definition.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+import org.eclipse.{lsp4j => l}
+import scala.meta.internal.mtags.MtagsEnrichments._
+import scala.meta.pc.OffsetParams
+import scala.meta.internal.jdk.CollectionConverters._
+import scala.meta.internal.metals.CompilerOffsetParams
+
+class TypeDefinitionProvider(val compiler: MetalsGlobal, params: OffsetParams) {
+  import compiler._
+
+  def typedTree: Option[(RichCompilationUnit, Position, Tree)] = {
```
tgodzik: Do we need all 3 as return values?  Those can be created separately when needed and just have the Tree returned here.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+    gtree match {
+      case Apply(fun, List(_: Function)) => isForSynthetic(fun)
+      case TypeApply(fun, _) => isForSynthetic(fun)
+      case gtree: Select if isForComprehensionSyntheticName(gtree) => true
+      case _ => false
+    }
+  }
+
+  private def typedHoverTreeAt(pos: Position): Tree = {
+    val typedTree = typedTreeAt(pos)
```
tgodzik: This conflicts a bit with the method named `typedParams`, maybe rename?

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+      val pos = unit.position(params.offset())
+      val tree = typedHoverTreeAt(pos)
+      Some(unit, pos, tree)
+    }
+  }
+
+  def typeSymbol: Option[(RichCompilationUnit, Position, Symbol)] = {
+    val tri = typedTree
+    tri match {
+      case Some((t1, t2, tree))
```
tgodzik: `t1` and `t2` are not really used here, we can create them later.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+          if tree.symbol.isTypeSymbol || tree.symbol.isMethod || tree.symbol.isConstructor =>
+        Some(t1, t2, tree.symbol)
+      case Some((t1, t2, tree)) if tree.tpe.isDefined =>
```
tgodzik: I tihnk using simple `if/else` might be simpler here. We can `flatMap` over typedTree itself.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+        val expTree = expandRangeToEnclosingApply(tree.pos)
+        if (expTree.tpe.isDefined) Some(t1, t2, expTree.tpe.typeSymbol)
+        else None
+      case _ => None
+    }
+  }
+
+  def typeDefinition: List[l.Location] = {
+    typeSymbol match {
+      case Some((_, _, sym))
```
tgodzik: Would just flatMap over typeDefinition result.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+          if sym.pos.source.file != null && !sym.pos.source.file.isVirtual =>
+        val pos = sym.pos
+
+        val src = scala.io.Source.fromFile(sym.pos.source.path)
```
tgodzik: Use `Files.readAllLines()`, it's more reliant.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+            typedTreeAt(value.pos)
+          case None =>
+            head
+        }
+      case _ =>
+        EmptyTree
+    }
+  }
+
+  lazy val isForName: Set[compiler.Name] = Set[Name](
```
tgodzik: This should be on top of the file since it's a value

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+            }
+          case None =>
+            List(new l.Location(pos.source.path, pos.toLSP))
+        }
+      case _ => Nil
+    }
+  }
+
+  def expandRangeToEnclosingApply(pos: Position): Tree = {
+    def tryTail(enclosing: List[Tree]): Option[Tree] = enclosing match {
```
tgodzik: Try to do it `@tailrec` maybe?

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,152 @@
...
+    }
+  }
+
+  lazy val isForName: Set[compiler.Name] = Set[Name](
+    nme.map,
+    nme.withFilter,
+    nme.flatMap,
+    nme.foreach
+  )
+  def isForSynthetic(gtree: Tree): Boolean = {
```
tgodzik: These are copied from HoverProvider - can we just reuse those methods? Or some of them may be moved to an object.

tests/cross/src/test/scala/tests/pc/TypeDefinitionSuite.scala
```diff
@@ -0,0 +1,42 @@
...
+      |
+      |object Main {
+      | def tst(m: Main): Unit = {}
+      |
+      |  tst(new M@@ain(2))
+      |}""".stripMargin,
+    "Main.scala/\\2:6/\\2:9"
+  )
+
+  check(
```
tgodzik: Maybe add tests additionaly for:
-  named parameter -> it should go to the parameter inside the method
-  on pattern match - it should go to corresponding unapply if it exists otherwise to the object

Anything more you can come up with while testing.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
```diff
@@ -0,0 +1,83 @@
...
+    s"|$uri/\\${start.getLine}:${start.getCharacter}/\\${end.getLine}:${end.getCharacter}|"
+  }
+
+  def check(
+      name: String,
+      code: String,
+      expected: String,
+      duration: Duration = Duration("3 min")
+  ): Unit = {
+    val filename = {
```
tgodzik: I don't think the filename here matters - you can set it explicitly to be a default value of `Main.scala`

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
```diff
@@ -0,0 +1,83 @@
...
+            paramsT._1,
+            paramsT._2
+          )
+        )
+        .get
+        .asScala
+        .toList match {
+        case h :: _ =>
+          val loc = locationToString(h)
+            .dropWhile(_ == '|')
```
tgodzik: Do we need to do this for both? Maybe just compare raw ones.

gabro:
@sswistun-vl when you have time, you can rebase this on master to get the updated CI going (so far it's still running the old one)

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
sswistun-vl: What do you mean by that?

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,80 @@
...
+      None
+    } else {
+      val unit = addCompilationUnit(
+        code = params.text(),
+        filename = params.filename(),
+        cursor = None
+      )
+      val pos = unit.position(params.offset())
+      val tree = typedHoverTreeAt(pos)
+      Some(tree.asInstanceOf[compiler.Tree])
```
tgodzik: Do we need to cast it still?

tests/unit/src/test/scala/tests/TypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,163 @@
...
+  )
+
+  check(
+    "pattern-match",
+    """
+      |case class CClass(str: String) {}
+      |
+      |object Main {
+      |  CClass("test") match {
+      |    case CClass(st@@r) =>
```
tgodzik: Let's add tests also for unapply in case it is defined and in case it isn't.

tests/unit/src/test/scala/tests/TypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,163 @@
+package tests
```
tgodzik: Let's also add tests when multiple files are involved.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  def typeDefinition: List[l.Location] = {
+    typeSymbol.map {
+      case sym
+          if sym.pos.source.file != null && !sym.pos.source.file.isVirtual =>
+        val pos = sym.pos
+
+        val src = Files.readAllLines(sym.pos.source.file.file.toPath).asScala
+
+        val nParams = CompilerOffsetParams(
```
marek1840: what are `nParams` ? My initial intuition was that the `n` stands for `number` (as in "amount")

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,80 @@
...
+          pos.source.path,
+          src.mkString,
+          pos.start
+        )
+
+        val provider = new PcDefinitionProvider(compiler, nParams)
+        val res = provider.definition()
```
marek1840: `res` is a redundant variable, let's rename it to `val locations = provider.definition().locations()`

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,80 @@
...
+        val unit = unitOfFile(pos.source.file)
+        unit.body.find(_.symbol == sym) match {
+          case Some(value) =>
+            value match {
+              case t: DefTree =>
+                List(new l.Location(pos.source.path, t.namePos.toLSP))
+            }
+          case None =>
+            List(new l.Location(pos.source.path, pos.toLSP))
+        }
```
marek1840:
```scala
val some_sensible_name = unit.body
  .find(_.symbol == sym) 
  .collect { case t: DefTree => t.namePos }
  .getOrElse(pos)

List(new l.Location(pos.source.path, some_sensible_name.toLSP)
```

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+    new l.Location(uri, range)
+  }
+
+  def locationToString(loc: l.Location): String = {
+    val (uri, start, end) =
+      (loc.getUri, loc.getRange.getStart, loc.getRange.getEnd)
+    s"|$uri/\\${start.getLine}:${start.getCharacter}/\\${end.getLine}:${end.getCharacter}|"
+  }
+
+  def check(
```
marek1840: we can change it a bit to make the call site a bit nicer: 
```scala 
def check(name:String)(code:String, expected:String,duration:Duration=(...))
```

now, we call it like:
```scala
check("foo")(
  code = ???,
  expected = ???
)
```

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  def locationFromString(str: String): l.Location = {
+    def positionFromString(str: String): l.Position = {
+      val ints = str.split(":").map(_.toInt)
+      new l.Position(ints.head, ints.last)
+    }
+
+    val split = str.split("""/\\""")
+    assertEquals(split.length, 3, "Incorrect location string format")
```
marek1840: print the actual content and maybe a hint on why it is incorrent

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+  def check(
+      name: String,
+      code: String,
+      expected: String,
+      duration: Duration = Duration("3 min")
+  ): Unit = {
+    val filename = "Main.scala"
+
+    val raw = "package a\n" + code
+    val paramsT = params(raw, filename)
```
marek1840: what does the `T` stand for ?

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+      pc.typeDefinition(
+          CompilerOffsetParams(
+            filename,
+            paramsT._1,
+            paramsT._2
+          )
+        )
+        .get
+        .asScala
+        .toList match {
```
marek1840: extract to a variable

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+        case h :: _ =>
+          assertEquals(locationToString(h), expected)
+        case _ =>
+          fail("typeDefinition didn't produce any results")
```
marek1840: maybe also print what was expected?

tests/cross/src/test/scala/tests/pc/TypeDefinitionSuite.scala
```diff
@@ -0,0 +1,42 @@
...
+    "basic",
+    """
+      |class Main(i: Int) {
+      |
+      |}
+      |
+      |object Main {
+      |  val ts@@t = new Main(2)
+      |}""".stripMargin,
+    "|Main.scala/\\2:6/\\2:9|"
```
marek1840: maybe something more readable like `Main.scala [2:6 -> 2:9]` or some similar format?

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -700,6 +700,47 @@ final class TestingServer(
...
+    def locationToString(loc: l.Location): String = {
+      val (uri, start, end) =
+        (loc.getUri, loc.getRange.getStart, loc.getRange.getEnd)
+      s"|$uri/\\${start.getLine}:${start.getCharacter}/\\${end.getLine}:${end.getCharacter}|"
+    }
+
+    for {
+      typeDefinitions <- typeDefinition(filename, query)
+    } yield {
+      for (location <- typeDefinitions)
```
marek1840: why `for` inside a `for`?

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
```diff
@@ -57,6 +56,11 @@
      */
     public abstract CompletableFuture<Optional<Hover>> hover(OffsetParams params);
 
+    /**
+     * Returns location of the expression's type definition at the given position.
```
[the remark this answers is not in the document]
sswistun-vl: It does, but typeDefinition in MetalsLanguageServer returns list, so the one in PresentationCompiler return the same type.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
sswistun-vl: It stands for tuple.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
sswistun-vl: assertEquals prints both expected and obtained results in case of fail.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,80 @@
...
+        cursor = None
+      )
+      val pos = unit.position(params.offset())
+      val tree = typedHoverTreeAt(pos)
+      Some(tree.asInstanceOf[compiler.Tree])
+    }
+  }
+
+  def typeSymbol: Option[Symbol] = {
+    val t = typedTree
```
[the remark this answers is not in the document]
sswistun-vl: Fixed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,80 @@
...
+        if (expTree.tpe.isDefined) Some(expTree.tpe.typeSymbol)
+        else None
+      case _ => None
+    }
+  }
+
+  def typeDefinition: List[l.Location] = {
+    typeSymbol.map {
+      case sym
+          if sym.pos.source.file != null && !sym.pos.source.file.isVirtual =>
```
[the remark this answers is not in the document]
sswistun-vl: Fixed.

tests/unit/src/test/scala/tests/TypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,163 @@
...
+  )(run: => Future[Unit]): Unit = {
+    if (BaseSuite.isWindows) {
+      // src.zip is missing on Appveyor which breaks definition tests.
+      ignore(name) {}
+    } else {
+      super.testAsync(name, maxDuration)(run)
+    }
+  }
+
+  def check(name: String, testCase: String, expected: String): Unit = {
```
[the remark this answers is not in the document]
sswistun-vl: Fixed.

tests/unit/src/test/scala/tests/TypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,163 @@
...
+  override def serverConfig: MetalsServerConfig =
+    super.serverConfig.copy(
+      statistics = new StatisticsConfig("diagnostics")
+    )
+
+  override def testAsync(
+      name: String,
+      maxDuration: Duration = Duration("3min")
+  )(run: => Future[Unit]): Unit = {
+    if (BaseSuite.isWindows) {
```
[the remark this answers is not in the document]
sswistun-vl: Fixed.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -156,6 +157,27 @@ class Compilers(
         .asScala
     }.getOrElse(Future.successful(new CompletionList()))
 
+  def typeDefinition(
+      params: TextDocumentPositionParams,
+      token: CancelToken,
+      interactiveSemanticdbs: InteractiveSemanticdbs
+  ): Future[List[Location]] =
+    withPC(params, Some(interactiveSemanticdbs)) { (pc, pos) =>
```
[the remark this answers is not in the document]
sswistun-vl: Fixed.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -156,6 +157,27 @@ class Compilers(
...
+            pos.start,
+            token
+          )
+        )
+        .asScala
+        .map(_.asScala)
+    }.getOrElse {
+        Future.successful(List())
+      }
+      .map(_.toList)
```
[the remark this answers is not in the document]
sswistun-vl: Fixed.

mtags/src/main/scala/scala/meta/internal/pc/ProviderCommon.scala
```diff
@@ -0,0 +1,82 @@
+package scala.meta.internal.pc
+
+trait ProviderCommon {
```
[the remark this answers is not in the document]
sswistun-vl: Fixed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
sswistun-vl: Fixed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
sswistun-vl: Fixed.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
sswistun-vl: Fixed.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
sswistun-vl: Fixed.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
sswistun-vl: Fixed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
sswistun-vl: Fixed.

mtags/src/main/scala/scala/meta/internal/pc/ProviderCommon.scala
```diff
@@ -0,0 +1,82 @@
...
+  val compiler: MetalsGlobal
+  import compiler._
```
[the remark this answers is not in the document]
sswistun-vl: I've tried to keep it explicit, but it causes type conflicts.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
marek1840: Does it work for `fail` too?

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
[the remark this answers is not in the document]
marek1840: If it *always* returns single location, the the type should reflect that. If a list is needed at the call site, then it can be wrapped there

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,76 @@
...
+      case sym if !isFileVirtual(sym) =>
+        val pos = sym.pos
+
+        val src = Files.readAllLines(sym.pos.source.file.file.toPath).asScala
+
+        val params = CompilerOffsetParams(
+          pos.source.path,
+          src.mkString,
+          pos.start
+        )
```
marek1840:
```scala
  val content = Files.readAllBytes(sym.pos.source.file.file.toPath)`
  val params = CompilerOffsetParams(
    pos.source.path,
    new String(content),
    pos.start
)

```

tests/cross/src/test/scala/tests/pc/TypeDefinitionSuite.scala
```diff
@@ -0,0 +1,39 @@
+package tests.pc
+
+object TypeDefinitionSuite extends BaseTypeDefinitionSuite {
+
+  check("basic")(
```
marek1840: what is `basic` ? Local variable?

tests/cross/src/test/scala/tests/pc/TypeDefinitionSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+      |
+      |}
+      |
+      |object Main {
+      |  val ts@@t = new Main(2)
+      |}""".stripMargin,
+    "Main.scala [2:6 -> 2:9]"
+  )
+
+  check("method")(
```
marek1840: maybe rename to `constructor` ?
Then, the next case can be renamed to `method`

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -792,6 +792,45 @@ final class TestingServer(
...
+  def typeDefinition(
+      filename: String,
+      query: String
+  ): Future[List[l.Location]] = {
+    for {
+      (_, params) <- offsetParams(filename, query, workspace)
+      definitions <- server
+        .typeDefinition(params)
+        .asScala
+        .map(l => l.asScala.toList)
```
marek1840: this map can be moved to the `yield` clause which would increase the readability as:
a) the server.typeDefinition(...).asScala chain can be put in a single line
b) the result transformations would be in a single place: `yield { definitions.asScala.toList } `

tests/unit/src/test/scala/tests/TypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+
+    str match {
+      case regex(uri, startL, startC, endL, endC) =>
+        createLocation(
+          workspace.toURI.resolve(uri).toString.replaceFirst("/", "///"),
+          (startL.toInt, startC.toInt),
+          (endL.toInt, endC.toInt)
+        )
+      case _ =>
+        fail("Incorrect location string format")
```
marek1840: print the malformed string, so the user doesn't have to guess how it looked like

mtags/src/main/scala/scala/meta/internal/pc/AbstractProvider.scala
```diff
@@ -0,0 +1,82 @@
...
+              case Nil => None
+              case Select(_, _) :: next =>
+                tryTail(next)
+              case _ =>
+                None
+            }
+          case _ =>
+            None
+        }
+    }
```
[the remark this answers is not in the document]
sswistun-vl: Changed.

mtags/src/main/scala/scala/meta/internal/pc/AbstractProvider.scala
```diff
@@ -0,0 +1,82 @@
...
+    lastVisistedParentTrees match {
```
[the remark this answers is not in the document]
sswistun-vl: Fixed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,76 @@
...
+        Some(tree.children.head.tpe.typeSymbol)
+      case tree =>
+        val expTree = expandRangeToEnclosingApply(tree.pos)
+        if (expTree.tpe.isDefined) Some(expTree.tpe.typeSymbol)
+        else None
+      case _ => None
+    }
+  }
+
+  private def isFileVirtual(sym: compiler.Symbol) =
```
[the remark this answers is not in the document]
sswistun-vl: Changed.

tests/cross/src/test/scala/tests/pc/TypeDefinitionSuite.scala
sswistun-vl: Fixed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
sswistun-vl: Renamed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,76 @@
...
+    sym.pos.source.file == null || sym.pos.source.file.isVirtual
+
+  def typeDefinition: List[l.Location] = {
+    typeSymbol.map {
+      case sym if !isFileVirtual(sym) =>
```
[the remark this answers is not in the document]
sswistun-vl: Fixed.

tests/unit/src/test/scala/tests/TypeDefinitionLspSuite.scala
sswistun-vl: Changed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,76 @@
...
+      case tree
+          if tree.symbol.isTypeSymbol || tree.symbol.isMethod || tree.symbol.isConstructor =>
+        Some(tree.symbol)
+      case tree if tree.tpe.isDefined =>
+        Some(tree.tpe.typeSymbol)
+      case tree if tree.children.nonEmpty =>
+        Some(tree.children.head.tpe.typeSymbol)
+      case tree =>
+        val expTree = expandRangeToEnclosingApply(tree.pos)
+        if (expTree.tpe.isDefined) Some(expTree.tpe.typeSymbol)
```
olafurpg: Beware that `expTree.tpe` can be null here

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+import org.eclipse.{lsp4j => l}
+
+class BaseTypeDefinitionSuite extends BasePCSuite {
+
+  override def beforeAll(): Unit = {
+    indexJDK()
+    indexScalaLibrary()
+  }
+
+  def locationFromString(str: String): l.Location = {
```
olafurpg: Let's try to use the same testing infrastructure as "textDocument/definition" instead of coming up with a new expect syntax.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+          filename,
+          paramsT._1,
+          paramsT._2
+        )
+      )
+      .get
+      .asScala
+      .toList
+
+    test(name) {
```
olafurpg: Move this to the top of the method, otherwise one failing test case will block the subsequent test cases from starting.

tests/cross/src/test/scala/tests/pc/TypeDefinitionSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+    """
+      |object Main {
+      |  def tst(): Unit = {}
+      |
+      |  ts@@t()
+      |}""".stripMargin,
+    "Main.scala [3:6 -> 3:8]"
```
olafurpg: I would expect this to go to the definition of `Unit` instead of `tst()`

tests/cross/src/test/scala/tests/pc/TypeDefinitionSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+  )
```
olafurpg: Some test cases to consider
```scala
List(1).hea@@d // should go to `Int.scala` instead of `def head`

class F@@oo(val x: Int) // should go to `Foo`

va@@l x = 42 // should not go anywhere

val x = 4@@2
// can go to Int.scala, but it's OK to not go anywhere either IMO.
// Hover doesn't show types for literals because it gets noisy.

for {
  x <- List(1)
  i@@f x > 1 // I'm not sure what's the expected behavior
} println(x)

"".stripS@@uffix("foo") // should go to Predef.scala

def foo[T](param: T): T = para@@m // should go to `T` type parameter

```

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
olafurpg: In general, I have found it helpful to test "goto definition" by running it on all possible tokens instead of writing individual test cases for custom positions.

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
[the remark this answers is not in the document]
olafurpg: One tricky case where there are multiple type definitions

```scala
def overload(a: Int): Int = a
def overload(a: String): String = a

overloa@@d // should show both `Int.scala` and `String.scala`
```

mtags/src/main/scala/scala/meta/internal/pc/AbstractProvider.scala
```diff
@@ -0,0 +1,71 @@
+package scala.meta.internal.pc
+
+trait AbstractProvider {
```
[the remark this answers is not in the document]
sswistun-vl: Fixed, please tell me if it's what you've meant when you have time.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,76 @@
...
+        else None
+      case _ => None
+    }
+  }
+
+  private def hasVirtualSource(sym: compiler.Symbol) =
+    sym.pos.source.file == null || sym.pos.source.file.isVirtual
```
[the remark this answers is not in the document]
sswistun-vl: Fixed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def typeDefinition: List[l.Location] = {
```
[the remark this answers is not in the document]
sswistun-vl: Changed.

mtags/src/main/scala/scala/meta/internal/pc/AbstractProvider.scala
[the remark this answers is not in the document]
olafurpg: We should remove `AbstractProvider` and move these utility methods into a helper trait named `Trees` that is mixed into the `MetalsGlobal` class the same way as `AutoImports`, `Signatures` and `Types`, see

https://github.com/scalameta/metals/blob/3c09c7a2d91afe4689895594badb71c1195da64f/mtags/src/main/scala/scala/meta/internal/pc/Signatures.scala#L11

https://github.com/scalameta/metals/blob/3c09c7a2d91afe4689895594badb71c1195da64f/mtags/src/main/scala/scala/meta/internal/pc/MetalsGlobal.scala#L33

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,76 @@
...
+import scala.meta.internal.mtags.MtagsEnrichments._
+import scala.meta.pc.OffsetParams
+import scala.meta.internal.jdk.CollectionConverters._
+import scala.meta.internal.metals.CompilerOffsetParams
+
+class TypeDefinitionProvider(val compiler: MetalsGlobal, params: OffsetParams)
+    extends AbstractProvider {
+  import compiler._
+
+  def typedTree: Option[Tree] = {
```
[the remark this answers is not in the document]
sswistun-vl: Changed. I've structured it after HoverProvider, so it may require some work as well.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
sswistun-vl: Could you give me example of suite you have in mind?

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
sswistun-vl: I've added null check there.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
sswistun-vl: I've added it to prevent clashing with params method.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
sswistun-vl: Like that?

sswistun-vl:
olafurpg thanks for this review, I'm going through it right now.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
olafurpg: I was referring to something like `DefinitionLspSuite`

olafurpg: https://github.com/scalameta/metals/blob/cf7b95265a6019a99b0dea323caef47096dd8a58/tests/unit/src/test/scala/tests/DefinitionLspSuite.scala#L59-L78

but it seems to only be available in lsp suites

olafurpg: Instead of using that structure, we can reuse `PcDefinitionSuite`, see https://github.com/scalameta/metals/blob/cf7b95265a6019a99b0dea323caef47096dd8a58/tests/cross/src/test/scala/tests/pc/PcDefinitionSuite.scala#L52

olafurpg: As a general rule of thumb, avoid hardcoding line:column numbers in expect tests. Numbers are harder review than something like "Int.scala"

sswistun-vl: Ok, I see what you have in mind.

mtags/src/main/scala/scala/meta/internal/pc/AbstractProvider.scala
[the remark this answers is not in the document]
sswistun-vl: Ok, I've changed it.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
sswistun-vl: I've changed cross tests for goto-type-definition, now they share code with cross tests for goto-definition.

sswistun-vl:
Unit and cross tests for goto-typeDefinition are suppose to fail right now - there's a problem similar to the one with type arguments in https://github.com/scalameta/scalameta/pull/1910 - compiler doesn't provide information about named arguments.

sswistun-vl:
I've solved the problem with named arguments, but please tell me if this solution is acceptable as it's based on tokens.

tests/cross/src/test/scala/tests/pc/TypeDefinitionSuite.scala
sswistun-vl: Changed.

tests/cross/src/test/scala/tests/pc/TypeDefinitionSuite.scala
sswistun-vl: Shouldn't ```"".stripS@@uffix("foo")``` go to String?

sswistun-vl:
Regarding named arguments - https://github.com/scalameta/scalameta/pull/1919 provides better way to handle occurrences for them.

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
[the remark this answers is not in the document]
sswistun-vl: Unfortunately compiler returns error in such case.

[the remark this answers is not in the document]
tgodzik: I don't think it can show both. If we invoke one function then goto definition will only go to one. The same should happen with type definition.

tests/cross/src/test/scala/tests/pc/TypeDefinitionSuite.scala
tgodzik: I think it should since, that method is not define din Predef.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+          Some(sel.tpe.typeSymbol)
+        case app @ Apply(fun, args) if args.nonEmpty =>
+          val (rStart, rEnd) = (fun.pos.start, args.last.pos.end)
+
+          val txt = app.pos.source.content
+            .slice(rStart, rEnd)
+            .fold("")(_ + _.toString)
+            .toString
+          val tokens = txt.tokenize.get.tokens.toList
+            .filter(t => {
```
tgodzik: Let's just use `find` here.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+              val (pStart, pEnd) =
+                (app.pos.start + t.pos.start, app.pos.start + t.pos.end)
+              pStart <= params.offset && pEnd >= params.offset
```
tgodzik:
```suggestion
              val treeOffset = app.pos.start
              pStart <= params.offset - treeOffset && pEnd >= params.offset - treeOffset
```

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+        case tree: Tree =>
+          val expTree = expandRangeToEnclosingApply(tree.pos)
+          if (expTree.tpe != null && expTree.tpe.isDefined)
+            Some(expTree.tpe.typeSymbol)
+          else None
+        case _ => None
+      }
+  }
+
+  def typeDefinition(params: OffsetParams): List[l.Location] =
```
tgodzik: Add brackets `{}` here maybe

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+      .toList
+      .flatten
+
+  private def getSymbolDefinition(sym: Symbol): List[l.Location] = {
+    val file = sym.pos.source.file
+    val uri =
+      try {
+        file.toURL.toURI.toString
+      } catch {
+        case _: NullPointerException =>
```
tgodzik: Can't you just check for null? The only thing that can throw it is file.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+        if (unit != null) {
+          val trees = unit.body
+            .filter(_.symbol == sym)
+            .filter(_.pos != null)
+            .filter(_.isDef)
+          trees
+            .map(t => new l.Location(uri, t.pos.toLSP))
+        } else fallbackToSemanticDB(sym)
+      } else fallbackToSemanticDB(sym)
+    } catch {
```
tgodzik: There is already a lot of checks, do we need this?

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+    val uri =
+      try {
+        file.toURL.toURI.toString
+      } catch {
+        case _: NullPointerException =>
+          sym.pos.source.path
+      }
+
+    try {
+      if (file != null || compiler.unitOfFile.contains(file)) {
```
tgodzik: I think we can incorporate the changes from https://github.com/scalameta/metals/pull/1174/files - we look for the file containing the symbol there.

tests/cross/src/main/scala/tests/pc/BasePcDefinitionSuite.scala
```diff
@@ -1,59 +1,20 @@
...
 import tests.BasePCSuite
-import scala.meta.internal.metals.CompilerOffsetParams
-import org.eclipse.lsp4j.TextEdit
-import org.eclipse.{lsp4j => l}
-import scala.meta.internal.jdk.CollectionConverters._
-import scala.meta.internal.mtags.MtagsEnrichments._
-import scala.meta.internal.metals.TextEdits
 
 abstract class BasePcDefinitionSuite extends BasePCSuite {
+  val test: (String, String) => (String, String) =
```
tgodzik: Can we rename `test` to something more meaningful? Maybe `runCheck`?

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
```diff
@@ -0,0 +1,27 @@
+package tests.pc
+
+import tests.BasePCSuite
+import scala.concurrent.duration.Duration
+
+class BaseTypeDefinitionSuite extends BasePCSuite {
+  val test: (String, String) => (String, String) =
```
tgodzik: Can we rename this to to something more meaningful?

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
```diff
@@ -0,0 +1,27 @@
+package tests.pc
+
+import tests.BasePCSuite
+import scala.concurrent.duration.Duration
+
+class BaseTypeDefinitionSuite extends BasePCSuite {
```
tgodzik: Doesn't look like we need this as a separate class altogether - it's only utility methods used in one suite.

tests/cross/src/test/scala/tests/pc/TypeDefinitionSuite.scala
```diff
@@ -0,0 +1,102 @@
...
+      |}
+      |""".stripMargin
+  )
+
+  check("if")(
+    """
+      |object Main {
+      | for {
+      |   x <- List(1)
+      |   i/*scala/collection/generic/FilterMonadic# FilterMonadic.scala*/@@f x > 1 // I'm not sure what's the expected behavior
```
tgodzik: Better to comment above the check. I think it's kind of ok? It's how this if is defined.

tests/unit/src/main/scala/tests/BaseTypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,111 @@
+package tests
+
+import scala.meta.internal.metals.TextEdits
+import org.eclipse.{lsp4j => l}
+import scala.concurrent.Future
+import scala.meta.inputs.Input
+import scala.meta.inputs.Position
+import scala.meta.internal.mtags.MtagsEnrichments._
+
+class BaseTypeDefinitionLspSuite extends BaseLspSuite("typeDefinition") {
```
tgodzik: Might not make sense to have a separate base class if this is used in one suite. Better to add those methods as private to the bottom of the suite. I don't imagine this hierachy will be used.

tests/unit/src/main/scala/tests/BaseTypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,111 @@
...
+  def check(
+      name: String
+  )(query: String, expectedLocs: List[String] = Nil): Unit = {
+    testAsync(name) {
+      cleanWorkspace()
+      val code =
+        query
+          .replaceAll("(@@)?(<<)?(>>)?", "")
+          .replaceAll("""\/\*[^\*]+\*\/""", "")
+      val files = query.lines
```
tgodzik: You can check this after running FileLayout. Should be simpler then.

tests/unit/src/main/scala/tests/BaseTypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,111 @@
...
+        .filter(_.startsWith("/"))
+        .filter(_.filter(!_.isWhitespace) != "/metals.json")
```
tgodzik: We can have it separate here by default like in RenameLspSuie

tests/unit/src/main/scala/tests/BaseTypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,111 @@
...
+import org.eclipse.{lsp4j => l}
+import scala.concurrent.Future
+import scala.meta.inputs.Input
+import scala.meta.inputs.Position
+import scala.meta.internal.mtags.MtagsEnrichments._
+
+class BaseTypeDefinitionLspSuite extends BaseLspSuite("typeDefinition") {
+  def check(
+      name: String
+  )(query: String, expectedLocs: List[String] = Nil): Unit = {
```
tgodzik: Is `expectedLocs` used still?

tests/unit/src/main/scala/tests/BaseTypeDefinitionLspSuite.scala
```diff
@@ -0,0 +1,111 @@
...
+        _ = if (!server.server.isInitialized.get())
+          fail("Build tool not initialized")
+      } yield ()
+    }
+  }
+
+  def prepareDefinition(original: String): String =
+    original.replaceAll("(<<)?(>>)?", "")
+
+  def locationsToCode(
```
tgodzik: This is already used in the PCSuite I think? Can we reuse the same code?

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -862,6 +862,103 @@ final class TestingServer(
...
+            content.replaceAll("""/\*[^\*]*\*/""", ""),
+            content.lines.find(_.contains("@@")).collect {
+              case uriRegex(uri) => uri
+            }
+          )
+        case _ => (filenameStr, queryStr, None)
+      }
+    } else (filenameStr, queryStr, None)
+
+    val extractedExp = if (expectedLocs.isEmpty) {
```
tgodzik: Can we move most of the logic from here to the TestSuite - it seems that it can be simplified a lot really.

tests/unit/src/main/scala/tests/BaseTypeDefinitionLspSuite.scala
sswistun-vl: `ext-library` uses it. It makes it possible to specify exact position in another file.

tests/unit/src/main/scala/tests/BaseTypeDefinitionLspSuite.scala
sswistun-vl: `ext-library` uses it. It makes it possible to specify exact position in another file.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+  val ignoredTags: List[String] = List(
+    "val",
+    "var"
+  )
+
+  def tree(params: OffsetParams): Option[Tree] =
+    if (params.isWhitespace)
+      None
+    else {
+      Some(createCompilationUnit(params)._3)
```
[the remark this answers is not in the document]
sswistun-vl: Fixed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
sswistun-vl: Changed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
sswistun-vl: Added.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
sswistun-vl: Changed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+      }
+      .flatMap {
+        case sel: Select
+            if sel.symbol.asMethod.returnType.typeSymbol.isTypeParameter =>
+          Some(sel.tpe.typeSymbol)
+        case app @ Apply(fun, args) if args.nonEmpty =>
+          val (rStart, rEnd) = (fun.pos.start, args.last.pos.end)
+
+          val txt = app.pos.source.content
+            .slice(rStart, rEnd)
```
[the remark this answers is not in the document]
sswistun-vl: Changed.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
sswistun-vl: Changed.

tests/unit/src/main/scala/tests/BaseTypeDefinitionLspSuite.scala
sswistun-vl: I think I removed duplicates of it.

tests/cross/src/main/scala/tests/pc/BasePcDefinitionSuite.scala
sswistun-vl: Changed.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
sswistun-vl: Changed.

tests/cross/src/main/scala/tests/pc/BaseTypeDefinitionSuite.scala
sswistun-vl: Changed.

tests/unit/src/main/scala/tests/BaseTypeDefinitionLspSuite.scala
sswistun-vl: Changed.

tgodzik:
Thanks for the fixes! I will finish up it later after Christmas, since @sswistun-vl  will not have time to work on it now.

tgodzik:
Closing this PR until someone has time to finish the work.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,136 @@
...
+            .find(t => {
+              val (pStart, pEnd) =
+                (app.pos.start + t.pos.start, app.pos.start + t.pos.end)
+              pStart <= params.offset && pEnd >= params.offset
+            }) match {
+            case Some(t) =>
+              app.symbol.asMethod.paramLists.flatten
+                .find(_.nameString.trim == t.text)
+                .map(_.tpe.typeSymbol)
+            case _ => None
```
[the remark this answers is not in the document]
alekseiAlefirov: maybe fixing `typedHoverTreeAt` is possible to untangle this code a bit?
Also, @sswistun-vl  [mentioned](https://github.com/scalameta/metals/pull/978#issuecomment-562541418) solution for problem of named parameter in [semanticDB](https://github.com/scalameta/scalameta/pull/1919). Not sure though, how it can save working with compiler here.

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,136 @@
...
+          }
+        case tree: Tree if tree.symbol.isMethod =>
+          Some(tree.symbol.asMethod.returnType.typeSymbol)
+        case tree: Tree if tree.symbol.isTypeSymbol =>
+          Some(tree.symbol)
+        case tree: Tree if tree.tpe.isDefined =>
+          Some(tree.tpe.typeSymbol)
+        case tree: Tree if tree.children.nonEmpty =>
+          Some(tree.children.head.tpe.typeSymbol)
+        case vd: compiler.ValDef if vd.rhs.tpe != null =>
```
alekseiAlefirov, line 83: `vd.rhs.tpe != null` is not enough here. it should be also not Error and not `NoType`. Seems to be one of the reason `basicMultifile` test fails (for me it does, after rebase).

mtags/src/main/scala/scala/meta/internal/pc/TypeDefinitionProvider.scala
```diff
@@ -0,0 +1,136 @@
...
+      val typeSym = semanticdbSymbol(sym)
+      search.definition(typeSym).asScala.toList
+    }
+  }
+
+  object Xtensions {
+
+    implicit class XtensionPosition(pos: Position) {
+      def includes(point: Int): Boolean = pos.start <= point && pos.end >= point
+    }
```
alekseiAlefirov, line 133: not used

tgodzik:
Just as a side note @alekseiAlefirov will try to finish up this PR and make sure it's all working properly.

kpbochenek:
Curious what should actually `textDocument/typeDefinition` return? 
How will this be represented in VS Code?

e.g. I have code `val tmp: List[String] = ???` and if I execute this command on `tmp` in VS Code what would happen?

tgodzik:
@kpbochenek It should go to the definition of `List`

PR #979, opened 2019-10-10
Add system property to disable project/metals.sbt generation

olafurpg:
This is intended to be used in the Bloop sbt build that uses a custom
way to install the sbt-bloop plugin.


jvican:
Thanks for doing this! LGTM

olafurpg:
Ignoring unrelated CI failure, GitHub actions seems to have been down in the past week :/ maybe we should re-enable azure 🤔

PR #982, opened 2019-10-11
Install bloop server in CI

sswistun-vl:
Previously, the CI tests were run using an embedded Bloop server. Recently the tests started to time in CI. This PR changes the tests in CI to use an installed Bloop server instead of the embedded Bloop server.

The tests are still failing but no longer timing out, which is an improvmement.


.travis.yml
```diff
@@ -9,34 +9,88 @@ stages:
   - name: release
     if: (branch = master AND type = push) OR (tag IS present)
 
+before_install:
+  - mkdir bloop-server
+  - touch -f ~/Library/Caches/org.scalameta.metals/lsp.trace.json
```
tgodzik: You need the linux path, not the mac one:
```
touch ~/.cache/metals/lsp.trace.json # text editor
touch ~/.cache/metals/bsp.trace.json # build server
```

.travis.yml
```diff
@@ -9,34 +9,86 @@ stages:
...
     - env: TEST="scalafmt+scalafix+docs"
-      before_script: nvm install 'lts/*' && nvm use 'lts/*'
+      before_script:
+      - nvm install 'lts/*' && nvm use 'lts/*'
       script:
         - ./bin/scalafmt --test
         - sbt scalafixCheck docs/docusaurusCreateSite
     - env: TEST="sbt 'slow/testOnly -- tests.sbt'"
+      before_script:
+        - wget https://github.com/scalacenter/bloop/releases/download/v1.3.3/install.py
```
marek1840: can we somehow derive the version from the build config?

.gitignore
```diff
@@ -43,3 +43,7 @@ website/yarn.lock
 website/node_modules
 website/i18n/*
 !website/i18n/en.json
+
+project/metals.sbt
+.bloop/*
+.metals/*
```
tgodzik: Let's remove those - that's fixed in latest master

bin/travis/test.sh
```diff
@@ -0,0 +1,17 @@
...
+source tools.sh
+
+bloopDir="bloop-server"
+[ -d "$bloopDir" ] && rm -rf "$bloopDir"
+mkdir "$bloopDir"
+
+bloopVersion=$(get_bloop_version)
+bloopURL="https://github.com/scalacenter/bloop/releases/download/v$bloopVersion/install.py"
+
+wget "$bloopURL"
```
marek1840: no need to call python if wget fails
```suggestion
wget "$bloopURL" && python install.py --dest $bloopDir
```

bin/travis/test.sh
```diff
@@ -0,0 +1,17 @@
...
+python install.py --dest $bloopDir
+
+$bloopDir/bloop ng-stop
```
marek1840: If those are dependant task, we should nto try to execute the second one if first one finishes with errors
```suggestion
$bloopDir/bloop ng-stop && $bloopDir/bloop server &>$bloopDir/bloop.log & 
```

It's worth considering this script to be something like:
```bash
install_bloop(){
  (...)
}
start_bloop(){
  (...)
}

install_bloop && start_bloop && sbt "$1"
```

bin/travis/tools.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
marek1840: there's no need for separate file (for now ;) ). It can be a function in `test.sh` or even a direct variable in `install_bloop` function:
```sh
bloopVersion=$(sed -n 's|val\s*bloop\s*=\s*"\(.*\)"\s*$|\1|p' build.sbt)
```

bin/travis/test.sh
```diff
@@ -0,0 +1,18 @@
...
+
+[ -d "$bloopDir" ] && rm -rf "$bloopDir"
+mkdir "$bloopDir"
+
+bloopURL="https://github.com/scalacenter/bloop/releases/download/v$(bloop_version)/install.py"
+
+wget "$bloopURL" && python install.py --dest $bloopDir
+
+$bloopDir/bloop ng-stop
+$bloopDir/bloop server &>$bloopDir/bloop.log &
```
jvican: You can replace all of this script by coursier launching bloopgun-core, the invocation will start bloop in the background of it isn’t running and then connect to the server and run the nailgun command you specified

bin/travis/test.sh
sswistun-vl: ng-stop only stops server if it's running - it should fail

bin/travis/test.sh
sswistun-vl: Do you have any manual for it at hand? I can't find any.

tgodzik: `coursier launch ch.epfl.scala:bloopgun-core_2.12:1.3.4 -- <command>` will work

bin/travis/test.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+function bloop_version {
+  cat build.sbt | grep "val bloop" | sed 's|[^0-9.]||g'
```
marek1840:
```suggestion
  grep "val bloop" build.sbt | sed 's|[^0-9.]||g'
```

bin/travis/test.sh
```diff
@@ -0,0 +1,12 @@
...
+}
+
+wget -O bin/coursier https://git.io/coursier-cli && chmod +x bin/coursier \
+&& bin/coursier launch ch.epfl.scala:bloopgun-core_2.12:$(bloop_version) -- server &>bloop.log &
```
jvican: You don't need to run `bloop server` here, you can instead run:

`bin/coursier launch ch.epfl.scala:bloopgun-core_2.12:$(bloop_version) -- about`

That invocation will make sure of starting the server in the background if there isn't one running. You'll lose access to the logs but it's unclear why you want to have access to them here, the server logs do not contain any useful information and are mostly noise coming from the nailgun server.

.gitignore
```diff
@@ -44,6 +44,4 @@ website/node_modules
 website/i18n/*
 !website/i18n/en.json
 
-project/metals.sbt
-.bloop/*
-.metals/*
\ No newline at end of file
+project/metals.sbt
```
olafurpg: Should this change be included?

tgodzik: That's fine, we want to have it gitignored

olafurpg:
LGTM 👍 the failing tests are unrelated to this change and apparently @tgodzik is working on a separate fix for them.

PR #985, opened 2019-10-11
Clarify Emacs installation instructions

sswistun-vl:
Corrected errors in installation tutorial

### DIFF
--- a/docs/editors/emacs.md
+++ b/docs/editors/emacs.md
@@ -14,7 +14,7 @@ Metals works in Emacs thanks to the the
 
 ## Installation
 
-To use Metals in Emacs, place this snippet in your Emacs configuration to load
+To use Metals in Emacs, place this snippet in your Emacs configuration (for example .emacs.d/init.el) to load
 `lsp-mode` along with its dependencies:
 
 ```el
@@ -22,6 +22,8 @@ To use Metals in Emacs, place this snippet in your Emacs configuration to load
 
 ;; Add melpa to your packages repositories
 (add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/") t)
+
+(package-initialize)
 
 ;; Install use-package if not already installed
 (unless (package-installed-p 'use-package)


tgodzik:
I asked on gitter and @kschwarz1116  suggested:
```
maybe it makes more sense to provide a link to the MELPA website and describe what to install using it, instead of adding it to the .emacs changes? https://melpa.org/#/getting-started If I were new to emacs and getting started, I think that would be more helpful than just getting a block of code to put somewhere on my machine.
```
I think we can keep the code but let people know about the getting started guide.

olafurpg:
cc/ @JesusMtnez could you take a look?

JesusMtnez:
Current Emacs instructions, with the fixes suggested in this PR, not only suggest how to install metals support for Emacs, but also suggests to use `package.el` package manager and `use-package` helper. But they are not the only options to manage your Emacs packages. 

Other packages documentations usually offer how to configure it using just plain `package.el`([example](https://github.com/bbatsov/projectile/tree/70d13691aa06b9afa4476e9765b88657b70efcbe#installation)), without `use-package`. Usually, they add more documentation to include helpers like `use-package`, `straight.el`, `el-get`,... 

What they don't usually do is provide the `package.el` configuration to add `melpa` or `melpa-stable` repositories, they just indicate the package is available through `melpa` if you want to use it.

So, this changes are necessary to fix `package.el` setup, but maybe we could improve our documentation to avoid _forcing_ users to understand `package.el` and `use-package`.

sswistun-vl:
Main purpose of this PR was to correct tutorial that didn't work for me. In my opinion it's far more important to make it work than to educate someone, but feel free to correct that as I don't have knowledge about Emacs to do it informatively.

JesusMtnez:
IMHO, merging this PR does no harm to the documentation, as @sswistun-vl said, it will fix to make it possible to copy/paste and go. But we may consider to update this configuration documentation to be more complete about how to configure metals for Emacs in the future, in another PR of course.

olafurpg:
> Main purpose of this PR was to correct tutorial that didn't work for me. In my opinion it's far more important to make it work than to educate someone

@sswistun-vl I encourage you to formulate your arguments more carefully in the future. @JesusMtnez did not argue in favor of the opposite point even if you implicitly hint that he thinks "it's more important to educate someone than to make it work". This debate style is not tolerated when contributing to any Scalameta projects.

JesusMtnez:
Just to clarify, I was not trying to say we should educate about how to configure/use Emacs. On the contrary, what I was trying to say was that the current configuration is not offering a way to configure metals using Emacs vanilla with its default package manager, but instead, it is _forcing_ the user to understand _use-package_, which is not a requirement to use metals at all. 

That is why I proposed to change or add to this documentation the process to install, configure and use metals with Emacs vanilla, at least. That way, anyone using other package managers or helpers could adapt it easily if they want to.

@sswistun-vl Sorry if I did not express myself correctly :smile: but I was not trying to force my opinion here at all.

sswistun-vl:
Me neither, I was just saying that we should present working configuration first and then show how to create custom one - that way both fresh and seasoned Emacs user can easily benefit from it :)

PR #986, opened 2019-10-11
(Spac)Emacs continuously crashing due to compilation of swap files

quiquedev:
it seems that the regex of the docu for the scala-mode is too much general for emacs so swap files (starting with .#) are compiled by metals and this one crashes, throwing a runtime exception, when the swap file is not there

this new regex is maybe too restrictive but seems to be valid for most of usecases


quiquedev:
@tgodzik you welcome :) this got me crazy for weeks!

PR #988, opened 2019-10-14
Re-introduce Azure pipelines for more stable CI.

olafurpg:
GitHub Actions seems to be timing out once-in-a-while. This commit adds
back Azure Pipelines to run the same tests in parallel with Github Actions.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,6 @@
+jobs:
+- job: Linux
+  pool:
+    vmImage: 'Ubuntu-16.04'
+  steps:
+  - script: ./sbt unit/test


tgodzik:
Looks like Azure is having issues too :/

tgodzik:
Seems neither Azure nor Github Actions work properly- would it make sense to move these tests to Travis? It seems most reliable.

olafurpg:
Travis frequently times out with OOM for `unit/test`, that was the original motivation to start using Azure. I think it's fine to run Azure + GH Actions for now and ignore failures in one of them as long as the other succeeds.

PR #989, opened 2019-10-14
Make find implementation run on a separate thread for each file

tgodzik:


metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -35,7 +37,7 @@ final class ImplementationProvider(
     buildTargets: BuildTargets,
     buffer: Buffers,
     definitionProvider: DefinitionProvider
-) {
+)(implicit ec: ExecutionContext) {
```
hepin1989: will limit the thread size?

tgodzik: It's using`Executors.newCachedThreadPool()`, so it tries to reuse existing threads, but it doesn't limit them.

tgodzik: @olafurpg what do you think? Should we maybe add a thread limit on the ThreadPool ?

olafurpg: I'm concerned about submitting so many runnables to the Metals server execution context. I would prefer to use parallel collections `.foreach` for this use case, collecting results in a `j.u.c.ConcurrentLinkedQueue`. I think parallel collections use their own execution context which is by default suitable for CPU bound work. I don't have a good experience using non-`foreach` combinators in parallel collections.

tgodzik: Changed to parallel collections, thanks!

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -177,29 +180,37 @@ final class ImplementationProvider(
...
-      )
-      implOccurrence <- findDefOccurrence(implDocument, implSymbol, source)
-      range <- implOccurrence.range
-      revised <- distance.toRevised(range.toLSP)
-    } yield new Location(file.toUri.toString, revised)
+      parallelMap = findImplementation(
+        symbolClass.symbol,
+        classContext
+      ).par
+    } parallelMap.foreach {
```
olafurpg: nit: please wrap expression in block
```suggestion
    } {
      parallelMap.foreach {
      ...
    }
```

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -177,29 +180,37 @@ final class ImplementationProvider(
...
+            parentSymbol,
+            implDocument,
+            symbolClass,
+            classContext,
+            implReal
+          )
+          implOccurence <- findDefOccurrence(implDocument, implSymbol, source)
+          range <- implOccurence.range
+          revised <- distance.toRevised(range.toLSP)
+        } allLocations.add(new Location(file.toUri.toString, revised))
```
olafurpg: nit: wrap expression in block

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -177,29 +180,37 @@ final class ImplementationProvider(
...
-      revised <- distance.toRevised(range.toLSP)
-    } yield new Location(file.toUri.toString, revised)
+      parallelMap = findImplementation(
+        symbolClass.symbol,
+        classContext
+      ).par
+    } parallelMap.foreach {
+      case (file, locations) =>
+        val implPath = AbsolutePath(file)
+        for {
```
olafurpg: readability: should we maybe move this for comprehension into a separate method?

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -177,29 +180,37 @@ final class ImplementationProvider(
...
-        implReal
-      )
-      implOccurrence <- findDefOccurrence(implDocument, implSymbol, source)
-      range <- implOccurrence.range
-      revised <- distance.toRevised(range.toLSP)
-    } yield new Location(file.toUri.toString, revised)
+      parallelMap = findImplementation(
+        symbolClass.symbol,
+        classContext
+      ).par
```
olafurpg: I'm not sure how the parallel map works, I generally only use parallel arrays

```scala
val map = findImplentation(...)
keys.toArray.par.foreach { file =>
  val locations = map(file)
  // ...
}
```

tgodzik: Shouldn't it work basically the same?

tgodzik: From what I checked the only real issue is with collections like List, which are sequential by nature. Map should be fine since it allows easily for concurrent access.

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
tgodzik: I can actually move it back in since `foreach` is being invoked by default there.

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
tgodzik: After a bit more research, Arrays seem to be most efficient really, so changed to that.

tgodzik:
> LGTM 
> 
> Do you maybe have any numbers from akka/akka before and after?

Around twice from what I calculated quickly.
But usually it's around 200 ms while before around 500ms

olafurpg:
Un-scientific results from running "find implementation" on the `Actor.receive` method in akka/akka

```
// before
[Trace - 11:33:37 AM] Sending response 'textDocument/implementation - (19)'. Processing request took 1376ms

// after
[Trace - 11:28:40 AM] Sending response 'textDocument/implementation - (24)'. Processing request took 846ms
```
To reproduce, 

* setup metals in the akka/akka repo and index everything.
* enable lsp.trace.json, create the file if it doesn't exist (exact path varies between OS)
```
tail -f /Users/lgeirsson/Library/Caches/org.scalameta.metals/lsp.trace.json | grep textDocument/
```
* Run "Reload Window" to start a clean server.
* Run "go to implementation" on the `Actor` trait to warm up the server.
* Then run "go to implementation" on the `def receive` method

PR #990, opened 2019-10-15
Fix issue when on type formatting threw an exception on last newline

tgodzik:


tgodzik:
The tests get stuck on github actions and not sure why, but this cannot be caused by this PR - will merge.

PR #994, opened 2019-10-16
Keep code lenses after error until next successful compilation

### LINKED ISSUE #992: Code lenses don't disappear after recompiling with error
**Describe the bug**
Code lenses are not being refreshed after compilation error.

**To Reproduce**
1. Open following source code as a file and compile it:
```scala
object Foo {
  def main(args: Array[String]): Unit = ???
}
```
2. wait to see the `run` code lens
3. remove trailing `}` and save the file
4. observe the code lens not disappearing despite the error being reported correctly

**Expected behavior**
The code lens should disappear in `4.`

**Search terms**
lens disappear compilation error

marek1840:
Closes #992 
Previously, code lens would not disappear when recompilation would fail.
Now, we refresh the code lenses in an open tab also after failure

### DIFF
--- a/tests/unit/src/main/scala/tests/BaseLspSuite.scala
+++ b/tests/unit/src/main/scala/tests/BaseLspSuite.scala
@@ -1,6 +1,5 @@
 package tests
 import java.nio.file.Files
-import java.util.Comparator
 import java.util.concurrent.Executors
 import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContextExecutorService
--- a/tests/unit/src/main/scala/tests/BaseLspSuite.scala
+++ b/tests/unit/src/main/scala/tests/BaseLspSuite.scala
@@ -1,5 +1,6 @@
 package tests
 import java.nio.file.Files
+import java.util.Comparator
 import java.util.concurrent.Executors
 import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContextExecutorService
@@ -85,6 +86,13 @@ abstract class BaseLspSuite(suiteName: String) extends BaseSuite {
       .resolve("e2e")
       .resolve(suiteName)
       .resolve(name.replace(' ', '-'))
+
+    if (path.isDirectory) {
+      val files = Files.walk(path.toNIO)
+      try files.sorted(Comparator.reverseOrder()).forEach(Files.delete(_))
+      finally files.close()
+    }
+
     Files.createDirectories(path.toNIO)
     path
   }
--- a/tests/unit/src/main/scala/tests/BaseLspSuite.scala
+++ b/tests/unit/src/main/scala/tests/BaseLspSuite.scala
@@ -87,12 +87,7 @@ abstract class BaseLspSuite(suiteName: String) extends BaseSuite {
       .resolve(suiteName)
       .resolve(name.replace(' ', '-'))
 
-    if (path.isDirectory) {
-      val files = Files.walk(path.toNIO)
-      try files.sorted(Comparator.reverseOrder()).forEach(Files.delete(_))
-      finally files.close()
-    }
-
+    RecursivelyDelete(path)
     Files.createDirectories(path.toNIO)
     path
   }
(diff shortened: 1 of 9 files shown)


tests/unit/src/test/scala/tests/CodeLensesLspSuite.scala
```diff
@@ -159,6 +159,38 @@ object CodeLensesLspSuite extends BaseLspSuite("codeLenses") {
...
+          |  def main(args: Array[String]): Unit = ???
+          |}
+          |""".stripMargin
+      )
+      _ <- server.didSave("a/src/main/scala/Main.scala")(
+        text => text.replace("}", "")
+      )
+      _ <- assertCodeLenses(
+        "a/src/main/scala/Main.scala",
+        """object Main {
```
olafurpg: Do I understand correctly that the code lens is removed due to the compile error?

I'm not sure this is desirable behavior. I would expect the code lens to stay stable until a successful compilation invalidates the main/test class.

olafurpg: Reasoning: I would like the "run" and "test" code lenses to be as stable as possible to avoid jittery code lenses. If the user requests to run a main or test on broken code then we could do a `window/showMessage` or something like that to explain that it's not possible to run the code because it's not compiling.

marek1840: I can see a value in that and it is consistent with the approach "if it doesn't compile, first - make it compile"

tests/unit/src/main/scala/tests/TestingClient.scala
```diff
@@ -64,10 +65,24 @@ final class TestingClient(workspace: AbsolutePath, buffers: Buffers)
       None
   }
 
+  private var refreshedOnIndex = false
+  var refreshModelHandler: Unit => Unit = { _: Unit =>
```
tgodzik: Can you use the `identity` function?

tests/unit/src/main/scala/tests/TestingClient.scala
```diff
@@ -64,10 +65,24 @@ final class TestingClient(workspace: AbsolutePath, buffers: Buffers)
...
+  }
+
   override def metalsExecuteClientCommand(
       params: ExecuteCommandParams
   ): Unit = {
     clientCommands.addLast(params)
+    params.getCommand match {
+      case ClientCommands.RefreshModel.id =>
+        if (refreshedOnIndex)
+          refreshModelHandler()
```
tgodzik: Nit: Could you keep the braces consistent?

tests/unit/src/main/scala/tests/TestingClient.scala
```diff
@@ -64,10 +65,24 @@ final class TestingClient(workspace: AbsolutePath, buffers: Buffers)
...
+        else {
+          refreshedOnIndex = true
+        }
+      case _ =>
+        println()
```
tgodzik: Is this needed?

marek1840:
Other tests are still flaky but at least those related to code lenses seem to be fixed

tests/unit/src/main/scala/tests/BaseLspSuite.scala
```diff
@@ -85,6 +86,13 @@ abstract class BaseLspSuite(suiteName: String) extends BaseSuite {
       .resolve("e2e")
       .resolve(suiteName)
       .resolve(name.replace(' ', '-'))
+
+    if (path.isDirectory) {
```
tgodzik: Isn't this done in `cleanWorkspace`?

marek1840: Right, I will inline it in here

tests/unit/src/main/scala/tests/TestingClient.scala
```diff
@@ -268,3 +281,9 @@ final class TestingClient(workspace: AbsolutePath, buffers: Buffers)
   }
 
 }
+
+object TestingClient {
+  trait Listener {
+    def onRefreshModel(): Unit
```
tgodzik: Is this used?

Issue #762, opened 2019-04-18
Support Java 11


raboof:
The documentation and runtime warnings make it quite clear that Metals requires Java 11.

At some point it would be good if Metals could support Java 11. It might be good to at least start collecting notes on what issues are still holding back support for Java 11?

olafurpg:
Bloop v1.3.0 is out now with JDK 11 support https://github.com/scalacenter/bloop/releases/tag/v1.3.0.

This means it's possible to compile on JDK 11 when Bloop is installed on the computer even if the Metals server runs on JDK 8. We should be able to make the Metals server run on JDK 11 with a few minor changes, such as #739.

olafurpg:
We just merged #765 that goes a long way towards enabling Java 11 support. Fuzzy symbol search (workspace/symbol, which is also used for auto-import completions) does not include JDK 11 symbols, otherwise most other functionality should work on Java 11.

olafurpg:
The PR #792 fixes the last regression for users running Metals on Java 11. Java 11 support will still remain experimental. There are several issues that we should try to solve before advertising official Java 11 support: 

- run Java 11 tests on CI, maybe as a daily cron job
- add a configuration option in VS Code to choose a preferred java version
- automatically detect the situation when the Metals server and installed Bloop server are running on incompatible Java versions and report a helpful message to the user with instructions on how to fix the problem.

DanVicarel:
The VS Code Metals experience is a bit confusing with regards to Java 11 support.

1. The [VS Code page on the Metals website](https://scalameta.org/metals/docs/editors/vscode.html) lists "Java 8 or 11 provided by OpenJDK or Oracle" as one of the requirements.
2. That same page says "Metals only works with Java 8 so this executable cannot point to another version such as Java 11" under `Configure Java version`.
3. When I install Metals in a container with `$JAVA_HOME` pointed at `/usr/local/openjdk-11`, I get a warning from Metals saying "Unable to find Java 8 home. To fix this problem, update the 'Java Home' setting to point to a Java 8 home directory".

So what exactly is the status of this issue? Thanks in advance.

tgodzik:
Fixing the docs issue here: https://github.com/scalameta/metals/pull/883

As for Java Home, we only try to find by default java 8, for java 11 you need to specify it in Metals settings. To fix this we would need to look for 11 here: https://github.com/scalameta/metals-vscode/blob/master/src/getJavaHome.ts

DanVicarel:
[...]

PR #995, opened 2019-10-16
A documentation change in support of Java 11 configuration

andyczerwonka:
The current documentation suggests that Java 11 is supported, but it actually needs to be manually configured at the moment. This PR hopefully makes that more clear.

See https://github.com/scalameta/metals/issues/762 and https://twitter.com/andyczerwonka/status/1184560663215034369 for details.

### DIFF
--- a/docs/editors/vscode.md
+++ b/docs/editors/vscode.md
@@ -42,26 +42,20 @@ variables.
 
 ## Configure Java version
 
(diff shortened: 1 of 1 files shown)


olafurpg:
Our CI has failing recently, we are investigating how to fix it in #982

PR #1004, opened 2019-10-19
Switch to to using Bloop Launcher

tgodzik:
Previously, we were connecting via running bloop with the bsp command, which required Python to start Bloop if it isn't exist.

Now, we connect using the Bloop Launcher, which uses bloopgun underneath, which doesn't require Python and is implemented inside Bloop project itself.


tgodzik:
Fixed the test broken by merging the two yesterday's PRs

build.sbt
```diff
@@ -248,6 +248,8 @@ lazy val metals = project
       // for BSP
       "org.scala-sbt.ipcsocket" % "ipcsocket" % "1.0.0",
       "ch.epfl.scala" % "bsp4j" % V.bsp,
+      "ch.epfl.scala" % "bloop-launcher_2.12" % V.bloop,
```
olafurpg: Use %% to avoid the _2.12 suffix.

build.sbt
```diff
@@ -248,6 +248,8 @@ lazy val metals = project
...
+      "ch.epfl.scala" % "bloop-launcher-core_2.12" % V.bloop,
```
olafurpg: Do we need both the launcher and launcher-core?

tgodzik: I think so, I had `ClassMissingException` when I didn't include core.

tgodzik:
> I wonder if the test.sh script is still needed for CI? If I understand correctly, the launcher will do it for us.

It should work, I will remove it and test.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -1,326 +1,78 @@
...
+        serverStarted
+      )
 
-  private def defaultRetries: Int =
-    if (config.bloopProtocol.isTcp) {
-      10
-    } else {
-      2
-    }
+    val job = ec.submit(new Runnable {
```
gabro: super tiny, but this can probably be a SAM

```scala
ec.submit(_ => launcher.runLauncher(...))
```

tgodzik: It would need to be something like:
```scala
    val runServer: Runnable = () =>
      launcher.runLauncher(
        BuildInfo.bloopVersion,
        skipBspConnection = false,
        Nil
      )
    val job = ec.submit(runServer)
```

I can change it, but doesn't seem like much of an improvement, because we need the explicit type there :/

gabro: Ah, I recall @olafurpg had a similar issue, does it work if you do

```scala
val job = ec.submit[Unit](_ => launcher.runLauncher(
  BuildInfo.bloopVersion,
  skipBspConnection = false,
  Nil
))
```
?

The `Unit` annotation should help disambiguate `Runnable` from `Callable`.

Anyway, it's a super tiny thing, sorry for digressing ^^

tgodzik: Doesn't work either, but thanks for the suggestion. I would love to have it simpler than what it currently is. It feels a bit too Java currently :smile:

gabro: ahah, ok, I surrender.

build.sbt
jvican: You should only use `bloop-launcher`, bloop-launcher-core contains the unshaded classes. What `ClassNotFoundException` did you get?

tgodzik: I removed it and currently it just freezes:
```
INFO  time: initialize in 0.2s
Starting the bsp launcher for bloop...
```

but previously it was I think `UnixDomainSocket` - for sure a class from `sockets`

tgodzik:
> This looks 
> 
> @tgodzik I've noticed that `BloopServers` does not take `MetalsServerConfig` as input anymore. Two questions:
> 
> * was there any bloop-specific parameter in there that we can remove (e.g. `bloopProtocol`)?
> * was there any parameter that we should respect instead (e.g. `verbose`)?

Removed `bloopProtocol` and some more no longer needed classes. However, not exactly sure how to add `verbose` via launcher - there is no room for CLI options there.

build.sbt
olafurpg: What happens if you only use `bloop-launcher-core`? I don't think we need the shaded classes, it would be good to share the coursier dependency for example

tgodzik: It works with just `bloop-launcher-core`, so leave it as the single dependency?

olafurpg: My vote goes to use `bloop-launcher-core` and deal with binary conflicts like we do with any other 3rd party dep

tgodzik: I agree, I think it's fine from Metals perspective.

jvican: Sounds good but keep in mind that you will then need to manage any dependency breakage that happens upstream. Not a big deal but worth thinking about

olafurpg: Maybe we should start using https://github.com/scalacenter/sbt-missinglink 🤔

gabro: https://github.com/scalameta/metals/issues/1007

tgodzik:
Looks like the tests started timing out again - I reverted the commit removing the script.

olafurpg:
🎉

PR #1019, opened 2019-10-25
Suppress model-refresh if debugging is unsupported

### LINKED ISSUE #1016: Command metals-model-refresh should only be sent to clients with debugging support
**Describe the bug**

When using the latest SNAPSHOT of Metals with Emacs we get the following warning
```
Warning (lsp-mode): Unknown metals client command: metals-model-refresh
```
The warning origins from
https://github.com/scalameta/metals/blob/68459a7a9fede5cb56685012a2aa8f01a86a8125/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1536

**Expected behavior**

No warning because the Emacs client doesn't support debugging.

**Additional context**

https://gitter.im/scalameta/metals?at=5db154c72f8a034357c9cd58

marek1840:
closes #1016 

Previously, model-refresh command was sent to the client even though it didn't support debugging.
Now, metals will refrain from sending in in such cases.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
@@ -100,8 +100,6 @@ final class ConfiguredLanguageClient(
   ): Unit = {
     if (config.executeClientCommand.isOn) {
       params.getCommand match {
-        case _ if config.executeClientCommand.isOff =>
-        // ignore
         case ClientCommands.RefreshModel() if !debuggingSupported =>
         // ignore
         case _ =>
--- a/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
@@ -18,10 +18,16 @@ import scala.meta.internal.metals.MetalsEnrichments._
  * from window/logMessage are always visible in the UI while in VS Code the logs are hidden by default.
  */
 final class ConfiguredLanguageClient(
-    _underlying: MetalsLanguageClient,
+    initial: MetalsLanguageClient,
     config: MetalsServerConfig
 )(implicit ec: ExecutionContext)
-    extends DelegatingLanguageClient(_underlying, config) {
+    extends DelegatingLanguageClient(initial) {
+
+  @volatile private var debuggingSupported = true
+
+  def configure(capabilities: ClientExperimentalCapabilities): Unit = {
+    debuggingSupported = capabilities.debuggingProvider
+  }
 
   override def shutdown(): Unit = {
     underlying = NoopLanguageClient
@@ -93,7 +99,14 @@ final class ConfiguredLanguageClient(
       params: ExecuteCommandParams
   ): Unit = {
     if (config.executeClientCommand.isOn) {
-      underlying.metalsExecuteClientCommand(params)
+      params.getCommand match {
+        case _ if config.executeClientCommand.isOff =>
+        // ignore
+        case ClientCommands.RefreshModel.id if !debuggingSupported =>
+        // ignore
+        case _ =>
+          underlying.metalsExecuteClientCommand(params)
+      }
     }
   }
 
--- a/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
@@ -25,8 +25,9 @@ final class ConfiguredLanguageClient(
 
   @volatile private var debuggingSupported = true
 
-  def configure(capabilities: ClientExperimentalCapabilities): Unit = {
+  override def configure(capabilities: ClientExperimentalCapabilities): Unit = {
     debuggingSupported = capabilities.debuggingProvider
+    super.configure(capabilities)
   }
 
   override def shutdown(): Unit = {
--- a/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
@@ -102,7 +102,7 @@ final class ConfiguredLanguageClient(
       params.getCommand match {
         case _ if config.executeClientCommand.isOff =>
         // ignore
-        case ClientCommands.RefreshModel.id if !debuggingSupported =>
+        case ClientCommands.RefreshModel() if !debuggingSupported =>
         // ignore
         case _ =>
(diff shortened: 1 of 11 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageClient.scala
```diff
@@ -1,19 +1,30 @@
...
 import org.eclipse.lsp4j.MessageType
 import org.eclipse.lsp4j.jsonrpc.services.JsonNotification
 import org.eclipse.lsp4j.jsonrpc.services.JsonRequest
 import org.eclipse.lsp4j.services.LanguageClient
-
 import scala.meta.internal.tvp._
 import scala.meta.internal.metals.MetalsEnrichments._
 
 trait MetalsLanguageClient extends LanguageClient with TreeViewClient {
+  @volatile private var debuggingSupported = true
```
olafurpg: Any logic related to configuring what gets forwarded to the language client should be handled in `ConfigureLanguageClient`.

As a general rule of thumb, I would recommend avoiding `val` and `var` inside traits.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -494,7 +498,7 @@ final class TestingServer(
       _ = client.refreshModelHandler = handler
       // first compilation, to trigger the handler
       _ <- server.compilations.compileFiles(List(path))
-      lenses <- codeLenses.future
+      lenses <- codeLenses.future.withTimeout(30, TimeUnit.SECONDS)
```
olafurpg: what is the motivation for this change? 30 seconds is a long timeout.

tests/unit/src/test/scala/tests/UnsupportedDebuggingLspSuite.scala
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.concurrent.TimeUnit
+import scala.concurrent.TimeoutException
+import scala.meta.internal.metals.ClientCommands
+import scala.meta.internal.metals.ClientExperimentalCapabilities
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.util.Failure
+import scala.util.Success
+
+object UnsupportedDebuggingLspSuite
```
olafurpg: 👍 Thank you for writing a test case!

metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
```diff
@@ -93,7 +99,14 @@ final class ConfiguredLanguageClient(
       params: ExecuteCommandParams
   ): Unit = {
     if (config.executeClientCommand.isOn) {
-      underlying.metalsExecuteClientCommand(params)
+      params.getCommand match {
+        case _ if config.executeClientCommand.isOff =>
+        // ignore
+        case ClientCommands.RefreshModel.id if !debuggingSupported =>
```
olafurpg: Does this work? I think there's an `unapply`
```suggestion
        case ClientCommands.RefreshModel() if !debuggingSupported =>
```

metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
```diff
@@ -93,7 +99,14 @@ final class ConfiguredLanguageClient(
       params: ExecuteCommandParams
   ): Unit = {
     if (config.executeClientCommand.isOn) {
-      underlying.metalsExecuteClientCommand(params)
+      params.getCommand match {
+        case _ if config.executeClientCommand.isOff =>
```
olafurpg: Why is this line needed? We already guard against `isOn` above

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -200,6 +199,14 @@ class MetalsLanguageServer(
     MetalsLogger.setupLspLogger(workspace, redirectSystemOut)
     val clientExperimentalCapabilities =
       ClientExperimentalCapabilities.from(params.getCapabilities)
+
+    languageClient.underlying match {
+      case client: ConfiguredLanguageClient =>
+        client.configure(clientExperimentalCapabilities)
```
olafurpg: Would it make sense to add this method to `MetalsLanguageClient` with a default implementation that ignores the argument?

metals/src/main/scala/scala/meta/internal/metals/NoopLanguageClient.scala
```diff
@@ -39,9 +39,6 @@ abstract class NoopLanguageClient extends MetalsLanguageClient {
   override def metalsTreeViewDidChange(
       params: TreeViewDidChangeParams
   ): Unit = ()
-  // override def metalsTreeViewNodeReveal(
```
olafurpg: Why is this change needed?

metals/src/main/scala/scala/meta/metals/Main.scala
```diff
@@ -35,9 +34,8 @@ object Main {
         .setRemoteInterface(classOf[MetalsLanguageClient])
         .setLocalService(server)
         .create()
-      val underlyingClient = launcher.getRemoteProxy
-      val client = new ConfiguredLanguageClient(underlyingClient, config)(ec)
```
olafurpg: What was the motivation for this change?

metals/src/main/scala/scala/meta/internal/metals/NoopLanguageClient.scala
marek1840: As a rule of thumb, I don't like to keep a commented out code in the codebase. If the code is unused, then it can and probably should be removed.

metals/src/main/scala/scala/meta/metals/Main.scala
marek1840: The introduced `UnsupportedDebuggingLspSuite` tests the behavior of the `ConfigurationLanguageClient` which is responsible for suppressing the model-refresh notification. Without this change, we couldn't test it, as the `ConfigurationLanguageClient` was created only inside `Main` object, and then injected to the language server (hence not present in the test suite).

metals/src/main/scala/scala/meta/internal/metals/NoopLanguageClient.scala
olafurpg: I agree! I thought the method was removed in this PR but I didn't see it was commented out.

metals/src/main/scala/scala/meta/metals/Main.scala
olafurpg: I see, makes sense, thanks for the explanation!

PR #1020, opened 2019-10-27
Reenable and fix tests on Windows

tgodzik:
After moving most test to GithubActions it turned out it might be possible to reenable the Windows tests. However, during the time we were not running the tests it seems some of them broke.

Fixes https://github.com/scalameta/metals/issues/878


tgodzik:
@marek1840 It seems that running and testing doesn't work on windows.

tests/unit/src/test/scala/tests/CodeLensesLspSuite.scala
```diff
@@ -188,7 +188,7 @@ object CodeLensesLspSuite extends BaseLspSuite("codeLenses") {
   def check(name: String, library: String = "")(expected: String): Unit = {
     testAsync(name) {
       cleanWorkspace()
-      val original = expected.replaceAll("<<.*>>\\W", "")
+      val original = expected.replaceAll("<<.*>>\\W+", "")
```
tgodzik: This broke on windows due to "\n\r"

tests/unit/src/test/scala/tests/DebugProtocolSuite.scala
```diff
@@ -5,8 +5,23 @@ import java.util.concurrent.TimeUnit.SECONDS
 import ch.epfl.scala.bsp4j.ScalaMainClass
 import ch.epfl.scala.bsp4j.DebugSessionParamsDataKind
 import scala.meta.internal.metals.MetalsEnrichments._
+import scala.concurrent.duration.Duration
+import scala.concurrent.Future
 
 object DebugProtocolSuite extends BaseLspSuite("debug-protocol") {
```
tgodzik: Debug is not working - ignored

tests/unit/src/test/scala/tests/DocumentHighlightLspSuite.scala
```diff
@@ -25,7 +25,7 @@ object DocumentHighlightLspSuite extends BaseLspSuite("documentHighlight") {
     """
       |object Main {
       |  val abc = 123
-      |  abc.<<toInt@@>>
+      |  abc.<<to@@Int>>
```
tgodzik: Seems that doesn't work with cursor at last position - doesn't seem like a big issue and might not make sense to debug until someone reports. Looks to be working outside of tests.

metals/src/main/scala/scala/meta/internal/metals/FoldingRangeExtractor.scala
```diff
@@ -173,8 +175,9 @@ final class FoldingRangeExtractor(
           startToken <- term.expr.findFirstTrailing(_.is[Token.KwCatch])
           lastCase <- term.catchp.lastOption
           endToken <- lastCase.findFirstTrailing(_.is[Token.RightBrace])
-        } yield
-          Position.Range(tree.pos.input, startToken.pos.end, endToken.pos.end)
+          tryPos = Position
```
tgodzik: Formatting here was a bit crazy and worked differently on Windows.

tgodzik:
I think this should be ready to merge. I ignored a couple of tests, but even having windows tests in a limited capacity will be helpful (and it's not a crazy amount of ignored tests).

metals/src/main/scala/scala/meta/internal/metals/FoldingRangeExtractor.scala
olafurpg: What's going on 😨

tests/unit/src/test/scala/tests/DocumentHighlightLspSuite.scala
olafurpg: Hmm, I wonder what's going on here 🤔

PR #1023, opened 2019-10-28
trace the debug adapter protocol messages

### LINKED ISSUE #1022: Trace the DAP messages
Currently there is no way to know what messages were exchanged between the dap client, proxy and server.

We should add a tracing mechanism like one used for logging the messages from lsp and bsp.

marek1840:
closes #1022

Previously the dap messages were were not being traced.
Now, we trace them on both ends in: `dap-client.trace.json` and `tap-server.trace.json` files

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/EndpointLogger.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/EndpointLogger.scala
@@ -1,0 +1,72 @@
+package scala.meta.internal.metals.debug
+
+import java.io.PrintWriter
+import java.time.Clock
+import java.time.format.DateTimeFormatter
+import java.util.Collections
+import java.util.function.Consumer
+import com.google.gson.GsonBuilder
+import org.eclipse.lsp4j.jsonrpc.MessageConsumer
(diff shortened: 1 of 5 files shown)


metals/src/main/scala/scala/meta/internal/metals/debug/EndpointLogger.scala
```diff
@@ -0,0 +1,72 @@
...
+    endpoint.listen { message =>
+      try log(Received, message)
+      finally messageConsumer.consume(message)
+    }
+  }
+
+  override def cancel(): Unit = endpoint.cancel()
+
+  private def log(direction: Direction, message: Message): Unit = {
+    logger.println(s"[Trace][$time] $direction ${typeOf(message)}:")
```
tgodzik: Does it work with the inspector tool? It would be nice to keep it compatible.

marek1840: Sadly, no - the format is too much of a hassle to implement it manually. Reusing the implementation present in the `lsp4j.jsonrpc` library is also not viable because it is tightly coupled with their launcher implementation.

At one point we might want to change the format to accomodate the inspector but let's delay it for now as it is better to have some logs than none at all.

PR #1026, opened 2019-10-28
Change mtags to use project matrix

tgodzik:


build.sbt
```diff
@@ -174,13 +173,14 @@ lazy val interfaces = project
       "implNote:a:Implementation Note:"
     )
   )
+  .jvmPlatform(scalaVersions = V.supportedScalaVersions)
```
olafurpg: How often does this project build when you run `publishLocal`?

build.sbt
```diff
@@ -379,7 +380,7 @@ lazy val cross = project
       if211 = List("-Ywarn-unused:imports")
     )
   )
-  .dependsOn(mtest, mtags)
+  .dependsOn(mtest, mtags.jvm("2.12.8"))
```
olafurpg: I would expect `mtest` to also be a `projectMatrix` 🤔

tgodzik:
I think making this work will be much more complex than it's worth.

Issue #1008, opened 2019-05-06
Automatic implementation for abstract methods in concrete classes


dgalichet:
**Is your feature request related to a problem? Please describe.**

Not related to a problem

**Describe the solution you'd like**

This could take place automatically or with quick action.

```scala
  trait Foo {
    def foo(i: Int): String
  }
// before
  val foo = new Foo {}

// after
  val foo = new Foo {
    def foo(i: Int): String = ???
  }
```

**Describe alternatives you've considered**

This could be triggered using completion (same way that for the exhaustive pattern match), or using a quick action.

**Additional context**

**Search terms**:

olafurpg:
I think it would be a good idea to start by implementing this as a completion. We can additionally support this feature as a code action.

The label of this completion could be something like `implement all methods` which works similar to `match (exchaustive)`.

A good place to start looking at to implement this feature would be here https://github.com/scalameta/metals/blob/a5050d7c17d011183dad832ba786a7a7edfa9fc0/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala#L1183-L1194

We can create a new `CompletionOverrideAllSuite` that works similarly as https://github.com/scalameta/metals/blob/a5050d7c17d011183dad832ba786a7a7edfa9fc0/tests/cross/src/test/scala/tests/pc/CompletionOverrideSuite.scala

I'm not sure in what order this completion should appear 🤔 Should it appear at the top of the completion results or should it appear below the indvidual abstract methods?

olafurpg:
One tricky situation that we should try to handle is when different methods introduce different symbol in scope with the same name. For example
```scala
package a { class Foo }
package b { class Foo }
trait Foo {
  def one: a.Foo
  def two: b.Foo
}
class Impl extends Foo {
  // implement all methods: ensure we don't insert conflicting imports
}
```
We can do this by using the same `ShortenedNames` instance when rendering the signatures of all methods https://github.com/scalameta/metals/blob/a5050d7c17d011183dad832ba786a7a7edfa9fc0/mtags/src/main/scala/scala/meta/internal/pc/Signatures.scala#L104

After rendering all signatures, the `ShortenedNames` instance make sure we don't introduce conflicting imports.

ckipp01:
> I'm not sure in what order this completion should appear 🤔 Should it appear at the top of the completion results or should it appear below the individual abstract methods? 

My initial thought is at the top?

olafurpg:
[...]

PR #1031, opened 2019-10-30
Add completion to add all abstract members

### LINKED ISSUE #1029: Override abstract var not behaving as expected
**Describe the bug**
If you have the following code below and try to trigger a completion for the abstract `var foo`, you aren't able to trigger it by beginning to type `var`. However, you are able to trigger the override completion with beginning to type `def` or `override`. If you do then trigger the completion for it, it will be implemented as `def foo: Int = ???`

```scala
package example

object Hello extends App {
  trait Foo {
    var foo: Int
  }

  val test = new Foo {
    // trigger override here
  }
}
```

**To Reproduce**
Just pace the example object from, move to where it says _trigger override here_, and then attempt to get the correct `var foo` override implementation.

**Expected behavior**
I would expect that the override completion be triggered by starting to type `override` or `var` and that the implementation would both show the correct signature and implementation of a `var` and not a `def`.

**Screenshots**
Screenshot 2019-10-30 at 15 05 12
Screenshot 2019-10-30 at 15 05 33
Screenshot 2019-10-30 at 15 05 46

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code and Vim
- Metals version: v0.7.6-Most recent snapshot

**Search terms**
absract members, var, completions

ckipp01:
So I believe I have the implemented correctly now that it will work for all abstract members -- defs, vals, and vars. If you have 5, but 2 implemented, then the implement all will implement the remaining 3.

I am not sure about a few things though...
1) The way I did the actual text edit with the strings of the other text edits
2) I wasn't positive about what you mentioned in the issue about this https://github.com/scalameta/metals/issues/1008#issuecomment-544505366, but I made a test with that scenario, and it seems to work as expected?
3) Other tests that I may be missing.

This should close #1029 and also address #1008. However, regarding #1008, I'm not sure if we also want to implement it as well where if you are instantiating a class or something that has the abstract members, it will gives you the option to automatically create the class not just with an empty `{}` but rather with all the members. I wasn't sure if that should be grouped with this PR or not.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Configs.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Configs.scala
@@ -72,9 +72,9 @@ object Configs {
           "metals.hover.documentation",
           default = true
         ),
-        isMagicIndentClient = MetalsServerConfig.binaryOption(
-          "metals.magic-indent-client",
-          default = false
+        snippetVscodeIndent = MetalsServerConfig.binaryOption(
+          "metals.snippet-vscode-indent",
+          default = true
         ),
         isSignatureHelpDocumentationEnabled = MetalsServerConfig.binaryOption(
           "metals.signature-help.documentation",
--- a/metals/src/main/scala/scala/meta/internal/metals/Configs.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Configs.scala
@@ -72,8 +72,8 @@ object Configs {
           "metals.hover.documentation",
           default = true
         ),
-        snippetVscodeIndent = MetalsServerConfig.binaryOption(
-          "metals.snippet-vscode-indent",
+        snippetAutoIndent = MetalsServerConfig.binaryOption(
+          "metals.snippet-auto-indent",
           default = true
         ),
         isSignatureHelpDocumentationEnabled = MetalsServerConfig.binaryOption(
--- a/metals/src/main/scala/scala/meta/internal/metals/Configs.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Configs.scala
@@ -72,6 +72,10 @@ object Configs {
           "metals.hover.documentation",
           default = true
         ),
+        isMagicIndentClient = MetalsServerConfig.binaryOption(
+          "metals.magic-indent-client",
+          default = false
+        ),
         isSignatureHelpDocumentationEnabled = MetalsServerConfig.binaryOption(
           "metals.signature-help.documentation",
           default = true
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -1231,14 +1231,23 @@ trait Completions { this: MetalsGlobal =>
 
           val allAbstractMembers = overrideMembers
             .filter(_.sym.isAbstract)
-            .map(_.edit)
+
+          val (allAbstractEdits, allAbstractImports) =
+            allAbstractMembers.foldLeft(
+              (List.empty[l.TextEdit], Set.empty[l.TextEdit])
+            ) { (editsAndImports, overrideDefMember) =>
+              val edits = overrideDefMember.edit :: editsAndImports._1
+              val imports = overrideDefMember.autoImports.toSet ++ editsAndImports
+                ._2
+              (edits, imports)
+            }
 
           if (allAbstractMembers.length > 1 && overrideDefMembers.length > 1) {
             val necessaryIndent = if (metalsConfig.snippetAutoIndent()) {
               ""
             } else {
               val amount =
-                allAbstractMembers.head.getRange.getStart.getCharacter
+                allAbstractEdits.head.getRange.getStart.getCharacter
               " " * amount
             }
 
@@ -1246,14 +1255,14 @@ trait Completions { this: MetalsGlobal =>
               prefix,
               new l.TextEdit(
                 range,
-                allAbstractMembers
+                allAbstractEdits
                   .map(_.getNewText)
-                  .reverse
                   .mkString(s"\n${necessaryIndent}")
               ),
               completionsSymbol("implement"),
               label = Some("Implement all members"),
-              detail = Some(s" (${allAbstractMembers.length} total)")
+              detail = Some(s" (${allAbstractMembers.length} total)"),
+              additionalTextEdits = allAbstractImports.toList
             )
 
             implementAll :: overrideDefMembers
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -15,6 +15,7 @@ import scala.meta.internal.semanticdb.Scala._
 import scala.meta.internal.tokenizers.Chars
 import scala.meta.pc.PresentationCompilerConfig.OverrideDefFormat
 import scala.util.control.NonFatal
+import scala.collection.immutable.Nil
 
 /**
  * Utility methods for completions.
@@ -1217,6 +1218,7 @@ trait Completions { this: MetalsGlobal =>
           val overrideMembers = typed.tpe.members.iterator.toList
             .filter(isOverridableMethod)
             .map(OverrideCandidate.apply)
+            .map(_.toMember)
 
           val overrideDefMembers: List[OverrideDefMember] =
             overrideMembers
@@ -1226,21 +1228,17 @@ trait Completions { this: MetalsGlobal =>
                   candidate.filterText
                 )
               }
-              .map(_.toMember)
-              .toList
 
           val allAbstractMembers = overrideMembers
-            .map(_.toMember)
             .filter(_.sym.isAbstract)
-            .toList
-            .map(_.edit.getNewText)
+            .map(_.edit)
 
           if (allAbstractMembers.length > 1 && overrideDefMembers.length > 1) {
             val implementAll: TextEditMember = new TextEditMember(
               prefix,
               new l.TextEdit(
                 range,
-                allAbstractMembers.reverse.mkString("\n")
+                allAbstractMembers.map(_.getNewText).reverse.mkString(s"\n")
               ),
               completionsSymbol("implement"),
               label = Some("Implement all members"),
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -1,18 +1,20 @@
 package scala.meta.internal.pc
 
-import java.net.URI
 import java.lang.StringBuilder
-import org.eclipse.{lsp4j => l}
-import scala.meta.internal.semanticdb.Scala._
-import scala.collection.mutable
-import scala.meta.internal.mtags.MtagsEnrichments._
-import scala.util.control.NonFatal
-import scala.meta.internal.jdk.CollectionConverters._
-import scala.collection.mutable.ListBuffer
-import scala.meta.pc.PresentationCompilerConfig.OverrideDefFormat
+import java.net.URI
 import java.nio.file.Paths
 import java.util.logging.Level
+
+import org.eclipse.{lsp4j => l}
+
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+import scala.meta.internal.jdk.CollectionConverters._
+import scala.meta.internal.mtags.MtagsEnrichments._
+import scala.meta.internal.semanticdb.Scala._
 import scala.meta.internal.tokenizers.Chars
+import scala.meta.pc.PresentationCompilerConfig.OverrideDefFormat
+import scala.util.control.NonFatal
 
 /**
  * Utility methods for completions.
@@ -1035,7 +1037,8 @@ trait Completions { this: MetalsGlobal =>
      * @param t the enclosing template for the class/object/trait we are implementing.
      * @param pos the position of the completion request, points to `_CURSOR_`.
      * @param text the text of the original source code without `_CURSOR_`.
-     * @param defn the method (either `val` or `def`) that we are implementing.
+     * @param start the position start of the completion.
+     * @param isCandidate the determination of whether the symbol will be a possible completion item.
      */
     case class Override(
         name: Name,
@@ -1047,9 +1050,9 @@ trait Completions { this: MetalsGlobal =>
     ) extends CompletionPosition {
       val prefix: String = name.toString.stripSuffix(CURSOR)
       val typed: Tree = typedTreeAt(t.pos)
-      val isDecl = typed.tpe.decls.toSet
+      val isDecl: Set[Symbol] = typed.tpe.decls.toSet
       val range: l.Range = pos.withStart(start).withEnd(pos.point).toLSP
-      val lineStart: Int = pos.source.lineToOffset(pos.line - 1)
+      val lineStart: RunId = pos.source.lineToOffset(pos.line - 1)
 
       // Returns all the symbols of all transitive supertypes in the enclosing scope.
       // For example:
@@ -1207,16 +1210,47 @@ trait Completions { this: MetalsGlobal =>
       }
 
       override def contribute: List[Member] = {
-        if (start < 0) Nil
-        else {
-          typed.tpe.members.iterator
+        if (start < 0) {
+          Nil
+        } else {
+
+          val overrideMembers = typed.tpe.members.iterator.toList
             .filter(isOverridableMethod)
             .map(OverrideCandidate.apply)
-            .filter { candidate =>
-              CompletionFuzzy.matchesSubCharacters(prefix, candidate.filterText)
-            }
+
+          val overrideDefMembers: List[OverrideDefMember] =
+            overrideMembers
+              .filter { candidate =>
+                CompletionFuzzy.matchesSubCharacters(
+                  prefix,
+                  candidate.filterText
+                )
+              }
+              .map(_.toMember)
+              .toList
+
+          val allAbstractMembers = overrideMembers
             .map(_.toMember)
+            .filter(_.sym.isAbstract)
             .toList
+            .map(_.edit.getNewText)
+
+          if (allAbstractMembers.length > 1 && overrideDefMembers.length > 1) {
+            val implementAll: TextEditMember = new TextEditMember(
+              prefix,
+              new l.TextEdit(
+                range,
+                allAbstractMembers.reverse.mkString("\n")
+              ),
+              completionsSymbol("implement"),
+              label = Some("Implement all members"),
+              detail = Some(s" (${allAbstractMembers.length} total)")
+            )
+
+            implementAll :: overrideDefMembers
+          } else {
+            overrideDefMembers
+          }
         }
       }
     }
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -1234,7 +1234,7 @@ trait Completions { this: MetalsGlobal =>
             .map(_.edit)
 
           if (allAbstractMembers.length > 1 && overrideDefMembers.length > 1) {
-            val necessaryIndent = if (metalsConfig.snippetVscodeIndent()) {
+            val necessaryIndent = if (metalsConfig.snippetAutoIndent()) {
               ""
             } else {
               val amount =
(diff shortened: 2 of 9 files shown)


ckipp01:
Seeing the full test suite run I realized I'm missing some cases like the override keywords when the members are in abstract classes etc. I'll address those. Sorry, I should have looked closer at those before issuing the pr. For now, I'll just say I knew they were failing and just wanted some earlier feedback 😏 .

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1183,14 +1191,42 @@ trait Completions { this: MetalsGlobal =>
...
+          val allAbstractMembers = typed.tpe.members.iterator
             .filter(isOverridableMethod)
             .map(OverrideCandidate.apply)
-            .filter { candidate =>
-              CompletionFuzzy.matchesSubCharacters(prefix, candidate.filterText)
-            }
             .map(_.toMember)
+            .filter(_.sym.isAbstract)
             .toList
+            .map(_.edit.getNewText)
```
olafurpg: I wonder if we could somehow reuse the rendering of the signatures? Alternatively, could we introduce some cache to avoid computing the type signatures + imports twice?

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1183,14 +1191,42 @@ trait Completions { this: MetalsGlobal =>
...
+          if (allAbstractMembers.length > 1 && overrideMembers.length > 1) {
+            val implementAll: TextEditMember = new TextEditMember(
+              prefix,
+              new l.TextEdit(
+                range,
+                allAbstractMembers.reverse.mkString("\n")
+              ),
+              completionsSymbol("implement"),
+              label = Some("Implement all members"),
+              detail = Some(s"(${allAbstractMembers.length} members)")
```
olafurpg: "all members (3 members)" feels a bit redundant
```suggestion
              detail = Some(s" (${allAbstractMembers.length} total)")
```

tests/cross/src/test/scala/tests/pc/CompletionOverrideAllSuite.scala
```diff
@@ -0,0 +1,180 @@
...
+      |  def one: a.Foo
+      |  def two: b.Foo
+      |}
+      |object Main {
+      |  val x = new Foo {
+      |    def@@
+      |  }
+      |}
+      |""".stripMargin,
+    """|Implement all members(2 members)
```
olafurpg: How does the code look like after running the completion?

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1183,14 +1191,42 @@ trait Completions { this: MetalsGlobal =>
       override def contribute: List[Member] = {
         if (start < 0) Nil
         else {
-          typed.tpe.members.iterator
+
+          val overrideMembers: List[OverrideDefMember] =
+            typed.tpe.members.iterator
```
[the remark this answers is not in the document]
ckipp01: So i'm not 100% sure what's happening here. I originally tried to reuse this iterator, but couldn't get it to work. For example, when I change this block to the below

```scala
          val members = typed.tpe.members.iterator

          val overrideMembers: List[OverrideDefMember] =
            members
              .filter(isOverridableMethod)
              .map(OverrideCandidate.apply)
              .filter { candidate =>
                CompletionFuzzy.matchesSubCharacters(
                  prefix,
                  candidate.filterText
                )
              }
              .map(_.toMember)
              .toList

          val allAbstractMembers = members
            .filter(isOverridableMethod)
            .map(OverrideCandidate.apply)
            .map(_.toMember)
            .filter(_.sym.isAbstract)
            .toList
            .map(_.edit.getNewText)
```
the individual overrides still appear in the completions, but the `allAbstractMembers` won't appear at all in the completion list. I'm not fully sure I understand why. Any insight on the best way to reuse this?

tests/cross/src/test/scala/tests/pc/CompletionOverrideAllSuite.scala
ckipp01: It will end up looking like this
```scala
package a { class Foo }
package b { class Foo }

trait Foo {
  def one: a.Foo
  def two: b.Foo
}
object Main {
  val x = new Foo {
    def one: a.Foo = ???
    def two: b.Foo = ???
  }
}
```

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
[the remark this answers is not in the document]
olafurpg: Iterators can only be consumed once. You can work around it with

```diff
-          val members = typed.tpe.members.iterator
+          val members = typed.tpe.members.iterator.toList
```

tests/cross/src/test/scala/tests/pc/CompletionOverrideAllSuite.scala
olafurpg: we should have a test case to ensure it produces that code, if we render the methods separately there's a risk we produce
```
import a.Foo
import b.Foo
...
   def one: Foo = ???
   def two: Foo = ???
```

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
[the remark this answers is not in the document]
ckipp01: Ahh cool. I didn't know that. Thanks!

tests/cross/src/test/scala/tests/pc/CompletionOverrideAllSuite.scala
ckipp01: Sounds good. I added a few more tests in for the various cases to actually check what will be produced.

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
ckipp01: This was the one point I didn't feel like addressed. I saw there were a couple of other places in the completions were cache was implemented but was sort of lost on how to implement it here.

tests/cross/src/test/scala/tests/pc/CompletionOverrideAllSuite.scala
```diff
@@ -0,0 +1,362 @@
...
+    """|package example
+       |
+       |trait Foo {
+       |  def foo: Int
+       |  def bar: Int
+       |}
+       |object Main {
+       |  val x = new Foo {
+       |    def foo: Int = ${0:???}
+       |def bar: Int = ${0:???}
```
ckipp01: So the formatting is a bit odd in these. I'm assuming because either range or type formatting isn't being triggered when these lines are being put in here? I'm not sure if I should just rely on when you're in the editor and the line is put there that it will then automatically be formated by (I think the range formatting?), or if I need to also calculate the amount of indent there should be there?

olafurpg: I would expect the completions to include the indentation here and I think we already calculate the necessary indentation for similar features.

tests/cross/src/test/scala/tests/pc/CompletionOverrideAllSuite.scala
```diff
@@ -0,0 +1,366 @@
...
+       |""".stripMargin,
+    """|Implement all members (2 total)
+       |""".stripMargin,
+    filter = _.contains("Implement")
+  )
+
+  // Note: ckipp01 the indentation for the completions on the edits
+  // look a bit odd, but it seems the tests aren't respecting the actual
+  // indentation as these all work locally as expected, and when the actual
+  // indent is added, the necessary indent is unnecessarily doubled
```
gabro: If the test behavior is different than the actual one, this may be a bug in 
https://github.com/scalameta/metals/blob/f7dc2c5d6732bb484420451cccc8e386807d49fd/tests/mtest/src/main/scala/tests/TextEdits.scala#L13-L34

ckipp01: I'll take a closer look at this here to see if I can find anything to explain what's happening. Thanks!

ckipp01:
I think it may be best to have someone take a look at the what is going on with the tests and the indentation. I'm sort of just going in circles. I can't seem to figure out why the behavior is different locally than in the test, but I can't get them to act the same.

gabro:
@ckipp01 I've taken a quick look, and it seems the indentation in the `newText` field of the completion item is not correct. For example, here's the `newText` in the `simple-edit` test case:

```scala
"def foo: Int = ${0:???}\ndef bar: Int = ${0:???}"
```

Notice the absence of spaces after `\n`.

So the rendering in tests seems correct. The fact that you see a correct behavior when testing in the editor could be due to the editor itself

ckipp01:
> Notice the absence of spaces after \n

hmm, yea I do see that. Any idea why then when I do grab the indentation and add it after the \n that it double indents in the editors?

gabro:
@ckipp01 this seems relevant https://github.com/Microsoft/language-server-protocol/issues/83

tgodzik:
> @ckipp01 this seems relevant [microsoft/language-server-protocol#83](https://github.com/microsoft/language-server-protocol/issues/83)

We could drop the indentation when it's being used in VS Code. The setting can be specified in `MetalsServerConfig`. 

I think I've seen the comment about somewhere, so I might be repeating myself, but I think this will be fine do for VS Code.

gabro:
@tgodzik yep, I agree that’s a good solution. Apparently coc.vim does the same, so it’s not only vscode.
We briefly discussed this on Discord yesterday

tgodzik:
The failing test seems to be breaking only for 2.13.0, so best to change the Scala version in sbt and run the test. It might be fixed in 2.13.1, in that case I would just ignore it for 2.13.0. We do it in a couple of different cases.

```
sbt
> ++2.13.0
> cross/testOnly -- tests.pc.CompletionOverrideConfigSuite
```

ckipp01:
So tonight I went on a little journey to see what editors input this "magic indent" and we know that VS Code did as well as coc.nvim.

Atom also seems to add it...


Emacs seems to not add it...


but then I came across something really bizarre... I haven't really used Sublime before and I gave it a shot, and I could not figure out why I couldn't get the "Implement all members" to appear. Randomly, I somehow just typed `e` which will consistently bring up the "Implement all members"... but then when it is selected, it just inputs that text. Any idea what in the world is going on, and why that is happening in Sublime?

tgodzik:
@ckipp01 Thanks for looking into this! Seems Sublime uses `label` as text to be inserted, it seems to be a bug inside the plugin. We might want to use insertText (or label if all else fails) for Sublime itself. But not sure if it's a large issue right now, since that is a client issue.

For any formatting issues we might want to add formatting command to be run after the completion, which should solve most issues.

```
/**
	 * An optional command that is executed *after* inserting this completion. *Note* that
	 * additional modifications to the current document should be described with the
	 * additionalTextEdits-property.
	 */
	command?: Command;
```

gabro:
Paging also @ayoub-benali about the Sublime LSP bug

ayoub-benali:
@ckipp01 have you tried using this setting in the LSP plugin ? 
`"prefer_label_over_filter_text": true`

ckipp01:
> @ckipp01 have you tried using this setting in the LSP plugin ?

Yea, I have that setting set to true.

ayoub-benali:
@tgodzik sublime LSP plugin tries `textEdit`
see https://github.com/tomv564/LSP/blob/4f9c91c6186e4f59ad073329e54b7ff29dd7661d/plugin/core/completion.py#L32

ayoub-benali:
@ckipp01 there could be a regression in the plugin then. For example the Exhaustive match feature used to work and now it doesn't for me. I assume this PR behaves the same

ckipp01:
> @ckipp01 there could be a regression in the plugin then. For example the Exhaustive match feature used to work and now it doesn't for me. I assume this PR behaves the same

Ironically, when testing this, the exhaustive match worked correctly for me which just confused me more about why this one was behaving differently.

ayoub-benali:
ok I will investigate the sublime text behavior later today

ckipp01:
So I took a stab at adding in the `isMagicIndentClient` which I use to determine whether the indent should be added or not. This now allows the tests to act normal and have the spacing you would expect and also to work in the editors that don't seem to do the magic indent. _However_, I screwed something up with the way the value is being set because even in the editors where I set it to true, it seems that the value is never actually being set and always defaulting. So, I know I'm missing something. Help me Obi Wan Kenobi, what am I missing?

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -34,7 +34,10 @@ abstract class BasePCSuite extends BaseSuite {
   val scalaLibrary: Seq[Path] = PackageIndex.scalaLibrary
   def extraClasspath: Seq[Path] = Nil
   def scalacOptions: Seq[String] = Nil
-  def config: PresentationCompilerConfig = PresentationCompilerConfigImpl()
+  def config: PresentationCompilerConfig =
+    PresentationCompilerConfigImpl().copy(
+      isMagicIndentClient = false
```
gabro: `false` is already the default, so you can leave this unchanged

gabro:
> even in the editors where I set it to true, it seems that the value is never actually being set and always defaulting

@ckipp01 how are you testing this? The configuration is based on the `metals.client` property being sent by the clients

ckipp01:
> @ckipp01 how are you testing this? The configuration is based on the metals.client property sent by the clients

I use `publishLocal`, and then restart the server after making sure VS Code is set to the latest SNAPSHOT. It's for sure using the latest snapshot (I tested with having it input random text to make sure it was using the latest).  Same with vim, I bootstrap the latest snapshot before opening the project to test.

gabro:
@ckipp01 I don't have brilliant ideas, but the config is initialized here

https://github.com/scalameta/metals/blob/1f6c327d0b66a96c4f28b58103feff1090332030/metals/src/main/scala/scala/meta/metals/Main.scala#L20-L21

so you may want to check what happens there and whether the expected system properties are set. You can use a `scribe.info()` as a glorified `println` 🤓

metals/src/main/scala/scala/meta/internal/metals/Configs.scala
```diff
@@ -72,6 +72,10 @@ object Configs {
           "metals.hover.documentation",
           default = true
         ),
+        isMagicIndentClient = MetalsServerConfig.binaryOption(
+          "metals.magic-indent-client",
+          default = false
```
olafurpg: I think it would be better to default use current behavior, which is `true`. I have a feeling a lot of people on vim/emacs don't use the `client` property so that would be a regression for them.

metals/src/main/scala/scala/meta/internal/metals/Configs.scala
```diff
@@ -72,6 +72,10 @@ object Configs {
           "metals.hover.documentation",
           default = true
         ),
+        isMagicIndentClient = MetalsServerConfig.binaryOption(
+          "metals.magic-indent-client",
```
olafurpg: I'm not a fan of the "magic indent" name. How about `metals.snippet-vscode-indent=true`? The behavior of the option would be determined by how VS Code treats indentation in completion `new Text` snippets.

ckipp01: Sure, I can change that. One argument to use magic-indent though is that it follows other patters such as the one used here https://github.com/spring-projects/sts4/blob/master/headless-services/commons/commons-language-server/src/main/java/org/springframework/ide/vscode/commons/languageserver/completion/VscodeCompletionEngineAdapter.java#L287-L289 and also language that has already been used when talking about the issue on the LSP GitHub page. I'm just worried that having vscode in the name is very vscode-centric when it's not just vscode that does it. For example, atom also uses the same concept.

metals/src/main/scala/scala/meta/internal/metals/Configs.scala
ckipp01: Sounds good. I changed the default to true

ckipp01:
Alright, I believe I have it all figured out now. I was correctly setting the value based on the client... but not passing it into the compiler options like I needed to. That's now fixed and I address the naming things you brought up @olafurpg. Playing around with it locally with VS Code, emacs, atom, and vim all seem to be working as expected!

tests/cross/src/test/scala/tests/pc/CompletionOverrideAllSuite.scala
```diff
@@ -0,0 +1,361 @@
...
+       |  }
+       |}
+       |""".stripMargin,
+    """|Implement all members (3 total)
+       |""".stripMargin,
+    filter = _.contains("Implement")
+  )
+
+  checkEdit(
+    "mixed-partial-edit",
```
olafurpg: What happens if a method we override needs an import?
```scala
trait Foo {
  def file: java.io.File 
}
class Bar {
  def file@@ // implement all members
}
```

ckipp01:
> The test failure seems legitimate
> 
> ```scala
> 2019-11-07T20:37:28.3751896Z ===========
> 2019-11-07T20:37:28.3752152Z => Obtained
> 2019-11-07T20:37:28.3752409Z ===========
> 2019-11-07T20:37:28.3752659Z     """|package example
> 2019-11-07T20:37:28.3752919Z        |
> 2019-11-07T20:37:28.3753267Z        |trait Foo {
> 2019-11-07T20:37:28.3753610Z        |  def foo: Int
> 2019-11-07T20:37:28.3753865Z        |  val bar: Int
> 2019-11-07T20:37:28.3754171Z        |  var car: Int
> 2019-11-07T20:37:28.3754456Z        |}
> 2019-11-07T20:37:28.3754828Z        |object Main {
> 2019-11-07T20:37:28.3755164Z        |  val x = new Foo {
> 2019-11-07T20:37:28.3755424Z        |    def foo: Int = ${0:???}
> 2019-11-07T20:37:28.3756297Z        |    val bar: Int = ${0:???}
> 2019-11-07T20:37:28.3757119Z        |    def car: Int = ${0:???}
> 2019-11-07T20:37:28.3757534Z        |  }
> 2019-11-07T20:37:28.3757820Z        |}
> 2019-11-07T20:37:28.3758114Z 
> 2019-11-07T20:37:28.3758340Z 
> 2019-11-07T20:37:28.3758670Z =======
> 2019-11-07T20:37:28.3759035Z => Diff
> 2019-11-07T20:37:28.3759569Z =======
> 2019-11-07T20:37:28.3760233Z ---âˆ™
> 2019-11-07T20:37:28.3760789Z +++âˆ™
> 2019-11-07T20:37:28.3761374Z @@ -11,3 +11,3 @@
> 2019-11-07T20:37:28.3761741Z      val bar: Int = ${0:???}
> 2019-11-07T20:37:28.3762210Z -    def car: Int = ${0:???}
> 2019-11-07T20:37:28.3762583Z +    var car: Int = ${0:???}
> 2019-11-07T20:37:28.3762896Z    }
> ```
> 
> I'm not sure what Scala version that's on, I think 2.11.12. Previously, the test names in `cross/test` included the Scala version but that seems to have regressed since we moved to GitHub Actions 🤔

I was assuming the test failure was what @tgodzik was referring to up above when he said
> The failing test seems to be breaking only for 2.13.0, so best to change the Scala version in sbt and run the test. It might be fixed in 2.13.1, in that case I would just ignore it for 2.13.0. We do it in a couple of different cases.

as I can seem to reproduce it locally when running the tests or when actually testing the functionality.

tests/cross/src/test/scala/tests/pc/CompletionOverrideAllSuite.scala
ckipp01: Just tested it locally and it import it, but let me add a test to actually check it.

tgodzik:
> I was assuming the test failure was what @tgodzik was referring to up above when he said

Did you manage to confirm that it only breaks for 2.13.0? I think the failure I saw was a bit different, so might be a different issue altogether.

tests/cross/src/test/scala/tests/pc/CompletionOverrideAllSuite.scala
ckipp01: Actually nevermind, it does for a single member but not like I thought it would with the implement all. Good catch, thanks.

ckipp01:
I was able to fix the issue where imports weren't being added if they were necessary for the implement all members.

I have verified that the tests in cross that are failing only fail for 2.11.12. The `isVar` seems to not be working the same way. We had a conversation regarding this here https://discordapp.com/channels/632642981228314653/632652693013528589/642698819967909899 on discord. I'm sort of at a loss of how to fix it.

olafurpg:
@ckipp01 do the tests maybe get fixed if you rebase on top of master? The issue with vars got fixed in #1056 by @gabro

gabro:
yep, @ckipp01 confirmed in the other PR that tests pass when rebased 🎉

ckipp01:
Soooo happy to see all the green check marks finally, ha! Thanks for the work on #1056 @gabro that seemed to do the trick!

Issue #1027, opened 2019-10-28
Error highlighting is intermittent in Emacs


marcsaegesser:
**Describe the bug**
Compiler errors not always indicated in Emacs. Sometimes the red-squigglies appear along with error messages in the margin. Other times compiler errors are just not highlighted at all. Potentially related, there are times where errors are highlighted and when the error is fixed the red-squigglies do not go away.

This behavior is intermittent. Some times it works and then it stops.

I have created a simple project that I used to demonstrate the issue it can be found [here](https://github.com/marcsaegesser/ScalaMetalsTest). In the root directory is the minimal Emacs init.el file that I used. It is taken directly from the Metals documentation for Emacs with the small addition of disabling SBT supershell because that kills sbt-mode.

This directory also contains the metals.log file in case that is useful. The screen shot file is also there. The screen shows that the compiler errors appear in the sbt-mode window but the source code does not indicate any errors.

**To Reproduce**
1. I removed the existing .bloop and .metals directories so that I was starting from scratch.
2. I opened the Main.scala file in Emacs and selected "I" when prompted to import the project.
3. I split the window and started sbt-mode
4. In the sbt-mode window I started a continuous compile with ~compile
5. I entered two lines of code, each with a type error. This time the errors were highlighted.
6. I deleted those two lines and entered them again. This time the errors where not highlighted.
7. The sbt-mode windows shows the compiler errors that are not highlighted in the source code.

**Expected behavior**
Compiler errors should always be highlighted.

**Screenshots**
![ScreenShot](https://user-images.githubusercontent.com/1134304/67726531-24e61a00-f9b4-11e9-97bc-556446adedfe.png)

**Installation:**
- Linux
- Editor: Emacs
- Metals version: v0.7.6+170-9eddd91f-SNAPSHOT

**Additional context**

**Search terms**
red squiggly highlighting

olafurpg:
Thank you for reporting! I'm unable to reproduce in VS Code.

This might be a duplicate of https://github.com/scalameta/metals/issues/956.

> I deleted those two lines and entered them again. This time the errors where not highlighted.

Did you save the file between the changes? Metals triggers compilation on file save.

I see there's a `.#Main.scala` file in the screenshot, what happens if you configure emacs to place those generated files in another directory?

marcsaegesser:
Yes, I saved the file but it actually didn't matter. Again, the behavior is intermittent so it's difficult to provide concrete reproduction steps.

[...]

PR #1032, opened 2019-10-30
Update sample Emacs configuration to avoid AutoSave files in src dir

marcsaegesser:
See issue #1027. AutoSave files in source directories break the
compile and in Emacs can cause error highlighting to vanish in some
sourc files.

This change also shows a work-around to prevent the SBT 1.3 supershell
from killing sbt-mode.

### DIFF
--- a/docs/editors/emacs.md
+++ b/docs/editors/emacs.md
@@ -33,8 +33,11 @@ To use Metals in Emacs, place this snippet in your Emacs configuration (for exam
 (require 'use-package)
 
 ;; Enable defer and ensure by default for use-package
+;; Keep auto-save/backup files separate from source code:  https://github.com/scalameta/metals/issues/1027
 (setq use-package-always-defer t
-      use-package-always-ensure t)
+      use-package-always-ensure t
+      backup-directory-alist `((".*" . ,temporary-file-directory))
+      auto-save-file-transforms `((".*" ,temporary-file-directory t)))
 
 ;; Enable scala-mode and sbt-mode
 (use-package scala-mode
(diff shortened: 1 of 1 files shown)


docs/editors/emacs.md
```diff
@@ -33,8 +33,11 @@ To use Metals in Emacs, place this snippet in your Emacs configuration (for exam
 (require 'use-package)
 
 ;; Enable defer and ensure by default for use-package
+;; Keep auto-save/backup files separate from source code:  https://github.com/scalameta/metals/issues/1027
 (setq use-package-always-defer t
-      use-package-always-ensure t)
+      use-package-always-ensure t
+      backup-directory-alist `((".*" . ,temporary-file-directory))
+      auto-save-file-transforms `((".*" ,temporary-file-directory t)))
```
ag91: hey, I think you mean `auto-save-file-name-transforms`, because `auto-save-file-transforms` does not seem to be defined in emacs.

ag91:
thanks @marcsaegesser ! I added a comment about a variable, but the rest seems good to me. The fix for sbt 1.3 is very welcome: hopefully they will add it to sbt-mode itself.

docs/editors/emacs.md
marcsaegesser: Ugh. You are correct, the variable name should be auto-save-file-name-transforms. I had it correct in my personal emacs config and then messed it up putting the changes into the metals repo. Thank you for catching my boneheaded mistake.

gabro:
I've taken the liberty to fix the typo, thanks again @marcsaegesser for the contribution!

PR #1033, opened 2019-10-31
Update SemanticDB version to 4.2.5 and enable previously broken test

tgodzik:


olafurpg:
The test failure for `unit/test` is legitimate, you can fix it by running `sbt save-expect`

olafurpg:
Looks like the 4.2.4 release didn't complete the full cross-build, I have triggered a new v4.2.5 release.

tgodzik:
> Looks like the 4.2.4 release didn't complete the full cross-build, I have triggered a new v4.2.5 release.

Will try with the new one, I haven't had time to take a look at it, but will make sure everything is working with 4.2.5 on Monday

olafurpg:
The release job is running at https://travis-ci.org/scalameta/scalameta/jobs/606391878 🍿 let's hope it completes before the Travis timeout. Might be a good idea to migrate the Scalameta release job to GitHub Actions to avoid the timeouts

olafurpg:
Also probably time to upgrade the scalameta build to sbt v1.3 so that we can use the latest sbt-sonatype to speed up the publish step

olafurpg:
Release succeeded, artifacts should be on Maven in the next ~30 minutes

olafurpg:
Release is up now on Maven, all artifacts seem to have uploaded successfully

tgodzik:
Test failures seem legit, will take a look on Monday.

tgodzik:
@olafurpg Also I think 2.12.7 is not published, but I think we should most likely drop it and deprecate 2.12.8.

olafurpg:
I'm in favor of dropping 2.12.7 and deprecating 2.12.8/2.12.9

tests/slow/src/test/scala/tests/gradle/GradleLspSuite.scala
```diff
@@ -284,7 +284,7 @@ object GradleLspSuite extends BaseImportSuite("gradle-import") {
       _ = assertNoDiff(
         client.messageRequests.peekLast(),
         // only 3 projects since no empty test targets are created for gradle
```
marek1840: > only 3

now it's 5 ;)

tgodzik: Updated the comment not to include a number - that was not important really.

PR #1041, opened 2019-11-03
Implement basic worksheet mode using mdoc and text decorations

### LINKED ISSUE #1040: Worksheet mode
It would be nice if there was a "worksheet mode" that allows you to interactively evaluate toplevel statements inside the editor and see the results. 

Some challenges:

- how to extract static types and runtime values? We could build on top of the Scala REPL or mdoc (https://scalameta.org/mdoc/)
- what will be the JSON-RPC protocol? We could maybe build on top of http://dotty.epfl.ch/docs/usage/worksheet-mode-implementation-details.html or use the Debug Adapter Protocol
- how to display values in the editor? We can use VS Code decorations https://code.visualstudio.com/api/references/vscode-api#TextEditorDecorationType but that approach may not work well with multi-line output.

olafurpg:
Previously, there was no way for users to interactively explore APIs and evaluate small programs when coding with Metals. This commit adds a new "worksheet mode" for files with the extension `.worksheet.sc`. Programs with this file extension are evaluated on file save and with evaluated results are displayed inline alongside the code using a new "Decoration Protocol".

## Snappy feedback

Simple code usually compiles and evaluates pretty fast, even if you do a bit of I/O

## Robust against slow programs

Slow running programs can be cancelled with a single click, even infinite loops.

## Infinite stream are evaluated lazily

Screenshot 2019-11-03 at 19 23 23

Thanks to http://www.lihaoyi.com/PPrint/

## Runtime crashes are reported with red squiggles

A full stack trace is included in the error message.

Screenshot 2019-11-03 at 18 19 29

## Hover message for large values
Large values are truncated to a single line and a restricted column width. Hover over the decoration to see the full evaluated value.

Screenshot 2019-11-03 at 19 24 23

## Acknowledgements
Kudos to @Duhemm for pioneering Scala worksheets in the Dotty Language Server (http://dotty.epfl.ch/docs/usage/worksheet-mode-implementation-details.html). The implementation in this PR is heavily inspired by worksheets in the Dotty Language Server.

Big thanks to @gabro for exploring decorations in the VS Code API and giving lots of useful feedback.

Fixes #1040.


olafurpg:
There are several features that are not implemented in this PR but would be super cool to support in the future

* evaluate parts of the document (e.g., notebook cells) to avoid repeating slow computation. I have some ideas how to support this with mdoc while ensuring that cells always evaluate in the right order.
* displaying rich media objects (plots, charts, progress bars, etc). This might be blocked by https://github.com/microsoft/vscode/issues/3220. We could maybe build on top of the Notebook support for Python in VS Code 🤔 
* writing documentation in the script. I believe it would be quite straightforward to extend the mdoc-based implementation to support "markdown worksheets" in a similar fashion as R Markdown https://rmarkdown.rstudio.com/

olafurpg:
Accompanying changes to the VS Code extension https://github.com/scalameta/metals-vscode/pull/159

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
```diff
@@ -0,0 +1,352 @@
...
+      token: CancelToken
+  ): Future[Array[DecorationOptions]] = {
+    if (path.extension != "sc") {
+      Future.successful(Array.empty)
+    } else {
+      reporter.reset()
+      val result = new CompletableFuture[Array[DecorationOptions]]()
+      token.onCancel().asScala.foreach {
+        case java.lang.Boolean.TRUE =>
+          // Don't publish any
```
gabro: the comment seems 'interrupted', was it meant to be

```suggestion
          // Don't publish any result on cancelation
```

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
```diff
@@ -0,0 +1,352 @@
...
+          result.complete(Array.empty)
+        case _ =>
+      }
+      def runEvaluation(): Unit = {
+        cancelables.add(Cancelable(() => result.complete(Array.empty)))
```
gabro: the cancelation logic is repeated here and at line 90, is this intentional? Are these two different kinds of cancelation that can happen?

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
```diff
@@ -0,0 +1,352 @@
...
+
+  private def evaluateWorksheet(
+      path: AbsolutePath,
+      token: CancelToken
+  ): Array[DecorationOptions] = {
+    val commentHeader = " // "
+    val input = path.toInputFromBuffers(buffers)
+    val decorations = for {
+      ctx <- getContext(path)
+      source <- dialects.Sbt1(input).parse[Source].toOption
```
gabro: I understand the logic but seeing the `Sbt1` dialect here is a bit confusing. Can we have a `Worksheet` dialect instead? It can be a type alias, it's purely for readability.

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
```diff
@@ -0,0 +1,352 @@
...
+            Iterator("\n// ", line)
+          }
+        ).flatten.mkString
+        DecorationOptions(
+          range,
+          new l.MarkedString("scala", hoverMessage),
+          ThemableDecorationInstanceRenderOptions(
+            after = ThemableDecorationAttachmentRenderOptions(
+              contentText,
+              color = "green",
```
gabro: It's fine for this first iteration, but I wonder if we can let the editors decide on how to render these.

It feels strange to decide on styling details server-side.

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
```diff
@@ -0,0 +1,352 @@
...
+    for {
+      target <- buildTargets.inverseSources(path)
+      info <- buildTargets.scalaTarget(target)
+      scala <- info.info.asScalaBuildTarget
+      scalaVersion = scala.getScalaVersion
+      isSupported = ScalaVersions.isSupportedScalaVersion(scalaVersion)
+      _ = {
+        if (!isSupported) {
+          scribe.warn(
+            s"worksheet: unsupported Scala version '${scalaVersion}', to fix this problem use Scala version '${BuildInfo.scala212}' instead."
```
gabro: do worksheets only work with Scala 2.12? If not, shouldn't we recommend 2.13 instead?

mtags/src/main/scala/scala/meta/internal/mtags/MtagsEnrichments.scala
```diff
@@ -98,6 +98,15 @@ trait MtagsEnrichments {
         case _ => false
       }
     }
+    def isWorksheet: Boolean = {
+      filename.endsWith(".worksheet.sc")
+    }
+    def isScalaOrScalaScript: Boolean = {
+      extension match {
+        case "scala" | "sc" => true
```
gabro: I'm a bit confused in general on whether this feature supports `.sc` or `.worksheet.sc`.

mtags/src/main/scala/scala/meta/internal/pc/MetalsGlobal.scala
```diff
@@ -384,7 +385,11 @@ class MetalsGlobal(
       case _ => code
     }
     val unit = newCompilationUnit(codeWithCursor, filename)
-    val richUnit = new RichCompilationUnit(unit.source)
+    val source =
+      if (filename.endsWith(".sc"))
```
gabro: `filename.isScalaScript` for consistency?

metals/src/main/scala/scala/meta/internal/implementation/GlobalClassTable.scala
```diff
@@ -37,7 +37,7 @@ final class GlobalClassTable(
     for {
       buildTargetId <- buildTargets.inverseSources(source)
       scalaTarget <- buildTargets.scalaTarget(buildTargetId)
-      classpath = new Classpath(scalaTarget.classpath)
+      classpath = new Classpath(scalaTarget.jarClasspath)
```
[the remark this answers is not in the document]
tgodzik: This actually makes sense, the name was a bit misleading before - I added it while doing implmentation.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -169,6 +180,7 @@ class MetalsLanguageServer(
   private var doctor: Doctor = _
   var httpServer: Option[MetalsHttpServer] = None
   var treeView: TreeViewProvider = NoopTreeViewProvider
+  var worksheetProvider: WorksheetProvider = NoopWorksheetProvider
```
tgodzik: Would it be possible to just use Option instead?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1610,6 +1640,31 @@ class MetalsLanguageServer(
     )
   }
 
+  private def onWorksheetChangedUnbatched(
```
tgodzik: Can we move this inside the WorksheetProvider ?

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
```diff
@@ -0,0 +1,352 @@
...
+      }
+    }
+    sh.schedule(
+      promptUserToCancel,
+      userConfig().worksheetCancelTimeout,
+      TimeUnit.SECONDS
+    )
+  }
+
+  private def evaluateWorksheet(
```
tgodzik: This method looks a bit complex - can we move some functionalities to separate methods? It's a bit hard to review it, since a lot of the logic seems rendering specific.

docs/editors/decoration-protocol.md
```diff
@@ -0,0 +1,110 @@
+---
+id: decoration-protocol
+title: Decoration Protocol
```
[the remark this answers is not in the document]
olafurpg: Sounds good, I added the version v0.1.0 and a changelog at the bottom of the document. I'm not sure we need to include the version in the actual JSON-RPC communication however 🤔

docs/editors/decoration-protocol.md
```diff
@@ -0,0 +1,110 @@
...
+  renderOptions: ThemableDecorationInstanceRenderOption;
+}
+```
+
+## ThemableDecorationInstanceRenderOption
+
+```ts
+export interface ThemableDecorationInstanceRenderOption {
+  /** The decoration to display next to the given range. */
+  after?: ThemableDecorationAttachmentRenderOptions;
```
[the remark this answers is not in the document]
olafurpg: Yes, the implementation mirrors the VS Code API but we currently only use a fraction of it. I would prefer to wait with expanding the actual protocol as we start using more parts of it.

docs/editors/new-editor.md
```diff
@@ -347,6 +347,12 @@ To enable Metals extensions, start the main process with the system property
 Metals implements several custom JSON-RPC endpoints related to rendering tree
 views in the editor client, the [Tree View Protocol](tree-view-protocol.md).
 
+### Decoration Protocol
+
+Metals implements an LSP extension called the "Decoration Protocol" to rendering
```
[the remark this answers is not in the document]
olafurpg: Good catch, fixed!

metals/src/main/scala/scala/meta/internal/decorations/DecorationClient.scala
```diff
@@ -0,0 +1,34 @@
...
+import org.eclipse.lsp4j.Range
+import org.eclipse.lsp4j.MarkedString
+
+trait DecorationClient {
+  @JsonNotification("metals/decorationTypeDidChange")
+  def metalsDecorationTypeDidChange(
+      params: DecorationTypeDidChange
+  ): Unit
+  @JsonNotification("metals/publishDecorations")
+  def metalsDecorationRangesDidChange(
```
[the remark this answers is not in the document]
olafurpg: Good catch, the name was originally `metals/decorationRangesDidChange` but then I changed to `metals/publishDecorations` since the API feels closer to `textDocument/publishDiagnostics` than `workspace/didChangeConfiguration`

metals/src/main/scala/scala/meta/internal/decorations/DecorationClient.scala
```diff
@@ -0,0 +1,34 @@
...
+      params: PublishDecorationsParams
+  ): Unit
+}
+
+case class DecorationTypeDidChange(
```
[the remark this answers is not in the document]
olafurpg: I removed the method since it's never used.

metals/src/main/scala/scala/meta/internal/decorations/DecorationClient.scala
```diff
@@ -0,0 +1,34 @@
...
+    @Nullable overviewRulerLane: java.lang.Integer = null
+)
+
+case class DecorationOptions(
+    range: Range,
+    @Nullable hoverMessage: MarkedString = null,
+    @Nullable renderOptions: ThemableDecorationInstanceRenderOptions = null
+)
+
+case class PublishDecorationsParams(
```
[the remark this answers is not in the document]
olafurpg: The endpoint name is `metals/publishDecorations`

metals/src/main/scala/scala/meta/internal/implementation/GlobalClassTable.scala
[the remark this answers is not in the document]
olafurpg: It took me a while to actually debug this issue, I just saw the extension method `.classpath` and didn't realize it filtered out directory entries

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -224,7 +194,7 @@ class Compilers(
       .orElse(interactiveSemanticdbs.flatMap(_.getBuildTarget(path)))
     target match {
       case None =>
-        if (path.toLanguage.isScala) Some(ramboCompiler)
+        if (path.isScalaOrScalaScript) Some(ramboCompiler)
```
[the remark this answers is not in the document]
olafurpg: Renamed to `isScalaFilename`

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
olafurpg: That's a great question! I refactored to use `Option[WorksheetProvider]` and it came out much nicer

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
olafurpg: I simplified this implementation a bit. For consistency with `onBuildChanged` I'd like to keep it in this file for now

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -25,7 +25,9 @@ case class UserConfiguration(
     scalafmtConfigPath: RelativePath =
       UserConfiguration.default.scalafmtConfigPath,
     symbolPrefixes: Map[String, String] =
-      UserConfiguration.default.symbolPrefixes
+      UserConfiguration.default.symbolPrefixes,
+    screenWidth: Int = 120,
```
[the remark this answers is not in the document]
olafurpg: Good idea, done.

metals/src/main/scala/scala/meta/internal/tvp/MetalsTreeViewProvider.scala
```diff
@@ -199,7 +199,13 @@ class MetalsTreeViewProvider(
   ): MetalsTreeViewChildrenResult = {
     val children: Array[TreeViewNode] = params.viewId match {
       case Help =>
+        pprint.log(BuildInfo.metalsVersion)
```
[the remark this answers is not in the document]
olafurpg: Yes, I also removed the new tree view node with the version. That can be left for a separate PR

metals/src/main/scala/scala/meta/internal/worksheets/MdocStoreReporter.scala
```diff
@@ -0,0 +1,53 @@
...
+import scala.meta._
+import scala.meta.internal.metals.MetalsEnrichments._
+import mdoc.internal.io.ConsoleReporter
+import scala.collection.mutable
+import org.eclipse.lsp4j.Diagnostic
+import org.eclipse.lsp4j.DiagnosticSeverity
+import java.io.ByteArrayOutputStream
+import java.io.PrintStream
+
+class StoreReporter() extends ConsoleReporter(System.out) {
```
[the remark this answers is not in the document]
olafurpg: Good catch! This was forgotten in my manual rename implementation, can't wait to have "rename symbol" 😄

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
olafurpg: Fixed.

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
olafurpg: I made both call the same method to complete the empty result.

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
olafurpg: I agree, I should probably have opened a "draft PR". I had already refactored this method into smaller methods before seeing your review

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
olafurpg: I added a `Worksheet` alias. We could consider adding this to Scalameta itself.

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
olafurpg: I agree, but my understanding is that there's no way for us to access the colors for things like comments https://github.com/Microsoft/vscode/issues/32813

olafurpg: We could hardcode this value in the client but I'm not sure that's any better 🤔 Or maybe it is better to send some enum to the client
```scala
object Colors {
  val comment = 1
  val identifier = 2
}
```
and then have the VS Code client translate that into colors

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
olafurpg: Worksheets currently only work with 2.12.10. To support 2.11/2.13 I would prefer to move some of the logic here into mdoc and publish a stable `mdoc-interfaces` Java library. I felt it would be better to leave that for a separate PR

mtags/src/main/scala/scala/meta/internal/mtags/MtagsEnrichments.scala
olafurpg: Completions work in `*.sc` files even if they are not `*.worksheet.sc`. The "eval on file save and display decorations" feature only works for `*.worksheet.sc`

olafurpg: The new `isScalaFilename` method name could make it clearer

* "scala filename": completions/hover/... work
* "scala worksheet": eval with mdoc and display decorations

mtags/src/main/scala/scala/meta/internal/pc/MetalsGlobal.scala
olafurpg: Done.

metals/src/main/scala/scala/meta/internal/worksheets/MetalsWorksheetProvider.scala
gabro: My first idea would be to use tmLanguage classes, but then I don’t know how easy is for the client to use them

olafurpg:
> What about using workspaceEdit with adding the results as comments? This is also not ideal, but I worry about adding LSP extensions that there will be no one to actually implement those for anything else.

@tgodzik I think this approach is worth exploring as a fallback for other editors. We might be able to include a magic marker for the inserted comments so it's easy to tell which comments are generated by worksheets and which comments are written by the user.

metals/src/main/scala/scala/meta/internal/metals/TokenEditDistance.scala
```diff
@@ -42,6 +42,8 @@ final class TokenEditDistance private (
    *   nearest token in the original document instead.
    */
   def toRevised(range: l.Range): Option[l.Range] = {
+    pprint.log(isNoMatch)
```
tgodzik: Looks like a leftover?

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -0,0 +1,389 @@
...
+
+  def decorations(
+      path: AbsolutePath,
+      token: CancelToken
+  ): Future[Array[DecorationOptions]] = {
+    reporter.reset()
+    val result = new CompletableFuture[Array[DecorationOptions]]()
+    def completeEmptyResult() = result.complete(Array.empty)
+    token.onCancel().asScala.foreach {
+      case java.lang.Boolean.TRUE => completeEmptyResult()
```
tgodzik: maybe just: 
```
foreach{
  cancelled => if(cancelled) completeEmptyResult()
}
```

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -0,0 +1,389 @@
...
+
+  private def renderHoverMessage(
+      statement: Statement,
+      margin: Int,
+      isEmptyValue: Boolean
+  ): String = {
+    val out = new StringBuilder()
+    if (!isEmptyValue) {
+      statement.binders.iterator.foreach { binder =>
+        out
```
tgodzik: Maybe this would be a bit more clear:
```
val color = binder.tpe.render(TPrintColors.BlackWhite)
out.append(s"\n${binder.name}: $color: ")
``` 
?

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -0,0 +1,389 @@
...
+  }
+
+  private def renderDecoration(statement: Statement): DecorationOptions = {
+    val pos = statement.position
+    val range = new l.Range(
+      new l.Position(pos.startLine, pos.startColumn),
+      new l.Position(pos.endLine, pos.endColumn)
+    )
+    val margin = math.max(
+      20,
```
tgodzik: Move `20` to a value? Seems like a magic number

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -0,0 +1,389 @@
...
+    )
+  }
+
+  private def getContext(path: AbsolutePath): Option[Context] = {
+    for {
+      target <- buildTargets.inverseSources(path)
+      info <- buildTargets.scalaTarget(target)
+      scala <- info.info.asScalaBuildTarget
+      scalaVersion = scala.getScalaVersion
+      isSupported = ScalaVersions.isSupportedScalaVersion(scalaVersion)
```
gabro: You mentioned worksheets work only in 2.12.10, but this is checking against all versions supported by Metals, isn't it?

docs/editors/new-editor.md
```diff
@@ -369,6 +375,8 @@ _Request_:
 interface MetalsSlowTaskParams {
   /** The name of this slow task */
   message: string;
+  /** If true, don't open the logs output. */
+  noLogs?: boolean;
```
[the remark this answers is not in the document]
gabro: ☝️ @olafurpg what do you think about this?

metals/src/main/scala/scala/meta/internal/metals/TokenEditDistance.scala
olafurpg: Good catch, removed.

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
olafurpg: Done

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
olafurpg: Not really IMO. When using string builders it's usually best to append individual strings

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
olafurpg: Good idea, done

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
olafurpg: Great catch! Indeed. Changed the condition to check against the same Scala version

docs/editors/new-editor.md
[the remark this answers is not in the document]
olafurpg: Renamed to `quietLogs` to mean that even if there may be logs from this task, they can be ignored (i.e., the output channel doesn't need to be shown).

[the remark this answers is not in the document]
olafurpg: Better?

olafurpg:
I would normally ignore the flaky test failure (it was passing before with the same code) but the failing test case is from `WorksheetLspSuite`
```
 X tests.worksheets.WorksheetLspSuite.completion 4510ms 
No more data in the server stdin, exiting...
  tests.TestFailedException: failed assertion at D:\a\metals\metals\tests\unit\src\test\scala\tests\worksheets\WorksheetLspSuite.scala:30
failed assertion at D:\a\metals\metals\tests\mtest\src\main\scala\tests\DiffAssertions.scala:24
Obtained empty output!
  tests.DiffAssertions$.colored(DiffAssertions.scala:107)
  tests.BaseSuite.assertNoDiff(BaseSuite.scala:100)
  tests.worksheets.WorksheetLspSuite$.$anonfun$new$4(WorksheetLspSuite.scala:30)
  scala.util.Success.$anonfun$map$1(Try.scala:255)
  scala.util.Success.map(Try.scala:213)
  scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
```
I'll try to investigate...

olafurpg:
Ignoring unrelated flaky test failure 
```diff
X tests.TreeViewLspSuite.no-op 2280ms 
  tests.TestFailedException: failed assertion at /home/runner/work/metals/metals/tests/unit/src/test/scala/tests/TreeViewLspSuite.scala:487
===========
=> Obtained
===========
    """|b/ +


=======
=> Diff
=======
--- obtained
+++ expected
@@ -1,1 +1,1 @@
-b/ +
+
```

olafurpg:
For people who want to try out the new worksheets

* Upgrade VS Code extension, may require a restart to pick up the new release
* Set "Server version" setting to 0.7.6+208-90bc4909-SNAPSHOT

Feedback welcome!

Issue #679, opened 2019-04-17
Implement textDocument/rename


olafurpg:
The rename request is sent from the client to the server to perform a workspace-wide rename of a symbol. https://microsoft.github.io/language-server-protocol/specification#textDocument_rename

Some challenges:
-  overridden methods, should we rename only the focused symbol or all overrides + their implementations? IntelliJ lets you choose the behavior but this UI is not available in LSP (although we can optionally implement it for VS Code).
- broken code, should we require the sources to compile successfully? I vote for always renaming even in the presence of broken code.

olafurpg:
Blocked by fix in https://github.com/scalameta/scalameta/pull/1859 to handle renaming of overridden methods in anonymous classes

olafurpg:
Up for debate: should "rename symbol" work for symbols that are not defined in the workspace? For example, should it be possible to rename references to a class defined in library dependencies? In IntelliJ, the symbol must be defined in the workspace but sometimes this limitation might be undesirable.

gabro:
@olafurpg I'm a bit puzzled by the use case?

If the symbol is defined in a dependency, wouldn't renaming references to that symbol just break the usage site? 
I'm sure I'm missing something, can you provide an example?

olafurpg:
I think you are right, we should restrict rename for symbols that are defined in the workspace.

cb372:
Is anyone currently working on this?

Just as an anecdotal data point, I just watched a colleague try VS Code + Metals for the first time, discover that they couldn't rename symbols, and switch back to IntelliJ within 5 minutes 😢

tgodzik:
@olafurpg started to work on some necessary work with `goto implementation`, but not sure when he will be able to work on this. We will most likely plan something after he gets back.

PR #1048, opened 2019-11-06
Implement textDocument/rename

tgodzik:
Previously, rename was not supported in any way. User would have to manually rename everything by hand and that could prove really tedious. Now, we offer renaming of any symbol with a small number of exclusions:
- unapply
- equals
- hashCode
- unary_!
- !

We can also add/remove other exclusions if needed.

There are also some special cases and several approaches that need to be mentioned:

1. When apply is renamed we add full name to each invocation

renamed

2. Symbols can't be renamed while compiling, since we need full and correct semanticDB information - an error will pop up as a message.
3. Methods ending with colon can only be renamed to another name ending with colon. 
4. Renaming an overriden method will rename both parent symbol and current symbol:
rename-inherit

5. Symbols from outside workspace cannot be renamed.
6. Companion objects and classes are renamed.
companion

Currently not working:
- type reference such as `classOf[MyClass]` or `[T <: MyClass]`
- we don't check operator precedence when they are renamed
- we rename all parent's overriden methods by default - we might want to ask about that, however MessageRequest doesn't seem a pretty way to do it (might be mistaken).

Fixes https://github.com/scalameta/metals/issues/679

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -255,12 +255,13 @@ object MetalsEnrichments
 
     def isWorkspaceSource(workspace: AbsolutePath): Boolean =
       isLocalFileSystem(workspace) &&
-        !isInReadonlyDirectory(workspace)
+        !isInReadonlyDirectory(workspace) &&
+        path.toNIO.startsWith(workspace.toNIO)
 
-    def isLocalFileSystem(workspace: AbsolutePath) =
+    def isLocalFileSystem(workspace: AbsolutePath): Boolean =
       workspace.toNIO.getFileSystem == path.toNIO.getFileSystem
 
-    def isInReadonlyDirectory(workspace: AbsolutePath) =
+    def isInReadonlyDirectory(workspace: AbsolutePath): Boolean =
       path.toNIO.startsWith(
         workspace.resolve(Directories.readonly).toNIO
       )
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -250,10 +250,20 @@ object MetalsEnrichments
       }
     }
     def isDependencySource(workspace: AbsolutePath): Boolean =
-      workspace.toNIO.getFileSystem == path.toNIO.getFileSystem &&
-        path.toNIO.startsWith(
-          workspace.resolve(Directories.readonly).toNIO
-        )
+      isLocalFileSystem(workspace) &&
+        isInReadonlyDirectory(workspace)
+
+    def isWorkspaceSource(workspace: AbsolutePath): Boolean =
+      isLocalFileSystem(workspace) &&
+        !isInReadonlyDirectory(workspace)
+
+    def isLocalFileSystem(workspace: AbsolutePath) =
+      workspace.toNIO.getFileSystem == path.toNIO.getFileSystem
+
+    def isInReadonlyDirectory(workspace: AbsolutePath) =
+      path.toNIO.startsWith(
+        workspace.resolve(Directories.readonly).toNIO
+      )
 
     /**
      * Writes zip file contents to disk under $workspace/.metals/readonly.
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -21,12 +21,18 @@ import org.eclipse.lsp4j.Location
 import scala.meta.internal.metals.MetalsLanguageClient
 import org.eclipse.lsp4j.MessageParams
 import org.eclipse.lsp4j.MessageType
+import org.eclipse.lsp4j.{Range => LSPRange}
+import org.eclipse.lsp4j.jsonrpc.messages.{Either => LSPEither}
 import scala.meta.internal.metals.Buffers
 import java.net.URI
 import java.nio.file.Paths
 import scala.meta.internal.metals.TextEdits
 import scala.meta.internal.metals.Compilations
 import scala.meta.internal.metals.ReferencesResult
+import org.eclipse.lsp4j.TextDocumentEdit
+import org.eclipse.lsp4j.VersionedTextDocumentIdentifier
+import org.eclipse.lsp4j.ResourceOperation
+import org.eclipse.lsp4j.RenameFile
 
 final class RenameProvider(
     referenceProvider: ReferenceProvider,
@@ -39,6 +45,22 @@ final class RenameProvider(
     buffers: Buffers,
     compilations: Compilations
 ) {
+
+  def prepareRename(params: TextDocumentPositionParams): Option[LSPRange] = {
+    if (!compilations.currentlyCompiling.isEmpty) {
+      client.showMessage(isCompiling)
+      None
+    } else {
+      val source = params.getTextDocument.getUri.toAbsolutePath
+      val symbolOccurence =
+        definitionProvider.symbolOccurence(source, params)
+      for {
+        (occurence, semanticDb) <- symbolOccurence
+        if canRenameSymbol(occurence.symbol, None)
+        range <- occurence.range
+      } yield range.toLSP
+    }
+  }
 
   def rename(params: RenameParams): WorkspaceEdit = {
     if (!compilations.currentlyCompiling.isEmpty) {
@@ -59,9 +81,9 @@ final class RenameProvider(
 
       val allReferences = for {
         (occurence, semanticDb) <- symbolOccurence.toIterable
-        if canRenameSymbol(occurence.symbol, params.getNewName())
+        if canRenameSymbol(occurence.symbol, Option(params.getNewName()))
         parentSymbols = implementationProvider
-          .symbolParent(occurence.symbol, semanticDb)
+          .topMethodParents(occurence.symbol, semanticDb)
         txtParams <- if (parentSymbols.isEmpty) List(textParams)
         else parentSymbols.map(toTextParams)
         currentReferences = referenceProvider.references(
@@ -76,19 +98,21 @@ final class RenameProvider(
         loc <- refResult.locations
       } yield loc
 
-      val isApply =
-        symbolOccurence.exists(occ => occ._1.symbol.desc.name.value == "apply")
+      def isOccurence(fn: String => Boolean): Boolean = {
+        symbolOccurence.exists {
+          case (occ, _) => fn(occ.symbol)
+        }
+      }
 
       val allChanges = for {
         (uri, locs) <- allReferences.toList.distinct.groupBy(_.getUri())
       } yield {
         val textEdits = for (loc <- locs) yield {
-          textEdit(isApply, loc, params.getNewName())
+          textEdit(isOccurence, loc, params.getNewName())
         }
         Seq(uri -> textEdits.toList)
       }
       val fileChanges = allChanges.flatten.toMap
-
       val (openedEdits, closedEdits) = fileChanges.partition {
         case (file, edits) =>
           val path = AbsolutePath(Paths.get(new URI(file)))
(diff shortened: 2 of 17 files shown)


metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -0,0 +1,275 @@
...
+    val forbiddenMethods = Set("equals", "hashCode", "unapply", "unary_!", "!")
+    val desc = symbol.desc
+    val name = desc.name.value
+    val isForbidden = forbiddenMethods(name)
+    if (isForbidden) {
+      client.logMessage(forbiddenRename(name, newName))
+    }
+    val colonNotAllowed = name.endsWith(":") && !newName.endsWith(":")
+    if (colonNotAllowed) {
+      forbiddenColonRename(name, newName)
```
olafurpg: Calling this message has no side-effect
```suggestion
      client.logMessage(forbiddenColonRename(name, newName))
```

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -0,0 +1,275 @@
...
+    }
+    symbolIsLocal(symbol) && (!desc.isMethod || (!colonNotAllowed && !isForbidden))
+  }
+
+  private def symbolIsLocal(symbol: String): Boolean = {
+
+    lazy val isFromWorkspace = index
```
olafurpg: Instead of lazy val you can write a method
```suggestion
    def isFromWorkspace = index
```

tests/unit/src/test/scala/tests/RenameSuite.scala
```diff
@@ -0,0 +1,432 @@
...
+       |  def <<method>>(asf: T): Int
+       |}
+       |
+       |trait Alphabet extends Methodable[String] {
+       |  def <<method>>(adf: String) = 123
+       |}
+       |
+       |object Main {
+       |  val a = new Alphabet {
+       |    override def <<me@@thod>>(adf: String): Int = 321
```
olafurpg: 😍

tests/unit/src/test/scala/tests/RenameSuite.scala
```diff
@@ -0,0 +1,432 @@
...
+             |    ],
+             |    "libraryDependencies": [
+             |      "org.scalatest::scalatest:3.0.5",
+             |      "io.circe::circe-generic:0.12.0"
+             |    ]
+             |  }
+             |}
+             |${input.replaceAll(allMarkersRegex, "")}""".stripMargin
+        )
+        _ <- Future.sequence(
```
olafurpg: To stress that we correctly use edit distance, can we artificially add a newline to the top of all open files before applying the rename? The output of the rewrites should be the same, but it will stress that we use token edit distance since the line numbers have shifted.

tests/unit/src/test/scala/tests/RenameSuite.scala
```diff
@@ -0,0 +1,432 @@
...
+          "No `@@` was defined that specifies cursor position"
+        )
+      }
+
+    testAsync(name) {
+      cleanWorkspace()
+      for {
+        _ <- server.initialize(
+          s"""/metals.json
+             |{"a":
```
olafurpg: Can we add test cases that stress the following scenarios?

* Rename symbol across different build targets
* Rename symbol across Scala/Java (just document that Java files are left unchanged)
* Rename symbol in project with compile error still works (we use an old semanticdb snapshot)

tgodzik: Sure, will add that to the tests. It would be super useful to have that explicit.

tgodzik:
Thanks @olafurpg for the review! I will try to address the comments soon, though I am thinking we should merge this after the next release so that we introduce implementation and test/run first. What do you think?

> Did you look into implementing textDocument/prepareRename? https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#textDocument_prepareRename I believe it could help provide faster validation to the user but I'm not 100% sure how it works.

I am not  sure if it helps us a lot. I was hoping there would be some kind of additional window or something like that - it would enable use to choose exact options for rename. However, currently it will do a very similar thing to what we do in rename. I will see if it helps us now.

> Can we use the `documentChanges` field in `WorkspaceEdit` to rename the files? One feature I love in IntelliJ is that it renames the file to match the classname when that class/companion is the only toplevel definition in the file.

Och, I wanted to do that, but totally forgot. Should be simple enough to do.

gabro:
> I am not sure if it helps us a lot. I was hoping there would be some kind of additional window or something like that - it would enable use to choose exact options for rename. However, currently it will do a very similar thing to what we do in rename. I will see if it helps us now.
> 
> >

As far I as understand, `prepareRename` may enable us to reject a rename before the rename input is prompted to the user. With the current flow, the user will first input the new name, and then get an error, which can be slightly annoying.

tgodzik:
> > I am not sure if it helps us a lot. I was hoping there would be some kind of additional window or something like that - it would enable use to choose exact options for rename. However, currently it will do a very similar thing to what we do in rename. I will see if it helps us now.
> > >
> 
> As far I as understand, `prepareRename` may enable us to reject a rename before the rename input is prompted to the user. With the current flow, the user will first input the new name, and then get an error, which can be slightly annoying.

I will check that! That would be much more useful indeed, haven't thought of that.

olafurpg:
Yes, the idea with `prepareRename` would be to give the user faster feedback on when the rename is not supported, ideally, before they type out the new name. For example, the validation to ensure we don't rename during compilation, that we don't rename `equals`, etc.

olafurpg:
>  I am thinking we should merge this after the next release so that we introduce implementation and test/run first. What do you think?

I'm not sure if that's necessary. From the looks of it, this PR is quite complete already with a solid implementation and test cases. I don't think we have to rush for the next release, it's OK if we wait a few more weeks before having a large stable release with a lot of the exciting stuff that's is almost ready (rename symbol, worksheets, breakpoint debugging, pants support)

mdedetrich:
> I'm not sure if that's necessary. From the looks of it, this PR is quite complete already with a solid implementation and test cases. I don't think we have to rush for the next release, it's OK if we wait a few more weeks before having a large stable release with a lot of the exciting stuff that's is almost ready (rename symbol, worksheets, breakpoint debugging, pants support)

I am getting so excited right now...

tests/unit/src/test/scala/tests/RenameSuite.scala
tgodzik: > Rename symbol across different build targets

Added 

> Rename symbol across Scala/Java

Also added

> Rename symbol in project with compile error still works

Added with a type error.

tests/unit/src/test/scala/tests/RenameSuite.scala
tgodzik: Each file is now modified in tests.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala

tgodzik:
@olafurpg I added some more tests as suggested and file renaming in case that a top level symbol matches the file name.

Let me know if there is something more to fix!

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -0,0 +1,342 @@
...
+    symbolIsLocal(symbol) && (!desc.isMethod || (!colonNotAllowed && !isForbidden))
+  }
+
+  private def symbolIsLocal(symbol: String): Boolean = {
+
+    def isFromWorkspace =
+      index
+        .definition(MSymbol(symbol))
+        .exists { definition =>
+          workspace.toNIO.getFileSystem == definition.path.toNIO.getFileSystem &&
```
olafurpg: Duplicate logic from `isDependencySource` in https://github.com/scalameta/metals/blob/f54ca4951c4f0ba16b87c13d7308f475305ac81f/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala#L252

tests/unit/src/main/scala/tests/TestRanges.scala
```diff
@@ -27,4 +30,23 @@ object TestRanges extends RangeReplace {
...
+      code: String,
+      workspaceEdit: WorkspaceEdit
+  ): Option[String] = {
+    for {
+      validLocations <- workspaceEdit
+        .getDocumentChanges()
+        .asScala
+        .find(
+          change =>
+            change.isLeft && change.getLeft.getTextDocument.getUri
```
olafurpg: nit: 
```suggestion
            change.isLeft &&
            change.getLeft.getTextDocument.getUri.contains(file)
```

Scalafmt doesn't insert newlines in infix operators

tests/unit/src/test/scala/tests/RenameSuite.scala
```diff
@@ -0,0 +1,519 @@
...
+    newName = "truth"
+  )
+
+  renamed(
+    "self-type",
+    """|/a/src/main/scala/a/Main.scala
+       |package a
+       |trait <<A@@BC>>
+       |trait Alphabet{
+       |  this: <<ABC>> =>
```
olafurpg: :+1:

tests/unit/src/test/scala/tests/RenameSuite.scala
```diff
@@ -0,0 +1,519 @@
...
+    "java-unchanged",
+    """|/a/src/main/java/a/Other.java
+       |package a;
+       |public class Other{
+       |
+       |}
+       |/a/src/main/scala/a/Main.scala
+       |package a
+       |object Main{
+       |  val other = new <<Oth@@er>>()
```
olafurpg: Should we consider erroring in this case?

tests/unit/src/test/scala/tests/RenameSuite.scala
```diff
@@ -0,0 +1,519 @@
...
+       |
+       |object Main{
+       |  "foo" match {
+       |    case <<Foo>>(s) => ()
+       |  }
+       |}
+       |""".stripMargin,
+    newName = "Bar"
+  )
+
```
olafurpg: Missing test case: named parameter for case class fields

```scala
case class Name(value: String)
Name(value = "42")
  .copy(value = "43")
  .value
new Name(value = "44")
```

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: The second part is negated, so I didn't really see any sensible way to reuse it. I will move ti the extension method and extract the two conditions.

tests/unit/src/test/scala/tests/RenameSuite.scala
tgodzik: We could, but the other way would be to use Metals to at least rename Scala occurrences and rename the rest by hand or via different LSP. 

I thinking maybe add a warning in that case?

tests/unit/src/test/scala/tests/RenameSuite.scala
tgodzik: It seems that references are not being found for 42 nad 44. Added that test case.

tests/unit/src/main/scala/tests/TestRanges.scala
tgodzik: Fixed

tests/unit/src/test/scala/tests/RenameSuite.scala
tgodzik: Seems to be the same issue as https://github.com/scalameta/scalameta/issues/1169

tgodzik:
There are 3 test cases that are not behaving correctly due to some issues in SemanticDB. I will create the issues there and add information about them to this PR.

@sswistun-vl might be able to work on some of those issues.

The `macro-annotation` test case might be probably fixed by a work around, which I will try to follow up with separately.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -250,10 +250,20 @@ object MetalsEnrichments
...
     }
     def isDependencySource(workspace: AbsolutePath): Boolean =
-      workspace.toNIO.getFileSystem == path.toNIO.getFileSystem &&
-        path.toNIO.startsWith(
-          workspace.resolve(Directories.readonly).toNIO
-        )
+      isLocalFileSystem(workspace) &&
+        isInReadonlyDirectory(workspace)
+
+    def isWorkspaceSource(workspace: AbsolutePath): Boolean =
```
olafurpg: `path.startsWith(workspace)` should strictly speaking be a condition as well

tgodzik: Added!

tests/unit/src/test/scala/tests/RenameSuite.scala
tgodzik: I added a warning in a case like that.

tgodzik:
@olafurpg Thanks again for the review! I fixed the issue with scalafix and added a warning when renaming a symbol from a java file. 

I agree that we can merge this PR and rework the last issues later on. For sure some things will come up when people start using the feature.

olafurpg:
I'm not sure why but the worksheet completion test seems flaky on Windows

```
X tests.worksheets.WorksheetLspSuite.completion 5755ms 
```
Might be best to disable it for now and I'll open up a ticket that I will followup on later.

olafurpg:
I can disable it in a followup PR, feel free to merge despite CI failure

tgodzik:
> I can disable it in a followup PR, feel free to merge despite CI failure

I can try to take a look at some point. I have everything setup on a Windows machine at home.

ckipp01:
Really excited to start using this! 🎉

PR #1050, opened 2019-11-07
Add Coursier cache for Github Actions

sswistun-vl:
Before, all dependencies were downloaded before each CI job. Now, the dependencies are cached so that the jobs can start compiling and running faster.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,32 +18,28 @@ jobs:
       - name: Run unit tests
         run: bin/test.sh "unit/test"
         shell: bash
-      - uses: actions/cache@v1
-        with:
-          path: ${%LOCALAPPDATA%\Coursier\Cache\v1}
-          key: ${{ runner.os }}-coursier-ci
-          restore-keys: |
-            ${{ runner.os }}-coursier-
   linux:
     name: Linux unit tests
     runs-on: ubuntu-latest
     steps:
       - uses: actions/cache@v1
         with:
-          path: ~/.cache/coursier/v1
-          key: ${{ runner.os }}-coursier-ci
+          path: ~/.cache/coursier/v1/https/repo1.maven.org/maven2/com
+          key: ${{ runner.os }}-coursier-com-ci
           restore-keys: |
-            ${{ runner.os }}-coursier-
+            ${{ runner.os }}-coursier-com
+      - uses: actions/cache@v1
+        with:
+          path: ~/.cache/coursier/v1/https/repo1.maven.org/maven2/org
+          key: ${{ runner.os }}-coursier-org-ci
+          restore-keys: |
+            ${{ runner.os }}-coursier-org
       - uses: actions/checkout@v1
       - uses: olafurpg/setup-scala@v5
       - name: Run unit tests
         run: bin/test.sh unit/test
-      - uses: actions/cache@v1
-        with:
-          path: ~/.cache/coursier/v1
-          key: ${{ runner.os }}-coursier-ci
-          restore-keys: |
-            ${{ runner.os }}-coursier-
+      - name: List cache dir
+        run: du -sch ~/.cache/coursier/v1/https/repo1.maven.org/maven2/*|sort -h
   sbt:
     name: Sbt integration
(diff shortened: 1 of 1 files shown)


sswistun-vl:
I thought that CI didn't run because it was draft, but it won't run even now.

olafurpg:
CI runs for draft PRs

tgodzik:
I think Github doesn't understand the changed configuration - had the same when I misspelled something. It doesn't even report any error.

olafurpg:
It's usually best to experiment/iterate on changes like these in tiny repos like https://github.com/olafurpg/sbt-ci-release and only once that's working do the same for this repo

gabro:
I think `with` needs to be at the same indentation level as `uses`

tgodzik:
> It's usually best to experiment/iterate on changes like these in tiny repos like https://github.com/olafurpg/sbt-ci-release and only once that's working do the same for this repo

I asked @sswistun-vl if he could figure out how to use caching here. We can for sure also experiment in sbt-ci-release, but I was hoping for some quick gains in build times and see if it even make sense.

olafurpg:
@tgodzik I was just suggesting to experiment with the new cache feature in sbt-ci-release since it's a smaller project and it might provide a faster feedback cycle while iterating on changes. 

This current PR tries to enable caching for all jobs, and one syntax error in one place could fail the whole CI. The smaller the change, the easier it will be to track down the problem

tgodzik:
> @tgodzik I was just suggesting to experiment with the new cache feature in sbt-ci-release since it's a smaller project and it might provide a faster feedback cycle while iterating on changes.
> 
> This current PR tries to enable caching for all jobs, and one syntax error in one place could fail the whole CI. The smaller the change, the easier it will be to track down the problem

Och right, that makes a lot of sense, If we can't fix it quickly then we should for sure experiment in another smaller repo.

sswistun-vl:
Ok, now CI starts but actions/cache@v1 fails. Thanks for noticing this mistake.

sswistun-vl:
It seems like resulting cache is too big for Actions Cache feature.

tgodzik:
> It seems like resulting cache is too big for Actions Cache feature.
> We could try to compress it manually before uploading it, but it would add few additional steps.

We could create multiple caches I think. The limit is for an individual cache.

.github/workflows/ci.yml
```diff
@@ -7,23 +7,53 @@ jobs:
...
+      - uses: actions/cache@v1
+        with:
+          path: ${%LOCALAPPDATA%\Coursier\Cache\v1}
+          key: ${{ runner.os }}-coursier-ci
+          restore-keys: |
+            ${{ runner.os }}-coursier-
       - uses: actions/checkout@v1
       - uses: olafurpg/setup-scala@v5
       - name: Run unit tests
         run: bin/test.sh "unit/test"
```
tgodzik: Can we actually have a build matrix with different commands to run? There seems to be more and more repeated code.

sswistun-vl: I'll work on it when I manage to get cache to work.

olafurpg: I think it's better to generate `ci.yml` from build.sbt, I experimented with that approach here https://github.com/scalameta/metals/pull/940

tgodzik: Sbt seems like an overkill. It can easily be done via script or I think it would be much better to solve it using the actions syntax. Why would using `sbt` be neccessary? We do not use any information from the build itself.

olafurpg: Using sbt would allow us to generate a custom `cross/test` job for `crossScalaVersions`

tgodzik:
I am closing this PR since 400 MB for cache is not enough and it seems to be a bigger problem as discussed here: https://github.com/actions/cache/issues/6

We can reiterate on this once the limit is bumped.

imbsky:
There are still some concerns, but the cache limit has been bumped to 2GB.

tgodzik:
> There are still some concerns, but the cache limit has been bumped to 2GB.

Cool! We will try it out then, Thanks!

PR #1051, opened 2019-11-08
Add an empty line after the package declaration before import

### LINKED ISSUE #1054: Insert an empty line after the package declaration and before the first auto-import
**Describe the bug**
If I create a new scala file, and declare a package declaration. The first auto-import would insert this import after the  package declaration like this:
```
package com.example
import scala.annotation.tailrec
```

**To Reproduce**
1. Create a new scala file.
2. Add a package declaration.
3. Write a simple code to trigger the first auto-import.

**Expected behavior**
A new empty line should be inserted before this import.
```
package com.example

import scala.annotation.tailrec
```
Most people write Scala use this code style.

**Installation:**
- Operating system: Linux
- Editor: Visual Studio Code/Emacs
- Metals version: v0.7.0

**Additional context**
Though scalafix can (may, I haven't try) fix this problem, but I think the Metals should provide this good code style by default.

ckipp01:
This will close #1054

Checks to see if there are any imports and if not, but there is a package declaration, it will add an extra new line to add padding between the package declaration and the first import.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/AutoImports.scala
@@ -10,7 +10,7 @@ trait AutoImports { this: MetalsGlobal =>
    * @param offset the offset where to place the import.
    * @param indent the indentation at which to place the import.
    * @param padTop whether the import needs to be padded on top
-   *               in the case that it is the first one after the pacage def
+   *               in the case that it is the first one after the package def
    */
   case class AutoImportPosition(
       offset: Int,
@@ -54,20 +54,22 @@ trait AutoImports { this: MetalsGlobal =>
           case Some(pkg)
               if pkg.symbol != rootMirror.EmptyPackage ||
                 pkg.stats.headOption.exists(_.isInstanceOf[Import]) =>
-            val lastImport: (Tree, Boolean) = pkg.stats
+            val lastImport = pkg.stats
               .takeWhile(_.isInstanceOf[Import])
-              .lastOption // if there are no imports, return true to pad top of import
-              .fold[(Tree, Boolean)]((pkg.pid, true))((tree => (tree, false)))
+              .lastOption
+
+            val padTop = lastImport.isEmpty
+            val lastImportOrPkg = lastImport.getOrElse(pkg.pid)
 
             Some(
               new AutoImportPosition(
-                pos.source.lineToOffset(lastImport._1.pos.focusEnd.line),
+                pos.source.lineToOffset(lastImportOrPkg.pos.focusEnd.line),
                 text,
-                lastImport._2
+                padTop
               )
             )
           case _ =>
-            Some(AutoImportPosition(0, 0, false))
+            Some(AutoImportPosition(0, 0, padTop = false))
         }
     }
   }
--- a/mtags/src/main/scala/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/AutoImports.scala
@@ -9,10 +9,16 @@ trait AutoImports { this: MetalsGlobal =>
    *
    * @param offset the offset where to place the import.
    * @param indent the indentation at which to place the import.
+   * @param padTop whether the import needs to be padded on top
+   *               in the case that it is the first one after the pacage def
    */
-  case class AutoImportPosition(offset: Int, indent: Int) {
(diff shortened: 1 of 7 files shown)


mtags/src/main/scala/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -48,18 +54,20 @@ trait AutoImports { this: MetalsGlobal =>
           case Some(pkg)
               if pkg.symbol != rootMirror.EmptyPackage ||
                 pkg.stats.headOption.exists(_.isInstanceOf[Import]) =>
-            val lastImport = pkg.stats
+            val lastImport: (Tree, Boolean) = pkg.stats
               .takeWhile(_.isInstanceOf[Import])
-              .lastOption
-              .getOrElse(pkg.pid)
+              .lastOption // if there are no imports, return true to pad top of import
+              .fold[(Tree, Boolean)]((pkg.pid, true))((tree => (tree, false)))
```
gabro: is the explicit type annotation here necessary?

mtags/src/main/scala/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -48,18 +54,20 @@ trait AutoImports { this: MetalsGlobal =>
           case Some(pkg)
               if pkg.symbol != rootMirror.EmptyPackage ||
                 pkg.stats.headOption.exists(_.isInstanceOf[Import]) =>
-            val lastImport = pkg.stats
+            val lastImport: (Tree, Boolean) = pkg.stats
```
gabro: I would use a pattern here so to avoid the tuple accessors below, which are hard to read

```suggestion
            val (lastImportOrPkg, padTop): (Tree, Boolean) = pkg.stats
```

mtags/src/main/scala/scala/meta/internal/pc/AutoImports.scala
tgodzik: Maybe just use separate values for each?
```scala
val lastImportOpt = pkg.stats
              .takeWhile(_.isInstanceOf[Import])
              .lastOption
val noImport = lastImport.isEmpty
val lastImport = lastImportOpt.getOrElse(pkg.pid)

```

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1096,7 +1096,7 @@ trait Completions { this: MetalsGlobal =>
       val baseAutoImport: Option[AutoImportPosition] =
         autoImportPosition(pos, text)
       val autoImport: AutoImportPosition = baseAutoImport.getOrElse(
-        AutoImportPosition(lineStart, inferIndent(lineStart, text))
+        AutoImportPosition(lineStart, inferIndent(lineStart, text), false)
```
tgodzik: `padTop = false` ?

mtags/src/main/scala/scala/meta/internal/pc/AutoImports.scala
ckipp01: Ahh that's much nicer. Thanks!

mtags/src/main/scala/scala/meta/internal/pc/AutoImports.scala
ckipp01: Without it I get a type mismatch where it wants a RefTree on the right where it's just a Tree.

gabro: ah, I would've assumed it were unnecessary since the entire expression is annotated anyway. Too bad

ckipp01: I suppose that is much easier to read 👍

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
ckipp01: Shoot, forgot that 😬. Will fix.

ckipp01: Actually, this seems to be what I intended after looking closer. If I understand this correct, during the completion, if there is no `baseAutoImport` from above, then it puts the import inline like it does in the exhaustive match down around line 1230. So it can be assumed since that's happening that no padding should be added.

PR #1056, opened 2019-11-10
Handle abstract var overrides

### LINKED ISSUE #1029: Override abstract var not behaving as expected

gabro:
Fixes #1029 

Previously we would prevent overriding `var` members altogether, recently there seems to be been a regression where we suggest `def` completions for `var` members (#1029); now we allow overriding abstract `var` members, while still rejecting concrete `var` overrides.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -452,7 +452,7 @@ trait Completions { this: MetalsGlobal =>
           pos,
           text,
           valdef.pos.start,
-          _.isStable
+          _ => true
         )
       case (m @ Match(_, Nil)) :: parent :: _ =>
         CompletionPosition.CaseKeyword(m.selector, editRange, pos, text, parent)
@@ -1076,6 +1076,13 @@ trait Completions { this: MetalsGlobal =>
         isVisited
       }
 
+      // NOTE(gabro): sym.isVar does not work consistently across Scala versions
+      // Specifically, it behaves differently between 2.11 and 2.12/2.13
+      // This check is borrowed from
+      // https://github.com/scala/scala/blob/f389823ef0416612a0058a80c1fe85948ff5fc0a/src/reflect/scala/reflect/internal/Symbols.scala#L2645
+      def isVarSetter(sym: Symbol): Boolean =
+        !sym.isStable && !sym.isLazy && sym.isAccessor
+
       // Returns true if this symbol is a method that we can override.
       def isOverridableMethod(sym: Symbol): Boolean = {
         sym.isMethod &&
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -452,7 +452,7 @@ trait Completions { this: MetalsGlobal =>
           pos,
           text,
           valdef.pos.start,
-          _.isStable
+          _ => true
```
gabro: `vars` are not stable symbols, but we can still override them in some cases

tests/cross/src/test/scala/tests/pc/CompletionOverrideSuite.scala
```diff
@@ -648,10 +648,60 @@ object CompletionOverrideSuite extends BaseCompletionSuite {
...
+       |abstract class Val {
+       |  var hello1: Int
+       |  val hello2: Int
+       |  def hello3: Int
+       |}
+       |class Main extends Val {
+       |  override val hello@@
+       }
+       |""".stripMargin,
+    // NOTE(gabro): we inlcude also var and def completions despite the user typed a val
```
gabro: I honestly couldn't make this work, but it's not too terrible as UX

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1087,7 +1094,7 @@ trait Completions { this: MetalsGlobal =>
         !sym.isEffectivelyFinal &&
         !sym.name.endsWith(CURSOR) &&
         !sym.isConstructor &&
-        !sym.isMutable &&
+        (!isVarSetter(sym) || (isVarSetter(sym) && sym.isAbstract)) &&
```
gabro: allow overriding vars, but only if they're abstract

ckipp01:
Awesome! This is exactly what I needed! I just locally rebased on this and my issues with #1031 go away entirely and all tests pass for 2.11, 2.12, and 2.13. Thanks for looking at this 👏

tests/cross/src/test/scala/tests/pc/CompletionOverrideSuite.scala
olafurpg: I think this UX is totally fine, you don't always know if the supermethod is a def, val or a var

PR #1057, opened 2019-11-10
Respect client capabilities for completion item snippets

### LINKED ISSUE #1055: Metals returns snippets even when a client doesn't support them
**Describe the bug**

The protocol doesn't require clients to support snippet completion. For those clients, the servers are supposed to only provide plain text completion.

One such client is ycmd, whose capabilities you can see [here](https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/language_server/language_server_protocol.py#L236)
As you can see, the capabilities don't contain `{ 'completionItem': { 'snippetSupport': True } }` and, according to the protocol, the lack of  `'snippetSupport'` is supposed to be treated as `'snippetSupport': False`.

Unfortunately, completion request returns:

    2019-11-09 00:13:32,982 - DEBUG - RX: Received message: b'{"jsonrpc":"2.0","id":20,"result":{"isIncomplete":false,"items":[{"label":"getRemoteEndpoint(): RemoteEndpoint","kind":8,"detail":"(): RemoteEndpoint","preselect":true,"sortText":"00000","filterText":"getRemoteEndpoint","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"getRemoteEndpoint"},"data":{"symbol":"org/eclipse/lsp4j/jsonrpc/Launcher#getRemoteEndpoint().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"getRemoteProxy(): MetalsLanguageClient","kind":8,"detail":"(): MetalsLanguageClient","sortText":"00001","filterText":"getRemoteProxy","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"getRemoteProxy"},"data":{"symbol":"org/eclipse/lsp4j/jsonrpc/Launcher#getRemoteProxy().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"startListening(): Future[Void]","kind":8,"detail":"(): Future[Void]","sortText":"00002","filterText":"startListening","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":29}},"newText":"startListening"},"data":{"symbol":"org/eclipse/lsp4j/jsonrpc/Launcher#startListening().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"+(other: String): String","kind":2,"detail":"(other: String): String","sortText":"00003","filterText":"+","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"+"},"data":{"symbol":"scala/Predef.any2stringadd#`+`().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"formatted(fmtstr: String): String","kind":2,"detail":"(fmtstr: String): String","sortText":"00004","filterText":"formatted","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"formatted"},"data":{"symbol":"scala/Predef.StringFormat#formatted().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"parse[U](implicit convert: Convert[Launcher[MetalsLanguageClient],Input], parse: Parse[U], dialect: Dialect): Parsed[U]","kind":2,"detail":"[U](implicit convert: Convert[Launcher[MetalsLanguageClient],Input], parse: Parse[U], dialect: Dialect): Parsed[U]","sortText":"00005","filterText":"parse","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"parse"},"data":{"symbol":"scala/meta/parsers/Api#XtensionParseInputLike#parse().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"show[Style[X] \\u003c: scala.meta.prettyprinters.Show[X]](implicit style: Style[Launcher[MetalsLanguageClient]]): String","kind":2,"detail":"[Style[X] \\u003c: scala.meta.prettyprinters.Show[X]](implicit style: Style[Launcher[MetalsLanguageClient]]): String","sortText":"00006","filterText":"show","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"show"},"data":{"symbol":"scala/meta/prettyprinters/Api#XtensionShow#show().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"tokenize(implicit convert: Convert[Launcher[MetalsLanguageClient],Input], tokenize: Tokenize, dialect: Dialect): Tokenized","kind":2,"detail":"(implicit convert: Convert[Launcher[MetalsLanguageClient],Input], tokenize: Tokenize, dialect: Dialect): Tokenized","sortText":"00007","filterText":"tokenize","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"tokenize"},"data":{"symbol":"scala/meta/tokenizers/Api#XtensionTokenizeInputLike#tokenize().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"asInstanceOf[T0]: T0","kind":2,"detail":"[T0]: T0","sortText":"00008","filterText":"asInstanceOf","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"asInstanceOf"},"data":{"symbol":"scala/Any#asInstanceOf().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"equals(x$1: Any): Boolean","kind":8,"detail":"(x$1: Any): Boolean","sortText":"00009","filterText":"equals","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"equals"},"data":{"symbol":"java/lang/Object#equals().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"getClass(): Class[_]","kind":8,"detail":"(): Class[_]","sortText":"00010","filterText":"getClass","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"getClass"},"data":{"symbol":"java/lang/Object#getClass().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"hashCode(): Int","kind":8,"detail":"(): Int","sortText":"00011","filterText":"hashCode","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"hashCode"},"data":{"symbol":"java/lang/Object#hashCode().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"isInstanceOf[T0]: Boolean","kind":2,"detail":"[T0]: Boolean","sortText":"00012","filterText":"isInstanceOf","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"isInstanceOf"},"data":{"symbol":"scala/Any#isInstanceOf().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"synchronized[T0](x$1: T0): T0","kind":2,"detail":"[T0](x$1: T0): T0","sortText":"00013","filterText":"synchronized","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"synchronized"},"data":{"symbol":"java/lang/Object#synchronized().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}},{"label":"toString(): String","kind":8,"detail":"(): String","sortText":"00014","filterText":"toString","insertTextFormat":2,"textEdit":{"range":{"start":{"line":43,"character":15},"end":{"line":43,"character":15}},"newText":"toString"},"data":{"symbol":"java/lang/Object#toString().","target":"file:/home/bstaletic/aur/metals/src/metals-0.7.6/metals/?id\\u003dmetals"}}]}}'

The completion items contain `"inserTextFormat":2`, which indicates snippets and triggers [an assert in ycmd](https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/language_server/language_server_completer.py#L2251).

**To Reproduce**
1. Install YouCompleteMe for vim
2. Unfortunately, ycmd still doesn't support `showMessageRequest`, so one will need to apply [this patch](https://gist.github.com/bstaletic/bdb40ed788de1b85d1ef5b7c4c02199e).
3. Put `let g:ycm_language_server = [ { 'name': 'metals-showMessageRequest', 'filetypes': [ 'scala', 'sbt' ],   'cmdline': [ 'metals-vim' ], 'project_root_files': [ 'build.sbt' ] } ]` in your vimrc.
4. Open a scala file and wait for metals to finish building
5. Try to complete
6. The stack trace of the assert can be found in one of the logs visible in `:YcmDebugInfo`.

**Expected behavior**

**Screenshots**

**Installation:**
- Operating system: Linux
- Editor: Vim with YouCompleteMe
- Metals version: v0.7.6

**Additional context**

**Search terms**

gabro:
Fixes #1055 

Previously we always used snippets in completion items, regardless of whether the client declared that it supported them. Now we respect the capabilities declared by the client.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -509,6 +509,16 @@ object MetalsEnrichments
         )
       } yield hierarchicalDocumentSymbolSupport.booleanValue).getOrElse(false)
 
+    def supportsCompletionSnippets: Boolean =
+      (for {
+        params <- initializeParams
+        capabilities <- Option(params.getCapabilities)
+        textDocument <- Option(capabilities.getTextDocument)
+        completion <- Option(textDocument.getCompletion)
+        completionItem <- Option(completion.getCompletionItem)
+        snippetSupport <- Option(completionItem.getSnippetSupport())
+      } yield snippetSupport.booleanValue).getOrElse(false)
+
   }
 
   implicit class XtensionPromise[T](promise: Promise[T]) {
--- a/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
@@ -16,7 +16,8 @@ import org.eclipse.{lsp4j => l}
 
 class CompletionProvider(
     val compiler: MetalsGlobal,
-    params: OffsetParams
+    params: OffsetParams,
+    clientSupportsSnippets: Boolean
 ) {
   import compiler._
 
@@ -88,13 +89,13 @@ class CompletionProvider(
           item.setDetail(detail)
         }
         val templateSuffix =
-          if (!isSnippet) ""
+          if (!isSnippet || !clientSupportsSnippets) ""
           else if (completion.isNew &&
             r.sym.dealiased.requiresTemplateCurlyBraces) " {}"
           else ""
 
         val typeSuffix =
-          if (!isSnippet) ""
+          if (!isSnippet || !clientSupportsSnippets) ""
           else if (completion.isType && r.sym.hasTypeParams) "[$0]"
           else if (completion.isNew && r.sym.hasTypeParams) "[$0]"
           else ""
@@ -111,7 +112,11 @@ class CompletionProvider(
             item.setFilterText(symbolName)
         }
 
-        item.setInsertTextFormat(InsertTextFormat.Snippet)
+        if (clientSupportsSnippets) {
+          item.setInsertTextFormat(InsertTextFormat.Snippet)
+        } else {
+          item.setInsertTextFormat(InsertTextFormat.PlainText)
+        }
 
         r match {
           case i: TextEditMember =>
@@ -151,7 +156,9 @@ class CompletionProvider(
                 case head :: Nil if head.forall(_.isImplicit) =>
                   () // Don't set ($0) snippet for implicit-only params.
                 case _ =>
-                  item.setTextEdit(textEdit(baseLabel + "($0)"))
+                  if (clientSupportsSnippets) {
+                    item.setTextEdit(textEdit(baseLabel + "($0)"))
+                  }
                   metalsConfig
                     .parameterHintsCommand()
                     .asScala
--- a/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
@@ -16,8 +16,7 @@ import org.eclipse.{lsp4j => l}
 
 class CompletionProvider(
     val compiler: MetalsGlobal,
-    params: OffsetParams,
-    clientSupportsSnippets: Boolean
+    params: OffsetParams
 ) {
   import compiler._
 
@@ -45,6 +44,8 @@ class CompletionProvider(
     )
     val pos = unit.position(params.offset)
     val isSnippet = isSnippetEnabled(pos, params.text())
+    val clientSupportsSnippets =
+      compiler.metalsConfig.isCompletionSnippetsEnabled()
     val (i, completion, editRange, query) = safeCompletionsAt(pos)
     val start = inferIdentStart(pos, params.text())
     val end = inferIdentEnd(pos, params.text())
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -19,6 +19,9 @@ import scala.meta.internal.tokenizers.Chars
  */
 trait Completions { this: MetalsGlobal =>
 
+  val clientSupportsSnippets: Boolean =
+    metalsConfig.isCompletionSnippetsEnabled()
+
   /**
    * A member for symbols on the classpath that are not in scope, produced via workspace/symbol.
    */
@@ -765,11 +768,11 @@ trait Completions { this: MetalsGlobal =>
           if (text.charAt(lit.pos.start - 1) != 's')
             List(new l.TextEdit(lit.pos.withEnd(lit.pos.start).toLSP, "s"))
           else Nil
-        val dolarEdits = for {
+        val dollarEdits = for {
           i <- lit.pos.start to (lit.pos.end - CURSOR.length())
           if text.charAt(i) == '$' && i != interpolator.dollar
         } yield new l.TextEdit(pos.source.position(i).withEnd(i).toLSP, "$")
-        interpolatorEdit ++ dolarEdits
+        interpolatorEdit ++ dollarEdits
       }
 
       def newText(sym: Symbol): String = {
@@ -793,6 +796,7 @@ trait Completions { this: MetalsGlobal =>
 
       val filter: String =
         text.substring(lit.pos.start, pos.point - interpolator.name.length)
+
       override def contribute: List[Member] = {
         metalsScopeMembers(pos).collect {
           case s: ScopeMember
@@ -985,7 +989,7 @@ trait Completions { this: MetalsGlobal =>
           allParams.exists(param => param.name.startsWith(prefix))
         val isExplicitlyCalled = suffix.startsWith(prefix)
         val hasParamsToFill = allParams.count(!_.hasDefault) > 1
-        if ((shouldShow || isExplicitlyCalled) && hasParamsToFill) {
+        if ((shouldShow || isExplicitlyCalled) && hasParamsToFill && clientSupportsSnippets) {
           val editText = allParams.zipWithIndex
             .collect {
               case (param, index) if !param.hasDefault =>
@@ -1202,7 +1206,11 @@ trait Completions { this: MetalsGlobal =>
         private def signature = printer.defaultMethodSignature()
         private def edit = new l.TextEdit(
           range,
-          s"$filterText$signature = $${0:???}"
+          if (clientSupportsSnippets) {
+            s"$filterText$signature = $${0:???}"
+          } else {
+            s"$filterText$signature = ???"
+          }
         )
       }
 
@@ -1259,7 +1267,11 @@ trait Completions { this: MetalsGlobal =>
           "match",
           new l.TextEdit(
             editRange,
-            "match {\n\tcase$0\n}"
+            if (clientSupportsSnippets) {
+              "match {\n\tcase$0\n}"
+            } else {
+              "match"
+            }
           ),
           completionsSymbol("match"),
           label = Some("match"),
@@ -1273,7 +1285,11 @@ trait Completions { this: MetalsGlobal =>
               tail
                 .map(_.edit.getNewText())
                 .mkString(
-                  s"match {\n\t${head.edit.getNewText} $$0\n\t",
+                  if (clientSupportsSnippets) {
+                    s"match {\n\t${head.edit.getNewText} $$0\n\t"
+                  } else {
+                    s"match {\n\t${head.edit.getNewText}\n\t"
+                  },
                   "\n\t",
                   "\n}"
                 )
@@ -1408,7 +1424,10 @@ trait Completions { this: MetalsGlobal =>
         if (definitions.isTupleType(parents.selector)) {
           result += new TextEditMember(
             "case () =>",
-            new l.TextEdit(editRange, "case ($0) =>"),
+            new l.TextEdit(
+              editRange,
+              if (clientSupportsSnippets) "case ($0) =>" else "case () =>"
+            ),
             parents.selector.typeSymbol,
             label = Some(s"case ${parents.selector} =>"),
             command = metalsConfig.parameterHintsCommand().asScala
@@ -1466,8 +1485,10 @@ trait Completions { this: MetalsGlobal =>
         val label = s"case $pattern =>"
         new TextEditMember(
           filterText = label,
-          edit =
-            new l.TextEdit(editRange, label + (if (isSnippet) " $0" else "")),
+          edit = new l.TextEdit(
+            editRange,
+            label + (if (isSnippet && clientSupportsSnippets) " $0" else "")
+          ),
           sym = sym,
           label = Some(label),
           additionalTextEdits = autoImports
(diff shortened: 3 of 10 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -509,6 +509,16 @@ object MetalsEnrichments
         )
       } yield hierarchicalDocumentSymbolSupport.booleanValue).getOrElse(false)
 
+    def supportsCompletionSnippets: Boolean =
```
olafurpg: 🙈

olafurpg: Wish this was easier to write...

olafurpg:
```scala
try initializeParams.getCapabilities.getTextDocument.getCompletion.getCompletionItem.getSnippetSupport.booleanValue
catch { case _: NPE => false }
```
 😅

mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -16,7 +16,8 @@ import org.eclipse.{lsp4j => l}
 
 class CompletionProvider(
     val compiler: MetalsGlobal,
-    params: OffsetParams
+    params: OffsetParams,
+    clientSupportsSnippets: Boolean
```
olafurpg: FYI, `compiler.metalsConfig` already has this info

bstaletic:
Hello everyone. I tried out this patch briefly with ycmd and completions started working. @gabro Thanks again for working on this.

mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
gabro: ah, makes sense! I even thought about it, but then I forgot 😄 I’ll fix it

gabro:
@tgodzik thanks for the review and woops, yes, I've missed a lot of cases from `Completion.scala`! I've addressed all the ones I could spot and taken some UX decisions (I'll comment inline) that would need some validation from you and @olafurpg.

---

> Hello everyone. I tried out this patch briefly with ycmd and completions started working. @gabro Thanks again for working on this.

Thanks for testing this, @bstaletic!

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
gabro: poor man's monad :P

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -985,7 +989,7 @@ trait Completions { this: MetalsGlobal =>
           allParams.exists(param => param.name.startsWith(prefix))
         val isExplicitlyCalled = suffix.startsWith(prefix)
         val hasParamsToFill = allParams.count(!_.hasDefault) > 1
-        if ((shouldShow || isExplicitlyCalled) && hasParamsToFill) {
+        if ((shouldShow || isExplicitlyCalled) && hasParamsToFill && clientSupportsSnippets) {
```
gabro: this disables the `Autofill with default values` completion when the client does not support snippets, as per @tgodzik. Please review that it makes sense and that we can't find a good completion that works without snippets

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1259,7 +1267,11 @@ trait Completions { this: MetalsGlobal =>
           "match",
           new l.TextEdit(
             editRange,
-            "match {\n\tcase$0\n}"
+            if (clientSupportsSnippets) {
+              "match {\n\tcase$0\n}"
+            } else {
+              "match"
```
gabro: avoid the braces if we have no snippets. Inserting braces will result in a cursor out of place (`{}<CURSOR>`), which seems bad.

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1273,7 +1285,11 @@ trait Completions { this: MetalsGlobal =>
               tail
                 .map(_.edit.getNewText())
                 .mkString(
-                  s"match {\n\t${head.edit.getNewText} $$0\n\t",
+                  if (clientSupportsSnippets) {
+                    "match {\n\t${head.edit.getNewText} $$0\n\t"
+                  } else {
+                    "match {\n\t${head.edit.getNewText}\n\t"
```
gabro: in this case we're ok with an "annoying" cursor position (`match {...}<CURSOR>`) since the exhaustive match feature is quite useful regardless (and it's opt-in anyway)

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1408,7 +1424,10 @@ trait Completions { this: MetalsGlobal =>
         if (definitions.isTupleType(parents.selector)) {
           result += new TextEditMember(
             "case () =>",
-            new l.TextEdit(editRange, "case ($0) =>"),
+            new l.TextEdit(
+              editRange,
+              if (clientSupportsSnippets) "case ($0) =>" else "case () =>"
```
gabro: not brilliant, but again, it's opt-in.

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
bstaletic: From my point of view, it is bad to end up with `foo{}<CURSOR>` after completion. I ran into a similar case with Apple's swift server and it forced me to do [this](https://github.com/bstaletic/ycmd/blob/swift/ycmd/completers/swift/swift_completer.py#L106-L110) - overriding completion algorithm for swift to strip `()` from `foo()`, even if there's something in between `(` and `)`.

gabro: thanks for the feedback! I'm actively avoiding that pattern, except in some cases where the completion is still very useful and/or it's opt-in (e.g. it's not the default case)

gabro:
@bstaletic curious to hear your input about one particular case: Scala allows interpolating strings with params and if you have a "simple" param you can do `$param` whereas if it's a "complicated" expression, you have to surround it with braces. Metals allows you to complete on the "simple" expression and it inserts braces for you:

```scala
// from
s"$buffer.app<CURSOR>"

//to
s"${buffer.append(<CURSOR>)}"
```

This is not possible without snippets, so what would you prefer as a language client?

1. ```s"$buffer.app<CURSOR>"``` -> no completions
2. ```s"$buffer.app<CURSOR>"``` -> s"${buffer.append}<CURSOR>"
3. ```s"$buffer.app<CURSOR>"``` -> s"$buffer.append" <- not what you want, this is potentially a bug

I'm leaning towards 1 but the current implementation in this PR would yield 2

bstaletic:
I'm leaning towards 1 as well. If a user complains that such a case isn't getting completions I can point the user to the logs and say no completions are returned by metals.

The option 2 doesn't seem good for the same reason `foo()` doesn't. On the other hand, I've never written a line of scala.

Tell me if I understand option 3 and scala parsing rules correctly. Without braces `$` in string interpolation refers to a single identifier. In case of `s"$buffer.append` (without braces) it would be parsed as `s"${buffer}.append"`. In general case this might compile and is therefore impossible to catch with errors/warnings. If all of the above is correct, I'd definitely avoid option 3.

@puremourning Do you have anything to add?

bstaletic:
To add to the above, I won't mind option 2 either. If option 1 is too much trouble, oh well, I won't lose any sleep over it.

gabro:
> Tell me if I understand option 3 and scala parsing rules correctly. Without braces `$` in string interpolation refers to a single identifier. In case of `s"$buffer.append` (without braces) it would be parsed as `s"${buffer}.append"`. In general case this might compile and is therefore impossible to catch with errors/warnings. If all of the above is correct, I'd definitely avoid option 3.

Correct.


Thanks for the feedback @bstaletic, I'll whether option 1 is worth the hassle :)

puremourning:
> @puremourning Do you have anything to add?

Idiomatically it should be:

*`${buffer.app` -> `${buffer.append`, with the suggestions being of the form `append`, `apple`, `appropriate`
* `$buffer.app` -> `${buffer.append`, with the suggestions all being of the form `{buffer.append`, `{buffer.apple`, `{buffer.appropriate`

ycmd would handle the correct prefix selection (assuming compliant `TextEdit`s are returned).

Basically, I don't think you should insert anything to the "right" for the word being completed, ever. Some servers love to try and insert `;` or `()` or whatever, but this is almost always more wrong than right, or at least not consistent enough to be useful.

IM(ns?)HO !

gabro:
> Basically, I don't think you should insert anything to the "right" for the word being completed, ever. Some servers love to try and insert `;` or `()` or whatever, but this is almost always more wrong than right, or at least not consistent enough to be useful.

Thanks for the feedback, that's the general direction I'm following when there's no snippet support in the client.

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
tgodzik: Let's disable it for now, I will think it over to see if it makes sense without the snippets.

gabro:
@puremourning / @bstaletic in the end, is quite finicky to tweak that particular use case, and it's quite a niche use case so - for now - I'll merge this as is (option 2 above)

PR #1065, opened 2019-11-14
Implement code action for importing missing symbols

gabro:
This is draft PR to validate a possible architecture for code actions (aka refactors).

For now, I'm focusing on code actions that resolve existing diagnostics (i.e. quickfixes), but the architecture could be expanded to more general refactors.

This PR implements a working prototype ~(although it's a bit of a hack)~ for the "add missing import" quickfix, which gets triggered by the relevant compiler error.

If we're happy with the general structure, I can move on with the testing infrastructure for code actions and refining the actual implementation.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
@@ -1,0 +1,28 @@
+package scala.meta.internal.metals
+
+import org.eclipse.{lsp4j => l}
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.concurrent.Future
+import scala.meta.pc.CancelToken
+import scala.concurrent.ExecutionContext
+
+final class CodeActionProvider(
+    compilers: Compilers
+) extends QuickFixes {
+
+  def codeActions(
+      params: l.CodeActionParams,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+    for {
+      quickFixes <- QuickFix.fromDiagnostics(
+        params.getContext().getDiagnostics().asScala.toSeq,
+        params,
+        compilers,
+        token
+      )
+    } yield quickFixes
+
+  }
+
+}
--- a/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
@@ -1,26 +1,26 @@
 package scala.meta.internal.metals
 
 import org.eclipse.{lsp4j => l}
-import scala.meta.internal.metals.MetalsEnrichments._
-import scala.concurrent.Future
 import scala.meta.pc.CancelToken
 import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
 
 final class CodeActionProvider(
     compilers: Compilers
-) extends QuickFixes {
+) {
 
   def codeActions(
       params: l.CodeActionParams,
       token: CancelToken
   )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+
     for {
-      quickFixes <- QuickFix.fromDiagnostics(
-        params.getContext().getDiagnostics().asScala.toSeq,
-        params,
-        compilers,
-        token
-      )
+      quickFixes <- Future
+        .sequence(
+          List(QuickFix.ImportMissingSymbol)
+            .map(_.contribute(params, compilers, token))
+        )
+        .map(_.flatten)
     } yield quickFixes
 
   }
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
@@ -25,6 +25,8 @@ import scala.meta.pc.PresentationCompiler
 import scala.meta.pc.SymbolSearch
 import scala.concurrent.Future
 import java.{util => ju}
+import org.eclipse.lsp4j.CodeActionParams
+import scala.meta.pc.AutoImportsResult
 
 /**
  * Manages lifecycle for presentation compilers in all build targets.
@@ -150,6 +152,20 @@ class Compilers(
       pc.complete(CompilerOffsetParams.fromPos(pos, token)).asScala
     }.getOrElse(Future.successful(new CompletionList()))
 
+  def autoImports(
+      params: CodeActionParams,
+      name: String,
+      token: CancelToken
+  ): Future[ju.List[AutoImportsResult]] = {
+    val textDocumentPositionParams = new TextDocumentPositionParams(
+      params.getTextDocument(),
+      params.getRange().getEnd()
+    )
+    withPC(textDocumentPositionParams, None) { (pc, pos) =>
+      pc.autoImports(name, CompilerOffsetParams.fromPos(pos, token)).asScala
+    }.getOrElse(Future.successful(new ju.ArrayList))
+  }
+
   def hover(
       params: TextDocumentPositionParams,
       token: CancelToken,
@@ -162,6 +178,7 @@ class Compilers(
     }.getOrElse {
       Future.successful(Option.empty)
     }
+
   def definition(
       params: TextDocumentPositionParams,
       token: CancelToken
@@ -178,6 +195,7 @@ class Compilers(
           )
         }
     }.getOrElse(Future.successful(DefinitionResult.empty))
+
   def signatureHelp(
       params: TextDocumentPositionParams,
       token: CancelToken,
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
@@ -25,7 +25,6 @@ import scala.meta.pc.PresentationCompiler
 import scala.meta.pc.SymbolSearch
 import scala.concurrent.Future
 import java.{util => ju}
-import org.eclipse.lsp4j.CodeActionParams
 import scala.meta.pc.AutoImportsResult
 
 /**
@@ -153,15 +152,11 @@ class Compilers(
     }.getOrElse(Future.successful(new CompletionList()))
 
   def autoImports(
-      params: CodeActionParams,
+      params: TextDocumentPositionParams,
       name: String,
       token: CancelToken
   ): Future[ju.List[AutoImportsResult]] = {
-    val textDocumentPositionParams = new TextDocumentPositionParams(
-      params.getTextDocument(),
-      params.getRange().getEnd()
-    )
-    withPC(textDocumentPositionParams, None) { (pc, pos) =>
+    withPC(params, None) { (pc, pos) =>
       pc.autoImports(name, CompilerOffsetParams.fromPos(pos, token)).asScala
     }.getOrElse(Future.successful(new ju.ArrayList))
   }
--- a/metals/src/main/scala/scala/meta/internal/metals/QuickFix.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/QuickFix.scala
@@ -29,34 +29,25 @@ object QuickFix {
           diagnostic: l.Diagnostic,
           name: String
       ): Future[Seq[CodeAction]] = {
-
-        // TODO(gabro): this is hack. Instead of computing the auto-imports for a name at a range,
-        // we run completions starting from the end of the range, and filter the completions that
-        // match exactly the name we're looking for
-        val completionParams = new l.CompletionParams(
-          params.getTextDocument(),
-          params.getRange().getEnd()
-        )
-        compilers.completions(completionParams, token).map { completions =>
-          scribe.info(completions.getItems().toString())
-          scribe.info(name)
-          completions.getItems().asScala.collect {
-            case completionItem if completionItem.getFilterText() == name =>
-              val pkg = completionItem.getDetail().trim()
-              val edit = new l.WorkspaceEdit()
+        compilers
+          .autoImports(params, name, token)
+          .map { imports =>
+            imports.asScala.map { i =>
               val uri = params.getTextDocument().getUri()
-              val changes = Map(uri -> completionItem.getAdditionalTextEdits())
+              val edit = new l.WorkspaceEdit(Map(uri -> i.edits).asJava)
 
               val codeAction = new l.CodeAction()
-              codeAction.setTitle(s"Import '$name' from package '$pkg'")
+
+              codeAction.setTitle(
+                s"Import '$name' from package '${i.packageName}'"
+              )
               codeAction.setKind(l.CodeActionKind.QuickFix)
               codeAction.setDiagnostics(List(diagnostic).asJava)
-
-              edit.setChanges(changes.asJava)
               codeAction.setEdit(edit)
+
               codeAction
+            }
           }
-        }
       }
 
       Future
--- a/metals/src/main/scala/scala/meta/internal/metals/QuickFix.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/QuickFix.scala
@@ -5,7 +5,6 @@ import scala.meta.pc.CancelToken
 import org.eclipse.{lsp4j => l}
 import scala.concurrent.ExecutionContext
 import scala.meta.internal.metals.MetalsEnrichments._
-import org.eclipse.lsp4j.{CodeAction, CodeActionParams}
 
 trait QuickFix {
   def contribute(
@@ -20,17 +19,21 @@ object QuickFix {
   object ImportMissingSymbol extends QuickFix {
 
     override def contribute(
-        params: CodeActionParams,
+        params: l.CodeActionParams,
         compilers: Compilers,
         token: CancelToken
-    )(implicit ec: ExecutionContext): Future[Seq[CodeAction]] = {
+    )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
 
       def importMissingSymbol(
           diagnostic: l.Diagnostic,
           name: String
-      ): Future[Seq[CodeAction]] = {
+      ): Future[Seq[l.CodeAction]] = {
+        val textDocumentPositionParams = new l.TextDocumentPositionParams(
+          params.getTextDocument(),
+          diagnostic.getRange.getEnd()
+        )
         compilers
-          .autoImports(params, name, token)
+          .autoImports(textDocumentPositionParams, name, token)
           .map { imports =>
             imports.asScala.map { i =>
               val uri = params.getTextDocument().getUri()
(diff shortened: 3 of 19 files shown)


metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -0,0 +1,28 @@
...
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+    for {
+      quickFixes <- QuickFix.fromDiagnostics(
+        params.getContext().getDiagnostics().asScala.toSeq,
+        params,
+        compilers,
+        token
+      )
+    } yield quickFixes
```
gabro: the idea here is that we would have some other refactors joining in. For example:

```scala
for {
  quickFixes <- QuickFix.fromDiagnostics(...)
  classRefactors <- ClassRefactor.fromParams(...)
  ...
} yield quickFixes ++ classRefactors ++ ...
```

metals/src/main/scala/scala/meta/internal/metals/QuickFixes.scala
```diff
@@ -0,0 +1,82 @@
...
+
+import scala.concurrent.Future
+import scala.meta.pc.CancelToken
+import org.eclipse.{lsp4j => l}
+import scala.concurrent.ExecutionContext
+import scala.meta.internal.metals.MetalsEnrichments._
+
+trait QuickFixes {
+  trait QuickFix {
+    def contribute(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]]
```
tgodzik: I wonder if we should have contribute returning and `Option[CodeAction]`, we could have a lsit of quickfixes and just flatMap over them to get the code actions.

`contribute` would also take other parameters from below that we could need in any quickFix code action.

gabro: The issue is that some computations are asynchronous (hence `Future`) and there may be different imports available for an unknown symbol (hence `Seq`).

Or did I not understand your proposal?

tgodzik: Or just the Seq would actually also work:
```scala
trait QuickFix {
    def contribute( 
        diagnostics: Seq[l.Diagnostic],
        params: l.CodeActionParams,
        compilers: Compilers,
        token: CancelToken)
   (implicit ec: ExecutionContext): Future[Seq[l.CodeAction]]
}
```

and then we could do:

```scala
List(ImportMissingSymbol, Other).flatMap(_.contribute)
```
It would be easier to keep the actions separate in their own files.

metals/src/main/scala/scala/meta/internal/metals/QuickFixes.scala
```diff
@@ -0,0 +1,82 @@
...
+      params: l.CodeActionParams,
+      compilers: Compilers,
+      token: CancelToken
+  ) extends QuickFix {
+
+    override def contribute(
+        implicit ec: ExecutionContext
+    ): Future[Seq[l.CodeAction]] = {
+
+      // TODO(gabro): this is hack. Instead of computing the auto-imports for a name at a range,
```
tgodzik: To replace this hack we would need most likely to add an additional method to the java interfaces.

metals/src/main/scala/scala/meta/internal/metals/QuickFixes.scala
tgodzik: @gabro Sure it should be with futures with Seq. Let me fix the examples a bit.

tgodzik: I just meant it would be useful to have an interface describing what you can use in the quickfix and just add it to the list.

metals/src/main/scala/scala/meta/internal/metals/QuickFixes.scala
gabro: Yep, I’ve actually started on that, but then decide to get some early feedback first!

metals/src/main/scala/scala/meta/internal/metals/QuickFixes.scala
gabro: ahh, I see. That might work, I'll have a try! :)

gabro: Take a look now, I think it's close to what you mean @tgodzik

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  def codeActions(
+      params: l.CodeActionParams,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+
+    for {
+      quickFixes <- Future
+        .sequence(
+          List(QuickFix.ImportMissingSymbol)
```
gabro: this will grow into a list of all quickfixes, e.g.

```scala
List(QuickFix.ImportMissingSymbol, QuickFix.ImplementAbstractMembers, QuickFix.RemoveUnusedImport)
```

metals/src/main/scala/scala/meta/internal/metals/QuickFixes.scala
```diff
@@ -0,0 +1,72 @@
...
+}
+
+object QuickFix {
+
+  object ImportMissingSymbol extends QuickFix {
+
+    override def contribute(
+        params: CodeActionParams,
+        compilers: Compilers,
+        token: CancelToken
```
gabro: as suggested by @tgodzik, `contribute` will "document" which dependencies are needed by  any `QuickFix`

tgodzik:
I was thinking about how to make it work properly with the presentation compiler and I think a lot of the actions will need to get data from the presentation compiler, so it would be nice to do it via single interface and maybe create some kind of a query interface?

Would be maybe something like:
```java
...
java.util.List[TextEdit] query(kind: QueryType, filter: Predicate<String, Boolean>) 
...
public enum QueryType{
  FindSymbol, Params
}
```
Just a thought.

gabro:
@tgodzik / @olafurpg whenever you have time, I could use a review of the last commit, where I remove the "completions hack" and replace it with an `ImportProvider`, which loosely follows the implementation of `CompletionProvider`.

I'm not sure that's the way to go, so I could benefit from some early feedback before I proceed with polishing the implementation.

(PS, I'm surely doing something wrong since the imports include `_root_` and the ranges of the text edits are often wrong, but I haven't investigated yet)

mtags/src/main/scala/scala/meta/internal/pc/AutoImportsProvider.scala
```diff
@@ -0,0 +1,148 @@
...
+              context,
+              value
+            )
+            edits
+        }
+        AutoImportsResultImpl(pkg, edits.asJava)
+    }
+  }
+
+  private class CompilerSearchVisitor(
```
gabro: this is analogous to the homonymous class in `CompletionProvider`, except we visit symbols directly (instead of normalizing everything into a completion `Member`)

mtags/src/main/scala/scala/meta/internal/pc/AutoImportsProvider.scala
```diff
@@ -0,0 +1,148 @@
...
+    def shouldVisitPackage(pkg: String): Boolean =
+      packageSymbolFromString(pkg).isDefined
+
+    override def isCancelled: Boolean =
+      false
+
+  }
+
+  // FIXME(gabro): this is copy-pasted from CompletionProvider, we should de-dupe it
+  private def loadSymbolFromClassfile(
```
gabro: could it make sense to move this utility into `MetalsGlobal`?

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -150,6 +152,20 @@ class Compilers(
...
     }.getOrElse(Future.successful(new CompletionList()))
 
+  def autoImports(
+      params: CodeActionParams,
+      name: String,
+      token: CancelToken
+  ): Future[ju.List[AutoImportsResult]] = {
+    val textDocumentPositionParams = new TextDocumentPositionParams(
+      params.getTextDocument(),
+      params.getRange().getEnd()
```
gabro: code actions are triggered on a range, but the presentation compiler needs a position, so I'm arbitrarily picking the end of the range.

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
```diff
@@ -63,6 +63,11 @@
      */
     public abstract CompletableFuture<DefinitionResult> definition(OffsetParams params);
 
+    /**
+     * Return the necessary imports for a symbol at the given position.
+     */
+    public abstract CompletableFuture<List<AutoImportsResult>> autoImports(String name, OffsetParams params);
```
tgodzik: Maybe we could do something more general:
```
public abstract CompletableFuture<List<CodeActionResult>> codeAction(String input, ActionType actionType, OffsetParams params);

enum ActionType{
  Import, Implement, TypeAnnotation
}
```
otherwise we will be adding a lot of different methods for each action.

mtags/src/main/scala/scala/meta/internal/pc/AutoImportsProvider.scala
tgodzik: Might make sense to use the same logic with an additional transformer (TextEdit => T) function that might be identity by default.

mtags/src/main/scala/scala/meta/internal/pc/AutoImportsProvider.scala
tgodzik: Maybe via additional trait? I think it makes sense though.

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
gabro: Good point: I'm torn about this. From one end, I share your concern, from another I'm not sure of what code actions will be implemented so I'm wary of generalizing the API without knowing the use cases.

gabro:
I've added some basic tests, that are currently failing due to the issue with the import scope I've mentioned above.

mtags/src/main/scala/scala/meta/internal/pc/AutoImportsProvider.scala
gabro: Done, take a look!

mtags/src/main/scala/scala/meta/internal/pc/AutoImportsProvider.scala
gabro: Done 👍

mtags/src/main/scala/scala/meta/internal/pc/AutoImportsProvider.scala
```diff
@@ -0,0 +1,75 @@
...
+
+  def autoImports(): List[AutoImportsResult] = {
+    val unit = addCompilationUnit(
+      code = params.text(),
+      filename = params.filename(),
+      cursor = Some(params.offset())
+    )
+    val pos = unit.position(params.offset)
+    // make sure the compilation unit is loaded
+    typedTreeAt(pos)
```
gabro: @olafurpg is there a smarter way of doing this? If I omit it, `autoImportPosition` fails because `pos.source` isn't loaded yet.

olafurpg: There isn't. This is needed.

gabro: Fair enough! ¯\\\_(ツ)\_/¯

olafurpg:
This is very exciting! I believe this feature may be the most commonly requested feature judging by questions on gitter/discord.

I won't be able to review this PR in depth anytime soon . If you think the implementation is solid enough, then I think it's fine if we merge it nevertheless and we can leave improvements to the future. What do you think?

mtags/src/main/scala/scala/meta/internal/pc/AutoImportsProvider.scala
```diff
@@ -0,0 +1,75 @@
...
+      }
+      false
+    }
+
+    val visitor = new CompilerSearchVisitor(name, context, visit)
+
+    search.search(name, buildTargetIdentifier, visitor)
+
+    def isExactMatch(sym: Symbol, name: String): Boolean =
+      sym.name.dropLocal.decoded == name
```
gabro: this is needed since `search` above will return partially matching results, whereas we only want exact results for symbols.

@olafurpg is there a different way we could do this? I tried, but I couldn't find a way to instruct `search` to only return exact matches

olafurpg: It's possible to do exact searches and the index is optimized for this use-case, I can't remember the most relevant place to look into, maybe

https://github.com/scalameta/metals/blob/24cfabf16cae3d6187a094e0711d71c125e9e79f/mtags/src/main/scala/scala/meta/internal/metals/WorkspaceSymbolQuery.scala#L35

gabro: I saw that, but it seems `exact` still produces a list of alternatives. Anyway, it's fast enough not to be a concern (it's upper bound by completions, by definition)

gabro:
> I won't be able to review this PR in depth anytime soon . If you think the implementation is solid enough, then I think it's fine if we merge it nevertheless and we can leave improvements to the future. What do you think?

I think it's decent enough to unblock some more experiments on code actions.

I'll add an LSP test suite, then we can probably merge.

If you have time, can you just comment on the general approach of adding `autoImports` as a method on the Java interface `PresentationCompiler`, as opposed to have a more general method for querying the compiler, as proposed by @tgodzik?

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
tgodzik: Ok, let's leave it then and see if generalizing it will be useful in the future.

tests/cross/src/test/scala/tests/pc/AutoImportsSuite.scala
```diff
@@ -0,0 +1,119 @@
...
+           |""".stripMargin
+    )
+  )
+
+  checkEdit(
+    "basic-edit",
+    """|package a
+       |
+       |object A {
+       |  <<Future>>.successful(2)
```
tgodzik: Could you add a case with a partial import?
For example:
`<<concurrent.Future>>`

tests/cross/src/test/scala/tests/pc/AutoImportsSuite.scala
```diff
@@ -0,0 +1,119 @@
+package tests.pc
+
+import tests.BaseCodeActionSuite
+import scala.meta.pc.AutoImportsResult
+import scala.meta.internal.metals.CompilerOffsetParams
+import scala.meta.internal.jdk.CollectionConverters._
+import scala.meta.internal.metals.TextEdits
+
+object AutoImportsSuite extends BaseCodeActionSuite {
```
tgodzik: I would also create unit tests:
- CodeActionsDetectionSuite - it would detect what code actions are available for given offsets and if they should be
- RunCodeActionSuite - it would run given code action at a given position

I am ok with combining those suites, but for sure I would add some unit tests, which might be useful for people writing simpler code actions.

metals/src/main/scala/scala/meta/internal/metals/QuickFix.scala
```diff
@@ -0,0 +1,66 @@
...
+              codeAction.setEdit(edit)
+
+              codeAction
+            }
+          }
+      }
+
+      Future
+        .sequence(params.getContext().getDiagnostics().asScala.collect {
+          case d @ ScalacDiagnostic.SymbolNotFound(name) =>
```
tgodzik: Can we quickly return if the diagnostic range does not contain the current position?

tests/cross/src/test/scala/tests/pc/AutoImportsSuite.scala
gabro: Doing it as we speak ;)

metals/src/main/scala/scala/meta/internal/metals/QuickFix.scala
tgodzik: `encloses` should be easy to use here.

tests/cross/src/test/scala/tests/pc/AutoImportsSuite.scala
gabro: I don't think that works, since we get a diagnostic on the package name which we can't resolve, it seems. For example:

image

hovering on `time` won't yield any quick action (no completions are returned either)

tests/unit/src/test/scala/tests/CodeActionLspSuite.scala
```diff
@@ -0,0 +1,57 @@
...
+        |""".stripMargin,
+    """|package a
+       |
+       |import scala.concurrent.Future
+       |
+       |object A {
+       |  val f = Future.successful(2)
+       |}
+       |""".stripMargin
+  )
```
gabro: @tgodzik this is my proposal. You have three main elements:

- the input text (using `@@` as the cursor position)
- the expected list of code actions (I'm checking the `title` property of code actions)
- the expected text after applying one of the code actions (I'm selecting the first one by default, but it's configurable)

This can be expanded to support other common test cases, but I think it's a decent start

tests/cross/src/test/scala/tests/pc/AutoImportsSuite.scala
olafurpg: this is because the classpath index doesn't support querying packages. It should be an easy fix to implement

gabro: I imagined. Should we consider that out of scope for this PR?

tests/unit/src/test/scala/tests/CodeActionLspSuite.scala
tgodzik: That's exactly what I was thinking about. Looks awesome!

metals/src/main/scala/scala/meta/internal/metals/QuickFix.scala
gabro: done, take a look!

tests/cross/src/test/scala/tests/pc/AutoImportsSuite.scala
olafurpg: It’s best to implement that as a separate pr. I estimate it’s a simple fix

fmv1992:
Hey, so how can we execute auto import on vim?

Thanks!

gabro:
@fmv1992 you must invoke the coc-codeaction command. See here for the recommended keybindings https://scalameta.org/metals/docs/editors/vim.html

fmv1992:
Thanks! This plugin is great. I was able to do this by disabling other code completion customization and using the one from that link!

PR #1075, opened 2019-11-18
Use exact match when computing file renames

### LINKED ISSUE #1074: File is renamed even when the renamed toplevel symbol partially matches the filename
**Describe the bug**
Reported on Gitter by @kubukoz 

> renaming a top-level symbol can rename the file if the file ends with (symbol).scala.
> Because of that, if we're in a file like "FooBar.scala" and have this code:
> 
> ```scala
> object Bar
> 
> object FooBar
> ```
> 
> if we rename `Bar`, the file will also be renamed. 
> to be honest I'd look for exact matches of filename.scala and object filename / class filename etc.

**Installation:**
- Metals version: v0.7.6-master

**Additional context**

**Search terms**
rename, filename, toplevel

gabro:
Fixes #1074 

Previously we would rename files whose name partially matched the symbol being renamed. Now we compute those renames using an exact match, to avoid e.g. renaming a file named `FooBar.scala` when renaming the symbol `Bar` in that file.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -32,7 +32,6 @@ import org.eclipse.lsp4j.TextDocumentEdit
 import org.eclipse.lsp4j.VersionedTextDocumentIdentifier
 import org.eclipse.lsp4j.ResourceOperation
 import org.eclipse.lsp4j.RenameFile
-import java.io.File
 
 final class RenameProvider(
     referenceProvider: ReferenceProvider,
@@ -149,7 +148,7 @@ final class RenameProvider(
       .find { file =>
         isOccurence(str => {
           (str.desc.isType || str.desc.isTerm) &&
-            new File(file).getName() == str.desc.name.value + ".scala"
+            Paths.get(file).filename == str.desc.name.value + ".scala"
         })
       }
       .map { file =>
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -323,7 +323,7 @@ final class RenameProvider(
   ): MessageParams = {
     val renamed = name.map(n => s"to $n").getOrElse("")
     val message =
-      s"""|Cannot rename $old $renamed since it will change the semantics and
+      s"""|Cannot rename from $old to $renamed since it will change the semantics
           |and might break your code""".stripMargin
     new MessageParams(MessageType.Error, message)
   }
@@ -341,7 +341,7 @@ final class RenameProvider(
   ): MessageParams = {
     val renamed = name.map(n => s"to $n").getOrElse("")
     val message =
-      s"""|Cannot rename from $old $renamed since it will change the semantics and
+      s"""|Cannot rename from $old to $renamed since it will change the semantics and
           |and might break your code.
           |Only rename to names ending with `:` is allowed.""".stripMargin
     new MessageParams(MessageType.Error, message)
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -148,7 +148,7 @@ final class RenameProvider(
       .find { file =>
         isOccurence(str => {
           (str.desc.isType || str.desc.isTerm) &&
-            Paths.get(file).filename == str.desc.name.value + ".scala"
+            Paths.get(new URI(file)).filename == str.desc.name.value + ".scala"
         })
       }
       .map { file =>
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -148,7 +148,7 @@ final class RenameProvider(
       .find { file =>
         isOccurence(str => {
           (str.desc.isType || str.desc.isTerm) &&
-            Paths.get(new URI(file)).filename == str.desc.name.value + ".scala"
+            file.endsWith(s"/${str.desc.name.value}.scala")
         })
       }
       .map { file =>
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -32,6 +32,7 @@ import org.eclipse.lsp4j.TextDocumentEdit
 import org.eclipse.lsp4j.VersionedTextDocumentIdentifier
 import org.eclipse.lsp4j.ResourceOperation
 import org.eclipse.lsp4j.RenameFile
+import java.io.File
 
 final class RenameProvider(
     referenceProvider: ReferenceProvider,
@@ -147,9 +148,8 @@ final class RenameProvider(
     fileChanges
       .find { file =>
         isOccurence(str => {
-          (str.desc.isType || str.desc.isTerm) && file.endsWith(
-            str.desc.name.value + ".scala"
-          )
+          (str.desc.isType || str.desc.isTerm) &&
+            new File(file).getName() == str.desc.name.value + ".scala"
         })
       }
       .map { file =>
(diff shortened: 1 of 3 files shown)


gabro:
@tgodzik I'm currently unable to run tests locally (they all fail, I haven't investigated, but as you know my Mac is cursed), so I'm using the CI to get some feedback

tests/unit/src/test/scala/tests/RenameSuite.scala
```diff
@@ -279,6 +279,29 @@ object RenameSuite extends BaseLspSuite("rename") {
...
+
+  renamed(
+    "filename-exact-match-2",
+    """|/a/src/main/scala/a/Main.scala
+       |package a
+       |object Main
+       |object <<The@@Main>>
+       |""".stripMargin,
+    newName = "Tree",
+    fileRenames = Map.empty
```
gabro: I know this is the default, but I wanted to be explicit here

tests/unit/src/test/scala/tests/RenameSuite.scala
```diff
@@ -279,6 +279,29 @@ object RenameSuite extends BaseLspSuite("rename") {
...
+       |object <<Ma@@in>>
+       |object TheMain
+       |""".stripMargin,
+    newName = "Tree",
+    fileRenames =
+      Map("a/src/main/scala/a/Main.scala" -> "a/src/main/scala/a/Tree.scala")
+  )
+
+  renamed(
+    "filename-exact-match-2",
```
tgodzik: :heart_eyes: Thanks @gabro!

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -147,9 +147,8 @@ final class RenameProvider(
     fileChanges
       .find { file =>
         isOccurence(str => {
-          (str.desc.isType || str.desc.isTerm) && file.endsWith(
-            str.desc.name.value + ".scala"
-          )
+          (str.desc.isType || str.desc.isTerm) &&
+            Paths.get(file).filename == str.desc.name.value + ".scala"
```
tgodzik: Using just `s"/${str.desc.name.value}.scala"` will also work since it's a uri, but not sure which is the best way.

gabro: Ah, I wasn't sure. I assumed we would get the full path here, e.g. `/src/main/scala/foo/FooBar.scala`, is that not the case?

tgodzik:
> @tgodzik I'm currently unable to run tests locally (they all fail, I haven't investigated, but as you know my Mac is cursed), so I'm using the CI to get some feedback

Let me know if you need help with that - I might be able to help? I had some issue that broke all tests previously.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: It should be a full uri: `file://<path-to-workspace>/a/src/main/scala/foo/FooBar.scala`

You might need to use `Paths.get(new URI(file)))`, though not 100% sure

gabro: `Paths.get` has an overload that accepts strings. Here's my quick test in the REPL:

```scala
scala> import java.nio.file.Paths
import java.nio.file.Paths

scala> Paths.get("/foo/bar/baz/Foo.scala").getFileName().toString
res1: String = Foo.scala
```

(our `.filename` is short for `.getFileName().toString`)

gabro:
The failure on Windows is relevant:

```
X tests.RenameSuite.across-targets 8002ms 
  java.nio.file.InvalidPathException: Illegal char <:> at index 4: file:///D:/a/metals/metals/tests/unit/target/e2e/rename/across-targets/a/src/main/scala/a/Main.scala
```

gabro:
Although I can't reproduce in a REPL on macOS 🤔 

```scala
scala> Paths.get("file:///D:/a/metals/metals/tests/unit/target/e2e/rename/across-targets/a/src/main/scala/a/Main.scala")
res4: java.nio.file.Path = file:/D:/a/metals/metals/tests/unit/target/e2e/rename/across-targets/a/src/main/scala/a/Main.scala
```

tgodzik:
@gabro We do `AbsolutePath(Paths.get(new URI(file)))` and it seems to work with the same input. Might be worth fixing this way.

gabro:
Thanks for looking into this, I think we're hitting on this JDK bug https://bugs.openjdk.java.net/browse/JDK-8232925 (linked from https://github.com/quarkusio/quarkus/issues/3592, which describes a similar problem). Adding the `new URI` step probably introduces a step or URL encoding which fixes the issue. I'll give it a try and see if it fixes tests on Windows

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -147,9 +147,8 @@ final class RenameProvider(
     fileChanges
       .find { file =>
         isOccurence(str => {
-          (str.desc.isType || str.desc.isTerm) && file.endsWith(
-            str.desc.name.value + ".scala"
-          )
+          (str.desc.isType || str.desc.isTerm) &&
+            Paths.get(new URI(file)).filename == str.desc.name.value + ".scala"
```
olafurpg: Since we know `file` is a URI, can we do this instead?
```suggestion
           file.endsWith(s"/${str.desc.name.value}.scala")
```

gabro: That's what @tgodzik suggested too. If you both think it's a good idea, I'll fix it this way!

tests/unit/src/test/scala/tests/RenameSuite.scala
```diff
@@ -279,6 +279,29 @@ object RenameSuite extends BaseLspSuite("rename") {
       Map("a/src/main/scala/a/Main.scala" -> "a/src/main/scala/a/Tree.scala")
```
olafurpg: Unrelated to this PR, but I just noticed that this test suite should be called `RenameLspSuite`! cc/ @tgodzik

gabro: I can squeeze a fix in ;-)

tgodzik: You can use rename for the renaming of RenameLspSuite :O

olafurpg: 🤯

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -324,14 +323,14 @@ final class RenameProvider(
   ): MessageParams = {
     val renamed = name.map(n => s"to $n").getOrElse("")
     val message =
-      s"""|Cannot rename $old $renamed since it will change the semantics and 
+      s"""|Cannot rename $old $renamed since it will change the semantics and
```
kubukoz: Double `and`

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -342,8 +341,8 @@ final class RenameProvider(
   ): MessageParams = {
     val renamed = name.map(n => s"to $n").getOrElse("")
     val message =
-      s"""|Cannot rename from $old $renamed since it will change the semantics and 
-          |and might break your code. 
+      s"""|Cannot rename from $old $renamed since it will change the semantics and
```
kubukoz: `from $old to $renamed`?

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
gabro: thanks :) I hadn't actually changed those lines in this PR, but I'll take the chance to fix it

tests/unit/src/test/scala/tests/RenameSuite.scala
gabro: Done!

I've renamed the object using the rename command, forgot to rename the file... but "rename" had my back! 😍

PR #1077, opened 2019-11-19
Avoid renaming synthetic occurrences

### LINKED ISSUE #1072: "Rename symbol" inlines methods to their implicit parameters
**Describe the bug**

When renaming a value that's used as an implicit parameter, the value is inlined at the position of the call to the method taking it implicitly.

**To Reproduce**
1. Compile this:

```scala
object foo {
  implicit val x: Int = ???

  def foo[A](implicit a: A): (A, A) = (a, a)

  foo[Int]
}
```

2. Rename `x` to anything:

image

3. Watch the result:

```scala
object foo {
  implicit val y: Int = ???

  def foo[A](implicit a: A): (A, A) = (a, a)

  y
}
```

**Expected behavior**

The call site of the method taking implicits isn't changed (unless the parameter is passed explicitly).

**Installation:**
- Metals version: current master (eecfc753249b34ecf64e54d57f510eab1dd617e1)

**Additional context**

If the argument is passed explicitly, everything is fine. If it's passed both implicitly and explicitly, like in:

```scala
object foo {
  implicit val x: Int = ???

  def foo[A](a: A)(implicit a2: A): (A, A) = (a, a2)

  foo(x)
}
```

Renaming results in:

image

**Search terms**
Rename symbol, implicit, inline, parameter

gabro:
Fixes #1072 

Previously we would rename all references of a symbol, including synthetic ones. This resulted in bugs such as #1072. Now we ignore synthetics references when performing a rename.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
@@ -67,7 +67,7 @@ final class ReferenceProvider(
   def references(
       params: ReferenceParams,
       checkMatchesText: Boolean = false,
-      includeSynthetics: Boolean = true
+      includeSynthetics: SymbolOccurrence => Boolean = _ => true
   ): ReferencesResult = {
     val source = params.getTextDocument.getUri.toAbsolutePath
     semanticdbs.textDocument(source).documentIncludingStale match {
@@ -86,7 +86,7 @@ final class ReferenceProvider(
               alternatives,
               params.getContext.isIncludeDeclaration,
               checkMatchesText,
-              includeSynthetics
+              includeSynthetics(occurrence)
             )
             ReferencesResult(occurrence.symbol, locations)
           case None =>
--- a/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
@@ -66,7 +66,8 @@ final class ReferenceProvider(
 
   def references(
       params: ReferenceParams,
-      checkMatchesText: Boolean = false
+      checkMatchesText: Boolean = false,
+      includeSynthetics: Boolean = true
   ): ReferencesResult = {
     val source = params.getTextDocument.getUri.toAbsolutePath
     semanticdbs.textDocument(source).documentIncludingStale match {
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -86,7 +86,8 @@ final class RenameProvider(
         txtParams <- if (parentSymbols.isEmpty) List(textParams)
         else parentSymbols.map(toTextParams)
         currentReferences = referenceProvider.references(
-          toReferenceParams(txtParams)
+          toReferenceParams(txtParams),
+          includeSynthetics = false
```
olafurpg: This will miss references to synthetic `.apply`, which we do want to rename

gabro: right, I forgot to "de-focus" the tests locally. I'll think of a solution

gabro: I've pushed a fix, which is the simplest thing I could think of: decide whether to include or not the synthetics based on the occurrence.

This way we can decide for which symbols we want to include the synthetics, e.g. we do want them for `apply`

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -86,7 +86,8 @@ final class RenameProvider(
         txtParams <- if (parentSymbols.isEmpty) List(textParams)
         else parentSymbols.map(toTextParams)
         currentReferences = referenceProvider.references(
-          toReferenceParams(txtParams)
+          toReferenceParams(txtParams),
+          includeSynthetics = _.symbol.desc.name.value == "apply"
```
gabro: Include the synthetics occurrences of `apply` because we want to rename them, but exclude anything else

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -64,7 +64,10 @@ final class ReferenceProvider(
     resizeReferencedPackages()
   }
 
-  def references(params: ReferenceParams): ReferencesResult = {
+  def references(
+      params: ReferenceParams,
+      includeSynthetics: SymbolOccurrence => Boolean = _ => true
```
tgodzik: Let's do the function as suggested by @olafurpg :

`Synthetic => Boolean `

tgodzik: @olafurpg I took the liberty to change it myself - is this what you meant?

tgodzik:
@olafurpg I rebased the change and this should be ok to merge, what do you think?

PR #1079, opened 2019-11-21
Add a 'use named arguments' refactoring code action

cb372:
This PR adds a refactoring code action, inspired by @gabro's #1065.

I chose "use named arguments" because it's a reasonably simple refactoring, I would use it myself, and IntelliJ's similar feature was mentioned in https://github.com/scalameta/metals-feature-requests/issues/18.

Nov-21-2019 10-52-50

## Disclaimer

I Have No Idea What I'm Doing.

I mashed this together using whatever tools I could find in the `scala.meta.internal.metals` package.

## How it works

1. Check if the tree under the cursor is a method application. If so, note the position of the name of the method being applied.
2. Using SemanticDB, find the symbol being referenced at that position.
3. Use symbol search to find the documentation for that symbol. That gives us the argument names.
4. Using the tree from step 1, go through each argument. For each one that is not already a named argument, generate an edit to prefix it with `$argumentName = `.

This could easily be tweaked to precisely match the behaviour of IntelliJ's "Use named argument for current and subsequent arguments" action if desired.

## Notes

Use of symbol search to get symbol documentation seems to be a bit temperamental. Sometimes the `documentation` method returns `None` for symbols defined in the workspace, while it consistently returns documentation for symbols in `scala.Predef`. Restarting VSCode fixes it. I haven't dug into that.


metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -0,0 +1,44 @@
...
+    symbolSearch: MetalsSymbolSearch
+) {
+
+  def codeActions(
+      params: l.CodeActionParams,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+
+    for {
+      quickFixes <- Future
```
PeterPerhac: you don't need the for-comprehension here

metals/src/main/scala/scala/meta/internal/metals/Refactorings.scala
```diff
@@ -0,0 +1,128 @@
...
+
+      def findMethodApplyTreeUnderCursor(
+          root: Tree,
+          range: Position
+      ): Option[Term.Apply] = {
+        root.collect {
+          case t @ Term.Apply(_, _)
+              if t.pos.start <= range.start && t.pos.end >= range.end =>
+            t
+        }.headOption
```
PeterPerhac: can use `collectFirst` here

tgodzik:
Looks cool! It would be awesome if people were able to easily contribute new actions. I think we can get back to it once @gabro 's PR is merged.

tgodzik:
@cb372 Hey Chris, https://github.com/scalameta/metals/pull/1065 has been merged, so this PR is unblocked :tada: You can rebase and let us know when you'll need a review :smiley:

cb372:
Thanks @tgodzik. This is now ready for review.

I'd like to write some tests but I don't really know how. I need to take a closer look at the tests @gabro wrote in #1065.

metals/src/main/scala/scala/meta/internal/metals/Refactorings.scala
```diff
@@ -0,0 +1,128 @@
...
+      def buildEdits(
+          methodApplyTree: Term.Apply,
+          paramNames: List[String]
+      ): List[l.TextEdit] = {
+        methodApplyTree.args.zip(paramNames).flatMap {
+          case (Term.Assign(_, _), _) =>
+            // already a named argument, no edit needed
+            None
+          case (term, paramName) =>
+            val position =
```
tgodzik: The position here might need to be adjusted using EditDistance - you can get it from the `DefinitionProvider.positionOccurence`

tgodzik:
You can add tests inside `CodeActionLspSuite` - this has 3 parts:
1. Code before code action with cursor indicated as `@@`.
2. Code actions detected at the cursor postion.
3. Code after applying the code action.

metals/src/main/scala/scala/meta/internal/metals/Refactorings.scala
cb372: I don't think we need to do any adjustment here? Because we are getting the offsets from a tree given to us by `Trees`, which uses `Buffers` to take account of unsaved edits.

I'll try to write a test for files with unsaved edits to confirm this. That's the only test I haven't been able to write yet.

tgodzik: Right! I think it should be fine then. This is being tested for example in https://github.com/scalameta/metals/blob/2475a952a4eaf7f3b5e41246e091c3757f244908/tests/unit/src/test/scala/tests/RenameLspSuite.scala#L604-L609
but this test doesn't need to be added inside of this PR I think.

cb372:
@tgodzik I'm pretty happy with this now. I've been testing manually in VSCode and it seems solid.

I had to add one nasty little symbol-rewriting hack, but if you're happy with that, I think this is ready to go.

cb372:
Forgot to say: thank you for your helpful comments!

tests/unit/src/test/scala/tests/CodeActionLspSuite.scala
```diff
@@ -26,6 +27,279 @@ object CodeActionLspSuite extends BaseLspSuite("codeAction") {
        |""".stripMargin
   )
 
+  check(
+    "use-named-arguments-basic",
```
gabro: since these tests are going to grow for each code action, maybe we can split these into:

- `BaseCodeActionLspSuite` (which simply defines the `check` method)
- `ImportMissingSymbolLspSuite`
- `UseNamedArgumentsLspSuite`

This should have the benefits of:
- keeping the file length manageable
- keeping the suite more focused (you can run the entire suite when iterating on the specific action)
- removing the prefix from the test name (`basic`, instead of `use-named-arguments-basic`)

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -2,27 +2,60 @@ package scala.meta.internal.metals
...
-    for {
-      quickFixes <- Future
-        .sequence(
-          List(QuickFix.ImportMissingSymbol)
-            .map(_.contribute(params, compilers, token))
+    def quickfixes =
+      List(QuickFix.ImportMissingSymbol)
+        .map(_.contribute(params, compilers, token))
+
+    def refactorings =
```
gabro: if we go with my proposal below, we can unify these `contribute` calls can go together

metals/src/main/scala/scala/meta/internal/metals/Refactorings.scala
```diff
@@ -0,0 +1,182 @@
...
+
+import scala.concurrent.{Future, ExecutionContext}
+import scala.meta._
+import scala.meta.pc.CancelToken
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.mtags._
+import scala.meta.internal.semanticdb.TextDocument
+import org.eclipse.{lsp4j => l}
+
+trait Refactoring {
```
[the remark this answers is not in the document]
tgodzik: I think that the distinction only makes sense on the frontend really.  There is nothing really that distinguishes quick fixes from refactoring inside Metals.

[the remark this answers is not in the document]
tgodzik: Although, might be good to actually have it separate and force the correct type for `refactorings`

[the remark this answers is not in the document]
gabro: > force the correct type for refactorings value

what do you mean?



> I think that the distinction only makes sense on the frontend really.

I agree with this, hence my proposal to unify the interface into `CodeActionKind`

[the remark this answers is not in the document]
tgodzik: > > force the correct type for refactorings value
> 
> what do you mean?
> 
> > I think that the distinction only makes sense on the frontend really.
> 
> I agree with this, hence my proposal to unify the interface into `CodeActionKind`

I just meant we could do:
```
def refactorings: Seq[Refactoring] =
```
to make sure everything is correctly returned and we could set the action kind for each of the refactorings. It would just kind of be a logical divide there.

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
cb372: I agree it's probably a good idea to get rid of the `QuickFix` and `Refactoring` traits and just have a single `CodeAction` trait to cover both.

But we should still avoid doing unnecessary work when the client sets the `only` param. e.g. if the client requests only quick fixes, we shouldn't waste cycles on calculating all possible refactorings. So we could either keep separate manually-curated lists, like I did here, or the `CodeAction` trait could have a `def kind: CodeActionKind` and we could filter based on that before we call `contribute`. I think the latter is probably cleaner.

/cc @tgodzik

metals/src/main/scala/scala/meta/internal/metals/Refactorings.scala
```diff
@@ -0,0 +1,128 @@
...
+      scribe.debug("Running contribute for UseNamedArguments")
+
+      def findMethodApplyTreeUnderCursor(
+          root: Tree,
+          range: Position
+      ): Option[Term.Apply] =
+        root
+          .collect {
+            case t @ Term.Apply(_, _)
+                if t.pos.start <= range.start && t.pos.end >= range.end =>
```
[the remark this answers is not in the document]
cb372: Thanks, I'll have a look

metals/src/main/scala/scala/meta/internal/metals/Refactorings.scala
```diff
@@ -0,0 +1,182 @@
...
+          .lastOption
+
+      def findSymbolTree(tree: Tree): Name = tree match {
+        case x @ Term.Name(_) => x
+        case x @ Type.Name(_) => x
+        case Term.Select(_, x) => x
+        case Term.Apply(x, _) => findSymbolTree(x)
+        case Term.ApplyType(x, _) => findSymbolTree(x)
+        case Type.Apply(x, _) => findSymbolTree(x)
+        case Init(x, _, _) => findSymbolTree(x)
```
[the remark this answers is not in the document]
cb372: errr, probably...

I wrote this by manually trying a bunch of different kinds of method/constructor calls and looking at the scalameta tree, so there's no guarantee I covered every possible case. Scala always has the potential to surprise.

I'll change this method to return an Option and add a catch-all `case _ => None`.

metals/src/main/scala/scala/meta/internal/metals/Refactorings.scala
```diff
@@ -0,0 +1,182 @@
...
+          symbol
+        }
+      }
+
+      def buildEdits(
+          tree: Tree,
+          paramNames: List[String],
+          editDistance: TokenEditDistance
+      ): List[l.TextEdit] = {
+        // TODO write tests to confirm whether we need to offset by edit distance or not
```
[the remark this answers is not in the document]
cb372: Based on discussion with @tgodzik I was leaving this as a "TODO one day when we get around to it". But I'll have a go at writing a test as part of this PR. If it's too hard, I'll remove the TODO and open an issue.

tests/unit/src/test/scala/tests/CodeActionLspSuite.scala
cb372: Yes, I was thinking that as well.

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
gabro: Ah, absolutely. I didn't think of `only` so I guess you can ignore my comment and we should invoke `contribute` only after we check that parameter.

metals/src/main/scala/scala/meta/internal/metals/Refactorings.scala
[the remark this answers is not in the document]
gabro: Regardless of the type, we can still "force" the code action `kind` when grouping them:

```scala
def quickfixes = List(...)
  .flatMap(_.contribute(...))
  .map(_.setKind(CodeActionKind.QuickFix))
```

I don't think we need to encode it in the type since it's a fairly innocuous mistake and the risk of getting this wrong is low.

metals/src/main/scala/scala/meta/internal/metals/CodeAction.scala
```diff
@@ -0,0 +1,27 @@
...
+
+trait CodeAction {
+
+  /**
+   * This should be one of the String constants
+   * listed in [[org.eclipse.lsp4j.CodeActionKind]]
+   */
+  def kind: String
+
+  def contribute(
```
olafurpg: Can we reduce the number of parameters in this method? It would be nice to bundle these below into a single class so that we can add more components in the future without updating a ton of signatures

* compilers
* trees
* buffers
* semanticdbs
* symbol search
* definition provider

We should try to design these APIs as if they're public and try to keep them as stable as possible.

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -2,27 +2,56 @@ package scala.meta.internal.metals
...
-            .map(_.contribute(params, compilers, token))
-        )
-        .map(_.flatten)
-    } yield quickFixes
+    val allActions = List(
+      QuickFixes.ImportMissingSymbol,
+      Refactorings.UseNamedArguments
+    )
+
+    val isRequestedKind: CodeAction => Boolean =
```
olafurpg: nit: can we refactor this into `def isRequestedKind(..): Boolean` method?

olafurpg: alternatively, inline this into `filter(..)`

metals/src/main/scala/scala/meta/internal/metals/Refactorings.scala
```diff
@@ -0,0 +1,175 @@
...
+import scala.meta._
+import scala.meta.pc.CancelToken
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.mtags._
+import scala.meta.internal.semanticdb.TextDocument
+import org.eclipse.{lsp4j => l}
+
+object Refactorings {
+
+  object UseNamedArguments extends CodeAction {
```
olafurpg: Can we move each refactoring into a separate source file and into the package `scala.meta.internal.codeactions`? We may end up with hundreds of code actions, which would be nice to maintain in separate source files.

gabro: 👍 I was thinking the same, makes sense! 
If only we had a “move package” code action 😅

metals/src/main/scala/scala/meta/internal/metals/CodeAction.scala
gabro: I thought about this as well, and one alternative way of doing this is to have each CodeAction as a class, which take the necessary “global” parameters upon initialization and the “per request” parameters in the contribute method.

Something like

```
class ImportMissingSymbol(
  compilers: Compilers,
  // ...
) extends CodeAction {
  override def contribute(params: CodeActionParams): Future[List[l.CodeAction]] = // ...
```

gabro: This would allow each code action to manage their required dependencies independently. I think I was too eager in trying to generalize a common interface.

metals/src/main/scala/scala/meta/internal/metals/Refactorings.scala
cb372: Done 👍

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
cb372: Done

metals/src/main/scala/scala/meta/internal/metals/CodeAction.scala
cb372: I've done as @gabro suggested. The `CodeAction` trait's `contribute` method now takes only two arguments, the `params: CodeActionParams` and the cancel token. Each code action is now a class that takes its dependencies as constructor parameters.

I agree we could maybe try to group the dependencies into a small number of approachable APIs that abstract over them, but I feel it's too soon to do that. So far we only have 2 code actions, and they each depend on completely different functionality, so it's not yet clear how to define sensible abstractions. Maybe revisit this once we've built a couple more code actions and commonalities start to emerge? @olafurpg

mtags/src/main/scala/scala/meta/internal/mtags/MtagsEnrichments.scala
```diff
@@ -311,11 +311,18 @@ trait MtagsEnrichments {
         range.getEnd.getCharacter
       )
     def encloses(position: l.Position): Boolean = {
```
cb372: Unless I'm misunderstanding how ranges work, I think this logic was incorrect for multi-line ranges (e.g. a range from char 10 of line 10 to char 20 of line 12 should enclose the position at char 5 of line 11) so I rewrote it. 

I noticed the same mistake in a few other `encloses` methods in this file and in `MetalsEnrichment.scala`. I can fix those too, but probably better to do that in a separate PR.

metals/src/main/scala/scala/meta/internal/metals/Refactorings.scala
cb372: I've confirmed via manual testing in VSCode that the code action does the right thing when the file has unsaved edits. I tried to write a unit test for it, but after 2 hours of battling I only succeeded in writing a test that proves the opposite 😢 

I'll open an issue for writing the test.

metals/src/main/scala/scala/meta/internal/metals/codeactions/UseNamedArguments.scala
```diff
@@ -0,0 +1,164 @@
...
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+
+    def findMethodApplyOrCtorTreeUnderCursor(
+        root: Tree,
+        cursorPos: l.Range
+    ): Option[Tree] =
+      root
+        .collect {
+          case t @ Term.Apply(_, _) if t.pos.toLSP.encloses(cursorPos) => t
```
gabro: nit:

```suggestion
          case t: Term.Apply if t.pos.toLSP.encloses(cursorPos) => t
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/UseNamedArguments.scala
```diff
@@ -0,0 +1,164 @@
...
+          case t @ Init(_, _, _) if t.pos.toLSP.encloses(cursorPos) => t
```
gabro: nit:

```suggestion
          case t: Init if t.pos.toLSP.encloses(cursorPos) => t
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/UseNamedArguments.scala
```diff
@@ -0,0 +1,164 @@
...
+        symbolOccurrence <- resolvedSymbol.occurrence
+        symbol = tweakSymbol(symbolOccurrence.symbol)
+        symbolDocumentation <- symbolSearch
+          .documentation(symbol)
+          .asScala
+        parameterNames = symbolDocumentation
+          .parameters()
+          .asScala
+          .map(_.displayName())
+          .toList
```
gabro: This part here is using `symbolSearch` and `symbolDocumentation` to lookup the signature information of the method under the cursor.

This is very similar to the job `SignatureHelpProvider` does, but the main difference is that `SignatureHelpProvider` uses the presentation compiler, so it should work even with code you just typed or when the buffer doesn't parse or compile.

I also think `SignatureHelpProvider` does a very similar job at finding the enclosing `apply` (see `EnclosingApply`).

Bottom line, I see a chance of building the core of this action on top of the presentation compiler, possibly re-using some of the infrastructure provided by `SignatureHelpProvider`, or even just using that provider directly.

As a quick POC, you can maybe try invoking `compilers.signatureHelp` and see whether that's enough information to build this action.

metals/src/main/scala/scala/meta/internal/metals/codeactions/UseNamedArguments.scala
```diff
@@ -0,0 +1,164 @@
...
+      }).flatten.toSeq
+    }
+
+  }
+
+}
+
+object UseNamedArguments {
+
+  val title = "Use named arguments"
```
gabro:
```suggestion
  val label = "Use named arguments"
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/UseNamedArguments.scala
gabro: Paging @olafurpg, as soon as you have time

metals/src/main/scala/scala/meta/internal/metals/codeactions/UseNamedArguments.scala
cb372: I don't really mind either way, but FYI lsp4j (and the LSP spec) calls it a title.

metals/src/main/scala/scala/meta/internal/metals/codeactions/UseNamedArguments.scala
cb372: Thanks, I'll give it a try

cb372: @gabro Wow, that certainly simplified things!

metals/src/main/scala/scala/meta/internal/metals/codeactions/UseNamedArguments.scala
gabro: ah, nevermind then, can we rename the `ImpostMissingSymbol` one for consistency?

metals/src/main/scala/scala/meta/internal/metals/codeactions/UseNamedArguments.scala
```diff
@@ -0,0 +1,127 @@
...
+
+    val path = params.getTextDocument().getUri().toAbsolutePath
+
+    val textDocPosParams = new l.TextDocumentPositionParams(
+      params.getTextDocument,
+      params.getRange.getStart
+    )
+
+    val parameterNames: Future[List[String]] = compilers
+      .signatureHelp(textDocPosParams, token, interactiveSemanticdbs)
```
olafurpg: Is there a way to distinguish between code actions that are explicitly requested by the users and code action requests that are automatically triggered by the editor?

I am concerned about triggering this method on every `codeAction` request since VS Code requests for code actions on every cursor move. The signature help method triggers typechecking on the same compiler thread as other completion/signatureHelp/hover requests from the user. I'm not sure how well this approach will scale as we introduce more code actions 🤔

The "import missing symbol" code action doesn't suffer from the same problems because it only does expensive analysis when the cursor is over a "no such symbol" diagnostic.

gabro: to clarify my intention when bringing up `signatureHelp`: I wouldn't merge it as is. I would rather extract the relevant bits off the `SignatureHelpProvider` and provide a more general API to get the active signature info (`signatureHelp` does a lot more, which we don't need for this code action and it also returns data in an inconvenient format).

cb372: >Is there a way to distinguish between code actions that are explicitly requested by the users and code action requests that are automatically triggered by the editor?

Unfortunately not. On cursor move VS Code sends a code actions request with `only = null`, i.e. it requests all possible actions. Then if the user explicitly requests "Refactor..." or "Quick Fix...", it reuses the info it has already retrieved. I don't think it ever sends a code actions request in which it specifies a value for `only`.

I'm not sure how much work it would be, but as a long term solution would it make sense for `Compilers` (or some layer on top of that) to cache the typechecking results? It could be computed once per `didChange` and then reused by various code actions, or whoever else might need it.

@olafurpg What if we reverted to the previous implementation, using `definitionProvider` and `symbolSearch` instead of the presentation compiler? The code was more cumbersome, but would that alleviate your performance concerns? It was traversing a pre-built scalameta tree, then looking up the symbol in semanticdb, and finally looking up the symbol's definition in the mtags index, all of which should be faster than typechecking?

@gabro That sounds reasonable but IIUC, even if we made the API nicer, any implementation that uses the presentation compiler hits the problem of triggering an expensive typecheck every time we generate the code action.

Not quite sure how to proceed from here...

olafurpg: @gabro suggested a nice solution in an offline conversation that the code action only needs to analyze whether the refactoring can be applied at the given position. If the refactoring is eligible, we return an ID for a command that performs the expensive analysis only if user explicitly chooses to apply the code action.

I have a feeling it will be best add one method to the `PresentationCompiler` API that returns what code actions are available at a given position. This method will need to be aware of all available code actions. For "add named argument", this analysis involves checking only if the cursor is above an argument to a method call.

What do you think?

gabro: By the way, I've discovered that code actions that return a `codeEdit` directly are not necessarily supported by all clients, since they were added in a later revision of LSP. We should honor the `codeActionLiteralSupport` sent by the client.

At a first glance, all major LSP clients seem to support this (I've checked VS Code, coc.nvim, sublime's LSP, emacs-lsp), but we should honor the setting anyway.

I can take care of it in a separate PR.

---

For this specific PR, I have a proposal: can we split it into a PR that contains the refactoring to the codebase and a PR that implements the specific code action?
I think the refactoring is pretty uncontroversial and we could merge it as is, whereas the code action requires a bit more work to nail the PC integration.

What do you think @cb372?

Also, thank you for your patience: we're figuring this out as we go 😅

cb372:
I'm closing this PR. Maybe we can come back to it once you've got a scalable architecture for code actions in place.

I hope the PR was useful at least for the discussion it provoked. I certainly had fun 😄

gabro:
Thank you @cb372 the discussion and the refactor were immensely useful and I think we have a good idea on how to proceed now. I'll ping you here once we're done with the basics and we can give this another go!

PR #1080, opened 2019-11-21
Update documentation

ckipp01:
There were a few things I noticed that should be updated in the docs, so I took the opportunity to just do a comb through and update all the things I thought should be updated, correct some small grammar mistakes, and remove some things that were no longer valid.

We're missing a few things such as a section on worksheets and symbol renames, but I wasn't sure if that was intentional since it's not in the latest stable release.

### DIFF
--- a/docs/editors/overview.md
+++ b/docs/editors/overview.md
@@ -149,10 +149,6 @@ functionality.
 </tbody>
 </table>
 
-> Metals is a new project with limited features. If you are learning Scala or
-> are looking for a rich IDE experience it is recommended to use IntelliJ
-> instead.
-
 ## Installation
 
 **Single click**: Metals is easy to install and requires minimal configuration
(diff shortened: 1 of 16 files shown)


docs/editors/overview.md
```diff
@@ -345,7 +338,6 @@ extensions, see [integrating a new editor](new-editor.md).
 
 Metals does not support the following features:
 
-- Rename symbol
```
tgodzik: I think we should remove this only once the stable release is out - I have a number of updates that will go together with that.

docs/editors/vim.md
```diff
@@ -103,17 +103,17 @@ recommended default:
...
+In order to get comment highlighting, please add:
 
 ```vim
 autocmd FileType json syntax match Comment +\/\/.\+$+
 ```
 
 ### Generating metals binary
 
-If you now start Vim in a Scala project then it will fail since the `metals-vim`
+If you now start Vim in a Scala projec, it will fail since the `metals-vim`
```
tgodzik: projec -> project

docs/editors/overview.md
ckipp01: Sounds good. I'll add it back in.

olafurpg:
This is awesome! Thank you so much 😊

Issue #1081, opened 2019-11-22
onTypeFormatting not working as expected on multi-line strings on coc.nvim


ckipp01:
**Describe the bug**
If you have the following code and you hit enter at the `<enter>` position, I'd expect the MultilineStringFormatter to insert a `|` on the new line. This works as expected in VS Code when the relevante settings are turned on, but with the relevant settings turned on in coc.nvim, no `|` is inserted.

Code example:
```scala
package example

object Example extends App {

  """|<enter>
     |""".stripMargin
}
```

**To Reproduce**
1. Copy the code above
2. Make sure coc.nvim and metals are installed correctly
3. Mark `"coc.preferences.formatOnType"` to true in your `coc-settings.json`
4. Hit enter in the marked position

**Expected behavior**
I'd like to get a `|` automatically inserted like is done in VS Code. I'd expect that the response in the logs from coc.nvim should be the same as those in VS Code rather than an empty array.

**Installation:**
- Operating system: macOS
- Editor: Vim
- Metals version: v0.7.6-Latest-Snapshot

**Additional context**
Logs from VS  Code
Request:
```json
[Trace - 02:32:55 PM] Received request 'textDocument/onTypeFormatting - (5)'
Params: {
  "position": {
    "line": 6,
    "character": 2
  },
  "ch": "\n",
  "textDocument": {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/src/main/scala/example/Hello.scala"
  },
  "options": {
    "tabSize": 2,
    "insertSpaces": true
  }
}
```
Response:
```json
[Trace - 02:32:55 PM] Sending response 'textDocument/onTypeFormatting - (5)'. Processing request took 6ms
Result: [
  {
    "range": {
      "start": {
        "line": 6,
        "character": 2
      },
      "end": {
        "line": 6,
        "character": 2
      }
    },
    "newText": "   |"
  }
]
```

Logs from nvim with coc.nvim
Request:
```json
[Trace - 02:31:38 PM] Received request 'textDocument/onTypeFormatting - (3)'
Params: {
  "position": {
    "line": 7,
    "character": 0
  },
  "ch": "\n",
  "textDocument": {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/src/main/scala/example/Hello.scala"
  },
  "options": {
    "tabSize": 2,
    "insertSpaces": true
  }
}
```
Response:
```json
[...]

PR #1082, opened 2019-11-23
add in more documentatin for vim setup

ckipp01:
Part of this is related to #1081 since `onTypeFormatting` wasn't working as expected with coc.nvim. There are some extra configuration settings that are needed to ensure this works as expected which aren't clearly documented. There was also a small bug in coc.vim that they fixed really quickly. So props to them. This now works as expected like so:

I also added a few custom commands into the configuration and updated the vim docs to utilize them. They will make manually triggering build imports, connecting to the build server, and running doctor much easier by providing `:MetalsDoctor`, `:MetalsImport`, and `:MetalsConnect` commands like so:

### DIFF
--- a/docs/editors/vim.md
+++ b/docs/editors/vim.md
@@ -98,7 +98,9 @@ recommended default:
       "rootPatterns": ["build.sbt"],
       "filetypes": ["scala", "sbt"]
     }
-  }
+  },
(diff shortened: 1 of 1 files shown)


tgodzik:
@ckipp01 This is amazing! Thanks!

PR #1084, opened 2019-11-25
Remove sbt-coursier

alexarchambault:
It's not necessary since sbt 1.3.x, that uses coursier automatically.


tgodzik:
Tests for 2.12.9 seem to be reliably failing here, not sure why. Though, I would merge this and try to fix them separately.

tgodzik:
Merged it, I will try to figure out or disable tests later today.

PR #1087, opened 2019-11-27
Update to the latest setup-scala.

olafurpg:
This release automatically sets the environment variable `CI=true` so
that CI-related logic gets picked up the same way it does on Travis.
I'm surprised this variable is not set by default in GitHub Actions.


olafurpg:
This makes it easier to see what Scala version failed `+cross/test`

Screenshot 2019-11-27 at 17 29 46

PR #1070, opened 2019-11-17
Make goto definition work for worksheets, fixes #1067

olafurpg:
Previously, goto definition didn't work in `*.worksheet.sc` files
because there were several guards in the implementation against
`*.scala` filenames. Now, goto definition treats `*.sc` files the same
way as `*.scala` files.

PR #1088, opened 2019-11-28
Worksheets support for non-VS Code editors

alekseiAlefirov:
!! The branch of this PR is on top of `olafurpg:worksheets-definition`. #1070 needs to be merged first before this one can be rebased and merged as well.

Current implementation of worksheets functionality (or, to be more accurate, publishing of worksheet evaluation) is VS-Code specific. This PR adds universal solution for other editors with basic LSP implementation.
worksheet-vim

The solution is using `workspace/applyEdit` method.

Additionally, providing package header functionality has been fixed (it also uses `workspace/applyEdit`). Before it failed in some editors (like Vim and Eclipse), because paths of files to edit were sent, not URIs. While regular paths are supported in, for instance, VS Code and Sublime (which is why, probably, the bug had not been spotted earlier), URI is the default parameter for WorkspaceEdit object.


alekseiAlefirov:
Closed to make a draft #1089 first, while #1070 to be merged yet.

PR #1090, opened 2019-11-28
Fix issue with no source dependencies being downloaded by Bloop

tgodzik:
Previously, no source dependencies would be downloaded due to a typo and a misconfiguration.

Now, the typo is fixed and proper Bloop settings applied to all projects.


tgodzik:
Issue with cross tests is known - I am fixing it in a separate PR.

PR #1091, opened 2019-11-28
Fix cross macro tests for older Scala versions

tgodzik:
The issue seems to have migrated - not exactly sure why. It happened after a coursier upgrade - classpath issue? However, since it's for older Scala versions I am inclined to just fix it here.


olafurpg:
It’s an odd case 🤔 we’ll probably drop 2.12.9 soon so this workaround is fine for now

PR #1092, opened 2019-11-29
Decode for better readability

ckipp01:
When I first came across this I found it really hard to understand at first glance until @olafurpg kindly explained it to me. If I understood him correctly, I _believe_ this will work as expected and I think it's much more readable.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
@@ -378,18 +378,19 @@ class CompletionProvider(
       val isTypeMember = kind == CompletionListKind.Type
       params.checkCanceled()
       val matchingResults = completions.matchingResults { entered => name =>
+        val decoded = entered.decoded
+
         /** NOTE(tgodzik): presentation compiler bug https://github.com/scala/scala/pull/8193
          *  should be removed once we drop support for 2.12.8 and 2.13.0
          *  in case we have a comment presentation compiler will see it as the name
          *  CompletionIssueSuite.issue-813 for more details
          */
         val realEntered =
-          if (entered.startsWith("$div$div") ||
-            entered.startsWith("$div$times")) { // start of a comment: "//" or "/*"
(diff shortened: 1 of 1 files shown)


mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -378,18 +378,18 @@ class CompletionProvider(
...
             // we reverse the situation and look for the word from start of complition to either '.' or ' '
-            val reversedString = entered.toString().reverse
-            val lastSpace = reversedString.indexOfSlice("0200u$") // reversed $u0020 = ' '
-            val lastDot = reversedString.indexOfSlice("E200u$") // reversed $u002E = '.'
+            val reversedString = decoded.reverse
+            val lastSpace = reversedString.indexOfSlice(" ")
+            val lastDot = reversedString.indexOfSlice(".")
             val startOfWord =
               if (lastSpace < lastDot && lastSpace >= 0) lastSpace else lastDot
             reversedString.slice(0, startOfWord).reverse
```
ckipp01: The only thing I wasn't 100% sure on is here do we need to also encode this again before we pass it into `matchesSubCharacters` or `matches` since the `entered.toString`?

tgodzik: Not sure either, but might be good to encode it again. Situations like: `weird_method@Name`

tgodzik:
I really just planned to remove it ones we drop support for 2.13.0 and 2.12.8 - it is a bit ugly, but fixed in the compiler itself.

I agree it looks much more readable now, thanks! Although I would be glad to get rid of it altogether :sweat_smile:

mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
olafurpg: We always want to fuzzy match against the decoded name, it would be weird to include `::` if the user types `foo.$colon$col@@`

PR #1095, opened 2019-12-01
Fix renames in case of macro annotations

tgodzik:
Previously, renames would not work correctly in case of macro annotations - it would not find annotated classes definition and create an additional reference pointing to the annotation itself.

Now, we always try to find the definition separately for non local symbols and we check all references to be at the right names.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -27,7 +27,6 @@ import java.net.URI
 import java.nio.file.Paths
 import scala.meta.internal.metals.TextEdits
 import scala.meta.internal.metals.Compilations
-import scala.meta.internal.metals.ReferencesResult
 import org.eclipse.lsp4j.TextDocumentEdit
 import org.eclipse.lsp4j.VersionedTextDocumentIdentifier
 import org.eclipse.lsp4j.ResourceOperation
@@ -76,9 +75,6 @@ final class RenameProvider(
         params.getPosition()
       )
 
-      val refParams =
-        toReferenceParams(params.getTextDocument(), params.getPosition())
-
       val symbolOccurence =
         definitionProvider.symbolOccurence(source, textParams)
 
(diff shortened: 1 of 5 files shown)


metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -89,16 +85,27 @@ final class RenameProvider(
...
-        currentReferences = referenceProvider.references(
-          toReferenceParams(txtParams)
-        )
+        isLocal = occurence.symbol.isLocal
+        currentReferences = referenceProvider
+          .references(
+            // we can't get definition by name for local symbols
+            toReferenceParams(txtParams, includeDeclaration = isLocal),
+            // local symbol will not contain a proper name
+            strictCheck = !isLocal
```
olafurpg: Can you clarify what `strictCheck = true` does?

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -89,16 +85,27 @@ final class RenameProvider(
...
+          )
+          .locations
+        definitionLocation = if (parentSymbols.isEmpty)
```
olafurpg: nit: I always wrap if/else in blocks inside for comprehensions to workaround wonky scalafmt formatting
```suggestion
        definitionLocation = {if (parentSymbols.isEmpty)
        ...
        
        }
```

tests/unit/src/test/scala/tests/RenameLspSuite.scala
```diff
@@ -391,6 +391,36 @@ object RenameLspSuite extends BaseLspSuite("rename") {
     breakingChange = (str: String) => str.replaceAll("Int", "String")
   )
 
+  renamed(
+    "macro-annotation",
+    """|/a/src/main/scala/a/Main.scala
+       |package a
+       |import io.circe.generic.JsonCodec
+       |trait LivingBeing
+       |@JsonCodec sealed trait <<An@@imal>> extends LivingBeing
```
olafurpg: What happens if you rename an `Animal` reference from another file?

tgodzik: That works - I can add another test for that.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: I added a check that looks into what is underneath a range - anything there should be contained in the symbol itself. So string `Animal` should be contained in symbol `a/Animal#` 

Not the best solution, but I haven't figured out how else to filter out the odd reference under `@JsonCodec`

tgodzik: Also I do the check only for renames - I think it's better to sacrifice a it of performance for correctness.

olafurpg: Sounds like a good idea! How does this play with token edit distance? I thought we already kind of do something like this 🤔 Otherwise, how about renaming `strictCheck` into `checkMatchesText`?

tests/unit/src/test/scala/tests/RenameLspSuite.scala
olafurpg: It would be good to add a test for this I think just in case.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: Sure, I will also check with the edit distances - might not work currently because of that

tgodzik: I checked - token distance is not an issue since we use the reference range to find the name in the snapshot. Token distance is taken into account later on.

This is also tested in the RenameLspSuite since we simulate an edit there.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: Renamed to `checkMatchesText`

tests/unit/src/test/scala/tests/RenameLspSuite.scala
tgodzik: Added another case, fortunately works :sweat_smile:

PR #1139, opened 2019-12-04
Remove Deprecated Usage Of String#lines Method

isomarcte:
In JDK11 a method on `String` was added named `lines` which returned a `java.util.stream.Stream`. This collided with the enrichment method from Scala's `StringOps` class by the same name which returned an `Iterator`. The result, for metals, is that it was not possible to build metals using JDK11 or greater.

This commit uses the new enrichment method `linesIterator` which was added to the Scala standard library as a replacement. This now enables metals to compile on JDK11.

See: https://github.com/scala/bug/issues/11125


olafurpg:
LGTM 👍 Thank you! We should probably start running tests in JDK 11 by default and run only one job on JDK 8 🤔

PR #1153, opened 2019-12-05
Disable file hashing for improved startup performance.

olafurpg:
Opening a separate PR for this change (it's currently part of #1145 ) to see whether it impacts the CI.

The file hashing feature in the directory watcher makes file watching very expensive to start in large directories.
See http://scalameta.org/metals/blog/2019/01/22/bloom-filters.html#indexing-time for more details on how expensive this step is.
We can safely disable the file hashing feature because it's fine if we get duplicate file watching notifications.
Our mtags indexer is fast enough that it's not a big deal if we end up tokenizing the same file twice.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
@@ -106,6 +106,7 @@ final class FileWatcher(
       .builder()
       .paths(directories)
       .listener(new DirectoryListener())
+      .fileHashing(false)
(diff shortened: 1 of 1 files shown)


olafurpg:
As I suspected, this change does affect the correctness of the file watcher on Linux and not just performance. 

Screenshot 2019-12-05 at 10 37 14


I'm very tempted to try and use a different library for file watching, primarily https://github.com/swoval/swoval For now I will drop this change from #1145 to avoid unwanted regressions.

PR #935, opened 2019-09-19
Add basic support for the Pants build tool

Sohamr:
Previously, Metals didn't work with the Pants build tool https://www.pantsbuild.org/. This PR adds support for Pants that works similarly to the build tool support with sbt, Gradle, Maven and Mill.

* the user needs to configure the `pants-targets` setting to specify what parts of their Pants build should be exported to Metals/Bloop.
* the user is prompted to import a Pants workspace on first startup
* when `BUILD` files change, Metals will prompt to the user to "import changes"

This PR is still a WIP. There are several issues related to handling large number of individual source files that belong to a Pants target.

PR #1155, opened 2019-12-05
Use atomic file move operations.

olafurpg:
Following up on https://github.com/scalameta/metals/pull/935#discussion_r354406977

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -38,6 +38,7 @@ import scala.meta.tokens.Token
 import scala.util.Properties
 import scala.{meta => m}
 import java.nio.file.StandardOpenOption
+import scala.util.control.NonFatal
 
 /**
  * One stop shop for all extension methods that are used in the metals build.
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -344,7 +345,27 @@ object MetalsEnrichments
...
+      try {
+        Files.move(
+          tmp,
+          path.toNIO,
+          StandardCopyOption.REPLACE_EXISTING,
+          StandardCopyOption.ATOMIC_MOVE
+        )
+      } catch {
+        case NonFatal(_) =>
+          Files.move(tmp, path.toNIO, StandardCopyOption.REPLACE_EXISTING)
```
tgodzik: So there is still a risk we will have a non-atomic operation? Shouldn't we retry the atomic move again or resign altogether.

olafurpg: Some file systems don’t support atomic moves so the best we can do is fall back to a normal move.

gabro: Could there be a more specific exception to catch here?

olafurpg: The docs don't specify what exception will be thrown so I didn't want to be too specific.

olafurpg:
Ignoring unrelated CI failures.

```
 X tests.TreeViewLspSuite.no-op 3539ms 
  tests.TestFailedException: failed assertion at /Users/runner/runners/2.162.0/work/metals/metals/tests/unit/src/test/scala/tests/TreeViewLspSuite.scala:499
```

PR #1163, opened 2019-12-06
Add breakpoint support while debugging

marek1840:
Previously, there were no possibility to add a breakpoint while debugging. 
Now, this changes: user can add a breakpoint to `scala` or `java` file and expect the debugger to stop there, presenting the stack frames and their variables.

add-breakpoints-support


metals/src/main/scala/scala/meta/internal/metals/debug/RuntimeSourceAdapter.scala
```diff
@@ -0,0 +1,59 @@
...
+  private def searchAsClassPathSymbol(source: Source): Option[String] = {
+    val container =
+      source.getPath.stripSuffix(source.getName).replaceAll("/", ".")
+    val pattern = source.getName
+      .stripSuffix(".scala")
+      .stripSuffix(".java")
+
+    val symbols = for {
+      target <- targets.view
+      symbol <- symbolProvider.searchExact(pattern, Some(target))
```
olafurpg: It's quite heavyweight to go via `WorkspaceSymbolProvider` to find the location of a definition. Did you try using the `DefinitionIndex` directly? It's possible to compute mtags "toplevels" on the fly given the source file.

olafurpg: WorkspaceSymbolProvider supports fuzzy symbol searching, which I don't think we need here

marek1840: You are right, using DefinitionProvider works! It has one regression in regards to the WorksapceSymbolProvider but I think it can be addressed later: ambiguous source files, like:
- `foo/a/b/Foo.scala`
- `bar/a/b/Foo.scala`
where `foo` and `bar` are different, unrelated modules (let's say with variants of classes for different scala versions or something like that).

In this case, the `OnDemandSymbolIndex` fails us, because the top-level symbols from the latter module would overwrite the symbols from the earlier

olafurpg: The same limitation applies for “goto definition” and this issue is totally fixable.

metals/src/main/scala/scala/meta/internal/metals/ClassPath.scala
```diff
@@ -0,0 +1,35 @@
+package scala.meta.internal.metals
+import scala.meta.internal.semanticdb.SymbolOccurrence
+
+object ClassPath {
```
olafurpg: Can we give this a more specific name? For example: `TypeSignature.scala` with `TypeSignature.fromSymbolOccurrence`. There is already a `scala.meta.io.Classpath`, which is a wrapper around a JVM classpath such as `a.jar:/user/b/classes`

metals/src/main/scala/scala/meta/internal/metals/ClassPath.scala
```diff
@@ -0,0 +1,35 @@
...
+  final case class TypeSignature(value: String) extends AnyVal
```
olafurpg: Did you measure that allocations from this class were causing problems? If not, let's remove `AnyVal`
```suggestion
  final case class TypeSignature(value: String)
```

metals/src/main/scala/scala/meta/internal/metals/ClassPath.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  def toTypeSignature(definition: SymbolOccurrence): TypeSignature = {
+    import scala.meta.internal.semanticdb.Scala._
+    @scala.annotation.tailrec
+    def qualifier(symbol: String, acc: List[String]): String = {
+      val owner = symbol.owner
+      if (owner == Symbols.RootPackage || owner == Symbols.EmptyPackage) {
```
olafurpg: What about the case when `owner == Symbols.None`?

metals/src/main/scala/scala/meta/internal/metals/ClassPath.scala
```diff
@@ -0,0 +1,35 @@
+package scala.meta.internal.metals
+import scala.meta.internal.semanticdb.SymbolOccurrence
+
+object ClassPath {
+  final case class TypeSignature(value: String) extends AnyVal
+
+  def toTypeSignature(definition: SymbolOccurrence): TypeSignature = {
+    import scala.meta.internal.semanticdb.Scala._
+    @scala.annotation.tailrec
+    def qualifier(symbol: String, acc: List[String]): String = {
```
olafurpg: Did you try using `ownerChain`? It's easy to end in infinite recursion when walking up owner chains

```
      "a/b/c#".ownerChain.map(_.desc.value)
res0: List[String] = List("_root_", "a", "b", "c")

```

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -411,6 +411,13 @@ object MetalsEnrichments
 
     def toAbsolutePath: AbsolutePath =
       AbsolutePath(Paths.get(URI.create(value.stripPrefix("metals:")))).dealias
+
+    def findAll(pattern: String): Iterator[Int] = {
```
olafurpg: Did you try using `findAllMatchIn`?

Screenshot 2019-12-15 at 11 33 17

metals/src/main/scala/scala/meta/internal/metals/WorkspaceSymbolProvider.scala
```diff
@@ -31,14 +31,28 @@ final class WorkspaceSymbolProvider(
   var inDependencies: ClasspathSearch =
     ClasspathSearch.fromClasspath(Nil, bucketSize)
 
-  def search(query: String): Seq[l.SymbolInformation] = {
-    search(query, () => ())
+  def search(
```
olafurpg: Are these changes used?

tests/unit/src/main/scala/scala/meta/internal/metals/debug/Debugger.scala
```diff
@@ -47,6 +64,88 @@ final class Debugger(server: RemoteServer)(implicit ec: ExecutionContext) {
...
+        val args = new NextArguments()
+        args.setThreadId(threadId)
+        server.next(args).asScala.ignoreValue
+      case cause =>
+        val error = s"Unsupported debug step $cause"
+        Future.failed(new IllegalStateException(error))
+    }
+  }
+
+  def stackFrame(threadId: lang.Long): Future[StackFrame] = {
```
olafurpg: nit: let's use `import java.lang{Long => JLong}` instead

tests/unit/src/main/scala/scala/meta/internal/metals/debug/DummyServer.scala
```diff
@@ -0,0 +1,31 @@
+package scala.meta.internal.metals.debug
+import org.eclipse.lsp4j.jsonrpc.MessageConsumer
+import org.eclipse.lsp4j.jsonrpc.messages.Message
+import org.eclipse.lsp4j.jsonrpc.messages.ResponseMessage
+import scala.collection.mutable
+
+final class DummyServer(
```
olafurpg: For consistency with `TestingServer` and `TestingClient`?
```suggestion
final class TestingDebugServer(
```

olafurpg:
Would it be possible to add explicit "debug" code lenses next to "run" and "test"? It would be nice to be able to ignore the breakpoints while when clicking on "run"/"test"

metals/src/main/scala/scala/meta/internal/metals/ClassPath.scala
marek1840: You are right, let's avoid the confusion. Also, Classpath doesn't reflect the intention behind this object very well. In the end, I went with `JvmSignatures` to distinguish it form other kinds (for example semanticdb signatures)

tests/unit/src/main/scala/scala/meta/internal/metals/debug/Debugger.scala
marek1840: Good catch. In fact, we should use the Scala `Long` here.

metals/src/main/scala/scala/meta/internal/metals/ClassPath.scala
marek1840: Thanks, I didn't know the method!

metals/src/main/scala/scala/meta/internal/metals/ClassPath.scala
marek1840: With your next suggestion (`ownerChain`) it should be covered

metals/src/main/scala/scala/meta/internal/metals/WorkspaceSymbolProvider.scala
marek1840: Not after switching to the `DefinitionProvider`, no. Thanks for catching - reverted.

marek1840:
> I'm struggling to get the "Variables" view to work as expected when running tests. It works great when running main methods

It was fixed in https://github.com/scalacenter/bloop/pull/1115

marek1840:
>Would it be possible to add explicit "debug" code lenses next to "run" and "test"? It would be nice to be able to ignore the breakpoints while when clicking on "run"/"test"

Certainly, I will look into it - possibly as a separate PR, since this one is already way too big.

tests/unit/src/main/scala/scala/meta/internal/metals/debug/DummyServer.scala
marek1840: good catch. In fact, we should use the Scala `Long` here

metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
```diff
@@ -0,0 +1,88 @@
...
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.mtags.Mtags
+import scala.meta.internal.semanticdb.SymbolOccurrence
+
+private[debug] final class SetBreakpointsRequestHandler(
+    server: ServerAdapter,
+    adapters: MetalsDebugAdapters
+)(implicit ec: ExecutionContext) {
+
+  // TODO the partitioning does not work for cases when the same file has other types defined
```
olafurpg: I'm not sure I understand this TODO, could we open a followup ticket?

metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathProvider.scala
```diff
@@ -0,0 +1,53 @@
...
+      .stripSuffix(".java")
+      .replace("\\", "/") // adapt windows paths to the expected format
+
+    val symbols = for {
+      symbol <- Set(symbolBase + ".", symbolBase + "#")
+      definition <- definitionProvider.fromSymbol(symbol).asScala
+    } yield definition.getUri.toAbsolutePath
+
+    if (symbols.isEmpty) {
+      scribe.warn(s"No matching symbol on the classpath for [$symbolBase]")
```
olafurpg: nit: for consistency with logging statements in the codebase
```suggestion
      scribe.warn(s"no definition for symbol: $symbolBase)
```

metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathProvider.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  private def searchAsSourceFile(source: Source): Option[AbsolutePath] = {
+    val files = for {
+      target <- targets.view
+      sourceFile <- buildTargets.buildTargetSources(target)
+      if sourceFile.filename == source.getName
+    } yield sourceFile
+
+    if (files.isEmpty) {
+      scribe.warn(s"No matching source file for $source")
```
olafurpg:
```suggestion
      scribe.warn(s"no matching source file: $source")
```

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
```diff
@@ -0,0 +1,468 @@
...
+import tests.BaseDapSuite
+import scala.meta.internal.metals.debug.DebugStep._
+import scala.meta.internal.metals.debug.DebugFileLayout
+import scala.meta.internal.metals.debug.StepNavigator
+
+object BreakpointDapSuite extends BaseDapSuite("debug-breakpoint") {
+
+  // disabled, because finding enclosing class for the breakpoint line is not working
+  // see [[scala.meta.internal.metals.debug.SetBreakpointsRequestHandler]]
+  assertFileBreakpoints("preceding-class", disabled = true)(
```
olafurpg: Interesting, this should be totally fixable inside `Mtags.toplevels`. I've considered extending the parser to emit global def/val/var symbols as well. Can we open a ticket for this?

olafurpg: This is BTW a totally fine limitation for now, we'll make it clear in the release announcement that the debugging support has known issues

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
```diff
@@ -0,0 +1,468 @@
...
+         |object Main {
+         |    def main(args: Array[String]): Unit = {
+         |      not.matching.Foo.call()
+         |    }
+         |}
+         |""".stripMargin
+    )
+  )
+
+  // TODO OnDemandSymbolIndex doesn't distinguish between build targets
```
olafurpg: This is a known limitation that also applies to "goto definition". We can open a followup ticket and remove this TODO

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
```diff
@@ -0,0 +1,468 @@
...
+         |object Main {
+         |    def main(args: Array[String]): Unit = {
+         |      call()
+         |    }
+         |}
+         |""".stripMargin
+    )
+  )
+
+  assertProjectBreakpoints("not-matching-filename")(
```
olafurpg: 😍

olafurpg: So much blood, sweat and tears to deliver this functionality 💪

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
```diff
@@ -0,0 +1,468 @@
...
+                |    x <- List()
+                |  } {
+                |>>    println(x)
+                |    }
+                |  }
+                |}
+                |""".stripMargin
+  )
+
+  // TODO investigate what happens
```
olafurpg: Let's open a followup ticket or remove the TODO comment.

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
```diff
@@ -0,0 +1,468 @@
...
+                |  def main(args: Array[String]): Unit = {
+                |    for {
+                |>>    x <- List()
+                |    } println(x)
+                |  }
+                |}
+                |""".stripMargin
+  )
+
+  // TODO investigate what happens
```
olafurpg: Let's open followup ticket

tests/unit/src/test/scala/tests/debug/StackFrameDapSuite.scala
```diff
@@ -0,0 +1,159 @@
...
+        Scope.local(
+          Variable("this: Main$"),
+          Variable("args: String[]"),
+          Variable("foo: Foo")
+        )
+      )
+    )
+  )
+
+  def assertStackFrame(
```
olafurpg: Can we add a test case with symbolic identifiers? I noticed that `::` was displayed as `$colon$colon` in the variable view

Screenshot 2019-12-16 at 11 03 07

No need to change the current behavior, just document it with a test case.

tests/unit/src/test/scala/tests/debug/StepDapSuite.scala
```diff
@@ -0,0 +1,172 @@
...
+         |    foo()
+         |  }
+         |}
+         |""".stripMargin
+    ),
+    main = "a.Main",
+    instrument = steps =>
+      steps
+        .at("a/src/main/scala/Main.scala", line = 5)(StepOut)
+        .at("a/src/main/scala/Main.scala", line = 9)(Continue)
```
olafurpg: Could we avoid hardcoding the line numbers here somehow? For example, using comment markers 
```scala
             |>>  println() // debug: step 1
             
// ...
           |    foo() // debug: step 2
// ...
                .at("a/src/main/scala/Main.scala", line = "step 1")(StepOut)
                .at("a/src/main/scala/Main.scala", line = "step 2")(Continue)
```

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
```diff
@@ -0,0 +1,468 @@
...
+
+    testAsync(name) {
+      cleanWorkspace()
+      val fileLayouts = sources.map(DebugFileLayout.apply)
+
+      val layout =
+        s"""|/metals.json
+            |{ "a": {}, "b": {} }
+            |
+            |${fileLayouts.map(_.layout).mkString("\n")}
```
tgodzik: Can we reuse FileLayout here?

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProxy.scala
```diff
@@ -34,31 +42,57 @@ private[debug] final class DebugProxy(
   }
 
   private def listenToServer(): Unit = {
-    Future(server.listen(handleServerMessage)).andThen { case _ => cancel() }
+    Future(server.onReceived(handleServerMessage))
+      .andThen { case _ => cancel() }
   }
 
   private val handleClientMessage: MessageConsumer = {
+    case null =>
```
tgodzik:
```suggestion
   message => Option(message).foreach{
```

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
marek1840: Opened https://github.com/scalameta/metals/issues/1195

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
marek1840: Opened https://github.com/scalameta/metals/issues/1196

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
marek1840: Fixed, it turns out one needs a non-empty list to iterate over (and there was `for (i <- List())`)

tests/unit/src/test/scala/tests/debug/StepDapSuite.scala
marek1840: That would be problematic for cases when a single line is hit multiple times. Also, developer would have to either declare the methods in order of breakpoints or somehow indicate the order within the commit markers which I think would be less readable than having all of the steps in one place. Also, we would need to encode the step type in the comment and it all becomes not worth the hassle in my opinion. 
The test cases are small enough that they shouldn't be changed that often, so let's hope we won't have too many reasons to do so.

metals/src/main/scala/scala/meta/internal/metals/JvmSignatures.scala
```diff
@@ -0,0 +1,36 @@
...
+
+          val delimiter =
+            if (desc.isPackage) "."
+            else if (desc.isTerm) "$"
+            else if (tail.nonEmpty) "$" // nested class
+            else ""
+          loop(tail, fqcn.append(delimiter))
+      }
+    }
+    val fqcn = loop(definition.symbol.ownerChain, new StringBuilder)
```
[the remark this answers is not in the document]
marek1840: Added a test case for that

tests/unit/src/test/scala/tests/debug/StepDapSuite.scala
olafurpg: Fair enough, as a general principle I try to avoid hardcoded line numbers in test cases because the line numbers change when you change the code.

marek1840:
Looks like mill tests are still flaky.

metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
olafurpg: @marek1840 can we link to the issue here?

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
olafurpg: Did we open a followup ticket?

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
olafurpg: Awesome!

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
olafurpg: Can you elaborate?

tests/unit/src/test/scala/tests/debug/StackFrameDapSuite.scala
olafurpg: Did we add a test case with symbolic identifier?

marek1840: Yup: https://github.com/scalameta/metals/pull/1163/files#diff-e39244aad5f9193c28b3779716ed886bR124

tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala
marek1840: Nope, I was iterating over a `List()` which is empty, hence the breakpoint was not effective

marek1840:
I have associated all of the remaining TODOs with their github issues so that the places which need to be changed are easy to find.

metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
marek1840: done!

metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
```diff
@@ -0,0 +1,86 @@
...
+)(implicit ec: ExecutionContext) {
+
+  // TODO: https://github.com/scalameta/metals/issues/1195
+  def apply(
+      request: SetBreakpointsArguments
+  ): Future[SetBreakpointsResponse] = {
+    val path = adapters.adaptPath(request.getSource.getPath)
+    val input = path.toAbsolutePath.toInput
+    val occurrences = Mtags.allToplevels(input).occurrences
+    val groups = request.getBreakpoints.groupBy { breakpoint =>
```
jvican: Are we sending fqcns only for symbols from third-party deps? Or are we sending fqcns for symbols for Java and Scala sources defined in the project? Are we following the recommendation in [the PR introducing support for breakpoints in Bloop](https://github.com/scalacenter/bloop/pull/1087)?

marek1840: We are avoiding extra branching by not checking whether the source is defined within the workspace or is a part of a dependency. We are confident this approach is working due to the extensive suite of tests we provide alongside. If you know a case when this will fail, please don't hesitate to point to it.

Or did you have another recommendation in mind?

jvican: My recommendation is not to send `dap-fqcn` for files defined inside user-defined projects. It should only be used for third-party dependencies.

Bloop and Zinc are the source of truth here. Regardless of how confident we're from Metals doing the right thing, we should be relying on the source of truth if that's possible to avoid issues/discrepancies in the future.

marek1840: Fair enough. For now, though, we will rely on the compiler to emit `LineNumberTable` according to the JVM spec. 
If the compiler, at any point, proves to be unreliable by messing the line numbers, then I believe neither bloop nor zinc would help us.

jvican: I feel pretty strongly this is not the way to go. You should not be reimplementing this part of the logic. Metals should be reusing the breakpoint logic for user-defined sources implemented in Bloop.

esHack:
It would be great if we some documentation on how to run debugger using metals

tgodzik:
> It would be great if we some documentation on how to run debugger using metals

There will be additional code lenses for debugging with breakpoints, we also need to merge the VS Code extension PR for this to work.

metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
marek1840: The logic you say I am reimplementing is required for the library sources anyway, since bloop solution works only for workspace sources. 
Since it is required either way, right now, I don't see a reason for falling back to bloop and not using it also for user-defined sources. It does not mean there is no such reason but only that I don't see it currently - can you help me in this regard?

jvican: > The logic you say I am reimplementing is required for the library sources anyway, since bloop solution works only for workspace sources.
  
Yes, the whole reason why we have `dap-fqcn` is so that IDEs that manage library sources can pass in the resolved fully qualified class names.
  
>  I don't see a reason for falling back to bloop and not using it also for user-defined sources. It does not mean there is no such reason but only that I don't see it currently - can you help me in this regard?
  
Yes, the point I'm trying to make is that the logic for library sources is necessary since Bloop doesn't support it. But the logic for workspace sources isn't. There is a reference implementation that it's as good as it gets: it's using the same resolution mechanism the JVM uses for debugging (what's happening under the hood is that the JVM gets our fully qualified class name we pass it and undoes some of the work Bloop has done and finds the class file the loaded class was loaded from in the classloader).
  
At a high level, the process looks like this:
  
Bloop: `$WORKSPACE_SOURCE_FILE` -> `$WORKSPACE_CLASS_FILE` -> `$FQCN` ->
JVM: `$FQCN` -> `$WORKSPACE_CLASS_FILE`
  
The JVM uses the `LineNumberTable` of the `$WORKSPACE_CLASS_FILE` to correctly set the breakpoint. This is why it's so important that `WORKSPACE_CLASS_FILE` matches the one the JVM picks and why the Bloop implementation is the source of the truth -- because it mimics exactly the logic of the JVM.
  
What Metals is doing here might be correct based on the semantic-level information you have. But it's creating a layer of indirection that should be avoided as it could produce issues or inconsistent results in the future. I think it's a good practice to rely on the source of truth if it's possible and in this case it is.

PR #1164, opened 2019-12-06
Try fixing the Windows CI

gabro:
Let's see whether this fixes the CI on Windows.

My suspicion is that the recent rewrite of the Windows Sbt launcher (https://github.com/sbt/sbt-launcher-package/commit/b5fd78f74bf1dc38bbc4203067dabb0f54dd5ea7#diff-0e467aba72fad1d1010d8efae1c500d6) has a regression in the handling of `.jvmopts`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,12 +12,15 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
+    env:
+      JAVA_OPTS: -Xss4m -Xms1G -Xmx8G -XX:ReservedCodeCacheSize=1024m -XX:+TieredCompilation -XX:+CMSClassUnloadingEnabled -Dfile.encoding=UTF-8 -Djna.nosys=true
     steps:
       - uses: actions/checkout@v1
-
       - uses: olafurpg/setup-scala@v7
       - name: Run unit tests
-        run: bin/test.sh unit/test
+        run: |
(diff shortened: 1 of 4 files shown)


gabro:
mm, I guess I was wrong. Windows CI is still failing even with 1.3.3

olafurpg:
The logs make it look like the issue is not caused by sbt but by the `bash` script mode on Windows.
```
Resolution error: Error downloading ch.epfl.scala:bloopgun-core_2.12:1.3.42511815
  not found: C:\Users\runneradmin\.ivy2\local\ch.epfl.scala\bloopgun-core_2.12\1.3.42511815\ivys\ivy.xml
  not found: https://repo1.maven.org/maven2/ch/epfl/scala/bloopgun-core_2.12/1.3.42511815/bloopgun-core_2.12-1.3.42511815.pom
```

I wonder if GitHub Actions just updated their `windows-latest` image 🤔

gabro:
@olafurpg ah, that might actually be. On the documentation, you can see

image

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners

which hints at the fact that something has changed on December 3rd (recently)

gabro:
@olafurpg I’ve found an issue with a regex in the test script which didn’t play well with the “snapshot” version of bloop.
I’ve fixed it in the latest commit, but it seems unrelated to the JVM failure

olafurpg:
Back to green CI 🎉

PR #1165, opened 2019-12-06
Fix Runtime Error In Metals Bench

isomarcte:
A path to file needed to run the `CompletionBench` benchmark was hard coded as scala `2.12.8`. Since this is no longer the version of Scala used to compile `metals` this fails at runtime.

This was changed to use sbt-buildinfo to populate the correct version of scala for the given build from the build definition directly.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -451,6 +451,7 @@ lazy val slow = project
 
 lazy val bench = project
   .in(file("metals-bench"))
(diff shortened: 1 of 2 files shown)


olafurpg:
Ignoring unrelated CI failures. The Windows issues should already be fixed in #1164

PR #1168, opened 2019-12-08
Fix Windows CI Tests

isomarcte:
Windows CI tests are failing because the `.jvmopts` file in the CI environment has DOS style line endings, e.g. `\r\n`. This _is not the case in VSC_ and I am unsure how/why it is getting modified when running in the Windows CI environment.

Nevertheless, the result is that this bash function is reading values in with carriage returns at the end.

https://github.com/paulp/sbt-extras/blob/59a9a1d24b1134587d1512012788421867d690fa/sbt#L495
```bash
readConfigFile() {
  local end=false
  until $end; do
    read -r || end=true
    [[ $REPLY =~ ^# ]] || [[ -z $REPLY ]] || echo "$REPLY"
  done <"$1"
}
```

This, `-Xss4m` becomes `-Xss4m\r`. This is also why the error message looks odd, e.g. `Invalid thread stack size: -Xss4m`. The invalid part of that is the `\r` which is not literally printed.

As work around we run a `sed -i 's/\r//g'` over the `.jvmopts` file before starting the build.

This should be considered a short term work around. One of two different changes should take place long term.

* `sbt-extras` should be updated to handle `\r` in the `.jvmopts` file.
    * This is a pretty small change, but I am unsure if the `sbt-extras` maintainers want this behavior. It would be perfect valid to require Unix line endings.
* `metals` needs to figure out why the Windows CI environment is mangling the files to have DOS line endings.
    * Just because DOS line endings are standard on Windows, doesn't mean that files should be modified _in anyway_.

----

#

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,14 +12,11 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-    env:
-      JAVA_OPTS: -Xss4m -Xms1G -Xmx8G -XX:ReservedCodeCacheSize=1024m -XX:+TieredCompilation -XX:+CMSClassUnloadingEnabled -Dfile.encoding=UTF-8 -Djna.nosys=true
     steps:
       - uses: actions/checkout@v1
       - uses: olafurpg/setup-scala@v7
       - name: Run unit tests
         run: |
-          rm .jvmopts
           bin/test.sh unit/test
         shell: bash
   sbt:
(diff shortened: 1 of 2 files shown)


isomarcte:
@gabro @olafurpg I see you guys fixed this issue with a different work around, going through the environment. 

Do you want me to close this or do you want to go back to using the `.jvmopts` now that we've figured out why it was failing.

gabro:
Hey @isomarcte, thanks for looking into this and for finding the root cause. I much prefer your workaround since it addresses the problem in a more specific way.

About the reason, I think what’s to blame is the bash shell on the latest windows workers (which appear to have been updated recently)

gabro:
Ah, I just thought of something: could it be that git converts the line endings on checkout? We can test this using a .gitattributes file that sets the autocrlf setting

gabro:
Here it is, this seems to support my thesis https://github.community/t5/GitHub-Actions/git-config-core-autocrlf-should-default-to-false/td-p/30445

isomarcte:
@gabro good call on the `.gitattributes` settings. Closing this PR in favor of #1170

PR #1169, opened 2019-12-08
Fix renames for edited companion objects and case class apply/c…

tgodzik:
*Fix renames for edited companion objects*
Previously, when renaming in a workspace with edited files with macro annotated classes some symbols would not be renamed properly. Now, we make sure that token distance is taken into account in those files. 

*Fix renaming copy/apply occurrence*
Previously, all copy/apply occurences for case classes would be renamed to the new class name. Now, we do not rename those symbols and additionally `textDocument/references` don't return them.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
@@ -113,24 +113,6 @@ final class ReferenceProvider(
           info.symbol.owner,
           Descriptor.Type(info.displayName)
         )
-    // Returns true if `info` is a synthetic `copy` or `apply` of the occurrence class symbol.
-    def isCopyOrApplyMethod(info: SymbolInformation): Boolean =
-      info.isMethod &&
-        isCopyOrApply(info.displayName) &&
-        occ.symbol == (Symbol(info.symbol) match {
-          case GlobalSymbol(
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -113,24 +113,6 @@ final class ReferenceProvider(
           info.symbol.owner,
           Descriptor.Type(info.displayName)
         )
-    // Returns true if `info` is a synthetic `copy` or `apply` of the occurrence class symbol.
-    def isCopyOrApplyMethod(info: SymbolInformation): Boolean =
```
tgodzik: @olafurpg I removed it since it didn't seem intuitive to find apply/copy for case class symbols. I can change that to be behind a flag - I am not sure what was the motivations for adding them to references in the first place :thinking:

olafurpg: The motivation was mostly for testing purposes, to preserve a bijection between definition/references.

I’m fine with removing this behavior but it may require adjustments in ReferenceLspSuite

tgodzik: Fixed the tests in ReferenceLspSuite, should be alright.

PR #1170, opened 2019-12-08
Disable EOL Conversion On /.jvmopts

isomarcte:
Add a `.gitattributes` file to prevent end of line (EOL) conversion on the `/.jvmopts` file. This was the causing windows based CI failures as the JVM options were being parsed with a carriage return at the end of them.

This commit also re-enables use of the `.jvmopts` file in the Windows CI environment.

----

#

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/.jvmopts eol=lf
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,14 +12,11 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-    env:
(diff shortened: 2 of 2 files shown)


gabro:
Thanks for following up! Would you mind adding a comment explaining that we do this because sbt-extras requires LF even on Windows?

olafurpg:
The windows failure looks unrelated to this PR

```

[error occurred during error reporting (null), id 0xc0000005]

#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (malloc) failed to allocate 1114736 bytes for Chunk::new
# An error report file with more information is saved as:
# D:\a\metals\metals\tests\unit\hs_err_pid2276.log
```
I suspect we need to drop the `-Xmx4G` line from .jvmopts

isomarcte:
> Thanks for following up! Would you mind adding a comment explaining that we do this because sbt-extras requires LF even on Windows?

@gabro sorry I didn't see this until just now. Do you want me to add a comment still? To be clear, it isn't really a `sbt-extras` problem, it's more of a `bash` problem with how it splits a file on `\n` when parsing in that function.

gabro:
that’s fine, I just wanted to provide some context, but this PR is easily searchable anyway. Thanks again

PR #1171, opened 2019-12-08
Remove Usage Of Deprecated `lines` Method

isomarcte:
Another instance of the deprecated `lines` method was recently added to the code base. This method causes metals to be not build on JDK >= 11.

This commit also updates the CI to build with JDKs 8 and 11 using AdpotOpenJDK. The majority of tests are JDK 11 as default, with a single test using JDK8 as a legacy compatibility smoke test. This should help prevent future additions of this method.

Finally, this commit drops cross building with deprecated scala versions, as they are unable to parse class files generated with bytecode for JDKs > 8.

See: ce9c890c27efec49fd8ad78746e2f406bfd89d3b and 4d78f9cf5a2e6c49d8a6e2474a49a8bbb99e216c for more context.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,10 +11,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macOS-latest]
+        os: [windows-latest, macOS-latest, ubuntu-latest]
+        java: [8, 11]
+        exclude:
+          - os: windows-latest
+            java: 8
+          - os: macOS-latest
+            java: 8
     steps:
       - uses: actions/checkout@v1
       - uses: olafurpg/setup-scala@v7
+        with:
+          java-version: "adopt@1.${{ matrix.java }}"
       - name: Run unit tests
         run: |
           bin/test.sh unit/test
@@ -25,6 +33,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: olafurpg/setup-scala@v7
+        with:
+          java-version: "adopt@1.11"
       - name: Run Sbt tests
         run: bin/test.sh 'slow/testOnly -- tests.sbt'
   maven:
@@ -33,6 +43,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: olafurpg/setup-scala@v7
+        with:
+          java-version: "adopt@1.11"
       - name: Run Maven tests
         run: bin/test.sh 'slow/testOnly -- tests.maven'
   gradle:
@@ -41,6 +53,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: olafurpg/setup-scala@v7
+        with:
+          java-version: "adopt@1.11"
       - name: Run Gradle tests
         run: bin/test.sh 'slow/testOnly -- tests.gradle'
   mill:
@@ -49,6 +63,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: olafurpg/setup-scala@v7
+        with:
+          java-version: "adopt@1.11"
       - name: Run Mill tests
         run: bin/test.sh 'slow/testOnly -- tests.mill'
   pants:
@@ -56,7 +72,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - uses: olafurpg/setup-scala@v5
+      - uses: olafurpg/setup-scala@v7
+        with:
+          java-version: "adopt@1.11"
       - name: Run Pants tests
         run: bin/test.sh 'slow/testOnly -- tests.pants'
   feature:
@@ -65,6 +83,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: olafurpg/setup-scala@v7
+        with:
+          java-version: "adopt@1.11"
       - name: Run LSP tests
         run: bin/test.sh 'slow/testOnly -- tests.feature'
   cross:
@@ -73,6 +93,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: olafurpg/setup-scala@v7
+        with:
+          java-version: "adopt@1.11"
       - name: Run cross tests
         run: sbt +cross/test
   checks:
@@ -81,6 +103,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: olafurpg/setup-scala@v7
+        with:
+          java-version: "adopt@1.11"
       - uses: actions/setup-node@v1
       - run: git fetch --depth=10000
       - run: git fetch --tags
--- a/tests/mtest/src/main/scala/tests/BaseSuite.scala
+++ b/tests/mtest/src/main/scala/tests/BaseSuite.scala
@@ -1,11 +1,15 @@
 package tests
 
+import scala.collection.mutable
 import scala.concurrent.Await
 import scala.concurrent.Future
 import scala.concurrent.duration.Duration
 import scala.language.experimental.macros
-import scala.meta.io.AbsolutePath
+import scala.meta.internal.metals.JdkSources
+import scala.meta.internal.metals.Testing
 import scala.meta.internal.mtags
+import scala.meta.internal.semver.SemVer
+import scala.meta.io.AbsolutePath
 import scala.reflect.ClassTag
 import scala.util.Properties
 import utest.TestSuite
@@ -16,9 +20,6 @@ import utest.framework.TestCallTree
 import utest.framework.Tree
 import utest.ufansi.Attrs
 import utest.ufansi.Str
-import scala.meta.internal.metals.JdkSources
-import scala.meta.internal.metals.Testing
-import scala.collection.mutable
 
 /**
  * Test suite that replace utest DSL with FunSuite-style syntax from ScalaTest.
@@ -37,6 +38,12 @@ class BaseSuite extends TestSuite {
   def beforeAll(): Unit = ()
   def afterAll(): Unit = ()
   def intercept[T: ClassTag](exprs: Unit): T = macro Asserts.interceptProxy[T]
+  def isValidScalaVersionForEnv(scalaVersion: String): Boolean =
+    this.isJava8 || SemVer.isCompatibleVersion(
+      BaseSuite.minScalaVersionForJDK9OrHigher,
+      scalaVersion
+    )
+
   def assertNotEmpty(string: String): Unit = {
     if (string.isEmpty) {
       fail(
@@ -192,4 +199,8 @@ class BaseSuite extends TestSuite {
       .getOrElse(default)
     postProcess(result)
   }
+}
+
+object BaseSuite {
+  val minScalaVersionForJDK9OrHigher: String = "2.12.10"
 }
--- a/tests/slow/src/test/scala/tests/feature/CompletionCrossLspSuite.scala
+++ b/tests/slow/src/test/scala/tests/feature/CompletionCrossLspSuite.scala
@@ -5,11 +5,17 @@ import tests.BaseCompletionLspSuite
 
 object CompletionCrossLspSuite
     extends BaseCompletionLspSuite("completion-cross") {
-  testAsync("basic-211") {
-    basicTest(V.scala211)
+
+  if (super.isValidScalaVersionForEnv(V.scala211)) {
+    testAsync("basic-211") {
+      basicTest(V.scala211)
+    }
   }
-  testAsync("basic-213") {
-    basicTest(V.scala213)
+
+  if (super.isValidScalaVersionForEnv(V.scala213)) {
+    testAsync("basic-213") {
+      basicTest(V.scala213)
+    }
   }
   testAsync("match-211") {
     matchKeywordTest(V.scala213)
--- a/tests/slow/src/test/scala/tests/feature/DefinitionCrossLspSuite.scala
+++ b/tests/slow/src/test/scala/tests/feature/DefinitionCrossLspSuite.scala
@@ -7,12 +7,16 @@ import scala.concurrent.Future
 object DefinitionCrossLspSuite
     extends BaseCompletionLspSuite("definition-cross") {
 
-  testAsync("2.11") {
-    basicDefinitionTest(BuildInfo.scala211)
+  if (super.isValidScalaVersionForEnv(BuildInfo.scala211)) {
+    testAsync("2.11") {
+      basicDefinitionTest(BuildInfo.scala211)
+    }
   }
 
-  testAsync("2.13") {
-    basicDefinitionTest(BuildInfo.scala213)
+  if (super.isValidScalaVersionForEnv(BuildInfo.scala213)) {
+    testAsync("2.13") {
+      basicDefinitionTest(BuildInfo.scala213)
+    }
   }
 
   def basicDefinitionTest(scalaVersion: String): Future[Unit] = {
--- a/tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
+++ b/tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
@@ -12,6 +12,12 @@ abstract class BaseWorksheetLspSuite(scalaVersion: String)
     Some(ClientExperimentalCapabilities(decorationProvider = true))
   override def userConfig: UserConfiguration =
     super.userConfig.copy(worksheetScreenWidth = 40, worksheetCancelTimeout = 1)
+
+  override final def test(name: String)(fun: => Any): Unit =
+    if (super.isValidScalaVersionForEnv(this.scalaVersion)) {
+      super.test(name)(fun)
+    }
+
   if (!isWindows)
     testAsync("completion") {
       for {
--- a/tests/unit/src/test/scala/tests/TreeViewLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/TreeViewLspSuite.scala
@@ -1,8 +1,49 @@
 package tests
 
+import scala.collection.SortedSet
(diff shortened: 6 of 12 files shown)


isomarcte:
The failures in CI seem to be timeout issues unrelated to these changes.

.github/workflows/ci.yml
```diff
@@ -12,9 +12,15 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
+        java-version:
+          - "adopt@1.11"
+          - "adopt@1.8"
+          - "openjdk@1.11"
```
olafurpg: What’s the motivation to have two JDK 11 entries?

tests/unit/src/test/scala/tests/TreeViewLspSuite.scala
```diff
@@ -1,9 +1,50 @@
...
 import scala.meta.internal.tvp.TreeViewProvider
 
 object TreeViewLspSuite extends BaseLspSuite("tree-view") {
 
+  /** The libraries we expect to find for tests in this file.
+   *
+   * @note this value changes depending on the JVM version in use as some JAR
+   *       files have moved to become modules on JVM > 8.
+   */
+  val expectedLibraries: SortedSet[String] = {
```
olafurpg: Thank you for fixing this! 👏

The biggest reason for delaying the upgrade to jdk 11 in our ci was that tests needed to be updated.

.github/workflows/ci.yml
olafurpg: We should include the jdk name in the job name to make it easier to distinguish in the web ui.

gabro:
Thank you @isomarcte for working on this!

@olafurpg what about testing on JDK 11 by the default, and only keeping one “canary” job on jdk 8? My concern is not to slow down the CI too much

olafurpg:
@gabro I'm in favor of that approach 👍

.github/workflows/ci.yml
isomarcte: @olafurpg I'm fine with just one. I added two because I had noticed you were using AdoptOpenJDK for the 1.8 builds. I thought it may be wise to use OpenJDK's "official" implementation when available, which it is for 11, and I also wanted to keep parity with what you were already using, hence AdoptOpenJDK. 

That being said, using just one of the two is probably fine.

.github/workflows/ci.yml
```diff
@@ -5,16 +5,38 @@ on:
       - master
   pull_request:
 jobs:
+  ubuntuUnitWithJDK8And11:
+    name: ubuntu-latest tests
```
gabro: can you include the jdk version in the name?

```suggestion
    name: ubuntu-latest tests (${{  matrix.java-version }})
```

or something like this

isomarcte:
@gabro @olafurpg this should be good to go to go now. I've changed it to the ubuntu tests with JDK8 and JDK11, while the windows and macOS tests only use JDK11. 

I've also changed it to only use `"openjdk1.11"` so we'll only have one JDK11.

Finally, I switched to use the system property `"java.vm.specification.version"` as I did some reading around and that seems like the most technically correct value to use.

.github/workflows/ci.yml
isomarcte: yep

isomarcte:
@gabro can you take a look again. I update the other tests to use JDK 11 as well since we want that to be the global default.

.github/workflows/ci.yml
```diff
@@ -57,6 +85,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: olafurpg/setup-scala@v5
```
isomarcte: @gabro @olafurpg I noticed this is on v5 not v7. Is that intentional or did it just get missed?

gabro: just missed, I believe.

isomarcte: Updated

isomarcte:
@olafurpg @gabro tests are failing for the deprecated Scala versions on JDK > 8. This is due to the scala compiler's class file parser not having support for parsing some of the constant pool tags, in particular `CONSTANT_MODULE` on these old compiler versions.

What would you like to do? I'd personally be in favor of dropping them. I'm pretty sure metals won't be able to work in general when being used with JDK > 8 class files and Scala < 2.12.10.

isomarcte:
> I'd personally be in favor of dropping them.

Dropping them from the cross compilation tests that is.

olafurpg:
I'm fine with not testing JDK>8 on Scala <2.12.10

isomarcte:
@olafurpg done. This should be good to go now.

isomarcte:
@olafurpg the mill tests caused the JVM to core dump. I think they were working on earlier versions of this PR. Can you get those debug logs? I don't think I have access to get them...

olafurpg:
@isomarcte I'm unable to access the core dumps

isomarcte:
@olafurpg K. I see some non-JDK 11 compliant scala versions being used in the "slow" tests that I missed. Let me fix those and see if that changes things.

olafurpg:
FWIW, I recall also hitting on JVM crashes when running the tests locally on JDK 11

isomarcte:
@olafurpg I think I've got it all sorted out now.

* As for the JVM core dumps.
    * I think whatever build of `openjdk` that jabba is getting is wonky. I've changed it to `adopt` and things are fine. I noticed the version jabba is grabbing is _very slightly_ out of date, `11.0.2` vs. `11.0.5`.
* The remaining test failures seem to be related to timeouts talking to bloop. I don't think they are related to this PR.

isomarcte:
Looks like there are still issues with the LSP integration tests though. I'm looking into it.

isomarcte:
@olafurpg @gabro all tests are green.

isomarcte:
I rebased the commit and update the commit message. Hopefully all tests stay green.

.github/workflows/ci.yml
```diff
@@ -5,16 +5,36 @@ on:
       - master
   pull_request:
 jobs:
+  ubuntuUnitWithJDK8And11:
+    name: ubuntu-latest tests ${{ matrix.java-version }}
```
olafurpg: Instead of duplicating the unit tests for ubuntu, could we try to use `exclude` instead? 🤔 

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

```yml
    os: [macos, windows, ubuntu]
    java: [8, 11]
    exclude:
      - os: windows
        java: 8
      - os: macos
        java: 8
```

tests/slow/src/test/scala/tests/feature/WorksheetCrossLspSuite.scala
```diff
@@ -2,5 +2,4 @@ package tests.feature
 
 import scala.meta.internal.metals.{BuildInfo => V}
 
-object Worksheet211LspSuite extends tests.BaseWorksheetLspSuite(V.scala211)
```
olafurpg: Instead of removing these tests for all JVM versions, could we skip them on newer java versions?

tests/unit/src/main/scala/tests/JVMVersion.scala
```diff
@@ -0,0 +1,117 @@
...
+/** ADT used to model the current version of the runtime JVM.
+ *
+ * It is needed to know the current version for several of the test cases as
+ * there are some changes to the classpath between different JVM versions.
+ *
+ * For example, on JDK > 8 several standard jar files are now stored as java
+ * modules and will now longer show up as standalone jars,
+ * e.g. `charsets.jar`.
+ */
+sealed trait JVMVersion extends Product with Serializable {
```
olafurpg: Did you try using `scala.util.Properties.isJavaAtLeast`?
```scala
(Scala 2.12.9 Java 1.8.0_222)
@ scala.util.Properties.isJavaAtLeast("9")
res1: Boolean = false


(Scala 2.12.9 Java 11.0.5)
@ scala.util.Properties.isJavaAtLeast("11")
res2: Boolean = true
```

.github/workflows/ci.yml
isomarcte: Sure. Updated.

tests/slow/src/test/scala/tests/feature/WorksheetCrossLspSuite.scala
isomarcte: Yes. I've updated the code to discriminate on the JDK version and Scala versions.

tests/unit/src/main/scala/tests/JVMVersion.scala
isomarcte: I did not, I did not know about that. I've updated the code, removed the `JVMVersion` class and am now using that predicate.

isomarcte:
@olafurpg Updated again. Should be good to go now.

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
```diff
@@ -5,323 +5,335 @@ import scala.meta.internal.metals.UserConfiguration
...
-           |1.to(10).toVector
-           |res1: Vector[Int] = Vector(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
-           |val List(a, b) = List(42, 10)
-           |a: Int = 42
-           |b: Int = 10
-           |""".stripMargin
-      )
-    } yield ()
-  }
+  scalaVersions.foreach((scalaVersion: String) =>
```
olafurpg: Can we keep the logic simpler here? It’s possible to override “testAsync” and skip tests based on a condition. I’m not a fan of nesting tests this deep

tests/slow/src/test/scala/tests/feature/CompletionCrossLspSuite.scala
```diff
@@ -5,9 +5,6 @@ import tests.BaseCompletionLspSuite
 
 object CompletionCrossLspSuite
     extends BaseCompletionLspSuite("completion-cross") {
-  testAsync("basic-211") {
```
olafurpg: Can we reinstate this test?

olafurpg: It’s fine if not, but I’m curious to know why we can’t keep the 2.11 tests.

tests/slow/src/test/scala/tests/feature/DefinitionCrossLspSuite.scala
```diff
@@ -7,10 +7,6 @@ import scala.concurrent.Future
 object DefinitionCrossLspSuite
     extends BaseCompletionLspSuite("definition-cross") {
 
-  testAsync("2.11") {
```
olafurpg: Ditto

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
```diff
@@ -5,323 +5,335 @@ import scala.meta.internal.metals.UserConfiguration
...
+    ScalaVersion
+      .fromString(scalaVersion)
+      .fold(
+        (e: ScalaVersion.ScalaVersionParseError) => fail(e.errorMessage), {
+          (sv: ScalaVersion) =>
+            if (ScalaVersion.validScalaVersionForCurrentJDK(sv)) {
```
olafurpg: Would be nice to avoid adding a custom “ScalaVersion” class with its own parsing logic just for this condition. Could we solve this problem instead with a hardcoded set of Scala version strings that don’t work in Java 11? 🤔

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
```diff
@@ -5,323 +5,335 @@ import scala.meta.internal.metals.UserConfiguration
 import scala.meta.internal.metals.MetalsSlowTaskResult
 import scala.concurrent.Promise
 
-abstract class BaseWorksheetLspSuite(scalaVersion: String)
+abstract class BaseWorksheetLspSuite(scalaVersions: Set[String])
```
olafurpg: For consistency with the rest of the codebase, I would prefer to keep the old signature and duplicate instead subclasses.

tests/slow/src/test/scala/tests/feature/CompletionCrossLspSuite.scala
isomarcte: @olafurpg I initially removed these because after this PR we will not be running the cross Scala tests on JDK 8. The first version of Scala that supports class file parsing of JDK > 8 bytecode is 2.12.10. Notably, there is no 2.11.x version which supports this AFAIK. 

I can add them back in, and then we can either.

* Discriminate on the JDK version and not run the tests with 2.11.x if we are > 8
* Discriminate on the JDK version and even if it is > 8 only have it emit 8 compatible bytecode.

I'm on the fence about the utility of this myself.

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
isomarcte: I'll revert that.

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
isomarcte: We can do what you suggest, or we can use more simple parsing logic. It is my opinion both are a bit more brittle and difficult to debug in the case of typos in the input. 

I'll throw up two other branches so we can discuss, but I'll defer to what ever you'd like.

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
isomarcte: Certainly.

tests/slow/src/test/scala/tests/feature/CompletionCrossLspSuite.scala
isomarcte: Nevermind, I realized just now that we get the discrimination for free as part of the `BaseCompletionLspSuite` changes. Sorry, I'm still learning my way around the code base.

isomarcte:
@olafurpg I've update the PR to address your requested changes except that I've left `ScalaVersion` in place for now. 

I'll create two more branches to compare to pros/cons of different discrimination strategies on the scala version string. I'm going to be busy for the next 24 hours, so I may not have them up for a day or more.

tests/mtest/src/main/scala/tests/ScalaVersion.scala
```diff
@@ -0,0 +1,99 @@
...
+ * We need to skip some tests if we are not on a very particular scala
+ * version. This class enables us to discriminate on that.
+ *
+ * It extends `Ordered` so that we can do stuff like,
+ * {{{
+ * scala> ScalaVersion(2L, 12L, 10L, "") < ScalaVersion(2L, 12L, 9L, "")
+ * res0: Boolean = false
+ * }}}
+ */
+final case class ScalaVersion(
```
olafurpg: This class looks like a duplicate of `BuildTool.isCompatibleVersion` https://github.com/scalameta/metals/blob/abca8703638fe2768e0058288ae0e3decbe42e07/metals/src/main/scala/scala/meta/internal/builds/BuildTool.scala#L51

We could add a new package `scala.meta.internal.semver` with a basic helper to compare version numbers. Down the road we can hopefully replace that package with a proper library implementation like Coursier

tests/mtest/src/main/scala/tests/BaseSuite.scala
```diff
@@ -37,6 +37,21 @@ class BaseSuite extends TestSuite {
...
   def intercept[T: ClassTag](exprs: Unit): T = macro Asserts.interceptProxy[T]
+  def isValidScalaVersionForEnv(scalaVersion: String): Boolean =
+    ScalaVersion
+      .fromString(scalaVersion)
+      .fold(
+        (e: ScalaVersion.ScalaVersionParseError) => this.fail(e.errorMessage),
+        (sv: ScalaVersion) => ScalaVersion.validScalaVersionForCurrentJDK(sv)
+      )
+
+  def guardOnScalaVersion(scalaVersion: String)(f: => Unit): Unit =
```
olafurpg: nit: let's remove this helper and use `if (isValidScalaVersion(..)` instead

tests/mtest/src/main/scala/tests/ScalaVersion.scala
isomarcte: Do you want me to move `isCompatibleVersion` from `BulidTool` into the `scala.meta.internal.semver` package you propose? Or do you want me to add a dependency on `metals` in the `mtest` project?

tests/mtest/src/main/scala/tests/BaseSuite.scala
isomarcte: Once we decide on what to do here (https://github.com/scalameta/metals/pull/1171#discussion_r357985278) I'll remove this.

tests/mtest/src/main/scala/tests/ScalaVersion.scala
olafurpg: Let's move the `isCompatibleVersion` method into a `internal.semver` package inside the `mtags` project so that it can be used from mtest

isomarcte: Updated.

tests/mtest/src/main/scala/tests/BaseSuite.scala
isomarcte: Updated.

isomarcte:
@olafurpg should be ready for review again.

isomarcte:
Oops. Looks like I left some unused imports. Should be fixed now. 

Also, FWIW, I removed the `Try` wrapper in the body of `isCompatibleVersion`. Unless I'm _really_ missing something, those inequality checks will never throw.

PR #1174, opened 2019-12-09
Sort auto-completed exhaustive match'es by declaration order

tanishiking:
https://github.com/scalameta/metals-feature-requests/issues/49

This PR enables metals to sort auto-completed exhaustive match cases by declaration order, instead of the order of `Symbol#knownDirectSubclasses`

- if symbols of subclasses' position information are not available, load all the symbols of the file that contains the definition of the symbol (in declaration order), and sort subclasses by those symbols.
- Otherwise, sort cases based on the position information without loading symbols.
  - This case happens if the definition of `abstract sealed class ...` is located in the same file with the target `match ...` statement.
  - In that case, we can sort subclasses quite cheaply.

---

Here's a test for match exhaustive completion of `scala.meta.internal.docstrings.Block`
https://github.com/scalameta/metals/blob/7d33c4b1d13cbc5ad5bb19de807c241525969603/mtags/src/main/scala/scala/meta/internal/docstrings/Body.scala#L45-L59

match

---

Thank you for giving a kind instruction for implementing this @olafurpg !  It's insightful and made me more comfortable to read through the metals' codebase :) https://github.com/scalameta/metals-feature-requests/issues/49#issuecomment-544236184


mtags-interfaces/src/main/java/scala/meta/pc/VirtualFile.java
```diff
@@ -0,0 +1,18 @@
...
+    /**
+     * The path name of the virtual file.
+     */
+    String path();
+
+    /**
+     * The content of the virtual file.
+     */
+    String value();
+}
```
tanishiking: Java version of `Input.VirtualFile` for defining a method that returns `Input.VirtualFile` equivalent information in `SymbolSearch.java`.

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -103,6 +103,33 @@ object CompletionMatchSuite extends BaseCompletionSuite {
     filter = _.contains("exhaustive")
   )
 
+  checkEdit(
+    "Sort auto-completed exhaustive match keywords by declaration order",
```
gabro: we generally use short meaningful names for the tests (e.g. `exhaustive-sorting`), so that they can be used as targets for `testOnly`.

tests/mtest/src/main/scala/tests/TestingSymbolSearch.scala
```diff
@@ -46,6 +48,23 @@ class TestingSymbolSearch(
     }
   }
 
+  override def definitionSource(symbol: String): ju.Optional[VirtualFile] = {
+    index.definition(Symbol(symbol)) match {
+      case None =>
+        ju.Optional.empty()
+      case Some(value) =>
+        import java.nio.file.Files
```
gabro: nit: move this in the toplevel import block, for consistency

mtags/src/main/scala/scala/meta/internal/metals/MetalsVirtualFile.scala
```diff
@@ -0,0 +1,16 @@
+package scala.meta.internal.metals
+
+import scala.meta.pc.VirtualFile
+import scala.meta.inputs.Input
+
+case class MetalsVirtualFile(
```
gabro: I believe we use the `Impl` suffix for similar instances, so this would be `VirtualFileImpl`, for consistency.

mtags/src/main/scala/scala/meta/internal/metals/MetalsVirtualFile.scala
```diff
@@ -0,0 +1,16 @@
...
+    path: String,
+    value: String
+) extends VirtualFile
+object MetalsVirtualFile {
```
gabro: nit: add a newline in between class and object

tanishiking:
Thank you for taking a look at this @gabro ! I just added some commits based on your feedback :)

mtags-interfaces/src/main/java/scala/meta/pc/SymbolSearch.java
```diff
@@ -19,6 +19,11 @@
      */
     List<Location> definition(String symbol);
 
+    /**
+     * Returns the file where the symbol is defined, if any.
+     */
+    Optional<VirtualFile> definitionSource(String symbol);
```
olafurpg: I'm wondering if we should change this API to directly return the toplevels 🤔 
```suggestion
    List<String> definitionSourceToplevels(String symbol);
```
This would allow the Metals server to potentially cache the results between different invocations. The `Mtags.toplevels` method is on the other hand very fast (~900k lines/sec) I have a feeling that a lot of `match (exhaustive)` completions are on the same types (for example `scala/Option#`).

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -103,6 +103,33 @@ object CompletionMatchSuite extends BaseCompletionSuite {
     filter = _.contains("exhaustive")
   )
 
+  checkEdit(
+    "exhaustive-sorting",
```
olafurpg: Can we add another cross test case for 3rd party defined symbols? Cross tests have the ability to index library dependencies, see `indexScalaLibrary()`

tests/unit/src/main/scala/tests/BaseCompletionLspSuite.scala
```diff
@@ -125,4 +160,60 @@ abstract class BaseCompletionLspSuite(name: String) extends BaseLspSuite(name) {
       )
     } yield ()
   }
+
+  def matchKeywordTest(scalaVersion: String): Future[Unit] = {
```
olafurpg: 💯

mtags-interfaces/src/main/java/scala/meta/pc/SymbolSearch.java
tanishiking: > I'm wondering if we should change this API to directly return the toplevels 🤔

Sure! That makes more sense, and now we can remove the java version of `VirtualFile` and `VirtualFileImpl` 😄

tanishiking: Thank you so much for your input @olafurpg !

I just added a straightforward implementation of the cache for `definitionSourceToplevels` in https://github.com/scalameta/metals/pull/1174/commits/a70886e97250f94890b922cb89662fb911f3eaf5

Though it definitely makes sense to introduce cache for `definitionSourceToplevels` :+1:, I have one concern about the cache invalidation:
In https://github.com/scalameta/metals/pull/1174/commits/a70886e97250f94890b922cb896 the cache in `MetalsSymbolSearch` will be cleared only when the [workspace is updated](https://github.com/scalameta/metals/blob/2475a952a4eaf7f3b5e41246e091c3757f244908/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L390-L407).
If we add some new subclasses to a `sealed abstract class X`, metals server would be unable to reflect the changes until the workspace is updated.

---

To avoid this, it might be better to do:

- cache `definitionSourceToplevels` result by the symbol of definition's (absolute) path
  - `private val cache = new TrieMap[AbsolutePath, ju.List[String]]()`
- Everytime the project is compiled, clear the cache for `definitionSourceToplevels` at [Compilers#didCompile](https://github.com/scalameta/metals/blob/2475a952a4eaf7f3b5e41246e091c3757f244908/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala#L109-L122) before restart the PresentationCompiler instance.
  - so the metals server can reflect the change to the completion items.
  - to do this, we have to add cache cleanup method to `SymbolSearch`
- The cache items we have to clear are the items whose key (`AbsolutePath`) is located under the project's workspace.
  - We don't need to clear the symbol cache from the 3rd party definitions, because they won't change without updating the workspace.

What do you think about this?

olafurpg: Cache invalidation is a good question. Can we maybe reuse the toplevels indexing for WorkspaceSymbolProvider? That index is already updated on file save.

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -103,6 +107,63 @@ object CompletionMatchSuite extends BaseCompletionSuite {
...
+      """package sort
+        |object App {
+        |  Option(1) matc@@
+        |}
+        |""".stripMargin,
+      if (!isScala211)
+        """package sort
+          |object App {
+          |  Option(1) match {
+          |\tcase Some(value) => $0
```
olafurpg: Yay, really excited to finally get the `Some` case at the top

olafurpg:
Feel free to ignore the unrelated test failure, I'm hoping #1186 fixes that

metals/src/main/scala/scala/meta/internal/metals/MetalsSymbolSearch.scala
```diff
@@ -16,13 +18,27 @@ class MetalsSymbolSearch(
     wsp: WorkspaceSymbolProvider,
     defn: DefinitionProvider
 ) extends SymbolSearch {
+  // cache for definitionSourceTopLevels
+  private val cache = new TrieMap[String, ju.List[String]]()
```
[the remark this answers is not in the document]
tanishiking: done in https://github.com/scalameta/metals/pull/1174/commits/a68027a5e8832ac885c8515cb22cc9b7cb33593c and now the completion seems much faster (especially for workspace sources)!

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1297,7 +1297,39 @@ trait Completions { this: MetalsGlobal =>
...
+          } else {
+            // Read all the symbols in the source that contains
+            // the definition of the symbol in declaration order
+            val defnSymbols = search
+              .definitionSourceToplevels(semanticdbSymbol(tpe.typeSymbol))
+              .asScala
+            if (defnSymbols.length > 0)
+              subclassesResult
+                .sortBy(sym => {
+                  defnSymbols.indexOf(semanticdbSymbol(sym))
```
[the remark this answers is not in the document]
tanishiking: Thanks! 👍  https://github.com/scalameta/metals/pull/1174/commits/707a24f40d27f1bb528fa719742b417e89a06b06

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1597,6 +1599,7 @@ class MetalsLanguageServer(
       semanticDBIndexer.reset()
       treeView.reset()
       worksheetProvider.reset()
+      symbolSearch.reset()
```
tanishiking: Clear symbolSearch cache (for dependency sources) on indexing workspace

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -103,6 +107,63 @@ object CompletionMatchSuite extends BaseCompletionSuite {
...
+          |\tcase None =>
+          |}
+          |}
+          |""".stripMargin
+      else
```
tgodzik: I don't think we need the else here? `!isScala211` is already tested at the top

tanishiking: Oops, we don't need it. I meant like this https://github.com/scalameta/metals/pull/1174/commits/0d5f96fb977b6c73c44b8b5c061480393ecb1c82, thank you for pointing it out :)

PR #1176, opened 2019-12-09
Minor tweaks to CI jobs names

gabro:
Just some proposed changes for (I think) clarity.

We may also take the chance to rename the `slow` project into `lsp` or `integration`. Thoughts?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
 jobs:
   unit:
-    name: ${{ matrix.os }} tests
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -76,7 +76,7 @@ jobs:
       - name: Run cross tests
         run: sbt +cross/test
   checks:
-    name: Scalafmt/Scalacheck/Docs
```
tgodzik: :smile: Somehow I never realized what I wrote here

PR #1177, opened 2019-12-09
Ignore flaky BspSwitchSuite and reenable FoldingRangeSuite on Windows.

tgodzik:
BspSwitchSuite is consistently failing on CI for Windows, but FoldingRangeSuite might be reenabled after recent fixes to the reliability of the directory watcher within tests.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,6 +1,6 @@
 -Xss4m
 -Xms1G
--Xmx4G
+-Xmx2G
 -XX:ReservedCodeCacheSize=1024m
 -XX:+TieredCompilation
 -XX:+CMSClassUnloadingEnabled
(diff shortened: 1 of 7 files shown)


tgodzik:
Now they're going OOM :/

gabro:
Could it be the -Xmx4G option which is a bit much?

olafurpg:
Yes, I think the `-Xmx4g` option may be too much, iirc the GitHub Actions machines only have 2gb

tgodzik:
> Yes, I think the `-Xmx4g` option may be too much, iirc the GitHub Actions machines only have 2gb

Hmm... according to https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners they have 7gb. But it might be Bloop server taking up some of it.

gabro:
✅ all green!

PR #1181, opened 2019-12-10
Improve Pants to Bloop export step.

olafurpg:
Previously, the Pants to Bloop export step was implemented with fairly
tricky code that was difficult to reason about. This commit cleans up
the implementation so that it's more correct, easier to maintain and it
also runs faster.

PR #1183, opened 2019-12-10
Ignore pants ivy sources without a default scope

lbordowitz:
In my build dependencies, I have:

```json
{
    "libraries": {
            "io.netty:netty-transport-native-epoll:4.1.38.Final": {
            "linux-x86_64": "/home/lbordowitz/.ivy2/pants/io.netty/netty-transport-native-epoll/jars/netty-transport-native-epoll-4.1.38.Final-linux-x86_64.jar"
        },
    },
}
```

There's no `default` key provided, and metals fails every time I try to import it. This fix causes the import to ignore those entries that don't have a "default-scoped" target.


olafurpg:
Thank you for this contribution! I believe this issue is fixed by #1181 (along with a host of other issues)

PR #1194, opened 2019-12-16
Update The Getting Started Documentation For Java 11

isomarcte:


isomarcte:
@gabro not really. As of the merging of #1171 CI is building this with JDK 11. Not all code which builds with JDK 8 will build with JDK 11. That was the impetus for 

* #1171 
* #1148 
* #1139 

If people are using JDK 8 they may think everything is fine, but then it will fail in CI. 

So I think we should encourage them to only dev on `metals` with JDK 11 or >.

isomarcte:
That if metals builds on JDK 11 it should build JDK 8, but not the other way around.

olafurpg:
@isomarcte i still develop Metals on JDK 8 and edo day to continue to do so for at least the next year. Our contributing docs should say that either JDK 8 or 11 is fine.

olafurpg:
> That if metals builds on JDK 11 it should build JDK 8, but not the other way around.

The same is true for contributors who work on JDK 11 and use APIs that are not available on JDK 8

isomarcte:
> The same is true for contributors who work on JDK 11 and use APIs that are not available on JDK 8  

That's a fair point. I'll update it.

isomarcte:
Updated

PR #1197, opened 2019-12-17
add in documenation for coc-metals

ckipp01:
Hopefully, this will help with the review of `coc-metals`. I did my best to update all of the documentation and include images of all the new things that `coc-metals` offers vim users.

I think it's at a point where it's very stable for downloading new versions, managing configuration options, and using commands to restart the server etc. I've found the experience much more pleasant than managing it by hand. I think it'd be cool to have this included in the next release.

I think that the extension should probably be looked at by someone on the team first to make sure it's sane, but everything seems to be working alright, and the code is very similar to that of the vs code extension. In the future, we may be able to extract out a common module to publish and share as well between the two of them.

[Package is here](https://www.npmjs.com/package/coc-metals)
[Repo is here](https://github.com/ckipp01/coc-metals)

EDIT: Removed the draft status since I think this is ready to go 👍

### DIFF
--- a/docs/contributors/updating-website.md
+++ b/docs/contributors/updating-website.md
@@ -17,10 +17,8 @@ For running the website locally, you'll need:
 
 In addition to Docusaurus, we preprocess the markdown files using:
 
-- [sbt-docusaurus](https://olafurpg.github.io/sbt-docusaurus/), to publish the
-  Docusaurus page to GitHub from the sbt shell.
-- [mdoc](https://github.com/scalameta/mdoc), to type-check and interpret Scala
-  code fences.
+- [mdoc](https://github.com/scalameta/mdoc), to type-check, interpret Scala
+  code fences, and to generate the site using it's [built-in Docusaurus support](https://scalameta.org/mdoc/docs/docusaurus.html).
 
 The first step is to then preprocess the markdown files. You can do it with:
 
--- a/docs/contributors/updating-website.md
+++ b/docs/contributors/updating-website.md
@@ -18,7 +18,7 @@ For running the website locally, you'll need:
 In addition to Docusaurus, we preprocess the markdown files using:
 
 - [mdoc](https://github.com/scalameta/mdoc), to type-check, interpret Scala
-  code fences, and to generate the site using it's [built-in Docusaurus support](https://scalameta.org/mdoc/docs/docusaurus.html).
+  code fences, and to generate the site using its [built-in Docusaurus support](https://scalameta.org/mdoc/docs/docusaurus.html).
 
 The first step is to then preprocess the markdown files. You can do it with:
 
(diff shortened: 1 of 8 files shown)


docs/contributors/updating-website.md
```diff
@@ -17,10 +17,8 @@ For running the website locally, you'll need:
 
 In addition to Docusaurus, we preprocess the markdown files using:
 
-- [sbt-docusaurus](https://olafurpg.github.io/sbt-docusaurus/), to publish the
-  Docusaurus page to GitHub from the sbt shell.
-- [mdoc](https://github.com/scalameta/mdoc), to type-check and interpret Scala
-  code fences.
+- [mdoc](https://github.com/scalameta/mdoc), to type-check, interpret Scala
```
olafurpg: 👍

docs/editors/vim.md
```diff
@@ -34,13 +22,12 @@ sure you have the correct version installed.
 ```sh
 # If using Vim
 vim --version | head
-VIM - Vi IMproved 8.1
```
olafurpg: Should we recommend nvim over vim?

docs/editors/vim.md
```diff
@@ -245,70 +220,113 @@ Or use the default recommended mapping `<space> a`.
...
-![Run Doctor](https://i.imgur.com/yelm0jd.png)
+## Available Configuration Options
+
+The following configuration options are currently available. The easiest way to set these configurations is to enter `:CocConfig` or `:CocLocalConfig` to set your global or local configuration settings respectively.
+If you'd like to get autocompletion help for the configuration values you can install [coc-json](https://github.com/neoclide/coc-json).
+
+
+   Configuration Option     |      Description
+----------------------------|---------------------------
+`metals.serverVersion`      | The version of the Metals server artifact. Requires reloading the window.
```
olafurpg: This table can be automatically generated using mdoc similarly as it's done in `new-editor.md` https://github.com/scalameta/metals/blame/b1cc89c7b6bd081e772057b7bbb45377cc705796/docs/editors/new-editor.md#L312

It would be nice to avoid duplicating the descriptions here

docs/editors/vim.md
ckipp01: 8.2 is working pretty well right now. All of the same core functionality needed for a pleasant LSP experience is there, albiet I've found some of it to be not as smooth. Either way, coc.nvim promises to work with both, so I'd say for now, we can leave both in.

docs/editors/vim.md
ckipp01: Ahh, good call. I'll update this to use mdoc.

docs/contributors/updating-website.md
```diff
@@ -17,10 +17,8 @@ For running the website locally, you'll need:
...
+  code fences, and to generate the site using it's [built-in Docusaurus support](https://scalameta.org/mdoc/docs/docusaurus.html).
```
gabro:
```suggestion
  code fences, and to generate the site using its [built-in Docusaurus support](https://scalameta.org/mdoc/docs/docusaurus.html).
```

ckipp01:
> I would also go through the same page and make sure that the table is up-to-date in general (e.g. there are a couple of notes about Vim that I'm not sure they are try anymore)

Most of these are up to date still, but as some more things get added, i'll continue to check back here and update.

PR #1198, opened 2019-12-17
Several improvements to Pants export step

olafurpg:
See individual commits for details. In summary:

* faster `bloop test` for ScalaTest targets
* increased compatibility with older versions of Pants
* improved UX for Pants targets that use recursive `**/*.scala` globs

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/pantsbuild/PantsGlobs.scala
+++ b/metals/src/main/scala/scala/meta/internal/pantsbuild/PantsGlobs.scala
@@ -1,0 +1,60 @@
+package scala.meta.internal.pantsbuild
+
+import java.nio.file.Path
+import ujson.Value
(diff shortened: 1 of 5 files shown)


metals/src/main/scala/scala/meta/internal/pantsbuild/PantsGlobs.scala
```diff
@@ -0,0 +1,60 @@
...
+        PantsGlobs(include, exclude)
+      case _ =>
+        PantsGlobs(Nil, Nil)
+    }
+  }
+
+  private def globsFromObject(value: Value): List[String] = value match {
+    case Obj(obj) =>
+      obj.get("globs") match {
+        case Some(arr: Arr) =>
```
wiwa: why not `map` for these case matches?

PR #1199, opened 2019-12-18
Fetch classes after taskFinish

### LINKED ISSUE #1189: Fetch main and test classes as soon as possible
Currently, we have to wait until bloop finished the entire compilation before we can ask for a main or test classes in a build target, even if this build target was already compiled (e.g. it was a dependency module being compiled first). This is caused by some nondeterministic behavior in bloop fixed via https://github.com/scalacenter/bloop/issues/1067.

We should no longer wait for the whole compilation to speed up the code lens retrieval times to enhance the user experience.

marek1840:
Previously, we had to wait for the whole compilation to finish, before fetching the classes for a build target.
Now, those classes are fetched as soon as the build target gets compiled successfully. This results in a much more smooth experience for the user.

Closes #1189


tgodzik:
This is currently not yet fixed, the issue in Bloop has been reopened.

Issue #1010, opened 2019-10-22
Add support for breakpoint debugging


marek1840:
Adding supporting for breakpoints consists of two parts:
1. Adapting the `setBreakpoints` request sent to the server
2. Resolving the source path for `stackTrace` responses received from the server

### Adapting `setBreakpoints` request
Bloop [cannot resolve FQCN for third-party sources](https://github.com/scalacenter/bloop/issues/1057), so metals needs to map the breakpoints for a given file `/foo/bar/Baz.scala` to the fully qualified name of the class they are specified in (e.g. `foo.bar.Baz$` for objects). These new paths must then be sent to Bloop.
One caveat here is that a single file can contain multiple classes, so we must partition the client request into possibly multiple synthetic requests and then gather the responses into one before sending them to the client.

The source URI for the requests [expected by bloop](https://github.com/scalacenter/bloop/pull/1087) is `dap-fqcn:$FULLY_QUALIFIED_CLASS_NAME`

### Adapting `stackTrace` response
The stack frames returned from bloop are of the form: 
` { name: $SOURCE_FILE_NAME, path: $PACKAGE_PREFIXED_SOURCE_FILE_NAME }`
This does not match the real source paths expected by the client, so we need to adapt those.

To resolve the real source path we must either: 
1. Find the symbol, using the package-prefixed source file name. Since we know the packages, we can derive a symbols possibly matching the expected type. Note, that this will fail for types not matching the source file name.
2. Find the source file for the given `buildTarget` using the source file name. 

The second approach should be used as a complement to the first one to recover from types not matching their source file name.

jvican:
Excellent ticket description, I had missed the fact that `SetBreakpointsRequest` holds a sequence of breakpoints in my discussion at https://github.com/scalacenter/bloop/issues/1057. The proposed implementation looks good to me :+1:

jvican:
An important consideration when implementing this is that clients only send one unique `setBreakpointRequests` per URI. This is really important because if there is more than one the semantics of DAP establish that only the last processed `setBreakpointRequests` takes effect, e.g. the previous ones will be ignored.
  
I recommend to pass to Bloop only those fully qualified class names belonging to third-party sources and not passing them for user-defined Scala sources, which Bloop supports out-of-the-box via a reliable implementation.

tgodzik:
This is already done by @marek1840

PR #1200, opened 2019-12-18
Add separate "debug" code lens

### LINKED ISSUE #1188: Add separate "debug" code lens
After #1010, breakpoints will always be sent to the JVM. We want to add a way to launch an application without any breakpoints.

This should be done by adding separate code lens: `debug` and (possibly `debug test` for tests?)

marek1840:
Closes #1188 
Previously, there was no way to run program and ignore the breakpoints.
Now, we introduce a separate command `StartRunSession` which ignores the breakpoint requests altogether.
Both commands (`StartRunSession` and `StartDebugSession`) are shown to the user as separate code lenses.

TODO: rebase after #1010 gets merged

### DIFF
--- a/tests/unit/src/main/scala/tests/CodeLensesTextEdits.scala
+++ b/tests/unit/src/main/scala/tests/CodeLensesTextEdits.scala
@@ -3,18 +3,21 @@ package tests
 import org.eclipse.{lsp4j => l}
 
 object CodeLensesTextEdits {
-  def apply(ranges: Seq[l.CodeLens]): List[l.TextEdit] = {
-    ranges.map(convert).toList
+  def apply(lenses: Seq[l.CodeLens]): Iterable[l.TextEdit] = {
+    for {
+      (line, lenses) <- lenses.groupBy(_.getRange.getStart.getLine)
+      textEdit = convert(line, lenses)
+    } yield textEdit
   }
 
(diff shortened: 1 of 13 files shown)


tests/unit/src/main/scala/tests/CodeLensesTextEdits.scala
```diff
@@ -3,18 +3,21 @@ package tests
...
   }
 
-  private def convert(lens: l.CodeLens): l.TextEdit = {
-    val range = lens.getRange
-    val pos = new l.Position(range.getStart.getLine, 0)
-    new l.TextEdit(new l.Range(pos, pos), print(lens))
+  private def convert(line: Int, lenses: Seq[l.CodeLens]): l.TextEdit = {
+    val pos = new l.Position(line, 0)
+    val edit = lenses.map(print).mkString
+    new l.TextEdit(new l.Range(pos, pos), edit + "\n")
```
tgodzik: Will it not be a problem on windows by any chance?

marek1840: Changed to `System.lineSeparator`

marek1840:
The failure seems unrelated:
```
[error] Failed tests:
[error] 	tests.worksheets.WorksheetLspSuite
```

PR #1208, opened 2019-12-19
Send SetBreakpointRequest directly to bloop for workspace sources

marek1840:


metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathProvider.scala
```diff
@@ -12,6 +12,10 @@ final class SourcePathProvider(
     buildTargets: BuildTargets,
     targets: List[BuildTargetIdentifier]
 ) {
+  def isWorkspacePath(path: AbsolutePath): Boolean = {
```
olafurpg: Should we use this instead?
https://github.com/scalameta/metals/blob/59ca771c9d1c1f3ca3d34df0be84215bc86e4a40/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala#L257

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProxy.scala
```diff
@@ -60,9 +61,27 @@ private[debug] final class DebugProxy(
...
+      val path = adapters.adaptPath(args.getSource.getPath).toAbsolutePath
+      if (sourcePathProvider.isWorkspacePath(path)) {
+        server
+          .sendPartitioned(List(request))
+          .map(_.map(DebugProtocol.parseResponse[SetBreakpointsResponse]))
+          .map(_.head.get)
+          .map { response =>
+            val breakpoints = response.getBreakpoints
+            if (breakpoints != null) {
+              breakpoints.foreach(_.setSource(args.getSource))
```
olafurpg: What are we doing here?

marek1840: java-debug is not setting the sources for breakpoints, so we have to do it manually :(

marek1840:
It seems that the [bloop breakpoint support](https://github.com/scalacenter/bloop/pull/1087) doesn't sit well with our suite of tests - it seems the jvm is not stopping at the breakpoints we specify. @jvican , could you help me understand why?

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProxy.scala
jvican: Hmm, why is `java-debug` setting the sources for breakpoints in the former case and not now?

marek1840: By former case you mean the library sources? I have to set them there as well: https://github.com/scalameta/metals/blob/master/metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala#L50

marek1840:
> How are you partitioning the requests? 

I am not - I send a "partitioning" of a single request (https://github.com/scalameta/metals/pull/1208/files#diff-734f6ad19137cce99449cbe2221fb139R68). I do so, because I still need to set the sources and no other method in the current implementation returns a `Future[Response]`. The `server.send` method is not returning a response.

Of course, once we get this working, the implementation will have to be tidied up, but for now it is sufficient workaround of a known limitation.

jvican:
I'm not sure I fully understand your answer but are you then not following this contract " The contract java-debug has to set breakpoints correctly is that it needs to receive all breakpoints from one file together. "? The test case I linked should give you a hint of why it doesn't work, as that tests multi-breakpoints in both Scala and Java sources.

marek1840:
> are you then not following this contract (...)

I am, I am just passing the request to Bloop as I get it. Only one request goes to bloop (for a single source, that is).

marek1840:
As for the offline question:

> Where is the logic to read LineNumberTable in Metals?

Which, I believe was caused by me being unclear in this comment: https://github.com/scalameta/metals/pull/1163#discussion_r359731298

We are not checking this table ourselves. What we do is to resolve the fully qualified name using the source file [here](https://github.com/scalameta/metals/blob/master/metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala#L23-L33) and [then use it](https://github.com/scalameta/metals/blob/master/metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala#L64-L75) as a `dap-fqcn` uri.

The silent assumption here is that the compiler populates the `LineNumberTable` correctly -  I should have probably phrased that more clearly.

jvican:
> The silent assumption here is that the compiler populates the LineNumberTable correctly - I should have probably phrased that more clearly.
  
FTR, I don't think neither the Scala nor Java compiler is using the `LineNumberTable` class file annotation.

marek1840:
> I don't think neither the Scala nor Java compiler is using the LineNumberTable class file annotation.

Didn't know that, thanks. 
Anyway, to be more precise: we assume the jvm knows how to set the breakpoint for a specific file at line L (be it with the help of a LineNumberTable or via any other mechanism). 
My implementation assumes that lines will be resolved correctly and therefore it focuses only on resolving the fully qualified class name using the sources we have.

jvican:
> therefore it focuses only on resolving the fully qualified class name using the sources we have.

Yes, that's my understanding. See my reply in the other PR for why we should avoid doing that for workspace sources.

tgodzik:
Closing after discussing @marek1840 - this approach is currently not working.

jvican:
This will be fixed in 1.4.0-RC2. I’ll comment here when the release is out.

Issue #1142, opened 2019-12-04
Use `BLOOP_EXPORT_CLASSIFIER_JARS` when importing build


olafurpg:
When upgrading to the next version of Bloop, we should use the new environment variable added in https://github.com/scalacenter/bloop/pull/1113 instead of `set bloopExportClassifierJars in Global = Some(Set("sources"))`

PR #1212, opened 2019-12-20
Use bloop.export-jar-classifiers property

joriscode:
This is my first contribution attempt, let me know if this corresponds to the issue (#1142).

At the moment, the only test I've carried on is using my local version:
- with VSCode (`publishLocal`)
- compiling a `build.sbt` file with `javacOptions.in(Global) ++= Seq("-Dbloop.export-jar-classifiers=\"sources\"")`

### DIFF
--- a/docs/build-tools/sbt.md
+++ b/docs/build-tools/sbt.md
@@ -49,10 +49,10 @@ addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "@BLOOP_VERSION@")
 Next, run:
 
 ```
-sbt "set bloopExportJarClassifiers := Some(Set(\"source\"))" bloopInstall
+sbt -Dbloop.export-jar-classifiers=sources bloopInstall
 ```
 
-to generate the Bloop JSON configuration files. You can also set the 
+to generate the Bloop JSON configuration files. You can also set the
 `bloopExportJarClassifiers` setting inside your main build.sbt file, but using
 the above command will do it automatically for you in the current sbt session.
 
--- a/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
@@ -25,7 +25,7 @@ case class SbtBuildTool(
 
   override def args(workspace: AbsolutePath): List[String] = {
     val sbtArgs = List[String](
-      "set bloopExportJarClassifiers in Global := Some(Set(\"sources\"))",
+      "-Dbloop.export-jar-classifiers=sources",
       "bloopInstall"
     )
     val allArgs = userConfig().sbtScript match {


docs/build-tools/sbt.md
```diff
@@ -49,11 +49,11 @@ addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "@BLOOP_VERSION@")
 Next, run:
 
 ```
-sbt "set bloopExportJarClassifiers := Some(Set(\"source\"))" bloopInstall
+sbt -Dbloop.export-jar-classifiers="sources" bloopInstall
```
joriscode: "source" was singular here, I think it was a typo.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -25,7 +25,7 @@ case class SbtBuildTool(
 
   override def args(workspace: AbsolutePath): List[String] = {
     val sbtArgs = List[String](
-      "set bloopExportJarClassifiers in Global := Some(Set(\"sources\"))",
+      "-Dbloop.export-jar-classifiers=\"sources\"",
```
joriscode: Do we want the extra quote signs `\"`?

docs/build-tools/sbt.md
```diff
@@ -95,7 +95,7 @@ libraries.
 // One of:
 //   ~/.sbt/0.13/build.sbt
 //   ~/.sbt/1.0/build.sbt
-bloopExportJarClassifiers.in(Global) := Some(Set("sources"))
```
tgodzik: This is still the preferred way. The java property is just for the situation where we want to run from command line.

docs/build-tools/sbt.md
tgodzik: that was a typo that I previously lost 2 hours because of :sweat_smile:

docs/build-tools/sbt.md
```diff
@@ -49,11 +49,11 @@ addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "@BLOOP_VERSION@")
...
 
 ```
-sbt "set bloopExportJarClassifiers := Some(Set(\"source\"))" bloopInstall
+sbt -Dbloop.export-jar-classifiers="sources" bloopInstall
 ```
 
-to generate the Bloop JSON configuration files. You can also set the 
-`bloopExportJarClassifiers` setting inside your main build.sbt file, but using
+to generate the Bloop JSON configuration files. You can also set the property
+`bloop.export-jar-classifiers` inside your main build.sbt file, but using
```
tgodzik: Let's keep it as is. It's better to add the setting inside of the `build.sbt` than each time when running

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
tgodzik: It seems to not work in the Sbt integration tests currently. Might be because of of the additional quotes?

joriscode: Happy new year!
I'm back from holidays and going to work on this as soon as possible!

joriscode:
@tgodzik I could not reproduce the errors locally... but it appears that they are gone with `0.8`, except `ubuntu-latest jdk-11 unit tests`.

PR #1214, opened 2019-12-20
Support breakpoints in java static methods and nested classes

marek1840:
Previously, setting breakpoints in java static methods or nested/anonymous classes didn't work.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
@@ -8,6 +8,8 @@ import scala.concurrent.Future
 import scala.meta.internal.metals.JvmSignatures
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.mtags.Mtags
+import scala.meta.internal.semanticdb.Language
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
```diff
@@ -19,9 +21,24 @@ private[debug] final class SetBreakpointsRequestHandler(
...
-    val input = path.toAbsolutePath.toInput
-    val occurrences = Mtags.allToplevels(input).occurrences
+    val path = adapters.adaptPath(request.getSource.getPath).toAbsolutePath
+
+    val topLevels = Mtags.allToplevels(path.toInput)
+    val occurrences = path.toLanguage match {
+      case Language.JAVA =>
+        def isTypeSymbol(symbol: SymbolInformation): Boolean = {
+          val kind = symbol.kind
+          kind.isClass || kind.isInterface
```
tgodzik: What about enums?

marek1840: Good catch, thanks!
It turns out enums have a `class` kind, so they are already covered. Nevertheless, I have added a test case for them.

PR #1218, opened 2019-12-21
Update Sublime settings regarding window/showMessage

ayoub-benali:
Previously LSP displayed a dialog box for `window/showMessage`, latest version (0.9.3 via PR https://github.com/tomv564/LSP/pull/811) displays that in the status bar.

Additionally: 
- make `metals/status` go as `window/showMessage`
- removed some outdated comments.

### DIFF
--- a/docs/editors/new-editor.md
+++ b/docs/editors/new-editor.md
@@ -168,7 +168,7 @@ Possible values:
 
 - `off` (default): don't start a server with the Metals HTTP client.
 - `on`: start a server with the [Metals HTTP client] to interact with the server
-  through a basic web UI.
+  through a basic web UI. This option is needed for editor clients that don't support necessary requests such as `window/showMessageRequest`.
 
 ### `-Dmetals.icons`
 
--- a/docs/editors/new-editor.md
+++ b/docs/editors/new-editor.md
@@ -114,7 +114,7 @@ Possible values:
 - `log-message`: translate `metals/status` notifications to `window/logMessage`
   notifications. Used by vim-lsc at the moment.
 - `show-message`: translate `metals/status` notifications to
-  `window/showMessage` notifications. Used by coc.nvim at the moment.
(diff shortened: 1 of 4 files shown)


docs/editors/new-editor.md
```diff
@@ -168,9 +168,7 @@ Possible values:
 
 - `off` (default): don't start a server with the Metals HTTP client.
 - `on`: start a server with the [Metals HTTP client] to interact with the server
-  through a basic web UI. This option is needed for editor clients like Sublime
```
tgodzik: Let's add back `This option is needed for editor clients that don't support necessary requests such as
  `window/showMessageRequest`.`

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -144,9 +144,8 @@ object MetalsServerConfig {
       case "sublime" =>
         base.copy(
           isHttpEnabled = true,
-          // Sublime text opens an invasive alert dialogue for window/showMessage
-          // and window/showMessageRequest.
-          showMessage = ShowMessageConfig.logMessage,
+          showMessage = ShowMessageConfig.on,
```
gabro: I would remove this line so that it uses the default as the other editors (the default is `"on"` unless overridden by a system property) 

Actually, we can probably remove `ShowMessageConfig` and `ShowMessageRequestConfig` altogether since as far as I know they were only used by Sublime Text.

I doubt anyone uses those system properties directly, and if they do, I would be curious to know why and better handle this on the server directly.

gabro: (It's fine to just remove the line for this PR, I can follow up with a different PR cleaning up the config options)

gabro: actually, I'll merge as is and follow up directly.

PR #443, opened 2018-12-31
update sublime doc to reflect the changes in the LSP plugin

ayoub-benali:
given the new [release](https://github.com/tomv564/LSP/releases/tag/0.8.0) for the LSP plugin, I updated the doc. 

@olafurpg do we need to create a gif for the "import changes" prompt ?

PR #1222, opened 2019-12-26
Remove showMessage and showMessageRequest configuration options

gabro:
These were only customized by the Sublime Text LSP client, which now
has proper support for the `window/showMessage` and
`window/showMessageRequest` notifications (see https://github.com/scalameta/metals/pull/443 and https://github.com/scalameta/metals/pull/1218)

I think these options served their purpose of getting the most popular editors on-board and can now be removed.


gabro:
I've also removed `isNoInitialized` since it was unused

gabro:
I agree with the split sentiment, but I ultimately consider those settings a list of quirks that should decrease over time, since standard LSP support for editors keeps getting better and better

PR #1227, opened 2019-12-27
Updates to vim section

ckipp01:
Updates to vim documentation

- Added in information about alternative ways to install
- Changed out Doctor information for an embedded version
- A new section about statusline integration
- Fixed some typos

### DIFF
--- a/docs/editors/vim.md
+++ b/docs/editors/vim.md
@@ -16,8 +16,9 @@ which will provide the most complete implementation of LSP and Metals-specific h
 ## Installing Vim
 
 The coc.nvim plugin requires either **Vim >= 8.1** or **Neovim >= 0.3.1**. Make
-sure you have the correct version installed. While it works with both Vim and Neovim, the
-experience is a bit smoother with Neovim.
+sure you have the correct version installed. While it works with both Vim and Neovim,
+we recommend using Neovim since it provides a smoother experience with some of the features such
+as code actions.
 
 ```sh
 # If using Vim
@@ -179,8 +180,14 @@ command like so:
 :CocInstall https://github.com/ckipp01/coc-metals
 ```
 If you'd like to use the latest changes on master, but would prefer managing the plugin using a plugin
-manager to download the extension, make sure you run `:CocUninstall coc-metals` if you've had
-coc-metals installed before. Then, if you are using [`vim-plug`](https://github.com/junegunn/vim-plug)
+manager to download the extension, make sure you run the below snippet to uninstall the old version
+first.
+
+```vim
+:CocUninstall coc-metals
+```
+
+Then, if you are using [`vim-plug`](https://github.com/junegunn/vim-plug)
 for example, enter the following into where you manage your plugins:
 
 ```vim
@@ -309,11 +316,11 @@ The Metals server is shutdown when you exit vim as usual.
 `coc.nvim` has multiple ways to integrate with various statusline plugins. You can find instructions
 for each of them located [here](https://github.com/neoclide/coc.nvim/wiki/Statusline-integration).
 Two noteworthy things that they add are the ability to see diagnostic information in the current
-buffer...
+buffer.
 
 ![Diagnostic statusline](https://i.imgur.com/7uNYTYl.png)
 
-... and also progress information for longer standing processes.
+And it also shows progress information for longer standing processes.
 
 ![Progress item](https://i.imgur.com/AAWZ4o4.png)
 
@@ -337,8 +344,8 @@ let g:airline_section_c = '%f%{CocExtensionStatus()}'
 ```
 ## Formatting on save
 
-If you'd like to have `:w` format using Metals + Scalafmt, then make sure you have the following in
-your `:CocConfig`.
+Add the following configuration to `:CocConfig` if you'd like to have `:w` format using Metals and
+Scalafmt.
 
 ```json
 "coc.preferences.formatOnSaveFiletypes": ["scala"]
@@ -375,3 +382,5 @@ binary does not exist yet.
 The `-Dmetals.client=vim-lsc` flag is important since it configures Metals for
 usage with the `vim-lsc` client.
 
+If you are using `coc.nvim` without the `coc-metals` extension, you'll need to include the
+`-Dmetals.http=true` flag when you are bootstrapping Metals in order to use Doctor.
--- a/docs/editors/vim.md
+++ b/docs/editors/vim.md
@@ -18,7 +18,7 @@ which will provide the most complete implementation of LSP and Metals-specific h
 The coc.nvim plugin requires either **Vim >= 8.1** or **Neovim >= 0.3.1**. Make
 sure you have the correct version installed. While it works with both Vim and Neovim,
 we recommend using Neovim since it provides a smoother experience with some of the features such
-as code actions.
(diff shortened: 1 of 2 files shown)


docs/editors/vim.md
```diff
@@ -17,7 +16,8 @@ which will provide the most complete implementation of LSP and Metals specific h
 ## Installing Vim
 
 The coc.nvim plugin requires either **Vim >= 8.1** or **Neovim >= 0.3.1**. Make
-sure you have the correct version installed.
+sure you have the correct version installed. While it works with both Vim and Neovim, the
+experience is a bit smoother with Neovim.
```
ckipp01: I originally wasn't going to mention anything about a preference for Neovim, but after spending some time with the new Vim 8.2, I feel pretty confident in saying that Neovim is a smoother experience. It's snappier with larger codebases for one, but another thing leading me to say this is that code actions seem sort of unusable at the moment with Vim 8.2 and coc.nvim (see https://github.com/neoclide/coc.nvim/issues/1404)

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -133,7 +133,6 @@ object MetalsServerConfig {
       case "coc.nvim" =>
         base.copy(
           statusBar = StatusBarConfig.showMessage,
-          isHttpEnabled = true,
```
ckipp01: I've taken this out and there are also some other changes that I'll make to the configuration like `isInputBoxEnabled` being turned on and `executeClientCommand` also being on for `coc-metals`. However, I'm turning those on in the extension itself in order to allow someone to still bootstrap Metals themselves for coc.nvim and not use the extension. If they do that, then client commands and input boxes won't be implemented allowing the default settings in here to still take effect.

docs/editors/vim.md
olafurpg: I'm fine with recommending neovim over vim. How about?
```suggestion
While it works with both Vim and Neovim, we recommend using Neovim since it provides the most feature-complete experience with Metals.
```

We might want to list what features are currently working more smoothly in neovim. You never know, maybe somebody will take on the heroic task of improving the situation for vim.

docs/editors/vim.md
```diff
@@ -165,20 +165,30 @@ inoremap <silent><expr> <cr> pumvisible() ? coc#_select_confirm()
...
-make sure you run `:CocUninstall coc-metals` to remove it. Then, if you are using [`vim-plug`](https://github.com/junegunn/vim-plug)
+
+If you'd like to use the latest changes on master, you can also use the url of the repo in the
+command like so:
+
+```vim
+:CocInstall https://github.com/ckipp01/coc-metals
+```
+If you'd like to use the latest changes on master, but would prefer managing the plugin using a plugin
+manager to download the extension, make sure you run `:CocUninstall coc-metals` if you've had
```
olafurpg: >  make sure you run `:CocUninstall coc-metals`

This is a bit hidden inside the paragraph. It might be good to extract it into a code fence like this
````
```
:CocUninstall coc-metals
```
````

File: docs/editors/vim.md [hunk 4]
```diff
@@ -235,9 +245,10 @@ window allowing you to search for `metals.doctor-run` command.
 
 ![Run Doctor Command](https://i.imgur.com/QaqhxF7.png)
 
-This command opens your browser with a table like this.
+This command opens an embedded doctor in your preview window. If you're not familiar with
+having multiple windows, you can use `<C-w> + w` to jump into it.
 
-![Run Doctor](https://i.imgur.com/yelm0jd.png)
+![Embedded Doctor](https://i.imgur.com/McaAFv5.png)
```
olafurpg: 😻

docs/editors/vim.md
```diff
@@ -284,23 +296,62 @@ To close a buffer and return to the previous buffer, run the following command.
...
 :wq
 ```
 
-This step clean ups resources that are used by the server.
+## Statusline integration
+
+`coc.nvim` has multiple ways to integrate with various statusline plugins. You can find instructions
+for each of them located [here](https://github.com/neoclide/coc.nvim/wiki/Statusline-integration).
+Two noteworthy things that they add are the ability to see diagnostic information in the current
+buffer...
```
olafurpg: What do the triple dots mean?
```suggestion
buffer.
```

olafurpg: Aha, it seems to be to imply the sentence continues. I generally try to avoid letting sentences go over images since it's difficult to read the sentence that way

docs/editors/vim.md
```diff
@@ -284,23 +296,62 @@ To close a buffer and return to the previous buffer, run the following command.
...
+
+```vim
+function! CocExtensionStatus() abort
+  return get(g:, 'coc_status', '')
+endfunction
+let g:airline_section_c = '%f%{CocExtensionStatus()}'
+```
+## Formatting on save
+
+If you'd like to have `:w` format using Metals + Scalafmt, then make sure you have the following in
```
olafurpg:
```suggestion
Add the following configuration to `:CocConfig` if you'd like to have `:w` format using Metals and Scalafmt
```

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
olafurpg: Should we introduce a `-Dmetals.client=coc-metals` instead? There might still be coc.nvim users that are not using coc-metals.

ckipp01: I've thought about that, and I would say that if it moves away any further, that might be a good idea. For now, I think just setting the extra needed values in `coc-metals` should suffice, and then that way we don't have to add another entry for it. Using coc.nvim without coc-metals should still work fine with what is currently in here.

Maybe once I dig into the tree view stuff, then it'll be a good idea to split it out.

docs/editors/vim.md
ckipp01: The only thing I don't like about this change is that technically it doesn't provide a more _feature-complete_ experience, since they both allow for the same features. However, in Neovim, they seem to just work smoother without any hangs. Some of it is purely performance related.

ckipp01: I'll massage this a bit though.

docs/editors/vim.md
ckipp01: Sounds good. I'll change this around to not have these.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
ckipp01: Thinking about it more, I'll just do it right away. I'll end up doing it anyways, so why not now.

ckipp01:
Alright, I made the recommended changes to the wording etc, and I also added in a new `coc-metals` option to the `MetalsServerConfig`. I wasn't going to do it right away, but this will just continue to change so I might as well do it now.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
ckipp01: See my other comment about not doing this now.

PR #1232, opened 2019-12-30
Discard 'code' from BSP diagnostics

### LINKED ISSUE #1220: Diagnostics have a confusing "code"
**Describe the bug**

Compile errors are displayed in a confusing way

Screenshot 2019-12-23 at 21 33 17

Notice the "bloop(val map: Map(..." part. I expected that to be only "bloop"

The culprit seems to be the "code" field for diagnostic

```json
[Trace - 09:32:17 PM] Sending notification 'textDocument/publishDiagnostics'
Params: {
  "uri": "file:///Users/lgeirsson/dev/geirsson.com/mark/src/main/scala/blog/Post.scala",
  "diagnostics": [
    {
      "range": {
        "start": {
          "line": 5,
          "character": 30
        },
        "end": {
          "line": 5,
          "character": 40
        }
      },
      "severity": 1,
      "code": "  val map: Map[Int, String] \u003d expression",
      "source": "bloop",
      "message": "type mismatch;\n found   : scala.collection.immutable.Map[Int,Boolean]\n required: Map[Int,String]"
    }
  ]
}
```
I think the "code" field is only supposed to contain information like "E20" for the category of the error

gabro:
Fixes #1220 

This is a temporary fix that discards the `code` property from BSP diagnostics. If we decide to change the `code` property semantics in BSP, then we can revert this fix. See the discussion here https://github.com/scalacenter/bloop/pull/1135

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -482,8 +482,9 @@ object MetalsEnrichments
         diag.getRange.toLSP,
         fansi.Str(diag.getMessage).plainText,
         diag.getSeverity.toLSP,
-        if (diag.getSource == null) "scalac" else diag.getSource,
-        diag.getCode
+        if (diag.getSource == null) "scalac" else diag.getSource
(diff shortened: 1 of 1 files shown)


gabro:
Windows CI failure is unrelated

PR #1235, opened 2019-12-31
Add in server property for doctor-format and implement json format

### LINKED ISSUE #1234: Allow Doctor to return multiple formats
Currently, Doctor only returns HTML. If you have a client that would like to display this information differently (a textual representation for example), they have to parse the HTML, which isn't ideal.

It'd be great if we could create a server property like `doctor.format=html|json|markdown”`, which would allow for the client to set which type it'd like to receive.

I'm currently building this into `coc-metals` and I'm parsing the HTML, and then grabbing the information to display. It's not ideal as you have to create the correct parsing for each case in a sort of a hacky way.

**Search terms**:
Doctor

ckipp01:
fixes #1234 

This will allow for a new server property to be set `metals.doctor-format` that currently has two options, html and json. This will allow for clients like `coc-metals` to more easily parse the information and display it textually.

Here are a couple examples of the current format:

```json
{
  "name": "Metals Doctor",
  "headerText": "These are the installed build targets for this workspace. One build target corresponds to one classpath. For example, normally one sbt project maps to two build targets: main and test.",
  "messages": [
    {
      "title": "⚠️  No build targets were detected in this workspace so most functionality won't work.",
      "recommendations": [
        "Make sure the workspace directory '/Users/ckipp/Documents/scala-workspace/test-project' matches the root of your build.",
        "Try removing the directories .metals/ and .bloop/, then restart metals And import the build again."
      ]
    }
  ]
}
```
and

```json
{
  "name": "Metals Doctor",
  "headerText": "These are the installed build targets for this workspace. One build target corresponds to one classpath. For example, normally one sbt project maps to two build targets: main and test.",
  "targets": [
    [
      {
        "Build Target": "root-test",
        "Scala": "2.12.10",
        "Diagnostics": "✅ ",
        "Goto definition": "✅ ",
        "Completions": "✅ ",
        "Find references": "✅ ",
        "Recommendation": ""
      },
      {
        "Build Target": "root",
        "Scala": "2.12.10",
        "Diagnostics": "✅ ",
        "Goto definition": "✅ ",
        "Completions": "✅ ",
        "Find references": "✅ ",
        "Recommendation": ""
      }
    ]
  ]
}
```


metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -63,10 +64,14 @@ final class Doctor(
       onServer: MetalsHttpServer => Unit
   ): Unit = {
     if (config.executeClientCommand.isOn) {
-      val html = buildTargetsHtml()
+      val output = config.doctorFormat match {
```
olafurpg: nit: can we use if/else instead of pattern matching here?

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -125,10 +130,14 @@ final class Doctor(
 
     def isMaven: Boolean = workspace.resolve("pom.xml").isFile
     def hint() =
-      if (isMaven)
-        "enable SemanticDB following instructions on the " +
-          "<a href=https://scalameta.org/metals/docs/build-tools/maven.html>Metals website</a>"
-      else s"run 'Build import' to enable code navigation."
+      if (isMaven) {
+        val website =
+          if (config.doctorFormat.value == "html")
```
olafurpg: ?
```suggestion
          if (config.doctorFormat.isHtml)
```

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -297,4 +327,49 @@ final class Doctor(
       )
     }
   }
+
+  private def extractTargetInfo(target: ScalaTarget) = {
```
olafurpg: Can we introduce a case class that wraps all the columns here? It would be great to avoid passing around a `Tuple7`

metals/src/main/scala/scala/meta/internal/metals/DoctorFormatConfig.scala
```diff
@@ -0,0 +1,16 @@
+package scala.meta.internal.metals
+
+final case class DoctorFormatConfig(value: String) {
+  def isHtml: Boolean = value == "html"
+  def isJson: Boolean = value == "json"
+  def isMarkdown: Boolean = value == "markdown"
```
olafurpg: Remove?

metals/src/main/scala/scala/meta/internal/metals/DoctorFormatConfig.scala
```diff
@@ -0,0 +1,16 @@
...
+}
+
+object DoctorFormatConfig {
+  def html = new DoctorFormatConfig("html")
+  def json = new DoctorFormatConfig("json")
+  def markdown = new DoctorFormatConfig("markdown")
```
olafurpg: Remove?

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -119,7 +122,6 @@ object MetalsServerConfig {
       case "coc.nvim" =>
         base.copy(
           statusBar = StatusBarConfig.showMessage,
-          isHttpEnabled = true,
```
olafurpg: Is this change intentional?

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
ckipp01: _Makes `isHtml` method... and then doesn't use it._ 

Yes, ha, I'll use this.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
ckipp01: Sure. I originally was also going to do Markdown, which led me down the match route, but then decided against it. I'll switch this.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
ckipp01: Ah, no. Sorry, was playing around with this for testing. I'll revert it.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -204,37 +211,65 @@ final class Doctor(
...
+          "recommendations" -> Arr(noBuildTargetRecOne, noBuildTargetRecTwo)
+        )
+      )
+      ujson.write(json)
+    } else {
+      val targetResults = targets.sortBy(_.info.getBaseDirectory).map {
+        target =>
+          val targetInfo = extractTargetInfo(target)
+          ujson.Obj(
+            "Build Target" -> targetInfo.name,
```
gabro: Minor, but in order to make these json properties more appealing for a JS client, can we use `camelCase` keys?

gabro: Actually, what do you think about defining a case class for the entire result and define ujson codecs for them?

metals/src/main/scala/scala/meta/internal/metals/DoctorTargetInfo.scala
```diff
@@ -0,0 +1,10 @@
+package scala.meta.internal.metals
+
+final case class DoctorTargetInfo(
+    name: String,
+    scalaVersion: String,
+    definitionStatus: String,
```
gabro: I propose to have an enum here for the status. I think sending emojis is not super friendly in the JSON API (if the client wants to displays something custom they would have to match on the emoji character).

ckipp01:
Thanks for the quick review @olafurpg!

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
ckipp01: Regarding the `camelCase`, totally.

Regarding the codecs approach. I guess the result we want to return the `targets` array is just the return value of `extractTargetInfo`, so we already have the case class. (Unless you mean case class for the entire return value).

I would normally be all for doing the codecs, but do you think in this situation it's useful? There's really only a few different things that the api can return right now, and it's pretty contained and straight-forward in here. I think with the html approach we tried to keep is super simple, and I tried to do the same with JSON. Again, I'm not opposed to it, but what does it really gain us here by making the codes vs just being explicit with it here?

metals/src/main/scala/scala/meta/internal/metals/DoctorTargetInfo.scala
ckipp01: If not the emoji characters, what would the enum values be here?

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
gabro: > Unless you mean case class for the entire return value

Yep, I also meant that.

> making the codes vs just being explicit 

To be clear: the codec can be explicit (I don't meant to auto-derive it or anything) and live in the companion object of the respective case classes. My main concern is that now that we have different formats, keeping the encoding logic together with the data logic will make the doctor implementation a bit hard to read.

metals/src/main/scala/scala/meta/internal/metals/DoctorTargetInfo.scala
gabro: I would keep it simple and do:

```scala
sealed abstract class DoctorStatus
object DoctorStatus {
  case object Ok extends DoctorStatus
  case object Error extends DoctorStatus
}
```

and then serialize it as `"Ok"` and `"Error"` in the json response and as unicode characters in the html response.

gabro: The use-case I'm assuming here is that the client won't display this value as-is but it will choose how to represent the information.

If we are sure clients will always be ok in displaying the emoji directly, I say it's fine to leave it as is

ckipp01: Yea, that's a good point. For now, I know I'm going to just use it as is in `coc-metals`. My opinion is to leave it as if now, and later on of someone wants to display something custom based off those results, then we can easily change it then, but no need to now.

ckipp01:
Alright, I took at stab at what you were mentioning @gabro. I'll be honest, I no idea if that was the approach you were thinking. I'd very much like some feedback on how to improve it.

metals/src/main/scala/scala/meta/internal/metals/DoctorTargetInfo.scala
gabro: Sounds good, let's see whether some other client uses this first 👍

ckipp01:
> @ckipp01 👍 thanks for refactoring this, I think it reads better now. I might take a stab at using `Writer` from upickle, but that's definitely out of scope now.

oooo, I didn't even know about this as I was only looking at the uJson package. Thanks for pointing that out!

olafurpg:
ujson doesn’t include a serialization type class or macros, it’s just a plain JSON library. The higher-level functionality is available in upickle

Issue #1236, opened 2020-01-01
Worksheets hang on diagnostics without positions


olafurpg:
**Describe the bug**

Worksheets don't evaluate when the document has unpositioned diagnostics.
**To Reproduce**
1. Open `Foo.worksheet.sc` file
2. type
```scala
trait X[F[_]]
val x = 3
```
3. save file

**Expected behavior**

I expected evaluation to complete.

A workaround is to add the following line

```scala
import scala.language.higherKinds
```

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: 0.7.6+641-25d169c8-SNAPSHOT

**Additional context**

Reported here https://gitter.im/scalameta/metals?at=5e0c6abafd580457e7a7bd18
**Search terms**

Evaluating worksheet

ckipp01:
I was looking at the #1237, and I don't actually fully understand why this happens. Why is it that if we have the same trait in your example in a file, it will compile fine without any diagnostics being returned, but then in a worksheet, there are diagnostics returned without a position?

gabro:
It does return a diagnostic, a warning without a position, if I understand this correctly

olafurpg:
The scala compiler sometimes reports diagnostics with empty positions for things like “there are deprecation warnings, enable -deprecation to see details”. 

In LSP, diagnostics must have a location to place such messages so I believe we ignore them in normal scala files.

gabro:
Closed by #1237

PR #1237, opened 2020-01-01
Fix issue #1236, make worksheets handle unpositioned diagnostics.

olafurpg:
Previously, worksheets did not gracefully handle diagnostics that had no
positions resulting in hanging "Evaluating worksheet" notification
windows. Now, we filter out such diagnostics so that the worksheet
finishes evaluation. We also ensure that unexpected exceptions no longer
cause hanging "Evaluating worksheet" notification windows.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
@@ -139,7 +139,11 @@ class WorksheetProvider(
       // infinite loops.
       val thread = new Thread(s"Evaluating Worksheet ${path.filename}") {
         override def run(): Unit = {
-          result.complete(evaluateWorksheet(path, token))
+          try result.complete(evaluateWorksheet(path, token))
+          catch {
+            case e @ (NonFatal(_) | InterruptException()) =>
+              result.completeExceptionally(e)
+          }
         }
       }
       interruptThreadOnCancel(path, result, thread)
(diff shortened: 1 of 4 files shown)


olafurpg:
Opened https://github.com/scalameta/metals/pull/1238 disabling the unrelated flaky test failure.

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -220,9 +224,16 @@ class WorksheetProvider(
       input = path.toInputFromBuffers(buffers)
     } yield mdoc.evaluateWorksheet(input.path, input.value)
   }.map { worksheet =>
+    val toPublish = worksheet
+      .diagnostics()
+      .iterator()
+      .asScala
+      .filterNot(_.position().isNone)
```
tgodzik: Should we not at least try to publish those diagnostics at the start? Will it not just hide it from the user? We could show it at the start of the file maybe?

olafurpg: Good point, I changed these diagnostics to display in the beginning of the file instead.

olafurpg:
> This also means we'll start showing these diagnostics in regular Scala files, right?

No, this change is limited to worksheets.

olafurpg:
Test failure in 2.13 looks legitimate
```
X tests.feature.Worksheet213LspSuite.no-position 6679ms 
  tests.TestFailedException: failed assertion at /home/runner/work/metals/metals/tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala:344
failed assertion at /home/runner/work/metals/metals/tests/mtest/src/main/scala/tests/DiffAssertions.scala:24
Obtained empty output!
  tests.DiffAssertions$.colored(DiffAssertions.scala:107)
```

PR #1240, opened 2020-01-02
Add in upickle to use for json doctor

ckipp01:
After your comments @gabro I ended up taking a further look into the `write` options with upickle. I think this is a lot cleaner, but I also realize we are now fully pulling in upickle instead of just ujson. Are we ok with that? @olafurpg I also saw that you were using ujson in some of your pants stuff that looked like it _may_ also benifit from the read/write of upickle, but didn't really want to touch it unless you think I should also try to convert some of that stuff to use this.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -284,7 +284,7 @@ lazy val metals = project
       // for debugging purposes, not strictly needed but nice for productivity
       "com.lihaoyi" %% "pprint" % "0.5.7",
       // For exporting Pants builds.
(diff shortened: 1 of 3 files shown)


gabro:
I agree with @olafurpg, when I suggested to use the automatic encoders I thought upickle was already on the classpath. If it's not, I don't think it holds its weight for the few cases we have.
The manual approach is fine (although I would still separate encoding logic from business as much as possible)

ckipp01:
Sounds good. I'll go ahead and close this then. 👍

PR #1245, opened 2020-01-04
Fix multi-line range formatting when on first line

### LINKED ISSUE #1244: When pasting at the start of  a multiline string pipe is added unnecessarily
**Describe the bug**
When pasting in a multiline string with existing `|` a new `|` is added before the string itself/

**To Reproduce**
1.Make sure `On paste formatting` is enabled
2.Create multiline string like:
```
s"""| this is@@ a multiline
      | string
      |""".stripMargin
```
3. Paste something not multiline at the position of `@@`
4. `|` will be added before the `s"""`

**Expected behavior**
Nothing should be added there.

**Screenshots**

image

**Installation:**
- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: newest snapshot

**Search terms**
multiline on paste formatting pipe

ckipp01:
Closes #1244 

When I originally worked on this (my first Metals ticket 🎉 ), I totally forgot about the case where you paste on the first line of the multi-line string. This adds in a check for that and also a test for it.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
@@ -145,7 +145,7 @@ final class MultilineStringFormattingProvider(
       line: Int,
       lines: Array[String],
       defaultIndent: String
-  ): TextEdit = {
+  ): Option[TextEdit] = {
     val zeroPos = new Position(line, 0)
     val lineText = lines(line)
     val firstChar = lineText.trim.headOption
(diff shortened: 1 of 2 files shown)


ckipp01:
``` 
[error] Failed tests:
[error] 	tests.BspSwitchLspSuite`
```
Test failure seems to be unrelated.

metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -159,14 +159,24 @@ final class MultilineStringFormattingProvider(
...
             val pipePos = new Position(line, firstPipeIndex)
-            new TextEdit(new Range(zeroPos, pipePos), defaultIndent)
+            val textEdit =
+              new TextEdit(new Range(zeroPos, pipePos), defaultIndent)
+            Some(textEdit)
         }
       case _ =>
-        val newText = defaultIndent + "|"
-        new TextEdit(new Range(zeroPos, zeroPos), newText)
+        val isFirstLineOfMultiLine = lineText.trim.contains("\"\"\"|")
```
olafurpg: What happens if the line doesn’t have a pipe? It’s optional

ckipp01: If there is no pipe on the line it gets filtered out beforehand and the line never enters the `formatPipeLine` method. I suppose I should have added a test for this as well. I did just check it locally and that indeed is how it works.

ckipp01: I just added another test for this https://github.com/scalameta/metals/pull/1249

PR #1249, opened 2020-01-04
add in another test for multi-line string formatting

ckipp01:
Just for good measure, I've added in another test to ensure that this is working in both situations where there is a pipe on the first line and when there isn't a pipe.

### DIFF
--- a/tests/unit/src/test/scala/tests/RangeFormattingSuite.scala
+++ b/tests/unit/src/test/scala/tests/RangeFormattingSuite.scala
@@ -62,7 +62,7 @@ object RangeFormattingSuite extends BaseLspSuite("rangeFormatting") {
   )
 
   check(
-    "paste-on-fist-line",
+    "paste-on-first-line-with-pipe",
     s"""
        |object Main {
        |  val str = '''| hi @@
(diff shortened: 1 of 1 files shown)


tests/unit/src/test/scala/tests/RangeFormattingSuite.scala
```diff
@@ -62,7 +62,7 @@ object RangeFormattingSuite extends BaseLspSuite("rangeFormatting") {
   )
 
   check(
-    "paste-on-fist-line",
+    "paste-on-fist-line-with-pipe",
```
gabro:
```suggestion
    "paste-on-first-line-with-pipe",
```

👊

tests/unit/src/test/scala/tests/RangeFormattingSuite.scala
```diff
@@ -78,6 +78,23 @@ object RangeFormattingSuite extends BaseLspSuite("rangeFormatting") {
        |}""".stripMargin
   )
 
+  check(
+    "paste-on-fist-line-without-pipe",
```
gabro: ✊

tests/unit/src/test/scala/tests/RangeFormattingSuite.scala
ckipp01: Nope, not fixing, it was like that before :trollface:

PR #1250, opened 2020-01-04
Support scaladoc auto completion on type `/**`

### LINKED ISSUE #1252: Support Scaladoc template auto completion
When I try to write scaladoc on method definitions with intellij-scala-plugin, it auto-completes the `@param parameterName` and `@return` based on the method definition below the cursor like this.

scaladocintellij

And this feature is available in TypeScript (VsCode + TSServer) too.
jsdoc

This is my favorite feature of intellij-scala-plugin and tsserver, and It would be awesome if the feature is available in metals as well:

- When the user type `/**`, show the popup for selecting completions
  - `/** */ Scaladoc comment`
- if the user hit the `/** */ Scaladoc comment`, auto-complete the scaladoc template based on the defined method just below the cursor.

For example,

```scala
/**| <- cursor here
def test(x: String, y: String): Boolean = {
  // ...
}
```

to

```scala
/**
  *
  * @param x | <- move cursor to here
  * @param y
  * @return
  */
def test(x: String, y: String): Boolean = {
  // ...
}
```

Currently, I write `@param` and `@return` by hand, but if metals auto-complete them, it would be easier to write detailed scaladoc, and potentially motivate people to write scaladoc.

**Additional context**

**Search terms**:
scaladoc, completion, docstring

tanishiking:
Closes #1252

This PR enables the metals server to auto-complete scaladoc for **method definition** and **class definition** on type `/**`

vscode-scaladoc

### Implement Overview
- Enable to fire `textDocument/completion` on type `*` by setting `triggerCharacters` on `initialize`.
- If the line only have a `/**`, try to complete scaladoc
  - Close the comment by inserting `*/` behind the cursor before adding the source to the compilation unit, so that metals can traverse whole AST in the source.
  - Find an associated member definition by traverse the AST for the source, and ~collect all `MemberDef` that is defined below the `/**`.~, and get the closest `MemberDef` that is defined below the `/**` as an associated definition.
  - Generate scaladoc based on the definition.

#### CursorPosition
<details>
<summary>Outdated :)</summary>

~Currently, after the completionEdit, the cursor will move to behind the first parameter like,

```scala
/**@@
def test(x: Int, y: Int): Int = ???
```

to

```scala
/**
  *
  * @param x | <- move the cursor to here like tsserver
  * @param y
  * @return
  */
def test(x: Int, y: Int): Int = ???
```

Is it ok to move the cursor here? 

```scala
/**
  * | <- maybe we should move cursor here instead? like intellij-scala-plugin
  *
  * @param x
  * @param y
  * @return
  */
def test(x: Int, y: Int): Int = ???
```

</details>

```scala
/**
  * | <- move cursor here instead? like intellij-scala-plugin, is it ok?
  *
  * @param x
  * @param y
  * @return
  */
def test(x: Int, y: Int): Int = ???
```

#### Scaladoc or Javadoc? 
Now, this PR add scaladoc with Javadoc style with gutter asterisks aligned in column three like:

```scala
/**
  * | <- move cursor here
  *
  * @param x
  * @return
  */
def method(x: Int): Int = ???
```

https://docs.scala-lang.org/style/scaladoc.html

But, IMO it's common to write scaladoc with Javadoc style, with gutter asterisks aligned in column two:

```scala
/**
 * | <- move cursor here
 *
 * @param x
 * @return
 */
def method(x: Int): Int = ???
```

Should we complete the doc that format instead?

---

- I haven't checked the behavior on Windows yet, and I'm going to confirm the behavior later (after setting up the windows working environment) and check it with other language server client than VSCode

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -494,7 +494,7 @@ class MetalsLanguageServer(
       capabilities.setCompletionProvider(
         new CompletionOptions(
           config.compilers.isCompletionItemResolve,
-          List(".").asJava
+          List(".", "*").asJava
         )
       )
       capabilities.setWorkspaceSymbolProvider(true)
--- a/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
@@ -27,9 +27,8 @@ class CompletionProvider(
         // Insert potentially missing `}` to avoid "unclosed literal" error in String interpolator..
         CURSOR + "}"
       case '*'
-          if params
-            .text()
-            .charAt(i - 1) == '*' && params.text().charAt(i - 2) == '/' =>
+          if params.text().charAt(i - 1) == '*' &&
+            params.text().charAt(i - 2) == '/' =>
         // Insert potentially missing `*/` to avoid comment out all codes after the "/**".
         CURSOR + "*/"
       case _ =>
--- a/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
@@ -26,6 +26,12 @@ class CompletionProvider(
       case '{' if params.text().charAt(i - 1) == '$' =>
         // Insert potentially missing `}` to avoid "unclosed literal" error in String interpolator..
         CURSOR + "}"
+      case '*'
+          if params
+            .text()
+            .charAt(i - 1) == '*' && params.text().charAt(i - 2) == '/' =>
+        // Insert potentially missing `*/` to avoid comment out all codes after the "/**".
+        CURSOR + "*/"
       case _ =>
         // Default _CURSOR_ instrumentation.
         CURSOR
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -16,6 +16,7 @@ import scala.meta.internal.tokenizers.Chars
 import scala.meta.pc.PresentationCompilerConfig.OverrideDefFormat
 import scala.util.control.NonFatal
 import scala.collection.immutable.Nil
+import scala.reflect.internal.ModifierFlags
 
 /**
  * Utility methods for completions.
@@ -711,7 +712,11 @@ trait Completions { this: MetalsGlobal =>
     private def process(t: MemberDef): Unit = {
       // if the node's location is below the current associate def,
       // we don't have to visit the node because it can't be an associated def.
-      if (associatedDef.map(t.pos.point <= _.pos.point).getOrElse(true)) {
+      if (associatedDef
+          .map(cur =>
+            t.pos.isDefined && cur.pos.isDefined && t.pos.point <= cur.pos.point
+          )
+          .getOrElse(true)) {
         if (t.pos.isDefined && t.pos.start >= pos.start) associatedDef = Some(t)
         if (treePos(t).includes(pos)) super.traverse(t)
       }
@@ -1802,17 +1807,33 @@ trait Completions { this: MetalsGlobal =>
       class ConstructorFinder(clazz: ClassDef) extends Traverser {
         private var found: Option[DefDef] = scala.None
         def getConstructor: Option[DefDef] = {
-          found = scala.None
-          clazz.impl.body.foreach(traverse)
-          found
+          // Don't try to find the constructor for trait/abstract definition
+          // because they don't have a constructor
+          if (!clazz.mods.hasFlag(
+              ModifierFlags.ABSTRACT |
+                ModifierFlags.TRAIT |
+                ModifierFlags.INTERFACE
+            )) {
+            found = scala.None
+            clazz.impl.body.foreach(traverse)
+            found
+          } else {
+            scala.None
+          }
         }
         override def traverse(tree: Tree): Unit = {
-          tree match {
-            case constructor: DefDef
-                if constructor.name == termNames.CONSTRUCTOR =>
-              found = Some(constructor)
-            case _ =>
-              super.traverse(tree)
+          if (found
+              .map(cur =>
+                tree.pos.isDefined && cur.pos.isDefined && tree.pos.point <= cur.pos.point
+              )
+              .getOrElse(true)) {
+            tree match {
+              case constructor: DefDef
+                  if constructor.name == termNames.CONSTRUCTOR =>
+                found = Some(constructor)
+              case _ =>
+                super.traverse(tree)
+            }
           }
         }
       }
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -430,7 +430,9 @@ trait Completions { this: MetalsGlobal =>
           new AssociatedMemberDefFinder(pos).findAssociatedDef(unit.body)
         }
         associatedDef
-          .map(definition => CompletionPosition.Scaladoc(editRange, definition))
+          .map(definition =>
+            CompletionPosition.Scaladoc(editRange, definition, pos, text)
+          )
           .getOrElse(CompletionPosition.None)
       case (ident: Ident) :: (a: Apply) :: _ =>
         fromIdentApply(ident, a)
@@ -1682,21 +1684,59 @@ trait Completions { this: MetalsGlobal =>
      * @param editRange the range in the original source file.
      * @param associatedDef the memberDef associated with the scaladoc to complete.
      *                      This class will construct scaladoc based on the params of this definition.
+     * @param pos the position of the completion request.
+     * @param text the text of the original source code.
      */
     case class Scaladoc(
         editRange: l.Range,
-        associatedDef: MemberDef
+        associatedDef: MemberDef,
+        pos: Position,
+        text: String
     ) extends CompletionPosition {
+      // The indent for gutter asterisks aligned in column three.
+      // |/**
+      // |  *
+      private val scaladocIndent = "  "
+
       override def contribute: List[Member] = {
-        val lines = scaladocLines(associatedDef)
-        // add `* @return` line only if the definition is a method def.
+        val necessaryIndent = inferIndent(pos, text)
+        val indent = s"${necessaryIndent}${scaladocIndent}"
+
+        val params: List[ValDef] = getParams(associatedDef)
+        val cursor =
+          if (clientSupportsSnippets) " $0" else ""
+
+        val scaladocParamLines: String = params
+          .map { param =>
+            s"${indent}* @param ${param.name}"
+          }
+          .mkString("\n")
+
+        // Add `* @return` only if the associatedDef is method definition.
         val returnLine =
-          if (lines.isEmpty && clientSupportsSnippets) "* @return $0"
-          else "* @return"
-        val newText =
-          if (associatedDef.isInstanceOf[DefDef])
-            (Seq("", "*") ++ lines ++ Seq(returnLine, "*/")).mkString("\n  ")
-          else (Seq("", "*") ++ lines ++ Seq("*/")).mkString("\n  ")
+          if (associatedDef.isInstanceOf[DefDef]) s"${indent}* @return" else ""
+
+        // Construct the following new text.
+        // """
+        //
+        //   * $0 <- move cursor here. Add an empty line below here, only if there's @param or @return line.
+        //   *
+        //   * @param param1
+        //   * @param param2
+        //   * @return
+        //   */
+        // """
+        val newText: String =
+          s"""|${indent}*${cursor}
+              |${if (scaladocParamLines.isEmpty && returnLine.isEmpty) ""
+             else s"${indent}*"}
+              |${scaladocParamLines}
+              |${returnLine}
+              |${indent}*/""".stripMargin
+            .split("\n")
+            .filter(!_.isEmpty) // remove empty lines: scaladocParamLines and returnLine can be empty
+            .mkString("\n", "\n", "")
+
         List(
           new TextEditMember(
             "Scaladoc Comment",
@@ -1710,55 +1750,64 @@ trait Completions { this: MetalsGlobal =>
           )
         )
       }
-    }
 
-    /**
-     * Returns the parameter lines of scaladoc besed on the given memberDef
-     * like ["* @param param1 $0", "* @param param2"].
-     *
-     * @param memberDef The memberDef to construct scaladoc.
-     */
-    private def scaladocLines(memberDef: MemberDef): List[String] = {
-      memberDef match {
-        case DefDef(_, _, _, vparamss, _, _) =>
-          vparamss.flatten.zipWithIndex
-            .map {
-              case (valdef, idx) => {
-                // /**
-                //   * @param param1 | <- move cursor to here.
-                //   * @param param2
-                //   */
-                if (idx == 0 && clientSupportsSnippets)
-                  s"* @param ${valdef.name} $$0"
-                else s"* @param ${valdef.name}"
-              }
+      // Infers the indentation at the completion position by counting the number of leading
+      // spaces in the line.
+      // For example:
+      // |"""
+      // |object A {
+      // |  /**<COMPLETE> // inferred indent is 4 spaces
+      // |  def foo(x: Int) = ???
+      // |}
+      // |"""
+      private def inferIndent(pos: Position, text: String): String = {
+        if (metalsConfig.snippetAutoIndent()) {
+          ""
+        } else {
+          val line =
+            try {
+              text.split(System.lineSeparator())(pos.line - 1)
+            } catch { case NonFatal(_) => "" }
+          line.takeWhile(ch => ch != '/')
+        }
+      }
+
+      /**
+       * Returns the parameters of the given memberDef
+       *
+       * @param memberDef The memberDef to construct scaladoc.
+       */
+      private def getParams(memberDef: MemberDef): List[ValDef] = {
+        memberDef match {
+          case defdef: DefDef =>
+            defdef.vparamss.flatten
+          case clazz: ClassDef =>
+            // If the associated def is a class definition,
+            // retrieve the constructor from the class, and caluculate the lines
+            // from the constructor definition instead.
+            new ConstructorFinder(clazz).getConstructor match {
+              case Some(defdef) => getParams(defdef)
+              case scala.None => Nil
             }
-        case clazz @ ClassDef(_, _, _, _) =>
-          // If the associated def is a class definition,
-          // retrieve the constructor from the class, and caluculate the lines
-          // from the constructor definition instead.
-          new ConstructorFinder(clazz).getConstructor match {
-            case Some(defdef) => this.scaladocLines(defdef)
-            case scala.None => Nil
-          }
-        case _ => Nil
+          case _ => Nil
+        }
       }
-    }
 
-    class ConstructorFinder(clazz: ClassDef) extends Traverser {
-      def getConstructor: Option[DefDef] = {
-        this.constructor = scala.None
-        clazz.impl.body.foreach(traverse)
-        constructor
-      }
-      private var constructor: Option[DefDef] = scala.None
-      override def traverse(tree: Tree): Unit = {
-        tree match {
-          case constructor @ DefDef(_, name, _, _, _, _)
-              if name == termNames.CONSTRUCTOR =>
-            this.constructor = Some(constructor)
-          case _ =>
-            super.traverse(tree)
+      class ConstructorFinder(clazz: ClassDef) extends Traverser {
+        def getConstructor: Option[DefDef] = {
+          this.constructor = scala.None
+          clazz.impl.body.foreach(traverse)
+          constructor
+        }
+        private var constructor: Option[DefDef] = scala.None
+        override def traverse(tree: Tree): Unit = {
+          tree match {
+            case constructor @ DefDef(_, name, _, _, _, _)
+                if name == termNames.CONSTRUCTOR =>
+              this.constructor = Some(constructor)
+            case _ =>
+              super.traverse(tree)
+          }
         }
       }
     }
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -685,16 +685,16 @@ trait Completions { this: MetalsGlobal =>
    *            This class will find the associated member def based on this pos.
    */
   class AssociatedMemberDefFinder(pos: Position) extends Traverser {
-    private var defs: List[MemberDef] = Nil
+    private var associatedDef: Option[MemberDef] = None
 
     /**
      * Collect all the member definitions whose position is
      * below the given `pos`. And then return the closest member definiton.
      */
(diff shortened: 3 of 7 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -494,7 +494,7 @@ class MetalsLanguageServer(
       capabilities.setCompletionProvider(
         new CompletionOptions(
           config.compilers.isCompletionItemResolve,
-          List(".").asJava
+          List(".", "*").asJava
```
tanishiking: Enable to fire `textDocument/completion` on type `*`

mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -26,6 +26,12 @@ class CompletionProvider(
       case '{' if params.text().charAt(i - 1) == '$' =>
         // Insert potentially missing `}` to avoid "unclosed literal" error in String interpolator..
         CURSOR + "}"
+      case '*'
+          if params
+            .text()
+            .charAt(i - 1) == '*' && params.text().charAt(i - 2) == '/' =>
+        // Insert potentially missing `*/` to avoid comment out all codes after the "/**".
+        CURSOR + "*/"
```
tanishiking: If we don't close the comment here, the whole code below the `/**` will be commented out, and we cannot traverse the AST (retrieved from a compilation unit) below the comment

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -669,6 +676,42 @@ trait Completions { this: MetalsGlobal =>
...
+    }
+    override def traverse(t: Tree): Unit = {
+      t match {
+        case typedef @ TypeDef(_, _, _, _) => process(typedef)
+        case clsdef @ ClassDef(_, _, _, _) => process(clsdef)
+        case defdef @ DefDef(_, _, _, _, _, _) => process(defdef)
+        case moduledef @ ModuleDef(_, _, _) => process(moduledef)
+        case pkgdef @ PackageDef(_, _) => process(pkgdef)
+        case valdef @ ValDef(_, _, _, _) => process(valdef)
+        case _ if treePos(t).includes(pos) => super.traverse(t)
```
tanishiking: What we construct scaladoc for is just `ClassDef` and `DefDef`, but collecting all the `MemberDef`  to avoid the following pattern.

```scala
/**@@
type Foo = Int // If we don't collect TypeDef

def foo(x: Int): Int = ???
```

to

```scala
/**
   *
   * @param x
   * @return
   */
type Foo = Int // If we don't collect TypeDef

def foo(x: Int): Int = ???
```

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -669,6 +676,42 @@ trait Completions { this: MetalsGlobal =>
...
+    /**
+     * Collect all the member definitions whose position is
+     * below the given `pos`. And then return the closest member definiton.
+     */
+    def findAssociatedDef(root: Tree): Option[MemberDef] = {
+      defs = Nil
+      traverse(root)
+      defs.sortBy(_.pos.point).headOption
+    }
+    override def traverse(t: Tree): Unit = {
```
tanishiking: Since just traversing the AST for the source, if the source has long lines, it takes a bit time before returning the completion response...

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1632,6 +1675,93 @@ trait Completions { this: MetalsGlobal =>
...
+
+    /**
+     * Returns the parameter lines of scaladoc besed on the given memberDef
+     * like ["* @param param1 $0", "* @param param2"].
+     *
+     * @param memberDef The memberDef to construct scaladoc.
+     */
+    private def scaladocLines(memberDef: MemberDef): List[String] = {
+      memberDef match {
+        case DefDef(_, _, _, vparamss, _, _) =>
```
gabro:
```suggestion
        case d: DefDef =>
```

and then `d.vparamss`

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1632,6 +1675,93 @@ trait Completions { this: MetalsGlobal =>
...
+    class ConstructorFinder(clazz: ClassDef) extends Traverser {
+      def getConstructor: Option[DefDef] = {
+        this.constructor = scala.None
+        clazz.impl.body.foreach(traverse)
+        constructor
+      }
+      private var constructor: Option[DefDef] = scala.None
+      override def traverse(tree: Tree): Unit = {
+        tree match {
+          case constructor @ DefDef(_, name, _, _, _, _)
```
gabro: similar suggestion, it's a bit easier to read by doing:

```suggestion
          case constructor: DefDef
```

and then

```scala
if constructor.name == termNames.CONSTRUCTOR
```

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1867,4 +1997,12 @@ trait Completions { this: MetalsGlobal =>
     }
   }
 
+  def isScaladocCompletion(pos: Position, text: String): Boolean = {
+    val line =
+      try {
+        text.split(System.lineSeparator())(pos.line - 1)
+      } catch { case _: Throwable => "" }
```
gabro: Do we need to catch `Throwable` here? It's probably fine if we catch `NonFatal` instead 

```suggestion
      } catch { case NonFatal(_) => "" }
```

Also, what about wrapping the entire method in `try/catch` and directly returning `false` instead?

tests/cross/src/test/scala/tests/pc/CompletionScaladocSuite.scala
```diff
@@ -0,0 +1,194 @@
...
+    "scaladoc-methoddef-edit",
+    """|object A {
+       |  /**@@
+       |  def test1(param1: Int, param2: Int): Int = ???
+       |  def test2(param1: Int, param2: Int, param3: Int): Int = ???
+       |}
+       |""".stripMargin,
+    """|object A {
+       |  /**
+       |  *
```
gabro: shouldn't the `*` be aligned with the first `*` in `/**` (following Scaladoc style)?

tests/cross/src/test/scala/tests/pc/CompletionScaladocSuite.scala
```diff
@@ -0,0 +1,194 @@
...
+  checkEdit(
+    "scaladoc-valdef-edit",
+    """|object A {
+       |  /**@@
+       |  val x = 1
+       |}
+       |""".stripMargin,
+    """|object A {
+       |  /**
+       |  *
```
gabro: where's the cursor in this case?

tests/cross/src/test/scala/tests/pc/CompletionScaladocSuite.scala
```diff
@@ -0,0 +1,194 @@
...
+       |  * @param param1 $0
+       |  * @param param2
+       |  * @return
+       |  */
+       |  def test1(param1: Int, param2: Int): Int = ???
+       |  def test2(param1: Int, param2: Int, param3: Int): Int = ???
+       |}
+       |""".stripMargin,
+    filter = _.contains("/** */")
```
gabro: do we need this filter? I would expect this to be the only completion returned.

Same for all the other tests below

tests/cross/src/test/scala/tests/pc/CompletionScaladocSuite.scala
```diff
@@ -0,0 +1,194 @@
...
+       |  def test(x: Int): Int = {
+       |    /**@@
+       |    def nest(y: Int) = ???
+       |  }
+       |}
+       |""".stripMargin,
+    """|object A {
+       |  def test(x: Int): Int = {
+       |    /**
+       |  *
```
gabro: indentation is off? We probably need something like `inferIndent` that is used for auto imports completions

tests/cross/src/test/scala/tests/pc/CompletionScaladocSuite.scala
```diff
@@ -0,0 +1,194 @@
...
+    "scaladoc-classdef-edit",
+    """|object A {
+       |  /**@@
+       |  class Test1(param1: Int, param2: Int) {}
+       |  class Test2(param1: Int, param2: Int, param3: Int) {}
+       |}
+       |""".stripMargin,
+    """|object A {
+       |  /**
+       |  *
```
gabro: The [official docs](https://docs.scala-lang.org/style/scaladoc.html#classes) also suggest the `@constructor` tag for classes. What do you think?

gabro:
About your questions, I think it makes sense to use Scaladoc style and I would personally expect the cursor to move before the `@param` block, so that I can start typing a description.

tests/cross/src/test/scala/tests/pc/CompletionScaladocSuite.scala
gabro: This actually works fine in VSCode, and I think I know what's going on here. 

VSCode and other LSP clients automatically add indentation when returning a multi-line `textEdit`. Other clients don't.

In a similar case, we add indentation based on `metalsConfig.snippetAutoIndent`, see here for an example:
 https://github.com/scalameta/metals/blob/25641a7ca25593a45243ade7eae084ec9e46c2a4/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala#L1300-L1306

tests/cross/src/test/scala/tests/pc/CompletionScaladocSuite.scala
gabro: nevermind, see my comments below

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1632,6 +1675,93 @@ trait Completions { this: MetalsGlobal =>
...
+                if (idx == 0 && clientSupportsSnippets)
+                  s"* @param ${valdef.name} $$0"
+                else s"* @param ${valdef.name}"
+              }
+            }
+        case clazz @ ClassDef(_, _, _, _) =>
+          // If the associated def is a class definition,
+          // retrieve the constructor from the class, and caluculate the lines
+          // from the constructor definition instead.
+          new ConstructorFinder(clazz).getConstructor match {
```
olafurpg: Did you try `clazz.symbol.primaryConstructor`? In general, when using compiler APIs you want to operate on types and symbols instead of trees (unlike scalameta where we have mostly just trees)

gabro:
FYI, I've just used this feature to author the Scaladoc in this PR 😄 

https://github.com/scalameta/metals/pull/1254/

tanishiking:
Thank you so much for feedbacks! @olafurpg @gabro 
It's still working in progress, but I'm going to address feedback in a few days :)

tests/cross/src/test/scala/tests/pc/CompletionScaladocSuite.scala
tanishiking: I think that's a good idea:+1: I'm going to implement it later!

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1632,6 +1675,93 @@ trait Completions { this: MetalsGlobal =>
...
+     */
+    case class Scaladoc(
+        editRange: l.Range,
+        associatedDef: MemberDef
+    ) extends CompletionPosition {
+      override def contribute: List[Member] = {
+        val lines = scaladocLines(associatedDef)
+        // add `* @return` line only if the definition is a method def.
+        val returnLine =
+          if (lines.isEmpty && clientSupportsSnippets) "* @return $0"
```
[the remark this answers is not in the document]
tanishiking: I meant to move the cursor to `@return | <- here` only if there are no parameters.
But now I'm going to move the cursor to the following position, we don't need this logic anymore.

```scala
/**
   * | <- here instead of after the @param
   *
   * @param
   */
```

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
tanishiking: Oh, I haven't tried it yet, I'm gonna give it a try!

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -669,6 +676,42 @@ trait Completions { this: MetalsGlobal =>
...
+    private var defs: List[MemberDef] = Nil
+
+    /**
+     * Collect all the member definitions whose position is
+     * below the given `pos`. And then return the closest member definiton.
+     */
+    def findAssociatedDef(root: Tree): Option[MemberDef] = {
+      defs = Nil
+      traverse(root)
+      defs.sortBy(_.pos.point).headOption
```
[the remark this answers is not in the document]
tanishiking: Thanks for your input! Done in https://github.com/scalameta/metals/pull/1250/commits/e36da2748d88988c311ad0d9fa03ef217de61995

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1632,6 +1675,93 @@ trait Completions { this: MetalsGlobal =>
...
+      override def contribute: List[Member] = {
+        val lines = scaladocLines(associatedDef)
+        // add `* @return` line only if the definition is a method def.
+        val returnLine =
+          if (lines.isEmpty && clientSupportsSnippets) "* @return $0"
+          else "* @return"
+        val newText =
+          if (associatedDef.isInstanceOf[DefDef])
+            (Seq("", "*") ++ lines ++ Seq(returnLine, "*/")).mkString("\n  ")
+          else (Seq("", "*") ++ lines ++ Seq("*/")).mkString("\n  ")
```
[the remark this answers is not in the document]
tanishiking: :+1: to `StringBuilder` and done in https://github.com/scalameta/metals/pull/1250/commits/e051eb5f80ea69b6b377c448705491de5a2ac836

mtags/src/main/scala/scala/meta/internal/pc/ScaladocCompletion.scala
```diff
@@ -0,0 +1,199 @@
...
+   *
+   * @param pos The cursor position
+   * @param text The original text of the source file.
+   */
+  protected def isScaladocCompletion(pos: Position, text: String): Boolean = {
+    try {
+      pos.isDefined &&
+      text.charAt(pos.point - 3) == '/' &&
+      text.charAt(pos.point - 2) == '*' &&
+      text.charAt(pos.point - 1) == '*'
```
gabro: maybe you can do

```scala
text.slice(pos.point - 3, 3) == "/**"
```

(I would double check the indexes :P )

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1632,6 +1675,93 @@ trait Completions { this: MetalsGlobal =>
...
+          // from the constructor definition instead.
+          new ConstructorFinder(clazz).getConstructor match {
+            case Some(defdef) => this.scaladocLines(defdef)
+            case scala.None => Nil
+          }
+        case _ => Nil
+      }
+    }
+
+    class ConstructorFinder(clazz: ClassDef) extends Traverser {
```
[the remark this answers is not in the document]
tanishiking: Done in https://github.com/scalameta/metals/pull/1250/commits/b6de5d98ad7ea7973b9363b3f72a4dc8c0b705ee 👍 
Is it ok to split those files in this way?

If it's ok, I'd like to split other modules into another file in another PR (like MatchKeyword Completion related logics into `MatchKeywordCompletion.scala`)(in this case, maybe we should dig a directory like `completions` and put those files in `pc/completions/ScaladocCompletion.scala`)

as mentioned in https://github.com/scalameta/metals/issues/672

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1867,4 +1997,12 @@ trait Completions { this: MetalsGlobal =>
     }
   }
 
+  def isScaladocCompletion(pos: Position, text: String): Boolean = {
```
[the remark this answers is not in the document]
tanishiking: Done:+1:  https://github.com/scalameta/metals/pull/1250/commits/82618d1e3529cee4adc76c461ee79a45ad13583f

tests/cross/src/test/scala/tests/pc/CompletionScaladocSuite.scala
```diff
@@ -197,4 +199,88 @@ object CompletionScaladocSuite extends BaseCompletionSuite {
        |}
        |""".stripMargin
   )
+
+  checkEdit(
+    "scaladoc-defdef-returns-unit",
```
gabro: since this is the `CompletionScaladocSuite` I think you can avoid the `scaladoc-` prefix in all tests

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
tanishiking: Thanks! I could simplify the logic with `clazz.symbol.primaryConstructor` :) in https://github.com/scalameta/metals/pull/1250/commits/4f936d3f99045a68e038ebcab5774bf599a24f4e

tests/cross/src/test/scala/tests/pc/CompletionScaladocSuite.scala
tanishiking: done in https://github.com/scalameta/metals/pull/1250/commits/f4e93f44cc33c1fb3ff0a670e17575be5ca9d9f6

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1632,6 +1675,93 @@ trait Completions { this: MetalsGlobal =>
...
+    /**
+     * Returns the parameter lines of scaladoc besed on the given memberDef
+     * like ["* @param param1 $0", "* @param param2"].
+     *
+     * @param memberDef The memberDef to construct scaladoc.
+     */
+    private def scaladocLines(memberDef: MemberDef): List[String] = {
+      memberDef match {
+        case DefDef(_, _, _, vparamss, _, _) =>
+          vparamss.flatten.zipWithIndex
```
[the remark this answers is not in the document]
tanishiking: Ah, I got something like 

```
/**
  *
  * @param evidence$1
  * @return
  */
def foo[T: Ordering] = ...
 ```

which is added around [here](https://github.com/scala/scala/blob/ba9701059216c629410f4f23a2175d20ad62484b/src/compiler/scala/tools/nsc/ast/parser/TreeBuilder.scala#L135).

Filtered out those synthetic params here https://github.com/scala/scala/blob/ba9701059216c629410f4f23a2175d20ad62484b/src/compiler/scala/tools/nsc/ast/parser/TreeBuilder.scala#L135 👍

File: mtags/src/main/scala/scala/meta/internal/pc/ScaladocCompletion.scala [hunk 22]
```diff
@@ -0,0 +1,199 @@
...
+        case clazz: ClassDef =>
+          // If the associated def is a class definition,
+          // retrieve the constructor from the class, and caluculate the lines
+          // from the constructor definition instead.
+          clazz.symbol.primaryConstructor.paramss.flatten
+            .filter(sym =>
+              // FIXME: All synthetic params should be filtered out by checking SYNTHETIC flag.
+              // evidence params for classs constructor look like they don't have SYNTHETIC flag.
+              !sym.isSynthetic &&
+                !sym.name.startsWith(nme.EVIDENCE_PARAM_PREFIX)
```
gabro: if you're happy with the workaround, I would change the `FIXME` into `NOTE`. We tend to avoid `FIXME` comments as much as possible (with a few exceptions that point to specific GitHub issues)

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -336,7 +336,7 @@ trait Completions { this: MetalsGlobal =>
   // variable but it avoids repeating traversals from the compiler
   // implementation of `completionsAt(pos)`.
   var lastVisistedParentTrees: List[Tree] = Nil
-  sealed abstract class CompletionPosition {
+  abstract class CompletionPosition {
```
tanishiking: Removed a `sealed` modifier to split `Scaladoc` related logic into another file.
I think we cannot avoid removing this for splitting each CompletionPosition logics into other files, and I think it's ok with removing this modifier 🤔

mtags/src/main/scala/scala/meta/internal/pc/ScaladocCompletion.scala
tanishiking: I believed that we can filter out all synthetic parameter like `evidence$1: Ordering[T]` which is appended here https://github.com/scala/scala/blob/ba9701059216c629410f4f23a2175d20ad62484b/src/compiler/scala/tools/nsc/ast/parser/TreeBuilder.scala#L135
But only with `isSynthetic` filter, I couldn't filter out `evidence$1` from params and [this test](https://github.com/scalameta/metals/pull/1250/files#diff-a92ad42d0d97b9389b28fec32531915dR267) failed 🤔 

I think I have to dig deeper about this phenomenon, but I just patched it with filtering by `EVIDENCE_PARAM_PREFIX` this for now.

mtags/src/main/scala/scala/meta/internal/pc/ScaladocCompletion.scala
olafurpg: The current solution is fine.

mtags/src/main/scala/scala/meta/internal/pc/ScaladocCompletion.scala
tanishiking: I'm not still 100% sure, but maybe this happens because the scalac type checker doesn't keep tree modifiers in sync with symbol flags for a class constructor. (actually, there's only `IMPLICIT` flag remains on the symbol of `evidence` on class constructor).

Maybe I can create an issue on `scala/bug` and fix (or maybe feature?) this,  but let me keep this workaround and note this behavior on a comment for now :)

---


- when a symbol for a method signature is created, modifiers that are defined as [ValueParameterFlags](https://github.com/scala/scala/blob/8514c3066ac8f71e676369e6163a1b2e6303f965/src/reflect/scala/reflect/internal/Flags.scala#L303-L307) survive during type checking.
  - Assigning modifiers to value parameters [here]( https://github.com/scala/scala/blob/4f04e38f1d0a74048b4ea350e7c7ff377b6fa3d4/src/compiler/scala/tools/nsc/typechecker/Namers.scala#L1332-L1336)
- but, it looks like type-checker doesn't use `ValueParameterFlags` as modifiers mask for a class constructor. Just assigning [inConstructorFlags]( https://github.com/scala/scala/blob/4f04e38f1d0a74048b4ea350e7c7ff377b6fa3d4/src/compiler/scala/tools/nsc/typechecker/Namers.scala#L409) (I'm not 100% sure...), and all modifiers from the tree don't remain in symbol definition.

FYI: The PR that makes `evidence` (for method definition symbol) SYNTHETIC is here https://github.com/scala/scala/pull/2132

gabro:
The `@constructor` tag was my idea (I've read it in the guide), but I have very little experience with Scaladoc. If it's not commonly used, then 👍 for not adding it.

tanishiking:
Thank you so much for the review and finishing up this PR @olafurpg @gabro !
I'm so excited to use this feature in the latest release 😄

Issue #1253, opened 2020-01-04
Exhaustive match does not fully-qualify shadowed symbols


gabro:
**Describe the bug**
Exhaustive match does not fully-qualify shadowed symbols

**To Reproduce**
This test fails

```
  checkEdit(
    "exhaustive-fully-qualify",
    """
      |package example
      |
      |object None
      |
      |object Main {
      |  Option(1) match@@
      |}""".stripMargin,
    """
      |package example
      |
      |object None
      |
      |object Main {
      |  Option(1) match {
      |\tcase Some(value) => $0
      |\tcase scala.None =>
      |}
      |}""".stripMargin,
    filter = _.contains("exhaustive")
  )
```

the actual result is

```scala
package example

import scala.None

object None

object Main {
  Option(1) match {
    case Some(value) =>
    case None =>
  }
}
```

which tries to auto-import `scala.None` (and it still produces a compile error)

**Expected behavior**
Fully-qualify `scala.None` in the pattern.

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: master

**Additional context**

**Search terms**
shadow, exhaustive match, auto-import

gabro:
This boils down to this call to `context.nameIsInScope` returning `false` when I would expect it to return `true` because `None` is in scope:

https://github.com/scalameta/metals/blob/597898e65d756622e282ad2f80642802155c218c/mtags/src/main/scala/scala/meta/internal/pc/Signatures.scala#L75

By the way, when selecting `None` using scope completions it fully qualifies it correctly.

@olafurpg any pointer?

PR #1254, opened 2020-01-05
Failing test case for #1253

gabro:
Following an offline conversation with @olafurpg, this is a stab at "failing tests", inspired from https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#failing-tests

### DIFF
--- a/tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
+++ b/tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
@@ -49,7 +49,7 @@ abstract class BaseCompletionSuite extends BasePCSuite {
   }
 
   def checkEditLine(
-      name: String,
+      name: TestOptions,
       template: String,
       original: String,
       expected: String,
--- a/tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
+++ b/tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
@@ -70,17 +70,16 @@ abstract class BaseCompletionSuite extends BasePCSuite {
   }
 
   def checkEdit(
-      name: String,
+      name: TestOptions,
       original: String,
       expected: String,
       filterText: String = "",
       assertSingleItem: Boolean = true,
       filter: String => Boolean = _ => true,
       command: Option[String] = None,
-      compat: Map[String, String] = Map.empty,
-      expectFailure: Boolean = false
+      compat: Map[String, String] = Map.empty
   )(implicit filename: sourcecode.File, line: sourcecode.Line): Unit = {
-    test(name, expectFailure) {
+    test(name) {
       val items = getItems(original).filter(item => filter(item.getLabel))
       if (items.isEmpty) fail("obtained empty completions!")
       if (assertSingleItem && items.length != 1) {
--- a/tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
+++ b/tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
@@ -68,7 +68,9 @@ abstract class BaseCompletionSuite extends BasePCSuite {
       command = command
     )
   }
-  def checkEdit(
+
+  private def checkEditCore(
+      testMethod: (String) => (=> Unit) => Unit,
       name: String,
       original: String,
       expected: String,
@@ -78,7 +80,7 @@ abstract class BaseCompletionSuite extends BasePCSuite {
       command: Option[String] = None,
       compat: Map[String, String] = Map.empty
   )(implicit filename: sourcecode.File, line: sourcecode.Line): Unit = {
-    test(name) {
+    testMethod(name) {
       val items = getItems(original).filter(item => filter(item.getLabel))
       if (items.isEmpty) fail("obtained empty completions!")
       if (assertSingleItem && items.length != 1) {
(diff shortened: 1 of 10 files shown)


tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -184,4 +184,29 @@ object CompletionMatchSuite extends BaseCompletionSuite {
       "2.11" -> "match"
     )
   )
+
+  // https://github.com/scalameta/metals/issues/1253
+  checkEditBug(
```
gabro: this test passes because it fails 🙃

tests/mtest/src/main/scala/tests/BaseSuite.scala
```diff
@@ -149,6 +149,11 @@ class BaseSuite extends TestSuite {
       myTests += FlatTest(name, () => fun)
     }
   }
+
+  def failingTest(name: String)(fun: => Any): Unit = {
+    test(name)(intercept[TestFailedException](fun))
```
gabro: I'm tempted to do

```scala
    test(utest.ufansi.Color.LightRed(s"EXPECTED FAILURE - $name").toString()) {
      intercept[TestFailedException](fun)
    }
```

and obtain

image


but then it becomes hard to select a single test to run.

olafurpg: I agree it’s better to leave the test name unchanged.

tests/mtest/src/main/scala/tests/BaseSuite.scala
```diff
@@ -149,6 +149,11 @@ class BaseSuite extends TestSuite {
       myTests += FlatTest(name, () => fun)
     }
   }
+
+  def failingTest(name: String)(fun: => Any): Unit = {
```
olafurpg: “testFailing”?

tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
```diff
@@ -101,6 +103,52 @@ abstract class BaseCompletionSuite extends BasePCSuite {
     }
   }
 
+  def checkEdit(
```
olafurpg: This signature is so complicated it’s probably better to keep adding more parameters instead of duplicating the method. What do you think?

tests/mtest/src/main/scala/tests/BaseSuite.scala
```diff
@@ -158,6 +163,15 @@ class BaseSuite extends TestSuite {
     }
   }
 
+  def failingAsyncTest(name: String, maxDuration: Duration = Duration("10min"))(
```
olafurpg: I think we can avoid some duplicate logic by encoding the failure bit where we collect the tests

tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
gabro: yeah, I agree. I was uneasy too with the amount of duplication

gabro:
@olafurpg new attempt, which just parametrizes `test` and `testAsync` (since we rarely use them directly, it probably makes more sense this way than exposing another "primitive")

gabro:
Here's how RenameLspSuite could be changed to "document" tests that are expected to fail https://github.com/scalameta/metals/commit/c9cc928d6b5007ea791c0380ad1c641631c6e34e /cc @tgodzik

tests/mtest/src/main/scala/tests/BaseSuite.scala
```diff
@@ -144,15 +144,26 @@ class BaseSuite extends TestSuite {
 
   def isTestSuiteEnabled: Boolean = true
 
-  def test(name: String)(fun: => Any): Unit = {
+  def test(name: String, expectFailure: Boolean = false)(fun: => Any): Unit = {
```
olafurpg: This is an improvement! I am not sure it's a good idea to add such a specific parameter to this signature however. I have wanted to add tag support down the road so that we can classify some tests for example as flaky. What do you think about this approach instead?
```suggestion
class Tag
case object ExpectFailure extends Tag
  def test(name: String, tags: Tag*)(fun: => Any): Unit = {
    val isExpectFailure = tags.contains(ExpectFailure)
```

olafurpg: Another design I have thought of could be to replace `name: String` with a custom data structure

```scala
case class TestData(name: String, isExpectedFailure: Boolean, tags: List[Tag]) {
  def isFailing: TestData = copy(isExpectedFailure = true)
  def tag(t: Tags): TestData.copy(tags = t :: tags)
}
object TestData {
  implicit def generate(name: String): TestData = TestData(name, false, Nil)
}
def test(data: TestData)(fun: => Unit): Unit = ...
```

The implicit conversion will make existing tests continue to work unchanged and we can more easily introduce new APIs to tweak the test metadata

```scala
test("issue-42".isFailing.tag(Flaky)) { ... }
```

olafurpg: Just validated this compiles and works as expected
Screenshot 2020-01-05 at 17 17 27

gabro: I like the design! I've submitted a proposal that implements it

tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
```diff
@@ -68,8 +68,9 @@ abstract class BaseCompletionSuite extends BasePCSuite {
       command = command
     )
   }
+
   def checkEdit(
-      name: String,
+      name: TestOptions,
```
gabro: tests suites that need tagging support, can gradually make this conversion

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -184,4 +184,29 @@ object CompletionMatchSuite extends BaseCompletionSuite {
       "2.11" -> "match"
     )
   )
+
+  // https://github.com/scalameta/metals/issues/1253
+  checkEdit(
+    "exhaustive-fully-qualify".expectedToFail,
```
gabro: here's an example. You could technically also do 

```suggestion
    "exhaustive-fully-qualify".tag(Tag.ExpectedFailure),
```

but since we'll reasonably have a few tags, I opted for a utility "DSL" method

tests/unit/src/test/scala/tests/CompressionSuite.scala
```diff
@@ -6,7 +6,8 @@ import scala.meta.internal.metals.ClassfileElementPart
 object CompressionSuite extends BaseSuite {
   def checkRoundtrip(a: String): Unit = {
     val nonempty = a.trim.linesIterator.map(_.trim).filterNot(_.isEmpty).toArray
-    test(nonempty.headOption.getOrElse("<empty>")) {
+    val testName = nonempty.headOption.getOrElse("<empty>")
+    test(testName) {
```
gabro: this was needed in order to please the type checker, which was inferring `Serializable` when inlined.

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
olafurpg: I think it's fine to have a method here to ease discovery.

tests/unit/src/test/scala/tests/CompressionSuite.scala
olafurpg: What 😟 That's surprising. Looks like a compiler bug in that case. Is it worth adding a `def test(name: String, tags: Tag*)` overload to avoid similarly confusing failures for basic cases?

gabro:
Failing test is not related

PR #1255, opened 2020-01-05
typo - evaluating

ckipp01:
Screenshot 2020-01-05 at 21 35 12

_why is this only finding one occurance... I swear I saw more... wait a second_


gabro:
The metals/status notification from the screenshot looks wrong too!

gabro:
Ah no wait, this is it! Sorry 😅

Issue #1190, opened 2019-12-16
Improve experience of upgrading Bloop 


olafurpg:
Currently, Metals explains to the user that that there's an old version of Bloop running on the computer. However, this message doesn't explain how to fix the problem.

It would be nice if the message explained more details how to upgrade, something like:

> You have Bloop v1.3.4 installed and Metals requires at least Bloop v1.4.0.
> If you installed bloop via a system package manager (brew, aur, scoop), please upgrade manually.
> If not, select "Turn off old server".
> Options: I'll update manually | Turn off old server | Not now

ckipp01:
I played around with this today, and I'm curious about your thoughts on how you'd like this to actually work. I downloaded and ran an old version of bloop, and then tried to import a project. As expected, I see this.

<img width="757" alt="Screenshot 2019-12-20 at 14 54 16" src="https://user-images.githubusercontent.com/13974112/71259578-09a2e580-2339-11ea-9fb5-1655c092f71a.png">

After clicking more info, I'm then brought to Doctor, where I see the relevant information.

<img width="746" alt="Screenshot 2019-12-20 at 14 58 49" src="https://user-images.githubusercontent.com/13974112/71259754-70c09a00-2339-11ea-8d4e-c14a1b6de74b.png">

So, in this scenario, would those options you have listed...
> Options: I'll update manually | Turn off old server | Not now
appear in Doctor?

Also, what would be the difference between `I'll update manually`, and `Not Now`?

olafurpg:
@ckipp01 those options would appear in the original `window/showMessageRequest`. Instead of pointing people to the doctor, we let the users fix the problem right away in the request.

> Also, what would be the difference between I'll update manually, and Not Now?

None. By convention, we try to always keep a "Not now" option that's safe to choose but comes at the cost of degraded user experience. We could drop the "I'll update manually" option but it wouldn't be so clear that's how the user fixes the problem 🤔

PR #1256, opened 2020-01-06
Restart Bloop automatically if newer version is needed

tgodzik:
Previously, when detected that running build server is using an old Bloop version, we would just display a warning without explaining exactly what to do.

Now, we suggest to restart the Bloop server if Metals uses a newer version and this restart can be requested by the user.

Fixes https://github.com/scalameta/metals/issues/1190


metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -29,6 +31,25 @@ final class BloopServers(
...
+      dummyIn,
+      System.out,
+      System.err,
+      Shell.default
+    )
+    val result = cli.run(Array("ng-stop")) == 0
+    if (!result) {
+      scribe.warn("There were issues stopping the Bloop server.")
+      scribe.warn(
+        "If it doesn't start back up you can run `build-restart` command manually."
```
ckipp01:
```suggestion
        "If it doesn't start back up you can run the `build-restart` command manually."
```

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -28,6 +28,14 @@ final class Doctor(
   private val hasProblems = new AtomicBoolean(false)
   private var bspServerName: Option[String] = None
   private var bspServerVersion: Option[String] = None
+  val minimumBloopVersion = "1.3.5"
```
ckipp01: I know this was in here before, but is there a way we can just pull this from the version that we're using rather than having it hardcoded? 🤔

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -138,6 +138,37 @@ class Messages(icons: Icons) {
...
+    def dismissForever: MessageActionItem =
+      new MessageActionItem("Don't show again")
+    def params(
+        bloopVersion: String,
+        minimumBloopVersion: String
+    ): ShowMessageRequestParams = {
+
+      val params = new ShowMessageRequestParams()
+      params.setMessage(
+        s"""|You have a Bloop $bloopVersion installed and Metals requires at least Bloop $minimumBloopVersion.
```
ckipp01:
```suggestion
        s"""|You have Bloop $bloopVersion installed and Metals requires at least Bloop $minimumBloopVersion.
```

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: I moved it to build.sbt. It might actually make sense to tell users to use the same one that we have in the project, since we upgrade the version when we need it. Good idea!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -29,6 +31,25 @@ final class BloopServers(
...
+  def shutdownServer(): Boolean = {
+    val dummyIn = new ByteArrayInputStream(new Array(0))
+    val cli = new BloopgunCli(
+      BuildInfo.bloopVersion,
+      dummyIn,
+      System.out,
+      System.err,
+      Shell.default
+    )
+    val result = cli.run(Array("ng-stop")) == 0
```
olafurpg: What is the difference between ngstop and exit?

tgodzik: I think that is the underlying nailgun command, might be better to use exit. I was testing it previously and it didn't seem to work with exit, but that might have been a different thing.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: fixed

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: Switched to current Bloop version.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
tgodzik: Fixed

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: Switched to `exit` since it works with it and that is a Bloop command.

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
```diff
@@ -1,8 +1,13 @@
 package scala.meta.internal.semver
 
 object SemVer {
+
   def isCompatibleVersion(minimumVersion: String, version: String): Boolean = {
-    (minimumVersion.split('.'), version.split('.')) match {
+
+    def splitVersion(v: String) =
+      v.replaceAll("(-|\\+).+$", "").split('.')
```
tgodzik: Fixed it to work with `1.4.0-RC` etc. and snapshot versions.

tests/unit/src/main/scala/scala/meta/internal/metals/debug/TestDebugger.scala
```diff
@@ -78,7 +78,7 @@ final class TestDebugger(
   def shutdown: Future[Unit] = {
     for {
       _ <- terminated.future
-      _ <- debugger.shutdown
+      _ <- debugger.shutdown(60)
```
tgodzik: The test on windows were timing out and they usually are slower.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1653,6 +1658,26 @@ class MetalsLanguageServer(
       .foreach(_ => languageClient.refreshModel())
   }
 
+  private def checkRunningBloopVersion(bspServerVersion: String) = {
```
olafurpg: nit: should we move this into a separate file?

tgodzik: I think we should do it as a larger refactor of the file. I can see a number of similar methods that might be put there.

PR #1272, opened 2020-01-08
Fix some incorrect range-encloses checks and add tests for them

cb372:
* Fix the logic in various "does this range enclose that position/range" checks, particularly for multi-line ranges
* Add unit tests
* Minor refactoring: renaming; combining two implicit classes into one

### DIFF
--- a/tests/unit/src/test/scala/tests/MtagsEnrichmentsSuite.scala
+++ b/tests/unit/src/test/scala/tests/MtagsEnrichmentsSuite.scala
@@ -1,6 +1,7 @@
 package tests
 
 import org.eclipse.{lsp4j => l}
+import scala.meta.internal.{semanticdb => s}
 import scala.meta.internal.mtags.MtagsEnrichments._
 
 object MtagsEnrichmentsSuite extends BaseSuite {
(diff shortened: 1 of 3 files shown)


tests/unit/src/test/scala/tests/MtagsEnrichmentsSuite.scala
```diff
@@ -0,0 +1,89 @@
+package tests
+
+import org.eclipse.{lsp4j => l}
+import scala.meta.internal.{semanticdb => s}
+import scala.meta.internal.mtags.MtagsEnrichments._
+
+object MtagsEnrichmentsSuite extends BaseSuite {
+
+  test("XtensionLspRange.encloses - single line") {
```
ckipp01: I may be wrong on this, but doesn't the defining the test name like this then not allow the user to target a single test with `testOnly`? For example I tried `unit/testOnly tests.MtagsEnrichmentsSuite."XtensionLspRange.encloses - single line"`, which won't work. Is there a way to do this with a space? Either way, it might be easier to follow the convention of the other tests with shorter dash seperated names in order to more easily target them.

cb372: Ah, I wasn't aware of the naming convention. I've renamed the tests.

PR #1276, opened 2020-01-09
Add caching to github actions for Cousier

tgodzik:
After recent changes it should now be possible to cache up to 2 GB, so we should be fine to cache the Coursier directories.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,27 @@ jobs:
       - uses: olafurpg/setup-scala@v7
         with:
           java-version: "adopt@1.${{ matrix.java }}"
+      - uses: actions/cache@v1
+        if: startsWith(runner.os, 'Linux')
+        with:
+          path: ~/.cache/coursier/v1
+          key: ${{ runner.os }}-coursier-${{ hashFiles('**/*.sbt') }}
+          restore-keys: |
+            ${{ runner.os }}-coursier-
+      - uses: actions/cache@v1
+        if: startsWith(runner.os, 'Windows')
+        with:
+          path: ~\AppData\Local\Coursier\Cache\v1
+          key: ${{ runner.os }}-coursier-${{ hashFiles('**/*.sbt') }}
+          restore-keys: |
+            ${{ runner.os }}-coursier-
+      - uses: actions/cache@v1
+        if: startsWith(runner.os, 'macOS')
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -23,6 +23,27 @@ jobs:
       - uses: olafurpg/setup-scala@v7
         with:
           java-version: "adopt@1.${{ matrix.java }}"
+      - uses: actions/cache@v1
```
olafurpg: How about using `COURSIER_CACHE=$HOME/.cache/coursier` so we can use the same directory for all operating systems?

tgodzik: I was thinking about it, will change before going out of draft.

tgodzik:
We should also maybe cache `.sbt` directory, anything else that anyone can think about? 

I am also thinking that it might not make sense to have separate caches for each operating system.

tgodzik:
Might not be ready to review - just trying out caching.

.github/workflows/ci.yml
```diff
@@ -18,97 +18,83 @@ jobs:
...
+  integration:
+    name: ${{ matrix.name }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        type:
+          [sbt, maven, gradle, pants, mill, feature, cross, scalafmt, scalafix]
+        include:
+          - type: sbt
```
tgodzik: It should be easy now to split the unit tests, but I don't want to do too much in this PR. Will follow up later

tgodzik:
> Do we have an idea for how much this speeds up the CI? What is the workflow for clearing the cache?

Clearing the cache is not possible yet: https://github.com/actions/cache/issues/2  
Though it's only coursier and sbt that's being cached - should be minimal risk. We can always remove caching if it proves problematic.

> I'm a bit uneasy about adding caching since it adds one possible cause for flaky failures. My impression is that downloads are not the bottleneck in the CI builds. The slowest jobs are running `unit/test`, which is mostly bottlenecked by `*LspSuite`

It's not a crazy speedup for sure. Something less than a minute. However, caching will be more useful when we split the tests into multiple runs - that's my plan after the changes in the mtests PR. So altogether the speed up should be good and I think useful.

Overall, I would opt to use it, but caching was not the thing that I spend the most in this PR, so if you super object to it I will not argue :smile:

PR #1284, opened 2020-01-11
Don't insert parentheses when importing

### LINKED ISSUE #1281: Import completions for methods should not insert parentheses
Seems to be related to #251, but in reverse.

Edit: oops, didn't notice that's the old completions.

**Describe the bug**

When importing a method with parameters, parentheses are inserted. That's illegal in Scala and shouldn't happen.

**To Reproduce**
1. Create an object with a method taking parameters
2. Import that object's method in another object
3. See the parens appear

Here's an example of a file exhibiting this:

```scala
package example

object demo {
  def method(a: Int) = a
}

object usage {
  import demo.m<tab>
}
```

**Expected behavior**
No parentheses are inserted in imports.

**Screenshots**

image

image

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: 0.7.6+720-7b699bd6-SNAPSHOT

**Search terms**
parentheses, suggestions, completions, imports, import

kubukoz:
Previously, if completions were used for importing a method that has parameters, its name would appear with parentheses. Now, the parentheses won't be inserted if the completion is requested in an import statement.

Will close #1281.

~Leaving this as a draft for now, as I wasn't able to make tests fail locally, hopefully I can confirm what's going on in CI.~

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
@@ -44,23 +44,28 @@ class CompletionProvider(
       cursor = Some(params.offset),
       cursorName = cursorName
     )
+
     val pos = unit.position(params.offset)
     val isSnippet = isSnippetEnabled(pos, params.text())
-    val clientSupportsSnippets =
-      compiler.metalsConfig.isCompletionSnippetsEnabled()
+
     val (i, completion, editRange, query) = safeCompletionsAt(pos)
+
     val start = inferIdentStart(pos, params.text())
     val end = inferIdentEnd(pos, params.text())
     val oldText = params.text().substring(start, end)
     val stripSuffix = pos.withStart(start).withEnd(end).toLSP
+
     def textEdit(newText: String) = {
       if (newText == oldText) new l.TextEdit(stripSuffix, newText)
       else new l.TextEdit(editRange, newText)
     }
+
     val history = new ShortenedNames()
+
     val sorted = i.results.sorted(memberOrdering(query, history, completion))
     lazy val importPosition = autoImportPosition(pos, params.text())
     lazy val context = doLocateImportContext(pos, importPosition)
+
     val items = sorted.iterator.zipWithIndex.map {
       case (r, idx) =>
         params.checkCanceled()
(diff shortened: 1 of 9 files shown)


mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -39,44 +39,49 @@ class CompletionProvider(
       cursor = Some(params.offset),
       cursorName = cursorName
     )
+
     val pos = unit.position(params.offset)
     val isSnippet = isSnippetEnabled(pos, params.text())
-    val clientSupportsSnippets =
```
ckipp01: I believe this is needed a bit further down to get the correct suffixes?

kubukoz: We have access to it already from a val in a trait that's mixed in here (`Completions#clientSupportsSnippets`). I believe it's the same value...

mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -144,10 +149,11 @@ class CompletionProvider(
                 item.setAdditionalTextEdits(edits.asJava)
                 item.setTextEdit(textEdit(short + suffix))
             }
-          case _ =>
+          case _ if importPosition.isEmpty =>
```
kubukoz: I'm not entirely sure this is accurately coupled with "the current position is an import"... could there be any other case in which importPosition is empty?

mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
ckipp01: Ahh, I didn't notice that. Cool!

mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
olafurpg: Judging by the implementation of `autoImportPosition` it only returns `None` for import expressions, but there's nothing in the contract of that method which prevents this implementation detail from changing in the future. It would be better do something custom instead which implements exactly what we want for this purpose https://gist.github.com/olafurpg/d15833f9b519ac286e36a5daad8f07a5

kubukoz: I'll give this a try :)

PR #1289, opened 2020-01-12
Update recommended mappings with the latest

ckipp01:
There has been some updates to the recommended mappings both from `coc.nvim` and a change in the way I do expands for worksheets.

While changing this up, I noticed this is a huge chunk of stuff to copy here in the docs. In the readme of `coc-metals` I ended up taking this out and instead just linking to a [recommended mapping file](https://github.com/scalameta/coc-metals/blob/master/coc-mappings.vim). Do you think that would be a good idea here to not clog up the docs so much, or do you think it's a good idea having this all right in the docs on the site?

### DIFF
--- a/docs/editors/vim.md
+++ b/docs/editors/vim.md
@@ -249,7 +249,7 @@ Update the `metals.sbtScript` setting to use a custom `sbt` script instead of th
 default Metals launcher if you need further customizations like reading environment
 variables.
 
-![Sbt Launcher](https://i.imgur.com/kbxNKzI.png)
+![Sbt Launcher](https://i.imgur.com/meciPTg.png)
 ```
 
 ## Configure Java version
@@ -259,7 +259,7 @@ The `coc-metals` extension uses by default the `JAVA_HOME` environment variable 
 
 If no `JAVA_HOME` is detected you can then Open Settings by following the instructions or do it at a later time by using `:CocConfig` or `:CocConfigLocal` which will open up your configuration where you can manually enter your JAVA_HOME location.
 
-![Enter Java Home](https://i.imgur.com/wVThrMq.png)
+![Enter Java Home](https://i.imgur.com/wK07Vju.png)
 
 ## Using latest Metals SNAPSHOT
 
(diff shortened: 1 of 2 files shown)


olafurpg:
We can automatically in-line the contents of a file. Would that help?

ckipp01:
> We can automatically in-line the contents of a file. Would that help?

Possibly, but I guess it's not super important. Thinking about it some more, maybe it is best to keep them separate, and have the one on the website document for the latest stable, the one located in master can contain stuff not yet in stable.

docs/editors/vim.md
```diff
@@ -375,16 +422,12 @@ This step cleans up resources that are used by the server.
...
-
-- [`vim-lsc`](https://github.com/natebosch/vim-lsc/): simple installation and
-  low resource usage but limited functionality (no auto-import, cancellation,
-  formatting, folding).
-- [`LanguageClient-neovim`](https://github.com/autozimu/LanguageClient-neovim/):
-  client written in Rust.
-- [`vim-lsp`](https://github.com/prabirshrestha/vim-lsp): simple installation
-  but limited functionality (no auto-import, cancellation and no prompt for
-  build import).
+with these alternative LSP clients. Keep in mind that they have varying levels of LSP support.
```
ckipp01: I was working with `vim-lsc` this morning and realized these points weren't really accurate anymore since a lot of functionality has been updated. Same with `vim-lsp`. I think it might be best to just list what they are, and not really mention features for each one, but rather put a general note to keep in mind that there are varying levels of support.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -110,9 +110,7 @@ object MetalsServerConfig {
         )
       case "vim-lsc" =>
         base.copy(
-          // window/logMessage output is always visible and non-invasive in vim-lsc
-          statusBar = StatusBarConfig.logMessage,
-          // Not strictly needed, but helpful while this integration matures.
+          statusBar = StatusBarConfig.showMessage,
```
ckipp01: I was testing this today and realized that with `logMessage` things are correctly getting logged, but the user is getting zero feedback about what it happening. I think it's much more preferable to have this set to `showMessage`.

ckipp01: I use this some more today, and I'm going to revert this. While it's hard without getting feedback, with this being set this way, the user receives way too much. .

PR #1307, opened 2020-01-14
Use actual mill version in predef

jodersky:
As of recent versions of mill, the `$ivy` importers support
substituting `$MILL_VERSION` with the current version of mill.

The changes presented here modify the predef import to use the literal
`$MILL_VERSION` instead of a the value read from mill's version
file. While both these values are usually equivalent, they can differ
when using a local mill wrapper script (e.g. when developing and
testing mill locally).

In essence, the changes ensure that the correct bloop plugin will
always be used.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
@@ -50,18 +50,18 @@ case class MillBuildTool(userConfig: () => UserConfiguration)
   override def digest(workspace: AbsolutePath): Option[String] =
     MillDigest.current(workspace)
 
-  override def minimumVersion: String = "0.4.0"
+  override def minimumVersion: String = "0.6.0"
 
   override def recommendedVersion: String = version
 
-  override def version: String = "0.5.2"
+  override def version: String = "0.6.0"
 
   override def toString(): String = "Mill"
 
   def executableName = "mill"
 
   private def predefScript(millVersion: String) =
-    s"import $$ivy.`com.lihaoyi::mill-contrib-bloop:$millVersion`".getBytes()
+    "import $ivy.`com.lihaoyi::mill-contrib-bloop:$MILL_VERSION`".getBytes()
 
   private def predefScriptPath(millVersion: String): Path = {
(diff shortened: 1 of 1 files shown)


jodersky:
Looks like this is blocked until bloop integration with mill is fixed. I.e. until lihaoyi/mill#756 is merged

tgodzik:
The tests are failing though
```
INFO  java.lang.NoClassDefFoundError: io/circe/RootEncoder
INFO  	at mill.contrib.bloop.BloopImpl$BloopOps$bloop$.$anonfun$config$3(BloopImpl.scala:81)
INFO  	at scala.collection.mutable.HashMap.getOrElseUpdate(HashMap.scala:86)
INFO  	at mill.moduledefs.Cacher.cachedTarget(Cacher.scala:12)
INFO  	at mill.moduledefs.Cacher.cachedTarget$(Cacher.scala:10)
INFO  	at mill.define.Module.cachedTarget(Module.scala:18)
INFO  	at mill.contrib.bloop.BloopImpl$BloopOps$bloop$.config(BloopImpl.scala:81)
INFO  	at mill.contrib.bloop.BloopImpl$BloopOps$bloop$.$anonfun$writeConfig$1(BloopImpl.scala:86)
INFO  	at scala.collection.mutable.HashMap.getOrElseUpdate(HashMap.scala:86)
```
@jodersky any idea what is happening here?

ckipp01:
> The tests are failing though

I believe this is because https://github.com/lihaoyi/mill/pull/756 is merged, but it will need to be released before this will work as expected.

krotton:
The fix is currently released in a Mill's snapshot, so changing to:

```scala
override def version: String = "0.5.9-6-484657"
```

fixes the issue immediately.

jodersky:
> I believe this is because lihaoyi/mill#756 is merged, but it will need to be released before this will work as expected.

Indeed. I think the best course of action is to wait until a new "blessed" mill release is out (I reckon it will be 0.5.10). Once it's out I'll update the minimum version in this PR.

jodersky:
A new stable release of mill is out, I updated the version settings and pushed again. Hopefully this time the build will pass.

PR #1309, opened 2020-01-15
Update screen-record of Scaladoc auto-completion in v0.8.0 release-note.

tanishiking:
This PR updates the screen-record of the Scaladoc auto-completion section in v0.8.0's release note.

The previous screen record came from the [Pull Request](https://github.com/scalameta/metals/pull/1250), but it is a bit different from real behavior: we don't add `* @constructor` line for scaladoc on a class definition. Sorry for the confusion!
(the behavior had updated in https://github.com/scalameta/metals/pull/1250/commits/cf8e6070092ac693251e01568b009a7f1bfce601, but I didn't update the screen-record.) 

### before
scaladoc

### after
scaladoc


tgodzik:
@tanishiking Thanks! I missed that!

PR #1310, opened 2020-01-15
Add an override for build tool plugin version and promote LSP settings

tgodzik:
Previously, there was not way to override all of the build plugins Bloop versions. Now we add a unified setting for doing so - one setting should be enough since for each workspace we can only use one build tool. 

Or do we want to have a separate setting for each build tool? That would be for 3 different ones: Sbt, Gradle and Maven.

I also promoted some setting to LSP ones.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
@@ -160,6 +160,26 @@ class Messages(icons: Icons) {
       params
     }
   }
+  object BloopVersionChange {
+    def reconnect: MessageActionItem =
+      new MessageActionItem("Restart Bloop")
+    def notNow: MessageActionItem =
+      new MessageActionItem("Not now")
+    def params(version: String): ShowMessageRequestParams = {
+      val params = new ShowMessageRequestParams()
+      params.setMessage(
+        s"Bloop version was updated, do you want to kill the running server and start Bloop $version?"
+      )
+      params.setType(MessageType.Warning)
+      params.setActions(
+        List(
+          reconnect,
+          notNow
+        ).asJava
+      )
+      params
+    }
+  }
 
   object IncompatibleBloopVersion {
     def manually: MessageActionItem =
@@ -170,15 +190,20 @@ class Messages(icons: Icons) {
       new MessageActionItem("Don't show again")
     def params(
         bloopVersion: String,
-        minimumBloopVersion: String
+        minimumBloopVersion: String,
+        isChangedInSettings: Boolean
     ): ShowMessageRequestParams = {
 
       val params = new ShowMessageRequestParams()
+      val additional =
+        if (isChangedInSettings)
+          "You will also need to remove or update the `bloopVersion` setting"
+        else ""
       params.setMessage(
         s"""|You have Bloop $bloopVersion installed and Metals requires at least Bloop $minimumBloopVersion.
             |If you installed bloop via a system package manager (brew, aur, scoop), please upgrade manually.
             |If not, select "Turn off old server". A newer server will be started automatically afterwards.
-            |""".stripMargin
+            |""".stripMargin + s"\n$additional"
       )
       params.setType(MessageType.Warning)
       params.setActions(
--- a/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
@@ -168,7 +168,7 @@ class Messages(icons: Icons) {
     def params(version: String): ShowMessageRequestParams = {
       val params = new ShowMessageRequestParams()
       params.setMessage(
-        s"Bloop version was updated, do you want to kill the running server and start Bloop $version?"
+        s"Bloop version was updated, do you want to restart the running Bloop server?"
       )
       params.setType(MessageType.Warning)
       params.setActions(
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -829,6 +829,8 @@ class MetalsLanguageServer(
                 case item if item == Messages.BloopVersionChange.reconnect =>
                   bloopServers.shutdownServer()
                   autoConnectToBuildServer().ignoreValue
+                case _ =>
+                  Future.successful(())
               }
           } else if (userConfig.pantsTargets != old.pantsTargets) {
             slowConnectToBuildServer(forceImport = false).ignoreValue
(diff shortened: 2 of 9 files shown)


metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -105,6 +107,20 @@ object UserConfiguration {
         |`src/main/scala:: src/main/java::`. Syntax such as `src/{main,test}::`
         |is not supported.
         |""".stripMargin
+    ),
+    UserConfigurationOption(
+      "bloop-generate-sbt",
```
olafurpg: I'm on the fence about this setting since the plugin file is required for "import build" to work. I feel like we should instead have some setting to disable "import build" entirely, for users who prefer to do it manually 🤔 Any thoughts?

Alternatively, we should have an option to use a global sbt-bloop plugin instead of project-local. But I'm not sure that's a good direction.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -105,6 +107,20 @@ object UserConfiguration {
...
+      """true""",
+      "true",
+      "Generate Bloop plugin file",
+      """Option that can specified to false if the user doesn't want to generate an sbt file adding the Bloop sbt plugin."""
+    ),
+    UserConfigurationOption(
+      "bloop-plugin-version",
```
olafurpg: What about?
```suggestion
      "bloop-version",
```

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
tgodzik: > I'm on the fence about this setting since the plugin file is required for "import build" to work. I feel like we should instead have some setting to disable "import build" entirely, for users who prefer to do it manually thinking Any thoughts?

If someone has a Bloop plugin inside their workspace already and they really don't want to generate the Sbt file, I think we should let them. They would still want to have an automatic import on sbt file changes. And it's a really clear setting that I don't think anyone will click unless they really want to.

>Alternatively, we should have an option to use a global sbt-bloop plugin instead of project-local. But I'm not sure that's a good direction.

That would be just back to the previous behavior and from what I understand it wouldn't play well with offloading.

Overall, I think this would be needed, because we are basically modifying the user's build definition and we should have an option of user telling us that they don't want it.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
tgodzik: The user might think it will also change the Bloop server version, and that should be done manually by them. 

Or maybe that setting should also apply to the server version? What do you think?

olafurpg: Makes sense

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
olafurpg: Can we rename the setting to something like “I already have bloop installed”?

tgodzik: Changed to `bloop-sbt-already-installed`

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
tgodzik: Ok, so now it's possible to change Bloop version and restart the server when the setting change.

I think that's a pretty neat feature especially if a user wants to leverage new Bloop and Metals wer enot yet released. What do you think?

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -160,6 +160,26 @@ class Messages(icons: Icons) {
...
   }
+  object BloopVersionChange {
+    def reconnect: MessageActionItem =
+      new MessageActionItem("Restart Bloop")
+    def notNow: MessageActionItem =
+      new MessageActionItem("Not now")
+    def params(version: String): ShowMessageRequestParams = {
+      val params = new ShowMessageRequestParams()
+      params.setMessage(
+        s"Bloop version was updated, do you want to kill the running server and start Bloop $version?"
```
olafurpg: ?
```suggestion
        "Bloop version was updated, do you want to restart the running Bloop server?"
```
I'm not sure we need to mention `version`

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -819,7 +819,18 @@ class MetalsLanguageServer(
...
           }
-          if (userConfig.pantsTargets != old.pantsTargets) {
+          if (userConfig.bloopVersion != old.bloopVersion) {
+            languageClient
+              .showMessageRequest(
+                Messages.BloopVersionChange.params(userConfig.bloopVersion)
+              )
+              .asScala
+              .flatMap {
+                case item if item == Messages.BloopVersionChange.reconnect =>
```
olafurpg: This will crash with `MatchError` when the user doesn't want to reconnect.

tgodzik: Thanks, fixed!

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
tgodzik: Fixed

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -105,6 +107,21 @@ object UserConfiguration {
         |`src/main/scala:: src/main/java::`. Syntax such as `src/{main,test}::`
         |is not supported.
         |""".stripMargin
+    ),
+    UserConfigurationOption(
+      "bloop-sbt-already-installed",
+      "false",
+      "false",
+      "Don't generate Bloop plugin file for Sbt",
```
olafurpg: For better or worse, the correct capitalization for sbt is always lowercase, even when starting new sentences.
```suggestion
      "Don't generate Bloop plugin file for sbt",
```

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -105,6 +107,21 @@ object UserConfiguration {
...
+      "Option that can specified to true if the user doesn't want to generate an sbt file adding the Bloop sbt plugin."
```
olafurpg:
```suggestion
      "If true, Metals will not generate a `project/metals.sbt` file under the assumption that sbt-bloop is already manually installed in the sbt build. Build import will fail with a 'not valid command bloopInstall' error in case Bloop is not manually installed in the build when using this option."
```

Issue #1315, opened 2020-01-16
jvmopts and coursier repositories are not forwarded do Bloop


kczulko:
**Describe the bug**
When I've updated my metals installation to 0.8.0 (and 0.8.0 snapshot) I'm unable to properly import my project. I suspect that the issue may be related with my custom cacerts. For previous version (0.7.6) it worked fine. 

When I run my emacs, after some time I get this error:
```
INFO  started: Metals version 0.8.0+17-572f819c-SNAPSHOT in workspace '<my_workspace>'
INFO  time: initialize in 0.76s
INFO  running '/nix/store/wilpyg2acq8fglcchd1gh2340gl1hb8f-jetbrainsjdk-152b1248.6/jre/bin/java -Djline.terminal=jline.UnsupportedTerminal -Dsbt.log.noformat=true -Dfile.encoding=UTF-8 -Djavax.net.ssl.trustStore=<path_to_my_cacerts> -Djavax.net.ssl.trustStorePassword=changeit -Xss2048K -jar /tmp/metals1763162161274481289/sbt-launch.jar -Dbloop.export-jar-classifiers=sources bloopInstall'

...

[D] Waiting for a connection at local:///tmp/bsp-launcher1339486761166939038/bsp.socket...
The server is listening for incoming connections at local:///tmp/bsp-launcher1339486761166939038/bsp.socket...
Starting thread that pumps stdin and redirects it to the bsp server...
Starting thread that pumps server stdout and redirects it to the client stdout...
INFO  tracing is disabled for protocol BSP, to enable tracing of incoming and outgoing JSON messages create an empty file at <my_home>/.cache/metals/bsp.trace.json
ERROR Timeout waiting for 'build/initialize' response
ERROR Failed to connect with build server, no functionality will work.
java.util.concurrent.TimeoutException
	at java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1771)
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1915)
	at scala.meta.internal.metals.BuildServerConnection$.initialize(BuildServerConnection.scala:259)
	at scala.meta.internal.metals.BuildServerConnection$.$anonfun$fromSockets$1(BuildServerConnection.scala:203)
	at scala.util.Success.$anonfun$map$1(Try.scala:255)
	at scala.util.Success.map(Try.scala:213)
	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
[...]

Issue #1301, opened 2020-01-14
TimeoutException & download error on build\initialize


mrt181:
**Describe the bug**
When VS Code starts I get the following output

```log
Java home: C:\Program Files\AdoptOpenJDK\jdk-11.0.5.10-hotspot\
Using JVM options set in c:\Users\itisme\repos\gtm\skynet\.jvmopts
INFO  logging to file C:\Users\itisme\repos\gtm\skynet\.metals\metals.log
INFO  started: Metals version 0.8.0 in workspace 'C:\Users\itisme\repos\gtm\skynet'
INFO  time: initialize in 0.88s
Initializing Scala Debugger
INFO  skipping build import with status 'Installed'
INFO  tracing is disabled for protocol BSP, to enable tracing of incoming and outgoing JSON messages create an empty file at C:\Users\itisme\AppData\Local\scalameta\metals\cache\bsp.trace.json
ERROR Timeout waiting for 'build/initialize' response
ERROR Failed to connect with build server, no functionality will work.
java.util.concurrent.TimeoutException
	at java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1886)
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2021)
	at scala.meta.internal.metals.BuildServerConnection$.initialize(BuildServerConnection.scala:259)
	at scala.meta.internal.metals.BuildServerConnection$.$anonfun$fromSockets$1(BuildServerConnection.scala:203)
	at scala.util.Success.$anonfun$map$1(Try.scala:255)
	at scala.util.Success.map(Try.scala:213)
	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
	at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
	at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.lang.Thread.run(Thread.java:834)

WARN  Stopped configuration of SemanticDB in Scala 2.13.1 projects: Error downloading org.scalameta:semanticdb-scalac_2.13.1:4.3.0
[...]

PR #1317, opened 2020-01-16
include .jvmopts and coursier_repositories in Bloop

tgodzik:
Previously, we didn't forward any of those options to Bloop and this caused multiple issues with downloading sematicDB. Now we forward both, which is especially need both for proxies and custom repositories usage.

@jvican Can you confirm that it looks alright to you? -Dcoursier.repositories="" should work judging by [this line](https://github.com/coursier/coursier/blob/3742b82ecef50017230b45e8177f97740b850863/modules/coursier/jvm/src/main/scala/coursier/PlatformResolve.scala#L45).

Fixes https://github.com/scalameta/metals/issues/1315
Fixes https://github.com/scalameta/metals/issues/1301


tgodzik:
@jvican when testing locally it doesn't seem any additional options are picked up

tgodzik:
> @jvican when testing locally it doesn't seem any additional options are picked up

Ok, after some investigation it seems that the properties are not being forwarded, so this will not work :/

PR #1323, opened 2020-01-17
Fix version of sbt-metals for pre 0.8.0 instructions

gabro:
Minor, but it currently shows the snapshot version which does not exist, since we stopped publishing `sbt-metals`.

image


tgodzik:
Thanks @gabro ! I think we can remove it later, since we will not want anyone using 0.7.6 or earlier at some point.

gabro:
Yeah, let’s wait while we tackle potential 0.8.0 and then remove the section altogether

Issue #1303, opened 2020-01-14
Autocompletion is broken on conflicting "encoding" scalac flag


dpfeiffer:
**Describe the bug**
For any project where the build contains the play sbt-plugin in `plugins.sbt` and the project has the `PlayScala` plugin enabled autocompletion does not work.

The `metals.log` file contains an interesting error

```
SEVERE: requirement failed: List(utf8)
java.lang.IllegalArgumentException: requirement failed: List(utf8)
	at scala.Predef$.require(Predef.scala:281)
```

When hunting this down I figured out that the only thing which sets anything regarding the string `utf8` I found that the play sbt-plugin sets the `scalacOption` `-encoding` to `utf8`. See: https://github.com/playframework/playframework/blob/ddf3a7ee4285212ec665826ec268ef32b5a76000/dev-mode/sbt-plugin/src/main/scala/play/sbt/PlaySettings.scala#L51

When explicitly removing the encoding set by play, everything goes back to normal. (See code snippet below where `scalacOptions` modification is commented.

**To Reproduce**
1. Create a new sbt project containing the play sbt-plugin

`plugins.sbt`
```scala
addSbtPlugin("com.typesafe.play" % "sbt-plugin"    % "2.7.4")
```

`build.sbt`
```scala
lazy val root = project
  .in(file("."))
  .settings(
    name := "test"
    // scalacOptions := scalacOptions.value.filter(o => o != "-encoding" && o != "utf8"  && o != "utf-8"),
  )
  .enablePlugins(PlayScala)
```

2. Create a new scala source file
3. Try very simple autocomplete e.g. `"".toUpp....`

**Expected behavior**
Autocomplete should be working

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.8.0

gabro:
Thanks for reporting!

I've tried to reproduce this but I couldn't. Here's my attempt: https://github.com/gabro/metals-issue-1303

As you can see from the screenshot, autocompletions seem to work fine

![image](https://user-images.githubusercontent.com/691940/72350906-9abe6f80-36df-11ea-8a29-39d009cee123.png)

Would you mind trying to play around with my example (or come up with a different one) and see whether you can reproduce this? I suspect there's something else other than the `encoding` option

dpfeiffer:
[...]

PR #1324, opened 2020-01-17
Change require to warning for the presentation compiler

tgodzik:
That require was causing some issue when `utf-8` was used and it actually doesn't cause the presentation compiler to break.

Fixes https://github.com/scalameta/metals/issues/1303

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -179,7 +179,9 @@ case class ScalaPresentationCompiler(
     val (isSuccess, unprocessed) =
       settings.processArguments(options, processAll = true)
     require(isSuccess, unprocessed)
-    require(unprocessed.isEmpty, unprocessed)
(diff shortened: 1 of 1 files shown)


mtags/src/main/scala/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -179,7 +179,7 @@ case class ScalaPresentationCompiler(
     val (isSuccess, unprocessed) =
       settings.processArguments(options, processAll = true)
     require(isSuccess, unprocessed)
-    require(unprocessed.isEmpty, unprocessed)
+    logger.warning(s"Unknown compiler options: ${unprocessed.mkString(", ")}")
```
olafurpg:
```suggestion
if (unprocessed.nonEmpty) {
  logger.warning(s"Unknown compiler options: ${unprocessed.mkString(", ")}")
}
```

tgodzik: omg! Thanks, I haven't noticed it. :fearful:

PR #1277, opened 2020-01-09
Replace utest with MUnit

gabro:
Previously we used a heavily customized `BaseSuite` on top of utest. We didn't really use any feature of utest, except some pretty printing utility for the stack traces.

@olafurpg recently created https://github.com/olafurpg/funsuite which (not coincidentally) is extremely similar to our custom `BaseSuite`, and it provides some extra goodies that help in simplifying our suites.

This PR:

- replaces utest with funsuite
- uses some nice features from funsuite, such as `.flaky`, `.fail` and `assume(...)` to simplify our test suites
- adds `(implicit loc Location)` to all our custom `check...` and `assert...` methods in order to benefit from the fancy pretty printing in funsuite
- de-duplicates and refactors our custom assertions so to re-use as much as possible from funsuite
- turns test suites into classes (this is still open for debate, @olafurpg)

Still missing:
- there's a single test that uses `intercept` from utest, which is currently missing from funsuite
- there are two utilities (`assertNoDiffOrShowObtained` and `unifiedDiff`) that are present in funsuite but are not exposed by the public API (and I think they should)

Main benefits of this PR:
- we don't have to maintain an untested testing library inside Metals
- funsuite offers some utilities that simplify our test suites
- funsuite it's based on JUnit, so I expect it will be easier to do a few integrations we discussed about (e.g. reporting flaky failures)

PR #1339, opened 2020-01-22
Added Command to create new worksheet

alekseiAlefirov:
Issue: https://github.com/scalameta/metals-vscode/issues/179
(PR for the vscode extension: https://github.com/scalameta/metals-vscode/pull/183)

From the explorer context menu:
new-worksheet-explorer

From Command Palette:
new-worksheet-command-palette


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1249,6 +1250,24 @@ class MetalsLanguageServer(
             val msg = s"Invalid arguments: $args. Expecting: $argExample"
             Future.failed(new IllegalArgumentException(msg)).asJavaObject
         }
+      case ServerCommands.NewScalaWorksheet() =>
+        val args = params.getArguments.asScala
+        def f(dir: URI, name: String) = Future {
+          val path = dir.toString.toAbsolutePath.resolve(name + ".worksheet.sc")
+          //TODO: report failure, like file already exists?
```
tgodzik: Maybe we can just use `client.showMessage`?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1249,6 +1250,24 @@ class MetalsLanguageServer(
             val msg = s"Invalid arguments: $args. Expecting: $argExample"
             Future.failed(new IllegalArgumentException(msg)).asJavaObject
         }
+      case ServerCommands.NewScalaWorksheet() =>
```
tgodzik: Would it make sense to move it to a CommandProvider?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1249,6 +1250,24 @@ class MetalsLanguageServer(
...
+        val args = params.getArguments.asScala
+        def f(dir: URI, name: String) = Future {
```
tgodzik: We should use a better name than just `f`

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alekseiAlefirov: Done + `scribe.error` added

alekseiAlefirov: Screenshot from 2020-01-24 17-10-31

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alekseiAlefirov: While body of this method deserves (imho) being refactored to another file/class, it probably worth to be done in another PR

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alekseiAlefirov: `createWorksheet` now

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -130,6 +130,17 @@ object ServerCommands {
        |""".stripMargin
   )
 
+  val NewScalaWorksheet = new Command(
+    "new-scala-worksheet",
+    "Create new scala worksheet",
+    """|Create and open new scala worksheet file,
```
tgodzik: Nitpick, maybe without the comma and in one line?

```
    """|Create and open new scala worksheet file at the argument location with the argument + ".worksheet.sc" name.
```

alekseiAlefirov: :+1:

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1249,6 +1250,31 @@ class MetalsLanguageServer(
             val msg = s"Invalid arguments: $args. Expecting: $argExample"
             Future.failed(new IllegalArgumentException(msg)).asJavaObject
         }
+      case ServerCommands.NewScalaWorksheet() =>
```
tgodzik: Also, I think we can start with new worksheet, but even create a `WizardProvider` and `NewCommands` that would generalize over creating templates.

tests/unit/src/test/scala/tests/NewScalaWorksheetLspSuite.scala
```diff
@@ -0,0 +1,25 @@
+package tests
+
+import scala.meta.internal.metals.ServerCommands
+import scala.meta.internal.metals.MetalsEnrichments._
+
+object NewScalaWorksheetLspSuite extends BaseLspSuite("new-scala-worksheet") {
```
olafurpg: Heads up, when rebasing on top of master you will need to change this `object` into a `class`. See #1277

alekseiAlefirov:
Added command to create new class/object/trait file:
NewScalaClass

olafurpg:
That looks amazing! What do you think about having all the options in the same quickopen? Worksheet/class/object/trait/..

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1262,6 +1260,55 @@ class MetalsLanguageServer(
...
+                    MessageType.Error,
+                    s"Cannot create worksheet:\n ${e.toString()}"
+                  )
+            }
+            result
+          case _ =>
+            val msg = s"Invalid arguments: $args."
+            Future.failed(new IllegalArgumentException(msg))
+        }).asJavaObject
+      case ServerCommands.NewScalaClass() =>
```
tgodzik: Is it possible to może most of the logic to the new provider you created?

tgodzik: Te code grew a bit and we need to reduce the amount of code in MetalsLanguageServer I think.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1262,6 +1260,55 @@ class MetalsLanguageServer(
             val msg = s"Invalid arguments: $args. Expecting: $argExample"
             Future.failed(new IllegalArgumentException(msg)).asJavaObject
         }
+      case ServerCommands.NewScalaWorksheet() =>
+        val args = params.getArguments.asScala
+        (args match {
+          case Seq(JsonString.Option(directory), JsonString(name)) =>
```
tgodzik: Can the parameters just contain a single object that we would translate to a class using `JsonParser`? We wouldn't need to do the crazy unapplies.

And we could use the kind parameter for everything. There wouldn't need to be separate logic for creating worksheets and classes.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1917,4 +1964,19 @@ class MetalsLanguageServer(
     })
   }
 
+  private object JsonString {
```
tgodzik: Overall, I am not sure we need the unapplies here. You can just as easily match on JsonString.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1917,4 +1964,19 @@ class MetalsLanguageServer(
...
+    def unapply(json: Any): Option[String] = json match {
+      case p: JsonPrimitive if p.isString() => Some(p.getAsString())
```
tgodzik: Why you have: `p: JsonPrimitive if p.isString()` and below you can match on JsonString?

alekseiAlefirov: when I match on JsonString below, I use this exactly method `unapply`.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alekseiAlefirov: @tgodzik 
Maybe it's worth to create full `metals/newScalaFile` endpoint method? I'm not really sure which one is more appropriate in what situations.

tgodzik: I think a command is better since it can also work for other clients. Although not 100 % sure

ckipp01: I'd be in favor of a `newScalaFile` endpoint since looking at it, there is a lot that is shared. They can also both share the same params and just have kind be either trait, object, class, or worksheet.

alekseiAlefirov:
Made one new command `new-scala-file` instead of previous two.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alekseiAlefirov: Made one new command `new-scala-file` instead of previous two, and `MetalsNewScalaFileParams` case class with all it's needed arguments.

tgodzik:
@alekseiAlefirov `tests.AddPackageLspSuite` is failing due to changes in the PR, it's probably something simple. And you should also run `scalafixAll` on the codebase. I will review shortly

alekseiAlefirov:
Oh, yes. I forgot to mention that I moved providing package declaration from `didOpen` to the new command. So it won't happen when regular new file is created. This seems to make sense, as user either uses a template provider, or go raw to set everything by themselves. IDEA does it the same way. Hope that's OK.

Yep, and some tests to provide package object is failing. So I'm thinking about adding `package object` to the command as well. What do you think?

olafurpg:
>  I'm thinking about adding package object to the command as well

That sounds like a great idea!

olafurpg:
I think it would be nice to keep the current behavior where the package gets auto-inserted when you create a new file even if it's not through the "New scala file" command.

alekseiAlefirov:
Okay, there's a bit of difficulty to request apply edits both from the new `new-scala-file` command and on `didOpen` (sends package declaration now): VSCode does not apply edits until the text document is open, hence `didOpen` is triggered with empty `TextDocument`,  regardless if there have been edits to apply (btw this behavior is not reflected in `TestingClient`). So for the `new-scala-file` there going to be two edits apply, conflicting by their content and (possibly) their placement in the file.
To deal with it we can:
- Do not provide package on the the new command executed.
So the order is: on `new-scala-file` edits with empty class templates requested -> client opens document -> edits with package info requested -> client inserts package info at start of the document, where empty class is already inserted. The question is, is it guaranteed that the client receives requests in the same order, that they have been sent? If not (which may result in wrong order of the new empty class declaration and the package declaration), it can be probably fixed by not requesting the empty class edit applied, but sending this edit as a response to initial request, and then manually applied (or. to be more accurate, queued to be applied) on client side before the document is open.
Also, this approach assumes though quite default, but specific handling from the client (request to execute command and then open the new document). Also, I am not sure, if it's general for all LSP clients to have this `applyEdits` and `openTextDocument` order behavior, as VSCode does.
- Provide package with the new command and save the new file by the server, not sending request to apply edits. This seems a bit out of line with LSP-logic and inconsistent with `applyEdits` request sent, which is already there on `didOpen`. Though at the same time, the new file is already created by the server side (at this moment in this PR), and manipulations like this are already being performed at some places in Metals.

Just my thoughts out loud :thinking:, there probable have been and to be more, but that's what I've gathered up at the moment. After writing this down, I'm more about just writing the whole new content  for the new class (package object, etc.) on the server side. This seems simple and expects minimum behavior from a client.

alekseiAlefirov:
- fixed providing package for regular empty scala file
- added creation of new package object;
- template text now inserted on server site.
new-package-object

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -199,6 +216,13 @@ object ServerCommands {
...
     CancelCompile,
     BspSwitch,
-    StartDebugAdapter
+    StartDebugAdapter,
+    NewScalaFile
   )
 }
+
+case class MetalsNewScalaFileParams(
+    @Nullable directory: String,
```
tgodzik: We should be able to just pack it into Option? Or is it a gson limitation?

alekseiAlefirov: I just followed other params case class declaration, like `MetalsStatusParams`. I would guess, it's a lsp4j limitation, not sure. Maybe @olafurpg  can comment on this ?
Then I do `Option(directory)`, yes.

tgodzik:
@olafurpg what do you think about the current state?

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -0,0 +1,96 @@
...
+  def createNewFile(directory: Option[URI], name: String, kind: String)(
+      implicit ec: ExecutionContext
+  ): Future[URI] = kind match {
+    case "class" | "object" | "trait" =>
+      createClass(directory, name, kind)
+    case "package-object" =>
+      createPackageObject(directory)
+    case "worksheet" =>
+      createWorksheet(directory, name)
+    case _ => Future.failed(new IllegalArgumentException)
```
olafurpg:
```suggestion
    case invalid => Future.failed(new IllegalArgumentException(invalid))
```

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
olafurpg: We're using gson to serialize to JSON and I think you need custom gson adapters to support `Option[T]`

olafurpg: The `@Nullable` annotation isn't required, but I used it to document field where I would have preferred to use `Option[T]`

alekseiAlefirov:
> The implementation looks good to me  Great work @alekseiAlefirov 

Thanks :)
 
> What is the behavior now when users create new files through the file explorer?

new-package-object

olafurpg:
@alekseiAlefirov what is the behavior if the user runs "File: new file"?

alekseiAlefirov:
> @alekseiAlefirov what is the behavior if the user runs "File: new file"?

It's the same as in master: empty `.scala` files get package declaration, and `package.scala` ones get package object declaration additionally.

ckipp01:
Really excited to give this a try! Great job @alekseiAlefirov! 🎉

alekseiAlefirov:
@ckipp01 thank you :smile:.
Still, there's client PR to get approved and merged
https://github.com/scalameta/metals-vscode/pull/183

Issue #1161, opened 2019-12-06
Upgrade to the latest nuprocess and directory-watcher


olafurpg:
See #1158. We should update nuprocess and directory-watcher together. I'll take care of this upgrade to ensure that there are no regressions when running tests locally on my computer.

PR #1340, opened 2020-01-22
Update Nu Process to 1.2.5 and directory watcher to 0.9.9

tgodzik:
Fixes https://github.com/scalameta/metals/issues/1161

Can anyone using MacOS confirm?


tgodzik:
Tested it on a mac and seems to work fine. Let me know if anyone finds issues - I can revert that.

tgodzik:
Ok, actually might be better to run one of the faster build tool integration tests on MacOs, since they use NuProcess. So if there is any issues it should come up there.

PR #1343, opened 2020-01-23
Updated "unsupported features"

rtar:
Renaming symbols, running tests and debugging is actually supported now


olafurpg:
Thank you! This is a good idea

PR #1352, opened 2020-01-24
Create a symlink from workspace .bloop to out .bloop for BloopP…

wiwa:
This will allow users to use `bloop` commands (e.g. `bloop test ...`) directly from their workspace.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
+++ b/metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
@@ -1,38 +1,38 @@
 package scala.meta.internal.pantsbuild
 
 import bloop.config.{Config => C}
-import java.nio.file.Paths
-import java.nio.file.Files
-import java.nio.charset.StandardCharsets
-import java.nio.file.Path
-import scala.collection.mutable
-import ujson.Value
-import scala.util.Success
-import scala.util.Failure
-import scala.util.Try
-import scala.meta.internal.metals.BuildInfo
-import scala.meta.internal.metals.Timer
-import scala.meta.internal.metals.Time
-import java.nio.file.NoSuchFileException
-import scala.util.Properties
-import coursierapi.Dependency
-import scala.concurrent.ExecutionContext
-import scala.meta.internal.metals.MetalsEnrichments._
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
```diff
@@ -218,6 +213,19 @@ object BloopPants {
...
   }
 
+  private def symlinkToOut(args: Args): Unit = {
+    val workspaceBloop = args.workspace.resolve(".bloop")
+
+    if (!Files.exists(workspaceBloop) || Files.isSymbolicLink(workspaceBloop)) {
+      val outBloop = args.out.resolve(".bloop")
+      val actionWord =
+        if (Files.isSymbolicLink(workspaceBloop)) "overwriting" else "creating"
+      scribe.info(s"$actionWord symlink from $workspaceBloop to $outBloop")
```
olafurpg: Let's drop this logging statement

metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
```diff
@@ -1,38 +1,32 @@
 package scala.meta.internal.pantsbuild
 
 import bloop.config.{Config => C}
-import java.nio.file.Paths
-import java.nio.file.Files
+import ch.epfl.scala.bsp4j.BuildTargetIdentifier
```
olafurpg: Can you please configure IntelliJ to 

* sort everything alphabetically
* don't group prefix with `import a.{b,c}`

olafurpg:
we can ignore the unrelated CI failure, it should no longer be a problem after we merge #1277

metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
wiwa: Done.

metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
wiwa: done

PR #1359, opened 2020-01-27
[Draft] Split unit tests into packages to make them faster

tgodzik:


.github/workflows/ci.yml
```diff
@@ -6,13 +6,14 @@ on:
...
   unit:
-    name: ${{ matrix.os }} jdk-${{ matrix.java }} unit tests
+    name: ${{ matrix.os }} jdk-${{ matrix.java }} unit ${{matrix.type}} tests
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: [windows-latest, macOS-latest, ubuntu-latest]
         java: [8, 11]
+        type: [build, debug, extensions, navigation, refactor, tools, util]
```
olafurpg: We should be able to speed up the CI by ~2x by only introducing two types instead of 7 types.

olafurpg:
PS. We can now tag tests that are only flaky on Windows.

.github/workflows/ci.yml
tgodzik: I kind went overboard here :smile: 2 or 3 might be enough.

olafurpg:
BTW, it's possible to configure in build.sbt that only a subset of classes should run https://www.scala-sbt.org/1.x/docs/Testing.html#Filter+classes

We could use this to automatically split the test groups without moving classes into different packages

olafurpg:
I'm also curious if we can parallelize some of the test runner by using `testGrouping`. If we create two groups of tests then sbt will fork two JVMs and run the tests in parallel making use of the available CPUs on GitHub Actions.

tgodzik:
Thanks for the pointers @olafurpg ! I will dig into them and close the PR for now.

Not using packages might be useful if we want to split the tests equally.

PR #1361, opened 2020-01-28
Add new page on website with overiew of test suites in the code…

olafurpg:
Examples:

**The new sidebar**
Screenshot 2020-01-28 at 10 14 18

**Tests sorted by duration**
Screenshot 2020-01-28 at 10 10 24

**Tests sorted by flakyness** (the flakyness score in the `R` column is calculated incorrectly, I'll need to fix that)
Screenshot 2020-01-28 at 10 11 47

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -476,9 +476,10 @@ lazy val docs = project
     skip.in(publish) := true,
     moduleName := "metals-docs",
     mdoc := run.in(Compile).evaluated,
+    munitRepository := Some("scalameta/metals"),
     libraryDependencies ++= List(
       "org.jsoup" % "jsoup" % "1.12.1"
     )
   )
   .dependsOn(metals)
-  .enablePlugins(DocusaurusPlugin)
+  .enablePlugins(DocusaurusPlugin, MUnitReportPlugin)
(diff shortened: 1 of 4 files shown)


olafurpg:
Fun fact, the Metals codebase has ~2.6k test cases! We have ~22k lines of code (including whitespace and comments) only for tests

```
❯ git ls-files | grep "/test/" | grep ".*.scala$" | grep -v "/resources/"  | xargs wc -l | sort -n
...
   22055 total
``

olafurpg:
The number of tests is actually smaller than 2.6k, I just realized we generate a unique test name for the presentation compiler tests (which run against 6 different Scala versions) and we have a few hundred of those. The actual number of unique tests is closer to ~1.2k I think but not entirely sure

olafurpg:
The report is now live here https://scalameta.org/metals/docs/contributors/tests.html

olafurpg:
Please share any feedback on what data you would be interested in having displayed. This is the first iteration. I just put something quick together for a proof-of-concept. The underlying data source has more information, including

* JVM version
* operating system
* scala version
* structured exceptions for failed tests (including skipped tests due to flakiness etc)
* timestamps

PR #1374, opened 2020-01-30
Include ScalaBuildTarget in ScalaTarget 

tgodzik:
Previously, we had to check if `ScalaTarget` is a Scala target, which was a bit confusing. Now, we include the target inside the ScalaTarget class and avoid running `asScalaBuildTarget` unneccessarily.

Overall, I think this makes sense and it annoyed me a bit when dealing with it. :sweat_smile:


tgodzik:
The failures seem to be legitimate, I will take a look.

tgodzik:
Ok, it was basically a type - used scalaVersion for scalaBinaryVersion.

PR #1379, opened 2020-02-03
CodeAction: insert missing abstract members

tanishiking:
Close https://github.com/scalameta/metals-feature-requests/issues/59
Related: https://github.com/scalameta/metals-feature-requests/issues/67

## Overview
- Since metals v0.8.0, [it can complete all abstract missing members](https://scalameta.org/metals/blog/2020/01/10/cobalt.html#completion-to-add-all-abstract-members).
- This PR implements the [CodeAction](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#textDocument_codeAction) `implement methods` as the following screen-record, and it enables metals to implement all missing symbol by clicking diagnostics.

codeaction2

## Implementation overview
- Define code action `ImplementAbstractMethods.scala`
  - Add two kind of `ScalacDiagnostic` to `ScalacDiagnostic.scala` to capture the diagnostics
- Move `OverrideCompletion` related logic from `Completions.scala` to `OverrideCompletions.scala` and refactor a bit, so that CodeAction can re-use the procedures. https://github.com/scalameta/metals/commit/f44c7af43041fcd57f1801e6a2b964afb80c4522
- Calculate the completions using those procedures from `OverrideCompletions#implementAllAt()`
- Calculate insert position and indent
  - insert implements into the top of the body of class/object.
  - if there's any other declaration in the class/object, follow its indent, otherwise, fall back to **2 space indent**.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
@@ -10,8 +10,8 @@ import scala.collection.JavaConverters._
 final class CodeActionProvider(
     compilers: Compilers
 ) {
-
   val allActions: List[CodeAction] = List(
+    new ImplementAbstractMethods(compilers),
     new ImportMissingSymbol(compilers)
   )
 
@@ -19,7 +19,6 @@ final class CodeActionProvider(
       params: l.CodeActionParams,
       token: CancelToken
   )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
-
     def isRequestedKind(action: CodeAction): Boolean =
       Option(params.getContext.getOnly) match {
         case Some(only) =>
--- a/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
@@ -11,7 +11,7 @@ final class CodeActionProvider(
     compilers: Compilers
 ) {
   val allActions: List[CodeAction] = List(
-    new ImplementAbstractMethods(compilers),
+    new ImplementAbstractMembers(compilers),
     new ImportMissingSymbol(compilers)
   )
 
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
@@ -3,11 +3,30 @@ package scala.meta.internal.metals
 import org.eclipse.{lsp4j => l}
 
 object ScalacDiagnostic {
+
   object SymbolNotFound {
     private val regex = """not found: (value|type) (\w+)""".r
     def unapply(d: l.Diagnostic): Option[String] = d.getMessage() match {
       case regex(_, name) => Some(name)
       case _ => None
     }
+  }
+  object MissingImplementation {
+    // https://github.com/scala/scala/blob/4c0f49c7de6ba48f2b0ae59e64ea94fabd82b4a7/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala#L566
+    private val regex = """needs to be abstract""".r
+    def unapply(d: l.Diagnostic): Option[String] =
+      regex.findFirstMatchIn(d.getMessage()) match {
+        case Some(_) => Some(d.getMessage())
+        case _ => None
+      }
+  }
+  object ObjectCreationImpossible {
+    // https://github.com/scala/scala/blob/4c0f49c7de6ba48f2b0ae59e64ea94fabd82b4a7/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala#L564
+    private val regex = """object creation impossible""".r
+    def unapply(d: l.Diagnostic): Option[String] =
+      regex.findFirstMatchIn(d.getMessage()) match {
+        case Some(_) => Some(d.getMessage())
+        case None => None
+      }
   }
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
@@ -13,20 +13,19 @@ object ScalacDiagnostic {
   }
   object MissingImplementation {
     // https://github.com/scala/scala/blob/4c0f49c7de6ba48f2b0ae59e64ea94fabd82b4a7/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala#L566
-    private val regex = """needs to be abstract""".r
+    private val regex = """(?s).*needs to be abstract.*""".r
     def unapply(d: l.Diagnostic): Option[String] =
-      regex.findFirstMatchIn(d.getMessage()) match {
-        case Some(_) => Some(d.getMessage())
+      d.getMessage() match {
+        case regex() => Some(d.getMessage())
         case _ => None
       }
   }
   object ObjectCreationImpossible {
     // https://github.com/scala/scala/blob/4c0f49c7de6ba48f2b0ae59e64ea94fabd82b4a7/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala#L564
-    private val regex = """object creation impossible""".r
-    def unapply(d: l.Diagnostic): Option[String] =
-      regex.findFirstMatchIn(d.getMessage()) match {
-        case Some(_) => Some(d.getMessage())
-        case None => None
-      }
+    private val regex = """(?s).*object creation impossible.*""".r
+    def unapply(d: l.Diagnostic): Option[String] = d.getMessage() match {
+      case regex() => Some(d.getMessage())
+      case _ => None
+    }
   }
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImplementAbstractMembers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImplementAbstractMembers.scala
@@ -7,7 +7,7 @@ import scala.concurrent.ExecutionContext
 import scala.meta.internal.metals._
 import scala.meta.internal.metals.MetalsEnrichments._
 
-class ImplementAbstractMethods(compilers: Compilers) extends CodeAction {
+class ImplementAbstractMembers(compilers: Compilers) extends CodeAction {
 
   override def kind: String = l.CodeActionKind.QuickFix
 
@@ -23,15 +23,15 @@ class ImplementAbstractMethods(compilers: Compilers) extends CodeAction {
         .collect {
           case d @ ScalacDiagnostic.ObjectCreationImpossible(_)
               if params.getRange().overlapsWith(d.getRange()) =>
-            autoImport(d, params, token)
+            implementAbstractMembers(d, params, token)
           case d @ ScalacDiagnostic.MissingImplementation(_)
               if params.getRange().overlapsWith(d.getRange()) =>
-            autoImport(d, params, token)
+            implementAbstractMembers(d, params, token)
         }
     )
   }
 
-  private def autoImport(
+  private def implementAbstractMembers(
       diagnostic: l.Diagnostic,
       params: l.CodeActionParams,
       token: CancelToken
@@ -41,14 +41,14 @@ class ImplementAbstractMethods(compilers: Compilers) extends CodeAction {
       diagnostic.getRange.getStart()
     )
     compilers
-      .autoImplement(textDocumentPositionParams, token)
+      .implementAbstractMembers(textDocumentPositionParams, token)
       .map { edits =>
         val uri = params.getTextDocument().getUri()
         val edit = new l.WorkspaceEdit(Map(uri -> edits).asJava)
 
         val codeAction = new l.CodeAction()
 
-        codeAction.setTitle(ImplementAbstractMethods.title)
+        codeAction.setTitle(ImplementAbstractMembers.title)
         codeAction.setKind(l.CodeActionKind.QuickFix)
         codeAction.setDiagnostics(List(diagnostic).asJava)
         codeAction.setEdit(edit)
@@ -58,7 +58,7 @@ class ImplementAbstractMethods(compilers: Compilers) extends CodeAction {
   }
 }
 
-object ImplementAbstractMethods {
+object ImplementAbstractMembers {
   def title: String =
     s"Implement methods"
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImplementAbstractMembers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImplementAbstractMembers.scala
@@ -59,6 +59,5 @@ class ImplementAbstractMembers(compilers: Compilers) extends CodeAction {
 }
 
 object ImplementAbstractMembers {
-  def title: String =
-    s"Implement methods"
+  def title: String = "Implement all members"
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImplementAbstractMethods.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImplementAbstractMethods.scala
@@ -1,0 +1,64 @@
+package scala.meta.internal.metals.codeactions
+
+import scala.concurrent.Future
+import scala.meta.pc.CancelToken
+import org.eclipse.{lsp4j => l}
+import scala.concurrent.ExecutionContext
+import scala.meta.internal.metals._
+import scala.meta.internal.metals.MetalsEnrichments._
+
+class ImplementAbstractMethods(compilers: Compilers) extends CodeAction {
+
+  override def kind: String = l.CodeActionKind.QuickFix
+
+  override def contribute(
+      params: l.CodeActionParams,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+    Future.sequence(
+      params
+        .getContext()
+        .getDiagnostics()
+        .asScala
+        .collect {
+          case d @ ScalacDiagnostic.ObjectCreationImpossible(_)
+              if params.getRange().overlapsWith(d.getRange()) =>
+            autoImport(d, params, token)
+          case d @ ScalacDiagnostic.MissingImplementation(_)
+              if params.getRange().overlapsWith(d.getRange()) =>
+            autoImport(d, params, token)
+        }
+    )
+  }
+
+  private def autoImport(
+      diagnostic: l.Diagnostic,
+      params: l.CodeActionParams,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[l.CodeAction] = {
+    val textDocumentPositionParams = new l.TextDocumentPositionParams(
+      params.getTextDocument(),
+      diagnostic.getRange.getStart()
+    )
+    compilers
+      .autoImplement(textDocumentPositionParams, token)
+      .map { edits =>
+        val uri = params.getTextDocument().getUri()
+        val edit = new l.WorkspaceEdit(Map(uri -> edits).asJava)
+
+        val codeAction = new l.CodeAction()
+
+        codeAction.setTitle(ImplementAbstractMethods.title)
+        codeAction.setKind(l.CodeActionKind.QuickFix)
+        codeAction.setDiagnostics(List(diagnostic).asJava)
+        codeAction.setEdit(edit)
+
+        codeAction
+      }
+  }
+}
+
+object ImplementAbstractMethods {
+  def title: String =
+    s"Implement methods"
+}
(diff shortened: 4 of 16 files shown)


mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1045,252 +1035,6 @@ trait Completions { this: MetalsGlobal =>
       }
     }
 
-    /**
```
tanishiking: Just moved to `OverrideCompletions.scala`

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
```diff
@@ -0,0 +1,461 @@
...
+import org.eclipse.{lsp4j => l}
+
+import scala.collection.mutable
+import scala.collection.immutable.Nil
+
+trait OverrideCompletions { this: MetalsGlobal =>
+
+  private val DefaultIndent = 2
+
+  class OverrideDefMember(
```
tanishiking: The followings code comes from `Completions.scala`

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
```diff
@@ -0,0 +1,461 @@
...
+
+          implementAll :: overrideDefMembers
+        } else {
+          overrideDefMembers
+        }
+      }
+    }
+  }
+
+  private def getMembers(
```
tanishiking: Extracted the logic to calculate the list of `OverrideDefMember` so that this method can be used from both
- `case class Override`
- `implementAll()`

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
```diff
@@ -0,0 +1,461 @@
...
+    }
+  }
+
+  // NOTE(gabro): sym.isVar does not work consistently across Scala versions
+  // Specifically, it behaves differently between 2.11 and 2.12/2.13
+  // This check is borrowed from
+  // https://github.com/scala/scala/blob/f389823ef0416612a0058a80c1fe85948ff5fc0a/src/reflect/scala/reflect/internal/Symbols.scala#L2645
+  private def isVarSetter(sym: Symbol): Boolean =
+    !sym.isStable && !sym.isLazy && sym.isAccessor
+
```
tanishiking: The following codes are new ones.

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
```diff
@@ -0,0 +1,461 @@
...
+   * insert implementations onto the top of the body.
+   *
+   * @param text the text of the original source code.
+   * @param t the enclosing template for the class/object/trait we are implementing.
+   */
+  private def inferEditPosition(text: String, t: Template): Position = {
+    val start = t.pos.start
+    val end = t.pos.end
+    val offset = text.indexOf('{', start) + 1
+    t.pos.withStart(offset).withEnd(offset)
```
tanishiking: Find the position of the opening brace for the Template. Since `Template` always has both the opening and closing braces, it's safe, I guess.

Does anybody have a better idea to find a good position to insert implements? 🤔

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
```diff
@@ -0,0 +1,461 @@
...
+      // the closing brace is indented.
+      //
+      // object {
+      //   class Foo extends Bar {
+      //     override def foo = ???
+      //   }
+      // }
+      val end =
+        if (t.pos.source.offsetToLine(t.pos.start) ==
+            t.pos.source.offsetToLine(t.pos.end)) "\n" + " " * necessaryIndent
```
tanishiking: For the class/object that has defined in one line like this:

```scala
class Foo extends Bar {}
```

We have to add a newline and proper indent at the end of implements for the sake of pretty result like:

```scala
object X {
  class Foo extends Bar {
    override def foo(x: Int): Int = ??? 
  }
}
```

---

Otherwise, the result looks something like 

```scala
object X {
  class Foo extends Bar {
    override def foo(x: Int): Int = ???}
}

// or 

object X {
  class Foo extends Bar {
    override def foo(x: Int): Int = ???
}
}
```

tanishiking:
Hmm, CI fails with 

> ==> X tests.pc.AutoImplementMethodsSuite.initializationError  0.002s org.junit.runners.model.InvalidTestClassError: Invalid test class 'tests.pc.AutoImplementMethodsSuite':

but, I couldn't reproduce it in my local machine. I'll take a look at this.

gabro:
@tanishiking thank you so much for this contribution! I gave it a quick look and it looks great, I hope I’ll have time to review this properly later today.

As per the CI, we’re now using MUnit as our testing library and it requires suites to be classes not objects. I think that may be the cause of the error you’re seeing

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
```diff
@@ -0,0 +1,461 @@
...
+
+      // infer indent for implementations
+      // if there's declaration in the class/object, follow its indent.
+      // otherwise the indent default to 2
+      val indent = typed.tpe.decls
+        .filter(sym =>
+          !sym.isSynthetic &&
+            !sym.isPrimaryConstructor &&
+            sym.pos.line != t.pos.line // filter out explicit primary constructor `class Foo(x: Int) ...`
+        )
```
tanishiking: Find the explicit declaration in the class/object so that we can infer the indent from them.

tanishiking:
> As per the CI, we’re now using MUnit as our testing library and it requires suites to be classes not objects. I think that may be the cause of the error you’re seeing

Thank you for your input!
After rebased onto master, could reproduce it, and fixed that 👍

tests/cross/src/test/scala/tests/pc/AutoImplementMethodsSuite.scala
```diff
@@ -0,0 +1,506 @@
...
+import scala.meta.internal.metals.TextEdits
+
+class AutoImplementMethodsSuite extends BaseCodeActionSuite {
+
+  checkEdit(
+    "classdef",
+    """|package a
+       |
+       |object A {
+       |  trait Base {
```
tgodzik: Do we deal with generic classes like:
```
trait A[T]{
  def method : T
}

class B extends A[Int]{}
```

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
```diff
@@ -0,0 +1,461 @@
...
+   * Get the position of the opening brace of given Template.
+   * insert implementations onto the top of the body.
+   *
+   * @param text the text of the original source code.
+   * @param t the enclosing template for the class/object/trait we are implementing.
+   */
+  private def inferEditPosition(text: String, t: Template): Position = {
+    val start = t.pos.start
+    val end = t.pos.end
+    val offset = text.indexOf('{', start) + 1
```
tgodzik: What if we don't have {} and it's just `class Foo extends Bar` ?

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
```diff
@@ -0,0 +1,461 @@
...
+
+  def implementAllAt(pos: Position, text: String): List[l.TextEdit] = {
+    // make sure the compilation unit is loaded
+    typedTreeAt(pos)
+
+    lastVisistedParentTrees match {
+
+      // class Foo extends Bar {}
+      // ~~~~~~~~~~~~~~~~~~~~~~~~
+      case (c: ClassDef) :: _ =>
```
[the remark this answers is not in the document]
tanishiking: Oops, I missed that situation, we have to add the following case:

```scala
case (m: ModuleDef) :: _ =>
  // ...
```

to capture object's definition.

tests/cross/src/test/scala/tests/pc/AutoImplementMethodsSuite.scala
tanishiking: Yes:smile:

generics

- Syntactically, `class B extends A[Int] {}` can be captured by this case: https://github.com/scalameta/metals/blob/2969ba9e3f907ff769469cd020720a6145de35ea/mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala#L303-L313
- After that, the methods to implement are calculated based on semantics.

So, yes, we deal with that case.

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
tanishiking: I missed dealing with that case. Now, in that case, we don't add any implements on the class/object.

Probably, we should add necessary braces and complete it to

```scala
trait Bar {
  def bar(x: Int): Int
}

class Foo extends Bar

// to

class Foo extends Bar {
  override def  bar(x: Int): Int
}
```

tanishiking: done https://github.com/scalameta/metals/pull/1379/commits/4edcdc60dbfd90b00b1d3eafca9de07e598fa353 :)
For the object creation: `val foo = new Foo`, the diagnostic message is `trait Foo is abstract; cannot be instantiated`. It doesn't match "missing implementation" kind of error message, and so there's no quickfix available for that situation

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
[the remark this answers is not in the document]
tanishiking: done in https://github.com/scalameta/metals/pull/1379/commits/0bd0092ecb6890e62d2b137b085fed55d4db5473 :)

tests/cross/src/test/scala/tests/pc/AutoImplementMethodsSuite.scala
tanishiking: Added test in https://github.com/scalameta/metals/pull/1379/commits/0bd0092ecb6890e62d2b137b085fed55d4db5473

tanishiking:
Thank you so much for your feedback!
Answered and addressed them with 0bd0092 and 4edcdc6 :)

tgodzik:
> Thank you so much for your feedback!
> Answered and addressed them with [0bd0092](https://github.com/scalameta/metals/commit/0bd0092ecb6890e62d2b137b085fed55d4db5473) and [4edcdc6](https://github.com/scalameta/metals/commit/4edcdc60dbfd90b00b1d3eafca9de07e598fa353) :)

Thanks! It looks good! I will finish my review tomorrow, but I don't see anything that is missing really.

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -10,16 +10,15 @@ import scala.collection.JavaConverters._
 final class CodeActionProvider(
     compilers: Compilers
 ) {
-
   val allActions: List[CodeAction] = List(
+    new ImplementAbstractMethods(compilers),
```
gabro: Members instead of Methods?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImplementAbstractMethods.scala
```diff
@@ -0,0 +1,64 @@
...
+              if params.getRange().overlapsWith(d.getRange()) =>
+            autoImport(d, params, token)
+          case d @ ScalacDiagnostic.MissingImplementation(_)
+              if params.getRange().overlapsWith(d.getRange()) =>
+            autoImport(d, params, token)
+        }
+    )
+  }
+
+  private def autoImport(
```
gabro: copy/paste? Should be implementAbstractMembers or something like it

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
```diff
@@ -68,6 +69,11 @@
      */
     public abstract CompletableFuture<List<AutoImportsResult>> autoImports(String name, OffsetParams params);
 
+    /**
+     * Return the missing implements and imports for the symbol at the given position.
+     */
+    public abstract CompletableFuture<List<TextEdit>> autoImplement(OffsetParams params);
```
gabro: maybe implementAbstactMembers for consistency?

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
```diff
@@ -0,0 +1,506 @@
...
+          inferEditPosition(text, t).toLSP,
+          t,
+          text,
+          true,
+          _ => true
+        )
+
+      // object Foo extends Bar {}
+      // ~~~~~~~~~~~~~~~~~~~~~~~~
+      case (m: ModuleDef) :: _ =>
```
gabro: all the cases are nearly identical in their implementation, the only difference is how they extract a tree. Can we maybe extract it to a utility method taking a Tree as input?

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
```diff
@@ -0,0 +1,506 @@
...
+import scala.meta.pc.PresentationCompilerConfig.OverrideDefFormat
+
+import org.eclipse.{lsp4j => l}
+
+import scala.collection.mutable
+import scala.collection.immutable.Nil
+
+trait OverrideCompletions { this: MetalsGlobal =>
+
+  private val DefaultIndent = 2
```
gabro: defaultIndent (lowercase D)?

tanishiking: I personally usually use upper camel case for constants https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods, and [metals seem that it also follow this naming convention](https://github.com/scalameta/metals/blob/60e3d151dd80f0def1a122ff54e5c2f9e805e6cb/metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala#L12).

I don't have a strong opinion around a naming convention, but what do you think about this? :)

gabro: I don't mind either, let's leave it like this 👍

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
tanishiking: done in https://github.com/scalameta/metals/pull/1379/commits/761a52952d7d9138cba683bbbd1c638723889903
Thank you for your advice, it's much clear :)

metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
```diff
@@ -3,11 +3,30 @@ package scala.meta.internal.metals
...
     }
   }
+  object MissingImplementation {
+    // https://github.com/scala/scala/blob/4c0f49c7de6ba48f2b0ae59e64ea94fabd82b4a7/src/compiler/scala/tools/nsc/typechecker/RefChecks.scala#L566
+    private val regex = """needs to be abstract""".r
+    def unapply(d: l.Diagnostic): Option[String] =
+      regex.findFirstMatchIn(d.getMessage()) match {
+        case Some(_) => Some(d.getMessage())
+        case _ => None
+      }
```
gabro: nit: this could be (for consistency)

```scala
private val regex = """.*needs to be abstract.*""".r
def unapply(d: l.Diagnostic): Option[String] = d.getMessage() match {
  case regex() => Some(d.getMessage())
  case _ => None
}
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImplementAbstractMembers.scala
```diff
@@ -0,0 +1,64 @@
...
+        codeAction.setEdit(edit)
+
+        codeAction
+      }
+  }
+}
+
+object ImplementAbstractMembers {
+  def title: String =
+    s"Implement methods"
```
gabro:
```suggestion
    "Implement all members"
```

to align with the corresponding completion.

(also no need to break this into two lines)

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
```diff
@@ -0,0 +1,489 @@
...
+   * `class Foo extends Bar` => retuning position would be right after `Bar`.
+   *
+   * @param text the text of the original source code.
+   * @param t the enclosing template for the class/object/trait we are implementing.
+   */
+  private def inferEditPosition(text: String, t: Template): Position = {
+    val start = t.pos.start
+    val end = t.pos.end
+    val offset = text.indexOf('{', start)
+    if (offset > 0 && offset < t.pos.end)
```
tgodzik:
```suggestion
    if (hasBody(text,t))
```
We could reuse logic from `hasBody`

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
```diff
@@ -0,0 +1,489 @@
...
+      // ```
+      val start = if (shouldCompleteBraces) s" {\n${indent}" else s"\n${indent}"
+      val end = if (shouldCompleteBraces) s"${lastIndent}}" else lastIndent
+      val implementAll = new l.TextEdit(
+        range,
+        allAbstractEdits
+          .map(_.getNewText)
+          .mkString(
+            start,
+            s"\n${indent}",
```
tgodzik: Let's maybe add another newline between members?

It's usually what people would want I think.

So instead of:
```
class A{
  def a = ???
  def b = ???
}
```
we would have:
```
class A{
  def a = ???
  
  def b = ???
}
```

I would also consider maybe newline at the start or end.

tanishiking: Personally, I prefer the dense result (without empty lines), but meanwhile, the intellij-scala-plugin seems to add empty lines between methods, and it's probably better to add empty lines as you suggested 👍 so that users from IntelliJ can easily get used to the metals.

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
tanishiking: done https://github.com/scalameta/metals/pull/1379/commits/15d8e1feba9aed6cce18d9f6b74382f464b8be1e

Modified the `hasBody`'s return type to `Option[Int]` so that we can reuse `hasBody` from `inferEditPosition`.

metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
tanishiking: done https://github.com/scalameta/metals/pull/1379/commits/15d8e1feba9aed6cce18d9f6b74382f464b8be1e 👍 

Added `(?s)` to enable `.` to capture newlines (because the diagnostic usually have multiple lines).
https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#DOTALL

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImplementAbstractMembers.scala
tanishiking: https://github.com/scalameta/metals/pull/1379/commits/15d8e1feba9aed6cce18d9f6b74382f464b8be1e 👍

mtags/src/main/scala/scala/meta/internal/pc/OverrideCompletions.scala
ckipp01: So one thing to think about with this is now we have the "completion" version giving you the condensed version with no spaces and the "code action" version giving extra spaces. I sort of think we should be consistent.

tanishiking: > I sort of think we should be consistent.

Ah, that's correct, I'm still leaning toward add spaces for both codeaction and completions.
What do you think about this ?

tanishiking: For reference, added spaces between abstract members for completion-version https://github.com/scalameta/metals/pull/1379/commits/e100253dce0a50a1201510f9bec11fbe8f298bbb :)
We can revert it if we don't need it.

ckipp01: I don't have a strong preference on it. I just thing whichever one we choose we should do it the same in both. I'd say leave them they way you have them 👍

tanishiking:
Thank you so much for reviewing @tgodzik @gabro !

PR #1382, opened 2020-02-03
Rename file only if renamed symbol is directly enclosed by a package (closes #1380)

### LINKED ISSUE #1380: Renaming nested symbol renames file
**Describe the bug**

When a nested symbol matches the file name, and the symbol is renamed, the file is renamed too, regardless of whether the enclosing symbol is named like the file or not.

**To Reproduce**
1. Create a file `Events.scala`:

```scala
object Events {
  object Foo {
    object Events {}
  }
}
```
2. Rename `Events.Foo.Events` to X (anything)
3. The file is now `X.scala`.

**Expected behavior**

The file isn't renamed if:
- approach A: the symbol isn't top level
- ~approach B: there is a top-level symbol with the same name, but it's not the one we're renaming.~

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.8.0

**Search terms**
rename, nested, file, top level

kubukoz:
Previously, if a symbol was renamed that wasn't a top-level definition in the file but had a matching name, the file could be renamed. Now, the file will only be renamed if the symbol is only enclosed by a package.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -176,10 +176,11 @@ final class RenameProvider(
   ): Option[LSPEither[TextDocumentEdit, ResourceOperation]] = {
     fileChanges
       .find { file =>
-        isOccurence(str => {
+        isOccurence { str =>
+          str.owner.isPackage &&
           (str.desc.isType || str.desc.isTerm) &&
-            file.endsWith(s"/${str.desc.name.value}.scala")
-        })
+          file.endsWith(s"/${str.desc.name.value}.scala")
+        }
       }
       .map { file =>
         val newFile =
--- a/tests/unit/src/test/scala/tests/RenameLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/RenameLspSuite.scala
@@ -502,6 +502,17 @@ class RenameLspSuite extends BaseLspSuite("rename") {
     newName = "Animal"
   )
 
+  renamed(
+    "nested-symbol",
+    """|/a/src/main/scala/a/Main.scala
+       |package a
+       |object Foo {
+       |  object <<Ma@@in>>
+       |}
+       |""".stripMargin,
+    newName = "Child"
+  )
+
   def renamed(
       name: String,
       input: String,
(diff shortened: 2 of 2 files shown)


tests/unit/src/test/scala/tests/RenameLspSuite.scala
```diff
@@ -502,6 +502,17 @@ class RenameLspSuite extends BaseLspSuite("rename") {
...
 
+  renamed(
+    "nested-symbol",
+    """|/a/src/main/scala/a/Main.scala
+       |package a
+       |object Foo {
+       |  object <<Ma@@in>>
+       |}
+       |""".stripMargin,
+    newName = "Child"
```
kubukoz: Should I explicitly say here that the map of renamed files is empty, or can we assume that because of the parameter's default value?

gabro: I would be explicit, since that’s the purpose of this test case

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -176,10 +176,11 @@ final class RenameProvider(
   ): Option[LSPEither[TextDocumentEdit, ResourceOperation]] = {
     fileChanges
       .find { file =>
-        isOccurence(str => {
+        isOccurence { str =>
+          str.owner.isPackage &&
```
gabro: what happens if the file has no package definition?

olafurpg: I believe it returns “empty/“

tests/unit/src/test/scala/tests/RenameLspSuite.scala
kubukoz: I see there is one test already that passes `fileRenames = Map.empty`, so I'll do the same for explicitness.

kubukoz:
Sure, I'll do that :) give me a sec

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
kubukoz: I can see it's `_empty_/` :)

kubukoz:
All good on my side.

kubukoz:
I think this is still an incomplete solution. Consider this:

```scala
//Foo.scala
package Foo {
  package b {
    final case class Foo(x: String)
  }
}
```

Renaming the nested Foo will still trigger a file rename.

tgodzik:
@kubukoz  I think the behavior here is correct, normally you don't name the file the same as the package, but as the class. The logic for renaming the file works only as long as we follow those principles, you don't really need to name the file the same as the class/object, but we do it nonetheless.

There can be some edge cases, but I don't think it's worth covering every possible one.

PR #1386, opened 2020-02-04
Add .scalafmt.conf symbolic link when exporting Pants build.

olafurpg:
Previously, when exporting the Pants build to a separate directory the
`.scalafmt.conf` file in the original build did not get detected in the
exported build. Now, we insert a symbolic link to the config in the
original Pants build.


olafurpg:
Judging the the test reports we have a flaky test in `tests.debug.BreakpointDapSuite.case-class-unapply`

Issue #1287, opened 2020-01-12
Improve efficiency of go to implementation for non workspace symbols


tgodzik:
Currently, when looking for implementations of non workspace sources like RuntimeException, we load all information about the classpath. 

To make if faster we could use https://github.com/classgraph/classgraph

Otherwise some hand crafted solution might be needed.

joriscode:
Hi @tgodzik, I know very little of the code base of the project for the moment but I'm interested in this ticket. I'm currently reading metals doc, and https://scalameta.org/docs/semanticdb/specification.html#symbol. Is `ImplementationProvider.implementations` the correct entrypoint to this ticket?

tgodzik:
@joriscode Awesome! This is quite an isolated issue, so you are welcome to take a look. 

Most things that need to modified are in `GlobalClassTable` - it creates the context for situations where we need information about the dependencies inheritance. So that we for example know that `RuntimeException` inherits from `Exception`.

What we want is to replace the GlobalSymbolTable with just information about the class inheritance that we could get via classgraph.

The workflow I would recommend is running  the `ImplementationLspSuite.java-classes` test which actually checks the functionality for non workspace symbols. 
```
sbt unit/testOnly --  tests.ImplementationLspSuite.java-classes
```

We want to first confirm that we can get a much better performance from `classgraph` - but I am almost 100% sure that current implementation is way too slow.

Let me know if you need any more info!

tgodzik:
@joriscode did you manage to take a look at this? Or is it a harder problem than we thought?

joriscode:
@tgodzik Tbh, I had little time to look at it and I won't have more time for the coming week. However, I'm still interested in solving this issue. It goes without saying, that if this is blocking somehow the project or someone, I happily withdraw myself from the issue.

tgodzik:
@joriscode It's not blocking, but I am worried that as a result of https://github.com/scalameta/metals/pull/1388 finding references might get slower. However, I doubt if anyone else will be able to take a look at it. 

Also, I did some thinking and I suppose what we need to do here is to load information about class signatures from classgraph and translate them to semanticdb ClassSignature complete with filed/method signatures.

pvid:
Hi @tgodzik! I've been looking into this a bit - I don't think that switching to classgraph would be a good approach if you'd want to collect `ClassSignature` objects. Reimplementing a translation from classgraph objects to semanticdb could be quite error prone and hard to maintain.

[...]

PR #1388, opened 2020-02-05
Find References searches also for base method occurrences

### LINKED ISSUE #1063: Offer to 'Find all references' on overridden method
When doing a 'find all references' (in vscode) on a method that overrides a method from a superclass, only references to the method on the subclass are returned. It would be useful to have the option to also see references to the method on the superclass. Also easily jumping to the method in the superclass would be helpful.

kpbochenek:
Fixes #1063 

According to the discussion in related issue it would be good to provide references of overridden / super methods when searching for references of given method. 

Example code:
```scala
object Example {

  class A { def fx(): Unit = () }
  class B extends A { override def fx(): Unit = () }
  class C extends B { override def fx(): Unit = () }
  class D extends C
  class E extends D { override def fx(): Unit = () }

  class X { def fx(): Unit = () }

  val a = new A()
  a.fx()

  val b = new B();
  b.fx()

  val c = new C();
  c.fx()

  val d = new D();
  d.fx()

  val e = new E();
  e.fx()
  
}
```

With current behaviour when we search for a references to 'b.fx()' we will get 2 results:
2020-02-06-090121_290x107_scrot

With a new behaviour searching references to the same function we will get same results plus references to overridden methods (from superclasses and subclasses):
2020-02-06-090224_283x264_scrot

Detailed changes:
- fix typos (symbolOccurence -> symbolOccurrence, ineheritance -> inheritance)
- change functions to need only position and not full Params object if operate only on position
- get rid of using TextParams API object internally in favor of `PositionInFile`
- add `referencesSync` (only for tests) to not have to deal with async behaviour in tests.
- move functionality of searching for all subclasses / superclasses from RenameProvider -> ReferenceProvider and put it into `allInheritanceReferences` function
- reuse this function in searching for references and renaming.
- test checking mentioned functionality: https://github.com/scalameta/metals/pull/1388/files#diff-bd990ebb68b0756f1af9b16a82841ef8R216
- reference results are sorted ONLY in tests to provide deterministic results.
- fix searching for parent class when parent class exists in not currently opened file.

Here (https://github.com/scalameta/metals/pull/1388/files#diff-2c60d9818336621db1490d63594edff5R151) I had to add declaration to expected results, because in tests `server.references(...)` creates underneath `new ReferenceContext(true)` (where true is for includeDeclaration) which indicates declaration should be added to results.

Finding references is not following inheritance tree in case of functions defined in `methodsSearchedWithoutInheritance` (e.g. toString, equals, hashCode, clone)
2020-02-06-170525_527x179_scrot


tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala
```diff
@@ -142,12 +142,15 @@ class WorkspaceSymbolLspSuite extends BaseLspSuite("workspace-symbol") {
       _ = server.workspaceSymbol("scala.None")
       option = ".metals/readonly/scala/Option.scala"
       _ <- server.didOpen(option)
-      references <- server.references(option, "object None")
+      references = server.references(option, "object None")
```
tgodzik: doesn't references return future?

kpbochenek: Before my changes it called `references` on MetalsLanguageServer and returned Future. But this method only wrapped `referencesSync` in `CancelTokens` to introduce cancelability. 
We don't need cancelability here so I changed it to use directly `referencesSync` and not jump into async world not necessarily. 
But it is just cosmetic thing, I can revert that to Future if you want.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -197,9 +174,11 @@ final class RenameProvider(
       loc <- definitionProvider
         .fromSymbol(companionSymbol)
         .asScala
-      if loc.getUri().isScalaFilename
+      if loc.getUri.isScalaFilename
       companionLocs <- referenceProvider
-        .references(toReferenceParams(loc, includeDeclaration = false))
+        .references(
+          referenceProvider.toReferenceParams(loc, includeDeclaration = false)
```
tgodzik: `toReferenceParams` will not be needed here if we switch to maybe using `locations` in `references` method ?

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -390,13 +387,18 @@ final class TestingServer(
...
+    val f = server
       .didOpen(
         new DidOpenTextDocumentParams(
           new TextDocumentItem(uri, extension, 0, text)
         )
       )
       .asScala
+    f.onFailure {
+      case error =>
+        println(s"Failed to open file! ${filename} with error ${error}")
```
tgodzik: Why do we need it? If a future returns a failure then the tests should fail with this error anyway. :thinking:

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -214,6 +199,18 @@ final class TestingServer(
...
     )
   }
+
+  def assertReferenceDiff(
+      filename: String,
+      substring: String,
+      expectedDiff: String
+  )(implicit loc: munit.Location): Unit = {
+    Assertions.assertNoDiff(
+      references(filename, substring),
```
tgodzik: You can sort the results here, so that the tests don't fail.

tests/unit/src/test/scala/tests/ReferenceLspSuite.scala
```diff
@@ -212,4 +214,174 @@ class ReferenceLspSuite extends BaseLspSuite("reference") {
...
+          |  val d = new D();
+          |  d.fx()
+          |
+          |  val e = new E();
+          |  e.fx()
+          |}
+          |
+          |""".stripMargin
+      )
+      expectedDiff = """
```
tgodzik: Maybe move """ to newline?

tests/unit/src/test/scala/tests/ReferenceLspSuite.scala
```diff
@@ -212,4 +214,174 @@ class ReferenceLspSuite extends BaseLspSuite("reference") {
...
+          |}
+          |/b/src/main/scala/b/B.scala
+          |package b
+          |object B {
+          |  val bc = new a.C()
+          |  bc.fx()
+          |}
+          |""".stripMargin
+      )
+      expectedDiff = """
```
tgodzik: Move to next line?

tests/unit/src/test/scala/tests/RenameLspSuite.scala
```diff
@@ -515,7 +516,7 @@ class RenameLspSuite extends BaseLspSuite("rename") {
   )
 
   def renamed(
-      name: String,
+      name: TestOptions,
```
tgodzik: This change doesn't seem neccessary. Why use TestOptions here?

tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala
```diff
@@ -142,10 +142,13 @@ class WorkspaceSymbolLspSuite extends BaseLspSuite("workspace-symbol") {
...
       _ = assertNoDiff(
         references,
-        """|a/src/main/scala/a/A.scala:4:24: info: reference
+        """|.metals/readonly/scala/Option.scala:527:13: info: reference
```
tgodzik: I think this change is a regression. It seems we do include declaration by default now?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1018,9 +1018,12 @@ class MetalsLanguageServer(
       params: ReferenceParams
   ): CompletableFuture[util.List[Location]] =
     CancelTokens { _ =>
-      referencesResult(params).locations.asJava
+      referencesSync(params)
```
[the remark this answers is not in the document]
kpbochenek: it is used twice

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -65,31 +70,27 @@ final class ReferenceProvider(
...
-              distance,
-              occurrence,
-              alternatives,
-              params.getContext.isIncludeDeclaration,
-              checkMatchesText,
-              includeSynthetics
-            )
-            ReferencesResult(occurrence.symbol, locations)
+            val symbolName = occurrence.symbol.desc.name.value
+            if (RenameProvider.forbiddenModifyMethods.contains(symbolName)) {
```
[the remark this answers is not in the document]
kpbochenek: It was based on: https://github.com/scalameta/metals/issues/1063#issuecomment-553131812
To double check, should I remove this check here?

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -98,13 +99,142 @@ final class ReferenceProvider(
...
+    val referenceParams = new ReferenceParams()
+    referenceParams.setPosition(pos)
+    referenceParams.setTextDocument(textDoc)
+    val context = new ReferenceContext()
+    context.setIncludeDeclaration(includeDeclaration)
+    referenceParams.setContext(context)
+    referenceParams
+  }
+
+  def toReferenceParams(
```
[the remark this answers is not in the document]
kpbochenek: not in current state, it is used in RenameProvider too

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -98,13 +99,142 @@ final class ReferenceProvider(
...
+        implLoc <- implementation.implementations(textParams)
+        locParams = toReferenceParams(implLoc, includeDeclaration = true)
+        loc <- references(locParams).locations
+      } yield loc
+    } else {
+      Nil
+    }
+  }
+
+  def allRefs(
```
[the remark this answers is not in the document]
kpbochenek: no, it is used by MetalsLanguageServer too, and second one by RenameProvider

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -98,13 +99,142 @@ final class ReferenceProvider(
...
+      symbolOccurrence: SymbolOccurrence,
+      doc: TextDocument,
+      fnIncludeSynthetics: Synthetic => Boolean,
+      textParams: TextDocumentPositionParams
+  ): Seq[Location] = {
+    def toTextParams(location: Location): TextDocumentPositionParams = {
```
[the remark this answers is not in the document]
kpbochenek: I think yes, I just moved this part of code from RenameProvider as it was.
If I move it to outside of this method then `parentSymbols.map(toTextParams)` below will have to change because `toTextParams` will need to get additional params. Move it anyway?

tests/unit/src/main/scala/tests/TestingServer.scala
kpbochenek: it returns a String :/

tests/unit/src/test/scala/tests/RenameLspSuite.scala
kpbochenek: For testing to be able to add ".only" at the end of single test. It only applied to TestOptions.

tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala
kpbochenek: As described in PR:
Here (https://github.com/scalameta/metals/pull/1388/files#diff-2c60d9818336621db1490d63594edff5R151) I had to add declaration to expected results, because in tests server.references(...) creates underneath new ReferenceContext(true) (where true is for includeDeclaration) which indicates declaration should be added to results.

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -98,13 +99,142 @@ final class ReferenceProvider(
...
+        (
+          loc.getUri,
+          loc.getRange.getStart.getLine,
+          loc.getRange.getStart.getCharacter
+        )
+      )
+  }
+
+  def references(
+      params: ReferenceParams,
```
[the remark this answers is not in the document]
kpbochenek: This is an object used in lsp API, what do you mean here?

tests/unit/src/test/scala/tests/ReferenceLspSuite.scala
kpbochenek: this is formated by scalafix.

tests/unit/src/test/scala/tests/ReferenceLspSuite.scala
kpbochenek: this is formated by scalafix.

tgodzik:
@olafurpg It's ready to review, let me know what you think. :)

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -98,13 +109,97 @@ final class ReferenceProvider(
     }
   }
 
+  private def implementations(
```
tgodzik: Let's move the private methods to the bottom.

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -233,6 +327,39 @@ final class ReferenceProvider(
     }
   }
 
+  def companionReferences(sym: String): Seq[Location] = {
```
tgodzik: Isn't this needed only in the RenameProvider?

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -50,45 +52,26 @@ import scala.collection.mutable
...
-import scala.meta.internal.metals.WindowStateDidChangeParams
-import scala.meta.internal.metals.Directories
 import scala.meta.internal.metals.MetalsEnrichments._
-import scala.meta.internal.metals.MetalsLanguageServer
-import scala.meta.internal.metals.MetalsServerConfig
 import scala.meta.internal.metals.PositionSyntax._
-import scala.meta.internal.metals.ProgressTicks
-import scala.meta.internal.metals.Time
-import scala.meta.internal.metals.UserConfiguration
+import scala.meta.internal.metals._
```
tgodzik: Let's avoid the wildcard imports.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -3,17 +3,15 @@ package tests
 import java.io.IOException
 import java.net.URI
 import java.nio.charset.StandardCharsets
-import java.nio.file.FileVisitResult
-import java.nio.file.Files
-import java.nio.file.Path
-import java.nio.file.Paths
-import java.nio.file.SimpleFileVisitor
+import java.nio.file._
```
tgodzik: Let's avoid the wildcard import here too.

tests/unit/src/test/scala/tests/ReferenceLspSuite.scala
```diff
@@ -212,4 +214,176 @@ class ReferenceLspSuite extends BaseLspSuite("reference") {
       _ = server.assertReferenceDefinitionBijection()
     } yield ()
   }
+
+  test("method-hierarchy") {
+    cleanWorkspace()
+    val expectedDiff =
```
tgodzik: I was wrong, let's put it after the code definition, since it's more natural to read.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -945,7 +948,11 @@ class MetalsLanguageServer(
       params: TextDocumentPositionParams
   ): CompletableFuture[util.List[DocumentHighlight]] =
     CancelTokens { _ =>
-      documentHighlightProvider.documentHighlight(params)
+      val positionInFile = PositionInFile(
```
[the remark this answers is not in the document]
kpbochenek: ok

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -60,7 +60,7 @@ final class DefinitionProvider(
           warnings.noSemanticdb(path)
           DefinitionResult.empty
       }
-    if (fromSemanticdb.locations.isEmpty()) {
+    if (fromSemanticdb.locations.isEmpty) {
```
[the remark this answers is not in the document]
kpbochenek: okay I will keep things as they are in other places.

As a side note in scala world there is actually established convention:
https://docs.scala-lang.org/style/naming-conventions.html#parentheses
Less clear for using java in scala because you cannot forbid calling method with () but still good rule is 
side effects -> parenthesis
pure -> no parenthesis

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -65,31 +68,50 @@ final class ReferenceProvider(
...
-              distance,
-              occurrence,
-              alternatives,
-              params.getContext.isIncludeDeclaration,
-              checkMatchesText,
-              includeSynthetics
-            )
-            ReferencesResult(occurrence.symbol, locations)
+            val symbolName = occurrence.symbol.desc.name.value
+            if (ReferenceProvider.methodsSearchedWithoutInheritance.contains(
```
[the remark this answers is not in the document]
kpbochenek: okay

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -98,13 +120,97 @@ final class ReferenceProvider(
...
 
+  def allInheritanceReferences(
+      symbolOccurrence: SymbolOccurrence,
+      doc: TextDocument,
+      filePosition: FilePosition,
+      fnIncludeSynthetics: Synthetic => Boolean
+  ): Seq[Location] = {
+    val parentSymbols = implementation
+      .topMethodParents(doc, symbolOccurrence.symbol)
+    val txtParams: Seq[FilePosition] = {
```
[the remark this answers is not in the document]
kpbochenek: ahh good catch!

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -68,64 +60,43 @@ final class RenameProvider(
...
-        isLocal = occurence.symbol.isLocal
-        currentReferences = referenceProvider
-          .references(
-            // we can't get definition by name for local symbols
-            toReferenceParams(txtParams, includeDeclaration = isLocal),
-            // local symbol will not contain a proper name
-            checkMatchesText = !isLocal,
-            includeSynthetics = includeSynthetic
+      val symbolOccurrence =
+        definitionProvider.symbolOccurrence(params)
```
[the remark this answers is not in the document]
kpbochenek: There is still prepareRename that uses overloaded version with `params`

olafurpg:
@tgodzik is the performance penalty for every request (even for non overridden methods) or only for requests on overridden methods?

olafurpg:
Currently the performance of references has not been an issue (it’s really fast) whereas missing results (and occasionally incorrect results) has been a bigger issue in my experience.

tgodzik:
> Currently the performance of references has not been an issue (it’s really fast) whereas missing results (and occasionally incorrect results) has been a bigger issue in my experience.

Ok, I think we can optimize is later if needed and rename already requires loading classpath information. However, now that I think about it. Do we actually look correctly for methods coming from a dependency? Can we add a test case?

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -177,13 +212,18 @@ final class ImplementationProvider(
   private def methodInParentSignature(
       sig: ClassSignature,
       childInfo: SymbolInformation,
+      globalSymbolTable: Option[GlobalSymbolTable],
       childASF: Map[String, String] = Map.empty
-  ): Seq[Location] = {
+  ): Seq[Either[Location, SymbolInformation]] = {
     sig.parents.flatMap {
-      case parentSym: TypeRef =>
+      case parentSym: TypeRef if parentSym.symbol != "scala/AnyRef#" =>
```
tgodzik: Why do we need this if?

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -82,93 +83,127 @@ final class ImplementationProvider(
...
     } yield {
       classInfo.signature match {
-        case sig: ClassSignature =>
-          methodInParentSignature(sig, currentInfo)
-        case _ => Nil
+        case classSignature: ClassSignature =>
+          methodInParentSignature(
+            classSignature,
+            currentInfo,
+            globalTable.globalSymbolTableFor(
```
tgodzik: Could we extract this to a val.

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -328,10 +372,10 @@ final class ImplementationProvider(
         loc =>
           // we are not interested in local symbols from outside the workspace
           (loc.symbol.isLocal && loc.file.isEmpty) ||
-          // local symbols ineheritance should only be picked up in the same file
+          // local symbols inheritance should only be picked up in the same file
           (loc.symbol.isLocal && loc.file != currentPath)
       }
-      directImplementations.toSet ++ directImplementations
+      directImplementations ++ directImplementations
```
tgodzik: ` directImplementations ++ directImplementations` I think we might only need  `directImplementations.toSet` - this actually never made sense :fearful: my bad

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -208,15 +185,29 @@ final class RenameProvider(
...
       loc <- definitionProvider
         .fromSymbol(companionSymbol)
         .asScala
-      if loc.getUri().isScalaFilename
+      if loc.getUri.isScalaFilename
       companionLocs <- referenceProvider
-        .references(toReferenceParams(loc, includeDeclaration = false))
+        .currentSymbolReferences(
+          locationToFilePosition(loc),
+          includeDeclaration = false
```
tgodzik: We should have `canSkipExactMatch = true`

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -98,13 +124,118 @@ final class ReferenceProvider(
...
+      canSkipExactMatchCheck: Boolean = true
+  ): Seq[Location] = {
+    val parentSymbols = implementation
+      .topMethodParents(doc, symbolOccurrence.symbol)
+
+    if (failWhenReachingDependencySymbol && parentSymbols.exists(_.isRight)) {
+      Seq.empty
+    } else {
+
+      val topParentSymbols: Seq[Either[FilePosition, SymbolInformation]] = {
```
tgodzik: Name might actually be missleading, maybe `mainDefinitions` ?

kpbochenek: ok

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
kpbochenek: We get AnyRef as parents, it makes sense not to analyze it

tgodzik:
We need to block this PR until https://github.com/scalameta/metals/issues/1287 is solved. We don't have an easy way to detect inheritance inside the dependencies and I am afraid this will cause `find references` less responsive. This can be done either by using classgraph or by our custom solution similar to searching for top levels.

TLDR: We need a faster and less memory heavy way to check inheritance in dependencies.

PR #1390, opened 2020-02-05
Change config options to match coc

ckipp01:
This pr changes the way the available config options for metals is displayed in the vim editor page. They are displayed with dashes by default, but when a user is setting them using `coc-metals` they are displayed in camel-case when in `:CocConfig`. This was on oversight when I re-did the docs. This corrects that (only on the vim editor page).

### DIFF
--- a/metals-docs/src/main/scala/docs/UserConfigurationModifier.scala
+++ b/metals-docs/src/main/scala/docs/UserConfigurationModifier.scala
@@ -62,7 +62,7 @@ class UserConfigurationModifier extends StringModifier {
                |**Example**:
                |```json
                |{
-               |  "metals.${option.key}": "${option.example}"
+               |  "metals.${option.camelCaseKey}": "${option.example}"
                |}
                |```
                |""".stripMargin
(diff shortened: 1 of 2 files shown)


metals-docs/src/main/scala/docs/UserConfigurationModifier.scala
```diff
@@ -62,7 +62,7 @@ class UserConfigurationModifier extends StringModifier {
                |**Example**:
                |```json
                |{
-               |  "metals.${option.key}": "${option.example}"
+               |  "metals.${option.camelCaseKey}": "${option.example}"
```
ckipp01: You can't see it without expanding, but this is in the `lsp-config-coc` case, so it will only be used there.

metals/src/main/scala/scala/meta/internal/metals/UserConfigurationOption.scala
```diff
@@ -15,4 +15,9 @@ case class UserConfigurationOption(
   def headerID: String = {
     title.toLowerCase().replace(' ', '-')
   }
+  def camelCaseKey: String =
+    key.split('-').toList match {
```
tgodzik: You can actually pattern match on an Array:
```
  case Array(head, tail @_*) =>
    head ++ tail.flatMap(_.capitalize)
```
Not 100 % sure if more readable, but an interesting side note.

ckipp01: Ahh, nice. I _think_ in terms of readability, I'd prefer the other one, but thanks!

PR #1397, opened 2020-02-06
Do not accept empty string values from the lsp settings

tgodzik:


metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -133,11 +133,16 @@ object UserConfiguration {
     val base: JsonObject =
       Option(json.getAsJsonObject("metals")).getOrElse(new JsonObject)
 
-    def getKey[A](key: String, f: JsonElement => Option[A]): Option[A] = {
+    def getKey[A](
+        key: String,
+        f: JsonElement => Option[A],
+        isEmpty: JsonElement => Boolean = _ => false
```
olafurpg: Instead of passing in `isEmpty = ...`, the `f` method can implement the filter instead.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -155,8 +160,12 @@ object UserConfiguration {
               None
             }, Some(_))
```
olafurpg:
```suggestion
            }, Some(_).filter(_.nonEmpty))
```

olafurpg:
There is BTW a `UserConfigurationSuite`, we should add a test case for this

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
tgodzik: Hmm... that will not work since the f method is invoked after we get the value.

tgodzik:
> There is BTW a `UserConfigurationSuite`, we should add a test case for this

Added test cases and switched around the logic to make this simpler. First properties and the settings.

PR #1399, opened 2020-02-06
Enable verbose sbt loggers in CI.

olafurpg:
Previously, we used the default MUnit test arguments when testing in CI.
In some cases, the CI logs haven't included recently what test suite
is failing so this change is an attempt to try and fix that by enabling
verbose loggers and telling MUnit to use the sbt loggers instead of
`System.out.println()`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,6 +11,8 @@ def crossSetting[A](
     case _ => otherwise
   }
 
+val MUnitFramework = new TestFramework("munit.Framework")
+
 inThisBuild(
   List(
     version ~= { dynVer =>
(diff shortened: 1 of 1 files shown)


tgodzik:
Though I don't really understand what happened with the failure:
```
[info] tests.DebugProtocolSuite.disconnect started
[error] ==> X tests.DebugProtocolSuite.disconnect  1.642s java.lang.IllegalStateException: ERROR: transport error 202: send failed: Broken pipe
```

No data at all compared to other tests, :/

olafurpg:
Feel free to merge if you like the new format

I don't know what's going on there either :/

PR #1405, opened 2020-02-07
Return all rename file changes up to a threshold

tgodzik:
Previously, we would only send edits for open files when renaming. This causes an issue if the user wants to see a preview of all files.

Now, this changed to return up to threshold number of files for VS Code, so that preview can work correctly. The behavior is unchanged for other editors, since it might be problematic to save all files there.

preview

Tested on ZIO and Akka, where there were over 200 files changed. It seems VS code is much more optimized now.

ZIO example:

preview2

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
@@ -3,8 +3,6 @@ package scala.meta.internal.metals
 import scala.meta.internal.metals.Configs._
 import scala.meta.internal.pc.PresentationCompilerConfigImpl
 import scala.meta.pc.PresentationCompilerConfig.OverrideDefFormat
-import scala.util.Success
-import scala.util.Try
 
 /**
  * Configuration parameters for the Metals language server.
@@ -57,14 +55,8 @@ final case class MetalsServerConfig(
       "metals.warnings",
       default = true
     ),
-    openFilesOnRenames: Boolean = MetalsServerConfig.binaryOption(
-      "metals.open-files-on-rename",
-      default = false
-    ),
-    renameFileThreshold: Int = MetalsServerConfig.intOption(
-      "metals.rename-file-threshold",
-      default = 300
-    ),
+    openFilesOnRenames: Boolean = false,
+    renameFileThreshold: Int = 300,
     bloopEmbeddedVersion: String = System.getProperty(
       "bloop.embedded.version",
       BuildInfo.bloopVersion
@@ -97,25 +89,11 @@ object MetalsServerConfig {
       case Some("false" | "off") => false
       case Some(other) =>
         scribe.warn(
-          s"Invalid property, required 'true|on' or 'false|off', but got '$other'. Using the default value '$default'"
+          s"Invalid value for property '$key', required 'true|on' or 'false|off', but got '$other'. Using the default value '$default'"
         )
         default
       case None => default
     }
-  def intOption(key: String, default: Int): Int = {
-    sys.props.get(key) match {
-      case Some(property) =>
-        Try(property.toInt) match {
-          case Success(value) => value
-          case _ =>
-            scribe.warn(
-              s"Invalid property, required Int, but got '$property'. Using the default value '$default'"
-            )
-            default
-        }
-      case None => default
-    }
-  }
 
   def base: MetalsServerConfig = MetalsServerConfig()
   def default: MetalsServerConfig = {
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
@@ -3,6 +3,8 @@ package scala.meta.internal.metals
 import scala.meta.internal.metals.Configs._
 import scala.meta.internal.pc.PresentationCompilerConfigImpl
 import scala.meta.pc.PresentationCompilerConfig.OverrideDefFormat
+import scala.util.Success
+import scala.util.Try
 
 /**
  * Configuration parameters for the Metals language server.
@@ -55,6 +57,14 @@ final case class MetalsServerConfig(
       "metals.warnings",
       default = true
     ),
+    openFilesOnRenames: Boolean = MetalsServerConfig.binaryOption(
+      "metals.open-files-on-rename",
+      default = false
+    ),
+    renameFileThreshold: Int = MetalsServerConfig.intOption(
+      "metals.rename-file-threshold",
+      default = 300
+    ),
     bloopEmbeddedVersion: String = System.getProperty(
       "bloop.embedded.version",
       BuildInfo.bloopVersion
@@ -67,6 +77,8 @@ final case class MetalsServerConfig(
     List[String](
       s"glob-syntax=$globSyntax",
       s"status-bar=$statusBar",
+      s"open-files-on-rename=$openFilesOnRenames",
+      s"rename-file-threshold=$renameFileThreshold",
       s"slow-task=$slowTask",
       s"execute-client-command=$executeClientCommand",
       s"compilers=$compilers",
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -55,6 +57,14 @@ final case class MetalsServerConfig(
       "metals.warnings",
       default = true
     ),
+    openFilesOnRenames: Boolean = MetalsServerConfig.binaryOption(
+      "metals.open-files-on-renames",
```
ckipp01: Nit: All the other names seem to be singular. Could we do `open-files-on-rename`?

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -45,9 +46,11 @@ final class RenameProvider(
     workspace: AbsolutePath,
     client: MetalsLanguageClient,
     buffers: Buffers,
-    compilations: Compilations
+    compilations: Compilations,
+    metalsConfig: MetalsServerConfig
 ) {
 
+  private val renameFileTreshold = 100
```
ckipp01: I don't think this is used anywhere is it since we are taking the threshold from the `metalsConfig`?

tgodzik: Good catch, I forgot to remove that!

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
tgodzik: Sure, I agree it makes sense - changed

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -85,6 +97,11 @@ object MetalsServerConfig {
       case "false" | "off" => false
       case _ => default
     }
+  def intOption(key: String, default: Int): Int =
```
kpbochenek: I can see it is similar to `binaryOption` but it is not a good idea to silently ignore user properties if they are not parsed correctly. 
I would at least log an error/warning that parameter could not be parsed and default value is used.

tgodzik:
@kpbochenek The settings are actually explained in the settings menu.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
tgodzik: Fixed, added log messages for binary and int options.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -55,6 +57,14 @@ final case class MetalsServerConfig(
       "metals.warnings",
       default = true
     ),
+    openFilesOnRenames: Boolean = MetalsServerConfig.binaryOption(
+      "metals.open-files-on-rename",
+      default = false
+    ),
+    renameFileThreshold: Int = MetalsServerConfig.intOption(
+      "metals.rename-file-threshold",
+      default = 300
```
tgodzik: 300 is around the moment that VS Code is not yet totally killed and the previews still make sense. Around 1000 everything went bonkers.

tgodzik:
@gabro @olafurpg what do you think about this change?

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -80,11 +92,30 @@ final case class MetalsServerConfig(
...
-    System.getProperty(key) match {
-      case "true" | "on" => true
-      case "false" | "off" => false
-      case _ => default
+    sys.props.get(key) match {
+      case Some("true" | "on") => true
+      case Some("false" | "off") => false
+      case Some(other) =>
+        scribe.warn(
+          s"Invalid property, required 'true|on' or 'false|off', but got '$other'. Using the default value '$default'"
```
gabro: if we're adding this warning, I would recommend also including the `key` name

```suggestion
          s"Invalid value for property '$key', required 'true|on' or 'false|off', but got '$other'. Using the default value '$default'"
```

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -80,11 +92,30 @@ final case class MetalsServerConfig(
...
+      case None => default
+    }
+  def intOption(key: String, default: Int): Int = {
+    sys.props.get(key) match {
+      case Some(property) =>
+        Try(property.toInt) match {
+          case Success(value) => value
+          case _ =>
+            scribe.warn(
+              s"Invalid property, required Int, but got '$property'. Using the default value '$default'"
```
gabro: same as above

```suggestion
              s"Invalid value for property '$key', required Int, but got '$property'. Using the default value '$default'"
```

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -135,14 +136,21 @@ final class RenameProvider(
...
       }
       val fileChanges = allChanges.flatten.toMap
-      val (openedEdits, closedEdits) = fileChanges.partition {
-        case (file, edits) =>
-          val path = AbsolutePath(Paths.get(new URI(file)))
-          buffers.contains(path)
-      }
+      val (openedEdits, closedEdits) =
+        if (!metalsConfig.openFilesOnRenames || fileChanges.keySet.size >= metalsConfig.renameFileThreshold) {
+          if (metalsConfig.openFilesOnRenames) {
```
gabro: I find this check a bit confusing: we're checking for `!metalsConfig.openFilesRenames` and immediately after for `metalsConfig.openFilesOnRenames`. Can we simplify this somehow?

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -135,14 +136,21 @@ final class RenameProvider(
...
+            client.showMessage(fileThreshold(fileChanges.keySet.size))
+          }
+          fileChanges.partition {
+            case (path, edits) =>
```
gabro: nit:

```suggestion
            case (path, _) =>
```

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -357,6 +367,16 @@ final class RenameProvider(
     )
   }
 
+  private def fileThreshold(
+      files: Int
+  ): MessageParams = {
+    val message =
+      s"""|Renamed symbol is present in over $files files.
+          |It will be renamed outside the editor in the background, 
+          |otherwise the editor might become unresponsive.""".stripMargin
```
gabro: Suggestion (feel free to discard):

s"""|The renamed symbol appears in over $files files.
    |It will be renamed without opening the files, 
    |to prevent the editor from becoming unresponsive.""".stripMargin

olafurpg:
One alternative to system preferences is to hardcore the value per editor. Once we support the options for clients then we’re bound to support it for a long time.

gabro:
I actually had the same doubt. Maybe hardcoding is fine for the time being

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
tgodzik: Fixed!

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
tgodzik: Removed it since, the method is no longer used.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: I extraced `shouldRenameInBackground` value and it should be a bit clearer I think.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: Thanks, fixed!

tgodzik:
@olafurpg @gabro I hardcoded the values and fixed the messages. Should be good to go!

olafurpg:
Thank you @tgodzik

PR #1407, opened 2020-02-08
Remove check for focused document on non-supporting editors

### LINKED ISSUE #1258: After switching focus to another file, worksheets no longer work in non vscode editors
**Describe the bug**
We'll use vim for example. If I open `test.worksheet.sc`, throw down some code, save, and then see the feature in action and I see the evaluations appear. Awesome. I change a couple of things, save again, and boom, new evaluations appear. I then jump into another file in a new buffer, then after a bit decide I want to test something else in `test.worksheet.sc`, so I go back to it. I change some stuff, save, and no updated evaluations appear :(

It looks like we are using `onChangeActiveTextEditor` in the vs code extension to capture the change in focus on the document, and then evaluating the worksheets and publishing the decorations. However, this is reliant on the focus being set. In other editors, this doesn't seem to be working. I've posted a gif below in both Vim and Sublime to illustrate this.

vim

sublime

**To Reproduce**
1. Open a worksheet and evaluate
2. Open another file and do some stuff
3. Go back to the worksheet and try to evaluate

**Expected behavior**
I would expect to be able to go back to the worksheet and get new evaluations/decorations returned on save.

**Installation:**
- Operating system: macOS
- Editor: Vim and Sublime (I'm assuming all non VS Code editors)
- Metals version: v0.7.6-LATEST-SNAPSHOT

**Search terms**
worksheets, decorations

ckipp01:
Fixes #1258 

~This may not be the most elegant solution, but it seems to do the trick. This PR introduces a private`clientSupportsDidFocus` var in `MetalsLanguageServer` that defaults to false. This will help clients that don't support `didFocusTextDocument` to be able to move back and forth between scala files and worksheets correctly. In it's current state, this won't work once you leave the worksheet. This fixes that issue.~

~It does feel a bit eh doing it like this, so one alternative would be to just make this a client config value which defaults to false, and then for example in `coc-metals` and `vs code` we can have it set to true. Then, any other client that starts to support this, like the `lsp-mode` emacs folks https://github.com/emacs-lsp/lsp-mode/pull/1353, then they'd also have to pass in the property set to true. I sort of think this might be cleaner, but then it introduces another property we'd have to set. Thoughts?~

EDIT: A way simpler solution. This pr just adds in a check to see if the `focusDocument` is defined. If so, it means that the editor supports `didFocusTextDocument`. If that is true, we do an extra check to ensure that the worksheet is focused before re-publishing. If not, then we always re-publish.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1816,8 +1816,12 @@ class MetalsLanguageServer(
   ): Future[Unit] = {
     paths
       .find { path =>
-        focusedDocument.contains(path) &&
-        path.isWorksheet
+        if (focusedDocument.isDefined) {
+          focusedDocument.contains(path) &&
+          path.isWorksheet
+        } else {
+          path.isWorksheet
+        }
       }
       .fold(Future.successful(()))(
         worksheetProvider.evaluateAndPublish(_, EmptyCancelToken)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -116,6 +116,7 @@ class MetalsLanguageServer(
   private val openedFiles = new ActiveFiles(time)
   private val messages = new Messages(config.icons)
   private val languageClient = new DelegatingLanguageClient(NoopLanguageClient)
+  private var clientSupportsDidFocus = false
```
olafurpg: I just realized, can we use “focusedDocument.isDefined” instead?

ckipp01: ahhhh, good call. Yea, that is way simpler and also does the trick. Just pushed this up.

PR #1412, opened 2020-02-09
Use workspace level Gradle wrapper if present

Arthurm1:
For build import: choose workspace Gradle wrapper if present over downloading version 5.3.1
Update minimum Gradle version to 4.3 as that is the minimum Bloop now supports

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
@@ -11,11 +11,6 @@ case class GradleBuildTool(userConfig: () => UserConfiguration)
 
   private val initScriptName = "init-script.gradle"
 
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
```diff
@@ -76,7 +77,11 @@ case class GradleBuildTool(userConfig: () => UserConfiguration)
       case Some(script) =>
         script :: cmd
       case None =>
-        embeddedGradleLauncher.toString() :: cmd
+        val workspaceGradle = workspaceGradleLauncher(workspace)
+        if (workspaceGradle.toFile.exists())
+          workspaceGradle.toString() :: cmd
```
olafurpg: “workspaceGradle.isFile” should work

PR #1413, opened 2020-02-10
Enable metals/status for emacs.

kurnevsky:
Emacs supports it now: https://github.com/emacs-lsp/lsp-mode/pull/1402

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
@@ -148,6 +148,7 @@ object MetalsServerConfig {
         )
       case "emacs" =>
         base.copy(
+          statusBar = StatusBarConfig.on,
           executeClientCommand = ExecuteClientCommandConfig.on,
           compilers = base.compilers.copy(
             snippetAutoIndent = false


kurnevsky:
There is only one alternative implementation https://github.com/joaotavora/eglot and with "emacs" we already enabled client commands that are not supported by `eglot`. So I'd rather add another entry for `eglot`, or even remove `emacs` and add `emacs-eglot` and `emacs-lsp-mode`.

ckipp01:
> even remove emacs and add emacs-eglot and emacs-lsp-mode.

I think that this is probably the best option if no one else objects. I'd want to ensure we aren't  screwing up the display of messages for eglot users by doing this if they don't have a status integration.

Are you by chance able to split this up like you mentioned?

olafurpg:
I would propose to keep `emacs` unchanged for backwards compatibility and rather introduce a new `emacs-lsp-mode` client name.

Ideally, Metals should enable metals/status if the client declares the capability during `initialize` (like how we deal with other extensions). The reason we use system properties for metals/status is because it was one of the first extensions and I didn't know clients could include capabilities in the experimental section.

ckipp01:
Sounds good. 👍 

Then just to be crystal clear

[ ] Address @olafurpg 's point about the overview
[ ] Leave the curent emacs client as is, and create a new emacs-lsp-mode client name with then necessary changes

_Unless I'm completely misunderstanding this_

kurnevsky:
> I would propose to keep emacs unchanged for backwards compatibility and rather introduce a new emacs-lsp-mode client name.

Should I remove `executeClientCommand` and add `isHttpEnabled` to `emacs` entry then? Eglot doesn't support client commands and can't display doctor window. But then if users remain use `-Dmetals.client=emacs` with `lsp-mode` some things will stop working.

ckipp01:
> > I would propose to keep emacs unchanged for backwards compatibility and rather introduce a new emacs-lsp-mode client name.
> 
> Should I remove `executeClientCommand` and add `isHttpEnabled` to `emacs` entry then? Eglot doesn't support client commands and can't display doctor window. But then if users remain use `-Dmetals.client=emacs` with `lsp-mode` some things will stop working.

I'd say yes, if that's the level of support Eglot offers, then remove `executeClientCommand` and add `isHttpEnabled` for `emacs` and then create `emacs-lsp-mode` with the correct properties for `lsp-mode`.

kurnevsky:
So anyway we need to change `emacs` entry either for `eglot` or `lsp-mode`. Then probably it'd be better to add `emacs-eglot` as `lsp-mode` supposedly has more users?

Also as an idea - maybe just leave "default" client for that as a client with least supported features set? That'd mean to change default value for `isHttpEnabled`.

ckipp01:
> So anyway we need to change `emacs` entry either for `eglot` or `lsp-mode`. Then probably it'd be better to add `emacs-eglot` as `lsp-mode` supposedly has more users?
> 
> Also as an idea - maybe just leave "default" client for that as a client with least supported features set? That'd mean to change default value for `isHttpEnabled` and `snippetAutoIndent`.

Well if I understood what you said correctly, the changes that we'd make to the `emacs` won't have anny effect on Eglot users since those features aren't supported. However, if we rename it, then as @olafurpg mentioned, we lose backwards compatibility since they would then fall to `base`. 

Regarding what you said about "default", we actually have that right below the change you requested. It will default to `base`.

kurnevsky:
But anyway we loose backward compatibility, either for `lsp-mode` or `eglot`. If we decide `emacs` stands for `eglot` and remove `executeClientCommand` and add `isHttpEnabled` there that will mean some features will stop work for `lsp-mode`.

It's not that I'm strongly against of - just have a feeling that not all of options are considered. So if everyone is fine with that I'll update PR.

ckipp01:
> But anyway we loose backward compatibility, either for lsp-mode or eglot. If we decide emacs stands for eglot and remove executeClientCommand and add isHttpEnabled there that will mean some features will stop work for lsp-mode.

But a new `emacs-lsp-mode` would be created no? Therefore all of the `lsp-mode` users will now fall under this category moving forward correct?

kurnevsky:
Right, so it's either lsp-mode users will have to change the way they install metals or eglot users. And for either of them backward compatibility will be lost. And it's not very clear why we decide lsp-mode users should install it with new key but not eglot users :)

ckipp01:
> Right, so it's either lsp-mode users will have to change the way they install metals or eglot users. And for either of them backward compatibility will be lost. And it's not very clear why we decide lsp-mode users should install it with new key but not eglot users :)

Ahhh, of course. Sorry I was misunderstanding and assuming that the key would be automatically set my `lsp-mode`, but no. So either way we are breaking backwards compatibility for some users. However, this will need to happen sooner or later I'm assuming. _If_ that is indeed the case, and most users are `lsp-mode` users, maybe the smaller impact is actually the opposite of what I initially said, and leave `emacs` for `lsp-mode` users and then create an `emacs-eglot` client. Sorry for the confusion.

kurnevsky:
Actually lsp-mode has a new feature of installing server itself, but it's not implemented for metals yet. So theoretically it can set this key itself. But it's only for installation and we need to decide what key to set anyway. So what would be the decision? :)

ckipp01:
> So what would be the decision? :)

My opinion would be to make your changes to `emacs` and make the assumption that that is `lsp-mode`. Please also update the `overview.md` page as well. Then we can make another `lsp-eglot` if necessary.

ckipp01:
Hey @kurnevsky, now that #1414 is merged in, how would you like to proceed? We should be versioning in not too long again, so do you think it's preferable on the emacs side to just wait for that, and then add the new setting in via `clientExperimentalProperties`?

kurnevsky:
Yes, added them in https://github.com/emacs-lsp/lsp-mode/pull/1470 - won't harm to set them without waiting of metals release. It this pr I'll update docs later. One question left - what should we do about existing `metals.client=emacs` property? Should we remove it?

ckipp01:
Sounds good @kurnevsky. Just took a look at the pr that you got merged in `lsp-mode`. I think with that, there is no need to merge this in since `statusBarProvider` is being set anyways.

> won't harm to set them without waiting of metals release. 

 What you're seeing in master right now with experimental capabilities was merged in before the 0.8.1 release, so they should be there in the latest stable. :)

Thanks again for the conversation on this.

ckipp01:
> what should we do about existing metals.client=emacs property? Should we remove it?

Also sorry, forgot to answer this. For now, it will need to stay.

PR #1414, opened 2020-02-10
Move some properties to clientExperimentalCapabilities

ckipp01:
Following up a comment at @olafurpg said in here https://github.com/scalameta/metals/pull/1413, we actually have a few properties that we could just treat as `clientExperimentalCapabilities`. Just to see how easy this would be, I've taken the `inputBox` ripped it out of the server properties and listed it as an experimental feature. This was surprisingly easy, and there are multiple other settings that can be done this way. I also like that we minimize the amount of server properties.

This is the comment that sparked the idea:
> Ideally, Metals should enable metals/status if the client declares the capability during initialize (like how we deal with other extensions).

Here is the minimal changes that need to happen in a client. (~plus a small change in metals-languageserver to just remove the property~ not true, nothing should need to change there):

https://github.com/scalameta/coc-metals/pull/74
https://github.com/scalameta/metals-vscode/pull/207

If this is something we want to move towards, I'm happy to make the necessary changes in the vs code extension, ~metals-languageserver~, and also start hammering away at the other properties that can be treated as extensions.


ckipp01:
After a quick glance, we should be able to do this for the following properties:
- statusBar
- didFocus
- slowTask
- InputBox

kurnevsky:
Nice!
One more property to move to capabilities: executeClientCommand

metals/src/main/scala/scala/meta/internal/metals/SlowTaskConfig.scala
```diff
@@ -1,15 +0,0 @@
-package scala.meta.internal.metals
-
-final case class SlowTaskConfig(value: String) {
-  def isOff: Boolean = value == "off"
-  def isOn: Boolean = value == "on"
-  def isStatusBar: Boolean = value == "status-bar"
```
ckipp01: I didn't know this existed before. AFAIK, no clients use this. Is it safe to assume I can just remove this, and the remove the following check in `ConfiguredLanguageClient.scala`:

```scala
      if (config.slowTask.isStatusBar) {
        metalsStatus(MetalsStatusParams(params.message))
      }
```

ckipp01:
> One more property to move to capabilities: executeClientCommand
Thanks, added it above.

> Ideally, we should not specify editor when installing metals.
There is still some properties that don't really fit as `clientExperimentalCapabilities`, and even if there did, unfortunately we can't guarantee that every client will be able to set experimental capabilities.

kurnevsky:
> we can't guarantee that every client will be able to set experimental capabilities.

Why not? These experimental capabilities are useful mostly in the case when clients implement metals specific features - they have to be extendable and setting experimental capabilities shouldn't be too difficult.

kurnevsky:
> There is still some properties that don't really fit as clientExperimentalCapabilities

According to the spec experimental capabilities is an `any` which means it can have anything we need: https://microsoft.github.io/language-server-protocol/specifications/specification-current/

metals/src/main/scala/scala/meta/internal/metals/SlowTaskConfig.scala
olafurpg: Fine by me 👍

ckipp01:
> Thank you for looking into this! I agree this is a desirable change. Are we able to continue supporting the system properties so that existing clients don't break?

Sure, that shouldn't be a problem. In the case that a client as the same value set but set differently, which should win? The capabilities or the property?

ckipp01:
> > There is still some properties that don't really fit as clientExperimentalCapabilities
> 
> According to the spec experimental capabilities is an `any` which means it can have anything we need: https://microsoft.github.io/language-server-protocol/specifications/specification-current/

It's not so much about the _possiblity_ of whether it can or can't be done, but more about whether all the clients actually provide an easy way to do it, or really a way to do it at all. We support clients with varying levels of LSP support.

kurnevsky:
I'm still not sure how is that possible for client to implement custom metals extension but not able to set experimental capabilities. I believe for such clients default setting should suffice.

But anyway as @olafurpg noted we might have both capabilities and properties for backward compatibility.

My point is that if a client can set capabilities then it should be able to enable all necessary extensions itself without user interaction.

olafurpg:
@ckipp01 I think I’m ok with whichever wins as long as properties continue to work when capabilities are not set. It’s less of a concern  when clients set both properties since there’s nobody doing that right now.

ckipp01:
The following server properties have all been added to `ClientCapabilities` under `experimental`. However, the ability to set using properties still exists for backwards compatibility. I'll add in a couple notes in places to explain why I did what I did. I also added in a block of documentation regarding this.

 - statusBar
 - didFocus
 - slowTask
 - InputBox
 - execute-client-command
 - doctorFormat

There are corresponding prs for the shared library and various clients as well that can be found in the following places:
https://github.com/scalameta/metals-languageclient/pull/34
https://github.com/scalameta/metals-vscode/pull/207
https://github.com/scalameta/coc-metals/pull/74

metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
```diff
@@ -35,13 +35,13 @@ final class ConfiguredLanguageClient(
   }
 
   override def metalsStatus(params: MetalsStatusParams): Unit = {
-    if (config.statusBar.isOn) {
+    if (config.statusBar.isOn || clientCapabilities.statusBarIsOn) {
```
ckipp01: I tried to order these in level of "support". So if either is fully turned on, we'll go with that, then showMessage, and then LogMessage.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -69,10 +70,11 @@ final class Doctor(
       clientCommand: Command,
       onServer: MetalsHttpServer => Unit
   ): Unit = {
-    if (config.executeClientCommand.isOn) {
+    if (config.executeClientCommand.isOn || clientExperimentalCapabilities.executeClientCommandProvider) {
       val output =
-        if (config.doctorFormat.isHtml) buildTargetsHtml()
-        else buildTargetsJson()
+        if (config.doctorFormat.isJson || clientExperimentalCapabilities.doctorFormatIsJson)
```
ckipp01: I flipped the order of these since both default to html. So, I assume that if either is set to Json, it was intentionally set, and we then go with that.

docs/editors/new-editor.md
```diff
@@ -117,6 +119,9 @@ Possible values:
   `window/showMessage` notifications. Used by coc.nvim and sublime at the
   moment.
 
+*Usage of `statusBarProvider` in `ClientCapabilities.experimental` is
```
olafurpg: :+1:

ckipp01:
I have some consistent test failures. Looking into them 🤔

PR #1418, opened 2020-02-11
Send stackTraces as paths rather than URIs

danilbykov:
I'm trying to launch debugger from vim. I use this plugin
https://github.com/puremourning/vimspector

It starts and does first initialization query:
```
{
  "type": "request",
  "seq": 1,
  "command": "initialize",
  "arguments": {
    ....
    "pathFormat": "path",
    ....
  }
}
```
But even if it sets `pathFormat` to "path", metals returns path as URI in stackTraces (I get this log from dap-client.trace.json):
```
{
  "type": "response",
  "command": "stackTrace",
  "body": {
    "stackFrames": [
      {
        "id": 1,
        "source": {
          "name": "Test.scala",
          "path": "file:///home/dbykov/reps/test/src/main/scala/test/Test.scala",
          "sourceReference": 0
        },
      },
      ....
   ]
}
```

This breaks vimspector. I check dap-client.trace.json from python debugger and it returns these paths in path format.

So I made small change to send path depending on `pathFormat`. It fixes vimspector. I also did quick test in VScode and it still works.

What do you think?

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/MetalsDebugAdapters.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/MetalsDebugAdapters.scala
@@ -1,14 +1,17 @@
 package scala.meta.internal.metals.debug
 
+import java.nio.file.Paths
 import org.eclipse.lsp4j.debug.InitializeRequestArguments
 import org.eclipse.lsp4j.debug.{
   InitializeRequestArgumentsPathFormat => PathFormat
 }
-import java.nio.file.Paths
+import scala.meta.io.AbsolutePath
 
 final class MetalsDebugAdapters {
   private var lineAdapter: Long => Long = identity
-  private var pathAdapter: String => String = identity
+  private var clientToServerPathAdapter: String => String = identity
+  private var serverToClientPathAdapter: AbsolutePath => String =
+    p => p.toURI.toString
 
   def initialize(clientConfig: InitializeRequestArguments): Unit = {
     val shouldAdaptLines =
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/debug/MetalsDebugAdapters.scala
```diff
@@ -20,15 +23,19 @@ final class MetalsDebugAdapters {
...
       // ignore
     }
   }
 
   def adaptLine(line: Long): Long = lineAdapter(line)
 
-  def adaptPath(path: String): String = pathAdapter(path)
+  def adaptPath(path: String): String = clientToServerPathAdapter(path)
+
+  def adataPathForClient(path: AbsolutePath) = serverToClientPathAdapter(path)
```
tgodzik: Also maybe it should be `adaptPathForClient`? And maybe `adaptPath` -> `AdaptPathForServer`

gabro:
@olafurpg @danilbykov mentioned in the PR description that vscode was tested as well, so I think so

metals/src/main/scala/scala/meta/internal/metals/debug/MetalsDebugAdapters.scala
danilbykov: Done.

danilbykov:
Yes, I checked debug for my simple test application in vscode. I was able to move between files by clicking on lines in CALL STACK window.

But I can only test on Linux (Archlinux actually).

olafurpg:
Oh I’m sorry I missed that. Awesome! Thank you for this contribution 🙏

PR #1423, opened 2020-02-12
Refactor: split Completions.scala into smaller files

tanishiking:
> split Completions.scala into smaller files, it's at ~1.6k lines now with a lot of unrelated functionality grouped in one place

https://github.com/scalameta/metals/issues/672

---

This PR split `Completions.scala` into smaller files and reduced the size of `Completions.scala` from 1.6K lines to around 800 lines.

- move each concrete class of `CompletePosition` (e.g. `MatchKeyword`, `InterpolatorType`) into another files.
  - We already split some of them into smaller files
    - ScaladocCompletions.scala https://github.com/scalameta/metals/pull/1250#discussion_r363080599
    - OverrideCompletions.scala https://github.com/scalameta/metals/pull/1379
- move those files into `mtags/src/main/scala/scala/meta/internal/pc/completions/*.scala`

---

BTW, I used `ImportMissingSymbol` feature of Metals a lot for this PR, it's amazing :tada:

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
@@ -1,16 +1,11 @@
 package scala.meta.internal.pc
 
-import java.lang.StringBuilder
-import java.net.URI
-import java.nio.file.Paths
 import java.util.logging.Level
 
 import org.eclipse.{lsp4j => l}
 
 import scala.collection.mutable
-import scala.collection.mutable.ListBuffer
 import scala.meta.internal.jdk.CollectionConverters._
-import scala.meta.internal.mtags.MtagsEnrichments._
 import scala.meta.internal.semanticdb.Scala._
 import scala.meta.internal.tokenizers.Chars
 import scala.util.control.NonFatal
@@ -411,14 +406,14 @@ trait Completions { this: MetalsGlobal =>
         ident: Ident,
         apply: Apply
     ): CompletionPosition = {
-      if (CompletionPosition.hasLeadingBrace(ident, text)) {
+      if (hasLeadingBrace(ident, text)) {
         if (isCasePrefix(ident.name)) {
-          CompletionPosition.CaseKeyword(EmptyTree, editRange, pos, text, apply)
+          CaseKeyword(EmptyTree, editRange, pos, text, apply)
         } else {
           CompletionPosition.None
         }
       } else {
-        CompletionPosition.Arg(ident, apply, pos, text, completions)
+        Arg(ident, apply, pos, text, completions)
       }
     }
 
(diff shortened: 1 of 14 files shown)


mtags/src/main/scala/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,411 @@
...
+import org.eclipse.{lsp4j => l}
+
+import scala.collection.immutable.Nil
+import scala.meta.internal.pc.{Identifier, MetalsGlobal}
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+import scala.meta.internal.mtags.MtagsEnrichments._
+import scala.meta.internal.jdk.CollectionConverters._
+
+trait MatchCaseCompletions { this: MetalsGlobal =>
```
tanishiking: Contains `match` and `case` related logic.

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -690,708 +569,8 @@ trait Completions { this: MetalsGlobal =>
       override def isNew: Boolean = true
```
tanishiking: Leaving `Type`, `New`, and `None` in `object CompletionPosition` because those CompletionPositions are small enough, and I couldn't see the benefit of splitting them into smaller files.
However, it might be better to split them into smaller files for consistency. 🤔

tanishiking:
Oops, interpolator related tests failed... I'll fix it.

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
tgodzik: I think it would be better to be consistent here, but that's just my opinion. I am fine either way :)

mtags/src/main/scala/scala/meta/internal/pc/Completions.scala
```diff
@@ -1,16 +1,11 @@
 package scala.meta.internal.pc
 
-import java.lang.StringBuilder
```
tgodzik: Can we also move Completions class to the new package?

tanishiking:
My pleasure 😄 Thank you so much for reviewing!

> I think it would be better to be consistent here, but that's just my opinion. I am fine either way :)
> https://github.com/scalameta/metals/pull/1423#discussion_r378997306

I agree with this and moved those classes into small files as well https://github.com/scalameta/metals/pull/1423/commits/949b6e466ec4516c3e685e0b36962b7acb44f9f4
However, the names like `New` and `Type` conflict with the class names of Tree API, so renamed the concrete class of `CompletionPosition` to `XXXCompletion` (e.g. `NewCompletion`, `ScaladocCompletion`) What do you think about this?

tanishiking:
Thanks!

PR #1427, opened 2020-02-13
Fix issue with multiple pipes in string

tgodzik:
The issue explained in the test case.


metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -41,7 +41,7 @@ final class MultilineStringFormattingProvider(
   }
 
   private def determineDefaultIndent(sourceText: String, start: Int): String = {
-    val lastPipe = sourceText.lastIndexBetween('|', upperBound = start)
+    val lastPipe = sourceText.indexOf('|')
```
tgodzik: Since we check previously if there is a pipe in the line, the lastPipe inex will be correct.

Issue #1421, opened 2020-02-12
Rename variable removes backticks in pattern matching


LGLO:
**Describe the bug**

Having:
```scala
def f(x: Int): Unit = x match {
      case `myVar` => ()
      case _ => ()
    }
```
when I rename `myVar` to `xxx`, then backticks inside case are removed. This changes semantic of pattern match.

**To Reproduce** Steps to reproduce the behavior:

1. Rename any `val` that is used in `case` clause in pattern matching

**Expected behavior**

Backticks are not removed.

**Search terms**

rename backticks

gabro:
Hi, thanks for reporting! I’ve moved this issue to the main Metals issue tracker since this is not vscode specific.

I estimate it's an easy fix /cc @tgodzik

LGLO:
Thank you, sir!

PR #1430, opened 2020-02-13
Use proper range for references with backticks

tgodzik:
Previously rename would not work with backticks. Now, the reference provider returns proper ranges not including the backticks, we an exact match is needed for renames.

Fixes https://github.com/scalameta/metals/issues/1421

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
@@ -258,12 +258,22 @@ final class ReferenceProvider(
       if isSymbol(reference.symbol)
       if !reference.role.isDefinition || isIncludeDeclaration
       range <- reference.range.toList
-      if !checkMatchesText || reference.symbol.contains(
-        findName(range, snapshot.text)
-      )
     } {
-      add(range)
+
+      /* We skip checking if the symbol name matches exactly
+       * in case of finding references, where false positives
+       * are ok and speed is more important. This was needed
+       * for some issues with macro annotations, so with renames we
+       * must be sure that a proper name is replaced.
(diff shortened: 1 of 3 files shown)


tests/unit/src/test/scala/tests/RenameLspSuite.scala
```diff
@@ -502,18 +575,6 @@ class RenameLspSuite extends BaseLspSuite("rename") {
     newName = "Animal"
   )
 
-  renamed(
```
tgodzik: Moved up. This section is for known issues with rename

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -275,15 +284,40 @@ final class ReferenceProvider(
     buf.result()
   }
 
+  private def findRealRange(
+      range: s.Range,
+      text: String,
+      symbol: String
+  ): Option[s.Range] = {
+    val name = findName(range, text)
+    val isBackticked = name.charAt(0) == '`'
```
kpbochenek: you could extract `removeBackticks` function that would do the check and substring if needed.
it would be:
`val realName = removeBackticks(name)`
It could also be used in RenameProvider as removing backticks it the same there

tgodzik: I tried it, but I still need to check for backticks and change the range, so I would check it double. This is on the critical path, so I'd rather leave it as is.

kpbochenek: ahh right isBackticked is needed later for another check, good as it is then :+1:

PR #1435, opened 2020-02-14
Make sure we properly handle overflow events even if the path is null

tgodzik:
Previously, if the overflow event had a null path we wouldn't some references might get lost and things like renames wouldn't work properly. Now, when we are faced with an OVERFLOW event we reindex all build targets.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
@@ -11,6 +11,7 @@ import java.util.concurrent.Executors
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.io.AbsolutePath
 import java.{util => ju}
+import io.methvin.watcher.DirectoryChangeEvent.EventType
 
 /**
  * Handles file watching of interesting files in this build.
@@ -39,32 +40,25 @@ final class FileWatcher(
 ) extends Cancelable {
 
   private val directoryExecutor = Executors.newFixedThreadPool(1)
-  private val fileExecutor = Executors.newFixedThreadPool(1)
 
   ThreadPools.discardRejectedRunnables(
     "FileWatcher.executor",
     directoryExecutor
   )
-  ThreadPools.discardRejectedRunnables("FileWatcher.fileExecutor", fileExecutor)
 
   private var activeDirectoryWatcher: Option[DirectoryWatcher] = None
-  private var activeFileWatcher: Option[DirectoryWatcher] = None
 
   private var directoryWatching: CompletableFuture[Void] =
     new CompletableFuture()
-  private var fileWatching: CompletableFuture[Void] = new CompletableFuture()
 
   override def cancel(): Unit = {
     stopWatching()
     directoryExecutor.shutdown()
-    fileExecutor.shutdown()
     activeDirectoryWatcher.foreach(_.close())
-    activeFileWatcher.foreach(_.close())
   }
 
   def restart(): Unit = {
     val sourceDirectoriesToWatch = new util.LinkedHashSet[Path]()
-    val sourceFilesToWatch = new util.ArrayList[Path]()
     val createdSourceDirectories = new util.ArrayList[AbsolutePath]()
     def watch(path: AbsolutePath, isSource: Boolean): Unit = {
       if (!path.isDirectory && !path.isFile) {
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -202,6 +202,7 @@ final class ReferenceProvider(
         scalaPath <- SemanticdbClasspath
           .toScala(workspace, AbsolutePath(path))
           .iterator
+        if scalaPath.exists
```
tgodzik: If we removed some files, we would still be looking for references inside them, since the semanticDB files will not be deleted.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -857,33 +860,40 @@ class MetalsLanguageServer(
...
-    } else if (isScalaOrJava && !savedFiles.isRecentlyActive(path)) {
-      event.eventType() match {
-        case EventType.CREATE =>
-          buildTargets.onCreate(path)
-        case _ =>
-      }
-      onChange(List(path)).asJava
-    } else if (path.isSemanticdb) {
+    if (event.eventType() == EventType.OVERFLOW && event.path() == null) {
       CompletableFuture.completedFuture {
```
tgodzik: OVERFLOW event usually had null as path, so it was never addressed. I tested it by creating 1k files automatically.

metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
```diff
@@ -129,22 +123,14 @@ final class FileWatcher(
...
     }
-    activeFileWatcher.foreach(_.close())
-    fileWatching.cancel(false)
     directoryWatching.cancel(false)
   }
 
   class DirectoryListener extends DirectoryChangeListener {
     override def onEvent(event: DirectoryChangeEvent): Unit = {
-      if (!Files.isDirectory(event.path())) {
+      if (event.eventType() == EventType.OVERFLOW) {
```
kpbochenek: we could add a comment why it has dedicated branch that is doing the same logic.

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
kpbochenek: Can we simply delete this file?

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
```diff
@@ -39,6 +40,20 @@ class SemanticdbIndexer(
...
+  /**
+   * Handle EventType.OVERFLOW, meaning we lost file events, when we don't know the path.
+   * We walk up the file tree for all targets `META-INF/semanticdb` directory
+   * and re-index all of its `*.semanticdb` children.
+   */
+  def onOverflow(): Unit = {
+    buildTargets.scalacOptions.foreach { item =>
+      val targetroot = item.targetroot
+      if (!targetroot.isJar) {
+        onChangeDirectory(targetroot.resolve(Directories.semanticdb).toNIO)
```
kpbochenek: Something to take into consideration, we use FixedThreadPool(1) to handle events, because reindexing is sync, we will stop processing events till reindexing is done.

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
tgodzik: That file is managed by Bloop and we shouldn't remove it ourselves.

metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
tgodzik: added

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
tgodzik: changed to using future there

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -799,7 +802,7 @@ class MetalsLanguageServer(
       } {
         val event =
           new DirectoryChangeEvent(EventType.MODIFY, generatedFile.toNIO, 1)
-        didChangeWatchedFiles(event)
+        didChangeWatchedFiles(event).get()
```
tgodzik: Since it's for the test, it's better that we wait.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -857,33 +860,39 @@ class MetalsLanguageServer(
...
-            semanticDBIndexer.onChange(event.path())
-          case EventType.OVERFLOW =>
-            semanticDBIndexer.onOverflow(event.path())
+          case EventType.CREATE =>
+            buildTargets.onCreate(path)
+          case _ =>
         }
+        onChange(List(path)).asJava
+      } else if (path.isSemanticdb) {
+        Future {
```
tgodzik: It's now run in a Future in order not to block the event watcher

PR #1439, opened 2020-02-15
Mark most often failing tests as flaky and fix DefinitionLspSuite test

tgodzik:
The tests marked a flaky were failing quite often and causing a lot of noise. We might need to add more later on.

I also fixed the `missing-compiler-plugin` test and unmarked as flaky. If it proves to be still problematic I will create a follow up PR to remark as flaky.

BuildServerConnectionLspSuite seems to be stable now, so I also reinstated it.


tests/unit/src/test/scala/tests/DefinitionLspSuite.scala
```diff
@@ -182,7 +182,7 @@ class DefinitionLspSuite extends BaseLspSuite("definition") {
   // requires org.scalamacros:macroparadise and io.spire-match:kind-projector.
   // Navigation continues to mostly work, except for areas that have compilation
   // errors.
-  test("missing-compiler-plugin".flaky) {
+  test("missing-compiler-plugin") {
```
olafurpg: The purpose of this test was to document a known limitation when the dependency classpath was missing symbols that are referenced in the library sources. It looks like we bumped up the cats dependency to a version that brings in simulacrum 🤔

Either way, we could remove this test instea d

tgodzik:
> Thank you for doing this. I was just looking at the test reports today and noticed we still have quite a few flaky tests with 2-3% failure rate.

I think we have marked all the biggest culprits already, but will follow up if anything else shows up.

tests/unit/src/test/scala/tests/DefinitionLspSuite.scala
tgodzik: Ok cool! I removed it!

PR #1447, opened 2020-02-17
Add new LSP extension 'metals/pickInput' to implement the "Create Scala file" command

alekseiAlefirov:
After new-scala-file command has been added to metals https://github.com/scalameta/metals/pull/1339 , client-side service was discussed https://github.com/scalameta/metals-vscode/pull/183 , and it was decided to move and implemet quick-pick communication to server.

### DIFF
--- a/docs/editors/new-editor.md
+++ b/docs/editors/new-editor.md
@@ -671,6 +671,7 @@ rather then using server properties. The currently available settings for
       "debuggingProvider": boolean,
       "decorationProvider": boolean,
       "inputBoxProvider": boolean,
+      "pickInputProvider": boolean,
       "didFocusProvider": boolean,
       "slowTaskProvider": boolean,
       "executeClientCommandProvider": boolean,
--- a/docs/editors/new-editor.md
+++ b/docs/editors/new-editor.md
@@ -141,6 +141,16 @@ Possible values:
 
 *Usage of `inputBoxProvider` in `ClientCapabilities.experimental` is preferable.*
 
+### `-Dmetals.pick-input`
+
+Possible values:
+
+- `off` (default): the `metals/pickInput` request is not supported. In this case,
+  Metals fallbacks to `window/showMessageRequest`.
+- `on`: the `metals/pickInput` request is fully supported.
+
+*Usage of `pickInputProvider` in `ClientCapabilities.experimental` is preferable.*
+
 ### `-Dmetals.execute-client-command`
 
 Possible values:
@@ -537,6 +547,73 @@ export interface MetalsInputBoxParams {
 ```ts
 export interface MetalsInputBoxResult {
   value?: string;
+  cancelled?: boolean;
+}
+```
+
+### `metals/pickInput`
+
+The Metals pick input request is sent from the server to the client to let the
+user provide a string value by picking one out of a number of given options. It is similar to `window/showMessageRequest`, but the `metals/pickInput` request has richer parameters, that can be used to filter items to pick, like `description` and `detail`.
+
+_Request_:
+
+- method: `metals/pickInput`
+- params: `MetalsPickInputParams` defined as follows. It partially matches [`QuickPickOptions`](https://code.visualstudio.com/api/references/vscode-api#QuickPickOptions) in the Visual Studio Code API, but it also contains `items` of [`MetalsPickItem`](https://code.visualstudio.com/api/references/vscode-api#QuickPickItem), which, in it's turn, partially matches `QuickPickItem`, but these interfaces do not contain options for picking many items:
+
+```ts
+export interface MetalsPickInputParams {
+  /**
+   * An array of items that can be selected from.
+   */
+  items: MetalsPickItem[];
+  /**
+   * An optional flag to include the description when filtering the picks.
+   */
+  matchOnDescription?: boolean;
+  /**
+   * An optional flag to include the detail when filtering the picks.
+   */
+  matchOnDetail?: boolean;
+  /**
+   * An optional string to show as place holder in the input box to guide the user what to pick on.
+   */
+  placeHolder?: string;
+  /**
+   * Set to `true` to keep the picker open when focus moves to another part of the editor or to another window.
+   */
+  ignoreFocusOut?: boolean;
+}
+
+export interface MetalsPickItem {
+  /**
+   * An id for this items that should be return as a result of the picking.
+   */
+  id: string;
+  /**
+   * A human readable string which is rendered prominent.
+   */
+  label: string;
+  /**
+   * A human readable string which is rendered less prominent.
+   */
+  description?: string;
+  /**
+   * A human readable string which is rendered less prominent.
+   */
+  detail?: string;
+  /**
+   * Always show this item.
+   */
+  alwaysShow?: boolean;
+}
+```
+
+- result: `MetalsPickInputResult` defined as follows:
+
+```ts
+export interface MetalsPickInputResult {
+  itemId?: string;
   cancelled?: boolean;
 }
 ```
--- a/docs/editors/new-editor.md
+++ b/docs/editors/new-editor.md
@@ -541,22 +541,22 @@ export interface MetalsInputBoxResult {
 }
 ```
 
-### `metals/pickInput`
+### `metals/quickPick`
 
-The Metals pick input request is sent from the server to the client to let the
-user provide a string value by picking one out of a number of given options. It is similar to `window/showMessageRequest`, but the `metals/pickInput` request has richer parameters, that can be used to filter items to pick, like `description` and `detail`.
+The Metals quick pick request is sent from the server to the client to let the
+user provide a string value by picking one out of a number of given options. It is similar to `window/showMessageRequest`, but the `metals/quickPick` request has richer parameters, that can be used to filter items to pick, like `description` and `detail`.
 
 _Request_:
 
-- method: `metals/pickInput`
-- params: `MetalsPickInputParams` defined as follows. It partially matches [`QuickPickOptions`](https://code.visualstudio.com/api/references/vscode-api#QuickPickOptions) in the Visual Studio Code API, but it also contains `items` of [`MetalsPickItem`](https://code.visualstudio.com/api/references/vscode-api#QuickPickItem), which, in it's turn, partially matches `QuickPickItem`, but these interfaces do not contain options for picking many items:
+- method: `metals/quickPick`
+- params: `MetalsQuickInputParams` defined as follows. It partially matches [`QuickPickOptions`](https://code.visualstudio.com/api/references/vscode-api#QuickPickOptions) in the Visual Studio Code API, but it also contains `items` of [`MetalsQuickPickItem`](https://code.visualstudio.com/api/references/vscode-api#QuickPickItem), which, in it's turn, partially matches `QuickPickItem`, but these interfaces do not contain options for picking many items:
 
 ```ts
-export interface MetalsPickInputParams {
+export interface MetalsQuickPickParams {
   /**
    * An array of items that can be selected from.
    */
-  items: MetalsPickItem[];
+  items: MetalsQuickPickItem[];
   /**
    * An optional flag to include the description when filtering the picks.
    */
@@ -575,7 +575,7 @@ export interface MetalsPickInputParams {
   ignoreFocusOut?: boolean;
 }
 
-export interface MetalsPickItem {
+export interface MetalsQuickPickItem {
   /**
    * An id for this items that should be return as a result of the picking.
    */
@@ -599,10 +599,10 @@ export interface MetalsPickItem {
 }
 ```
 
-- result: `MetalsPickInputResult` defined as follows:
+- result: `MetalsQuickPickResult` defined as follows:
 
 ```ts
-export interface MetalsPickInputResult {
+export interface MetalsQuickPickResult {
   itemId?: string;
   cancelled?: boolean;
 }
@@ -661,7 +661,7 @@ rather then using server properties. The currently available settings for
       "debuggingProvider": boolean,
       "decorationProvider": boolean,
       "inputBoxProvider": boolean,
-      "pickInputProvider": boolean,
+      "quickPickProvider": boolean,
       "didFocusProvider": boolean,
       "slowTaskProvider": boolean,
       "executeClientCommandProvider": boolean,
--- a/docs/editors/new-editor.md
+++ b/docs/editors/new-editor.md
@@ -141,16 +141,6 @@ Possible values:
 
 *Usage of `inputBoxProvider` in `ClientCapabilities.experimental` is preferable.*
 
-### `-Dmetals.pick-input`
-
-Possible values:
-
-- `off` (default): the `metals/pickInput` request is not supported. In this case,
-  Metals fallbacks to `window/showMessageRequest`.
-- `on`: the `metals/pickInput` request is fully supported.
-
-*Usage of `pickInputProvider` in `ClientCapabilities.experimental` is preferable.*
-
 ### `-Dmetals.execute-client-command`
 
 Possible values:
--- a/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
@@ -106,16 +106,16 @@ final class ConfiguredLanguageClient(
     }
   }
 
-  override def metalsPickInput(
-      params: MetalsPickInputParams
-  ): CompletableFuture[MetalsPickInputResult] = {
-    if (clientCapabilities.pickInputProvider) {
-      underlying.metalsPickInput(params)
+  override def metalsQuickPick(
+      params: MetalsQuickPickParams
+  ): CompletableFuture[MetalsQuickPickResult] = {
+    if (clientCapabilities.quickPickProvider) {
+      underlying.metalsQuickPick(params)
     } else {
       showMessageRequest(
         toShowMessageRequestParams(params)
       ).asScala
-        .map(item => MetalsPickInputResult(itemId = item.getTitle()))
+        .map(item => MetalsQuickPickResult(itemId = item.getTitle()))
         .asJava
     }
   }
@@ -129,7 +129,7 @@ final class ConfiguredLanguageClient(
   }
 
   private def toShowMessageRequestParams(
-      params: MetalsPickInputParams
+      params: MetalsQuickPickParams
   ): ShowMessageRequestParams = {
     val result = new ShowMessageRequestParams()
     result.setMessage(params.placeHolder)
--- a/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
@@ -106,12 +106,33 @@ final class ConfiguredLanguageClient(
     }
   }
 
+  override def metalsPickInput(
+      params: MetalsPickInputParams
+  ): CompletableFuture[MetalsPickInputResult] = {
+    if (config.isPickInputEnabled) {
+      underlying.metalsPickInput(params)
+    } else {
+      showMessageRequest(params).asScala
+        .map(item => MetalsPickInputResult(itemId = item.getTitle()))
+        .asJava
+    }
+  }
+
   override def metalsPublishDecorations(
       params: PublishDecorationsParams
   ): Unit = {
     if (clientCapabilities.decorationProvider) {
       underlying.metalsPublishDecorations(params)
     }
+  }
+
+  private implicit def metalsPickInputParams2ShowMessageRequestInputParams(
+      params: MetalsPickInputParams
+  ): ShowMessageRequestParams = {
+    val result = new ShowMessageRequestParams()
+    result.setMessage(params.placeHolder)
+    result.setActions(params.items.map(item => new MessageActionItem(item.id)))
+    result
   }
 
 }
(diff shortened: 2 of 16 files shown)


metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -7,51 +7,120 @@ import MetalsEnrichments._
...
-    case "worksheet" =>
-      createWorksheet(directory, name)
-    case invalid => Future.failed(new IllegalArgumentException(invalid))
+  ): Future[Unit] = {
+    val directory = directoryUri
+      .map(_.toString.toAbsolutePath)
+      .orElse(focusedDocument.map(_.parent))
+
+    val newlyCreatedFile =
+      askForKind
```
alekseiAlefirov: I've made some kind of DSL here, to cover `Future[Option[_]]` handling (making this method more or readable). `Future` is for about remote requests, and `Option` reflect that the request can be cancelled (thus, nothing should be returned eventually).
Not sure, if it's good here.

tests/unit/src/main/scala/tests/TestingClient.scala
```diff
@@ -71,6 +71,7 @@ final class TestingClient(workspace: AbsolutePath, buffers: Buffers)
     _: ShowMessageRequestParams =>
       None
   }
+  var inputBoxHandler: MetalsInputBoxParams => Option[MetalsInputBoxResult] = _
```
alekseiAlefirov: while picking kind of file can be expressed via `showMessageRequest`, asking for name of the new file cannot, so I had to add this handler for testing purposes.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -106,6 +111,8 @@ object MetalsServerConfig {
           openFilesOnRenames = true,
           executeClientCommand = ExecuteClientCommandConfig.on,
           globSyntax = GlobSyntaxConfig.vscode,
+          isInputBoxEnabled = true,
```
alekseiAlefirov: I had to change it as well, because entering name for the new file cannot be expressed via `showMessageRequest`. This resulted into a bit changed UI when formatting and no scalafmt version is set.
Also, I would expect for such configs to be set by client, but okay, I mimicked it with `isPickInputEnabled`.

docs/editors/new-editor.md
```diff
@@ -134,6 +134,14 @@ Possible values:
   Metals tries to fallback to `window/showMessageRequest` when possible.
 - `on`: the `metals/inputBox` request is fully supported.
 
+### `-Dmetals.pick-input`
```
ckipp01: Related to some of the work I started in here: https://github.com/scalameta/metals/pull/1414, is it possible to not introduce this as a property, but rather as a `clientExperimentalCapabilities`? I feel like this would fit well there.

alekseiAlefirov: Agreed, as I said in my previous comment, I would expect it to be a client config, rather then a server one. I've just repeated after `input-box`. If it's to be changed -- then this is to be changed as well :)

tests/unit/src/main/scala/tests/TestingClient.scala
tgodzik: Can you add a default implementation returning some random name? You can have `var nameToPick = "Foo"` and replace that if needed. You also have None as the default if it's an issue.

Otherwise, we will get a null pointer in case of some particular changes.

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
```diff
@@ -2,17 +2,36 @@ package tests
...
+  override def serverConfig: MetalsServerConfig =
+    super.serverConfig.copy(isInputBoxEnabled = true)
+
   test("new-worksheet") {
     cleanCompileCache("a")
     RecursivelyDelete(workspace.resolve("a"))
     Files.createDirectories(
       workspace.resolve("a/src/main/scala/").toNIO
     )
+    client.showMessageRequestHandler = { params =>
```
tgodzik: Maybe a utility closure:
```
def createMessageRequestHander(name : String) : MessageRequestParams => Option[MessageRequestResult] =
{ params =>
      if (NewScalaFile.isSelectTheKindOfFile(params)) {
        params.getActions().asScala.find(_.getTitle() == name)
      } else {
        None
      }
    }
```

then you can just use:
`client.showMessageRequestHandler = createMessageRequestHander("worksheet") `

tgodzik: Or even better create a check method:
```
def check(picked : String, expectedOutput : String, name : String = "Foo", shouldCheckForName : Boolean)
```
then you can just have:
```
check(
  "object",
 """| object Foo {
     |}
     |""".stripMargin
```

createMessageRequestHande might not be needed then.

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -66,31 +135,80 @@ class NewFilesProvider(
...
         |}
         |""".stripMargin
 
 }
+
+object DialogEnrichments {
+
+  implicit class DialogContinuation[A](state: Future[Option[A]]) {
+
+    def continueWith[B](
```
tgodzik: Looks good, this is almost a MonadTransformer haha.

Maybe `flatMapOption` and `mapOption` ? This will be a bit more natural.

And these methods might be actually useful to add to `MetalsEnrichments`, I remember situations when we were combining Futures and Options.

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -66,31 +135,80 @@ class NewFilesProvider(
...
+  }
+
   private def createFile(
       path: AbsolutePath
-  )(implicit ec: ExecutionContext): Future[URI] = {
+  )(implicit ec: ExecutionContext): Future[AbsolutePath] = {
     val result = Future {
-      Files.createFile(path.toNIO).toUri()
+      AbsolutePath(
+        Files.createFile(path.toNIO)
```
tgodzik: You can use `touch` from `XtensionAbsolutePathBuffers`

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -7,51 +7,120 @@ import MetalsEnrichments._
...
+  }
+
+  private def createClass(
+      directory: Option[AbsolutePath],
+      name: String,
+      kind: String
+  )(
+      implicit ec: ExecutionContext
+  ): Future[AbsolutePath] = {
+    val path = directory.getOrElse(workspace).resolve(name + ".scala")
```
tgodzik: What happens if the name is `a/b/c/A` - I think we should create the directories there. Most likely `createFile` should also create directories? Package will then be added automatically which is neat.

docs/editors/new-editor.md
```diff
@@ -529,6 +537,73 @@ export interface MetalsInputBoxResult {
...
+```ts
+export interface MetalsPickInputParams {
+  /**
+   * An array of items that can be selected from.
+   */
+  items: MetalsPickItem[];
+  /**
+   * An optional flag to include the description when filtering the picks.
+   */
+  matchOnDescription?: boolean;
```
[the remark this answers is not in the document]
alekseiAlefirov: good question to discuss.
I've added them to match, yes, to be able to use the most of VSCode API.
But if we don't want them - we can remove them, they'll be initialized to default values on VSCode side.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -106,6 +111,8 @@ object MetalsServerConfig {
...
+          isPickInputEnabled = true,
```
ckipp01: As a follow-up to my previous comment, and similiar to what @tgodzik mentioned, if you rebase on master now, you'll see the pattern I followed in #1414. Moving forward for _most cases_, we shouldn't need to add new server properties, but rather list them under `ClientExperimentalCapabilities`.

docs/editors/new-editor.md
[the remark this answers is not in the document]
tgodzik: It should be fine, we can leave it as is. Unless they might cause issues for implementing this in other clients?

metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
```diff
@@ -117,4 +129,13 @@ final class ConfiguredLanguageClient(
     }
   }
 
+  private implicit def metalsPickInputParams2ShowMessageRequestInputParams(
```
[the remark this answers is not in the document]
alekseiAlefirov: done, changed to
```
private def toShowMessageRequestParams(
      params: MetalsPickInputParams
  ): ShowMessageRequestParams
```

docs/editors/new-editor.md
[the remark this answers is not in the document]
alekseiAlefirov: I think, it won't be no problem, thanks to JS agility. Even now it got implicitly converted from Metals interface to VSCode one.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageClient.scala
```diff
@@ -112,3 +123,33 @@ case class MetalsInputBoxResult(
     @Nullable value: String = null,
     @Nullable cancelled: java.lang.Boolean = null
 )
+
+case class MetalsPickInputParams(
```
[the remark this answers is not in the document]
alekseiAlefirov: I just followed `MetalsInputBoxParams`, defined in this file before.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1266,27 +1273,18 @@ class MetalsLanguageServer(
...
-                    s"Cannot create file:\n ${e.toString()}"
-                  )
-            }
-            result
+          case Seq(directory: JsonPrimitive) if directory.isString =>
+            newFilesProvider.createNewFileDialog(
+              Some(directory.getAsString()).map(new URI(_))
+            )
+          case Seq(_: JsonNull) =>
+            newFilesProvider.createNewFileDialog(None)
```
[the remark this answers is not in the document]
alekseiAlefirov: done

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -66,31 +135,80 @@ class NewFilesProvider(
...
+      path
     }
   }
 
+  private def openFile(path: AbsolutePath): Unit = {
+    client.metalsExecuteClientCommand(
+      new ExecuteCommandParams(
+        ClientCommands.GotoLocation.id,
+        List(
+          new Location(path.toString(), new org.eclipse.lsp4j.Range()): Object
```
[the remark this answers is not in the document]
alekseiAlefirov: `Range` fixed.
Good catch with `toURI`, thanks!
`:Object` can not be ommitted, since it's Java List eventually, and it's not covariant. It's the same with already existing cases of calling `ClientCommands.GotoLocation`

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -382,4 +382,15 @@ class Messages(icons: Icons) {
         s"Please upgrade to Scala $recommended."
     }
   }
+
+  object NewScalaFile {
+    def selectTheKindOfFileMessage = "Select the kind of file to create"
+    def enterNameMessage(kind: String): String = s"Enter name for the new $kind"
+
+    def isSelectTheKindOfFile(params: ShowMessageRequestParams): Boolean =
```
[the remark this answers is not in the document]
alekseiAlefirov: It's the same with other `is...` methods in this file - they are used only in tests.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -382,4 +382,15 @@ class Messages(icons: Icons) {
         s"Please upgrade to Scala $recommended."
     }
   }
+
+  object NewScalaFile {
+    def selectTheKindOfFileMessage = "Select the kind of file to create"
+    def enterNameMessage(kind: String): String = s"Enter name for the new $kind"
```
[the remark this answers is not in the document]
alekseiAlefirov: fixed

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -7,51 +7,120 @@ import MetalsEnrichments._
...
+        case _ => None
+      }
+  }
+
+  private def createClass(
+      directory: Option[AbsolutePath],
+      name: String,
+      kind: String
+  )(
+      implicit ec: ExecutionContext
```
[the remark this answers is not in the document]
alekseiAlefirov: done

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
alekseiAlefirov: not sure, if I got you correctly, but I've done something there - please, check.

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
alekseiAlefirov: Yeah, I also thought about `OptionT` :D . Done

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
alekseiAlefirov: `touch` does not return the path, and although I already have the path, I like to reuse output of the function.

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
alekseiAlefirov: it fails. Not really sure, it looks neat to ask for the name for the new class `foo/Bar` to create `class Bar`at `foo/Bar.scala`

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
alekseiAlefirov: I've added client properies, but server ones are still there, cause `https://github.com/scalameta/metals-vscode/pull/207` is not being pushed.

ckipp01: But that's because those properties already exist and are relied upon via clients that are already using them. This is a brand new one that we are introducing, so I'm not sure I understand why we also need to support this with client properties since nothing is relying on it? We can just only support it via `ClientExperimentalCapabilities`, and introduce it in vscode via experimentl.

docs/editors/new-editor.md
tgodzik: Let's use the recently merged way defining capabilities.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
alekseiAlefirov: `isPickInputEnabled` removed now, and capability added to the client in https://github.com/scalameta/metals-vscode/pull/183. Though `isInputBoxEnabled = true` still in this place, I am not sure, if it should not be here now.

ckipp01: Yea,  for now that's fine since that property already existed anyways. Once we release, and the the vs code pr gets merged, we can then remove the `isInputBoxEnabled` since then it will be set via experimental from vs code.

docs/editors/new-editor.md
```diff
@@ -541,6 +541,73 @@ export interface MetalsInputBoxResult {
 }
 ```
 
+### `metals/pickInput`
```
olafurpg: Would it make sense to name this `metals/quickPick`? In general I try to follow the same naming convention as in the VS Code APIs

alekseiAlefirov: Yeah, I was not sure about the name, so if you think, this one's better, okay then. Renamed.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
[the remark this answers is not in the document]
tgodzik: That's what I meant, I think they can just be move to the test class.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -602,4 +602,25 @@ object MetalsEnrichments
...
+
+    def mapOption[B](
+        f: A => Future[B]
+    )(implicit ec: ExecutionContext): Future[Option[B]] =
+      state.flatMap(
+        _.fold(Future.successful(Option.empty[B]))(f(_).liftOption)
+      )
+  }
+
+  implicit class OptionFutureLift[A](state: Future[A]) {
```
tgodzik: There should already be `XtensionScalaFuture`

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
tgodzik: I would even remove it from here, since we want to drop the properties.

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
tgodzik: Could we implement it here? I think we just need to do `getParent.mkdirs`

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
tgodzik: Can we change touch to return the path?

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
```diff
@@ -155,26 +132,64 @@ class NewFilesLspSuite extends BaseLspSuite("new-files") {
...
+        workspace.resolve("a/src/main/scala/foo/package.scala"),
         """|package object foo {
            |  
            |}
            |""".stripMargin
       )
+    } yield ()
+  }
 
+  private def check(
```
tgodzik: I like the direction, but I thought of something similar to RenameLspSuite and other, where we would even invoke `test` from here.

alekseiAlefirov:
@tgodzik 
>> isInputBoxEnabled = true,

> I would even remove it from here, since we want to drop the properties.

What should I do with the client part https://github.com/scalameta/metals-vscode/pull/183 ?
- I could add `inputBox` capability, but it's already in @ckipp01 's PR https://github.com/scalameta/metals-vscode/pull/207
- I could rebase on that PR, so it is adequate to this PR without the `isInputBoxEnabled` line
- I could leave it, just wait for that PR to be merged.

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
alekseiAlefirov: done

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
alekseiAlefirov: done

alekseiAlefirov:
@tgodzik PR reworked,
> That's what I meant, I think they can just be move to the test class.

Moved.
> I would even remove it from here, since we want to drop the properties.

removed, but see previous comment.

> Could we implement it here? I think we just need to do getParent.mkdirs

done.

> Can we change touch to return the path?

Nah, I just returned path there (and changed `Files.create` to `touch`).

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -382,4 +382,10 @@ class Messages(icons: Icons) {
         s"Please upgrade to Scala $recommended."
     }
   }
+
+  object NewScalaFile {
+    def selectTheKindOfFileMessage = "Select the kind of file to create"
+    def enterNameMessage(kind: String): String =
+      s"Enter the name for the new $kind"
```
ckipp01: Super nitpicky, but I just tested this, and is it possible to just add a `:` in here?
```suggestion
      s"Enter the name for the new $kind:"
```

alekseiAlefirov: it seems not to be going look good here:
image

tgodzik: Let's leave it as is, I don't think it adds that much visually and it's not really being used in UI anywhere.

ckipp01: Hmmm, yea I see what you're saying. That's tricky, because in `coc-metals` there isn't a nice break, so it seems like there should really be a `:`. No worries, I'll just add it in on the vim side.

Screenshot 2020-02-24 at 14 01 42

ckipp01: Also, this feature works great 🎉 !!

alekseiAlefirov: thanks :relieved:

PR #1450, opened 2020-02-19
Index by scala filename to avoid duplicates

tgodzik:
Previously, if we had two targets for one file, like for example JS and JVM, we would get duplicated results due to 2 different semanticDB files being created. Now we index by path to the scala file itself, which is not duplicate.


tgodzik:
Confirmed with @olafurpg that it's not a good way to fix it.

PR #1452, opened 2020-02-19
Make use of latest Pants and Bloop improvements

olafurpg:
Previously, the `fastpass` command-line was compatible with most recent
versions of Pants. Now, it requires Pants v1.24 or newer. The motivation
for this change was to take advantage of the new
`./pants export-dep-as-jar` goal that was added specifically to support
an exporter tool like fastpass.

The benefits of this changes include:

* faster export times
* sources.jar for non-exported targets fixing navigation bugs
* simpler logic overall

Additionally, this commit
* changes the generated Bloop JSON configuration to use the new `sourcesGlobs` field from https://github.com/scalacenter/bloop/pull/1176 (this change was necessary in order to consume `./pants export-dep-as-jar`
* upgrades to the latest Metaconfig in order to support tab completions.


metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
```diff
@@ -538,69 +503,58 @@ private class BloopPants(
...
-            case (entry, i) => exportClasspathJar(target, entry, s"-$i")
-          }.toList
-        } else {
-          Nil
+  private def toImmutableJar(filename: String, path: Path): Path = {
+    // NOTE(olafur): Jars that live inside $WORKSPACE/.pants.d get overwritten
+    // by Pants during compilation. We copy these jars over to the Bloop
+    // directory so that the Bloop incremental cache is unaffected by the
+    // `./pants compile` tasks that the user is running.
+    if (path.startsWith(mutableJarsHome)) {
```
olafurpg: Just found an issue because this method doesn't handle directory entries. Will need to fix before merge

olafurpg:
Ignoring unrelated failure
```
[error] ==> X tests.debug.BreakpointDapSuite.java-nested  2.141s java.lang.IllegalStateException: ERROR: JDWP Unable to get JNI 1.2 environment, jvm->GetEnv() return code = -2
```

PR #1454, opened 2020-02-19
Recognize .srcjar extension as jar files

carmen-stripe:
- Bazel uses .srcjar extension to recognize files that need to be
unpacked and compiled.
- For this convention, `rules_scala`'s emits protobuf source jars
with .srcjar extension.
- As is, providing the path to a .srcjar file in the resolution field
of a Bloop configuration file does not work in Metals because it isn't
recognized as a jar file.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -306,7 +306,7 @@ object MetalsEnrichments
 
     def isJar: Boolean = {
       val filename = path.toNIO.getFileName.toString
-      filename.endsWith(".jar")
+      filename.endsWith(".jar") || filename.endsWith(".srcjar")
     }
 
     /**


olafurpg:
Thank you for this contribution! Could you maybe explain a bit about the motivation behind using another extension than `.jar` for jar files? Would it be possible to rename the files before passing them to Bloop/Metals? A lot of tools that work with JVM classpaths assume that jars use the `.jar` file extension, for better or worse.

carmen-stripe:
`.srcjar` is a convention used within Bazel. It is hardcoded within the rules and it will break expectations in the standard libraries to rename them (e.g [com_google_protobuf](https://github.com/protocolbuffers/protobuf/blob/master/protobuf.bzl#L321), [rules_scala](https://github.com/bazelbuild/rules_scala/blob/master/scala_proto/private/scalapb_aspect.bzl#L39), [rules_java](https://docs.bazel.build/versions/master/be/java.html#java_library.srcs)). 

If we want to rename them, we'd have to duplicate all `.srcjar` files with `.jar` extensions locally to ensure compatibility with Bazel. That seems like an expensive and somewhat hack-y solution.

The `.srcjar` file type is the same as `.jar` when I check it using the `file` command. `tar` and `jar` also behave as expected. Bloop rejects the `.srcjar` extension, but I don't think Bloop checks the `resolution` field in the configuration files. If Metals calls tools that reject the `.srcjar` extension, then we'd prefer a different solution. But AFAICT only Metals does, so I think the change should be in Metals because it is the least disruptive solution.

A cleaner approach might be to rename the function `isValidJar(...)` and check for the `.jar` file extension or if `buildTools.isBazel` and the file has `.srcjar` extension. That would convey within Metals that it is a Bazel-specific convention.

I closed this PR because I realized it's a bit too early to start merging Bazel-specific needs into Metals. My main motivation for drafting this PR was to run the CI tests. I am not sure why I failed the MacOS check, but things are working locally so I'll keep prototyping Bazel-Metals integration with my branch. We can hold off on this for now and create an issue when we're more certain this is the solution we want. :)

PR #1455, opened 2020-02-20
add properties back in

ckipp01:
I went a little crazy with #1414 and since we aren't going to merge in https://github.com/scalameta/metals-vscode/pull/207 or https://github.com/scalameta/coc-metals/pull/74 until we do a release, these will need to stay in here. If not, when a user is using a snapshot of Metals, they won't have the necessary values set since they aren't being passed via experimental yet until those are merged. So, we need these for now.


tgodzik:
@ckipp01  Is this connected to a warning I am getting on the newest snapshots?
```
WARN  Unable to run doctor. To fix this problem enable -Dmetals.http=true
```
I plan to work on a release next week, seems we have a lot to release now.

ckipp01:
> @ckipp01 Is this connected to a warning I am getting on the newest snapshots?

Yea, this is exactly the reason @tgodzik

PR #1457, opened 2020-02-20
Fix issues discovered while testing out IntelliJ integration.

olafurpg:
This commit fixes some IntelliJ regressions that were caused by the
changes in https://github.com/scalameta/metals/pull/1452


olafurpg:
Ignoring unrelated failure on macOS
```
[error] Failed tests:
[error] 	tests.debug.StepDapSuite

```

PR #1463, opened 2020-02-24
add openFilesOnRenameProvider to experimental

ckipp01:
This setting was merged in right around when I was doing #1414, so I didn't include it. However, this can also go into `ClientExperimentalCapabilities`. It's also very VS Code specific, so we may be able to just fully remove it from the ServerConfig options.

To summarize, this pr moves the `metalsconfig.openFilesOnRenames` to an experimental provider. I'll also add this into the vscode extension.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -35,6 +35,7 @@ import scala.meta.internal.async.ConcurrentQueue
 import scala.meta.internal.semanticdb.Synthetic
 import scala.meta.internal.semanticdb.SelectTree
 import scala.meta.internal.metals.MetalsServerConfig
+import scala.meta.internal.metals.ClientExperimentalCapabilities
 
 final class RenameProvider(
     referenceProvider: ReferenceProvider,
@@ -46,7 +47,8 @@ final class RenameProvider(
     client: MetalsLanguageClient,
     buffers: Buffers,
     compilations: Compilations,
-    metalsConfig: MetalsServerConfig
+    metalsConfig: MetalsServerConfig,
+    clientExperimentalCapabilities: ClientExperimentalCapabilities
 ) {
 
   private var awaitingSave = new ConcurrentLinkedQueue[() => Unit]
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -145,11 +147,11 @@ final class RenameProvider(
       }
       val fileChanges = allChanges.flatten.toMap
       val shouldRenameInBackground =
-        !metalsConfig.openFilesOnRenames ||
+        (!metalsConfig.openFilesOnRenames && !clientExperimentalCapabilities.openFilesOnRenameProvider) ||
           fileChanges.keySet.size >= metalsConfig.renameFileThreshold
       val (openedEdits, closedEdits) =
         if (shouldRenameInBackground) {
-          if (metalsConfig.openFilesOnRenames) {
+          if (metalsConfig.openFilesOnRenames || clientExperimentalCapabilities.openFilesOnRenameProvider) {
```
tgodzik: Could we move it to a value? We can use that in both places then.

ckipp01: Sure! Fixed.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -144,12 +146,14 @@ final class RenameProvider(
         Seq(uri.toAbsolutePath -> textEdits.toList)
       }
       val fileChanges = allChanges.flatten.toMap
+      val openFilesOnRenameProvider =
+        clientExperimentalCapabilities.openFilesOnRenameProvider
```
tgodzik: We could even do:
```
val openFilesOnRenames = metalsConfig.openFilesOnRenames || clientExperimentalCapabilities.openFilesOnRenameProvider
```

tgodzik: Sorry for not being specific :(

ckipp01: Ah, duh. Of course. No worries, that was my bad, just quickly doing it and not thinking.

PR #1466, opened 2020-02-25
'StartDebugAdapter' command now accepts user's parameters (launch.json)

alekseiAlefirov:
Issue: https://github.com/scalameta/metals-vscode/issues/217

```json
"configurations": [
        {
          "type": "scala",
          "request": "launch",
          "name": "Foo",

          "mainClass": "bar.Foo",
          "buildTarget": "foo",
          "args": ["42"],
          "jvmOptions": ["..."]
      }
    ]
```
In this PR `debug-adapter-start` command is modified to accept more type of parameters like configuration above.
launch_debug

Client side PR: https://github.com/scalameta/metals-vscode/pull/223


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1302,6 +1302,28 @@ class MetalsLanguageServer(
...
         }).asJavaObject
+      case ServerCommands.ResolveDebugClass() =>
+        //TODO: polish this
+        val args = params.getArguments().asScala
+        val je = args.head.asInstanceOf[JsonElement]
+        val p = je.as[DebugClassParams].get
+        val (target, clazz) = (for {
+          //TODO: would be good to have error messages, maybe with use of Either
+          info <- buildTargets.findByDisplayName(
+            Option(p.project).getOrElse("root")
```
tgodzik: Should we have any default here?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1302,6 +1302,28 @@ class MetalsLanguageServer(
...
+            _.getClassName == p.mainClass
+          )
+        } yield (info.getId(), clazz)).get
+
+        val result = {
+          val dataKind = b.DebugSessionParamsDataKind.SCALA_MAIN_CLASS
+          val data = clazz.toJson
+          new b.DebugSessionParams(List(target).asJava, dataKind, data)
+        }
+        Future { result }.asJavaObject
```
tgodzik: Is it possible to just run debug here instead of returning params?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: Or maybe we can search through all build targets if project is not defined. In case there are two identical ones we can show a warning. `Running a.Main from 'A' project, but there is another main class with that name`

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
olafurpg: We may want to wrap the entire expression in a future to unblock future requests.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -211,6 +219,12 @@ object ServerCommands {
...
     StartDebugAdapter,
-    NewScalaFile
+    NewScalaFile,
+    ResolveDebugClass
   )
 }
+
+case class DebugClassParams(
+    mainClass: String,
+    @Nullable project: String = null
```
olafurpg: For better or worse we should probably use “build target” instead of “project” here

alekseiAlefirov: but will that be clear for users, what it is?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alekseiAlefirov: @tgodzik ,
yeah, I'm actually thinking about maybe removing this new command and just adding another possible `Params` class(es) for `StartDebugAdapter`.
So beside `DebugSessionParams`, it could take something like
```
DebugSessionUnresolvedMainClassParams {
	mainClass: string,
	buildTarget?: string,
	args?: string[],
	jvmOptions?: string[]
}

DebugSessionUnresolvedTestClassParams {
	testClass: string,
	buildTarget?: string
}
```

also, in future for https://github.com/scalameta/metals-feature-requests/issues/81 it could process `TextDocumentPositionParams` as well.
What do you think?

alekseiAlefirov:
`debug-resolve-params` command removed, while `debug-adapter-start` modified to accept more type of parameters to support user-set debug configurations.
Still **there are problems** with newly created tests, presumably because TestingServer do not fully reflect real Metals setup.
Header description of this PR edited.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugServer.scala
```diff
@@ -105,6 +112,155 @@ object DebugServer {
...
+        val dataKind =
+          b.DebugSessionParamsDataKind.SCALA_MAIN_CLASS
+        val data = clazz.toJson
+        new b.DebugSessionParams(
+          singletonList(target),
+          dataKind,
+          data
+        )
+    }
+  }
```
marek1840: Could you split the method somehow?

metals/src/main/scala/scala/meta/internal/metals/debug/DebugServer.scala
```diff
@@ -105,6 +112,155 @@ object DebugServer {
...
+        val dataKind =
+          b.DebugSessionParamsDataKind.SCALA_TEST_SUITES
+        val data = singletonList(clazz).toJson
+        new b.DebugSessionParams(
+          singletonList(target),
+          dataKind,
+          data
+        )
+    }
+  }
```
marek1840: Could you split the method somehow?

tests/unit/src/test/scala/tests/DebugProtocolSuite.scala
```diff
@@ -106,4 +110,66 @@ class DebugProtocolSuite extends BaseLspSuite("debug-protocol") {
...
+           |  }
+           |}
+           |/a/src/main/scala/a/Foo.scala
+           |package a
+           |class Foo extends org.scalatest.FunSuite {
+           |  test("foo") {}
+           |}
+           |""".stripMargin
+      )
+      _ = Thread.sleep(1000) //it fails otherwise
```
marek1840: Could you elaborate on the failure cause?

tests/unit/src/test/scala/tests/DebugProtocolSuite.scala
```diff
@@ -106,4 +110,66 @@ class DebugProtocolSuite extends BaseLspSuite("debug-protocol") {
...
+           |/a/src/main/scala/a/Main.scala
+           |package a
+           |object Main {
+           |  def main(args: Array[String]) = {
+           |    print(args(0))
+           |  }
+           |}
+           |""".stripMargin
+      )
+      _ = Thread.sleep(1000) //it fails otherwise
```
marek1840: Could you elaborate on the failure cause?

marek1840: Also, this is a place where possible flakiness may come from due to some underlying race conditions.

marek1840:
Also, increasing the timeout to `10000` fixed the failing tests on my machine. To remove the flakiness, you could wait until the workspace indexing finishes before proceeding to start the debug session - this would probably require adding some code around indexing to know when it finishes. The other solution could be to reuse the `TestingClient.refreshModelHandler` in a similar fashion it is used for CodeLenses, where we facing the same issue.

The underlying cause is that the BuildTargetClasses are fetched initially as a part of the indexing and `server.initialize` does not wait for the indexing to finish as there is no need to in most of the cases.

alekseiAlefirov:
Thank you for the review and the advice @marek1840 , will try that.
Yes, `BuildTargetClasses` don't get propagated with the actual data by the time, but I also have encountered following situations *sometimes* (it is really non-deterministic and hard to catch):
- `BuildTargets` do not get updated as well
- Seems like this indices get updated, but `DebugServer.start` just hangs - although it's really hard to reproduce, I think, it happens when I run metals-vscode debug (so, run debug inside debug) quite regularly. I am not sure, maybe I do something wrong, but just sharing this knowledge -- maybe you can also comment on this, as with `BuildTargetClasses`.

marek1840:
> BuildTargets do not get updated as well

`BuildTargets` also get filled during indexing, so the issue should disappear after waiting for indexing to finish (by any of the means described in the previous comment). If the issue still happens after that, we can focus on it then.

As for the server hanging - I am at a loss but can try to debug it if you share the exact sequence of steps (I can at least try, even if it would mean repeating them a few times to encounter this singularity ;) ). Also, a thread dump could help a great deal in finding the cause.

alekseiAlefirov:
> if you share the exact sequence of steps

basically, you run metals-vscode, run debug, and in newly opened code instance open a simple main class with one-line `main` method. Put a breakpoint onto this line, and run debug via codelens. For me it stucks somewhere never reaching the breakpoint and the pause button does not do anything. When I abort it, there's an error message saying `no debug adapter found`.

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -28,6 +28,32 @@ final class BuildTargetClasses(
...
+    index
+      .mapValues(
+        _.testClasses.values.find(_ == name)
+      )
+      .toList
+      .collect {
+        case (target, Some(clazz)) => clazz -> target
+      }
+  }
+
```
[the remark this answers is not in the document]
alekseiAlefirov: Done

metals/src/main/scala/scala/meta/internal/metals/debug/DebugServer.scala
alekseiAlefirov: I've moved finding class and build target data into new class `BuildTargetClassesFinder`.
`findMainClassAndItsBuildTarget` and `findTestClassAndItsBuildTarget` can also probably can be refactored, especially since they follow the same scenario, just main/test is different, but I'm a bit afraid, it eventually get complex-easy (instead of a simple-hard).

metals/src/main/scala/scala/meta/internal/metals/debug/DebugServer.scala
alekseiAlefirov: > I've moved finding class and build target data into new class BuildTargetClassesFinder.
findMainClassAndItsBuildTarget and findTestClassAndItsBuildTarget can also probably can be refactored, especially since they follow the same scenario, just main/test is different, but I'm a bit afraid, it eventually get complex-easy (instead of a simple-hard).

alekseiAlefirov:
@marek1840 I've made some reworks, and fixed flakiness of tests by including indexing build targets into `MetalsLanguageServer#indexWorkspace` and making use of `indexingPromise` in the tests.

alekseiAlefirov:
Don't see though, why it still fails on CI, the tests are being passed now on my machine :thinking:

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -28,6 +28,27 @@ final class BuildTargetClasses(
...
+    findClassesBy(_.mainClasses.values.find(_.getClassName() == name))
+
+  def findTestClassByName(
+      name: String
+  ): List[(String, b.BuildTargetIdentifier)] =
+    findClassesBy(_.testClasses.values.find(_ == name))
+
+  private def findClassesBy[A](
+      f: Classes => Option[A]
+  ): List[(A, b.BuildTargetIdentifier)] = {
```
tgodzik: Would `Option` not be enough at least here?

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClassesFinder.scala
```diff
@@ -0,0 +1,111 @@
...
+      buildTargets
+        .findByDisplayName(targetName)
+        .fold[Try[List[(String, b.BuildTarget)]]] {
+          Failure(
+            new BuildTargetNotFoundException(targetName)
+          )
+        } { target =>
+          buildTargetClasses
+            .classesOf(target.getId())
+            .testClasses
```
tgodzik: This looks like the only difference between the 2 functions is here and two lines below. Can w reuse the code and just add maybe a function `Classes => Option[T]`?

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -107,7 +108,26 @@ object ServerCommands {
...
         |```
-    """.stripMargin
+        |
+        |or DebugUnresolvedMainClassParams object
+        |Example:
+        |```json
+        |{
+        |   mainClass: "com.foo.App",
+        |   buildTarget: "foo",
+        |   args: ["bar"]
```
tgodzik: Are we able to also add environment variables? Also, let's add `jvmOptions` to the example.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProxy.scala
```diff
@@ -152,7 +152,7 @@ private[debug] object DebugProxy {
       endpoint: RemoteEndpoint,
       name: String
   ): RemoteEndpoint = {
-    val trace = GlobalTrace.setup(name)
+    val trace = GlobalTrace.setupTracePrinter(name)
```
tgodzik: This should fixed the issues with missing logs.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugServer.scala
```diff
@@ -133,6 +221,42 @@ object DebugServer {
...
   }
 
+  private def withRebuildRetry[A](
+      f: () => Try[A],
+      compilations: Compilations,
+      buildTargets: BuildTargets,
+      buildTargetClasses: BuildTargetClasses
+  )(implicit ec: ExecutionContext): Future[A] = {
+    Future.fromTry(f()).recoverWith {
+      case _: ClassNotFoundException | _: BuildTargetNotFoundException =>
```
tgodzik: `BuildTargetNotFoundException` will not be needed here. User would need reimport the build to change them.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugServer.scala
```diff
@@ -105,6 +114,85 @@ object DebugServer {
     }
   }
 
+  def resolveMainClassParams(
+      params: DebugUnresolvedMainClassParams,
+      buildTargetClassesFinder: BuildTargetClassesFinder,
```
tgodzik: Let's maybe move these new methods to `BuildTargetClassesFinder` with other than `params` method parameters added to the constructor? Also, with client instead of showWarningMessage.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugServer.scala
```diff
@@ -133,6 +221,42 @@ object DebugServer {
...
+      f: () => Try[A],
+      compilations: Compilations,
+      buildTargets: BuildTargets,
+      buildTargetClasses: BuildTargetClasses
+  )(implicit ec: ExecutionContext): Future[A] = {
+    Future.fromTry(f()).recoverWith {
+      case _: ClassNotFoundException | _: BuildTargetNotFoundException =>
+        val allTargets = buildTargets.all.toSeq.map(_.info.getId())
+        for {
+          _ <- compilations.compileTargets(allTargets)
```
tgodzik: I think even if we manage to find the build target we should still run compile on it.

So, do the same, but on single target first. If that fails, compile all targets.

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
alekseiAlefirov: actually, I've made it a List motivated by your idea to report multiple source targets https://github.com/scalameta/metals/pull/1466#discussion_r384661635

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
alekseiAlefirov: the input for this `Params` class is based on `ScalaMainClass` params, and it does not take environment variables.
About jvmOpts -- can you propose one?

tgodzik: @marek1840 Would make sense to add them to BSP? 
@alekseiAlefirov  Maybe `-Dfile.encoding=UTF-8`?

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
tgodzik: Ach, nvm. I just thought it's `toList` from Option. Nevermind then :sweat_smile:

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClassesFinder.scala
alekseiAlefirov: okay, I've added
```
private def findClassAndBuildTarget[A](
      className: String,
      buildTarget: Option[String],
      findClassesByName: String => List[(A, b.BuildTargetIdentifier)],
      classesByBuildTarget: b.BuildTargetIdentifier => Iterable[A],
      getClassName: A => String
  ): Try[List[(A, b.BuildTarget)]]
```
to generalize them

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
alekseiAlefirov: added

metals/src/main/scala/scala/meta/internal/metals/debug/DebugServer.scala
alekseiAlefirov: good one, fixed

metals/src/main/scala/scala/meta/internal/metals/debug/DebugServer.scala
alekseiAlefirov: I am not sure. This method takes some debug parameters to resolve them to other parameters and reports about running debug on one class, if multiple found. So it's more about debug, then about finding, which is extracted to `Finder`

metals/src/main/scala/scala/meta/internal/metals/debug/DebugServer.scala
alekseiAlefirov: sure, added narrowed recovery for `ClassNotFoundInBuildTargetException`

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1295,27 +1296,53 @@ class MetalsLanguageServer(
...
-                buildTargets,
-                buildServer
-              )
-            } yield {
-              cancelables.add(server)
-              DebugSession(server.sessionName, server.uri.toString)
-            }
-
-            session.asJavaObject
+        val debugSessionParamsParser = new JsonParser.Of[b.DebugSessionParams]
```
tgodzik: Let's maybe move those parsers to an object, no need to recreate them each time.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugServer.scala
tgodzik: Then maybe this should not actually be an object but maybe `DebugProvider`  class? Start also takes a lot of parameters, which are not specific to the actual function invocation.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alekseiAlefirov: created `DebugParametersJsonParsers` with these parsers

metals/src/main/scala/scala/meta/internal/metals/debug/DebugServer.scala
alekseiAlefirov: `DebugProvider` class created, all code from `DebugServer` object moved to it.

alekseiAlefirov:
@tgodzik agreed on that `DebugServer` object needed to be a class. `DebugProvider` class created.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -122,6 +123,8 @@ class MetalsLanguageServer(
   private val languageClient = new DelegatingLanguageClient(NoopLanguageClient)
   var userConfig: UserConfiguration = UserConfiguration()
   val buildTargets: BuildTargets = new BuildTargets()
+  private val buildTargetClassesFinder =
```
tgodzik, line 126: This is unused now.

alekseiAlefirov: thank you! fixed

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -0,0 +1,235 @@
...
+  def resolveTestClassParams(
+      params: DebugUnresolvedTestClassParams
+  )(implicit ec: ExecutionContext): Future[b.DebugSessionParams] = {
+    withRebuildRetry(() => {
+      buildTargetClassesFinder
+        .findTestClassAndItsBuildTarget(
+          params.testClass,
+          Option(params.buildTarget)
+        )
+    }).map {
```
tgodzik, line 137: This will fail on Nil, could you handle that case?

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -0,0 +1,235 @@
...
+      params: DebugUnresolvedMainClassParams
+  )(implicit ec: ExecutionContext): Future[b.DebugSessionParams] = {
+    withRebuildRetry(() =>
+      buildTargetClassesFinder
+        .findMainClassAndItsBuildTarget(
+          params.mainClass,
+          Option(params.buildTarget)
+        )
+    ).map {
+      case (clazz, target) :: others =>
```
tgodzik, line 107: This will also fail on Nil, could you handle that case?

alekseiAlefirov: as I commented on `BuildTargetClassFinder#findMainClassAndItsBuildTarget` (above method def), it either returns `Success` with non-empty list, or `Failure`. So it should not be the case. I wish we had a type-level solution for that.

kpbochenek: Solution for NonEmptyList? You have `scalaz.NonEmptyList`, `cats.data.NonEmptyList`, everything is already included in dependencies

alekseiAlefirov: hmm, last time I checked (several months ago), `cats` just got eliminated from Metals, and it was said it's better to have less dependencies. Okay, thanks, @kpbochenek !

alekseiAlefirov: okay, seems like there's no cats used in the main code, and @tgodzik recommended not to start.
^ fixed with `throw NoSuchElement(Exception)` for the `Nil` case

tgodzik:
@ckipp01 This might be helpful for `coc-metals` and `vim-inspector` -> you will able to run a simplified command without code lenses.

ckipp01:
Yes! I'll have to dive into this a bit more and then that'd be awesome. Because right now if it's reliant on code-lenses, then only nvim users will benefit.

tgodzik:
> Yes! I'll have to dive into this a bit more and then that'd be awesome. Because right now if it's reliant on code-lenses, then only nvim users will benefit.

Exactly, with this PR however we will no longer be reliant on code lenses.

PR #1467, opened 2020-02-25
Don't delete bloop.settings.json when exporting Pants builds.

olafurpg:
Previously, Metals deleted the `bloop.settings.json` file after
exporting a Pants build.


olafurpg:
Closing since I won't be able to look into the issues this week.

PR #1470, opened 2020-02-27
Export packagePrefix to bloop

lukaszwawrzyk:
See https://github.com/scalacenter/bloop/pull/1183


metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
```diff
@@ -404,12 +422,22 @@ private class BloopPants(
         Some(List(baseDirectory))
       }
 
+    val sourceRoots = {
+      val sourceDirs = sources ++ sourcesGlobs.toSeq.flatten.map(_.directory)
```
olafurpg: I believe the following is sufficient
```suggestion
      val sourceDirs = List(baseDirectory)
```

PR #1472, opened 2020-02-28
Fix issue when warning about Bloop version change 

tgodzik:
Previously. the warning about Bloop version change would pop up always when starting the workspace when the user had a non default setting. Now, the default is changed to None, which stops the consistent warning.


metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
```diff
@@ -21,9 +21,8 @@ case class GradleBuildTool(userConfig: () => UserConfiguration)
       ""
     }
   }
-  private val versionToUse = userConfig().bloopVersion
 
-  private val initScript =
+  private def initScript(versionToUse: String) =
```
tgodzik: Since we are using string interpolation here it will be safer to check the type.

Issue #1471, opened 2020-02-28
Javadoc not updated after change


kpbochenek:
**Describe the bug**
Function documentation displayed on mouse hover is not updated.
I can reproduce it in VS Code and Emacs that's why I raise it here, but it might be editor specific, must be checked better.

**To Reproduce**
Type in VS Code/Emacs this:
```scala
object Scaladoc {
  def fx(x: String): String = x

  fx("test")
}
```
You can see when hovering mouse over `fx("test")` it displays only signature.

Add scaladoc comment:
```scala
object Scaladoc {
  /**
    * Description of a fx
    *
    * @param x Super important param
    * @return the same
    */
  def fx(x: String): String = x

  fx("test")
}
```

Hover mouse over `fx("test")` and still only signature.

Reload window and hover again -> proper scaladoc is displayed.

Also when you already have doc displayed and change its content it doesn't refresh what is displayed after hover.

![Screenshot 2020-02-28 at 09 54 47](https://user-images.githubusercontent.com/10478402/75525642-7bba3980-5a10-11ea-877f-dd28f14619d2.png)

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code and Emacs (the same behaviour)
- Metals version: v0.8.1

**Search terms**
hover, doc

tgodzik:
I think I've seen the issue before, but pretty sure that we haven't fixed it.

I think Scaladoc are only indexed at the start and that's why?

tanishiking:
I also hit this behavior and took a quick look at this.

> I think Scaladoc are only indexed at the start and that's why?

As @tgodzik mentioned, the cause of this is a lack of cache expiration in `scala.meta.internal.metals.Docstrings`: once the symbol (and it's docstring) had cached to `Docstrings.cache`, it never expired.
 https://github.com/scalameta/metals/blob/60859ccb7dfa19a077cc71bf8604adde976676fd/mtags/src/main/scala/scala/meta/internal/metals/Docstrings.scala#L26-L37

I guess we can fix this "bug" by reindexing symbols to `Docstrings` on save as `WorkspaceSymbols` does. 🤔

PR #1473, opened 2020-03-01
Expire docstrings' cache on save

tanishiking:
Fixes https://github.com/scalameta/metals/issues/1471

### Problem
Since the cache in `scala.meta.internal.metals.Docstrings` never expired, once the docstring for the symbol has cached, it won't update until the metals server restart.
That's why even if we update Javadoc comments, the change won't reflect the docstrings showed on hover as https://github.com/scalameta/metals/issues/1471.

### How this PR fixed the problem
- Implemented `Docstrings#didChange(symbols: Seq[String]): Unit` that expires the cache for the docstrings of given (scalameta) symbols.
- Call `Docstrings#didChange` method on save (from `indexSourceFile`).

docstrings

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1579,6 +1579,7 @@ class MetalsLanguageServer(
           }
       }
       workspaceSymbols.didChange(source, symbols)
+      symbolDocs.didChange(symbols.map(_.symbol))
     } catch {
       case NonFatal(e) =>
         scribe.error(source.toString(), e)
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1579,7 +1579,10 @@ class MetalsLanguageServer(
           }
       }
       workspaceSymbols.didChange(source, symbols)
-      symbolDocs.didChange(symbols.map(_.symbol))
+
+      // Since the `symbols` here are toplevel symbols,
+      // we cannot use `symbols` for expiring the cache for all symbols in the source.
+      symbolDocs.expireSymbolDefinition(source)
     } catch {
       case NonFatal(e) =>
         scribe.error(source.toString(), e)
--- a/mtags/src/main/scala/scala/meta/internal/metals/Docstrings.scala
+++ b/mtags/src/main/scala/scala/meta/internal/metals/Docstrings.scala
@@ -12,6 +12,7 @@ import scala.meta.internal.semanticdb.Language
 import scala.meta.pc.SymbolDocumentation
 import scala.util.control.NonFatal
 import scala.meta.internal.mtags.GlobalSymbolIndex
+import scala.meta.internal.semanticdb.SymbolOccurrence
 
 /**
  * Implementation of the `documentation(symbol: String): Option[SymbolDocumentation]` method in `SymbolSearch`.
@@ -38,18 +39,25 @@ class Docstrings(index: GlobalSymbolIndex) {
   }
 
   def didChange(symbols: Seq[String]): Unit = {
-    symbols.foreach(sym => indexSymbol(sym))
+    symbols.foreach(sym => indexSymbol(sym, Some(expireCacheSymbol)))
   }
 
   private def cacheSymbol(doc: SymbolDocumentation): Unit = {
     cache(doc.symbol()) = doc
   }
 
-  private def indexSymbol(symbol: String): Unit = {
+  private def expireCacheSymbol(occ: SymbolOccurrence): Unit = {
+    cache.remove(occ.symbol)
+  }
+
+  private def indexSymbol(
+      symbol: String,
+      occFn: Option[SymbolOccurrence => Unit] = None
+  ): Unit = {
     index.definition(Symbol(symbol)) match {
       case Some(defn) =>
         try {
-          indexSymbolDefinition(defn)
+          indexSymbolDefinition(defn, occFn)
         } catch {
           case NonFatal(e) =>
             logger.log(Level.SEVERE, defn.path.toURI.toString, e)
@@ -58,14 +66,17 @@ class Docstrings(index: GlobalSymbolIndex) {
     }
   }
 
-  private def indexSymbolDefinition(defn: SymbolDefinition): Unit = {
+  private def indexSymbolDefinition(
+      defn: SymbolDefinition,
+      occFn: Option[SymbolOccurrence => Unit]
+  ): Unit = {
     defn.path.toLanguage match {
       case Language.JAVA =>
         JavadocIndexer
           .foreach(defn.path.toInput)(cacheSymbol)
       case Language.SCALA =>
         ScaladocIndexer
-          .foreach(defn.path.toInput)(cacheSymbol)
+          .foreach(defn.path.toInput)(cacheSymbol, occFn)
       case _ =>
     }
   }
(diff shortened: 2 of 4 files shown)


tanishiking:
Oops, what we should do on save is just expiring cache, not updating cache (to save the memory usage)
I'll change it tomorrow.

mtags/src/main/scala/scala/meta/internal/metals/ScaladocIndexer.scala
```diff
@@ -17,118 +17,135 @@ import scala.meta.pc.SymbolDocumentation
  */
 class ScaladocIndexer(
     input: Input.VirtualFile,
-    fn: SymbolDocumentation => Unit
+    fn: SymbolDocumentation => Unit,
+    occFn: Option[SymbolOccurrence => Unit]
```
tanishiking: If we specify `occFn` to `ScalafmtIndexer`, do not calculate docstrings because this function is usually used for expiring the symbol.

mtags/src/main/scala/scala/meta/internal/metals/Docstrings.scala
```diff
@@ -54,14 +66,17 @@ class Docstrings(index: GlobalSymbolIndex) {
...
 
-  private def indexSymbolDefinition(defn: SymbolDefinition): Unit = {
+  private def indexSymbolDefinition(
+      defn: SymbolDefinition,
+      occFn: Option[SymbolOccurrence => Unit]
+  ): Unit = {
     defn.path.toLanguage match {
       case Language.JAVA =>
         JavadocIndexer
           .foreach(defn.path.toInput)(cacheSymbol)
```
tanishiking: Though we won't expire the cache for Javadoc symbols, it might be better to pass `occFn` function to `JavadocIndexer` for consistency?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1579,6 +1579,7 @@ class MetalsLanguageServer(
           }
       }
       workspaceSymbols.didChange(source, symbols)
+      symbolDocs.didChange(symbols.map(_.symbol))
```
tgodzik: Could we invoke `indexSymbolDefinition` inside the foreach above? I think we have all the needed information and skip the additional steps.

mtags/src/main/scala/scala/meta/internal/metals/ScaladocIndexer.scala
tgodzik: Why do you need to expire the symbol? Wouldn't we just be good with updating it?

tgodzik: Och ok, we update the docstrings on demand, I didn't see it.

mtags/src/main/scala/scala/meta/internal/metals/Docstrings.scala
```diff
@@ -37,15 +38,26 @@ class Docstrings(index: GlobalSymbolIndex) {
     }
   }
 
+  def didChange(symbols: Seq[String]): Unit = {
+    symbols.foreach(sym => indexSymbol(sym, Some(expireCacheSymbol)))
```
tgodzik: Why don't we just do:
`symbols.foreach(cache.remove)` 
?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: Ok, I think did change is fine here, but added a comment in Docstrings

tanishiking:
@tgodzik Thank you so much for your feedback!

---

> Why do you need to expire the symbol? Wouldn't we just be good with updating it?

I believe it's better to expire here over updating in terms of memory usage because:

- The cache for a docstring is calculated only when it is referred via `definition`
  - https://github.com/scalameta/metals/blob/60859ccb7dfa19a077cc71bf8604adde976676fd/mtags/src/main/scala/scala/meta/internal/metals/Docstrings.scala#L31
  - therefore, the cache contains the only symbols which referred (by HoverProvider) before.
- If we update the cache for symbols every time we save a file, metals will cache all symbols in files we've saved, and metals consume quite a lot of memory, I guess.

---

> Why don't we just do:
> symbols.foreach(cache.remove)
> ?

The symbols calculated in [MetalsLanguageServver#indexSourceFile](https://github.com/scalameta/metals/blob/60859ccb7dfa19a077cc71bf8604adde976676fd/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1563) are all toplevel symbols:

`a/A.` (maybe `a/A#` ?) for

```scala
package a
object A {
  /**
   * test
   */
  def foo(x: Int): Int = x
}
```

(doesn't contain `a/A.foo().` nor `a/Def.foo().(x)`)

To invalidate the cache for  `a/A.foo().`, we have to collect symbol occurrences in a file by [ScaladocIndexer](https://github.com/scalameta/metals/blob/60859ccb7dfa19a077cc71bf8604adde976676fd/mtags/src/main/scala/scala/meta/internal/metals/Docstrings.scala#L62-L64) (and JavaIndexer).

**That's why we call `indexSymbol` to collect symbol occurrences in a file.**

- It might be better to call `Scaladoc/JavadocIndexer` from `didChange` directly because `didChange `doesn't index the symbols.
- And I'm going to write some scaladoc comment on `didChange` :)

mtags/src/main/scala/scala/meta/internal/metals/Docstrings.scala
```diff
@@ -54,14 +66,17 @@ class Docstrings(index: GlobalSymbolIndex) {
     }
   }
 
-  private def indexSymbolDefinition(defn: SymbolDefinition): Unit = {
+  private def indexSymbolDefinition(
```
tgodzik: Can we have an explicit method to expire and use: 
```scala
class Deindexer(
    input: Input.VirtualFile
) extends ScalaMtags(input) {
  override def visitOccurrence(
      occ: SymbolOccurrence,
      sinfo: SymbolInformation,
      owner: String
  ): Unit = {
    cache.remove(occ.symbol)
  }
}

  def expireSymbolDefinition(path : AbsolutePath): Unit = {
    path.toLanguage match {
      case Language.SCALA =>
        new Deindexer(path.toInput).indexRoot()
      case _ =>
    }
  }
```
`expireSymbolDefinition` can be explicitely run from `didChange`

This is more explicit and less confusing than passing the function.

tgodzik:
Maybe you could add part of the explanation you posted to `Docstrings`? This would be useful for the future in case someone else was confused like me :sweat_smile:

tanishiking:
Now, ready for review :)

tanishiking:
Oops, windows test is failing, I'll check it later...

mtags/src/main/scala/scala/meta/internal/metals/ScaladocIndexer.scala
```diff
@@ -125,6 +125,14 @@ class ScaladocIndexer(
 
 object ScaladocIndexer {
 
+  /**
+   * Extracts Scaladoc from Scala source code.
+   *
+   * @param fn callback function for calculated SymbolDocumentation
+   * @param occFn (optional) callback function for SymbolOccurrence
```
kpbochenek: I think after last changes you no longer have 'occFn'

mtags/src/main/scala/scala/meta/internal/metals/Docstrings.scala
kpbochenek: This should be actually `ScalaDeindexer` right?
Is it okay to ignore java files in this case?

mtags/src/main/scala/scala/meta/internal/metals/ScaladocIndexer.scala
tanishiking: Thanks! Fixed :)

Issue #1477, opened 2020-03-04
Multiple multi-line strings in a file throw off formatting


ckipp01:
**Describe the bug**
If you have multiple multi-line strings in a file, the `onTypeFormatting` is screwed up.

**Expected behavior**
```scala
  val test = """|
                | Here is my working multi-line
                |""".stripMargin

  val otherTest = """|
                     | Here is another one
                     |""".stripMargin
```
**Actual behavior**
```scala
  val test = """|
                | Here is my working multi-line
                |""".stripMargin

  val otherTest = """|
                | This one is wrong
                |""".stripMargin
```

**To Reproduce**
1. Create one multi-line string
2. Start creating another multi-line string and try to get onTypeFormatting working correctly
3. See error

**Screenshots**
![2020-03-04 14 34 35](https://user-images.githubusercontent.com/13974112/75884542-51f47e80-5e25-11ea-80ef-4e835b9da8ae.gif)

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code and Vim
- Metals version: v0.8.1

**Search terms**
multi-line strings, onTypeFormatting, formatting

PR #1478, opened 2020-03-04
remove -1 from determining if a new line is before pos

### LINKED ISSUE #1469: Newline at the beginning of the file throws an exception
**Describe the bug**
```
SEVERE: Internal error: java.lang.StringIndexOutOfBoundsException: String index out of range: -1
java.util.concurrent.CompletionException: java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:308)
	at java.util.concurrent.CompletableFuture.uniAccept(CompletableFuture.java:661)
	at java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:646)
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:488)
	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1990)
	at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1(CancelTokens.scala:38)
	at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1$adapted(CancelTokens.scala:36)
Caused by: java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.lang.String.charAt(String.java:658)
	at scala.collection.immutable.StringOps$.apply$extension(StringOps.scala:41)
	at scala.meta.internal.metals.MetalsEnrichments$XtensionString.lastIndexBetween(MetalsEnrichments.scala:411)
	at scala.meta.internal.metals.MultilineStringFormattingProvider.pipeInScope(MultilineStringFormattingProvider.scala:86)
	at scala.meta.internal.metals.MultilineStringFormattingProvider.$anonfun$withToken$1(MultilineStringFormattingProvider.scala:131)
	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:659)
	at scala.util.Success.$anonfun$map$1(Try.scala:255)
	at scala.util.Success.map(Try.scala:213)
	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
	at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
	at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
	... 4 more
```
**To Reproduce**
1. Go to any file
2. Move cursor to the start of the file
3. Press enter

**Expected behavior**
Nothing happens.

**Installation:**
- Metals version: v0.8.1

**Search terms**
on type formatting exception

ckipp01:
This pr removes logic that was creating an exception from being thrown when a new line was being added as the first thing in a file.

Closes #1469

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
@@ -78,7 +78,7 @@ final class MultilineStringFormattingProvider(
       newlineAdded: Boolean
   ): Boolean = {
     val newLineBeforePos =
-      text.lastIndexBetween('\n', upperBound = pos.start - 1)
+      text.lastIndexBetween('\n', upperBound = pos.start)
     val pipeSearchStop =
       if (newlineAdded)
         text.lastIndexBetween('\n', upperBound = newLineBeforePos - 1)


metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -78,7 +78,7 @@ final class MultilineStringFormattingProvider(
       newlineAdded: Boolean
   ): Boolean = {
     val newLineBeforePos =
-      text.lastIndexBetween('\n', upperBound = pos.start - 1)
```
ckipp01: If I understand this correctly, this isn't needed here since we can just calculate starting at the position start rather then the position start -1 to determine the last index between the new line and the position start. This is what was throwing the exception when we were trying to get the index using the value here, which was -1 when in the beginning of the file.

ckipp01:
Not working as expected. This actually introduces another issue.

metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
tgodzik: We probably just need `Math.max(pos.start - 1, 0)`

ckipp01: I'm working on a fix for #1477, so I'll include the fix for both in another pr.

kpbochenek: When pressing enter we search for `newLineBeforePos`(where our newline was added). Then we search for '\n' before that and assign it to `pipeSearchStop` and it gives us a range [pipeSearchStop, newLineBeforePos] which is our previous line. Then we check if it contains '|' and if that is the case we do formatting(add '|' in current line properly aligned).

In current code it fails because it starts to search `pipeSearchStop` from position -1, I think we should simply add `Math.max(0, xxx - 1)` in both cases where we do search for \n

kpbochenek: @tgodzik you were faster with a comment, ignore mine :P

PR #1481, opened 2020-03-05
Add ability to specify zipkin url for fastpass

wiwa:
We currently can only submit traces to Bloop's default of `http://127.0.0.1:9411/api/v2/spans`, unless we manually write to `~/.bloop/.jvmopts`. This change will allow `BloopPants` to write to `~/.bloop/.jvmopts` automatically. Binary versions of `fastpass` can be built with the correct server url statically embedded.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/pantsbuild/commands/ZipkinUrls.scala
+++ b/metals/src/main/scala/scala/meta/internal/pantsbuild/commands/ZipkinUrls.scala
@@ -1,0 +1,44 @@
+package scala.meta.internal.pantsbuild.commands
+
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.StandardOpenOption
+import scala.meta.io.AbsolutePath
+
+object ZipkinUrls {
+
+  def url: Option[String] =
+    Option(System.getProperty("metals.zipkin.server.url"))
+
+  def updateZipkinServerUrl(): Boolean = {
+    ZipkinUrls.url match {
+      case Some(newUrl) =>
+        import scala.meta.internal.metals.MetalsEnrichments._
+        val homedir = AbsolutePath(System.getProperty("user.home"))
+        val jvmopts = homedir.resolve(".bloop").resolve(".jvmopts")
+        val oldOptions =
+          if (jvmopts.isFile) jvmopts.readText.linesIterator.toList
+          else Nil
+        val zipkin = "-Dzipkin.server.url=(.*)".r
+        val oldUrl = oldOptions.collectFirst { case zipkin(url) => url }
+        if (oldUrl.contains(newUrl)) {
+          false
+        } else {
+          val otherOptions =
+            oldOptions.filterNot(_.startsWith("-Dzipkin.server.url")).toList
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/pantsbuild/commands/ZipkinUrls.scala
```diff
@@ -0,0 +1,52 @@
...
+          val idx = lines.indexWhere(_.startsWith("-Dzipkin.server.url="))
+          if (idx >= 0) {
+            val oldUrl = lines(idx).split('=')(1)
+            // Case 1: Update an existing URL
+            if (oldUrl != newUrl) {
+              lines(idx) = valueAssignment
+              lines
+            } else {
+              // Case 2: Old and new URL are the same; break out
+              return false
```
olafurpg: nit: let's avoid nested returns since they can be difficult to reason about (for example when inside a `map` lambda).

We occasionally use early `return` in the Metals codebase right at the start of a method implementation to avoid nested indentation.

metals/src/main/scala/scala/meta/internal/pantsbuild/commands/ZipkinUrls.scala
```diff
@@ -0,0 +1,52 @@
...
+
+object ZipkinUrls {
+
+  def url: Option[String] =
+    Option(System.getProperty("metals.zipkin.server.url"))
+
+  def updateZipkinServerUrl(): Boolean = {
+    ZipkinUrls.url match {
+      case Some(newUrl) =>
+        val homedir = Paths.get(System.getProperty("user.home"))
```
olafurpg:
```suggestion
        import MetalsEnrichments._
        val homedir = AbsolutePath(System.getProperty("user.home"))
        val jvmopts = homedir.resolve(".bloop").resolve(".jvmopts")
        val oldOptions = 
          if (jvmopts.isFile) jvmopts.readText.linesIterator.toList
          else Nil
        val zipkin = "-Dzipkin.server.url=(.*)".r
        val oldUrl = oldOptions.collectFirst { case zipkin(url) => url }
        if (oldUrl.contains(newUrl)) {
          false
        } else {
          val otherOptions = oldOptions.filterNot(_.startWith("-D.zipkin.server.url")).toList
          val zipkinOption = s"-Dzipkin.server.url=$newUrl"
          val allOptions = zipkinOption :: otherOptions
          Files.write(...)
```

metals/src/main/scala/scala/meta/internal/pantsbuild/commands/ZipkinUrls.scala
```diff
@@ -0,0 +1,52 @@
...
+            }
+          } else {
+            // Case 3: .jvmopts exists but no old URL
+            lines :+ valueAssignment
+          }
+        } else {
+          // Case 4: .jvmopts does not exist
+          Array(valueAssignment)
+        }
+        scribe.info(s"Updating Zipkin server URL to $newUrl")
```
olafurpg: We generally try to format logs like this
```suggestion
        scribe.info(s"zipkin: new server URL '$newUrl'")
```

PR #1486, opened 2020-03-10
Skip coursier download in fastpass create with --coursier-binary option

wiwa:
Users will now be able to optionally specify a path to their `coursier` binary to skip downloading it.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/pantsbuild/commands/SharedCommand.scala
+++ b/metals/src/main/scala/scala/meta/internal/pantsbuild/commands/SharedCommand.scala
@@ -47,7 +47,7 @@ object SharedCommand {
           }
           1
         case Success(count) =>
-          IntelliJ.writeBsp(export.project)
+          IntelliJ.writeBsp(export.project, export.export.coursierBinary)
           val targets = LogMessages.pluralName("Pants target", count)
           export.app.info(
             s"exported ${targets} to project '${export.project.name}' in $timer"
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/pantsbuild/commands/CreateCommand.scala
```diff
@@ -39,19 +39,35 @@ object CreateCommand extends Command[CreateOptions]("create") {
...
-        1
-      case None =>
-        val project = Project.create(name, create.common, create.targets)
-        SharedCommand.interpretExport(
-          Export(project, create.open, app).copy(export = create.export)
-        )
+    val coursierBinary = create.coursierBinary.map(AbsolutePath(_))
+    val pathError = coursierBinary
+      .map { path =>
+        if (!path.isFile) {
```
olafurpg: We can skip this validation. It's OK if the path does not exist. We only need this flag to allow testing fastpass in a CI environment that doesn't have access to internet and the bsp JSON files are only used by IntelliJ (not CI).

metals/src/main/scala/scala/meta/internal/pantsbuild/commands/SharedCommand.scala
```diff
@@ -21,9 +21,13 @@ import java.nio.charset.StandardCharsets
 import bloop.bloopgun.core.Shell
 import scala.concurrent.Promise
 import scala.meta.internal.metals.BuildInfo
+import java.nio.file.Path
 
 object SharedCommand {
-  def interpretExport(export: Export): Int = {
+  def interpretExport(
+      export: Export,
+      coursierBinary: Option[Path] = None
```
olafurpg: If we move `coursierBinary` into `ExportOptions` then we can avoid adding this parameter here and then the flag becomes available to all commands that trigger an export: create/refresh/amend

PR #1487, opened 2020-03-10
Goto super method in hierarchy of inheritance

### LINKED ISSUE #1416: Add an ability to go to super class/method
This is well know feature that might be useful for a lot of users. 

We already look recursively to find the top class in the ImplementationProvider, we could reuse to just go one level up. 

This could be implemented using code lenses although still not sure how it should look exactly. 

If anyone is interested in contributing this feature please let me know. 

Search by hand

**Additional context**
Suggested by @olafurpg 

**Search terms**:
Implementation super method class

kpbochenek:
Fixes #1416

## Goto super method + lenses
Preview of how it looks like in VSCode:
Screenshot 2020-03-10 at 09 24 41

During requesting lenses metals will also calculate if any definition of a method in requested file overrides another method, if yes it will attach lens to this position to be able to go to super method. Lenses also link to methods in external dependencies and are displayed in files from libraries.

They are configurable from `UserConfiguration`, example from VSCode:

Screenshot 2020-03-09 at 16 24 21

If they are found to be disturbing your workflow they can be easily disabled(they are enabled by default). Goto Super Method lenses should be quick to calculate but if you notice significant slowdown you can also disable them(when disabled they are not only not displayed but also not calculated at all)

There is also exposed server command: `goto-super-method` that can be mapped and used through shortcut. As example from VSCode PR (in client it is `metals.go-to-super-method`):
Screenshot 2020-03-10 at 09 38 49

## Goto Super Method in hierarchy

Preview how it looks like in VSCode:
super-hierarchy

Metals server exposes command `super-method-hierarchy` that can be invoked on a method definition. It calculated super method hierarchy and if more than one position is found it displays quickPick window (same as from new scala file) with super methods and when user selects one entry it jumps to that definition. 

As example from VSCode it can be bind to a shortcut this way:
Screenshot 2020-03-10 at 09 48 55

## Internal

At core of those changes is implementation of a function to perform linearization of any provided class. It is used in this PR to find first next super method or list super methods according to linearization rules but it can also be used in a future for another features.

Other changes:
- a few typos fixed
- tests for both functionalities
- related changes in VSCode-metals (https://github.com/scalameta/metals-vscode/pull/233)


metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+    }
+  }
+
+  def getSuperMethodHierarchy(
+      methodSymbolInformation: SymbolInformation,
+      documentWithPath: TextDocumentWithPath,
+      symbolRole: SymbolOccurrence.Role,
+      findSymbol: String => Option[SymbolInformation]
+  ): Option[List[SymbolInformation]] = {
+    if (symbolRole.isDefinition && (methodSymbolInformation.isMethod || methodSymbolInformation.isField)) {
```
tgodzik: Extract to a method?

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+  private def matchMethodInClass(
+      superClass: SymbolWithAsSeenFrom,
+      baseMethodSymbolInformation: SymbolInformation,
+      baseMethodInfo: MethodSignature,
+      findSymbol: String => Option[SymbolInformation]
+  ): Option[SymbolInformation] = {
+    superClass.symbolInformation.signature match {
+      case classSig: ClassSignature =>
+        classSig.getDeclarations.symlinks
+          .map(methodSlink =>
```
tgodzik: `methodSlink` -> `methodSymbolLink` ? Let's use full names. And also below.

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+  ): List[SymbolWithAsSeenFrom] = {
+    if (skip.exists(_.symbol == symbolInformation.symbol)) {
+      List.empty
+    } else {
+      skip += symbolInformation
+      symbolInformation.signature match {
+        case classSignature: ClassSignature =>
+          val parents = classSignature.parents
+            .collect { case x: TypeRef => x }
+            .filterNot(typeRef => stopSymbols.contains(typeRef.symbol))
```
tgodzik: Join the two filers?

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+      baseClassSignature = baseClassSymbolInformation.signature
+        .asInstanceOf[ClassSignature]
+      SymbolWithAsSeenFrom(superClass, asSeenFrom) <- calculateClassSuperHierarchyWithCache(
+        baseClassSymbolInformation,
+        cache,
+        findSymbol
+      )
+      if superClass.signature.isInstanceOf[ClassSignature]
+      classSig = superClass.signature.asInstanceOf[ClassSignature]
+      methodSlink <- classSig.getDeclarations.symlinks
```
tgodzik: methodSlink -> methodSymbolLink

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+                  classSignature.typeParameters
+                )
+                val currentASF =
+                  AsSeenFrom.translateAsSeenFrom(asSeenFrom, parentASF)
+                getSuperClasses(si, findSymbol, skip, currentASF)
+            }
+            .toList
+            .reverse
+            .flatten
+          val outASF =
```
tgodzik: Do we need to do it here? `toRealNames` should really be translated at the point we are comparing two methods from two specific classes.

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+        findSymbol
+      )
+      if superClass.signature.isInstanceOf[ClassSignature]
+      classSig = superClass.signature.asInstanceOf[ClassSignature]
+      methodSlink <- classSig.getDeclarations.symlinks
+      mSymbolInformation <- findSymbol(methodSlink).toIterable
+      if mSymbolInformation.isMethod
+      methodSignature = mSymbolInformation.signature
+        .asInstanceOf[MethodSignature]
+      newASF = AsSeenFrom.toRealNames(
```
tgodzik: Either this is no needed or the one inside `calculateClassSuperHierarchy`

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
```diff
@@ -61,7 +89,16 @@ final class DebugCodeLensProvider(
...
+            val docWithPath = TextDocumentWithPath(textDocument, path)
+            val gotoSuperMethod =
+              createSuperMethodCommand(
+                docWithPath,
+                symbol,
+                occurrence.role,
+                emptyLensGoSuperCache(),
+                search
+              )
+            main ++ tests ++ gotoSuperMethod
```
tgodzik: I was thinking of having it calculated with a method like `SemanticDBDoc => CodeLenses`, which could be in a separate trait. Both implementations would in a list and we could join their results.

tests/unit/src/test/scala/tests/CodeLensesLspSuite.scala
```diff
@@ -186,7 +188,61 @@ class CodeLensesLspSuite extends BaseLspSuite("codeLenses") {
     } yield ()
   }
 
-  def check(name: String, library: String = "")(
+  check("go-to-super-method-lenses")(
```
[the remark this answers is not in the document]
kpbochenek: can you provide example what you mean?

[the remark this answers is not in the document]
tgodzik: Something like
```
trait A{
  def abc : String
  def xyz: Int
}

class B extends A{
  def abc : String = ""
  def xyz: Int = 0
}

```

tests/unit/src/test/scala/tests/SuperHierarchyLspSuite.scala
```diff
@@ -0,0 +1,103 @@
...
+      _ <- server.didOpen("a/src/main/scala/a/A.scala")
+      _ = assertNoDiagnostics()
+    } yield {
+      val path = server.toPath("a/src/main/scala/a/A.scala").toURI.toString
+      val context = parse(code)
+      server.assertSuperMethodHierarchy(path, context(pos), expected)
+    }
+  }
+
+  private def parse(
```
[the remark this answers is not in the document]
kpbochenek: then for each marker I would have to copy whole code and convert to position which also iterates whole code again.
Are you okay with this performance penalty?

tests/unit/src/test/scala/tests/SuperMethodLspSuite.scala
```diff
@@ -0,0 +1,329 @@
...
+      code: String,
+      uri: String
+  ): (Map[Int, (Position, String)], Map[Int, Option[Int]]) = {
+    val (mapping, asserts) = parse(code)
+    (mapping.mapValues((_, uri)), asserts)
+  }
+
+  private def parse(
+      code: String
+  ): (Map[Int, Position], Map[Int, Option[Int]]) = {
```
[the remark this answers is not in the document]
kpbochenek: I can understand searching for <<1>> <<2>> until not found, but searching for every <<X->Y>> ?
I might not get idea, can you provide more defails how it should work?

tests/unit/src/test/scala/tests/SuperHierarchyLspSuite.scala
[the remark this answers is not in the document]
tgodzik: I don't believe the performance would be much worse in this case, I prefer readability in the tests here. No need to have a super optimized code for test.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -195,6 +197,33 @@ final class TestingServer(
...
   }
 
+  def assertGotoSuperMethod(
+      pos: Int,
+      maybeSuperPos: Option[Int],
+      context: Map[Int, (l.Position, String)]
+  )(implicit loc: munit.Location): Unit = {
+    val command = GoToSuperMethodParams(context(pos)._2, context(pos)._1)
+    val maybeFoundLocation =
+      server.goToSuperMethod.getGoToSuperMethodLocation(command)
```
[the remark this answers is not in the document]
kpbochenek: I can do checking for super method, but for hierarchy where I invoke QuickPick it is much harder.
I will take a look if it is something doable

tests/unit/src/test/scala/tests/SuperMethodLspSuite.scala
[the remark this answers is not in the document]
tgodzik: Ok, I was thinking we could make it simpler, but after thinking it might be more complex to rework.

tests/unit/src/test/scala/tests/SuperHierarchyLspSuite.scala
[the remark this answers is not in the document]
tgodzik: Ok nvm me, I think it might not be much easier to implement, let's leave it.

metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodLensesProvider.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  override def codeLenses(
+      textDocumentWithPath: TextDocumentWithPath
+  ): Seq[l.CodeLens] = {
+    val textDocument = textDocumentWithPath.textDocument
+    val path = textDocumentWithPath.filePath
+
+    val search =
+      makeSymbolSearchMethod(
+        makeGlobalClassTable(textDocumentWithPath.filePath),
```
tgodzik: This should be lazy if possible

tests/unit/src/test/scala/tests/CodeLensesLspSuite.scala
```diff
@@ -188,7 +190,63 @@ class CodeLensesLspSuite extends BaseLspSuite("codeLenses") {
...
+      |
+      |trait Cersei extends Tywin {
+      |<< payTheirDebts>>
+      |override def payTheirDebts = false
+      |<< trueLannister>>
+      |override def trueLannister = false
+      |}
+      |
+      |class Joffrey extends Lannister with Jamie with Cersei {
+      |<< payTheirDebts>>
```
tgodzik: Is it possible to fix the formatting here?

metals/src/main/scala/scala/meta/internal/metals/codelenses/CodeLenses.scala
```diff
@@ -0,0 +1,9 @@
+package scala.meta.internal.metals.codelenses
+
+import org.eclipse.{lsp4j => l}
+import scala.meta.internal.implementation.TextDocumentWithPath
+
+trait CodeLenses {
```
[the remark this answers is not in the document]
olafurpg: `CodeLenseProvider`?

olafurpg:
This is super impressive work @kpbochenek I agree the configuration option to enable/disable the code lenses is merited in this case, even if we try to avoid user-facing configuration options. I'm really excited to try this out 😄

ckipp01:
Is it possible to completely separate this from being dependent on having `debuggingProvider` set to true? I was trying to get this to work with `coc-metals` which currently doesn't have `debuggingProvider` set to true, which in this changes causing the `findLens` to result in nothing. I couldn't figure out why this was happening for an embarrassingly long time before it clicked and I turned the value to true, which immoderately made the code lenses appear. Since various clients might not provide debugger support, I think it's important to be able to offer this feature regardless of that setting.

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
```diff
@@ -17,134 +12,36 @@ trait CodeLensProvider {
...
   def apply(
-      buildTargetClasses: BuildTargetClasses,
-      buffers: Buffers,
-      buildTargets: BuildTargets,
-      compilations: Compilations,
+      codeLensProviders: List[CodeLenses],
       semanticdbs: Semanticdbs,
       capabilities: ClientExperimentalCapabilities
   )(implicit ec: ExecutionContext): CodeLensProvider = {
     if (!capabilities.debuggingProvider) Empty
```
ckipp01: Regarding my comment, I think this is where a change would need to happen. We aren't providing any `codeLensProviders` if the `debuggingProvider` is false, but really the superMethod code lens shouldn't be dependent on this.

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -0,0 +1,276 @@
...
+      findClassInfo(
+        msi.symbol,
+        msi.symbol.owner,
+        documentWithPath.textDocument,
+        findSymbol
+      )
+    val methodInfo = msi.signature.asInstanceOf[MethodSignature]
+
+    val result = for {
+      si <- classSymbolInformation.toIterable
```
ckipp01: Can this just be `symbolInformation` rather than `si`

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -0,0 +1,276 @@
...
+      findSymbol: String => Option[SymbolInformation]
+  ): Option[List[SymbolInformation]] = {
+    val classSymbolInformationMaybe =
+      findClassInfo(
+        msi.symbol,
+        msi.symbol.owner,
+        documentWithPath.textDocument,
+        findSymbol
+      )
+    val methodInfo = msi.signature.asInstanceOf[MethodSignature]
```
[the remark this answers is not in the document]
kpbochenek: If it is a field it is still MethodSignature it turns out, lenses are generated for fields too :)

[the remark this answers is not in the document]
kpbochenek: I moved this casting to method calling this function where I do check `isDefinitionOfMethodField`

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -0,0 +1,276 @@
...
+
+    classSymbolInformationMaybe.map(classSymbolInformation => {
+      calculateClassSuperHierarchy(classSymbolInformation, findSymbol)
+        .flatMap(matchMethodInClass(_, msi, methodInfo, findSymbol))
+    })
+  }
+
+  private def matchMethodInClass(
+      superClass: SymbolWithAsSeenFrom,
+      baseMethodSymbolInformation: SymbolInformation,
```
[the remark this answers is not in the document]
kpbochenek: it is but it has `Signature` tpe, and I need `MethodSignature`. I could do casting here but I prefer to keep casting where check is performed and simply provide it here with correct tpe (but yes it is duplicated).
If `SymbolInformation` was generic with `SymbolInformation[X <: Signature]` I could say here I need `baseMethodSymbolInformation: SymbolInformation[MethodSignature]` but I can't with current types.

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
tgodzik: That should be checked inside the RunTest code lense provider for sure. Thanks for finding it @ckipp01 !

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -0,0 +1,276 @@
...
+      findSymbol: String => Option[SymbolInformation]
+  ): Option[SymbolInformation] = {
+    superClass.symbolInformation.signature match {
+      case classSig: ClassSignature =>
+        classSig.getDeclarations.symlinks
+          .map(methodSymbolLink =>
+            for {
+              mSymbolInformation <- findSymbol(methodSymbolLink)
+              if mSymbolInformation.isMethod
+              methodSignature = mSymbolInformation.signature
```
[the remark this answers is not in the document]
kpbochenek: that is a nice trick! changed :+1:

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
kpbochenek: I changed `classSymbolInformation` -> `classSymbolInformationOption`
and `si` -> `classSymbolInformation`
what you think?

metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodLensesProvider.scala
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  private def createSuperMethodCommand(
+      docWithPath: TextDocumentWithPath,
+      symbol: String,
+      role: SymbolOccurrence.Role,
+      cache: LensGoSuperCache,
+      findSymbol: String => Option[SymbolInformation]
+  ): Option[l.Command] = {
+    if (userConfig().superMethodLensesEnabled) {
```
kpbochenek: removed this as `enabled` is checked in other place, this is old check which I forgot to remove

tests/unit/src/test/scala/tests/CodeLensesLspSuite.scala
kpbochenek: As answered previously, I pasted in previous answer which regex needs to be modified and what rules it must confront to. I could not quickly make it work and and don't want to spend hours working on regex(or parse this manually) that could only improve formatting in tests a little.
If you have can give me regex that will work I will add it and update tests.

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
kpbochenek: ahh okay now I get what this `debuggingProvider` was for!

I moved it to where I create a list of lens providers. 
If debugging is off I only create `superMethodProvider` but if debugging is on I create both.

```scala
    val runTestLensesProvider: List[CodeLenses] = if (clientExperimentalCapabilities.debuggingProvider) {
      List(new RunTestLensesProvider(buildTargetClasses, buffers, buildTargets))
    } else {
      List.empty
    }
```

metals/src/main/scala/scala/meta/internal/metals/codelenses/CodeLenses.scala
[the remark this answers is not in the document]
kpbochenek: Yes we have unfortunate clash of names here.
There is already a class `CodeLensProvider` that acts as aggregator for single lenses generators and is similar to other providers we have in metals.
Then there is a hierarchy of classes that provide specific lenses:
CodeLenses(or CodeLens as Tomek suggests):
- RunTestLensesProvider <: CodeLenses
- SuperMethodLensesProvider <: CodeLenses
... (others in future maybe?)

So maybe `CodeLensProvider` will be this `provider` class and here I will do hierarchy:
CodeLens
RunTestCodeLens <: CodeLens
SuperMethodCodeLens <: CodeLens

What do you think?

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
kpbochenek: Pushed a change, could you double check it is okay?
https://github.com/scalameta/metals/pull/1487/files#diff-f8c05eebbf12c9c21a7d568f09b500eaR395-R402

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -130,6 +130,43 @@ object ServerCommands {
...
+       |  document: "file:///home/dev/foo/Bar.scala",
+       |  position: {line: 5, character: 12}
+       |}
+       |```
+       |""".stripMargin
+  )
+
+  val SuperMethodHierarchy = new Command(
+    "super-method-hierarchy",
+    "Go to super method/field definition",
```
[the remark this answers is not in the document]
kpbochenek: added :+1:

kpbochenek:
Updated and commented on all requests.

There should probably still be some documentation added for users but I would like to firstly get this one merged, then give it a moment to settle and verify usability is okay, then update docs in another PR.

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -0,0 +1,280 @@
...
+      methodSymbolInformation: SymbolInformation,
+      documentWithPath: TextDocumentWithPath,
+      symbolRole: SymbolOccurrence.Role,
+      findSymbol: String => Option[SymbolInformation],
+      cache: LensGoSuperCache
+  ): Option[String] = {
+    if (isDefinitionOfMethodField(symbolRole, methodSymbolInformation)) {
+      findSuperForMethodOrFieldChecked(
+        methodSymbolInformation,
+        methodSymbolInformation.signature.asInstanceOf[MethodSignature],
```
olafurpg: Can we remove all usage of `asInstanceOf`? The idiomatic way to do this here would be to use pattern matching
```scala
info.signature match {
  case m: MethodSignature =>
    ...
    Some(...)
  case _ =>
    None
```

metals/src/main/scala/scala/meta/internal/metals/CodeLensProvider.scala
ckipp01: Woo hoo! Works now 👍

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
ckipp01: yea, I think that'd be clear.

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
kpbochenek: Okay, moved this to `isDefinitionOfMethodField` which returns `Option` and refactored as `for` comprehension. Same with similar situation below :+1:

kpbochenek:
@ckipp01 do you think you could try test command `"metals.super-method-hierarchy"` how it behaves/looks in vim?
I tried it for VSCode and what I needed to do was to add a command that can be bound to a shortcut which will call metals with:
name of command: "metals.super-method-hierarchy"
arguments: 
```
            {
              document: <current document uri>,  (string)
              position: <curson position>  (org.eclipse.lsp4j.Position(line: Int, character: Int)
            }
```
If you are on overridden method it should show you popup(the same as for NewScalaFile) with list of super methods you want to go (it is good to test if there is C.x <: B.x <: A.x, call it on C.x and see a list: [B.x, A.x])

I assumed you might be familiar with vim <-> metals (coc-vim?) stuff but if I am wrong do you know who I could talk to about this?

metals/src/main/scala/scala/meta/internal/metals/codelenses/CodeLenses.scala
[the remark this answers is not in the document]
kpbochenek: also do you prefer to just reply to comment, leave it unresolved and you get notification or should I tag you always it requires your feedback? @tgodzik @olafurpg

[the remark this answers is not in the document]
tgodzik: I think leave it unresolved, next time I take a look at the PR I will see what needs to be addressed.

ckipp01:
> @ckipp01 do you think you could try test command `"metals.super-method-hierarchy"` how it behaves/looks in vim?
> I tried it for VSCode and what I needed to do was to add a command that can be bound to a shortcut which will call metals with:
> name of command: "metals.super-method-hierarchy"
> arguments:
> 
> ```
>             {
>               document: <current document uri>,  (string)
>               position: <curson position>  (org.eclipse.lsp4j.Position(line: Int, character: Int)
>             }
> ```
> 
> If you are on overridden method it should show you popup(the same as for NewScalaFile) with list of super methods you want to go (it is good to test if there is C.x <: B.x <: A.x, call it on C.x and see a list: [B.x, A.x])
> 
> I assumed you might be familiar with vim <-> metals (coc-vim?) stuff but if I am wrong do you know who I could talk to about this?

Sure. The implementation ends up being a tad different since the coc.nvim api doesn't have `registerTextEditorCommand`, but just `registerCommand`. However, the situation you outlined looks like this:



I'll probably also just make a shortcut to easily be able to trigger it instead of using the command pallete.

kpbochenek:
@ckipp01 thanks for checking :+1: if it works with 2 editors it should work with all  :D

tests/unit/src/test/scala/tests/CodeLensesLspSuite.scala
tgodzik: Ok, let's leave it for now. Sorry for missing the previous answer!

olafurpg:
@tgodzik I think it's OK to merge now and include it in the release notes. It's a very impressive feature!

metals/src/main/scala/scala/meta/internal/metals/codelenses/CodeLenses.scala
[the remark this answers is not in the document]
kpbochenek: renamed as mentioned in previous comment

metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodLensesProvider.scala
kpbochenek: changed as discussed

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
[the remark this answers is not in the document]
kpbochenek: changed to pattern match :+1:

metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodCodeLens.scala
```diff
@@ -0,0 +1,100 @@
...
+    val path = textDocumentWithPath.filePath
+
+    val search = implementationProvider.defaultSymbolSearch(
+      textDocumentWithPath
+    )
+    val distance =
+      TokenEditDistance.fromBuffer(path, textDocument.text, buffers)
+
+    for {
+      occurrence <- textDocument.occurrences
```
tgodzik: I wonder if it would be faster to go over symbols, check if they are field or method and only as the last step find occurence?

tgodzik:
The question remains on how we go around and merge it? If we merge this PR this will cause lense to appear which do not do anything. We need the merge in the VS Code plugin for that.

@olafurpg @gabro @ckipp01  What do you think? 

1. Merge here and just wait for the next release. If somebody uses snapshots just tell them those don't work?

2. Or merge with the default setting to false? Later just before the release we enable them.

3. Or merge the VS Code PR and release without the server update, which will in turn create commands that do not do anything.

I think 3 is most reasonable for now.

ckipp01:
> The question remains on how we go around and merge it? If we merge this PR this will cause lense to appear which do not do anything. We need the merge in the VS Code plugin for that.
> 
> @olafurpg @gabro @ckipp01 What do you think?
> 
>     1. Merge here and just wait for the next release. If somebody uses snapshots just tell them those don't work?
> 
>     2. Or merge with the default setting to false? Later just before the release we enable them.
> 
>     3. Or merge the VS Code PR and release without the server update, which will in turn create commands that do not do anything.
> 
> 
> I think 3 is most reasonable for now.

I think we should just do nightly releases and not worry about it 😏 ...

Jokes aside, yea we ran into this in the last big release as well, and I think it will continue to happen 🤔 . For coc-metals I error on the side of merging things in that require the snapshot of metals since it's easy to use both the master branch of coc-metals and a snapshot or vice versa. However, in VS Code we have the tricky situation where I assume the average user doesn't want to build the VS Code extension locally in order to use a Metals snapshot.

Would it be possible to do a hybrid of one, and turn the feature off by default unless explicitly turned on by the client? That way the lenses won't appear? Then when we do merge and release the vs code client the default "on" will trigger it and the lenses will then appear?

tgodzik:
> Would it be possible to do a hybrid of one, and turn the feature off by default unless explicitly turned on by the client? That way the lenses won't appear? Then when we do merge and release the vs code client the default "on" will trigger it and the lenses will then appear?

I think that's a great idea! We can update the vscode plugin and label it to wait for the release.

PR #1490, opened 2020-03-10
New `create` option:--no-root-project option

tpasternak:
In some cases, metals generate synthetic '*-project-root' modules. The
new `--no-root-project' unconditionally prevents from it.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/pantsbuild/commands/ExportOptions.scala
+++ b/metals/src/main/scala/scala/meta/internal/pantsbuild/commands/ExportOptions.scala
@@ -15,6 +15,10 @@ case class ExportOptions(
         "If unspecified, coursier will be downloaded automatically."
     )
     coursierBinary: Option[Path] = None,
+    @Description(
+      "Unconditionally prevent generation of '*-project-root' modules"
+    )
+    noRootProject: Boolean = true,
     @Hidden()
     mergeTargetsInSameDirectory: Boolean = false
 )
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/pantsbuild/commands/ExportOptions.scala
```diff
@@ -9,6 +9,10 @@ case class ExportOptions(
       "Don't download *-sources.jar for 3rd party dependencies."
     )
     noSources: Boolean = false,
+    @Description(
+      "Unconditionally prevent generation of '*-project-root' modules"
+    )
+    noRootProject: Boolean = false,
```
olafurpg: Should this perhaps be enabled by default? The end goal is to avoid configuration options to tweak the export step, the defaults should "just work"

tpasternak: Ok, so maybe it would be better to make it fully CLI-controlled (without `isBaseDirectory(root.toNIO)` condition)?

olafurpg: The `isBaseDirectory` check is required to make IntelliJ functional when generating project roots. I'm leaning towards removing synthetic projects entirely since they seem to be causing correctness issues when testing, which is more serious than "project files view not perfect". We can fix the project files view separately, maybe via a BSP extension.

PR #1491, opened 2020-03-10
Fix Bloop sending a restart message when custom version of Bloop is used

tgodzik:
Previously, we were being asked to restart Bloop each time we opened a new window and the Bloop default version was changed. This is due to the fact that those settings are only being sent later on and not part of the default settings. Now, we only ask to restart if we know that the running version is different than the one specified. There might be a race condition, where those settings are sent after the connection to Bloop is established, but there is really no way around it.

I was sure I fixed it previously, but this was not the case.

There are be 3 scenarios when we ask to restart Bloop:
- bloop version is too old - that's a separate case after establishing a connection
- if the version doesn't correspond to the one set in the settings - nothing happens if no setting for Bloop was defined
- if the version was defined previously and changed to None, which means we should reverse to default

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -841,7 +841,11 @@ class MetalsLanguageServer(
           val expectedBloopVersion = userConfig.currentBloopVersion
           val correctVersionRunning =
             buildServer.map(_.version).contains(expectedBloopVersion)
-          if (buildServer.nonEmpty && !correctVersionRunning) {
+          val allVersionsDefined = buildServer.nonEmpty && userConfig.bloopVersion.nonEmpty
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -838,12 +838,13 @@ class MetalsLanguageServer(
           if (userConfig.symbolPrefixes != old.symbolPrefixes) {
             compilers.restartAll()
           }
-          if (userConfig.bloopVersion != old.bloopVersion) {
+          val expectedBloopVersion = userConfig.currentBloopVersion
+          val correctVersionRunning =
+            buildServer.map(_.version).contains(expectedBloopVersion)
```
olafurpg: Will this trigger a warning if I'm running a custom version of Bloop locally? Ideally, it would be good to only trigger this when the user has explicitly configured a bloop version through the settings before (I personally don't change this setting).

PR #1498, opened 2020-03-13
Handle fatal exceptions when evaluating worksheets.

### LINKED ISSUE #1456: Non Terminating Worksheet Evaluation for SO Condition
**Describe the bug**
Code that returns an SO in a normal App causes a non-terminating `Evaluated Worksheet` message and does not return an SO.

**To Reproduce**
Below code should reproduce issue within a `worksheet.sc` file.

```
def bigData = (1 to 1000000).toStream
def foldLeft[A, B](l: List[A], z: B)(f: (B, A) => B): B = l match {
  case Nil    => z
  case h :: t => foldLeft(t, f(z, h))(f)
}
def foldRightViaFoldLeft[A, B](l: List[A], z: B)(s: (A, B) => B): B =
  foldLeft(l, (x: B) => x)((e, f) => (x: B) => e(s(f, x)))(z)
foldRightViaFoldLeft(bigData.toList, 0)(_ + _) // doesn't terminate
```

**Installation:**
- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: 
```
metals.serverVersion": "0.8.0+294-4dcdec07-SNAPSHOT
```

olafurpg:
Fixes #1456. Previously, fatal exceptions were not caught while
evaluating worksheet making them stay stuck in "evaluating". Now,
fatal exceptions are caught in mdoc and are reported inline
in worksheets like normal non-fatal exceptions.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
@@ -23,7 +23,6 @@ import java.util.concurrent.TimeUnit
 import scala.meta.internal.metals.MutableCancelable
 import scala.meta.internal.metals.StatusBar
 import scala.meta.internal.pc.InterruptException
-import scala.util.control.NonFatal
 import java.util.concurrent.Executors
 import scala.meta.internal.metals.Diagnostics
 import scala.meta.internal.metals.Timer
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -120,11 +119,12 @@ class WorksheetProvider(
       }
     }
     val onError: PartialFunction[Throwable, Option[EvaluatedWorksheet]] = {
-      case NonFatal(e) =>
-        scribe.error(s"worksheet: $path", e)
-        None
       case InterruptException() =>
         None
+      case e: Throwable =>
+        // NOTE(olafur): we catch all exceptions because of https://github.com/scalameta/metals/issues/1456
```
kpbochenek: catching all exceptions is handled in mdoc and they are converted to NonFatal errors which are already handled, this case is not related to mentioned issue, right?

olafurpg: It's correct that they should be caught by mdoc now but I added this case nevertheless to be sure. If there's something fatally broken about the JVM I suspect it will show up in the logs regardless.

PR #1504, opened 2020-03-15
Add in new scalafmt default version

ckipp01:
This pr updates the default scalafmt version to the latest 2.4.2 so that when a user formats for the first time, that is the default version that will get added to the `.scalafmt.conf` file that is created. I also changed this to the version that the Metals codebase is using by doing the slim bootstrap outlined on the website and put it in the `bin` directory to replace the old one. I then ran it on the code base.


olafurpg:
LGTM 👍 I regret some of the diffs but it’s nice that Scalafmt has started taking more control over line breaks leaving less manual decisions to users.

PR #1507, opened 2020-03-16
Do not include scala boot library

tgodzik:
Previously. the 2.12.10 library was indexed in a target even if the target used only 2.13.1 Scala version. Now, only the right libraries are included.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
@@ -27,6 +27,8 @@ import scala.concurrent.Future
 import java.{util => ju}
 import scala.meta.pc.AutoImportsResult
 import org.eclipse.lsp4j.TextEdit
+import scala.util.Try
+import scala.meta.internal.pc.EmptySymbolSearch
 
 /**
  * Manages lifecycle for presentation compilers in all build targets.
@@ -63,11 +65,15 @@ class Compilers(
     scribe.info(
       "no build target: using presentation compiler with only scala-library"
     )
-    val compiler = configure(new ScalaPresentationCompiler()).newInstance(
-      s"metals-default-${mtags.BuildInfo.scalaCompilerVersion}",
-      PackageIndex.scalaLibrary.asJava,
-      Nil.asJava
-    )
+    val ramboSearch =
+      Try(new RamboSymbolSearch(workspace, buffers))
+        .getOrElse(EmptySymbolSearch)
+    val compiler =
+      configure(new ScalaPresentationCompiler(), ramboSearch).newInstance(
+        s"metals-default-${mtags.BuildInfo.scalaCompilerVersion}",
+        PackageIndex.scalaLibrary.asJava,
+        Nil.asJava
+      )
     ramboCancelable = Cancelable(() => compiler.shutdown())
     compiler
   }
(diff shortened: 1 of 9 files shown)


olafurpg:
The CI failure looks related?
```
 [error] ==> X tests.CompletionLspSuite.rambo  0.795s munit.FailException: /home/runner/work/metals/metals/tests/unit/src/test/scala/tests/CompletionLspSuite.scala:220
[error] 219:      )
[error] 220:      _ <- assertCompletion(
[error] 221:        "Properties@@",
[error] diff assertion failed
[error] => Obtained
[error] "Properties - java.util"
[error]  Properties - java.util
[error] +Properties - scala.util
```

tgodzik:
> The CI failure looks related?
> 
> ```
>  [error] ==> X tests.CompletionLspSuite.rambo  0.795s munit.FailException: /home/runner/work/metals/metals/tests/unit/src/test/scala/tests/CompletionLspSuite.scala:220
> [error] 219:      )
> [error] 220:      _ <- assertCompletion(
> [error] 221:        "Properties@@",
> [error] diff assertion failed
> [error] => Obtained
> [error] "Properties - java.util"
> [error] => Diff (- obtained, + expected)
> [error]  Properties - java.util
> [error] +Properties - scala.util
> [error]     at munit.Assertions.fail(Assertions.scala:134)
> ```

It does look related, I will take a look!

tgodzik:
@olafurpg The reason for the failure was that the rambo compiler did depend on the implicit boot Scala library.

I decided to add a separate symbol search for the rambo compiler, which setups everything, otherwise it did depend on the workspace one. It's possible to got to scala dependencies now, which would be a nice added bonus.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -63,11 +63,13 @@ class Compilers(
     scribe.info(
       "no build target: using presentation compiler with only scala-library"
     )
-    val compiler = configure(new ScalaPresentationCompiler()).newInstance(
-      s"metals-default-${mtags.BuildInfo.scalaCompilerVersion}",
-      PackageIndex.scalaLibrary.asJava,
-      Nil.asJava
-    )
+    val ramboSearch = new RamboSymbolSearch(workspace)
```
olafurpg: This constructor can fail now, which is not great since the rambo compiler is a lazy val. We should fallback to an empty symbol search on exception.

metals/src/main/scala/scala/meta/internal/metals/RamboSymbolSearch.scala
```diff
@@ -0,0 +1,86 @@
...
+import scala.meta.internal.mtags.Mtags
+
+class RamboSymbolSearch(workspace: AbsolutePath) extends SymbolSearch {
+
+  private val dependencySourceCache =
+    new TrieMap[AbsolutePath, ju.List[String]]()
+
+  private val scalaVersion = BuildInfo.scala212
+
+  private val fetch = Fetch
```
olafurpg: Let's move this to the `Embedded` object where we download other jars. We should also update the `DownloadDependencies` class to cache scala-library sources.

metals/src/main/scala/scala/meta/internal/metals/RamboSymbolSearch.scala
```diff
@@ -0,0 +1,86 @@
...
+  private val docs = new Docstrings(index)
+
+  def documentation(symbol: String): ju.Optional[SymbolDocumentation] =
+    docs.documentation(symbol)
+
+  def definition(x: String): ju.List[Location] = {
+    index
+      .definition(Symbol(x))
+      .map { definition =>
+        val destinationPath = definition.path.toFileOnDisk(workspace)
```
olafurpg: This logic looks duplicated from some other location, could we share the code somehow?

PR #1510, opened 2020-03-17
Add support for Scala 2.12.11

tgodzik:
This PR adds support for Scala 2.12.11, ~~however we need to wait with the stable release until https://github.com/scalameta/scalameta/pull/2004 is merged.~~

~~Semanticdb is causing Bloop to behave rather flakily, but that is under 4.3.4 version too.~~

Updated to 4.3.6, which should have everything fixed.

### DIFF
--- a/bin/test-release.sh
+++ b/bin/test-release.sh
@@ -10,6 +10,7 @@ coursier fetch \
   org.scalameta:mtags_2.13.1:$version \
   org.scalameta:mtags_2.13.0:$version \
   org.scalameta:mtags_2.12.10:$version \
+  org.scalameta:mtags_2.12.11:$version \
   org.scalameta:mtags_2.12.9:$version \
   org.scalameta:mtags_2.12.8:$version \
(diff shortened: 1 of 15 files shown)


tests/cross/src/test/scala/tests/pc/CompletionBacktickSuite.scala
```diff
@@ -29,7 +29,7 @@ class CompletionBacktickSuite extends BaseCompletionSuite {
   )
 
   check(
-    "comment",
+    "comment".only,
```
kpbochenek: remove

kpbochenek: I was pretty sure I also once forgotten to remove `.only` and CI fails if it finds this, super cool feature :)
(but this time I can't find it in logs, weird)

tgodzik: Fixed!

PR #1522, opened 2020-03-21
Update lsp4j to enable more 3.15 support

ckipp01:
Since there was a couple issues that mentioned things that LSP 3.15 supports (such as progress), I figured it'd be a good idea to update to the latest 0.9.0 if lsp4j which implements most of 3.15. I also went through and removed/updated a few of the deprecated things.

In short this pr:

- updates the lsp4j library
- removes and replaces some deprecation with the lsp4j update

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
@@ -214,8 +214,7 @@ final class Diagnostics(
         range,
         d.getMessage,
         d.getSeverity,
-        d.getSource,
-        d.getCode
+        d.getSource
       )
     }
     if (result.isEmpty) {
--- a/metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
+++ b/metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
@@ -6,6 +6,7 @@ import scala.meta.internal.decorations.DecorationOptions
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.decorations.ThemableDecorationInstanceRenderOptions
 import scala.meta.internal.decorations.ThemableDecorationAttachmentRenderOptions
+import scala.meta.internal.pc.HoverMarkup
 import MdocEnrichments._
 import scala.meta.internal.decorations.PublishDecorationsParams
 import scala.meta.io.AbsolutePath
@@ -16,8 +17,6 @@ import org.eclipse.lsp4j.MarkupKind
 
 class DecorationWorksheetPublisher() extends WorksheetPublisher {
 
-  private val scalaMarkdownTics = "```scala"
-  private val endTics = "```"
   private val commentHeader = " // "
 
   override def publish(
@@ -45,7 +44,7 @@ class DecorationWorksheetPublisher() extends WorksheetPublisher {
           s.position().toLsp,
           new MarkupContent(
             MarkupKind.MARKDOWN,
-            List(scalaMarkdownTics, s.details, endTics).mkString("\n")
+            HoverMarkup(s.details)
           ),
           ThemableDecorationInstanceRenderOptions(
             after = ThemableDecorationAttachmentRenderOptions(
--- a/metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
+++ b/metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
@@ -7,15 +7,18 @@ import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.decorations.ThemableDecorationInstanceRenderOptions
 import scala.meta.internal.decorations.ThemableDecorationAttachmentRenderOptions
 import MdocEnrichments._
-import org.eclipse.lsp4j.MarkedString
 import scala.meta.internal.decorations.PublishDecorationsParams
 import scala.meta.io.AbsolutePath
 import org.eclipse.lsp4j.Position
 import org.eclipse.lsp4j.Hover
+import org.eclipse.lsp4j.MarkupContent
+import org.eclipse.lsp4j.MarkupKind
 
 class DecorationWorksheetPublisher() extends WorksheetPublisher {
 
-  val commentHeader = " // "
+  private val scalaMarkdownTics = "```scala"
+  private val endTics = "```"
+  private val commentHeader = " // "
 
   override def publish(
       languageClient: MetalsLanguageClient,
@@ -40,7 +43,10 @@ class DecorationWorksheetPublisher() extends WorksheetPublisher {
       .map { s =>
         new DecorationOptions(
           s.position().toLsp,
-          new MarkedString("scala", s.details()),
+          new MarkupContent(
+            MarkupKind.MARKDOWN,
+            List(scalaMarkdownTics, s.details, endTics).mkString("\n")
+          ),
           ThemableDecorationInstanceRenderOptions(
             after = ThemableDecorationAttachmentRenderOptions(
               commentHeader + s.summary(),
--- a/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
@@ -3,6 +3,7 @@ package scala.meta.internal.pc
 import scala.meta.internal.mtags.MtagsEnrichments._
 import org.eclipse.lsp4j.CompletionItem
 import org.eclipse.lsp4j.CompletionItemKind
+import org.eclipse.lsp4j.CompletionItemTag
 import org.eclipse.lsp4j.CompletionList
 import org.eclipse.lsp4j.InsertTextFormat
 import scala.meta.internal.jdk.CollectionConverters._
@@ -216,7 +217,7 @@ class CompletionProvider(
         item.setKind(completionItemKind(member))
         item.setSortText(f"${idx}%05d")
         if (member.sym.isDeprecated) {
-          item.setDeprecated(true)
+          item.setTags(List(CompletionItemTag.Deprecated).asJava)
         }
         // NOTE: We intentionally don't set the commit character because there are valid scenarios where
         // the user wants to type a dot '.' character without selecting a completion item.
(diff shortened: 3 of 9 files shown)


ckipp01:
Related test failure
```
[error] ==> X tests.hover.HoverDefnSuite.2.12.10-package  0.002s munit.FailException: /home/runner/work/metals/metals/tests/cross/src/test/scala/tests/hover/HoverDefnSuite.scala:162
[error] 161:
[error] 162:  check(
[error] 163:    "package",
[error] diff assertion failed
[error] => Obtained
[error] "package b.pkg"
[error] +```scala
[error]  package b.pkg
[error] +```
```
However, the hover still works as expected since now scala is being set a bit different. Let me see if there is a way to ensure that Scala is being correctly passed in the hover suite.

metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
```diff
@@ -40,7 +43,10 @@ class DecorationWorksheetPublisher() extends WorksheetPublisher {
       .map { s =>
         new DecorationOptions(
           s.position().toLsp,
-          new MarkedString("scala", s.details()),
+          new MarkupContent(
+            MarkupKind.MARKDOWN,
+            List(scalaMarkdownTics, s.details, endTics).mkString("\n")
```
ckipp01: I was going to do a multiline string here, but scalafmt kept breaking the string causing compilation to fail. I'll report an issue to scalafmt about it.

metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
```diff
@@ -40,7 +43,10 @@ class DecorationWorksheetPublisher() extends WorksheetPublisher {
       .map { s =>
         new DecorationOptions(
           s.position().toLsp,
-          new MarkedString("scala", s.details()),
```
ckipp01: @deprecated use MarkupContent instead. from the LSP docs.

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_hover

mtags/src/main/scala/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -216,7 +217,7 @@ class CompletionProvider(
         item.setKind(completionItemKind(member))
         item.setSortText(f"${idx}%05d")
         if (member.sym.isDeprecated) {
-          item.setDeprecated(true)
```
ckipp01: deprecated Use `tags` instead if supported. From LSP docs.

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion

mtags/src/main/scala/scala/meta/internal/pc/HoverProvider.scala
```diff
@@ -1,17 +1,19 @@
...
 import scala.meta.pc.OffsetParams
 import scala.reflect.internal.{Flags => gf}
-import scala.meta.internal.jdk.CollectionConverters._
 import scala.util.control.NonFatal
 
 class HoverProvider(val compiler: MetalsGlobal, params: OffsetParams) {
   import compiler._
 
+  private val scalaMarkdownTics = "```scala"
+  private val endTics = "```"
```
ckipp01: I have this in here twice. Also in the `DecorationWorksheetPublisher`. Should I just leave as is, or do you think I should combine them and put them somewhere? Mainly I wasn't sure where to stick them.

mtags/src/main/scala/scala/meta/internal/pc/HoverProvider.scala
```diff
@@ -175,14 +177,13 @@ class HoverProvider(val compiler: MetalsGlobal, params: OffsetParams) {
...
-          List(
-            JEither.forRight[String, MarkedString](
-              new MarkedString(
-                "scala",
-                s"${symbol.javaClassSymbol.keyString} ${symbol.fullName}"
-              )
-            )
-          ).asJava
+          new MarkupContent(
+            MarkupKind.MARKDOWN,
```
ckipp01: I misunderstood this at first and put `scala`, but it looks like this should be `plainText` or `markdown` according to the spec. Then , you put ````scala` in the actual value. They give an example in the LSP docs.

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#markupContent

metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
```diff
@@ -214,8 +214,7 @@ final class Diagnostics(
         range,
         d.getMessage,
         d.getSeverity,
-        d.getSource,
-        d.getCode
```
ckipp01: In the spec it says that code `can be omitted`, and I don't believe we use it at all. Plus, now it's a `Either[String, Number]`, but it was also returning `null` often, so I just removed it. Let me know if I'm missing something, and we actually need this.

metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
```diff
@@ -40,7 +43,10 @@ class DecorationWorksheetPublisher() extends WorksheetPublisher {
...
+          new MarkupContent(
```
ckipp01: Also, this breaks from the [Decoration Protocol 0.1.0](https://scalameta.org/metals/docs/editors/decoration-protocol.html) with having the hover message being `MarkupContent` rather than `MarkedString`. I can update the docs, however, how do we version the decoration protocol? Do we need to keep history of the v0.1.0, or am I ok to just update it and make it v0.2.0? After we figure this out, I'll also need to update the vs code and coc-metals clients to match this.

This is also causing the integration test suite to fail because we are just looking at the raw string. However, I don't really see this any different than the other hovers, so I think it actually make sense to surround it with ````scala` tics. Thoughts?

kpbochenek:
I quickly looked at `v0.9.0` release notes and there is a section
`Parameters for the following requests have been changed from TextDocumentPositionParams` that says all `textDocument/XXX` methods have their arguments unique types and not generic `TextDocumentPositionParams`, we could also update that. 
It looks like they replaced arguments(https://github.com/eclipse/lsp4j/pull/412/files), shouldn't that make code not compile without adopting metals to those changes? Confusing :thinking:

metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
kpbochenek: There is a comment in conversion `bsp.Diagnostics -> lsp.Diagnostics`:
```
        // We omit diag.getCode since Bloop's BSP implementation uses 'code' with different semantics
        // than LSP. See https://github.com/scalacenter/bloop/issues/1134 for details
```
Issue is still open but was assigned 4 days ago, maybe it is worth just doing `Left(d.getCode)` ? (more complex as it is java and must handle null, but concept the same)
This way when you update conversion bsp->lsp it will be pushed to client.

What is this code for is actually a good question anyway.

mtags/src/main/scala/scala/meta/internal/pc/HoverProvider.scala
kpbochenek: maybe we could have method like `makeScalaMarkupContent(body)` that would handle all boilerplate(making markupContent with kind Markdown and wrapping body in scala tag). We could use that in both places.

ckipp01: Sounds good. I realized we had a `HoverMarkup` object that I could reuse, and just added another apply to it.

metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
ckipp01: I sort of just assumed that it was ok to update the protocol. I've gone ahead and updated the decoration protocol docs and updated the changelog just in case someone was using the protocol.

I also updated the worksheet suite to include the new hover text having ```scala tics. I've tested this manually to ensure there is no regression in the way things appear in both VS Code an coc-metals. As far as I can tell, it behaves exactly the same.

metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
ckipp01: Thanks for pointing out the conversation. I'm hesitant to change something to account for the changes in bloop when they aren't there yet. Since on the metals side we aren't using `code` for anything (again as far as I can tell), do you think it might be best to leave it out for now and first implement it if needed when there when it's finished and concrete just to ensure we don't have to potentially make the changes for the current case, and then change them again later? I'd rather just do that once.

tgodzik: I think we can drop code for now, it's not really useful anyway. This usually just means the code for the error/warning type. That can be useful in some automatic tooling, but here we just need to present data to the user. Let's drop it.

metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
tgodzik: I think it's fine to do 0.2.0, but probably best if @olafurpg confirms. The only issue I see is that will make the snapshot versions stop working. Maybe we should have the client send the version it is using? And send one or the other (MarkupContent or MarkupString)?

Might be more effort to change it than it's worth, we might need to remove MarkedString at some point, but s it deprecated itself?

metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
tgodzik: Do we need to use MarkupContent though? MarkedString is deprecated for Hover, but will it be removed altogether?

metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
ckipp01: Talked about his offline, but even with the current version of the extensions, this still works fine. The old shape that was returned was below:

```json
  "hoverMessage": {
        "language": "scala",
        "value": "x: Int \u003d 3"
      }
```
Now, the new format is below:
```json
```json
"hoverMessage": {
        "kind": "markdown",
        "value": "```scala\nx: Int \u003d 3\n```"
      },
```
The current extension is able to handled these both.

metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
ckipp01: Talked about his offline and we are good to move forward 👍

ckipp01:
> LGTM! @ckipp01 confirmed it will work without the changes to the VS Code plugin, so I think we are fine to merge this.

Sounds good! Again, if I did miss something, we can quickly revert.

Issue #1367, opened 2020-01-29
Support for Scala 3


tgodzik:
Scala 3 is currently not support in Metals and we'd like to change that.

This issue is used to track any progress on that front.

tgodzik:
Related PR in Bloop: https://github.com/scalacenter/bloop/pull/1165

tgodzik:
We've just merged an initial support for Scala 3 for 0.23-RC1 version, which should provide similar features to the Dotty Language Server. However there is still loads of work to do:

Still missing:
- [ ]  shorter type names and exact method parameter names
- [ ]  exact type when it comes to generics
- [ ]  scaladocs not showing for everything
- [ ]  lacking all advanced completions and ordering/filtering
- [ ] shorter type names and exact param names
- [ ] type signatures are not handled by dotty
- [ ] pattern matching unapply method signature help
- [ ] the most outer signature help is usually found

Known issues:
- [ ]  rambo Presentation Compiler sometimes loaded for Dotty after reload
- [ ]  Bloop sometimes loads without the proper settings - especially after Bloop restart
- [ ]  warnings from Bloop don't seem to be cleaned out - only after compile error
- [ ]  some features like rename will not work with Scala 2 non-parsable code - this is due to Mtags class heavily using the Scalameta parser

Demo of what is currently working:

![demo](https://user-images.githubusercontent.com/3807253/79450980-0ddcc800-7fe6-11ea-973f-4313fa916f70.gif)

bishabosha:
This is working very nicely for me in my initial testing. However, I found that the signature tooltip only appears for me on an enum constant if the enum is contained within an outer wrapper, and the tooltip does not show the correct definition for the first enum constant, I believe its overlapping with the invisible `$values` method which returns an array, which I saw in the semanticdb Symbols section always appears immediately before the first constant.

```scala
package example

object Outer:
  enum TestEnum:
    case A // class TestEnum: Array[example.Outer.TestEnum]
    case B // val B: example.Outer.TestEnum
    case C // val C: example.Outer.TestEnum

  enum TestEnum2:
    case Prod(a: String) // class Prod(a: String): example.Outer.TestEnum2
    case A // class TestEnum: Array[example.Outer.TestEnum]
    case B // val B: example.Outer.TestEnum
    case C // val C: example.Outer.TestEnum

enum TestEnum3:
  case A // no tooltip
  case B // no tooltip
  case C // no tooltip

```

tgodzik:
[...]

PR #1523, opened 2020-03-21
Add support for Scala 3

tgodzik:
This PR adds the basic support for Scala 3 with some more advanced features still missing. The plan is to add them incrementally after we manage to merge this PR. Metals should already be perfectly usable with Scala 3.

Biggest changes:
- moved FoldingProvider, MultilineStringFormattingProvider, didChange, didClose, documentSymbols behind the presentation compiler interface
- added an implementation for Dotty using mostly the Dotty Language Server code (mtags3)
- added common source set for Scala 2 and Scala 3 (mtags-common), it's added as unmanaged source set
- use the presentation compiler interface with additional classloader for tests in `test/cross`
- made most tests work for dotty - some are ignored, since some advanced features are missing
- anything in the `compat` Maps in tests shows what can be improved

Still missing:

*hovers*
- shorter type names and exact param names
- exact type in generics
- some docs not showing

*completions*
- lacking all advanced completions and ordering/filtering
- shorter type names and exact param names
- documentation from scaladocs

*method signature*
- type signatures are not handled by dotty
- short type names
- pattern matching unapply signatures
- the most outer signature help is usually found

*parser*
- ~~there might be some deduplicate errors coming from parser (still need to check)~~ looks to be fixed

Everything that is ignored is currently marked with:
`override def excludedScalaVersions = Set("0.22.0-RC1")`
or via ignore on a single test.

Connected to https://github.com/scalameta/metals/issues/1367

dotty-demo


tgodzik:
Looks like we are blocked on https://github.com/scalameta/metals/pull/1522, there seem to be some weird issues with the current version of LSP4J and as far as I checked it's fixed in 0.9.0

tgodzik:
Getting closer, but it seems when running on JDK 11, we're getting now:
```
Exception in thread "pool-6-thread-2" java.lang.NoClassDefFoundError: javax/tools/DiagnosticListener
```

build.sbt
```diff
@@ -27,10 +27,6 @@ inThisBuild(
       // https://github.com/scala/bug/issues/10448
       "-Ywarn-unused:imports"
     ),
-    addCompilerPlugin(
```
tgodzik: I add it per project, because it fails when added to a dotty project.

olafurpg: The following setting should work
```scala
libraryDependencies ++= {
  if (isScala2) List(compilerPlugin(...))
  else List()
}
```

olafurpg: Not under `inThisBuild` however, it needs to be added as a shared setting

```scala
val sharedSettings = List(
  libraryDependencies ++= {
    if (isScala2) List(compilerPlugin(...))
    else List()
  }
)

lazy val mtags = project
  .settings(
    sharedSettings,
    ...
```

build.sbt
```diff
@@ -198,11 +210,34 @@ val genyVersion = Def.setting {
   else "0.4.2"
 }
 
+lazy val mtags3 = project
```
olafurpg: It would be nice if we can fold this into the `mtags` module with `crossScalaVersions = List(2.12, 2.13, 0.23)`. That would allow us to share more code between scala 2 and 3. We may have to move Scala 2 specific code into a `scala-2` source directory but that's fine.

You can see how we configuire this setup in MUnit here https://github.com/scalameta/munit/blob/f22d9c3d841668f3aebaeaf0819438ca53fab7c5/build.sbt#L125-L146

tgodzik: @olafurpg Any idea how to disable the scalafix plugin based on Scala version? That's why I initially moved it into two modules. They do reuse a common directory for any common code, which I think is a bit simpler to do.

Is there actual need to have it in a single module?

tests/slow/src/test/scala/tests/feature/SyntaxErrorLspSuite.scala
```diff
@@ -1,7 +1,9 @@
-package tests
+package tests.feature
```
tgodzik: Moved it since it depends on publishing the 2.13 version, which slow test already do. Not sure why it worked before.

tgodzik:
Windows is failing due to the bug fixed already in scalameta, we might need to release a new version with the fix.

olafurpg:
There's already a new Scalameta release with a fix for Windows

build.sbt
tgodzik: Added separateSettings to all 2.12 project.

build.sbt
tgodzik: Also we don't have fiddle with the settings and numerous edge cases if we keep them in separate modules. 

I sharing code currently via:
```
 Compile / unmanagedSourceDirectories += file("mtags-common") / "src/main/scala",
```

tgodzik:
Ok, it looks like there is only a flaky HoverLspSuite test failure, which is unrelated.

tests/cross/src/test/scala/tests/pc/MacroCompletionSuite.scala
```diff
@@ -120,9 +148,7 @@ class MacroCompletionSuite extends BaseCompletionSuite {
     compat = Map(
       "2.11" ->
         """|fold[X](fa: Int => X, fb: String => X): X
-           |""".stripMargin,
-      // NOTE(tgodzik): the presentation compiler returns empty results here in 2.12.9
-      "2.13.0" -> ""
```
tgodzik: It seems that the issues were finally fixed - probably because we are using a much more limited classpath now.

tests/cross/src/test/scala/tests/pc/SignatureHelpDocSuite.scala
```diff
@@ -114,17 +106,7 @@ class SignatureHelpDocSuite extends BaseSignatureHelpSuite {
...
       "2.12.8" -> foldOlderDocs2,
-      "2.12.7" -> foldOlderDocs2,
-      "2.13" ->
-        s"""|$foldLatestDocs
-            |**Parameters**
-            |- `f`: the function to apply if nonempty.
-            |- `ifEmpty`: the expression to evaluate if empty.
-            |fold[B](ifEmpty: => B)(f: Int => B): B
-            |        ^^^^^^^^^^^^^
-            |  @param ifEmpty String the expression to evaluate if empty.
```
tgodzik: The difference was caused by the different implementation of collections, which caused parameters to be returned in a different order.

build.sbt
olafurpg: Do you need to disable the Scalafix plugin? As long as you don't run `scalafix` after `++0.21.0...` it's fine. We currently only run Scalafix on 2.12 (not 2.11 or 2.13). We can update sbt-scalafix to special handle Dotty so that the plugin doesn't do anything by default, it's a small fix.

It's desirable to have a single mtags module:

* no custom logic to resolve mtags3 instead of mtags
* idiomatic code reuse via `mtags/src/main/{scala,scala-2,scala-3,...}` folder structure. I have almost always regretted customizing my sbt build outside of these conventions.

Dotty is in my view just another Scala version and it we should try to use `crossScalaVersions` where possible to support it.

olafurpg: My biggest motivation to have a single mtags module is that we already use `crossScalaVersions` to abstract over 2.11/2.12/2.13. It would be a step backwards to add a different abstraction to support 0.x or 3.x

tgodzik: > no custom logic to resolve mtags3 instead of mtags

We will still need custom logic to resolve it, but I can see how hard it would to merge them.

tgodzik: Merged the directories together, but kept the modules separate, otherwise it will not be possible to have language support in Scala 3 build target, since targets correspond to modules.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -113,14 +122,80 @@ class Compilers(
...
+
   def didCompile(report: CompileReport): Unit = {
     if (report.getErrors > 0) {
       cache.get(report.getTarget).foreach(_.restart())
     } else {
       // Restart PC for all build targets that depend on this target since the classfiles
       // may have changed.
       for {
-        target <- buildTargets.inverseDependencies(report.getTarget)
+        target <- buildTargets.inverseDependencies(report.getTarget) + report
```
tgodzik: InverseDependencies doesn't contain the current target,

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -238,6 +240,7 @@ class MetalsLanguageServer(
       () => userConfig,
       config
     )
+    metalsServerConfig.getAndUpdate(config => config.fromInitParams(params))
```
tgodzik: Needed for folding option only.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -191,23 +216,6 @@ final class Doctor(
 
   def allTargets(): List[ScalaTarget] = buildTargets.all.toList
 
-  private def problemSummary: Option[String] = {
```
tgodzik: Changed to just print unsupported versions if there is an issue, since semanticDB is added automatically by Bloop, so if there is a problem then it means user is using an unsupported version.

metals/src/main/scala/scala/meta/internal/metals/Trees.scala
```diff
@@ -1,58 +0,0 @@
-package scala.meta.internal.metals
```
tgodzik: This was moved.

mtags/shared/src/main/scala-2/scala/meta/internal/mtags/MtagsEnrichments.scala
```diff
@@ -487,4 +503,65 @@ trait MtagsEnrichments {
       else Generator()
     }
   }
+
+  implicit class XtensionToken(token: m.Token) {
```
tgodzik: Moved mostly from MetalsEnrichments

mtags/shared/src/main/scala-2/scala/meta/internal/pc/ScalaCompilerAccess.scala
```diff
@@ -0,0 +1,84 @@
+package scala.meta.internal.pc
+
+import java.util.concurrent.ScheduledExecutorService
+import scala.tools.nsc.reporters.StoreReporter
+import scala.tools.nsc.interactive.ShutdownReq
+import scala.util.control.NonFatal
+import scala.meta.pc.PresentationCompilerConfig
+import scala.concurrent.ExecutionContextExecutor
+
+class ScalaCompilerWrapper(global: MetalsGlobal)
```
tgodzik: Extracted code from CompilerAccess to reuse it also with Scala 3

mtags/shared/src/main/scala-3/scala/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -0,0 +1,538 @@
...
+import scala.meta.internal.pc.DefinitionResultImpl
+import scala.meta.internal.pc.CompilerAccess
+import scala.meta.pc.VirtualFileParams
+import scala.meta.pc.PresentationCompilerConfig
+import scala.meta.pc.PresentationCompiler
+import scala.meta.pc.SymbolSearch
+import scala.meta.pc.DefinitionResult
+import scala.meta.pc.OffsetParams
+
+case class ScalaPresentationCompiler(
```
tgodzik: The code here is mostly copied from Dotty Language Server and will be mostly likely reworked heavily soon, so I wouldn't concentrate on this too much. Tests are a better  documentation showing what's currently working.

mtags/shared/src/main/scala-3/scala/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -0,0 +1,538 @@
...
+  def completionItemResolve(
+      item: CompletionItem,
+      symbol: String
+  ): CompletableFuture[CompletionItem] = {
+    CompletableFuture.completedFuture(
+      null
+    )
+  }
+
+  // TODO NOT IMPLEMENTED
```
tgodzik: Some of the features are not yet implemented, but Metals work perfectly well without them, so I would rather merge this without those features and work on them later. Plus it can be later worked on by multiple people when needed.

mtags/shared/src/main/scala-3/scala/scala/meta/internal/pc/SymbolPrinter.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import dotty.tools.dotc.printing.PlainPrinter
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.Symbols._
+import dotty.tools.dotc.core.Names._
+import dotty.tools.dotc.core.Types._
+import dotty.tools.dotc.core.Flags
+import scala.language.implicitConversions
+
+class SymbolPrinter(implicit ctx: Context) extends PlainPrinter(ctx) {
```
tgodzik: This uses PlainPrinter from Dotty, which doesn't have the needed method exposed. There is a lot of utility classes that me might end up copying or at least extending from.

tests/cross/src/main/scala/tests/BaseCodeActionSuite.scala
```diff
@@ -1,18 +1,18 @@
...
+import scala.meta.internal.pc.EmptyCancelToken
 
 abstract class BaseCodeActionSuite extends BasePCSuite {
 
   def cancelToken: CancelToken = EmptyCancelToken
 
   def params(
       code: String
-  ): (String, String, Int) = {
+  )(implicit testEnvironment: TestEnvironment): (String, String, Int) = {
```
tgodzik: Added `TestEnvironment`, which basically is the context of the test with the PresenationCompiler itself, symbolIndex and the scalaVersion itself.

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -1,121 +1,203 @@
...
+    val search = new TestingSymbolSearch(
+      ClasspathSearch.fromClasspath(myclasspath),
+      new Docstrings(index),
+      workspace,
+      index
+    )
+
+    val scalacOpts = scalacOptions(myclasspath)
+
+    val pc = TestEnvironment
```
tgodzik: Changed the way we run cross suite in order to be able to easily extend the existing tests without the need for having all test dependencies work in Scala 3. I now start a compiler for each version and create tests for all supported version.

As an added bonus, this way found a number of issues, which were missed because cross tests were not tested the same as mtags would eventually be used. This way is closer to that.

tests/cross/src/test/scala/tests/pc/CompletionBacktickSuite.scala
```diff
@@ -37,37 +43,10 @@ class CompletionBacktickSuite extends BaseCompletionSuite {
         |""".stripMargin,
     """|`///`: Int
        |""".stripMargin,
-    filterText = "///"
-  )
-
-  checkEdit(
```
tgodzik: Moved to BackTick suite, where Dotty tests are ignored.

tests/cross/src/test/scala/tests/pc/CompletionDocSuite.scala
```diff
@@ -537,25 +536,7 @@ class CompletionDocSuite extends BaseCompletionSuite {
...
-           |
-           | Pass a different value for rethrow if you want to probably
-           | unwisely allow catching control exceptions and other throwables
-           | which the rest of the world may expect to get through.
-           |
-           |**Type Parameters**
-           |- `T`: result type of bodies used in try and catch blocks
-           |
-           |**Parameters**
-           |- `fin`: Finally logic which if defined will be invoked after catch logic
```
tgodzik: In this and above the difference was actually only the parameter order, which is different when run with Scala 2.13 and 2.12. Probably due to reimplementation of standard collections. The tests now show what will really be in Metals, which is run in 2.12.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: Actually, it still make sense to print warning about lacking semanticDB in case of using other BSP servers. Did a fallback, since I think reporting unsupported Scala versions is more important.

tgodzik:
I updated to Scala 3 version 0.23.0-RC1, which contains some needed fixes. I also made the tests a bit more resilient, so that we don't have to change them all the time.

tests/mtest/src/main/scala/tests/BaseSuite.scala
```diff
@@ -58,16 +51,18 @@ class BaseSuite extends munit.FunSuite with Assertions {
...
       default: String,
       compat: Map[String, String],
-      scalaVersion: String = this.scalaVersion
+      scalaVersion: String
   ): String = {
     val postProcess = compatProcess
-      .get(scalaBinary(scalaVersion))
-      .orElse(compatProcess.get(scalaVersion))
+      .collect {
+        case (ver, process) if scalaVersion.startsWith(ver) => process
```
tgodzik: Changed this to startsWith, so that we can use "0." -> for Scala 3. We can also just add a condition for major version, opinions?

kpbochenek:
Why moving `mtags/src/...` -> `mtags/shared/src/...`?
In mtags there is only `shared` directory and scala sources for each version are in /src/main/scala-X.Y.Z

tgodzik:
> Why moving `mtags/src/...` -> `mtags/shared/src/...`?
> In mtags there is only `shared` directory and scala sources for each version are in /src/main/scala-X.Y.Z

I followed the convention set in Munit.

kpbochenek:
not super needed but when you will be about to merge could you just quickly check with any tool(let's say Intellij :P ) if there are any unused fields/methods etc. after those changes?

kpbochenek:
Did you check with JDK8 / JDK9+? Are there any problems we should work on here?

tgodzik:
> Did you check with JDK8 / JDK9+? Are there any problems we should work on here?

There were some issues with JDK 11, but those got fixed. And also tests are running on both JDK 8 and JDK 11

tests/cross/src/test/scala/tests/hover/HoverDefnSuite.scala
```diff
@@ -180,10 +202,10 @@ class HoverDefnSuite extends BaseHoverSuite {
       |  }
       |}
       |""".stripMargin,
-    """```scala
-      |head: Int
-      |```
-      |""".stripMargin
+    "head: Int".hover,
```
kpbochenek, line 205: this is a change also for scala2, something was also fixed for scala2 here?

tests/mtest/src/main/scala/tests/BaseSuite.scala
kpbochenek: startsWith looks okay, (collect + headOption == collectFirst)

olafurpg:
> I followed the convention set in Munit.

The `shared/` convention is only used for jvm/js/native cross-building. I don't think we need it for scala 2/3 cross-building

mtags/shared/src/main/scala-3/scala/scala/meta/internal/pc/ScalaPresentationCompiler.scala
kpbochenek: ok :+1:

.github/workflows/ci.yml
```diff
@@ -100,7 +100,7 @@ jobs:
             name: LSP integration tests
             os: ubuntu-latest
           - type: cross
-            command: sbt +cross/test
+            command: sbt cross/test
```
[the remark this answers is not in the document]
tgodzik: We are cross testing, but loading compiler for each version inside the tests.

build.sbt
```diff
@@ -27,10 +40,6 @@ inThisBuild(
       // https://github.com/scala/bug/issues/10448
       "-Ywarn-unused:imports"
     ),
-    addCompilerPlugin(
-      "org.scalameta" % "semanticdb-scalac" % V.scalameta cross CrossVersion.full
-    ),
-    scalacOptions += s"-P:semanticdb:sourceroot:${baseDirectory.in(ThisBuild).value}",
```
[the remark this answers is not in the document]
tgodzik: That's actually the default and we can't add it to all projects with Scala 3 here, since it will not use the semanticDB plugin.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -235,8 +235,11 @@ object BuildServerConnection {
   ): InitializeBuildResult = {
     val extraParams = BloopExtraBuildParams(
       BuildInfo.scalametaVersion,
-      BuildInfo.supportedScalaVersions.asJava
+      BuildInfo.supportedScalaVersions
```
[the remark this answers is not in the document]
tgodzik: It does, but there is no need to send supportedScalaVersions, which is basically an information on which Scala versions are supported by that Scalameta version.

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -254,11 +254,14 @@ class WorksheetProvider(
...
         scalaVersion = info.scalaVersion
-        isSupported = ScalaVersions.isSupportedScalaVersion(scalaVersion)
+        isSupported = ScalaVersions
+          .isSupportedScalaVersion(scalaVersion) && !ScalaVersions
+          .isScala3Version(scalaVersion)
         _ = {
           if (!isSupported) {
             scribe.warn(
-              s"worksheet: unsupported Scala version '${scalaVersion}', to fix this problem use Scala version '${BuildInfo.scala212}' instead."
+              s"worksheet: unsupported Scala version '${scalaVersion}', using Scala version '${BuildInfo.scala212}' without classpath instead.\n" +
```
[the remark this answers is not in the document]
tgodzik: It's the one used in Metals and we're most likely more sure about it. But there is no particular reason aside from that.

mtags/shared/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -126,25 +120,24 @@ final class MultilineStringFormattingProvider(
...
-        } else Nil
+  )(fn: (String, meta.Position) => List[TextEdit]): List[TextEdit] = {
+    val source = textId.getUri.toAbsolutePath
+    if (source.exists) {
+      val pos = range.getStart.toMeta(
+        Input.VirtualFile(source.toString(), sourceText)
+      )
+      if (pipeInScope(pos, sourceText, newlineAdded)) {
+        val tokens =
+          Input.VirtualFile(source.toString(), sourceText).tokenize.toOption
```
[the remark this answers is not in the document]
tgodzik: We should be able, but not 100 % sure yet. We might need to keep it in the new place.

mtags/shared/src/main/scala-2/scala/meta/internal/metals/TokenEditDistance.scala
```diff
@@ -125,7 +127,7 @@ final class TokenEditDistance private (
             }
           Some(revised.toLSP)
         case (start, end) =>
-          scribe.warn(s"stale range: $start $end")
+          logger.warning(s"stale range: $start $end")
```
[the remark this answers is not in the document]
tgodzik: I don't think scribe is accessible in mtags.

tests/cross/src/test/scala/tests/hover/HoverDefnSuite.scala
tgodzik: It's just a simpler way of adding the additional quotes. It wasn't used in this test case, but doesn't change anything.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: Folding was moved into the presentation compiler, so we need to update the presentation compiler config.

tgodzik:
Thanks for the review @kpbochenek ! I will apply some fixes and rebase next week.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
olafurpg: Hmm, I wonder if this change fixes weird issues I've hit on where "connect to build server" have fixed it 🤔 

nit: please try to manually break on infix operators
```suggestion
        target <- buildTargets.inverseDependencies(report.getTarget) +
          report.getTarget()
```

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -132,8 +132,7 @@ final class DefinitionProvider(
       snapshot: TextDocument
   ): ResolvedSymbolOccurrence = {
     // Convert dirty buffer position to snapshot position in "source"
-    val sourceDistance =
-      TokenEditDistance.fromBuffer(source, snapshot.text, buffers)
+    val sourceDistance = buffers.tokenEditDistance(source, snapshot.text)
```
olafurpg, line 135: Nice improvement to readability 👍

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -152,10 +153,15 @@ final class Doctor(
...
           s"Scala $scalaVersion is not yet supported"
         }
+
       } else {
-        s"Code navigation is not supported for this compiler version, change to " +
-          s"Scala ${BuildInfo.scala213} or ${BuildInfo.scala212} and " +
-          s"run 'Build import' to enable code navigation."
+        val versionToUpgradeTo =
+          if (ScalaVersions.isScala3Version(scalaVersion)) {
+            s"Scala3 ${BuildInfo.scala3}"
```
olafurpg, line 160: ?
```suggestion
            s"Scala ${BuildInfo.scala3}"
```

olafurpg:
PS. I am concerned about git rebase conflicts with #1538. I propose we try to get both of these features merged once we're happy with the overall architecture and leave smaller improvements to separate PRs.

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -1,121 +1,203 @@
...
+        .fetch()
+        .asScala
+        .map(_.toPath())
+
+    val extraScala211Classpath = ClasspathLoader
+      .getURLs(this.getClass.getClassLoader)
+      .filter(!_.toString().contains("scala"))
+      .map(url => Paths.get(url.toURI))
+
+    // pprint.log(extraScala211Classpath)
```
ckipp01, line 108: Leftover `pprint` can be removed.

tests/cross/src/test/scala/tests/pc/AutoImplementAbstractMembersSuite.scala
```diff
@@ -6,9 +6,16 @@ import tests.BaseCodeActionSuite
 import scala.meta.internal.metals.CompilerOffsetParams
 import scala.meta.internal.metals.TextEdits
 import scala.meta.internal.jdk.CollectionConverters._
+import java.net.URI
+import tests.TestEnvironment
+import scala.meta.internal.metals.ScalaVersions
 
 class AutoImplementAbstractMembersSuite extends BaseCodeActionSuite {
 
+  // @tgodzik currently not implemented for Dotty
```
ckipp01, line 15: Should we tag these with TODOs?

tgodzik:
> Is there a way to get back `+cross/test` instead of the new behavior where `cross/test` runs tests for all Scala versions? 

I think that should be perfectly doable. I wasn't sure about the change myself, but it was more correct. Sacrificing an easy of testing might not be that good idea. And I think I have some better udnerstanding of sbt to make it happen now :sweat_smile:

mtags/shared/src/main/scala-2/scala/meta/internal/metals/Trees.scala
```diff
@@ -0,0 +1,73 @@
...
+    dialect(fileUri).map { d =>
+      val input = Input.VirtualFile(Paths.get(fileUri).toString(), text)
+      d(input).parse[Source]
+    }
+  }
+  private def dialect(fileUri: URI): Option[Dialect] = {
+    Option(PathIO.extension(Paths.get(fileUri))).collect {
+      case "scala" => dialects.Scala
+      case "sbt" => dialects.Sbt
+      case "sc" => dialects.Sbt
```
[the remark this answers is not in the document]
tgodzik: I think it's a workaround, until we properly support Sbt I don't think it add much. This also just moved and is the current behaviour.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -508,14 +478,26 @@ object MetalsEnrichments
...
         .getOrElse(item.getClassDirectory.toAbsolutePath)
     }
-    def isSemanticdbEnabled: Boolean =
-      item.getOptions.asScala.exists { opt =>
-        opt.startsWith("-Xplugin:") && opt
-          .contains("semanticdb-scalac")
+
+    def isSemanticdbEnabled(scalaVersion: String): Boolean = {
+      if (ScalaVersions.isScala3Version(scalaVersion)) {
+        item.getOptions.asScala.exists { opt => opt == "-Ysemanticdb" }
```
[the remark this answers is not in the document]
tgodzik: Yes, we don't need to check for the plugin, since we don't need one.

build.sbt
```diff
@@ -164,25 +174,48 @@ lazy val V = new {
       .collect { case Some((a, b)) => s"$a.$b" }
       .toList
       .distinct
-  def supportedScalaVersions =
-    nonDeprecatedScalaVersions ++ deprecatedScalaVersions
-  def deprecatedScalaVersions =
+
+  // Scala 2
+  def deprecatedScala2Versions =
     Seq(scala211, "2.12.8", "2.12.9", "2.13.0")
```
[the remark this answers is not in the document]
tgodzik: `2.13.0` was released a while ago and had a couple of issue, so I don't think we can recommend that. `2.13.1` was quickly released with the fixes. `2.12.10` is still in use by a lot of codebases and it`s only recently that the new version was released.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
tgodzik: Fixed!

tests/mtest/src/main/scala/tests/BaseSuite.scala
tgodzik: Thanks, fixed!

tgodzik:
> Why moving `mtags/src/...` -> `mtags/shared/src/...`?
> In mtags there is only `shared` directory and scala sources for each version are in /src/main/scala-X.Y.Z

I removed the shared part, and just keep each Scala version in it's own directory.

@olafurpg I also changed back to the way we were doing cross tests previously. I had to juggle a bit with the source files to make it happen and make dummy implementations for Scala 3 for now (for things that are either way not used)

tests/cross/src/test/scala/tests/pc/AutoImplementAbstractMembersSuite.scala
tgodzik: Marked msot with TODOs

tests/cross/src/main/scala/tests/BasePCSuite.scala
tgodzik: Fixed, thanks!

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -245,7 +229,17 @@ object MetalsEnrichments
...
+     */
+    def toInputFromBuffers(buffers: Buffers): m.Input.VirtualFile = {
+      buffers.get(path) match {
+        case Some(text) => Input.VirtualFile(path.toString(), text)
+        case None => path.toInput
+      }
+    }
+
+    /**
+     * Resolve each path segment individually to prevent
```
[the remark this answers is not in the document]
tgodzik: Thanks for noticing, not sure how that disappeared. Fixed!

build.sbt
```diff
@@ -384,23 +474,71 @@ lazy val testSettings: Seq[Def.Setting[_]] = List(
...
+  val newState = Project
+    .extract(state.value)
+    .appendWithSession(
+      List(
+        scalaVersion.in(mtags) := scalaV,
+        useSuperShell.in(ThisBuild) := false
+      ),
+      state.value
+    )
+  val (s, _) = Project
```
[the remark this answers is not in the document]
tgodzik: You are right, it's removed.

mtags/shared/src/main/scala-2/scala/meta/internal/metals/Trees.scala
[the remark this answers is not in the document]
ckipp01: Sounds good!

build.sbt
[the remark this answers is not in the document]
ckipp01: Alright, that makes sense.

.gitignore
```diff
@@ -7,6 +7,7 @@
 .lib/
 dist/*
 target/
+target3/
```
olafurpg, line 10: We should find an alternative way to achieve this that follows sbt naming conventions. For synthetic JS/Native pure cross-projects they do `project.in(file(".js"))`.

tgodzik: We need a new target directory if we want a new module in the same file, otherwise sbt doesn't compile. I don't think there is a naming convention for that.

Even if we just use the module for adding build target without making it publishable.

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -1,105 +1,153 @@
...
+  def testPc(name: TestOptions, ignoredScalaVersions: Set[String] = Set.empty)(
+      fn: PresentationCompiler => Unit
+  ): Unit = {
+    if (excludedScalaVersions(scalaVersion)) {
+      test(name.ignore) {}
+    } else {
+      val testOptions: TestOptions =
+        name.withName(s"${name.name}_$scalaVersion")
+      val possiblyIgnored =
+        if (ignoredScalaVersions(scalaVersion)) testOptions.ignore
```
olafurpg, line 140: This can be replaced with a custom tag + test transform
```scala
case class IgnoreScalaVersion(versions: String*) extends Tag("NoScalaVersion")
  override def munitTestTransforms = super.munitTestTransforms ++ List(
    new TestTransform("IgnoreScalaVersion", { test =>
      val isIgnoredScalaVersion = test.tags
         .collect { case IgnoreScalaVersion(versions) => versions.toSet }
         .getOrElse(Set.empty)
      if (isIgnoredScalaVersion(scalaVersion)) test.ignore
      else test
    })
  )

test("foo".tag(IgnoreScalaVersion("2.11.12")) { ... }
```

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -1,105 +1,153 @@
...
+
+  val tmp: AbsolutePath = AbsolutePath(Files.createTempDirectory("metals"))
+
+  override def afterAll(): Unit = {
+    presentationCompiler.shutdown()
+    RecursivelyDelete(tmp)
+    executorService.shutdown()
+  }
+
+  def testPc(name: TestOptions, ignoredScalaVersions: Set[String] = Set.empty)(
```
olafurpg, line 131: I think we can replace `testPc` with the default MUnit `test` if we use test transforms. See following comments.

Do we need the `fn: PresentationCompiler => Unit` argument? It looks like we can access `presentationCompiler` directly

Also, what is the difference between an ignored scala version and excluded Scala version? My understanding is that `ignoredScalaVersions` is configurable per-test and `excludedScalaVersions` is configurable per-suite

olafurpg:
We may want to add dotty to the list of tests here https://github.com/scalameta/metals/blob/f97f1abfe945f237522fabe9a89e665bf7fd7a95/tests/slow/src/test/scala/tests/feature/CompletionCrossLspSuite.scala#L10

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
```diff
@@ -161,4 +200,11 @@
      * detail that the compiler is `null` when it has been reset.
      */
     public abstract boolean isLoaded();
+
+    protected CompletionList emptyCompletion(){
```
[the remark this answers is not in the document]
tgodzik: It was not necessary, but since it was useful to have in both PresentationCompiler implementations I thought it would good to have it here. Especially since this is java code we are interacting with.

I will put it into a separate object.

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -1,105 +1,153 @@
...
+      else
+        ScalaDependencies.scala2(scalaVersion)
+
+    val fetch = Fetch
+      .create()
+
+    val allDependencies = scalaDependencies ++ extraDependencies(scalaVersion)
+
+    allDependencies.foreach(fetch.addDependencies(_))
+    val libraries: Seq[Path] = fetch
```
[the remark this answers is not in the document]
tgodzik: Mostly it was due to the fact I was doing tests with a single Scala 2.12 library on classpath previously, I can switch back.

tgodzik:
> We may want to add dotty to the list of tests here

Added https://github.com/scalameta/metals/pull/1523/files#diff-05bb4e66315624fc56d2a3b998e0ea78R21-R56. 

Thanks for noticing the issue!

@olafurpg Thanks for the review! I addressed I think all the comments and I will still have a look at the changes to make sure nothing is leftover from the previous approaches to testing.

tests/cross/src/main/scala/tests/BasePCSuite.scala
tgodzik: Removed it altogether, with the additional tags that you suggested it was no longer neccessary.

tests/cross/src/main/scala/tests/BasePCSuite.scala
tgodzik: Added it here: https://github.com/scalameta/metals/pull/1523/files#diff-713c81e0230fa147e415ec5325807f59R133-R146

Had to change the last part to:
```
https://github.com/scalameta/metals/pull/1523/files#diff-713c81e0230fa147e415ec5325807f59R133-R146
```

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -1,105 +1,153 @@
...
+    executorService.shutdown()
+  }
+
+  def testPc(name: TestOptions, ignoredScalaVersions: Set[String] = Set.empty)(
+      fn: PresentationCompiler => Unit
+  ): Unit = {
+    if (excludedScalaVersions(scalaVersion)) {
+      test(name.ignore) {}
+    } else {
+      val testOptions: TestOptions =
```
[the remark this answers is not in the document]
tgodzik: Added here: https://github.com/scalameta/metals/pull/1523/files#diff-713c81e0230fa147e415ec5325807f59R133-R146

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -1,105 +1,153 @@
...
+  override def afterAll(): Unit = {
+    presentationCompiler.shutdown()
+    RecursivelyDelete(tmp)
+    executorService.shutdown()
+  }
+
+  def testPc(name: TestOptions, ignoredScalaVersions: Set[String] = Set.empty)(
+      fn: PresentationCompiler => Unit
+  ): Unit = {
+    if (excludedScalaVersions(scalaVersion)) {
```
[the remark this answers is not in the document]
tgodzik: Thanks! Added here: https://github.com/scalameta/metals/pull/1523/files#diff-713c81e0230fa147e415ec5325807f59R131

Wonder if we should change Munit to color the ignored suites to yellow? What do you think?

tests/cross/src/test/scala/tests/pc/AutoImportsSuite.scala
```diff
@@ -6,9 +6,16 @@ import scala.meta.internal.metals.CompilerOffsetParams
 import scala.meta.internal.jdk.CollectionConverters._
 import scala.meta.internal.metals.TextEdits
 import munit.Location
+import java.nio.file.Paths
+import tests.BuildInfoVersions
+import scala.meta.pc.PresentationCompiler
 
 class AutoImportsSuite extends BaseCodeActionSuite {
 
+  // @tgodzik TODO currently not implemented for Dotty
```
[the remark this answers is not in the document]
tgodzik: Removed it, those tests are really a documentation of what still needs to be done.

tests/cross/src/main/scala/tests/BasePCSuite.scala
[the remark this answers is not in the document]
tgodzik: Changed it back to using classpath, also taking into account Dotty library:
https://github.com/scalameta/metals/pull/1523/files#diff-713c81e0230fa147e415ec5325807f59R38-R42

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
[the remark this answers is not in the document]
tgodzik: It's here: https://github.com/scalameta/metals/pull/1523/files#diff-8d0d759de71bb32f0a0f4734991ae91dR1-R14

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -223,4 +208,49 @@ object Embedded {
...
     new URLClassLoader(allURLs, parent)
   }
+
+  def newScala3PresentationCompilerClassLoader(
+      info: ScalaBuildTarget,
+      scalac: ScalacOptionsItem
+  ): URLClassLoader = {
+    val pc = Dependency.of(
+      "org.scalameta",
+      s"mtags3_${ScalaVersions.dropVendorSuffix(info.getScalaVersion)}",
```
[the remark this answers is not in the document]
tgodzik: Thanks for noticing! It was working because I had everything published before.

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -1,105 +1,151 @@
...
-import java.util.concurrent.ScheduledExecutorService
 import scala.collection.Seq
+import scala.util.control.NonFatal
 import scala.meta.pc.PresentationCompiler
+import scala.meta.internal.pc.ScalaPresentationCompiler
+import scala.meta.internal.metals.PackageIndex
+import munit.Tag
 
 abstract class BasePCSuite extends BaseSuite {
-  def thisClasspath: Seq[Path] =
```
olafurpg, line 30: The technique of inspecting `this.getClass.getClassLoader` to extract scala-library.jar works when you run `+cross/test`. It's significantly faster compared to using `coursier.Fetch.create()`.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -286,7 +360,8 @@ class Compilers(
         config.compilers.copy(
           _symbolPrefixes = userConfig().symbolPrefixes,
           isCompletionSnippetsEnabled =
-            initializeParams.supportsCompletionSnippets
+            initializeParams.supportsCompletionSnippets,
+          isFoldOnlyLines = initializeParams.foldOnlyLines
```
tgodzik, line 364: Moved the configuration for `foldOnlyLines` here

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -119,9 +119,12 @@ object Embedded {
       dep: Dependency,
       scalaVersion: String
   ): Fetch = {
+
     val resolutionParams = ResolutionParams
       .create()
-      .forceVersions(
+
+    if (!ScalaVersions.isScala3Version(scalaVersion))
```
tgodzik, line 126: Removed the redundant method for create Scala 3 presentation compiler and just added this single condtion. Everything else is the same.

tgodzik:
@olafurpg I looked through the code again and removed some not necessary changes, but I think we should be good to merge unless you have some further comments. What do you think?

ckipp01:
This is a huge accomplishment! :tada:
Great job on this @tgodzik 👏

PR #1525, opened 2020-03-22
Add "case class" option in NewFilesProvider

### LINKED ISSUE #1524: Add "case class" option to NewFilesProvider
Case classes are very common in Scala so I think it would make sense to have them as an explicit option in the NewFilesProvider.

They also should have a slightly different template than regular classes, using parenthesis instead of braces.

gabro:
Closes #1524 

Previously we would offer the choice to create:

- class
- object
- trait
- package object
- worksheet

now we also include "case class" right after "class"

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
@@ -23,6 +23,8 @@ class NewFilesProvider(
 ) {
 
   private val classPick = MetalsQuickPickItem(id = "class", label = "Class")
+  private val caseClassPick =
+    MetalsQuickPickItem(id = "case-class", label = "Case class")
   private val objectPick = MetalsQuickPickItem(id = "object", label = "Object")
   private val traitPick = MetalsQuickPickItem(id = "trait", label = "Trait")
   private val packageObjectPick =
@@ -38,7 +40,8 @@ class NewFilesProvider(
     val newlyCreatedFile =
       askForKind
         .flatMapOption {
-          case kind @ (classPick.id | objectPick.id | traitPick.id) =>
+          case kind @ (classPick.id | caseClassPick.id | objectPick.id |
+              traitPick.id) =>
             askForName(kind)
               .mapOption(
                 createClass(directory, _, kind)
@@ -67,6 +70,7 @@ class NewFilesProvider(
         MetalsQuickPickParams(
           List(
             classPick,
+            caseClassPick,
             objectPick,
             traitPick,
             packageObjectPick,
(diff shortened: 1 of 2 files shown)


tgodzik:
Test failure looks related:
```
[error] ==> X tests.NewFilesLspSuite.new-trait-new-dir  0.735s munit.FailException: /home/runner/work/metals/metals/tests/unit/src/test/scala/tests/NewFilesLspSuite.scala:143
[error] 142:        assert(expectedFilePathAbsolute.exists)
[error] 143:        assertNoDiff(
[error] 144:          expectedFilePathAbsolute.readText,
[error] diff assertion failed
[error] => Obtained
[error]     """|package bar
[error]        |
[error]        |trait bar/Baz {
[error]        |  
[error]        |}
[error]        |""".stripMargin
[error]  ∙
[error] -trait bar/Baz {
[error] +trait Baz {
[error]    ∙
```

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -102,9 +106,12 @@ class NewFilesProvider(
     val path = directory.getOrElse(workspace).resolve(name + ".scala")
     //name can be actually be "foo/Name", where "foo" is a folder to create
     val className = directory.getOrElse(workspace).resolve(name).filename
+    val template = kind match {
+      case caseClassPick.id => caseClassTemplate(name)
```
tgodzik: You should use `className` here and below.

gabro: Or course, thanks for spotting the bug! I’ll fix it soon

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -181,4 +188,7 @@ class NewFilesProvider(
         |}
         |""".stripMargin
 
+  private def caseClassTemplate(name: String): String =
+    s"case class $name()"
```
olafurpg: Add final?

gabro: I thought about it, but it seemed a bit too opinionated? On the other hand, I never use final on case classes because I’m too lazy to type it so this may be what makes me start using it 😅

olafurpg: I feel like the custom wizard is more merited if it’s more than just adding “case”

gabro: Hey, it also adds different parenthesis! :D 

Anyway, added `final`, I think it makes sense

gabro:
Failure on Windows looks unrelated

```
[error] ==> X tests.HoverLspSuite.update-docstrings  5.026s munit.FailException: D:\a\metals\metals\tests\unit\src\test\scala\tests\HoverLspSuite.scala:94 Obtained empty output!
[error] 93:      _ <- server.didSave("a/src/main/scala/a/Def.scala")(s => s)
[error] 94:      _ <- server.assertHover(
[error] 95:        "a/src/main/scala/a/Main.scala",
```

PR #1526, opened 2020-03-22
Sort references before returning them

### LINKED ISSUE #1520: Return references in order
**Question**

This was originally brought to my attention via @poslegm on Discord, and I think it's a valid point. Let's say you have two files:

A.scala
```scala
package example.A

trait AA
trait BB extends AA
```
B.scala
```scala
package example.B

import example.A.AA

class CC extends AA
class DD extends AA
class EE extends AA
```
If I do a go to to definition on `AA` in `A.scala` I should get a list of definitions of all the places that `AA` is found. For example, here is the response from Metals.

<details><summary>Metals Output</summary><br>

```json
[Trace - 10:19:23 PM] Sending response 'textDocument/definition - (51)'. Processing request took 5ms
Result: [
  {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/A/A.scala",
    "range": {
      "start": {
        "line": 4,
        "character": 17
      },
      "end": {
        "line": 4,
        "character": 19
      }
    }
  },
  {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/B/B.scala",
    "range": {
      "start": {
        "line": 2,
        "character": 17
      },
      "end": {
        "line": 2,
        "character": 19
      }
    }
  },
  {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/B/B.scala",
    "range": {
      "start": {
        "line": 5,
        "character": 17
      },
      "end": {
        "line": 5,
        "character": 19
      }
    }
  },
  {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/B/B.scala",
    "range": {
      "start": {
        "line": 4,
        "character": 17
      },
      "end": {
        "line": 4,
        "character": 19
      }
    }
  },
  {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/B/B.scala",
    "range": {
      "start": {
        "line": 6,
        "character": 17
      },
      "end": {
        "line": 6,
        "character": 19
      }
    }
  }
]
```
</details>
In some clients (coc.nvim) for example, these are displayed as is, showing the following out of order results.

Screenshot 2020-03-20 at 22 32 47

The same happens when you do a find references on the same symbol.

<details><summary>Metals Output</summary><br>

```json
[Trace - 10:19:52 PM] Sending response 'textDocument/references - (53)'. Processing request took 4ms
Result: [
  {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/A/A.scala",
    "range": {
      "start": {
        "line": 4,
        "character": 17
      },
      "end": {
        "line": 4,
        "character": 19
      }
    }
  },
  {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/B/B.scala",
    "range": {
      "start": {
        "line": 2,
        "character": 17
      },
      "end": {
        "line": 2,
        "character": 19
      }
    }
  },
  {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/B/B.scala",
    "range": {
      "start": {
        "line": 5,
        "character": 17
      },
      "end": {
        "line": 5,
        "character": 19
      }
    }
  },
  {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/B/B.scala",
    "range": {
      "start": {
        "line": 4,
        "character": 17
      },
      "end": {
        "line": 4,
        "character": 19
      }
    }
  },
  {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/B/B.scala",
    "range": {
      "start": {
        "line": 6,
        "character": 17
      },
      "end": {
        "line": 6,
        "character": 19
      }
    }
  }
]
```
</details>
Screenshot 2020-03-20 at 22 32 56

** Actual question **

Does it make sense / is it possbile to correctly order these before they go back to the client? VS Code orders these correctly, but this isn't working as expected for `coc-metals` users.

**Installation:**

- Operating system: macOS
- Editor: Vim
- Metals version: v0.8.3

**Search terms**

references out of order, definition out of order

ckipp01:
This is a naive approach to close #1520 which sorts the references before they are returned. @tgodzik you mentioned changing to a sorted collection, but I wasn't sure how to do a sorted builder, or if that was what you actually meant. I didn't want to get rid of the builder as I assumed that it was used for performance. So this approach just sorts the collection right as it's about to be returned.

### DIFF
--- a/tests/mtest/src/main/scala/tests/Assertions.scala
+++ b/tests/mtest/src/main/scala/tests/Assertions.scala
@@ -1,5 +1,6 @@
 package tests
 
+import org.eclipse.lsp4j
 import munit.Location
 import scala.meta.io.AbsolutePath
 
@@ -57,6 +58,15 @@ trait Assertions extends munit.Assertions {
     }
   }
 
+  def assertSimpleLocationOrdering(locations: List[lsp4j.Location]): Unit = {
+    val grouped = locations.groupBy(_.getUri)
+    grouped.foreach {
+      case (_, group) => {
+        val lineStarts = group.map(_.getRange.getStart.getLine)
+        assert(lineStarts == lineStarts.sorted)
+      }
+    }
+  }
 }
 
 object Assertions extends Assertions
--- a/tests/unit/src/main/scala/tests/BaseRangesSuite.scala
+++ b/tests/unit/src/main/scala/tests/BaseRangesSuite.scala
@@ -1,0 +1,63 @@
+package tests
+
+import munit.Location
+import scala.concurrent.Future
+
+abstract class BaseRangesSuite(name: String) extends BaseLspSuite(name) {
+
+  def assertCheck(
+      filename: String,
+      edit: String,
+      expected: Map[String, String],
+      base: Map[String, String]
+  ): Future[Unit]
+
+  def check(name: String, input: String)(implicit loc: Location): Unit = {
+    val files = FileLayout.mapFromString(input)
+    val (filename, edit) = files
+      .find(_._2.contains("@@"))
+      .map {
+        case (fileName, code) =>
+          (fileName, code.replaceAll("(<<|>>)", ""))
+      }
+      .getOrElse {
+        throw new IllegalArgumentException(
+          "No `@@` was defined that specifies cursor position"
+        )
+      }
+    val expected = files.map {
+      case (fileName, code) =>
+        fileName -> code.replaceAll("@@", "")
+    }
+    val base = files.map {
+      case (fileName, code) =>
+        fileName -> code.replaceAll("(<<|>>|@@)", "")
+    }
+
+    test(name) {
+      cleanWorkspace()
+      for {
+        _ <- server.initialize(
+          s"""/metals.json
+             |{"a":
+             |  {
+             |    "compilerPlugins": [
+             |      "org.scalamacros:::paradise:2.1.1"
+             |    ],
+             |    "libraryDependencies": [
+             |      "org.scalatest::scalatest:3.0.5",
+             |      "io.circe::circe-generic:0.12.0"
+             |    ]
+             |  }
+             |}
+             |${input
+               .replaceAll("(<<|>>|@@)", "")}""".stripMargin
+        )
+        _ <- Future.sequence(
+          files.map(file => server.didOpen(s"${file._1}"))
+        )
+        _ <- assertCheck(filename, edit, expected, base)
+      } yield ()
+    }
+  }
+}
--- a/tests/unit/src/main/scala/tests/TestingServer.scala
+++ b/tests/unit/src/main/scala/tests/TestingServer.scala
@@ -268,6 +268,26 @@ final class TestingServer(
     })
   }
 
+  def assertReferences(
+      filename: String,
+      query: String,
+      expected: Map[String, String],
+      base: Map[String, String]
+  )(implicit loc: munit.Location): Future[Unit] = {
+    for {
+      references <- references(filename, query, base)
+    } yield {
+      references.foreach {
+        case (file, obtained) =>
+          val expectedImpl = expected(file)
+          Assertions.assertNoDiff(
+            obtained,
+            expectedImpl
+          )
+      }
+    }
+  }
+
   def assertReferenceDefinitionBijection()(
       implicit loc: munit.Location
   ): Unit = {
@@ -279,6 +299,7 @@ final class TestingServer(
       compare.definitionFormat
     )
   }
+
   def assertReferenceDefinitionDiff(
       expectedDiff: String
   )(implicit loc: munit.Location): Unit = {
@@ -942,6 +963,22 @@ final class TestingServer(
       implementations <- server.implementation(params).asScala
     } yield {
       TestRanges.renderLocationsAsString(base, implementations.asScala.toList)
+    }
+  }
+
+  def references(
+      filename: String,
+      query: String,
+      base: Map[String, String]
+  ): Future[Map[String, String]] = {
+    for {
+      (_, params) <- offsetParams(filename, query, workspace)
+      refParams = new ReferenceParams(new ReferenceContext(true))
+      _ = refParams.setPosition(params.getPosition())
+      _ = refParams.setTextDocument(params.getTextDocument())
+      references <- server.references(refParams).asScala
+    } yield {
+      TestRanges.renderLocationsAsString(base, references.asScala.toList)
     }
   }
 
--- a/tests/unit/src/main/scala/tests/TestingServer.scala
+++ b/tests/unit/src/main/scala/tests/TestingServer.scala
@@ -275,8 +275,11 @@ final class TestingServer(
       base: Map[String, String]
   )(implicit loc: munit.Location): Future[Unit] = {
     for {
-      references <- references(filename, query, base)
+      referenceLocations <- getReferenceLocations(filename, query, base)
     } yield {
+      Assertions.assertSimpleLocationOrdering(referenceLocations)
+      val references =
+        TestRanges.renderLocationsAsString(base, referenceLocations)
       references.foreach {
         case (file, obtained) =>
           val expectedImpl = expected(file)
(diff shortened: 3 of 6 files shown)


ckipp01:
> I was thinking `SortedSet` would be useful here, but either way this will normally be a small collection. One file usually doesn't contain that many occurrences.
> 
> It might be good to test it though, maybe use the code I did before?
> 
> [tgodzik/metals@master...better-reference-suite](https://github.com/tgodzik/metals/compare/master...better-reference-suite)
> 
> What do you think?

Thanks for linking me to that. Yea, when I looked at testing this to be honest I had a hard time. It wasn't extremely clear what the best way to test this is seeing that in the `TestingServer` we seem to construct the references a bit differently? I'm referring to this here:

https://github.com/scalameta/metals/blob/5f1214270a52d12396dfe9c6a4e345372978150c/tests/unit/src/main/scala/tests/TestingServer.scala#L217-L269

Regarding the stuff you linked, that's not merged in right? So would it be carrying all of that stuff over and adding it in? If so, it seems like a lot of stuff just to test that a collection is ordered. I guess I'm just trying to think of how to simply and easily test that `private def referenceLocations` correctly returns a sorted collection 🤔 , but unsure of the best way to do that in our test setup.

tgodzik:
> Regarding the stuff you linked, that's not merged in right? So would it be carrying all of that stuff over and adding it in? If so, it seems like a lot of stuff just to test that a collection is ordered. I guess I'm just trying to think of how to simply and easily test that private def referenceLocations correctly returns a sorted collection thinking , but unsure of the best way to do that in our test setup.

It's not merged yet, there was no need for it yet. However, I think this would be useful for writing tests in the future and will be useful here. The way we test references now is not useful when testing some edge cases, this would help with that.  We need to check the diff between references and definitions, which really is not intuitive.

There is even one failing test in that diff, which I totally forgot about, and would be super difficult to test without the changes.

ckipp01:
> > Regarding the stuff you linked, that's not merged in right? So would it be carrying all of that stuff over and adding it in? If so, it seems like a lot of stuff just to test that a collection is ordered. I guess I'm just trying to think of how to simply and easily test that private def referenceLocations correctly returns a sorted collection thinking , but unsure of the best way to do that in our test setup.
> 
> It's not merged yet, there was no need for it yet. However, I think this would be useful for writing tests in the future and will be useful here. The way we test references now is not useful when testing some edge cases, this would help with that. We need to check the diff between references and definitions, which really is not intuitive.
> 
> There is even one failing test in that diff, which I totally forgot about, and would be super difficult to test without the changes.

That makes sense. Would you prefer to get that branch merged in and have me rebase on it, or should I just copy that stuff over into my branch?

kpbochenek:
With this solution sorting will be also applied to rename results.

Do we want to sort rename results? (There is a VSCode rename preview, that could be a reason?)

RenameProvider searches for all classes in hierarchy which in this case produce list for a given file e.g. [A.fun(), A.fun(), AA.fun(), AA.fun()] (all results for class A properly sorted, then AA <: A properly sorted, but whole list won't be sorted by position in file only elements for the same class)

ckipp01:
> With this solution sorting will be also applied to rename results.
> 
> Do we want to sort rename results? (There is a VSCode rename preview, that could be a reason?)

Yea, I could see this maybe as a good thing? It wouldn't harm right?

ckipp01:
I rebased this on top of the work that @tgodzik did for the better references suite. I still need to add in some more checks yet for ordering now.

tests/mtest/src/main/scala/tests/Assertions.scala
```diff
@@ -57,6 +58,24 @@ trait Assertions extends munit.Assertions {
...
+    def sortByLocationPosition(
+        l1: lsp4j.Location,
+        l2: lsp4j.Location
+    ): Boolean = {
+      l1.getRange.getStart.getLine < l2.getRange.getStart.getLine
+    }
+
+    grouped.foreach {
+      case (_, group) =>
+        assertNoDiff(
```
ckipp01: So I'm a bit stuck with this. I feel like this isn't doing anything since I'm just using the same logic to sort in `ReferenceProvider` as we are to test it. I know this isn't right. However, again, the way the tests are, it doesn't make it easy to be explicit about what exactly is going to be returned. Do you think it makes sense to change the input of `check` to include another parameter that is just a string representation of the expected references to ensure that they are in order?

tests/unit/src/main/scala/tests/BaseRangesSuite.scala
```diff
@@ -0,0 +1,63 @@
...
+      .map {
+        case (fileName, code) =>
+          (fileName, code.replaceAll("(<<|>>)", ""))
+      }
+      .getOrElse {
+        throw new IllegalArgumentException(
+          "No `@@` was defined that specifies cursor position"
+        )
+      }
+    val expected = files.map {
```
ckipp01: So I originally had written a test that sorted the `Location`s that all came back and asserted that they were in the same order. However, I realized (I hope correctly), that this will actually implicitly test the same thing. If for some reason the references returned are out of order, this expected mapping with the string representation of what we gave the test won't match later on when we assert that the references returned match this. Let me know if this is indeed a good enough test for the actual ordering.

tgodzik: I don't think it will work that way though. We just replace all the ranges regardless of their sorting. But you can just add an assert . I will point to where to add it.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -873,6 +894,22 @@ final class TestingServer(
...
+      base: Map[String, String]
+  ): Future[Map[String, String]] = {
+    for {
+      (_, params) <- offsetParams(filename, query, workspace)
+      refParams = new ReferenceParams(new ReferenceContext(true))
+      _ = refParams.setPosition(params.getPosition())
+      _ = refParams.setTextDocument(params.getTextDocument())
+      references <- server.references(refParams).asScala
+    } yield {
+      TestRanges.renderLocationsAsString(base, references.asScala.toList)
```
tgodzik: You can just do `assert(references.sortWith(...) == references)`.

ckipp01: So I originally had that, but then I’m asserting the order is correct by the same logic I actually do the sorting with in the reference provider.

tgodzik: The other option is to just have a separate assert that we can write with the expected results like:
assertReferences(
  "A.scala:5-6"
  "A.scala:8-9"
)

And translate entries to locations.

ckipp01: Blah. Yea, this is frustrating that just testing the order is this difficult. I added the assertion back in right before you said this, so I will take it back out and go back to the drawing board.

tests/mtest/src/main/scala/tests/Assertions.scala
```diff
@@ -57,6 +58,15 @@ trait Assertions extends munit.Assertions {
     }
   }
 
+  def assertSimpleLocationOrdering(locations: List[lsp4j.Location]): Unit = {
```
ckipp01: So, I just took a step back and tried to make this as simple as possible. In order to actually be able to just leave the new reference suite the way you had it @tgodzik I just grab the locations when I have them, group them by uri, grab all the line starts, and then assert that they are in order for each uri. The alternative to this would be having to make an entirely different way to do the check. I feel like this is the easiest way to do a simple check on the ordering. What do you think?

ckipp01:
Test failure on windows seems unrelated
```
 [error] ==> X tests.HoverLspSuite.basic-rambo  2.423s munit.FailException: D:\a\metals\metals\tests\unit\src\test\scala\tests\HoverLspSuite.scala:35 Obtained empty output!
[error] 34:      )
[error] 35:      _ <- server.assertHover(
[error] 36:        "a/src/main/scala/a/Main.scala",
```

PR #1528, opened 2020-03-23
Add "Create new symbol '$name'..." code action

gabro:
This PR is a proposal to add a "Create new..." code action which is a quickfix for a "missing symbol" diagnostic.

Previously we would propose to import a symbol (if available) in response to "missing symbol" errors. Now we also propose to create a new class/trait/object

Here's a demo:

As you can see in the demo, this is very useful when sketching down a domain model, which requires creating a bunch of case classes in a top-down fashion.

What do you all think?

(This is currently based on top of #1525, but I'll rebase once it's merged)

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
@@ -195,11 +195,13 @@ class NewFilesProvider(
     )
   }
 
-  private def classTemplate(kind: String, name: String): String =
+  private def classTemplate(kind: String, name: String): String = {
+    val indent = "  "
     s"""|$kind $name {
-        |
+        |$indent
         |}
         |""".stripMargin
+  }
 
   private def caseClassTemplate(name: String): String =
     s"final case class $name()"
--- a/metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
@@ -32,7 +32,10 @@ class NewFilesProvider(
   private val worksheetPick =
     MetalsQuickPickItem(id = "worksheet", label = "Worksheet")
 
-  def createNewFileDialog(directoryUri: Option[URI]): Future[Unit] = {
+  def createNewFileDialog(
+      directoryUri: Option[URI],
+      name: Option[String]
+  ): Future[Unit] = {
     val directory = directoryUri
       .map(_.toString.toAbsolutePath)
       .orElse(focusedDocument().map(_.parent))
@@ -42,12 +45,12 @@ class NewFilesProvider(
         .flatMapOption {
           case kind @ (classPick.id | caseClassPick.id | objectPick.id |
               traitPick.id) =>
-            askForName(kind)
+            getName(kind, name)
               .mapOption(
                 createClass(directory, _, kind)
               )
           case worksheetPick.id =>
-            askForName(worksheetPick.id)
+            getName(worksheetPick.id, name)
               .mapOption(
                 createWorksheet(directory, _)
               )
@@ -96,6 +99,16 @@ class NewFilesProvider(
         case name if !name.cancelled => Some(name.value)
         case _ => None
       }
+  }
+
+  private def getName(
+      kind: String,
+      name: Option[String]
+  ): Future[Option[String]] = {
+    name match {
+      case Some(_) => Future.successful(name)
+      case None => askForName(kind)
+    }
   }
 
   private def createClass(
@@ -184,7 +197,7 @@ class NewFilesProvider(
 
   private def classTemplate(kind: String, name: String): String =
     s"""|$kind $name {
-        |  
+        |
         |}
         |""".stripMargin
 
--- a/metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
@@ -106,8 +106,8 @@ class NewFilesProvider(
       name: Option[String]
   ): Future[Option[String]] = {
     name match {
-      case Some(_) => Future.successful(name)
-      case None => askForName(kind)
+      case Some(v) if v.trim.length > 0 => Future.successful(name)
+      case _ => askForName(kind)
     }
   }
 
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewFile.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewFile.scala
@@ -19,7 +19,7 @@ class CreateNewFile() extends CodeAction {
 
     def createNewFile(diagnostic: l.Diagnostic, name: String): l.CodeAction = {
       val codeAction = new l.CodeAction()
-      codeAction.setTitle(CreateNewFile.title)
+      codeAction.setTitle(CreateNewFile.title(name))
       codeAction.setKind(l.CodeActionKind.QuickFix)
       codeAction.setDiagnostics(List(diagnostic).asJava)
       val directory = Paths
@@ -42,5 +42,5 @@ class CreateNewFile() extends CodeAction {
 }
 
 object CreateNewFile {
-  val title = "Create new..."
+  def title(name: String): String = s"Create new symbol '$name'..."
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewFile.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewFile.scala
@@ -1,0 +1,46 @@
+package scala.meta.internal.metals.codeactions
+
+import scala.meta.internal.metals._
+import scala.meta.internal.metals.MetalsEnrichments._
+import org.eclipse.{lsp4j => l}
+import scala.meta.pc.CancelToken
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import java.net.URI
+import java.nio.file.Paths
+
+class CreateNewFile() extends CodeAction {
+  override def kind: String = l.CodeActionKind.QuickFix
+
+  override def contribute(
+      params: l.CodeActionParams,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+
+    def createNewFile(diagnostic: l.Diagnostic, name: String): l.CodeAction = {
+      val codeAction = new l.CodeAction()
+      codeAction.setTitle(CreateNewFile.title)
+      codeAction.setKind(l.CodeActionKind.QuickFix)
+      codeAction.setDiagnostics(List(diagnostic).asJava)
+      val directory = Paths
+        .get(URI.create(params.getTextDocument().getUri()))
+        .getParent()
+        .toString()
+      codeAction.setCommand(ServerCommands.NewScalaFile.toLSP(List(null, name)))
+      codeAction
+    }
+
+    val codeActions = params.getContext().getDiagnostics().asScala.collect {
+      case d @ ScalacDiagnostic.SymbolNotFound(name)
+          if d.getRange().encloses(params.getRange()) =>
+        createNewFile(d, name)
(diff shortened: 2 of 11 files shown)


metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -23,28 +23,34 @@ class NewFilesProvider(
...
   private val traitPick = MetalsQuickPickItem(id = "trait", label = "Trait")
   private val packageObjectPick =
     MetalsQuickPickItem(id = "package-object", label = "Package Object")
   private val worksheetPick =
     MetalsQuickPickItem(id = "worksheet", label = "Worksheet")
 
-  def createNewFileDialog(directoryUri: Option[URI]): Future[Unit] = {
+  def createNewFileDialog(
+      directoryUri: Option[URI],
+      name: Option[String]
```
gabro: I've changed this to also accept a `name`, so that we can skip the name-picking dialog and just ask for a file kind.

This allows the code action to fully re-use the existing provider's flow, which I think is really nice

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewFile.scala
```diff
@@ -0,0 +1,46 @@
...
+        .get(URI.create(params.getTextDocument().getUri()))
+        .getParent()
+        .toString()
+      codeAction.setCommand(ServerCommands.NewScalaFile.toLSP(List(null, name)))
+      codeAction
+    }
+
+    val codeActions = params.getContext().getDiagnostics().asScala.collect {
+      case d @ ScalacDiagnostic.SymbolNotFound(name)
+          if d.getRange().encloses(params.getRange()) =>
```
gabro: this is `encloses` and not `overlapsWith`, because I don't think it makes sense to show the "Create new..." action when the user has selected a range which contains multiple `SymbolNotFound` diagnostics.

tests/unit/src/main/scala/tests/TestingClient.scala
```diff
@@ -342,8 +343,21 @@ final class TestingClient(workspace: AbsolutePath, buffers: Buffers)
...
+      command.getCommand(),
+      command.getArguments().asScala: _*
+    )
+  }
+
+  def applyCodeAction(codeAction: CodeAction, server: TestingServer): Unit = {
+    val edit = codeAction.getEdit()
+    val command = codeAction.getCommand()
+    if (edit != null) applyWorkspaceEdit(edit)
+    if (command != null) executeServerCommand(command, server)
```
gabro: This is now as per the LSP specification: if `edit` and `command` are both present, first `edit` is applied then `command` is executed.

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewFile.scala
```diff
@@ -0,0 +1,46 @@
...
+        createNewFile(d, name)
+    }
+
+    Future.successful(codeActions)
+
+  }
+}
+
+object CreateNewFile {
+  val title = "Create new..."
```
olafurpg: `create new symbol 'NAME'`?

tests/unit/src/test/scala/tests/codeactions/CreateNewFileLspSuite.scala
```diff
@@ -0,0 +1,79 @@
...
+import org.eclipse.lsp4j.ShowMessageRequestParams
+import scala.meta.internal.metals.MetalsEnrichments._
+
+class CreateNewFileLspSuite extends BaseCodeActionLspSuite("createNew") {
+
+  checkNewFile(
+    "basic",
+    """|package a
+       |
+       |case class School(name: String, location: <<Location>>)
```
tgodzik: Maybe let's add at least one other test case for other class kind?

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -98,6 +101,16 @@ class NewFilesProvider(
       }
   }
 
+  private def getName(
+      kind: String,
+      name: Option[String]
+  ): Future[Option[String]] = {
+    name match {
+      case Some(_) => Future.successful(name)
+      case None => askForName(kind)
```
tgodzik: Let's also maybe ask if the name is empty? I think that breaks in the current behaviour.

gabro: what you mean? Empty string? I can add a test but can you explain why do you think it breaks the current functionality?

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewFile.scala
gabro: I thought about adding something like that, but I feel like "symbol" is not very user-friendly (it's very much compiler lingo).

Anything else would require listing out the options ("Create new trait 'NAME'", "Create new class 'NAME'"), but think it would become too heavy.

I feel like "Create new..." is clear when presented in context and it avoids the problem of listing out all the options.

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -182,11 +195,13 @@ class NewFilesProvider(
     )
   }
 
-  private def classTemplate(kind: String, name: String): String =
+  private def classTemplate(kind: String, name: String): String = {
+    val indent = "  "
     s"""|$kind $name {
-        |  
+        |$indent
```
gabro: My editor is set to strip out whitespaces on save, so I've been struggling with this.

I think it's much clearer this way, anyway.

tests/unit/src/test/scala/tests/codeactions/CreateNewFileLspSuite.scala
gabro: done

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewFile.scala
olafurpg: I agree "symbol" is not idea but it's used in "rename symbol" and "go to symbol in file" in VS Code so I think it's OK to use it here.

gabro: You're right, it's used around, at least in VSCode. I'll make the change

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
gabro: Anyway, I've added handling blank values

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewFile.scala
gabro: Done!

image

olafurpg: Nice! I think it's a bit more self documenting what this option does with the name. Should we drop the `...` suffix?

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
olafurpg: I agree this is clearer. I also have my editor remove trailing whitespace on save, intellij has this enabled by default, so we should not rely on it in our tests

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewFile.scala
gabro: The `...` is used to convey the fact that the action starts a flow and it doesn't take effect immediately. 

I've seen this pattern used quite often, including in VSCode. See for instance

image

The commands ending with `...` start a flow that requires further interaction to complete.

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
ckipp01: I also like this as I get a warning in vim about trailing whitespaces all the time ha.

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewFile.scala
```diff
@@ -0,0 +1,46 @@
...
+        createNewFile(d, name)
+    }
+
+    Future.successful(codeActions)
+
+  }
+}
+
+object CreateNewFile {
+  def title(name: String): String = s"Create new symbol '$name'..."
```
ckipp01:
```suggestion
  def title(name: String): String = s"Create new symbol: '$name'..."
```

I'm not 100% how this is displayed, but this may make sense depending on how/where the title is displayed?

ckipp01: Ignore this, I see that the value is surrounded by ' ' now, so it makes sense.

gabro:
The failed test on Windows looks relevant, but I think it's flaky (especially because it succeeded on Linux and macOS): probably a race condition that caused the diagnostic not to disappear on time.

```
2020-03-23T14:14:44.3224758Z [error] ==> X tests.codeactions.CreateNewSymbolLspSuite.trait  6.271s munit.FailException: D:\a\metals\metals\tests\unit\src\test\scala\tests\codeactions\CreateNewSymbolLspSuite.scala:30
2020-03-23T14:14:44.4127307Z [error] 29:
2020-03-23T14:14:44.4133840Z [error] 30:  checkNewSymbol(
2020-03-23T14:14:44.4134203Z [error] 31:    "trait",
2020-03-23T14:14:44.4134455Z [error] diff assertion failed
2020-03-23T14:14:44.4134698Z [error] => Obtained
2020-03-23T14:14:44.4134967Z [error]     """|a/src/main/scala/a/A.scala:3:43: error: not found: type Location
2020-03-23T14:14:44.4135562Z [error]        |case class School(name: String, location: Location)
2020-03-23T14:14:44.4135868Z [error]        |                                          ^^^^^^^^
2020-03-23T14:14:44.4136127Z [error]        |""".stripMargin
2020-03-23T14:14:44.4136585Z [error] => Diff (- obtained, + expected)
2020-03-23T14:14:44.4136910Z [error] -a/src/main/scala/a/A.scala:3:43: error: not found: type Location
2020-03-23T14:14:44.4137203Z [error] -case class School(name: String, location: Location)
2020-03-23T14:14:44.4137468Z [error] -                                          ^^^^^^^^
2020-03-23T14:14:44.4148767Z [error] +
2020-03-23T14:14:44.4149080Z [error]     at munit.Assertions.fail(Assertions.scala:134)
```

kpbochenek:
Do you think it would be possible to move cursor into `final case class MyClass(<<here>>)` ? 

From your gif it seems like you want to be there anyway and I would like to be there too if I created a new class :) (not sure how hard it is though and if it is metals or more editor specific action?)

gabro:
@kpbochenek that's absolutely what I would like, and I think it should be possible by working with the range here https://github.com/scalameta/metals/blob/8141a99766998d87cf9ad1a33f430d70b9215a48/metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala#L192-L193

I would do it in a separate issue since it's quite independent from this PR.

gabro:
Ok, now the CI is failing for an unrelated test (I've fixed the previous race condition, which was my fault)

```
2020-03-23T15:26:33.1840250Z [error] ==> X tests.HoverLspSuite.basic-rambo  2.375s munit.FailException: D:\a\metals\metals\tests\unit\src\test\scala\tests\HoverLspSuite.scala:35 Obtained empty output!
2020-03-23T15:26:33.1840824Z [error] 34:      )
2020-03-23T15:26:33.1841093Z [error] 35:      _ <- server.assertHover(
2020-03-23T15:26:33.1841345Z [error] 36:        "a/src/main/scala/a/Main.scala",
2020-03-23T15:26:33.1873104Z [error]     at munit.Assertions.fail(Assertions.scala:134)
```

Merging

PR #1530, opened 2020-03-23
Create synthetic project root module only if source dir does not collide with existing modules

tpasternak:
image

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
+++ b/metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
@@ -262,10 +262,18 @@ private class BloopPants(
     )
     val isBaseDirectory =
       projects.iterator.filter(_.sources.nonEmpty).map(_.directory).toSet
+
+    def existsProjectDefinedAtRoot(
+        projects: Seq[C.Project],
+        root: AbsolutePath
+    ): Boolean =
+      projects.exists(_.directory == root.toNIO)
+
     // NOTE(olafur): generate synthetic projects to improve the file tree view
     // in IntelliJ. Details: https://github.com/olafurpg/intellij-bsp-pants/issues/7
     val syntheticProjects: List[C.Project] = sourceRoots.flatMap { root =>
-      if (isBaseDirectory(root.toNIO) || args.export.noRootProject) {
+      if (isBaseDirectory(root.toNIO) ||
+        existsProjectDefinedAtRoot(projects, root)) {
         Nil
       } else {
         val name = root
--- a/metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
+++ b/metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
@@ -262,9 +262,25 @@ private class BloopPants(
     )
     val isBaseDirectory =
       projects.iterator.filter(_.sources.nonEmpty).map(_.directory).toSet
+    // NOTE(olafur): generate synthetic projects to improve the file tree view
+    // in IntelliJ. Details: https://github.com/olafurpg/intellij-bsp-pants/issues/7
+    val syntheticProjects: List[C.Project] = sourceRoots.flatMap { root =>
+      if (isBaseDirectory(root.toNIO) || args.export.noRootProject) {
+        Nil
+      } else {
+        val name = root
+          .toRelative(AbsolutePath(workspace))
+          .toURI(isDirectory = false)
+          .toString()
+        // NOTE(olafur): cannot be `name + "-root"` since that conflicts with the
+        // IntelliJ-generated root project.
+        List(toEmptyBloopProject(name + "-project-root", root.toNIO))
+      }
+    }
     val generatedProjects = new mutable.LinkedHashSet[Path]
-    val byName = projects.map(p => p.name -> p).toMap
-    projects.foreach { project =>
+    val allProjects = syntheticProjects ::: projects
+    val byName = allProjects.map(p => p.name -> p).toMap
+    allProjects.foreach { project =>
       // NOTE(olafur): we probably want to generate projects with empty
       // sources/classpath and single dependency on the parent.
       if (!export.cycles.parents.contains(project.name)) {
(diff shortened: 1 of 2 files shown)


tpasternak:
This is a solution proposed by @lukaszwawrzyk . The old "-project-root" syntetic modules are back, but they're generated only when there is no regular target with the same source root. Initially, seems to work.

metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
```diff
@@ -261,9 +261,27 @@ private class BloopPants(
     )
     val isBaseDirectory =
       projects.iterator.filter(_.sources.nonEmpty).map(_.directory).toSet
+    // NOTE(olafur): generate synthetic projects to improve the file tree view
+    // in IntelliJ. Details: https://github.com/olafurpg/intellij-bsp-pants/issues/7
+    val syntheticProjects: List[C.Project] = sourceRoots.flatMap { root =>
+      if (isBaseDirectory(root.toNIO) || projects.exists(
+          p => p.sources.exists(_.getParent == root.toNIO)
```
olafurpg: Does this line have an effect since we primarily use sources globs?

metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
```diff
@@ -620,7 +638,9 @@ private class BloopPants(
         path.filename != "bloop.settings.json" &&
         !generatedProjects(path.toNIO)
       }
-      .foreach { path => Files.deleteIfExists(path.toNIO) }
+      .foreach { path =>
```
olafurpg: It looks like you're using an older version of Scalafmt, which is causing these diffs 🤔

metals/src/main/scala/scala/meta/internal/pantsbuild/commands/SharedCommand.scala
```diff
@@ -89,27 +84,6 @@ object SharedCommand {
     }
   }
 
-  def symlinkProjectViewRoots(project: Project): Unit = {
```
olafurpg: Can we still keep these symlinks? I'd like to get user feedback on whether they're useful or not

metals/src/main/scala/scala/meta/internal/pantsbuild/commands/SwitchCommand.scala
```diff
@@ -54,38 +52,23 @@ object SwitchCommand extends Command[SwitchOptions]("switch") {
...
-        app.info(
-          s"$verb project '${project.name}'${runBloopProjects(project.common)}"
-        )
+        app.info(s"switched to project '${project.name}', $runBloopProjects")
         0
       }
     }
   }
 
-  private def runBloopProjects(common: SharedOptions): String = {
```
olafurpg: This is still needed, we launch IntelliJ immediately before the bloop build is exported. We don't want to print a message saying "run bloop projects" when that command won't list anything

metals/src/main/scala/scala/meta/internal/pantsbuild/commands/SharedCommand.scala
tpasternak: Sure, this was just a proposal, It's up to you which option you prefer

metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
tpasternak: Right, I fixed it, but I'm still not sure if it's enough

metals/src/main/scala/scala/meta/internal/pantsbuild/commands/SwitchCommand.scala
tpasternak: I force-pushed this PR without reversion of `Minor polish for Pants integration`

metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
```diff
@@ -261,9 +261,28 @@ private class BloopPants(
     )
     val isBaseDirectory =
       projects.iterator.filter(_.sources.nonEmpty).map(_.directory).toSet
+    // NOTE(olafur): generate synthetic projects to improve the file tree view
+    // in IntelliJ. Details: https://github.com/olafurpg/intellij-bsp-pants/issues/7
+    val syntheticProjects: List[C.Project] = sourceRoots.flatMap { root =>
+      if (isBaseDirectory(root.toNIO) || projects.exists(p =>
+          p.sources.exists(_.getParent == root.toNIO)
```
olafurpg: Let's refactor the `projects.exists(...)` expression into a local def to make it more readable

metals/src/main/scala/scala/meta/internal/pantsbuild/BloopPants.scala
```diff
@@ -262,9 +262,38 @@ private class BloopPants(
...
     val isBaseDirectory =
       projects.iterator.filter(_.sources.nonEmpty).map(_.directory).toSet
+
+    def existsProjectDefinedAtRoot(
+        projects: Seq[C.Project],
+        root: AbsolutePath
+    ): Boolean =
+      projects.exists(p =>
+        p.sources.exists(_.getParent == root.toNIO)
+          || p.sourcesGlobs.exists(_.exists(_.directory == root.toNIO))
```
olafurpg: Is this condition necessary? The `.directory` field is always `baseDirectory` https://github.com/scalameta/metals/blob/775d2bff9c37acab065a9a465e6984db7ad1c03e/metals/src/main/scala/scala/meta/internal/pantsbuild/PantsGlobs.scala#L34

tpasternak: right

tpasternak: changed the whole condition `projects.exists(_.directory == root.toNIO)`

olafurpg: This case is already covered by `isBaseDirectory`, which contains `Project.directory` entries. I'm not sure this condition needs to exist.

I suspect the problem with the synthetic modules approach is that we're generating modules with a resource directory. I feel like we need a different approach. I think it's best to close this for now and I will try to think of something else.

PR #1531, opened 2020-03-23
Small clean-ups

alexarchambault:
This PR makes changes that I think can help new comers grok the metals code base. I stumbled upon those when diving into it myself for https://github.com/scalameta/metals-feature-requests/issues/80.

It mainly
- marks as `private` some methods that can be marked so,
- removes unused fields, methods, or method parameters, and
- ~adds the `case` modifier to some stateless classes, and~ removes `case` from some stateful classes (it seems it's a convention you're following most of the time - this applies it to more classes).


metals/src/main/scala/scala/meta/internal/metals/WorkspaceSearchVisitor.scala
```diff
@@ -28,7 +27,7 @@ class WorkspaceSearchVisitor(
     workspace: AbsolutePath,
     query: WorkspaceSymbolQuery,
     token: CancelChecker,
-    index: OnDemandSymbolIndex,
+    index: scala.meta.internal.mtags.GlobalSymbolIndex,
```
kpbochenek: It would be more consistent to keep it as `index: GlobalSymbolIndex` and do proper import (also similar in def below)

metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
```diff
@@ -53,14 +53,14 @@ final class BuildTools(
...
-  def isMill: Boolean = workspace.resolve("build.sc").isFile
+  private def isMill: Boolean = workspace.resolve("build.sc").isFile
   def isGradle: Boolean =
     workspace.resolve("build.gradle").isFile || workspace
       .resolve("build.gradle.kts")
       .isFile
   def isMaven: Boolean = workspace.resolve("pom.xml").isFile
   def isPants: Boolean = workspace.resolve("pants.ini").isFile
-  def isBazel: Boolean = workspace.resolve("WORKSPACE").isFile
+  private def isBazel: Boolean = workspace.resolve("WORKSPACE").isFile
```
tgodzik: I would leave out private in this case - it's inconsistent with the other `is` methods.

metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
```diff
@@ -21,7 +21,7 @@ import scala.concurrent.ExecutionContext
  *                            details according to BSP spec:
  *                            https://github.com/scalacenter/bsp/blob/master/docs/bsp.md#default-locations-for-bsp-connection-files
  */
-final class BuildTools(
+final case class BuildTools(
```
tgodzik: No need for case class here.

metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
```diff
@@ -24,7 +24,7 @@ object ScalaVersions {
   val isLatestScalaVersion: Set[String] =
     Set(BuildInfo.scala213, BuildInfo.scala212, BuildInfo.scala211)
 
-  def recommendedVersion(scalaVersion: String): String = BuildInfo.scala212
+  def recommendedVersion: String = BuildInfo.scala212
```
tgodzik: The change here should most likely be to use the `scalaVersion` to get the binary version and the recommended version for this binary version. It would be BuildInfo.scala212 for 2.11 and 2.12. BuildInfo.scala213 for 2.13.

metals/src/main/scala/scala/meta/internal/metals/WorkspaceSymbolProvider.scala
```diff
@@ -22,7 +21,7 @@ final class WorkspaceSymbolProvider(
     val workspace: AbsolutePath,
     statistics: StatisticsConfig,
     val buildTargets: BuildTargets,
-    val index: OnDemandSymbolIndex,
+    val index: scala.meta.internal.mtags.GlobalSymbolIndex,
```
tgodzik: -> `GlobalSymbolIndex` ?

metals/src/main/scala/scala/meta/internal/tvp/MetalsTreeViewProvider.scala
```diff
@@ -20,7 +19,7 @@ class MetalsTreeViewProvider(
     languageClient: MetalsLanguageClient,
     buildTargets: BuildTargets,
     compilations: () => TreeViewCompilations,
-    definitionIndex: OnDemandSymbolIndex,
+    definitionIndex: scala.meta.internal.mtags.GlobalSymbolIndex,
```
tgodzik: -> `GlobalSymbolIndex` ?

metals/src/main/scala/scala/meta/internal/metals/WorkspaceSearchVisitor.scala
alexarchambault: Thanks, should be fixed now. Seems I had underlooked that commit.

metals/src/main/scala/scala/meta/internal/metals/WorkspaceSymbolProvider.scala
alexarchambault: Thanks, should be fixed now. I had underlooked that commit I think.

metals/src/main/scala/scala/meta/internal/tvp/MetalsTreeViewProvider.scala
alexarchambault: Should be fixed too.

metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
alexarchambault: Yes, fixed.

metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
alexarchambault: Just reverted the initial change, and added a commit to do that.

alexarchambault:
About the case class / not case class thing: I'm fine reverting that if you're not used to that convention.

The `case` modifiers I had added were valid I think, they were only added to classes that don't break the case class semantics (That is, we can substitute an instance of a case class to another as long as `equals` say they're equals. Implying that they don't hold any meaningful field beyond their parameters. Which also means they don't actually hold a state, they're only possibly given references to stateful things, that were created elsewhere.)

olafurpg:
The CI failures are legitimate

```
 [error] /home/runner/work/metals/metals/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala:14:29: scala.meta.internal.metals.codeactions.ImplementAbstractMembers.type does not take parameters
[error]     ImplementAbstractMembers(compilers),
[error]                             ^
```

metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
alexarchambault: Should be fixed.

alexarchambault:
> The CI failures are legitimate

Just fixed those, the CI is green now.

PR #1534, opened 2020-03-24
Update to latest munit

### LINKED ISSUE #1527: Migrate to latest munit
There were a few breaking changes with munit v0.5.0 that affect us. For example, instead of overriding `munitRunTest` we should probably be utilizing the newly introduced `munitTestTransforms` for things like "FlakyWindows" and other "Flaky" tests. We can also use this for example in the `BasePCSuite` where we preface tests with the Scala version name if it's not 2.12.

I've been working on this already, so I figured I'd make an issue for it just so we're aware of it and someone else doesn't start work on it.

ckipp01:
The pr updates to the latest munit. It follows the new pattern of using `TestTransforms` that this commit demonstrates and outlines https://github.com/scalameta/munit/commit/f2d277df78d159ef644727416fdf8281e1e1e4f2.

Closes #1527

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -155,7 +155,7 @@ lazy val V = new {
   val mavenBloop = bloop
   val mdoc = "2.1.3"
   val scalafmt = "2.4.2"
-  val munit = "0.4.3"
+  val munit = "0.7.1"
   // List of supported Scala versions in SemanticDB. Needs to be manually updated
   // for every SemanticDB upgrade.
   def supportedScalaBinaryVersions =
--- a/tests/cross/src/main/scala/tests/BasePCSuite.scala
+++ b/tests/cross/src/main/scala/tests/BasePCSuite.scala
@@ -70,13 +70,15 @@ abstract class BasePCSuite extends BaseSuite {
     JdkSources().foreach(jdk => index.addSourceJar(jdk))
   }
 
-  override def munitNewTest(test: Test): Test = {
-    val testName =
-      if (isCI && BuildInfo.scalaCompilerVersion != BuildInfoVersions.scala212)
-        s"${BuildInfo.scalaCompilerVersion}-${test.name}"
-      else test.name
-    test.withName(testName)
-  }
+  override def munitTestTransforms: List[TestTransform] =
+    super.munitTestTransforms :+
+      new TestTransform("Add Compiler Version", { test =>
+        val testName =
+          if (isCI && BuildInfo.scalaCompilerVersion != BuildInfoVersions.scala212)
+            s"${BuildInfo.scalaCompilerVersion}-${test.name}"
+          else test.name
+        test.withName(testName)
+      })
 
   def indexScalaLibrary(): Unit = {
     val sources = Fetch()
--- a/tests/mtest/src/main/scala/tests/BaseSuite.scala
+++ b/tests/mtest/src/main/scala/tests/BaseSuite.scala
@@ -5,8 +5,8 @@ import scala.meta.internal.metals.JdkSources
 import scala.meta.internal.mtags
 import scala.meta.internal.semver.SemVer
 import scala.util.Properties
+import munit.Flaky
 import munit.Tag
-import munit.TestOptions
 
 class BaseSuite extends munit.FunSuite with Assertions {
 
@@ -32,18 +32,19 @@ class BaseSuite extends munit.FunSuite with Assertions {
 
   override def munitTimeout: Duration = Duration("10min")
 
-  // NOTE(olafur): always ignore flak test failures.
+  // NOTE(olafur): always ignore flaky test failures.
   override def munitFlakyOK = true
-  override def munitRunTest(
-      options: TestOptions,
-      body: => Any
-  ): Any = {
-    if (Properties.isWin && options.tags(FlakyWindows)) {
-      munitFlaky(options, body)
-    } else {
-      super.munitRunTest(options, body)
-    }
-  }
+
+  override def munitTestTransforms: List[TestTransform] =
+    super.munitTestTransforms ++ List(
+      new TestTransform(
+        "FlakyWindows",
+        test =>
+          if (test.tags(FlakyWindows) && Properties.isWin) test.tag(Flaky)
+          else test
+      ),
+      munitFlakyTransform
+    )
 
   private def scalaVersion: String =
     Properties.versionNumberString
--- a/tests/unit/src/main/scala/tests/BaseLspSuite.scala
+++ b/tests/unit/src/main/scala/tests/BaseLspSuite.scala
@@ -1,9 +1,9 @@
 package tests
+
 import java.nio.file.Files
 import java.util.concurrent.Executors
 import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContextExecutorService
-import scala.concurrent.Future
 import scala.meta.internal.io.PathIO
 import scala.meta.internal.metals.Buffers
 import scala.meta.internal.metals.ClientExperimentalCapabilities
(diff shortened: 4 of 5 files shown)


build.sbt
```diff
@@ -155,7 +155,7 @@ lazy val V = new {
   val mavenBloop = bloop
   val mdoc = "2.1.3"
   val scalafmt = "2.4.2"
-  val munit = "0.4.3"
+  val munit = "0.6.0"
```
gabro: We have released 0.7.1 this morning, which is unchanged except for the newly added ScalaCheck integration

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -70,13 +70,16 @@ abstract class BasePCSuite extends BaseSuite {
...
 
-  override def munitNewTest(test: Test): Test = {
-    val testName =
-      if (isCI && BuildInfo.scalaCompilerVersion != BuildInfoVersions.scala212)
-        s"${BuildInfo.scalaCompilerVersion}-${test.name}"
-      else test.name
-    test.withName(testName)
-  }
+  override def munitTestTransforms
+      : List[TestTransform] = super.munitTestTransforms ++ List(
```
gabro: minor, but can you manually break the line after `=` so that it does not break on the type annotation?

gabro: also, no need to wrap this in a `List`, you can do

```scala
super.munitTestTransforms :+ new TestTransform(...)
```

tests/mtest/src/main/scala/tests/BaseSuite.scala
```diff
@@ -5,11 +5,19 @@ import scala.meta.internal.metals.JdkSources
...
+import munit.TestValues
+import scala.concurrent.Future
+import scala.util.Failure
+import scala.util.Success
+import scala.concurrent.ExecutionContextExecutor
 
 class BaseSuite extends munit.FunSuite with Assertions {
 
+  implicit val ec: ExecutionContextExecutor =
+    scala.concurrent.ExecutionContext.global
```
gabro: you can use `munitExecutionContext` which is in scope

tests/mtest/src/main/scala/tests/BaseSuite.scala
```diff
@@ -32,18 +40,41 @@ class BaseSuite extends munit.FunSuite with Assertions {
...
+        val rerunCount = test.tags
+          .collectFirst { case Rerun(n) => n }
+          .getOrElse(1)
+        if (rerunCount == 1) test
+        else {
+          test.withBody(() => {
+            Future.sequence(1.to(rerunCount).map(_ => test.body()).toList)
+          })
+        }
+      }),
```
gabro: is this a new tag we want to use? Do we use it anywhere?

tests/mtest/src/main/scala/tests/BaseSuite.scala
```diff
@@ -32,18 +40,41 @@ class BaseSuite extends munit.FunSuite with Assertions {
...
+      new TestTransform("Flaky Windows", { test =>
+        if (Properties.isWin && test.tags.contains(FlakyWindows)) {
```
gabro: minor: you could do

```suggestion
        if (Properties.isWin && test.tags(FlakyWindows)) {
```

tests/mtest/src/main/scala/tests/BaseSuite.scala
```diff
@@ -32,18 +40,41 @@ class BaseSuite extends munit.FunSuite with Assertions {
...
+          test.withBodyMap(_.transformCompat {
+            case Success(value) => Success(value)
+            case Failure(exception) =>
+              if (munitFlakyOK) {
+                Success(new TestValues.FlakyFailure(exception))
+              } else {
+                throw exception
+              }
+          }(munitExecutionContext))
```
gabro: @olafurpg is there a way to reuse the default implementation (like we did in 0.5 using `munitFlaky`?

build.sbt
ckipp01: Ahhh you guys are moving too fast 😆

tests/mtest/src/main/scala/tests/BaseSuite.scala
gabro: I would actually recommend using it inline, so that we don't accidentally use it implicitly anywhere else

tests/mtest/src/main/scala/tests/BaseSuite.scala
ckipp01: We don't actually use it anywhere, but the old implementation of `FlakyTest` had the ability to pass in a number for max retries. Since I removed that I wanted to replace it with something that offered the same functionality. Do you think I should just remove it?

gabro: I personally think so, since it's unused. We can easily add it back if we need it

tests/mtest/src/main/scala/tests/BaseSuite.scala
ckipp01: I can't figure out where to put this inline... Can you give me a hint. 😬

ckipp01: Actually it doesn't matter. I'm going to remove "Rerun" anyways....

tests/mtest/src/main/scala/tests/BaseSuite.scala
olafurpg: `munitFlakyTransform` is the default transform, which you can apply as a `Test => Test` function

olafurpg: An alternative approach is to add the `Flaky` tag based on `Properties.isWin`

```scala
new TestTransform("FlakyWindows", test => if (test.tags(FlakyWindows) && Properties.isWin) test.tag(Flaky) else test)
```
As long as that transform comes before `munitFlakyTransform` then you don't need to re-implement the `.flaky` logic

ckipp01: Ah, cool. I'll do that.

tests/mtest/src/main/scala/tests/BaseSuite.scala
```diff
@@ -32,18 +32,21 @@ class BaseSuite extends munit.FunSuite with Assertions {
...
-      body: => Any
-  ): Any = {
-    if (Properties.isWin && options.tags(FlakyWindows)) {
-      munitFlaky(options, body)
-    } else {
-      super.munitRunTest(options, body)
-    }
-  }
+
+  case class Rerun(count: Int) extends munit.Tag("Rerun")
```
gabro: Unused?

tests/mtest/src/main/scala/tests/BaseSuite.scala
```diff
@@ -32,18 +32,21 @@ class BaseSuite extends munit.FunSuite with Assertions {
...
+
+  override def munitTestTransforms: List[TestTransform] =
+    super.munitTestTransforms ++ List(
```
gabro: :+ new TestTransform ?

ckipp01: But this is a list of the new `FlakyWindows` and `munitFlakyTransform`.

tests/mtest/src/main/scala/tests/BaseSuite.scala
ckipp01: Ahh yea. Good catch, I forgot to remove it.

tests/mtest/src/main/scala/tests/BaseSuite.scala
gabro: Ah, I hadn’t noticed. Sorry

PR #1535, opened 2020-03-24
Add auto-install of vscode liveshare extension to fastpass

wiwa:
This change will allow users of `fastpass` to automatically get Microsoft's `Live Share Extension Pack` installed for VSCode, which offers a novel way to pair program remotely on code using the hosting person's environment.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/pantsbuild/VSCode.scala
+++ b/metals/src/main/scala/scala/meta/internal/pantsbuild/VSCode.scala
@@ -23,6 +23,7 @@ object VSCode {
       scribe.info(s"updated: $settings")
       val code = codeCommand()
       exec(code, "--install-extension", "scalameta.metals")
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/pantsbuild/VSCode.scala
```diff
@@ -23,6 +23,7 @@ object VSCode {
       scribe.info(s"updated: $settings")
       val code = codeCommand()
       exec(code, "--install-extension", "scalameta.metals")
+      exec(code, "--install-extension", "ms-vsliveshare.vsliveshare-pack")
```
olafurpg: I'm not sure if it's the best UX to do this during `fastpass open --vscode PROJECT_NAME`. That command is specifically for launching VS Code, it would be a regression for normal VS Code users to re-run that command on every `fastpass open`. I wonder if it's better to add a new subcommand `fastpass liveshare PROJECT_NAME` 🤔 I'm not sure what the best UX is..

PR #1540, opened 2020-03-25
Allow multiple "Create new symbol" actions on a range

gabro:
Previously we would only add "Create new symbol" actions if the diagnostic range enclosed the code action range.

Now we suggest the action if the ranges overlap, which means we can suggest the action multiple times (one per missing symbol in the range).

This also fixes https://github.com/scalameta/coc-metals/issues/134 /cc @ckipp01

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewSymbol.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewSymbol.scala
@@ -35,7 +35,7 @@ class CreateNewSymbol() extends CodeAction {
 
     val codeActions = params.getContext().getDiagnostics().asScala.collect {
       case d @ ScalacDiagnostic.SymbolNotFound(name)
-          if d.getRange().encloses(params.getRange()) =>
+          if params.getRange().overlapsWith(d.getRange()) =>
         createNewSymbol(d, name)
     }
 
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateNewSymbol.scala
```diff
@@ -35,7 +35,7 @@ class CreateNewSymbol() extends CodeAction {
 
     val codeActions = params.getContext().getDiagnostics().asScala.collect {
       case d @ ScalacDiagnostic.SymbolNotFound(name)
-          if d.getRange().encloses(params.getRange()) =>
+          if params.getRange().overlapsWith(d.getRange()) =>
```
ckipp01: It funny because I remember reading your comment about this on the first pr, and it totally didn't even register that this wasn't going to work in coc-metals 😆 . Thanks for fixing this!

ckipp01: O, and I did test this locally, and it works great :)

gabro: Cool, thanks for confirming!

gabro:
Windows CI failure looks spurious

```
2020-03-25T15:12:31.8522036Z [error] ==> X tests.HoverLspSuite.update-docstrings  4.977s munit.FailException: D:\a\metals\metals\tests\unit\src\test\scala\tests\HoverLspSuite.scala:94 Obtained empty output!
2020-03-25T15:12:31.8522628Z [error] 93:      _ <- server.didSave("a/src/main/scala/a/Def.scala")(s => s)
2020-03-25T15:12:31.8522945Z [error] 94:      _ <- server.assertHover(
2020-03-25T15:12:31.8523142Z [error] 95:        "a/src/main/scala/a/Main.scala",
2020-03-25T15:12:31.8610657Z No more data in the client stdin, exiting...
```

Merging

PR #1541, opened 2020-03-25
Change go to parent lenses to true by default

tgodzik:
I just realised that I was totally mistaken before and the code lenses will work without the VS Code release.

@kpbochenek To confirm.


kpbochenek:
confirmed :+1: (misleading part was that I added a command to vscode when lenses pointed to location and those changes were needed for lenses to work, but after review I changed lenses to use symbol not location and new command in vscode was no longer needed - hence no changes needed in vscode to make lenses work)

olafurpg:
Feel free to ignore the Pants failure, I'm investigating why it started failing  https://pantsbuild.slack.com/archives/C046T6T9U/p1585214536013100

PR #1546, opened 2020-03-26
Specify cursor position when creating new files

### LINKED ISSUE #1529: Move cursor to a more convenient location when creating a new file
See the conversation here https://github.com/scalameta/metals/pull/1528#issuecomment-602694276

gabro:
Closes #1529 

Previously we would set the cursor position at the beginning of the file when creating a new file. Now we place the cursor in a more convenient position, depending on the type of file.

Demo:

### DIFF
--- a/tests/unit/src/test/scala/tests/NewFileTemplateSuite.scala
+++ b/tests/unit/src/test/scala/tests/NewFileTemplateSuite.scala
@@ -7,7 +7,7 @@ import org.scalacheck.Prop.forAll
 
 class NewFileTemplateSuite extends BaseSuite with ScalaCheckSuite {
 
-  test("it requires exactly one cursor marker") {
+  test("cursor-marker-errors") {
     intercept[IllegalArgumentException] {
       NewFileTemplate("no cursor markers")
     }
@@ -16,7 +16,7 @@ class NewFileTemplateSuite extends BaseSuite with ScalaCheckSuite {
     }
   }
 
-  property("cursor marker position is computed correctly") {
+  property("cursor-marker-position") {
     val template =
       s"""|package a
           |
--- a/tests/unit/src/test/scala/tests/NewFileTemplateSuite.scala
+++ b/tests/unit/src/test/scala/tests/NewFileTemplateSuite.scala
@@ -1,0 +1,33 @@
+package tests
+
+import munit.ScalaCheckSuite
+import scala.meta.internal.metals.NewFileTemplate
+import org.scalacheck.Gen
+import org.scalacheck.Prop.forAll
+
+class NewFileTemplateSuite extends BaseSuite with ScalaCheckSuite {
+
+  test("it requires exactly one cursor marker") {
+    intercept[IllegalArgumentException] {
+      NewFileTemplate("no cursor markers")
+    }
+    intercept[IllegalArgumentException] {
+      NewFileTemplate("many cursor @@ markers @@")
+    }
+  }
+
+  property("cursor marker position is computed correctly") {
+    val template =
+      s"""|package a
+          |
+          |case class Foo()
+          |""".stripMargin
+    val cursorOffsetGen = Gen.chooseNum(0, template.length)
+    forAll(cursorOffsetGen) { cursorOffset =>
+      val templateWithCursor = template.patch(cursorOffset, "@@", 0)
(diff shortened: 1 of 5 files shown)


tests/unit/src/test/scala/tests/NewFileTemplateSuite.scala
```diff
@@ -0,0 +1,33 @@
+package tests
+
+import munit.ScalaCheckSuite
+import scala.meta.internal.metals.NewFileTemplate
+import org.scalacheck.Gen
+import org.scalacheck.Prop.forAll
+
```
gabro: /cc @olafurpg dogfooding the new scalacheck module 👇

tests/unit/src/test/scala/tests/NewFileTemplateSuite.scala
```diff
@@ -0,0 +1,33 @@
...
+class NewFileTemplateSuite extends BaseSuite with ScalaCheckSuite {
+
+  test("it requires exactly one cursor marker") {
```
tgodzik: Should we stick to short names as in other suites?

tests/unit/src/test/scala/tests/NewFileTemplateSuite.scala
tgodzik: Do we really need Scalacheck here though? It's supposed to be used over a set of possible values and here it is just normal testing. I have never found a usage for property based testing and maybe it would be better to keep the tests consistent?

tests/unit/src/test/scala/tests/NewFileTemplateSuite.scala
gabro: Ah, absolutely, I completely forgot. Fixed

tests/unit/src/test/scala/tests/NewFileTemplateSuite.scala
gabro: > It's supposed to be used over a set of possible values and here it is just normal testing

I'm using Scalacheck to test "all" possible positions of the cursor inside a (fixed) document.


> Do we really need Scalacheck here though?

It's not strictly necessary, but it just gives me a little more confidence that the computation is not right by chance but it holds for "all" possible cursor positions.

That said, if you prefer we can remove it.

tgodzik: > I'm using Scalacheck to test "all" possible positions of the cursor inside a (fixed) document.

I am totally blind and have not seen it. It's a perfectly valid scenario to use it in. :+1:

gabro:
```
2020-03-27T08:11:21.6652922Z [error] ==> X tests.HoverLspSuite.basic-rambo  2.558s munit.FailException: D:\a\metals\metals\tests\unit\src\test\scala\tests\HoverLspSuite.scala:35 Obtained empty output!
2020-03-27T08:11:21.6653700Z [error] 34:      )
2020-03-27T08:11:21.6654013Z [error] 35:      _ <- server.assertHover(
2020-03-27T08:11:21.6654296Z [error] 36:        "a/src/main/scala/a/Main.scala",
2020-03-27T08:11:21.6654871Z [error]     at munit.Assertions.fail(Assertions.scala:178)
```

Unrelated flaky test (we should probably mark it as `FlakyWindows`)

Merging

tests/unit/src/test/scala/tests/NewFileTemplateSuite.scala
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  property("cursor-marker-position") {
+    val template =
+      s"""|package a
+          |
+          |case class Foo()
+          |""".stripMargin
+    val cursorOffsetGen = Gen.chooseNum(0, template.length)
+    forAll(cursorOffsetGen) { cursorOffset =>
```
olafurpg: What is the difference between `forAll` here and `0.to(template.length).foreach { .. }`?

gabro: ah, good point: in this case just the number of iterations (100 with ScalaCheck). I can revert to just using a range

olafurpg: I think `1.to(100)` would be clearer and more consistent with the rest of the tests. Property based testing is more helpful when you can't tests all possible inputs.

PR #1547, opened 2020-03-26
Add support to query remote language server as fallback.

olafurpg:
Previously, Metals only supported responding to definition/requests from a local index that was built by Metals. For some large codebases it's not practical to index all the code locally making it difficult to use Metals.

Now, Metals supports querying a remote language server as a fallback when the local index is not able to respond to definition/references requests. This feature is especially helpful in a large monorepo where there exists a lot of code but the user is only actively working on a small part of the codebase. With a remote language server, users can query the remote language server for source files that the user is not editing without building a local index.

Big thanks to @ShaneDelmore who has implemented a remote language server that is backed by a SQL index of SemanticDBs. Our early experiments show that this approach is scalable to large codebases and response times are ~200-300ms for most "goto definition" requests on a good internet connection. One "find references" requests that returned ~30k results took 9 seconds to respond with my 4G tethered internet connection from the countryside in Iceland.

This remote language server is currently tied to our internal Pants/CI infrastructure so there isn't much code that's worth sharing publicly at this point. We hope this initial PR sparks enough interest for people to start experimenting with implementing remote language servers that integrate with other build tools such as sbt/Gradle/Maven.

### DIFF
--- a/docs/contributors/remote-language-server.md
+++ b/docs/contributors/remote-language-server.md
@@ -1,0 +1,146 @@
+---
+id: remote-language-server
+title: Remote Language Servers
+---
+
+Metals has experimental support to offload certain requests to a remote language
+server. This feature can be used to navigate large codebases when its not
+possible to index all the code on a local computer.
+
+## Difference from local language servers
+
+There are some important differences between local and remote language servers:
+
+- Instead of JSON-RPC, a remote language server responds to HTTP POST requests
+  with an `application/json` header and a JSON-formatted body.
+- Instead of using absolute `file://` URIs, a remote language server uses
+  relative `source://` URIs. For example, the absolute URI
+  `file://path/to/workspace/src/main/scala/Address.scala` becomes the relative
+  URI `source://src/main/scala/Address.scala` when communicating with a remote
+  language server.
+
+## Methods
+
+Each remote language server method expects a JSON-formatted body of type
+`JsonRpcRequest<T>`.
+
+```ts
+interface JsonRpcRequest<T> {
+  /** The JSON-RPC method name, for example textDocument/definition */
+  method: string;
+
+  /** The parameter for the JSON-RPC method, for example `TextDocumentPositionParams` */
+  params: T;
+
+  /** The ID of this request, can be any integer number. */
+  id: number;
+}
+```
+
+### `textDocument/definition`
+
+The `textDocument/definition` request is sent from the client to the server to
+get the list of definitions for a given position.
+
+_Request_:
+
+- method: `textDocument/definition`
+- params: `JsonRpcRequest<TextDocumentPositionParams>`, where
+  `TextDocumentPositionParams` is defined in LSP.
+
+_Response_:
+
+- result: `Location[]`, as defined in LSP.
+
+_Example request_:
+
+```sh
+curl --location --request POST 'http://remote-language-server.com' \
+--header 'Content-Type: application/json' \
+--data-raw '{
+  "method": "textDocument/definition",
+  "params": {
+    "textDocument": {
+      "uri": "source://src/main/scala/Address.scala"
+    },
+    "position": {
+      "line": 5,
+      "character": 10
+    }
+  },
+  "id": 10
+}'
+```
+
+_Example response_:
+
+```json
+[
+  {
+    "uri": "source://src/main/scala/User.scala",
+    "range": {
+      "start": { "line": 61, "character": 15 },
+      "end": { "line": 61, "character": 31 }
+    }
+  }
+]
+```
+
+### `textDocument/references`
+
+The `textDocument/references` request is sent from the client to the server to
+get the list of all references to a symbol at a given position.
+
+_Request_:
+
+- method: `textDocument/references`
+- params: `JsonRpcRequest<ReferenceParams>`, where `ReferenceParams` is defined
+  in LSP.
+
+_Response_:
+
+- result: `Location[]`, as defined in LSP.
+
+_Example request_:
+
+```sh
+curl --location --request POST 'http://remote-language-server.com' \
+--header 'Content-Type: application/json' \
+--data-raw '{
+  "method": "textDocument/references",
+  "params": {
+    "textDocument": {
+      "uri": "source://src/main/scala/Address.scala"
+    },
+    "position": {
+      "line": 5,
+      "character": 10
+    },
+    "context": {
+      "includeDeclaration": true
+    }
+  },
+  "id": 10
+}'
+```
+
+_Example response_:
+
+```json
+[
+  {
+    "uri": "source://src/main/scala/User.scala",
+    "range": {
+      "start": { "line": 61, "character": 15 },
+      "end": { "line": 61, "character": 31 }
+    }
+  },
+  {
+    "uri": "source://src/main/scala/Country.scala",
+    "range": {
+      "start": { "line": 62, "character": 16 },
+      "end": { "line": 62, "character": 32 }
+    }
+  }
+]
+```
--- a/docs/contributors/remote-language-server.md
+++ b/docs/contributors/remote-language-server.md
@@ -12,7 +12,12 @@ possible to index all the code on a local computer.
 There are some important differences between local and remote language servers:
 
 - Instead of JSON-RPC, a remote language server responds to HTTP POST requests
-  with an `application/json` header and a JSON-formatted body.
+  with an `application/json` header and a JSON-formatted body. The reason HTTP
+  is chosen over JSON-RPC is because it makes the remote language server
+  accesible from more clients, for example via `curl`. A caveat with using HTTP
+  instead of JSON-RPC is that it's not possible for the remote language server
+  to push notification down to the client. In the future, we could consider
+  using JSON-RPC via websockets instead.
 - Instead of using absolute `file://` URIs, a remote language server uses
   relative `source://` URIs. For example, the absolute URI
   `file://path/to/workspace/src/main/scala/Address.scala` becomes the relative
--- a/docs/contributors/remote-language-server.md
+++ b/docs/contributors/remote-language-server.md
@@ -4,7 +4,7 @@ title: Remote Language Servers
 ---
 
 Metals has experimental support to offload certain requests to a remote language
-server. This feature can be used to navigate large codebases when its not
+server. This feature can be used to navigate large codebases when it's not
 possible to index all the code on a local computer.
 
 ## Difference from local language servers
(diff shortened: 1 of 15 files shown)


metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -61,17 +63,28 @@ final class DefinitionProvider(
...
+    val fromSnapshot = fromSemanticdb match {
+      case Some(doc) =>
+        definitionFromSnapshot(path, params, doc)
+      case _ =>
+        DefinitionResult.empty
+    }
+    val fromCompiler =
+      if (fromSnapshot.locations.isEmpty()) {
+        if (remote.isEnabledForPath(path)) {
+          Future(remote.definition(params).getOrElse(fromSnapshot))
```
gabro: I know `requests` is a synchronous http library (have you considered other async options, by the way?) but wouldn't it make sense for the API for `RemoteLanguageServer` to return futures?

Also, I'm not sure it has a huge impact, but it would probably make sense to mark it as `blocking` as in

```suggestion
          Future(blocking(remote.definition(params).getOrElse(fromSnapshot)))
```

docs/contributors/remote-language-server.md
```diff
@@ -0,0 +1,146 @@
...
+Metals has experimental support to offload certain requests to a remote language
+server. This feature can be used to navigate large codebases when its not
+possible to index all the code on a local computer.
+
+## Difference from local language servers
+
+There are some important differences between local and remote language servers:
+
+- Instead of JSON-RPC, a remote language server responds to HTTP POST requests
+  with an `application/json` header and a JSON-formatted body.
```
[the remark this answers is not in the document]
olafurpg: It's a good question, I elaborated on the motivation to choose HTTP over JSON-RPC.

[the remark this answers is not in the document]
gabro: Thanks for the clarification 👍

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
olafurpg: I considered using an async client like sttp but chose requests because it has a smaller footprint (total ~100kb). I refactored the `RemoteLanguageServer` API to return futures so that we can replace requests with an async client in the future. I didn't refactor `ReferencesProvider` to become async since it requires changing a large chunk of code and I'm not convinced it's worth the effort right now.

gabro: Thanks for the explanation. I agree it's probably not worth it to refactor `ReferencesProvider` right now

metals/src/main/scala/scala/meta/internal/remotels/RemoteLanguageServer.scala
```diff
@@ -0,0 +1,114 @@
...
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext
+
+class RemoteLanguageServer(
+    workspace: () => AbsolutePath,
+    config: () => UserConfiguration,
+    buffers: Buffers,
+    buildTargets: BuildTargets
+)(implicit ec: ExecutionContext) {
+  def isEnabled: Boolean = config().remoteLanguageServer.isDefined
```
ckipp01: I think this is only used in here and can probably be marked `private`?

docs/contributors/remote-language-server.md
```diff
@@ -0,0 +1,151 @@
+---
+id: remote-language-server
+title: Remote Language Servers
+---
+
+Metals has experimental support to offload certain requests to a remote language
+server. This feature can be used to navigate large codebases when its not
```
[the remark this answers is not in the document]
olafurpg: Good catch. Fixed.

docs/contributors/remote-language-server.md
```diff
@@ -0,0 +1,151 @@
...
+  },
+  {
+    "uri": "source://src/main/scala/Country.scala",
+    "range": {
+      "start": { "line": 62, "character": 16 },
+      "end": { "line": 62, "character": 32 }
+    }
+  }
+]
+```
```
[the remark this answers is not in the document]
olafurpg: Haha, I have thought about this as well. I added a new "Open questions" section that documents some concerns I have about this approach and the to stress that this is an experimental feature.

metals/src/main/scala/scala/meta/internal/remotels/RemoteLanguageServer.scala
olafurpg: I inlined this method into `isEnabledForPath`

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -61,17 +63,28 @@ final class DefinitionProvider(
...
+    val fromSnapshot = fromSemanticdb match {
+      case Some(doc) =>
+        definitionFromSnapshot(path, params, doc)
+      case _ =>
+        DefinitionResult.empty
+    }
+    val fromCompiler =
+      if (fromSnapshot.locations.isEmpty()) {
+        if (remote.isEnabledForPath(path)) {
+          remote.definition(params).map(_.getOrElse(fromSnapshot))
```
tgodzik: I think we can do something like:

```scala
   val fromSemanticdb =
      semanticdbs.textDocument(path).documentIncludingStale
    val fromSnapshot = fromSemanticdb match {
      case Some(doc) =>
        definitionFromSnapshot(path, params, doc)
      case _ =>
        DefinitionResult.empty
    }
    val fromIndex =
      if (fromSnapshot.locations.isEmpty() && remote.isEnabledForPath(path)) {
          remote.definition(params).map(_.getOrElse(fromSnapshot))
      } else {
         fromSnapshot
        }
      } 
    if (fromIndex.locations.isEmpty()) {
      compilers().definition(params, token)
    } else {
      if(fromIndex.isEmpty)  warnings.noSemanticdb(path)
      Future.successful(fromIndex)
    }
```

This way it will also work for rambo mode, otherwise we assume that anything outside the build targets doesn't need any kind of compiler.

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -94,7 +99,14 @@ final class ReferenceProvider(
...
         }
       case None =>
-        ReferencesResult.empty
+        // NOTE(olafur): we block here instead of returning a Future because it
+        // requires a significant refactoring to make the reference provider and
+        // its dependencies (including rename provider) asynchronous. The remote
+        // language server returns `Future.successful(None)` when it's disabled
+        // so this isn't even blocking for normal usage of Metals.
+        Await
+          .result(remote.references(params), Duration(1, TimeUnit.MINUTES))
```
tgodzik: Shouldn't we have the timeout configured in MetalsServerConfig? Or at least keep it somewhere in a value like  `RemoteLanguageServer.defaultTimeout`

metals/src/main/scala/scala/meta/internal/remotels/RemoteLanguageServer.scala
```diff
@@ -0,0 +1,114 @@
...
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext
+
+class RemoteLanguageServer(
+    workspace: () => AbsolutePath,
+    config: () => UserConfiguration,
+    buffers: Buffers,
+    buildTargets: BuildTargets
+)(implicit ec: ExecutionContext) {
+  def isEnabledForPath(path: AbsolutePath): Boolean =
```
tgodzik: `isOutsideBuildTargets` ? `isEnabledForPath` is rather misleading as it looks like there is some kind of configuration.

Together with using `Option[RemoteLanguageServer]` it would be the same result, but clearer.

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -44,7 +45,8 @@ final class DefinitionProvider(
     icons: Icons,
     statusBar: StatusBar,
     warnings: Warnings,
-    compilers: () => Compilers
+    compilers: () => Compilers,
+    remote: RemoteLanguageServer
```
tgodzik: Maybe `Option[RemoteLanguageServer]` ? It will be more clear that it might not be available.

metals/src/main/scala/scala/meta/internal/remotels/RemoteLanguageServer.scala
tgodzik: Though probably then it's not possible to just change the setting. Maybe just return None in references if `remoteLanguageServer` is not defined?

olafurpg: I prefer to keep distinguishing when the remote server is enabled vs. returning empty results. The key motivation to keep the distinction is that I don't want to use the remote language server to be enabled for files that are exported in the build graph.

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
olafurpg: This was a great suggestion! Applied (with minor changes)

metals/src/main/scala/scala/meta/internal/remotels/RemoteLanguageServer.scala
tgodzik: Sure, I just felt it's a bit counter intuitive. `isEnabledForPath` actually checks if file is not in buildTarget. So anything outside the source sets also fits the bill.

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -61,17 +63,28 @@ final class DefinitionProvider(
...
+      if (fromSnapshot.isEmpty && remote.isEnabledForPath(path)) {
+        remote.definition(params).map(_.getOrElse(fromSnapshot))
+      } else {
+        Future.successful(fromSnapshot)
+      }
+    fromIndex.flatMap { result =>
+      if (result.isEmpty) {
+        compilers().definition(params, token)
+      } else {
+        if (result.isEmpty && fromSemanticdb.isEmpty) {
```
tgodzik: Is ` fromSemanticdb.isEmpty` needed?  If the `result.isEmpty` is true then ` fromSemanticdb.isEmpty` must also be true.

olafurpg: We can get an empty result when the cursor is over a non-symbol position even if there’s a SemanticDB file. My goal was to try and isolate the “SemanticDB world” from the “remote language server world”

Issue #1548, opened 2020-03-26
Incorrect warning about scala version


ColOfAbRiX:
**Describe the bug**
After the upgrade to 0.8.3 I keep seeing a warning notification message on the bottom right of the window:
```
You are using a legacy Scala version 2.13.0, which might not be supported in future versions of Metals. Please upgrade to Scala 2.12.11.
```
and this is clearly incorrect (my project uses scala 2.13.0 and the warning tells me to "upgrade" to 2.12.11).

**To Reproduce**
I'm not sure. I'm using Metals 0.8.3, VSCode 1.41.3, Windows 10, SBT 1.3.8.
My build.sbt file is a pretty long multi project file and I ended up trying:
```
scalaVersion := "2.13.0"
Global / scalaVersion := "2.13.0"
ThisBuild / scalaVersion := "2.13.0"
```
but that made no difference. I don't see or find anything unusual in the output logs.

**Expected behavior**
This warning shouldn't appear as my Scala version is higher than the recommended one

**Screenshots**

**Installation:**
- Operating system: Windows 10
- Editor: Visual Studio Code
- Metals version: v0.8.3

ckipp01:
Thanks for reporting! It looks like the `deprecatedVersionWarning` def in `Doctor.scala` is pulling in the wrong recommended Scala version. It looks like the "recommended" Scala version we are pulling in by default is just `BuildInfo.212`. I'll post the relevant code/locations below incase someone wants to grab this since it should be a quick fix.

**Doctor where the message is formed**
https://github.com/scalameta/metals/blob/68cd1c945768f38b168cb03319649201d62831bd/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala#L174-L190

**Where we grab the version from**
https://github.com/scalameta/metals/blob/68cd1c945768f38b168cb03319649201d62831bd/metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala#L27

tgodzik:
I do already change it in the Scala 3 PR, so I might move it from there or maybe we can wait until it's merged?

ckipp01:
Ahh, perfect. We can probably just wait for that branch to merged _unless_ we are going to cut a release before then? What do you think?

tgodzik:
We would want to release before we merge it most likely. The Scala 3 support would be 0.9.0 in my book. I will move it to a separate PR then.

ColOfAbRiX:
Thank you guys!

PR #1549, opened 2020-03-27
Fix recommendations for Scala versions and add better messages.

tgodzik:
Previously, we would recommened Scala 2.12.11 even if the user is already using 2.13. Also, when using an unsupported Scala version we would just say that project is misconfigured, which is the same in case of a missing semanticDB. 

Now, we recommend the last version for each binary except 2.11. And we also display if any Scala version is unsupported in the warning.

Closes https://github.com/scalameta/metals/issues/1548

image

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
@@ -1,6 +1,7 @@
 package scala.meta.internal.metals
 
 import scala.meta.internal.mtags
+import scala.meta.internal.semver.SemVer
 
 object ScalaVersions {
 
@@ -24,15 +25,30 @@ object ScalaVersions {
   val isLatestScalaVersion: Set[String] =
     Set(BuildInfo.scala212, BuildInfo.scala213)
 
-  def recommendedVersion(scalaVersion: String): String = {
+  def latestBinaryVersionFor(scalaVersion: String): Option[String] = {
     val binaryVersion = scalaBinaryVersionFromFullVersion(scalaVersion)
     isLatestScalaVersion
       .find(latest =>
         binaryVersion == scalaBinaryVersionFromFullVersion(latest)
       )
-      .getOrElse {
-        BuildInfo.scala213
-      }
+  }
+
+  def recommendedVersion(scalaVersion: String): String = {
+    latestBinaryVersionFor(scalaVersion).getOrElse {
+      BuildInfo.scala213
+    }
+  }
+
+  def isFutureVersion(scalaVersion: String): Boolean = {
+    latestBinaryVersionFor(scalaVersion)
+      .map(latest =>
+        latest != scalaVersion && SemVer
+          .isCompatibleVersion(latest, scalaVersion)
+      )
+      .getOrElse(
+        isLatestScalaVersion
+          .forall(ver => SemVer.isCompatibleVersion(ver, scalaVersion))
+      )
   }
 
   def isCurrentScalaCompilerVersion(version: String): Boolean =
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
@@ -22,9 +22,18 @@ object ScalaVersions {
     }
 
   val isLatestScalaVersion: Set[String] =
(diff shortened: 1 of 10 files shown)


metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
```diff
@@ -22,9 +22,18 @@ object ScalaVersions {
...
+    Set(BuildInfo.scala212, BuildInfo.scala213)
+
+  def recommendedVersion(scalaVersion: String): String = {
+    val binaryVersion = scalaBinaryVersionFromFullVersion(scalaVersion)
+    isLatestScalaVersion
+      .find(latest =>
+        binaryVersion == scalaBinaryVersionFromFullVersion(latest)
+      )
+      .getOrElse {
+        BuildInfo.scala212
```
ckipp01: Just out of curiosity, if we do fall into this, why do we recommend 12 over 13?

tgodzik: Good question, mostly because we are using it Metals I guess, so we are most sure about. We could do the newest available version.

ckipp01: I'm sort of on the side that if we are in this scenario, we should recommend 13. /shrug

tgodzik: Switched to 2.13

tgodzik:
I added also a warning if a not yet supported version is released, which will be useful in any future upgrade. What do you all think?

ckipp01:
> I added also a warning if a not yet supported version is released, which will be useful in any future upgrade. What do you all think?

Nice, that makes sense 👍

kpbochenek:
Awesome work! (LGTM!)

2020-03-29-120044_1243x322_scrot

gabro:
One minor concern I have is about the 2.11 message: if you're on a minor version of 2.11 it might be bit strange to be told to upgrade to 2.13.

Nothing blocking anyway, truly great work!

tgodzik:
> One minor concern I have is about the 2.11 message: if you're on a minor version of 2.11 it might be bit strange to be told to upgrade to 2.13.
> 
> Nothing blocking anyway, truly great work!

I think you are right, let's recommend 2.11.12, first so that the user is less discouraged. A warning for later be displayed later on anyway.

tgodzik:
In case of 2.11.x messages  the message now looks:
```
You are using Scala versions 2.11.9, 2.12.1, 2.12.2, which are not supported in this version of Metals. Please upgrade to Scala versions 2.12.11, 2.13.1 or alternatively to legacy Scala 2.11.12.
```

I also refactored the logic for getting recommended versions and added a test suite, because I wanted to document what messages are being show in some particular cases. @gabro What do you think?

I am taking my time with this PR, because it's been very often that the messages were wrong here and users reported it often.

tgodzik:
When we refactor the Doctor we should most likely create an additional LSP suite for it to totally control the possible output.

tests/unit/src/test/scala/tests/MessagesSuite.scala
```diff
@@ -0,0 +1,47 @@
...
+        " Please upgrade to Scala version 2.12.11."
+    )
+  }
+
+  test("unspported-multiple2") {
+    assertDiffEqual(
+      Messages.UnsupportedScalaVersion
+        .message(Seq("2.11.9", "2.12.2", "2.12.1")),
+      "You are using Scala versions 2.11.9, 2.12.1, 2.12.2, which are not supported in this version of Metals." +
+        " Please upgrade to Scala version 2.12.11 or alternatively to legacy Scala 2.11.12."
```
ckipp01: 🤔 I guess I'm still not convinced that suggesting to a user that is currently on 2.12.2 that they switch to 2.11.12 is a good idea. In this specific situation, what would be the benefit of suggesting this current message over just `Please upgrade to Scala versions 2.12.11`?

tgodzik: They are both on `2.11.9` and `2.12.2`, it might be another module they have in the project that is for Scala native and we cannot recommend they upgrade to 2.12. This should be a rare situation though.

ckipp01: ahhhh, sorry, I misunderstood the test. Yea, then this is fine.

PR #1551, opened 2020-03-27
Add FlakyWindows tags to the hover suite

ckipp01:
This test has been failing quite often on windows. Following up on the comment on one of the recents prs, I'm marking the `basic-rambo` test in the hover suite as `FlakyWindows`

### DIFF
--- a/tests/unit/src/test/scala/tests/HoverLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/HoverLspSuite.scala
@@ -4,7 +4,7 @@ import scala.meta.internal.metals.Directories
 
 class HoverLspSuite extends BaseLspSuite("hover") with TestHovers {
 
-  test("basic") {
+  test("basic".tag(FlakyWindows)) {
     for {
       _ <- server.initialize(
         """/metals.json
@@ -22,7 +22,7 @@ class HoverLspSuite extends BaseLspSuite("hover") with TestHovers {
     } yield ()
   }
 
-  test("basic-rambo") {
+  test("basic-rambo".tag(FlakyWindows)) {
     for {
       _ <- server.initialize(
         """|/a/src/main/scala/a/Main.scala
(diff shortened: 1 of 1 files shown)


tgodzik:
I think the whole suite is flaky on Windows though, getting empty output sometimes. We might need to do a retry for hover or try to fix it on Windows at some point.

ckipp01:
> I think the whole suite is flaky on Windows though, getting empty output sometimes. We might need to do a retry for hover or try to fix it on Windows at some point.

Is it possible to "tag" a suite in munit?

olafurpg:
@ckipp01 I think you can edit the comment because you're a member of the Scalameta organization.

There's no way to tag an entire test suite but you can implement it with test transform, something like this
```
val flakyWindowsSuiteTransform = new TestTransform("FlakyWindowsSuite", _.tag(FlakyWindows))
override def munitTestTransforms = super.munitTestTransforms ++ List(flakyWindowsSuiteTransform)
```

ckipp01:
> @ckipp01 I think you can edit the comment because you're a member of the Scalameta organization.
> 
> There's no way to tag an entire test suite but you can implement it with test transform, something like this
> 
> ```
> val flakyWindowsSuiteTransform = new TestTransform("FlakyWindowsSuite", _.tag(FlakyWindows))
> override def munitTestTransforms = super.munitTestTransforms ++ List(flakyWindowsSuiteTransform)
> ```

Ah, of course. This is a great idea. I've gone ahead and done this.

ckipp01:
Pr to make hovers suite flaky on windows.... hover suite fails build. 🙄 

```
[error] ==> X tests.HoverLspSuite.basic-rambo  2.524s munit.FailException: D:\a\metals\metals\tests\unit\src\test\scala\tests\HoverLspSuite.scala:43 Obtained empty output!
[error] 42:      )
[error] 43:      _ <- server.assertHover(
[error] 44:        "a/src/main/scala/a/Main.scala",
```
@olafurpg did I do this wrong?

olafurpg:
The order matter so the new transform needs to come first 😅 my bad sorry!

ckipp01:
> The order matter so the new transform needs to come first 😅 my bad sorry!

haha, figures. No worries. I went ahead and instead just tagged them all in the suite since there is only a few of them.

PR #1556, opened 2020-03-28
Update lsp4j-debug

ckipp01:
This pr updates the lsp4j-debug dependency.

### DIFF
--- a/tests/unit/src/main/scala/scala/meta/internal/metals/debug/Debugger.scala
+++ b/tests/unit/src/main/scala/scala/meta/internal/metals/debug/Debugger.scala
@@ -76,7 +76,7 @@ final class Debugger(server: RemoteServer)(implicit ec: ExecutionContext) {
     server.setBreakpoints(args).asScala
   }
 
-  def step(threadId: Long, nextStep: DebugStep): Future[Unit] = {
+  def step(threadId: Int, nextStep: DebugStep): Future[Unit] = {
     nextStep match {
       case DebugStep.Continue =>
         val args = new ContinueArguments()
@@ -100,7 +100,7 @@ final class Debugger(server: RemoteServer)(implicit ec: ExecutionContext) {
     }
   }
 
-  def stackFrame(threadId: Long): Future[StackFrame] = {
+  def stackFrame(threadId: Int): Future[StackFrame] = {
     for {
       frame <- stackTrace(threadId).map(_.getStackFrames.head)
       scopes <- scopes(frame.getId).map(_.getScopes)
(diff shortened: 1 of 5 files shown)


tests/unit/src/main/scala/scala/meta/internal/metals/debug/Debugger.scala
```diff
@@ -122,20 +122,20 @@ final class Debugger(server: RemoteServer)(implicit ec: ExecutionContext) {
 
   def stackTrace(thread: Long): Future[StackTraceResponse] = {
     val args = new StackTraceArguments
-    args.setThreadId(thread)
-    args.setLevels(1L)
+    args.setThreadId(thread.toInt)
+    args.setLevels(1L.toInt)
```
kpbochenek: 1L.toInt -> 1 
:)

ckipp01: 😬

tests/unit/src/main/scala/scala/meta/internal/metals/debug/Debugger.scala
```diff
@@ -122,20 +122,20 @@ final class Debugger(server: RemoteServer)(implicit ec: ExecutionContext) {
 
   def stackTrace(thread: Long): Future[StackTraceResponse] = {
```
kpbochenek: I would change `thread: Long` -> `threadId -> Int` (threadId is Int now)

ckipp01:
> In release notes of lsp4j: `In DebugProtocol fields which used to be defined as Long are now Integer`
> 
> I think we should update our internal variables to be `Int` for `threadId`, `frameId`, `variableReference` and `breakpoint->id` (all that changed Long -> Integer). Right now you only convert right before using lsp4j objects but internally we still represent them as Longs. Might be better to avoid `.toLong` and `.toInt` altogether
> 
> What I mean e.g.:
> `private val breakpoints = TrieMap.empty[Long, Breakpoint]` (should be [Int, Breakpoint]) etc.

Yea, that's probably a much better approach. I'll work on changing this over.

tests/unit/src/main/scala/scala/meta/internal/metals/debug/Debugger.scala
ckipp01: Do you think it also makes sense to then change both `def scopes` and also `def variables` to be `Int`s as well?

kpbochenek: yes because right now it is: Int from LSP -> convert to Long -> modify other thing -> convert to Int -> put into LSP (we could simply avoid `convert` step)

ckipp01: Sounds good. I changed these as well.

ckipp01:
Unrelated test failure on windows
```
 [error] Failed: Total 955, Failed 1, Errors 0, Passed 950, Skipped 4
[error] Failed tests:
[error] 	tests.NewFilesLspSuite
[error] (unit / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

tgodzik:
> Unrelated test failure on windows
> 
> ```
>  [error] Failed: Total 955, Failed 1, Errors 0, Passed 950, Skipped 4
> [error] Failed tests:
> [error] 	tests.NewFilesLspSuite
> [error] (unit / Test / test) sbt.TestsFailedException: Tests unsuccessful
> ```

Looks like Windows specific flakiness, we can ignore it.
```
java.nio.file.AccessDeniedException: D:\a\metals\metals\tests\unit\target\e2e\new-files\new-case-class\a\src\main\scala\foo\Foo.scala
```

ckipp01:
> LGTM! It would be good if you could also try to check manually if debug/run still works before merging.

Yup, tested it locally to ensure that it was indeed working.

tgodzik:
> > LGTM! It would be good if you could also try to check manually if debug/run still works before merging.
> 
> Yup, tested it locally to ensure that it was indeed working.

Thanks!!

PR #1558, opened 2020-03-28
Add in the scala steward conf file

ckipp01:
This pr adds in a conf file for scalasteward. I recently added metals back in [here](https://github.com/scala-steward-org/repos/pull/253). I figured setting the frequency to weekly? to start out with was a good middle ground? What do you guys think? Also, is there any dependencies we want ignored by steward? You can see the configuration options [here](https://github.com/fthomas/scala-steward/blob/master/docs/repo-specific-configuration.md).

### DIFF
--- a/.scalasteward.conf
+++ b/.scalasteward.conf
@@ -1,0 +1,1 @@
+pullRequests.frequency = "0 0 1,15 * ?"


gabro:
Thanks for doing this! I think weekly is still a  bit much (also considering the release cadence), maybe biweekly?

ckipp01:
> Thanks for doing this! I think weekly is still a bit much (also considering the release cadence), maybe biweekly?

Sure, that makes sense. I've gone ahead and changed it from `@weekly` to `0 0 1,15 * *` which will just trigger on the 1st and 15th of the month.

.scalasteward.conf
```diff
@@ -0,0 +1 @@
+pullRequests.frequency = "0 0 1,15 * *"
```
fthomas: Unfortunately the library for cron expressions used by Scala Steward won't accept that string. If "day of month" is specified, "day of week" needs to be `?`:
```suggestion
pullRequests.frequency = "0 0 1,15 * ?"
```

ckipp01: Ahh, ok. Thanks! I switched it.

.scalasteward.conf
```diff
@@ -0,0 +1 @@
+pullRequests.frequency = "0 0 1,15 * ?"
```
olafurpg: I'm OK with monthly or bi-weekly if you think that is too infrequent.

ckipp01: Maybe let's try bi-weekly. If we get annoyed, we can change it to monthly.

PR #1559, opened 2020-03-28
Remove scalacheck

gabro:
Followup of https://github.com/scalameta/metals/pull/1546#discussion_r399216837

I was a bit too eager to add Scalacheck to test the new MUnit module, but it wasn't really necessary.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -445,10 +445,8 @@ lazy val unit = project
     testSettings,
     libraryDependencies ++= List(
       "io.get-coursier" %% "coursier" % V.coursier, // for jars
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -445,10 +445,8 @@ lazy val unit = project
     testSettings,
     libraryDependencies ++= List(
       "io.get-coursier" %% "coursier" % V.coursier, // for jars
-      "org.scalameta" %% "testkit" % V.scalameta,
```
gabro: I also took the liberty of removing this: it's not used anymore since we introduced MUnit

gabro:
Both CI errors look spurious

PR #1561, opened 2020-03-28
Remove deprecations

ckipp01:
A bit more house-keeping.

This pr removes a couple of the other deprecations from the lsp4j update recently. I also just went ahead and updated the other stuff that I noticed when adding the `-deprecation` flag. However, the two that I didn't update seem to be a bit more involved, so I left them as is.

The remaining deprecated warnings are:

1. The usage of `Thread.stop()` if various places
2. The usage of `PlatformTokenizerCache.megaCache.clear()`

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/worksheets/WorkspaceEditWorksheetPublisher.scala
+++ b/metals/src/main/scala/scala/meta/internal/worksheets/WorkspaceEditWorksheetPublisher.scala
@@ -3,18 +3,20 @@ package scala.meta.internal.worksheets
 import scala.meta.internal.metals.MetalsLanguageClient
 import mdoc.interfaces.EvaluatedWorksheet
 import scala.meta.io.AbsolutePath
-import org.eclipse.lsp4j.WorkspaceEdit
 import org.eclipse.lsp4j.ApplyWorkspaceEditParams
+import org.eclipse.lsp4j.Hover
+import org.eclipse.lsp4j.MarkupContent
+import org.eclipse.lsp4j.MarkupKind
+import org.eclipse.lsp4j.Position
+import org.eclipse.lsp4j.Range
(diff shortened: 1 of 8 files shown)


metals/src/main/scala/scala/meta/internal/worksheets/WorkspaceEditWorksheetPublisher.scala
```diff
@@ -44,11 +46,10 @@ class WorkspaceEditWorksheetPublisher(buffers: Buffers)
...
       message <- getHoverMessage(snapshotPosition, messages.hovers)
     } yield new Hover(
-      List(
-        JEither.forRight[String, MarkedString](
-          new MarkedString("scala", message)
-        )
-      ).asJava
+      new MarkupContent(
+        MarkupKind.MARKDOWN,
+        HoverMarkup(message)
```
kpbochenek: I don't remember where exactly but when you did similar change you wrapped whole body in ```scala block, shouldn't it be similar here?

ckipp01: I ended up making that happen in the apply method of `HoverMarkup`, so it is happening here :)

kpbochenek: perfect! 🥇

olafurpg:
> The usage of Thread.stop() if various places

I believe this method is a no-op on newer JDK versions. I'm wondering if we can just remove it 🤔 

> The usage of PlatformTokenizerCache.megaCache.clear()

That can be removed, it's no longer needed

ckipp01:
> > The usage of PlatformTokenizerCache.megaCache.clear()
> 
> That can be removed, it's no longer needed

Sounds good. I went ahead and removed these then.

PR #1568, opened 2020-03-30
Make sure we detect scala App trait when using hover. 

tgodzik:
Previously, when using hover over the Scala App trait no information would be returned. Now, it's properly displayed with all needed docs. This also works with a number of other cases in for example Play projects.

hover-App

Removing the condition altogether would cause anonymous classes to be shown like, so I just changed it to detect the specific scenario.

anon-hover


tgodzik:
Both errors are flaky. Created an issue for flaky debug tests, will try to gather some more information there.

PR #1572, opened 2020-03-31
Fix: don't show super lenses on synthetic methods e.g. toString

kpbochenek:
Forgot to include ` "scala/Any#"` in list of ignored parent classes

@tgodzik thanks for reporting!

### DIFF
--- a/tests/unit/src/test/scala/tests/SuperMethodLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/SuperMethodLspSuite.scala
@@ -48,7 +48,21 @@ class SuperMethodLspSuite extends BaseLspSuite("gotosupermethod") {
     checkSuperMethod(code)
   }
 
-  test("type inheritance") {
+  test("synthetic-methods") {
+    val code =
+      """
+        |package a
+        |case class A() {
+        |  override def <<1->0>>toString: String = "A"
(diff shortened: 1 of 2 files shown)


tests/unit/src/test/scala/tests/SuperMethodLspSuite.scala
```diff
@@ -70,7 +84,7 @@ class SuperMethodLspSuite extends BaseLspSuite("gotosupermethod") {
     checkSuperMethod(code)
   }
 
-  test("anonymousclass") {
+  test("anonymous class") {
```
olafurpg: nit: we try to avoid spaces in test names to make it easier to  run a single test from the command-line
```suggestion
  test("anonymous-class") {
```

kpbochenek: `testOnly tests.SuperMethodLspSuite -- --tests "*anonymous class*"` works just fine :)

kpbochenek: I looked at munit docs and with recommended commands there even with '-' it doesn't work :/

kpbochenek: anyway all test names changed in this file

PR #1576, opened 2020-04-01
Add in ... to signifigy truncation in worksheets

### LINKED ISSUE #1575: Add visual indicator to multi-line worksheet outputs that have been truncated
I can see full output by hovering over decorators but it would be nice to have a visual symbol indicating that output was truncated. Currently, there is no way to distinguish whether or not an output was truncated. 

Something like this would be helpful:
Current display: `for (num <- 1 to 10) println(num) // 1`
Potential Solution: `for (num <- 1 to 10) println(num) // 1...`

**Search terms**:
worksheet metals vscode

ckipp01:
This pr adds in a way to signify in worksheets that a lint is truncated or continued. For example instead of just ending, you'll now see that the line will end in a `...`. It's not added to the end to lengthen the line, but if what we are showing is not complete, we drop the last 3 chars and replace them.

Closes #1575

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/worksheets/MdocEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/worksheets/MdocEnrichments.scala
@@ -1,6 +1,7 @@
 package scala.meta.internal.worksheets
 
 import org.eclipse.{lsp4j => l}
+import mdoc.interfaces.EvaluatedWorksheetStatement
 import mdoc.{interfaces => i}
 
 object MdocEnrichments {
@@ -47,6 +48,17 @@ object MdocEnrichments {
         "mdoc"
       )
     }
+  }
+
+  /**
+   * Determines whether or not an evaluated worksheet statement summary is
+   * complete or not. If it is complete, it just returns the summary, otherwise
+   * it will chop off the last 3 chars and replace them with … to signify
+   * continuation
+   */
+  def truncatify(statement: EvaluatedWorksheetStatement): String = {
+    if (statement.isSummaryComplete()) statement.summary()
+    else statement.summary() + "…"
   }
 
 }
--- a/tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
+++ b/tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
@@ -77,7 +77,7 @@ abstract class BaseWorksheetLspSuite(scalaVersion: String)
            |import java.nio.file.Files
            |val name = "Susan" // "Susan"
            |val greeting = s"Hello $name" // "Hello Susan"
-           |println(greeting + "\nHow are you?") // Hello Susan
+           |println(greeting + "\nHow are you?") // Hello Susan…
            |1.to(10).toVector // Vector(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
            |val List(a, b) = List(42, 10) // a=42, b=10
            |""".stripMargin
@@ -108,7 +108,7 @@ abstract class BaseWorksheetLspSuite(scalaVersion: String)
            |import java.nio.file.Files
            |val name = "Susan" // "Susan"
            |val greeting = s"Hello $name" // "Hello Susan"
-           |println(greeting + "\nHow are you?") // Hello Susan
+           |println(greeting + "\nHow are you?") // Hello Susan…
            |1.to(10).toVector // Vector(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
            |val List(a, b) = List(42, 10) // a=42, b=10
            |""".stripMargin
--- a/metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
+++ b/metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
@@ -48,7 +48,7 @@ class DecorationWorksheetPublisher() extends WorksheetPublisher {
           ),
           ThemableDecorationInstanceRenderOptions(
             after = ThemableDecorationAttachmentRenderOptions(
-              commentHeader + s.summary(),
+              commentHeader + truncatify(s),
               color = "green",
               fontStyle = "italic"
             )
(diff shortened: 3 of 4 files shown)


tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
```diff
@@ -77,7 +77,7 @@ abstract class BaseWorksheetLspSuite(scalaVersion: String)
            |import java.nio.file.Files
            |val name = "Susan" // "Susan"
            |val greeting = s"Hello $name" // "Hello Susan"
-           |println(greeting + "\nHow are you?") // Hello Susan
+           |println(greeting + "\nHow are you?") // Hello Su...
```
ckipp01, line 80: So this is debatable whether or not this is good. On one side, I really like it. It shows that there is something more than you're seeing, causing the user to hover on it and indeed see that it goes 2 lines. Even though this is a short line, without it, the user might not do the hover. I'd be curious of your thoughts on it.

kpbochenek: what is an argument against `// Hello Susan...`? appending 3 chars to the first line without dropping anything?
Can lengthening a line causing any troubles here?

ckipp01: So in this exact situation, there is no trouble that can happen by appending it. However, the reason I went for that in general is so that when we hit the cutoff of how many line we display (for very long evaluations), then it won't make them _even_ longer. So the idea would be that even if it's cutoff, the `...` won't go past the limit that we've set.

However, by doing that, I didn't special case it for situations like the test case where we are manually breaking the line. Tbh, I've actually never personally did this in a worksheet, but have came across the long lines version of it many times. So I didn't even think about this situation until my test failed.

kpbochenek: aaaa ok we manually assume screenWidth  `settings.screenWidth` and we already have truncated line here during processing :D

2020-04-02-085304_1116x107_scrot

About those dots I thought we could in second case append `...` after end if it won't exceed `screenWidth` what do you think?

I think this could be small improvement but if it is not something quick to do just ignore it :+1:

metals/src/main/scala/scala/meta/internal/worksheets/MdocEnrichments.scala
```diff
@@ -49,4 +50,15 @@ object MdocEnrichments {
     }
   }
 
+  /**
+   * Determines whether or not an evaluated worksheet statement summary is
+   * complete or not. If it is complete, it just returs the summary, otherwise
```
kpbochenek, line 55: typo, 'returs -> returns'

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
ckipp01: Yea that does make sense. Let me see if I can change that. One thing I'm hitting right away is that is that I'm not actually seeing `screenWidth` make any difference. At first I thought that it was due to `coc-metals`, but even in intelliJ whether I have vs code small or full screen, it makes no difference and cuts at the same place. Am I missing something with that setting?

Screenshot 2020-04-02 at 10 21 31

For example in the image, is that suppose to be cutting it off before the end?

ckipp01: Ahhh yea, it looks like we don't pass in any screen information.

https://github.com/scalameta/metals/blob/37f4c238c0cc7d83f3362f7269cf63b2c7a4175d/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala#L270-L275

ckipp01: We could add a `withScreenWidth`, and `withScreenHeight` here.

kpbochenek: I don't think vscode uses any truncation mechanisms here, all truncation is done in mdoc according to this:
https://github.com/scalameta/mdoc/blob/master/mdoc/src/main/scala/mdoc/internal/cli/Settings.scala#L136

`screenWidth` I think is a constant and it is not real width of vscode/vim screen.

check this method in mdoc:
`private def renderDecoration(statement: Statement): EvaluatedWorksheetStatement = {`

maybe this truncation should actually happend in mdoc? :question:

ckipp01: @olafurpg made a good point about how clients may display truncation differently, so it's probably best for clients to handle it, which makes sense. 

Looking more into this, our test never hit this in the past because before when doing worksheets in vscode, we never added `...` like we did in the editors that didn't support virtual text. So if we simply add that, those tests will fail, since the only check we are doing is to see if the evaluation is complete. I also think this is the only check we can sensibly do as well without getting a whole bunch more information.

That said, the more I think about it, the more I think this is probably valid behavior.

kpbochenek: Yeah you are right truncation should not be done in `mdoc` 
User will see an output is truncated and that's what is important.

As my after thought we could also do:
adding `margin: Int` to `EvaluatedWorksheetStatement` and in `truncatify`:
```scala
      if (statement.summary().length +3 < statement.margin) statement.summary() + "..."
      else statement.summary().dropRight(3) + "..."
```
But this is a change in mdoc and it makes it more complicated than it should be.

metals/src/main/scala/scala/meta/internal/worksheets/MdocEnrichments.scala
```diff
@@ -49,4 +50,15 @@ object MdocEnrichments {
...
 
+  /**
+   * Determines whether or not an evaluated worksheet statement summary is
+   * complete or not. If it is complete, it just returns the summary, otherwise
+   * it will chop off the last 3 chars and replace them with ... to signify
+   * continuation
+   */
+  def truncatify(statement: EvaluatedWorksheetStatement): String = {
+    if (statement.isSummaryComplete()) statement.summary()
+    else statement.summary().dropRight(3) + "..."
```
olafurpg, line 61: What about using the horizontal ellipsis unicode character? I think we can also skip `dropRight()` in that case so that the full summary is preferved from MUnit
```suggestion
    else statement.summary().dropRight(3) + "…"
```

ckipp01: Ahh, that's a great idea.

ckipp01:
> LGTM 👍 Another related change I have wanted to implement is to use a unicode character instead of `//` to indicate the start of code lenses

That would be great. I already do this in `coc-metals`, and I replace the `//` with `‣`.

PR #1582, opened 2020-04-02
Add properties for bloop zipkin tracing to fastpass

wiwa:
This change adds the properties introduced in https://github.com/scalacenter/bloop/pull/1231 to fastpass.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/pantsbuild/commands/BloopGlobalSettings.scala
+++ b/metals/src/main/scala/scala/meta/internal/pantsbuild/commands/BloopGlobalSettings.scala
@@ -5,7 +5,7 @@ import scala.util.control.NonFatal
 import ujson.Obj
 import java.nio.file.Paths
 import java.nio.file.Path
-import scala.meta.internal.zipkin.ZipkinUrls
+import scala.meta.internal.zipkin.ZipkinProperties
 import ujson.Str
 
 object BloopGlobalSettings {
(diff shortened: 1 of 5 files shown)


metals/src/main/scala/scala/meta/internal/pantsbuild/commands/BloopZipkinTraceProperties.scala
```diff
@@ -0,0 +1,18 @@
+package scala.meta.internal.pantsbuild.commands
+
+import scala.meta.internal.pantsbuild.commands.BloopGlobalSettings.Property
+
+object BloopZipkinTraceProperties {
```
olafurpg, line 5: Can we move all these to `ZipkinUrls`? Note that it lives in the `internal.zipkin` package to make it easier to configure for native-image

metals/src/main/scala/scala/meta/internal/pantsbuild/commands/BloopGlobalSettings.scala
```diff
@@ -49,4 +63,29 @@ object BloopGlobalSettings {
     }
 
   }
+
+  case class Property(metalsProperty: String) {
```
olafurpg, line 67: Can we move this to the `zipkin` package and make it a toplevel class? Might be good to also move the `optionValue` and `updateOption` methods to belong to `Property` somehow

PR #1584, opened 2020-04-04
Add VersionedDocsModifier

gabro:
This PR adds a custom mdoc modifier that inserts a warning about unreleased features in the documentation.

The idea is that we can freely merge documentation for unreleased features, and the warning will be displayed until we release a stable version.

## Demo

Assuming the latest stable is 0.8.3, the demo shows how the warning is displayed when using `mdoc:since:0.8.4` but it's not displayed when using `mdoc:since:0.8.2`

### DIFF
--- a/metals-docs/src/main/scala/docs/VersionedDocsModifier.scala
+++ b/metals-docs/src/main/scala/docs/VersionedDocsModifier.scala
@@ -1,0 +1,25 @@
+package docs
+
+import mdoc.Reporter
+import mdoc.StringModifier
+import scala.meta.inputs.Input
+import scala.meta.internal.metals.BuildInfo
+import scala.meta.internal.semver.SemVer.Version
+
+class VersionedDocsModifier extends StringModifier {
+  val name = "since"
+
+  override def process(
+      version: String,
+      code: Input,
+      reporter: Reporter
+  ): String = {
+    val unreleased =
+      Version.fromString(version) >= Version.fromString(BuildInfo.metalsVersion)
+
+    if (unreleased) {
+      s"> 🚧 This documentation section describes a feature that is currently unreleased. The expected release version is $version\n"
+    } else ""
+
+  }
+}
--- a/metals-docs/src/main/scala/docs/VersionedDocsModifier.scala
+++ b/metals-docs/src/main/scala/docs/VersionedDocsModifier.scala
@@ -19,7 +19,9 @@ class VersionedDocsModifier extends StringModifier {
 
     if (unreleased) {
       s"> 🚧 This documentation section describes a feature that is currently unreleased. The expected release version is $version\n"
-    } else ""
+    } else {
+      ""
+    }
 
   }
 }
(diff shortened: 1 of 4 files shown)


metals-docs/src/main/scala/docs/VersionedDocsModifier.scala
```diff
@@ -0,0 +1,25 @@
...
+      version: String,
+      code: Input,
+      reporter: Reporter
+  ): String = {
+    val unreleased =
+      Version.fromString(version) >= Version.fromString(BuildInfo.metalsVersion)
+
+    if (unreleased) {
+      s"> 🚧 This documentation section describes a feature that is currently unreleased. The expected release version is $version\n"
+    } else ""
```
gabro, line 22: This is meant to make sure we don't accidentally leave the warning visible when the corresponding version has been published.

Of course we could just remove the annotation instead, this is a safeguard.

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
```diff
@@ -2,18 +2,26 @@ package scala.meta.internal.semver
...
+      val Array(major, minor, patch) =
+        version.replaceAll("(-|\\+).+$", "").split('.').map(_.toInt)
+
+      Version(major, minor, patch)
     }
   }
+
+  def isCompatibleVersion(minimumVersion: String, version: String): Boolean = {
+    Version.fromString(version) >= Version.fromString(minimumVersion)
+  }
```
gabro, line 26: I refactored this to support a slightly more general use case.

We still compare versions only by x.y.z, ignoring any other following modifier. This has some edge cases such as `1.0.0+1` being equal to `1.0.0-SNAPSHOT` and `1.0.0+56-abcdefg`, but it still covers all our uses cases for now

website/i18n/en.json
```diff
@@ -53,7 +56,7 @@
         "sidebar_label": "Debug Adapter Protocol"
       },
       "editors/decoration-protocol": {
-        "title": "Decoration Protocol v0.1.0",
+        "title": "Decoration Protocol v0.2.0",
```
ckipp01, line 59: 😬 thanks for this!

metals-docs/src/main/scala/docs/VersionedDocsModifier.scala
olafurpg: Alternatively, we can keep these and change the message into something like
```suggestion
    } else {
      "**Available since**: v0.8.4, released April 2020.\n"
    }
```
We can probably drill out the release date from git. Leaving it empty is also fine 👍

olafurpg: Nit, I try to keep if/else branches consistent wrt to braces
```suggestion
    } else {
      ""
    }
```

gabro:
@olafurpg for now I think it's fine to leave it blank (I'm not sure what's the advantage of marking the release version of features in the docs).

That said, if we leave the modifier in place, we can always change our mind and update it a posteriori

PR #1592, opened 2020-04-08
Create a new file next to an existing one if user selected a file 

tgodzik:
This will allow us to have the `New Scala File` option also when right clicking on a file.

new-file


tgodzik:
I also added a test just in case.

Issue #1464, opened 2020-02-24
Local type aliases not detected in goto implementation


tgodzik:
**Describe the bug**

ImplementationLspSuite.scala
```
check(
    "type-local",
    """|/a/src/main/scala/a/Main.scala
       |object Test {
       |  def main {
       |    class @@A
       |    class <<B>> extends A
       |    type C = A
       |    class <<D>> extends C
       |  }
       |}
       |""".stripMargin
  )
```

```
=> Diff (- obtained, + expected)
     type C = A
-    class D extends C
+    class <<D>> extends C
   }
```

**Search terms**
implementation search type alias

ckipp01:
Just so someone else doesn't grab this, @eilite has been looking at this, and will shoot in a pr.

tgodzik:
Closed via https://github.com/scalameta/metals/pull/1598

PR #1598, opened 2020-04-10
goto Implementations with local alias types

eilite:
This PR addresses https://github.com/scalameta/metals/issues/1464

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
@@ -338,17 +338,29 @@ final class ImplementationProvider(
       file: Path
   ): Map[Path, Set[ClassLocation]] = {
 
+    def isTypeAlias(symbol: String) =
+      classContext
+        .findSymbol(symbol)
+        .exists(_.kind == SymbolInformation.Kind.TYPE)
+
     def loop(symbol: String, currentPath: Option[Path]): Set[ClassLocation] = {
       val directImplementations = classContext.getLocations(symbol).filterNot {
         loc =>
-          // we are not interested in local symbols from outside the workspace
-          (loc.symbol.isLocal && loc.file.isEmpty) ||
-          // local symbols inheritance should only be picked up in the same file
-          (loc.symbol.isLocal && loc.file != currentPath)
+          // we keep type aliases
+          !isTypeAlias(loc.symbol) &&
+          (
+            // we are not interested in local symbols from outside the workspace
+            (loc.symbol.isLocal && loc.file.isEmpty) ||
+            // local symbols inheritance should only be picked up in the same file
+            (loc.symbol.isLocal && loc.file != currentPath)
+          )
       }
       directImplementations ++ directImplementations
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -338,17 +338,29 @@ final class ImplementationProvider(
...
+            // local symbols inheritance should only be picked up in the same file
+            (loc.symbol.isLocal && loc.file != currentPath)
+          )
       }
       directImplementations ++ directImplementations
         .flatMap { loc =>
-          val allPossible = loop(loc.symbol, loc.file)
+          val allPossible = loop(
+            loc.symbol,
+            loc.file.orElse(Some(file).filter(_ => isTypeAlias(loc.symbol)))
```
tgodzik, line 362: It seems we are missing the file due to [here](https://github.com/scalameta/metals/blob/39fa3cc3acbed1b7e252747e520ce42ad542260f/metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala#L490)

I think this was done in order not to show types as implementations. We could add the file the same way we add it in the method above and move checking for type alias in [here](https://github.com/scalameta/metals/blob/39fa3cc3acbed1b7e252747e520ce42ad542260f/metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala#L323)

We should use the `implDocument` to search for the `SymbolInformation` and check if it is a type alias.

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -457,6 +468,12 @@ object ImplementationProvider {
       .find(sym => sym.symbol == symbol)
   }
 
+  def isTypeAlias(
+      symbol: String,
+      findSymbol: String => Option[SymbolInformation]
```
tgodzik, line 473: We could use the already existing `implDocument` and have it passed here. Then:
`ImplementationProvider.findSymbol(symbol, implDocument).exists(_.kind == SymbolInformation.Kind.TYPE)`

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -320,7 +320,18 @@ final class ImplementationProvider(
         classContext,
         implReal
       )
-      implOccurrence <- findDefOccurrence(implDocument, implSymbol, source)
+      symbolSearch = defaultSymbolSearch(
+        implPath,
+        implDocument
+      )
+      filteredImplSymbol <- Some(implSymbol).filterNot(
```
tgodzik, line 327: We could just do `if ! isTypeAlias(implSymbol, implDocument) `

eilite: Oh yeah of course !

tgodzik:
Just noticed, that one symbol was unused, so I removed it and I think we are ready to merge.

PR #1606, opened 2020-04-12
Properly set the arguments field in sever commands documentation

ayoub-benali:
For several commands the `arguments` field was left with its default value `null`
while the description mentioned arguments to be sent as seen bellow:

image

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -162,12 +162,10 @@ object ServerCommands {
        |
        |Note: document in json argument must be absolute path.
        |""".stripMargin,
-    """|```json
-       |{
+    """|{
        |  document: "file:///home/dev/foo/Bar.scala",
        |  position: {line: 5, character: 12}
        |}
-       |```
        |""".stripMargin
   )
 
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -161,9 +161,8 @@ object ServerCommands {
        |If symbol under cursor is invalid or does not override anything then command is ignored.
        |
        |Note: document in json argument must be absolute path.
-       |
-       |Arguments:
-       |```json
+       |""".stripMargin,
+    """|```json
```
ckipp01, line 165: So it looks like this one is still a bit screwy with this change.

Screenshot 2020-04-13 at 10 49 55

ayoub-benali: probably because it is a multi line string. 
What is the sbt command to generate the doc ?

ckipp01: You can find instructions on how to generate the docs and view them here https://scalameta.org/metals/docs/contributors/updating-website.html

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -182,26 +179,22 @@ object ServerCommands {
        |QuickPick will show up only if more than one result is found.
        |
        |Note: document in json argument must be absolute path.
-       |
-       |Arguments:
-       |```json
-       |{
+       |""".stripMargin,
+    """|{
```
ckipp01, line 183: I still think we need to figure out a way for this to be a code block potentially. I find the arguments here pretty hard to read.

Screenshot 2020-04-16 at 14 56 00

ckipp01:
Unrelated DAP failure.

ayoub-benali:
Thanks for the fix

PR #1607, opened 2020-04-12
Don't ask .scalafmt.conf creation again

takezoe:
When "Format on Compile" or "Format on Save" is enabled in the editor-wide settings, Metals asks `.scalafmt.conf` creation every time. It would be nice if the dialog has a volatile "Don't ask again" option for `.scalafmt.conf` creation.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
@@ -22,6 +22,7 @@ import scala.meta.internal.metals.Messages.MissingScalafmtVersion
 import scala.meta.internal.metals.MetalsEnrichments._
 import java.io.OutputStream
 import java.io.OutputStreamWriter
+import java.util.concurrent.TimeUnit
 
 /**
  * Implement text formatting using Scalafmt
@@ -34,7 +35,8 @@ final class FormattingProvider(
     clientConfig: ClientConfiguration,
     statusBar: StatusBar,
     icons: Icons,
-    workspaceFolders: List[AbsolutePath]
+    workspaceFolders: List[AbsolutePath],
+    tables: Tables
 )(implicit ec: ExecutionContext)
     extends Cancelable {
   override def cancel(): Unit = {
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
```diff
@@ -158,21 +158,28 @@ final class FormattingProvider(
     }
   }
 
+  private var dontAskFileCreation: Boolean = false
```
olafurpg, line 161: This var won't persist if the Metals server restarts. Normally we implement "Don't show again" by persisting the user's request in a sql database. Example: https://github.com/scalameta/metals/blob/e2320b5b4d7f365c6064ccc3ecb2a105b579a109/metals/src/main/scala/scala/meta/internal/metals/DismissedNotifications.scala

olafurpg:
FWIW, I personally enable "format on save" per workspace and have it disabled globally. It might be good to update the website to explain what the recommended workflow is

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
takezoe: That's intentional behavior. If we persist it, we have to provide a way to re-enable it too. So, I thought asking once in each session should be better. I didn't know that we can configure formatting behavior per workspace.

olafurpg: It can be unexpected for the user to select "Don't show again", restart VS Code and then get the .scalafmt.conf prompt again.

I agree it would be nice to have a way to re-enable prompts where the user selected "Don't show again", the only way to achieve that right now is to `rm -rf .metals`. This functionality be implemented separately from this PR and it would then work for all "Don't show again" prompts.

takezoe: How about asking "Not now" instead?

olafurpg: We also use the SQL database to handle "Not now" responses to dismiss this prompt for a X number of minutes. For example, here we handle "Not now" for the "Import build" prompt https://github.com/scalameta/metals/blob/abca8703638fe2768e0058288ae0e3decbe42e07/metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala#L163

> Being asked every time when I save files even after selecting "Not now" is annoying. I just want to avoid it.

I agree this is annoying and I'm open to merge this feature but I'm not convinced it's a reliable fix to store this state in the Metals server process because that means the user gets the prompt again when they restart Metals. The reliable fix for this problem is to use the SQL table, which is consistent with how we handle other prompts

olafurpg:
@takezoe Have you tried to override the "format on save" option only for Scala like this?
```json
❯ cat .vscode/settings.json
{
  "[scala]": {
    "editor.formatOnSave": false
  }
}
```

takezoe:
I want to use scalafmt in some projects, but don't want in some projects. For example, most of projects in our company are configured to use scalafmt, but some projects aren't.

takezoe:
Updated to use `DismissedNotifications` to store the dismiss state.

takezoe:
@olafurpg Could you review this again?

olafurpg:
We can merge this after the git conflicts have been resolved. It was unlucky timing to hit on a conflict, `FormattingProvider` has been pretty stable for a while

takezoe:
Thanks for your comment! I rebased to the latest master.

takezoe:
Thanks for your comment! I squashed all commits in this pull request to a single commit and rebased to the latest master.

takezoe:
Thanks for your comment! I squashed all commits in this pull request to a single commit and rebased to the latest master.

takezoe:
Thanks for your comment! I squashed all commits in this pull request to a single commit and rebased to the latest master.

takezoe:
Thanks for your comment! I squashed all commit in this pull request to a single commit and rebased to the latest master.

PR #1608, opened 2020-04-13
(WIP) Go to Type Definition

alekseiAlefirov:
This PR is an attempt to finish what's started here: https://github.com/scalameta/metals/pull/978

Note:
- `TypeDefinitionLspSuite` is good place to review and discuss, what the feature should do and what should not

TODO:
- fix `todo`s with pprint -- seams like unreachable code, but maybe it's reachable under some scala versions.


tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -1011,17 +1011,56 @@ final class TestingServer(
...
+        .asJava
+        .get()
+    }, markNoSymbol = false)
+  }
+
+  private def toDefinitionsTextDocument(
+      path: AbsolutePath,
+      getDefinition: TextDocumentPositionParams => DefinitionResult,
+      //in case no resolved symbol gotten
+      markNoSymbol: Boolean = true
```
alekseiAlefirov, line 1055: While I wanted to reuse, what `DefinitionLspSuite` does, I was not sure about it testing not actual endpoint method `definition`, which returns a list of locations, but some inner method, which returns a bit richer `DefinitionResult`, to use resolved `symbol` for the tests sake - this a bit breaks the `integration` in testing. So, without the `symbol` I had to tinker this method a bit.

tgodzik:
I am going to close this since it's not really highly anticipated feature and I will not have time to properly review and maybe fix it. There is a great number of other issues we need to work on and I feel like this is too low priority to work on.

At first, I though it would be an easy beginners feature, but och how wrong was I :cry: This is way more complex than I imagined at first and it was my fault for proposing that we work on it.

We can go back to it if there is enough interest. Let me know if you feel like I am making a wrong decision here.

gabro:
I think it's a fair decision, it's a highly complicated feature to implement (it seems) and not a super requested feature.

By the way, is there even a feature request for it?

tgodzik:
> I think it's a fair decision, it's a highly complicated feature to implement (it seems) and not a super requested feature.
> 
> By the way, is there even a feature request for it?

I moved the issue back : https://github.com/scalameta/metals-feature-requests/issues/153

ckipp01:
> I am going to close this since it's not really highly anticipated feature and I will not have time to properly review and maybe fix it. There is a great number of other issues we need to work on and I feel like this is too low priority to work on.
> 
> At first, I though it would be an easy beginners feature, but och how wrong was I 😢 This is way more complex than I imagined at first and it was my fault for proposing that we work on it.

So looking through this a bit and seeing that there are already tests etc, how much work do you envision being left here?

> We can go back to it if there is enough interest. Let me know if you feel like I am making a wrong decision here.

No, I think it's the right choice, especially seeing some of the tickets about general stability popping up. 👍

tgodzik:
> So looking through this a bit and seeing that there are already tests etc, how much work do you envision being left here?

It's hard to say, I haven't had the time to properly look through it. We would need someone to take ownership of the code to push it forward for sure.

PR #1611, opened 2020-04-14
Handle return to line in strings in OnTypeFormatting

mlachkar:
* Indentation is not working correctly (but the test doesn't show the issue :(

metals-1586873460408

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
@@ -53,10 +53,9 @@ object MultilineStringFormattingProvider {
   }
 
   private def indent(
-      sourceText: String,
+      splitLines: Array[String],
       position: Position
   ): TextEdit = {
-    val splitLines = sourceText.split('\n')
     // position line -1 since we are checking the line before when doing onType
     val defaultIndent = determineDefaultIndent(splitLines, position.getLine - 1)
     val existingSpaces = position.getCharacter()
@@ -112,40 +111,49 @@ object MultilineStringFormattingProvider {
 
   private def multilineStringInTokens(
       tokens: Tokens,
-      startPos: meta.Position,
-      endPos: meta.Position,
+      startPos: StartPosition,
+      endPos: EndPosition,
       sourceText: String,
       newlineAdded: Boolean
   ): Boolean = {
     if (pipeInScope(startPos, endPos, sourceText, newlineAdded)) {
-      def shouldFormatMultiString: PartialFunction[(Token, Int), Boolean] = {
-        case (token: Constant.String, index: Int) =>
-          isMultilineString(sourceText, token) && hasStripMarginSuffix(
-            index,
-            tokens
-          ) && inToken(startPos, endPos, token)
-      }
-
-      def shouldFormatInterpolationString
-          : PartialFunction[(Token, Int), Boolean] = {
-        case (token: Interpolation.Start, index: Int)
-            if token.start < startPos.start => {
-          var endIndex = index + 1
-          while (!tokens(endIndex)
-              .isInstanceOf[Interpolation.End]) endIndex += 1
-          isMultilineString(sourceText, token) && hasStripMarginSuffix(
-            endIndex,
-            tokens
-          )
-        }
-      }
       tokens.zipWithIndex
         .exists(
-          shouldFormatMultiString orElse shouldFormatInterpolationString orElse {
+          shouldFormatMultiString(sourceText, tokens, startPos, endPos) orElse
+            shouldFormatInterpolationString(sourceText, tokens, startPos) orElse {
             case _ => false
           }
         )
     } else false
+  }
+
+  private def shouldFormatMultiString(
+      sourceText: String,
+      tokens: Tokens,
+      start: StartPosition,
+      end: EndPosition
+  ): PartialFunction[(Token, Int), Boolean] = {
+    case (token: Constant.String, index: Int) =>
+      isMultilineString(sourceText, token) && hasStripMarginSuffix(
+        index,
+        tokens
+      ) && inToken(start, end, token)
+  }
+  private def shouldFormatInterpolationString(
+      sourceText: String,
+      tokens: Tokens,
+      start: StartPosition
+  ): PartialFunction[(Token, Int), Boolean] = {
+    case (token: Interpolation.Start, index: Int)
+        if token.start < start.start => {
+      var endIndex = index + 1
+      while (!tokens(endIndex)
+          .isInstanceOf[Interpolation.End]) endIndex += 1
+      isMultilineString(sourceText, token) && hasStripMarginSuffix(
+        endIndex,
+        tokens
+      )
+    }
   }
 
   private def withToken(
@@ -170,23 +178,34 @@ object MultilineStringFormattingProvider {
   }
 
   private def doubleQuoteNotClosed(
-      lineBefore: String
-  ): Boolean = lineBefore.count(_ == '"') % 2 != 0
+      splitLines: Array[String],
+      position: Position
+  ): Boolean = {
+    val lineBefore = splitLines(position.getLine - 1)
+    lineBefore.count(_ == '"') % 2 != 0
+  }
 
   private def fixStringNewline(
-      newlineAdded: Boolean,
       position: Position,
-      lineBefore: String
+      splitLines: Array[String]
   ): List[TextEdit] = {
     val precedentLine = position.getLine - 1
-    val lastChar = lineBefore.size
-    val endPrecedentLine = new Position(precedentLine, lastChar)
+    val lastCharOfLineBefore = splitLines(precedentLine).size
+    val endPrecedentLine = new Position(precedentLine, lastCharOfLineBefore)
     val textEditPrecedentLine = new TextEdit(
       new Range(endPrecedentLine, endPrecedentLine),
       "\"" + " " + "+"
     )
+    val isPrecedentLineFirstLineOfString =
+      if (precedentLine > 1)
+        !splitLines(precedentLine - 1).trim.lastOption.contains('+')
+      else true
+    val defaultIndent = splitLines(precedentLine).takeWhile(_ == ' ').size
+    val indent =
+      if (isPrecedentLineFirstLineOfString) defaultIndent + 2 else defaultIndent
+    val zeroPos = new Position(position.getLine, 0)
     val textEditcurrentLine =
-      new TextEdit(new Range(position, position), "  " + "\"")
+      new TextEdit(new Range(zeroPos, position), " " * indent + "\"")
     List(textEditPrecedentLine, textEditcurrentLine)
   }
 
@@ -235,7 +254,8 @@ object MultilineStringFormattingProvider {
     val range = new Range(params.getPosition, params.getPosition)
     val doc = params.getTextDocument()
     val newlineAdded = params.getCh() == "\n"
-
+    val splitLines = sourceText.split('\n')
+    val position = params.getPosition
     withToken(doc, sourceText, range) {
       (startPos, endPos, sourceText, tokens) =>
         tokens match {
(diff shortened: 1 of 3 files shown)


tests/unit/src/test/scala/tests/OnTypeFormattingSuite.scala
```diff
@@ -168,7 +168,7 @@ class OnTypeFormattingSuite extends BaseLspSuite("onTypeFormatting") {
        |  | a multiline
        |  | string
        |  '''.stripMargin
-       |
+       |
```
ckipp01: Ha, I have the same thing and my vim setup removes empty spaces at the end of a line with I format. This pr that @gabro did is a clever way to get around it and it also has the benefit of being more explicit that we want an `indent` there. You can probably utilize the same idea.

metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -114,56 +114,86 @@ final class MultilineStringFormattingProvider(
...
+      val startPos = range.getStart.toMeta(virtualFile)
+      val endPos = range.getEnd.toMeta(virtualFile)
+      Future { fn(startPos, endPos, sourceText, virtualFile.tokenize.toOption) }
+    } else Future.successful(Nil)
+  }
+  private def doubleQuoteNotClosed(
+      sourceText: String,
+      position: Position
+  ): Boolean = {
+    val splitLines = sourceText.split('\n')
```
tgodzik, line 172: Could we maybe operate or indexes here?
Find the right line and then check `"` until the next `\n`?

It might be more efficient to not use split on the whole file. I am worried about the performance with large files.

Or alternatively, we could create a method `findLine` that would just do the substring on the needed line. What do you think?

metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -114,56 +114,86 @@ final class MultilineStringFormattingProvider(
...
+  }
 
+  private def tryToIndent(
+      sourceText: String,
+      newlineAdded: Boolean,
+      position: Position
+  ): List[TextEdit] = {
+    if (newlineAdded && doubleQuoteNotClosed(sourceText, position)) {
+      val precedentLine = position.getLine - 1
+      val lines = sourceText.split('\n')
```
tgodzik, line 184: We are splitting both here and in `doubleQuoteNotClosed` 

Maybe we can calculate `lineBefore` only once?

metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -114,56 +114,86 @@ final class MultilineStringFormattingProvider(
...
+  private def doubleQuoteNotClosed(
+      sourceText: String,
+      position: Position
+  ): Boolean = {
+    val splitLines = sourceText.split('\n')
+    val lineBefore = splitLines(position.getLine - 1)
+    lineBefore.count(_ == '"') % 2 != 0
+  }
 
+  private def tryToIndent(
```
tgodzik, line 177: `fixStringNewline` ?

metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -114,56 +114,86 @@ final class MultilineStringFormattingProvider(
...
-          val tokens =
-            virtualFile.tokenize.toOption
-          tokens
-            .filter(multilineStringInTokens(_, startPos, endPos, sourceText))
-            .map(_ => fn(sourceText, startPos))
-            .getOrElse(Nil)
-        } else Nil
-      } else Nil
-    }
+  private def withTokenOpt(textId: TextDocumentIdentifier, range: Range)(
```
tgodzik, line 151: I don't think we need to change the name to `withTokenOpt` since we don't have any other method like that.

metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
mlachkar: ok. I named it like that because I imagined inside tryIndent, if String, we do that, if you type """ , do that ....

metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
mlachkar: ok. maybe the class MultilineStringFormattingProvider, could have the lines of the current file.. so the split and all if these operations are only done once?

tgodzik: Could be also, though most likely `substring(startOfLine, endOfLine)` would be fastest.

mlachkar:
Result on VSCode has the correct indentation, but the test has incorrect indentation.

metals-1587683817869

ckipp01:
> Result on VSCode has the correct indentation, but the test has incorrect indentation.

I'm not 100% sure that this is the case, but one thing to consider if you keep running into this is the [magic indentation](https://github.com/Microsoft/language-server-protocol/issues/83) that VS Code (and some others) add on multi-line text edits. I ran into this when I did the "Add all members" completion and it drove me nuts because it worked fine in the editor, but the spacing in the tests were off. I lead me to to have to check the editor to determine the correct spacing.

https://github.com/scalameta/metals/blob/1a75a6e6b01840ec3c6896137d9ec1ffc625f6f7/mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala#L76-L83

Again, not sure that this is what is happening here, but just keep it in mind.

tgodzik:
The tests seem to be failing on windows, could you take a look:
```
[error] ==> X tests.OnTypeFormattingSuite.string-tree-lines  4.354s munit.FailException: D:\a\metals\metals\tests\unit\src\test\scala\tests\OnTypeFormattingSuite.scala:277
[error] 276:  // in VSCode it behaves normally and both "test" and "2" are aligned
[error] 277:  check(
[error] 278:    "string-tree-lines",
[error] diff assertion failed
[error] => Obtained
[error]     """|object Main {
[error]        |  val str = "test1" +
[error]        |    "test" +
[error]        |    "2"
[error]        |}
[error]        |""".stripMargin
[error]      "test" +
[error] -    "2"
No more data in the client stdin, exiting...
[error] +  "2"
No more data in the server stdin, exiting...
[error]  }
[info] Test run tests.OnTypeFormattingSuite finished: 1 failed, 0 ignored, 16 total, 74.015s
```

mlachkar:
Yes, I noticed that. On Windows the test does what it should do. ( I will check what I can do ). Thanks

mlachkar:
the failing test is tests.DebugProtocolSuite.test-unresolved-params because of a timeout of 10 min.

mtags/src/main/scala-2/scala/meta/internal/metals/Test.sc
```diff
@@ -0,0 +1,2 @@
+val string = "5454  ".trim
```
tgodzik, line 1: I think this is not needed?

mlachkar: Indeed, no. thanks

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -109,57 +111,104 @@ object MultilineStringFormattingProvider {
...
+    val endPrecedentLine = new Position(precedentLine, lastCharOfLineBefore)
+    val textEditPrecedentLine = new TextEdit(
+      new Range(endPrecedentLine, endPrecedentLine),
+      "\"" + " " + "+"
+    )
+    val isPrecedentLineFirstLineOfString =
+      if (precedentLine > 1)
+        !splitLines(precedentLine - 1).trim.lastOption.contains('+')
+      else true
+    val defaultIndent = splitLines(precedentLine).takeWhile(_ == ' ').size
```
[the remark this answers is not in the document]
mlachkar: I didn't know this method, thanks

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -109,57 +130,111 @@ object MultilineStringFormattingProvider {
...
+      fn(startPos, endPos, sourceText, virtualFile.tokenize.toOption)
     } else Nil
   }
 
+  private def doubleQuoteNotClosed(
+      splitLines: Array[String],
+      position: Position
+  ): Boolean = {
+    val lineBefore = splitLines(position.getLine - 1)
+    !getIndexOfLastQuote(lineBefore).exists(_._2)
```
tgodzik, line 204: MAybe we can make it a bit more clear:
`getIndexOfLastQuote(lineBefore).exists{ case (_, quoteClosed) => !quoteClosed }`

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -109,57 +130,111 @@ object MultilineStringFormattingProvider {
...
+      "\"" + " " + "+"
+    )
+    val defaultIndent = previousLine.prefixLength(_ == ' ')
+    val indent =
+      if (previousLineNumber > 1 && !splitLines(previousLineNumber - 1).trim.lastOption
+          .contains('+'))
+        defaultIndent + 2
+      else defaultIndent
+    val interpolationString = {
+      val quoteIndex = getIndexOfLastQuote(previousLine).map(_._1).getOrElse(-1)
```
tgodzik, line 226: Maybe this can be a bit more clear also?
```
 getIndexOfLastQuote(previousLine).map{
  case (lastQuoteIndex) => 
    if (quoteIndex >0 && previousLine(quoteIndex - 1) == 's') "s" else ""
}.getOrElse("")
```

tests/unit/src/test/scala/tests/OnTypeFormattingSuite.scala
```diff
@@ -129,8 +130,38 @@ class OnTypeFormattingSuite extends BaseLspSuite("onTypeFormatting") {
...
        |object Main {
        |  val number = 102
-       |  val str = s"|
-       |  $$number".stripMargin
+       |  val str = s"|" +
+       |    s"$$number".stripMargin
+       |}""".stripMargin
+  )
+
+  check(
```
tgodzik, line 138: Let's add a couple of test cases with escaped quotes and other symbols to make sure the new algorithm is working.

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -49,11 +52,29 @@ object MultilineStringFormattingProvider {
...
+      val char = line(i)
+      if (char == '"') {
+        if (!escaped) {
+          lastQuote = i
+          quoteClosed = !quoteClosed
+        } else {
+          escaped = !escaped
+        }
+      }
+      if (char == '\\') escaped = !escaped
```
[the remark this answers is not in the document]
mlachkar: ok

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
mlachkar: ok

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
mlachkar: ok

tests/unit/src/test/scala/tests/OnTypeFormattingSuite.scala
mlachkar: ok

PR #1624, opened 2020-04-15
Fix NewFileLspSuite on Windows

tgodzik:
Getting weird access denied exceptions. Might be a race condition between folder being opened and used.
```
INFO  Executing command [new-scala-file]
	at sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:85)
	at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
[error] ==> X tests.NewFilesLspSuite.new-trait-new-dir  2.929s java.nio.file.AccessDeniedException: C:\Users\RUNNER~1\AppData\Local\Temp\metals14694223127987732815Baz.scala -> D:\a\metals\metals\tests\unit\target\e2e\new-files\new-trait-new-dir\a\src\main\scala\bar\Baz.scala
	at sun.nio.fs.WindowsFileSystemProvider.newByteChannel(WindowsFileSystemProvider.java:235)
[info] tests.NewFilesLspSuite.new-package-object started
	at java.nio.file.Files.newByteChannel(Files.java:370)
	at java.nio.file.Files.newByteChannel(Files.java:421)
	at java.nio.file.Files.readAllBytes(Files.java:3205)
	at scala.meta.internal.io.PlatformFileIO$.slurp(PlatformFileIO.scala:45)
	at scala.meta.internal.io.FileIO$.slurp(FileIO.scala:24)
	at scala.meta.internal.mtags.MtagsEnrichments$XtensionAbsolutePathMetals.toInput(MtagsEnrichments.scala:168)
	at scala.meta.internal.metals.Docstrings.expireSymbolDefinition(Docstrings.scala:58)
	at scala.meta.internal.metals.MetalsLanguageServer.indexSourceFile(MetalsLanguageServer.scala:1625)
	at scala.meta.internal.metals.MetalsLanguageServer.$anonfun$reindexWorkspaceSources$2(MetalsLanguageServer.scala:1592)
	at scala.meta.internal.metals.MetalsLanguageServer.$anonfun$reindexWorkspaceSources$2$adapted(MetalsLanguageServer.scala:1589)
	at scala.collection.Iterator.foreach(Iterator.scala:943)
	at scala.collection.Iterator.foreach$(Iterator.scala:943)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
	at scala.meta.internal.metals.MetalsLanguageServer.reindexWorkspaceSources(MetalsLanguageServer.scala:1589)
	at scala.meta.internal.metals.MetalsLanguageServer.$anonfun$onChange$2(MetalsLanguageServer.scala:962)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:659)
```

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
@@ -11,6 +11,7 @@ import org.eclipse.lsp4j.ExecuteCommandParams
 import org.eclipse.lsp4j.Location
 import org.eclipse.lsp4j.Range
 import scala.meta.internal.metals.Messages.NewScalaFile
+import java.nio.file.FileAlreadyExistsException
 
 class NewFilesProvider(
     workspace: AbsolutePath,
@@ -164,15 +165,22 @@ class NewFilesProvider(
       name: String
   ): Future[(AbsolutePath, Range)] = {
     val path = directory.getOrElse(workspace).resolve(name + ".worksheet.sc")
-    createFile(path).map((_, new Range))
+    createFileAndWriteText(path, NewFileTemplate.empty)
   }
 
-  private def createFile(
-      path: AbsolutePath
-  ): Future[AbsolutePath] = {
-    val result = Future {
-      path.touch()
-      path
+  private def createFileAndWriteText(
+      path: AbsolutePath,
+      template: NewFileTemplate
+  ): Future[(AbsolutePath, Range)] = {
+    val result = if (path.exists) {
+      Future.failed(
+        new FileAlreadyExistsException(s"The file $path already exists.")
+      )
+    } else {
+      Future {
+        path.writeText(template.fileContent)
+        (path, template.cursorPosition.toLSP)
+      }
     }
     result.failed.foreach {
       case NonFatal(e) => {
@@ -184,16 +192,6 @@ class NewFilesProvider(
       }
     }
     result
-  }
-
-  private def createFileAndWriteText(
-      path: AbsolutePath,
-      template: NewFileTemplate
-  ): Future[(AbsolutePath, Range)] = {
-    createFile(path).map { _ =>
-      path.writeText(template.fileContent)
-      (path, template.cursorPosition.toLSP)
-    }
   }
 
   private def openFile(path: AbsolutePath, cursorRange: Range): Unit = {
--- a/tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
@@ -9,6 +9,8 @@ import scala.meta.internal.metals.MetalsInputBoxResult
 import scala.meta.internal.metals.ClientExperimentalCapabilities
 import scala.meta.internal.metals.MetalsInputBoxParams
 import org.eclipse.lsp4j.ShowMessageRequestParams
+import munit.TestOptions
+import java.nio.file.FileAlreadyExistsException
 
 class NewFilesLspSuite extends BaseLspSuite("new-files") {
   override def experimentalCapabilities
@@ -100,15 +102,37 @@ class NewFilesLspSuite extends BaseLspSuite("new-files") {
                        |""".stripMargin
   )
 
+  check("existing-file")(
+    Some("a/src/main/scala/foo"),
+    "class",
+    Some("Other"),
+    "a/src/main/scala/foo/Other.scala",
+    expectedContent = s"""|package foo
+                          |
+                          |class Other {
+                          | val withContent = true
+                          |}
+                          |""".stripMargin,
+    existingFiles = s"""|/a/src/main/scala/foo/Other.scala
+                        |package foo
+                        |
+                        |class Other {
+                        | val withContent = true
+                        |}
+                        |""".stripMargin,
+    expectedException = List(classOf[FileAlreadyExistsException])
+  )
+
   private def indent = "  "
 
-  private def check(testName: String)(
+  private def check(testName: TestOptions)(
       directory: Option[String],
       pickedKind: String,
       name: Option[String],
       expectedFilePath: String,
       expectedContent: String,
-      existingFiles: String = ""
+      existingFiles: String = "",
+      expectedException: List[Class[_]] = Nil
   ): Unit = test(testName) {
     val directoryUri = directory.fold(null.asInstanceOf[String])(
       workspace.resolve(_).toURI.toString()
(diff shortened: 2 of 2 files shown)


kpbochenek:
Also error from latest build:
```
ERROR D:\a\metals\metals\tests\unit\target\e2e\new-files\new-trait-new-dir\a\src\main\scala\bar\Baz.scala
INFO  time: indexed workspace in 1.38s
INFO  tests.TestingServer#executeCommand
java.nio.file.FileSystemException: D:\a\metals\metals\tests\unit\target\e2e\new-files\new-trait-new-dir\a\src\main\scala\bar\Baz.scala: The process cannot access the file because it is being used by another process.
INFO  Executing command [new-scala-file]

[info]   + new-trait-new-dir 3.601s
	at sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
[info] tests.NewFilesLspSuite.new-package-object started
	at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
	at sun.nio.fs.WindowsFileSystemProvider.newByteChannel(WindowsFileSystemProvider.java:235)
```

Isn't this actually a bug not a flakiness?

tgodzik:
@kpbochenek You might be right, I think there is a possibility there was a race condition, where we try to create a file in two ways. We should be fine to just use `writeText`, which already creates everything. Let's see if it helps.

tgodzik:
First run clean, reruning...

tgodzik:
Second one clear, 3 time is the charm.

tgodzik:
It looks like the issue is fixed, but not 100% sure. This change is still valid even if it doesn't fix anything.

kpbochenek:
Looking at implementation of `writeText` it looks like it creates all necessary parent directories, creates tmp file, puts code there and then performs `move` where a file should actually be.

With that in mind I think `createFile` call is not necessary here.
Also exception showing here might be caused by metals/bloop locking a freshly created file, then we create temp file and put content there, and when we perform move we fail because a file is used by bloop/metals?

kpbochenek:
Also I just realized we don't check anywhere if a file already exists if we want to create a new class :fearful: 

Which means you can easily erase your whole file by creating a class with the same name :fearful: 

Do you agree `new-scala-file` should fail if file already exists? (and not erase currently existing file? :fearful: )

alekseiAlefirov:
> Do you agree new-scala-file should fail if file already exists? (and not erase currently existing file? fearful )

yes, I think, it should fail (and mention it to the user) and not rewrite file.

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
```diff
@@ -102,7 +103,7 @@ class NewFilesLspSuite extends BaseLspSuite("new-files") {
 
   private def indent = "  "
 
-  private def check(testName: String)(
+  private def check(testName: TestOptions)(
```
alekseiAlefirov, line 106: just out of curiosity:
1) why?
2) Is there implicit conversion happening? I see, there is `TestOptionsConversions` at munit, but it does not look imported and available...

alekseiAlefirov:
@tgodzik so you saying, that two `Files.createFile` sequential calls eventually get asynchronous?
Sounds like a source of flakiness :neutral_face:

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
tgodzik: I can revert, but this enables adding tags and ignoring.

tgodzik:
> Looking at implementation of writeText it looks like it creates all necessary parent directories, creates tmp file, puts code there and then performs move where a file should actually be.
> With that in mind I think createFile call is not necessary here.
> Also exception showing here might be caused by metals/bloop locking a freshly created file, then we create temp file and put content there, and when we perform move we fail because a file is used by bloop/metals?

I think that was an issue, trying to create a file 2 different ways at the same time. Windows might have locked some directories and that's why the exception. I change to just using `writeText`, which should be enough.

> Also I just realized we don't check anywhere if a file already exists if we want to create a new class 
> Which means you can easily erase your whole file by creating a class with the same name fearful
> Do you agree new-scala-file should fail if file already exists? (and not erase currently existing file? fearful )

I will add a check, good idea!

> @tgodzik so you saying, that two Files.createFile sequential calls eventually get asynchronous?
> Sounds like a source of flakiness 

The lock might not be released in time, that might have sometimes cuased flaky issues.

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
ckipp01: We also do this in other places, so we probably don't have to revert.

tgodzik:
I added a check to see if a file exists and an accompanying test case.

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
```diff
@@ -164,15 +164,22 @@ class NewFilesProvider(
...
-    val result = Future {
-      path.touch()
-      path
+  private def createFileAndWriteText(
+      path: AbsolutePath,
+      template: NewFileTemplate
+  ): Future[(AbsolutePath, Range)] = {
+    val result = if (path.exists) {
+      Future.failed(
+        new IllegalArgumentException(s"The file $path already exists.")
```
alekseiAlefirov, line 176: maybe `FileAlreadyExistsException`?

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
```diff
@@ -100,15 +101,36 @@ class NewFilesLspSuite extends BaseLspSuite("new-files") {
...
   )
 
+  val filePath = "a/src/main/scala/foo/Other.scala"
+  val existingFileContents =
+    """|package foo
+       |
+       |class Other {
+       | val withContent = true
+       |}
+       |""".stripMargin
```
alekseiAlefirov, line 111: these two declarations look a bit out here, among `check`s. I would put it somewhere down with appropriate names (fixture style), or remove a new line between them and the check below, to express they are related.

tgodzik: I moved them into the check, I think it's fine to have it duplicated there and looks more like the other test cases.

metals/src/main/scala/scala/meta/internal/metals/NewFilesProvider.scala
tgodzik: Good idea! Changed

Issue #1614, opened 2020-04-14
Migrate client config flags to initializationOptions


ayoub-benali:
**Is your feature request related to a problem? Please describe.**

ST4 is in alpha since several months, multiple releases have been published and they fix several limitation the previous version has regarding LSP. By default to handle the two versions we would need a new accepted value for `metals.client`.

**Describe the solution you'd like**

But @ckipp01 had a good suggestion on [gitter](https://gitter.im/scalameta/metals?at=5e947094d021c45cc7e1f1ab) and since there was already some work in https://github.com/scalameta/metals/pull/1414. Metals should read client configuration flags from `initializationOptions`. This way metals client extensions can enable or disable features depending on the client version without having to add to Metals server properties each time. 
 
I can make a PR and would handle initially `isCompletionItemDetailEnabled`, `isExitOnShutdown` but I can add others like `statusBar`, `metals.input-box` if you have other suggestions.

**Describe alternatives you've considered**

Add a new value for `metals.client`

ckipp01:
@ayoub-benali I think this is a good approach, and if I would have known better at the time I would have done the ones in #1414 as `initializationOptions` rather than `clientExperimentalCapabilities`. However, that doesn't prevent us from moving in that direction. I think as long as we support both for some time we may be able to migrate them all over to `initializationOptions` which could in theory almost remove the need for a lot of the server properties.

ayoub-benali:
I can make the flags introduced in https://github.com/scalameta/metals/pull/1414 also available in `initializationOptions` with fallback to `clientExperimentalCapabilities`. What do you think ?

ckipp01:
Yea so just reading it over again [here](https://microsoft.github.io/language-server-protocol/specification#initialize), I think yea you could probably include all of the ones below as `initializationOptions`.

- statusBar
- didFocus
- slowTask
- InputBox
- execute-client-command
- doctorFormat

Plus the extra ones you mentioned as well. As you said, we'd still want to fall back to the ones that exist as `clientExperimentalCapabilities` and also fall back to to them being set as server properties to not break clients that may be using them.

PR #1626, opened 2020-04-15
Handle initialization options

ayoub-benali:
fixes https://github.com/scalameta/metals/issues/1614

Migrated from `clientExperimentalCapabilities`
- [x] statusBar
- [x] didFocus
- [x] slowTask
- [x] InputBox
- [x] quickPick
- [x] execute-client-command
- [x] doctorFormat

Migrated from server properties 
- [x] exit-on-shutdown
- [x] completion-item.detail
- [x] completion-item.documentation
- [x] isHoverDocumentationEnabled
- [x] hover.documentation
- [x] signature-help.documentation
- [x] completion-item.resolve
- [x] http

The following have been left out because they are none boolean settings so we have to decide on a priority if setting is set both ways: 
- icons
- glob-syntax
- completionCommand
- signature-help.command
- overrideDefFormat

- [x] Update doc to make initialization options the recommended way for this flags

### DIFF
--- a/docs/editors/new-editor.md
+++ b/docs/editors/new-editor.md
@@ -78,7 +78,7 @@ example, in vim-lsc the `window/logMessage` notification is always displayed in
 the UI so `-Dmetals.status-bar=log-message` can be configured to direct
 higher-priority messages to the logs. However, whenever possible, if the client
 supports the ability to add in `experimental` items to the `ClientCapabilities`
-interface and/or to `InitializationOptions`, this is preferable.
+interface or to `InitializationOptions`, this is preferable.
 
 ### `-Dmetals.verbose`
 
--- a/docs/editors/new-editor.md
+++ b/docs/editors/new-editor.md
@@ -284,6 +284,9 @@ Possible values:
 - `off`: the client does not add any indentation when receiving a multi-line
   textEdit
 
+*Usage of `doctorProvider` in `InitializationOptions.compilerOptions.snippetAutoIndent` is preferable.*
+
+
 ### `-Dmetals.doctor-format`
 
 Format that you'd like Doctor to return information in.
--- a/docs/editors/new-editor.md
+++ b/docs/editors/new-editor.md
@@ -677,11 +677,11 @@ The currently available settings for `InitializationOptions` are listed below.
       "isExitOnShutdown" : boolean,
       "isHttpEnabled": boolean,
       "compilerOptions":{
-        "isCompletionItemDetailEnabled": boolean
-        "isCompletionItemDocumentationEnabled": boolean
-        "isHoverDocumentationEnabled": boolean
-        "snippetAutoIndent": boolean
-        "isSignatureHelpDocumentationEnabled": boolean
+        "isCompletionItemDetailEnabled": boolean,
+        "isCompletionItemDocumentationEnabled": boolean,
+        "isHoverDocumentationEnabled": boolean,
+        "snippetAutoIndent": boolean,
+        "isSignatureHelpDocumentationEnabled": boolean,
         "isCompletionItemResolve": boolean
       }
     }
--- a/docs/editors/new-editor.md
+++ b/docs/editors/new-editor.md
@@ -78,7 +78,7 @@ example, in vim-lsc the `window/logMessage` notification is always displayed in
 the UI so `-Dmetals.status-bar=log-message` can be configured to direct
 higher-priority messages to the logs. However, whenever possible, if the client
 supports the ability to add in `experimental` items to the `ClientCapabilities`
-interface, this is preferable.
+interface and/or to `InitializationOptions`, this is preferable.
 
 ### `-Dmetals.verbose`
 
@@ -119,7 +119,7 @@ Possible values:
   `window/showMessage` notifications. Used by coc.nvim and sublime at the
   moment.
 
-*Usage of `statusBarProvider` in `ClientCapabilities.experimental` is
+*Usage of `statusBarProvider` in `InitializationOptions` is
 preferable.*
 
 ### `-Dmetals.slow-task`
@@ -131,6 +131,9 @@ Possible values:
 - `status-bar`: the `metals/slowTask` request is not supported, but send updates
   about slow tasks via `metals/status`.
 
+*Usage of `slowTaskProvider` in `InitializationOptions` is
+preferable.*
+
 ### `-Dmetals.input-box`
 
 Possible values:
@@ -139,7 +142,7 @@ Possible values:
   Metals tries to fallback to `window/showMessageRequest` when possible.
 - `on`: the `metals/inputBox` request is fully supported.
 
-*Usage of `inputBoxProvider` in `ClientCapabilities.experimental` is preferable.*
+*Usage of `inputBoxProvider` in `InitializationOptions` is preferable.*
 
 ### `-Dmetals.execute-client-command`
 
@@ -151,8 +154,7 @@ Possible values:
 - `on`: the `metals/executeClientCommand` notification is supported and all
   [Metals client commands](#metals-client-commands) are handled.
 
-*Usage of `executeClientCommandProvider` in `ClientCapabilities.experimental` is
-preferable.*
+*Usage of `executeClientCommandProvider` in `InitializationOptions` is preferable.*
 
 ### `-Dmetals.http`
 
@@ -162,6 +164,8 @@ Possible values:
 - `on`: start a server with the [Metals HTTP client] to interact with the server
   through a basic web UI. This option is needed for editor clients that don't
   support necessary requests such as `window/showMessageRequest`.
+
+*Usage of `isHttpEnabled` in `InitializationOptions` is preferable.*
 
 ### `-Dmetals.icons`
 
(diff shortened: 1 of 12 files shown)


metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
```diff
@@ -0,0 +1,37 @@
+package scala.meta.internal.metals
+import com.google.gson.JsonElement
+import org.eclipse.{lsp4j => l}
+
+final case class InitializationOptions(
+    statusBar: String = "off",
```
ckipp01, line 6: I'd recommend trying to stay as close to the names we use in `ClientExperimentalCapabilities`.

https://github.com/scalameta/metals/blob/8dc77885d48f5ebd7b97f59b17c6ef850beb7c73/metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala#L6-L16

ayoub-benali:
@ckipp01 while I am at it. Do you think other compiler flags likes `isCompletionItemResolve` should be read from `initializationOptions` as well? 
Because I see them set [here](https://github.com/scalameta/metals/blob/0a3ec5b82439b7da7936b770ce6d250dfe8fbd44/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala#L136) for example for vim

Basically what in `PresentationCompilerConfig.scala`

ckipp01:
> @ckipp01 while I am at it. Do you think other compiler flags likes `isCompletionItemResolve` should be read from `initializationOptions` as well?
> Because I see them set [here](https://github.com/scalameta/metals/blob/0a3ec5b82439b7da7936b770ce6d250dfe8fbd44/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala#L136) for example for vim
> 
> Basically what in `PresentationCompilerConfig.scala`

Yea, I honestly think those can be as well.

ckipp01:
So one thing to think about that maybe @tgodzik or someone else has thoughts on. Some of the ones I question about migrating are ones that deal directly with LSP extensions that we have created since they aren't technically part of LSP, I'm not 100% sure if they should be part of `initializationOptions` or `clientExperimentalCapabilities`. Some of them for example I know that I'll keep in `clientExperimentalCabailities` since they need to be set dynamically. For example in coc-metals a user can turn on or off `metals/status` depending on whether or not they have a statusline integration. Therefore, that needs to remain in `clientExperimentalCapablities` because `initializationOptions` can't be dynamically set in coc.nvim. @gabro I feel like you also have good input on LSP stuff. Any thoughts on this?

ayoub-benali:
@ckipp01 you mean for `coc-nvim` `metals/status` support could be dropped in the middle of an LSP session ?

ckipp01:
> @ckipp01 you mean for `coc-nvim` `metals/status` support could be dropped in the middle of an LSP session ?

Not in the middle of an LSP session, but in the middle of the sessions for example the user could turn it off, which would trigger a server restart with the new setting. set in `clientExperimentalCapabilities`. However if that setting is hardcoded in the `package.json` file as an `initializationOption`, it won't be able to be updated by the user in that way.

ayoub-benali:
> However if that setting is hardcoded in the package.json file as an initializationOption

I see now that this is the core issue. `initializationOption` can only be set there ?

ayoub-benali:
@ckipp01 I updated the field names and the PR description. There are settings I couldn't migrate because they aren't Boolean so we would need to prioritize their initializationOption value over the server settings, or similar.

Please let me know what you think so I can work on the doc once the settings settled.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -357,12 +357,18 @@ class Compilers(
...
       .withScheduledExecutorService(sh)
       .withConfiguration(
-        config.compilers.copy(
-          _symbolPrefixes = userConfig().symbolPrefixes,
-          isCompletionSnippetsEnabled =
-            initializeParams.supportsCompletionSnippets,
-          isFoldOnlyLines = initializeParams.foldOnlyLines
-        )
+        initializeParams
+          .map(p => {
```
ckipp01, line 361:
```suggestion
          .map(params => {
```

metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
```diff
@@ -85,7 +92,7 @@ final class ConfiguredLanguageClient(
   override def metalsExecuteClientCommand(
       params: ExecuteCommandParams
   ): Unit = {
-    if (config.executeClientCommand.isOn || clientCapabilities.executeClientCommandProvider) {
+    if (config.executeClientCommand.isOn || clientCapabilities.executeClientCommandProvider || initializationOptions.executeClientCommandProvider) {
```
ckipp01, line 95: In a lot of these boolean checks can you pull them out into a variable like you did below? For example this one being like `isExecuteClientCommandProvider` or something like that. Now that we have three different boolean checks in here, it's getting a bit messy.

ckipp01:
> I see now that this is the core issue. initializationOption can only be set there ?

I'm not 100% sure to be honest. I need to look more into it.

## A bit of context

Just for the others reviewing this, I'll give a bit more backstory to why this pr exists. Initially we set everything with properties. A While back [here](https://github.com/scalameta/metals/pull/1414) I had a pr where I moved a chunk of stuff from server properties to `clientExperimentalCapablities`. In hindsight I think I should have moved them to `initializationOptions` rather than `clientExperimentalCapablities`. Both can be used in the same way to really set any value, but if you look at the [lsp spec](https://microsoft.github.io/language-server-protocol/specification#initialize) you see that the way `clientExperimentalCapablities` is talked about is like the following:

> The experimental can be used to pass experimental capabilities under development.

I think the things we are passing in there aren't really _experimental_ so I question the correctness of having them there. Looking around at some other LSP integrations, I've noticed that many make it somewhat easy to set `initializationOptions`, but not so easy to set `clientExperimentalCapablities`. I found this the case with `kak-lsp` when looking into Kakuaone.

I'm propose that we communicate that we are moving away from things being in `clientExperimentalCapablities` (maybe even deprecate the values we have in there if that makes sense?) and document that the way to set values should be in `initializationOptions` rather than `clientExperimentalCapablities` unless like the spec says, they are in development and are actually experimental.

In theory, we could set everything we need to this way and completely move away from having to even set a client as a property. I think this would be a good goal.

@ayoub-benali thanks a lot for working on this.

@kurnevsky I'm also tagging you on here because you were involved with the emacs fold integrating this stuff as `clientExperimentalCapablities`. Just want to give you a heads up on this conversation as it's a direction we are discussing moving to.

ayoub-benali:
> Looking around at some other LSP integrations, I've noticed that many make it somewhat easy to set initializationOptions, but not so easy to set clientExperimentalCapablities.

It was exactly the case for sublime LSP package, I had to add it https://github.com/sublimelsp/LSP/pull/971 this might also happen for other clients.

ayoub-benali:
> Though I think it would make sense to have an update method inside MetalsServerConfig and we could have all the logic for resolving conflicts in there

Yes but the biggest issue is that `config` in `MetalsLanguageServer` is immutable while `clientExperimentalCapabilities` and `initializationOptions` are mutables. They get initialized with default values and then later get mutated once the initialization request is done. But in the mean time their references are passed around ...

A proper solution where all these settings would get merged in a single place would require quite some refactoring of `MetalsLanguageServer`

docs/editors/new-editor.md
```diff
@@ -78,7 +78,7 @@ example, in vim-lsc the `window/logMessage` notification is always displayed in
 the UI so `-Dmetals.status-bar=log-message` can be configured to direct
 higher-priority messages to the logs. However, whenever possible, if the client
 supports the ability to add in `experimental` items to the `ClientCapabilities`
-interface, this is preferable.
+interface and/or to `InitializationOptions`, this is preferable.
```
ckipp01, line 81:
```suggestion
interface or to `InitializationOptions`, this is preferable.
```
I'd say just use **or**. We don't want confusion over whether they should set them both or not.

metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
```diff
@@ -25,22 +25,31 @@ final class ConfiguredLanguageClient(
...
-    if (config.statusBar.isOn || clientCapabilities.statusBarIsOn) {
+    val statusBarIsOn = config.statusBar.isOn ||
+      clientCapabilities.statusBarIsOn ||
+      initializationOptions.statusBarIsOn
+
+    if (statusBarIsOn) {
       underlying.metalsStatus(params)
     } else if (params.text.nonEmpty && !pendingShowMessage.get()) {
-      if (config.statusBar.isShowMessage || clientCapabilities.statusBarIsShowMessage) {
+      if (config.statusBar.isShowMessage || clientCapabilities.statusBarIsShowMessage || initializationOptions.statusBarIsShowMessage) {
```
ckipp01, line 50: Nit: Can you do the same for this one and the one below and separate this out into a variable like you did for `statusBarIsOn`. Then we can just do checks on `statusBarIsShow` and `statusBarIsLog`.

docs/editors/new-editor.md
ayoub-benali: I was trying to say that some setting should be part of `InitializationOptions` other in `experimental`

ckipp01: Understood 👍

PR #1538, opened 2020-03-25
Add early support for Ammonite scripts

alexarchambault:
~Just checking if the CI is green for now, will add more details once it is. (This probably needs some more clean-up / tweaking too.)~

Fixes https://github.com/scalameta/metals/issues/1766

PR #1628, opened 2020-04-16
Initial support for sbt meta projects.

dos65:
Implements definition, hover and completion for `*.sbt` files.
Errors highlighting and quick-fixes aren't supported yet.

Project importing from BSP was done by copying bloop config file into the root `.bloop` directory. This part might be changed after merge #1538 to reuse changes around BSP connections.

For a now, I'm leaving it just to get a review because it requires at least to wait when bloop will be released with [fix for sbt project](https://github.com/scalacenter/bloop/pull/1219)

metals_sbt


tgodzik:
@dos65 Closing this for now until the Bloop PR is reworked. I think we are on the right track here and will be much easier with the proposed changes to Bloop. Thanks a lot for the great work!

PR #1630, opened 2020-04-16
Show date in logs

mlachkar:
I think it can be useful to show date in metals logs
```
2020.04.16 11:31:34 INFO  Connected to Build server v1.4.0-RC1-162-888454e1
2020.04.16 11:31:34 INFO  time: indexed workspace in 0.54s
2020.04.16 11:31:51 INFO  shutting down Metals
```


ritschwumm:
just one small thing - is the date really formatted with dots? i think that's a bit confusing, as dots usually signify a day.month.year field order. as a user, i'd expect year-month-day like in ISO-8601 dates.

mlachkar:
This the only format provided by scribe (if we want to use their date)

olafurpg:
LGTM 👍 Thank you!

PR #1637, opened 2020-04-17
Use DefinitionProvider instead of index in RenameProvider 

tgodzik:
Previously, we would use GlobalSymbolIndex, which might not always work for Scala 3 in case of non-parsable code. Now, we use the definitionProvider, which will in turn ask the PresentationCompiler if the definition is missing. For Scala 2 it will still work mostly the same.

The test might fail locally due to issue being fixed https://github.com/scalacenter/bloop/pull/1253, which makes the test fail on even tries :sweat_smile:

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -12,9 +12,7 @@ import scala.meta.internal.metals.MetalsEnrichments._
 import org.eclipse.lsp4j.Position
 import org.eclipse.lsp4j.ReferenceContext
 import scala.meta.internal.metals.DefinitionProvider
-import scala.meta.internal.mtags.{Symbol => MSymbol}
 import scala.meta.internal.semanticdb.Scala._
-import scala.meta.internal.mtags.GlobalSymbolIndex
 import scala.meta.io.AbsolutePath
 import org.eclipse.lsp4j.Location
 import scala.meta.internal.metals.MetalsLanguageClient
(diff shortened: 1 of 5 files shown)


metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -35,138 +33,152 @@ import scala.meta.internal.semanticdb.Synthetic
...
-      for {
-        (occurence, semanticDb) <- symbolOccurence
-        if canRenameSymbol(occurence.symbol, None)
-        range <- occurence.range
-      } yield range.toLSP
+      definitionProvider.definition(source, params, token).map { definition =>
+        val symbolOccurence =
+          definitionProvider.symbolOccurrence(source, params.getPosition)
+        for {
+          (occurence, _) <- symbolOccurence
```
kpbochenek, line 67: rename `symbolOccurence` -> `symbolOccurrence` let's fix this once and for all in this file :muscle:

kpbochenek:
a little unrelated but when I looked at `def definition(` it contains:
```
if (result.isEmpty) {
        compilers().definition(params, token)
      } else {
        if (result.isEmpty && fromSemanticdb.isEmpty) {
          warnings.noSemanticdb(path)
        }
        Future.successful(result)
      }
```
in second `if` result cannot be empty

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -279,21 +295,21 @@ final class RenameProvider(
     if (colonNotAllowed) {
       client.showMessage(forbiddenColonRename(name, newName))
     }
-    symbolIsLocal(symbol) && (!desc.isMethod || (!colonNotAllowed && !isForbidden))
+    symbolIsLocal(symbol, definitionPath) && (!desc.isMethod || (!colonNotAllowed && !isForbidden))
   }
 
-  private def symbolIsLocal(symbol: String): Boolean = {
+  private def symbolIsLocal(
```
kpbochenek, line 301: maybe we could rename it to `isLocalSymbol` ?

also inside it is `symbol.startsWith("local")` but we can simply reuse `symbol.isLocal` which is performing exactly this check.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -279,21 +295,21 @@ final class RenameProvider(
...
-        .definition(MSymbol(symbol))
-        .exists { definition =>
-          val isLocal = definition.path.isWorkspaceSource(workspace)
-          if (isLocal && definition.path.isJava) {
-            client.showMessage(javaSymbol(symbol.desc.name.value))
-          }
-          isLocal
-        }
+    def isFromWorkspace = {
+      val isLocal = definitionPath.isWorkspaceSource(workspace)
```
kpbochenek, line 307: for me `isLocal` has a meaning that symbol is `localXXX`(anonymous in a given file) and here we talk about `isInWorkspace` ?

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -267,7 +279,11 @@ final class RenameProvider(
     }
   }
 
-  private def canRenameSymbol(symbol: String, newName: Option[String]) = {
+  private def canRenameSymbol(
+      symbol: String,
```
kpbochenek, line 283: I think you try to bind symbol with AbsolutePath and propagate them together through multiple functions, maybe it is worth making case class containing those 2 things and just pass one thing. Also it will be easier to see this path is related to this symbol

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -35,138 +33,152 @@ import scala.meta.internal.semanticdb.Synthetic
...
-        isLocal = occurence.symbol.isLocal
-        currentReferences = referenceProvider
-          .references(
-            // we can't get definition by name for local symbols
-            toReferenceParams(txtParams, includeDeclaration = isLocal),
-            canSkipExactMatchCheck = false,
-            includeSynthetics = includeSynthetic
+        val allReferences = for {
+          (occurence, semanticDb) <- symbolOccurence.toIterable
+          definitionLoc <- definition.locations.asScala
```
kpbochenek, line 111: I would not go with `<-` here as it means if you get more locations all references and will be calculated twice for each location

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
kpbochenek: Okay actually I would go with symbol <-> List[AbsolutePath] as it looks like there can be multiple definitions of a symbol. This would fix a problem with `<-` above

kpbochenek: And then you can decide in `symbolIsLocal` if given symbol with all definition locations is local or not. 
Previously we did `.exists` which means at least one location must be in workspace, but maybe `.all` would make more sense here?
I don't know many multi def locations examples, is it only class/trait + object scenario?

kpbochenek:
From what I see you basically replace `definitionProvider.fromSymbol(occurence.symbol)` with `definitionProvider.definition(source, params, token).`
Two questions:
1. Would it make sense to use index and in case of no result fallback to DefProvider?
2. There are multiple places where `definitionProvider.fromSymbol`, would it make sense to also change them?

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: I don't think we have multiple definitions anywhere currently really. And even if we get one, I can't imagine one being inside workspace and the other outside.

I can rework it a bit, but I don't think we need to add a case class just for 2 methods.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: I should most likely go with headOption, but we don't have really currently a situation with multiple definitions.

tgodzik:
>From what I see you basically replace definitionProvider.fromSymbol(occurence.symbol) with definitionProvider.definition(source, params, token).
>Two questions:

>Would it make sense to use index and in case of no result fallback to DefProvider?
DefProvider does it already, uses index to locate the SemanticDB with the definition and then searches that SemanticDB. If it's not available there then we fallaback to PC

> There are multiple places where definitionProvider.fromSymbol, would it make sense to also change them?

It might make sense, but before Dotty this wasn't needed. We might need to go through all the functionalities and confirm they're working sensibly in all scenarios.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: Moved `isWorkspaceSymbol` up, so that we don't propagate the path now.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: Changed to `isWorkspaceSymbol`

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: fixed

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala

mlachkar:
LGTM too.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -35,138 +33,154 @@ import scala.meta.internal.semanticdb.Synthetic
...
-          .references(
-            // we can't get definition by name for local symbols
-            toReferenceParams(txtParams, includeDeclaration = isLocal),
-            canSkipExactMatchCheck = false,
-            includeSynthetics = includeSynthetic
+        val allReferences = for {
+          (occurence, semanticDb) <- symbolOccurrence.toIterable
+          definitionLoc <- definition.locations.asScala.headOption.toIterable
+          definitionPath = definitionLoc.getUri().toAbsolutePath
+          if canRenameSymbol(occurence.symbol, Option(newName)) &&
```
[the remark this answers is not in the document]
tgodzik: Will fix it maybe next time, shouldn't be a big issue.

PR #1639, opened 2020-04-17
Implement textDocument/documentSymbol for Scala 3

### LINKED ISSUE #1636: Implement textDocument/documentSymbol for Scala 3
In the first implementation we will most likely use the Dotty parser, but move to Scalameta one after https://github.com/scalameta/scalameta/issues/2022 is finished.

gabro:
Closes #1636 

This PR implements the `DocumentSymbolProvider` for Scala 3.

A few things to notice:
- we're using the dotty presentation compiler to parse the source. I've tried using the `Parser` directly but I couldn't make positions work for some reason

- that said, using the PC grants us parsing error recovery, so we get an outline even if the buffer parses only partially

- I still think it will make sense to switch back to the Scalameta parser once it's available for Scala 3, since it's way easier to work with in term of syntactic trees

- I haven't tried using the typed trees for this, which may be more convenient (working with symbols instead of trees)

This is still WIP, since it misses a few edge cases and tests.

### DIFF
--- a/mtags/src/main/scala-3/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
+++ b/mtags/src/main/scala-3/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
@@ -1,0 +1,176 @@
+package scala.meta.internal.metals
+
+import java.util
+import org.eclipse.lsp4j.DocumentSymbol
+import org.eclipse.lsp4j.SymbolKind
+import org.eclipse.{lsp4j => l}
+import scala.meta.internal.mtags.MtagsEnrichments._
+import scala.meta.internal.jdk.CollectionConverters._
+import java.net.URI
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.util.SourceFile
+import dotty.tools.dotc.util.SourcePosition
+import dotty.tools.dotc.parsing.Parsers.Parser
+import dotty.tools.dotc.ast.untpd._
+import dotty.tools.dotc.printing.Showable
+import dotty.tools.dotc.core.{Flags => f}
+import java.{util => ju}
+
+/**
+ *  Retrieves all the symbols defined in a document
+ */
+class DocumentSymbolProvider() {
+
+  def documentSymbols()(implicit ctx: Context): util.List[DocumentSymbol] = {
+    val tree = ctx.run.units.head.untpdTree
+    new SymbolTraverser().symbols(tree)
+  }
+
+  class SymbolTraverser extends UntypedTreeTraverser {
+    var owner: DocumentSymbol = new DocumentSymbol(
+      "root",
+      SymbolKind.Namespace,
+      new l.Range(new l.Position(0, 0), new l.Position(0, 0)),
+      new l.Range(new l.Position(0, 0), new l.Position(0, 0)),
+      "",
+      new ju.ArrayList[DocumentSymbol]()
+    )
+    def symbols(
+        tree: Tree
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala-3/scala/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -186,13 +195,17 @@ case class ScalaPresentationCompiler(
...
   ): CompletableFuture[ju.List[DocumentSymbol]] = {
-    CompletableFuture.completedFuture(
-      List.empty[DocumentSymbol].asJava
-    )
+    compilerAccess.withInterruptableCompiler(
+      List.empty[DocumentSymbol].asJava,
+      params.token
+    ) { access =>
+      val driver = access.compiler()
+      driver.run(params.uri, params.text)
```
gabro, line 206: here we're running the PC: I've tried parsing the text directly but then I got all `-1` for the positions...

mtags/src/main/scala-3/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
```diff
@@ -0,0 +1,176 @@
...
+          // val (name, selection) = (t.constr, t.parents) match {
+          //   case (t, Nil => ("(anonymous)", t.sourcePos)
+          //   case inits =>
+          //     (inits.map(_.tpe.syntax).mkString(" with "), inits.head.pos)
+          // }
+          // if (t.templ.stats.nonEmpty) {
+          //   addChild(s"new $name", SymbolKind.Interface, t.pos, selection, "")
+          //   newOwner()
+          // } else continue()
+          continue() // TODO(gabro)
```
gabro, line 98: I still need to finish this

mtags/src/main/scala-3/scala/scala/meta/internal/metals/DocumentSymbolProvider.scala
```diff
@@ -0,0 +1,176 @@
...
+              .asScala
+              .forall(_.getName() != "catch"))
+            addChild("catch", SymbolKind.Struct, t.sourcePos, t.sourcePos, "")
+        case t: Try =>
+          if (t.cases.nonEmpty) {
+            addChild("try", SymbolKind.Struct, t.sourcePos, t.sourcePos, "")
+            newOwner()
+          } else {
+            continue()
+          }
```
gabro, line 131: This isn't working, yet

mtags/src/main/scala-3/scala/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -526,15 +538,12 @@ case class ScalaPresentationCompiler(
...
   private def toSource(uri: URI, sourceCode: String): SourceFile = {
     val path = Paths.get(uri)
-    val virtualFile = new VirtualFile(path.getFileName.toString, path.toString)
-    val writer = new BufferedWriter(
-      new OutputStreamWriter(virtualFile.output, "UTF-8")
-    )
-    writer.write(sourceCode)
-    writer.close()
-    new SourceFile(virtualFile, Codec.UTF8)
+    SourceFile.virtual(path.getFileName().toString, sourceCode)
```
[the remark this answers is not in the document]
tgodzik: I think this is actually copied form the Dotty Language Server, I mostly left utility methods intact.

gabro:
I won't likely have time to complete this anytime soon. It's been a fun experiment, but I think the safest route would be to stick with the Scalameta parser as soon as it's available for Scala 3

PR #1640, opened 2020-04-17
Detect Scala 3 main outer methods to create code lenses

tgodzik:
Previously, code lenses were only available on classes, now we also show them for main methods.

I am wondering if this should actually be generated differently in Dotty.

@bishabosha Main methods symbol is now shown as `foo/Foo$package.main().` while this is not something that can actually be run. I propose a bit of a hack here to generate a potential main method symbol. I am not sure if that is the right solution though. How is it being done in Dotty itself? Should we generate a different symbol for semanticDB if it's a main method?

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
@@ -13,7 +13,8 @@ import scala.meta.internal.semanticdb.Scala.Symbols
  * In-memory index of main class symbols grouped by their enclosing build target
  */
 final class BuildTargetClasses(
-    buildServer: () => Option[BuildServerConnection]
+    buildServer: () => Option[BuildServerConnection],
+    buildTargets: BuildTargets
 )(implicit val ec: ExecutionContext) {
   private val index = TrieMap.empty[b.BuildTargetIdentifier, Classes]
 
@@ -87,8 +88,17 @@ final class BuildTargetClasses(
     for {
       item <- result.getItems.asScala
       target = item.getTarget
+      buildTarget <- buildTargets.scalaTarget(target)
       aClass <- item.getClasses.asScala
-      symbol <- createSymbols(aClass.getClassName, List(Descriptor.Term))
+      descriptors = {
+        if (ScalaVersions.isScala3Version(buildTarget.scalaVersion))
+          List(Descriptor.Term, Descriptor.Type)
+        else List(Descriptor.Term)
+      }
+      symbol <- createSymbols(
+        aClass.getClassName,
+        descriptors
+      )
     } {
       classes(target).mainClasses.put(symbol, aClass)
     }
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -115,13 +115,14 @@ class MetalsLanguageServer(
   private val symbolDocs = new Docstrings(definitionIndex)
   var buildServer: Option[BuildServerConnection] =
     Option.empty[BuildServerConnection]
-  private val buildTargetClasses = new BuildTargetClasses(() => buildServer)
   private val savedFiles = new ActiveFiles(time)
   private val openedFiles = new ActiveFiles(time)
   private val messages = new Messages(config.icons)
   private val languageClient = new DelegatingLanguageClient(NoopLanguageClient)
   var userConfig: UserConfiguration = UserConfiguration()
   val buildTargets: BuildTargets = new BuildTargets()
+  private val buildTargetClasses =
+    new BuildTargetClasses(() => buildServer, buildTargets)
   private val remote = new RemoteLanguageServer(
     () => workspace,
     () => userConfig,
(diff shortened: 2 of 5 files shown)


bishabosha:
if i do this in empty package

```scala
@main def TopLevelMain: Unit = println("foo")
```
theres a symbol for `_empty_/TopLevelMain#main(). => static method main`, which is the actual entry point which then delegates to `_empty_/toplevel$package.TopLevelMain()`, is that usable?

tgodzik:
> if i do this in empty package
> 
> ```scala
> @main def TopLevelMain: Unit = println("foo")
> ```
> 
> theres a symbol for `_empty_/TopLevelMain#main(). => static method main`, which is the actual entry point which then delegates to `_empty_/toplevel$package.TopLevelMain().`, is that usable?

I might have not noticed that, should be much more helpful, thanks! Though probably Bloop should generate the additional `()`

tgodzik:
@bishabosha Looks like we do have in occurences:
```
SymbolOccurrence(Some(Range(2, 0, 2, 10)), "foo/mainMethod#", DEFINITION),
```
Shouldn't it actually be `foo/mainMethod.` ?

bishabosha:
> @bishabosha Looks like we do have in occurences:
> 
> ```
> SymbolOccurrence(Some(Range(2, 0, 2, 10)), "foo/mainMethod#", DEFINITION),
> ```
> 
> Shouldn't it actually be `foo/mainMethod.` ?

@tgodzik that looks like the wrapper class with the same name as the method in source code

tgodzik:
I think this needs to be fixed in Dotty, since the information we get in semanticDB is not really valid:
```
RunTestCodeLens.scala:70 textDocument.symbols: Vector(
  SymbolInformation("foo/mainMethod#main().", SCALA, METHOD, 4096, "main", Empty, Vector(), Empty),
  SymbolInformation("local0", SCALA, LOCAL, 1024, "error", Empty, Vector(), Empty),
  SymbolInformation("foo/mainMethod#", SCALA, CLASS, 8, "mainMethod", Empty, Vector(), Empty),
  SymbolInformation("foo/Foo$package.", SCALA, PACKAGE_OBJECT, 8, "foo", Empty, Vector(), Empty),
  SymbolInformation(
    "foo/mainMethod#`<init>`().",
    SCALA,
    CONSTRUCTOR,
    8192,
    "<init>",
    Empty,
    Vector(),
    Empty
  ),
  SymbolInformation(
    "foo/mainMethod#main().(args)",
    SCALA,
    PARAMETER,
    0,
    "args",
    Empty,
    Vector(),
    Empty
  ),
  SymbolInformation(
    "foo/Foo$package.mainMethod().",
    SCALA,
    METHOD,
    0,
    "mainMethod",
    Empty,
    Vector(),
    Empty
  )
)
RunTestCodeLens.scala:71 textDocument.occurrences: Vector(
  SymbolOccurrence(Some(Range(0, 8, 0, 11)), "foo/", DEFINITION),
  SymbolOccurrence(Some(Range(2, 0, 2, 11)), "foo/Foo$package.", DEFINITION),
  SymbolOccurrence(Some(Range(2, 1, 2, 5)), "scala/main#", REFERENCE),
  SymbolOccurrence(Some(Range(2, 10, 2, 20)), "foo/Foo$package.mainMethod().", DEFINITION),
  SymbolOccurrence(Some(Range(2, 24, 2, 28)), "scala/Unit#", REFERENCE),
  SymbolOccurrence(Some(Range(3, 2, 3, 9)), "scala/Predef.println(+1).", REFERENCE),
  SymbolOccurrence(Some(Range(2, 0, 2, 10)), "foo/mainMethod#", DEFINITION),
  SymbolOccurrence(Some(Range(2, 0, 2, 0)), "foo/mainMethod#`<init>`().", DEFINITION),
  SymbolOccurrence(Some(Range(2, 0, 2, 0)), "foo/Foo$package.mainMethod().", REFERENCE),
  SymbolOccurrence(Some(Range(2, 0, 2, 0)), "scala/util/CommandLineParser.showError().", REFERENCE),
  SymbolOccurrence(Some(Range(2, 0, 2, 0)), "local0", REFERENCE)
)
```

This PR is an OK workaround for now, but I would certainly want to fix how outer methods are actually represented in SemanticDB.

metals/src/main/scala/scala/meta/internal/metals/BuildTargetClasses.scala
```diff
@@ -87,8 +88,17 @@ final class BuildTargetClasses(
     for {
       item <- result.getItems.asScala
       target = item.getTarget
+      buildTarget <- buildTargets.scalaTarget(target)
       aClass <- item.getClasses.asScala
-      symbol <- createSymbols(aClass.getClassName, List(Descriptor.Term))
+      descriptors = {
+        if (ScalaVersions.isScala3Version(buildTarget.scalaVersion))
+          List(Descriptor.Term, Descriptor.Type)
```
tgodzik, line 95: In case of Scala 3 we get additional reference for Descriptor.Type when using main methods, so we can use that. In the future it could change to a method, but that depends on the changes in Dotty.

bishabosha:
What is the information in the symbol that you need to recognise it as the actual main method?

tgodzik:
> What is the information in the symbol that you need to recognise it as the actual main method?

We are being sent information by Bloop with the fully qualified class name, which we read and we get a name from. Based on that name we look for either class or object. Currently in Scala 3 we need to look for a class, but it might be better to look for a function if we want to properly represent the outer methods in semanticDB

Issue #1610, opened 2020-02-19
Add code action to auto-import all missing symbols when there are no ambiguities


olafurpg:
**Is your feature request related to a problem? Please describe.**

When copy-pasting code from one file to another I get compile errors about missing imports. I currently select the code and trigger the code action multiple times for each missing symbol

<img width="606" alt="Screenshot 2020-02-19 at 12 10 19" src="https://user-images.githubusercontent.com/1408093/74843373-b55db700-5323-11ea-8871-2067a0fe9d9c.png">

**Describe the solution you'd like**

It would be nice if there was an "import all missing symbols that are unambiguous" code action that would insert imports for all missing symbols in one stop. This code action could skip imports when there are multiple candidates with the same name.

**Describe alternatives you've considered**

Trigger the code action for each individual error.

ckipp01:
BTW @coline-lunatech is diving into this :)

PR #1652, opened 2020-04-21
Feature : Code action Import all missing symbols

coline-lunatech:
Add the code action "import all missing symbols that are unambiguous" for issue #1610

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
@@ -1,6 +1,7 @@
 package scala.meta.internal.metals.codeactions
 
 import scala.concurrent.Future
+import scala.collection.mutable
 import scala.meta.pc.CancelToken
 import org.eclipse.{lsp4j => l}
 import scala.concurrent.ExecutionContext
@@ -20,17 +21,18 @@ class ImportMissingSymbol(compilers: Compilers) extends CodeAction {
         diagnostic: l.Diagnostic,
         name: String
     ): Future[Seq[l.CodeAction]] = {
+
       val textDocumentPositionParams = new l.TextDocumentPositionParams(
         params.getTextDocument(),
         diagnostic.getRange.getEnd()
       )
+
       compilers
         .autoImports(textDocumentPositionParams, name, token)
         .map { imports =>
           imports.asScala.map { i =>
             val uri = params.getTextDocument().getUri()
             val edit = new l.WorkspaceEdit(Map(uri -> i.edits).asJava)
-
             val codeAction = new l.CodeAction()
 
             codeAction.setTitle(ImportMissingSymbol.title(name, i.packageName))
@@ -43,14 +45,34 @@ class ImportMissingSymbol(compilers: Compilers) extends CodeAction {
         }
     }
 
+    def importAllMissingSymbol(
+        codeActions: mutable.Buffer[l.CodeAction]
+    ): Seq[l.CodeAction] = {
+
+      if (codeActions.length > 1) {
+        val allSymbols: l.CodeAction = new l.CodeAction()
+        val uri = params.getTextDocument().getUri()
+        val diags = codeActions.flatMap(_.getDiagnostics().asScala)
+        val edits =
+          codeActions.flatMap(_.getEdit().getChanges().get(uri).asScala)
+
+        allSymbols.setTitle(ImportMissingSymbol.allSymbolsTitle)
+        allSymbols.setKind(l.CodeActionKind.QuickFix)
+        allSymbols.setDiagnostics(diags.asJava)
+        allSymbols.setEdit(new l.WorkspaceEdit(Map(uri -> edits.asJava).asJava))
+
+        codeActions += allSymbols
+      }
+      codeActions
+    }
+
     Future
       .sequence(params.getContext().getDiagnostics().asScala.collect {
         case d @ ScalacDiagnostic.SymbolNotFound(name)
             if params.getRange().overlapsWith(d.getRange()) =>
           importMissingSymbol(d, name)
       })
-      .map(_.flatten)
-
+      .map(actions => importAllMissingSymbol(actions.flatten))
   }
 
 }
@@ -60,4 +82,6 @@ object ImportMissingSymbol {
   def title(name: String, packageName: String): String =
     s"Import '$name' from package '$packageName'"
 
+  def allSymbolsTitle: String =
+    s"Import all missing symbols that are unambiguous"
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
@@ -1,7 +1,6 @@
 package scala.meta.internal.metals.codeactions
 
 import scala.concurrent.Future
-import scala.collection.mutable.Buffer
 import scala.meta.pc.CancelToken
 import org.eclipse.{lsp4j => l}
 import scala.concurrent.ExecutionContext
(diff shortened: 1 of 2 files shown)


tests/unit/src/test/scala/tests/codeactions/ImportMissingSymbolLspSuite.scala
```diff
@@ -78,26 +78,102 @@ class ImportMissingSymbolLspSuite
...
+        |${CreateNewSymbol.title("Future")}
+        |${CreateNewSymbol.title("Instant")}
+        |${CreateNewSymbol.title("ListBuffer")}
+        |${CreateNewSymbol.title("Future")}
+        |${CreateNewSymbol.title("ListBuffer")}
+        |""".stripMargin,
+    """|package a
+       |
+       |import java.time.Instant
+       |
```
coline-lunatech, line 170: I don't know why there's a space here ... Moreover tests sometimes pass and sometimes not cause the imports are reversed

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
```diff
@@ -43,14 +44,42 @@ class ImportMissingSymbol(compilers: Compilers) extends CodeAction {
         }
     }
 
+    def addImportAllMissingSymbol(
```
ckipp01, line 47:
```suggestion
    def importMissingSymbols(
```
Maybe this to better align with the other `importMissingSymbol` in the file.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
```diff
@@ -43,14 +44,42 @@ class ImportMissingSymbol(compilers: Compilers) extends CodeAction {
...
+        codeActions: Buffer[l.CodeAction]
+    ): Seq[l.CodeAction] = {
+      val uniqueCodeActions = codeActions
+        .groupBy(_.getDiagnostics())
+        .values
+        .filter(_.length <= 1)
```
ckipp01, line 53: What would be a scenario where we'd want `< 1`? Aren't we always targeting when there is `1` exactly?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
```diff
@@ -43,14 +44,42 @@ class ImportMissingSymbol(compilers: Compilers) extends CodeAction {
         }
     }
 
+    def addImportAllMissingSymbol(
+        codeActions: Buffer[l.CodeAction]
```
ckipp01, line 48: You can probably just leave this as a `Seq` instead of a buffer. Down below at the end of the file where this is `flattened` and passed it, it can just stay a `Seq`. Then in this method, you could change the logic a bit that instead of the `if` statement adding something to the buffer, you could do something like


```scala
val codeActionsToReturn = If(uniqueCodeActions.length > 1) {
  ... // create new code action
  codeActions :+ newCodeAction
} else {
  codeActions
}
codeActionsToReturn
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
```diff
@@ -60,4 +89,6 @@ object ImportMissingSymbol {
   def title(name: String, packageName: String): String =
     s"Import '$name' from package '$packageName'"
 
+  def allSymbolsTitle: String =
+    s"Import all missing symbols that are unambiguous"
```
ckipp01, line 93: I'm curious if showing the `unambiguous` is a bit verbose to the user.
```suggestion
    s"Import all possible missing symbols"
```
What do you think?

tests/unit/src/test/scala/tests/codeactions/ImportMissingSymbolLspSuite.scala
ckipp01: If you add a sort in the code action that returns them, does it fix this?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
coline-lunatech: yep that's true

tests/unit/src/test/scala/tests/codeactions/ImportMissingSymbolLspSuite.scala
coline-lunatech: I wanted to keep the "import all" above all the others but it does not work quite well anyway.. I'll try different ways

coline-lunatech: It will not fix anyway because it's not the order of actions but the order of TextEdits that could matter.
But actually it looks like it's kind of random, text edits order is always the same but they seems not to be applied in that order.

=> after some checks I found that it's only on tests. On files when I use the codeAction they are always applied in the same order. (Could be the textEdits.scala file)

coline-lunatech: About the spaces it's also a textEdits issue. If a file has no imports yet, imports textEdits are prepend with a \n. In our case we concatenate textEdits so they all are prepend with it..

I'm not sure I can fix this two issues

ckipp01: Ahhhh of course...
https://github.com/scalameta/metals/pull/1051
I should have known that.. seeing that I added it 😬 .

I suppose this is an edge case that you'll only hit on if you are triggering this code action from a file that has no imports yet. 🤔 I'm trying to think what the best approach to fixing this is. We could just check the text edits and strip out the leading `\n` on the second textEdit, but that will still fail in this test if it's flaky. I'm not 100% sure why the textEdits aren't applied in a uniform way in tests. @tgodzik may know more on that.

tgodzik: We do sort them by start to apply them properly:
```
      val positions = edits
        .map(edit => edit -> edit.getRange.toMeta(input))
        .sortBy(_._2.start)
```
in `TextEdits.scala`. I am not sure what will be the behaviour in case of multiple imports, but I believe we need to fix it there.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
```diff
@@ -43,14 +43,42 @@ class ImportMissingSymbol(compilers: Compilers) extends CodeAction {
...
+          .distinct
+
+        allSymbols.setTitle(ImportMissingSymbol.allSymbolsTitle)
+        allSymbols.setKind(l.CodeActionKind.QuickFix)
+        allSymbols.setDiagnostics(diags.asJava)
+        allSymbols.setEdit(new l.WorkspaceEdit(Map(uri -> edits.asJava).asJava))
+
+        allSymbols +: codeActions
+      } else { codeActions }
+      allCodeActions
```
tgodzik, line 72: This could be dropped. Just have the if-else as the last expression.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
```diff
@@ -43,14 +43,42 @@ class ImportMissingSymbol(compilers: Compilers) extends CodeAction {
...
+        val uri = params.getTextDocument().getUri()
+        val diags = uniqueCodeActions.flatMap(_.getDiagnostics().asScala)
+        val edits = uniqueCodeActions
+          .flatMap(_.getEdit().getChanges().get(uri).asScala)
+          .distinct
+
+        allSymbols.setTitle(ImportMissingSymbol.allSymbolsTitle)
+        allSymbols.setKind(l.CodeActionKind.QuickFix)
+        allSymbols.setDiagnostics(diags.asJava)
+        allSymbols.setEdit(new l.WorkspaceEdit(Map(uri -> edits.asJava).asJava))
```
tgodzik, line 68: I think we could just join all the textEdits into a single text edit, which might help with the tests and the additional newline.

coline-lunatech: I managed to fix everything that way without changing anything to the TextEdits.scala file. I don't know if this is the right way but it seems that imports are always coming in different order so maybe it is.

ckipp01:
Merging since the testing failure is unrelated.

```
[error] 	tests.debug.BreakpointDapSuite
```

ckipp01:
O, and one other nitpick to remember for the future @coline-lunatech, try to keep your commits a bit more concise. We don't have any strict rules on single commits vs multiple commits, but it's always a good idea to have your commits be able to stand alone and be descriptive. Commits like `more styling` can probably just be rolled into another commit to not make the commits to noisy.

PR #1656, opened 2020-04-22
update dialect of worksheet to scala213

### LINKED ISSUE #1591: Literal type annotations show an error in worksheets
**Describe the bug**
When using a literal type annotation within a worksheet Metals shows an error such as "identifier expected but string constant found" for string literals and "identifier expected but integer constant found" for integer literals.

I noticed a similar issue https://github.com/scalameta/metals/issues/787 where exactly same kind of errors had been fixed for regular `*.scala` files. Indeed, Metals does not show any errors for literal type annotations in normal project files.

**To Reproduce**
1. Add a line such as `val fortyTwo: 42 = 42` to a worksheet
2. See the error in the IDE

**Expected behavior**
No errors in the IDE.

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.8.3

mlachkar:
fixes #1591

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/metals/Trees.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/metals/Trees.scala
@@ -67,7 +67,9 @@ final class Trees {
     Option(PathIO.extension(Paths.get(fileUri))).collect {
       case "scala" => dialects.Scala
       case "sbt" => dialects.Sbt
-      case "sc" => dialects.Sbt
+      case "sc" =>
+        dialects.Scala213
+          .copy(allowToplevelTerms = true, toplevelSeparator = "")
     }
   }
 }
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala-2/scala/meta/internal/metals/Trees.scala
```diff
@@ -67,7 +67,9 @@ final class Trees {
     Option(PathIO.extension(Paths.get(fileUri))).collect {
       case "scala" => dialects.Scala
       case "sbt" => dialects.Sbt
-      case "sc" => dialects.Sbt
+      case "sc" =>
+        dialects.Scala213
```
tgodzik, line 71: We should still use `dialects.Sbt` for worksheets running in 2.12.x Scala version. You can use `
  BuildInfo.scalaCompilerVersion` to check

olafurpg: I think it's OK to use the 2.13 dialect for all worksheets. The users will still get syntax errors when they save a file that uses 2.13-only syntax with a 2.12 classpath

olafurpg: This dialect only affects the syntax errors that we display on keystroke. It's helpful for setting errors like "unbalanced parentheses" or "missing closing quote". It's not latency sensitive that the user gets a syntax errors when they for example use literal types (2.13-only syntax)

tgodzik: Yeah, you're actually right after thinking about it for a while.

Issue #1647, opened 2020-04-20
java.lang.NoClassDefFoundError: difflib/myers/Equalizer


bishabosha:
**Describe the bug**

when using metals 0.8.4+62-67ef4e74-SNAPSHOT with scala/scala repo, there is a missing class when compiling library subproject

**To Reproduce**
1. clone `git@github.com:scalacenter/scala.git`
2. checkout branch `tasty_reader`
2. open VSCode
3. Import Build
4. Open `src/compiler/scala/tools/nsc/tasty/TreeUnpickler.scala`

**Expected behavior**

~For type hints to start working~
Edit: type hints did start working, but i still see this error

**Screenshots**

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version:  0.8.4+62-67ef4e74-SNAPSHOT

**Additional context**

<details>
<summary><b>Log Output</b></summary>
<br/>
```
2020.04.20 11:05:21 INFO  compiling library (531 scala sources and 33 java sources)
[Error - 11:05:23 AM] Request textDocument/foldingRange failed.
  Message: Internal error.
  Code: -32603 
java.util.concurrent.CompletionException: java.lang.NoClassDefFoundError: difflib/myers/Equalizer
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1606)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.NoClassDefFoundError: difflib/myers/Equalizer
	at scala.meta.internal.metals.FoldingRangeProvider.$anonfun$getRangedFor$1(FoldingRangeProvider.scala:24)
	at scala.Option.map(Option.scala:242)
	at scala.meta.internal.metals.FoldingRangeProvider.getRangedFor(FoldingRangeProvider.scala:21)
	at scala.meta.internal.pc.ScalaPresentationCompiler.$anonfun$foldingRange$1(ScalaPresentationCompiler.scala:130)
[...]

PR #1657, opened 2020-04-22
Remove optimisation to reuse semanticdb jar from the classpath

tgodzik:
Previously, we would only download mtags module and reuse the semanticDB jars on the classpath, but that can cause other libraries not be fetched and cause unpredictable or easy to miss runtime issues.

Now, we fetch all the transitive dependencies of mtags.

Fixes https://github.com/scalameta/metals/issues/1647

### DIFF
--- a/tests/slow/src/test/scala/tests/feature/FoldingCrossLspSuite.scala
+++ b/tests/slow/src/test/scala/tests/feature/FoldingCrossLspSuite.scala
@@ -1,0 +1,43 @@
+package tests.feature
+
+import tests.BaseLspSuite
+import scala.meta.internal.metals.{BuildInfo => V}
+
+class FoldingCrossLspSuite extends BaseLspSuite("foldingRange-cross") {
+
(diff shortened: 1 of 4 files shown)


tests/slow/src/test/scala/tests/feature/FoldingCrossLspSuite.scala
```diff
@@ -0,0 +1,43 @@
+package tests.feature
+
+import tests.BaseLspSuite
+import scala.meta.internal.metals.{BuildInfo => V}
+
+class FoldingCrossLspSuite extends BaseLspSuite("foldingRange-cross") {
```
kpbochenek, line 6: Question: Is this test related to mentioned optimisation removal?

tgodzik: Yes actually, this issue came up when using Folding on Scala versions different than `2.12.11`

Issue #1641, opened 2020-04-17
Running when workspace not compiling causes it to hang indefinitely.


tgodzik:
**Describe the bug**
If the workspace has compilation error and we run a class the whole process my hang and it's not even possible to stop it in VS Code.

**To Reproduce**
1. Create a main class.
2. Add a configuration to `launch.json` in VS Code or equivalent.
3. Create a compile error.
4. Try to run.

**Expected behavior**
User is informed that there are compilation errors.

**Installation:**
- Operating system: macOS/Windows/Linux
- Editor: Visual Studio Code/Atom/Vim/Sublime/Emacs
- Metals version: v0.8.4

**Search terms**
run debug hang

PR #1658, opened 2020-04-23
Make sure that the workspace contains no errors when running 

tgodzik:
Previously, run/debug would hang if run in a broken workspace. Now, we display all issue about why we weren't able and make sure that we only allow run on a clean workspace.

Fixes https://github.com/scalameta/metals/issues/1641

When running in a workspace with errors:

debug-error1

When running non existing class:

debug-class-not-found

When running non exiting target:

debug-target-not-found

When running class not in a target, but the target was defined:

debug-class-in-target

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
@@ -52,6 +52,14 @@ final class ForwardingMetalsBuildClient(
   val updatedTreeViews: ju.Set[BuildTargetIdentifier] =
     ConcurrentHashSet.empty[BuildTargetIdentifier]
 
+  def buildHasErrors(buildTargetId: BuildTargetIdentifier): Boolean = {
+    buildTargets
+      .buildTargetTransitiveDependencies(buildTargetId)
+      .exists(hasReportedError.contains(_))
+  }
+
+  override def buildHasErrors: Boolean = !hasReportedError.isEmpty()
+
   def reset(): Unit = {
     cancel()
     updatedTreeViews.clear()
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
@@ -33,6 +33,12 @@ import scala.meta.internal.metals.JsonParser
 import scala.meta.internal.metals.JsonParser._
 import scala.meta.internal.metals.ClassNotFoundInBuildTargetException
 import scala.meta.internal.metals.Messages.UnresolvedDebugSessionParams
+import scala.meta.internal.metals.MetalsBuildClient
+import org.eclipse.lsp4j.ExecuteCommandParams
+import scala.meta.internal.metals.Messages
+import scala.meta.internal.metals.ClientCommands
+import scala.meta.internal.metals.StatusBar
+import scala.meta.internal.metals.BuildTargetNotFoundException
 
 class DebugProvider(
     definitionProvider: DefinitionProvider,
@@ -40,7 +46,10 @@ class DebugProvider(
     buildTargets: BuildTargets,
     buildTargetClasses: BuildTargetClasses,
     compilations: Compilations,
-    languageClient: MetalsLanguageClient
+    languageClient: MetalsLanguageClient,
+    buildClient: MetalsBuildClient,
+    messages: Messages,
+    statusBar: StatusBar
 ) {
 
   lazy val buildTargetClassesFinder = new BuildTargetClassesFinder(
@@ -97,13 +106,16 @@ class DebugProvider(
   def resolveMainClassParams(
       params: DebugUnresolvedMainClassParams
   )(implicit ec: ExecutionContext): Future[b.DebugSessionParams] = {
-    withRebuildRetry(() =>
+    val result = withRebuildRetry(() =>
       buildTargetClassesFinder
         .findMainClassAndItsBuildTarget(
           params.mainClass,
           Option(params.buildTarget)
         )
-    ).map {
+    ).flatMap {
+      case (clazz, target) :: others
+          if buildClient.buildHasErrors(target.getId()) =>
+        Future.failed(WorkspaceErrorsException)
       case (clazz, target) :: others =>
         if (others.nonEmpty) {
           reportOtherBuildTargets(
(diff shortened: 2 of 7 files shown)


metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
```diff
@@ -52,6 +52,8 @@ final class ForwardingMetalsBuildClient(
   val updatedTreeViews: ju.Set[BuildTargetIdentifier] =
     ConcurrentHashSet.empty[BuildTargetIdentifier]
 
+  def buildHasErrors = !hasReportedError.isEmpty()
```
tgodzik, line 55: I wonder if we should just check all targets that we depend on when running a particular class :thinking:

ckipp01: Probably right? Because if not the same issue will still occur won't it?

metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
```diff
@@ -52,6 +52,14 @@ final class ForwardingMetalsBuildClient(
   val updatedTreeViews: ju.Set[BuildTargetIdentifier] =
     ConcurrentHashSet.empty[BuildTargetIdentifier]
 
+  def buildHasErrors(buildTargetId: BuildTargetIdentifier): Boolean = {
+    buildTargets
+      .buildTargetTransitiveDependencies(buildTargetId)
+      .exists(id => hasReportedError.contains(id))
```
kpbochenek, line 58: if you don't find it less readable you can do `exists(hasReportedError.contains)`

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -138,24 +151,58 @@ class DebugProvider(
...
-        }
-        new b.DebugSessionParams(
-          singletonList(target.getId()),
-          b.DebugSessionParamsDataKind.SCALA_TEST_SUITES,
-          singletonList(clazz).toJson
+        //should not really happen due to
+        //`findMainClassAndItsBuildTarget` succeeding with non-empty list
+        case Nil => throw new ju.NoSuchElementException(params.testClass)
+      }
+      .recover(reportErrors)
```
kpbochenek, line 175: `failed.foreach` and above too

metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
tgodzik: It's a java set, so I can't do it the other way.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
tgodzik: Fixed

metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
tgodzik: changed to `exists(hasReportedError.contains(_))`

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -97,13 +106,16 @@ class DebugProvider(
...
+    val result = withRebuildRetry(() =>
       buildTargetClassesFinder
         .findMainClassAndItsBuildTarget(
           params.mainClass,
           Option(params.buildTarget)
         )
     ).map {
+      case (clazz, target) :: others
+          if buildClient.buildHasErrors(target.getId()) =>
+        throw WorkspaceErrorsException
```
mlachkar, line 118: why not return Future.failed(WorkspaceErrorsException) and flatMap before ?

tests/unit/src/test/scala/tests/DebugProtocolSuite.scala
```diff
@@ -174,4 +268,49 @@ class DebugProtocolSuite extends BaseLspSuite("debug-protocol") {
...
+      result <- server
+        .startDebuggingUnresolved(
+          new DebugUnresolvedTestClassParams(
+            "a.Foo"
+          ).toJson
+        )
+        .recover {
+          case WorkspaceErrorsException =>
+            server.statusBarHistory
+          case other =>
```
mlachkar, line 308: not necessary.

tests/unit/src/test/scala/tests/DebugProtocolSuite.scala
```diff
@@ -144,6 +146,98 @@ class DebugProtocolSuite extends BaseLspSuite("debug-protocol") {
...
+          new DebugUnresolvedMainClassParams(
+            "a.Main",
+            "a",
+            singletonList("Foo")
+          ).toJson
+        )
+        .recover {
+          case WorkspaceErrorsException =>
+            server.statusBarHistory
+          case other =>
```
mlachkar, line 232: not necessary

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
tgodzik: I would need to add Future() to the cases before and this does basically the same thing.

tests/unit/src/test/scala/tests/DebugProtocolSuite.scala
tgodzik: Fixed!

tests/unit/src/test/scala/tests/DebugProtocolSuite.scala
tgodzik: Fixed!

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
mlachkar: you can add Future.successful for the others. 
Isn't throwing and catching exceptions costly?

tgodzik: I don't think there is much difference. But I guess using Future.failed is more idiomatic. I will change.

PR #1661, opened 2020-04-24
Use onType formatting to create multiline string 

mlachkar:
This is just to show how we could use onTypeFormartting to do some simple actions
hello-1587727221284


mlachkar:
The code is quiet experimental. I am moving the logic of identation inside directly multilineStringInTokens2 (since we have all the necessary information here (like if there is a pipe or not, stripMargin or not ) 
I also added a new character that call onTypeFormatting, but tests works only for return to line, so we need to adapt tests for those use cases. 


So it should look like this
hello-1587731705379

tgodzik:
@mlachkar We didn't merge this, right? I am a bit confused here.

colineto:
@tgodzik it has been merged. I just rebased it's in it.

PR #1666, opened 2020-04-27
Turn off go to parent code lenses by default

tgodzik:
Turning them off until we can get the situation under control.


tgodzik:
> LGTM
> 
> We'll have to make sure to do this in the extension as well since they are defaulted to true there, which will still just turn this on right away.

Right, I should have done it right away:

https://github.com/scalameta/metals-vscode/pull/284

PR #1671, opened 2020-04-28
Update Scala 3 version to 0.23.0

tgodzik:
Let's move to a stable version before the release.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,6 +6,8 @@ def isCI = System.getenv("CI") != null
 def isScala211(v: Option[(Long, Long)]): Boolean = v.contains((2, 11))
 def isScala212(v: Option[(Long, Long)]): Boolean = v.contains((2, 12))
 def isScala213(v: Option[(Long, Long)]): Boolean = v.contains((2, 13))
+def isScala023(v: Option[(Long, Long)]): Boolean = v.contains((0, 23))
+def isScala024(v: Option[(Long, Long)]): Boolean = v.contains((0, 24))
 def isScala2(v: Option[(Long, Long)]): Boolean = v.exists(_._1 == 2)
 def isScala3(v: Option[(Long, Long)]): Boolean =
   v.exists(_._1 == 0) || v.exists(_._1 == 3)
@@ -14,11 +16,13 @@ def crossSetting[A](
     scalaVersion: String,
     if211: List[A] = Nil,
     if2: List[A] = Nil,
+    ifLaterThan211: List[A] = Nil,
     if3: List[A] = Nil
 ): List[A] =
   CrossVersion.partialVersion(scalaVersion) match {
     case partialVersion if isScala211(partialVersion) => if211 ::: if2
-    case partialVersion if isScala2(partialVersion) => if2
+    case partialVersion if isScala212(partialVersion) => ifLaterThan211 ::: if2
+    case partialVersion if isScala213(partialVersion) => ifLaterThan211 ::: if2
     case partialVersion if isScala3(partialVersion) => if3
     case _ => Nil
   }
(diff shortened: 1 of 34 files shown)


olafurpg:
MUnit currently only supports one Dotty version at a time but we can probably support the latest two versions if there is demand. The upgrade from 0.23 to 0.24 didn't require any changes in the implementation.

gabro:
I don’t see much value in supporting multiple dotty versions when the release cycle is of 6 weeks.

tgodzik:
> MUnit currently only supports one Dotty version at a time but we can probably support the latest two versions if there is demand. The upgrade from 0.23 to 0.24 didn't require any changes in the implementation.

I think we should support a single Dotty version, but if we do that I think it would be clearer to only support stable releases. It's a bit weird to only support release candidates and not the releases themselves. Otherwise, I would support last stable and last release candidate.

olafurpg:
I think it's more weird to not support the latest release since I expect most Dotty users to be early adopters that want to stay on the latest. Still, I think it's a bit rough to not give any sliding window to complete the upgrade so it might make sense to support the latest stable + candidate. That gives people 6 weeks to upgrade to the latest Dotty

tgodzik:
> I think it's more weird to not support the latest release since I expect most Dotty users to be early adopters that want to stay on the latest. Still, I think it's a bit rough to not give any sliding window to complete the upgrade so it might make sense to support the latest stable + candidate. That gives people 6 weeks to upgrade to the latest Dotty

I updated two include the last 2 Dotty versions, though now we will need the munit release. Is it ok to release now or is there anything we are waiting for?

mtags/src/main/scala-0.24/scala/scala/meta/internal/pc/CompilerInterfaces.scala
```diff
@@ -0,0 +1,94 @@
...
+import dotty.tools.dotc.reporting.Diagnostic.Error
+import dotty.tools.dotc.interfaces.Diagnostic
+import dotty.tools.dotc.util.SourceFile
+import dotty.tools.dotc.util.SourcePosition
+import dotty.tools.io.VirtualFile
+import org.eclipse.{lsp4j => l}
+
+// note(@tgodzik) the plan is to be able to move the methods here back to Dotty compiler
+// so that we can provide easier compatibility with multiple Scala 3 versions
+object CompilerInterfaces {
```
tgodzik, line 24: So this is my first attempt to move some of the methods we use to a separate class. In time we might be able to move those methods into Dotty itself and allow for better compatibility with different Scala versions.

Plus this part was incompatible between 0.23 and 0.24

olafurpg:
Here is a PR adding 0.23.0 + 0.24.0-RC1 support in MUnit https://github.com/scalameta/munit/pull/117

tgodzik:
> Here is a PR adding 0.23.0 + 0.24.0-RC1 support in MUnit [scalameta/munit#117](https://github.com/scalameta/munit/pull/117)

Thanks! I updated the PR with the support for the two versions.

build.sbt
```diff
@@ -302,7 +312,7 @@ val mtagsSettings = List(
       crossSetting(
         scalaVersion.value,
         if211 = List("com.lihaoyi" %% "pprint" % "0.5.4"),
-        if2 = List("com.lihaoyi" %% "pprint" % "0.5.9"),
+        ifLaterThan211 = List("com.lihaoyi" %% "pprint" % "0.5.9"),
```
ckipp01, line 315: 👍

mtags/src/main/scala-0.23/scala/scala/meta/internal/pc/CompilerInterfaces.scala
```diff
@@ -0,0 +1,94 @@
...
+    val freshContext = driver.currentCtx.fresh.setReporter(parsingReporter)
+    val parser = new Parser(sourceFile)(freshContext)
+    parser.parse()
+    val diags = parsingReporter.allErrors
+    diags.flatMap(diagnostic).asJava
+  }
+
+  private def diagnostic(mc: MessageContainer): Option[l.Diagnostic] =
+    if (!mc.pos.exists)
+      None // diagnostics without positions are not supported: https://github.com/Microsoft/language-server-protocol/issues/249
```
ckipp01, line 43: Probably not ideal, but it _could_ depend on what the diagnostic is, but if it is relevant to an entire file for example, we could just detect it, give it a position at the beginning of the file and display it.

tgodzik: Good idea, this message comes from Dotty though. Might have copied it with the code :O

Issue #1668, opened 2020-04-27
`InitializationOptions` and `ClientExperimentalCapablities` can result in a null pointer exceptions


ckipp01:
**Describe the bug**
We haven't hit this yet because both `coc-metals` and the VS Code extension set all the `ClientExperimentalCapabilities`. However, if you remove one of them, that value will become `null` in the case class representing it. The same happens for `InitializationOptions`. When there are no `InitializationOptions`, some clients will not even send in the key. This will result in the `IntializationOptions.Default`, which is fine. However, if the client sends in an empty `{}`, then the result will be the following:

```
InitializationOptions.scala:37 x: InitializationOptions(null, null, null, null, null, null, null, null, null, null)
```

Then as you can guess, when we check for these values, it results in a null pointer exception.

**To Reproduce**
1. Remove a value in either `coc-metals` or the VS Code exception that we do a check against or use a client that sends an empty object for either of these.

**Expected behavior**
If the key doesn't exist, the default value should get applied. It shouldn't be an all or nothing.

You can see it's an all or nothing here:

https://github.com/scalameta/metals/blob/810a09c5584642ba7a5196d538342607615b88cb/metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala#L33-L34

And also here:

https://github.com/scalameta/metals/blob/810a09c5584642ba7a5196d538342607615b88cb/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala#L34-L35

**Search terms**
clientExperimentalCapabilities, initializationOptions

ayoub-benali:
@ckipp01 I don't know how easy it would be to handle missing keys or values in Gson lib but I think these settings should also changed to `Option` with default `None` to have better semantic. 
So that when merging values from server properties, InitializationOptions, ClientExperimentalCapablities we have better context.

dangerousben:
This affects emacs lsp-mode, given the following (trimmed) init request:

```
{
  "processId": null,
  "rootPath": "/home/ben/src/typelevel/cats",
  "rootUri": "file:///home/ben/src/typelevel/cats",
  "capabilities": {
    "experimental": {
      "decorationProvider": true,
      "debuggingProvider": true,
      "didFocusProvider": true,
      "executeClientCommandProvider": true,
      "doctorProvider": "html",
      "statusBarProvider": "on"
    }
  },
  "clientInfo": {
    "name": "emacs",
[...]

PR #1674, opened 2020-04-28
Create a `ClientConfig` to contain all configuration

ckipp01:
One configuration to rule them all...

So before I go too crazy, I figured I'd put this up here.

This is my interpretation of what we talked about offline to address #1668. I think this is a nice way to instead of passing around `clientExperimentalCapabilities` the `config` and also the `initializationOptions`, I'd like to just create a `clientConfig` class that contains all three. Then all checks from that point on will go through `languageClient.clientConfiguration` which provides all of the necessary things like `isInputBoxEnabled` etc. Plus in here is where any null checks would happen.

Feedback welcome before I continue on this way.

FYI: Totally unrelated, but as I was digging in I noticed the `Message` class was a class, but didn't really get why it was a class with a single parameter that we were giving vs code icons to but not using. So it was essentially stateless with no parameters. So I just turned it into an `Object`.


kpbochenek:
What guys saying in issue that `initializationOptions` are null is correct:

When running emacs with master metals:
```
2020.04.28 20:31:59 INFO  Experimental capabilities ClientExperimentalCapabilities(null,true,true,null,null,true,null,true,null,html,on)
2020.04.28 20:31:59 INFO  Initialization options null
```

vs a line with deserialization fix from gitter:

```
2020.04.28 20:40:47 INFO  Experimental capabilities ClientExperimentalCapabilities(null,true,true,null,null,true,null,true,null,html,on)
2020.04.28 20:40:47 INFO  Initialization options InitializationOptions(off,false,false,false,false,false,html,false,false,CompilerInitializationOptions(true,true,true,true,true,true))
```

It is null because of what they correctly found here:
```
initializeParams.getInitializationOptions() match {
      case json: JsonElement =>
        json.as[InitializationOptions].getOrElse(Default)
      case _ =>
        Default
    }
```
Docs say: `   * @return an object of type T from the json. Returns {@code null} if {@code json} is {@code null}.`
You get `JsonNull` -> it is subtype of JsonElement -> convert to null -> get Try(null) -> no exception then you get `Success(null)` -> getOrElse returns `null`.
You end up with variable that you assume exists but is null.
Wrapping ` json.as[InitializationOptions]` in Option would make most sense.
For both InitializationOptions and ClientExperimentalCapabilities.

Also I would change all fields in `ClientExperimentalCapabilities` and `InitializationOptions` to be `Option[X]`
Then 
`  private def choose[T](a: T, b: T, default: T): T =`
is changed to:
`  private def choose[T](a: Option[T], b: Option[T], default: T): T =`

Then I would add tests(for both configs):
- deserialize empty string to config -> verify all options set to None
- deserialize partial config -> verify not set params are None, others are properly set

Then you might need custom deserialized for `scala.Option` it should be one liner according to docs:
https://github.com/google/gson/blob/master/UserGuide.md#writing-a-deserializer

We could also be more consistent about naming, keep all methods returning boolean be in format of `isSomethingXXX`, `isStatusBarOn` etc

`initializationOptions.statusBarIsOn == !initializationOptions.statusBarIsOff` is false I would do work here to make it act correctly. I see we use String to pass boolean value which is problematic but we could say 'On' is true, rest is false, or 'On' is true, 'Off' is false, otherwise fail(which sounds nice because if you type `of` as client in your config you get error about bad config and not silently use default - or at least log warning as we do in MetalsServerConfig)
Tried to see if this holds for `def statusBarIsOn()` for `ClientConfig`  but it is more complex :P I would expect it to also behave with logic On = !Off

ckipp01:
> What guys saying in issue that `initializationOptions` are null is correct:
> 
> When running emacs with master metals:
> 
> ```
> 2020.04.28 20:31:59 INFO  Experimental capabilities ClientExperimentalCapabilities(null,true,true,null,null,true,null,true,null,html,on)
> 2020.04.28 20:31:59 INFO  Initialization options null
> ```
> 
> vs a line with deserialization fix from gitter:
> 
> ```
> 2020.04.28 20:40:47 INFO  Experimental capabilities ClientExperimentalCapabilities(null,true,true,null,null,true,null,true,null,html,on)
> 2020.04.28 20:40:47 INFO  Initialization options InitializationOptions(off,false,false,false,false,false,html,false,false,CompilerInitializationOptions(true,true,true,true,true,true))
> ```
> 
> It is null because of what they correctly found here:
> 
> ```
> initializeParams.getInitializationOptions() match {
>       case json: JsonElement =>
>         json.as[InitializationOptions].getOrElse(Default)
>       case _ =>
>         Default
>     }
> ```
> 
> Docs say: ` * @return an object of type T from the json. Returns {@code null} if {@code json} is {@code null}.`
> You get `JsonNull` -> it is subtype of JsonElement -> convert to null -> get Try(null) -> no exception then you get `Success(null)` -> getOrElse returns `null`.
> You end up with variable that you assume exists but is null.
> Wrapping ` json.as[InitializationOptions]` in Option would make most sense.
> For both InitializationOptions and ClientExperimentalCapabilities.
> 
> Also I would change all fields in `ClientExperimentalCapabilities` and `InitializationOptions` to be `Option[X]`
> Then
> ` private def choose[T](a: T, b: T, default: T): T =`
> is changed to:
> ` private def choose[T](a: Option[T], b: Option[T], default: T): T =`
> 
> Then I would add tests(for both configs):
> 
>     * deserialize empty string to config -> verify all options set to None
> 
>     * deserialize partial config -> verify not set params are None, others are properly set
> 
> 
> Then you might need custom deserialized for `scala.Option` it should be one liner according to docs:
> [google/gson:UserGuide.md@`master`#writing-a-deserializer](https://github.com/google/gson/blob/master/UserGuide.md#writing-a-deserializer) [](/google/gson/blob/HEAD@{2020-04-29T08:24:55Z}/UserGuide.md#writing-a-deserializer)
> 
> We could also be more consistent about naming, keep all methods returning boolean be in format of `isSomethingXXX`, `isStatusBarOn` etc
> 
> `initializationOptions.statusBarIsOn == !initializationOptions.statusBarIsOff` is false I would do work here to make it act correctly. I see we use String to pass boolean value which is problematic but we could say 'On' is true, rest is false, or 'On' is true, 'Off' is false, otherwise fail(which sounds nice because if you type `of` as client in your config you get error about bad config and not silently use default - or at least log warning as we do in MetalsServerConfig)
> Tried to see if this holds for `def statusBarIsOn()` for `ClientConfig` but it is more complex :P I would expect it to also behave with logic On = !Off

Thanks a lot for this @kpbochenek. I'll work on pretty much doing all of this, switching them to Options, fixing the `JsonElement` match and work on the custom deserializer. Regarding the `On` and `Off` as string values, I believe we do that because those aren't the only options. We can also provide `log-message` and `show-message`.

ckipp01:
@kpbochenek thinking about this a bit more. I get making all the fields `Option[T]`, but I'm not sure I see the benefit of wrapping `ClientExperimentalCapabilities` and `InitializationOptions` in `Options`. We always want/need those to be set, so having them be a `None` just wouldn't work. If we could instead just catch that `null` like the person mentioned in Gitter:

```scala
  def from(
      initializeParams: l.InitializeParams
  ): InitializationOptions = {
    import scala.meta.internal.metals.JsonParser._
    initializeParams.getInitializationOptions() match {
      case _: JsonNull => Default
      case json: JsonElement =>
        json.as[InitializationOptions].getOrElse(Default)
      case _ =>
        Default
    }
  }
```
Would this accomplish the same thing? Where if the entire thing is `null` we just get the `Default` and then if individual values aren't set, then we get `None` for those?

kpbochenek:
Ahh I see all methods check the same field `statusBarProvider` right now I get it :+1:  

Yes catching `JsonNull` works equally well :+1: (even better for now)

---
Now that I think about it `json.as[InitializationOptions]` returns `Try[InitializationOptions]` and the flaw is there, as it can return you `Success(null)` which breaks whole typesafety idea. 
It should actually fail instead of returning null or return type should be `Try[Option[InitializationOptions]]`

But deserialization is used across metals and for now might be better not to touch it :)

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
```diff
@@ -4,37 +4,45 @@ import org.eclipse.{lsp4j => l}
...
-    decorationProvider: java.lang.Boolean = false,
-    inputBoxProvider: java.lang.Boolean = false,
-    quickPickProvider: java.lang.Boolean = false,
-    didFocusProvider: java.lang.Boolean = false,
-    slowTaskProvider: java.lang.Boolean = false,
-    executeClientCommandProvider: java.lang.Boolean = false,
-    openFilesOnRenameProvider: java.lang.Boolean = false,
-    doctorProvider: String = "html",
-    statusBarProvider: String = "off"
+    debuggingProvider: Option[Boolean] = Some(false),
```
ayoub-benali, line 7: I believe defaults `None` is better semantically to match the `null` from Gson, this way in `ClientConfiguration` we can better merge values, especially the none Boolean ones

ckipp01:
So I'm stuck and not loving gson at the moment. Let's take a look at the following:

```scala
import java.lang.reflect.Type
import com.google.gson.GsonBuilder
import com.google.gson.JsonDeserializationContext
import com.google.gson.JsonDeserializer
import com.google.gson.JsonElement
import com.google.gson.JsonNull

class OptionDeSerializer extends JsonDeserializer[Option[Any]] {
  override def deserialize(
      json: JsonElement,
      typeOfT: Type,
      context: JsonDeserializationContext
  ): Option[Any] = {
    json match {
      case JsonNull.INSTANCE => None
      case x: JsonElement =>
        Some(x)
      case _ => None
    }
    Some(json)
  }
}

val gsonBuilder = new GsonBuilder
gsonBuilder.registerTypeAdapter(classOf[Option[Any]], new OptionDeSerializer)
val gson = gsonBuilder.create

case class Test(a: Option[Boolean], b: Option[String])

val json = "{'a': true, 'b': 'hi'}"
val json2 = "{'a': true}"
val json3 = "{}"

gson.fromJson(json, classOf[Test]) <-- returns Test(true, "hi")
gson.fromJson(json2, classOf[Test]) <-- returns Test(true, null)
gson.fromJson(json3, classOf[Test]) <-- returns Test(null, null)
```

I can't figure out how to get a deserializer to just give me `None`. I'm a bit stuck on how to move forward with this.

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
ckipp01: Ahh good point. I'll change this... if I can actually get the `null`s to be turned into `None`s...

kpbochenek:
Yeah it looks like gson creates empty instance and fills in only fields found in json :/
You can make it work by defining default constructor:
```scala
case class Test(debug: Option[Boolean], title: Option[String]) {
  def this() = this(
    a = Some(false),
    b = Some("default_title")
  )
}
```
But! I start to dislike it more and more as it is also relies on how gson works internally :disappointed:

Maybe just a step back and @ckipp01 your initial approach was how we should do this? Agree initialConfig, experimentalCapabilities and initializationOptions can contain null fields, but they are private in scope of ClientConfiguration and nulls never leave outside of this facade?

Sorry for the confusion :disappointed:

ckipp01:
> Yeah it looks like gson creates empty instance and fills in only fields found in json :/
> You can make it work by defining default constructor:
> 
> ```scala
> case class Test(debug: Option[Boolean], title: Option[String]) {
>   def this() = this(
>     a = Some(false),
>     b = Some("default_title")
>   )
> }
> ```
> 
> But! I start to dislike it more and more as it is also relies on how gson works internally 😞
> 
> Maybe just a step back and @ckipp01 your initial approach was how we should do this? Agree initialConfig, experimentalCapabilities and initializationOptions can contain null fields, but they are private in scope of ClientConfiguration and nulls never leave outside of this facade?
> 
> Sorry for the confusion 😞

I even though I could make custom deserializers for `ClientExperimentalCapabilities` and `InitializationOptions`... but then after trying realized that if given `{}` it won't even hit the deserializer.

No worries.  I appreciate the insight, and if anything I've learned that I never want to use gson 😆 .

ckipp01:
So one other thing that's biting me right now is precedence. Without really thinking about it right now, I have everything checking `InitializationOptions` first, and the `ClientExperimental`, and finally the initial config. However, if the user sends in nothing, then they'll get the `Default` `InitializationOptions`, and with my current logic, they will always win. So I wanted to again do what @ayoub-benali mentioned with starting them all out as `None`, but then I'm still stuck with the `null` issue.

ckipp01:
So, I think this is finally ready. Took me a bit longer than expected. One of the things that still baffles me is this, which took me on a wild ride.

In a worksheet, this evaluates as follows:
```scala
val gson = new Gson()

case class Test(a: String, b: String) {
  def this() = this("no", "no")
}

val json = "{'a': 'yes'}"
gson.fromJson(json, classOf[Test]) <-- Test("yes", null)
```
However, the same thing not in a worksheet evaluates as follows:

```scala
Test("yes", "no")
```
I don't get why these are different when ran in a program or in a worksheet.

To summarize what I have done:

- Took the messages class and just make it an object
- Make default contructors for `ClientExperimentalCapabilities` and `InitializationOptions` in order to not have `null`s anymore. Thanks for the original suggestion on this @kpbochenek even though I wrote it off since it didn't work when I tried it in a worksheet 😬 . Thank you @tgodzik for trying it in a non-worksheet and saving the day.
- I centralized all configuration from server properties, `ClientExperimentalCapabilities` and `InitializationOptions` in a `Client Configuration` class. Wherever I could and where it made sense, I always tried to access setting through there.
- I removed the original `config` from a few classes that weren't using it at all.

kpbochenek:
What about this `case _: JsonNull => Default` part? is it intentionally not there anymore or got lost in a process? :P

Could you add tests for `ClientExperimentalCapabilities.from(...)` for empty string and not all properties present? checking that defined properties are actually present in resulted object and not present are defaults

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
```diff
@@ -3,18 +3,23 @@ import com.google.gson.JsonElement
...
+    inputBoxProvider: Boolean,
+    openFilesOnRenameProvider: Boolean,
+    quickPickProvider: Boolean,
+    slowTaskProvider: Boolean,
+    statusBarProvider: String,
+    treeViewProvider: Boolean
 ) {
+  def this() =
+    this(
+      false, false, false, "html", false, false, false, false, false, "off",
```
kpbochenek, line 20: they are all false so not hard to recognize what is what, but could we name them? decorationProvider = false , etc

ckipp01:
> What about this `case _: JsonNull => Default` part? is it intentionally not there anymore or got lost in a process? :P
> 

I wasn't able to ever get it to fall into this case. If we pass in `{}`, then that's a valid `JsonElement` which before resulted in `Class(null, null, ...)`. If it's not passed in at all, then it's not a `JsonElement`, and should get the default from the `_` case. I don't think adding the `JsonNull` case will change any behavior unless I'm totally missing something.

> Could you add tests for `ClientExperimentalCapabilities.from(...)` for empty string and not all properties present? checking that defined properties are actually present in resulted object and not present are defaults

I don't believe an empty string is valid for this is it? According to the spec these should all be valid `Objects` being passed in. So `""` would be invalid input right?

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
ckipp01: Ah totally. I meant to do this and forgot 👍

kpbochenek:
```
  def from(
      initializeParams: l.InitializeParams
  ): InitializationOptions = {
    import scala.meta.internal.metals.JsonParser._
    scribe.info(s"getInitializationOptions() => ${initializeParams.getInitializationOptions()}")
    initializeParams.getInitializationOptions() match {
      case json: JsonElement =>
        json.as[InitializationOptions].getOrElse(Default)
      case _ =>
        Default
    }
  }
```

```
2020.05.01 13:39:48 INFO  getExperimental {"decorationProvider":true,"debuggingProvider":null,"didFocusProvider":true,"executeClientCommandProvider":true,"doctorProvider":"html","statusBarProvider":"on","treeViewProvider":true}
2020.05.01 13:39:48 INFO  getInitializationOptions() => null
2020.05.01 13:39:48 INFO  time: initialize in 0.31s
May 01, 2020 1:39:48 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint fallbackResponseError
SEVERE: Internal error: java.lang.NullPointerException
java.util.concurrent.CompletionException: java.lang.NullPointerException
...
```

Emacs :heart:

ckipp01:
> ```
>   def from(
>       initializeParams: l.InitializeParams
>   ): InitializationOptions = {
>     import scala.meta.internal.metals.JsonParser._
>     scribe.info(s"getInitializationOptions() => ${initializeParams.getInitializationOptions()}")
>     initializeParams.getInitializationOptions() match {
>       case json: JsonElement =>
>         json.as[InitializationOptions].getOrElse(Default)
>       case _ =>
>         Default
>     }
>   }
> ```
> 
> ```
> 2020.05.01 13:39:48 INFO  getExperimental {"decorationProvider":true,"debuggingProvider":null,"didFocusProvider":true,"executeClientCommandProvider":true,"doctorProvider":"html","statusBarProvider":"on","treeViewProvider":true}
> 2020.05.01 13:39:48 INFO  getInitializationOptions() => null
> 2020.05.01 13:39:48 INFO  time: initialize in 0.31s
> May 01, 2020 1:39:48 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint fallbackResponseError
> SEVERE: Internal error: java.lang.NullPointerException
> java.util.concurrent.CompletionException: java.lang.NullPointerException
> ...
> ```
> 
> Emacs ❤️

So the experimental `null` was because awesomely I forgot a default 😬, which I caught when naming the parameters. As for the `getInitializationOptions` being `null` here. That shouldn't matter right? In this case, it's won't match on the `JsonElement`, so it should fall into the bottom case right? Are we sure that NPE didn't come from the `null` that snuck into experimental. Like if you pull again and try do you still get it? I need to spin up emacs... painful to say as a vimmer.

kpbochenek:
ahh wait I can test that on your branch, give me a sec :D

ckipp01:
> ahh wait I can test that on your branch, give me a sec :D

I think this should be fine. For example, `coc-metals` does the same thin, it passes in no `InitializationParams` which result in a null.

```
2020.05.01 14:47:25 INFO  getInitializationOptions() => null
```
And I don't get a NPE. I think it was the other `null` from the missing parameter that was causing it.

kpbochenek:
Yes I still get NPE.
`In this case, it's won't match on the JsonElement` -> It actually matches JsonElement :)

ckipp01:
> Yes I still get NPE.
> `In this case, it's won't match on the JsonElement` -> It actually matches JsonElement :)

🤔 I'm so confused why this doesn't give me an NPE then for `coc-metal`. Even if it is a `JsonElement` wouldn't it not have any keys, and then just get all the defaults?

ckipp01:
> Yes I still get NPE.
> `In this case, it's won't match on the JsonElement` -> It actually matches JsonElement :)

Is there a chance somehow that these `null`s are different? I just tested again, and the `null` that I'm seeing is not falling into the `JsonElement`.

kpbochenek:
Checked more, and whatever emacs sends or doesn't send it ends up as `JsonNull`:
```
2020.05.01 14:11:59 INFO  getInitializationOptions() => null
ClassOf class com.google.gson.JsonNull
```
:exploding_head:

tgodzik:
> Checked more, and whatever emacs sends or doesn't send it ends up as `JsonNull`:
> 
> ```
> 2020.05.01 14:11:59 INFO  getInitializationOptions() => null
> ClassOf class com.google.gson.JsonNull
> ```

Maybe the difference if in one case the field is never sent and in the other it's sent as null?

ckipp01:
So I'm at a total loss as to why, but we have two cases.

1. `coc-metal` leaves out the `InitializationOptions` key, which results in a `null`. However, this `null` is not recognized as a `JsonElement`. This makes some sense I think.
2. emacs 🙈 also doesn't send this key in and it seems to parse as a `null`, _but_ for some reason, it _does_ match as `JsonElement`.

Some details to back up my crazy claim:

Emacs:

As you see... no `InitalizationOptions`. I can paste the entire thing if you want, but they aren't in here.
```
  "processId": 66672,
  "rootPath": "/Users/ckipp/Documents/scala-workspace/metals-minimal-sbt",
  "rootUri": "file:///Users/ckipp/Documents/scala-workspace/metals-minimal-sbt",
  "capabilities": {
    "workspace": {
      "applyEdit": true,
      "workspaceEdit": {
        "documentChanges": true,
        "resourceOperations": [
          "create",
          "rename",
          "delete"
        ]
      },
...
```

This will be processed by lsp4j as the following:

```
 processId = 68596
  rootPath = "/Users/ckipp/Documents/scala-workspace/metals-minimal-sbt"
  rootUri = "file:///Users/ckipp/Documents/scala-workspace/metals-minimal-sbt"
  initializationOptions = null
```
This `null` will fall into the `JsonElement` case...

Vim:
```
[Trace - 03:25:36 PM] Received request 'initialize - (0)'
Params: {
  "processId": 67344,
  "rootPath": "/Users/ckipp/Documents/scala-workspace/metals-minimal-sbt",
  "rootUri": "file:///Users/ckipp/Documents/scala-workspace/metals-minimal-sbt",
  "capabilities": {
    "workspace": {
      "applyEdit": true,
      "workspaceEdit": {
        "documentChanges": true,
        "resourceOperations": [
          "create",
          "rename",
          "delete"
 ...
```
Pretty much the same as emacs... no `InitializationOption`. This will be processed by lsp4j as the following:

```
 processId = 68433
  rootPath = "/Users/ckipp/Documents/scala-workspace/metals-minimal-sbt"
  rootUri = "file:///Users/ckipp/Documents/scala-workspace/metals-minimal-sbt"
  initializationOptions = null
```
Looks familiar right? _However_, this will not fall into the `JsonElement` case, but the catch all.

I have no idea why these `null`s are different. I'm not even sure how to write a test to catch these as according to what is being passed in, neither are passing in `InitializationOptions`, so it makes no sense why the `null`s are different.


/shrug

tgodzik:
@ckipp01 I think it's fine if we just add the case for null, I wouldn't bother about it too much,

ckipp01:
> @ckipp01 I think it's fine if we just add the case for null, I wouldn't bother about it too much,

Sounds good. This should be ready to go then.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -238,21 +238,19 @@ class MetalsLanguageServer(
     scribe.info(
       s"started: Metals version ${BuildInfo.metalsVersion} in workspace '$workspace'"
     )
-    clientExperimentalCapabilities =
-      ClientExperimentalCapabilities.from(params.getCapabilities)
-    initializationOptions = InitializationOptions.from(params)
 
-    languageClient.configure(clientExperimentalCapabilities)
-    languageClient.configure(initializationOptions)
+    languageClient.configure(
```
tgodzik, line 242: Maybe we can just do:
```
 clientConfig.clientExperimentalCapabilities= ClientExperimentalCapabilities.from(params.getCapabilities)
 clientConfig.initializationOptions = InitializationOptions.from(params)
```

we will not need the additional configure methods.

tests/unit/src/test/scala/tests/BaseTablesSuite.scala
```diff
@@ -15,7 +18,15 @@ abstract class BaseTablesSuite extends BaseSuite {
   override def beforeEach(connect: BeforeEach): Unit = {
     workspace = AbsolutePath(Files.createTempDirectory("metals"))
     time.reset()
-    tables = new Tables(workspace, time, MetalsServerConfig())
+    tables = new Tables(
+      workspace,
+      time,
+      new ClientConfiguration(
```
tgodzik, line 24: We could have `ClientConfiguration.Default`, which would set up everything.

tests/unit/src/test/scala/tests/StatusBarSuite.scala
```diff
@@ -3,27 +3,28 @@ package tests
...
-  import scala.meta.internal.metals.StatusBarConfig
   var status = new StatusBar(
     () => client,
     time,
     ProgressTicks.dots,
-    Icons.default,
-    StatusBarConfig.default,
-    ClientExperimentalCapabilities.Default,
-    InitializationOptions.Default
+    new ClientConfiguration(
```
tgodzik, line 23: `ClientConfiguration.Default` ?

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
```diff
@@ -1,15 +1,15 @@
...
 import scala.concurrent.Promise
+import scala.meta.internal.metals.MetalsSlowTaskResult
+import scala.meta.internal.metals.ClientExperimentalCapabilities
 
 abstract class BaseWorksheetLspSuite(scalaVersion: String)
     extends BaseLspSuite("worksheet") {
   override def experimentalCapabilities
       : Option[ClientExperimentalCapabilities] =
-    Some(ClientExperimentalCapabilities(decorationProvider = true))
+    Some(ClientExperimentalCapabilities.Default.copy(decorationProvider = true))
```
tgodzik, line 12: `new ClientExperimentalCapabilities(decorationProvider = true) ` ?

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
```diff
@@ -6,17 +6,16 @@ import scala.meta.internal.metals.MetalsEnrichments._
...
 import munit.TestOptions
 import java.nio.file.FileAlreadyExistsException
+import scala.meta.internal.metals.ClientExperimentalCapabilities
 
 class NewFilesLspSuite extends BaseLspSuite("new-files") {
   override def experimentalCapabilities
       : Option[ClientExperimentalCapabilities] =
-    Some(ClientExperimentalCapabilities(inputBoxProvider = true))
-
+    Some(ClientExperimentalCapabilities.Default.copy(inputBoxProvider = true))
```
tgodzik, line 18: `new ClientExperimentalCapabilities(inputBoxProvider = true)` ?

tests/unit/src/main/scala/tests/BaseLspSuite.scala
```diff
@@ -19,6 +18,7 @@ import scala.meta.internal.metals.SlowTaskConfig
 import scala.util.control.NonFatal
 import munit.Ignore
 import munit.Location
+import scala.meta.internal.metals.ExecuteClientCommandConfig
```
tgodzik, line 21: Revert the changes in this file maybe?

metals/src/main/scala/scala/meta/internal/metals/JsonParser.scala
```diff
@@ -1,11 +1,11 @@
 package scala.meta.internal.metals
 
-import com.google.gson.Gson
```
tgodzik, line 3: We can revert changes in this file.

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
ckipp01: Yea, this won't work anymore because of the type of constructor we have.

tests/unit/src/test/scala/tests/StatusBarSuite.scala
ckipp01: Yea that's a good idea since we do this is a couple places. I'll go ahead and add this in.

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
tgodzik: Ach, yeah. You are right!

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
ckipp01: Ahhh, that's true.

ckipp01:
Unrelated test failure only on windows so I'm going to go ahead and merge.

```
[error] 	tests.DebugProtocolSuite
```

olafurpg:
@ckipp01 The difference between worksheets and main apps is that worksheets wrap the code inside an inner class, which looks like GSON is unable to reflectively load
```scala
❯ cat foo.scala
object a extends App {
  import com.google.gson.Gson
  val gson = new Gson()
  val json = "{'a': 'yes'}"
  class InnerClass {
    case class Test1(a: String, b: String) {
      def this() = this("no", "no")
    }
    def example1 = gson.fromJson(json, classOf[Test1])
  }
  case class Test2(a: String, b: String) {
    def this() = this("no", "no")
  }
  def example2 = gson.fromJson(json, classOf[Test2])
  println(new InnerClass().example1)
  println(example2)
}
❯ cs launch scala -- -cp $(cs fetch -p com.google.code.gson:gson:2.8.6) foo.scala
Test1(yes,null)
Test2(yes,no)
```

Issue #1354, opened 2020-01-26
Missing bloop disconnection alert when exiting bloop


jvican:
**Describe the bug**
Metals doesn't detect Bloop server has disconnected.

**To Reproduce**
1. Open a project with Metals, open `build.sbt` and let Metals connect to Bloop. Don't open a Scala or Java source file.
2. In a terminal, run `bloop exit`.
3. Switch back to Metals, open a Scala file. Metals will "Load the presentation compiler".
4. When you try to do something like go to definition, Metals will say something like "Wait until compilation is finished and try again".

**Expected behavior**
I expect Metals to detect the server connection is not up anymore and that it connects to it automatically or reports me with a connection error.

**Installation:**
- Operating system: macOS
- Editor: vim / coc-metals
- Metals version: v0.8.0

**Search terms**
bloop, disconnection, exit, reconnection, server

ckipp01:
Just hit on this today. I'm plan to start working a fix for this.

tgodzik:
> Just hit on this today. I'm plan to start working a fix for this.

If you have any issues with that let me know, I might be able to help out.

ckipp01:
Sure! I haven't really touched any of the bloop related stuff yet, so I'm taking it as an opportunity to familiarize myself with it. I'll probably follow-up with you when I get stuck 😆

PR #1677, opened 2020-04-29
Reworked Bloop connection and Tree View

tgodzik:
Previously, we would only reconnect if the user asked for it and often this would not show up quickly.
Now, we make the reconnection work better by:
- reconnecting as soon as the build server is disconnected (fixes https://github.com/scalameta/metals/issues/1354)
- making reconnection automatic unless the user specifically asks for it not to be via system property `metals.ask-to-reconnect=true`
- moving compilation outside the initial build server connection - longer running compilations would then produce invalid exceptions and break the next connection
- reindexing and compiling the workspace on reconnection (this is needed since workspaces would not have all the classpath updated)

Also in this PR:
-  a new clean-compile command was introduced, which together with the restart bloop command was added to tree view
- tree view now has 4 parts - one for help, one for project view, one for commands and one for running compilations

The reworked tree view would look like this:

image

I am also thinking of:
- adding icons to compiling projects, I've seen it break in some cases
- partly reindexing the workspace - not sure if that is sensible since Bloop has a different directory for each client and when reconnecting we are effectively a different client

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
@@ -29,7 +29,8 @@ final class BloopServers(
     workspace: AbsolutePath,
     client: MetalsBuildClient,
     languageClient: LanguageClient,
-    tables: Tables
+    tables: Tables,
+    config: MetalsServerConfig
 )(implicit ec: ExecutionContextExecutorService) {
 
   def shutdownServer(): Boolean = {
@@ -61,7 +62,8 @@ final class BloopServers(
         client,
         languageClient,
         () => connectToLauncher(bloopVersion),
-        tables
+        tables,
+        config
       )
       .map(Option(_))
   }
@@ -96,6 +98,7 @@ final class BloopServers(
         serverStarted
       )
 
+    val finished = Promise[Unit]()
     val job = ec.submit(new Runnable {
       override def run(): Unit = {
         launcher.runLauncher(
@@ -103,6 +106,7 @@ final class BloopServers(
           skipBspConnection = false,
           Nil
         )
+        finished.success(())
       }
     })
 
@@ -117,7 +121,8 @@ final class BloopServers(
             clientOut.close()
           },
           Cancelable(() => job.cancel(true))
-        )
+        ),
+        finished
       )
     }
   }
--- a/metals/src/main/scala/scala/meta/internal/metals/BspServers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BspServers.scala
@@ -14,6 +14,7 @@ import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.mtags.MD5
 import scala.meta.io.AbsolutePath
 import scala.util.Try
+import scala.concurrent.Promise
 
 /**
  * Implements BSP server discovery, named "BSP Connection Protocol" in the spec.
@@ -26,7 +27,8 @@ final class BspServers(
     client: MetalsLanguageClient,
     buildClient: MetalsBuildClient,
     tables: Tables,
-    bspGlobalInstallDirectories: List[AbsolutePath]
+    bspGlobalInstallDirectories: List[AbsolutePath],
+    config: MetalsServerConfig
 )(implicit ec: ExecutionContextExecutorService) {
 
   def newServer(): Future[Option[BuildServerConnection]] = {
@@ -70,6 +72,12 @@ final class BspServers(
         s"${details.getName} input stream"
       )
 
+      val finished = Promise[Unit]()
+      Future {
+        process.waitFor()
+        finished.success(())
+      }
+
       Future.successful {
         SocketConnection(
           details.getName(),
@@ -77,7 +85,8 @@ final class BspServers(
           input,
           List(
             Cancelable(() => process.destroy())
-          )
+          ),
+          finished
         )
       }
 
@@ -88,7 +97,8 @@ final class BspServers(
       buildClient,
       client,
       newConnection,
-      tables
+      tables,
+      config
     )
   }
 
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
@@ -18,10 +18,12 @@ import scala.meta.io.AbsolutePath
 import scala.util.Try
 import com.google.gson.Gson
 import MetalsEnrichments._
-import java.io.IOException
 import org.eclipse.lsp4j.services.LanguageClient
 import java.util.concurrent.atomic.AtomicReference
+import scala.concurrent.Promise
+import scala.concurrent.ExecutionContext
 import org.eclipse.lsp4j.jsonrpc.JsonRpcException
+import java.io.IOException
 
 /**
  * An actively running and initialized BSP connection.
@@ -32,11 +34,20 @@ class BuildServerConnection private (
     ],
     initialConnection: BuildServerConnection.LauncherConnection,
     languageClient: LanguageClient,
-    tables: Tables
+    tables: Tables,
+    config: MetalsServerConfig
 )(implicit ec: ExecutionContextExecutorService)
     extends Cancelable {
 
   @volatile private var connection = Future.successful(initialConnection)
+  initialConnection.onConnectionFinished(reconnect)
+
+  private val isShuttingDown = new AtomicBoolean(false)
+  private val onReconnection =
+    new AtomicReference[BuildServerConnection => Future[Unit]](_ =>
+      Future.successful(())
+    )
+
   private val _version = new AtomicReference(initialConnection.version)
 
   private val ongoingRequests =
@@ -47,13 +58,21 @@ class BuildServerConnection private (
   // the name is set before when establishing conenction
   def name: String = initialConnection.socketConnection.serverName
 
+  def onReconnection(
+      index: BuildServerConnection => Future[Unit]
+  ): Unit = {
+    onReconnection.set(index)
+  }
+
   /** Run build/shutdown procedure */
   def shutdown(): Future[Unit] = connection.map { conn =>
     try {
-      conn.server.buildShutdown().get(2, TimeUnit.SECONDS)
-      conn.server.onBuildExit()
-      // Cancel pending compilations on our side, this is not needed for Bloop.
-      cancel()
+      if (isShuttingDown.compareAndSet(false, true)) {
+        conn.server.buildShutdown().get(2, TimeUnit.SECONDS)
+        conn.server.onBuildExit()
+        // Cancel pending compilations on our side, this is not needed for Bloop.
+        cancel()
+      }
     } catch {
       case e: TimeoutException =>
         scribe.error(
@@ -70,6 +89,10 @@ class BuildServerConnection private (
 
   def compile(params: CompileParams): CompletableFuture[CompileResult] = {
     register(server => server.buildTargetCompile(params))
+  }
+
+  def clean(params: CleanCacheParams): CompletableFuture[CleanCacheResult] = {
+    register(server => server.buildTargetCleanCache(params))
   }
 
   def mainClasses(
@@ -118,21 +141,49 @@ class BuildServerConnection private (
   }
 
   private def askUser(): Future[BuildServerConnection.LauncherConnection] = {
-    val notification = tables.dismissedNotifications.ReconnectBsp
-    if (!notification.isDismissed) {
-      val params = Messages.DisconnectedServer.params()
-      languageClient.showMessageRequest(params).asScala.flatMap {
-        case response if response == Messages.DisconnectedServer.reconnect =>
-          reestablishConnection()
-        case response if response == Messages.DisconnectedServer.notNow =>
-          notification.dismiss(5, TimeUnit.MINUTES)
-          connection
+    if (config.askToReconnect) {
+      val notification = tables.dismissedNotifications.ReconnectBsp
+      if (!notification.isDismissed) {
+        val params = Messages.DisconnectedServer.params()
+        languageClient.showMessageRequest(params).asScala.flatMap {
+          case response if response == Messages.DisconnectedServer.reconnect =>
+            reestablishConnection()
+          case response if response == Messages.DisconnectedServer.notNow =>
+            notification.dismiss(5, TimeUnit.MINUTES)
+            connection
+          case _ =>
+            connection
+        }
+      } else {
+        connection
+      }
+    } else {
+      reestablishConnection()
+    }
+  }
+
+  private def reconnect(): Future[BuildServerConnection.LauncherConnection] = {
+    val original = connection
+    if (!isShuttingDown.get()) {
+      synchronized {
+        // if the future is different then the connection is already being reestablished
+        if (connection eq original) {
+          connection = askUser().map { conn =>
+            // version can change when reconnecting
+            _version.set(conn.version)
+            ongoingRequests.addAll(conn.cancelables)
+            conn.onConnectionFinished(reconnect)
+            conn
+          }
+          connection.foreach(_ => onReconnection.get()(this))
+        }
+        connection
       }
     } else {
       connection
     }
-  }
 
+  }
   private def register[T](
       action: MetalsBuildServer => CompletableFuture[T]
   ): CompletableFuture[T] = {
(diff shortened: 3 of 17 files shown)


tgodzik:
@nilskp I changed the TreeView to make it more explicit that the commands are clickable and add two more useful commands. Will this be helpful?

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -96,13 +98,15 @@ final class BloopServers(
         serverStarted
       )
 
+    val finished = Promise[Unit]()
```
tgodzik, line 101: This promise is used to determine whether the launcher is finished.

metals/src/main/scala/scala/meta/internal/metals/BspServers.scala
```diff
@@ -70,14 +72,21 @@ final class BspServers(
         s"${details.getName} input stream"
       )
 
+      val finished = Promise[Unit]()
+      Future {
+        process.waitFor()
```
tgodzik, line 77: Not 100% sure about this solution, need to test it with another BSP server.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -32,11 +34,19 @@ class BuildServerConnection private (
...
     initialConnection: BuildServerConnection.LauncherConnection,
     languageClient: LanguageClient,
-    tables: Tables
+    tables: Tables,
+    config: MetalsServerConfig
 )(implicit ec: ExecutionContextExecutorService)
     extends Cancelable {
 
   @volatile private var connection = Future.successful(initialConnection)
+  connection.foreach { conn => conn.onConnectionFinished(reconnect) }
```
tgodzik, line 43: This uses the promise from the launcher to reconnect when the connection is finished.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -47,6 +57,12 @@ class BuildServerConnection private (
   // the name is set before when establishing conenction
   def name: String = initialConnection.socketConnection.serverName
 
+  def onReconnection(
+      index: BuildServerConnection => Future[Unit]
+  ): Unit = {
+    onReconnection.set(index)
```
tgodzik, line 63: We need to reindex the workspace and this is to make sure that we will only try to reindex once we finished the indexing at least once.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -118,21 +138,45 @@ class BuildServerConnection private (
...
-    val notification = tables.dismissedNotifications.ReconnectBsp
-    if (!notification.isDismissed) {
-      val params = Messages.DisconnectedServer.params()
-      languageClient.showMessageRequest(params).asScala.flatMap {
-        case response if response == Messages.DisconnectedServer.reconnect =>
-          reestablishConnection()
-        case response if response == Messages.DisconnectedServer.notNow =>
-          notification.dismiss(5, TimeUnit.MINUTES)
-          connection
+    if (config.askToReconnect) {
```
tgodzik, line 141: Depends now on the user configuration.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -118,21 +138,45 @@ class BuildServerConnection private (
...
+      // if the future is different then the connection is already being reestablished
+      if (connection eq original) {
+        connection = askUser().map { conn =>
+          // version can change when reconnecting
+          _version.set(conn.version)
+          ongoingRequests.addAll(conn.cancelables)
+          conn.onConnectionFinished(reconnect)
+          conn
+        }
+        connection.foreach(_ => onReconnection.get()(this))
```
tgodzik, line 174: Here we actually specify that we want to reindex.

metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
```diff
@@ -66,6 +66,26 @@ final class Compilations(
     cascadeBatch.cancelCurrentRequest()
   }
 
+  def recompile(): Future[Unit] = {
```
tgodzik, line 69: Used to fully recompile a workspace.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1483,33 +1487,56 @@ class MetalsLanguageServer(
...
-      disconnectOldBuildServer()
-      val message =
-        "Failed to connect with build server, no functionality will work."
-      val details = " See logs for more details."
-      languageClient.showMessage(
-        new MessageParams(MessageType.Error, message + details)
-      )
-      scribe.error(message, e)
-      BuildChange.Failed
+      .flatMap(compileAllOpen)
```
tgodzik, line 1539: Moved compiling after recover, since this caused to disconnect the build server if anything happened during the compilation.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1483,33 +1487,56 @@ class MetalsLanguageServer(
...
+        maybeBuild <- timed("connected to build server") {
+          if (buildTools.isBloop) bloopServers.newServer(userConfig)
+          else bspServers.newServer()
+        }
+        result <- maybeBuild match {
+          case Some(build) =>
+            val result = connectToNewBuildServer(build)
+            build.onReconnection { reconnected =>
+              connectToNewBuildServer(reconnected)
+                .flatMap(compileAllOpen)
```
tgodzik, line 1516: We also compile, which will effectively just move all the old artifacts to the new BSP directory.

metals/src/main/scala/scala/meta/internal/tvp/TreeViewNode.scala
```diff
@@ -24,18 +24,21 @@ case class TreeViewNode(
...
+      icon: String = TreeViewNode.command
+  ): TreeViewNode =
     TreeViewNode(
       viewId = "commands",
       nodeUri = s"metals://command/${command.id}",
       label = command.title,
       command = MetalsCommand(
         command.title,
-        command.id,
+        "metals." + command.id,
```
tgodzik, line 37: Somehow this didn't work with the new command, since the ones declared all have the `metals.` prefix

This might not be needed, but I haven't been able to make the new command work otherwise.

I think we could also just use an echo command for all server commands here.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -66,6 +66,10 @@ final case class MetalsServerConfig(
       "bloop.embedded.version",
       BuildInfo.bloopVersion
     ),
+    askToReconnect: Boolean = MetalsServerConfig.binaryOption(
```
ckipp01, line 69: Instead of introducing this as a server property, can we just introduce it in `InitializationOptions`?

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -76,6 +76,16 @@ object ServerCommands {
        |""".stripMargin
   )
 
+  val CleanCompile = new Command(
+    "compile-clean",
+    "Recompile workspace",
+    """|Recompile of build targets in this workspace.
```
ckipp01, line 82:
```suggestion
    """|Recompile all build targets in this workspace.
```

ckipp01:
> @ckipp01 Could you maybe take a look at this from the coc-metals perspective? Not sure if a lot of changes will be need or whether I should do something differently.

First try shows the `MetalsBuild` section just fine, but the `MetalsCompile` and is empty and `MetalsProject` doesn't show up :/. Not sure if it's just a small thing or a larger thing that needs to get fixed on the coc-metals side without diving further in.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
tgodzik: Client editor should never set this. It's only to be set by a user that knows what they are doing. Usually via `.jvmopts` file in the workspace. The default is fine for all clients.

tgodzik:
Seems to be an issue with shutting down currently. The connection might get automatically reestablised :sweat_smile:  Will fix it tomorrow.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
ckipp01: ahhh ok, that makes sense. 👍

tgodzik:
> > @ckipp01 Could you maybe take a look at this from the coc-metals perspective? Not sure if a lot of changes will be need or whether I should do something differently.
> 
> First try shows the `MetalsBuild` section just fine, but the `MetalsCompile` and is empty and `MetalsProject` doesn't show up :/. Not sure if it's just a small thing or a larger thing that needs to get fixed on the coc-metals side without diving further in.

Project just changed the name, so maybe we can adjust the names of views too? MetalsCompile should be empty unless we are compiling anything.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala

metals/src/main/scala/scala/meta/internal/metals/BspServers.scala
tgodzik: Added a test for it in Bill Suite, seems to work pretty well.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
olafurpg: I'm not sure I understand the logic here, what is the difference between this and the logic below?

```suggestion
  initialConnection.onConnectionFinished(reconnect)
```

tgodzik: `onConnectionFinished` is coming from the process or the launcher to say that the connection is finished.  `onReconnection` is used to later reindex the workspace.

The alternative is to do it in the MetalsLanguageServer and just do a full reconnect, but then it will be less seamless for the user.

tgodzik:
> Should "Recompile workspace" be called "Clean compile workspace" instead?

I was thinking about it, but recompile seemed to be clear that we are redoing the whole compilation. I can change don't feel strongly about it

> Should we use the sync icon  for "Import build" for consistency with IntelliJ "Refresh project"? We can use something like a stop icon  for "Restart build server" instead.

Good idea! I wasn't really sure about those two.

> Can we move "Restart build server" to the bottom of the command list in the tree view?

Sure!

> (brainstorming, maybe for a separate PR) I wonder if we should have some form of "Clean import build" which removes `.bloop/` before importing the build  I worry however that we end up having too many troubleshooting commands and users blindly run them all.

Honestly, I have never needed to do that myself. Either clean compile or restart server should help with most situations. There is no need to remove the .json files in .bloop ever, so the only thing we need to do is remove the compile artifacts, which the clean compile should do.

tgodzik:
@olafurpg  I updated the icons for import and restart server. I used `debug-stop` for restart, since stop gives us the same icon as for cancel. I also updated the VSCode PR so that if we don't find an icon we try to use the default ones from VS Code.

The image in the description got updated too.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
tgodzik: I think this is a bit better than a full reconnect, because we will not lose any requests to the build server.

metals/src/main/scala/scala/meta/internal/tvp/MetalsTreeViewProvider.scala
```diff
@@ -75,6 +76,12 @@ class MetalsTreeViewProvider(
     languageClient.metalsTreeViewDidChange(
       TreeViewDidChangeParams(
         Array(
+          TreeViewNode(
```
ckipp01, line 79: Sort of nitpicky, but I'm trying to debug what is going on with `coc-metals` with the branch, and one things I noticed is that this is _technically_ against the spec that we have published. In the `TreeViewNode` class, `viewId` and `lable` aren't marked as `@Nullable`, and in the [spec](https://scalameta.org/metals/docs/editors/tree-view-protocol.html#tree-view-node) it's the same thing. I'm getting an undefined that I'm trying to chase down, and I don't think it's coming from here, but this is the sort of thing that could cause it and confuse someone since there should be somewhat of a guarantee that there is a `label` here. The same for the other two `TreeViewNode`'s below this.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
marek1840: then maybe it could be called: `launched` instead of `finished`?

metals/src/main/scala/scala/meta/internal/tvp/MetalsTreeViewProvider.scala
```diff
@@ -299,18 +317,16 @@ class MetalsTreeViewProvider(
...
+      icon = "compile"
     )
   }
 
   private def ongoingCompilationNode: TreeViewNode = {
-    val size = compilations().size
-    val counter = if (size > 0) s" ($size)" else ""
     TreeViewNode(
       Compile,
-      "metals://ongoing-compilations",
```
tgodzik, line 311: @ckipp01 Maybe this part is a problem with the `undefined` ? I hoped the node is discovered by id, but maybe we check something else in vim? This is the only big change, the rest is just moving things around

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -47,9 +58,16 @@ class BuildServerConnection private (
...
+  def onReconnection(
+      index: BuildServerConnection => Future[Unit]
+  ): Unit = {
+    onReconnection.set(index)
+  }
+
   /** Run build/shutdown procedure */
   def shutdown(): Future[Unit] = connection.map { conn =>
     try {
+      isShuttingDown.set(true)
```
marek1840, line 70: by using `compareAndSet(false, true` you could guarantee here that the method will be called exactly once, hence making it idempotent.

metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
marek1840: if the `fully` part is more important, then maybe it should be called `compileAll` ?

metals/src/main/scala/scala/meta/internal/tvp/MetalsTreeViewProvider.scala
ckipp01: No, I found the issue. The TVP panel in coc-metals was sort of built with the assumption that apart from the initial view it was never empty. Not that things have been shifted around the the compilations view will return an empty array just isn't accounted for. So I need to adjust the logic for this since I have logic built-in to find the offset of node to update, which is wrong now.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1483,33 +1488,56 @@ class MetalsLanguageServer(
...
-      }
-      result <- maybeBuild match {
-        case Some(build) =>
-          connectToNewBuildServer(build)
-        case None =>
-          Future.successful(BuildChange.None)
-      }
-      _ = {
-        treeView.init()
+    def compileAllOpen: BuildChange => Future[BuildChange] = {
```
marek1840, line 1491: could be `compileOpenFiles` at first I was not 100% if it meant open files.

metals/src/main/scala/scala/meta/internal/tvp/MetalsTreeViewProvider.scala
tgodzik: Ok, cool. Thanks for taking a look at it!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: `finished` is updated when the launcher finishes and not when it launches, so I think this is better.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
tgodzik: Good idea! Fixed it!

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: changed to `compileAllOpenFiles`

metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
tgodzik: Changed to `recompileAll`

metals/src/main/scala/scala/meta/internal/tvp/MetalsTreeViewProvider.scala
tgodzik: Added TreeViewNode.empty that sets both the id and label.

PR #1684, opened 2020-05-03
Move additional settings to InitializationOptions

ckipp01:
Previously debuggingProvider, decorationProvider, openFilesOnRenameProvider,
and treeViewProvider were only able to be set via `ClientExperimentalCapabilities`.
Now these are also able to be set via `InitializationOption`, which for some clients
are much easier to set. This is the case for the built-in LSP support in Nvim

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ClientConfiguration.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ClientConfiguration.scala
@@ -55,7 +55,8 @@ class ClientConfiguration(
       experimentalCapabilities.quickPickProvider
 
   def isOpenFilesOnRenameProvider(): Boolean =
-    experimentalCapabilities.openFilesOnRenameProvider ||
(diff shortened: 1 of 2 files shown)


ayoub-benali:
Shouldn't the doc be updated ? at least here: https://github.com/scalameta/metals/blob/master/docs/editors/new-editor.md#initializationoptions-and-experimental

ckipp01:
> Shouldn't the doc be updated ? at least here: [/docs/editors/new-editor.md@`master`#initializationoptions-and-experimental](https://github.com/scalameta/metals/blob/master/docs/editors/new-editor.md#initializationoptions-and-experimental) [](/scalameta/metals/blob/HEAD@{2020-05-04T09:33:05Z}/docs/editors/new-editor.md#initializationoptions-and-experimental)

Ahhh, good call @ayoub-benali, I totally missed that. I'll send in a pr to update these. Thanks for pointing it out!

Issue #1660, opened 2020-04-06
Code action or completion to change a single line string into multiline one


tgodzik:
**Is your feature request related to a problem? Please describe.**
It's pretty difficult currently to write a multiline string, since quotes are always completed in twos and 
everything is a bit finicky.
 
**Describe the solution you'd like**
It would be good to allow completions (maybe after writing the first two quotes) that would allow to insert `"""|""".stripMargin` or just `"""`. 

**Describe alternatives you've considered**
Add a code action on a single string to convert it to multiline.

**Additional context**

**Search terms**:
multiline stripMargin quotes triple

tgodzik:
@coline-lunatech invited you to the repository, we can assign the issue to you after you accept :)

kpbochenek:
As an inspiration this is how it looks from Intellij side(which I also find super useful):
![multiline_string](https://user-images.githubusercontent.com/10478402/80120915-f3799000-858b-11ea-879a-7f4e5bbeb2d9.gif)

coline-lunatech:
Awesome thanks for the inputs

tgodzik:
I think that Intellij is forcing users a bit too much, I wouldn't automatically add pipes and `stripMargin` using onTypeFormatting. It's something we were thinking about but decided against.

Code action gives the users most choice I think.

tgodzik:
I think we can concentrate on the code action here and improve typing of mutliline strings as suggested by @mlachkar here: https://github.com/scalameta/metals/pull/1661

PR #1685, opened 2020-05-04
code action to make single string multiline with strip margin

colineto:
New code Action to transform any single string into a stripMargin multiline one. https://github.com/scalameta/metals/issues/1660

stripMargin


metals/src/main/scala/scala/meta/internal/metals/codeactions/StringToMultiline.scala
```diff
@@ -0,0 +1,66 @@
...
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+
+    val uri = params.getTextDocument.getUri
+    val path = uri.toAbsolutePath
+    val position = params.getRange.getStart
+
+    Future
+      .successful {
+        path.toInput.tokenize.toOption match {
```
tgodzik, line 26: You should use the content from the buffers here. `path.toInput` will pick up the filesystem version. We keep the most recent version of the file in the `Buffers`.

colineto: okay perfect I'll try that :)

metals/src/main/scala/scala/meta/internal/metals/codeactions/StringActions.scala
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  def toStripMarginEnd(str: String): String = {
+    quotify(str.replace(""""""", "'''.stripMargin"))
+  }
+
+}
+
+object StringActions {
+  def title: String = "Convert to stripMargin"
```
tgodzik, line 91:
```suggestion
  def title: String = "Convert to multiline string"
```
?

tests/unit/src/test/scala/tests/codeactions/StringActionsLspSuite.scala
```diff
@@ -0,0 +1,77 @@
...
+       |  val str = "<<>>"
+       |}
+       |""".stripMargin,
+    s"${StringActions.title}",
+    """|package a
+       |
+       |object A {
+       |  val str = '''|'''.stripMargin
+       |}
+       |""".stripMargin.replace("'", """"""")
```
tgodzik, line 21: You can create an additional value:
`val tripleQuote = "\"\"\""`

tests/unit/src/test/scala/tests/codeactions/StringActionsLspSuite.scala
```diff
@@ -0,0 +1,77 @@
...
+       |
+       |object A {
+       |  val str = "<<>>"
+       |}
+       |""".stripMargin,
+    s"${StringActions.title}",
+    """|package a
+       |
+       |object A {
+       |  val str = '''|'''.stripMargin
```
tgodzik, line 19:
```suggestion
       |  val str = $tripleQuote|$tripleQuote.stripMargin
```

tests/unit/src/test/scala/tests/codeactions/StringActionsLspSuite.scala
```diff
@@ -0,0 +1,77 @@
...
+    s"${StringActions.title}",
+    """|package a
+       |
+       |object A {
+       |  val text = "text"
+       |  val str = s'''|this is an interpolation ${text}'''.stripMargin
+       |}
+       |""".stripMargin.replace("'", """"""")
+  )
+
```
tgodzik, line 76: Let's add a test case with an already existing multiline string to check if we don't propose it flakily.

And maybe a test case with multiple strings on the same line?

colineto: Thanks a lot for all your comments!
I've refactored everything. it looks better from my point of view :)

I just have a problem with multilines when testing. 
when I test that
```scala
"""|package a
   |
   |object A {
   |  val str = s'''|Hello
   |                |there'''.stripMargin + '''the <<cursor>> is actually here'''
   |}
   |""".stripMargin.replace("'", "\"")
```
it actually gives me that 
```scala
"""|package a
   |
   |object A {
   |  val str = s'''|Hello
   |there'''.stripMargin + '''the <<cursor>> is actually here'''
   |}
   |""".stripMargin.replace("'", "\"")
```
I'm already testing all the triple quotes cases so I'm not sure this test is useful

tests/unit/src/test/scala/tests/codeactions/StringActionsLspSuite.scala
colineto: I find it less readable actually, do you think it is really important ?

tests/unit/src/test/scala/tests/codeactions/StringActionsLspSuite.scala
tgodzik: Hmm... does it work in the editor? Maybe it's a problem with range formatting?

tests/unit/src/test/scala/tests/codeactions/StringActionsLspSuite.scala
tgodzik: Yeah, you're right. It's more readable this way. Let's leave it.

tests/unit/src/test/scala/tests/codeactions/StringActionsLspSuite.scala
tgodzik: I think we can skip this test then.

metals/src/main/scala/scala/meta/internal/metals/codeactions/StringActions.scala
```diff
@@ -0,0 +1,97 @@
...
+            val stringActions = tokens.collect {
+              case token: Token.Constant.String
+                  if (token.pos.toLSP.overlapsWith(range) && isNotTripleQuote(
+                    token
+                  )) =>
+                stripMarginAction(uri, token.pos.toLSP)
+            }.toList
+
+            val interpolatedStringTokens = tokens.collect {
+              case start: Token.Interpolation.Start
```
tgodzik, line 37: We will find an earlier token here even from a different line. We should make sure there are no other string tokens between start and end. 

We can should first check the line beside the character and then find both the start and end that match the position.

metals/src/main/scala/scala/meta/internal/metals/codeactions/StringActions.scala
```diff
@@ -0,0 +1,97 @@
...
+    Future
+      .successful {
+        path.toInputFromBuffers(buffers).tokenize.toOption match {
+          case Some(tokens) =>
+            val stringActions = tokens.collect {
+              case token: Token.Constant.String
+                  if (token.pos.toLSP.overlapsWith(range) && isNotTripleQuote(
+                    token
+                  )) =>
+                stripMarginAction(uri, token.pos.toLSP)
```
tgodzik, line 33:
```suggestion
token
```
and then we can handle it together with interpolated strings, so no need to iterate over the tokens twice.

metals/src/main/scala/scala/meta/internal/metals/codeactions/StringActions.scala
```diff
@@ -0,0 +1,97 @@
...
+                  if (start.pos.toLSP.getStart.getCharacter <= range.getStart.getCharacter
+                    && isNotTripleQuote(start)) =>
+                start
+              case end: Token.Interpolation.End
+                  if (end.pos.toLSP.getEnd.getCharacter >= range.getEnd.getCharacter
+                    && isNotTripleQuote(end)) =>
+                end
+            }.toList
+
+            interpolatedStringTokens match {
```
tgodzik, line 47:
```suggestion
            interpolatedStringTokens.collect {
```

tgodzik, line 47: We could do something like:
```scala
def loop(coll : List[Token]): Seq[l.CodeAction] = {
  coll match {
    case token: Token.Constant.String :: _ =>  stripMarginAction(uri, token.pos.toLSP)
    case (s: Token.Interpolation.Start) :: (e: Token.Interpolation.End) :: other =>
      if ( s <= pos <= e)
         stripMarginAction(uri,  ....)
      else
        loop(other)
    case _ => Nil
  }
}
```

tests/unit/src/test/scala/tests/codeactions/StringActionsLspSuite.scala
colineto: okay we will skip that one, but we need some others :)

colineto:
There is definitely an issue when a string is behind another. Maybe checking the line as we did before will help, I'll try. 
Anyway this gives me another question. in a case of a + string : `val a = "" + ""` is it more useful to transform the two of them into a multiline or only one by one keeping the + between ?

tgodzik:
> There is definitely an issue when a string is behind another. Maybe checking the line as we did before will help, I'll try.

Thanks!

> Anyway this gives me another question. in a case of a + string : `val a = "" + ""` is it more useful to transform the two of them into a multiline or only one by one keeping the + between ?

I would opt only for changing one. By merging them together I think we are guessing too much about the user's intentions.

metals/src/main/scala/scala/meta/internal/metals/codeactions/StringActions.scala
```diff
@@ -0,0 +1,103 @@
...
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+
+    val uri = params.getTextDocument.getUri
+    val path = uri.toAbsolutePath
+    val range = params.getRange
+
+    Future
+      .successful {
+        val input = path.toInputFromBuffers(buffers)
+        Trees.defaultDialect(input).tokenize.toOption match {
```
tgodzik, line 28: We want to use the 2.13 dialect, otherwise the feature will not work in a file with Scala 2.13 specific syntax

colineto: okay perfect :)

Issue #1669, opened 2020-04-28
Automatic imports doesn't import in the right position when in a package object


ckipp01:
**Describe the bug**

Give the following code:

```scala
package example

package object example {
  def doSomething(instant: Inst@@ant, p: Period) = ???
}
```
If I trigger a code action on the `@@` and then accept to import `java.time.Instant`, I'd expect this to be imported between the package declaration and the package object.  However, this imports in the package object. So the result is below:

```scala
package example

package object example {

  import java.time.Instant
  def doSomething(instant: Instant, p: Period) = ???
}
```

**Screenshots**

![2020-04-28 10 09 47](https://user-images.githubusercontent.com/13974112/80463750-05b34f80-8939-11ea-905d-0524e02eccd7.gif)

**Search terms**
code action, imports, package object

gabro:
Does it do the same if you import it via completions?

ckipp01:
> Does it do the same if you import it via completions?

Yes, it's the same behavior with a completion or a code action.

PR #1686, opened 2020-05-04
fix auto import position when package object inside package

colineto:
fixes https://github.com/scalameta/metals/issues/1669
The tree is read from ground to top and get the first package he sees which is a package object in our case. Reversing the tree should do the trick. 
I'm not totally sure, sounds too easy :)

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
@@ -47,7 +47,7 @@ trait AutoImports { this: MetalsGlobal =>
     findLastVisitedParentTree(pos) match {
       case Some(_: Import) => None
       case _ =>
-        val enclosingPackage = lastVisitedParentTrees.collectFirst {
+        val enclosingPackage = lastVisitedParentTrees.reverse.collectFirst {
           case pkg: PackageDef => pkg
         }
         enclosingPackage match {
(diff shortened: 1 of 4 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -47,7 +47,7 @@ trait AutoImports { this: MetalsGlobal =>
     findLastVisitedParentTree(pos) match {
       case Some(_: Import) => None
       case _ =>
-        val enclosingPackage = lastVisitedParentTrees.collectFirst {
+        val enclosingPackage = lastVisitedParentTrees.reverse.collectFirst {
```
tgodzik, line 50: I wonder if we can make it a bit more efficient than using reverse to avoid creating a new list :thinking: 

Maybe we could create an extension for list in MtagsEnrichments?

Something like:

```scala
implicit class XtensionList[T](list: List[T]) {
  def collectLast[R](fun: PartialFunction[T, R]): Option[R] = {
    def loop(current: List[T]): Option[R] = {
      current match {
        case Nil => None
        case head :: tl =>
          loop(tl).orElse(fun.lift(head))
      }
    }
    loop(list)
  }
}
```

colineto: yes sure! could be useful and more readable

tgodzik:
Thanks for fixing this! It might seem small, but it's still a lot of work connected to actually finding the issue. 

> I'm not totally sure, sounds too easy :)

It's actually true for the most issues connected to PC. At least 1 hour of checking (if lucky) and then one small change :sweat_smile:

PR #1708, opened 2020-05-06
Bump Bloop version and update fastpass to new TraceSettings

wiwa:
Previously, Zipkin tracing properties were set as global `javaOptions` for Bloop open `fastpass` invocation`. However, users may wish Bloop to submit or annotate traces differently depending on the workspace.

This PR will allow `fastpass` to set Zipkin tracing properties on a workspace level via the `TraceSettings` in Bloop `WorkspaceSettings` introduced in https://github.com/scalacenter/bloop/pull/1262/


metals/src/main/scala/scala/meta/internal/zipkin/ZipkinProperties.scala
```diff
@@ -4,20 +4,20 @@ object ZipkinProperties {
 
   val zipkinServerUrl: Property = Property("metals.zipkin.server.url")
 
-  val debug: Property = Property("metals.bloop.trace.debug")
+  val debug: Property = Property("metals.bloop.tracing.debug")
```
jvican, line 7: Remember debug has changed! It's now `debugTracing` :smile:

olafurpg: Good catch!

olafurpg:
LGTM 👍 Thank you!

PR #1716, opened 2020-05-08
Support Pants' `strict_deps`

Duhemm:
**Note: I believe we need to wait for the corresponding changes to Pants to be merged and release. (notably https://github.com/pantsbuild/pants/pull/9485)**

Previously, `fastpass` was not calling Pants with
`--respect-strict-deps`, which means that Pants will not respect
`strict_deps` when doing the export.

The output of Pants' `export-dep-as-jar` also changed to include a clear
separation between compile-time and runtime classpaths.

This commit updates fastpass so that `--respect-strict-deps` is passed
to pants, and so that fastpass can understand the new output of
`export-dep-as-jar`. The resulting Bloop configuration files will
include the correct compile-time and runtime classpaths.


olafurpg:
Thank you for this contribution! The Pants test failure looks legitimate
```
ERROR: Unrecognized command line flags on scope 'export-dep-as-jar': --respect-strict-deps.
```

The Pants version is declared here https://github.com/scalameta/metals/blob/bab412a20a229d94fe9a6fe27334406aa454b1ad/tests/slow/src/test/scala/tests/pants/PantsLspSuite.scala#L71

/cc @bescobar what Pants version can we target that includes the `--respect-strict-deps` flag?

blorente:
The first version that includes that flag is `1.26.0.dev1`, in this commit: https://github.com/pantsbuild/pants/commit/2e9dcd4f8dcd72443d6abf2def0a5958d50610e6

Duhemm:
I updated the PR to use Pants `1.26.0rc2` and fallback to `libraries` is `compile_libraries` or `runtime_libraries` don't exist.

PR #1717, opened 2020-05-08
teach metals to accept configuration of range formatting

danking:
This PR provides an escape hatch to Emacs users who find the beahvior documented in
formatting can opt-in to this behavior.

My main concern is that this change appears not backwards compatible. I am not familiar
with Metals JSON deserialization system. Ideally the `InitializationOptions` would
default to `true` for `provideRangeFormater` if it is not present in the JSON payload
from the `initialize` request.

I will make a corresponding change to https://github.com/emacs-lsp/lsp-mode to support
this new configuration option.

Thank you for your time.

I tried to test this locally, but after running `sbt +publishLocal`, I can't build a Scala-metals binary as described under the vim section of the contributing docs (there is no Emacs section):
```
~/projects/metals/coursier bootstrap \                                
  --java-opt -Dmetals.client=emacs \
  org.scalameta:mtags_2.11.12:0.9.1-SNAPSHOT \
  -r bintray:scalacenter/releases \
  -o /usr/local/bin/metals-emacs-snapshot -f -M scala.meta.metals.Main
```

The Main class appears to be missing. Could you please advise as to how to get a working binary?

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -541,7 +541,9 @@ class MetalsLanguageServer(
       capabilities.setDocumentOnTypeFormattingProvider(
         new DocumentOnTypeFormattingOptions("\n")
       )
-      capabilities.setDocumentRangeFormattingProvider(true)
+      capabilities.setDocumentRangeFormattingProvider(
+        clientConfig.initializationOptions.provideRangeFormatting
+      )
       capabilities.setSignatureHelpProvider(
         new SignatureHelpOptions(List("(", "[").asJava)
       )
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -542,7 +542,7 @@ class MetalsLanguageServer(
         new DocumentOnTypeFormattingOptions("\n")
       )
       capabilities.setDocumentRangeFormattingProvider(
-        clientConfig.initializationOptions.provideRangeFormatting
+        initialConfig.allowMultilineStringFormatting
       )
       capabilities.setSignatureHelpProvider(
         new SignatureHelpOptions(List("(", "[").asJava)
--- a/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
@@ -18,7 +18,8 @@ final case class InitializationOptions(
     quickPickProvider: Boolean,
     slowTaskProvider: Boolean,
     statusBarProvider: String,
-    treeViewProvider: Boolean
+    treeViewProvider: Boolean,
+    provideRangeFormatting: Boolean
 ) {
   def this() =
     this(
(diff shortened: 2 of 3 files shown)


tgodzik:
> The Main class appears to be missing. Could you please advise as to how to get a working binary?

Instead of:
`org.scalameta:mtags_2.11.12:0.9.1-SNAPSHOT \` 

you should use:

`org.scalameta:metals_2.12:0.9.1-SNAPSHOT` 

Mtags do not contain main class indeed.

ckipp01:
So I think the first question that needs to be answered is whether or not Metals is actually a `documentRangeFormattingProvider` seeing that we only offer range formatting on multi-line strings. Honestly, imo an argument could be made that no, we shouldn't say that we provide that because we simply provided a small subset of that only on multi-line strings. It would probably be more appropriate to have a code action for format a multi-line string rather than having that be the only criteria to say that we implement document range formatting.

Again, I think that's the first thing that should be decided and clarified.

Secondly, I don't love the approach layed out here mainly because `documentRangeFormattingProvider` is a `ServerCapabilities` item, meaning that the server should be in charge of saying whether it does or doesn't support this. Having `InitializationOptions` set this seems a bit odd as the client is basically cancelling out what the server is saying that it can provide. I think if a client is attempting to do that, then it's pointing to something not being correctly set in the server. Again, this goes up to the top point of whether or not we should consider Metals a range formatting provider only by its ability to format the range of a multi-line string.

tgodzik:
> So I think the first question that needs to be answered is whether or not Metals is actually a `documentRangeFormattingProvider` seeing that we only offer range formatting on multi-line strings. Honestly, imo an argument could be made that no, we shouldn't say that we provide that because we simply provided a small subset of that only on multi-line strings. It would probably be more appropriate to have a code action for format a multi-line string rather than having that be the only criteria to say that we implement document range formatting.

We use range formatting here in order to support formattingOnPaste in VS Code. Thanks to that we can easily paste things in to multiline strings. The other option would be to paste, then select and run a code action. This is much less user friendly, so I would really not remove the `documentRangeFormattingProvider`

An argument not to do this change is that it really doesn't make sense to have different formatter for range formatting and for document formatting. You should just run formatting on a full file, otherwise you will get different results depending on the which method you use. Because of that it actually makes sense for Metals to implement it and just not do anything rather then allow inconsistent behaviour.

@ckipp01 maybe we could add a server property that just allows to turn of rangeFormattingProvider and later it can be defined in `.jvmopts`  or we could define it always for emacs. I don't think we can actually represent this properly using init options (I agree with @ckipp01 here)

Also jvm option might make sense since this is something that only a small subset of people will really care about it. It's something that @danking cares strongly about but not neccessarily everyone.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -541,7 +541,9 @@ class MetalsLanguageServer(
       capabilities.setDocumentOnTypeFormattingProvider(
         new DocumentOnTypeFormattingOptions("\n")
       )
-      capabilities.setDocumentRangeFormattingProvider(true)
+      capabilities.setDocumentRangeFormattingProvider(
```
tgodzik, line 544: So probably the best that we can do currently is add an option to MetalsServerConfig like:
```scala
  allowMultilineStringFormatting = MetalsServerConfig.binaryOption(
      "metals.allow-multiline-string-formatting",
      default = true
    ),
```
And then:
```scala      

capabilities.setDocumentRangeFormattingProvider(initialConfig.allowMultilineStringFormatting)
```

danking:
Great, I made the changes you requested. I'm excited to get this functionality in.

FWIW, there are projects that do not use `scalafmt` and consider it rude to reformat entire files when you're only changing a few lines. The fallback indenter does a remarkably good job getting the right indentation. In Emacs, you nearly always have something "selected" in the Emacs sense. When programming, I have a habit of invoking the reformat region command often, especially when cutting and pasting. For example, when manipulating the tests for our compiler which contains somewhat large and deeply nested AST definitions.

Issue #1664, opened 2020-04-25
cannot navigate to a scala file containing underscore inside of integer literal


bwiercinski:
**Describe the bug**

**To Reproduce**
1. Create new class
```scala
class Tests {
  val x = 100_000

  def dummy() = x
}
```
2. Somewhere else
```scala
object Tests2 {
  val x = new Tests
  x.dummy()
}
```
3. click Ctrl and hover `dummy()`
4. Nothing highlights, cannot navigate to definition
5. Remove underscore
6. Works

**Expected behavior**
Should navigate in both scenarios

**Installation:**
- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.8.4

**Additional context**
https://www.javatpoint.com/java-numeric-literals-with-underscore

**Search terms**
underscore long integer literal

tgodzik:
Thanks for reporting! I think I might have an idea of what's going on, will try to take a quick look.

PR #1721, opened 2020-05-11
Fix issues in files with Scala 2.13 specific syntax

tgodzik:
Fixes https://github.com/scalameta/metals/issues/1664

Previously, we would use 2.12 parser and tokenizer by default. Now we define the default dialect as 2.13 in one place and make sure it's used everywhere.

### DIFF
--- a/tests/slow/src/test/scala/tests/feature/DefinitionCrossLspSuite.scala
+++ b/tests/slow/src/test/scala/tests/feature/DefinitionCrossLspSuite.scala
@@ -19,6 +19,36 @@ class DefinitionCrossLspSuite
     }
   }
 
+  test("underscore") {
+    cleanDatabase()
+    for {
+      _ <- server.initialize(
+        s"""
+           |/metals.json
+           |{
+           |  "a": {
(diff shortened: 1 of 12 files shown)


tests/slow/src/test/scala/tests/feature/DefinitionCrossLspSuite.scala
```diff
@@ -19,6 +19,36 @@ class DefinitionCrossLspSuite
...
 
+  test("underscore") {
+    cleanDatabase()
+    for {
+      _ <- server.initialize(
+        s"""
+           |/metals.json
+           |{
+           |  "a": {
+           |    "scalaVersion": "${BuildInfo.scala213}"
```
kpbochenek, line 30: This is unrelated to switching to default scala 2.13, correct?
Or you wanted to test that default scala would be 2.13 by veryfing 2.13 only feature would be parsed?

tgodzik: That is related, I wanted to make sure we properly parse 2.13 syntax.

PR #1723, opened 2020-05-13
Feature codeAction strings to interpolation strings

colineto:
This is just a proposal for a code action that make it easier to interpolate strings.

For now it just add a `s` at start so adding a`$` inside will automatically autocomplete.

toSimpleInterpolation

But before knowing there was autocompletion inside interpolated string I made it add also the `${}` 

toInterpolationString

I'd like to know if you find it useful and in which way

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/StringActions.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/StringActions.scala
@@ -34,64 +34,78 @@ class StringActions(buffers: Buffers) extends CodeAction {
               )
               .collect {
                 case token: Token.Constant.String
-                    if (token.pos.toLSP.overlapsWith(range)
-                      && isNotTripleQuote(token)) =>
+                    if (token.pos.toLSP.overlapsWith(range)) =>
                   token
                 case start: Token.Interpolation.Start
-                    if (start.pos.toLSP.getStart.getCharacter <= range.getStart.getCharacter
-                      && isNotTripleQuote(start)) =>
+                    if (start.pos.toLSP.getStart.getCharacter <= range.getStart.getCharacter) =>
                   start
                 case end: Token.Interpolation.End
-                    if (end.pos.toLSP.getEnd.getCharacter >= range.getEnd.getCharacter
-                      && isNotTripleQuote(end)) =>
+                    if (end.pos.toLSP.getEnd.getCharacter >= range.getEnd.getCharacter) =>
                   end
               }
               .toList match {
               case (t: Token.Constant.String) :: _ =>
-                List(stripMarginAction(uri, t.pos.toLSP))
+                if (isNotTripleQuote(t))
+                  multilineAction(uri, t.pos.toLSP) ::
+                    List(interpolationAction(uri, t.pos.toLSP))
+                else List(interpolationAction(uri, t.pos.toLSP))
               case _ :: (t: Token.Constant.String) :: _ =>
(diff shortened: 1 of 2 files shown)


tgodzik:
This looks good, however I am not sure if we really need it. You can already use completions and it will add the `s` automatically. That does feel more natural for me.

What are other people's opinions?

ckipp01:
> This looks good, however I am not sure if we really need it. You can already use completions and it will add the `s` automatically. That does feel more natural for me.
> 
> What are other people's opinions?

Is this actually the case? I thought the completions only added a `{}`. Testing locally I'm unable to actually get a `s` inserted in both coc-metals and vs code.

No completion offered here:


Completion offered here and `{}` inserted:

gabro:
Really nice @colineto!

To answer @tgodzik'point, similarly to other cases I generally prefer:

- code actions for fixing warnings while I'm browsing code
- completions for actions that take place while I'm typing code

In this case, I think the code action is useful in these cases:

- if I forgot to trigger the completion that makes the string interpolated (it's quite easy to miss, see for the example above by @ckipp01)
- in response to the default Scalac warning "That looks like an interpolated expression! Did you forget the interpolator?"

So I think the feature it's useful!

More food for thought:

- what about the opposite? If there are no interpolations in the string maybe we could offer a code action of type `"refactor"` to remove the useless interpolator

gabro:
@ckipp01 in the first case, you need to complete `test`. If you go ahead and type the `.` you've missed the chance (it's not ideal, but IIRC the logic to make it work is quite tricky)

ckipp01:
> @ckipp01 in the first case, you need to complete `test`. If you go ahead and type the `.` you've missed the chance (it's not ideal, but IIRC the logic to make it work is quite tricky)

ahhhh indeed, the `s` is inserted when I complete $test. Ha, I never realized that.

Also, great job on this @colineto.

Thinking about it, if I was already in a string and realized I wanted an interpolated string and forgot to add the `s` (like in my gif), then I think I'd use this code actions. I also agree with @gabro on catching the Scalac warning.

gabro:
> I also agree with @gabro on catching the Scalac warning.

About this, I don't think we need to look for the specific warning if we offer this unconditionally. However, I would probably limit it to cases where the string contains an interpolated value

ckipp01:
One other thing that I'm noticing that @colineto brought up is that I can get this to work with coc-metals, but not with vs code???

coc-metals:



VS Code:


Does this work for others with VS Code?

tgodzik:
> Does this work for others with VS Code?

It does work for me, haven't had issues. " $tes<cursor> " and `ctrl + space`

ckipp01:
> > Does this work for others with VS Code?
> 
> It does work for me, haven't had issues. " $tes " and `ctrl + space`

Ahh, I didn't realize I had to manually trigger it, since I don't need to with coc-metals

mlachkar:
I think this feature is nice but only interesting for some use cases.
I do use in intellij Live template for that 


I don't find scala snippet (or have issues using them on Vscode)

colineto:
> It does work for me, haven't had issues. " $tes " and `ctrl + space`

Oh okay, i can trigger it.. wouldn't have proposed that if I knew 😊

I'm happy to see it can eventually be useful ! I'm just not sure to understand how yet 😄

tgodzik:
Closing this for now in favour of https://github.com/scalameta/metals-feature-requests/issues/133 

@colineto would be awesome to continue this if you ever have time again, but closing it in favour of an issue, where we can discuss it further. Thanks for the awesome work!

Issue #1553, opened 2020-03-27
Missing references for case class if search starts from different file


tgodzik:
**Describe the bug**
If we look for references of a case classes's synthetic companion object from a file that doesn't contain the definition, we will not find all the correct references.

This is due to the fact that we only look for alternative references inside the current file. We need to first find semanticDB for the definition of a symbol and then use that.

**To Reproduce**

Add test case to ReferencesLspSuite:
```scala
  check(
    "companion",
    """|/a/src/main/scala/a/Main.scala
       |package a
       |case class <<Main>>(name : String){
       |  val appl = <<Main>>.apply("")
       |}
       |/a/src/main/scala/a/Other.scala
       |package a
       |object Other{
       | val appl = <<M@@ain>>.apply("")
       |}
       |""".stripMargin
  )
```

**Expected behavior**
All references to the case classes are returned.

**Installation:**
- Operating system: macOS/Windows/Linux
- Editor: Visual Studio Code/Atom/Vim/Sublime/Emacs
- Metals version: v0.8.3

**Additional context**

**Search terms**
references rename case class

tgodzik:
Should be possible to reuse method from `ImplementationProvider`, which should msot likely be moved to definitionProvider:

```scala
  private def findSemanticDbForSymbol(symbol: String): Option[TextDocument] = {
    for {
      symbolDefinition <- index.definition(MSymbol(symbol))
      document <- findSemanticdb(symbolDefinition.path)
    } yield {
      document
    }
  }
```

faizk:
Ah, someone finally found this.

I ran into the exact same thing problem a couple of months ago, but in my case it as an `unapply` that wasn't getting found:

```scala
// /a/src/main/scala/a/Stuff.scala
package a
sealed trait Stuff
case class X(n: Int) extends Stuff // looking for references to this,
```
```scala
// /a/src/main/scala/a/Main.scala
package a
object Main {
   def n(stuff: Stuff): Option[Int] = stuff match {
     case X(n) => Some(n) // this one won't be found.
     case _ => None
   }
}
```
Before I could figure out how to appropriately express this as a failing unit-test, I mucked around with `ReferenceProvider.scala` trying to find out how synthetics are getting missed - in the process wrote this hack just to test the theory that synthetics across the file border weren't getting found.
 
https://github.com/faizk/metals/commit/e837a24a32997ae858ace5128969e59b931b647e

[...]

PR #1727, opened 2020-05-13
fix missing synthetic references 

faizk:
Fix for one aspect of #1553:

References to case-classes via synthetics (`apply` , `unapply`) occurring in compilation units that neither define not explicitly import the class - these will now be found as expected. 

This fix results in more semanticdb lookups since we now include all those that might contain any of the alternative in addition to just the main type symbol.

The other aspect that isn't fixed just as yet: when starting the search *from* a synthetic the search results won't include the class definition itself as one of the expected references results (though other matching synthetics are returned). - for now the failing tests that demonstrate this issue are coaxed so they'll pass anyway - I'll work on this next.

### DIFF
--- a/tests/unit/src/test/scala/tests/ReferenceLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/ReferenceLspSuite.scala
@@ -108,6 +108,113 @@ class ReferenceLspSuite extends BaseRangesSuite("reference") {
     } yield ()
   }
 
+  /** References should be found correctly regardless of the whether or not
+   *  the declaring code and the refering share the same compilation unit.
+   *
+   *  This checks both cases by first placing them in the same file and then
+   *  spread across different files. The package stays the same - this is
+   *  actually necessary to test that references to synthetics can be found
+   *  even in the absence of an explicit `import`, which is often the case when
+   *  the referencing code shares with the declaring code the same package
+   *  (but possibly a different compilation unit)
+   */
+  def checkInSamePackage(
+      name: String,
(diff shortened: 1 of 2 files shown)


kpbochenek:
You have to rebase properly(or merge) as github diff includes changes from other people commits

faizk:
> You have to rebase properly(or merge changes to your branch) as github diff includes changes from other people commits

oops, thanks.

faizk:
https://github.com/scalameta/metals/pull/1727/checks?check_run_id=697058563

not sure if I need to worry about these "cancelled" tests.

tests/unit/src/test/scala/tests/ReferenceLspSuite.scala
```diff
@@ -107,6 +107,113 @@ class ReferenceLspSuite extends BaseRangesSuite("reference") {
...
+       |  val ma = <<Main>>("a")
+       |}
+       |""".stripMargin,
+    """|object Other {
+       |  val mb = <<Main>>.apply("b")
+       |}
+       |""".stripMargin
+  )
+
+  checkInSamePackage( // FIXME: doesn't find the case class declaration
```
tgodzik, line 152: Let's maybe add the reference to https://github.com/scalameta/metals/issues/1553

faizk: OK, I'm going to reference the conversation where it becomes clear that we've got 2 bugs (and what the bugs are).

PR #1726, opened 2020-05-13
Bump scalafmt to 2.5.2

takezoe:

PR #1730, opened 2020-05-14
Remove the declaration of the experimental features in initialize and add `ServerInfo`

ckipp01:
Previously these used to be set so that during the intiliaze phase under `ServerCapabilities` so a client could check to see if a server supported something. Since we have now moved away from using the clientExerimentalFeatures it makes sense to no longer declare these. This doesn't mean that aren't able to be set, just that we aren't explicit to the clients that we support them anymore. Plus these are out of date anyways.

Also as of [3.15](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialize) the initializeParams has an option for `serverInfo` to be passed to the client which we haven't been using. This pr also adds in that field.


ckipp01:
Same flaky test that was mentioned here: https://github.com/scalameta/metals/pull/1726#issuecomment-628570147

PR #1731, opened 2020-05-14
Fix CrossDefinitionLspSuite and additional logs to the DAP suites

tgodzik:


tgodzik:
Merging this since, these are just changes to the tests. This will fix the issues with lsp tests, fix the output from DAP tests and add some additional logs to tests.

PR #1753, opened 2020-05-15
Use onTypeFormatting to create multiline string

mlachkar:
By default, strip margin is added, but it's configurable with 
`   "metals.enable-strip-margin-on-type-formatting" : true `

### DIFF
--- a/mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
+++ b/mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
@@ -50,6 +50,11 @@ public interface PresentationCompilerConfig {
     boolean isCompletionItemDetailEnabled();
 
     /**
+     * Returns false if the <code>UserConfiguration.enableStripMarginOnTypeFormatting</code> is configured to false.
+     */
+    boolean isStripMarginEnabledOnTypeFormatting();
+
+    /**
      * Returns true if the <code>CompletionItem.documentation</code> field should be populated.
      */
     boolean isCompletionItemDocumentationEnabled();
--- a/mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
+++ b/mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
@@ -52,7 +52,7 @@ public interface PresentationCompilerConfig {
     /**
      * Returns false if the <code>UserConfiguration.enableStripMarginOnTypeFormatting</code> is configured to false.
      */
-    boolean isStripMarginEnabledOnTypeFormatting();
+    boolean isStripMarginOnTypeFormattingEnabled();
 
     /**
      * Returns true if the <code>CompletionItem.documentation</code> field should be populated.
--- a/mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
@@ -94,6 +94,25 @@ object MultilineStringFormattingProvider {
       if (i < 0 || i > text.length) false else text(i) == quote
     )
 
+  private def indentWhenNoStripMargin(
+      startToken: Token,
+      endToken: Token,
+      splitLines: Array[String],
+      position: Position,
+      enableStripMargin: Boolean
+  ): List[TextEdit] = {
+    if (enableStripMargin && startToken.pos.startLine == position.getLine - 1 && endToken.pos.endLine == position.getLine) {
+      val positionOfLastQuote =
+        splitLines(position.getLine).lastIndexOf("\"")
+      val newPos =
+        new Position(position.getLine, positionOfLastQuote + 1)
+      List(
+        indent(splitLines, position),
+        new TextEdit(new Range(newPos, newPos), ".stripMargin")
+      )
+    } else List(indent(splitLines, position))
+  }
+
   private def indent(
       splitLines: Array[String],
       position: Position
@@ -127,9 +146,10 @@ object MultilineStringFormattingProvider {
   private def inToken(
       startPos: meta.Position,
       endPos: meta.Position,
-      token: Token
+      startToken: Token,
+      endToken: Token
   ): Boolean =
-    startPos.startLine >= token.pos.startLine && endPos.endLine <= token.pos.endLine
+    startPos.startLine >= startToken.pos.startLine && endPos.endLine <= endToken.pos.endLine
 
   private def pipeInScope(
       startPos: meta.Position,
@@ -159,6 +179,39 @@ object MultilineStringFormattingProvider {
     pipeBetweenLastLineAndPos != -1 || pipeBetweenSelection != -1
   }
 
+  def isStringOrInterpolation(
+      startPosition: StartPosition,
+      endPosition: EndPosition,
+      token: Token,
+      index: Int,
+      text: String,
+      tokens: Tokens,
+      newlineAdded: Boolean
+  )(
+      indent: (Token, Token, Int) => Option[List[TextEdit]]
+  ): Option[List[TextEdit]] = {
+    token match {
+      case startToken @ (_: Token.Constant.String | _: Interpolation.Start) =>
+        val endIndex = if (startToken.isInstanceOf[Interpolation.Start]) {
+          var endIndex = index + 1
+          while (!tokens(endIndex)
+              .isInstanceOf[Interpolation.End]) endIndex += 1
+          endIndex
+        } else index
+        val endToken = tokens(endIndex)
+        if (inToken(startPosition, endPosition, startToken, endToken)
+          && isMultilineString(startToken) &&
+          pipeInScope(
+            startPosition,
+            endPosition,
+            text,
+            newlineAdded
+          )) indent(startToken, endToken, index)
+        else None
+      case _ => None
+    }
+  }
+
   private def indentOnRangeFormatting(
       start: StartPosition,
       end: EndPosition,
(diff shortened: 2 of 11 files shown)


mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -23,19 +22,19 @@ object MultilineStringFormattingProvider {
...
-      case token: Token.Ident if token.value == stripMargin =>
-        true
-      case _ =>
-        false
-    }
+  def hasStripMarginSuffix(token: Token): Boolean =
+    !stripMargin.zipWithIndex
+      .exists {
+        case (char, index) =>
+          token.input.chars(token.pos.end + index + 1) != char
```
tgodzik, line 29: This might fail if we have a whitespace somewhere between like `"""|""" stripMargin` That's why previously we used a while loop to skip over dot or whitespace.

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -74,6 +73,18 @@ object MultilineStringFormattingProvider {
     if (lastQuote != -1) Some((lastQuote, quoteClosed)) else None
   }
 
+  private def fourQuotes(
```
tgodzik, line 76: We could have one method like:
```scala
private def hasNQuotes(start: Int, text: String, n : Int): Boolean = {
   (start until n).forall(text(_) == quote)
}
```
and use it also in `isMultilineString`

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -131,57 +150,113 @@ object MultilineStringFormattingProvider {
...
-      ) && tokens(endIndex).pos.end > endPosition.end
-    }
+      end: EndPosition,
+      tokens: Tokens,
+      newlineAdded: Boolean,
+      splitLines: Array[String],
+      position: Position,
+      text: String
+  ): List[TextEdit] = {
+    implicit val sourceText: String = text
```
tgodzik, line 231: Let's not use implicit here. They should never be used for simple types like strings or Boolean. It adds a level of complexity, you need to search for implicit rather than just have it there explicitly.  And this can also cause some unexpected bugs. Honestly, I think the simpler we can write the easier it will be for someone to read this in the future.

Let's just add explicit values. You can also just use currying or overloaded functions if you don't want too write those values multiple times.

```scala
val fun =  isStringOrInterpolation(
          start,
          end,
          andMultilineString = true,
          andHadStripMargin = true,
          andHasPipeInScope = true
        )(List(indent(splitLines, position))).orElse(...)

fun(sourceText, tokens, newlineAdded)
```

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -131,57 +150,113 @@ object MultilineStringFormattingProvider {
...
+            end,
+            andMultilineString = true,
+            andHadStripMargin = false,
+            andHasPipeInScope = true
+          ) {
+            val newPos =
+              new Position(position.getLine, position.getCharacter + 3)
+            List(
+              indent(splitLines, position),
+              new TextEdit(new Range(newPos, newPos), ".stripMargin")
```
tgodzik, line 254: I am not sure if we want to add `stripMargin` this automatically. It was previously done on purpose not to add it. The initial plan was to do it as in Intellij, but after some discussions it turns out it's not something that every developer wants and we shouldn't be forcing them.

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -251,6 +325,12 @@ object MultilineStringFormattingProvider {
     List(textEditPrecedentLine, textEditcurrentLine)
   }
 
+  private def repaceWithSixQuotes(pos: Position): List[TextEdit] = {
+    val pos1 = new Position(pos.getLine, pos.getCharacter - 3)
```
tgodzik, line 329: Will it work if someone adds a quote inside existing quotes? `"add new quote here"`

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
mlachkar: I didn't know we could have more than one space after the multiline string. 
Right now it works infact only if there is a space or "." after the multiline string.

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
mlachkar: Ok

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
mlachkar: I agree with you about implicit, but It makes the call so long. It was just a proposition.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -604,10 +604,17 @@ final class TestingServer(
       query: String,
       expected: String,
       autoIndent: String,
+      replaceWith: String,
```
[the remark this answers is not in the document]
mlachkar: ok

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
mlachkar: This can only be trigger if there are already 3 consecutive quotes and the fourth character being typed is a quote, and the file cannot be tokenized.

mlachkar: Actually I just found one case where it will introduce a bug.

If there is already line = """ whatever 
and then the user will try to write 3 quotes. 
Also, I wonder how quotes are being written in other editors. For vscode it's will always write two quotes by two quotes, but maybe 3 is enough ?

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
mlachkar: Ok, I didn't know that you changed your mind about the stripMargin. 
In my case, every time I add a pipe, it's because I will add stripMargin. It's quite annoying to write it, sometimes by error, the completion puts instead stringPrefix in VSCode, and it's quite slow for this case in Intellij. 
If the person doesn't want it, he will delete it. So it will depend on how many people do actually write the stripmargin compared to those who will delete it?

tgodzik: It's a tough choice really. For me it would be useful, but I worry some people might be opposed.

mlachkar:
Test failing is due to: 
```returned HTTP response code: 504 for URL: https://dl.bintray.com/sbt/sbt-plugin-releases/com.typesafe.sbt/sbt-native-packager/scala_2.12/sbt_1.0 ..```

File: tests/unit/src/main/scala/tests/TestingServer.scala (hunk 6 above)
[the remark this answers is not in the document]
Comment (tgodzik) [reply tests/unit/src/main/scala/tests/TestingServer.scala:607]: I think this might have been fixed in the other commit we had issues with. Could you change it again?

File: mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java [hunk 7]
```diff
@@ -49,6 +49,11 @@
      */
     boolean isCompletionItemDetailEnabled();
 
+    /**
+     * Returns false if the <code>UserConfiguration.enableStripMarginOnTypeFormatting</code> is configured to false.
+     */
+    boolean isStripMarginEnabledOnTypeFormatting();
```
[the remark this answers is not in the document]
mlachkar: ok

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -131,57 +159,110 @@ object MultilineStringFormattingProvider {
...
-      end: EndPosition
-  ): PartialFunction[(Token, Int), Boolean] = {
-    case (token: Constant.String, index: Int) =>
-      isMultilineString(sourceText, token) && hasStripMarginSuffix(
-        index,
-        tokens
-      ) && inToken(start, end, token)
+      andMultilineString: Boolean = true,
+      andHadStripMargin: Boolean = true,
+      andHasPipeInScope: Boolean = true
```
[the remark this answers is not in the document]
mlachkar: yes right !

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -131,57 +159,110 @@ object MultilineStringFormattingProvider {
...
+        ) orElse {
+          isStringOrInterpolation(
+            start,
+            end,
+            text,
+            newlineAdded,
+            tokens,
+            andHadStripMargin = false
+          ) {
+            if (enableStripMargin) {
```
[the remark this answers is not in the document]
mlachkar: The issue is that we dont know whre the token ends ... for the string it works, but for the interpolation, we loose the information of the EndToken.. I dont know how to do that, except If  I search again with a while for endToken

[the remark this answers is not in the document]
tgodzik: That's why I am thinking we should to that inside the case, where we have the end token. We can drop the partial function and just have a normal function returning a list of TextEdits and move searching for the end token from inside the case guard to the actual case. We return Nil if the end is not correct or `indent + stripMarginEdit(endToken.pos + 1)` otherwise

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -74,6 +79,40 @@ object MultilineStringFormattingProvider {
...
+
+  private def indentWhenNoStripMargin(
+      startToken: Token,
+      endToken: Token,
+      splitLines: Array[String],
+      position: Position,
+      enableStripMargin: Boolean
+  ): List[TextEdit] = {
+    if (enableStripMargin && startToken.pos.startLine == position.getLine - 1 && endToken.pos.endLine == position.getLine) {
+      val positionOfLastQuote =
```
tgodzik, line 105: You can now use endToken this way:
```scala
      val newPos =
        new scala.meta.inputs.Position.Range(
          endToken.input,
          endToken.end,
          endToken.end
        )

      List(
        indent(splitLines, position),
        new TextEdit(newPos.toLSP, ".stripMargin")
      )
```

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
```diff
@@ -251,6 +350,12 @@ object MultilineStringFormattingProvider {
     List(textEditPrecedentLine, textEditcurrentLine)
   }
 
+  private def repaceWithSixQuotes(pos: Position): List[TextEdit] = {
```
tgodzik, line 353:
```suggestion
  private def replaceWithSixQuotes(pos: Position): List[TextEdit] = {
```

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
tgodzik: if we have an existing string say:
`val a = ""`
and we add quotes before, we will get:
`val a = """"`
Maybe we should also create 6 quotes there?

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
mlachkar: yes .. sorry forgot this place

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
mlachkar: It's clearly not a normal usage of this feature (adding quotes in the beginning ). I don't think we should take this usage into account.
I don't know for the "find the end and fill until 6": it means it will only cover the 3 quotes case since you are not supposed to have 5 quotes except if you deleted two quotes

mlachkar: 4 seems like a good number for me, since 3 could be a typing error.

mlachkar:
I addressed all your comments and I will open a PR in vscode-metals for the configuration

mtags/src/main/scala-2/scala/meta/internal/metals/MultilineStringFormattingProvider.scala
tgodzik: Ok, let's leave it this way. I might have been overthinking this.

tests/unit/src/test/scala/tests/OnTypeFormattingSuite.scala
```diff
@@ -374,12 +445,21 @@ class OnTypeFormattingSuite extends BaseLspSuite("onTypeFormatting") {
              |/a/src/main/scala/a/Main.scala
       """.stripMargin + base
         )
+        _ <- if (!stripMarginEnabled)
+          server.didChangeConfiguration(
+            """{
+              |  "enable-strip-margin-on-type-formatting": flase
```
tgodzik, line 451:
```suggestion
              |  "enable-strip-margin-on-type-formatting": false
```

Issue #1345, opened 2020-01-23
False "Cannot rename while the code is compiling" warning


olafurpg:
**Describe the bug**

When I execute "rename symbol" I get a warning

> Cannot rename while the code is compiling since it could produce incorrect results.

even if there are no targets compiling. After waiting a couple minutes, I was able to run rename symbol.

**To Reproduce**
I'm only able to reproduce this on a fairly large internal codebase. Just opening here to register that I've hit on this issue and may look into it.

**Expected behavior**

Rename symbol should work when there is no ongoing compilation.

**Screenshots**

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.8.0

olafurpg:
I was able to reproduce this in an open source project https://github.com/scalameta/munit, demo

![2020-01-23 23 39 48](https://user-images.githubusercontent.com/1408093/73033399-c822c000-3e39-11ea-8464-0b49e50cda19.gif)

tgodzik:
@ltrojanowski will take a look at it. The idea is to actually try to run the compile if it's compiling to get the future, which should just quickly return if it's a nop compile.  It would be also good to see where is the race condition for the running compilations map.

PR #1754, opened 2020-05-15
compile before rename instead of checking if compilation is not ongoing

ltrojanowski:
The pull request relates to the issue described in https://github.com/scalameta/metals/issues/1345

I did try to reproduce the issue, both on the code base mentioned in the bug report and another big code base to which I have access, but I couldn't reproduce it. As such I don't really know what caused the behavior. I suspect a race condition, a possible solution could be to run the compilation at every rename. If the file has already been compiled (most of the time), the future would return very quickly, in case a compilation is required it would do so and then go on with the rename.

I don't personally think this is particularly elegant and would much rather fix the underlying source of the problem with Compilations returning an ongoing compilation, unfortunalty I don't see a reason for this behavior. If anyone has a hunch where to look for it I would be interested in your opinion.


metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -77,105 +78,121 @@ final class RenameProvider(
...
-              id.exists(_.symbol.desc.name.toString == "apply")
-            case _ => false
+    def compilationFinishedSuccessfully(
+        source: AbsolutePath
+    ): Future[Boolean] = {
+      if (compilations.currentlyCompiling.isEmpty) {
+        Future(true)
+      } else {
+        compilations.cascadeCompileFiles(Seq(source)).map { compilationResult =>
+          if (compilationResult.getStatusCode() == StatusCode.OK) {
```
tgodzik, line 88: We most likely don't have to fail in case of compile errors. SemanticDB files will still be available and rename will only fail in case of new non-compiled symbols, which is the current behavior.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -77,105 +78,121 @@ final class RenameProvider(
       params: RenameParams,
       token: CancelToken
   ): Future[WorkspaceEdit] = {
-    if (!compilations.currentlyCompiling.isEmpty) {
-      client.showMessage(isCompiling)
```
tgodzik, line 81: We should remove the `isCompiling` message

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
ltrojanowski: removed check of compilation result

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -77,11 +77,17 @@ final class RenameProvider(
...
+        source: AbsolutePath
+    ): Future[Unit] = {
+      if (compilations.currentlyCompiling.isEmpty) {
+        Future(())
+      } else {
+        compilations.cascadeCompileFiles(Seq(source)).map { _ => () }
+      }
+    }
+    val source = params.getTextDocument.getUri.toAbsolutePath
+    compilationFinished(source).flatMap { _ =>
```
tgodzik, line 90: This should also be added to the `prepareRename` method, which is normally run before rename itself.

ltrojanowski: fixed

PR #1758, opened 2020-05-17
Handle workspaces with multiple build files.

### LINKED ISSUE #1757: Better support for workspaces with multiple build tool files
Take a repo that has both an SBT build and a mill build. You can use [uPickle](https://github.com/lihaoyi/upickle) as an example. This is obviously a mill build with a very minimal old `build.sbt` file as well. In the current situation, Metals will detect that this is an SBT workspace, and prompt the user to import via SBT. There is really no way to have Metals recognize this as a Mill space apart from removing the sbt related files since we check for SBT first, so it always wins.

I'd like to be give the choice that when Metals recognizes a workspace has both Mill and SBT files to choose which I'd like to use.

Delete all the other builds files.

**Additional context**
We do something similar to this when there are multiple BSP servers. Some of the logic that would need to change is located here:

https://github.com/scalameta/metals/blob/a39ac72880bde64d337e5c553f6017dabd40d7b7/metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala#L87-L94

I imagine we could just collect those instead of finding the first true, see if there are more than one, and then prompt the user to decide.

**Search terms**:
mill, sbt, multiple build files

ckipp01:
Previously this always defaulted to the first found, many times leading
  to an SBT build if a workspace had both a `build.sbt` and a `build.sc`
  file. Now both are grabbed and the user is prompted to choose whether
  the space if for build tool A or build tool B. This choice is remembered
  and they are not asked about this again.

### Chosen Approach

I went back and forth on how the flow should go for this. I really wanted that when it was detected that multiple build definitions were found that it'd prompt like the following:

```
New workspace found. Would you like to import:?
1. Import sbt build
2. Import mill build
3. Not now
4. Don't show again
```

However, after doing that it was quite messy since you have to carry a list of `BuildTool` round to multiple methods, which all would have to handle if there was one more many. I found it way simpler and less of a change to simply prompt the user right away when multiple were found to get their intention for the workspace. Then all of the rest of the logic can remained unchanged. I'm still open to the alternative way that I outlined I guess, but I really think it will messier and more code without much benefit.

All that to say, I wanted to throw this up here to make sure this approach was ok before I started to write tests for this.

Closes #1757

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
@@ -14,6 +14,8 @@ import scala.meta.io.AbsolutePath
 import scala.meta.internal.process.ProcessHandler
 import scala.meta.internal.process.ExitCodes
 import scala.meta.internal.metals.Messages._
+import org.eclipse.lsp4j.MessageActionItem
+import scala.meta.internal.semver.SemVer
 
 /**
  * Runs `sbt/gradle/mill/mvn bloopInstall` processes.
@@ -166,6 +168,15 @@ final class BloopInstall(
     }
   }
 
+  def checkForChosenBuildTool(
+      buildTools: List[BuildTool]
+  ): Future[Option[BuildTool]] =
+    tables.buildTool.selectedBuildTool match {
+      case Some(chosen) =>
+        Future(buildTools.find(_.executableName == chosen))
+      case None => requestBuildToolChoice(buildTools)
+    }
+
   private def persistChecksumStatus(
       status: Status,
       buildTool: BuildTool
@@ -201,4 +212,38 @@ final class BloopInstall(
       }
   }
 
+  private def requestBuildToolChoice(
+      buildTools: List[BuildTool]
+  ): Future[Option[BuildTool]] = {
+    languageClient
+      .showMessageRequest(ChooseBuildTool.params(buildTools))
+      .asScala
+      .map { choice =>
+        val selectedBuildTool =
+          buildTools.find(buildTool =>
+            new MessageActionItem(buildTool.executableName) == choice
+          )
+        selectedBuildTool match {
+          case Some(buildTool) => {
+            val isCompatibleVersion = SemVer.isCompatibleVersion(
+              buildTool.minimumVersion,
+              buildTool.version
+            )
+            if (isCompatibleVersion) {
+              tables.buildTool.chooseBuildTool(choice.getTitle)
+              Some(buildTool)
+            } else {
+              scribe.warn(
+                s"Unsupported $buildTool version ${buildTool.version}"
+              )
+              languageClient.showMessage(
+                Messages.IncompatibleBuildToolVersion.params(buildTool)
+              )
+              None
+            }
+          }
+          case None => None
+        }
+      }
+  }
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
@@ -15,7 +15,6 @@ import scala.meta.internal.process.ProcessHandler
 import scala.meta.internal.process.ExitCodes
 import scala.meta.internal.metals.Messages._
 import org.eclipse.lsp4j.MessageActionItem
-import scala.meta.internal.semver.SemVer
 
 /**
  * Runs `sbt/gradle/mill/mvn bloopInstall` processes.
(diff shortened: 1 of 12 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1406,44 +1406,51 @@ class MetalsLanguageServer(
...
-          None
-        }
-      case None =>
-        if (!buildTools.isAutoConnectable) {
-          warnings.noBuildTool()
+          Future(None)
         }
-        None
+      }
+      case buildTools @ head :: tail =>
```
tgodzik, line 1432: `head :: tail` probably not needed.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1406,44 +1406,51 @@ class MetalsLanguageServer(
         .orNull
     }.asJava
 
-  private def supportedBuildTool(): Option[BuildTool] =
+  private def supportedBuildTool(): Future[Option[BuildTool]] =
     buildTools.loadSupported match {
```
tgodzik, line 1410: Maybe we should filter the ones with incompatible versions and show a warning for those?

metals/src/main/scala/scala/meta/internal/metals/BloopInstall.scala
```diff
@@ -201,4 +212,38 @@ final class BloopInstall(
...
+          buildTools.find(buildTool =>
+            new MessageActionItem(buildTool.executableName) == choice
+          )
+        selectedBuildTool match {
+          case Some(buildTool) => {
+            val isCompatibleVersion = SemVer.isCompatibleVersion(
+              buildTool.minimumVersion,
+              buildTool.version
+            )
+            if (isCompatibleVersion) {
```
tgodzik, line 232: This seems duplicated with `MetalsLanguageServer.supportedBuildTool`, if we filter them before, there will be no need for checking it again.

ckipp01:
> I added some small comments, but I think it's a good direction. Both versions you proposed seem OK to me and if this causes less changes then I don't think we should go the other way.

Sounds good 👍 I'll keep going this route then.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
ckipp01: So one thing I thought about while doing this is in a situation where a user has an sbt and a mill file for example, I think it's important for the choice to take priority. Because let's say they are on an unsupported sbt version, but they want sbt, then it's probably better for them to choose sbt and then get a warning and "no build tool" message rather than have sbt filtered out, and mill chosen even if that's not what they want.

What do you think?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
ckipp01: True, good point.

kpbochenek:
Cool improvement! :+1:

Can we also have a possibility to see which build system was used for import?
Maybe in a Metals Doctor? Just random thought.

A db will not be wiped when I do "Build import", right? It would be good to give user an option to reset chosen build tool, what do you think?

ckipp01:
> Cool improvement! 👍
> 
> Can we also have a possibility to see which build system was used for import?
> Maybe in a Metals Doctor? Just random thought.

Ah, I hadn't thought of that, but yea, that's a cool idea. Let me play around with this to make sure there is a good minimal way to display this.

> A db will not be wiped when I do "Build import", right? It would be good to give user an option to reset chosen build tool, what do you think?

Yea the decisions is persisted, so currently, apart from delete the `.metals/` directory, there wouldn't be a way to reset this. However, this is sort of a problem that exists in multiple other message requests as well though right?

kpbochenek:
> Yea the decisions is persisted, so currently, apart from delete the `.metals/` directory, there wouldn't be a way to reset this. However, this is sort of a problem that exists in multiple other message requests as well though right?

You are probably right, I don't know how it works for other messages :sweat_smile:

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: Sure, that's probably better.

ckipp01:
If a user has two build definitions and they had to choose one, a short message will be added to their doctor output just saying `Your <build tool> build definition has been imported`. Here is an example:

Screenshot 2020-05-18 at 18 05 42

tests/unit/src/main/scala/tests/TestingClient.scala
```diff
@@ -244,6 +260,9 @@ final class TestingClient(workspace: AbsolutePath, buffers: Buffers)
           CheckDoctor.moreInformation
         } else if (SelectBspServer.isSelectBspServer(params)) {
           params.getActions.asScala.find(_.getTitle == "Bob").get
+        } else if (isSameMessageFromList(ChooseBuildTool.params)) {
+          params.getActions.asScala.toList.headOption
```
tgodzik, line 264: Could we look for `sbt` instead? If the order changes at any point the test will fail.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1406,44 +1406,51 @@ class MetalsLanguageServer(
...
-        }
-      case None =>
-        if (!buildTools.isAutoConnectable) {
-          warnings.noBuildTool()
+          Future(None)
         }
-        None
+      }
+      case buildTools =>
+        bloopInstall.checkForChosenBuildTool(buildTools)
```
tgodzik, line 1433: Can we extract `isCompatibleVersion` to a function (maybe just inner function) and use it both for 2. and 3. case ? This way we will not need to do it in `checkForChosenBuildTool`

tests/unit/src/main/scala/tests/TestingClient.scala
ckipp01: Ah, good point 👍

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
ckipp01: Yup, done.

ckipp01:
> I would like to merge the Ammonite PR first, since it can have conflicts with this one.

No worries! I figured there would be some conflicts. I'll rebase once it's merged.

PR #1762, opened 2020-05-18
Keep modifiers when implementing abstract members

### LINKED ISSUE #1759: "Implement all members" ignores access modifier
**Describe the bug**
When a concrete class has unimplemented members, the quick fix to "Implement all members" ignores any access modifier, generally `protected` (and instead adds an `override`).

**To Reproduce**
Self explanatory.

**Expected behavior**
Keep the intended access modifier, and preferably not add the semantically inaccurate `override`.

**Additional context**
Personally I dislike having `override` for methods that don't, strictly speaking, override.
But it's a matter of style, so it would be nice to have that as an optional setting.

kpbochenek:
Fixes #1759

Not sure if this is a bug or intended behaviour but `flagString` generates modifiers like `protected[package x.y.z]` and they need to be stripped manually.

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
@@ -209,7 +209,11 @@ trait OverrideCompletions { this: MetalsGlobal =>
         // https://github.com/scalameta/metals/issues/565#issuecomment-472761240
         else ""
 
-      val _modifs = sym.flagString.replace("package ", "")
+      val _modifs = sym.flagString
+        .replace("package ", "")
+        .replace("object ", "")
+        .replace("trait ", "")
+        .replace("class ", "")
       val modifs =
         if (_modifs.isEmpty) ""
         else _modifs + " "
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
@@ -209,9 +209,10 @@ trait OverrideCompletions { this: MetalsGlobal =>
         // https://github.com/scalameta/metals/issues/565#issuecomment-472761240
         else ""
 
-      val lzy: String =
-        if (sym.isLazy) "lazy "
-        else ""
+      val _modifs = sym.flagString.replace("package ", "")
+      val modifs =
+        if (_modifs.isEmpty) ""
+        else _modifs + " "
 
       val keyword: String =
         if (isVarSetter(sym)) "var "
@@ -232,7 +233,7 @@ trait OverrideCompletions { this: MetalsGlobal =>
 
       val name: String = Identifier(sym.name)
 
-      val filterText: String = s"${overrideKeyword}${lzy}${keyword}${name}"
+      val filterText: String = s"${overrideKeyword}${modifs}${keyword}${name}"
 
       // if we had no val or def then filter will be empty
       def toMember = new OverrideDefMember(
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
@@ -213,11 +213,11 @@ trait OverrideCompletions { this: MetalsGlobal =>
         if (sym.isLazy) "lazy "
         else ""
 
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
```diff
@@ -213,6 +213,16 @@ trait OverrideCompletions { this: MetalsGlobal =>
         if (sym.isLazy) "lazy "
         else ""
 
+      val _modifs = sym.flagString
+        .replace("package ", "")
```
tgodzik, line 217: Do we need to do this? We are not doing overrides for any of those symbol types.

mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
```diff
@@ -233,6 +243,8 @@ trait OverrideCompletions { this: MetalsGlobal =>
       val name: String = Identifier(sym.name)
 
       val filterText: String = s"${overrideKeyword}${lzy}${keyword}${name}"
+      val insertText: String =
+        s"${overrideKeyword}${modifs}${keyword}${name}$signature"
```
tgodzik, line 247: Will this work if we want to override lazy vals? They cannot be abstract but we can override them and this compilation is also used for non-abstract symbols.

mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
kpbochenek:
```
Not sure if this is a bug or intended behaviour but flagString generates modifiers like protected[package x.y.z] and they need to be stripped manually.
```

mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
kpbochenek: lazy should be in modifs already, I wanted to just get rid of lzy and only use modifs but it is used to match against what user typed and matching against modifiers is probably not what user wants.
Gonna double check lazy is there :+1:

mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
tgodzik: What about `object`, `trait` or `class` ?

mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
tgodzik: Cool, let's add a test if there is none.

mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
kpbochenek: handled already

mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
kpbochenek: yes, there is already a test :+1:
```
  checkEditLine(
    "lazy",
    """|package q
       |abstract class Val {
       |  lazy val hello: Int = 2
       |}
       |class Main extends Val {
       |   ___
       |}
       |""".stripMargin,
    "override val hel@@",
    "override lazy val hello: Int = ${0:???}"
  )
```

mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
tgodzik: I meant, does those keywords can appear here even? You mentioned `protected[package x.y.z] `, but why would trait appear here?

kpbochenek:
```
scala> trait A {
     |   trait AB {
     |     private[A] val x = 3
     |   }
     | }
trait A
```

tgodzik: We could do:
```
      val _modifs =
        sym.flagString.replace(
          sym.privateWithin.toString(),
          sym.privateWithin.name.toString()
        )
```

Issue #1514, opened 2020-03-18
Can't go-to source for Java 11 code when Java is installed via brew


Billzabob:
**Describe the bug**
Can't go-to source for Java 11 code when Java is installed via brew.

**To Reproduce**
1. `brew cask install java11`
2. Use metals to go to definition for Java 11 specific code, Ex: `java.net.http.HttpClient`
3. Metals will say `No definition found for 'HttpClient'`

**Expected behavior**
Metals should take you to the source file.

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.8.1

**Additional context**
This is apparently only a problem when installing Java 11 via brew. It might be related to the `src.zip` file being located in `$JAVA_HOME/lib` instead of directly in `$JAVA_HOME`.

Previously metals was reporting false errors around the Java 11 code. I had to change the `blp-server` script to call `java` directly instead of specifically calling java8. Now metals shows that the code compiles but still doesn't allow goto.

**Search terms**
java11, brew, goto

ckipp01:
Hi @Billzabob, are you still having this issue with the newest version of Metals?

Billzabob:
Hi @ckipp01, yes I'm on 0.9.0 and it still fails for me. Metals says it doesn't compile and go-to doesn't work. I guess I can't say whether it's a Metals or a Bloop issue. When I try to compile with Bloop I get errors like this:
```
[E] [E1] src/main/scala/dissonance/utils.scala:5:17
[E]      object http is not a member of package java.net
[E]      L5: import java.net.http.HttpClient
[E]                 ^^^^^^^^^^^^^
```

tgodzik:
@Billzabob Do you have a repo that we could reproduce it on? Bloop should be able to compile any project and if there is an issue it would be awesome to try to fix it on out side.

Billzabob:
This is the one I'm working on and experiencing the issue with:
https://github.com/Billzabob/dissonance
I'm guessing it's not an issue with the repo though, it's probably some sort of environmental issue. It requires Java 11 to compile. I can simply do `sbt compile` and that works fine but I can't get bloop to compile. How does bloop determine what Java version to use? My JAVA_HOME is setup to use Java 11.

tgodzik:
I think it uses the `platform` section inside the json configuration files. And that is set based on the version that was used to run sbt and probably the same as Metals is run in. I can experiment a bit with your repo and see what's going on.

Billzabob:
[...]

PR #1763, opened 2020-05-18
Support java files inside java modules for go to definition

tgodzik:
Previously, we would not find any java files from dependencies in java modules. Now, we make sure we java symbols are added as non-trivial for java modules.

Fixes https://github.com/scalameta/metals/issues/1514

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
@@ -13,6 +13,7 @@ import scala.meta.internal.{semanticdb => s}
 import scala.meta.io.AbsolutePath
 import scala.util.control.NonFatal
 import java.util.zip.ZipError
+import scala.util.Properties
 
 /**
  * An implementation of GlobalSymbolIndex with fast indexing and low memory usage.
@@ -40,7 +41,7 @@ final class OnDemandSymbolIndex(
     onError: PartialFunction[Throwable, Unit] = PartialFunction.empty
 ) extends GlobalSymbolIndex {
   val mtags = new Mtags
-  private val sourceJars = new ClasspathLoader()
+  private val sourceJars = new ClasspathLoader(onError)
   var indexedSources = 0L
   def close(): Unit = sourceJars.close()
   private val onErrorOption = onError.andThen(_ => None)
@@ -192,7 +193,9 @@ final class OnDemandSymbolIndex(
         case Some(file) =>
           addMtagsSourceFile(file)
         case _ =>
-          loadFromSourceJars(trivialPaths(toplevel)).foreach(addMtagsSourceFile)
+          loadFromSourceJars(trivialPaths(toplevel))
+            .orElse(loadFromSourceJars(modulePaths(toplevel)))
+            .foreach(addMtagsSourceFile)
       }
     }
     if (!definitions.contains(symbol.value)) {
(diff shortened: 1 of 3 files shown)


tgodzik:
Seems that we are using up too much memory on window, which most likely because we are adding each class in the standard library. I will need to rethink the approach.

tgodzik:
Ok, this is good to review now. I changed to using the `getModule` method, because this way we don't have to implement everything from scratch. The only problem is that I need to use reflection here as that method is not available for Java 8.

tests/unit/src/test/scala/tests/JdkSourcesSuite.scala
```diff
@@ -1,9 +1,26 @@
...
 import scala.meta.internal.metals.JdkSources
+import scala.meta.internal.mtags.OnDemandSymbolIndex
+import scala.meta.internal.mtags.Symbol
 
 class JdkSourcesSuite extends BaseSuite {
   test("src.zip") {
     JdkSources.getOrThrow()
   }
+
+  test("index-src.zip") {
```
kpbochenek, line 12: index-modules-src.zip

mtags/src/main/scala/scala/meta/internal/mtags/ClasspathLoader.scala
```diff
@@ -77,6 +78,10 @@ final class ClasspathLoader() {
     loader.resolve(path)
   }
 
+  def loadClass(symbol: String): Option[Class[_]] = {
+    Try(loader.loadClass(symbol)).toOption
```
kpbochenek, line 82: I would do logging if error happens. It will help investigate if something will start failing.

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
```diff
@@ -231,11 +234,33 @@ final class OnDemandSymbolIndex(
...
+          val module = cls.getClass().getMethod("getModule").invoke(cls)
+          val moduleName = module.getClass().getMethod("getName").invoke(module)
+
+          List(
+            s"$moduleName/$noExtension.java",
+            s"$moduleName/$noExtension.scala"
+          )
+        }
+      } catch {
+        case NonFatal(_) => Nil
```
kpbochenek, line 257: I would do logging if error happens. It will help investigate if something will start failing.

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
```diff
@@ -231,11 +234,33 @@ final class OnDemandSymbolIndex(
...
     )
   }
 
+  private def modulePaths(toplevel: Symbol): List[String] = {
+    if (Properties.isJavaAtLeast("9")) {
+      val noExtension = toplevel.value.stripSuffix(".").stripSuffix("#")
+      val javaSymbol = noExtension.replace("/", ".")
+      try {
+        sourceJars.loadClass(javaSymbol).toList.flatMap { cls =>
+          // note(@tgodzik) Modules are only available in Java 9+, so we need to ivoke this reflectively
```
kpbochenek, line 247: typo, 'invoke'

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
```diff
@@ -231,11 +234,33 @@ final class OnDemandSymbolIndex(
...
+          val module = cls.getClass().getMethod("getModule").invoke(cls)
```
kpbochenek, line 248: for unnamed module it will return null

tests/unit/src/test/scala/tests/JdkSourcesSuite.scala
tgodzik: it's not neccessarily modules, we test it also for JDK 8, which doesn't have modules.

mtags/src/main/scala/scala/meta/internal/mtags/ClasspathLoader.scala
tgodzik: We do a lot of guessing here, some weirdly defined classes or cases where we will never find the symbol will just create additional noise.

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
tgodzik: The same as in the other case, this might just create noise for the user.

mtags/src/main/scala/scala/meta/internal/mtags/ClasspathLoader.scala
tgodzik: Added onError, which will only do `scribe.debug`, which should not add noise unless the user turns on the debug flag.

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
tgodzik: Added onError

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
tgodzik: Fixed

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
tgodzik: Added options everywhere to be on the safe side.

PR #1771, opened 2020-05-21
Add file watcher for the standalone ammonite scripts

tgodzik:
Previously, watching a single source file would result with getting updates for the entire directory. Now, since adding Ammonite scripts it's more important to only follow updates on certain files. Especially when we create *.sc files at a root of a workspace.


kpbochenek:
But you are still watching parent directory of given files, how does it improve anything? I don't see it :P

tgodzik:
The idea is that we don't invoke didChange on unrelated files. This is especially a problem in case of ammonite files, which can be created anywhere including root. This will cause everything in the workspace to be reported. This change makes sure we only report related files.

kpbochenek:
Cool that makes sense :+1:

One more question, does watching a directory means you are watching everything recursively or just files/directories directly in specified folder?

PR #1774, opened 2020-05-22
Update Vim docs to help recognize *.sc files as scala

s-u-m-e:
I struggled to get the worksheet working with the current recommended mappings. This allows worksheets to be recognized.

### DIFF
--- a/docs/editors/vim.md
+++ b/docs/editors/vim.md
@@ -58,7 +58,7 @@ following settings.
 Plug 'neoclide/coc.nvim', {'branch': 'release'}
 
 " Configuration for vim-scala
-au BufRead,BufNewFile *.scala,*.sc set filetype=scala
+au BufRead,BufNewFile *.sbt,*.sc set filetype=scala
 ```
 
 Run `:PlugInstall` to install the plugin. If you already have `coc.nvim`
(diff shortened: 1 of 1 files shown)


docs/editors/vim.md
```diff
@@ -58,7 +58,7 @@ following settings.
 Plug 'neoclide/coc.nvim', {'branch': 'release'}
 
 " Configuration for vim-scala
-au BufRead,BufNewFile *.sbt set filetype=scala
+au BufRead,BufNewFile *.sbt,*.sc set filetype=scala
```
ckipp01, line 61: Good catch. Can we actually change the note here to explain a bit as to why it is being done. Something like _Necessary for Vim to recognize sbt, script, and worksheet files as Scala._

Also, are then able to move this entire chunk down to the recommended mappings file? It's a bit better fitting there since this isn't related to a specific plugin, but rather a general setting.

docs/editors/vim.md
```diff
@@ -303,7 +303,7 @@ open buffer.
 
 ## Worksheets
 
-Metals allows users to create a `*.worksheet.sc` file and see evaluations right
+Metals allows users to create a `src/main/scala/*.worksheet.sc` file and see evaluations right
```
ckipp01, line 306: So I actually wouldn't change this, or at least not without further explanation. Worksheets actually still work at the root of your project. They won't have access to everything in your package like they would if they were in `src/main/scala`, but the worksheet still works and evaluates. So if you are going to change this at all, you may want to make that distinction.

PR #1646, opened 2020-04-20
Organize imports with Scalafix

olafurpg:
This PR replaces the `RemoveUnused` Scalafix rule with a new `OrganizeImports` rule that was created by @liancheng. Massive thanks to @liancheng! I have wanted this for a long time 😄 

Marked as a draft since `OrganizeImports` still doesn't remove unused imports but there is a draft PR that implements this functionality https://github.com/liancheng/scalafix-organize-imports/pull/9

Are everyone OK with enabling this rule once it can remove unused imports?

PR #1775, opened 2020-05-22
Organize imports with Scalafix

tgodzik:
Follow up from https://github.com/scalameta/metals/pull/1646/files


olafurpg:
cc/ @liancheng

PR #1776, opened 2020-05-22
Bumb Scalafmt version.

ckipp01:
Since we are going to merge in #1775, I figured this would be a good time to bump the version of scalafmt to 2.5.2 as well.

Wait until that pr is merged, and you should see the other two commits via @tgodzik disappear here, but I wanted to merge off that work for this.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,9 +1,9 @@
-version = "2.4.2"
+version = "2.5.2"
 project.git = true
-align = none
-docstrings = javadoc
-danglingParentheses = true
+align.preset=none
+align.stripMargin = true
 assumeStandardLibraryStripMargin = true
+docstrings = javadoc
 binPack.literalsExclude = ["Term.Name"]
 onTestFailure = """
 To fix this problem:
--- a/metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
@@ -1,10 +1,12 @@
 package scala.meta.internal.builds
 import java.nio.file.Files
 import java.nio.file.Path
+
+import scala.util.Properties
+
 import scala.meta.internal.metals.BuildInfo
 import scala.meta.internal.metals.UserConfiguration
 import scala.meta.io.AbsolutePath
-import scala.util.Properties
 
 case class GradleBuildTool(userConfig: () => UserConfiguration)
     extends BloopPluginBuildTool {
--- a/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
@@ -1,9 +1,11 @@
 package scala.meta.internal.builds
-import scala.meta.internal.metals.UserConfiguration
-import scala.meta.io.AbsolutePath
-import scala.util.Properties
 import java.nio.file.Files
 import java.nio.file.Path
+
+import scala.util.Properties
+
+import scala.meta.internal.metals.UserConfiguration
+import scala.meta.io.AbsolutePath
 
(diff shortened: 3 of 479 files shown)


.scalafmt.conf
```diff
@@ -1,8 +1,7 @@
-version = "2.4.2"
+version = "2.5.2"
 project.git = true
-align = none
+align.preset=none
 docstrings = javadoc
-danglingParentheses = true
```
ckipp01, line 5: As far as i can tell, the multiple settings of `danglingParentheses` all default to `true`.

https://scalameta.org/scalafmt/docs/configuration.html#danglingparentheses

ckipp01:
So there is a crazy amount of changes with this. Are there any addition settings that the rest of you want to add or take away with the bump to 2.5.x?

metals/src/main/scala/scala/meta/internal/builds/GradleBuildTool.scala
```diff
@@ -17,27 +17,27 @@ case class GradleBuildTool(userConfig: () => UserConfiguration)
     val isSnapshotVersion = BuildInfo.gradleBloopVersion.contains("+")
     if (isSnapshotVersion)
       """|maven{
-         |  url 'https://dl.bintray.com/scalacenter/releases'
-         |}""".stripMargin
+        |  url 'https://dl.bintray.com/scalacenter/releases'
```
ckipp01, line 20: I really dislike the way it does these multiline strings in this way. Thoughts?

metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
```diff
@@ -40,7 +40,9 @@ case class MillBuildTool(userConfig: () => UserConfiguration)
     // In some environments (such as WSL or cygwin), mill must be run using interactive mode (-i)
     val iOption = if (Properties.isWin) List("-i") else Nil
     val cmd =
-      iOption ::: "--predef" :: predefScriptPath(millVersion).toString :: "mill.contrib.Bloop/install" :: Nil
+      iOption ::: "--predef" :: predefScriptPath(
+        millVersion
```
tgodzik, line 44: This is a bit weird, could split:
```scala
iOption ::: "--predef" :: predefScriptPath(millVersion).toString ::
 "mill.contrib.Bloop/install" :: Nil
```

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -130,10 +130,10 @@ object SbtBuildTool {
       ""
     }
     s"""|// DO NOT EDIT! This file is auto-generated.
-        |// This file enables sbt-bloop to create bloop config files.
-        |$resolvers
-        |addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "$bloopSbtVersion")
-        |""".stripMargin
+       |// This file enables sbt-bloop to create bloop config files.
```
tgodzik, line 133: Should this be changed after adding the align setting?

ckipp01: Yea there still seems to be some formatting on multilines happening after the addition of the align setting. I'll look at this week and ensure everything is working fine and address the points above. Plus, some of the formatting on the multilines are also breaking tests.

ckipp01:
Also just a note on this. The issue with interpolated string that I have mentioned [here](https://github.com/scalameta/scalafmt/issues/1998) is sort of a blocker on us doing this update. We have a lot of multiline strings that are being checked in tests etc, so this breaks quite a bit. I plan on waiting for that issue to addressed, and then I'll continue on with this.

tgodzik:
@ckipp01 There is a new release, should we get back to it?

ckipp01:
> @ckipp01 There is a new release, should we get back to it?

Yes, I'm actually going to close this and open another to not deal with the rebase.

PR #1778, opened 2020-05-25
Remove logging from ClasspathLoader 

tgodzik:
Previously, we would if a class was not found and in most cases this is not an issue, since it's only needed for checking Java 9 + modules. Now, we only return an Option.

Previously I thought this would be under the debug flag, but this might have changed recently and these logs will not be too useful or even confuse the user.


kpbochenek:
so now we have ammonite flaky tests ech

tgodzik:
> so now we have ammonite flaky tests ech

Yeah, I need to take a look at that.

PR #1780, opened 2020-05-26
update bloop to 1.4.1

alexbalonperin:


alexbalonperin:
Two errors:
- `CI / macOS-latest jdk-11 unit tests` is failing because of a dependency resolution timeout. Rerunning it would probably fix the issue. I have confirmed that the file is there: https://dl.bintray.com/sbt/sbt-plugin-releases/com.eed3si9n/sbt-buildinfo/scala_2.12/sbt_1.0/0.9.0/ivys/ivy.xml
- `CI / Gradle MacOS integration (pull_request)` & `CI / Gradle integration (pull_request)` are most likely failing because of the issue described here: https://github.com/scalacenter/bloop/issues/1286

tgodzik:
Thanks for opening the PR! The gradle issue I think might be related to the issue in Bloop and we should only upgrade once it is fixed.

Is there any particular reason why you would want ti update the Bloop version in Metals? The current version is only a couple of commits behind and you can always change the version in the settings if you indeed need it.

alexbalonperin:
No particular reason really. I just thought it would be good to update to a stable version instead of a release candidate but it seems that 1.4.1 might not be that stable after all. Feel free to close the PR if it is premature.

tgodzik:
I will close it for now and upgrade once we fix the problem.

Issue #1765, opened 2020-05-15
Add -J to JVM options automatically when running


Pradhyo:
Our project uses [pureconfig](https://github.com/pureconfig/pureconfig) and a non-default config file path by setting `config.file` as described [here](https://github.com/lightbend/config#standard-behavior).

The path works fine and the project runs as expected in Eclipse.
![](https://i.stack.imgur.com/JABnf.png)

I tried these methods to set the value but it couldn't pick it up:

1. putting this in a `.jvmopts`in the root directory
`-Dconfig.file=/path/to/config/file`

2. Setting it in the extension settings
![image](https://user-images.githubusercontent.com/7895492/82101798-831de480-96db-11ea-85a5-43c3f8c6a1f7.png)

What worked was using [System.setProperty](https://www.tutorialspoint.com/java/lang/system_setproperty.htm) right before using `pureconfig.loadConfigOrThrow`

`System.setProperty("config.file", "/path/to/config/file")`

Pradhyo:
Got it - please close this issue as you see fit, I am good. Thanks again for the quick response!

singh-abhijeet:
> Thanks for reporting! The `server properties` are properties needed specifically for the Metals language server and they are not used when running. You need to specify a run configuration first:
> 
> ```
>     {
>       "type": "scala",
>       "request": "launch",
>       "name": "MyMain",
>       "mainClass": "com.example.MyMain",
>       "args": [],
>       "jvmOptions": ["-J-Dconfig.file=/path/to/config/file"]
>     }
> ```

@tgodzik Hi, trying this approach for passing `--add-exports` is giving error: 
`Unrecognized option: --add-exports java.base/sun.nio.ch=ALL-UNNAMED`

This is how my launch.json looks:
```
{
    "version": "0.2.0",
    "configurations": [
      // Main class configuration
      {
        "type": "scala",
        "request": "launch",
        // configuration name visible for the user
        "name": "Launch Main",
        // full name of the class to run
        "mainClass": "SparkStreamingMain",
        // optional arguments for the main class
        "args": [],
        // optional jvm properties to use
        "jvmOptions": ["--add-exports java.base/sun.nio.ch=ALL-UNNAMED"]
      }
    ]
  }
```

Let me know how to pass [--add-exports](https://stackoverflow.com/questions/40313090/java-9-options-add-exports-vs-xaddexports-not-recognized) via jvmOpts.

singh-abhijeet:
[...]

PR #1781, opened 2020-05-27
Automatically add -J to jvm properties and make sure run JVM targets fists

tgodzik:
This PR addresses 2 small issues:
- Add automatically `-J` for Bloop so that users don't have to remember about it - fixes https://github.com/scalameta/metals/issues/1765
- Make sure that we use a more relevant build target if none was specified, which was an issue in scalameta. When running tests we would pick up js tests first, which are not supported on the same level as jvm ones.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -52,6 +52,27 @@ final class BuildTargets() {
   private val buildTargetInference =
     new ConcurrentLinkedQueue[AbsolutePath => Seq[BuildTargetIdentifier]]()
 
+  val buildTargetsOrder: BuildTargetIdentifier => Int = {
+    (t: BuildTargetIdentifier) =>
+      var score = 1
+
+      val isSupportedScalaVersion = scalaInfo(t).exists(t =>
+        ScalaVersions.isSupportedScalaVersion(t.getScalaVersion())
+      )
+      if (isSupportedScalaVersion) score <<= 2
+
+      val isJVM = scalacOptions(t).exists(_.isJVM)
+      if (isJVM) score <<= 1
+
+      // note(@tgodzik) once the support for Scala 3 is on par with Scala 2 this can be removed
+      val isScala2 = scalaInfo(t).exists(info =>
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -226,6 +227,26 @@ class DebugProvider(
...
   }
 
+  private def translateJvmParams(
+      parameters: b.DebugSessionParams
+  ): b.DebugSessionParams = {
+    parameters.getData match {
+      case json: JsonElement if parameters.getDataKind == "scala-main-class" =>
+        json.as[b.ScalaMainClass].foreach { main =>
+          val translated = main.getJvmOptions().asScala.map { param =>
+            if (!param.startsWith("-J"))
```
kpbochenek, line 237: good idea to act smart here! :+1:

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -52,6 +52,27 @@ final class BuildTargets() {
...
+      if (isJVM) score <<= 1
+
+      // note(@tgodzik) once the support for Scala 3 is on par with Scala 2 this can be removed
+      val isScala2 = scalaInfo(t).exists(info =>
+        !ScalaVersions.isScala3Version(info.getScalaVersion())
+      )
+      if (isScala2) score <<= 1
+
+      score
+  }
```
ckipp01, line 74: So I _think_ I understand what is going on here, but why when used up above do we so a `sortBy` and below we use `max`?. Also, what's the benefit of using binary shift operators here? I've actually never came across these in the wild before 😬

tgodzik: We are using maxBy in a scenario that we know the list is not empty and it produces a single element. `sortBy` still produces a list, which is useful in the other scenario.

Binary shift is I think just an easy way to introduce priority.

ckipp01: Sounds good. Thanks for explaining.

Issue #1195, opened 2019-12-17
Fix resolving the FQCN for breakpoints


marek1840:
#1010 tries to resolve the fully qualified class name corresponding to the breakpoint position by considering the closes preceding type symbol as the enclosing class. This obviously fails for cases like:
```scala
object Foo {
  class Bar

  def foo() = {
>>  println() // here is the breakpoint
  }
}
```

Even though the breakpoint belongs to `Foo` type, we are using `Bar` which results in not stopping at the breakpoint.

PR #1783, opened 2020-05-28
Use parser for detecting enclosing class for a breakpoint

tgodzik:
Previously, we would use toplevels to find the closest class symbol for a breakpoint. Now we use parser to find the enclosing class.

This PR addresses two issues:
- fixes https://github.com/scalameta/metals/issues/1195
- proper breakpoint support for Dotty, which will work  much better with the Dotty parser involved


metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -311,6 +311,20 @@ class Compilers(
       pc.signatureHelp(CompilerOffsetParams.fromPos(pos, token)).asScala
     }.getOrElse(Future.successful(new SignatureHelp()))
 
+  def classForPosition(
```
tgodzik, line 314: Not sure about the time, maybe `enclosingClass` would be better.

metals/src/main/scala/scala/meta/internal/metals/JvmSignatures.scala
```diff
@@ -27,10 +27,20 @@ object JvmSignatures {
           loop(tail, fqcn.append(delimiter))
       }
     }
-    val fqcn = loop(definition.symbol.ownerChain, new StringBuilder)
+    val cls = findClass(definition.symbol)
+    val fqcn = loop(cls.ownerChain, new StringBuilder)
     TypeSignature(fqcn)
   }
 
+  private def findClass(symbol: String): String = {
```
tgodzik, line 35: We also take methods into account now, which in case of java files will mean that the breakpoints will work properly even with an inner class.

metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
```diff
@@ -69,13 +70,6 @@ private[debug] final class SetBreakpointsRequestHandler(
       breakpoint <- response.getBreakpoints
     } yield {
       breakpoint.setSource(originalSource)
-      // note(@tgodzik) this seems to happen from time to time, not exactly sure why
```
tgodzik, line 72: Checked recent test reports and this didn't even pop up in case of issue.

mtags/src/main/scala-3/scala/scala/meta/internal/mtags/MtagsEnrichments.scala
```diff
@@ -1,3 +1,14 @@
 package scala.meta.internal.mtags
 
-object MtagsEnrichments extends CommonMtagsEnrichments
+import dotty.tools.dotc.util.SourcePosition
+import org.eclipse.{lsp4j => l}
+
+object MtagsEnrichments extends CommonMtagsEnrichments{
+
+  def (pos: SourcePosition).toLSP: l.Range = {
```
tgodzik, line 8: Taken from PR by @gabro

tests/unit/src/main/scala/scala/meta/internal/metals/debug/TestDebugger.scala
```diff
@@ -65,7 +65,15 @@ final class TestDebugger(
       .map { response =>
         // the breakpoint notification we receive does not contain the source
         // hence we have to register breakpoints here
-        response.getBreakpoints.foreach(this.breakpoints.register)
+        response.getBreakpoints.foreach { brPoint =>
+          // note(@tgodzik) from time to time breakpoints are sent back without the source,
```
tgodzik, line 69: This should help with the flaky tests better, since we are not sure how it is possible that source is very rarely null.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
marek1840: `enclosingClass` definitely sounds better

tgodzik: changed!

Issue #1370, opened 2020-01-30
Navigation errors on a project with symlinks


aldtid:
**Question**

While working with Metals on VSCode in an SBT project with symlinks, some strange navigation error logs appear if a navigation is performed from a file that is in a directory pointed by a symlink:

https://github.com/aldtid/sbt-project-with-symlinks/blob/be6e1f067fce1783c1706acbe2a03bcc378cff16/metals.log#L100

However, navigation can still be made. As this is my first time trying Metals, I cannot say if these errors prevent a normal extension behavior or everything is working as expected. Are these logs relevant?

Thanks in advance.

**Installation:**

- **Operating system**: Linux (NixOS 19.09)
- **Editor**: Visual Studio Code
- **Metals version**: v0.8.0

**Additional context**

I have uploaded an example project with symlinks to easily reproduce this issue: https://github.com/aldtid/sbt-project-with-symlinks

**Search terms**

navigation, symlink

tgodzik:
Thanks for reporting and the reproduction :heart_eyes: !

If everything is working you can ignore the warning, most likely there is some logic that doesn't follow symlinks.

tgodzik:
Confirmed that Metals doesn't work well with symlinks - Bloop manages to compile everything alright, but we don't translate the links well.

aldtid:
Great, thank you for confirming it!

aldtid:
Is there any plan to fix this issue in next releases?

kpbochenek:
Let us take a look and see if it is something easily fixable :crossed_fingers:

aldtid:
Great, thank you.

kpbochenek:
So we get request for:
`/github/others/sbt-project-with-symlinks/entities/example-entity/core-piece/src/main/scala/com/example/Launcher.scala`
and we call .toAbsolutePath and get:
`/github/others/sbt-project-with-symlinks/pieces/core-piece/src/main/scala/com/example/Launcher.scala`

our workspace is: `/github/others/sbt-project-with-symlinks/entities/example-entity`

that's why we require semanticdb to be here(relative to workspace root):
`/github/others/sbt-project-with-symlinks/entities/example-entity/.bloop/corePiece/bloop-bsp-clients-classes/classes-Metals-zW7pjnPKSb-unn7NCA4LpQ==/META-INF/semanticdb/../../pieces/core-piece/src/main/scala/com/example/Launcher.scala.semanticdb`

[...]

PR #1784, opened 2020-05-28
Fix issues with symlinks in source directories

tgodzik:
Fixes https://github.com/scalameta/metals/issues/1370

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -51,6 +51,8 @@ final class BuildTargets() {
     ConcurrentHashSet.empty[AbsolutePath]
   private val buildTargetInference =
     new ConcurrentLinkedQueue[AbsolutePath => Seq[BuildTargetIdentifier]]()
+  // if workspace contains symlinks, original source items are kept here and source items dealiased
+  private val originalSourceItems = ConcurrentHashSet.empty[AbsolutePath]
 
   val buildTargetsOrder: BuildTargetIdentifier => Int = {
     (t: BuildTargetIdentifier) =>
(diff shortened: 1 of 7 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1801,7 +1801,7 @@ class MetalsLanguageServer(
         item <- i.sources.getItems.asScala
         source <- item.getSources.asScala
       } {
-        val sourceItemPath = source.getUri.toAbsolutePath
+        val sourceItemPath = source.getUri.toAbsolutePath(followSymlink = false)
```
tgodzik, line 1804: This is needed so that we are able to be able to figure out what the original source roots were.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -51,6 +51,8 @@ final class BuildTargets() {
     ConcurrentHashSet.empty[AbsolutePath]
   private val buildTargetInference =
     new ConcurrentLinkedQueue[AbsolutePath => Seq[BuildTargetIdentifier]]()
+  // if workspace contains symlinks, original source items are kept here and source items dealiased
```
kpbochenek, line 54: you only call on this collection `.dealias.toNIO`. You can already keep dealiased path here then a `find` in `originalInverseSourceItem` will not have to do it every time.

kpbochenek: ahh no ok I get it, ignore :D

tgodzik:
> It looks like it should not be a problem but could you double check for a bigger scala repo it doesn't have any performance problems when inside symlink / src symlinked?

Tested it on ZIO and seems to work fine, there shouldn't really be any regressions in terms of performance here. This is only an additional check in case of symlinks.

PR #1785, opened 2020-05-28
Add in setting to vim docs to help recognize *.sc as Scala files

ckipp01:
This pr is to address what https://github.com/scalameta/metals/pull/1774 was created for. It adds in a small reference to a setting in the docs to help Vim recognize `*.sc` as Scala files.

### DIFF
--- a/docs/editors/vim.md
+++ b/docs/editors/vim.md
@@ -6,8 +6,9 @@ title: Vim
 ![Vim demo](https://i.imgur.com/4BYHCCL.gif)
 
 Metals works with most LSP clients for Vim, but we recommend using the
-[coc-metals extension](https://github.com/scalameta/coc-metals) for [`coc.nvim`](https://github.com/neoclide/coc.nvim)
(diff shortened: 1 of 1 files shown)


docs/editors/vim.md
```diff
@@ -99,6 +97,9 @@ set shortmess+=c
 " Always show signcolumns
 set signcolumn=yes
 
+" Help Vim recognize *.sbt and *.sc as Scala files
+au BufRead,BufNewFile *.sbt,*.sc set filetype=scala
```
ckipp01, line 101: Everything above is just formatting changes (just text width), and I moved this section from above to here and added the `*.sc`.

kpbochenek: checking docs: https://scalameta.org/metals/docs/editors/vim.html
should you not update this?
`au BufRead,BufNewFile *.sbt set filetype=scala`

kpbochenek: ah nvm you moved it, sry :facepalm:

PR #1787, opened 2020-05-29
Allow for exhaustive pattern matching in case of sealed upper bound types

### LINKED ISSUE #1066: Exhaustive match not working with upper type bounds
**Describe the bug**
Give the following code, if you trigger a completion at the marked position, you'll see that you can complete a basic match, but there is no option for an "exhaustive match" like you'd expect since the `Test` is a sealed trait.

```
package example

object Example extends App {
  sealed trait Test
  final case object Foo extends Test
  final case object Bar extends Test

  def testExhaustive[T <: Test](test: T): Boolean = 
    test m@@
    
}
```

You can see this in the below image
Screenshot 2019-11-22 at 11 24 50

**To Reproduce**
1. Copy the code snippet above
2. Trigger a completion at the `@@`
3. See that no exhaustive match completion is available

**Expected behavior**
There should be an exhaustive match open that will auto-complete both the `Foo` and `Bar` cases

**Installation:**
- Operating system: macOS
- Editor: Neovim
- Metals version: Latest Snapshot

EDIT: I had a poor example and image. I've updated it now to be more clear.

**Search terms**
completion, exhaustive match, upper type bounds, metals is awesome

ltrojanowski:
Allow for exhaustive pattern matching in case of sealed upper bound types

Changes:
* added a test to check exhaustive match with upper type bounds
* update match case completion
* handle refinement type

fixes #1066

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
@@ -150,7 +150,7 @@ trait MatchCaseCompletions { this: MetalsGlobal =>
       text: String
   ) extends CompletionPosition {
     override def contribute: List[Member] = {
-      val tpe = prefix.widen
+      val tpe = prefix.widen.bounds.hi
       val members = ListBuffer.empty[TextEditMember]
       val importPos = autoImportPosition(pos, text)
       val context = doLocateImportContext(pos, importPos)
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
@@ -149,7 +149,7 @@ trait MatchCaseCompletions { this: MetalsGlobal =>
       pos: Position,
       text: String
   ) extends CompletionPosition {
-    private def getSubclassesResult(tpe: Type): List[Symbol] = {
+    private def subclassesForType(tpe: Type): List[Symbol] = {
       val subclasses = ListBuffer.empty[Symbol]
       if (tpe.typeSymbol.isRefinementClass) {
         val RefinedType(parents, _) = tpe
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -149,15 +149,24 @@ trait MatchCaseCompletions { this: MetalsGlobal =>
       pos: Position,
       text: String
   ) extends CompletionPosition {
+    private def getSubclassesResult(tpe: Type): List[Symbol] = {
```
tgodzik, line 152:
```suggestion
    private def subclassesResult(tpe: Type): List[Symbol] = {
```
will be enough.

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -242,4 +242,42 @@ class CompletionMatchSuite extends BaseCompletionSuite {
...
+      |
+      |sealed trait TestA
+      |case object Foo extends TestA
+      |case object Bar extends TestA
+      |sealed trait TestB
+      |case object Baz extends TestB
+      |case object Goo extends TestB
+      |
+      |object Main {
+      |  def testExhaustive[T <: TestA with TestB](test: T): Boolean =
```
tgodzik, line 259: Could we maybe add a test also with `def testExhaustive[T <: TestA](test: T): Boolean =`

To be sure it works without refinement too?

mtags/src/main/scala-2/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
ltrojanowski: this would cause a naming clash. There is val subclassesResult in the overridden contribute function

tgodzik: `subclassesForType` maybe then ?

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
ltrojanowski: Added

mtags/src/main/scala-2/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
ltrojanowski: updated

Issue #1705, opened 2020-05-06
Scala 3 debugging issues


tgodzik:
**Describe the bug**

There is a number of issues that came up when testing debugging in Scala 3 codebase:
- breakpoint in outer methods don't work
- when debugging the dotty example project several steps on `runExamples("Named Type Arguments")(NamedTypeArguments.test)` did nothing. It seems to got stuck for a bit on class's init method
- debugging on Windows (not sure) seems to make each step go into the class loader

**To Reproduce**
1. Go to https://github.com/lampepfl/dotty-example-project
2.Try to reproduce on of the issues.

**Expected behavior**
Debugging working as expected in Scala 3

**Installation:**
- Operating system: macOS/Windows/Linux
- Editor: Visual Studio Code/Atom/Vim/Sublime/Emacs
- Metals version: v0.9.0

**Additional context**
This is part of the Scala 3 support work

**Search terms**
scala 3 debug

tgodzik:
Found another one:
```
  assertBreakpoints("lambda")(
    source = """|/a/src/main/scala/a/Main.scala
                |package a
                |object Main {
                |  def main(args: Array[String]): Unit = {
                |    List(1).foreach{ e =>
                |>>    println(e)
                |    }
                |    System.exit(0);
                |  }
                |}
                |""".stripMargin
  )
```
We will stop twice, first for the main class and then for lambda. This might need to be fixed in Bloop, but not sure yet.

tgodzik:
I changed the title, since the only thing that seems problematic is debugging on windows, which does not seem to work currently at all.

PR #1808, opened 2020-06-05
Add support for breakpoints in toplevel methods in Dotty

tgodzik:
Addresses one of the points in https://github.com/scalameta/metals/issues/1705

Previously, we did't stop at breakpoints inside toplevel methods when debugging. Now, we use a proper name for a class name generated from an outer method

### DIFF
--- a/mtags/src/main/scala-3/scala/scala/meta/internal/metals/ClassFinder.scala
+++ b/mtags/src/main/scala-3/scala/scala/meta/internal/metals/ClassFinder.scala
@@ -4,7 +4,9 @@ import org.eclipse.lsp4j.Position
 import dotty.tools.dotc.ast.untpd.Tree
 import dotty.tools.dotc.ast.untpd.PackageDef
 import dotty.tools.dotc.ast.untpd.TypeDef
+import dotty.tools.dotc.ast.untpd.DefDef
 import dotty.tools.dotc.ast.untpd.ModuleDef
+import dotty.tools.dotc.ast.untpd.Template
 import dotty.tools.dotc.ast.untpd.UntypedTreeTraverser
 import dotty.tools.dotc.ast.untpd
 import dotty.tools.dotc.util.SourcePosition
@@ -14,9 +16,11 @@ import dotty.tools.dotc.core.Flags
 
 object ClassFinder {
 
-  class SymbolTraverser(offset: Int) extends UntypedTreeTraverser {
+  class SymbolTraverser(offset: Int, fileName: String)
+      extends UntypedTreeTraverser {
     var symbol = ""
     var isInnerClass: Boolean = false
+    var isOuter: Boolean = true
     override def traverse(tree: Tree)(implicit ctx: Context): Unit = {
       if (tree.sourcePos.exists && tree.sourcePos.start <= offset && offset <= tree.sourcePos.end) {
         val delimeter =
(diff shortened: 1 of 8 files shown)


mtags/src/main/scala-3/scala/scala/meta/internal/metals/ClassFinder.scala
```diff
@@ -14,9 +16,11 @@ import dotty.tools.dotc.core.Flags
 
 object ClassFinder {
 
-  class SymbolTraverser(offset: Int) extends UntypedTreeTraverser {
+  class SymbolTraverser(offset: Int, fileName: String)
+      extends UntypedTreeTraverser {
     var symbol = ""
     var isInnerClass: Boolean = false
+    var isOuter: Boolean = true
```
marek1840, line 23: What does it mean to "be outer"?

tests/slow/src/test/scala/tests/feature/CrossDebugSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+         |def helloWorld(): Unit = {
+         |  object Even {
+         |>>  def unapply(s: String): Boolean = s.size % 2 == 0
+         |  }
+         |
+         |  "even" match {
+         |    case s @ Even() => println(s"$s has an even number of characters")
+         |    case s          => println(s"$s has an odd number of characters")
+         |  }
+         |  System.exit(0)
```
marek1840, line 45: Do we need such a complex case? 
Could we instead do 
```scala
"foo" match {
  case Even() => ()
  case _ => ()
}
```
?

The action taken based on whether even matches in of no consequence to the test itself, therefore we could simplify the logic reducing the cognitive load on the reader. It builds up quickly with more tests to "parse"

tests/slow/src/test/scala/tests/feature/CrossDebugSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+  assertBreakpoints(
+    "outer-object",
+    main = Some("a.helloWorld")
+  )(
+    source = """|/a/src/main/scala/a/Main.scala
+                |package a
+                |
+                |@main 
+                |def helloWorld(): Unit = {
+                |  object Hello{
```
marek1840, line 60: this is called `outer` object in dotty instead of `inner` ? It is inner to the method in the end... :thinking: 
Is it accessible from outside the `helloworld`?

mtags/src/main/scala-3/scala/scala/meta/internal/metals/ClassFinder.scala
tgodzik: Might be better to have `toplevel` - I will change that.

tgodzik: Fixed!

tests/slow/src/test/scala/tests/feature/CrossDebugSuite.scala
tgodzik: Done

tests/slow/src/test/scala/tests/feature/CrossDebugSuite.scala
tgodzik: Changed to `object-in-toplevel-method`

tgodzik:
The failure is unrelated. might need to take a look at it sometime in the future is the failures are too frequent.

PR #1811, opened 2020-06-08
Pick the most valued build target instead of least for running or debugging

tgodzik:
I missed that when testing it in https://github.com/scalameta/metals/pull/1781 - sortBy will give us the lowest numbers first. 

Did some manual tests, which should be enough here, since unit testing this might prove very difficult.


tgodzik:
The failure is unrelated, might be another flakiness of the ammonite suite.

PR #1812, opened 2020-06-09
Add lsp-metals as required dependency in Emacs docs

ag91:
I had an issue using the metals backend this morning, and I solved by requiring the metals backend for lsp-mode. I think it would be useful to have it in the docs.


kpbochenek:
Thanks for contribution!

Good catch, metals support was moved to separate package https://github.com/emacs-lsp/lsp-metals/

PR #1818, opened 2020-06-12
Fix issues with semanticdb on windows

tgodzik:
Previously, we would not find semanticdb file if the relative path was done with `\\`. Now, we make sure that the slash is used all the time.

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/mtags/Semanticdbs.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/mtags/Semanticdbs.scala
@@ -61,7 +61,7 @@ object Semanticdbs {
   ): TextDocumentLookup = {
     val reluri = scalaRelativePath.toURI(false).toString
     val sdocs = loadTextDocuments(semanticdbPath)
-    sdocs.documents.find(_.uri == reluri) match {
+    sdocs.documents.find(_.uri.replace("\\", "/") == reluri) match {
       case None => TextDocumentLookup.NoMatchingUri(scalaPath, sdocs)
       case Some(sdoc) =>
         val text = FileIO.slurp(scalaPath, charset)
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala-2/scala/meta/internal/mtags/Semanticdbs.scala
```diff
@@ -61,7 +61,7 @@ object Semanticdbs {
   ): TextDocumentLookup = {
     val reluri = scalaRelativePath.toURI(false).toString
     val sdocs = loadTextDocuments(semanticdbPath)
-    sdocs.documents.find(_.uri == reluri) match {
+    sdocs.documents.find(_.uri.replace("\\", "/") == reluri) match {
```
tgodzik, line 64: Normally path was done `/src/main/scala/Main.scala`, but when Dotty is generating the file it's `src\main\scala\Main.scala`

ckipp01: So this change this would only matter if we are looking for semantic information that dotty produced while the user is still in a scala-2 workspace right?

olafurpg: This sounds like a bug in Dotty, the `uri` field should always return a URI. I think it's OK to do this replace on the Metals side just-in-case but we should make sure this is also fixed upstream in Dotty.

tgodzik: Sure, we plan to follow up with a change to Dotty itself too.

tgodzik: @ckipp01 This only is an issue if someone compiles Dotty code on Windows, it doesn't influence Scala-2 codebases at all.. We will not use the semanticdb information due to the relative uri mismatch

PR #1819, opened 2020-06-12
Update ammonite to fix issues with file watching

tgodzik:
Previously, ammonite would remove all directories inside the target, which would cause the file watching to break. Now it only removes relevant files. This might also help with the ammonite suite flakiness.

### DIFF
--- a/tests/unit/src/main/scala/tests/BaseAmmoniteSuite.scala
+++ b/tests/unit/src/main/scala/tests/BaseAmmoniteSuite.scala
@@ -144,7 +144,6 @@ abstract class BaseAmmoniteSuite(scalaVersion: String)
         }
         promise
       }
-      _ <- server.didSave("main.sc")(identity)
       _ <- server.didSave("main.sc") { _ =>
         s""" // scala $scalaVersion
            |import $$ivy.`com.github.alexarchambault::case-app:2.0.0-M16`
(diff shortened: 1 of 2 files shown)


tests/unit/src/main/scala/tests/BaseAmmoniteSuite.scala
```diff
@@ -139,6 +139,7 @@ abstract class BaseAmmoniteSuite(scalaVersion: String)
       refreshedPromise = {
         val promise = Promise[Unit]()
         server.client.refreshBuildHandler = { () =>
+          println("hello")
```
ckipp01, line 142: I'm assuming this is leftover?

tgodzik: yes :O I did it again!

tests/unit/src/main/scala/tests/BaseAmmoniteSuite.scala
```diff
@@ -469,8 +470,21 @@ abstract class BaseAmmoniteSuite(scalaVersion: String)
         )
       )
       .asScala
-      .map { res =>
+      .flatMap { res =>
         val code = server.textContents(path)
-        TestHovers.renderAsString(code, Option(res), includeRange = true)
+        val result =
```
tgodzik, line 475: I added a retry, because it can't figure out why is this happening. We've been having issues with hover anyway, so it can be just that, but more prominent in this situation. I also added additional logging that can help investigate.

tgodzik:
We're still having issues with the test, I will investigate further and maybe ignore if I am unable to come up with something soon.

tgodzik:
Ok, it seems that the ammonite tests haven't failed for the last 3 runs, so I am going to merge it. I will hold off on closing the issue though.

PR #1832, opened 2020-06-15
Add in spacing in vscode icons only.

ckipp01:
Since VS Code is the only one using an actual icon for spacing it's
better to put the spacing there since we are doing a replace on the
icon and times producing labels that have one empty space and other
times two empty spaces causing odd spacing in other clients. This
change will allow for more consistent spacing in those clients while
keeping the same spacing in VS Code.

For example what this looks like in coc-metals without the change :(

Screenshot 2020-06-15 at 16 25 11

And after the change :)

Screenshot 2020-06-15 at 16 29 07

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Icons.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Icons.scala
@@ -65,7 +65,7 @@ object Icons {
     override def check: String = "$(check) "
     override def findsuper: String = "$(arrow-up)"
     override def folder: String = "$(folder)"
-    override def github: String = "$(github)"
+    override def github: String = "$(github) "
   }
   case object atom extends Icons {
     private def span(id: String) = s"<span class='icon icon-$id'></span> "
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/Icons.scala
```diff
@@ -65,7 +65,7 @@ object Icons {
     override def check: String = "$(check) "
     override def findsuper: String = "$(arrow-up)"
     override def folder: String = "$(folder)"
```
tgodzik, line 67: Should we also do it for each icon used in the NewProjectProvider? Like for `folder` ?

ckipp01: When testing it locally the only one I noticed that seemed not aligned was the github icon. I just checked and changing the folder one doesn't seem to affect anything because it seems that we aren't adding an extra space for it when it's used. As far as I can tell, only the GitHub one needs to change here.

ckipp01:
Unrelated test failure
```
 [error] ==> X tests.RenameLspSuite.unary-bad2  2.224s java.lang.IllegalArgumentException: requirement failed: Build server did not initialize
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[info] tests.RenameLspSuite.java-classes started
	at java.lang.Thread.run(Thread.java:834)
```

PR #1833, opened 2020-06-16
Remove fastpass command line tool

olafurpg:
Previously, the Pants to Bloop export tool lived in the Metals codebase.
This project has now been moved into a separate repository
https://github.com/scalameta/fastpass


.github/workflows/ci.yml
```diff
@@ -87,10 +86,6 @@ jobs:
             command: bin/test.sh 'slow/testOnly -- tests.gradle.*'
             name: Gradle integration
             os: ubuntu-latest
-          - type: pants
```
tgodzik, line 90: Don't we want to keep the tests, but just add a dependency on fastpass?

tgodzik:
> [error] D:\a\metals\metals\tests\unit\src\test\scala\tests\NewProjectLspSuite.scala:18:12: object apache is not a member of package org
[error] import org.apache.commons.io.FileUtils
[error]            ^
[error] D:\a\metals\metals\tests\unit\src\test\scala\tests\NewProjectLspSuite.scala:237:9: not found: value FileUtils
[error]         FileUtils.deleteDirectory(tmpDirectory.toNIO.toFile())
[error]         ^

My bad, apache file utils must have been brought as a transitive dependency and I used it by mistake. We can change the usual delete to:
```

    def delete(): Unit = {
      if (path.isFile) {
        Files.delete(path.dealias.toNIO)
      } else {
        path.list.toList.foreach(_.delete())
      }
    }
```
and use that

website/blog/2020-01-10-cobalt.md
```diff
@@ -241,13 +239,6 @@ the user the option of automatic reconnection. This works both when the build
 server was down and when the connection was just interrupted. It should also
 solve the issues on MacOs with dropped connection.
 
-### Experimental build tool support for Pants
-
-Thanks to the recent work by [@olafurpg](https://github.com/olafurpg), there's
-now experimental support for the Pants build tool. The Pants integration is
-still evolving so it's not yet documented. The end goal is to support Pants in
-the same way as other build tools like sbt, Maven and Gradle.
-
```
ckipp01, line 250: Since these blogs somewhat serve as release notes and _sort of_ a changelog, do we really want to change this blog post to remove reference at all to Pants? This seems to imply that we are sort of _hiding_ away that we had experimental pants support at one time. My question is mainly just trying to understand the reason behind this.

.github/workflows/ci.yml
olafurpg: The current Pants integration for "import build" doesn't work reliably. It's better for users to run the fastpass refresh/create commands from the command-line instead.

website/blog/2020-01-10-cobalt.md
olafurpg: My reasoning to remove this part is because it appears when users search for "Pants" on the website. Since the Pants integration is now gone I figured it would be best if this section didn't show up when people search for it.

olafurpg:
@tgodzik I replaced `FileUtils.deleteDirectory` with `RecursivelyDelete` that is already part of the Metals codebase.

PR #1834, opened 2020-06-16
Changes for decorationProvider if `isHttpEnabled` is true.

ckipp01:
This has multiple changes to account for the case where a client may
have `isDecorationProvider` and also `isHttpEnabled` both set to true.

- Checks to make sure that `isHttpEnabled` wins before sending doctor
- Makes sure the publishDecorations is forwarded

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
@@ -72,7 +72,7 @@ final class Doctor(
       clientCommand: Command,
       onServer: MetalsHttpServer => Unit
   ): Unit = {
-    if (clientConfig.isExecuteClientCommandProvider) {
+    if (clientConfig.isExecuteClientCommandProvider && !clientConfig.isHttpEnabled) {
       val output =
         if (clientConfig.doctorFormatIsJson)
           buildTargetsJson()
@@ -88,7 +88,7 @@ final class Doctor(
           onServer(server)
         case None =>
           scribe.warn(
-            "Unable to run doctor. To fix this problem enable -Dmetals.http=true"
+            "Unable to run doctor. Make sure `isHttpEnabled` is set to `true`."
           )
       }
     }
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -88,7 +88,7 @@ final class Doctor(
           onServer(server)
         case None =>
           scribe.warn(
-            "Unable to run doctor. To fix this problem enable -Dmetals.http=true"
+            "Unable to run doctor. Server has not yet started."
```
ckipp01, line 91: If we change the above, then this message also becomes a bit more telling. Since it's no longer actually about havint `http=true`, but rather than the server isn't up yet. For example, this sometimes leads to confusion if a client has `httpEnabled`, when the first `metals-doctor-reload` happens right at startup, this falls in `None` showing the warning, even though the setting is true. This change will also avoid that and make the message clearer.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -72,7 +72,7 @@ final class Doctor(
       clientCommand: Command,
       onServer: MetalsHttpServer => Unit
   ): Unit = {
-    if (clientConfig.isExecuteClientCommandProvider) {
+    if (clientConfig.isExecuteClientCommandProvider && !clientConfig.isHttpEnabled) {
```
ckipp01, line 75: In this scenario I think that `isHttpEnabled` should win. Mainly due to the fact that while all clients _should_ support all the commands that the server can send, they may not. For example in nvim-metals, we are making the client as `isExecuteClientCommandProvider` to true because we need some stuff (for the new scala file command), but we still want the doctor to be in the browser. Without this change, doing that is impossible.

metals/src/main/scala/scala/meta/internal/metals/MetalsHttpClient.scala
```diff
@@ -184,7 +184,11 @@ final class MetalsHttpClient(
 
   override def metalsPublishDecorations(
       params: PublishDecorationsParams
-  ): Unit = ()
+  ): Unit = {
```
ckipp01, line 187: This was causing worksheet decorations not to work for clients that had `isHttpEnabled` set to true.

PR #1839, opened 2020-06-18
Update dotty to 0.25.0-RC-2 and 0.24.0

tgodzik:
Almost nothing got changed, mostly moved files around due to one source imcompatibilty.


tgodzik:
The failure is unrelated, I will go ahead and merge since this is just a version bump.

PR #1840, opened 2020-06-18
Bump up the scalafmt version for the codebase

ckipp01:


metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
```diff
@@ -25,8 +25,8 @@ class ShellRunner(
     userConfig: () => UserConfiguration,
     time: Time,
     statusBar: StatusBar
-)(
-    implicit executionContext: scala.concurrent.ExecutionContext
+)(implicit
```
ckipp01, line 28: Personally I'm a bit torn by this. I don't like having custom settings for scalafmt. The closer to vanilla the better. Howerver, I also dislike the way the implicit is for this. 

```
newlines.implicitParamListModifierPrefer   = before
```
I prefer that setting above, but then it adds another setting. Any preference on this? Again, I don't have a strong feeling, because my desire to not add another setting is the same as my dislike for this formatting.

metals-docs/src/main/scala/docs/Snapshot.scala
```diff
@@ -42,7 +42,8 @@ object Snapshot {
   private def current: Snapshot =
     Snapshot(BuildInfo.metalsVersion, LocalDateTime.now())
 
-  /** Returns the latest published snapshot release, or the current release if. */
+  /**
+   * Returns the latest published snapshot release, or the current release if. */
```
ckipp01, line 46: I also don't love these, but since they aren't `//` they make sense.

tgodzik: I think it's fine, but if you prefer otherwise, I don't really have a preference.

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
tgodzik: Let's leave the default in my opinion. I think it's better what was before.

ckipp01: 👍 sounds good.

metals-docs/src/main/scala/docs/Snapshot.scala
ckipp01: I'd say if you don't have a preference, then let's just leave them as is 👍

ckipp01:
> Looks good! Thanks for following up! If you want to change some settings, it's ok with me, but I am fine with the default formatting.

I have some legit failures, so I'm looking into them.

tgodzik:
> > Looks good! Thanks for following up! If you want to change some settings, it's ok with me, but I am fine with the default formatting.
> 
> I have some legit failures, so I'm looking into them.

Test cases might have also got formatted. You might need to run `save-expect` or `saveExpect` ?

ckipp01:
> `saveExpect`

TIL about `save-expect`

tgodzik:
Weirdly enough it also looks like sometimes failing to load up sbt:
```
2020.06.18 19:19:45 ERROR Getting org.scala-sbt sbt 1.2.6  (this may take some time)...
[error] ==> X tests.sbt.SbtLspSuite.no-sbt-version  600.072s java.util.concurrent.TimeoutException: Futures timed out after [10 minutes]
```

metals/src/main/scala/scala/meta/internal/builds/NewProjectProvider.scala
```diff
@@ -45,40 +45,43 @@ class NewProjectProvider(
...
-            .collect {
-              case matching if matching.groupCount == 2 =>
-                MetalsQuickPickItem(
-                  id = matching.group(1),
-                  label = s"${icons.github}" + matching.group(1),
-                  description = matching.group(2)
-                )
-            }
+  def allTemplatesFromWeb: Seq[MetalsQuickPickItem] =
+    synchronized {
```
olafurpg, line 49: These large diffs shouldn’t be necessary. I don’t remember the setting to keep the old behavior but I’m tempted to open an upstream issue about enabling it by default. No action needed here, just thinking out loud

ckipp01:
> Weirdly enough it also looks like sometimes failing to load up sbt:
> 
> ```
> 2020.06.18 19:19:45 ERROR Getting org.scala-sbt sbt 1.2.6  (this may take some time)...
> [error] ==> X tests.sbt.SbtLspSuite.no-sbt-version  600.072s java.util.concurrent.TimeoutException: Futures timed out after [10 minutes]
> ```

Indeed, and I don't really get why. I can't mimic this locally, and I'm unsure what's happening there.

tgodzik:
> > Weirdly enough it also looks like sometimes failing to load up sbt:
> > ```
> > 2020.06.18 19:19:45 ERROR Getting org.scala-sbt sbt 1.2.6  (this may take some time)...
> > [error] ==> X tests.sbt.SbtLspSuite.no-sbt-version  600.072s java.util.concurrent.TimeoutException: Futures timed out after [10 minutes]
> > ```
> 
> Indeed, and I don't really get why. I can't reproduce this locally, and I'm unsure what's happening there.
> 
> EDIT: I just got it locally has well.

Looks like this is happening in no build tool workspace :thinking:

tgodzik:
@ckipp01 I am trying a rerun to see if it's a flaky sbt download, it looks like it's and issue in tests that use non-default sbt version. We could change the version to be 1.3.12 everywhere, but it's still weird it's happening.

ckipp01:
> @ckipp01 I am trying a rerun to see if it's a flaky sbt download, it looks like it's and issue in tests that use non-default sbt version. We could change the version to be 1.3.12 everywhere, but it's still weird it's happening.

Agreed. I did re-run them, and I got the same result. I was going to wait for https://github.com/scalameta/metals/pull/1842 to get merged in, then rebase and try again. I thought on an off change there may have been something weird affecting it.

As for the 1.3.12 everywhere, sure we can do that, but that really shouldn't be causing this to fail though right?

tgodzik:
> As for the 1.3.12 everywhere, sure we can do that, but that really shouldn't be causing this to fail though right?

True, but we still should use newest sbt as the default. I will make the change anyway. I do not understand what is happening here at all though.

ckipp01:
Sort of an aside, there is also a couple `java.lang.StringIndexOutOfBoundsException`s when we run this new version of scalafmt locally on the codebase. I'll try to minimize an example and submit an issue to scalafmt about it.

tgodzik:
> Sort of an aside, there is also a couple `java.lang.StringIndexOutOfBoundsException`s when we run this new version of scalafmt locally on the codebase. I'll try to minimize an example and submit an issue to scalafmt about it.
> 
> ~It also tries to `metals/mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala`, which it shouldn't, so I may have missed a change in `project.excludeFilters`.~
> Ugh, I just changed this to no longer have `/scala/scala`, which is listed in the excludes, so of course it wouldn't target it correctly.

We should change to only run tests after successful scalafmt and scalafix, which should limit the reruns.

The tests seems to be passing now weirdly enough. Might have been a download issue in case of previous versions. I created https://github.com/scalameta/metals/pull/1843 so that we always use the same version.

ckipp01:
All Green! \o/ 🎉

PR #1842, opened 2020-06-19
Fix adding unmanaged source directories for mtags3

tgodzik:


build.sbt
```diff
@@ -259,20 +257,9 @@ def multiScalaDirectories(root: File, scalaVersion: String) = {
...
-  }
-  if (isScala213(partialVersion)) {
-    result += base / "scala-2.13"
-  }
-  if (isScala023(partialVersion)) {
-    result += base / "scala-0.23"
-  }
-  if (isScala024(partialVersion)) {
-    result += base / "scala-0.24"
+  partialVersion.collect {
```
tgodzik, line 260: This should simplify things and make it more future proof.

build.sbt
```diff
@@ -338,6 +325,7 @@ val mtagsSettings = List(
 lazy val mtags3 = project
   .in(file(".mtags"))
   .settings(
+    unmanagedSourceDirectories.in(Compile) := Seq(),
```
tgodzik, line 328: In order not to add `.mtags` directories, although it seems that sbt now created the `.mtags` directory anyway.

build.sbt
ckipp01: Ah, I like this. Good idea 👍

PR #1843, opened 2020-06-19
In case of no defined sbt version run bloopInstall with the newest one

tgodzik:
Previously, we would run sbt version defined in the launcher (1.2.6 previously), now we run the default version used by Metals.

I also updated the launcher to the one from the newest version.


tgodzik:
This needs to wait for https://github.com/scalameta/metals/pull/1840 - we finally got a green build on the scalafmt update!

PR #1844, opened 2020-06-19
Make other tests depend on formatting and scalafix

tgodzik:
So that tests are not run needlessly in case of formatting failure.


tgodzik:
> I personally think it's nice when reviewing a change to know whether the tests pass even if the formatting/scalafix checks pass. Those can be automatically fixed so I'm not so worried about those failures when reviewing but it's really helpful to know if the tests pass or not. I think it can result in a bad contributor experience if the CI fails fast with formatting/scalafix and we need to do an extra roundtrip of "can you please fix formatting before I review" comments.

Actually, it might be bad contributor experience, you are right. I was thinking more from my perspective, but I can just cancel the previous run anyway.

Issue #1770, opened 2020-05-21
Invalid $file completions for Ammonite


tgodzik:
**Describe the bug**
Ammonite generates Scala files inside `.ammonite` directory and the package structure corresponds to the directory structure. Because of that the presentation compiler sends completions like:
`$file.directory.script` which will not neccessarily be right as the `$file` completions work relatively.

**To Reproduce**
1. Create a script `somedirectory/Main.sc`
2. Import the script
3. Try to complete `$file.`
4. There will be a completion for `somedirectory`, which is invalid.

**Expected behavior**
Currently no completions for $file should work. Alternatively, we can special case $file in `*.sc` file and provide filesystem based completions.

**Screenshots**

**Installation:**
- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: 0.9.0+130-f0796498-SNAPSHOT

**Additional context**

![image](https://user-images.githubusercontent.com/3807253/82581822-6f69f680-9b91-11ea-994a-0380201f26f3.png)

**Search terms**
ammonite completions

PR #1846, opened 2020-06-19
Added completions for $file in ammonite scripts

tgodzik:
Also in this PR:
- fixes https://github.com/scalameta/metals/issues/1770
- adds ammonite script to new file provider

Completions:

ammonite

cc @alexarchambault

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1687,33 +1687,35 @@ class MetalsLanguageServer(
 
     try {
       val sourceToIndex0 = sourceToIndex(source, targetOpt)
-      val reluri = source.toIdeallyRelativeURI(sourceItem)
-      val input = sourceToIndex0.toInput
-      val symbols = ArrayBuffer.empty[WorkspaceSymbolInformation]
-      SemanticdbDefinition.foreach(input) {
-        case SemanticdbDefinition(info, occ, owner) =>
-          if (WorkspaceSymbolProvider.isRelevantKind(info.kind)) {
-            occ.range.foreach { range =>
-              symbols += WorkspaceSymbolInformation(
-                info.symbol,
-                info.kind,
-                range.toLSP
-              )
+      if (sourceToIndex0.exists) {
+        val reluri = source.toIdeallyRelativeURI(sourceItem)
+        val input = sourceToIndex0.toInput
+        val symbols = ArrayBuffer.empty[WorkspaceSymbolInformation]
+        SemanticdbDefinition.foreach(input) {
+          case SemanticdbDefinition(info, occ, owner) =>
+            if (WorkspaceSymbolProvider.isRelevantKind(info.kind)) {
+              occ.range.foreach { range =>
+                symbols += WorkspaceSymbolInformation(
+                  info.symbol,
+                  info.kind,
+                  range.toLSP
+                )
+              }
             }
-          }
-          if (
-            sourceItem.isDefined &&
-            !info.symbol.isPackage &&
-            (owner.isPackage || source.isAmmoniteScript)
-          ) {
-            definitionIndex.addToplevelSymbol(reluri, source, info.symbol)
-          }
-      }
-      workspaceSymbols.didChange(source, symbols)
+            if (
+              sourceItem.isDefined &&
+              !info.symbol.isPackage &&
+              (owner.isPackage || source.isAmmoniteScript)
+            ) {
+              definitionIndex.addToplevelSymbol(reluri, source, info.symbol)
+            }
+        }
+        workspaceSymbols.didChange(source, symbols)
 
-      // Since the `symbols` here are toplevel symbols,
-      // we cannot use `symbols` for expiring the cache for all symbols in the source.
-      symbolDocs.expireSymbolDefinition(sourceToIndex0)
+        // Since the `symbols` here are toplevel symbols,
+        // we cannot use `symbols` for expiring the cache for all symbols in the source.
+        symbolDocs.expireSymbolDefinition(sourceToIndex0)
+      }
     } catch {
       case NonFatal(e) =>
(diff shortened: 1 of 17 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -274,11 +276,22 @@ class CompletionProvider(
       def isNotLocalForwardReference: Boolean =
         !head.sym.isLocalToBlock ||
           !head.sym.pos.isAfter(pos)
+      def isFileAmmoniteCompletion() = {
```
tgodzik, line 279: This removes any default `$file` completions coming from how the packages in the generated files are organized.

ckipp01:
Just something I noticed right away when trying it out with [this repo](https://github.com/ckipp01/chronica) is that I'm getting a completion option which leads to the corresponding `.scala` file in the `.ammonite` directory. 

Screenshot 2020-06-19 at 21 40 17

```
[Trace - 09:43:34 PM] Sending response 'textDocument/completion - (39)'. Processing request took 17ms
Result: {
  "isIncomplete": false,
  "items": [
    {
      "label": "website ammonite.$file",
      "kind": 9,
      "detail": " ammonite.$file",
      "preselect": true,
      "sortText": "00000",
      "filterText": "website",
      "insertTextFormat": 2,
      "textEdit": {
        "range": {
          "start": {
            "line": 4,
            "character": 13
          },
          "end": {
            "line": 4,
            "character": 13
          }
        },
        "newText": "website"
      },
      "data": {
        "symbol": "ammonite/$file/website.",
        "target": "file:///Users/ckipp/Documents/scala-workspace/chronica/website.sc"
      }
    },
    {
      "label": "scripts",
      "kind": 9,
      "detail": " ammonite.$file",
      "sortText": "00001",
      "filterText": "scripts",
      "insertTextFormat": 2,
      "textEdit": {
        "range": {
          "start": {
            "line": 4,
            "character": 13
          },
          "end": {
            "line": 4,
            "character": 13
          }
        },
        "newText": "scripts"
      },
      "data": {
        "symbol": "ammonite/$file/scripts/",
        "target": "file:///Users/ckipp/Documents/scala-workspace/chronica/website.sc"
      }
    }
  ]
}
```
That top completion item, if I complete it, and then do a go to definition on it, I end up in `.ammonite/scala-2.12.11/amm-2.1.4-6-2179b35/website/src/ammonite/$file/website.sc.scala`.

tgodzik:
> That top completion item, if I complete it, and then do a go to definition on it, I end up in `.ammonite/scala-2.12.11/amm-2.1.4-6-2179b35/website/src/ammonite/$file/website.sc.scala`.

Hmm... that should be what happens before the change. Let me check out your repo.

ckipp01:
> Hmm... that should be what happens before the change. Let me check out your repo.

Ack! I'm not using my local snapshot.....

mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
ckipp01, line 279: Super nitpicky, but wrapping the expression in brackets isn't necessary here. I only bring it up because we don't do it for the method above. If you prefer it, then no worries.

mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -323,7 +336,12 @@ class CompletionProvider(
     else if (symbol.isClass) k.Class
     else if (symbol.isMethod) k.Method
     else if (symbol.isCaseAccessor) k.Field
-    else if (symbol.isVal && !symbolIsFunction) k.Value
+    else if (symbol.isVal && !symbolIsFunction)
+      r match {
```
ckipp01, line 340: I know this was here before, but since you're using it, may I propose instead of using `r` we just use `member` as the argument name for `completionItemKind`?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1687,33 +1687,35 @@ class MetalsLanguageServer(
...
-      SemanticdbDefinition.foreach(input) {
-        case SemanticdbDefinition(info, occ, owner) =>
-          if (WorkspaceSymbolProvider.isRelevantKind(info.kind)) {
-            occ.range.foreach { range =>
-              symbols += WorkspaceSymbolInformation(
-                info.symbol,
-                info.kind,
-                range.toLSP
-              )
+      if (sourceToIndex0.exists) {
```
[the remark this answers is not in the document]
tgodzik: This is actually only an issue if ammonite failed to generate the scala files, which also means we should watch the `.ammonite` directory, since they can be created after indexing :thinking:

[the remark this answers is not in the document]
tgodzik: Actually, it didn't fail to generate, ammonite just differently generates path if `^` is used, which should not normally be an issue, but we should leave the guard just in case: I will change the tests a bit so that it's more natural path and not `^.backToTheSameDirectory`

mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
tgodzik: Fixed!

mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
tgodzik: Fixed!

tgodzik:
@alexarchambault do you think this change is fine to merge?

alexarchambault:
I didn't give that a try a yet (I can a bit later today). This mostly touches the completion logic of metals, that you're much more knowledgeable about than me, so it's fine by me.

PR #1851, opened 2020-06-24
Add quiet log option

tgodzik:
Fixes https://github.com/scalameta/metals-vscode/issues/137

Whenever we run a slow taks, the metals logs might be shown for user and this can become a bit annoying.

This change makes sure that we do not show logs when:
- user explicitly doesn't want them
- there are no logs to show

I've set the logs to quiet in places that I don't really see any benefit to showing them (99% they finish super quickly and for example in VS Code it will hide the whole side panel). Especially that even in case of an error they are hidden afterwards. Alternatively we could use the UserConfiguration option everywhere. Opinions?

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
@@ -271,6 +271,7 @@ final class FormattingProvider(
       downloadingScalafmt = Promise()
       statusBar.trackSlowFuture(
         "Loading Scalafmt",
+        quietLogs = true,
         downloadingScalafmt.future
       )
       System.out
--- a/metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
@@ -38,7 +38,8 @@ case class UserConfiguration(
     pantsTargets: Option[List[String]] = None,
     superMethodLensesEnabled: Boolean = false,
     remoteLanguageServer: Option[String] = None,
-    enableStripMarginOnTypeFormatting: Boolean = true
+    enableStripMarginOnTypeFormatting: Boolean = true,
+    quietLogs: Boolean = false
 ) {
 
   def currentBloopVersion: String =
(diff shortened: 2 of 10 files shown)


metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
```diff
@@ -271,6 +271,7 @@ final class FormattingProvider(
       downloadingScalafmt = Promise()
       statusBar.trackSlowFuture(
         "Loading Scalafmt",
+        quietLogs = true,
```
tgodzik, line 274: No interesting logs will be shown here. And even in case of an error, the log will be hidden afterwards. This operation is normally too quick to make sense to open the logs.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -146,6 +147,15 @@ object UserConfiguration {
            |
            |""".stripMargin
       ),
+      UserConfigurationOption(
```
olafurpg, line 150: I'm curious if we can avoid this user configuration option by refining how we handle "import build". I would be OK with disabling the output panel popup for "import build" and instead attach a command to the status bar item to reveal the output panel.

tgodzik: That's an awesome idea! We could also make sure that the option is available also if the build import failed. Something like "go to logs".

tgodzik:
I experimented a bit with the vs code extension to make the behaviour overall less annoying, but i haven't had any luck, so I reverted to the additional option. 

However, what @olafurpg is suggesting seems much better. 

I will try that 😁

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
ckipp01: Yea, I also agree, I think if we were able to avoid the configuration option by providing an action item to go to the logs, that would be great.

tgodzik:
Closing this in favour of https://github.com/scalameta/metals-vscode/pull/336

@ckipp01 I don't think this in particular is an issue in Vim, so most likely just the VS Code PR is enough.

ckipp01:
> Closing this in favour of [scalameta/metals-vscode#336](https://github.com/scalameta/metals-vscode/pull/336)
> 
> @ckipp01 I don't think this in particular is an issue in Vim, so most likely just the VS Code PR is enough.

👍 Correct, this is a non issue in Vim, so unless anyone would specifically ask about this, I'll just leave it as is.

PR #1853, opened 2020-06-25
Add jvm parameters configuration for Ammonite

tgodzik:
We run a separate ammonite bsp server per workspace. Previously, there was no way to limit the memory used by it. Now, users can specify configuration options that limit the memory or make other changes to the JVM used by ammonite. Alternatively, .jvmopts will also be picked up as is the case with Metals server itself.

VS Code PR: https://github.com/scalameta/metals-vscode/pull/337

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -34,6 +34,7 @@ import scala.meta.internal.builds.ShellRunner
 import scala.meta.internal.implementation.ImplementationProvider
 import scala.meta.internal.implementation.Supermethods
 import scala.meta.internal.io.FileIO
+import scala.meta.internal.metals.Messages.AmmoniteJvmParametersChange
 import scala.meta.internal.metals.Messages.IncompatibleBloopVersion
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.ammonite.Ammonite
@@ -952,6 +953,24 @@ class MetalsLanguageServer(
                 case item if item == Messages.BloopVersionChange.reconnect =>
                   bloopServers.shutdownServer()
                   autoConnectToBuildServer().ignoreValue
+                case _ =>
+                  Future.successful(())
+              }
+          } else if (
+            userConfig.ammoniteJvmProperties != old.ammoniteJvmProperties && buildTargets.allBuildTargetIds
+              .exists(Ammonite.isAmmBuildTarget)
+          ) {
+            languageClient
+              .showMessageRequest(AmmoniteJvmParametersChange.params())
+              .asScala
+              .flatMap {
+                case item if item == AmmoniteJvmParametersChange.restart =>
+                  ammonite
+                    .stop()
+                    .asScala
+                    .flatMap { _ =>
+                      ammonite.start()
+                    }
                 case _ =>
                   Future.successful(())
               }
--- a/metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
@@ -34,6 +34,7 @@ case class UserConfiguration(
     worksheetCancelTimeout: Int = 4,
     bloopSbtAlreadyInstalled: Boolean = false,
     bloopVersion: Option[String] = None,
+    ammoniteJvmProperties: Option[List[String]] = None,
     superMethodLensesEnabled: Boolean = false,
     remoteLanguageServer: Option[String] = None,
     enableStripMarginOnTypeFormatting: Boolean = true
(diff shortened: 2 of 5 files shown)


metals/src/main/scala/scala/meta/internal/metals/ammonite/Ammonite.scala
```diff
@@ -246,13 +246,22 @@ final class Ammonite(
       case (command, script) =>
         val extraScripts = buffers.open.toVector
           .filter(path => path.isAmmoniteScript && path != script)
+        val jvmOpts = userConfig().ammoniteJvmProperties.getOrElse(
```
tgodzik, line 249: The priority is the ammonite options specified in the configuration, otherwise we use JVM_OPTS variable and `.jvmopts` file.

This is the same as in the case of Metals server. I wonder whether we should use jvmopts, maybe just the settings?

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -205,6 +215,28 @@ object UserConfiguration {
...
+            val parsed = elem.getAsJsonArray().asScala.flatMap { value =>
+              Try(value.getAsJsonPrimitive().getAsString()) match {
+                case Failure(exception) =>
+                  errors += s"json error: values in '$key' should have value of type string but obtained $value"
+                  None
+                case Success(value) =>
+                  Some(value)
+              }
+            }
+            Option(parsed.toList)
```
kpbochenek, line 233: you can do `Some(...)` (it is impossible to be null here.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -205,6 +215,28 @@ object UserConfiguration {
...
+          } else {
+            None
```
kpbochenek, line 235: It doesn't matter much but I would check if it actually has any value or not and if it has but is not JsonArray I would do `errors += ...`

It might be helpful to catch potential error when other clients will start sending this value but will serialize it e.g. as String not array.

 But this is super edge case, you can completely ignore this.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -107,6 +108,15 @@ object UserConfiguration {
           |separators (even on Windows).
           |""".stripMargin
       ),
+      UserConfigurationOption(
+        "ammonite-jvm-properties",
+        """`[]`.""",
+        "['-Xmx1G']",
```
kpbochenek, line 114: this is invalid json, it should be `["-Xmx1G"]`

metals/src/main/scala/scala/meta/internal/metals/ammonite/Ammonite.scala
kpbochenek: I would keep it as simple as possible and only use UseConfig.
You don't have to worry what if I have `.jvmopts` but don't want to use if for ammonite, is `.jvmopts` prioritized over JVM_OPTS etc.

Also I think JVM_OPTS are only used by CI and those systems, users normally specify JAVA_OPTS (e.g. `amm` script respects JAVA_OPTS not JVM_OPTS).

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -955,6 +956,25 @@ class MetalsLanguageServer(
...
+              .asScala
+              .flatMap {
+                case item if item == AmmoniteJvmParametersChange.restart =>
+                  ammonite
+                    .stop()
+                    .asScala
+                    .flatMap { _ =>
+                      ammonite.start()
+                    }
+                    .ignoreValue
```
kpbochenek, line 974: Not needed, `ammonite.start()` already returns `Future[Unit]`

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -107,6 +108,15 @@ object UserConfiguration {
           |separators (even on Windows).
           |""".stripMargin
       ),
+      UserConfigurationOption(
+        "ammonite-jvm-properties",
+        """`[]`.""",
```
kpbochenek, line 113: I noticed with other parameters there is always a dot in the end, any idea why? Just curious

tgodzik: Not sure if really needed, but this is rendered here: https://scalameta.org/metals/docs/editors/vim.html#gradle-script

metals/src/main/scala/scala/meta/internal/metals/ammonite/Ammonite.scala
tgodzik: I remove the alternatives and using only user config now.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
tgodzik: fixed!

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
tgodzik: Added an additional error when we don't have a list.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
tgodzik: fixed

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: Removed!

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
ckipp01: 🤔 I don't think the dot is needed.

PR #1856, opened 2020-06-25
Update mdoc to 2.2.3 and power completions with worksheet's classpath

tgodzik:
Previously, worksheets didn't allow any additional dependencies, only the ones in the workspace. Now, it's possible to bring in additional deps like:

```scala
import $dep.`com.lihaoyi::scalatags:0.9.0`
import scalatags.Text.all._

val htmlFile = html(
  body(
    p("This is a big paragraph of text")
  )
) // TypedTag("html",List(WrappedArray(TypedTag("body",List(WrappedArray(TypedTag("p", List(WrappedArray(StringFrag("This is…

htmlFile.render // "<html><body><p>This is a big paragraph of text</p></body></html>"
```

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
@@ -1,5 +1,6 @@
 package scala.meta.internal.metals
 
+import java.nio.file.Path
 import java.nio.file.Paths
 import java.util.Collections
 import java.util.concurrent.ScheduledExecutorService
@@ -74,10 +75,25 @@ class Compilers(
     Collections.synchronizedMap(
       new java.util.HashMap[BuildTargetIdentifier, PresentationCompiler]
     )
+  private val jworksheetsCache: ju.Map[AbsolutePath, PresentationCompiler] =
+    Collections.synchronizedMap(
+      new java.util.HashMap[AbsolutePath, PresentationCompiler]
+    )
+
   private val cache = jcache.asScala
 
+  private val worksheetsCache = jworksheetsCache.asScala
+
   // The "rambo" compiler is used for source files that don't belong to a build target.
-  lazy val ramboCompiler: PresentationCompiler = {
+  lazy val ramboCompiler: PresentationCompiler = createRamboCompiler(
+    PackageIndex.scalaLibrary,
+    "metals-default"
+  )
+
+  def createRamboCompiler(
+      classpath: Seq[Path],
+      name: String
+  ): PresentationCompiler = {
     scribe.info(
       "no build target: using presentation compiler with only scala-library"
     )
@@ -86,8 +102,8 @@ class Compilers(
         .getOrElse(EmptySymbolSearch)
     val compiler =
       configure(new ScalaPresentationCompiler(), ramboSearch).newInstance(
-        s"metals-default-${mtags.BuildInfo.scalaCompilerVersion}",
-        PackageIndex.scalaLibrary.asJava,
+        s"$name-${mtags.BuildInfo.scalaCompilerVersion}",
+        classpath.asJava,
         Nil.asJava
       )
     ramboCancelable = Cancelable(() => compiler.shutdown())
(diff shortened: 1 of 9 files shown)


metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -177,10 +182,12 @@ object Embedded {
   private def downloadDependency(
       dep: Dependency,
       scalaVersion: String,
-      classfiers: Seq[String] = Seq.empty
+      classfiers: Seq[String] = Seq.empty,
+      resolution: ResolutionParams = ResolutionParams.create()
   ): List[Path] =
     fetchSettings(dep, scalaVersion)
       .addClassifiers(classfiers: _*)
+      .withResolutionParams(resolution)
```
olafurpg, line 190: FWIW, it's possible to provide a custom coursier logger for the download progress https://github.com/scalameta/mdoc/blob/54f40f32ed946d3676adab35cd05bb27a6055401/mdoc-interfaces/src/main/scala/mdoc/interfaces/Mdoc.java#L17

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -111,7 +111,12 @@ object Embedded {
       scalaVersion: String,
       scalaBinaryVersion: String
   ): URLClassLoader = {
-    val jars = downloadMdoc(scalaVersion, scalaBinaryVersion)
+    val resolutionParams = ResolutionParams
+      .create()
+
+    resolutionParams.addExclusion("io.get-coursier", "interface")
```
olafurpg, line 117: Add comment why?

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -329,14 +345,60 @@ class Compilers(
...
+      path: AbsolutePath,
+      classpath: List[Path]
+  ): Unit = {
+    val created = for {
+      targetId <- buildTargets.inverseSources(path)
+      info <- buildTargets.scalaTarget(targetId)
+      isSupported = ScalaVersions.isSupportedScalaVersion(info.scalaVersion)
+      _ = {
+        if (!isSupported) {
+          scribe.warn(s"unsupported Scala ${info.scalaVersion}")
```
olafurpg, line 381: This code path didn't exist before, why is that? Would it be possible to somehow move the management of compilers to `Compilers.scala`? There seems to be some overlap, maybe we can generate `BuildTargetIdentifier` for worksheets?

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
tgodzik: The default one used by Mdoc should just report things to System.err, which should be fine, no?

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
tgodzik: Not sure what you mean? It's invoked after evaluating the worksheet in case of changed dependencies, which in essence is a build target change. I am not sure how otherwise to trigger on that particular event. 

Alternatively, we could have worksheets as a different target and update the build target if dependencies change, which will be effectively the same with more indirection. We do need to somehow restart the presentation compiler.

> There seems to be some overlap, maybe we can generate BuildTargetIdentifier for worksheets?

I was thinking about having a separate build target for each worksheet, but since the worksheets can be a part of another build target I decided against. It also seemed like it might cause more work really and I planned to have this for the release tomorrow.

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
tgodzik: Added

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -329,14 +345,68 @@ class Compilers(
...
+            buffers,
+            fallback = Some(search)
+          )
+          newCompiler(scalac, info.scalaInfo, classpath, worksheetSearch)
+        }
+      )
+    }
+
+    created.getOrElse {
+      jworksheetsCache.put(
```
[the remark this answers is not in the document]
kpbochenek: okay this is actually foreach if not present :D scala should add method like forempty {} :P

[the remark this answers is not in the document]
kpbochenek: < resolved >

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -223,13 +232,42 @@ class WorksheetProvider(
...
+  }
+
   private def evaluateWorksheet(
       path: AbsolutePath,
       token: CancelToken
   ): EvaluatedWorksheet = {
     val mdoc = getMdoc(path)
     val input = path.toInputFromBuffers(buffers)
-    val worksheet = mdoc.evaluateWorksheet(input.path, input.value)
+    val relativePath = path.toRelative(workspace)
```
kpbochenek, line 255: that was this incompatibility with newest mdoc? :laughing:

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
```diff
@@ -55,6 +55,52 @@ abstract class BaseWorksheetLspSuite(scalaVersion: String)
...
+  test("completion-imports") {
+    for {
+      _ <- server.initialize(
+        s"""
+           |/metals.json
+           |{
+           |  "a": {}
+           |}
+           |/a/src/main/scala/foo/Main.worksheet.sc
+           |import $$dep.`com.lihaoyi::scalatags:0.9.0`
```
kpbochenek, line 67: just curious, why double $$, when in PR description you use $dep?
Is it related to some replacements happening in tests?

tgodzik: $$ is just so that we have $ in a string interpolation, since $ is a special character.

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
tgodzik: That's the change for worksheet to be able to include dependencies.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
olafurpg: Sounds good. I agree with your conclusion to keep this as is

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
olafurpg: It might be nice to somehow display the progress via LSP but looking at the coursier logger api it might not be so easy. This is fine

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -329,14 +345,68 @@ class Compilers(
...
+      isSupported = ScalaVersions.isSupportedScalaVersion(info.scalaVersion)
+      _ = {
+        if (!isSupported) {
+          scribe.warn(s"unsupported Scala ${info.scalaVersion}")
+        }
+      }
+      if isSupported
+      scalac <- buildTargets.scalacOptions(targetId)
+    } yield {
+      jworksheetsCache.put(
```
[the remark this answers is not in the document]
tgodzik: Added `()`, but this should not be an issue anyway, since that value is never used.

tgodzik:
I am planning to merge this as is currently. We could try to make each worksheet a separate build target, but that would require some additional work and might just cause an additional level of complication. Worksheets require new PC each time dependencies change, which normally is done during indexing, but in case of worksheets we don't need to reindex the entire workspace.

 I think I will need to come back to this at some point, but it will require some careful consideration and time.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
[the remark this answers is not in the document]
tgodzik: Changed to explicit: Option[Unit], which looks better.

PR #1861, opened 2020-06-27
Bump default mill version

ckipp01:


ckipp01:
> LGTM! Thanks for updating it! IT would be useful to have it done automatically. Does anyone know if it's possible to make Scala Steward update something inside the codebase?

I'm not sure. I can take a peak into that. We could also just do a `defaultMillVersion` in the `build.sbt` and use build info.

tgodzik:
> > LGTM! Thanks for updating it! IT would be useful to have it done automatically. Does anyone know if it's possible to make Scala Steward update something inside the codebase?
> 
> I'm not sure. I can take a peak into that. We could also just do a `defaultMillVersion` in the `build.sbt` and use build info.

Probably the same then as we do with ammonite, but it's not super important. I was just thinking aloud.

PR #1864, opened 2020-06-28
Set BuildTool version via BuildInfo.

ckipp01:
Previously we set the recommended version for the various BuildTools
by having the version hardcoded in the BuildTool file. This moves
the version for Maven, Mill, and Gradle into the build.sbt file and
uses BuildInfo to access those versions like we currently do with
the SBT build tool. This will also allow us to get some help from our
friend ScalaSteward.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -207,6 +207,9 @@ lazy val V = new {
   val coursier = "2.0.0-RC6-21"
   val coursierInterfaces = "0.0.22"
   val ammonite = "2.1.4-6-2179b35"
+  val mill = "0.7.3"
+  val gradle = "6.1.0"
+  val maven = "3.6.3"
 }
 
 val genyVersion = Def.setting {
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -207,6 +207,9 @@ lazy val V = new {
   val coursier = "2.0.0-RC6-21"
   val coursierInterfaces = "0.0.22"
   val ammonite = "2.1.4-6-2179b35"
+  val mill = "0.7.3"
+  val gradle = "6.1.0"
```
ckipp01, line 211: I didn't set this to the latest since the [Bloop site](https://scalacenter.github.io/bloop/docs/build-tools/gradle) mentions that the latest test/supported in 6.1

ckipp01:
> I wonder if we need something like @alexarchambault did for ammonite:
> 
> ```
>       // Only here to have scala-steward update V.ammonite.
>       // Not actually used in tests.
>       "com.lihaoyi" %% "ammonite-util" % V.ammonite intransitive ()
> ```

Ahhh, of course. If not, how would Scala steward detect that this variable was actually being used in a dependency. That makes sense. I can change this as well.

ckipp01:
I'm actually not sure this is going to work. For example, it looks like all of the artifacts post v0.7 are only published with 2.13, so then when we include it and it tries to fetch it, it doesn't work 🤔

ckipp01:
Closing this for now. I'll need to look into this a bit more.

Issue #1838, opened 2019-01-23
Support sbt files


tpolecat:
It would be helpful if Metals supported `.sbt` files … I have no idea how hard this would be but I didn't see an issue for it so here you go :-)

Thanks as always for your hard work.

tgodzik:
In the end https://github.com/scalacenter/bloop/pull/1297 will most likely not get merged due to reasons specified by @jvican  We need to get to the approach with mutiple build connection, which we already do for Ammonite. This should be simpler though, since we are always conencting to the same build server.

@dos65 Will be taking an another stab at it. Please let us know if you encounter any issues. I've already seen that you are on the right track with the previous PR.

dos65:
I've reworked my previous PR in metals - #1865 . Now it works over multiply BSP connections.
But it anyway depends on scalacenter/bloop#1219

PR #1865, opened 2020-06-28
Sbt support - multiply bsp connections

dos65:
This PR fixes https://github.com/scalameta/metals/issues/1838 by establishing additional BSP connection for each Sbt meta project.

It adds the following for Sbt:
- completions
- go to definition
- hover

Doesn't yet add:
- diagnostics
- references
- rename

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -169,6 +169,7 @@ commands += Command.command("save-expect") { s =>
 lazy val V = new {
   val scala210 = "2.10.7"
   val scala211 = "2.11.12"
+  val `scala2.12.10` = "2.12.10"
   val scala212 = "2.12.12"
   val scala213 = "2.13.3"
   val scalameta = "4.3.20"
@@ -491,12 +492,14 @@ lazy val testSettings: Seq[Def.Setting[_]] = List(
 
 def crossPublishLocal(scalaV: String) =
   Def.task[Unit] {
+    val versionValue = version.in(ThisBuild).value
     // Runs `publishLocal` for mtags with `scalaVersion := $scalaV`
     val newState = Project
       .extract(state.value)
       .appendWithSession(
         List(
           scalaVersion.in(mtags) := scalaV,
+          version.in(ThisBuild) := versionValue,
           useSuperShell.in(ThisBuild) := false
         ),
         state.value
@@ -524,7 +527,7 @@ def publishBinaryMtags =
     .in(interfaces)
     .dependsOn(
       publishAllMtags(
-        List(V.scala211, V.scala212, V.scala213, V.scala3)
+        List(V.scala211, V.`scala2.12.10`, V.scala212, V.scala213, V.scala3)
       )
     )
 
--- a/build.sbt
+++ b/build.sbt
@@ -169,7 +169,7 @@ commands += Command.command("save-expect") { s =>
 lazy val V = new {
   val scala210 = "2.10.7"
   val scala211 = "2.11.12"
-  val `scala2.12.10` = "2.12.10"
+  val sbtScala = "2.12.10"
   val scala212 = "2.12.12"
   val scala213 = "2.13.3"
   val scalameta = "4.3.20"
@@ -527,7 +527,7 @@ def publishBinaryMtags =
     .in(interfaces)
     .dependsOn(
       publishAllMtags(
-        List(V.scala211, V.`scala2.12.10`, V.scala212, V.scala213, V.scala3)
+        List(V.scala211, V.sbtScala, V.scala212, V.scala213, V.scala3)
       )
     )
 
--- a/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
@@ -87,34 +87,32 @@ case class SbtBuildTool(
       workspace: AbsolutePath
   ): Unit = {
 
-    def containsScalaOrSbtFiles(dir: AbsolutePath): Boolean =
-      dir.list.exists { f =>
-        val name = f.filename
-        f.isFile && (f.isScala || f.isSbt) && name != "metals.sbt"
-      }
-
     def sbtMetaDirs(
         dir: AbsolutePath,
+        parentHasSbtSrc: Boolean,
         acc: List[AbsolutePath]
     ): List[AbsolutePath] = {
-      if (dir.exists && containsScalaOrSbtFiles(dir)) {
-        val nextMeta = dir.resolve("project")
-        sbtMetaDirs(nextMeta, nextMeta :: acc)
+
+      if (dir.exists) {
+        val (hasScalaSrc, hasSbtSrc) = dir.list.foldLeft((false, false)) {
+          case ((_, sbtFlag), f) if f.isScala => (true, sbtFlag)
+          case ((scalaFlag, _), f) if f.isSbt && f.filename != "metals.sbt" =>
+            (scalaFlag, true)
+          case (acc, _) => acc
+        }
+        val goNext = hasScalaSrc || hasSbtSrc || parentHasSbtSrc
+
+        if (goNext) {
+          val next = dir.resolve("project")
+          sbtMetaDirs(next, hasSbtSrc, dir :: acc)
+        } else acc
       } else {
         acc
       }
     }
 
-    // by default create the following directory structure
-    //  workspace/
-    //     project/
-    //       metals.sbt
-    //       project/
-    //          metals.sbt
-    // and check if there are more inner meta projects
     val mainMeta = workspace.resolve("project")
-    val metaMeta = mainMeta.resolve("project")
-    val dirs = sbtMetaDirs(metaMeta, List(mainMeta, metaMeta))
+    val dirs = sbtMetaDirs(mainMeta, true, List(mainMeta))
 
     dirs.foreach(writeSbtMetalsPlugin0)
   }
--- a/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
@@ -3,6 +3,7 @@ import java.nio.charset.StandardCharsets
 import java.nio.file.Files
 import java.util.Properties
 
+import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals._
 import scala.meta.io.AbsolutePath
 
@@ -85,12 +86,47 @@ case class SbtBuildTool(
   private def writeSbtMetalsPlugin(
       workspace: AbsolutePath
   ): Unit = {
+
+    def containsScalaOrSbtFiles(dir: AbsolutePath): Boolean =
+      dir.list.exists { f =>
+        val name = f.filename
+        f.isFile && (f.isScala || f.isSbt) && name != "metals.sbt"
+      }
+
+    def sbtMetaDirs(
+        dir: AbsolutePath,
+        acc: List[AbsolutePath]
+    ): List[AbsolutePath] = {
+      if (dir.exists && containsScalaOrSbtFiles(dir)) {
+        val nextMeta = dir.resolve("project")
+        sbtMetaDirs(nextMeta, nextMeta :: acc)
+      } else {
+        acc
+      }
+    }
+
+    // by default create the following directory structure
+    //  workspace/
+    //     project/
+    //       metals.sbt
+    //       project/
+    //          metals.sbt
+    // and check if there are more inner meta projects
+    val mainMeta = workspace.resolve("project")
+    val metaMeta = mainMeta.resolve("project")
+    val dirs = sbtMetaDirs(metaMeta, List(mainMeta, metaMeta))
+
+    dirs.foreach(writeSbtMetalsPlugin0)
+  }
+
+  private def writeSbtMetalsPlugin0(
+      projectDir: AbsolutePath
+  ): Unit = {
     if (userConfig().bloopSbtAlreadyInstalled) return
     val versionToUse = userConfig().currentBloopVersion
     val bytes = SbtBuildTool
       .sbtPlugin(versionToUse)
       .getBytes(StandardCharsets.UTF_8)
-    val projectDir = workspace.resolve("project")
     projectDir.toFile.mkdir()
     val metalsPluginFile = projectDir.resolve("metals.sbt")
     val pluginFileShouldChange = !metalsPluginFile.isFile ||
--- a/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
@@ -93,19 +93,20 @@ case class SbtBuildTool(
         acc: List[AbsolutePath]
     ): List[AbsolutePath] = {
 
+      val next = dir.resolve("project")
       if (dir.exists) {
-        val (hasScalaSrc, hasSbtSrc) = dir.list.foldLeft((false, false)) {
-          case ((_, sbtFlag), f) if f.isScala => (true, sbtFlag)
-          case ((scalaFlag, _), f) if f.isSbt && f.filename != "metals.sbt" =>
-            (scalaFlag, true)
-          case (acc, _) => acc
-        }
+        val files = dir.list.toList
+        val hasScalaSrc = files.exists(_.isScala)
+        val hasSbtSrc = files.exists(f => f.isSbt && f.filename != "metals.sbt")
         val goNext = hasScalaSrc || hasSbtSrc || parentHasSbtSrc
 
-        if (goNext) {
-          val next = dir.resolve("project")
+        if (goNext)
           sbtMetaDirs(next, hasSbtSrc, dir :: acc)
-        } else acc
+        else
+          acc
+
+      } else if (parentHasSbtSrc) {
+        next :: acc
       } else {
         acc
       }
@@ -125,7 +126,7 @@ case class SbtBuildTool(
     val bytes = SbtBuildTool
       .sbtPlugin(versionToUse)
       .getBytes(StandardCharsets.UTF_8)
-    projectDir.toFile.mkdir()
+    projectDir.toFile.mkdirs()
     val metalsPluginFile = projectDir.resolve("metals.sbt")
     val pluginFileShouldChange = !metalsPluginFile.isFile ||
       !metalsPluginFile.readAllBytes.sameElements(bytes)
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -264,9 +264,7 @@ final class BuildTargets(
   def workspaceDirectory(
       buildTarget: BuildTargetIdentifier
   ): Option[AbsolutePath] =
-    targetToConnection
-      .get(buildTarget)
-      .map(_.workspaceDirectory)
+    buildServerOf(buildTarget).map(_.workspaceDirectory)
 
   /**
    * Returns the first build target containing this source file.
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -15,6 +15,7 @@ import scala.util.control.NonFatal
 
 import scala.meta.internal.io.PathIO
 import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.ammonite.Ammonite
 import scala.meta.internal.mtags.Mtags
 import scala.meta.internal.mtags.Symbol
 import scala.meta.io.AbsolutePath
@@ -29,7 +30,9 @@ import ch.epfl.scala.bsp4j.WorkspaceBuildTargetsResult
 /**
  * In-memory cache for looking up build server metadata.
  */
-final class BuildTargets() {
+final class BuildTargets(
+    ammoniteBuildServer: BuildTargetIdentifier => Option[BuildServerConnection]
+) {
   private var workspace = PathIO.workingDirectory
   def setWorkspaceDirectory(newWorkspace: AbsolutePath): Unit = {
     workspace = newWorkspace
@@ -53,6 +56,9 @@ final class BuildTargets() {
     new ConcurrentLinkedQueue[AbsolutePath => Seq[BuildTargetIdentifier]]()
   // if workspace contains symlinks, original source items are kept here and source items dealiased
   private val originalSourceItems = ConcurrentHashSet.empty[AbsolutePath]
+
+  private val targetToConnection =
+    new mutable.HashMap[BuildTargetIdentifier, BuildServerConnection]
 
   val buildTargetsOrder: BuildTargetIdentifier => Int = {
     (t: BuildTargetIdentifier) =>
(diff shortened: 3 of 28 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1597,17 +1604,17 @@ class MetalsLanguageServer(
...
         }
         result <- maybeBuild match {
           case Some(build) =>
             val result = connectToNewBuildServer(build)
-            build.onReconnection { reconnected =>
-              connectToNewBuildServer(reconnected)
-                .flatMap(compileAllOpenFiles)
-                .ignoreValue
-            }
+            // TODO how to implement reconnection correctly ?
```
dos65, line 1612: I'm not sure how to implement this thing correctly for multiply connections.

When reconnect happens?
I've tried to kill `bloop.Server` but this part of code wasn't executed

mtags/src/main/scala-2/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -60,6 +60,15 @@ case class ScalaPresentationCompiler(
   val logger: Logger =
     Logger.getLogger(classOf[ScalaPresentationCompiler].getName)
 
+  val sourceModificator: SourceModificator = {
+    val autoImports =
+      if (config.autoImports().isPresent()) Some(config.autoImports().get())
```
tgodzik, line 65: There should be `.asScala` extension method available.

mtags/src/main/scala/scala/meta/internal/pc/SourceModificator.scala
```diff
@@ -0,0 +1,104 @@
...
+  ): DefinitionResult
+  def forHover(pos: OffsetParams)(
+      f: OffsetParams => Option[Hover]
+  ): Option[Hover]
+  def forCompletion(pos: OffsetParams)(
+      f: OffsetParams => CompletionList
+  ): CompletionList
+}
+
+object SourceModificator {
```
tgodzik, line 25: This looks good, but maybe we can do it alongside the Ammonite fixes in `Compilers.ammoniteInputPosOpt` and `Ammonite.adjustHoverResp` etc.

We wouldn't need to forward the auto imports lists and we could generalize the Ammonite approach.

tgodzik:
I checked out you PRs and it seems that a lot of the things are working really well. One thing that is still a problem is compiling, which might explain the issues you had. We currently don't find sources for any of the sbt file. There are two problems:
- we don't trigger compilation for *.sbt files due to:

https://github.com/scalameta/metals/blob/48b20201f3357738ce472c40dcae47b0cccc56fb/mtags/src/main/scala/scala/meta/internal/mtags/CommonMtagsEnrichments.scala#L224

It should be possible to just add .sbt, but best check if that has any unintended consequences.

-  sources for meta builds seem to be badly defined, though I am not an sbt expert:
```       
"sources": [
            "/home/tgodzik/Documents/workspaces/metals-sample/project/src/main/scala-2.12",
            "/home/tgodzik/Documents/workspaces/metals-sample/project/src/main/scala",
            "/home/tgodzik/Documents/workspaces/metals-sample/project/src/main/java",
            "/home/tgodzik/Documents/workspaces/metals-sample/project/src/main/scala-sbt-1.0",
            "/home/tgodzik/Documents/workspaces/metals-sample/project/target/scala-2.12/sbt-1.0/src_managed/main"
],
```
should probably be:

```       
"sources": [
            "/home/tgodzik/Documents/workspaces/metals-sample/build.sbt",
            "/home/tgodzik/Documents/workspaces/metals-sample/project/"
],
```


So when we compile for example `build.sbt` we will no find target for it. We should fix it in the Bloop sbt plugin most likely. We could skip adding `build.sbt` and just for any sbt files look for a target corresponding to the project dir, which might be useful when adding new files.

dos65:
@tgodzik  Thanks for the review. I will fix the comments a little bit later.

About compilation of `*.sbt` files.
I think that bloop does his work correctly. I mean `*.sbt` file shouldn't be presented in `sources` because it can't be compiled as a usual source file.
[Sbt does it internally](https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/internal/EvaluateConfigurations.scala) and it doesn't look like a typical compilation. 

For a now, I'm not sure what might be the best option to solve this problem.
For example, this thing might be handled by a special compiler for sbt files in Metals. However, I don't know how to solve the situation when a user opens Metals on a fresh project that contains errors in the build definition. There is no way to get the diagnostic and even start such a compiler internally without bloop files.

I think it would be better to investigate how it might be solved later as a separate issue.

tgodzik:
> @tgodzik Thanks for the review. I will fix the comments a little bit later.
> 
> About compilation of `*.sbt` files.
> I think that bloop does his work correctly. I mean `*.sbt` file shouldn't be presented in `sources` because it can't be compiled as a usual source file.
> [Sbt does it internally](https://github.com/sbt/sbt/blob/develop/main/src/main/scala/sbt/internal/EvaluateConfigurations.scala) and it doesn't look like a typical compilation.
> 
> For a now, I'm not sure what might be the best option to solve this problem.
> For example, this thing might be handled by a special compiler for sbt files in Metals. However, I don't know how to solve the situation when a user opens Metals on a fresh project that contains errors in the build definition. There is no way to get the diagnostic and even start such a compiler internally without bloop files.
> 
> I think it would be better to investigate how it might be solved later as a separate issue.

Sure, we can focus on the completions etc. We can follow up with an issue about the diagnostics. I didn't realise that Bloop doesn't actually support them for sbt.

tgodzik:
Additional thing that I noticed is that the some targets after the change in Bloop have scala version set to null, which is causing the server not to start in some cases. This should come up in the tests, but if not we should make sure it all works after the changes in Bloop.

Also, let's add some additional tests to SbtLspSuite to assert completions/hover the same way as in the BaseAmmoniteSuite.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -85,12 +85,41 @@ case class SbtBuildTool(
...
+    }
+
+    // by default create the following directory structure
+    //  workspace/
+    //     project/
+    //       metals.sbt
+    //       project/
+    //          metals.sbt
+    // and check if there are more inner meta projects
+    val mainMeta = workspace.resolve("project")
```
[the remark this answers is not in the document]
dos65: If they aren't then they will be created in function below.
https://github.com/scalameta/metals/blob/71ebf3d7b546d98483b4426d176d01a84bd70d69/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala#L117-L125

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -85,12 +85,41 @@ case class SbtBuildTool(
   private def writeSbtMetalsPlugin(
       workspace: AbsolutePath
   ): Unit = {
+
+    def sbtMetaDirs(
+        dir: AbsolutePath,
+        acc: List[AbsolutePath]
+    ): List[AbsolutePath] = {
+      val projectDir = dir.resolve("project")
```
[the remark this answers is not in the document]
dos65: fixed

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -123,13 +123,15 @@ class MetalsLanguageServer(
     new AtomicReference[b.BuildTargetIdentifier]()
   private val definitionIndex = newSymbolIndex()
   private val symbolDocs = new Docstrings(definitionIndex)
-  var buildServer: Option[BuildServerConnection] =
-    Option.empty[BuildServerConnection]
+  var bspSession: Option[BspSession] =
+    Option.empty[BspSession]
+  var targetToConnection =
+    new mutable.HashMap[b.BuildTargetIdentifier, BuildServerConnection]
```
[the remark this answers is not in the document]
dos65: I agree, it seems that `BuildTargets` is a better place for this thing.
Should I move it there?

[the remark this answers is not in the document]
tgodzik: That would be best. I would try to not add new things to MetalsLanguageServer as it is already quite big.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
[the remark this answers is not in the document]
tgodzik: Cool. One additional thing that I did notice is that when we create `metals.sbt` file then the next `bloopInstall` will create a next level and then the next during consequtive install, which we should not do. We should only create those files if `metals.sbt` is not the only thing in the project directory.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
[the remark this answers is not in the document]
dos65: Done

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
```diff
@@ -1,6 +1,7 @@
 package scala.meta.pc;
 
 import java.util.HashMap;
+import java.util.List;
```
tgodzik, line 4: This is most likely no longer needed.

mtags/src/main/scala/scala/meta/internal/pc/PresentationCompilerConfigImpl.scala
```diff
@@ -27,6 +27,7 @@ case class PresentationCompilerConfigImpl(
     timeoutDelay: Long = 20,
     timeoutUnit: TimeUnit = TimeUnit.SECONDS
 ) extends PresentationCompilerConfig {
+
```
tgodzik, line 30: The newline can be remove, no need to have changes in this file.

tests/unit/src/test/scala/tests/SbtSuite.scala
```diff
@@ -0,0 +1,82 @@
+package tests
+
+class SbtSuite extends BaseLspSuite("sbt") with ScriptsAssertions {
```
tgodzik, line 3: Maybe we can just add it to `SbtLspSuite` in slow tests?

tests/unit/src/test/scala/tests/SbtSuite.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  val scalaVersion = "2.12.10"
```
tgodzik, line 5: We can use BuildInfo, let's add to imports:
`import scala.meta.internal.metals.{BuildInfo => V}`

and then:
```suggestion
  val scalaVersion = V.scala212
```

tests/unit/src/test/scala/tests/SbtSuite.scala
dos65: Moved it into `SbtLspSuite`

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -462,27 +460,79 @@ class Compilers(
...
+      }
+  }
+
   private def withPC[T](
       params: TextDocumentPositionParams,
       interactiveSemanticdbs: Option[InteractiveSemanticdbs]
   )(fn: (PresentationCompiler, Position) => T): Option[T] = {
+    val f =
+      (c: PresentationCompiler, p: Position, adj: AdjustLspData) => fn(c, p)
+    withPCAndAdjustLsp(params, interactiveSemanticdbs)(f)
```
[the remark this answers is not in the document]
dos65: fixed

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -101,18 +107,20 @@ final class BuildTargets() {
   def allBuildTargetIds: Seq[BuildTargetIdentifier] =
     all.toSeq.map(_.info.getId())
   def all: Iterator[ScalaTarget] =
-    for {
-      (id, target) <- buildTargetInfo.iterator
-      scalac <- scalacTargetInfo.get(id)
-      scalaTarget <- target.asScalaBuildTarget
-    } yield ScalaTarget(target, scalaTarget, scalac)
+    buildTargetInfo.iterator.flatMap({ case (_, bt) => toScalaTarget(bt) })
```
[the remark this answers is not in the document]
dos65: fixed

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -85,12 +87,41 @@ case class SbtBuildTool(
...
+      if (projectDir.exists) {
+        val toCreate = projectDir.resolve("project")
+        sbtMetaDirs(projectDir, toCreate :: acc)
```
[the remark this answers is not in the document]
dos65: fixed

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
dos65: I'm not quite sure about this change.

https://github.com/scalameta/metals/blob/95e8daf51964c373b769f07788e2edd1570889ee/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1630-L1636

Would it be ok to leave it in this way?

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -226,6 +237,28 @@ final class BuildTargets() {
   ): Option[ScalacOptionsItem] =
     scalacTargetInfo.get(buildTarget)
 
+  /**
+   * Returns meta build target for `*.sbt`  files.
+   * It selects build target by directory of its connection
+   *   because `*.sbt` aren't included in `sourceFiles` set
+   */
+  def sbtBuildScalaTarget(file: AbsolutePath): Option[ScalaTarget] = {
```
tgodzik, line 245: Do we return the right sbt target for the Scala files located in `project` directories? I think if Bloop doesn't provide use with proper source roots we could add a custom logic for resolving target by source file in `inverseSources`. If no build target is found we can look if this file should belong to an sbt build target. What do you think?

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -85,12 +87,47 @@ case class SbtBuildTool(
   private def writeSbtMetalsPlugin(
       workspace: AbsolutePath
   ): Unit = {
+
+    def containsScalaOrSbtFiles(dir: AbsolutePath): Boolean =
+      dir.list.exists { f =>
+        val name = f.filename
+        f.isFile && (f.isScala || f.isSbt) && name != "metals.sbt"
```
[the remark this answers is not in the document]
dos65: No, they have the same requirements as sbt files.
I checked this thing

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
dos65: This method is only for `.*sbt` files.
For `*.scala` files from meta build `inverseSources` is used as usual.

I think when we start implementing compilation for `.*sbt` files it would be better to move this logic into `inverseSources` method. But for now, I'm not sure if there is a need for that.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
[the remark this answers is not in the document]
dos65: Now it's changed.

Actually, it was me who forgot to push that, Github is innocent )

tgodzik:
Looks like you still need to run `sbt scalafixAll`. 

I will check it all out tomorrow to see if we still need to fix something, but otherwise if no one else wants to review, I will try to merge it this week.

tgodzik:
I also added the description, which we totally forgot about. Please do update it if something is not yet correct.

tgodzik:
It looks like a number of tests are failing on:
```
[error] ==> X tests.QuickBuildSuite.basic  2.615s java.lang.IllegalArgumentException: Missing scheme
[error]     at java.nio.file.Paths.get(Paths.java:134)
[error]     at scala.meta.internal.metals.MetalsEnrichments$XtensionString.toAbsolutePath(MetalsEnrichments.scala:445)
[error]     at scala.meta.internal.metals.MetalsEnrichments$XtensionString.toAbsolutePath(MetalsEnrichments.scala:442)
[error]     at tests.TestingServer.$anonfun$toSemanticdbTextDocument$3(TestingServer.scala:1212)
[error]     at scala.collection.immutable.List.map(List.scala:293)
[error]     at tests.TestingServer.$anonfun$toSemanticdbTextDocument$1(TestingServer.scala:1207)
[error]     at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36)
[error]     at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33)
[error]     at scala.meta.tokens.Tokens.foreach(Tokens.scala:29)
[error]     at tests.TestingServer.toSemanticdbTextDocument(TestingServer.scala:1195)
[error]     at tests.TestingServer.$anonfun$workspaceDefinitions$2(TestingServer.scala:1245)
[error]     at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:285)
[error]     at scala.collection.Iterator.foreach(Iterator.scala:943)
[error]     at scala.collection.Iterator.foreach$(Iterator.scala:943)
[error]     at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
[error]     at scala.collection.IterableLike.foreach(IterableLike.scala:74)
[error]     at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
[error]     at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
[error]     at scala.collection.TraversableLike.map(TraversableLike.scala:285)
[error]     at scala.collection.TraversableLike.map$(TraversableLike.scala:278)
[error]     at scala.collection.AbstractTraversable.map(Traversable.scala:108)
[error]     at tests.TestingServer.workspaceDefinitions(TestingServer.scala:1244)
[error]     at tests.QuickBuildSuite.$anonfun$new$6(QuickBuildSuite.scala:53)
```

any idea what it  is about?

Also, the `.bloop` files are not being generated for the meta builds. `project/project/.bloop` etc We might need to still fix that inside Bloop. I am ok with merging without the meta build, but we should at least file an issue to figure out what is going on.

dos65:
@tgodzik 
I will check what happens with tests.

> Also, the .bloop files are not being generated for the meta builds. project/project/.bloop etc We might need to still fix that inside Bloop. 

There is nothing wrong with bloop. It generates `.bloop` directory for meta builds correctly.
Maybe I broke smth during rebase.

dos65:
@tgodzik

The problem was in `FilesystemSemanticdbs` because I added additional logic to select the workspace directory to support `*.scala` files for meta-builds. Now it's fixed.

BTW, how to trigger CI build?

tgodzik:
@dos65 the build is not running due to a conflict in `build.sbt`

tgodzik:
> @tgodzik
> 
> The problem was in `FilesystemSemanticdbs` because I added additional logic to select the workspace directory to support `*.scala` files for meta-builds. Now it's fixed.
> 

What about for example `project/other.sbt` file? There seems to be no `project/project/.bloop` generated so we are not establishing the connection for the meta builds. Should we file an issue in Bloop?

dos65:
> What about for example project/other.sbt file? There seems to be no project/project/.bloop

Oh, I've realized that we should go deeper and create an additional `project` directory.
Will fix that.

tgodzik:
> > What about for example project/other.sbt file? There seems to be no project/project/.bloop
> 
> Oh, I've realized that we should go deeper and create an additional `project` directory.
> Will fix that.

Project directory is there, but the Bloop sbt plugin doesn't seem to be generated the config files.

dos65:
> Project directory is there, but the Bloop sbt plugin doesn't seem to be generated the config files.

The implementation of `writeSbtMetalsPlugin` wasn't fully correct. 
Maybe it will be easier to illustrate with an example. 
```
build.sbt            # belongs to `meta-build`

project/              # `meta-build` directory
   
    metals.sbt     # enables `bloopInstall` for main build
    
    project/          # `meta-build-build` directory
       
       metals.sbt   # enables bloopInstall for `meta-build` <- without this bloop will not generate config
```
So, to work with `build.sbt` it's required to create + 2 inner project directories with `metals.sbt` file.

Fixed in [this commit](https://github.com/scalameta/metals/pull/1865/commits/cdaa176b1af277e6312e68e56d25eefebf0ca22d)

dos65:
@tgodzik
Finally, all problems were fixed. Tests pass!

build.sbt
```diff
@@ -169,12 +169,13 @@ commands += Command.command("save-expect") { s =>
 lazy val V = new {
   val scala210 = "2.10.7"
   val scala211 = "2.11.12"
+  val `scala2.12.10` = "2.12.10"
```
tgodzik, line 172: `sbtScala` maybe?

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -25,6 +26,7 @@ case class SbtBuildTool(
 
   override def version: String = workspaceVersion.getOrElse(recommendedVersion)
   override def args(workspace: AbsolutePath): List[String] = {
+    println()
```
tgodzik, line 29: Remove?

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -85,13 +87,50 @@ case class SbtBuildTool(
...
+
+    def sbtMetaDirs(
+        dir: AbsolutePath,
+        parentHasSbtSrc: Boolean,
+        acc: List[AbsolutePath]
+    ): List[AbsolutePath] = {
+
+      val next = dir.resolve("project")
+      if (dir.exists) {
+        val (hasScalaSrc, hasSbtSrc) = dir.list.foldLeft((false, false)) {
```
tgodzik, line 99: Maybe this could be easier to udnerstand:
```scala
val files = dir.list
val hasScalaSrc = files.exists(_.isScala)
val hasSbtSrc = files.exists(f => f.isSbt && f.filename != "metals.sbt" )
```
Granted it's not all in one go, but this list is rather small.

tgodzik:
It seems that completions are still not working:

image

.bloop directory is not generated in project/project and I am not getting anything in `other.sbt`. I am not sure though we can fix it in this PR alone.

dos65:
@tgodzik Yes you right. I reproduced this thing too on a fresh project.
It somehow depends on the existing state. If I run `import build` one more time it created previously missing `.bloop` directory.

tgodzik:
> @tgodzik Yes you right. I reproduced this thing too on a fresh project.
> It somehow depends on the existing state. If I run `import build` one more time it creates previously missing `.bloop` directory.

Weird, doesn't seem to happen for me. I can never get it generated. Should the file contain anything?

dos65:
> Weird, doesn't seem to happen for me. I can never get it generated. Should the file contain anything?

No, I added an empty `other.sbt` file. Maybe the difference is in sbt version. My sample project uses `1.3.8`

tgodzik:
> > Weird, doesn't seem to happen for me. I can never get it generated. Should the file contain anything?
> 
> No, I added an empty `other.sbt` file. Maybe the difference is in sbt version. My sample project uses `1.3.8`

Same here. The only things I am getting in the logs are:
```
2020.07.30 13:21:07 INFO  skipping build import with status 'Started'
2020.07.30 13:21:10 INFO  [info] Loading project definition from /home/tgodzik/Documents/workspaces/metals-sample/project/project/project
2020.07.30 13:21:10 INFO  [info] Loading settings for project metals-sample-build-build from metals.sbt ...
2020.07.30 13:21:11 INFO  [info] Loading project definition from /home/tgodzik/Documents/workspaces/metals-sample/project/project
2020.07.30 13:21:11 INFO  [info] Loading settings for project metals-sample-build from other.sbt,metals.sbt ...
2020.07.30 13:21:11 INFO  [info] Loading project definition from /home/tgodzik/Documents/workspaces/metals-sample/project
2020.07.30 13:21:13 INFO  [success] Generated .bloop/metals-sample-build.json
2020.07.30 13:21:13 INFO  [success] Total time: 2 s, completed Jul 30, 2020, 1:21:13 PM
2020.07.30 13:21:13 INFO  [info] Loading settings for project root from build.sbt ...
```

PR #1866, opened 2020-06-29
Gitignore .ammonite directory + update docs

kpbochenek:
It is more optional than `.metals` or `.bloop` because not everyone uses ammonite scripts but I thought it might be good to ignore `.ammonite` by default.

### DIFF
--- a/metals-docs/src/main/scala/docs/GenericModifier.scala
+++ b/metals-docs/src/main/scala/docs/GenericModifier.scala
@@ -10,18 +10,21 @@ class GenericModifier extends StringModifier {
 
   override def process(info: String, code: Input, reporter: Reporter): String =
     s"""
-       |## Gitignore `project/metals.sbt` `.metals/` and `.bloop/`
+       |## Files and Directories to include in your Gitignore
        |
-       |The Metals server places logs and other files in the `.metals/` directory. The
+       |The Metals server places logs and other files in the `.metals` directory. The
        |Bloop compile server places logs and compilation artifacts in the `.bloop`
        |directory. Bloop plugin that generates Bloop configuration is added in the 
-       |`project/metals.sbt` file. It's recommended to ignore these directories and file
+       |`project/metals.sbt` file. Working with Ammonite scripts will place
+       |compiled scripts into the `.ammonite` directory.
+       |It's recommended to exclude these directories and file
        |from version control systems like git.
        |
        |```sh
        |# ~/.gitignore
        |.metals/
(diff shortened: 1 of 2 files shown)


metals-docs/src/main/scala/docs/GenericModifier.scala
```diff
@@ -10,18 +10,21 @@ class GenericModifier extends StringModifier {
 
   override def process(info: String, code: Input, reporter: Reporter): String =
     s"""
-       |## Gitignore `project/metals.sbt` `.metals/` and `.bloop/`
+       |## Gitignore `project/metals.sbt` `.metals/` `.ammonite/` and `.bloop/`
```
ckipp01, line 13:
```suggestion
       |## Files and Directories to include in your Gitignore
```
What do you think of instead of listing them all in the actual header, we just change it to be more generic here, and then be more descriptive down below?

metals-docs/src/main/scala/docs/GenericModifier.scala
```diff
@@ -10,18 +10,21 @@ class GenericModifier extends StringModifier {
...
        |
-       |The Metals server places logs and other files in the `.metals/` directory. The
+       |The Metals server places logs and other files in the `.metals` directory. The
        |Bloop compile server places logs and compilation artifacts in the `.bloop`
        |directory. Bloop plugin that generates Bloop configuration is added in the 
-       |`project/metals.sbt` file. It's recommended to ignore these directories and file
+       |`project/metals.sbt` file. Working with ammonite scripts will place
```
ckipp01, line 18:
```suggestion
       |`project/metals.sbt` file. Working with Ammonite scripts will place
```

metals-docs/src/main/scala/docs/GenericModifier.scala
```diff
@@ -10,18 +10,21 @@ class GenericModifier extends StringModifier {
...
+       |compiled scripts into `.ammonite` directory.
```
ckipp01, line 19:
```suggestion
       |compiled scripts into the `.ammonite` directory.
```

metals-docs/src/main/scala/docs/GenericModifier.scala
```diff
@@ -10,18 +10,21 @@ class GenericModifier extends StringModifier {
...
+       |It's recommended to ignore these directories and file
```
ckipp01, line 20:
```suggestion
       |It's recommended to exclude these
```

PR #1881, opened 2020-07-01
Changes for default branch moving from master to main

ckipp01:
Following up on a conversation had offline, these would be the changes that would need to happen from the code perspective for the change from the default branch being main instead of master. There are just a couple other things that would need to happen from someone with admin access, which you can see [here](https://www.hanselman.com/blog/EasilyRenameYourGitDefaultBranchFromMasterToMain.aspx).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -140,7 +140,7 @@ onLoad.in(Global) ~= { old =>
       prePush,
       """#!/bin/sh
         |set -eux
-        |bin/scalafmt --diff
+        |bin/scalafmt --diff --diff-branch main
         |git diff --exit-code
         |""".stripMargin.getBytes()
     )
--- a/docs/editors/debug-adapter-protocol.md
+++ b/docs/editors/debug-adapter-protocol.md
@@ -15,16 +15,16 @@ capabilities exposed by the client.
 ### Via code lenses
 
 The editor needs to handle two commands in its language client extension:
-[`metals-run-session-start`](https://github.com/scalameta/metals/blob/master/metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala#L56)
+[`metals-run-session-start`](https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala#L56)
 and
-[`metals-debug-session-start`](https://github.com/scalameta/metals/blob/master/metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala#L78).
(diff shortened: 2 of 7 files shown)


build.sbt
```diff
@@ -141,7 +141,7 @@ onLoad.in(Global) ~= { old =>
       """#!/bin/sh
         |set -eux
         |bin/scalafmt --diff
```
olafurpg, line 143: I think this needs “-diff-branch=main”

ckipp01: ah, whoops, I put in on the line below with `git diff` 😬 

I'll switch this.

gabro:
As discussed offline, +1 on my side.
It’s a nice gesture that also adds clarity.

Thanks for following up

ckipp01:
Sounds good 👍 I'll probably leave this for @tgodzik to merge since it will need to be done at the same time as the switching default branch in the settings.

gabro:
I would also probably remove `master` after the switch, so that if someone is automatically building from master for some reason they won’t accidentally get an outdated version

kpbochenek:
`.travis.yml` needs change too

docs/editors/debug-adapter-protocol.md
```diff
@@ -15,16 +15,16 @@ capabilities exposed by the client.
...
 and
-[`metals-debug-session-start`](https://github.com/scalameta/metals/blob/master/metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala#L78).
+[`metals-debug-session-start`](https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala#L78).
 Those commands should get executed automatically by the lsp client once the user
 activates a code lens. The difference between them is that the former ignores
 all breakpoints being set while the latter respects them. The procedure of
 starting the run/debug session is as follows:
 
 Then we can request the debug adapter uri from the metals server using the
-[`debug-adapter-start`](https://github.com/scalameta/metals/blob/master/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala#L95)
```
kpbochenek, line 27: this line is no longer valid as content was shifted by adding `CleanCompile` command. I know it is completely unrelated to this PR but just noticed it :)

ckipp01:
> `.travis.yml` needs change too

Ah, good call. I missed that one. I always forget about Travis. Some time in the future I wouldn't be opposed to only using Actions and getting rid of Travis all together 😏

docs/editors/debug-adapter-protocol.md
ckipp01: ah, good catch. I'm actually planning on spending a bit of time today reading through the docs since there are a few things I noticed weren't small mistakes. I'll include this change in there.

tgodzik:
The changes live on the main branch now.

PR #1883, opened 2020-07-02
Add `--version` command line option

avdv:
I am using metals for emacs on a few systems. And more often than not, I am not sure which version is currently installed, or whether I should run `coursier bootstrap ...` to generate a new script.

This PR adds a `--version` command line option, which can also be called as `-v` or `-version` showing the current version of metals and additionally the supported Scala versions.

### DIFF
--- a/metals/src/main/scala/scala/meta/metals/Main.scala
+++ b/metals/src/main/scala/scala/meta/metals/Main.scala
@@ -6,6 +6,7 @@ import java.util.concurrent.Executors
 import scala.concurrent.ExecutionContext
 import scala.util.control.NonFatal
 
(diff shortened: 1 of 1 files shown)


kpbochenek:
Thanks for contribution! looks good, could you only run scalafmt as it failed on CI?

```
error: --test failed

To fix this problem:
1. run ./bin/scalafmt from the project root directory
2. `git add . && git commit --amend` OR `git commit -am "Run scalafmt"`
```

avdv:
> Thanks for contribution! looks good, could you only run scalafmt as it failed on CI?

Done.

PR #1892, opened 2020-07-04
Update Vim docs and fix various typos

ckipp01:


docs/build-tools/gradle.md
```diff
@@ -34,20 +30,16 @@ buildscript {
...
 
-Now we can run `gradle bloopInstall`, which will create a Bloop configuration
-files.
+Now we can run `gradle bloopInstall`, which will create all of the Bloop
+configuration files.
 
 This will enable us to work with Metals and all features should work.
-
-You can also always try to reach us on the
-[Metals gitter channel](https://gitter.im/scalameta/metals) in case of any
```
tgodzik, line 52: Why remove it?

ckipp01: Mainly for consistency. We don't have it on any of the other build tool pages, so we should either have it on all or none. Plus, this is also in the footer.

tgodzik: Ok, it's fine with me then!

PR #1893, opened 2020-07-06
Document -version flag and log clientType

kpbochenek:
Document -version flag and log clientType

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -253,8 +253,12 @@ class MetalsLanguageServer(
   private def updateWorkspaceDirectory(params: InitializeParams): Unit = {
     workspace = AbsolutePath(Paths.get(URI.create(params.getRootUri))).dealias
     MetalsLogger.setupLspLogger(workspace, redirectSystemOut)
+    val clientInfo = MetalsServerConfig.metalsClientType match {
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -253,8 +253,9 @@ class MetalsLanguageServer(
   private def updateWorkspaceDirectory(params: InitializeParams): Unit = {
     workspace = AbsolutePath(Paths.get(URI.create(params.getRootUri))).dealias
     MetalsLogger.setupLspLogger(workspace, redirectSystemOut)
+    val clientType = MetalsServerConfig.metalsClientType
     scribe.info(
-      s"started: Metals version ${BuildInfo.metalsVersion} in workspace '$workspace'"
+      s"started: Metals version ${BuildInfo.metalsVersion} in workspace '$workspace' for client $clientType"
```
ckipp01, line 258: So not really a blocker per se, but related to what we've been talking about, this doesn't need to be set, and it won't be in all cases. For example when I'm using `nivm-metals` I don't set this at all, so I my `metalsClientType` won't be set so it'll just show `for client default`. I'm not opposed to it, just unsure how useful that is or if that will be confusing for a user to see.

kpbochenek: changed clientType to `Option` and `for client ...` is only added if value was set.
That should make more sense.

PR #1897, opened 2020-07-08
Reset notifications popup

kpbochenek:
Few questions:
- command in html will be only supported by vscode, does any other editor uses html? If so how to show that only for VSCode/editors supporting commands in html?
- I used `showMessageRequest` but would it be better to use quickpick?

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
@@ -13,6 +13,7 @@ import scala.meta.internal.semver.SemVer
 import scala.meta.io.AbsolutePath
 
 import org.eclipse.lsp4j.ExecuteCommandParams
+import java.net.URLEncoder
 
 /**
  * Helps the user figure out what is mis-configured in the build through the "Run doctor" command.
@@ -71,6 +72,10 @@ final class Doctor(
       hasProblems.set(false)
       languageClient.showMessage(CheckDoctor.problemsFixed)
     }
+    executeRefreshDoctor()
+  }
+
+  def executeRefreshDoctor(): Unit = {
     executeDoctor(
       ClientCommands.ReloadDoctor,
       server => {
@@ -257,6 +262,22 @@ final class Doctor(
 
   def allTargets(): List[ScalaTarget] = buildTargets.all.toList
 
+  private def selectedBuildToolMessage(): Option[String] = {
+    tables.buildTool.selectedBuildTool().map { value =>
+      s"Your ${value} build definition has been imported "
+    }
+  }
+
+  private def selectedImportBuildMessage(): Option[String] = {
+    import scala.concurrent.duration._
+    tables.dismissedNotifications.ImportChanges.whenExpires().map { expire =>
+      val toWhen =
+        if (expire > 1000.days.toMillis) "forever"
+        else "temporairly"
+      s"You dismissed ${toWhen} to re-import your build on configuration changes popup "
+    }
+  }
+
   private def buildTargetsHtml(): String = {
     new HtmlBuilder()
       .element("h1")(_.text(doctorTitle))
@@ -266,11 +287,12 @@ final class Doctor(
 
   private def buildTargetsJson(): String = {
     val targets = allTargets()
-    val heading = tables.buildTool.selectedBuildTool() match {
-      case Some(value) =>
-        doctorHeading + s"\n\nYour ${value} build definition has been imported."
-      case None => doctorHeading
-    }
+    val buildToolHeading = selectedBuildToolMessage()
+    val importBuildHeading = selectedImportBuildMessage()
+
+    val heading =
+      List(buildToolHeading, importBuildHeading, Some(doctorHeading)).flatten
+        .mkString("\n\n")
 
     val results = if (targets.isEmpty) {
       DoctorResults(
@@ -295,19 +317,34 @@ final class Doctor(
     ujson.write(results)
   }
 
+  private def resetChoiceCommand(choice: String): String = {
+    val param = s"""["$choice"]"""
+    s"command:metals.reset-choice?${URLEncoder.encode(param)}"
+  }
+
   private def buildTargetsTable(html: HtmlBuilder): Unit = {
+    selectedBuildToolMessage().foreach { msg =>
+      html.element("p")(
+        _.text(msg)
+          .text("(")
+          .link(resetChoiceCommand(PopupChoiceReset.BuildTool), "Reset")
+          .text(")")
+      )
+    }
+
+    selectedImportBuildMessage().foreach { msg =>
+      html.element("p")(
+        _.text(msg)
+          .text("(")
+          .link(resetChoiceCommand(PopupChoiceReset.BuildImport), "Reset")
+          .text(")")
+      )
+    }
+
     html
       .element("p")(
         _.text(doctorHeading)
       )
-
-    tables.buildTool.selectedBuildTool() match {
-      case Some(value) =>
-        html.element("p")(
-          _.text(s"Your ${value} build definition has been imported.")
-        )
-      case None => ()
-    }
 
     val targets = allTargets()
     if (targets.isEmpty) {
--- a/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
@@ -1,5 +1,6 @@
 package scala.meta.internal.metals
 
+import java.net.URLEncoder
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
 
@@ -13,7 +14,6 @@ import scala.meta.internal.semver.SemVer
 import scala.meta.io.AbsolutePath
 
 import org.eclipse.lsp4j.ExecuteCommandParams
-import java.net.URLEncoder
 
 /**
  * Helps the user figure out what is mis-configured in the build through the "Run doctor" command.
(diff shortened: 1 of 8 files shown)


kpbochenek:
2020-07-08-102946_847x256_scrot

kpbochenek:
choice-reset

metals/src/main/scala/scala/meta/internal/metals/PopupChoiceReset.scala
```diff
@@ -0,0 +1,66 @@
...
+        } else {
+          Future.successful(())
+        }
+      }
+  }
+}
+
+object PopupChoiceReset {
+  final val BuildTool = "buildtool"
+  final val BuildImport = "buildimport"
```
tgodzik, line 65:
```suggestion
  final val BuildImport = "build import"
```

metals/src/main/scala/scala/meta/internal/metals/PopupChoiceReset.scala
```diff
@@ -0,0 +1,66 @@
...
+          reset(PopupChoiceReset.BuildImport)
+        } else {
+          Future.successful(())
+        }
+      }
+  }
+}
+
+object PopupChoiceReset {
+  final val BuildTool = "buildtool"
```
tgodzik, line 64:
```suggestion
  final val BuildTool = "build tool selection"
```

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1416,6 +1424,29 @@ class MetalsLanguageServer(
...
+              case Some(argObject: JsonPrimitive) =>
+                val arg = argObject.asInstanceOf[JsonPrimitive]
+                val value = arg.getAsString()
+                scribe.info(
+                  s"Executing ResetChoicePopup ${command} for choice ${value}"
+                )
+                popupChoiceReset.reset(value)
+
+              case _ =>
+                scribe.info(
```
tgodzik, line 1441:
```suggestion
                scribe.debug(
```

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1416,6 +1424,29 @@ class MetalsLanguageServer(
...
 
+      case ServerCommands.ResetChoicePopup() =>
+        val argsMaybe = Option(params.getArguments())
+        (argsMaybe
+          .map { args =>
+            args.asScala.headOption match {
+              case Some(argObject: JsonPrimitive) =>
+                val arg = argObject.asInstanceOf[JsonPrimitive]
+                val value = arg.getAsString()
+                scribe.info(
```
tgodzik, line 1435:
```suggestion
                scribe.debug(
```
no need to always write it.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -502,6 +503,13 @@ class MetalsLanguageServer(
       tables,
       clientConfig
     )
+    popupChoiceReset = new PopupChoiceReset(
```
tgodzik, line 506: Maybe include it in the Doctor?  I feel like it's part of the Doctor's functionality.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -257,6 +262,22 @@ final class Doctor(
...
 
+  private def selectedBuildToolMessage(): Option[String] = {
+    tables.buildTool.selectedBuildTool().map { value =>
+      s"Your ${value} build definition has been imported "
+    }
+  }
+
+  private def selectedImportBuildMessage(): Option[String] = {
+    import scala.concurrent.duration._
+    tables.dismissedNotifications.ImportChanges.whenExpires().map { expire =>
```
ckipp01, line 273:
```suggestion
    tables.dismissedNotifications.ImportChanges.whenExpires().map { expiration =>
```
Nit on wording

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -257,6 +262,22 @@ final class Doctor(
 
   def allTargets(): List[ScalaTarget] = buildTargets.all.toList
 
+  private def selectedBuildToolMessage(): Option[String] = {
+    tables.buildTool.selectedBuildTool().map { value =>
+      s"Your ${value} build definition has been imported "
```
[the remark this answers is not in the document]
kpbochenek: changed :+1:

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -257,6 +262,22 @@ final class Doctor(
...
+    }
+  }
+
+  private def selectedImportBuildMessage(): Option[String] = {
+    import scala.concurrent.duration._
+    tables.dismissedNotifications.ImportChanges.whenExpires().map { expire =>
+      val toWhen =
+        if (expire > 1000.days.toMillis) "forever"
+        else "temporairly"
```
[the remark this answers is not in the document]
kpbochenek: :sweat_smile:

metals/src/main/scala/scala/meta/internal/metals/PopupChoiceReset.scala
kpbochenek: :+1:

metals/src/main/scala/scala/meta/internal/metals/PopupChoiceReset.scala
kpbochenek: :+1:

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
kpbochenek: :+1:

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
kpbochenek: :+1:

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
kpbochenek: I always prefer smaller, single responsibility classes but I have no strong preference here, if you want I can move this logic to Doctor :+1:

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
kpbochenek: Thanks for testing it and catching this problem!
I will not display this reset link in case `isHttpEnabled=true` as it will be simplest for now.

Maybe in that case add additional line like `To reset those choices execute command: metals.choice-reset` but would users understand that? As probably commands in LSP are mapped differently in different editors.

ckipp01: Yea it's either have to be a very vague message about sending the command, or just don't mention it. Then we'll just have to make sure it's documented well in the docs.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -257,6 +262,22 @@ final class Doctor(
...
+      s"You dismissed ${toWhen} to re-import your build on configuration changes popup "
```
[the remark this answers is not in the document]
kpbochenek: It generates 2 situations:
- 1 minutes (I can add check `if expiration.minutes == 1 then '1 minute' else 'x minutes'`)
- 0 minutes (unsure what to do here?)

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
kpbochenek: Or maybe say something more general like: `You can reset your choices with command 'Reset Choice'` and assume users of different editors will be able to map it to their editor command?

metals/src/main/scala/scala/meta/internal/metals/PopupChoiceReset.scala
kpbochenek: Because it goes into html "build tool selection" is placed in a link as "build+tool+selection".
That's why I also added: `val value = arg.getAsString().replace("+", " ")`
This way we handle both inputs: `build+tool+selection` and `build tool selection` correctly. (same applies to import build)

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
kpbochenek: Ok I am not adding any additional message, we can add it into docs :+1:

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: It's fine, but at some point we need a proper solution for setting up all the classes better.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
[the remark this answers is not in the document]
tgodzik: We can leave as is.

kpbochenek:
@ckipp01 sorry to bother you (I need to finally learn vim/metals) could you double check it looks okay on vim side and reset-choice command works ok?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1416,6 +1424,29 @@ class MetalsLanguageServer(
...
+                popupChoiceReset.reset(value)
+
+              case _ =>
+                scribe.debug(
+                  s"Executing ResetChoicePopup ${command} in interactive mode."
+                )
+                popupChoiceReset.interactiveReset()
+            }
+          })
+          .getOrElse(Future.successful(()))
```
ckipp01, line 1447: So not a big deal, but what do you think about instead of just returning a `Future.successful()` here if there are no arguments, to instead do the `interactiveReset()`. Main reason being right now for clients that won't ever send in params, they have to register the command separately because it _has_ to be registerd with the arguments field of an empty array. So for `coc-metals` for example, that means I can't just add a one-liner here:
https://github.com/scalameta/coc-metals/blob/630450a7cd1e777ddf1392e672838de2e2965b7c/src/index.ts#L248-L260

and instead I'd need to do this:

```javascript
    registerCommand("metals.reset-choice-interactive", () => {
      client.sendRequest(ExecuteCommandRequest.type, {
        command: "reset-choice",
        arguments: [],
      });
    });
```
Again, it's a nitpick, but for me having to set arguments to an empty array all the time doesn't make sense. So it might be nice that if the server gets the `reset-choice` request with no params, it still works.

ckipp01:
And of course, a little eye 🍬 .

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
kpbochenek: Definitely worth changing it as you suggested :+1:

kpbochenek: changed

kpbochenek:
@tgodzik @ckipp01 thanks for review and testing!

Issue #1895, opened 2020-07-07
Towards no longer needing to set the `metals.client` property


ckipp01:
This has sort of been an ongoing conversation offline and also via pull
requests. Metals used to be fully configured via server properties. However, some
of this started to shift to `clientExperimentalCapabilities` in
https://github.com/scalameta/metals/pull/1414, and then further shifted to be
more focused on `initializationOptions` in
https://github.com/scalameta/metals/pull/1626 and
https://github.com/scalameta/metals/pull/1684.  This further aligns with the LSP
spec and also makes bootstrapping the server easier. The configuration of the
server now belongs to the client. With nvim-metals as a test, I was able to
fully configure a client without needing to set the `metals.client` property.
Moving forward, I think this is a good goal to strive for no matter the client.
In order to make this happen there is a few things I'd like to work on that I've
split up into 3 stages.

## Stage 1 - Migrate existing properties to `initializationOptions` 
 
Below are all of the properties that are not either part of the
`initializationOptions` or `clientExperimentalCapabilities`.

The following properties are ones that we would need the client to be able to
set in some way. I think they make sense to migrate over, and it won't be a
ton of work.

- `globSyntax`
- ~`isExitOnShutdown`~ this one was already in there and I missed it 😬 
- `renameFileThreshold`
- `icons`
- `signature-help.command` (This is in compilers, so at first thought I thought this was part of the `InitializationOptions.compilerOptions`, but it's actually not. So I'll add this in there.
- `completion.command` (Same as above)
- `override-def-format`

These are ones that exist currently as properties and should remain only to be
set as properties since they aren't "normally" set and probably shouldn't be set
unless you _really_ want to do a deep dive into Metals or are an advanced user.
I don't think it makes sense to have the client set these.

- `isVerbose`
- `isAutoServer`
- `remoteTimeout`
- `askToReconnect`
- `statistics`
- `allowMultilineStringFormatting`
- `pc.debug`

The following two aren't used anywhere in the code base that I could see. I
propose that we remove them.

- `isWarningsEnabled`
- `bloopEmbeddedVersion`

~The final one which I'm unsure what to do with is below. I personally don't feel
like this should be set as a property or really set at all by the client. The
only place we are using this is to set the `serverCapabilities`. I don't think
it really makes sense for a client to affect what the server says it supports. I
don't know the full story behind this setting, but I think it goes a bit against
the grain of how the `initialize` process should work.~

- ~`allowMultilineStringFormatting`~

[...]

PR #1898, opened 2020-07-09
Remove gson runtime reflection with manual reads for configuration.

ckipp01:
While this is a bit more verbose than what we had, it allows this to be
much more explicit and represent what we actually want to do. When this
was first done (https://github.com/scalameta/metals/pull/1674) I had a
bunch of issues with `null` being set for values that didn't exist, not
easily being able to use `Option` in the structure, and also needing
default constructors in order for it to work correct. I didn't love it,
and have wanted to revisit this for a while. I also figured this was a
good idea to take care of as a precursor to
https://github.com/scalameta/metals/issues/1895

By doing it the way we previously did it caused us to have to check
configuration like this:

 1. Is `InitializationOptions` set to a non-default value?
 2. Is `ClientExperimentalCapabilities` set to a non-default value?
 3. If so, take their value or use the default set in the initial
    config.

Instead, all we really wanted to know was really:

  1. Did `InitializationOptions` set anything?
  2. Did `ClientExperimentalCapabilities` set anything?
  3. If neither, _then_ take the default

Again, this is a bit more verbose but I think ultimately more clearer
and closer to what we actually want.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
@@ -24,12 +24,8 @@ final case class ClientExperimentalCapabilities(
     treeViewProvider: Option[Boolean]
 ) {
   def doctorFormatIsJson: Boolean = doctorProvider.exists(_ == "json")
-  def statusBarIsOn: Boolean = statusBarProvider.exists(_ == "on")
-  def statusBarIsOff: Boolean = statusBarProvider.exists(_ == "off")
-  def statusBarIsShowMessage: Boolean =
-    statusBarProvider.exists(_ == "show-message")
-  def statusBarIsLogMessage: Boolean =
-    statusBarProvider.exists(_ == "log-message")
+  def statusBarState: Option[StatusBarState.StatusBarState] =
+    statusBarProvider.flatMap(StatusBarState.fromString)
 }
 
 object ClientExperimentalCapabilities {
--- a/metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
@@ -1,58 +1,99 @@
 package scala.meta.internal.metals
-import com.google.gson.JsonElement
-import com.google.gson.JsonNull
+import com.google.gson.JsonObject
 import org.eclipse.{lsp4j => l}
 
+/**
+ * While all these values can be set here, they are here only for
+ * compatablity with clients that were setting them this way. From
+ * a client perspective it's preferable and recommended to use
+ * InitializationOptions instead. From a development perspective
+ * don't add something here unless it's truly a more "experimental"
+ * type feature.
+ */
 final case class ClientExperimentalCapabilities(
-    debuggingProvider: Boolean,
-    decorationProvider: Boolean,
-    didFocusProvider: Boolean,
-    doctorProvider: String,
-    executeClientCommandProvider: Boolean,
-    inputBoxProvider: Boolean,
-    openFilesOnRenameProvider: Boolean,
-    quickPickProvider: Boolean,
-    slowTaskProvider: Boolean,
-    statusBarProvider: String,
-    treeViewProvider: Boolean
+    debuggingProvider: Option[Boolean],
+    decorationProvider: Option[Boolean],
+    didFocusProvider: Option[Boolean],
+    doctorProvider: Option[String],
+    executeClientCommandProvider: Option[Boolean],
+    inputBoxProvider: Option[Boolean],
+    openFilesOnRenameProvider: Option[Boolean],
+    quickPickProvider: Option[Boolean],
+    slowTaskProvider: Option[Boolean],
+    statusBarProvider: Option[String],
+    treeViewProvider: Option[Boolean]
 ) {
-  def this() =
-    this(
-      debuggingProvider = false,
-      decorationProvider = false,
-      didFocusProvider = false,
-      doctorProvider = "html",
-      executeClientCommandProvider = false,
-      inputBoxProvider = false,
-      openFilesOnRenameProvider = false,
-      quickPickProvider = false,
-      slowTaskProvider = false,
-      statusBarProvider = "off",
-      treeViewProvider = false
-    )
-  def doctorFormatIsJson: Boolean = doctorProvider == "json"
-  def statusBarIsOn: Boolean = statusBarProvider == "on"
-  def statusBarIsOff: Boolean = statusBarProvider == "off"
-  def statusBarIsShowMessage: Boolean = statusBarProvider == "show-message"
-  def statusBarIsLogMessage: Boolean = statusBarProvider == "log-message"
+  def doctorFormatIsJson: Boolean = doctorProvider.exists(_ == "json")
+  def statusBarIsOn: Boolean = statusBarProvider.exists(_ == "on")
+  def statusBarIsOff: Boolean = statusBarProvider.exists(_ == "off")
+  def statusBarIsShowMessage: Boolean =
+    statusBarProvider.exists(_ == "show-message")
+  def statusBarIsLogMessage: Boolean =
+    statusBarProvider.exists(_ == "log-message")
 }
 
 object ClientExperimentalCapabilities {
-  val Default = new ClientExperimentalCapabilities()
+  import scala.meta.internal.metals.JsonParser._
+  val Default: ClientExperimentalCapabilities = ClientExperimentalCapabilities(
+    None,
+    None,
+    None,
+    None,
+    None,
+    None,
+    None,
+    None,
+    None,
+    None,
+    None
+  )
 
   def from(
       capabilities: l.ClientCapabilities
   ): ClientExperimentalCapabilities = {
-    import scala.meta.internal.metals.JsonParser._
     capabilities.getExperimental match {
-      // NOTE: (ckipp01) For some reason when some editors (emacs) leave out the
-      // `InitializationOptions` key, it is parsed as a `JsonNull` while others
-      // that leave it out it's parsed as a normal `null`. Why? I have no idea.
-      case _: JsonNull => Default
-      case json: JsonElement =>
-        json.as[ClientExperimentalCapabilities].getOrElse(Default)
+      case json: JsonObject =>
+        // Note (ckipp01) We used to do this with reflection going from the JsonElement straight to
+        // InitializationOptions, but there was a lot of issues with it such as setting
+        // nulls for unset values, needing a default constructor, and not being able to
+        // work as expected with Options. This is a bit more verbose, but it gives us full
+        // control over how the InitializationOptions are created.
+        extractToClientExperimentalCapabilities(json)
       case _ =>
         Default
     }
+  }
+
+  def extractToClientExperimentalCapabilities(
+      json: JsonObject
+  ): ClientExperimentalCapabilities = {
+    val jsonObj = json.toJsonObject
+    val debuggingProvider = jsonObj.getBooleanOption("debuggingProvider")
+    val decorationProvider =
+      jsonObj.getBooleanOption("decorationProvider")
+    val didFocusProvider = jsonObj.getBooleanOption("didFocusProvider")
+    val doctorProvider = jsonObj.getStringOption("doctorProvider")
+    val executeClientCommandProvider =
+      jsonObj.getBooleanOption("executeClientCommandProvider")
+    val inputBoxProvider = jsonObj.getBooleanOption("inputBoxProvider")
+    val openFilesOnRenameProvider =
+      jsonObj.getBooleanOption("openFilesOnRenameProvider")
+    val quickPickProvider = jsonObj.getBooleanOption("quickPickProvider")
+    val slowTaskProvider = jsonObj.getBooleanOption("slowTaskProvider")
+    val statusBarProvider = jsonObj.getStringOption("statusBarProvider")
+    val treeViewProvider = jsonObj.getBooleanOption("treeViewProvider")
+    ClientExperimentalCapabilities(
+      debuggingProvider,
+      decorationProvider,
+      didFocusProvider,
+      doctorProvider,
+      executeClientCommandProvider,
+      inputBoxProvider,
+      openFilesOnRenameProvider,
+      quickPickProvider,
+      slowTaskProvider,
+      statusBarProvider,
+      treeViewProvider
+    )
   }
 }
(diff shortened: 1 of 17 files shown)


tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -393,16 +393,33 @@ final class TestingServer(
...
+    val existingInitOptions =
+      initOptions.getClass.getDeclaredFields
+        .foldLeft(Map.empty[String, Any]) { (acc, field) =>
+          field.setAccessible(true)
+          acc + (field.getName -> field.get(initOptions))
+        }
+        .collect {
+          case (key, Some(value)) => key -> value
+        }
+        .asJava
```
ckipp01, line 416: So I'm open to some feedback here about how we should do this. The tests starting failing after the changes due to us just calling `.toJson` here, which would convert our `Option`s to `{}`. That would cause the `InitializationOptions` to just be default not matter what. I thought about writing something for gson that would turn an Option into the value it holds, but then for `None` we are stuck with null. Since most clients aren't just going to send a valid key with a `null` field for this, I wanted to actually mimic how they'd do it by _only_ sending in the keys that they would want to set. So in the end, this does just that and sends in an object only with the desired keys set. I think this is actually good for testing as it more resembles what happens in real life, but then I'm stuck with this gross thing because of how we build up the `InitializtionOptions` in the test and then convert them to json. Thoughts?

tests/slow/src/test/scala/tests/feature/DefinitionCrossLspSuite.scala
```diff
@@ -77,7 +77,7 @@ class DefinitionCrossLspSuite
       _ <- server.didOpen("scala/Predef.scala")
       _ = assertNoDiff(
         client.workspaceMessageRequests,
-        ""
+        "Preparing presentation compiler"
```
ckipp01, line 80: Unless I'm mistaken, this should have been happening all the time, but may not have been due to some misconfiguration. Mainly because looking at this, if this runs without the current compiler version, then it uses the embedded presentation compile, which should be happening here, and that's where we see the "Preparing presentation compiler". This was failing without this, and that makes sense right?

ckipp01:
I'm aware of the valid test failures. Chasing down why this change is causing an issue with the `FormattingLspSuite`.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -393,16 +393,35 @@ final class TestingServer(
...
+          field.setAccessible(true)
+          field.getName -> field.get(initOptions)
+        }
+        .collect {
+          case (key, Some(value)) => key -> value
+        }
+        .toMap
+        .asJava
+
+    params.setInitializationOptions(existingInitOptions.toJson)
```
ckipp01, line 419: We must have missed this when we migrated to `InitilizationOptions` because we were actually still passing the values for that into the `ClientCapabilities` experimental param. Now we just pass in an empty object there and do the values here.

ckipp01: I keep finding small fixes, but no the one that fixes my issue 😆

ckipp01:
Alright, finally got the tests all passing. So changing these actually makes things a bit stricter in tests with what we are setting need to be set etc. So there were a few things places that I needed to add a `InitializationOption` etc. This should be ready to be reviewed now.

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
```diff
@@ -1,58 +1,87 @@
...
-      statusBarProvider = "off",
-      treeViewProvider = false
-    )
-  def doctorFormatIsJson: Boolean = doctorProvider == "json"
-  def statusBarIsOn: Boolean = statusBarProvider == "on"
-  def statusBarIsOff: Boolean = statusBarProvider == "off"
-  def statusBarIsShowMessage: Boolean = statusBarProvider == "show-message"
-  def statusBarIsLogMessage: Boolean = statusBarProvider == "log-message"
+  def doctorFormatIsJson: Boolean = doctorProvider.exists(_ == "json")
+  def statusBarIsOn: Boolean = statusBarProvider.exists(_ == "on")
```
kpbochenek, line 27: if we had 'Option[Boolean]' here we could use `extract` in `ClientConfiguration.statusBarIsOn`

because right now `statusBarIsShowMessage == true` AND `statusBarIsLogMessage == true` is possible if you set one in experimental capabilities and second in initialization options
:thinking:

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
```diff
@@ -1,58 +1,87 @@
...
+  def statusBarIsOff: Boolean = statusBarProvider.exists(_ == "off")
```
kpbochenek, line 28: we are doing && in ClientConfiguration for this configuration but || for all others. does that mean it must be set in all 3 places to be off?

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
```diff
@@ -1,58 +1,87 @@
...
+    debuggingProvider: Option[Boolean],
+    decorationProvider: Option[Boolean],
+    didFocusProvider: Option[Boolean],
+    doctorProvider: Option[String],
+    executeClientCommandProvider: Option[Boolean],
+    inputBoxProvider: Option[Boolean],
+    openFilesOnRenameProvider: Option[Boolean],
+    quickPickProvider: Option[Boolean],
+    slowTaskProvider: Option[Boolean],
+    statusBarProvider: Option[String],
```
kpbochenek, line 23: https://github.com/scalameta/metals/blob/48b20201f3357738ce472c40dcae47b0cccc56fb/docs/editors/new-editor.md

Here we say `statusBarProvider` is Boolean :thinking:

tests/unit/src/test/scala/tests/NewProjectLspSuite.scala
```diff
@@ -180,7 +180,7 @@ class NewProjectLspSuite extends BaseLspSuite("new-project") {
         if (isSelectProject(params)) {
           onSelectProject(findAction)
         } else if (isOpenWindow(params)) {
-          findAction("yes")
```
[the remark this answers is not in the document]
ckipp01: This was my exact question! I have no idea how this working before.

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
ckipp01: > if we had `Option[Boolean]` here we could use extract in ClientConfiguration.statusBarIsOn

Well the tricky thing is that it's not a binary decision right? It can be on, off, show message, or log message. So that can't really be represented this way right? So I'm unsure how to do that.

> because right now statusBarIsShowMessage == true AND statusBarIsLogMessage == true is possible if you set one in experimental capabilities and second in initialization options

Sort of? I guess a client could set `statusBarProvider = "on"` in `InitializationOptions` and then `statusBarProvider = "log-message"` in `ClientExperimentalCapabilities`, but I _believe_ all of the checks would check of `statusBarIsOn` all of the time, which would always cause that one to win. Ideally, we are allowing illegal states to be represented by allowing for them to be set in both places. I sort of wish that we could remove the `ClientExperimentalCapabilities` since they can all be set via `InitializationOptions`. Is that allowed? If you think it's fine to potentially break a client that is being set that way, then I'm totally ok with removing them. I think if we are explicit and up front about this happening, then maybe we can? In reality, these are marked "experimental", which implies that they won't always be here.

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
ckipp01: Ah, good catch. This is left over from how the logic used to be. Basically "off" was the default, so the reason this was `&&` for this was to ensure that none of them were changed before. Now this can actually be set to `||` since `ClientExperimentalCapabilities` and `InitializationOptions` are no longer defaulting to "off", but rather `None`. I'll change this.

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
ckipp01: 🤔 that's a mistake in the docs then. Good catch. I'll change this there.

tests/unit/src/test/scala/tests/NewProjectLspSuite.scala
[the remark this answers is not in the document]
kpbochenek: it must have never hit this if branch :thinking:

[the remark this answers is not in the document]
ckipp01: I suppose yea.

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
ckipp01: Updated this.

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
ckipp01: Alright, thanks for explaining this. After you showed me I got what you meant 😆 These changes are now in this branch.

metals/src/main/scala/scala/meta/internal/metals/ClientExperimentalCapabilities.scala
ckipp01: Changed.

metals/src/main/scala/scala/meta/internal/metals/StatusBarState.scala
```diff
@@ -0,0 +1,19 @@
...
+  sealed trait StatusBarState
+  case object On extends StatusBarState
+  case object Off extends StatusBarState
+  case object ShowMessage extends StatusBarState
+  case object LogMessage extends StatusBarState
+
+  def fromString(value: String): Option[StatusBarState] =
+    value match {
+      case "on" => Some(On)
+      case "Off" => Some(Off)
```
kpbochenek, line 13: ahh sorry I typed it instead of copy-paste it, it should be "off" and I capitalize it wrongly :facepalm:

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -393,16 +393,35 @@ final class TestingServer(
...
+        slowTaskProvider = Some(true)
       )
     )
+
+    // Yes, this is a bit gross :/
+    // However, I want to only get the existing fields that are being set
+    // much like it'd be when a client actually sends this. This will just
+    // collect the fields that are set, get the values, and then make them into
+    // a map that will become a JsonObject to pass in as the InitializationOptions
+    val existingInitOptions =
```
tgodzik, line 407: I would just add a simple json library for tests that takes into account options for tests, would be much nicer.

tests/unit/src/test/scala/tests/NewProjectLspSuite.scala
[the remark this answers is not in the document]
tgodzik: My bad, it must have somehow only worked with new window provider turned off.

metals/src/main/scala/scala/meta/internal/metals/StatusBarState.scala
ckipp01: gosh, I didn't even notice this. Thanks for catching it! fixed!

mtags/src/main/scala/scala/meta/internal/pc/PresentationCompilerConfigImpl.scala
```diff
@@ -34,20 +34,33 @@ case class PresentationCompilerConfigImpl(
   override def completionCommand: Optional[String] =
     Optional.ofNullable(_completionCommand.orNull)
 
+  /**
+   * Used to update the compiler config after we recieve the InitializationOptions.
+   * If the user sets a value in the InitializationOptions, then the value will be
+   * captures and set here. If not, we just resort back tot he default that already
```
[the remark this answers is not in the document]
ckipp01: Fixed.

mtags/src/main/scala/scala/meta/internal/pc/PresentationCompilerConfigImpl.scala
```diff
@@ -34,20 +34,33 @@ case class PresentationCompilerConfigImpl(
...
+   * exists.
+   */
   def update(
       options: CompilerInitializationOptions
   ): PresentationCompilerConfigImpl =
     copy(
-      isCompletionItemDetailEnabled =
-        isCompletionItemDetailEnabled || options.isCompletionItemDetailEnabled,
+      isCompletionItemDetailEnabled = options.isCompletionItemDetailEnabled
+        .fold(this.isCompletionItemDetailEnabled)(identity),
```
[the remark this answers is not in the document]
ckipp01: Nope, I actually had something different first that was more involved than this, and then simplified, and then didn't even realize that what I had was just a `getOrElse`. I'll update this.

tests/unit/src/main/scala/tests/TestingServer.scala
ckipp01: Could you expand on what you mean by this?

tests/slow/src/test/scala/tests/feature/DefinitionCrossLspSuite.scala
ckipp01: Totally. Let me look a bit closer at this.

tests/unit/src/main/scala/tests/TestingServer.scala
tgodzik: Ach, that was totally not cleat. I just meant adding something like upickle. Automatically transforming case class into a string and then using gson to parse that.

ckipp01: So it looks like in order for uPickle to handle this correctly I would need to define a serializer into the companion object of `InitializationOptions`, but we don't include uPickle. I'm hesitant to add a library just to do this. What do you think?

tgodzik: I haven't noticed that, I was thinking of only adding a library to the test project. This way we wouldn't be adding additional dependencies to the project itself.

We can leave it as is, it was just an idea.

ckipp01:
Something still doesn't feel right. I'm still getting spurious errors in CI that is happening often enough that I'm convinced I'm missing something. I'm going to give this another look through before merging 🤔

ckipp01:
Only this failing now on windows:

```
[error] Failed: Total 1043, Failed 1, Errors 0, Passed 1037, Skipped 5
[error] Failed tests:
[error] 	tests.BillLspSuite
```
But I see it's failing in the [other branches as well](https://github.com/scalameta/metals/actions/runs/165075591) so I'm going to merge.

I also noticed that CI has been taking longer lately 🤔 I may try to look back to see when this started being quite a bit longer to see if I can pin point why this is.

Issue #1845, opened 2020-06-19
False positive for automatic completion suggestions


olafurpg:
**Describe the bug**

Metals auto-completes a suggestion that compiles successfully but has buggy runtime semantics.

**To Reproduce**

Currently, when completing in this case where Metals autocompletes `processFile(isResourceFile = isLargeBanana)`. 

```scala
val isLargeBanana = true

def processFile(isResourceFile: Boolean): Unit = ()

processFile(isResourceFil@@)
```
The `isLargeBanana` variable is unrelated to `isResourceFile` even if it has the same type `Boolean`. I don't think Metals should try to guess on my behalf and insert `isLargeBanana` directly.

**Expected behavior**

I expect Metals to complete only `isResourceFile = `. It should at least provide two different alternatives with one alternative that doesn't auto-fill the name.

**Screenshots**

In this particular example, I was gonna type manually `true`.

<img width="519" alt="Screenshot 2020-06-19 at 17 27 44" src="https://user-images.githubusercontent.com/1408093/85150698-291cbb80-b253-11ea-806b-2aa2a6b6b438.png">

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.9.0

**Additional context**

**Search terms**

tgodzik:
:thinking: This looks super weird. We always add the empty arg as the first completion:

```scala
      params.map(param =>
        new NamedArgMember(param)
      ) ::: findPossibleDefaults() ::: fillAllFields()
```

And I can't seem to reproduce:

![image](https://user-images.githubusercontent.com/3807253/85154972-72233e80-b258-11ea-97b7-1a97c9837aef.png)

Any idea why the default is not included? Does it happen for you on the smaller example?

olafurpg:
I made several failed attempts to minimize the example.

I'm still able to consistently reproduce it in the large codebase, the exact line of code is here https://github.com/scalameta/fastpass/commit/f689ce2a442a5b3ed0510a1f16a7040b7bce17cb

I looked at the LSP trace logs to confirm that Metals is returning only one completions (it's not filtered out by VS Code)

```json
[Trace - 10:47:53 AM] Received request 'textDocument/completion - (13300)'
Params: {
  "context": {
    "triggerKind": 1
  },
  "textDocument": {
    "uri": "file:///Users/lgeirsson/dev/fastpass/fastpass/src/main/scala/scala/meta/internal/fastpass/pantsbuild/BloopPants.scala"
  },
[...]

PR #1901, opened 2020-07-10
Don't filter out named parameters for methods defined later in a class

tgodzik:
Previously, we would treat named parameters for methods defined later in the code as local forward references. Now, we make sure that named parameters are not filtered out in that case.

Fixes https://github.com/scalameta/metals/issues/1845

### DIFF
--- a/tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
@@ -203,6 +203,19 @@ class CompletionArgSuite extends BaseCompletionSuite {
   )
 
   check(
+    "arg14",
+    s"""|object Main {
+        |  val isLargeBanana = true
+        |  processFile(isResourceFil@@)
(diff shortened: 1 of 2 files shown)


tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -202,6 +202,17 @@ class CompletionArgSuite extends BaseCompletionSuite {
...
   )
 
+  check(
+    "arg14",
+    s"""|object Main {
+        |  curry(1)(bana@@)
+        |  def curry(a: Int)(banana: Int): Int = ???
+        |}
+        |""".stripMargin,
+    """|banana = : Int
```
olafurpg, line 212: This test case doesn’t stress the case where there is a conflicting variable in scope, do you think it’s worth adding another “bananaFoo” val above to test that the default completions is still unchanged?

tgodzik: Changed the test case to show the actual situation from the issue.

PR #1902, opened 2020-07-11
Improve gradle-bloop integration

jvican:
Makes some improvements to the gradle-bloop integration. There's more information in the commit messages :smile:


tests/slow/src/test/scala/tests/gradle/GradleLspSuite.scala
```diff
@@ -71,6 +71,66 @@ class GradleLspSuite extends BaseImportSuite("gradle-import") {
     }
   }
 
+  test("basic-configured".only) {
```
kpbochenek, line 74:
```
'Only' tag is not allowed when `isCI=true`
[error] 73:
[error] 74:  test("basic-configured".only) {
```

PR #1906, opened 2020-07-13
small refactoring DoctorFormat

kpbochenek:
- Refactor DoctorFormat in a similar way to StatusBarState (no more '||' in ClientConfiguration)
- use match for DoctorFormat - in future if adding new format it will throw unexhausted match instead of failling into else state
- doctorFormat was unused in MetalsServerConfig - removed


kpbochenek:
unrelated test failure (OOM jvm error)

PR #1907, opened 2020-07-13
Add support for Scala 2.12.12

tgodzik:


.github/workflows/ci.yml
```diff
@@ -99,7 +99,7 @@ jobs:
             name: Scala cross tests
             os: ubuntu-latest
           - type: scalafix
-            command: sbt scalafixCheck docs/docusaurusCreateSite
+            command: sbt ++2.12.11 scalafixCheck docs/docusaurusCreateSite
```
tgodzik, line 102: I will follow up with an issue to update both scalafix and ammonite when the support for both is added for Scala 2.12.12

tgodzik:
> Shouldn't also version be updated in `sbt`, `test-release.sh` and `test-pom.xml`? just grepped where 2.12.11 was used.

Updated! Thanks for reminding me! I worked around other features. Scalafix and ammonite can be updated later. This doesn't influence the actual use of Metals

kpbochenek:
@tgodzik will not ammonite scripts use default scala version which is `2.12.12` now, but it would not be supported by ammonite?

tgodzik:
> @tgodzik will not ammonite scripts use default scala version which is `2.12.12` now, but it would not be supported by ammonite?

They use the default set by ammonite, we don't declare any default from what I've seen. I can double check.

.github/workflows/ci.yml
kpbochenek: :+1:

metals/src/main/scala/scala/meta/internal/metals/ammonite/Ammonite.scala
```diff
@@ -149,7 +149,8 @@ final class Ammonite(
       .getOrElse(
         AmmVersions(
           ammoniteVersion = BuildInfo.ammoniteVersion,
-          scalaVersion = BuildInfo.scala212
+          // TODO: https://github.com/scalameta/metals/issues/1908
+          scalaVersion = "2.12.11"
```
tgodzik, line 153: This is the default version for Ammonite

kpbochenek: :+1:

alexarchambault:
FYI Ammonite `2.1.4-11-307f3d8`, with scala `2.12.12` support, is now on Central.

tgodzik:
> FYI Ammonite `2.1.4-11-307f3d8`, with scala `2.12.12` support, is now on Central.

Thanks! That was really fast! I can now update everything normally :D

PR #1910, opened 2020-07-14
Bump scalafmt version up to 2.6.3.

ckipp01:
Since we're probably releasing today I figured it'd be a good idea to bump the version of scalafmt that the server recommends for people if they don't have it set in their conf file. I also bumped it for the codebase, which only caused one small change.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version = "2.6.1"
+version = "2.6.3"
 project.git = true
 align.preset = none
 align.stripMargin = true
--- a/build.sbt
+++ b/build.sbt
@@ -183,7 +183,7 @@ lazy val V = new {
   val gradleBloop = bloop
   val mavenBloop = bloop
   val mdoc = "2.2.3"
-  val scalafmt = "2.6.1"
+  val scalafmt = "2.6.3"
   val munit = "0.7.9"
   // List of supported Scala versions in SemanticDB. Needs to be manually updated
   // for every SemanticDB upgrade.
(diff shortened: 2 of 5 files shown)


kpbochenek:
maybe a message has changed?
```
ERROR scalafmt: /Users/runner/work/metals/metals/tests/unit/target/e2e/formatting/missing-version/.scalafmt.conf: missing setting 'version'. To fix this problem, add the following line to .scalafmt.conf: 'version=2.6.3'.
```
vs
```
    def messageRequestMessage: String =
      s"No Scalafmt version is configured for this workspace. " +
        s"To fix this problem, update .scalafmt.conf to include 'version=${BuildInfo.scalafmtVersion}'."
```

Maybe we should also provide in all tests `${BuildInfo.scalafmtVersion}` because currently they have hardcoded `1.5.1`

ckipp01:
> maybe a message has changed?
> 
> ```
> ERROR scalafmt: /Users/runner/work/metals/metals/tests/unit/target/e2e/formatting/missing-version/.scalafmt.conf: missing setting 'version'. To fix this problem, add the following line to .scalafmt.conf: 'version=2.6.3'.
> ```
> 
> vs
> 
> ```
>     def messageRequestMessage: String =
>       s"No Scalafmt version is configured for this workspace. " +
>         s"To fix this problem, update .scalafmt.conf to include 'version=${BuildInfo.scalafmtVersion}'."
> ```

I'm actually not sure what's going on. Something changed between 2.6.1 and 2.6.2 that cause all of these to result in a Boxed Error. It's not even making it to the check before it blows up.

> Maybe we should also provide in all tests `${BuildInfo.scalafmtVersion}` because currently they have hardcoded `1.5.1`

Yea I have this changed locally now. When I figure out what's going on above I'll push this up.

ckipp01:
It looks like this change is biting us: https://github.com/scalameta/scalafmt/commit/a7bb6920a8ba486cfbff74d056037dfeb36c7e33

If a user doesn't have something configured correctly and we try to "warm up" scalafmt like we have done up to this point, it now returns an exception here:

https://github.com/scalameta/metals/blob/020b9ac296169aa531ce9f2b226ababa1d36bca1/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala#L69-L79

We may have to rethink how we do this now. I'm actually a bit surprised by this change because it's sort of breaking behavior for the interface.

PR #1912, opened 2020-07-14
Remove remants of fastpass.

ckipp01:
This pr removes a couple things I noticed were still in here that pertain to fastpass. I also removed the `preparePr` alias I added a while ago since it's not that useful.


build.sbt
```diff
@@ -151,17 +151,8 @@ onLoad.in(Global) ~= { old =>
...
-execScalafmt := {
-  "bin/scalafmt" !
-}
-
 addCommandAlias("scalafixAll", "all compile:scalafix test:scalafix")
 addCommandAlias("scalafixCheck", "; scalafix --check ; test:scalafix --check")
-addCommandAlias(
-  "preparePr",
-  "scalafixAll ; execScalafmt"
-)
```
ckipp01, line 164: Not related to fastpass, but I also wanted to remove this. It wasn't a great idea since after running this scalafmt is still ran in the hook. So you needlessy run scalafmt twice. Good idea, bad execution. I still think it'd be nice to have something like this, but we'd need to think about it a bit more.

PR #1930, opened 2020-07-16
Add Bloop jvmopts in tests to limit memory used by it

tgodzik:
This should help address the `Out of memory` issues we've been seeing.


kpbochenek:
`Add Bloop jvmopts to limit memory used by it` -> `Add Bloop jvmopts in tests to limit memory used by it` correct?

tgodzik:
> `Add Bloop jvmopts to limit memory used by it` -> `Add Bloop jvmopts in tests to limit memory used by it` correct?

Yeah, it's basically just for test.

PR #1934, opened 2020-07-20
Add missing docs and move versions to a more prominent location.

tgodzik:
This PR adds the following:
- new Scala project documentation for vim and VS Code
- parent code lenses documentation for vim and VS Code
- fixes a number of smaller issues
- moves versions to a more prominent location

### DIFF
--- a/docs/editors/new-editor.md
+++ b/docs/editors/new-editor.md
@@ -235,7 +235,7 @@ Default value: `false`
 ##### `didFocusProvider`
 
 Boolean value to signify that the client supports the
-[`metals/didFocusTextDocument`](#metalsdid-focus-text-document) extension.
+[`metals/didFocusTextDocument`](#metalsdidfocustextdocument) extension.
 
 Default value: `false`
 
@@ -328,7 +328,7 @@ Default: `false`
 ##### `quickPickProvider`
 
 Boolean value to signify whether or not the client implements the
-[`metals/quickPick`](#metalsquick-pick) extensions.
+[`metals/quickPick`](#metalsquickpick) extensions.
 
 Default value: `false`
 
--- a/docs/editors/overview.md
+++ b/docs/editors/overview.md
@@ -4,6 +4,18 @@ title: Text Editors
 sidebar_label: Overview
 ---
 
+## Latest Metals server versions
+
+To find out how to set the version in your editor please check out the editor
+specific sections.
+
+```scala mdoc:releases
+
+```
+
+
+## Editor support
+
 Metals works with the following text editors with varying degree of
 functionality.
 
@@ -187,8 +199,8 @@ functionality.
     <td>Metals Extensions</td>
     <td align="center">✅</td>
     <td align="center">Status bar</td>
-    <td align="center">Status bar</td>
-    <td align="center">Status bar</td>
+    <td align="center">Status bar, Quick pick</td>
+    <td align="center">Status bar, Input box</td>
     <td align="center">Status bar</td>
     <td align="center">Status bar, Slow task</td>
   </tr>
@@ -372,7 +384,8 @@ using the Metals sidebar. This feature is only implemented in VS Code.
 **Tree view**: Editor client implements the
 [Tree View Protocol](tree-view-protocol.md).
 
-**Did focus**: Editor client implements the `metals/didFocusTextDocument` notification.
+**Did focus**: Editor client implements the `metals/didFocusTextDocument`
+notification.
 
 **Slow task**: Editor client implements the `metals/slowTask` request.
 
--- a/docs/editors/vim.md
+++ b/docs/editors/vim.md
@@ -323,7 +323,7 @@ Then, when on the line that you'd like to expand the decoration to get the hover
 
 coc-metals has a built-in implementation of the [Tree View
 Protocol](https://scalameta.org/metals/docs/editors/tree-view-protocol.html).
-If you have the [recommended mappings](coc-mappings.vim) copied, you'll notice
+If you have the [recommended mappings](vim.md#recommended-cocnvim-mappings) copied, you'll notice
 that in the bottom you'll have some TVP related settings. You can start by
 opening the TVP panel by using the default `<space> t`. Once open, you'll see
 there are three parts to the panel. The first being the `MetalsCompile` window
@@ -409,6 +409,10 @@ If you'd like to get autocompletion help for the configuration values you can
 install [coc-json](https://github.com/neoclide/coc-json).
 
 ```scala mdoc:user-config:lsp-config-coc
+```
+
+```scala mdoc:new-project:vim
+
 ```
 
 ## Enable on type formatting for multiline string formatting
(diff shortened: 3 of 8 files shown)


docs/editors/new-editor.md
```diff
@@ -235,7 +235,7 @@ Default value: `false`
 ##### `didFocusProvider`
 
 Boolean value to signify that the client supports the
-[`metals/didFocusTextDocument`](#metalsdid-focus-text-document) extension.
+[`metals/didFocusTextDocument`](#metalsdidfocustextdocument) extension.
```
tgodzik, line 238: The links here were broken.

docs/editors/overview.md
```diff
@@ -4,6 +4,21 @@ title: Text Editors
 sidebar_label: Overview
 ---
 
+## Latest Metals server versions
```
tgodzik, line 7: I feel like it will be much easier to find this way. Alternatively we could move it to the top of each editor?

docs/editors/vscode.md
```diff
@@ -120,29 +124,36 @@ without opening the sidebar.
...
+![on-type-setting](https://i.imgur.com/s6nT9rC.png)
 
-## Enable formatting on paste for multiline strings
+## Formatting on paste for multiline strings
 
 Whenever text is paste into a multiline string with `|` it will be properly
 formatted by Metals:
 
-![format-on-paste](https://i.imgur.com/yJLAIxQ.gif)
+![format-on-paste](https://i.imgur.com/fF0XWYC.gif)
```
tgodzik, line 151: Added better gifs, because the previous ones were bother me quite a bit :sweat_smile:

docs/editors/overview.md
```diff
@@ -187,7 +202,7 @@ functionality.
     <td>Metals Extensions</td>
     <td align="center">✅</td>
     <td align="center">Status bar</td>
-    <td align="center">Status bar</td>
+    <td align="center">Status bar, Quick pick</td>
     <td align="center">Status bar</td>
```
ayoub-benali, line 206:
```suggestion
    <td align="center">Status bar, Input box</td>
```
Sublime plugin support Input Box now: https://github.com/scalameta/metals-sublime/blob/master/plugin.py#L114

tgodzik: Thanks! I will add it! Should I add any of the documentation to sublime docs? The new project provider or parent code lenses?

ayoub-benali: These two aren't yet supposed yet :/
But "new scala file" was just released: https://github.com/scalameta/metals-sublime/pull/26

docs/editors/overview.md
ckipp01: I think this is fine as long as we don't remove it from the editor page since I feel like some people just skip the overview page and go right to their editor.

docs/editors/overview.md
```diff
@@ -4,6 +4,21 @@ title: Text Editors
 sidebar_label: Overview
 ---
 
+## Latest Metals server versions
+
+All editors use a separate Metals server underneath, which version can be
+changed idependently. It can be used to checkout newest features from the main
+branch.
```
ckipp01, line 11: In theory though, this _could_ change moving forward no? Now that everything can be configured via `InitializationOptions`, there is no need to have a `metals-vscode` and `metals-emacs` but a single `metals` for each desired version.

docs/editors/vim.md
```diff
@@ -323,7 +323,7 @@ Then, when on the line that you'd like to expand the decoration to get the hover
 
 coc-metals has a built-in implementation of the [Tree View
 Protocol](https://scalameta.org/metals/docs/editors/tree-view-protocol.html).
-If you have the [recommended mappings](coc-mappings.vim) copied, you'll notice
+If you have the [recommended mappings](vim.md#recommended-cocnvim-mappings) copied, you'll notice
```
ckipp01, line 326: Just out of curiosity, since it's the same page this doesn't need to be prefaced with `vim.md` right?

docs/editors/vim.md
```diff
@@ -411,6 +411,14 @@ install [coc-json](https://github.com/neoclide/coc-json).
 ```scala mdoc:user-config:lsp-config-coc
 ```
 
+```scala mdoc:parent-lenses:vim
```
ckipp01, line 414: I actually have a whole section on this further down. https://scalameta.org/metals/docs/editors/vim.html#goto-super-method

I'd say remove it and replace it with this, but then my readme and this document becomes quite a bit different. I don't love having them both, but I also don't like not having a detailed readme 🤔 I'm not really sure what would be best for this.

metals-docs/src/main/scala/docs/NewProjectModifier.scala
```diff
@@ -0,0 +1,50 @@
...
+      if (isVscode)
+        "\nThe same command will be invoked when clicking the \"New Scala Project\" button in the Metals view."
+      else ""
+    s"""|## Create new project from template
+        |
+        |It is possible using Metals to easily setup a new project using the exiting [giter8](https://github.com/foundweekends/giter8/wiki/giter8-templates) templates. 
+        |This is an equivalent to the `sbt new` command, which uses the same mechanism.
+        |There is a great number of templates already available and it should be easy to find something for yourself.
+        |To start the setup you can use the $newScalaProject command, which works as following:
+        |1. Choose the templateand then:
```
ckipp01, line 35:
```suggestion
        |1. Choose the template and then:
```

metals-docs/src/main/scala/docs/NewProjectModifier.scala
```diff
@@ -0,0 +1,50 @@
...
+        |    1. Use the proposed templates.
+        |    2. Choose "Discover more" and then choose from the list downloaded from Giter8 wiki page.
```
ckipp01, line 37:
```suggestion
        |    2. Choose "Discover more" and then choose from the list downloaded from the Giter8 wiki page.
```

docs/editors/overview.md
tgodzik: Hmm... right. MAybe then just remove the section.

docs/editors/vim.md
tgodzik: ach, right. I will remove it. I never remember the exact syntax needed, so I experiment a bit.

docs/editors/vim.md
tgodzik: Haven't noticed it. I can just add it to the VS Code section then.

docs/editors/overview.md
tgodzik: Moved the lower paragraph here.

docs/editors/vim.md
tgodzik: Actually, I am getting a warning without it, so I will leave it as is.

docs/editors/vim.md
tgodzik: Removed from vim section.

PR #1903, opened 2020-07-11
Migrate final properties over to `InitializationOptions`.

ckipp01:
This migrates the final remaining server properties over to being
`InitializationOptions`. It should now be fully possibly to configure
the server without the usage of setting the `client` property. The
following properties were migrated.

- `globSyntax` -> `globSyntax`
- `renameFileThreshold` -> `renameFileThreshold`
- `icons` -> `icons`
- `signature-help.command` -> `compilerOptions.parameterHintsCommand`
- `completion.command` -> `compilerOptions.completionCommand`
- `override-def-format` -> compilerOptions.overrideDefFormat`

The following properties were removed as they were no longer used.

- `isWarningsEnabled`
- `bloopEmbeddedVersion`

This is step 1 in #1895 
I'd also like to move around a few things in the new-editor document, but I'm still playing around with how to make it the most clear. However, this should be all of the code changes unless I'm missing something.

PR #1936, opened 2020-07-20
Status bar config "on" by default for sublime

ayoub-benali:
It seems like since: https://github.com/scalameta/metals/pull/1903  `-Dmetals.status-bar=on` is overridden by `-Dmetals.client=sublime` which forces `statusBar = StatusBarConfig.showMessage`.

This a short term fix until https://github.com/scalameta/metals-sublime is migrated into setting all intialization options without client name.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
@@ -171,7 +171,7 @@ object MetalsServerConfig {
       case "sublime" =>
         base.copy(
           isHttpEnabled = true,
-          statusBar = StatusBarConfig.showMessage,
+          statusBar = StatusBarConfig.on,
           icons = Icons.unicode,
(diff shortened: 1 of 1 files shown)


ckipp01:
Ahh, I actually goofed up here.

https://github.com/scalameta/metals/blob/cae9adee59d9e8718fe853be4dcad0cce92de56b/metals/src/main/scala/scala/meta/internal/metals/ClientConfiguration.scala#L26-L31

This should actually be:

```scala
  def statusBarState(): StatusBarState.StatusBarState =
    extract(
      initializationOptions.statusBarState,
      experimentalCapabilities.statusBarState,
      StatusBarState
        .fromString(initialConfig.statusBar.value)
        .getOrElse(StatusBarState.Off)
    )
```
In order to take into account the server property if set. Currently it just ignores it. Are you also able to make this change? You will no longer have to then change the default value for sublime if you want, but it's still fine to if you want.

PR #1938, opened 2020-07-22
Add -Ximport-suggestion-timeout 0 to the presentation compiler options 

tgodzik:
Import suggestions are not used currently and might cause additional overhead for the presentation compiler.

Suggested by @smarter

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -340,6 +340,9 @@ lazy val mtags3 = project
     unmanagedSourceDirectories.in(Compile) += baseDirectory
       .in(ThisBuild)
       .value / "mtags" / "src" / "main" / "scala",
+    managedSourceDirectories.in(Compile) += baseDirectory
+      .in(ThisBuild)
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -340,6 +340,9 @@ lazy val mtags3 = project
     unmanagedSourceDirectories.in(Compile) += baseDirectory
       .in(ThisBuild)
       .value / "mtags" / "src" / "main" / "scala",
+    managedSourceDirectories.in(Compile) += baseDirectory
+      .in(ThisBuild)
+      .value / "mtags" / "target" / ("scala-" + scalaVersion.value) / "src_managed" / "main",
```
ckipp01, line 345: So I see this is added in order to access `BuildInfo` in `ScalaPresentationCompiler.scala`, but why isn't it detected by default here?

tgodzik: It's the `mtags3` module, which I added so that we have the additional target. It uses `mtags3` directory by default, so every source dir from mtags needs to be added manually.

PR #1939, opened 2020-07-23
Update to Scala 3 to 0.25.0 version

tgodzik:


tgodzik:
> https://github.com/lampepfl/dotty/releases/tag/0.26.0-RC1
> 
> It looks like 0.26.0-RC1 was also released. Do we want to also add support for this right away?

Good idea, I haven't noticed it also got released.

PR #1944, opened 2020-07-23
Add in , as a trigger for SignatureHelp.

ckipp01:
This will allow users to enter a parameter or two, step out, and then back
in and still get signature help. Currently, if you leave the parameters
and step back in, you will no longer get signature help. Also some clients
won't update and send another request to the server while you are in the
parameter group and hit ,. Now with this being explicit they will.

Just some examples. In VS Code before:

In VS Code after:

In coc-metals before:

In coc-metals after:

Closes https://github.com/scalameta/coc-metals/issues/264


ckipp01:
Some flaky unrelated test failures.

PR #1945, opened 2020-07-24
Add more documentation about worksheets.

### LINKED ISSUE #1935: Add better documentation for worksheets
We should create a separate section that explains better what worksheets are and explain all the possibilities like for example adding imports.

ckipp01:
Closes #1935

### DIFF
--- a/metals-docs/src/main/scala/docs/WorksheetModifier.scala
+++ b/metals-docs/src/main/scala/docs/WorksheetModifier.scala
@@ -1,0 +1,83 @@
+package docs
+
+import scala.meta.inputs.Input
+
+import mdoc.Reporter
+import mdoc.StringModifier
+
+class WorksheetModifier extends StringModifier {
+  override val name: String = "worksheet"
+
+  override def process(
+      info: String,
+      code: Input,
+      reporter: Reporter
+  ): String = {
+
+    val (howYouSeeEvaluations, howToHover) = info match {
+      case "vscode" =>
+        "as a decoration at the end of the line." ->
+          "hover on the decoration to expand the decoration."
+      case "vim" =>
+        """|differently depending on whether you are using Neovim or Vim. If 
+           |using Vim, you will see them appear as comments at the end of
+           |the line. If you're using Nvim you will see this as virtual text.
+           |Keep in mind that if you're using coc-metals with Nvim you'll need
+           |to make sure `codeLens.enable` is set to `true`.""".stripMargin ->
+          """|hover on the comment if you're in Vim or expand the virtual text by
+             |using the `coc-metals-expand-decoration` command. The default for this
+             |is:
+             |```vim
+             |nmap <Leader>ws <Plug>(coc-metals-expand-decoration)
+             |```
+             |""".stripMargin
+      case _ =>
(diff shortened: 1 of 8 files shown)


metals-docs/src/main/scala/docs/WorksheetModifier.scala
```diff
@@ -0,0 +1,83 @@
...
+             |""".stripMargin
+      case _ =>
+        "as a comment as the end of the line." -> "hover on the comment to expand."
+    }
+
+    s"""|## Worksheets
+        |
+        |Worksheets are a great way to explore an api, try out an idea, or code
+        |up an example and quickly see the evaluated expression or result. Behind
+        |the schens worksheets are powered by the great work done in
```
tgodzik, line 42:
```suggestion
        |the scenes worksheets are powered by the great work done in
```

metals-docs/src/main/scala/docs/WorksheetModifier.scala
```diff
@@ -0,0 +1,83 @@
...
+        |[mdoc](https://scalameta.org/mdoc/).
+        |
+        |### Getting started with Worksheets
+        |
+        |To get started with a worksheet you can either use the `metals.new-scala-file`
+        |command and select *Worksheet* or create a file called `*.worksheet.sc`.
+        |This format is important as it is what tells Metals that it's meant to be
```
tgodzik, line 49:
```suggestion
        |This format is important as this is what tells Metals that it's meant to be
```
?

metals-docs/src/main/scala/docs/WorksheetModifier.scala
```diff
@@ -0,0 +1,83 @@
...
+        |treated as a worksheet and not just a Scala script. Where you create the
```
tgodzik, line 50: `It also matters where you create the worksheet` ?

metals-docs/src/main/scala/docs/WorksheetModifier.scala
```diff
@@ -0,0 +1,83 @@
...
+        |script also matters. If you'd like to use classes and values from your
+        |project, you need to make sure the worksheet is created inside of your `src`
+        |directory. You can still create a worksheet in other places, but you will
+        |only have access to the standard library.
```
tgodzik, line 54: `or the local library dependencies` ?

metals-docs/src/main/scala/docs/WorksheetModifier.scala
ckipp01: Fixed 👍

metals-docs/src/main/scala/docs/WorksheetModifier.scala
ckipp01: Changed.

metals-docs/src/main/scala/docs/WorksheetModifier.scala
ckipp01: I think both are actually correct.

metals-docs/src/main/scala/docs/WorksheetModifier.scala
ckipp01: A good catch. I'll add this in.

Issue #1943, opened 2020-07-23
`metals.enableStripMarginOnTypeFormatting` not being taken into account


ckipp01:
**Describe the bug**
When a user turns on `metals.enableStripMarginOnTypeFormatting`, and they hit enter at the following position:
```scala
val test = """|@@"""
```
This will add the pipe and the `.stripMargin` even though the setting is set to false via User Configuration.

**To Reproduce**
1. Create the above code snippet
2. Ensure `metals.enableStripMarginOnTypeFormatting` is set to false
2. Hit return at `@@`
3. See that `.stripMargin` is added

**Expected behavior**
I would expect `.stripMargin` not to be added when this setting is turned off.

**Installation:**
- Operating system: macOS
- Editor: Vim
- Metals version: v0.9.2

**Search terms**
onTypeFormatting, stripMargin, multiline string

tgodzik:
Thanks for reporting! Just checked and I can confirm. Maybe I could fix this during the livestream tomorrow :thinking:

ckipp01:
> Thanks for reporting! Just checked and I can confirm. Maybe I could fix this during the livestream tomorrow 🤔

That's a good idea! It should be small enough one to possibly get done as well in not too much time.

PR #1946, opened 2020-07-24
Fix enableStripMargin setting

tgodzik:
Previously, the enableStripMargin setting was not being updated. Now, it's evaluated lazily.

Fixes https://github.com/scalameta/metals/issues/1943

### DIFF
--- a/tests/unit/src/test/scala/tests/OnTypeFormattingSuite.scala
+++ b/tests/unit/src/test/scala/tests/OnTypeFormattingSuite.scala
@@ -3,6 +3,7 @@ package tests
 import scala.concurrent.Future
 
 import munit.Location
+import munit.TestOptions
 
 class OnTypeFormattingSuite extends BaseLspSuite("onTypeFormatting") {
   private val indent = "  "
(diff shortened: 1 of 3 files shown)


tests/unit/src/test/scala/tests/OnTypeFormattingSuite.scala
```diff
@@ -421,6 +422,28 @@ class OnTypeFormattingSuite extends BaseLspSuite("onTypeFormatting") {
     stripMarginEnabled = false
   )
 
+  check(
+    "add-stripMargin-with-config".only,
```
ckipp01, line 426: Remove `only`?

tgodzik:
Not sure why but the JDK 8 tests seem to not finish very often. I will have to take a look at that :/

tests/unit/src/test/scala/tests/OnTypeFormattingSuite.scala
gabro: we could probably have a Scalafix linting rule for this. It could be shipped with MUnit

tgodzik:
BillLspSuite is failing, which was ignored in another PR

Issue #1900, opened 2020-07-10
Import suggestions aren't deduplicated


kubukoz:
**Describe the bug**
The quick fixes for importing symbols can contain duplicates. In large files, this can result in way too many options spanning more than a single screen can fit.

**To Reproduce**
1. Write some code that needs an import to work
2. Write it twice
3. Select it all and open quick fixes (cmd+. on mac)
4. See duplicate import suggestions

**Expected behavior**
Suggestions are deduplicated before being shown.

**Screenshots**
![image](https://user-images.githubusercontent.com/894884/87192957-e8ffa480-c2f7-11ea-860e-736d8d5ffe09.png)

![image](https://user-images.githubusercontent.com/894884/87192964-ebfa9500-c2f7-11ea-8f46-e3957ecfdead.png)

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.9.1 / extension version 1.9.1

**Search terms**
Import, duplicated, duplicate, dupes, dupe, auto import

Update: Nice ticket number!

tgodzik:
Thanks for reporting! I can confirm the issue, the problem is that we translate each error to a new code action. We shouldprobably group by the message here:

https://github.com/scalameta/metals/blob/48b20201f3357738ce472c40dcae47b0cccc56fb/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala#L52

kubukoz:
I suppose that would be more useful than deduplicating just the suggestions, but also more expensive (since there's more information in a diagnostic, and also more diagnostics than just missing imports). Do you think it'll have a heavy impact?

PR #1947, opened 2020-07-24
Fix duplicated code actions

tgodzik:
Previously, when diagnostics were showing the same symbol code actions were duplicated. Now we make sure the for each unique symbol we have a single code action.

Fixes https://github.com/scalameta/metals/issues/1900

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/CodeAction.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/CodeAction.scala
@@ -3,6 +3,7 @@ package scala.meta.internal.metals
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 
+import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.pc.CancelToken
 
 import org.eclipse.{lsp4j => l}
@@ -20,4 +21,31 @@ trait CodeAction {
       token: CancelToken
   )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]]
 
+  implicit val actionDiagnosticOrdering: Ordering[l.CodeAction] =
+    new Ordering[l.CodeAction] {
+
+      override def compare(
+          x: l.CodeAction,
+          y: l.CodeAction
+      ): Int = {
+
+        (
+          x.getDiagnostics().asScala.headOption,
+          y.getDiagnostics().asScala.headOption
+        ) match {
+          case (Some(diagx), Some(diagy)) =>
+            val startx = diagx.getRange().getStart()
+            val starty = diagy.getRange().getStart()
+            val line = startx.getLine().compare(starty.getLine())
+            if (line == 0) {
+              startx.getCharacter().compare(starty.getCharacter())
+            } else {
+              line
+            }
+          case (Some(_), None) => 1
+          case (None, Some(_)) => -1
+          case _ => 0
+        }
+      }
+    }
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
@@ -18,6 +18,21 @@ class ImportMissingSymbol(compilers: Compilers) extends CodeAction {
       token: CancelToken
   )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
 
+    val uri = params.getTextDocument().getUri()
+
+    def joinActionEdits(actions: Seq[l.CodeAction]) = {
+      actions
+        .flatMap(_.getEdit().getChanges().get(uri).asScala)
+        .distinct
+        .groupBy(_.getRange())
+        .values
+        .map(_.sortBy(_.getNewText()).reduceLeft { (l, r) =>
+          l.setNewText((l.getNewText() + r.getNewText()).replace("\n\n", "\n"))
+          l
+        })
+        .toSeq
+    }
+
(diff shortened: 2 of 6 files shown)


tgodzik:
Turns out that we need to group the actions afterwards, otherwise we will not get all proper ranges for all the edits.

gabro:
Also, minor, but each code action has an associated list of diagnostics it fixes.
I believe it’s largely ignored by editors (I had initially hoped it would optimistically clear the diagnostic once the fix is applied, but it’s not the case in VSCode).

Does this PR change anything about it? I think it would be worth it to keep it “correct” in case editors decide to do anything with it (Sorry for the drive-by comment, but I’m on vacation and from mobile)

tgodzik:
> Also, minor, but each code action has an associated list of diagnostics it fixes.
> I believe it’s largely ignored by editors (I had initially hoped it would optimistically clear the diagnostic once the fix is applied, but it’s not the case in VSCode).
> 
> Does this PR change anything about it? I think it would be worth it to keep it “correct” in case editors decide to do anything with it (Sorry for the drive-by comment, but I’m on vacation and from mobile)

Some tests were failing so I actually changed it to keep all the diagnostics and all actions are calculated separately then joined together if the title is the same.

gabro:
Cool, thanks for the explanation 👍

tgodzik:
The tests seem to be passing finally. I had to change the code a bit, @ckipp01 could you take a look again?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
```diff
@@ -18,6 +18,21 @@ class ImportMissingSymbol(compilers: Compilers) extends CodeAction {
...
+    val uri = params.getTextDocument().getUri()
+
+    def joinActionEdits(actions: Seq[l.CodeAction]) = {
+      actions
+        .flatMap(_.getEdit().getChanges().get(uri).asScala)
+        .distinct
+        .groupBy(_.getRange())
+        .values
+        .map(_.sortBy(_.getNewText()).reduceLeft { (l, r) =>
+          l.setNewText(l.getNewText() + r.getNewText() replace ("\n\n", "\n"))
```
ckipp01, line 30: Super nitpicky, but just for consistency, can we avoid infix notation here? I know it was just copied from below.
```suggestion
          l.setNewText(l.getNewText() + r.getNewText().replace("\n\n", "\n"))
```

tests/unit/src/test/scala/tests/codeactions/ImportMissingSymbolLspSuite.scala
```diff
@@ -118,13 +145,13 @@ class ImportMissingSymbolLspSuite
...
        |""".stripMargin,
     s"""|${ImportMissingSymbol.allSymbolsTitle}
-        |${ImportMissingSymbol.title("Future", "scala.concurrent")}
-        |${ImportMissingSymbol.title("Future", "java.util.concurrent")}
-        |${ImportMissingSymbol.title("Instant", "java.time")}
         |${ImportMissingSymbol.title("ListBuffer", "scala.collection.mutable")}
+        |${ImportMissingSymbol.title("Instant", "java.time")}
+        |${ImportMissingSymbol.title("Future", "java.util.concurrent")}
+        |${ImportMissingSymbol.title("Future", "scala.concurrent")}
+        |${CreateNewSymbol.title("ListBuffer")}
```
ckipp01, line 152: Maybe not super important, but previously these suggestions seemed to be given from the top of the file down. These are now reversed starting from the bottom up. Do you think it'd be a good idea to keep them as is and shown top down?

tgodzik: Good idea! Added sorting by a first diagnostic

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
tgodzik: fixed!

tgodzik: Although funnily enough adding a `.` make the replace be applied only on `r`. Without the dot it was applied on the sum of the two strings.

metals/src/main/scala/scala/meta/internal/metals/CodeAction.scala
```diff
@@ -20,4 +20,30 @@ trait CodeAction {
...
+    override def compare(
+        x: l.CodeAction,
+        y: l.CodeAction
+    ): Int = {
+
+      (
+        x.getDiagnostics().asScala.headOption,
+        y.getDiagnostics().asScala.headOption
+      ) match {
+        case (Some(diagx), Some(diagy)) =>
```
ckipp01, line 34: Officially starting to call all diagnostics now `diagy` and `diagies`. 😆

metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
ckipp01: Ah, yea I guess that makes sense though. Good catch.

kubukoz:
@tgodzik any chance there could be a quick small release with this? :D

tgodzik:
> @tgodzik any chance there could be a quick small release with this? :D

I might be doing a new release with the updated Dotty version this week. No promises though.

kubukoz:
Sounds good, thanks

PR #1949, opened 2020-07-26
Fix `filterText` value for completion item in string interpolation

### LINKED ISSUE #1703: VSCode: Code completion in string interpolation is weird
**Describe the bug**

When I trigger the code completion in the string interpolation, its behavior is weird. Some completion proposals which should be available are not shown on the list of completion proposals.

For example, in the following case, `dir` is listed as expected:

Screen Shot 2020-05-07 at 0 17 26

However, if I type `d`, `dir` disappeared from the list somehow.

Screen Shot 2020-05-07 at 0 23 31

As long as I investigated, `MetalsLanguageServer` has always returned a proper list. So, it might not be a Metals issue.

It works properly in the following cases:

- Filter text doesn't contain `{` (`${d` doesn't work, but `$d` works)

Screen Shot 2020-05-07 at 0 29 38

- String literal has `s` prefix and Scala code is inside `${...}`

Screen Shot 2020-05-07 at 0 24 10

**To Reproduce**

Trigger the code completion in the string interpolation on VSCode.

**Expected behavior**

Code completion works properly in the string interpolation.

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.9.0

dos65:
Fixes #1703.

It seems, that this problem happened because of the `filterText` value in completion item that started with `$` or `${`.
For example:
```scala
// input
val fooBar = "test"
"${fooBa@@"

// returned completion item
CompletionList [
  isIncomplete = false
  items = SeqWrapper (
    CompletionItem [
      label = "fooBar"
      kind = Value
      tags = null
      detail = ": String"
      documentation = null
      deprecated = null
      preselect = true
      sortText = "00000"
      filterText = ""${fooBar " // should be "fooBar"
     ... 
```

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/InterpolatorCompletions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/InterpolatorCompletions.scala
@@ -48,13 +48,12 @@ trait InterpolatorCompletions { this: MetalsGlobal =>
         .append('}')
         .toString
     }
-    val filter: String =
-      text.substring(ident.pos.start - 1, cursor.point - query.length)
+
     override def contribute: List[Member] = {
       metalsTypeMembers(ident.pos).collect {
         case m if CompletionFuzzy.matches(query, m.sym.name) =>
           val edit = new l.TextEdit(pos, newText(m.sym))
-          val filterText = filter + m.sym.name.decoded
+          val filterText = m.sym.getterName.decoded
           new TextEditMember(filterText, edit, m.sym)
       }
     }
@@ -127,15 +126,12 @@ trait InterpolatorCompletions { this: MetalsGlobal =>
       out.toString
     }
 
-    val filter: String =
-      text.substring(lit.pos.start, pos.point - interpolator.name.length)
-
     override def contribute: List[Member] = {
       metalsScopeMembers(pos).collect {
         case s: ScopeMember
             if CompletionFuzzy.matches(interpolator.name, s.sym.name) =>
           val edit = new l.TextEdit(nameRange, newText(s.sym))
-          val filterText = filter + s.sym.name.decoded
+          val filterText = s.sym.getterName.decoded
           new TextEditMember(
             filterText,
             edit,
--- a/tests/cross/src/test/scala/tests/pc/CompletionInterpolatorSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/CompletionInterpolatorSuite.scala
@@ -21,7 +21,7 @@ class CompletionInterpolatorSuite extends BaseCompletionSuite {
        |  def message = s"Hello $myName$0, you are welcome"
        |}
        |""".stripMargin,
-    filterText = "\"Hello $myName"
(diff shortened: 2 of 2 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/completions/InterpolatorCompletions.scala
```diff
@@ -127,15 +126,12 @@ trait InterpolatorCompletions { this: MetalsGlobal =>
...
-    val filter: String =
-      text.substring(lit.pos.start, pos.point - interpolator.name.length)
-
     override def contribute: List[Member] = {
       metalsScopeMembers(pos).collect {
         case s: ScopeMember
             if CompletionFuzzy.matches(interpolator.name, s.sym.name) =>
           val edit = new l.TextEdit(nameRange, newText(s.sym))
-          val filterText = filter + s.sym.name.decoded
+          val filterText = s.sym.getterName.decoded
```
tgodzik, line 134: `s.sym.name.decoded` will not be enough ?

tgodzik: From what I see this should be enough or is there a particular reason you used `getterName` ?

tgodzik:
I've been thinking about why the filter text was used this way and I think the issue was the filter text was used on the whole part of the string `"part before the completion$hello@@ "` using something similar to `startsWith`. Looks like this is no longer the case with VS Code or VS Code has more complex logic for this now, however I am wondering if this will be also handled well by other editors? @ckipp01 can you confirm this will work with Vim?

ckipp01:
> @ckipp01 can you confirm this will work with Vim?

Yup, just double checked and it works as expected both in coc-metals and nvim-metals 👍

mtags/src/main/scala-2/scala/meta/internal/pc/completions/InterpolatorCompletions.scala
dos65: I've just copied its usage from `newText` methods:
https://github.com/scalameta/metals/blob/99367756fe15268a192cd2e194f27e9e2f643fef/mtags/src/main/scala-2/scala/meta/internal/pc/completions/InterpolatorCompletions.scala#L110-L113

I'm don't know if `getterName` usage is required here, but at least it should use `dropLocal` since `s.sym.name.decoded` may contain space at the end.

kpbochenek:
Will merge it later today if no more questions :)

kpbochenek:
`[error] This test is based on https://github.com/scala/scalatest-example.g8, it might have changed.`

kpbochenek:
https://github.com/scala/scalatest-example.g8/commit/2e8d027156af637a52cdba3ce65728325433196b

tgodzik:
> `[error] This test is based on https://github.com/scala/scalatest-example.g8, it might have changed.`

Good hint! :smile: 

This is already fixed in master.

tgodzik:
I rebased it - let's see how it goes.

PR #1950, opened 2020-07-28
Enable BuildInfo for mtags3

ckipp01:
I didn't realize this before I cleaned and then tried to compile again and then kept hitting on this:
```
[error] -- [E006] Not Found Error: /Users/ckipp/Documents/scala-workspace/scalameta-org/metals/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala:100:47
[error] 100 |      if (SemVer.isCompatibleVersion("0.25.0", BuildInfo.scalaCompilerVersion))
[error]     |                                               ^^^^^^^^^
[error]     |                                               Not found: BuildInfo
[error] one error found
```
We didn't have the `BuildInfo` plugin enabled for `mtag3` which is causing this to fail for someone if they do a fresh check and try to build or if you clean everything locally and try to build again since we are accessing `BuildInfo` here in `mtags3`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -340,9 +340,6 @@ lazy val mtags3 = project
     unmanagedSourceDirectories.in(Compile) += baseDirectory
       .in(ThisBuild)
       .value / "mtags" / "src" / "main" / "scala",
-    managedSourceDirectories.in(Compile) += baseDirectory
-      .in(ThisBuild)
-      .value / "mtags" / "target" / ("scala-" + scalaVersion.value) / "src_managed" / "main",
     moduleName := "mtags3",
     scalaVersion := V.scala3,
     target := baseDirectory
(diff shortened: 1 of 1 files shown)


ckipp01:
> This should have been fixed by [#1938 (files)](https://github.com/scalameta/metals/pull/1938/files), but this might actually be better. We should remove this part though:
> 
> ```
>     managedSourceDirectories.in(Compile) += baseDirectory
>       .in(ThisBuild)
>       .value / "mtags" / "target" / ("scala-" + scalaVersion.value) / "src_managed" / "main",
> ```

Sounds good. I'll remove that then. And yea, it confused me a bit when we added that. I had a co-worker pull and try to build fresh and they also couldn't build so I think this route is good. I'll remove the `managedSourceDirectories`.

PR #1951, opened 2020-07-28
Update Scala 3 version to 0.26.0-RC1

tgodzik:


mtags/src/main/scala-0.24/scala/meta/internal/pc/Symbols.scala
```diff
@@ -0,0 +1,9 @@
+package scala.meta.internal.pc
+
+import dotty.tools.dotc.core.Symbols._
+import dotty.tools.dotc.core.Contexts._
+import dotty.tools.dotc.core.Annotations.AnnotInfo
+
+object Symbols {
```
tgodzik, line 7: `dotty.tools.dotc.core.Annotations.AnnotInfo` was removed in 0.26

mtags/src/main/scala-3/scala/meta/internal/pc/Scala3CompilerWrapper.scala
```diff
@@ -11,8 +11,8 @@ class Scala3CompilerWrapper(driver: InteractiveDriver)
   override def compiler(): InteractiveDriver = driver
 
   override def resetReporter(): Unit = {
-    val ctx = driver.currentCtx
-    ctx.reporter.removeBufferedMessages(ctx)
+    implicit val ctx = driver.currentCtx
```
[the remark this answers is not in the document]
gabro: I think you need to do

```scala
ctx.reporter.removeBufferedMessages(using ctx)
```

https://dotty.epfl.ch/docs/reference/contextual/using-clauses.html

mtags/src/main/scala-3/scala/meta/internal/pc/CompilerInterfaces.scala
```diff
@@ -31,8 +31,9 @@ object CompilerInterfaces {
     val sourceFile = toSource(uri, text)
     // we need to have a separate reporter otherwise errors are not cleared
     val parsingReporter = new StoreReporter(null)
-    val freshContext = driver.currentCtx.fresh.setReporter(parsingReporter)
-    val parser = new Parser(sourceFile)(freshContext)
+    implicit val freshContext =
+      driver.currentCtx.fresh.setReporter(parsingReporter)
+    val parser = new Parser(sourceFile)
```
ckipp01, line 36: Since we are switching this maybe a good idea to change over to Dotty syntax?
```suggestion
	import dotty.tools.dotc.core.Contexts.FreshContext
	
    given fc as FreshContext =
      driver.currentCtx.fresh.setReporter(parsingReporter)
    val parser = new Parser(sourceFile)
```

mtags/src/main/scala-3/scala/meta/internal/pc/Scala3CompilerWrapper.scala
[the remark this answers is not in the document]
ckipp01: Either that or change the implicit to `given ctx as Context`.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -118,12 +117,13 @@ case class ScalaPresentationCompiler(
       val driver = access.compiler()
       val uri = params.uri
       driver.run(uri, params.text)
-      implicit def ctx: Context = driver.currentCtx
 
+      implicit def ctx: Context = driver.currentCtx
```
ckipp01, line 121: Same as above we can change these implicits here to `given` s.

mtags/src/main/scala-3/scala/meta/internal/pc/Scala3CompilerWrapper.scala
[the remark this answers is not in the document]
tgodzik: You are right, though not sure why it changed now.

mtags/src/main/scala-3/scala/meta/internal/pc/CompilerInterfaces.scala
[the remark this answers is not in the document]
tgodzik: Fixed, though no scalafmt anymore :disappointed:

tgodzik: Added using here.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
tgodzik: Ok, changed it too.

tgodzik:
Entire suites started failing, but this can't be caused by this PR weirdly enough. Tests have become quite unstable and no idea why.

Issue #1654, opened 2020-04-21
DocumentText's text is sometimes empty


tgodzik:
**Describe the bug**
I am getting sometimes an exception like: 

```
SEVERE: Internal error: java.lang.IllegalArgumentException: 2 is not a valid line number, allowed [0..0]
java.util.concurrent.CompletionException: java.lang.IllegalArgumentException: 2 is not a valid line number, allowed [0..0]
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:331)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:346)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:704)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088)
	at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1(CancelTokens.scala:38)
	at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1$adapted(CancelTokens.scala:36)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.IllegalArgumentException: 2 is not a valid line number, allowed [0..0]
	at scala.meta.internal.inputs.InternalInput.lineToOffset(InternalInput.scala:38)
	at scala.meta.internal.inputs.InternalInput.lineToOffset$(InternalInput.scala:33)
	
```

**To Reproduce**
This doesn't seem to be easy to reproduce, but it looks like `textDocument.text` is giving us an empty string despite having occurrences etc. I think this might just be a corrupted semanticDB file, but should most likely handle this situation:

`textDocument.text.isEmpty && textDocument.occurences.nonEmpty`

and fallback to file on the filesystem.

**Expected behavior**
No exception, everything operating as normal.

**Installation:**
- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.8.4

**Search terms**
exception line offset

PR #1953, opened 2020-07-28
Update Bloop version to support broken workspaces 

tgodzik:
Previously, if a workspace was opened and it contained errors it would result in broken  completions, references etc. Now, Bloop copies old artifacts by default if there was a previously successful compilation.

Also fixes https://github.com/scalameta/metals/issues/1654


metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -253,16 +253,9 @@ class DestinationProvider(
...
-      defnOriginalInput = Input.VirtualFile(
-        defnPathInput.path,
-        destinationDoc.text
-      )
       destinationPath = symbolDefinition.path.toFileOnDisk(workspace)
-      destinationDistance = TokenEditDistance(
-        defnOriginalInput,
-        defnPathInput
-      )
+      destinationDistance =
```
tgodzik, line 257: This is essencialy the same code.

mtags/src/main/scala-2/scala/meta/internal/metals/TokenEditDistance.scala
```diff
@@ -289,6 +289,8 @@ object TokenEditDistance {
     if (!isScala) {
       // Ignore non-scala Files.
       unchanged
+    } else if (originalInput.value.isEmpty() || revisedInput.value.isEmpty()) {
```
tgodzik, line 292: This could break otherwise. This situation means that we most likely do not know how the file looked when the semanticdb was compiled. This will result in anything depending on edit distance not to work, but will not create a risk for incorrect results.

tests/slow/src/test/scala/tests/sbt/SbtLspSuite.scala
```diff
@@ -419,4 +420,18 @@ class SbtLspSuite extends BaseImportSuite("sbt-import") {
       )
     } yield ()
   }
+
+  test("min-sbt-version") {
```
tgodzik, line 424: This is just in case, Bloop will be dropping the support for sbt `0.13`, so this test should fail if anyone wants to update to that Bloop version. We will need to freeze Bloop version for sbt 0.13 in that case

gabro: At some point Metals should probably drop 0.13 as well. The times are ripe, I think

mtags/src/main/scala-2/scala/meta/internal/metals/TokenEditDistance.scala
tgodzik: Also if one of them is empty we cannot match any tokens.

tests/unit/src/test/scala/tests/CodeLensLspSuite.scala
```diff
@@ -171,6 +171,7 @@ class CodeLensLspSuite extends BaseCodeLensLspSuite("codeLenses") {
            |  def main(args: Array[String]): Unit = ???
            |}""".stripMargin
       )
+      _ <- server.didOpen("a/src/main/scala/Main.scala")
```
tgodzik, line 174: This is needed so that we save a fingerprint of a working file, which we produce a semanticDB for. SemanticDB has positions that might need to be adjusted in case of a broken/unsaved code and we need the difference between last working and current to figure out how to adjust those positions.

tgodzik:
Some things we might follow up this with is getting `positionOccurence` from the compiler and adding a fallback for TokenEditDistance when the symbol is located at the same position as in the smenaticDB file.

PR #1954, opened 2020-07-29
Bump default mill version and also add it to BuildInfo

ckipp01:
I know we talked about this before. It's maybe not the cleanest, but it will allow for Scalasteward to just pick these up in the future.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -216,6 +216,7 @@ lazy val V = new {
   val coursier = "2.0.0-RC6-23"
   val coursierInterfaces = "0.0.22"
   val ammonite = "2.1.4-11-307f3d8"
+  val mill = "0.8.0"
 }
 
 val genyVersion = Def.setting {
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -574,9 +576,10 @@ lazy val unit = project
       "io.get-coursier" %% "coursier" % V.coursier, // for jars
       "ch.epfl.scala" %% "bloop-config" % V.bloop,
       "org.scalameta" %% "munit" % V.munit,
-      // Only here to have scala-steward update V.ammonite.
-      // Not actually used in tests.
-      "com.lihaoyi" %% "ammonite-util" % V.ammonite intransitive ()
+      // The dependencies listed below are only listed so Scala Steward
+      // will pick them up and update them. They aren't actually used.
+      "com.lihaoyi" %% "ammonite-util" % V.ammonite intransitive (),
+      "com.lihaoyi" % "mill-contrib-testng" % V.mill intransitive ()
```
ckipp01, line 582: The package name seems odd, but it is indeed just `testng`. I figured this was a nice small one to include for this purpose.

ckipp01:
> Looks good! I wonder if there is also a way to not specify it in build.sbt, but have a special string? OR hint in the comments? I think I've seen a conversation about it somewhere.

I actually just created a question over there, so I'll wait to see. I'm assuming @fthomas has a better way to do this that we probably aren't thinking of.

ckipp01:
I'm just going to merge this for now, and revisit this one and the Ammonite one when I figure out a better way to do this.

PR #1965, opened 2020-08-01
Swap ordering in CancelCompletionSuite test to fix flakiness.

ckipp01:
So I believe this may have been an oversight when this was changed in https://github.com/scalameta/metals/commit/35af46a840e42c7d504259701e528f73c6787281#diff-56d26dd129140627b4c1c80ace3f1186, but when testing locally, this continually fails for me and this suite has also been somewhat consistently failing in tests. I'm not sure what's going on with the general flakiness of the suites, but this should at least help with this one as it seems the order needs to be swapped.

### DIFF
--- a/tests/cross/src/test/scala/tests/pc/CancelCompletionSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/CancelCompletionSuite.scala
@@ -91,12 +91,12 @@ class CancelCompletionSuite extends BaseCompletionSuite {
        |""".stripMargin,
     compat = Map(
       "0.25" ->
-        """|assert(assertion: Boolean @InlineParam): Unit
-           |assertFail(message: => Any): Nothing
-           |""".stripMargin,
-      "0.26" ->
         """|assertFail(message: => Any): Nothing
            |assert(assertion: Boolean @InlineParam): Unit
+           |""".stripMargin,
+      "0.26" ->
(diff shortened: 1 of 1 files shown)


kpbochenek:
I can see for scala2 we explicitly sort completions, but I cannot find it for dotty, does dotty sort them by default?
Or non determinism actually comes from results not being sorted at all?

tgodzik:
> I can see for scala2 we explicitly sort completions, but I cannot find it for dotty, does dotty sort them by default?
> Or non determinism actually comes from results not being sorted at all?

That might be it, we could add a flag to sort completions until we have a stable completion list.

tgodzik:
Looks like it's failing again. :thinking:  Though it hasn't been problematic before 0.26.0

ckipp01:
> I can see for scala2 we explicitly sort completions, but I cannot find it for dotty, does dotty sort them by default?
> Or non determinism actually comes from results not being sorted at all?

Ahh, good point. Let me look into this a bit.

>  Looks like it's failing again. 🤔 Though it hasn't been problematic before 0.26.0

Yea, I'm really sure what going on. It was a bit frustrating that locally I couldn't get the test to pass without this change, and now it fails here. I'll try to do what @kpbochenek and try to somewhat control the order of the completions like we do for Scala 2.x.

PR #1966, opened 2020-08-01
Analyze stacktrace

kpbochenek:
Stacktrace analyzer PoC

stacktrace-analyze

It creates `.metals/stacktrace.scala` file with stacktrace (provided in command, in vscode PR taken from clipboard like in Intellij)
Because stacktrace lines are in format: `scope.class.method(filename:line)` I try to convert `scope.class.method` into symbol and look for definition of it. If found I use path where definition is located and set line from stacktrace.
For very simple stacktraces it works okay, there might be issues with annonymous classes etc. (to be tested further).

If you are on a given line you can do action 'go to definition' that will move cursor to place where stacktrace line points to.
For every located symbol also there are added code lenses. 

For `goto-location` I added optional parameter `otherWindow` which says to client that it would be perfect if you moved cursor to given location but in another window/frame/tab. (you can see this in action in gif, clicking "jump" opens position on other window)

Right now please don't comment 'use different name for this variable' etc. but let me know what do you think about this concept generally?
If it looks good to you I will polish code and change status 'Draft' -> PR.

Complementary vscode PR:  https://github.com/scalameta/metals-vscode/pull/371

Not much changes on VSCode side so I don't expect problems on vim/emacs side but feedback for those editors would be good too

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -40,7 +40,7 @@ import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.ammonite.Ammonite
 import scala.meta.internal.metals.codelenses.RunTestCodeLens
 import scala.meta.internal.metals.codelenses.SuperMethodCodeLens
-import scala.meta.internal.metals.config.DoctorFormat
+import scala.meta.internal.metals.debug.BuildTargetClasses
 import scala.meta.internal.metals.debug.DebugParametersJsonParsers
 import scala.meta.internal.metals.debug.DebugProvider
 import scala.meta.internal.mtags._
@@ -435,8 +435,9 @@ class MetalsLanguageServer(
 
     stacktraceAnalyzer = new StacktraceAnalyzer(
       workspace,
+      buffers,
       definitionProvider,
-      clientConfig.icons.findsuper,
+      clientConfig.icons,
       clientConfig.isCommandInHtmlSupported
     )
 
@@ -791,43 +792,43 @@ class MetalsLanguageServer(
   @JsonNotification("textDocument/didOpen")
   def didOpen(params: DidOpenTextDocumentParams): CompletableFuture[Unit] = {
     val path = params.getTextDocument.getUri.toAbsolutePath
-    if (stacktraceAnalyzer.matches(path)) {
+    if (stacktraceAnalyzer.isStackTraceFile(path)) {
       CompletableFuture.completedFuture(())
     } else {
-    focusedDocument = Some(path)
-    openedFiles.add(path)
+      focusedDocument = Some(path)
+      openedFiles.add(path)
 
-    // Update md5 fingerprint from file contents on disk
-    fingerprints.add(path, FileIO.slurp(path, charset))
-    // Update in-memory buffer contents from LSP client
-    buffers.put(path, params.getTextDocument.getText)
-    val didChangeFuture = compilers.didChange(path)
+      // Update md5 fingerprint from file contents on disk
+      fingerprints.add(path, FileIO.slurp(path, charset))
+      // Update in-memory buffer contents from LSP client
+      buffers.put(path, params.getTextDocument.getText)
+      val didChangeFuture = compilers.didChange(path)
 
-    packageProvider
-      .workspaceEdit(path)
-      .map(new ApplyWorkspaceEditParams(_))
-      .foreach(languageClient.applyEdit)
+      packageProvider
+        .workspaceEdit(path)
+        .map(new ApplyWorkspaceEditParams(_))
+        .foreach(languageClient.applyEdit)
 
-    if (path.isDependencySource(workspace)) {
-      CancelTokens { _ =>
-        // trigger compilation in preparation for definition requests
-        interactiveSemanticdbs.textDocument(path)
-        // publish diagnostics
-        interactiveSemanticdbs.didFocus(path)
-        ()
+      if (path.isDependencySource(workspace)) {
+        CancelTokens { _ =>
+          // trigger compilation in preparation for definition requests
+          interactiveSemanticdbs.textDocument(path)
+          // publish diagnostics
+          interactiveSemanticdbs.didFocus(path)
+          ()
+        }
+      } else {
+        if (path.isAmmoniteScript)
+          ammonite.maybeImport(path)
+        val loadFuture = compilers.load(List(path))
+        val compileFuture =
+          compilations.compileFile(path)
+        Future
+          .sequence(List(didChangeFuture, loadFuture, compileFuture))
+          .ignoreValue
+          .asJava
       }
-    } else {
-      if (path.isAmmoniteScript)
-        ammonite.maybeImport(path)
-      val loadFuture = compilers.load(List(path))
-      val compileFuture =
-        compilations.compileFile(path)
-      Future
-        .sequence(List(didChangeFuture, loadFuture, compileFuture))
-        .ignoreValue
-        .asJava
     }
-  }
   }
 
   @JsonNotification("metals/didFocusTextDocument")
@@ -1086,14 +1087,7 @@ class MetalsLanguageServer(
       position: TextDocumentPositionParams
   ): CompletableFuture[util.List[Location]] =
     CancelTokens.future { token =>
-      val path = position.getTextDocument.getUri.toAbsolutePath
-      if (stacktraceAnalyzer.matches(path)) {
-        Future.successful {
-          stacktraceAnalyzer.definition(path, position.getPosition())
-        }
-      } else {
-        definitionOrReferences(position, token).map(_.locations)
-      }
+      definitionOrReferences(position, token).map(_.locations)
     }
 
   @JsonRequest("textDocument/typeDefinition")
@@ -1134,12 +1128,7 @@ class MetalsLanguageServer(
       params: TextDocumentPositionParams
   ): CompletableFuture[util.List[DocumentHighlight]] =
     CancelTokens { _ =>
-      val path = params.getTextDocument.getUri.toAbsolutePath
-      if (stacktraceAnalyzer.matches(path)) {
-        stacktraceAnalyzer.highlight(path, params.getPosition().getLine())
-      } else {
-        documentHighlightProvider.documentHighlight(params)
-      }
+      documentHighlightProvider.documentHighlight(params)
     }
 
   @JsonRequest("textDocument/documentSymbol")
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -214,6 +214,7 @@ class MetalsLanguageServer(
   var treeView: TreeViewProvider = NoopTreeViewProvider
   var worksheetProvider: WorksheetProvider = _
   var popupChoiceReset: PopupChoiceReset = _
+  var stacktraceAnalyzer: StacktraceAnalyzer = _
 
   private val clientConfig: ClientConfiguration =
     new ClientConfiguration(
@@ -431,9 +432,17 @@ class MetalsLanguageServer(
       () => userConfig,
       clientConfig
     )
+
+    stacktraceAnalyzer = new StacktraceAnalyzer(
+      workspace,
+      definitionProvider,
+      clientConfig.icons.findsuper
+    )
+
     codeLensProvider = new CodeLensProvider(
       List(runTestLensProvider, goSuperLensProvider),
-      semanticdbs
+      semanticdbs,
+      stacktraceAnalyzer
     )
     renameProvider = new RenameProvider(
       referencesProvider,
@@ -781,6 +790,8 @@ class MetalsLanguageServer(
   @JsonNotification("textDocument/didOpen")
   def didOpen(params: DidOpenTextDocumentParams): CompletableFuture[Unit] = {
     val path = params.getTextDocument.getUri.toAbsolutePath
+    if (stacktraceAnalyzer.matches(path))
+      return CompletableFuture.completedFuture(())
     focusedDocument = Some(path)
     openedFiles.add(path)
 
@@ -1071,7 +1082,14 @@ class MetalsLanguageServer(
       position: TextDocumentPositionParams
   ): CompletableFuture[util.List[Location]] =
     CancelTokens.future { token =>
-      definitionOrReferences(position, token).map(_.locations)
+      val path = position.getTextDocument.getUri.toAbsolutePath
+      if (stacktraceAnalyzer.matches(path)) {
+        Future.successful {
+          stacktraceAnalyzer.definition(path, position.getPosition())
+        }
+      } else {
+        definitionOrReferences(position, token).map(_.locations)
+      }
     }
 
   @JsonRequest("textDocument/typeDefinition")
@@ -1111,7 +1129,14 @@ class MetalsLanguageServer(
   def documentHighlights(
       params: TextDocumentPositionParams
   ): CompletableFuture[util.List[DocumentHighlight]] =
-    CancelTokens { _ => documentHighlightProvider.documentHighlight(params) }
+    CancelTokens { _ =>
+      val path = params.getTextDocument.getUri.toAbsolutePath
+      if (stacktraceAnalyzer.matches(path)) {
+        stacktraceAnalyzer.highlight(path, params.getPosition().getLine())
+      } else {
+        documentHighlightProvider.documentHighlight(params)
+      }
+    }
 
   @JsonRequest("textDocument/documentSymbol")
   def documentSymbol(
@@ -1359,7 +1384,18 @@ class MetalsLanguageServer(
         Future {
           compilers.restartAll()
         }.asJavaObject
-      case ServerCommands.GotoLocation() =>
+      case ServerCommands.GotoLocationForPosition() =>
+        Future {
+          // arguments are not checked but are of format singletonList(location: Location)
+          languageClient.metalsExecuteClientCommand(
+            new ExecuteCommandParams(
+              ClientCommands.GotoLocation.id,
+              params.getArguments()
+            )
+          )
+        }.asJavaObject
+
+      case ServerCommands.GotoLocationForSymbol() =>
         Future {
           for {
             args <- Option(params.getArguments())
@@ -1422,6 +1458,13 @@ class MetalsLanguageServer(
           DebugSession(server.sessionName, server.uri.toString)
         }
         session.asJavaObject
+
+      case ServerCommands.AnalyzeStacktrace() =>
+        Future {
+          val command = stacktraceAnalyzer.analyzeCommand(params)
+          command.foreach(languageClient.metalsExecuteClientCommand)
+          scribe.debug(s"Executing AnalyzeStacktrace ${command}")
+        }.asJavaObject
 
       case ServerCommands.GotoSuperMethod() =>
         Future {
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -41,6 +41,7 @@ import scala.meta.internal.metals.ammonite.Ammonite
 import scala.meta.internal.metals.codelenses.RunTestCodeLens
 import scala.meta.internal.metals.codelenses.SuperMethodCodeLens
 import scala.meta.internal.metals.debug.BuildTargetClasses
+import scala.meta.internal.metals.config.DoctorFormat
 import scala.meta.internal.metals.debug.DebugParametersJsonParsers
 import scala.meta.internal.metals.debug.DebugProvider
 import scala.meta.internal.mtags._
@@ -436,7 +437,8 @@ class MetalsLanguageServer(
     stacktraceAnalyzer = new StacktraceAnalyzer(
       workspace,
       definitionProvider,
-      clientConfig.icons.findsuper
+      clientConfig.icons.findsuper,
+      clientConfig.isCommandInHtmlSupported
     )
 
     codeLensProvider = new CodeLensProvider(
@@ -790,8 +792,9 @@ class MetalsLanguageServer(
   @JsonNotification("textDocument/didOpen")
   def didOpen(params: DidOpenTextDocumentParams): CompletableFuture[Unit] = {
     val path = params.getTextDocument.getUri.toAbsolutePath
-    if (stacktraceAnalyzer.matches(path))
-      return CompletableFuture.completedFuture(())
+    if (stacktraceAnalyzer.matches(path)) {
+      CompletableFuture.completedFuture(())
+    } else {
     focusedDocument = Some(path)
     openedFiles.add(path)
 
@@ -826,6 +829,8 @@ class MetalsLanguageServer(
         .asJava
     }
   }
+  }
+
   @JsonNotification("metals/didFocusTextDocument")
   def didFocus(uri: String): CompletableFuture[DidFocusResult.Value] = {
     val path = uri.toAbsolutePath
@@ -1386,7 +1391,8 @@ class MetalsLanguageServer(
         }.asJavaObject
       case ServerCommands.GotoLocationForPosition() =>
         Future {
-          // arguments are not checked but are of format singletonList(location: Location)
+          // arguments are not checked but are of format:
+          // singletonList(location: Location, otherWindow: Boolean)
           languageClient.metalsExecuteClientCommand(
             new ExecuteCommandParams(
               ClientCommands.GotoLocation.id,
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -792,42 +792,38 @@ class MetalsLanguageServer(
   @JsonNotification("textDocument/didOpen")
   def didOpen(params: DidOpenTextDocumentParams): CompletableFuture[Unit] = {
     val path = params.getTextDocument.getUri.toAbsolutePath
-    if (stacktraceAnalyzer.isStackTraceFile(path)) {
-      CompletableFuture.completedFuture(())
-    } else {
-      focusedDocument = Some(path)
-      openedFiles.add(path)
+    focusedDocument = Some(path)
+    openedFiles.add(path)
 
-      // Update md5 fingerprint from file contents on disk
-      fingerprints.add(path, FileIO.slurp(path, charset))
-      // Update in-memory buffer contents from LSP client
-      buffers.put(path, params.getTextDocument.getText)
-      val didChangeFuture = compilers.didChange(path)
+    // Update md5 fingerprint from file contents on disk
+    fingerprints.add(path, FileIO.slurp(path, charset))
+    // Update in-memory buffer contents from LSP client
+    buffers.put(path, params.getTextDocument.getText)
+    val didChangeFuture = compilers.didChange(path)
 
-      packageProvider
-        .workspaceEdit(path)
-        .map(new ApplyWorkspaceEditParams(_))
-        .foreach(languageClient.applyEdit)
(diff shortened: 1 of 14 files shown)


olafurpg:
This is an IntelliJ feature I have heavily missed in VS Code. I love this idea! 👏

ckipp01:
This is really cool @kpbochenek! 💯

tgodzik:
Looks really cool! This will be really useful.

I wonder if in VS Code we could use a webview for that? I think most things we can models as command links, so this would just be an HTML with added commands for each line.  The Scala file could be a fallback for other editors or maybe a normal HTML with URI links?

Also, we should not try to parse the file or we could make everything into a comment in order not to add false positive parsing errors.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -154,15 +154,34 @@ object ServerCommands {
        |""".stripMargin
   )
 
-  val GotoLocation = new Command(
+  val AnalyzeStacktrace = new Command(
+    "analyze-stacktrace",
+    "Analyze stacktrace",
+    """|<TODO>
```
kpbochenek, line 160: No vision what to put into description here ❓

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -780,6 +789,8 @@ class MetalsLanguageServer(
   @JsonNotification("textDocument/didOpen")
   def didOpen(params: DidOpenTextDocumentParams): CompletableFuture[Unit] = {
     val path = params.getTextDocument.getUri.toAbsolutePath
+    if (stacktraceAnalyzer.matches(path))
+      return CompletableFuture.completedFuture(())
```
kpbochenek, line 793: will convert it to proper if/else without return if this stays here

kpbochenek:
Screenshot 2020-08-03 at 21 21 26

kpbochenek:
As @tgodzik requested for VSCode there is a special handling using webview:

stacktrace-webview

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -431,9 +433,18 @@ class MetalsLanguageServer(
       () => userConfig,
       clientConfig
     )
+
+    stacktraceAnalyzer = new StacktraceAnalyzer(
+      workspace,
+      definitionProvider,
+      clientConfig.icons.findsuper,
+      clientConfig.doctorFormat() == DoctorFormat.Html
```
kpbochenek, line 441: this ideally should be a check for VSCode (or client that can execute commands inside html)

ckipp01:
> @ckipp01 would you be able/have time to give it a try with vim and let me know how does it look on that side?

For sure, I'll take a look at this today. Looking through it in the VS Code extension, there will simply have to be some things to work out with coc-metals. Even just accessing the clipboard is a lot more work, and then figuring out a good way to display this to the user may also be tricky since there isn't a webview per se. I'll play around with it and report back.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
ckipp01: 🤔 I wonder if we are broadening the `doctorFormat` too much? Or maybe the name no longer fully captures what it is being used for. In the scenario, if a user has `http` enabled and they also have the doctor format being Html, will it open in the browser?

kpbochenek: replaced it with dedicated setting :+1:

ckipp01:
So after a little exploring, this will probably be a decent amount of work to just get all the moving pieces working for vim. Oddly enough if you do something like:

```scala
object Test extends App {

  println("hi")
  println("hO")

  throw new Exception("noooooooo")
}
```
This will throw your exception and show your stacktrace in vscode. With vimspector, it will print `hi` and `h0` and then just stop. It won't even show the trace. So I need to dive into what is going on there before this is even a possibility. So don't let that slow the progress on this at all.

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,227 @@
...
+import scala.meta.io.AbsolutePath
+
+import com.google.gson.JsonPrimitive
+import org.eclipse.lsp4j.Location
+import org.eclipse.{lsp4j => l}
+
+class StacktraceAnalyzer(
+    workspace: AbsolutePath,
+    definitionProvider: DefinitionProvider,
+    gotoicon: String,
```
tgodzik, line 20: Let's use `Icons` instead of string here.

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,227 @@
...
+) {
+
+  def analyzeCommand(
+      commandParams: l.ExecuteCommandParams
+  ): Option[l.ExecuteCommandParams] = {
+    parseJsonParams(commandParams)
+      .flatMap(analyzeStackTrace)
+  }
+
+  def matches(path: AbsolutePath): Boolean =
```
tgodzik, line 31:
```suggestion
  def isStackTraceFile(path: AbsolutePath): Boolean =
```
?

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  private def readStacktraceFileLine(
+      path: AbsolutePath,
+      lineNumber: Int
+  ): Option[String] =
+    readStacktraceFile(path).drop(lineNumber).headOption
+
+  private def readStacktraceFile(path: AbsolutePath): List[String] = {
+    Source.fromFile(path.toString).getLines().toList
```
tgodzik, line 63: This should be in buffers, no need to read from Source.

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,227 @@
...
+    path == workspace.resolve(Directories.stacktrace)
+
+  def highlight(
+      path: AbsolutePath,
+      lineNumber: Int
+  ): java.util.List[l.DocumentHighlight] = {
+    val lineOpt = readStacktraceFileLine(path, lineNumber)
+    lineOpt match {
+      case Some(line) =>
+        val from = new l.Position(lineNumber, line.indexOf("at ") + 3)
```
tgodzik, line 41: I don't think document highlight makes sense unless we try to highlight lines referencing the same line/file

Highlighting one line doesn't seem to add anything.

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  private def adjustPosition(lineNumber: Int, pos: l.Position): Unit = {
+    pos.setLine(lineNumber)
+    pos.setCharacter(0)
+  }
+
+  private def tryGetLineNumberFromStacktrace(line: String): Try[Int] = {
+    Try(
+      // number is 1-based, but editors are 0-based
```
tgodzik, line 86: LSP is the only one that uses 0 based ones. Both editor itself and stacktrace are 1-based. Maybe we could make the comment more precise?

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,227 @@
...
+        lineNumberOpt.foreach { lineNumber =>
+          adjustPosition(lineNumber, location.getRange().getStart())
+          adjustPosition(lineNumber, location.getRange().getEnd())
+        }
+        location
+    }
+  }
+
+  private def getSymbolFromLine(line: String): String = {
+    line.substring(line.indexOf("at ") + 3, line.indexOf("("))
```
tgodzik, line 171: I don't think you can properly get symbol from stacktrace and the logic will always be flaky. Jumping to the line is enough. For example:
```
Exception in thread "main" java.lang.Exception
	at Main$.$anonfun$main$1(Main.scala:13)
	at Main$.$anonfun$main$1$adapted(Main.scala:10)
	at scala.collection.immutable.Range.foreach(Range.scala:190)
	at Main$.main(Main.scala:10)
	at Main.main(Main.scala)
```

It's not possible to map the full symbol `Main$.$anonfun$main$1` to anything similar in sematicDB. We can only map a part of it, which will make the logic flaky. I would drop going to definition.

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,227 @@
...
+  ): l.ExecuteCommandParams = {
+    new l.ExecuteCommandParams(
+      ClientCommands.GotoLocation.id,
+      List[Object](location, java.lang.Boolean.TRUE).asJava
+    )
+  }
+
+  private def getSymbolLocationFromLine(line: String): Option[l.Location] = {
+    val symbol = getSymbolFromLine(line)
+    definitionProvider.fromSymbol(convert(symbol)).asScala.headOption.map {
```
tgodzik, line 159: Why won't we just jump to the line? What is the definition location needed for if we adjust it anyway?

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,227 @@
...
+  private def analyzeStackTrace(
+      stacktrace: String
+  ): Option[l.ExecuteCommandParams] = {
+    if (commandsInHtmlSupported) {
+      Some(makeHtmlCommandParams(stacktrace))
+    } else {
+      val path = workspace.resolve(Directories.stacktrace)
+      val pathStr = path.toFile.toString
+
+      path.toFile.createNewFile()
```
tgodzik, line 134: Truncate existing file before?

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -154,15 +154,41 @@ object ServerCommands {
        |""".stripMargin
   )
 
-  val GotoLocation = new Command(
+  val AnalyzeStacktrace = new Command(
+    "analyze-stacktrace",
+    "Analyze stacktrace",
+    """|Converts provided stacktrace in parameter to a format that contains links to locations
```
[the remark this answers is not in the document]
kpbochenek: changed

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -154,15 +154,41 @@ object ServerCommands {
...
+       |of places where exception was raised.
```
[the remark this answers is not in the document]
kpbochenek: changed

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -154,15 +154,41 @@ object ServerCommands {
...
+       |
+       |Depending on a configuration parameter of a client(support-commands-in-html)
```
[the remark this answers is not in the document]
kpbochenek: fixed

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -154,15 +154,41 @@ object ServerCommands {
...
+       |if parametes is true then client is requested to display html with links
```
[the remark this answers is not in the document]
kpbochenek: fixed

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -154,15 +154,41 @@ object ServerCommands {
...
+       |already pointing to proper locations in user codebase.
+       |if parameter is false then client is requested to display simple scala file
```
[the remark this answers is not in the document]
kpbochenek: fixed :+1:

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
kpbochenek: ok

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
kpbochenek: renamed

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
kpbochenek: updated, not resolving it yet, let me know if it makes more sense

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
kpbochenek: to find out what file are we talking about

tgodzik: But we do have that information in the stack trace, no?

tgodzik: ach, ok I get it now. It's just the name of the file and not the full path, that's why you need it here.

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
tgodzik: Although we could do `toplevelFromLine` and just get the toplevel via:
```scala
private def getSymbolFromLine(line: String): String = {
  val toplevel = line.substring(line.indexOf("at ") + 3, line.indexOf("."))
  val descriptor = if(toplevel.endsWith("$")) List(Descriptor.Term) else List(Descriptor.Type)
  BuildTargetClasses.symbolFromClassName(toplevel.stripSuffix("$"), descriptor)
}
```

We might need to move the method from BuildTargetClasses if it makes sense. It's in the class, so some modification will be needed.

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
tgodzik: `// number is 1-based, but LSP uses 0-based line numbers` ?

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
kpbochenek: it is truncated by `new FileWriter` (by default `apppend=false`)

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
kpbochenek: okay dropping

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
kpbochenek: I think it was a nice visual aspect for going to definition as you always see relation 'highlight <-> gododefinition available'
but since goto def is dropped this can be dropped too

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
kpbochenek: changed

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
kpbochenek: haha sorry I wrote a comment and then pushed my change, you must have noticed unchanged comment, WDYT about a new one? :)

tests/unit/src/test/scala/tests/StacktraceParseSuite.scala
```diff
@@ -0,0 +1,62 @@
+package tests
+
+import scala.meta.internal.metals.StacktraceAnalyzer
+
+class StacktraceParseSuite extends BaseSuite {
+
+  test("convert-stacktrace-line-to-symbol") {
```
tgodzik, line 7: Maybe move test inside testConversion and make it all separate tests?

tests/unit/src/test/scala/tests/StacktraceParseSuite.scala
```diff
@@ -0,0 +1,62 @@
...
+
+    // --- CLASS ---
+    testConversion("p1.p2.Class.method", "p1/p2/Class#")
+    testConversion(
+      "p1.p2.OutsideClass$InsideClass.method",
+      List("p1/p2/OutsideClass#", "p1/p2/OutsideClass#InsideClass#")
```
tgodzik, line 13: We really just need "p1/p2/OutsideClass#", the other class must be located in the same file.

tgodzik (changes requested):
Great work! Just some more suggestions, but I think it's almost ready.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -154,15 +154,40 @@ object ServerCommands {
...
+    "analyze-stacktrace",
+    "Analyze stacktrace",
+    """|Converts provided stacktrace in the parameter to a format that contains links
+       |to locations of places where the exception was raised.
+       |
+       |If the configuration parameter of the client (support-commands-in-html) is true
+       |then client is requested to display html with links
+       |already pointing to proper locations in user codebase.
+       |Otherwise client will display simple scala file
+       |but with code lenses that on action move user to proper location in codebase.
```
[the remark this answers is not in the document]
kpbochenek: changed

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -154,15 +154,40 @@ object ServerCommands {
...
+       |""".stripMargin,
+    "[string], where the string is a stacktrace."
+  )
+
+  val GotoLocationForSymbol = new Command(
```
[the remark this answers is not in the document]
kpbochenek: changed

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -154,15 +154,40 @@ object ServerCommands {
...
     "goto",
-    "Goto location",
+    "Goto location for symbol",
     """|Move the cursor to the definition of the argument symbol.
        |
        |""".stripMargin,
     "[string], where the string is a SemanticDB symbol."
   )
 
+  val GotoLocationForPosition = new Command(
```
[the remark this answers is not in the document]
kpbochenek: changed

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,211 @@
...
+    readStacktraceFile(path)
+      .map(stacktraceLenses)
+      .getOrElse(Seq.empty)
+  }
+
+  private def readStacktraceFile(path: AbsolutePath): Option[List[String]] = {
+    buffers.get(path).map(_.split('\n').toList)
+  }
+
+  private def adjustPosition(lineNumber: Int, pos: l.Position): Unit = {
```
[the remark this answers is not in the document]
kpbochenek: changed

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,211 @@
...
+  private def makeGotoCommandParams(
+      location: Location
+  ): l.ExecuteCommandParams = {
+    new l.ExecuteCommandParams(
+      ClientCommands.GotoLocation.id,
+      List[Object](location, java.lang.Boolean.TRUE).asJava
+    )
+  }
+
+  private def getSymbolLocationFromLine(line: String): Option[l.Location] = {
```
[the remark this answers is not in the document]
kpbochenek: changed

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,211 @@
...
+        val lineNumberOpt = tryGetLineNumberFromStacktrace(line)
+        lineNumberOpt.foreach { lineNumber =>
+          adjustPosition(lineNumber, location.getRange().getStart())
+          adjustPosition(lineNumber, location.getRange().getEnd())
+        }
+        location
+      }
+  }
+
+  private def getSymbolFromLine(line: String): String = {
```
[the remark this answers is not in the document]
kpbochenek: changed

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,211 @@
...
+  ): l.ExecuteCommandParams = {
+    new l.ExecuteCommandParams(
+      ClientCommands.GotoLocation.id,
+      List[Object](location, java.lang.Boolean.TRUE).asJava
+    )
+  }
+
+  private def getSymbolLocationFromLine(line: String): Option[l.Location] = {
+    val symbol = getSymbolFromLine(line)
+    convert(symbol)
```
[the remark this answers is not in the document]
kpbochenek: that would work but meant different intent.
I want to find file where a symbol can be found, not do transformations on collection.

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,211 @@
...
+    val param = s"""["${uri}",${line},true]"""
+    s"command:metals.goto-path-uri?${URLEncoder.encode(param)}"
+  }
+
+  private def parseJsonParams(
+      commandParams: l.ExecuteCommandParams
+  ): Option[String] = {
+    for {
+      args <- Option(commandParams.getArguments)
+      argObject <- args.asScala.headOption
```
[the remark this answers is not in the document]
kpbochenek: doesn't compile

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,211 @@
...
+      arg = argObject.asInstanceOf[JsonPrimitive]
+      if arg.isString()
+      stacktrace = arg.getAsString()
+    } yield stacktrace
+  }
+}
+
+object StacktraceAnalyzer {
+
+  def convert(symbolIn: String): List[String] = {
```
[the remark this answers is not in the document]
kpbochenek: inlined

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,211 @@
...
+    def removeMethodPartFromSymbol(s: String) = s.split('.').init.mkString("/")
+    val symbol = removeMethodPartFromSymbol(symbolIn)
```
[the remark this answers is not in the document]
kpbochenek: inlined

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,211 @@
...
+      pathFile.createNewFile()
+      val fw = new FileWriter(pathStr)
+      try {
+        fw.write(stacktrace)
+      } finally {
+        fw.close()
+      }
+      val pos = new l.Position(0, 0)
+      val range = new l.Range(pos, pos)
+      val location = new l.Location(pathStr, range)
```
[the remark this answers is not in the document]
kpbochenek: renamed

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,211 @@
...
+      val pathStr = pathFile.toString
+
+      pathFile.createNewFile()
+      val fw = new FileWriter(pathStr)
+      try {
+        fw.write(stacktrace)
+      } finally {
+        fw.close()
+      }
+      val pos = new l.Position(0, 0)
```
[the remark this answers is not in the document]
kpbochenek: renamed

tests/unit/src/test/scala/tests/StacktraceParseSuite.scala
kpbochenek: changed

tests/unit/src/test/scala/tests/StacktraceParseSuite.scala
kpbochenek: changed

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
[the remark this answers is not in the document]
kpbochenek: changed

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -0,0 +1,211 @@
...
+      Some(makeHtmlCommandParams(stacktrace))
+    } else {
+      val path = workspace.resolve(Directories.stacktrace)
+      val pathFile = path.toFile
+      val pathStr = pathFile.toString
+
+      pathFile.createNewFile()
+      val fw = new FileWriter(pathStr)
+      try {
+        fw.write(stacktrace)
```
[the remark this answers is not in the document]
kpbochenek: updated and reverted `didOpen` workaround

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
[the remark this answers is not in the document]
kpbochenek: ok changed to pattern match

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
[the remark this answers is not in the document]
tgodzik: I just meant the map and flatMap below. Might have not been too clear :sweat_smile:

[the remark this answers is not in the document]
kpbochenek: yes but there is headOption in between which is important it has no corresponding action in for-comprehension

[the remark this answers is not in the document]
kpbochenek: toToplevelSymbol returns you a list of symbols that should be related to given stacktrace line.
What I want here is get first one that is found in semanticdb, get file, update position and return it.
I don't want to map every potential symbol to semanticdb one, update position for every one and in the end return first and drop others that I calculated

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
kpbochenek: it is mainly the same as you proposed, will mark it as resolved

tgodzik (approved):
LGTM! Thanks for taking the time to address my nitpicks!

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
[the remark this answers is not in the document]
tgodzik: Sure, I was about to write that it's fine to leave it as is.

PR #1967, opened 2020-08-02
Change implicits to Dotty syntax.

ckipp01:
While starting to poke around the Scala3 code I notice we have started to use the Dotty syntax for implicits, but only in some places. I'm sort of in favor of unifying the way we approach it. In Scala3 stuff I think it makes sense to be consistent and just fully switch over to the given/using syntax. What do you think?

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -13,6 +13,9 @@ To fix this problem:
 project.excludeFilters = [
   "test-workspace"
   "metals-bench/src/main/resources"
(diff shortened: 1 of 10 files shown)


ckipp01:
The `NewProjectLspSuite` is failing because the scalatest g8 template was updated. I have a fix included in #1968 for this.

ckipp01:
The tests still ran against the old version of main, however, the issue with the new-project is fixed in main. Merging.

PR #1968, opened 2020-08-02
Bump recommended scalafmt version and local scalafmt version to 2.6.4

ckipp01:


tests/unit/src/test/scala/tests/NewProjectLspSuite.scala
```diff
@@ -54,7 +54,7 @@ class NewProjectLspSuite extends BaseLspSuite("new-project") {
         |}
         |
         |/$name/src/test/scala/CubeCalculatorTest.scala
-        |class CubeCalculatorTest extends org.scalatest.FunSuite {
+        |class CubeCalculatorTest extends org.scalatest.funsuite.AnyFunSuite {
```
ckipp01, line 57: Not exactly related to this pr, but they must have updated the scalatest g8 template. I generated it locally and indeed this is now what it generates and it was failing the tests.

tgodzik: Thanks! It was a pretty stable template previously, we could change to the scalameta templates if this happens again.

Issue #1675, opened 2019-01-24
Scalafix in the editor


gabro:
### Is your feature request related to a problem? Please describe.
Scalafix is a tool for automatic refactors and linting rules.

Currently, Scalafix is primarily used in batch mode, fixing or linting an entire project.

However, Scalafix would be extremely useful in the editor as well. Linting rules that are failing could be reported inline and some rules can be translated to code actions (i.e. automatic fixes).

### Describe the solution you'd like
Integration of Scalafix in Metals, similarly to how the Scalafmt integration currently works.

Ideally the integration would use https://github.com/scalacenter/scalafix/blob/master/scalafix-interfaces/src/main/java/scalafix/interfaces/Scalafix.java

I would envision this to be broken down into at least two iterations:

- linting only (just report diagnostics for failing rules)
- code actions (translate Scalafix rules into LSP Code Actions, to provide automatic fixes in the editor)

### Describe alternatives you've considered
Being sad and not having Scalafix in the editor

### Search terms
scalafix metals linting refactor quickfix organize imports

perrin4869:
Is there a way to trigger scalafix rules other than `OrganizeImports` from metals at this point?

ckipp01:
Hey @perrin4869, at this point there is not.

PR #1971, opened 2020-08-05
Implement organize import using scalafix

mlachkar:
This PR request the new interfaces of Scalafix. The review is still on-going on scalafix side. 
It's a draft, some values are hard-coded. Only working for scala 2.12 projects.
The default [organize config](https://github.com/liancheng/scalafix-organize-imports) is used for now, which is:
```
OrganizeImports {
  coalesceToWildcardImportThreshold = 2147483647 # Int.MaxValue
  expandRelative = false
  groupExplicitlyImportedImplicitsSeparately = false
  groupedImports = Explode
  groups = ["re:javax?\\.", "scala.", "*"]
  importSelectorsOrder = Ascii
  importsOrder = Ascii
  removeUnused = true
}
```
We can think about taking the config found in .scalafix.conf if it exists, if not we will need to choose some default config for organize import code action. 

This is how it works (a shortcut is already bound to organize import and works)
hello-1596644426584

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -413,8 +413,6 @@ lazy val metals = project
       "org.scalameta" % "mdoc-interfaces" % V.mdoc,
       "org.scalameta" %% "scalafmt-dynamic" % V.scalafmt,
       "ch.epfl.scala" % "scalafix-interfaces" % V.scalafix,
-      "ch.epfl.scala" % "scalafix-cli_2.12.12" % V.scalafix,
-      "ch.epfl.scala" % "scalafix-core_2.12" % V.scalafix,
       // For reading classpaths.
       // for fetching ch.epfl.scala:bloop-frontend and other library dependencies
       "io.get-coursier" % "interface" % V.coursierInterfaces,
--- a/build.sbt
+++ b/build.sbt
@@ -413,6 +413,8 @@ lazy val metals = project
       "org.scalameta" % "mdoc-interfaces" % V.mdoc,
       "org.scalameta" %% "scalafmt-dynamic" % V.scalafmt,
       "ch.epfl.scala" % "scalafix-interfaces" % V.scalafix,
+      "ch.epfl.scala" % "scalafix-cli_2.12.12" % V.scalafix,
+      "ch.epfl.scala" % "scalafix-core_2.12" % V.scalafix,
       // For reading classpaths.
       // for fetching ch.epfl.scala:bloop-frontend and other library dependencies
       "io.get-coursier" % "interface" % V.coursierInterfaces,
--- a/build.sbt
+++ b/build.sbt
@@ -51,7 +51,7 @@ inThisBuild(
       "-target:jvm-1.8",
       "-Yrangepos"
     ) ::: scala212CompilerOptions,
-    scalafixDependencies += "com.github.liancheng" %% "organize-imports" % "0.4.0",
+    scalafixDependencies += "com.github.liancheng" %% "organize-imports" % V.organizeImportRule,
     organization := "org.scalameta",
     licenses := Seq(
       "Apache-2.0" -> url("http://www.apache.org/licenses/LICENSE-2.0")
@@ -186,6 +186,7 @@ lazy val V = new {
   val mdoc = "2.2.9"
   val scalafmt = "2.6.4"
   val munit = "0.7.12"
+  val scalafix = "0.9.20"
   // List of supported Scala versions in SemanticDB. Needs to be manually updated
   // for every SemanticDB upgrade.
   def supportedScalaBinaryVersions =
@@ -220,6 +221,7 @@ lazy val V = new {
   val coursierInterfaces = "0.0.25"
   val ammonite = "2.2.0-4-4bd225e"
   val mill = "0.8.0"
+  val organizeImportRule = "0.4.0"
 }
 
 val genyVersion = Def.setting {
@@ -410,6 +412,7 @@ lazy val metals = project
       "org.scala-lang.modules" %% "scala-java8-compat" % "0.9.1",
       "org.scalameta" % "mdoc-interfaces" % V.mdoc,
       "org.scalameta" %% "scalafmt-dynamic" % V.scalafmt,
+      "ch.epfl.scala" % "scalafix-interfaces" % V.scalafix,
       // For reading classpaths.
       // for fetching ch.epfl.scala:bloop-frontend and other library dependencies
       "io.get-coursier" % "interface" % V.coursierInterfaces,
@@ -444,6 +447,7 @@ lazy val metals = project
       "semanticdbVersion" -> V.semanticdb,
       "scalafmtVersion" -> V.scalafmt,
       "ammoniteVersion" -> V.ammonite,
+      "organizeImportVersion" -> V.organizeImportRule,
       "millVersion" -> V.mill,
       "supportedScalaVersions" -> V.supportedScalaVersions,
       "supportedScala2Versions" -> V.scala2Versions,
--- a/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
@@ -1,5 +1,6 @@
 package scala.meta.internal.metals
 
+import java.io.File
 import java.net.URLClassLoader
 import java.nio.file.Path
 import java.util.ServiceLoader
@@ -9,6 +10,8 @@ import scala.collection.concurrent.TrieMap
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.pc.ScalaPresentationCompiler
 import scala.meta.internal.worksheets.MdocClassLoader
+import scala.meta.io.AbsolutePath
+import scala.meta.io.Classpath
 import scala.meta.pc.PresentationCompiler
 
 import ch.epfl.scala.bsp4j.ScalaBuildTarget
@@ -73,6 +76,21 @@ final class Embedded(
     )
   }
 
+  def organizeImports(
+      scalaBinaryVersion: String,
+      scalafixClassLoader: ClassLoader
+  ): URLClassLoader = {
+    val toolClasspathJars =
+      statusBar.trackBlockingTask("Downloading organize import rule") {
+        Embedded.organizeImportRule(scalaBinaryVersion)
+      }
+    toClassLoader(
+      Classpath(toolClasspathJars.map(jar => AbsolutePath(jar))),
+      scalafixClassLoader
+    )
+
+  }
+
   private def serviceLoader[T](
       cls: Class[T],
       className: String,
@@ -91,6 +109,13 @@ final class Embedded(
     }
   }
 
+  private def toClassLoader(
+      classpath: Classpath,
+      scalafixClassLoader: ClassLoader
+  ): URLClassLoader = {
+    val urls = classpath.entries.map(_.toNIO.toUri.toURL).toArray
+    new URLClassLoader(urls, scalafixClassLoader)
+  }
 }
 
 object Embedded {
@@ -214,6 +239,7 @@ object Embedded {
     downloadDependency(semanticdbScalacDependency(scalaVersion), scalaVersion)
   def downloadMtags(scalaVersion: String): List[Path] =
     downloadDependency(mtagsDependency(scalaVersion), scalaVersion)
+
   def downloadMdoc(
       scalaVersion: String,
       scalaBinaryVersion: String,
@@ -224,6 +250,20 @@ object Embedded {
       scalaVersion,
       resolution = resolutionParams
     )
+
+  def organizeImportRule(scalaBinaryVersion: String): List[File] = {
+    val dep = Dependency.of(
+      "com.github.liancheng",
+      s"organize-imports_$scalaBinaryVersion",
+      BuildInfo.organizeImportVersion
+    )
+    Fetch
+      .create()
+      .withDependencies(dep)
+      .fetch()
+      .asScala
+      .toList
+  }
 
   def newPresentationCompilerClassLoader(
       info: ScalaBuildTarget,
--- a/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
@@ -258,10 +258,10 @@ object Embedded {
 
   def toClassLoader(
       classpath: Classpath,
-      scalafixClassLoader: ClassLoader
+      classLoader: ClassLoader
   ): URLClassLoader = {
     val urls = classpath.entries.map(_.toNIO.toUri.toURL).toArray
-    new URLClassLoader(urls, scalafixClassLoader)
+    new URLClassLoader(urls, classLoader)
   }
 
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
@@ -80,7 +80,7 @@ final class Embedded(
       scalafixClassLoader: ClassLoader
   ): URLClassLoader = {
     val toolClasspathJars =
-        Embedded.organizeImportRule(scalaBinaryVersion)
+      Embedded.organizeImportRule(scalaBinaryVersion)
     toClassLoader(
       Classpath(toolClasspathJars.map(AbsolutePath(_))),
       scalafixClassLoader
--- a/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
@@ -9,7 +9,6 @@ import scala.collection.concurrent.TrieMap
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.pc.ScalaPresentationCompiler
 import scala.meta.internal.worksheets.MdocClassLoader
-import scala.meta.io.AbsolutePath
 import scala.meta.io.Classpath
 import scala.meta.pc.PresentationCompiler
 
@@ -75,19 +74,6 @@ final class Embedded(
     )
   }
 
-  def organizeImports(
-      scalaBinaryVersion: String,
-      scalafixClassLoader: ClassLoader
-  ): URLClassLoader = {
-    val toolClasspathJars =
-      Embedded.organizeImportRule(scalaBinaryVersion)
-    toClassLoader(
-      Classpath(toolClasspathJars.map(AbsolutePath(_))),
-      scalafixClassLoader
-    )
-
-  }
-
   private def serviceLoader[T](
       cls: Class[T],
       className: String,
@@ -106,13 +92,6 @@ final class Embedded(
     }
   }
 
-  private def toClassLoader(
-      classpath: Classpath,
-      scalafixClassLoader: ClassLoader
-  ): URLClassLoader = {
-    val urls = classpath.entries.map(_.toNIO.toUri.toURL).toArray
-    new URLClassLoader(urls, scalafixClassLoader)
-  }
 }
 
 object Embedded {
@@ -275,6 +254,14 @@ object Embedded {
     val parent =
       new PresentationCompilerClassLoader(this.getClass.getClassLoader)
     new URLClassLoader(allURLs, parent)
+  }
+
+  def toClassLoader(
+      classpath: Classpath,
+      scalafixClassLoader: ClassLoader
+  ): URLClassLoader = {
+    val urls = classpath.entries.map(_.toNIO.toUri.toURL).toArray
+    new URLClassLoader(urls, scalafixClassLoader)
   }
 
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
@@ -1,6 +1,5 @@
 package scala.meta.internal.metals
 
-import java.io.File
 import java.net.URLClassLoader
 import java.nio.file.Path
 import java.util.ServiceLoader
@@ -81,11 +80,9 @@ final class Embedded(
       scalafixClassLoader: ClassLoader
   ): URLClassLoader = {
     val toolClasspathJars =
-      statusBar.trackBlockingTask("Downloading organize import rule") {
         Embedded.organizeImportRule(scalaBinaryVersion)
-      }
     toClassLoader(
-      Classpath(toolClasspathJars.map(jar => AbsolutePath(jar))),
+      Classpath(toolClasspathJars.map(AbsolutePath(_))),
       scalafixClassLoader
     )
 
@@ -251,18 +248,13 @@ object Embedded {
       resolution = resolutionParams
     )
 
-  def organizeImportRule(scalaBinaryVersion: String): List[File] = {
+  def organizeImportRule(scalaBinaryVersion: String): List[Path] = {
     val dep = Dependency.of(
       "com.github.liancheng",
       s"organize-imports_$scalaBinaryVersion",
       BuildInfo.organizeImportVersion
     )
-    Fetch
-      .create()
-      .withDependencies(dep)
-      .fetch()
-      .asScala
-      .toList
+    downloadDependency(dep, scalaBinaryVersion)
   }
 
   def newPresentationCompilerClassLoader(
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -514,6 +514,7 @@ class MetalsLanguageServer(
     scalafixProvider = ScalafixProvider(
       buildTargets,
       buffers,
+      userConfig.scalafixConfigPath,
       workspace,
       embedded,
       statusBar,
@@ -1173,12 +1174,10 @@ class MetalsLanguageServer(
       params: DocumentFormattingParams
   ): CompletableFuture[util.List[TextEdit]] =
     CancelTokens.future { token =>
-      {
-        formattingProvider.format(
-          params.getTextDocument.getUri.toAbsolutePath,
-          token
-        )
-      }
+      formattingProvider.format(
+        params.getTextDocument.getUri.toAbsolutePath,
+        token
+      )
     }
 
   @JsonRequest("textDocument/onTypeFormatting")
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -511,8 +511,15 @@ class MetalsLanguageServer(
       compilers,
       definitionIndex
     )
-    scalafixProvider =
-      ScalafixProvider(buildTargets, buffers, workspace, embedded, statusBar)
+    scalafixProvider = ScalafixProvider(
+      buildTargets,
+      buffers,
+      workspace,
+      embedded,
+      statusBar,
+      clientConfig.icons(),
+      languageClient
+    )
     codeActionProvider = new CodeActionProvider(
       compilers,
       buffers,
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -207,6 +207,7 @@ class MetalsLanguageServer(
   private var debugProvider: DebugProvider = _
   private var symbolSearch: MetalsSymbolSearch = _
   private var compilers: Compilers = _
+  private var scalafixProvider: ScalafixProvider = _
   def loadedPresentationCompilerCount(): Int =
     compilers.loadedPresentationCompilerCount()
   var tables: Tables = _
@@ -510,9 +511,12 @@ class MetalsLanguageServer(
       compilers,
       definitionIndex
     )
+    scalafixProvider =
+      ScalafixProvider(buildTargets, buffers, workspace, embedded, statusBar)
     codeActionProvider = new CodeActionProvider(
       compilers,
-      buffers
+      buffers,
+      scalafixProvider
     )
     doctor = new Doctor(
       workspace,
@@ -637,7 +641,11 @@ class MetalsLanguageServer(
       if (initializeParams.supportsCodeActionLiterals) {
         capabilities.setCodeActionProvider(
           new CodeActionOptions(
-            List(CodeActionKind.QuickFix, CodeActionKind.Refactor).asJava
+            List(
+              CodeActionKind.QuickFix,
+              CodeActionKind.Refactor,
+              CodeActionKind.SourceOrganizeImports
+            ).asJava
           )
         )
       } else {
@@ -1158,10 +1166,12 @@ class MetalsLanguageServer(
       params: DocumentFormattingParams
   ): CompletableFuture[util.List[TextEdit]] =
     CancelTokens.future { token =>
-      formattingProvider.format(
-        params.getTextDocument.getUri.toAbsolutePath,
-        token
-      )
+      {
+        formattingProvider.format(
+          params.getTextDocument.getUri.toAbsolutePath,
+          token
+        )
+      }
     }
 
   @JsonRequest("textDocument/onTypeFormatting")
@@ -1514,6 +1524,7 @@ class MetalsLanguageServer(
         ammonite.stop()
       case ServerCommands.NewScalaProject() =>
         newProjectProvider.createNewProjectFromTemplate().asJavaObject
+
       case cmd =>
         scribe.error(s"Unknown command '$cmd'")
         Future.successful(()).asJavaObject
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -512,7 +512,6 @@ class MetalsLanguageServer(
       definitionIndex
     )
     scalafixProvider = ScalafixProvider(
-      buildTargets,
       buffers,
       userConfig.scalafixConfigPath,
       workspace,
@@ -524,6 +523,7 @@ class MetalsLanguageServer(
     codeActionProvider = new CodeActionProvider(
       compilers,
       buffers,
+      buildTargets,
       scalafixProvider
     )
     doctor = new Doctor(
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
@@ -1,0 +1,126 @@
+package scala.meta.internal.metals
+
+import java.nio.file.Path
+import java.util
+import java.util.Collections
+import java.util.Optional
+
+import scala.collection.mutable
+import scala.util.Try
+
+import scala.meta._
+import scala.meta.internal.metals.MetalsEnrichments._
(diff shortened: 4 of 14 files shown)


gabro:
This is great! I can’t wait to use this as soon as it’s merged :)

Quick question: what happens if the project already declares the same rule (possibly a different version of it) in their scalafixDependencies (like Metals itself does)?

In general I would try to make so that running the rule in batch from sbt or from the editor via Metals would have the same effect.

This is crucial if the projects wants to enforce the rule via a CI, for example.

This may mean:
- using the rule configured by the project (if any) and loading it otherwise
- using the project’s config (if any) or falling back to the default one (like you suggested)

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -492,9 +493,13 @@ class MetalsLanguageServer(
       statusBar,
       compilers
     )
+    scalafixProvider = ScalafixProvider
+      .from("2.12", buildTargets, buffers, workspace)
+      .get // TODO fail + scalafixBinary should be get from BuildTargets
```
mlachkar, line 498: How to get the scala version here? 
Or should I have many scalafix instances downloaded (lazily) and cached (for each scalaVersion) in ScalafixProvider. 
Right now the only way I found the get the scalaVersion, is when I am formatting a file => 
```
      identifier <- buildTargets.inverseSources(file)
      scalaBuildTarget <- buildTargets.scalaInfo(identifier)
      scalaVersion = scalaBuildTarget.getScalaVersion
``` 
Another question: `ScalafixProvider.from(` is a method that can fail, if it's not able to fetch the artifact. Should it fail ? should the fetch be done here? We can also not fail, and in this case, we disable the organize import until the next restart?

tests/unit/src/main/scala/tests/QuickBuild.scala
```diff
@@ -148,6 +148,7 @@ case class QuickBuild(
         List(
           List(
             "-Yrangepos",
+            "-Ywarn-unused-import", //TODO remove this
```
mlachkar, line 151: I didn't understand completely how QuickBuild works. Should I create a file next to my test containing the scalacOptions? I saw it was reading `metals.json` under directory target, but this one is generated.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: > How to get the scala version here?

That's exactly how you can get the version right now. We need the information from the build target. It also has `scalaBinaryVersion`, which will be useful here.
However, I think you should also have scalafix instances populated lazily for each scala version. Usually it will really be one per workspace.

As for `ScalafixProvider.from( ` we should handle the case it failed to download and show a message to the user in that case via `client.showMessage()`. We can retry downloading in case the user tries to organize imports again, no need to disable it as long as we send a clear message what is wrong.

tests/unit/src/main/scala/tests/QuickBuild.scala
tgodzik: If you look at `BaseCodeActionLspSuite` there is:
```scala
  _ <- server.initialize(s"""/metals.json
                                  |{"a":{}}
                                  |/$path
                                  |${input
          .replace("<<", "")
          .replace(">>", "")}
                                  |""".stripMargin)
```

you can add scalacOptions there. It would look something like:
```scala
  _ <- server.initialize(s"""/metals.json
                                  |{"a":{"scalacOptions": [""-Ywarn-unused-import""]}}
                                  |/$path
                                  |${input
          .replace("<<", "")
          .replace(">>", "")}
                                  |""".stripMargin)
```

tgodzik:
> This is great! I can’t wait to use this as soon as it’s merged :)
> 
> Quick question: what happens if the project already declares the same rule (possibly a different version of it) in their scalafixDependencies (like Metals itself does)?
> 
> In general I would try to make so that running the rule in batch from sbt or from the editor via Metals would have the same effect.
> 
> This is crucial if the projects wants to enforce the rule via a CI, for example.
> 
> This may mean:
> 
> * using the rule configured by the project (if any) and loading it otherwise
> * using the project’s config (if any) or falling back to the default one (like you suggested)

I am not sure there is a way to make sure that the version of the rules or scalafix itself is the same in sbt and in Metals. Can it be defined inside the `.scalafix` file? Alternatively, we could add settings for each version so that users can change them manually if needed, but hopefully it will only be needed if the rule is broken.

I think especially in case of rules available via code actions it's not really important to check whether they are enabled in the codebase, since they really only impact one file and as long as they work correctly the version doesn't really matter. And it will just  be a certain set of rules that we are sure to work OK.

Add an additional command that would run all defined rules on a single file, which would be most fragile in case of version mismatch and we should be careful there for sure, but single code actions I think are fine even if they doesn't work sometime, which really should not happen an is basically Metals bug. I am do not know probably enough about scalafix, so I might be missing something.

metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
```diff
@@ -0,0 +1,37 @@
...
+      ec: ExecutionContext
+  ): Future[Seq[l.CodeAction]] = {
+    val uri = params.getTextDocument.getUri
+    val path = uri.toAbsolutePath
+    val edits: List[TextEdit] = scalafixProvider.organizeImports(path)
+
+    val codeAction = new l.CodeAction()
+    codeAction.setTitle(OrganizeImports.title)
+    codeAction.setKind(l.CodeActionKind.SourceOrganizeImports)
+    codeAction.setEdit(
```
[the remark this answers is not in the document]
mlachkar: ok! I didn't know what's the difference between command and edit

tests/unit/src/main/scala/tests/QuickBuild.scala
mlachkar: ah !! thanks

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
mlachkar: ok!

metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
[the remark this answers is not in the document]
tgodzik: Command just specifies what should be run if the action is selected, while edit already defines what should be done. 

The problem with recalculating code actions on cursor move is something that is really hidden. We should use command if possible.

mlachkar:
Blocked by https://github.com/scalacenter/scalafix/pull/1223

gabro:
@tgodzik 
> I think especially in case of rules available via code actions it's not really important to check whether they are enabled in the codebase, since they really only impact one file and as long as they work correctly the version doesn't really matter. And it will just be a certain set of rules that we are sure to work OK.

I guess my broader point is: let's make sure that the rules applied in the editor by Metals provide a result consistent to the ones enforced by a CI check (e.g. using `sbt-scalafix`)

Otherwise you may end up in a tricky situation you organize imports using Metals' code action, but this makes the CI break.

Specifically talking about OrganizeImports, this can happen in a couple of scenarios:
- same version of the rule, but different configs (e.g. Metals uses the default one, while the CI check uses the one in `.scalafix.conf`
- different versions of the rule. This is subtler, as different version of the rule may have different defaults, or behave differently due to bug fixes or other breaking changes.

I think the first scenario can be covered by picking up the `.scalafix.conf` (as discussed previously), while the second scenario is harder to fix since there's not an easy way to introspect the Scalafix rule classpath (used for example by sbt-scalafix).

For selected rules such as OrganizeImports this may not be an issue: the breaking changes are probably not going to happen often, Metals can simply stay up-to-date and users will need to align to the same version used by Metals in case of inconsistent results.

What do you all think?

tgodzik:
> What do you all think?

I think we should take into account `/scalafix.conf` for sure and we will not be merging without it. As for versions inside `sbt` I think the only thing we can do is to allow customizing them inside user settings for anything that can be problematic, which probably in the beginning means the scalafix version itself.

gabro:
@tgodzik it would solve the issue for the Scalafix version, but not for the rule version.

Anyway, I agree using that taking `scalafix.conf` into account is already a good approximation that will likely cover the vast majority of scenarios 👍

mlachkar:
So it takes into account scalafix.conf if present. 
It's code edit operation, but it's only computed when codeAction organize import is called. 
metals-1599218463844

ayoub-benali:
Really cool feature ! would it be a similar workflow in other editors ? 
One would trigger code actions anywhere in the code and "organize imports" would show up ?

tgodzik:
> Really cool feature ! would it be a similar workflow in other editors ?
> One would trigger code actions anywhere in the code and "organize imports" would show up ?

It should be handled in most editors, but each might have a different way of invoking it. Organize imports is a special rule, so for example in VS Code it can be invoked via shortcut or via menu as shown in the gif. Not 100% sure how are different code actions handled elsewhere, but I imagine it will not sit together with all the other rules.

We might need to experiment with it a bit.

ayoub-benali:
the code actions menu can also be invoked via keyboard shortcut in sublime LSP. I am just not sure if it can be invoked at any location in the code or just the import part. I will try once this PR merged.

ckipp01:
> the code actions menu can also be invoked via keyboard shortcut in sublime LSP. I am just not sure if it can be invoked at any location in the code or just the import part. I will try once this PR merged.

@ayoub-benali So I think this will be pretty highly specific to the client implementation. The `CodeActionKind` that this is fulfilling (`source.organizeImports`) is a predefined one, so _in theory_, it should be an action that already exists. That's how it is in coc.nvim, I don't need to do anything except look through the already available editor commands.



Really cool seeing this @mlachkar :)

metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
```diff
@@ -0,0 +1,50 @@
...
+          Map(uri -> edits.asJava).asJava
+        )
+      )
+      Future.successful {
+        Seq(codeAction)
+      }
+    } else Future.successful(Seq())
+  }
+
+  private def isSourceOrganizeImportCalled(params: CodeActionParams): Boolean =
```
ckipp01, line 41: Good idea to have this 👍

ckipp01 (commented):
Really great job @mlachkar! Just a few sugguestions.

mlachkar:
I just found one issue: since it relays on semanticDb file, we need to save the file then generates a new semanticdb file before organizing imports.

ckipp01:
One other thing to think about is how to handle worksheets and ammonite scripts. Currently you can call this on them, and you see the following in the logs `error: No files to fix`. That message a bit unclear. Do you think we should filter out worksheets and ammonite scripts and note even run it on them?

tgodzik:
> I just found one issue: since it relays on semanticDb file, we need to save the file then generates a new semanticdb file before organizing imports.

We should use edit distance for that, similar to how we do it in other places. You can get it from buffers.

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -73,6 +75,21 @@ final class Embedded(
     )
   }
 
+  def organizeImports(
+      scalaBinaryVersion: String,
+      scalafixClassLoader: ClassLoader
+  ): URLClassLoader = {
+    val toolClasspathJars =
+      statusBar.trackBlockingTask("Downloading organize import rule") {
```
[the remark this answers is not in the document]
gabro: I wonder if this logic could be centralized in `trackBlockingTask`. It seems a fairly general case

ckipp01:
Not sure why I can't add it via the linked issues on the sidebar, but just commenting to link this to #1675

kpbochenek:
Works for emacs out of the box :+1: 

organize-imports-emacs

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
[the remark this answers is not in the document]
mlachkar: I will just remove the tracking for this task. It's a small artifact.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,126 @@
...
+        .withToolClasspath(urlClassLoaderWithExternalRule)
+        .withConfig(scalafixConf)
+        .withRules(List(organizeImportRuleName).asJava)
+        .withPaths(List(file.toNIO).asJava)
+        .withSourceroot(workspace.toNIO)
+        .withScalacOptions(Collections.singletonList(scalacOption))
+        .evaluate()
+    }
+
+    val fixedOpt = resOpt
```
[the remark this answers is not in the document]
mlachkar: It's not really patches, but the content file after scalafix is evaluated on this file. I will change names.

mlachkar:
- I added an error message in case of ammonite file or worksheet. 
- Unfortunately organize import will be able to run only on saved files. 

Need help for improving the popup message: it lasts too long .. and It would be nice if it disappears when the file is saved? 
Thanks

build.sbt
```diff
@@ -410,6 +412,9 @@ lazy val metals = project
       "org.scala-lang.modules" %% "scala-java8-compat" % "0.9.1",
       "org.scalameta" % "mdoc-interfaces" % V.mdoc,
       "org.scalameta" %% "scalafmt-dynamic" % V.scalafmt,
+      "ch.epfl.scala" % "scalafix-interfaces" % V.scalafix,
+      "ch.epfl.scala" % "scalafix-cli_2.12.12" % V.scalafix,
+      "ch.epfl.scala" % "scalafix-core_2.12" % V.scalafix,
```
gabro, line 417: is there a specific reason to manually specify the Scala version rather than using `%%`?

Potentially this could be:

```suggestion
      "ch.epfl.scala" %% "scalafix-cli" % V.scalafix cross CrossVersion.full,
      "ch.epfl.scala" %% "scalafix-core" % V.scalafix,
```

but I might be missing some context

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+        s"Save ${file.toNIO.getFileName} before organizing imports"
+      )
+      Nil
+    } else {
+      val baseMatcher = scalafix.internal.v1.Args.baseMatcher
+      if (baseMatcher.matches(file.toNIO)) {
+        val scalafixConfPath = workspace.resolve(scalafixFileName)
+        val scalafixConf: Optional[Path] =
+          if (scalafixConfPath.isFile) Optional.of(scalafixConfPath.toNIO)
+          else Optional.empty()
```
gabro, line 48: Nit: in other places we generally use the Scala type and then convert to Java types on demand.

```suggestion
        val scalafixConf =
          if (scalafixConfPath.isFile) Some(scalafixConfPath.toNIO)
          else None
```

and then `scalafixConf.asJava` where needed

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+      )
+  }
+}
+
+object ScalafixProvider {
+  type ScalaBinaryVersion = String
+  type ScalaVersion = String
+
+  val organizeImportRuleName = "OrganizeImports"
+  val scalafixFileName = ".scalafix.conf"
```
gabro, line 145: Not sure how relevant this is, but the `.scalafmt.conf` file is configurable. See `scalafmtConfigPath` in `UserConfiguration`. I would instinctively expect this to be configurable as well.

What do you think?

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
```diff
@@ -0,0 +1,112 @@
...
+       |OrganizeImports.groups = [
+       |  "scala."
+       |  "re:javax?\\."
+       |  "*"
+       |]
+       |
+       |""".stripMargin
+
+  check(
+    "basic organize imports",
```
gabro, line 33: we generally keep test names short and without spaces, so that they're easier to target with a test filter. So something like:

```suggestion
    "basic",
```

and so on

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
```diff
@@ -0,0 +1,112 @@
...
+      |  val optional = Optional.empty()
+      |}
+      |""".stripMargin.replace("'", "\""),
+    kind = List(kind),
+    scalafixConf = scalafixConf,
+    scalacOptions = scalacOption
+  )
+
+  check(
+    "should not work on worksheet or Ammonite files",
```
gabro, line 90: Is this test documenting a missing feature or do we actively want to prevent this working in worksheets/ammonite files?

If that's the latter, then 👍 , otherwise I would suggest using the `.fail`.

It would look like

```scala
check("test-name".fail)
```

The main advantage of this is that we can still document the expected behavior, and explicitly say that this isn't working yet.

gabro (changes requested):
Amazing work @mlachkar! 😍
As far as I can tell, functionality-wise this looks really solid.

I've left a few "stylistic" comments, most of which are non-blocking

build.sbt
mlachkar: Thanks for noticing this. Those libs were added only for debugging purposes and should be removed. Thanks.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+    icons: Icons,
+    languageClient: MetalsLanguageClient
+) {
+  import ScalafixProvider._
+  private val scalafixCache = mutable.Map.empty[String, Scalafix]
+
+  def organizeImports(file: AbsolutePath): List[l.TextEdit] = {
+    val fileInput = file.toInput
+    val unsavedFile = file.toInputFromBuffers(buffers)
+    if (fileInput.text.diff(unsavedFile.text).nonEmpty) {
```
tgodzik, line 36: Why use `diff` instead of just equals or calculating md5 from something like:
```scala
def md5(text : String) = {
  val digest = MessageDigest.getInstance("MD5")
  digest.update(text)
  MD5.bytesToHex(digest.digest())
}
```

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+      scribe.info(s"Organize imports requires saving the file first")
+      languageClient.showMessage(
+        MessageType.Warning,
+        s"Save ${file.toNIO.getFileName} before organizing imports"
+      )
+      Nil
+    } else {
+      val baseMatcher =
+        FileSystems.getDefault.getPathMatcher("glob:**.{scala,sbt}")
+      if (baseMatcher.matches(file.toNIO)) {
```
tgodzik, line 46:
```suggestion
      if (file.isScala) {
```

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+      } else {
+        scribe.info(
+          s"""|Could not organize import for ${file.toNIO.getFileName}. Should end with .scala or .sbt"""
+        )
+        Nil
+      }
+    }
+  }
+
+  private def getTextEditsFrom(
```
tgodzik, line 88:
```suggestion
  private def textEditsFrom(
```

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+    } else {
+      Nil
+    }
+  }
+
+  private def getScalaVersionAndClassPath(
+      file: AbsolutePath
+  ): Option[(ScalaVersion, ScalaBinaryVersion, util.List[Path])] =
+    for {
+      identifier <- buildTargets.inverseSources(file)
```
tgodzik, line 104:
```suggestion
      buildId <- buildTargets.inverseSources(file)
```

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+      scalacOptions <- buildTargets.scalacOptions(identifier)
+      scalaBuildTarget <- buildTargets.scalaInfo(identifier)
```
tgodzik, line 106:
```suggestion
      scalaBuildTarget <- buildTargets.scalaTarget(identifier)
```
this should give you a nicer api overall.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+      classPath = scalacOptions.getClasspath.map(_.toAbsolutePath.toNIO)
+      _ = classPath.add(semanticdbTarget)
+    } yield (scalaVersion, scalaBinaryVersion, classPath)
+
+  private def configureApi(
+      api: Scalafix,
+      scalaVersion: ScalaVersion,
+      classPath: util.List[Path]
+  ): ScalafixArguments = {
+    api
```
tgodzik, line 119: This is only used once, we could move it to where it's invoked.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+      .newArguments()
+      .withScalaVersion(scalaVersion)
+      .withClasspath(classPath)
+  }
+
+  private def getOrUpdateScalafixCache(
```
tgodzik, line 125:
```suggestion
  private def scalafixCache(
```

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,147 @@
...
+      scalaBinaryVersion: ScalaBinaryVersion
+  ): Option[Scalafix] = {
+    scalafixCache
+      .get(scalaBinaryVersion)
+      .orElse(
+        statusBar.trackBlockingTask("Downloading scalafix") {
+          Try(Scalafix.fetchAndClassloadInstance(scalaBinaryVersion)).toOption
+            .map { api =>
+              scalafixCache.update(scalaBinaryVersion, api)
```
tgodzik, line 134: Should we not use concurrent map here? What if someone quickly runs optimize imports a couple of times? We should wait and not run additional times

tests/unit/src/test/scala/tests/codeactions/BaseCodeActionLspSuite.scala
```diff
@@ -15,21 +15,28 @@ abstract class BaseCodeActionLspSuite(suiteName: String)
...
     test(name) {
       cleanWorkspace()
       for {
         _ <- server.initialize(s"""/metals.json
-                                  |{"a":{}}
+                                  |{"a":{$scalacOptionsJson}}
+                                  |$scalafixConf
                                   |/$path
                                   |${input
           .replace("<<", "")
```
tgodzik, line 35: let's maybe moved this outside. It breaks the formatting quite weirdly.

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
```diff
@@ -0,0 +1,113 @@
...
+    """
+      |package a
+      |import scala.concurrent.Future
+      |import scala.concurrent.duration._
+      |
+      |object A {
+      |  val d = Duration(10, MICROSECONDS)
+      |  val k = Future.successful(1)
+      |}
+      |""".stripMargin.replace("'", "\""),
```
tgodzik, line 55: is this needed?
```suggestion
      |""".stripMargin,
```

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
```diff
@@ -0,0 +1,113 @@
...
+      |package a
+      |import scala.concurrent.duration._
+      |
+      |import java.util.Optional
+      |
+      |object A {
+      |  val d = Duration(10, MICROSECONDS)
+      |  val optional = Optional.empty()
+      |}
+      |""".stripMargin.replace("'", "\""),
```
tgodzik, line 83: Same here:
```suggestion
      |""".stripMargin,
```

tests/unit/src/test/scala/tests/codeactions/BaseCodeActionLspSuite.scala
```diff
@@ -15,21 +15,28 @@ abstract class BaseCodeActionLspSuite(suiteName: String)
       expectedActions: String,
       expectedCode: String,
       selectedActionIndex: Int = 0,
-      expectNoDiagnostics: Boolean = true
+      expectNoDiagnostics: Boolean = true,
+      fileName: Option[String] = None,
```
tgodzik, line 19:
```suggestion
      fileName: String = "A.scala",
```
should simplify it a bit.

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
```diff
@@ -0,0 +1,113 @@
...
+      |""".stripMargin,
+    s"${OrganizeImports.title}",
+    """
+      |import scala.concurrent.duration._
+      |import java.util.Optional
+      |
+      |object A {
+      | println("")
+      |}
+      |""".stripMargin.replace("'", "\""),
```
tgodzik, line 108:
```suggestion
      |""".stripMargin,
```

tgodzik (changes requested):
Thanks for the great work! I have a couple of comments, but nothing major. Great work with figuring the issue with unsaved files! That would have probably gone unnoticed.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
mlachkar: Weirdly the equality fails on windows.

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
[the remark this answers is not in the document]
mlachkar: We removed the tracking for this task. It's too small and it's just distracting. Actually there were other ideas, but I just decided to remove the tracking.https://github.com/scalameta/metals/pull/1971#discussion_r483623670

What would you suggest ?

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
mlachkar: And then for computing md5, you still need to read the file (to get the String). So I thought from a performance point of view, it's the same? 
I will implement this with md5.

tgodzik: Not 100% sure which one is faster, but md5 is used already to check if a file changed and diff seems a bit more involved.

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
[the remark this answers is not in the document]
tgodzik: Ach, that comment was an old one, but maybe we could just the status bar notification? It's not too distracting.

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
mlachkar: No, you are right. Thanks.. Copied from the wrong test ^^

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
mlachkar: so md5 is not working on windows ...

tgodzik: That's super weird :/ I will try out later at home on windows.

mlachkar: Thanks

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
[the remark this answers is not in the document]
mlachkar: It was already a status bar. no ?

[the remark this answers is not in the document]
tgodzik: Hmm... I think one method is just showing things in the bottom, while the long task shows a pop up in the right hand corner. `trackFuture` might be the one to use here.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,154 @@
...
+  private def scalaVersionAndClasspath(
+      file: AbsolutePath
+  ): Option[(ScalaVersion, ScalaBinaryVersion, util.List[Path])] =
+    for {
+      identifier <- buildTargets.inverseSources(file)
+      scalacOptions <- buildTargets.scalacOptions(identifier)
+      scalaBuildTarget <- buildTargets.scalaInfo(identifier)
+      scalaVersion = scalaBuildTarget.getScalaVersion
+      semanticdbTarget = scalacOptions.targetroot(scalaVersion).toNIO
+      scalaBinaryVersion = scalaBuildTarget.getScalaBinaryVersion
```
tgodzik, line 111: We should print a warning in case of Scala 3 versions, otherwise I think it will throw errors out.

tgodzik:
We should add a new row maybe to https://scalameta.org/metals/docs/editors/overview.html#additional-file-types so that it's documented where organize imports works.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
mlachkar: Forgot about scala 3..

mlachkar: It wont throw an exception, since it's cached in a try, but I will add an error message

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,184 @@
...
+    val unsavedFile = file.toInputFromBuffers(buffers)
+    if (isUnsaved(unsavedFile.text, fileInput.text)) {
+      scribe.info(s"Organize imports requires saving the file first")
+      languageClient.showMessage(
+        MessageType.Warning,
+        s"Save ${file.toNIO.getFileName} before organizing imports"
+      )
+      Nil
+    } else {
+      if (Seq("scala", "sbt").contains(file.`extension`)) {
```
tgodzik, line 46:
```suggestion
      if (Seq("scala", "sbt").contains(file.extension)) {
```

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,184 @@
...
+        val scalafixConfPath = workspace.resolve(scalafixConfigPath)
+        val scalafixConf =
+          if (scalafixConfPath.isFile) Some(scalafixConfPath.toNIO)
+          else None
+        val scalafixEvaluation = for {
+          (scalaVersion, scalaBinaryVersion, classPath) <-
+            scalaVersionAndClasspath(file)
+          _ <-
+            if (!ScalaVersions.isScala3Version(scalaVersion)) Some(())
```
tgodzik, line 55: Could we move it to `scalaVersionAndClasspath` there would be no need for `Some(())`. We can return None there if the scalaversion is wrong.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -0,0 +1,184 @@
...
+            .evaluate()
+        }
+
+        val newFileContentOpt = scalafixEvaluation
+          .flatMap(result => result.getFileEvaluations.headOption)
+          .flatMap(_.previewPatches().asScala)
+        newFileContentOpt.map(textEditsFrom(_, fileInput)).getOrElse(Nil)
+      } else {
+        scribe.info(
+          s"""|Could not organize import for ${file.toNIO.getFileName}. Should end with .scala or .sbt"""
```
tgodzik, line 89: This should not be needed if we check the file extension while contributing the code action

metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
```diff
@@ -0,0 +1,50 @@
...
+import org.eclipse.{lsp4j => l}
+
+final class OrganizeImports(scalafixProvider: ScalafixProvider)
+    extends CodeAction {
+  override def kind: String = OrganizeImports.kind
+
+  override def contribute(params: CodeActionParams, token: CancelToken)(implicit
+      ec: ExecutionContext
+  ): Future[Seq[l.CodeAction]] = {
+    if (isSourceOrganizeImportCalled(params)) {
```
tgodzik, line 22: Could we check here if the file ends with `.scala` or `.sbt` ?

I don't think we need to show warnings worksheets or ammonite scripts. Code action should not be added in that case.

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
tgodzik: This was not changed, did you push the code changes?

tgodzik (changes requested):
I think we are almost good to merge, I added some comments and there are still some that you missed.

Again, thanks for the great work!

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
mlachkar: ok ! Done.

metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
mlachkar: ok

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
mlachkar: ok.. But It should be noted somewhere that we expect files to be .scala or .sbt ... as a comment ?

tgodzik: We can leave it as a comment in the code action, but we also document it in the docs, so I think it should be fine

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
mlachkar: We decided to keep TrieMap after discussing it.

tgodzik (approved):
LGTM! We can merge it and test it out properly while it's in snapshot :man_dancing:

gabro (approved):
🎉🙌 great job @mlachkar, really excited to give this a try!

PR #1972, opened 2020-08-05
Adjust locations returned from the presentation compiler 

tgodzik:
Couple of issues I noticed only after merging the sbt support PR.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/AdjustLspData.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/AdjustLspData.scala
@@ -5,6 +5,7 @@ import scala.meta.pc.AutoImportsResult
 
 import org.eclipse.lsp4j.CompletionList
 import org.eclipse.lsp4j.Hover
+import org.eclipse.lsp4j.Location
 import org.eclipse.lsp4j.Position
 import org.eclipse.lsp4j.{Range => LspRange}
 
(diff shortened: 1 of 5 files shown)


tgodzik:
@dos65 could you take a look at the PR?

metals/src/main/scala/scala/meta/internal/metals/Warnings.scala
```diff
@@ -76,7 +76,7 @@ final class Warnings(
     }
     isReported match {
       case Some(()) =>
-      case None =>
+      case None if !path.isSbt =>
```
tgodzik, line 79: Otherwise we get warnings that are not neccessary here.

dos65: I started thinking to suppress these warnings for worksheets too.
But they produce warnings at a different place here:
https://github.com/scalameta/metals/blob/21fa274e18a1547ef511c41f3e036e03a89e5a00/metals/src/main/scala/scala/meta/internal/metals/Warnings.scala#L70-L73

Maybe, it would be better to move `BuildTargets.sbtBuildTarget` logic into `BuildTargets.inverseSources` as you orinially proposed and move this check at line 70 for `.sbt` and `.worksheet.sc`?

tgodzik: Sure, fixed both!

metals/src/main/scala/scala/meta/internal/metals/AdjustLspData.scala
```diff
@@ -51,7 +61,12 @@ object AdjustLspData {
 
   def create(f: Position => Position): AdjustLspData =
     new AdjustLspData {
-      override def adjustPos(pos: Position): Position = f(pos)
+      override def adjustPos(pos: Position): Position = {
```
tgodzik, line 64: This failed for any additional edits that were put into line 0.

Issue #1975, opened 2020-08-06
Renaming a class also renames the macro annotation above it


felher:
**Describe the bug**

Given the following code: 
```scala
package org.felher.renamemacro

import io.circe.generic.JsonCodec

@JsonCodec
final case class Main2(name: String)
```

Renaming `Main2` `Main` also renames `@JsonCodec` to `@Main`

**To Reproduce**

Clone https://github.com/felher/metalsrenamemacro .

1. Clone https://github.com/felher/metalsrenamemacro
2. Open src/main/scala/org/felher/renamemacro/Main2.scala
3. Rename Main2

**Expected behavior**

The annotation should not have been renamed.

**Installation:**
- Operating system: Ubuntu Linux
- Editor: Vim with coc-metals
- Metals version: v0.9.2

**Additional Information**
Might be related to #1095

PR #1976, opened 2020-08-06
Add checking exact name to all references() invocations

tgodzik:
Fixes https://github.com/scalameta/metals/issues/1975

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
@@ -314,7 +314,7 @@ final class ReferenceProvider(
     val realName =
       if (isBackticked) name.substring(1, name.length() - 1)
       else name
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -314,7 +314,7 @@ final class ReferenceProvider(
     val realName =
       if (isBackticked) name.substring(1, name.length() - 1)
       else name
-    if (symbol.isLocal || symbol.contains(realName)) {
+    if (symbol.isLocal || symbol.desc.name.toString == realName) {
```
tgodzik, line 317: This is because I figured contains is not good enough here. And this is only used in case of rename, which doesn't make references slower and we are fine with some flaky results there, but we do not want to break the code when renaming.

tgodzik:
JDK 8 tests are running OOM again - need to figure out what's happening there :(

PR #1977, opened 2020-08-06
Remove excluding packages

dr0i:
This is a proposal to solve https://github.com/scalameta/metals-feature-requests/issues/145.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/metals/CompressedPackageIndex.scala
+++ b/mtags/src/main/scala/scala/meta/internal/metals/CompressedPackageIndex.scala
@@ -29,26 +29,6 @@ case class CompressedPackageIndex(
 }
 
 object CompressedPackageIndex {
-  def isExcludedPackage(pkg: String): Boolean = {
-    // NOTE(olafur) At some point we may consider making this list configurable, I can
-    // imagine that some people wouldn't mind excluding more packages or including for
-    // example javax._.
-    pkg.startsWith("META-INF/") ||
-    pkg.startsWith("images/") ||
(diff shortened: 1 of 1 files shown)


tgodzik:
Thanks for the contribution! This could probably be instead included via UserConfiguration with the default being the current excluded packages. Other opinions?

ckipp01:
> Thanks for the contribution! This could probably be instead included via UserConfiguration with the default being the current excluded packages. Other opinions?

Yea I think this would be great. Keep the existing ones as defaults with option to remove/add more.

dr0i:
I hope that it will be then made very clear that there are libraries excluded because from my point of view the behaviour atm is irritating (and contradicting, since the excluded libraries are listed in the metals view in vscode).

gabro:
Hey @dr0i, thanks for the contribution. I too think this should be a configurable behavior.
The current default got us very far (since you're the first one raising the issue in over 2 years) but it makes sense for users to configure this bit.

> and contradicting, since the excluded libraries are listed in the metals view in vscode

On this I only partially agree: what we show in the completions isn't necessarily all that is available in the classpath: completions need to be quick and relevant, and they are a very different interaction from - say - exploring the classpath (which can be slower, but needs to be precise)

> the behaviour atm is irritating

I see it may not be optimal for your use case. Only one note: let's try to keep a positive tone. You're raising a valid issue, but fortunately it's not a complicated feature to implement :-)

ckipp01:
I'm going to go ahead and close this. Feel free to take a peak at #2012 which is more in line with the approach that we'd like to take for this.

PR #1978, opened 2020-08-07
Simplify logic to add metals.sbt files and make sure it's tested

tgodzik:
Previously, we wouldn't always generate `metals.sbt` needed for the Bloop plugin to generate `.bloop` configuration. The logic especially broke if the `project` directory didn't exist. Now, we make sure that there is a corresponding `metals.sbt` file for each level of *.sbt files excluding `metals.sbt` themselves.

To clarify what's going on:
- `project/metals.sbt` is needed for the `.bloop` config used for Scala files
- `project/project/metals.sbt` is needed for the `project/.bloop` config used for `*.sbt` files and `project/*.scala` files
- `project/project/project/metals.sbt` is needed for the `project/project/.bloop` config used for `project/*.sbt` files and `project/project/*.scala` files
- etc.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
@@ -56,7 +56,7 @@ case class SbtBuildTool(
         ).flatten
     }
     removeLegacyGlobalPlugin()
-    writeSbtMetalsPlugin(workspace)
+    writeSbtMetalsPlugins(workspace)
     allArgs
   }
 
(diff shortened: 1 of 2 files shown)


tgodzik:
@dos65 I finally understood what was happening and fixed up the logic a bit.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -83,42 +83,36 @@ case class SbtBuildTool(
...
+    ): Set[AbsolutePath] = {
+      if (meta.exists) {
+        val files = meta.list.toList
         val hasSbtSrc = files.exists(f => f.isSbt && f.filename != "metals.sbt")
-        val goNext = hasScalaSrc || hasSbtSrc || parentHasSbtSrc
-
-        if (goNext)
-          sbtMetaDirs(next, hasSbtSrc, dir :: acc)
-        else
+        if (hasSbtSrc) {
```
dos65, line 97: I think it should check the existence for `.scala` files too, not only `.sbt`.

However, it's probably it's hard to find a project that has sbt-scala sources somewhere deeper than in `.project/` directory

tgodzik: We are not really registering those files within the correct build target, I will fix up together later.

Issue #1981, opened 2020-08-09
Goto Definition not working as expected in sbt files.


ckipp01:
**Describe the bug**
Executing a goto definition on a `SettingKey` doesn't seem to ever bring you to the right definition. The hover on a specific key works fine, but the definition always is off.

**To Reproduce**
1. Create a new `Hello World` project with the New Scala Project functionality
2. Open the `build.sbt` and do a `BuildImport`.
3. Execute a goto definition on any `SettingKey`

**Expected behavior**
I would expect this to go to the right `SettingKey` definition site.

**Screenshots**
Notice that all three times I attempt this, they are all wrong.

![2020-08-09 20 11 33](https://user-images.githubusercontent.com/13974112/89738917-abb83f00-da7c-11ea-89e9-8c78dcb44611.gif)

**Installation:**
- Operating system: macOS
- Editor: Vim
- Metals version: `0.9.2+81-63198fd6-SNAPSHOT`

**Search terms**
goto definitionn in sbt

tgodzik:
Thanks for finding this! I think I know what is going on. I will be tidying up things this week for the release so I will fix this and the documentation issues.

PR #1984, opened 2020-08-10
Fix go to defintion for Sbt's Scala sources

tgodzik:
Fixes https://github.com/scalameta/metals/issues/1981

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/AdjustLspData.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/AdjustLspData.scala
@@ -41,12 +41,7 @@ trait AdjustLspData {
 
   def adjustLocations(
       locations: java.util.List[Location]
-  ): ju.List[Location] = {
-    locations.asScala.map { loc =>
-      loc.setRange(adjustRange(loc.getRange()))
-      loc
-    }.asJava
-  }
+  ): ju.List[Location]
 
   def adjustHoverResp(hover: Hover): Hover =
     if (hover.getRange == null)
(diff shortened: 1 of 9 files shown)


metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -488,9 +488,12 @@ class Compilers(
...
     val modifiedInput =
       originInput.copy(value = withOuter)
-    val adjustLspData = AdjustLspData.create(pos => {
-      new LspPosition(pos.getLine() - 1, pos.getCharacter() - ident.size)
-    })
+    val adjustLspData = AdjustLspData.create(
+      pos => {
+        new LspPosition(pos.getLine() - 1, pos.getCharacter() - ident.size)
+      },
+      filterOutLocations = { loc => loc.getUri().isWorksheet }
```
tgodzik, line 495: we shouldn't adjust positions outside of the worksheets.

metals/src/main/scala/scala/meta/internal/metals/AdjustLspData.scala
```diff
@@ -77,6 +72,24 @@ trait AdjustLspData {
   }
 }
 
+case class AdjustLspPos(
```
ckipp01, line 75: I normally don't love case classes that are verbs. What do you think of `AdjustedLspPos`?

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
ckipp01: 🤔 not work for now, but I'm wondering for clarity in the future if we should try to unify where we have adjustments done. For example we check here for worksheets and sbt and adjust accordingly. However, Ammonite also has some adjustments that we do, but those live in 	`Ammonite.scala`. I can see how that could be confusing from someone on the outside. Or, is it a good distinction to have that adjustments related to the PC are in `AdjustLspData` and other things live in their corresponding places? What do you think?

tgodzik: Yeah, I will move the adjustments to separate locations

metals/src/main/scala/scala/meta/internal/metals/AdjustLspData.scala
tgodzik: Fixed!

PR #1985, opened 2020-08-11
Add parallel tests

tgodzik:
Trying out to make the tests run faster via:
- splitting them to different shards based on duration available in the test reports
- adding Xmx to each fork
- limiting xmx used for sbt


kpbochenek (approved):
LGTM!

Also a few fixes in tests which are always great :+1:

We will see after multiple executions how it behaves but so far looks promising 💪

PR #1986, opened 2020-08-11
Don't run worksheets on Scala 0.25.0 or lower

tgodzik:


metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -291,18 +292,31 @@ class WorksheetProvider(
   }
 
   private def getMdoc(target: BuildTargetIdentifier): Option[Mdoc] = {
+
+    def isSupportedScala3Version(scalaVersion: String) = {
+      ScalaVersions.isScala3Version(scalaVersion) && SemVer.isCompatibleVersion(
+        "0.26.0",
```
ckipp01, line 298: Can we include a note either here or in the commit to why it has to be 0.26.0 or newer.

Issue #1667, opened 2020-04-27
Allow running main classes defined outside the build targets


chuanzhangjiang:
how can I debug my playframwork app in vs code?

tgodzik:
Thanks for reporting! From what I checked it's not currently possible to run play application from Metals. The play server main is not detected.

This might require some custom work, so I would suggest opening a feature request: https://github.com/scalameta/metals-feature-requests

Or alternatively upvoting https://github.com/scalameta/metals-feature-requests/issues/86
which would allow to connect to a running server.

chuanzhangjiang:
thanks!

dkovalenko:
Actually Bloop supports running play application, see: https://github.com/scalacenter/bloop/pull/717

Maybe

tgodzik:
@dkovalenko Actually you are right :O 

![image](https://user-images.githubusercontent.com/3807253/81423407-ad5c2900-9154-11ea-913f-06ee53474908.png)

The only problem is that we check if a given class is contained in the build target and if it is not we throw an exception. Maybe we could actually drop that check:
https://github.com/scalameta/metals/blob/5f2b935046e718409d633b1ed717c8319c1279fa/metals/src/main/scala/scala/meta/internal/metals/BuildTargetClassesFinder.scala#L56

dkovalenko:
Hi, just wondering if the fix was merged?
I still having trouble running play main using metals-vscode
<img width="396" alt="Screenshot 2020-07-27 at 13 12 25" src="https://user-images.githubusercontent.com/874918/88531125-09737280-d00b-11ea-965c-a18d98031fc3.png">
This is my configuration:
```
{
      "type": "scala",
      "request": "launch",
      "name": "Play main",
      "mainClass": "play.core.server.ProdServerStart",
      "args": [],
      "jvmOptions": [],
      "buildTarget": "rotator"
    }
```

tgodzik:
The fix was not yet merged. I will try to give it a go this week.

PR #1989, opened 2020-08-12
Allow running main classes from dependencies

tgodzik:
Previously, we would only allow running main classes defined inside workspace source, which information was supplied by Bloop. Now, we also try to find the class inside dependencies to allow running things like Play framework server:

play-run

Fixes https://github.com/scalameta/metals/issues/1667

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -2,6 +2,7 @@ package scala.meta.internal.metals
 
 import java.lang.{Iterable => JIterable}
 import java.net.URLClassLoader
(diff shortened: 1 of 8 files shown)


metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -52,8 +53,6 @@ final class BuildTargets(
     TrieMap.empty[AbsolutePath, BuildTargetIdentifier]
   private val isSourceRoot =
     ConcurrentHashSet.empty[AbsolutePath]
-  private val buildTargetInference =
```
tgodzik, line 55: This was not actually used anywhere. `sourceItemsToBuildTarget` is used for the same exact purpose.

Issue #1983, opened 2020-08-10
Update docs about sbt support.


ckipp01:
***Describe the Task***

Currently in the documentation we don't really have any information about sbt support. I know this isn't in a stable release yet, but we should make sure to add some information about it before that. I think there are a few _surprises_ people may have currently without some context. For example:

- What is supported and what is not (ex: completions, hover, goto definition supported)
- What else should be ignored now
 - This is relevant as I just had this happen to me where I sent in a pr, and realized I now had a `project/metals.sbt`, `project/project/metals.sbt`, and a `project/project/project/metals.sbt`. I didn't have the latter two ignored. Will these always be created? We should mention that they should be ignored as well.

tgodzik:
> This is relevant as I just had this happen to me where I sent in a pr, and realized I now had a project/metals.sbt, project/project/metals.sbt, and a project/project/project/metals.sbt. I didn't have the latter two ignored. Will these always be created? We should mention that they should be ignored as well.

They need to be created to generate `.bloop` config for the meta builds. I will add docs for it, best would be to just ignore `metals.sbt`

PR #1994, opened 2020-08-14
Update information about Sbt, Ammonite and worksheet support

tgodzik:
Resolves https://github.com/scalameta/metals/issues/1983

### DIFF
--- a/docs/editors/overview.md
+++ b/docs/editors/overview.md
@@ -13,7 +13,6 @@ specific sections.
 
 ```
 
-
 ## Editor support
 
 Metals works with the following text editors with varying degree of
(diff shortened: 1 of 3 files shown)


docs/editors/overview.md
```diff
@@ -403,6 +402,130 @@ Language Server Protocol (LSP). These extensions are not necessary for Metals to
...
+    <td align="center">Sbt scripts</td>
+    <td align="center">Worksheets</td>
+    <td align="center">Ammonite scripts</td>
+    <td align="center">Standalone Scala files</td>
+  </tr>
+</thead>
+<tbody>
+   <tr>
+    <td>Diagnostics</td>
+    <td align="center">Parsing errors</td>
```
ckipp01, line 423: Do you think it'd be a good idea to mark these with a `*` and then at the bottom of the table include an explanation of what `parsing errors` means here? I think it may not be super clear for people what exactly this means.

metals-docs/src/main/scala/docs/GenericModifier.scala
```diff
@@ -15,17 +15,20 @@ class GenericModifier extends StringModifier {
        |The Metals server places logs and other files in the `.metals` directory. The
        |Bloop compile server places logs and compilation artifacts in the `.bloop`
        |directory. Bloop plugin that generates Bloop configuration is added in the
```
ckipp01, line 17:
```suggestion
       |directory. The Bloop plugin that generates Bloop configuration is added in the 
```
I know you didn't change this line, but just spotted this.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -122,7 +122,7 @@ object UserConfiguration {
         "false",
         "false",
         "Don't generate Bloop plugin file for sbt",
-        "If true, Metals will not generate a `project/metals.sbt` file under the assumption that sbt-bloop is already manually installed in the sbt build. Build import will fail with a 'not valid command bloopInstall' error in case Bloop is not manually installed in the build when using this option."
+        "If true, Metals will not generate a `metals.sbt` files under the assumption that sbt-bloop is already manually installed in the sbt build. Build import will fail with a 'not valid command bloopInstall' error in case Bloop is not manually installed in the build when using this option."
```
ckipp01, line 125:
```suggestion
        "If true, Metals will not generate `metals.sbt` files under the assumption that sbt-bloop is already manually installed in the sbt build. Build import will fail with a 'not valid command bloopInstall' error in case Bloop is not manually installed in the build when using this option."
```

tgodzik: Fixed!

metals-docs/src/main/scala/docs/GenericModifier.scala
tgodzik: Thanks for noticing that!

docs/editors/overview.md
tgodzik: Changed to a tick with `*` with an explanation below.

PR #2004, opened 2020-08-17
Properly detect semanticdb target for Scala 3

### LINKED ISSUE #2003: Metals does not find semanticdb-target options for Scala 3
**Describe the bug**

In a Scala 3 project when the `semanticdb-target` option is set, Metals does not recognize it and it does not look for the semanticdb files in the right place.

**To Reproduce**

1. Create a Dotty project with sbt `1.4.0-M2`
2. Set `scalacOptions ++= Seq("-Ysemanticdb", "-semanticdb-target", "/home/adpi2/foo/target/meta")`
3. Start sbt
4. Open Metals

**Expected behavior**
When trying to navigate the code, Metals will print something like 
```
ERROR code navigation does not work for the file '/home/adpi2/foo/src/main/scala/foo/Main.scala' because the build target foo is missing the compiler option "-P:semanticdb:sourceroot:/home/adpi2/foo". To fix this problems, update the build settings to include this compiler option.
```
The expected behavior is that Metals extracts the semanticdb target folder from the compiler options and then it should be able to navigate the code.

**Installation:**
- Operating system: Linux
- Editor: Visual Studio Code

adpi2:
Fixes #2003

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -572,10 +572,20 @@ object MetalsEnrichments
         .map(uri => AbsolutePath(Paths.get(URI.create(uri))))
         .filter(p => Files.exists(p.toNIO))
     }
-    def targetroot: AbsolutePath = {
-      semanticdbFlag("targetroot")
-        .map(AbsolutePath(_))
-        .getOrElse(item.getClassDirectory.toAbsolutePath)
+    def targetroot(scalaVersion: String): AbsolutePath = {
+      if (ScalaVersions.isScala3Version(scalaVersion)) {
+        val options = item.getOptions.asScala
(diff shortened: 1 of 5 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -573,9 +573,17 @@ object MetalsEnrichments
         .filter(p => Files.exists(p.toNIO))
     }
     def targetroot: AbsolutePath = {
```
tgodzik, line 575: We could do it the same as the method below `isSemanticdbEnabled`
```
def targetroot(scalaVersion: String):
```

and check one or the other flag based on the Scala version.

adpi2: I pushed the revised version. As you can see I had to pass the `scalaVersion` from call site. Tell me what you think.

PR #2012, opened 2020-08-20
Customizable package exclusions

ckipp01:
This is a stab at making the package exclusions configurable. Before going much further I wanted to get some thoughts on the approach. Basically there is still the default set of exclusions that we had before, but now we allow for users to either remove some and add additional ones. We then pass the `userConfig().excludedPackages` to the various providers that are responsible for things like symbol searches etc. I sort of felt like I was passing stuff around a lot, so I wanted to see what your thoughts were on the approach or any red flags you may see before proceeding.

~Some 👁️ 🍬~ (Removing these to not cause confusions, since they don't work like this anymore.

The goal for this is to address #2013 since this is also something I hit on when working with Akka projects and want to filter out javadsl stuff.


tgodzik:
Took a quick look, I think it makes sense :+1:

kpbochenek:
Nice work!

WDYT about making a class 
```scala
class ExcludedPackageChecker(config: () => Config) {
  def isPackageExcluded(pkg: String): Boolean = ...
}
```
This way:
- you have exclusion logic in one place
- you can in future extend logic easily to not only support `startsWith` without changing whole chain of `excludedPackages: List[String]` in methods parameters

Maybe we could also cache `val ex = excludedPackages.map(_.replace(".", "/"))` to not execute that for each check we want to perform, although it is tricky because that would break possibility to dynamically change config :/

olafurpg:
This looks great! I think it would be useful to allow users to only add/remove to the defaults without having to copy the default settings. This allows us as maintainers to update the defaults in the future and have it picked up by users who only want to add/remove a single package. Something equivalent to sbt's `scalacOptions +=`, `scalacOptions -=`, and `scalacOptions :=` 🤔

ckipp01:
> WDYT about making a class
> 
> ```scala
> class ExcludedPackageChecker(config: () => Config) {
>   def isPackageExcluded(pkg: String): Boolean = ...
> }
> ```

Good idea! I was thinking about centralizing it somewhere, so this will work well.

> This looks great! I think it would be useful to allow users to only add/remove to the defaults without having to copy the default settings. This allows us as maintainers to update the defaults in the future and have it picked up by users who only want to add/remove a single package. Something equivalent to sbt's scalacOptions +=, scalacOptions -=, and scalacOptions := 🤔

Good point! From the client perspective, how do you view this working in settings as there seems to be _somewhat_ limited options? I believe we are somewhat limited to what we can do in JSON so a couple random thoughts:

1. Preface the strings with a `+` or `-`

```
+akka.actors.typed.javadsl
-javax
```

2. Maybe expose two settings for the user?

_Additional excludes_ and _Remove excludes_

3. Something I'm not thinking of?

I also think it's important that if we don't populate all the defaults, then it becomes _really_ important to make it super clear what the defaults are because I can already see confusion about, why is this not showing up!? I have nothing in my package exclusions?!

metals/src/main/scala/scala/meta/internal/metals/ExcludedPackagesHandler.scala
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  /**
+   * Should the given package be excluded from indexing
+   *
+   * @param pkg package to check against excluded list
+   */
+  def isExcludedPackage(pkg: String): Boolean = {
+    val packagesToExclude: List[String] = userConfig().excludedPackages match {
+      case Some(pkgs) => handleUserExclusions(pkgs)
```
olafurpg, line 38: I think this method-call here will need to be cached to avoid performance regressions. This is a fairly hot code path that's triggered for all completions and workspace/symbol.

ckipp01: That was my original thought as well, but I've put logging in here to test, and this only ever seems to be hit on when we index. It's appears to not be hit on when triggering completions, symbol searches etc.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -946,6 +950,9 @@ class MetalsLanguageServer(
         case Right(value) =>
           val old = userConfig
           userConfig = value
+          if (userConfig.excludedPackages != old.excludedPackages) {
+            slowConnectToBuildServer(forceImport = true).asJavaObject
```
ckipp01, line 954: So I'm not 100% sure what the best thing to do here is. This works as it just reconnects and re-indexes. Then the workspace symbol search, completions, etc reflect he newly indexed result. However, I'm not sure if there is a better way to re-index everything? Also, since there isn't a build tool recognized for the tests, this fails anyways. Any thoughts on whether this is the best thing to do here or if there is a better alternative?

tgodzik: We already do something similar when reconnecting automatically in the `MetalsLAnguageServer.scala` - not sure if helpful:
```scala
  session.mainConnection.onReconnection { newMainConn =>
              val updSession = session.copy(main = newMainConn)
              connectToNewBuildServer(updSession)
                .flatMap(compileAllOpenFiles)
                .ignoreValue
            }
```

metals/src/main/scala/scala/meta/internal/metals/ExcludedPackagesHandler.scala
olafurpg: You’re right. It’s probably fine as it is. I would still try to optimize the foldLeft, I’m generally paranoid with using immutable collections on hot paths like indexing.

ckipp01: 👍 totally. I was able to change this around a bit and cache it because when testing it on a large code base it was getting called hundreds of times. So that should be taken care of now.

ckipp01:
Alright, I think this is now ready for a more thorough look-over. This will allow a user to mainly add extra package exclusions via User Configuration to remove those suggestions from completions, symbol searches, and import missing symbol code actions. Basically when a change is detected, if the new excluded packages don't match the old, we clear the cache and re-index the classpath. If a user wants to remove an existing default exclusion, the can preface their entry with `--`. You can find a video showing the exclusion behavior here:

https://imgur.com/5CuoRCQ

tests/unit/src/test/scala/tests/CompletionLspSuite.scala
```diff
@@ -229,4 +229,54 @@ class CompletionLspSuite extends BaseCompletionLspSuite("completion") {
       )
     } yield ()
   }
+
+  test("with-exclusions") {
```
kpbochenek, line 233: nice test! 👍

kpbochenek (commented):
just added a few random thoughts.

Solid work 👍

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -49,6 +48,10 @@ object UserConfiguration {
 
   def default: UserConfiguration = UserConfiguration()
 
+  private val defaultExclusion = new ExcludedPackagesHandler(() =>
+    UserConfiguration()
+  ).defaultExclusions.mkString("\n").replace("/", ".").dropRight(1)
```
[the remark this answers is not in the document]
kpbochenek: `defaultExclusions.map(_.dropRight(1)).mkString("\n").replace("/", ".")`

metals-bench/src/main/scala/bench/ClasspathIndexingBench.scala
```diff
@@ -31,7 +31,10 @@ class ClasspathIndexingBench {
   @BenchmarkMode(Array(Mode.SingleShotTime))
   @OutputTimeUnit(TimeUnit.MILLISECONDS)
   def run(): Unit = {
-    ClasspathSearch.fromClasspath(classpath)
+    ClasspathSearch.fromClasspath(
+      classpath,
+      _ => true
```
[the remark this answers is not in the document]
ckipp01: 👍 sounds good. Added the defaults in here.

metals-bench/src/main/scala/bench/CompletionBench.scala
```diff
@@ -113,7 +113,12 @@ abstract class CompletionBench {
 
   def newSearch(): SymbolSearch = {
     require(libraries.nonEmpty)
-    new TestingSymbolSearch(ClasspathSearch.fromClasspath(classpath))
+    new TestingSymbolSearch(
+      ClasspathSearch.fromClasspath(
+        classpath,
+        _ => true
```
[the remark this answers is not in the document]
ckipp01: Added the default back in.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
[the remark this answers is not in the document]
ckipp01: ah duh, of course. good catch.

metals/src/main/scala/scala/meta/internal/metals/ExcludedPackagesHandler.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import scala.collection.mutable.ListBuffer
+
+/**
+ * This class handles packages that should be excluded from indexing causing
+ * them to not be recommended for completions, symbol searches, and code actions.
+ *
+ * @param userConfig access to user input for excluded packages
+ */
+class ExcludedPackagesHandler(userConfig: () => UserConfiguration) {
```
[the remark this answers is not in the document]
ckipp01: Thanks a lot for the review! I went ahead and took some of these suggestions because yea, it simplifies it for sure. I did still incorporate a cache though. Let me know what you think!

tests/unit/src/test/scala/tests/CompletionLspSuite.scala
```diff
@@ -229,4 +229,54 @@ class CompletionLspSuite extends BaseCompletionLspSuite("completion") {
...
+           |DurationLong - scala.concurrent.duration.package
+           |DurationLong - scala.concurrent.duration.package
+           |FiniteDuration - scala.concurrent.duration
+           |FiniteDurationIsOrdered - scala.concurrent.duration.FiniteDuration""".stripMargin,
+        includeDetail = false
+      )
+      _ <- server.didChangeConfiguration(
+        """{
+          |  "excluded-packages": [
+          |    "scala.concurrent/"
```
kpbochenek, line 268: user will type probably `scala.concurrent` without '/'

tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala
```diff
@@ -234,4 +234,54 @@ class WorkspaceSymbolLspSuite extends BaseLspSuite("workspace-symbol") {
...
+           |java.util.concurrent.ScheduledFuture
+           |java.util.concurrent.CompletableFuture
+           |java.util.concurrent.ScheduledThreadPoolExecutor#ScheduledFutureTask
+           |scala.collection.parallel.FutureThreadPoolTasks
+           |java.util.concurrent.RunnableScheduledFuture""".stripMargin
+      )
+      _ <- server.didChangeConfiguration(
+        """|{
+           |  "excluded-packages": [
+           |     "java.util/"
```
kpbochenek, line 271: user will type probably `java.util` without '/'

metals/src/main/scala/scala/meta/internal/metals/ExcludedPackagesHandler.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  def update(pkgs: List[String]): Unit = cachedExclusions = prepareCache(pkgs)
+
+  /**
+   * Should the given package be excluded from indexing
+   *
+   * @param pkg package to check against excluded list
+   */
+  def isExcludedPackage(pkg: String): Boolean = {
+    if (cachedExclusions.nonEmpty) {
```
kpbochenek, line 58: If we did:
`def isExcludedPackage(pkg: String): Boolean = cachedExclusions.exists(excluded => pkg.startsWith(excluded))`

It would not work only in 1 case where we created this class but did not call update yet, because `cachedExclusions` would be empty list.

But if we did:
`private var cachedExclusions: List[String] = prepareCache(pkgsToExclude)`

Then this should work? 

Fine as it now, just curious what you think?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -958,6 +962,10 @@ class MetalsLanguageServer(
         case Right(value) =>
           val old = userConfig
           userConfig = value
+          if (userConfig.excludedPackages != old.excludedPackages) {
+            userConfig.excludedPackages.foreach(excludedPackageHandler.update)
```
kpbochenek, line 966: so logic is inconsistent here now
if (new=None   !=   old=Some(a, b, c)) {
  don't update exclusions as we do that only for Some (foreach)
  but reindex classpath
}

Question what should happen on `None`? 
- treat it as 'reset' flag?
- treat it as 'keep previous' flag?

metals-bench/src/main/scala/bench/ClasspathIndexingBench.scala
```diff
@@ -31,7 +32,10 @@ class ClasspathIndexingBench {
   @BenchmarkMode(Array(Mode.SingleShotTime))
   @OutputTimeUnit(TimeUnit.MILLISECONDS)
   def run(): Unit = {
-    ClasspathSearch.fromClasspath(classpath)
+    ClasspathSearch.fromClasspath(
+      classpath,
+      new ExcludedPackagesHandler(None).isExcludedPackage
```
kpbochenek, line 37: maybe we could make `pkgsToExclude` default value 'None' and drop providing it in those test cases.

kpbochenek (commented):
just a few more questions than problems but I promise not to comment more :P

tests/unit/src/test/scala/tests/CompletionLspSuite.scala
ckipp01: Good point, I'll remove the `/`.

tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala
ckipp01: Same as above, I'll remove the `/`.

metals/src/main/scala/scala/meta/internal/metals/ExcludedPackagesHandler.scala
ckipp01: So my main reason for doing it this is that since this is created right after the user configuration, it will get the default user configuration value, which is `None`. Then if the user has any value set, it will run the update right away as soon as the initialize happens. My thinking was then that we'd run this twice no matter what whenever it was used. So my attempt to avoid that was to hold off on running the `prepareCache` until it's needed to index the first time, or when the `initialize` process happens and the change is picked up during the `didChangeConfiguration`. So I don't think it matters a ton, but again, was trying to avoid running `prepareCache` twice right in a row. I'm open to maybe changing this, but also don't see a downside of having it this way. WDYT?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
ckipp01: Ah, yea good catch. My mistake. I changed this a bit to now no matter what call update when the value has changed. If there are new packages (`Some`), then we call `prepateCache` again with the new pkgs. If the value is now a `None`, we just assign the default exclusions to the cache.

metals-bench/src/main/scala/bench/ClasspathIndexingBench.scala
ckipp01: Sure, I've gone ahead and changed this.

metals/src/main/scala/scala/meta/internal/metals/ExcludedPackagesHandler.scala
kpbochenek: Ahh yes right!, I did not notice it will be called twice in that case, all good :+1:

ckipp01:
> just a few more questions than problems but I promise not to comment more :P

haha, no worries. I appreciate all the input.

kpbochenek (approved):
LGTM :+1:

PR #2014, opened 2020-08-20
Switch file watching library to swoval

eatkins:
The directory watcher library used for file watching has some resource
management issues that can cause problems with other processes in osx.
In the paragraphs below, I copied the description from a similar change
that I proposed in bloop:

The directory watcher library used by metals is not well optimized for
mac os. When I use metals via coc.nvim, I find that the ~ command
often stops working in my sbt projects. The reason is likely that
directory watcher is exhausting the file watching resources provided by
the OS. The swoval file tree views library originally started out as a
fork of directory watcher to fix its resource management issues. It can
be used almost as a drop in replacement for directory watcher though the
api is somewhat different.

The problem is that the directory watcher is written to
implement the java nio watchservice which is itself based on inotify but
the apple file event system is very different from inotify so there is
an impedence mismatch. The inotify api only works on a particular path.
In mac os, the main file watching api watches the root of a file system
tree. This means that if you want to watch two subdirectories foo/bar
and foo/bar/baz, there is no need to create separate event streams for
foo/bar and foo/bar/baz, the events for foo/bar/baz will be
automatically detected by the stream for foo/bar. The swoval library
verifies that each path that is registered is not covered by an existing
stream. It also removes redundant streams if a new stream is opened that
covers an existing stream. Filtering of unwanted subdirectories of a
root directory is handled at the jvm application level.This stream
management is critical because mac os does not allow unlimited numbers
of streams. When adding a recursive directory watch to a
DirectoryWatcher, it creates a watch stream for every subdirectory which
can quickly cause os to stop accepting new watches. It also creates a
new thread for every subdirectory. Swoval will use one thread to handle
events for all of the directories that it is watching.

To minimize any potential downstream impact of this change, I made the
swoval file tree watcher still emit directory watcher
DirectoryChangeEvents. There isn't a 1:1 mapping from a
DirectoryChangeEvent to a swoval data structure anyway so it was
convenient to continue using it.

One additional advantage to swoval compared to directory watcher is that
it is designed to handle watching directories that may not yet exists.
There appeared to be some logic in FileWatcher.scala to create empty
directories in osx to ensure that any missing directories would be
monitored once they were created. I was able to remove that logic
because swoval already handles missing directories for any platform.
Otherwise, this change was straightforward.


eatkins:
This came up along with https://github.com/scalacenter/bloop/pull/1355. I didn't realize that metals also had its own internal file watcher until today I was reading the release notes of metals for something unrelated and saw a directory watcher bump.

eatkins:
I also noticed this comment in the code:
```
 * This class does not watch for changes in `*.sbt` files in the workspace directory and
 * in the `project/`. Those notifications are nice-to-have, but not critical. The library we are
 * using https://github.com/gmethvin/directory-watcher only supports recursive directory meaning
 * we would have to watch the workspace directory, resulting in a LOT of redundant file events.
```
The swoval file tree repository does support non-recursive directory monitoring so it would be possible to watch those files without generating redundant events.

eatkins:
I'm not sure what's going on with the integration tests. When I run them manually on my computer, they work.

tgodzik:
Thanks for the contribution! I am not familiar with swoval, are there maybe some statistics or comparison available?
I will take a closer look at the PR tomorrow, but since I have never used MacOS is there anyone that could confirm the issues directory watcher might have been causing? I wonder why it never popped up before.

Anyway, this might be a nice replacement especially with the enhanced capabilities.

The tests seem to be failing due to the file watcher change, I will check it out.

eatkins:
It's hard to provide a comparison because often no errors are logged or reported by any program, file watching just silently stops working. For the last few years, I was using intellij. When using intellij, I never had issues with `~` working in my sbt projects. In May, I was missing vim and decided to see how metals with coc.nvim would work. As soon as I made this change, `~` would periodically stop working and my sbt session would be spammed with errors from the os file watching service that looked like  `(FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)`. Literally the day that I noticed this, my laptop was stolen and I was forced to upgrade to osx catalana. With catalina, I no longer see the FSEvents errors but the `~` command just silently stops working when I use coc.nvim. To workaround this, I just kill all of the java processes on my machine and then restart sbt and things work again.

There was a related sbt watch issue that was reported a year ago but it was never cross linked over here: https://github.com/sbt/sbt/issues/4699. Note that their problems went away when they exited atom (using metals).

In case it isn't obvious, I am the sole author of swoval. I first started developing it as an sbt plugin because `~` in sbt did not work at all on mac with sbt 1.0.4. In the initial implementation, I used the directory-watcher library because I saw that play was using it and `~` seemed to work with play. I quickly started seeing the FSEvent errors described above and realized that directory watcher was not efficiently managing the event streams. Unfortunately, I didn't really document the directory watcher behavior or write tests that fail when too many streams are created. I designed the library to avoid creating too many streams so it's actually somewhat difficult to induce the behavior using swoval (it's not impossible though). I never figured out how many event streams were needed to cause the os to stop responding to file watch requests. That being said, let's say that you have 200 subdirectories in your project that contain or could contain source files, directory-watcher will register 200 event streams with the os. The file-tree-views implementation will probably only create one event stream for the root path of the project because file event streams provided by osx are always recursive so you only need to watch the root path. To create a non-recursive watch on osx, the library automatically filters out the paths that were not explicitly registered.

Directory watcher also creates a thread per subdirectory which isn't the most scalable approach: https://github.com/gmethvin/directory-watcher/blob/0aea5fdf60697c31008e866f117b59b174924992/core/src/main/java/io/methvin/watchservice/MacOSXListeningWatchService.java#L147
I ran jstack on my metals server running in the sbt build and it had many WatchService threads:
```
> jstack 38699 | grep tid | grep "WatchService" | wc -l
     346
```
Those threads also correspond to how many fs event streams are open so where swoval would likely have only a single stream in the root path, directory-watcher is creating 346.

FWIW, swoval file-tree-views is used to implement `~` in sbt > 1.3.0 and the linked issue above is the only issue where a user reported events not being detected. From a performance perspective, swoval doesn't need to hash the file contents to work correctly which can be significant depending on how many files you are monitoring. I really care a lot about the edit file/detect changes/rebuild loop so I invested a lot of time and energy in making swoval work as well as possible with minimal overhead.

eatkins:
WRT to the tests, I can't reproduce on my machine though I certainly believe that they nevertheless are responsible for the failing CI. I tried both on my mac as well as on a windows vm.

eatkins:
Also, I probably run into this because of my profligate vim usage and the fact that I actively work with a number of different repositories. I am constantly opening files in different projects and ctrl+z exiting. When I ran `ps ux | grep java | grep metals`, there were 5 metals servers running on my machine which is probably pretty unusual.

olafurpg:
Thank you for this contribution! I’m very much in favor of using swoval for file I/O operations 👍

I wonder if the issues you’re hitting on are caused by Metals using directory-watcher. Bloop only uses directory-watcher for “bloop test/compile —watch”. It doesn’t use file watching if you only interact with bloop
via Metals. I would be very keen on using swoval for file I/O in Metals as well.

olafurpg:
Did you confirm that this change fixes your issues?

tgodzik:
> Also, I probably run into this because of my profligate vim usage and the fact that I actively work with a number of different repositories. I am constantly opening files in different projects and ctrl+z exiting. When I ran ps ux | grep java | grep metals, there were 5 metals servers running on my machine which is probably pretty unusual.

Could you maybe post a stack trace in an issue? Looks like a bug.

@ckipp01 have encountered this? I think I fixed it on VS Code (not 100% sure yet), but it also might have something to do with the Metals server itself.

ckipp01:
> The directory watcher library used by metals is not well optimized for
mac os. When I use metals via coc.nvim, I find that the ~ command
often stops working in my sbt projects. 

> @ckipp01 have encountered this? I think I fixed it on VS Code (not 100% sure yet), but it also might have something to do with the Metals server itself.

I personally haven't hit on this, but I also don't use `~` often with my workflow. Today I'll pay attention and force myself to use it more to see if I come across any of the things mentioned, especially since I'm also on mac os.

Thanks a lot for the detailed write up @eatkins!

PR #2021, opened 2020-08-24
Clear fingerprints when semanticdb hash matches

tgodzik:
Not sure why this part of code was commented out, but it resulted in fingerprints never getting cleared, which means we had copies of all revisions in memory.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MutableMd5Fingerprints.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MutableMd5Fingerprints.scala
@@ -28,11 +28,12 @@ final class MutableMd5Fingerprints extends Md5Fingerprints {
       prints <- Option(fingerprints.get(path))
       fingerprint <- prints.asScala.find(_.md5 == md5)
     } yield {
-      // remove non-active md5 fingerprints.
-//      prints.clear()
+      // clear older fingerprints that no longer correspond to the semanticDB hash
+      prints.clear()
       prints.add(fingerprint)
       fingerprint.text
     }
   }
+
   override def toString: String = s"Md5FingerprintProvider($fingerprints)"
 }


ckipp01 (approved):
/shrug ha, that has been commented out for a long time. Good catch. LGTM

kpbochenek:
Did not get full picture yet, but it looks counter intuitive.

You can do 2 operations: add and lookup.
Assuming we have `A.scala` we can add multiple times it with different body:
add("A.scala", "text1")->MD5_1; add("A.scala", "text2")->MD5_2; add("A.scala", "text3")->MD5_3

Then you can do lookup:
lookup("A.scala", MD5_2) -> you get "text2" and entries MD5_1 and MD5_3 are removed.

What bothers me is that order of operations matter:
lookup("A.scala", MD5_2) -> OK
then lookup("A.scala" MD5_1) -> error
VS
lookup("A.scala", MD5_1) -> OK
then lookup("A.scala" MD5_2) -> error

I will try to dig how it is used exactly but could you explain a little how this fingerprinting is used in big picture? :)

kpbochenek:
Also in code:
```scala
        if (sdoc.md5 != md5) {
          fingerprints.lookupText(scalaPath, sdoc.md5) match {
            case Some(oldText) =>
              TextDocumentLookup.Stale(scalaPath, md5, sdoc.withText(oldText))
            case None =>
              TextDocumentLookup.Stale(scalaPath, md5, sdoc)
          }
        } else {
          TextDocumentLookup.Success(sdoc.withText(text))
        }
```
when requesting a stale md5 we cleanup others, but requesting newest one doesn't :/ 

Maybe cleanup should be explicit? (still digging how it works :P )

tgodzik:
> I will try to dig how it is used exactly but could you explain a little how this fingerprinting is used in big picture? :)

This actually just used only in `Semanticdbs.scala` to get the code that corresponds to a semanticdb document to create TokenEditDistance to calculate changes between current document and the one that the semanticdb was generated for. There is only one semanticdb document for any Scala file at any time, which means we are always asking for the last one. Anything else is not really important to us.

> when requesting a stale md5 we cleanup others, but requesting newest one doesn't :/

It might make sense also to cleanup in the other branch, but in reality we ask for semanticdb that often that there will often be stale documents and we'll be asking fingerprints.

kpbochenek (approved):
In that case LGTM :+1:

PR #2025, opened 2020-08-26
Bump memory for test forks to 2GB

tgodzik:
It seems 1GB for a fork was not enough on MacOS. And since we do not have multiple forks running it should be fine.
Current xmx for each process:
- bloop 2GB
- sbt 1GB
- test fork 2GB

With max 7GB it leaves 2GB spare for class space, which should be enough.

### DIFF
--- a/tests/unit/src/main/scala/tests/TestingServer.scala
+++ b/tests/unit/src/main/scala/tests/TestingServer.scala
@@ -1319,6 +1319,7 @@ final class TestingServer(
     val parents = (reveal.uriChain :+ null).map { uri =>
       server.treeView.children(TreeViewChildrenParams(reveal.viewId, uri))
     }
+
     val label = parents.iterator
       .flatMap { r =>
         r.nodes.iterator.map { n =>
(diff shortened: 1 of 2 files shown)


kpbochenek (approved):

ckipp01 (approved):

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -1339,7 +1339,16 @@ final class TestingServer(
           PrettyPrintTree(
             label(uri),
             parent.nodes
-              .map(n => PrettyPrintTree(label(n.nodeUri)))
+              .map { n =>
```
tgodzik, line 1342: One of the tests for TreeViewLSP suite is failing on Windows. Added something to debug the issue, hopefully it helps.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -1339,7 +1339,16 @@ final class TestingServer(
           PrettyPrintTree(
             label(uri),
```
kpbochenek, line 1340: it fails actually here, and it failed again :/

tgodzik: Right! Changed it. It's super weird it fails, labels are produced for exactly the same keys, which is weird this happens.

tgodzik: So it seems the label exists, but it's still not being found in the map :/

kpbochenek: it is in the map according to logs!!! WTF

tgodzik: Trying with stripping BOM

kpbochenek: nope 😭

tgodzik: Sooo it seems again, it was `c:` instead of `C:` - why Windows? Why!?

PR #2028, opened 2020-08-28
Add Maven template to the curated list of templates

tgodzik:
Managed to make it all work now out of the box with run and test available within seconds :tada:


kpbochenek (approved):
LGTM :)

Issue #2029, opened 2020-08-28
Navigating to dependencies code doesn't work in worksheets


keynmol:
**Describe the bug**

Navigating to dependencies code doesn't work in worksheets, while it works for Scala standard library

**To Reproduce**
1. Create a worksheet in VS code
2. Put some code with dependencies in it, e.g.

```scala
import $ivy.`software.amazon.smithy:smithy-model:1.0.10`
import software.amazon.smithy.utils.SimpleParser
import software.amazon.smithy.model.Model

val parser = new SimpleParser("hello")
val x = List(1, 2, 3)
```
3. Try clicking on List
4. Try clicking on SimpleParser
5. Navigating to SimpleParser does not work

**Expected behavior**
Navigation to Scala std sources works as well as sources from the dependency

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.9.3
- The folder doesn't contain anything apart from 2 worksheets and two text files, no build definition

**Search terms**
metals, worksheet, magic import, code navigation

_
![reproduce-metals-worksheet](https://user-images.githubusercontent.com/1052965/91563106-35e61b80-e936-11ea-985b-553052b33616.gif)

tgodzik:
Thanks for reporting! I think this is something we should take a look into before the next release. I will take a look and investigate first.

keynmol:
I also think it's related to the way autocomplete works with worksheets: in the example below I get autocomplete for imports, but no autocomplete when I just type a symbol.

![reproduce-metals-worksheet](https://user-images.githubusercontent.com/1052965/91566362-75fbcd00-e93b-11ea-9bb9-ce5a87b2f45b.gif)

_and sorry for the awful quality... if someone can give me the correct ffmpeg flags to convert .mov to .gif without it looking like the horror show from the 90s, I'll appreciate that :)_

tgodzik:
@keynmol The problem was connected, thanks for pointing that out. The fix is already in a PR. Worksheets inside sources should work correctly right now, which is a workaround for now.

PR #2030, opened 2020-08-28
Fix auto imports and go to sources for worksheets outside sources

tgodzik:
The sources and classpath was not forwarded to the symbol search for worksheets outside build targets.

Fixes https://github.com/scalameta/metals/issues/2029

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
@@ -87,19 +87,19 @@ class Compilers(
   // The "rambo" compiler is used for source files that don't belong to a build target.
   lazy val ramboCompiler: PresentationCompiler = createStandaloneCompiler(
     PackageIndex.scalaLibrary,
+    Try(StandaloneSymbolSearch(workspace, buffers))
+      .getOrElse(EmptySymbolSearch),
     "metals-default"
   )
 
-  def createStandaloneCompiler(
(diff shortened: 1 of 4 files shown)


kpbochenek (approved):
I think with this test change we stop testing completion for worksheets inside build target?
btw. this test tests completions, there is a test in line ~400 called "definition"

Maybe we could simply have 3 dedicated tests:

- gotodef-dependency-inside-target         <- this should already work
- gotodef-dependency-outside-target       <- this PR fixes this
- gotodef-targetdependency                      <- inside target we can also use project deps, correct?

This is something 'nice to have' and you can completely ignore it :)

tgodzik:
> I think with this test change we stop testing completion for worksheets inside build target?
> btw. this test tests completions, there is a test in line ~400 called "definition"

This test deals with the worksheet dependencies and I am avoiding adding additional tests and we still test completions in other tests. 
 
> * gotodef-dependency-inside-target         <- this should already work

I can add it back to the test, no problem.

> * gotodef-targetdependency                      <- inside target we can also use project deps, correct?

Already tested in another test.

tgodzik:
Moved the test to WorksheetLspSuite, which is tested for Scala 2.12.12 and that test was anyway only tested for that version. Also, split it into two tests one for outside target and one for inside it.

kpbochenek (approved):
I start to appreciate elasticity of munit more and more :D

LGTM :+1:

PR #2053, opened 2020-09-06
Fix missing classes from _empty_ package in treeView.

### LINKED ISSUE #1648: Missing class from `_empty_` package in TVP
**Describe the bug**
It seems that classes without any package are not seen in the Tree View. Once you move them into a package, they get shown.

**To Reproduce**
1. Create a class without a package
2. Try to look for it in the TreeView

**Expected behavior**
The created class should show up in the the Metals Tree View

**Installation:**
- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.8.4

**Search terms**
tree view empty package

ckipp01:
Previously if we were looking for members underneath `_empty_` we ended up
looking for them in `<uri>/_empty_` rather than just looking for them in the
root where they are stored. This adds in another check where when we look to
see if a symbol is in the root package, we also check to see if it's an empty
package, and if so, then just return the root value to look for it.

Closes #1648


tgodzik (approved):
LGTM! Thanks for fixing this!

Issue #1801, opened 2020-06-01
Flaky Ammonite test on windows


ckipp01:
**Describe the bug**
I've been seeing the following fail on windows pretty consistently.
```
[info] Test run tests.RangeFormattingWhenPastingSuite finished: 0 failed, 0 ignored, 15 total, 71.291s
Error deleting C:\Users\RUNNER~1\AppData\Local\Temp\ammonite-2.1.44086126562170075674.jar: java.nio.file.FileSystemException: C:\Users\RUNNER~1\AppData\Local\Temp\ammonite-2.1.44086126562170075674.jar: The process cannot access the file because it is being used by another process.

Error deleting C:\Users\RUNNER~1\AppData\Local\Temp\ammonite-2.1.42451691160055696164.jar: java.nio.file.FileSystemException: C:\Users\RUNNER~1\AppData\Local\Temp\ammonite-2.1.42451691160055696164.jar: The process cannot access the file because it is being used by another process.

Error deleting C:\Users\RUNNER~1\AppData\Local\Temp\ammonite-2.1.46551272268832771215.jar: java.nio.file.FileSystemException: C:\Users\RUNNER~1\AppData\Local\Temp\ammonite-2.1.46551272268832771215.jar: The process cannot access the file because it is being used by another process.

Error deleting C:\Users\RUNNER~1\AppData\Local\Temp\ammonite-2.1.41230435427312411412.jar: java.nio.file.FileSystemException: C:\Users\RUNNER~1\AppData\Local\Temp\ammonite-2.1.41230435427312411412.jar: The process cannot access the file because it is being used by another process.

Error deleting C:\Users\RUNNER~1\AppData\Local\Temp\ammonite-2.1.47523226631666151249.jar: java.nio.file.FileSystemException: C:\Users\RUNNER~1\AppData\Local\Temp\ammonite-2.1.47523226631666151249.jar: The process cannot access the file because it is being used by another process.
``` 

**Search terms**
CI, Ammonite

tgodzik:
I think this is more likely connected to:
```
2020.06.01 02:13:31 INFO  Connected to Ammonite Build server v2.1.4
2020.06.01 02:13:41 INFO  time: indexed workspace in 1.3s
2020.06.01 02:13:41 INFO  compiling main.sc
2020.06.01 02:13:48 INFO  time: compiled main.sc in 7.29s
2020.06.01 02:13:48 INFO  tests.TestingServer#didSave
[...]

PR #2054, opened 2020-09-07
Look into ammonite flakiness in tests to see if we can remove it

ckipp01:
A lot has changed since #1801 was created. I'm going to run the tests on this a bit and see if this may have solved itself.


ckipp01:
One full run with all ✅ . I'll do another just to make sure.

ckipp01:
Multiple runs with no issues. I think we are ok to remove the `flaky` tag here.

kpbochenek (approved):
Yup looking good :+1:

ckipp01:
There is one test failure, but it's not related to Amm so I'm going to go ahead and merge this.

PR #2056, opened 2020-09-07
Add in new version of Ammonite

### LINKED ISSUE #1852: Incorrect import in Ammonite script not detected
**Describe the bug**
If you import something in an Ammonite script that doesn't exist, the script still compiles and no error is shown to the user.

**To Reproduce**
1. Open an Ammonite script
2. Add a couple fake imports
3. Save
4. Don't see any errors.

For example, given a simple script with the following imports:

```scala
import $file.iDontExist, iDontExist._

import java.what
```

I would expect that both of these error and produce diagnostics that both `iDontExist` doesn't exist and also that `what` is not part of `java`. 

**Installation:**
- Operating system: macOS
- Editor: Vim
- Metals version: Latest snapshot

**Search terms**
Ammonite, scripts, incorrect imports

ckipp01:
This version contains a fix for #1852 so incorrect $file imports
now give diagnostics.

Closes #1852

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -217,7 +217,7 @@ lazy val V = new {
     "org.eclipse.lsp4j" % "org.eclipse.lsp4j.debug" % "0.9.0"
   val coursier = "2.0.0-RC6-25"
   val coursierInterfaces = "0.0.25"
-  val ammonite = "2.2.0"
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM

PR #2057, opened 2020-09-07
Don't search for runTest code lenses in Ammonite scripts and worksheets.

ckipp01:
Up to this point we search for all types of code lenses even in Ammonite scripts
where there won't be any run or test code lenses that appear. This will save us a
bit as we won't look through all the symbols in the textDocument for the main method.
It's probably not a huge deal, but no need to do it.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codelenses/RunTestCodeLens.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codelenses/RunTestCodeLens.scala
@@ -35,15 +35,20 @@ final class RunTestCodeLens(
   ): Seq[l.CodeLens] = {
     val textDocument = textDocumentWithPath.textDocument
     val path = textDocumentWithPath.filePath
-    val distance = buffers.tokenEditDistance(path, textDocument.text)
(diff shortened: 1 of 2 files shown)


ckipp01:
ugh, maybe I was too eager with removing the flaky...
```
[error] 	tests.AmmoniteSuite
```

kpbochenek:
Should worksheets be excluded in similar way?

ckipp01:
> Should worksheets be excluded in similar way?

Good point, yea they probably should be. I'll add that in.

kpbochenek (approved):
LGTM! invisible but nice improvements :+1:

(it is always worth try if tests are no longer flaky, a lot has changes for ammonite, but if they still fail, what can you do :( )

tgodzik (approved):
LGTM

tgodzik:
> ugh, maybe I was too eager with removing the flaky...
> 
> ```
> [error] 	tests.AmmoniteSuite
> ```

Yeah, it's really tricky :/ I spend way too much time trying to fix it.

ckipp01:
> Yeah, it's really tricky :/ I spend way too much time trying to fix it.

I just added it back in.

Issue #2024, opened 2020-08-25
Unable to use worksheets if using scalatags + `withDottyCompat` and Dotty


ckipp01:
**Describe the bug**
This is a super niche bug that I'm not 100% sure is even Metals, but could just lie somewhere the the combination listed in the title. If you are in a Dotty project and you use [scalatags](https://github.com/lihaoyi/scalatags) via `withDottyCompat` you get the same odd error in worksheets no matter what, even with the project compiling correctly. For example, in a worksheet as simple as below:

```scala
val x = 3
```
You'll see a diagnostic that is returned (always without a position, so it just appears at line 0) that looks something like this:

```
[mdoc] [E] test.worksheet.sc:9 (mdoc generated code) Found:    (App.this.x : Int)
Required: sourcecode.Text[A]

where:    A is a type variable with constraint 

    $doc.binder(x, 0, 0, 0, 9)

256
```

**To Reproduce**
1. Use the `New Scala Project` command to create a new project via the `lampepfl/dotty.g8` template
2. Use the default build definition, but add the following
```scala
    libraryDependencies += ("com.lihaoyi" %% "scalatags" % "0.9.1")
      .withDottyCompat(dottyVersion)
```
3. Create a worksheet in `src/main/scala`
4. Add anything to it and attempt to evaluate

**Expected behavior**
Since this is compiling correctly when used in a module I'd expect this to also work with worksheets.

**Installation:**
- Operating system: macOS
- Editor: Any
- Metals version: v0.9.3 and later

**Additional context**
I thought this may be in general when using `withDottyCompat`, but trying out also having other dependencies via `withDottyCompat` on the classpath didn't cause any issues. Using the same code and also having Doodle on the classpath with 2.13 also caused no issues for worksheets. So again, it's some odd combination with worksheets, dotty, and scalatags + `withDottyCompat`.

cc @eloots @noelwelsh

**Search terms**
worksheets, dotty

noelwelsh:
Wow, that's weird. The only unusual things that I can think of in Doodle are:

* using the Java2D / Swing APIs (which are standard Java, but most Scala programs don't touch them.)
* using Haoyi's Scalatags library, which *might* use some macros.

I don't see why these would cause issues if no Doodle code is being called.

ckipp01:
> using Haoyi's Scalatags library, which might use some macros.

[...]

PR #2058, opened 2020-09-09
Add support for dotty 0.27.0-RC1 and update mdoc

tgodzik:
The updated mdoc now uses `toString` and inbuilt mechanism for reporting types, which solves the issue with pprint dependency.

Fixes https://github.com/scalameta/metals/issues/2024

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -178,13 +178,14 @@ lazy val V = new {
   val bsp = "2.0.0-M4+10-61e61e87"
   val bloop = "1.4.3-27-dfdc9971"
   val scala3 = "0.26.0"
+  val scala3Candidate = "0.27.0-RC1"
   val bloopNightly = bloop
   val sbtBloop = bloop
   val gradleBloop = bloop
   val mavenBloop = bloop
-  val mdoc = "2.2.5"
+  val mdoc = "2.2.7"
   val scalafmt = "2.6.4"
-  val munit = "0.7.10"
+  val munit = "0.7.12"
   // List of supported Scala versions in SemanticDB. Needs to be manually updated
   // for every SemanticDB upgrade.
   def supportedScalaBinaryVersions =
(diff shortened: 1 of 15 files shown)


tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -460,7 +460,7 @@ class CompletionSuite extends BaseCompletionSuite {
   )
 
   check(
-    "commit1",
+    "commit1".tag(IgnoreScalaVersion("0.27.0-RC1")),
```
tgodzik, line 463: Not sure why, but in 0.27.0-RC1 completions stopped working from things in the standard library.

tgodzik: I will follow up in dotty and ask.

build.sbt
```diff
@@ -201,8 +202,8 @@ lazy val V = new {
   def scala2Versions = nonDeprecatedScala2Versions ++ deprecatedScala2Versions
 
   // Scala 3
-  def nonDeprecatedScala3Versions = Seq(scala3, "0.25.0")
-  def deprecatedScala3Versions = Seq("0.25.0-RC2", "0.26.0-RC1")
+  def nonDeprecatedScala3Versions = Seq(scala3, scala3Candidate)
```
tgodzik, line 205: It seems that we have a large regression in 0.27 and no completions coming from scala library are not displayed.
This way most large tests will be run on 0.26, which works and I will focus on fixing it upstream before the release.

tests/unit/src/main/scala/tests/QuickBuild.scala
```diff
@@ -117,7 +117,9 @@ case class QuickBuild(
           s"org.scala-lang:scala-library:$scalaVersion",
           s"org.scala-lang:scala-reflect:$scalaVersion"
         )
-      } ++ libraryDependencies
```
tgodzik, line 120: Fun fact, this never added anything for Scala 3, which is why some tests worked even with bad dependencies. CrossReferenceLspSuite and RenameCrossSuite were the only ones that were problematic.

build.sbt
tgodzik: Link to the regression https://github.com/lampepfl/dotty/issues/9760

romanowski (approved):

tests/unit/src/main/scala/tests/QuickBuild.scala
tgodzik: Ach! This is due to the fact that else was operating on the whole expression with the `++`

tgodzik:
This is blocking for @odersky so I will go ahead and merge. If anyone has any suggestions you can report them here, I will try to fix them later in a separate PR.

tests/unit/src/main/scala/tests/QuickBuild.scala
ckipp01: 👀 good catch ha

PR #2060, opened 2020-09-10
Fix wrong definition position for worksheets

tgodzik:
I haven't noticed that the test position were actually bad. They should point to 0 line after adjustments.


kpbochenek (approved):
Good catch :+1:

Jumping to local definitions looks okay with this fix.

Jumping to e.g. `println` doesn't because it is related to this issue https://github.com/lampepfl/dotty/issues/9760 correct?

tgodzik:
> Jumping to e.g. `println` doesn't because it is related to this issue [lampepfl/dotty#9760](https://github.com/lampepfl/dotty/issues/9760) correct?

I think it's due to symbol search not being used in ScalaPresentationCompiler, so this is on out side.

PR #2075, opened 2020-09-17
Change new-scala-file to enable quick creation of a file.

### LINKED ISSUE #2073: Add in functionality to quickly create a worksheet
Not necessarily a problem, but often a user wants to quickly create a worksheet where they are at. In order to do this currently (client agnostic), you need a few different steps to trigger the [create-new-scala-file](https://scalameta.org/metals/docs/editors/new-editor.html#create-new-scala-file) server command, choose the type of the file, and also the name. There is functionality to actually give the name as the second param ([Note, we need to update the params in the docs](https://scalameta.org/metals/docs/editors/new-editor.html#create-new-scala-file)), but that only cuts out one step. It'd be nice with one action to create a worksheet in the current directory, with the dirname as the name of the worksheet.

This was brought up by @tpolecat via the [Tweeters](https://twitter.com/tpolecat/status/1301896741604798471?s=20).

I'm creating this here now instead of the VS Code extension repo because while some work will need to be done in each client for this, it'd be nice to have a unified way to do it for any client that supports server commands.

There is a couple different ways that we could approach this.

1. We could edit the existing `new-scala-file` command to take in another argument of file type. Then if the dir and file type arguments are there, we could just automatically conclude the user wants a worksheet with the dirname as the name, and instead of sending the dialogue back, we can just create the file. This change would need to start here:
https://github.com/scalameta/metals/blob/02770a08c9c513f4b6427a92f9735b0421d8ad13/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1492-L1502

2. We could create a new `new-scala-worksheet` command that does exactly what the above does, but would be its own command.

Both of these options would allow the client to with one "execute command", get a new worksheet. This could be triggered via button, shortcut, or whatever the client can do.

**Search terms**:
worksheet, new-scala-file

ckipp01:
This changed the current functionality of the new-scala-file command to allow
for another argument in the third position, which signifies the file type. Previously,
we allowed for the name of the file to be in the args in the second position, but now
we also allow the file type to allow for a client to create a file with a single command.
This mainly will allow for the quick-creation of worksheets which can be bound to an
editor action or command.

This also changes up the NewFilesLspSuite to cover the other arguments. Previously the
functionality of the name being passed in wasn't covered. Now, that functionality and
also providing the fileType is also covered in the tests.

Ideally the args for this command would have been an JSON object allowing for the necessary
keys to be set or not, but since it wasn't that way, I hesitate to change it now since
some clients may be using the array with 2 items (directory and name). Implimenting it
this will will allow for backwards compatability with any clients that may already be
utilizing the command with both directory and name being passed in.

Example with it bound to a `quick-worksheet` command in `coc-metals`:

Closes #2073

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -255,7 +255,10 @@ object ServerCommands {
        |
        |Note: requires 'metals/inputBox' capability from language client.
        |""".stripMargin,
-    "[string], where the string is a directory location for the new file."
+    """|[string[]], where the first is a directory location for the new file.
+       |The second and third positions correspond to the file name and file type to allow for quick
+       |creation of a file if all are present.
+       |""".stripMargin
   )
 
   val NewScalaProject = new Command(
--- a/tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
@@ -20,234 +20,234 @@ class NewFilesLspSuite extends BaseLspSuite("new-files") {
     Some(InitializationOptions.Default.copy(inputBoxProvider = Some(true)))
 
   check("new-worksheet-picked")(
-    Some("a/src/main/scala/"),
-    Right(worksheet),
-    Right("Foo"),
-    "a/src/main/scala/Foo.worksheet.sc",
-    ""
+    directory = Some("a/src/main/scala/"),
+    fileType = Right(worksheet),
+    fileName = Right("Foo"),
+    expectedFilePath = "a/src/main/scala/Foo.worksheet.sc",
+    expectedContent = ""
   )
 
   check("new-worksheet-name-provided")(
-    Some("a/src/main/scala/"),
-    Left(worksheet),
-    Right("Foo"),
-    "a/src/main/scala/Foo.worksheet.sc",
-    ""
+    directory = Some("a/src/main/scala/"),
+    fileType = Left(worksheet),
+    fileName = Right("Foo"),
+    expectedFilePath = "a/src/main/scala/Foo.worksheet.sc",
+    expectedContent = ""
   )
 
   check("new-worksheet-fully-provided")(
-    Some("a/src/main/scala/"),
-    Left(worksheet),
-    Left("Foo"),
-    "a/src/main/scala/Foo.worksheet.sc",
-    ""
+    directory = Some("a/src/main/scala/"),
+    fileType = Left(worksheet),
+    fileName = Left("Foo"),
+    expectedFilePath = "a/src/main/scala/Foo.worksheet.sc",
+    expectedContent = ""
   )
 
   check("new-ammonite-script")(
     directory = Some("a/src/main/scala/"),
     fileType = Right(ammonite),
     fileName = Right("Foo"),
-    "a/src/main/scala/Foo.sc",
-    ""
+    expectedFilePath = "a/src/main/scala/Foo.sc",
+    expectedContent = ""
   )
 
   check("new-ammonite-script-name-provided")(
     directory = Some("a/src/main/scala/"),
     fileType = Right(ammonite),
     fileName = Left("Foo"),
-    "a/src/main/scala/Foo.sc",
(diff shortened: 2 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -255,7 +255,10 @@ object ServerCommands {
        |
        |Note: requires 'metals/inputBox' capability from language client.
        |""".stripMargin,
-    "[string], where the string is a directory location for the new file."
+    """|[string[]], where the first is a directory location for the new file.
```
ckipp01, line 258: We actually had this just as a string before, when in reality this as an Array with a URI in the first position and a filename in the second position.

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
```diff
@@ -19,26 +19,84 @@ class NewFilesLspSuite extends BaseLspSuite("new-files") {
...
-    Some("Foo"),
+    Left(worksheet),
+    Right("Foo"),
+    "a/src/main/scala/Foo.worksheet.sc",
+    ""
+  )
+
+  check("new-worksheet-fully-provided")(
+    Some("a/src/main/scala/"),
+    Left(worksheet),
```
tgodzik, line 40: Could you also use named parameters here and in the method above?

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
```diff
@@ -19,26 +19,84 @@ class NewFilesLspSuite extends BaseLspSuite("new-files") {
...
+    Left("Foo"),
+    "a/src/main/scala/Foo.worksheet.sc",
+    ""
+  )
+
+  check("new-ammonite-script")(
+    directory = Some("a/src/main/scala/"),
+    fileType = Right(ammonite),
```
tgodzik, line 48: Could we actually run the 3 cases in one check? Since the parameters are basically the same, we could just generate all the combinations `Right` `Right`, `Right` `Left` and `Left` `Left`

tgodzik (approved):
LGTM I just added some suggestions about the tests, but nothing major. Feel free to ignore.

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
ckipp01: For sure. Added.

tests/unit/src/test/scala/tests/NewFilesLspSuite.scala
ckipp01: I think possibly yea, but the first approach I took was to just have one check per file type, and then in the check create the file normally, and then also the other two ways, but then the check code became pretty hairy trying to get all the checked messages right in various places. So I actually then changed it to what I have now. It's more verbose, but I find it way easier to reason about.

ckipp01: I'm going to go ahead and merge this in. If you want I can revisit and try to refactor again, but I'm assuming it's fine given your original comment 😬

tpolecat:
Cool, how does this work in vs-code?

ckipp01:
> Cool, how does this work in vs-code?

As of right now, it doesn't. It's a secret plan I have to start sneaking features only in `coc-metals` to get more Vim users 😏 ...

But for real, it's not currently implemented in VS Code yet. There sort of needs to be a discussion had about the best way to do it. It would be quite simple to just tie it to a right click on folder and then click "Create Quick Worksheet", which would create the worksheet in the directory with the name of the directory as the worksheet. _However_ I'm not sure if that's the best way to VS Code users to do this, or if they would instead want to create it from the command palette or something else completely. I don't use VS Code, so I'm probably the worst person to make that decision. I'm counting on the others that use VS Code to chime in with their opinions.

tpolecat:
Command palette would be fine.

tgodzik:
We can add it to command palette for now and figure out the UI later. There might some new possibilities with submenus.

tgodzik:
@ckipp01 I will take a look at it before the release. Probably will finish it all on monday though.

PR #2083, opened 2020-09-19
Strip out `[E]` or `[error]` from stacktrace in analyzer.

ckipp01:
So in some clients (VS Code), the stacktrace will be printed without any
prefaces, which allows for the stacktrace to be correctly parsed and the lenses
created for them when utilyzing the stacktrace-anlyzer functionality. However,
for some clients that don't have a nice built-in way to run the application
in the client, they may be resorting to run via their build tool or with the
bloop-cli. This change ensures that a person can copy the stacktrace from
sbt or from bloop cli and still utilize the functionality.

Note that nothing special for Mill was needed since Mill also doesn't preface
the stacktrace output with any `[E]` or anything.


kpbochenek (approved):
good catch, thanks for fixing this!

LGTM :+1:

PR #2084, opened 2020-09-20
Add fetch related stuff in one call

ckipp01:
I'm not really sure what's going on, but in the last release, we are for some reason getting the following on the `git fetch --tags` call:

```
 * [new tag]           v0.4.0     -> v0.4.0
```
This is possibly the cause for the odd snapshot we currently have: `0.4.4+112-6154859a-SNAPSHOT`, which is a bit annoying since if you use it, it will warn you that you are using an out of date version every time you open Metals.

It actually seems that all the tags are pulled during the `git fetch --unshallow` call above if you look in the logs, https://github.com/scalameta/metals/runs/1139811242#step:6:47, but I figured this could also just all be done in the same `fetch` call. I don't know really why, but _maybe_ this could help? Thoughts?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,8 +11,7 @@ jobs:
       - uses: olafurpg/setup-scala@v7
       - uses: olafurpg/setup-gpg@v2
       - uses: coursier/cache-action@v3
-      - run: git fetch --unshallow
-      - run: git fetch --tags
+      - run: git fetch --tags --unshallow -f
       - name: Publish
         run: |
           sbt ci-release


kpbochenek:
Uh okay so '-f' has no meaning here as it is not used with `src:dst`, I would not add it.
We can clearly do those 2 steps in one line that looks completely okay.

But it sounds there is a problem that should be fixed but I don't see one :/ 
`git fetch --unshallow` fetched all other tags, one which is `* [new tag]           v0.9.3     -> v0.9.3` so there are all tags there.

Looking at docs: https://scalameta.org/metals/docs/editors/overview.html
Latest is 0.9.3 and snapshot is 0.9.3+110-9d268152-SNAPSHOT

Where is this `0.4.4+112-6154859a-SNAPSHOT` found?

ckipp01:
> Uh okay so '-f' has no meaning here as it is not used with `src:dst`, I would not add it.
> We can clearly do those 2 steps in one line that looks completely okay.
> 
> But it sounds there is a problem that should be fixed but I don't see one :/
> `git fetch --unshallow` fetched all other tags, one which is `* [new tag] v0.9.3 -> v0.9.3` so there are all tags there.
> 
> Looking at docs: [scalameta.org/metals/docs/editors/overview.html](https://scalameta.org/metals/docs/editors/overview.html)
> Latest is 0.9.3 and snapshot is 0.9.3+110-9d268152-SNAPSHOT
> 
> Where is this `0.4.4+112-6154859a-SNAPSHOT` found?

Yea I literally was just coming here to say that I don't think this was the issue. The `-f` came from a shameless copy from the way Scalafmt has it set up. This time it seemed to work fine actually, and yea, like you said the version is correct this time around. Maybe it was just a fluke that one time that the `0.4.4....` version was created? Either way. I'll close this.

kpbochenek:
ahh okay, FYI there is a new syntax how to fetch all by simply configuring github action and not executing custom commands:
https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches

PR #2085, opened 2020-09-20
Refactor NewFileProvider

ckipp01:
Today I sat down to write an Ammonite script and when I used the
new-scala-file command I saw the following message:

```
Enter the name for the new ammonite
```

This bugged me a bit since I wanted it to say "for the new Ammonite
Script". The same thing happens when you try to create a case-class. The
message that you see is:

```
Enter the name for the new case-class
```

I think it's nicer to show "Case Class" here. This along with a couple
other small things I noticed when I did
https://github.com/scalameta/metals/pull/2075 like just matching against
strings to get the fileType and not really having any sort of ADT for
the new filetypes. So I made some opinionated choices and refactored
some stuff related to the new-scala-file command into its own package.
This _I hope_ makes it a bit easier to work with and also has the added
benefit of showing the label in the message instead of the id like we
were before.


kpbochenek (approved):
Improving type safety always big yes :+1: :+1:

PR #2087, opened 2020-09-21
Bump other reference of coursier/cache

ckipp01:
I may or may not have missed this one...

However, the positive is that everything looks to have gone fine with the last one.


kpbochenek (approved):
LGTM :+1:

PR #2088, opened 2020-09-21
Add in top level-completions for empty file.

### LINKED ISSUE #1663: No completion suggestions at the beginning of the file
**Describe the bug**

**To Reproduce** Steps to reproduce the behavior:

1. Create a new Scala file.
2. Type "p".
3. No completion suggestions
4. If have a space before, completions occur.

**Expected behavior**

Should have suggestions like "package" etc.

**Screenshots**

image
image

**Installation:**

- Operating system: Windows
- VSCode version: 1.44.2
- VSCode extension version: 1.8.6
- Metals version: 0.8.4

**Additional context**

**Search terms**

ckipp01:
This adds some some somewhat sensible top-level completions when you attempt to
trigger a completion in an empty file. For example `p@@` will produce:

- package
- private
- protected
- import

All which would be valid compared to the behavior previously where nothing was
recommended.

Closes #1663

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
@@ -739,7 +739,7 @@ trait Completions { this: MetalsGlobal =>
   def inferIdentStart(pos: Position, text: String): Int = {
     def fallback: Int = {
       var i = pos.point - 1
-      while (i > 0 && Chars.isIdentifierPart(text.charAt(i))) {
+      while (i >= 0 && Chars.isIdentifierPart(text.charAt(i))) {
         i -= 1
       }
       i + 1
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
@@ -309,7 +309,7 @@ class CompletionProvider(
     }
     completions.foreach(visit)
     completion.contribute.foreach(visit)
-    buf ++= keywords(pos, editRange, latestParentTrees)
+    buf ++= keywords(pos, editRange, latestParentTrees, completion)
     val searchResults =
       if (kind == CompletionListKind.Scope) {
         workspaceSymbolListMembers(query, pos, visit)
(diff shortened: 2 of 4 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -739,7 +739,7 @@ trait Completions { this: MetalsGlobal =>
   def inferIdentStart(pos: Position, text: String): Int = {
     def fallback: Int = {
       var i = pos.point - 1
-      while (i > 0 && Chars.isIdentifierPart(text.charAt(i))) {
+      while (i >= 0 && Chars.isIdentifierPart(text.charAt(i))) {
```
ckipp01, line 742: So this was causing issues where if I was trying to do `p@@` to complete package, the lsp edit position start was never 0 here if you were on line 0 and character 1. It was returning an edit with the start and end in the same position causing the edit to append to the previous character instead of replace.

ckipp01:
🤔 so sort of interesting is the case where you have a class with nothing else and try to trigger a scaladoc completion.

```scala
/**@@
class Foo()
```

Where as in the past this would only suggest the scaladoc completion where as now this still has the scaladoc completion and it is preselected, but it also has `package, private, class etc`. That sort of makes sense actually, but what do you all think? In this scenario, should it _only_ be suggesting the scaladoc, or other toplevel suggestions?

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -340,6 +340,25 @@ class CompletionSuite extends BaseCompletionSuite {
     )
   )
 
+  check(
```
tgodzik, line 343: Could we move it to `CompletionKeywordSuite` ?

tgodzik:
>  so sort of interesting is the case where you have a class with nothing else and try to trigger a scaladoc completion.
> 
> ```scala
> /**@@
> class Foo()
> ```
> 
> Where as in the past this would only suggest the scaladoc completion where as now this still has the scaladoc completion and it is preselected, but it also has `package, private, class etc`. That sort of makes sense actually, but what do you all think? In this scenario, should it _only_ be suggesting the scaladoc, or other toplevel suggestions?

I think it would be better if it only suggested scaladoc completions in that case. Maybe we could check explicitely?

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
ckipp01: Ah, yes. Total oversight on my part. 👍

ckipp01:
Sorry it took so long for me to get back to this!

> I think it would be better if it only suggested scaladoc completions in that case. Maybe we could check explicitely?

I went ahead and added an explicit check to make sure that if we have a position completion that is a scaladoc completion, we don't do the other top level completions.

tgodzik (approved):
LGTM! Thanks for the fix!

PR #2091, opened 2020-09-23
Remove javax from default excluded packages

### LINKED ISSUE #2090: Why is `javax` excluded from indexing?
For a long time, I assumed that I didn't get completions for `javax.imageio` because of some bug in my setup.

Only after reading the release notes for 0.9.4 I realized that this was a feature-not-a-bug!

The `javax` package space contains many mainstream features like ImageIO, Swing and SQL.

Isn't excluding it by default going to be a poor experience for new Metals users who aren't aware of the exclusion list and the way to configure it?

**Search terms**

javax, imageio, swing, sql, exclusion

gabro:
Closes #2090

As discussed in the issue, the exclusion of javax isn't probably a great default anymore.
We even use some javax imports ourselves in Metals!

This PR removes `javax` from the default set of excluded packages. Users can opt-in to the exclusion using the recently released setting.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
@@ -116,8 +116,8 @@ object UserConfiguration {
         """`[]`.""",
         """["-Xmx1G"]""",
         "Ammonite JVM Properties",
-        """|Optional list of JVM properties to pass along to the Ammonite server. 
-           |Each property needs to be a separate item.\n\nExample: `-Xmx1G` or `-Xms100M`" 
+        """|Optional list of JVM properties to pass along to the Ammonite server.
+           |Each property needs to be a separate item.\n\nExample: `-Xmx1G` or `-Xms100M`"
            |""".stripMargin
       ),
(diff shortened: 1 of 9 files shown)


metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -139,7 +139,7 @@ object UserConfiguration {
             |Example:
             |
             |```js
-            |["--javax"]
+            |["--sun"]
```
gabro, line 142: I've picked another example, since javax is not excluded anymore

olafurpg (approved):
LGTM :+1: I can't remember the exact reason `javax/` was excluded by default but I think it just contained some generic class names that frequently conflicted with symbols that I was working with in the Metals codebase in the early days. I don't mind including it by default

ckipp01 (approved):
Looks good! LGTM!

tgodzik (changes requested):
Thanks for following up on the issue! There are some tests still failing, but it seems they are all connected to the changes.

tgodzik (approved):
LGTM!

PR #2097, opened 2020-09-25
Display used BuildServer in Doctor

kpbochenek:
No logic was changed in this PR (hopefully :D )

Added displaying BuildServer that is currently used and one that will be used after restart of metals.
To make it reliable I extracted resolving it to single method.


kpbochenek:
Obligatory screenshot(after removing .bloop dir)
2020-09-25-160128_372x130_scrot

tgodzik (approved):
LGTM! It's always good to be explicit, it will be useful for the sbt PR

PR #2098, opened 2020-09-25
Simplify TestHovers and remove warning

gabro:
While tinkering, I've noticed a deprecation warning due to a usage of `MarkedString`.

In `TestHovers` we still handled the case in which we returned a `MarkedString` or a `String` in `Hover` (now deprecated in lsp4j), but in the implementation we actually just use `MarkupContent` so that code branch is effectively unused.

I think it's safe to remove it, let's see whether the CI thinks otherwise

### DIFF
--- a/tests/mtest/src/main/scala/tests/TestHovers.scala
+++ b/tests/mtest/src/main/scala/tests/TestHovers.scala
@@ -1,7 +1,6 @@
 package tests
 
 import scala.meta.inputs.Input
-import scala.meta.internal.jdk.CollectionConverters._
 import scala.meta.internal.mtags.MtagsEnrichments._
 import scala.meta.internal.pc.HoverMarkup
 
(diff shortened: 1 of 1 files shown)


tests/mtest/src/main/scala/tests/TestHovers.scala
```diff
@@ -28,21 +27,7 @@ trait TestHovers {
...
-                e.asScala match {
-                  case Left(value) =>
-                    value
-                  case Right(marked) =>
-                    codeFence(marked.getValue, marked.getLanguage)
-                }
-              }
-              .mkString("\n")
-        }
+        val types = value.getContents.getRight.getValue()
```
gabro, line 30: the unhandled "left" case is the deprecated `MarkedString | String` case, which we never produce

ckipp01 (approved):
Looks good. Thanks for this! LGTM!

Issue #2099, opened 2020-09-25
Organize Imports does not work for Scala 2.11.12


pvid:
**Describe the bug**

Code action "Organize Imports" does not organize imports nor does it reports any errors in the UI, Doctor or metals.log

**To Reproduce**

1. Find a project using Scala 2.11.12 (For example https://github.com/scalapb/sparksql-scalapb/tree/v0.9.3 and manually change sparkSqlScalaPB scala version to 2.11.12)
2. Mess up imports in some file
3. Run organize imports
4. See that nothing happens and no error is reported

**Expected behavior**

Organize imports should work for Scala 2.11.12 or provide clear information why it is not working.

**Installation:**

- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: 0.9.4+15-a3d0d948-SNAPSHOT

**Additional context**

Hi guys! I am huge fan of the work you are doing here.

I have been trying out the new organize imports feature. I work on a lot of projects that use Spark 2.x as a dependency, which means that they still use Scala 2.11.12.

I also managed to do some debugging.  Metals tries to download artifact `com.github.liancheng::organize-imports_2.11:0.4.0`, but the `organize-imports` scalafix rule is not published for 2.11.

A possible solution is to use scalafix and scalafix rules compiled for 2.12 regardless of project version. I am actually using the `organize-imports` rule (compiled for 2.12) in a 2.11.12 project with the [gradle-scalafix](https://github.com/cosmicsilence/gradle-scalafix) plugin and it works fine.

The downside is that this approach would hinder usage of some scalafix rules that need the scala version in the build to match scala version of scalafix (`ExplicitResultTypes ` for example). For me personally, that is trade-off I would be willing to accept for organize imports to work.

Regardless of the solution, user should be informed when a Code Action fails . The current implementation does not propagate failure information to LSP client - failure to fetch artifacts results success with no edits.

**Search terms**

organize imports, scalafix

gabro:
Thank you for for trying the feature and reporting this issue!

At a quick glance, it seems that the easiest solution would be to cross-publish `organize-imports` for 2.11 as well. I took a stab at it, here's the PR https://github.com/liancheng/scalafix-organize-imports/pull/113. The major notable thing is that 2.11 won't allow accurate removal of unused imports (due to buggy or missing warnings), but it's probably something we can live with.

You also make a good point about error reporting, but I'll leave it to @tgodzik / @mlachkar to decide what's the best strategy for that.

[...]

PR #2101, opened 2020-09-28
Update organize-import rule to add Scala 2.11 support

gabro:
Partially addresses #2099 by adding 2.11 support for organize imports (the error reporting part is still missing)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -221,7 +221,7 @@ lazy val V = new {
   val coursierInterfaces = "0.0.25"
   val ammonite = "2.2.0-4-4bd225e"
   val mill = "0.8.0"
-  val organizeImportRule = "0.4.0"
+  val organizeImportRule = "0.4.2"
 }
 
 val genyVersion = Def.setting {


tgodzik (approved):
Cool! Thanks for working on this! LGTM

mlachkar (approved):
Where do you think the error should be shown?  is it in Metals logs?

gabro:
@mlachkar I think the precise error (failed to download Scalafix, or failed to download the rules) should be logged and we should also display a "user-friendly" error in the editor, something like ("Organize imports failed due to an unexpected problem, check the logs for more information")

tgodzik:
@gabro @mlachkar  should we add `-Ywarn-unused-import` for Scala 2.11 here https://github.com/scalameta/metals/blob/a3d0d9488bdfe3119a7ca34060e37f642cd0591b/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala#L61 ?

gabro:
I think it's already covered by the else case, isn't it?

tgodzik:
> I think it's already covered by the else case, isn't it?

Ach nevermind, I somehow thought it's a different flag.

Issue #2009, opened 2020-01-06
Show implicit arguments


julienrf:
**Is your feature request related to a problem? Please describe.**

I’m always frustrated when I call a method that takes implicit parameters and I don’t know which argument values have been inferred by the compiler.

**Describe the solution you'd like**

It would be nice to have an action making the implicit arguments visible for a given selection, or globally visible.

Ideally, this action should be implemented (and specified) as a BSP endpoint, so that it could be reused by other IDEs (e.g., IntelliJ).

**Describe alternatives you've considered**

IntelliJ provides a “Show implicit arguments” action that, given a text selection, shows the inferred implicit arguments (note that all the arguments and their own arguments are shown as a tree):

![Screenshot from 2019-12-27 14-01-59](https://user-images.githubusercontent.com/332812/71832926-2dda2280-30ac-11ea-861b-cf0313247fb6.png)

They also have a “Show implicit hints” action that reveals all the implicit arguments and implicit conversions:

![Screenshot from 2020-01-06 17-49-30](https://user-images.githubusercontent.com/332812/71833310-eef89c80-30ac-11ea-93bb-c04312b0eb08.png)

**Search terms**:

implicits

PR #2103, opened 2020-09-28
Show implicit arguments and type annotations for Scala files 

tgodzik:
Resolves https://github.com/scalameta/metals/issues/2009

metals-sample-1601562464506

Some issues that I found when working on this:
- hovers on decorations inside text will pop up only before that decoration - https://github.com/microsoft/vscode/issues/105302
- not sure how to do go to definition for the implicits - I was thinking of maybe adding code lenses above or adding an option in definitionProvider to also search for synthetics if the cursor matches the end of sythetic occurence. I decided that we can work on it later

This also requires a PR for VS Code: https://github.com/scalameta/metals-vscode/pull/413

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
+++ b/metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
@@ -1,0 +1,138 @@
+package scala.meta.internal.decorations
+
+import scala.meta.internal.metals.UserConfiguration
+import scala.meta.internal.metap.PrinterSymtab
+import scala.meta.internal.semanticdb.Print
+import scala.meta.internal.{semanticdb => s}
+import scala.meta.metap.Format
+
+object SemanticdbTreePrinter {
+
+  def printSyntheticInfo(
+      textDocument: s.TextDocument,
+      synthetic: s.Synthetic,
+      printSymbol: String => String,
+      userConfig: UserConfiguration,
+      simple: Boolean
+  ): Option[String] = {
+
+    lazy val symtab = PrinterSymtab.fromTextDocument(textDocument)
+
+    def printTree(t: s.Tree): Option[String] =
+      t match {
+        case s.Tree.Empty => None
+        case s.OriginalTree(_) => None
+        case s.TypeApplyTree(function, typeArguments) =>
+          Some(printTree(function).getOrElse("") + printTypeArgs(typeArguments))
+        case s.ApplyTree(function, arguments) =>
+          Some(printTree(function).getOrElse("") + printArgs(arguments))
+        case s.LiteralTree(constant) =>
+          Some(printConstant(constant))
+        case s.SelectTree(_, id) =>
+          id.flatMap(printTree)
+        case s.FunctionTree(parameters, body) =>
+          printTree(body).map(printed => printArgs(parameters) + "=>" + printed)
+        case s.IdTree(symbol) =>
+          Some(printSymbol(symbol))
+        case s.MacroExpansionTree(beforeExpansion, _) =>
+          printTree(beforeExpansion)
+      }
+
+    def printPrefix(t: s.Type): String = {
+      printType(t) match {
+        case "" => ""
+        case s => s"$s."
+      }
+    }
+    def printTypeArgs(typeArgs: Seq[s.Type]): String =
+      typeArgs match {
+        case Nil => ""
+        case _ => typeArgs.map(printType).mkString("[", ", ", "]")
+      }
+
+    def printArgs(args: Seq[s.Tree]): String =
+      args match {
+        case Nil => ""
+        case _ => args.flatMap(printTree).mkString("(", ", ", ")")
+      }
+
+    def printType(t: s.Type): String =
+      t match {
+        case s.Type.Empty => ""
+        case s.RepeatedType(tpe) =>
+          s"${printType(tpe)}*"
+        case s.SingleType(prefix, symbol) =>
+          s"${printPrefix(prefix)}${printSymbol(symbol)}"
+        case s.TypeRef(prefix, symbol, typeArguments) =>
+          s"${printPrefix(prefix)}${printSymbol(symbol)}${printTypeArgs(typeArguments)}"
+        case s.WithType(types) =>
+          types.map(printType).mkString(" with ")
+        case s.ConstantType(constant) =>
+          printConstant(constant)
+        case s.ByNameType(tpe) =>
+          s"=> ${printType(tpe)}"
+        case s.ThisType(symbol) =>
+          s"this.${printSymbol(symbol)}"
+        case s.IntersectionType(types) =>
+          types.map(printType).mkString(" & ")
+        case s.UnionType(types) =>
+          types.map(printType).mkString(" | ")
+        case s.SuperType(prefix, symbol) =>
+          s"super.${printPrefix(prefix)}${printSymbol(symbol)}"
+        case s.AnnotatedType(annots, tp) =>
+          val mapped = annots
+            .map(x => s"@${printType(x.tpe)}")
+            .reduceLeft((x, y) => s"$x $y")
+          s"$mapped ${printType(tp)}"
+        // this should not need to be printed but just in case we revert to semanticdb printer
+        case s.UniversalType(_, tpe) =>
+          if (simple)
+            s"[ ... ] => ${printType(tpe)}"
+          else Print.tpe(Format.Detailed, t, symtab)
+        case s.ExistentialType(tpe, _) =>
+          if (simple)
+            s"${printType(tpe)} forSome { ... }"
+          else Print.tpe(Format.Detailed, t, symtab)
+        case s.StructuralType(tpe, _) =>
+          if (simple)
+            s" ${printType(tpe)} { ... }"
+          else Print.tpe(Format.Detailed, t, symtab)
+      }
+
+    def printConstant(c: s.Constant): String =
+      c match {
+        case s.FloatConstant(value) => value.toString
+        case s.LongConstant(value) => value.toString
+        case s.DoubleConstant(value) => value.toString
+        case s.NullConstant() => "null"
+        case s.IntConstant(value) => value.toString
+        case s.CharConstant(value) => value.toString
+        case s.ByteConstant(value) => value.toString
+        case s.UnitConstant() => "()"
+        case s.ShortConstant(value) => value.toString
+        case s.Constant.Empty => ""
+        case s.BooleanConstant(value) => value.toString
+        case s.StringConstant(value) => value
+      }
+
+    synthetic.tree match {
+      /**
+       *  implicit val str = ""
+       *  def hello()(implicit a : String)
+       *  hello()<<(str)>>
+       */
+      case tree @ s.ApplyTree(_: s.OriginalTree, _)
+          if userConfig.implicitArgumentAnnotationsEnabled =>
+        printTree(tree)
+
+      /**
+       *  def hello[T](T object) = object
+       *  hello<<[String]>>("")
+       */
+      case tree @ s.TypeApplyTree(_: s.OriginalTree, _)
+          if userConfig.typeAnnotationsEnabled =>
+        printTree(tree)
+      case _ => None
+    }
+  }
+}
--- a/metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
@@ -1,0 +1,324 @@
+package scala.meta.internal.decorations
+
+import java.nio.charset.Charset
+import java.nio.file.Path
+import java.util.concurrent.atomic.AtomicReference
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.util.Try
+
+import scala.meta.internal.metals.Buffers
+import scala.meta.internal.metals.ClientConfiguration
+import scala.meta.internal.metals.Diagnostics
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.UserConfiguration
+import scala.meta.internal.mtags.Md5Fingerprints
+import scala.meta.internal.mtags.SemanticdbClasspath
+import scala.meta.internal.mtags.Semanticdbs
+import scala.meta.internal.pc.HoverMarkup
+import scala.meta.internal.semanticdb.Scala._
+import scala.meta.internal.semanticdb.TextDocument
+import scala.meta.internal.semanticdb.TextDocuments
+import scala.meta.io.AbsolutePath
+
+import org.eclipse.lsp4j.TextDocumentPositionParams
+import org.eclipse.{lsp4j => l}
+
+final class SyntheticsDecorationProvider(
+    workspace: AbsolutePath,
+    semanticdbs: Semanticdbs,
+    buffer: Buffers,
+    client: DecorationClient,
+    fingerprints: Md5Fingerprints,
+    charset: Charset,
+    diagnostics: Diagnostics,
+    focusedDocument: () => Option[AbsolutePath],
+    clientConfig: ClientConfiguration,
+    userConfig: () => UserConfiguration
+)(implicit ec: ExecutionContext) {
+
+  import SemanticdbTreePrinter.printSyntheticInfo
+
+  private object Document {
+    /* We update it with each compilation in order not read the same file on
+     * each change. When typing documents stay the same most of the time.
+     */
+    private val document = new AtomicReference[TextDocument]
+
+    def currentDocument: Option[TextDocument] = Option(document.get())
+
+    def set(doc: TextDocument): Unit = document.set(doc)
+  }
+
+  def publishSynthetics(path: AbsolutePath): Future[Unit] = {
+
+    // If focused document is not supported we can't be sure what is currently open
+    def isFocusedDocument =
+      focusedDocument().contains(path) || !clientConfig.isDidFocusProvider()
+
+    /**
+     * Worksheets currently do not use semanticdb, which is why this will not work.
+     * If at any time worksheets will support it, we need to make sure that the
+     * evaluation will not be replaced with implicit decorations.
+     */
+    if (!path.isWorksheet && isFocusedDocument) {
+      val textDocument = currentDocument(path)
+      textDocument match {
+        case Some(doc) =>
+          publishSyntheticDecorations(path, doc)
+        case _ =>
+          Future.successful(())
+      }
+
+    } else
+      Future.successful(())
+  }
+
+  def onChange(textDocument: TextDocuments, path: Path): Unit = {
+    for {
+      focused <- focusedDocument()
+      filePath = AbsolutePath(path)
+      sourcePath <- SemanticdbClasspath.toScala(workspace, filePath)
+      if sourcePath == focused || !clientConfig.isDidFocusProvider()
+      textDoc <- textDocument.documents.headOption
+      source <- fingerprints.loadLastValid(sourcePath, textDoc.md5, charset)
+    } {
+      val docWithText = textDoc.withText(source)
+      Document.set(docWithText)
+      publishSyntheticDecorations(sourcePath, docWithText)
+    }
+  }
+
+  def addSyntheticsHover(
+      params: TextDocumentPositionParams,
+      pcHover: Option[l.Hover]
+  ): Option[l.Hover] =
+    if (isSyntheticsEnabled) {
+      val path = params.getTextDocument().getUri().toAbsolutePath
+      val line = params.getPosition().getLine()
+      val newHover = currentDocument(path) match {
+        case Some(textDocument) =>
+          val edit = buffer.tokenEditDistance(path, textDocument.text)
+
+          val syntheticsAtLine = for {
+            synthetic <- textDocument.synthetics
+            range <- synthetic.range.toIterable
+            lspRange <- edit.toRevisedStrict(range).toIterable
+            if lspRange.getEnd.getLine == line
+            fullSnippet <- printSyntheticInfo(
+              textDocument,
+              synthetic,
+              toHoverString(textDocument),
+              userConfig(),
+              simple = false
+            ).toIterable
+          } yield (lspRange, fullSnippet)
+
+          if (syntheticsAtLine.size > 0) {
+            if (clientConfig.isInlineDecorationProvider()) {
+              createHoverAtPoint(
+                path,
+                syntheticsAtLine,
+                pcHover,
+                params.getPosition()
+              )
+            } else {
+              createHoverAtLine(path, syntheticsAtLine, pcHover).orElse(
+                pcHover
+              )
+            }
+          } else {
+            None
+          }
+        case None => None
+      }
+      newHover.orElse(pcHover)
+    } else {
+      pcHover
+    }
+
+  private def isSyntheticsEnabled: Boolean = {
+    userConfig().implicitArgumentAnnotationsEnabled || userConfig().typeAnnotationsEnabled
+  }
+
+  private def createHoverAtPoint(
+      path: AbsolutePath,
+      syntheticsAtLine: Seq[(l.Range, String)],
+      pcHover: Option[l.Hover],
+      position: l.Position
+  ): Option[l.Hover] = {
+    syntheticsAtLine
+      .find {
+        case (range, _) =>
+          range.getEnd().getCharacter() == position.getCharacter()
+      }
+      .flatMap {
+        case (_, synthetic) =>
+          addToHover(
+            pcHover,
+            "**Synthetics**:\n"
+              +
+                HoverMarkup(
+                  synthetic
+                )
+          )
+      }
+
+  }
+
+  private def createHoverAtLine(
+      path: AbsolutePath,
+      syntheticsAtLine: Seq[(l.Range, String)],
+      pcHover: Option[l.Hover]
+  ): Option[l.Hover] =
+    Try {
+      val line = syntheticsAtLine.head._1.getEnd().getLine()
+      for {
+        currentText <- buffer.get(path)
+        lineText <- currentText.linesIterator.drop(line).headOption
+        created <- createLine(syntheticsAtLine, pcHover, lineText)
+      } yield created
+    }.toOption.flatten
+
+  private def createLine(
+      syntheticsAtLine: Seq[(l.Range, String)],
+      pcHover: Option[l.Hover],
+      lineText: String
+  ) = {
+    val withEnd = syntheticsAtLine
+      .map {
+        case (range, str) => (range.getEnd().getCharacter(), str)
+      }
+      .sortBy(_._1) :+ (lineText.size, "")
+    val lineWithDecorations = withEnd
+      .foldLeft(("", 0)) {
+        case ((current, index), (char, text)) =>
+          val toAdd = lineText.substring(index, char) + text
+          (current + toAdd, char)
+      }
+      ._1
+
+    addToHover(
+      pcHover,
+      "**With synthetics added**:\n"
+        + HoverMarkup(
+          lineWithDecorations.trim()
+        )
+    )
+  }
+
+  private def addToHover(
+      pcHover: Option[l.Hover],
+      text: String
+  ): Option[l.Hover] = {
+    // Left is not handled currently, but we do not use it in Metals
+    if (pcHover.exists(_.getContents().isLeft())) {
+      None
+    } else {
+      val previousContent = pcHover
+        .filter(_.getContents().isRight())
+        .map(_.getContents().getRight.getValue + "\n")
+        .getOrElse("")
+      Some(
+        new l.Hover(
+          new l.MarkupContent(
+            l.MarkupKind.MARKDOWN,
+            previousContent + text
+          )
+        )
+      )
+    }
+  }
+
+  private def currentDocument(path: AbsolutePath): Option[TextDocument] = {
+    Document.currentDocument match {
+      case Some(doc) if workspace.resolve(doc.uri) == path => Some(doc)
+      case _ =>
+        val textDocument = semanticdbs
+          .textDocument(path)
+          .documentIncludingStale
+        textDocument.foreach(Document.set)
+        textDocument
+    }
+  }
+
+  private def toSymbolName(symbol: String, textDoc: TextDocument): String = {
+    if (symbol.isLocal)
+      textDoc.symbols
+        .find(_.symbol == symbol)
+        .map(_.displayName)
+        .getOrElse(symbol)
+    else symbol
+  }
+
+  private def toHoverString(textDoc: TextDocument)(symbol: String): String = {
+    toSymbolName(symbol, textDoc)
+      .replace("/", ".")
+      .stripSuffix(".")
+      .stripSuffix("#")
+      .stripPrefix("_empty_.")
+      .replace("#", ".")
+      .replace("()", "")
+  }
+
+  private def toDecorationString(
+      textDoc: TextDocument
+  )(symbol: String): String = {
+    if (symbol.isLocal)
+      textDoc.symbols
+        .find(_.symbol == symbol)
+        .map(_.displayName)
+        .getOrElse(symbol)
+    else symbol.desc.name.value
+  }
+
+  private def decorationOptions(
+      lspRange: l.Range,
+      decorationText: String
+  ) = {
+    // We don't add hover due to https://github.com/microsoft/vscode/issues/105302
+    new DecorationOptions(
+      lspRange,
+      renderOptions = ThemableDecorationInstanceRenderOptions(
+        after = ThemableDecorationAttachmentRenderOptions(
+          decorationText,
+          color = "grey",
+          fontStyle = "italic",
+          opacity = 0.7
+        )
+      )
+    )
+  }
+
+  private def publishSyntheticDecorations(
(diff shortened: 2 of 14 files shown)


metals/src/main/scala/scala/meta/internal/decorations/ImplicitDecorationProvider.scala
```diff
@@ -0,0 +1,368 @@
...
+    } yield {
+
+      val diag = new l.Diagnostic(
+        lspRange,
+        fullMessage
+      )
+      diag.setSeverity(l.DiagnosticSeverity.Information)
+      diag
+    }
+    diagnostics.onPublishDiagnostics(path, diags, isReset = true)
```
tgodzik, line 221: This is breaking the tests, I need to work on that.

ckipp01:
This is really awesome @tgodzik! I'll take a more closer look at the implementation details in the coming days, but I first wanted to do a quick look over, understand what's going on, and then try it out. I mainly just have a couple questions on the UX for other clients besides VS Code.

It's cool to see that there is an option for clients that aren't decoration providers. However, I was curious what the diagnostics would look like because I had an inkling that it _may_ be a lot of noise.

Here is an example of me opening a random file with Vim (doesn't support virtual text like Nvim), so it's not a decoration provider.

Screenshot 2020-09-28 at 21 51 12

IMO, that's really distracting, and currently there isn't a way to turn this off, which would be one of my first thoughts on seeing it. What do you think about making a userConfiguration setting to just turn this entire feature off?

Sort of related to that I think if I was a newcomer to Scala and I hovered on this, I'd sort of have no idea what this means:

Screenshot 2020-09-28 at 22 01 47

I wonder if there is a way that we can include some more info to give a _bit_ more context about what the user is setting here. What do you think?

The other large concern I have (this is just me thinking out loud), is for clients that up to this point could say they supported the the decoration protocol may not be able to anymore. For example in Nvim, you can't inline virtual text, you can only append it to the end of the line. [There has been a bit of work on it](https://github.com/neovim/neovim/pull/9496), but I don't see it crossing the finishline any time soon. Therefore, in both coc-metals, and nvim-metals I say I have support for the decoration protocol, but that's because up to this point I could get away with grabbing the end line number, and setting the virtual text that way. I really need to think about how to approach this now since in reality, I don't fully support the protocol, but I still obviously want to support worksheets, which works fine. I have quite a few issues with coc-metals that I'll need to figure out for this. Just one small example is that since I can't inline them, I only ever really account for one on each line, so even given something simple like this:

<details><summary>metals/publishDecorations</summary>

```
[Trace - 09:56:24 PM] Sending notification 'metals/publishDecorations'
Params: {
  "uri": "file:///Users/ckipp/Documents/scala-workspace/tester/src/main/scala/example/Main.scala",
  "options": [
    {
      "range": {
        "start": {
          "line": 19,
          "character": 2
        },
        "end": {
          "line": 19,
          "character": 18
        }
      },
      "hoverMessage": {
        "kind": "markdown",
        "value": "```scala\n(sourcecode.LineMacros.generate, sourcecode.FileNameMacros.generate)\n```"
      },
      "renderOptions": {
        "after": {
          "contentText": "(generate, generate)",
          "fontStyle": "italic",
          "color": "grey",
          "opacity": 0.7
        }
      }
    },
    {
      "range": {
        "start": {
          "line": 19,
          "character": 2
        },
        "end": {
          "line": 19,
          "character": 12
        }
      },
      "hoverMessage": {
        "kind": "markdown",
        "value": "```scala\n[java.lang.String]\n```"
      },
      "renderOptions": {
        "after": {
          "contentText": "[String]",
          "fontStyle": "italic",
          "color": "grey",
          "opacity": 0.7
        }
      }
    }
  ]
}

</details>

Looks like this in coc-metals:

<img width="274" alt="Screenshot 2020-09-28 at 22 20 01" src="https://user-images.githubusercontent.com/13974112/94481906-c5eadf80-01d8-11eb-849a-8d9ce657ed0e.png">

Whereas it's like this in VS Code

<img width="294" alt="Screenshot 2020-09-28 at 22 20 30" src="https://user-images.githubusercontent.com/13974112/94481944-d602bf00-01d8-11eb-844e-36963247975e.png">

I'm not familiar enough with the VS Code decoration api to know how it prioritizes what it displays vs what it puts in the hover, but I'm assuming it has to do with the range.

Either way,  all that to say, I know that is on me, and I need to figure it out on the client side. I just bring it up because I know there also may be other clients that up to the point have been able to say the support the decoration protocol, but now won't be able to say that. Which leads me to the question of is there any way to differentiate support for decorations in worksheets vs decorations here since one is quite simple and the other is quite involved.

File: metals/src/main/scala/scala/meta/internal/decorations/ImplicitDecorationProvider.scala [hunk 2]
```diff
@@ -0,0 +1,368 @@
...
+      if (clientConfig.isDecorationProvider()) {
+        publishDecorations(path, textDocument)
+      } else {
+        publishDiagnostics(path, textDocument)
+      }
+    }
+
+}
+
+object SemanticdbTreePrinter {
```
gabro, line 239: I would probably move this to a separate file

metals/src/main/scala/scala/meta/internal/decorations/ImplicitDecorationProvider.scala
```diff
@@ -0,0 +1,368 @@
...
+import scala.meta.internal.semanticdb.Scala._
+import scala.meta.internal.semanticdb.TextDocument
+import scala.meta.internal.semanticdb.TextDocuments
+import scala.meta.internal.{semanticdb => s}
+import scala.meta.io.AbsolutePath
+import scala.meta.metap.Format
+
+import org.eclipse.{lsp4j => l}
+
+final class ImplicitDecorationProvider(
```
gabro, line 29: About the name, does this only print implicit values? Doesn't it also print inferred type parameters and other synthetics?

gabro:
This looks great @tgodzik! I've left a couple of quick comments, but I'll hopefully have time to go back and review this more thoroughly

tgodzik:
Thanks for the feedback @ckipp01 ! I started to really overthink parts of the PR here, so I decided to at least open a PR to get some opinions. We can certainly change a lot still, especially when it comes to non-decoration clients.

> IMO, that's really distracting, and currently there isn't a way to turn this off, which would be one of my first thoughts on seeing it. What do you think about making a userConfiguration setting to just turn this entire feature off?

We do already turn both implicit arguments and inferred types by default off. So that shouldn't be an issue normally, though I would need to fix the commented part. I was also not super about the UI here really, but was thinking users would turn it on only if needed to maybe figure out some implicits? 

> Sort of related to that I think if I was a newcomer to Scala and I hovered on this, I'd sort of have no idea what this means:
> I wonder if there is a way that we can include some more info to give a _bit_ more context about what the user is seeing here. What do you think?

Yeah, I think we need to work on the alternative diagnostic mode, I will write up some ideas at the end.

> The other large concern I have (this is just me thinking out loud), is for clients that up to this point could say they supported the the decoration protocol may not be able to anymore. 
> I'm not familiar enough with the VS Code decoration api to know how it prioritizes what it displays vs what it puts in the hover, but I'm assuming it has to do with the range. So to account for this I'd either need to add multiple to the end of the line, which probably won't look good, or somehow display it in the hover.
> 
> Either way, all that to say, I know that is on me, and I need to figure it out on the client side, but I bring it up because I know there may also be other clients that up to the point have been able to say the support the decoration protocol, but now won't be able to say that. Which leads me to the question of is there any way to differentiate support for decorations in worksheets vs decorations here since one is quite simple and the other is quite involved. What do you think?
> 

We should add a separate option - something like `inlineDecorationsSupported`. We could maybe print a copy of the line on the right with added implicits? Or just fallback to the other option (diagnostic or whatever we figure out)

Some other ideas how we can display implicits for non-decoration clients:
- Code lenses that display specific information diags with code fragment including full line and added implicits 
- the same as above but display them on executeCommand
- we could only display information diags with a range limited to the end, maybe it will be less distracting?

I am going to experiment a bit with that.

kpbochenek (changes requested):
As always outstanding work :muscle: :+1: 

Just a question about dotty, I look at changes and it seems for me like it should just work? But I don't get decorations in dotty :( 
Curious if it should not work for dotty where is a difference compared to scala2?

tgodzik:
> As always outstanding work  
> 
> Just a question about dotty, I look at changes and it seems for me like it should just work? But I don't get decorations in dotty :(
> Curious if it should not work for dotty where is a difference compared to scala2?

There are not synthetics reported in dotty currently :/

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -169,6 +171,26 @@ object UserConfiguration {
            |
            |""".stripMargin
       ),
+      UserConfigurationOption(
+        "type-annotations-enabled",
+        "false",
```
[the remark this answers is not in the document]
tgodzik: My bad! I changed it for testing :sweat_smile:

metals/src/main/scala/scala/meta/internal/decorations/ImplicitDecorationProvider.scala
```diff
@@ -0,0 +1,368 @@
...
+  def publishDecorators(path: AbsolutePath): Future[Unit] = {
+
+    /**
+     * Worksheets currently do not use semanticdb, which is why this will not work.
+     * If at any time worksheets will support it, we need to make sure that the
+     * evaluation will not be replaced with implicit decorations.
+     */
+    if (
+      !path.isWorksheet && (focusedDocument().contains(path) || !clientConfig
+        .isDidFocusProvider())
```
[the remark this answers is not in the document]
tgodzik: If it's not the didFocusProvider then we cannot count on the file being the right one in the focusedDocument(), so we publish anyway. Might not be the best choice, but not sure what is better.

metals/src/main/scala/scala/meta/internal/decorations/ImplicitDecorationProvider.scala
```diff
@@ -0,0 +1,368 @@
...
+  ) = {
+
+    def decoration(range: l.Range, text: String, hover: l.MarkupContent) =
+      new DecorationOptions(
+        range,
+        hover,
+        ThemableDecorationInstanceRenderOptions(
+          after = ThemableDecorationAttachmentRenderOptions(
+            text,
+            color = "grey",
```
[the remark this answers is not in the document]
tgodzik: I think that's an error in VS Code, there is still a number of things that don't work too perfect.

metals/src/main/scala/scala/meta/internal/decorations/ImplicitDecorationProvider.scala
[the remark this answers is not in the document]
kpbochenek: oookay that makes perfect sense :+1:

kpbochenek:
Not 'reported' means they are not added to generated Semanticdb? 
Also when I look for references of `implicit x: String` it finds me usages as implicit parameter.

Do we use compiler for this feature?

tgodzik:
> Not 'reported' means they are not added to generated Semanticdb?
> Also when I look for references of `implicit x: String` it finds me usages as implicit parameter.
> 
> Do we use PC for this feature?

Things are generated in semanticdb but not the synthetics section. Maybe we can go around it for Dotty, or rather it would be good to fix in the compiler itself.

metals/src/main/scala/scala/meta/internal/decorations/ImplicitDecorationProvider.scala
```diff
@@ -0,0 +1,368 @@
...
+    clientConfig: ClientConfiguration,
+    userConfig: () => UserConfiguration
+)(implicit ec: ExecutionContext) {
+
+  import SemanticdbTreePrinter.printSyntheticInfo
+
+  /* We update it with each compilation in order not read the same file on
+   * each change. When typing documents stay the same most of the time.
+   */
+  val lastDocument = new AtomicReference[TextDocument]
```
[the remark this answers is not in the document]
tgodzik: I reworked that a bit. I don't really like `AtomicReference[Option[TextDocument]]`, so I put it all into a private object. Let me know if that is ok.

metals/src/main/scala/scala/meta/internal/decorations/ImplicitDecorationProvider.scala
```diff
@@ -0,0 +1,368 @@
...
+
+  def publishDecorators(path: AbsolutePath): Future[Unit] = {
```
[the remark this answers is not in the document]
tgodzik: Changed to `publishSynthetics` which makes more sense in this case.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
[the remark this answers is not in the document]
tgodzik: Fixed.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -169,6 +171,26 @@ object UserConfiguration {
...
+        "Should display type annotations for infered types",
+        """|When this option is enabled, for each method that can have infered types they are 
+           |displayed either as additional decorations if they are supported for the editor or 
+           |as info diagnostics otherwise.
+           |""".stripMargin
+      ),
+      UserConfigurationOption(
+        "implicit-argument-annotations-enabled",
+        "false",
```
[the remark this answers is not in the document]
tgodzik: Fixed

tests/unit/src/test/scala/tests/decorations/ImplicitDecorationsLspSuite.scala
```diff
@@ -0,0 +1,119 @@
...
+           |  implicit val ec: scala.concurrent.ExecutionContext = scala.concurrent.ExecutionContext.global
+           |  Future{
+           |    println("")
+           |  }
+           |}
+           |""".stripMargin
+      )
+      _ <- server.didChangeConfiguration(
+        """{
+          |  "implicit-argument-annotations-enabled": true,
```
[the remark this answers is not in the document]
tgodzik: Added!

metals/src/main/scala/scala/meta/internal/decorations/ImplicitDecorationProvider.scala
tgodzik: Moved!

metals/src/main/scala/scala/meta/internal/decorations/ImplicitDecorationProvider.scala
tgodzik: Renamed to `SyntheticsDecorationProvider`

tgodzik:
@ckipp01 Added `isInlineDecorationProvider`, so it should work alright in case of other editors. 

What do you think about the idea to duplicate the line with the additional text added in case of non inline decorations?
Would be something like:
`"foo".map(c => c.toUpper) // "foo".map[scala.Char, scala.Predef.String](c => c.toUpper)`

similarily for decorations:
```
           |a/src/main/scala/com/example/Main.scala:15:3: information: "foo".map[scala.Char, scala.Predef.String](c => c.toUpper)
           |  "foo".map(c => c.toUpper)
           |  ^^^^^^^^^
```

Alternatively we could make it work via additional lenses that publish the diagnostics/decorations on demand. Or just use diagnotics for both? I can try work on demo, what do you think?

tgodzik:
Or! We could actually just use hover for implicits without any additional diagnostics! Let me quickly demo that

metals/src/main/scala/scala/meta/internal/decorations/SyntheticDecorationProvider.scala
```diff
@@ -0,0 +1,240 @@
...
+        case _ =>
+          val textDocument = semanticdbs
+            .textDocument(path)
+            .documentIncludingStale
+          textDocument.foreach(Document.set)
+          textDocument
+      }
+      textDocument match {
+        case Some(doc) =>
+          publishImplicitInfo(path, doc)
```
ckipp01, line 74:
```suggestion
          publishSyntheticInfo(path, doc)
```
This isn't just publishing implicits right, but other synthetics?

metals/src/main/scala/scala/meta/internal/decorations/SyntheticDecorationProvider.scala
```diff
@@ -0,0 +1,240 @@
...
+    else symbol.desc.name.value
+  }
+
+  private def decorate(
+      lspRange: l.Range,
+      decorationText: String,
+      hoverMessage: String
+  ) = {
+
+    def decoration(range: l.Range, text: String, hover: l.MarkupContent) =
```
ckipp01, line 134:
```suggestion
    def getDecorationOptions(range: l.Range, text: String, hover: l.MarkupContent) =
```

metals/src/main/scala/scala/meta/internal/decorations/SyntheticDecorationProvider.scala
```diff
@@ -0,0 +1,240 @@
...
+  ): Unit = {
+    val decorations = for {
+      synthetic <- textDocument.synthetics
+      range <- synthetic.range.toIterable
+      edit = buffer.tokenEditDistance(path, textDocument.text)
+      decoration <- printSyntheticInfo(
+        textDocument,
+        synthetic,
+        toDecorationString(textDocument),
+        userConfig(),
```
ckipp01, line 179: I was curious why we passed in `userConfig()` here. TIL that we have `implicitArgumentAnnotationsEnable` and `typeAnnotationsEnabled` settings.

metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
```diff
@@ -0,0 +1,138 @@
+package scala.meta.internal.decorations
+
+import scala.meta.internal.metals.UserConfiguration
+import scala.meta.internal.metap.PrinterSymtab
+import scala.meta.internal.semanticdb.Print
+import scala.meta.internal.{semanticdb => s}
+import scala.meta.metap.Format
+
+object SemanticdbTreePrinter {
```
ckipp01, line 9: Super random, but I happened to look at the git blame for this file and it shows that this was added 2019-11-30? Why is it showing that?

ckipp01: Screenshot 2020-10-01 at 11 23 13

Actually I think somehow your force push is showing up as last year?

metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
```diff
@@ -0,0 +1,138 @@
...
+          if (simple)
+            s"${printType(tpe)} forSome { ... }"
+          else Print.tpe(Format.Detailed, t, symtab)
+        case s.StructuralType(tpe, _) =>
+          if (simple)
+            s" ${printType(tpe)} { ... }"
+          else Print.tpe(Format.Detailed, t, symtab)
+      }
+
+    def printCostant(c: s.Constant): String =
```
ckipp01, line 102:
```suggestion
    def printConstant(c: s.Constant): String =
```

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -198,6 +199,7 @@ class MetalsLanguageServer(
   private var renameProvider: RenameProvider = _
   private var documentHighlightProvider: DocumentHighlightProvider = _
   private var formattingProvider: FormattingProvider = _
+  private var syntheticsDecorator: SyntheticDecorationProvider = _
```
ckipp01, line 202: NIT: Can we align with plural or not?

```suggestion
  private var syntheticDecorator: SyntheticDecorationProvider = _
```
or 

```suggestion
  private var syntheticsDecorator: SyntheticsDecorationProvider = _
```

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -169,6 +171,26 @@ object UserConfiguration {
...
            |""".stripMargin
       ),
+      UserConfigurationOption(
+        "type-annotations-enabled",
+        "false",
+        "false",
+        "Should display type annotations for infered types",
+        """|When this option is enabled, for each method that can have infered types they are 
+           |displayed either as additional decorations if they are supported for the editor or 
+           |as info diagnostics otherwise.
```
ckipp01, line 181:
```suggestion
        """|When this option is enabled, each method that can have inferred types has them 
           |displayed either as additional decorations if they are supported by the editor or 
           |as info diagnostics.
```

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -169,6 +171,26 @@ object UserConfiguration {
...
+           |""".stripMargin
+      ),
+      UserConfigurationOption(
+        "implicit-argument-annotations-enabled",
+        "false",
+        "false",
+        "Should display implicit parameter at usage sites",
+        """|When this option is enabled, for each method that has implicit arguments they are 
+           |displayed either as additional decorations if they are supported for the editor or 
+           |as info diagnostics otherwise.
```
ckipp01, line 191:
```suggestion
        """|When this option is enabled, each method that has implicit arguments has them 
           |displayed either as additional decorations if they are supported by the editor or 
           |as info diagnostics.
```

ckipp01 (changes requested):
Again, great job on this @tgodzik! Finally had some more time to dig through this. I'll build again and make some changes to `coc-metals` to test this out and give some more feedback!

metals/src/main/scala/scala/meta/internal/decorations/SyntheticDecorationProvider.scala
```diff
@@ -0,0 +1,240 @@
...
+    // If focused document is not supported we can't be sure what is currently open
+    def isFocusedDocument =
+      focusedDocument().contains(path) || !clientConfig.isDidFocusProvider()
+
+    /**
+     * Worksheets currently do not use semanticdb, which is why this will not work.
+     * If at any time worksheets will support it, we need to make sure that the
+     * evaluation will not be replaced with implicit decorations.
+     */
+    if (!path.isWorksheet && isFocusedDocument) {
```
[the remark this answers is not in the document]
tgodzik: As long as we have synthetics it's fine to publish the decorations. Ammonite should be good and sbt will not have semanticdb produced,

metals/src/main/scala/scala/meta/internal/decorations/SyntheticDecorationProvider.scala
```diff
@@ -0,0 +1,240 @@
...
+    def currentDocument: Option[TextDocument] = Option(document.get())
+
+    def set(doc: TextDocument): Unit = document.set(doc)
+  }
+
+  def publishSynthetics(path: AbsolutePath): Future[Unit] = {
+
+    // If focused document is not supported we can't be sure what is currently open
+    def isFocusedDocument =
+      focusedDocument().contains(path) || !clientConfig.isDidFocusProvider()
```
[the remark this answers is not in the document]
tgodzik: So the idea behind is that if we don't provide focused document, we should try to publish anyway, since we don't know what the focused document is.

metals/src/main/scala/scala/meta/internal/decorations/SyntheticDecorationProvider.scala
tgodzik: This is just for invoking this method, so it doesn't really matter. It will be invoked each time with new user settings.

metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
tgodzik: haha, that's when I started looking into it :P so actually it's kind of right. Maybe I overdid it with the amends though.

metals/src/main/scala/scala/meta/internal/decorations/SyntheticDecorationProvider.scala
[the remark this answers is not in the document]
ckipp01: Ahh, ok, that makes sense then. 👍

metals/src/main/scala/scala/meta/internal/decorations/SyntheticDecorationProvider.scala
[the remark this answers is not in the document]
ckipp01: Sounds good 👍

metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
ckipp01: O wow haha, ok. I might be best to ammend the commit date to better reflect _this year_ 😆

metals/src/main/scala/scala/meta/internal/decorations/SyntheticDecorationProvider.scala
ckipp01: ha, yes, was just a comment when I looked at when the settings were added and say it was a year ago, so I thought they were in there the entire time without me every seeing them. 😆 but as I mentioned before, that was just the odd timestamp, so now worries!

metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
tgodzik: Amended it with current dat, thanks for noticing!

tests/unit/src/test/scala/tests/decorations/SyntheticHoverLspSuite.scala
```diff
@@ -0,0 +1,64 @@
...
+        """|**This line with synthetics added**:
+           |```scala
+           |hello()(com.example.Main.andy, boston)
+           |```
+           |""".stripMargin
+      )
+      _ <- server.assertHoverAtLine(
+        "a/src/main/scala/com/example/Main.scala",
+        "  \"foo\".map(c @@=> c.toUpper)",
+        """|**This line with synthetics added**:
```
tgodzik, line 56: @ckipp01 Change diagnostics to displaying the entire line with hover with shown additional synthetic info. What do you think?

metals/src/main/scala/scala/meta/internal/decorations/ImplicitDecorationProvider.scala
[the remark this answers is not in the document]
kpbochenek: ahh okay :+1:

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
tgodzik: Fixed and updated with info about hover.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
tgodzik: Fixed and updated with info about hover.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: Fixed to display plural in both.

metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
tgodzik: Fixed

metals/src/main/scala/scala/meta/internal/decorations/SyntheticDecorationProvider.scala
tgodzik: Changed

metals/src/main/scala/scala/meta/internal/decorations/SyntheticDecorationProvider.scala
tgodzik: Changed the name and actually dropped one method after removing diagnostics usage

metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
```diff
@@ -39,6 +39,7 @@ final case class InitializationOptions(
     compilerOptions: CompilerInitializationOptions,
     debuggingProvider: Option[Boolean],
     decorationProvider: Option[Boolean],
+    inlineDecorationProvider: Option[Boolean],
```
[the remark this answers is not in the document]
tgodzik: Fixed

metals/src/main/scala/scala/meta/internal/decorations/ImplicitDecorationProvider.scala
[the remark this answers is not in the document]
kpbochenek: okay :+1: , can we just make `new AtomicReference[TextDocument](null)` to explicitly state we allow nulls?
If it is clear for you just ignore, I am paranoic about nulls :P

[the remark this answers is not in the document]
tgodzik: AmoticReference starts with null so I would leave as is, but if you really want it I can add the null explicitely.

tests/unit/src/test/scala/tests/decorations/SyntheticHoverLspSuite.scala
ckipp01: So playing around with this, for me personally, this is _WAY_ nicer. I really like it. The only tricky part is that the user isn't notified that this extra information is here. However, like the original screen shot, if they are notified with the info diagnostic, imo it's way way too noisy. When playing around with it again before you added hovers it bummed me out because I really wanted to like it, but just having the info in randomly in the hover with no context wasn't super helpful. It was also super distracting to see the info diagnostics literally everywhere.

I think this solution is a much better approach 👍

ckipp01: Screenshot 2020-10-01 at 13 34 36

Here is what that example looks like in real life if you're curious.

kpbochenek (approved):
once again nice work :+1:

metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
```diff
@@ -0,0 +1,300 @@
...
+        val previousContent = pcHover
+          .filter(_.getContents().isRight())
+          .map(_.getContents().getRight.getValue + "\n")
+          .getOrElse("")
+
+        new l.Hover(
+          new l.MarkupContent(
+            l.MarkupKind.MARKDOWN,
+            previousContent
+              + "**This line with synthetics added**:\n"
```
ckipp01, line 166: What do you think of simply this?
```suggestion
              + "_**With synthetics added**_:\n"
```

tests/unit/src/test/scala/tests/decorations/SyntheticHoverLspSuite.scala
tgodzik: Yeah, diagnostics are way to noisy and could be quite fragile.

I wonder why the markdown is not working there, seemed to work ok in VS Code. Any idea?

ckipp01: And a larger example:

Screenshot 2020-10-01 at 13 39 35

metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
ckipp01: Actually hold off on this. I may be wrong.

From up above:

> I wonder why the markdown is not working there, seemed to work ok in VS Code. Any idea?

No need to worry about that, it's expected. Keep in mind that I'm in a terminal, so while some hacks are done in coc to make things bold etc, not everything will work as expected.

Also the bold in the italics doesn't work as expected, so don't do that. I'd just change the wording and keep the double `**`.

tests/unit/src/test/scala/tests/decorations/SyntheticHoverLspSuite.scala
ckipp01: > I wonder why the markdown is not working there, seemed to work ok in VS Code. Any idea?

I answered down below with more info, but basically I'm in a terminal ha. So there is no markdown rendering.

metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
tgodzik: Updated, thanks!

tgodzik:
I changed the hover to be also show on actually hover request and not added together with decorations due to https://github.com/microsoft/vscode/issues/105302

Now it looks like this:

metals-sample-1601562464506

ckipp01 (approved):
Really great work @tgodzik! LGTM 🚀

tgodzik:
I will wait with the merge until after the weekend, so if anyone wants to take a look feel free to do it :smiley:

tgodzik:
Go to definition was not covered by this PR, so I created a new feature here: https://github.com/scalameta/metals-feature-requests/issues/155

I would love some feedback on how we can add it in Metals, because it does not seem obvious.

Issue #2100, opened 2020-09-26
Organize import warning message about saving the file is a bit miselading


gabro:
I've been testing the organize imports feature, and I think the warning message we get when trying to organize imports of an unsaved file is a bit misleading.

If I understand correctly it's not really about the file being saved, rather the file being compiled (aka semanticdb has been produced). If that's the case, the warning message is a bit misleading because after saving the file the user also needs to wait for compilation to be over.

I found this while trying VSCode's `editor.codeActionsOnSave` command with this configuration:

```json
  "[scala]": {
    "editor.codeActionsOnSave": {
      "source.organizeImports": true
    }
  },
```

This will run the "organize imports" action right after the file is saved, but when I do that I still get the warning from Metals about saving the file first. If I save the file a second time, the action is ran correctly (because the file has been compiled in the meantime).

My proposal would be to do something similar to what we do in RenameProvider to monitor active compilations and trigger the warning based on that (rather than on the file being saved or not).

What do you think @mlachkar / @tgodzik?

---

As an aside, do we only need the semanticdb file of the current file for this to work? If that's the case maybe we can produce it using the PC and try to avoid this limitation?

tgodzik:
Thanks for testing it out! We've been meaning to also work on producing semanticdb for unsaved files and that might also unblock SBT files. The only problem is that we need to write that semanticdb files to the filesystem, but I think that shouldn't be that problematic.

We could do the same as for the rename provider for the time being, good idea!

PR #2104, opened 2020-09-29
Update Sublime doc regarding organize imports

ayoub-benali:
Currently it seems that [LSP](https://github.com/sublimelsp/LSP/) supports `source.organizeImports` only as code action on save.

### DIFF
--- a/docs/editors/sublime.md
+++ b/docs/editors/sublime.md
@@ -86,6 +86,17 @@ This paragraph contains a few tips & trick that can improve your daily productiv
 
 ### Optional LSP client tweaks
 
+If you want Metals to automatically organize the imports on file save set the following setting in the "Preferences > Preferences: LSP Settings":
+
+```json
+{
+  // ...
+  "lsp_code_actions_on_save": {
+    "source.organizeImports": true
+  }
+}
+```
+
 If you prefer to only enable Metals completions
 (without mixing them with the default ones from Sublime) set the following setting
 in the "Preferences > Preferences: LSP Settings":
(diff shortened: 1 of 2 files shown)


docs/editors/sublime.md
```diff
@@ -86,6 +86,17 @@ This paragraph contains a few tips & trick that can improve your daily productiv
 
 ### Optional LSP client tweaks
 
+If you want Metals to automatically organize the imports on file save set the following setting in the "Preferences > Preferences: LSP Settings":
+
+```json
+{
+  // ...
+  "lsp_code_actions_on_save": {
+    "source.organizeImports": true
```
gabro, line 95: unfortunately this will not work as expected.

Metals currently needs the file to have been compiled before being able to run Organize Imports, so this will go as follow:

- save
- get a warning about the file not being saved (confusing, I know, see https://github.com/scalameta/metals/issues/2100)
- wait for compilation to end
- save again
- now it works

See the linked issue for a more in-depth discussion

ayoub-benali: Hmm I see. So to get the feature working with the current Metals limitation the code action needs to be somehow exposed to the client like in Vim?

kpbochenek: `Currently it seems that LSP supports source.organizeImports only as code action on save.`
LSP in sublime? LSP specification defines it as a code action available anywhere.
You can simply bind it to shortcut and execute anytime you want at least in emacs/VSCode.

tgodzik: We might need to fix #2100 first I think before this will work sublime then.

ayoub-benali: OK I will add specific command to invoke `source.organizeImports` manually. 
But why isn't this code action available when invoking the general code action next to imports ?

kpbochenek: we have to start with something, we can definitely improve it

Issue #2116, opened 2020-05-09
JVM options and environmental variables in Vscode run configuration. 


pshemass:
**Is your feature request related to a problem? Please describe.**

When I run/debug application during development in many occasions I need to override system properties.

**Describe the solution you'd like**

It would be great if run configuration will support `vmArgs` and `env` the same way as java extension https://code.visualstudio.com/docs/java/java-debugging

tgodzik:
@mwz That is a separate issue, but should be easy to do.

We would need an `envFile` value next to the env one and we could read it in Metals and add to the rest of values. 

Feel free to create a PR or feature request. It would all be a very similar issue to https://github.com/scalameta/metals/pull/2118

mwz:
> We would need an `envFile` value next to the env one and we could read it in Metals and add to the rest of values.
> 
> Feel free to create a PR or feature request. It would all be a very similar issue to #2118

Ok, yeah it seems straightforward enough, let me have a look into that then. 👍

tgodzik:
@mwz Let me know if you have any issues! https://scalameta.org/metals/docs/contributors/getting-started.html should have all you need

PR #2118, opened 2020-10-01
Support environment variables when running or debugging

tgodzik:
Fixes https://github.com/scalameta/metals/issues/2116

Connected PR: https://github.com/scalameta/metals-vscode/pull/414


metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -373,7 +374,8 @@ case class DebugUnresolvedMainClassParams(
     mainClass: String,
     @Nullable buildTarget: String = null,
     @Nullable args: java.util.List[String] = null,
-    @Nullable jvmOptions: java.util.List[String] = null
+    @Nullable jvmOptions: java.util.List[String] = null,
+    @Nullable env: java.util.List[String] = null
```
ckipp01, line 378: Can we also update the docs here: https://scalameta.org/metals/docs/editors/new-editor.html#start-debug-adapter to show this in the example?

ckipp01 (approved):
Just a small note about updating the docs, but other than that, LGTM!

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
tgodzik: Good thinking! I think I will also change it to Map[String, String] if possible

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -128,7 +128,8 @@ object ServerCommands {
         |   mainClass: "com.foo.App",
         |   buildTarget: "foo",
         |   args: ["bar"],
-        |   jvmOptions: ["-Dfile.encoding=UTF-16"]
+        |   jvmOptions: ["-Dfile.encoding=UTF-16"],
+        |   env: ["NUM=123"]
```
ckipp01, line 132: Ah, I may have misunderstood the VS Code PR. So it's filled in as Map, but pass in as an Array of `key=value` right?

tgodzik: My bad! I we do send it as map, but it's changed to list when sent via bsp. Need to fix this.

ckipp01 (approved):
👍

tests/unit/src/test/scala/tests/DebugProtocolSuite.scala
```diff
@@ -39,18 +47,14 @@ class DebugProtocolSuite extends BaseDapSuite("debug-protocol") {
...
-        )
+        mainClass
       )
       _ <- debugger.initialize
       _ <- debugger.launch
       _ <- debugger.configurationDone
       _ <- debugger.shutdown
       output <- debugger.allOutput
-    } yield assertNoDiff(output, "FooBar")
+    } yield assertNoDiff(output, "FooBarFoo")
```
mwz, line 57: I've tried to adapt this test to reflect my changes from #2123, but this test keeps failing for me even when I try to run it locally from the main branch and I can't quite figure out yet why. 🤔 

When I run `sbt unit/testOnly *DebugProtocolSuite* -- *.start*` I get the following:
```diff
=> Diff (- obtained, + expected)
-FooBar
+FooBarFoo
```

Do I need to do anything special with my local setup or the way those tests are run to get this test to pass?

I tried running them via `bin/test.sh "unit/testOnly *DebugProtocolSuite* -- *.start*"` and I can see in the logs bloop being set to `v1.4.4-13` rather than `1.4.4-17` which is what you changed it to in this PR.

Am I missing something obvious here? Any help or pointers would be much appreciated.

tgodzik: no worries, we should not be probably using snapshot Bloop versions. I always creates additional confusion.

Issue #2102, opened 2020-09-28
Error reporting when organize imports fails


pvid:
**Describe the bug**

Follow-up to https://github.com/scalameta/metals/issues/2099. The issue with actual organizing of imports has been fixed. However, it would be useful to improve error reporting in case a different error occurs

**To Reproduce**

1. Find a project using Scala 2.11.12 (For example https://github.com/scalapb/sparksql-scalapb/tree/v0.9.3 and manually change sparkSqlScalaPB scala version to 2.11.12)
2. Mess up imports in some file
3. Run organize imports
4. See that nothing happens and no error is reported

**Expected behavior**

Metals should provide clear information why organize imports failed

**Installation:**

- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: 0.9.4+15-a3d0d948-SNAPSHOT

**Additional context**

Errors encountered during fetching of `organize-imports` artifact do not get reported because of the use of a `Try(...).toOption` pattern

**Search terms**

organize imports, scalafix

tgodzik:
Thanks for the follow up!

PR #2119, opened 2020-10-02
Make sure files are compiled when running scalafix 

tgodzik:
This fixes both:
- reporting errors - fixes https://github.com/scalameta/metals/issues/2102

Unfortunately we currently do not have any more information about the actual error message, it will need to be worked on further in scalafix. This hopefully enough for now, most error will be connected to semanticdb not being produced.

- making sure that compilation is run when using scalafix - fixes https://github.com/scalameta/metals/issues/2100

We are not yet able to organize imports in an unsaved file, but I will follow up with an issue for that.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
@@ -47,6 +47,16 @@ final class Compilations(
   def wasPreviouslyCompiled(buildTarget: b.BuildTargetIdentifier): Boolean =
     lastCompile.contains(buildTarget)
 
+  def compilationFinished(
+      source: AbsolutePath
+  ): Future[Unit] = {
+    if (currentlyCompiling.isEmpty) {
+      Future(())
+    } else {
+      cascadeCompileFiles(Seq(source))
+    }
+  }
+
   def compileTarget(
       target: b.BuildTargetIdentifier
   ): Future[b.CompileResult] = {
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
@@ -1,17 +1,20 @@
 package scala.meta.internal.metals
 
 import java.net.URLClassLoader
-import java.nio.file.Path
 import java.util.Collections
-import java.util.{List => JList}
 
 import scala.collection.concurrent.TrieMap
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.util.Failure
+import scala.util.Success
 import scala.util.Try
 
 import scala.meta._
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.io.AbsolutePath
 
+import org.eclipse.lsp4j.MessageParams
 import org.eclipse.lsp4j.MessageType
 import org.eclipse.{lsp4j => l}
 import scalafix.interfaces.Scalafix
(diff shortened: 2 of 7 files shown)


pvid:
🎉 could you please also take a look at lines [131](https://github.com/tgodzik/metals/blob/acbea734a7228fc75e3a9868142a428c28374295/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala#L131) and [148](https://github.com/tgodzik/metals/blob/acbea734a7228fc75e3a9868142a428c28374295/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala#L148) of `ScalafixProvider`? Errors encountered during fetching of artifacts are not reported. I had an issue with no errors reported with the latter when trying organize imports with 2.11 before a version of the rule for 2.11 was published

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -35,52 +39,76 @@ case class ScalafixProvider(
       scalaVersion: ScalaVersion,
       scalaBinaryVersion: ScalaBinaryVersion,
       classpath: JList[Path]
-  ): List[l.TextEdit] = {
+  ): Future[List[l.TextEdit]] = {
     val fileInput = file.toInput
     val unsavedFile = file.toInputFromBuffers(buffers)
```
ckipp01, line 44: I know you didn't change this, but reading through this, there is a chance that is isn't unsaved, but rather the same thing as `fileInput`? If so, I think naming this `unsavedFile` is a bit misleading.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -35,52 +39,76 @@ case class ScalafixProvider(
...
+        } yield textEditsFrom(patches, fileInput)
+
+        if (!scalafixEvaluation.exists(_.isSuccessful())) {
+          val scalafixError =
+            scalafixEvaluation.flatMap(_.getMessageError().asScala)
+
+          val message = scalafixError.getOrElse {
+            """|Unexpected error while running scalafix.
+               |Semanticdb might have been stale, which 
+               |would require successful compilation to be created""".stripMargin
```
ckipp01, line 95:
```suggestion
               |would require successful compilation to be created.""".stripMargin
```

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -35,52 +39,76 @@ case class ScalafixProvider(
...
+        edits.getOrElse(Nil)
+      }
     }
   }
 
+  private def warningMessage(
+      message: String
+  ): MessageParams = {
+    new MessageParams(MessageType.Error, message)
+  }
```
ckipp01, line 110: Sort of a nit,but we call this method `warningMessage`, but we give the message an `Error` type. Should we align those?

ckipp01 (changes requested):
These are some nice improvements @tgodzik 👍

Just a couple questions.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -35,52 +39,76 @@ case class ScalafixProvider(
...
+            .withConfig(scalafixConf.asJava)
+            .withRules(List(organizeImportRuleName).asJava)
+            .withPaths(List(file.toNIO).asJava)
+            .withSourceroot(workspace.toNIO)
+            .withScalacOptions(Collections.singletonList(scalacOption))
+            .evaluate()
+        }
+        val edits = for {
+          results <- scalafixEvaluation
+          if results.isSuccessful()
```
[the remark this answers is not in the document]
mlachkar: But in the other hand, it's ok since we only use scalafix for organizeimport, which doesn't report lint errors. And I will change this behavior in next version of scalafix to stick to that.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -35,52 +39,76 @@ case class ScalafixProvider(
...
-          .withScalacOptions(Collections.singletonList(scalacOption))
-          .evaluate()
-      }
+      compilations.compilationFinished(file).map { _ =>
+        val scalafixConfPath = workspace.resolve(scalafixConfigPath)
+        val scalafixConf =
+          if (scalafixConfPath.isFile) Some(scalafixConfPath.toNIO)
+          else None
+        val scalafixEvaluation = for {
+          api <- getScalafix(scalaBinaryVersion)
```
mlachkar, line 59: I would change this for comprehension from option to Try to get errors when failing downloading scalafix or the rule. It will also be easier to add error message on other part.
```
  private def getScalafix(
      scalaBinaryVersion: ScalaBinaryVersion
  ): Try[Scalafix] = {
    scalafixCache
      .get(scalaBinaryVersion)
      .toTry
      .orElse(
        statusBar.trackBlockingTask("Downloading scalafix") {
          Try(Scalafix.fetchAndClassloadInstance(scalaBinaryVersion))
            .map { api =>
              scalafixCache.update(scalaBinaryVersion, api)
              api
            }
        }
      )
  }

  private def getRuleClassLoader(
      scalaBinaryVersion: ScalaBinaryVersion,
      scalafixClassLoader: ClassLoader
  ): Try[URLClassLoader] = {
    val pathsOpt = organizeImportRuleCache
      .get(scalaBinaryVersion)
      .toTry
      .orElse(
        statusBar.trackBlockingTask("Downloading organize import rule") {
          Try(Embedded.organizeImportRule(scalaBinaryVersion))
            .map { paths =>
              organizeImportRuleCache.update(scalaBinaryVersion, paths)
              paths
            }
        }
      )
    pathsOpt.map(paths =>
      Embedded.toClassLoader(
        Classpath(paths.map(AbsolutePath(_))),
        scalafixClassLoader
      )
    )
  }
// to add with the rest of extension methods we have on Option.
  implicit class OptionExtension[A](val in: Option[A]) extends AnyVal {
    def toTry(e: => Throwable): Try[A] = in match {
      case Some(v) => Success(v)
      case None => Failure(e)
    }

    def toTry = toTry(new Exception("None.get "))
  
```

This way we will cover this error.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -35,52 +39,76 @@ case class ScalafixProvider(
...
+            .withPaths(List(file.toNIO).asJava)
+            .withSourceroot(workspace.toNIO)
+            .withScalacOptions(Collections.singletonList(scalacOption))
+            .evaluate()
+        }
+        val edits = for {
+          results <- scalafixEvaluation
+          if results.isSuccessful()
+          fileEvaluation <- results.getFileEvaluations().headOption
+          patches <- fileEvaluation.previewPatches().asScala
```
mlachkar, line 85: If we transform that to try for comprehension, we could do the error reporting here directly
fileEvaluation.previewPatches().asScala.toTry(new Exception(s"scalafix failed running on file $file with errors ${fileEvaluation.getErrors.toList}"))

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -35,52 +39,76 @@ case class ScalafixProvider(
...
+        } yield textEditsFrom(patches, fileInput)
+
+        if (!scalafixEvaluation.exists(_.isSuccessful())) {
+          val scalafixError =
+            scalafixEvaluation.flatMap(_.getMessageError().asScala)
```
mlachkar, line 90: You are right, getMessageError, is empty except if the arguments for scalafix are missing or some other parsing error.

mlachkar (commented):
I would change the for comprehension from Option to Try.
I am also fixing the issue you reported earlier on scalafix https://github.com/scalacenter/scalafix/pull/1251

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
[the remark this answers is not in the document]
tgodzik: Let me leave the note there maybe and we can followe up after the next scalafix release? I would try to fix even temporarilly if possible.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
tgodzik: Changed it to tries.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
tgodzik: Thanks, I changed it a bit, but stayed why Try as you suggested.

metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
```diff
@@ -47,6 +47,16 @@ final class Compilations(
   def wasPreviouslyCompiled(buildTarget: b.BuildTargetIdentifier): Boolean =
     lastCompile.contains(buildTarget)
 
+  def compilationFinished(
+      source: AbsolutePath
+  ): Future[Unit] = {
+    if (currentlyCompiling.isEmpty) {
+      Future(())
+    } else {
+      cascadeCompileFiles(Seq(source)).map { _ => () }
```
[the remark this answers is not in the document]
tgodzik: Thanks for noticing! Fixed!

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
tgodzik: Renamed it, you're right.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
tgodzik: Fixed!

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
tgodzik: changed the name and it now reflects the fact it's an error.

metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
```diff
@@ -32,18 +30,20 @@ final class OrganizeImports(
...
+        target <- buildTargets.scalaTarget(buildId)
+      } yield target
+      scalaTarget match {
+        case Some(target)
+            if !ScalaVersions.isScala3Version(target.scalaVersion) =>
+          organizeImportsEdits(file, target)
+        case Some(target)
+            if ScalaVersions.isScala3Version(target.scalaVersion) =>
+          scribe.info(
+            s"Organize import doesn't work on ${target.scalaVersion} files"
```
ckipp01, line 45: Do you think it's clear enough to the user that that this isn't supported in Scala 3 files? For example I triggered this in a Scala 3 file:

Screenshot 2020-10-06 at 08 04 36

And while we are logging this here that the rule isn't supported in this target version, do you think it might make sense to actually show this warning to the user to make it explicit why they are seeing the above message about the code action not being found?

ckipp01 (approved):
Great work on this! Just one final question about possibly showing the message to users in the case they try to use it in Scala 3, but other than that, LGTM!

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -22,65 +25,98 @@ case class ScalafixProvider(
...
-    languageClient: MetalsLanguageClient
-) {
+    languageClient: MetalsLanguageClient,
+    buildTargets: BuildTargets
+)(implicit ec: ExecutionContext) {
   import ScalafixProvider._
   private val scalafixCache = TrieMap.empty[ScalaBinaryVersion, Scalafix]
   private val organizeImportRuleCache =
-    TrieMap.empty[ScalaBinaryVersion, List[Path]]
+    TrieMap.empty[ScalaBinaryVersion, URLClassLoader]
```
tgodzik, line 36: Changed it to Classlaoder, since otherwise we create a new classlaoder on each run, which can really grow the memory.

mlachkar (approved):

metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
tgodzik: I think it's a bit different, since we are not adding the action at all and Metals is not expected to work in this case. Also, if someone has organize imports on save it would continuously show the message, because it will never work.

We can update it if it will be confusing for users.

PR #2121, opened 2020-10-03
Account for possible null value on PopupChoiceReset message request

ckipp01:
I randomly noticed this tonight when using the interactive reset choice functionality. According to [the spec](https://microsoft.github.io/language-server-protocol/specification.html#window_showMessageRequest), the showMessage request can return a `MessageActionItem` or `null`. We weren't accounting for the `null` and I was seeing null pointer exceptions being thrown when I cancelled the request. This just adds in a check.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/PopupChoiceReset.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/PopupChoiceReset.scala
@@ -51,7 +51,9 @@ class PopupChoiceReset(
       .showMessageRequest(choicesParams())
       .asScala
       .flatMap { item =>
-        if (item.getTitle() == PopupChoiceReset.BuildTool) {
+        if (item == null) {
+          Future.successful(())
+        } else if (item.getTitle() == PopupChoiceReset.BuildTool) {
           reset(PopupChoiceReset.BuildTool)
         } else if (item.getTitle() == PopupChoiceReset.BuildImport) {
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/PopupChoiceReset.scala
```diff
@@ -63,6 +65,6 @@ class PopupChoiceReset(
 }
 
 object PopupChoiceReset {
-  final val BuildTool = "build tool selection"
-  final val BuildImport = "build import"
+  final val BuildTool = "Build tool selection"
+  final val BuildImport = "Build import"
```
ckipp01, line 69: This is a super nitpick, but I sort of feel like these should be capitalized.

kpbochenek (approved):
Good catch! LGTM :+1:

kpbochenek:
Huh in many places we do `.showMessageRequest(...)` we don't account for possible `null` ;/

ckipp01:
> Huh in many places we do .showMessageRequest(...) we don't account for possible null ;/

🤔 yea you're right it seems. I think most of the time we never hit on it because it seems we do this a lot:
```scala
    languageClient
      .showMessageRequest(params)
      .asScala
      .map { item =>
        if (item == dontShowAgain) {
          notification.dismissForever()
        }
        Confirmation.fromBoolean(item == yes)
      }
```
Where we never attempt to access anything in the `item`, but simply compare it to the item.

PR #2123, opened 2020-10-04
Add support for loading env variables from a .env file.

mwz:
This PR adds support for loading environment variables from .env files.

This is still a WIP, but I wanted to make sure this approach is ok before I spend more time on implementing more tests and cleaning up the code.

I added a new `envFile` parameter to `DebugUnresolvedMainClassParams` which allows you to specify either a relative (to your project workspace) or an absolute path to an `.env` file in the `launch.json` config, e.g.:
```diff
 "version": "0.2.0",
  "configurations": [
    {
      "type": "scala",
      "name": "Debug ",
      "request": "launch",
      "mainClass": "Main",
+     "envFile": ".env"
    }
  ]
}
```

If the path is present, the file is parsed and the env variables are prepended to the env variables defined in the `env` object introduced in #2118 with the intention of making `env` override any variables from the `.env` file.

I've adapted the .env file parser regex from both [mefellows/sbt-dotenv](https://github.com/mefellows/sbt-dotenv) and [bkeepers/dotenv](https://github.com/bkeepers/dotenv).
It supports single/double-quoted and unquoted values as well as multiline values, both `=` and `:` separators as well as export statements, but it doesn't support value substitution (something that could potentially be added later if anyone needs it.)

Please let me know if you're happy with this approach and if there's anything major to address so far.

### DIFF
--- a/docs/editors/debug-adapter-protocol.md
+++ b/docs/editors/debug-adapter-protocol.md
@@ -59,8 +59,9 @@ identically named classes in different modules. A uri will be returned that can
 be used by the DAP client.
 
 `envFile` is an optional parameter, which allows you to specify a path to a
-`.env` file with additional environment variables. It supports single line as
-well as multi-line quoted values (without value substitution). Any variables
+`.env` file with additional environment variables. The path can be either
+absolute or relative to your project workspace. The parser supports single line
+as well as multi-line quoted values (without value substitution). Any variables
 defined in the `env` object take precedence over those from the `.env` file.
 
 ### Wiring it all together
--- a/docs/editors/debug-adapter-protocol.md
+++ b/docs/editors/debug-adapter-protocol.md
@@ -40,7 +40,8 @@ Apart from using code lenses, users can start a debug session by executing the
   "buildTarget": "foo",
   "args": ["bar"],
   "jvmOptions": ["-Dpropert=123"],
-  "env": { "RETRY": "TRUE" }
+  "env": { "RETRY": "TRUE" },
+  "envFile": ".env"
 }
 ```
 
@@ -56,6 +57,10 @@ Apart from using code lenses, users can start a debug session by executing the
 `buildTarget` is an optional parameter, which might be useful if there are
 identically named classes in different modules. A uri will be returned that can
 be used by the DAP client.
+
+`envFile` is an optional parameter, which allows you to specify a path to a
+`.env` file with additional environment variables. It supports single line as
+well as multi-line quoted values (without value substitution).
 
 ### Wiring it all together
 
--- a/docs/editors/debug-adapter-protocol.md
+++ b/docs/editors/debug-adapter-protocol.md
@@ -63,6 +63,30 @@ be used by the DAP client.
 absolute or relative to your project workspace. The parser supports single line
 as well as multi-line quoted values (without value substitution). Any variables
 defined in the `env` object take precedence over those from the `.env` file.
+Here's an example of a supported .env file:
+
+```bash
+# single line values
+key1=value 1
+key2='value 2'   # ignored inline comment
+key3="value 3"
+
+# multi-line values
+key4='line 1
+line 2'
+key5="line 1
+line 2"
+
+# export statements
+export key6=value 6
+
+# comma delimiter
+key7:value 6
+
+# keys cannot contain dots or dashes
+a.b.key8=value 8   # will be ignored
+a-b-key9=value 9   # will be ignored
+```
 
 ### Wiring it all together
 
--- a/docs/editors/debug-adapter-protocol.md
+++ b/docs/editors/debug-adapter-protocol.md
@@ -60,7 +60,8 @@ be used by the DAP client.
 
 `envFile` is an optional parameter, which allows you to specify a path to a
 `.env` file with additional environment variables. It supports single line as
-well as multi-line quoted values (without value substitution).
+well as multi-line quoted values (without value substitution). Any variables
+defined in the `env` object take precedence over those from the `.env` file.
 
 ### Wiring it all together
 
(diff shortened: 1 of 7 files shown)


tgodzik:
Thanks for the contribution! It looks good! I think that's the right direction to go. It might be also useful to add `envFile` to `DotEnvFileParserSuite.start` test to see if it all works nicely together. We would also need the VS Code extension update, but I can follow up with that one.

mwz:
I've updated tests and docs and I think this is now ready for review. Let me know if I missed anything or there are any other tests you'd like to see.

I'll follow up with a PR in metals-vscode.

tests/unit/src/test/scala/tests/debug/DotEnvFileParserSuite.scala
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  test("allow # in quoted values") {
+    assertDiffEqual(parse("""export KEY="v#l#e""""), Map("KEY" -> "v#l#e"))
+  }
+
+  test("parse multi-line single quoted values") {
+    val content = """KEY='line1
+                    |line2'""".stripMargin
+    assertDiffEqual(parse(content), Map("KEY" -> "line1\nline2"))
```
tgodzik, line 106: It seems the tests with `\n` are failing on windows. It might have something to do with additional `\r` there.

We could try doing `val separator = if (isWindows) "\r\n" else "\n"`, but not 100% if that will help. I try checking it out on windows if needed.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -143,22 +145,34 @@ class DebugProvider(
...
-            clazz.toJson
+        val envFromFile: Future[List[String]] =
+          if (params.envFile != null) {
+            val path = AbsolutePath(params.envFile)(workspace)
+            DotEnvFileParser
+              .parse(path)
+              .map(_.map { case (key, value) => s"$key=$value" }.toList)
+          } else Future.successful(List.empty)
+
+        envFromFile.flatMap { envFromFile =>
```
tgodzik, line 165: Maybe just `map` and we can skip `Future.successful` below?

tgodzik (changes requested):
Thanks for the great contribution! I left only two comments, which should be easily fixable.

tests/unit/src/test/scala/tests/debug/DotEnvFileParserSuite.scala
mwz: I completely forgot about Windows - I think it's a case of substituting all `\r\n` with `\n`. I'll try to fix that later today.

tgodzik: Thanks!

docs/editors/debug-adapter-protocol.md
```diff
@@ -57,6 +58,10 @@ Apart from using code lenses, users can start a debug session by executing the
 identically named classes in different modules. A uri will be returned that can
 be used by the DAP client.
 
+`envFile` is an optional parameter, which allows you to specify a path to a
+`.env` file with additional environment variables. It supports single line as
+well as multi-line quoted values (without value substitution).
```
kpbochenek, line 63: can we add how example file could look like to docs?
```
example.env:

# comment
key1=value1
key2='value2'    # comment too
key3="value3"
# key cannot contain dots or dashes
a.b.key4='value4'   # will be ignored
a-b-key5='value5'   # will be ignored
# multiline value
key6=' line1
       line2'
```

Just looked at tests and tried to extract essential things, if something else is nice, can be suprising might be worth adding here

docs/editors/debug-adapter-protocol.md
```diff
@@ -57,6 +58,10 @@ Apart from using code lenses, users can start a debug session by executing the
...
+
```
kpbochenek, line 64: Add clarification in docs what takes precedense in case of parameter defined in `envFile` and `env` parameter

docs/editors/debug-adapter-protocol.md
```diff
@@ -57,6 +58,10 @@ Apart from using code lenses, users can start a debug session by executing the
 identically named classes in different modules. A uri will be returned that can
 be used by the DAP client.
 
+`envFile` is an optional parameter, which allows you to specify a path to a
```
kpbochenek, line 61: Let's add information that provided path can be relative to root workspace (or absolute path - absolute path will work too?)

kpbochenek:
Thanks for contribution! so happy to see people adding nice things to metals :smile:

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
mwz: addressed in 78ec1b1

docs/editors/debug-adapter-protocol.md
mwz: good idea, I added an example in 1ad5709

docs/editors/debug-adapter-protocol.md
mwz: done in 21158cd
I've also updated one of the tests to reflect the precedence.

docs/editors/debug-adapter-protocol.md
mwz: Both relative and absolute paths are supported. I've mentioned about this in the docs as well as created an additional test to cover this (10dc7e7).

mwz:
@tgodzik @kpbochenek thank you both for your review.
I've addressed all of your comments and hopefully this is good to go now 😄

tgodzik (approved):
LGTM! Thanks for the contribution again!

PR #2124, opened 2020-10-05
Update sbt to 1.4.0

ckipp01:
This also syncs to the lastest sbt-extras

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.3.13
+sbt.version=1.4.0
--- a/sbt
+++ b/sbt
@@ -3,11 +3,39 @@
 # A more capable sbt runner, coincidentally also called sbt.
 # Author: Paul Phillips <paulp@improving.org>
 # https://github.com/paulp/sbt-extras
+#
+# Generated from http://www.opensource.org/licenses/bsd-license.php
+# Copyright (c) 2011, Paul Phillips. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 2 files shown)


ckipp01:
:/ Totally forgot about this: https://github.com/scalameta/munit/issues/210

Looks like we are hitting on the same thing.

tgodzik:
> :/ Totally forgot about this: [scalameta/munit#210](https://github.com/scalameta/munit/issues/210)
> 
> Looks like we are hitting on the same thing.

Yeah, we might need to hold off on the update for a bit unless we are able to fix it inside sbt build definition.

ckipp01:
> Yeah, we might need to hold off on the update for a bit unless we are able to fix it inside sbt build definition.

I'm looking around at the various discussions, but has it been pinned down what is actually causing this? If I remember correctly it was something with dotty docs? Is there an issue logged for it?

ckipp01:
I'm going to close this for now then.

tgodzik:
I think that will be fix in the next version of Scala 3 from what I see. Maybe 0.27.0 and 0.28.0-RC1.

kpbochenek:
We will be able to upgrade only when we completely drop 0.27.0-RC1? 
Do I correctly understand that mtags with dotty <= 0.27.0-RC1 cannot use sbt 1.4.0?

tgodzik:
> We will be able to upgrade only when we completely drop 0.27.0-RC1?
> Do I correctly understand that mtags with dotty <= 0.27.0-RC1 are not compatible with sbt 1.4.0?

I think so unless we fix it in build.sbt we will need to wait.

PR #2126, opened 2020-10-08
Do not show hover if the type is error

tgodzik:


romanowski (approved):

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -98,11 +98,7 @@ case class ScalaPresentationCompiler(
   }
 
   def newDriver: InteractiveDriver = {
-    val implicitSuggestionTimeout =
-      if (SemVer.isCompatibleVersion("0.25.0", BuildInfo.scalaCompilerVersion))
-        List("-Ximport-suggestion-timeout", "0")
-      else Nil
-
+    val implicitSuggestionTimeout = List("-Ximport-suggestion-timeout", "0")
```
tgodzik, line 101: We no longer support anything less than 0.25.0, so the option will always be available.

kpbochenek (approved):
Good Catch!

PR #2129, opened 2020-10-09
Fixed sbt completions coming from project/*.scala

### LINKED ISSUE #2106: Completions from sbt's Scala files only available after opening them
**Describe the bug**

When we load up for example a `build.sbt` file and try to go to definition on a symbol from another file like `project/Deps.scala` it will not work until the file is opened even it was compiled by sbt

**To Reproduce**

1. Create a sample sbt project
2. Add `project/Deps.scala`
```scala
object Deps {
    val scalatest = "3.0.8"
}
```
and to `build.sbt`:

```scala
libraryDependencies += "org.scalatest" %% "scalatest" % Deps.scalatest % Test
```
3. Close `project/Deps.scala`
4. Reload the project
5. Try to go to definition on `Deps.scalatest`

**Expected behavior**

Metals goes to the definition in the other file.

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.9.4

**Additional context**

This might be connected to the presentation compiler only loading the file after it is opened. We could try to make sure we have a proper classpath for sbt files, but that might not work perfectly if we didn't run sbt before.

A better solution will probably be connected to  https://github.com/scalameta/metals/issues/917

**Search terms**

sbt definition meta build

dos65:
I tested it locally but I'm not sure if it's possible to write a test on this thing.

I tried to comment [this line in SbtLspSuite](https://github.com/scalameta/metals/blob/73ac12e4d4b8f4146db0524c63e78ca960fe4bb9/tests/slow/src/test/scala/tests/sbt/SbtLspSuite.scala#L578) but it doesn't work 

Fix #2106

### DIFF
--- a/tests/slow/src/test/scala/tests/sbt/SbtLspSuite.scala
+++ b/tests/slow/src/test/scala/tests/sbt/SbtLspSuite.scala
@@ -588,4 +588,28 @@ class SbtLspSuite extends BaseImportSuite("sbt-import") with ScriptsAssertions {
     } yield ()
 
   }
+
+  test("sbt-meta-symbols") {
+    cleanWorkspace()
+    for {
+      _ <- server.initialize(
+        s"""|/build.sbt
+            |scalaVersion := MetaValues.scalaVersion
+            |/project/MetaValues.scala
+            |import scala.util.Success
+            |object MetaValues {
+            |  val scalaVersion = "$scalaVersion"
+            |}
+         """.stripMargin
+      )
+      _ <- server.didOpen("build.sbt")
+      _ = server.workspaceDefinitions
+      _ = assertNoDiff(
+        server.workspaceDefinitions,
+        """|/build.sbt
+           |scalaVersion/*Keys.scala*/ :=/*Structure.scala*/ MetaValues/*MetaValues.scala:1*/.scalaVersion/*MetaValues.scala:2*/
+           |""".stripMargin
+      )
+    } yield ()
+  }
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
@@ -136,7 +136,8 @@ final class Compilations(
 
   private def expand(path: AbsolutePath): Option[b.BuildTargetIdentifier] = {
     val isCompilable =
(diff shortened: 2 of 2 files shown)


tgodzik:
Thanks for the contribution! Out of curiosity, what is the case it solves?

tgodzik:
> This issue - [#2106 Completions from sbt's Scala files only available after opening them ](https://github.com/scalameta/metals/issues/2106)
> 
> As I got, `didOpen` calls `Compilations.compileFile` that should trigger compilation of buildTarget.
> 
> Without this fix, `expand` doesn't return buildTarget for `build.sbt` and that why nothing works until any of scala sources from meta build will be opened in the editor.

Sorry, haven't noticed that it's in the title :sweat_smile: A bit under the weather today. I will take a look at it a bit closer probably tomorrow or monday. Again, thanks for the contribution!

dos65:
@tgodzik Thx! It works with `server.didOpen("build.sbt")`.

However,  this test produces some error message in the log about `MetaValues.scala.sematicdb` file existence.
I will try to fix it and update PR.

tgodzik:
> @tgodzik Thx! It works with `server.didOpen("build.sbt")`.
> 
> However, this test produces some error message in the log about `MetaValues.scala.sematicdb` file existence.
> I will try to fix it and update PR.

We might just ignore those in sbt build files for now as the semanticdb files are not expected to be there currently.

dos65:
@tgodzik Ok. Added your test

tests/slow/src/test/scala/tests/sbt/SbtLspSuite.scala
```diff
@@ -585,7 +585,32 @@ class SbtLspSuite extends BaseImportSuite("sbt-import") with ScriptsAssertions {
            |}
            |""".stripMargin
       )
+      _ = println("DONE")
```
tgodzik, line 588: Could you just remove the `println`?

tgodzik (changes requested):
Thanks! We should be good to merge if you could remove the additional println?

tests/slow/src/test/scala/tests/sbt/SbtLspSuite.scala
dos65: Ough. Sorry - done

tgodzik (approved):
LGTM!

PR #2133, opened 2020-10-11
Add java sources to standalone worksheets and run compilation before evaluation if necessary.

### LINKED ISSUE #1433: Opening worksheet before opening a referenced file can result in compilation failures
**Describe the bug**
If you open a project and the first file you open is a worksheet, symbols referenced in it that are defined in the workspace won't be picked up, leading to compilation errors.

**To Reproduce**
1. Create a project with a file:

```scala
package demo

//Demo.scala
object Demo {
  val a = 5
}
```

2. Create a worksheet in the same directory:

```scala
//demo.worksheet.sc
import demo.Demo.a

println(a)
```

3. Reload window (or just close+open the project), making sure the worksheet is focused first

4. See compilation errors

**Expected behavior**
The module in which the worksheet is defined is compiled before the worksheet.

**Screenshots**

image

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.8.0

**Additional context**

lsp trace log since the point of reloading:

https://gist.github.com/kubukoz/d9430fac5d0065a22482b9ff5537a22a

**Search terms**
worksheet, compile, compilation order

ckipp01:
EDIT: This pr has changed a bit since I first started due to me totally overthinking what the actual issue was. So the aim for this now is to address two separate things.

1. Add java sources to the StandaloneSymbolSearch when we have a standalone worksheet so goto definition works on Java sources as you'd expect.
2. Close #1433 by checking to see if a worksheet belongs to a build target. If so, check to see if that build target has been compiled at all, and if not, compile it first.

~Now that you're able to put imports directly in your worksheet, I've
started to more often put worksheets either at the root of project, or
even in a directory of their own. I've started to keep worksheets and
just stick them in a `/worksheets` directory so I can refer to them
later on or share them. However, when starting to do this I realized
that while the goto definition worked fine in the buffer, if you had
something like this in your worksheet at the root of your project:~

```scala
import java.time.Instant

val x = Instant.no@@w()
```
~And you attempted to goto the definition of `now()`, you'd just get a
definition not found message. This is due to when we restart the
worksheet presentation compiler if there is no build target, we still
create a `StandaloneSymbolSearch`, but don't get it a workspace fallback
search if there is none. This would happen if you have a worksheet at
the root of your project, or in a standalone directory causing the
definition to not be found. The first change this pr does is to always
offer the fallback to the workspace search, which allows for `now()` to
be found in the scenario where you have a worksheet at the root, and it
correctly brings you to `Instant.java`.~

~However, if your worksheet was in its own directory completely apart
from a build, this still wouldn't work, because the
`StandaloneSymbolSearch` would be empty since we never indexed the
workspace. The second change introduces indexing the workspace when you
first evaluate a worksheet, and then then does again anytime is
determines that the workspace sources have changed. We do this at the
same time that we would restart the worksheet presentation compiler with
the new sources. This ensures that even in a standalone directory of
worksheets, that going to a definition can still be found in the
`destinationProvider` and correctly brings you to `Instant.java`. An
added benefit of this is that it also allows the user to use workspace
symbols while using worksheets.~

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
@@ -429,6 +429,11 @@ class Compilers(
     }
 
     created.getOrElse {
+      val sourcesWithJdk = JdkSources(userConfig().javaHome) match {
+        case Some(absPath) => absPath.toNIO :: sources
+        case None => sources
+      }
+
       jworksheetsCache.put(
         path,
         createStandaloneCompiler(
@@ -436,7 +441,7 @@ class Compilers(
           StandaloneSymbolSearch(
             workspace,
             buffers,
-            sources,
+            sourcesWithJdk,
             classpath,
             isExcludedPackage
           ),
--- a/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
@@ -19,6 +19,7 @@ import scala.meta.internal.metals.Buffers
 import scala.meta.internal.metals.BuildInfo
 import scala.meta.internal.metals.BuildTargets
 import scala.meta.internal.metals.Cancelable
+import scala.meta.internal.metals.Compilations
 import scala.meta.internal.metals.Compilers
 import scala.meta.internal.metals.Diagnostics
 import scala.meta.internal.metals.Embedded
@@ -62,7 +63,8 @@ class WorksheetProvider(
     diagnostics: Diagnostics,
     embedded: Embedded,
     publisher: WorksheetPublisher,
-    compilers: Compilers
+    compilers: Compilers,
+    compilations: Compilations
 )(implicit ec: ExecutionContext)
     extends Cancelable {
 
@@ -108,8 +110,21 @@ class WorksheetProvider(
       path: AbsolutePath,
       token: CancelToken
   ): Future[Unit] = {
-    evaluateAsync(path, token).map(
-      _.foreach(publisher.publish(languageClient, path, _))
+    val possibleBuildTarget = buildTargets.inverseSources(path)
+    val previouslyCompiled = compilations.previouslyCompiled.toSeq
+
+    (possibleBuildTarget match {
+      case Some(bdi)
+          if (previouslyCompiled.isEmpty || !previouslyCompiled.contains(
+            bdi
+          )) =>
+        compilations.compileTarget(bdi).ignoreValue
+      case _ =>
+        Future.successful(())
+    }).flatMap(_ =>
+      evaluateAsync(path, token).map(
+        _.foreach(publisher.publish(languageClient, path, _))
+      )
     )
   }
 
@@ -145,7 +160,9 @@ class WorksheetProvider(
       cancelables.cancel() // Cancel previous worksheet evaluations.
       val timer = new Timer(Time.system)
       result.asScala.foreach { _ =>
-        scribe.info(s"time: evaluated worksheet '${path.filename}' in $timer")
+        scribe.info(
+          s"time: evaluated worksheet '${path.filename}' in $timer"
+        )
       }
       cancelables.add(Cancelable(() => completeEmptyResult()))
       statusBar.trackFuture(
--- a/tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
+++ b/tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
@@ -475,6 +475,35 @@ abstract class BaseWorksheetLspSuite(scalaVersion: String)
     } yield ()
   }
 
+  test("root-outside-definition") {
+    assume(!isWindows, "This test fails unpredictably on Window")
+    for {
+      _ <- server.initialize(
+        s"""
+           |/metals.json
+           |{"a": {"scalaVersion": "$scalaVersion"}}
+           |/Main.worksheet.sc
+           |import java.time.Instant
+           |
+           |val x = Instant.now()
+           |""".stripMargin
+      )
+      _ <- server.didOpen("Main.worksheet.sc")
+      _ = assertNoDiff(
+        server.workspaceDefinitions,
+        getExpected(
+          """|/Main.worksheet.sc
+             |import java.time.Instant/*Instant.java*/
+             |
+             |val x/*L2*/ = Instant/*Instant.java*/.now/*Instant.java*/()
+             |""".stripMargin,
+          Map.empty,
+          scalaVersion
+        )
+      )
+    } yield ()
+  }
+
   test("no-position") {
     for {
       _ <- server.initialize(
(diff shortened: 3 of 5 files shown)


metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -433,13 +433,7 @@ class Compilers(
         path,
         createStandaloneCompiler(
           classpath,
-          StandaloneSymbolSearch(
```
tgodzik, line 436: Isn't the real issue is that we don't index JDK inside standalone symbols search? Not sure if we should allow workspace search, since it can show invalid results for auto imports etc.

tgodzik (commented):
Sorry for delay in reviewing! I was knocked out by cold for a few days.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
tgodzik: I think we would need to add `JdkSources(userConfig.javaHome)`

ckipp01: Ahhhh, yea you're totally correct. I _way_ overthought this, and I just tested it with adding `JdkSources(userConfig.javaHome)`, and it works as expected. I'll make this change, and then leave out the workspace search.

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -264,6 +265,7 @@ class WorksheetProvider(
     if (newDigest != previousDigest) {
       worksheetsDigests.put(path, newDigest)
       val sourceDeps = fetchDependencySources(worksheet.dependencies().asScala)
+      indexWorkspace()
```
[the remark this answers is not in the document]
ckipp01: Good point. I don't actually think the indexing is needed here now if we aren't going to do the workspace symbols, so I'll play around with triggering a compilation here instead to fix that issue.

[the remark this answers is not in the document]
ckipp01: So with playing around with it, indeed, if I trigger a full compile before the evaluation (like 10 lines above this), then opening a worksheet first that includes someone from another Scala file works. However, as you can imagine in a huge codebase, we don't want to do that. Do you think it's fine to instead just to trigger compilation in the build target that holds the worksheet?

ckipp01:
> Sorry for delay in reviewing! I was knocked out by cold for a few days.

No worries! Glad you got some rest, and I hope you're feeling better!

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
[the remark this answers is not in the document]
tgodzik: Should be fine to only compile the build target I think. It will try trigger any other build targets that are needed. That could still take a while, but I don't think there is any other choice here really? :thinking:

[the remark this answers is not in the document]
ckipp01: Alright, so I got this changed around now. I tested it in Metals, and it's actually not that bad. So I think this should work.

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -108,8 +110,20 @@ class WorksheetProvider(
...
+    val previouslyCompiled = compilations.previouslyCompiled.toSeq
+
+    (possibleBuildTarget match {
+      case Some(bdi) if previouslyCompiled.isEmpty =>
+        compilations.compileTarget(bdi).ignoreValue
+      case Some(bdi) if !previouslyCompiled.contains(bdi) =>
+        compilations.compileTarget(bdi).ignoreValue
+      case _ =>
+        Future.successful(())
+    }).flatMap(_ =>
```
ckipp01, line 123: There is still one flaw with that that I'm unsure to get around, _however_ I think it's related to some of the multiple other tickets we have about improving the standalone presentation compiler design. The main problem is that when you directly open a worksheet that is located in a src directory, it may not recognize the build target before triggering the evaluation. So this check here shows the build target as empty, then triggers an evaluation even though it really does belong to a build target. However, then as soon as it connects to the build server, then this ends up being hit again, it does recognize that the worksheet belongs to a build target, then correctly compiles the target, and corrects the worksheet evaluation. So when playing around with this, you _may_ see for a moment that the evaluation is incorrect since it thinks there is no build target, and therefore doesn't recognize your import, but then it quickly corrects.

tgodzik: Yeah, that's exactly the same issue as in other places. Probably we should wait for the indexing promise in this case.

tgodzik: I will work on this before the next release

ckipp01: Yea that's what I figured. For the time being, do you think this change is fine though? I do think it's a significant improvement on the current situation, and when testing it, you don't really hit on this issue at all unless you are open your worksheet as the first document in a large build. 

So most of the time, it should be fine, like so:

tgodzik: I think it's fine for now.

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -108,8 +110,20 @@ class WorksheetProvider(
...
       token: CancelToken
   ): Future[Unit] = {
-    evaluateAsync(path, token).map(
-      _.foreach(publisher.publish(languageClient, path, _))
+    val possibleBuildTarget = buildTargets.inverseSources(path)
+    val previouslyCompiled = compilations.previouslyCompiled.toSeq
+
+    (possibleBuildTarget match {
+      case Some(bdi) if previouslyCompiled.isEmpty =>
+        compilations.compileTarget(bdi).ignoreValue
```
tgodzik, line 118: Could we squash the case into one with `||` ?

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
```diff
@@ -475,6 +475,42 @@ abstract class BaseWorksheetLspSuite(scalaVersion: String)
...
+        server.workspaceDefinitions,
+        getExpected(
+          """|/Main.worksheet.sc
+             |import java.time.Instant/*Instant.java*/
+             |
+             |val x/*L2*/ = Instant/*Instant.java*/.now/*Instant.java*/()
+             |""".stripMargin,
+          Map(
+            V.scala3 ->
+              """|/Main.worksheet.sc
```
tgodzik, line 502: Isn't it exactly the same as the above string?

tgodzik (changes requested):
Thanks for the great work! I just have to minor suggestions.

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
ckipp01: Yup, fixed.

tgodzik (approved):
LGTM!

kubukoz:
Very nice, thank you!

PR #2148, opened 2020-10-15
Bump scalafmt up to 2.7.4

ckipp01:
Figured this might be a good time to do this seeing that we don't have any large prs open at the moment.


kpbochenek:
Now I understand frustration on scalafmt gitter about defaults changing left and right and constant need to heavily reformat code ;(

(but also a huge fan of syntax `xx.map { case yy => }` instead of putting case into separate line)

kpbochenek (changes requested):
Seems to be failing on all setups: `tests.DefinitionSuite`

ckipp01:
> Now I understand frustration on scalafmt gitter about defaults changing left and right and constant need to heavily reformat code ;(

Yea, agreed. I do wish these diffs could be smaller.

> (but also a huge fan of syntax xx.map { case yy => } instead of putting case into separate line)

Agreed again. I do like this change.

> Seems to be failing on all setups: tests.DefinitionSuite

Ach, I always forget to run `save-expect`. Let me fix this.

kpbochenek (approved):
LGTM :+1:

PR #2149, opened 2020-10-15
Bump setup-scala and cache-action actions

ckipp01:
We are getting a bunch of warnings in CI
```
The `add-path` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

I was hoping this might get rid of some of them, but I'm assuming the changes may need to actually happen in the actions themselves. I'll take a peak, but either way, good to upgrade.


tgodzik (approved):
LGTM!

Issue #2152, opened 2020-06-23
Ability to start sbt session before Metals connects using sbt -bsp


eed3si9n:
**Is your feature request related to a problem? Please describe.**

1. Open VS Code in a directory where I would like Metals to connect to sbt server.
2. If there's been previous session there will be `.bsp/sbt.json`.
3. Metals would then try to issue `java -cp sbt-launch.jar xsbt.boot.Boot -bsp`, which is sbt running in client mode.
4. When `sbt -bsp` does not find a running sbt session it will try to spawn sbt (server) in the background, which will take time and timeout.
5. Repeat to step 3 a few times.
6. To avoid this I'd delete `.bsp/sbt.json` but that doesn't feel right either.

**Describe the solution you'd like**

In addition to `sbt -bsp`, I would like a mechanism to suggest to Metals to open up a new terminal window and start a new sbt session by typing in "sbt", and wait for the server to come up. (The code exists here - https://github.com/sbt/sbt/blob/e323f1f713251968b3d4ecda8a7f0ed52593713a/vscode-sbt-scala/client/src/extension.ts#L34-L45)

**Describe alternatives you've considered**

Or maybe some hardcoded solution like a preference that starts sbt when sbt version is above 1.4.0-M1?

This could be convenient for testing etc purpose even if the user doesn't want to use sbt server as the build server.

**Additional context**

**Search terms**:

ckipp01:
I'm going to hijack this a bit, and then make follow-up tickets more than likely in the extension repos for actually opening sbt in another tab/window/etc depending on the client. _However_, I think this ticket is an important part of #2049 and I'm already hitting on it a bit as I'm working through it.

One of the potential "flows" from #2049 is the the following:

1. User clones a new sbt repo
2. User open up the repo in the editor of their choice
3. No `.bsp/sbt.json` exists yet. 
4. Then what???

Again following up on the conversation in #2049, but limiting it _only_ to the scenario above, I'm introducing a new server command (name is flexible), `build-server-start`, which allows for a single param of the build server that metals may know how to start. When this lands, it will just be `sbt` and `bloop`. However, I'm trying to keep thing as built-tool agnostic as possible because before too long we may be having similar conversation about how to treat [mill as a build server](https://github.com/lihaoyi/mill/pull/969). The temporary plan with this new command as we smooth everything out is the following:

1. User clones a new sbt repo
2. User open up the repo in the editor of their choice
3. No `.bsp/sbt.json` exists yet. 
[...]

Issue #2049, opened 2020-09-03
Allow to choose Bloop or BSP in case of build tools supporting both


tgodzik:
With sbt 1.4.0 we will get an automatic generation of `.bsp/sbt.json` file in case of running sbt. This will cause Metals to only pick up sbt in case of no `.bloop` files or only choose Bloop if those exist.

I propose:
1. When detecting both `.bloop` and `.bsp/sbt.json` files we show a message request with a choice and remember it as it is in case of  multiple build tools
2. When detecting sbt, we offer to either import Bloop or sbt, which will do the fist run, which will generate the needed json file - this should only be supported for sbt versions starting with 1.4.0
3. On connection to sbt server we should increase the timeout, since it might take a while for sbt server to be up.

It would be best to do it all before the release of sbt 1.4.0, since we current behavior might not work well there.

Thoughts? Ideas?

tgodzik:
Additional problem is reloading sbt server, but that should be solved with https://github.com/sbt/sbt/issues/5783

Or we could run manually `sbt -client reload` instead of reimporting.

tgodzik:
Another issue will be getting semanticdb to work with sbt. There is a new setting for sbt:

`Global / semanticdbEnabled := true`

we should at least display in the doctor to add it it build.sbt. Alternatively, we could try to work on something similar to Bloop, where we create a new file `.bloop/bloop.settings.json` and if it exists we automatically add the needed options. That however might be harder to push on sbt.

gabro:
From a UX point of view, I'm a bit worried that this will confuse users. Most users consider bloop as a Metals implementation details, so they wouldn't know what to choose between sbt and bloop.

I'm assuming this scenario happens when an existing Metals user upgrades to sbt 1.4.0?

What's the status of BSP in sbt? Is it on-par with bloop? What are the advantages of one over the other?

I think the ideal scenario would be for Metals to pick one and use the other as a fallback, because I don't think most users will have enough information to answer a "bloop vs sbt" question when prompted.

gabro:
Note that this is different than the multiple build tool question, since users of repos with multiple build tools (such as a Mill + Sbt project) are generally aware of this and can choose appropriately.

The specific BSP implementation  is instead a much more obscure detail, that's "IDE-specific"

ckipp01:
> I think the ideal scenario would be for Metals to pick one and use the other as a fallback, because I don't think most users will have enough information to answer a "bloop vs sbt" question when prompted.

[...]

Issue #2150, opened 2020-10-16
Header content exception in logs with sbt bsp


ckipp01:
**Describe the bug**

I'm not 100% sure where the actual issue lies with this, but as I'm working on #2049 and come across things I'll create issues for them. When connecting to sbt bsp server, this is basically filling up my logs:

<details>
<summary>Missing header Content-Length ....</summary>

```
java.lang.IllegalStateException: Missing header Content-Length in input "{"jsonrpc":"2.0","method":"build/taskStart","params":{"taskId":{"id":"2","parents":[]},"eventTime":1602794959623,"message":"Compiling bsp-test","dataKind":"compile-task","data":{"target":{"uri":"file:/Users/ckipp/Documents/scala-workspace/bsp-test/#bsp-test/Compile"}}}}Content-Length: 293
Content-Type: application/vscode-jsonrpc; charset=utf-8

"
        at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:91)
        at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:113)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)

Oct 15, 2020 10:49:19 PM org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer fireError
SEVERE: Missing header Content-Length in input "{"jsonrpc":"2.0","method":"build/publishDiagnostics","params":{"textDocument":{"uri":"file:/Users/ckipp/Documents/scala-workspace/bsp-test/src/main/scala/Main.scala"},"buildTarget":{"uri":"file:/Users/ckipp/Documents/scala-workspace/bsp-test/#bsp-test/Compile"},"diagnostics":[],"reset":true}}Content-Length: 318
Content-Type: application/vscode-jsonrpc; charset=utf-8

"
[...]

PR #2154, opened 2020-10-19
Enable smoother sbt bsp integration.

ckipp01:
## 📢  EDIT: A lot has changed since this was first opened, but I'm leaving this here to show my thought process about how I started this vs what it's like now. So if you want to know the full history of how this started, feel free to read this, if not scroll on down to the other large section with gifs and I'll explain the approach I landed on.

This pr is the first steps in addressing #2049 and #2152 by enabling the sbt bsp support to be much more integrated into Metals if a user would like choose it instead of using Bloop.

This is still heavily a WIP, I still have a bunch of TODOs as you'll see in the code, I need a bunch of testing, and even some docs updated. However, it's getting a bit large, so I'd like to throw this up here as it is working, and I'd love any feedback about the approach or things I may have missed before I get even farther into this.

Following along with the detailed conversations in the two threads listed above and also the POC that @kpbochenek started on (👍 thanks for that), I split the intended support into 4 paths.

#### Current behavior

The current behavior for this is that they would have to ignore the prompt to import the build, start up the sbt shell with `-Dsbt.semanticdb=true`, and then do a `build-connect` command to connect to it. 

#### New behavior

Again, following the discussion, until all the kinks get smoothed out, we have proper testing, and the UX is on par with that of the Bloop integration, we'll keep Bloop as the default choice. That part doesn't change, so upon opening the project, the user will still get a prompt to import the build (with Bloop). If they'd like to instead use sbt server, they'd just ignore the prompt, and execute a new command (that may be named different in the client) `build-server-start` with a single parameter "sbt". The idea in the future would be for this new command to be the way we can trigger to start a server manually for whatever reason. In the case of sbt, this is sort of special-cased as there is a `SbtServer.scala` class that will handle this by ensuring the sbt version is > 1.4.0, starts up a process of sbt, and then establishes a bsp connection.

Here is a gif showing that. The time is accurate to give you an idea of how long it will take. There will "progress" added so the user knows what's going on.

If you were to look at the logs in this situation, you'd see something like this, which will give you a good idea of what's going on behind the scenes:

```
2020.10.19 16:38:15 INFO  Started: Metals version 0.9.5-SNAPSHOT in workspace '/Users/ckipp/Documents/scala-workspace/bsp-test' .
2020.10.19 16:38:15 INFO  time: initialize in 0.31s
2020.10.19 16:38:16 INFO  Attempting quick connect to the build server
2020.10.19 16:38:16 WARN  Build server is not auto-connectable.
2020.10.19 16:38:17 INFO  no build target: using presentation compiler with only scala-library
2020.10.19 16:38:26 INFO  Attempting to connect to sbt BSP server...
2020.10.19 16:38:26 INFO  sbt 1.4.0 found for workspace.
2020.10.19 16:38:26 INFO  Suitable version of sbt found, attempting to connect...
2020.10.19 16:38:26 INFO  Installalling plugin to /Users/ckipp/Documents/scala-workspace/bsp-test/project/MetalsSbtBsp.scala
2020.10.19 16:38:26 INFO  Starting background sbt process...
2020.10.19 16:38:26 INFO  sbt process started: true
2020.10.19 16:38:37 INFO  sbt up and running, attempting to start a bsp session...
2020.10.19 16:38:38 INFO  tracing is enabled: /Users/ckipp/Library/Caches/org.scalameta.metals/bsp.trace.json
2020.10.19 16:38:38 INFO  Connected to Build server v1.4.0
2020.10.19 16:38:38 INFO  Processing
2020.10.19 16:38:38 WARN  There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
2020.10.19 16:38:39 INFO  Processing
2020.10.19 16:38:39 INFO  Processing
2020.10.19 16:38:39 INFO  Processing
2020.10.19 16:38:39 INFO  Processing
2020.10.19 16:38:39 INFO  Processing
2020.10.19 16:38:39 INFO  Processing
2020.10.19 16:38:39 WARN  There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
2020.10.19 16:38:40 INFO  time: imported build in 1.68s
2020.10.19 16:38:39 INFO  Processing
2020.10.19 16:38:42 INFO  time: indexed workspace in 1.72s
2020.10.19 16:38:42 INFO  Completed connection with a success
2020.10.19 16:38:42 INFO  Processing
2020.10.19 16:38:42 INFO  compiling bsp-test
```

#### Current behavior

This would currently be no different than above, the user would have to go through all the steps of connecting again.

#### New behavior
The users choice was saved up above, so that will be detected, sbt will be triggered to start, and then a bsp connection will be made. Some more work probably needs to happen here to ensure we don't have to many shells open that don't need to be.

#### Current behavior

Again, same as the others, they would need to basically remove their `.metals/` and `.bloop/` directories, and then repeat the process from the beginning to connect.

#### New behavior

Now the user is able to simply again trigger the new `start-build-server` command. As an example in the gifs you're seeing it bound to a `start-sbt-server` command, which sends in the necessary command to Metals. I'll open that pr and one in VS Code shortly after this if someone wants to take it for a test drive.

#### New behavior

I'm just going to skip the old behavior because that flow just doesn't exist right now. But there is now a couple ways to do this. Firstly, you could trigger an `reset-choice` to forget the build server choice, close your workspace, re-open, and then you'd be prompted again to import your build with Bloop.

Switching back and forth is also supported, as you can see here where I use `bsp-switch` to jump from the sbt server back to bloop (sbt is shutdown here):

There are also some other notable things:

- Existing commands like `build-disconnect`, `build-restart`, etc now have changed behavior to detect if the user is using Bloop or another bsp server and issue the commands accordingly.
- [`workspace/reload`](https://build-server-protocol.github.io/docs/specification.html#reload-request) is now introduced so when a build change is detected while using sbt bsp, it's fired. This also happens for example if the user triggers a manual `build-import`, but they are using sbt server rather than Bloop. ~Note that we do need to figure out some stuff with this and notifying the user if they reload a bad definition (https://github.com/sbt/sbt/issues/5988)~  (This has been fixed)
- The sbt plugin is currently just a scala file. I'm going to assume we want to actual publish that as a plugin, and then instead include a file including the plugin like we do with the `sbt-bloop` plugin?
- I refactored a lot of the BSP related stuff into a bsp package.

I also want to add while I'm a avid Bloop user and fan, I've been very impressed with the BSP support in sbt. I remember first trying it very early on and was worried because of some issues I hit on, but I didn't notice anything like that while playing around with this. This coming week I'll probably be using this exclusively at work to really test drive it. Looking forward to feedback.


metals/src/main/resources/MetalsSbtBsp.scala
```diff
@@ -0,0 +1,64 @@
+package metals
```
tgodzik, line 1: For sure it would be good to have it as a separate plugin, we did release an sbt plugin before and it should not be too much of an additional effort.

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -0,0 +1,200 @@
...
+        case ResolvedBspOne(details) =>
+          scribe.info(
+            s"Attempting to start a new connection to ${details.getName()} from previous choice..."
+          )
+          bspServers.newServer(workspace, details).map(Some(_))
+        case ResolvedMultiple(_, _) => Future.successful(None)
+      }
+    }
+
+    // TODO-BSP ensure the timing of this is correct, we don't want the metabuilds yet for sbt BSP
```
tgodzik, line 96: That's right, we meta builds are currently not supported, so we should most likely skip them. Especially that maybe later on they will be provided in the main connection, which would mean that we don't need to create any additional configuration.

tgodzik:
Thanks for the great work @ckipp01 ! I will give it a go and see if there is anything I could suggest. Overall, it looks like a very good approach.

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -0,0 +1,200 @@
...
+          // NOTE: we explicity start another sbt process here as simply using newServer
+          // here wil indeed start sbt, but not correctly star the bsp sessions. So before
+          // we start the session we ensure that sbt is running, and then connect.
+          val (_, processHandler) = sbtServer.runSbtShell()
+          processHandler.initialized.future.flatMap { _ =>
+            scribe.info(
+              s"sbt up and running, attempting to start a bsp session..."
+            )
+            bspServers.newServer(workspace, details).map(Some(_))
+          }
```
ckipp01, line 86: Depending on what happens here in https://github.com/sbt/sbt/issues/5993, this may not actually be needed. The issue I was hitting on and reported here https://github.com/scalameta/metals/issues/2150 may have been a red herring not due to file encoding but rather not being able to correctly parse the output when attempting to make a bsp connect. If that's the case, we don't actually need to start a new sbt shell here before making the bsp connect, and instead just allow that process handler to handle the whole thing.

metals/src/main/resources/MetalsSbtBsp.scala
ckipp01: Sounds good. That's what I'll plan on doing. I'll probably due a separate pr including the plugin, and then we can merge it in and dogfood it to this pr so it's available.

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
ckipp01: Alright, cool. Thanks for confirming 👍

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
ckipp01: So this actually works as expected in 1.4.1 without needed the `runSbtShell` call here, which is great, but it will be a problem for anyone trying with 1.4.0. Not sure to leave it in here or remove it.

ckipp01:
So from the original approach up above, I changed a few things that started to worry more the closer I looked at them. Mainly, I didn't want to special-case sbt so much, especially seeing how [the bsp support in mill was merged](https://github.com/lihaoyi/mill/pull/969), I kept trying to think of a nice way to after we smooth out the sbt bsp support to look ahead to the mill bsp support.

Ultimately this lead to the decision to get rid of the original `SbtServer` class that I had, and instead create a `BuildServerProvider` to be mixed in to a `BuildTool` the same way that `BloopPluginBuildTool` (which is now `BloopInstallProvider`) is done. So essentially all the build tools that support Bloop Install are now `<BuildToolName> extends BuildTool with BloopPluginProvider` and if they also provide bsp support, `with BuildServerProvider` also gets mixed in.

Another decision related to the above one was that instead of the `build-server-start` command, I instead landing on a different command called `generate-bsp-config`. This command is really only mean to be used in the case that you don't have a `.bsp/<entry>` file yet. This command will then check if the current build tool is a `BuildServerProvider`, and if so, generate the build server config, and then attempt to connect to it. If there are multiple build tools for the workspace (for example in the future Mill and sbt), then you'd be able to decide which to use as a build server as well. So to outline the 4 paths from above again:

### [Path 1] - User clones down a project, has never ran sbt on it before, opens it, and would like to use the sbt server.

The user will ignore the first import prompt, and instead trigger the `generate-bsp-config`. sbt is the only build tool in the workspace, so it runs `sbt bspConfig`, and after the successful generation of it, connects to the sbt build server.



### [Path 2] - User re-opens the workspace after they have previously chosen to use the sbt bsp server support

Same as above, the decision was saved when the user did the `generate-bsp-config` as sbt, and when you reopen the workspace, it will auto-connect to the sbt server.

### [Path 3] - User has already imported their build with Bloop, but now want to switch to sbt bsp

User has three choice.

1. They can execute the `generate-bsp-config` (but probably not necessary because it will already probably exist), but if it didn't they could execute this, sbt version would be checked, `.bsp/sbt.json` created, the decision saved, and then connected.

2. Since they `.bsp/sbt.json` already exists, the can simply do a `metals.bsp-switch`, which would detect sbt, ensure the plugin is there, and then connect to it while also saving the choice for next time.

3. While a user probably won't do this, the can use the `metals.build-disconnect` command, followed up by a `metals.build-connect` command as well.

### [Path 4] - User wants to switch back from sbt BSP to Bloop.

Same as the first examples, the user can again use the `metals.bsp-switch` command to switch back to Bloop. Or, they can also do a `metals.reset-choice` an select build-server, open the workspace again, and then they'll either be prompted to reimport if necessary or automatically connect to Bloop.

This seems to all be working pretty smoothly at this point, so dare I say the core part of this is now done. I'll focus in on cleaning a few things up ensuring things are properly tested and documented.

ckipp01:
So I think this is in a pretty good place now to have some more eyes on it and to be taken for a test drive. I've been playing around with it for a while, using it this last week some at work, and I'm overall pretty happy with it. I know this is a pretty large change, so from a functionality point of view, I've made an asciicast to show me starting with a fresh workspace, switching bsp servers to sbt, switching back to bloop again, reloading the workspace, playing with the `metals.build-*` commands, etc. Just basically showing you how it functions fully with either chosen build server. I'll also go through the code and leave some comments where I feel they might be needed for better explanations.

[asciicast](https://asciinema.org/a/367672)

metals/src/main/resources/MetalsSbtBsp.scala
ckipp01: My plan for this is to after it's to a point where we like it, send in a separate pr to add this to the build and to the release. Then after it's merged in, just use the snapshot that is produced and dogfeed it to this pr. Does that sound alright?

metals/src/main/scala/scala/meta/internal/builds/BloopInstallProvider.scala
```diff
@@ -0,0 +1,27 @@
+package scala.meta.internal.builds
+import scala.concurrent.Future
+
+import scala.meta.internal.metals.MetalsLanguageClient
+import scala.meta.io.AbsolutePath
+
+/**
+ * Helper trait for build tools that have a Bloop plugin
+ */
+trait BloopInstallProvider { this: BuildTool =>
```
ckipp01, line 10: This used to actually be the `BloopPluginBuildTool.scala`.  I've changed this to a trait mainly to allow it to be a mixing alongside of `BuildServerProvider.scala` for build tools that can also serve as build servers. My hope is that as we move forward to for example support Mill BSP, it will be much simpler to be able to just add the `BuildServerProvider` mixin to the `MillBuildTool`, add in the args for the bsp config generation, and that's literally almost all that would be needed.

metals/src/main/scala/scala/meta/internal/builds/BloopInstallResult.scala
```diff
@@ -1,30 +0,0 @@
-package scala.meta.internal.builds
-
-import scala.meta.internal.builds.Digest.Status
-
-sealed abstract class BloopInstallResult extends Product with Serializable {
```
ckipp01, line 5: This is maybe a bit opinionated, but I wanted to be able to re-use this for the reload result for build tools that support workspace reload from BSP. Since that's part of the spec, I decided to rename this `WorkspaceReloadResult` to better reflect it. As you'll see in the `WorkspaceReloadResult.scala` it's mentioned that it's used for both.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -132,13 +187,35 @@ case class SbtBuildTool(
...
 
+  val name = "sbt"
+
+  // Note: (ckipp01) first version was actually 1.4.0, but there is some issues
+  // with bsp discovery with it. So we instead are going with 1.4.1 being the
+  // first actually supported sbt version for bsp support in Metals.
+  val firstVersionWithBsp = "1.4.1"
+
+  def writeSbtBspPlugin(workspace: AbsolutePath): Unit = {
+    // TODO-BSP right now we just check to see if the file is there, but rather
```
ckipp01, line 205: I'll probably address this after we publish the plugin since I'll have to change this around a bit then.

metals/src/main/scala/scala/meta/internal/metals/ChosenBuildServers.scala
```diff
@@ -6,7 +6,9 @@ import java.sql.Timestamp
 import scala.meta.internal.metals.JdbcEnrichments._
 
 class ChosenBuildServers(conn: () => Connection, time: Time) {
-  def selectedServer(md5: String): Option[String] = {
+  final val explicit = "EXPLICIT"
```
ckipp01, line 9: This `EXPLICIT` stuff is probably just going to be temporary as long as we treat Bloop as the defacto standard. Once that changes, this distinction can just go away.

tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
```diff
@@ -0,0 +1,138 @@
...
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.ServerCommands
+import scala.meta.internal.metals.{BuildInfo => V}
+import scala.meta.io.AbsolutePath
+
+import tests.BaseImportSuite
+import tests.ScriptsAssertions
+
+/**
+ * TODO make a BaseLspSuite powered by sbt
```
ckipp01, line 17: This is probably the largest follow-up that I think I'd want to work on after this. Mainly because while this suite introduces _very_ basic test for the sbt server, our entire BaseLsp suite is focused on Bloop. I'd like to either refactor this or just introduce another base suite that is focused on sbt BSP support. This would need to happen before I'd ever even consider changing from Bloop to sbt as the default for sbt builds.

eatkins:
I was curious about this and I tried publishing a local version of this branch but it didn't seem to work. In a new project, there doesn't seem to be any option for `metals.bsp-switch` in the coc.nvim drop down after loading the project. Bloop was also clearly used for compilation. I did verify that the metals server was running the latest code.

As a side note, in addition to the `metals.sbtScript`, it would be nice to have the option of passing in the `--client` flag so that metals doesn't start a new sbt server if there is already one running. It actually would be nice to make `sbt --client` the default way of launching sbt from metals. Looking at the history, it seemed that the embedded launcher jar was used because of some problems with the launcher in windows back in 2018. `sbt --client` for sure won't work with sbt 1.4.1 and windows, but it should work on the other platforms and it should also work with windows in sbt 1.4.2. If not, we should probably investigate what the problem is on the sbt side. Even if `sbt --client` doesn't work on windows with metals with 1.4.2, I still think it might be a good idea to default to `sbt --client` on mac and linux for sbt versions >= 1.4.1 rather than using the embedded jar.

ckipp01:
> I was curious about this and I tried publishing a local version of this branch but it didn't seem to work. In a new project, there doesn't seem to be any option for `metals.bsp-switch` in the coc.nvim drop down after loading the project. Bloop was also clearly used for compilation. I did verify that the metals server was running the latest code.

Hey! Thanks for trying to take it for a spin. As mentioned up above, there are actually prs open in https://github.com/scalameta/coc-metals/pull/320 and https://github.com/scalameta/metals-vscode/pull/426 to help people explore the support since not all the command have been added to the extensions. Since you're using coc-metals, I actually just went ahead and merged in the `metals.bsp-switch` command since that has been in Metals for quite some time, just not used. The only command in the video you want have access to yet (unless you want to build coc-metals from source) is the new `metals.generate-bsp-config` command. However, since you'll already have the `.bsp/sbt.json` generated, you should be fine with just starting the project, and then using the `metals.bsp-switch` command. So if you're willing, pull in the latest master of coc-metals and you should be good to go.

> As a side note, in addition to the metals.sbtScript, it would be nice to have the option of passing in the --client flag so that metals doesn't start a new sbt server if there is already one running. It actually would be nice to make sbt --client the default way of launching sbt from metals. Looking at the history, it seemed that the embedded launcher jar was used because of some problems with the launcher in windows back in 2018. sbt --client for sure won't work with sbt 1.4.1 and windows, but it should work on the other platforms and it should also work with windows in sbt 1.4.2. If not, we should probably investigate what the problem is on the sbt side. Even if sbt --client doesn't work on windows with metals with 1.4.2, I still think it might be a good idea to default to sbt --client on mac and linux for sbt versions >= 1.4.1 rather than using the embedded jar.

Probably @olafurpg, @gabro, or @tgodzik will know a lot more since the decision was made so long ago, but it's a good point to bring up. I'll have to think about this a bit, and maybe this is a good discussion can start in another thread unless you think it's integral to consider those changes in this pr. I'm a bit hesitant at the moment since I imagine a lot of our users are still on sbt 1.3.x, so switching to default to `--client` should maybe happen, but I'm honestly unsure the best time to do that. What do you think?

tgodzik:
> Probably @olafurpg, @gabro, or @tgodzik will know a lot more since the decision was made so long ago, but it's a good point to bring up. I'll have to think about this a bit, and maybe this is a good discussion can start in another thread unless you think it's integral to consider those changes in this pr. I'm a bit hesitant at the moment since I imagine a lot of our users are still on sbt 1.3.x, so switching to default to `--client` should maybe happen, but I'm honestly unsure the best time to do that. What do you think?

The launcher is used in case that we don't know if the user has sbt installed and I think `-client` is valid also with the launcher (not sure). So this seems to be a separate problem altogether. We can start a new issue and discuss it there. I imagine using `sbtn` might also be something useful especially if we start the server itself before.

eatkins:
> Thanks for trying to take it for a spin @eatkins! As mentioned up above, there are actually prs open in scalameta/coc-metals#320 and scalameta/metals-vscode#426 to help people explore the support since not all the command have been added to the extensions.

Whoops. I guess I glazed over that part.

> The launcher is used if we don't know that the user has sbt installed and I think -client is valid also with the launcher (not sure). So this seems to be a separate problem altogether. We can start a new issue and discuss it there. I imagine using sbtn might also be something useful especially if we start the server itself before.

The launcher does support `-client` but it still spins up a jvm and takes about a second to start up (still better than spawning a new sbt process). The newer launchers scripts delegate `sbt --client` to `sbtn` so `sbt --client bloopInstall` can run very quickly. I manually overrode metals to run `sbt --client` instead of just `sbt` and it was very snappy. The reason to prefer `sbt --client` vs `sbtn` is that it should work even when `sbt` is not on the path, which isn't the case with `sbtn` (of course if `sbtn` is on the path, it's very unlikely that `sbt` isn't). That being said, I do appreciate that there is a bunch of work that might need to be done to transition metals to using `sbt --client` by default in some, if not all, cases.

tgodzik:
@eatkins Ach, you are talking about using sbt script instead of the launcher. I missed that, we could for sure try to change it (at least for linux/macos), but I would still leave it for another issue.

metals/src/main/resources/MetalsSbtBsp.scala
```diff
@@ -0,0 +1,64 @@
...
+      (old.toVector ++ sdbOptions ++
+        (if (ScalaInstance.isDotty(sv)) Some("-Ysemanticdb")
+         else None)).distinct
+    },
+    semanticdbTargetRoot := {
+      val in = semanticdbIncludeInJar.value
+      if (in) classDirectory.value
+      else semanticdbTargetRoot.value
+    },
+    semanticdbOptions ++= {
```
tgodzik, line 47: Do you know what is the purpose of having the separate `semanticdbOptions` ?

tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
tgodzik: I agree this will be needed, but for sure we can skip it in this PR. It's already a lot of great work!

metals/src/main/scala/scala/meta/internal/builds/WorkspaceReloadStatus.scala
```diff
@@ -0,0 +1,33 @@
...
+  def isFailed: Boolean = this.isInstanceOf[Failed]
+  def toChecksumStatus: Option[Status] =
+    Option(this).collect {
+      case WorkspaceReloadStatus.Rejected => Status.Rejected
+      case WorkspaceReloadStatus.Installed => Status.Installed
+      case WorkspaceReloadStatus.Cancelled => Status.Cancelled
+      case Failed(_) => Status.Failed
+    }
+}
+object WorkspaceReloadStatus {
```
tgodzik, line 25: Why not something like `WorkspaceConnectionStatus`. `Reload` seems to suggest that it has only something to do with the `workspace/reload` method, while this is not connected to that? Correct me if I am wrong, the name just feels confusing to me.

tgodzik (commented):
Wow! Thanks for the amazing A+ work you did! I tested it out locally and it seems that the UX is good if not perfect. I left a couple of first comments, I will try to take a longer and closer look tomorrow.

metals/src/main/resources/MetalsSbtBsp.scala
tgodzik: I see it's coming from sbt, but I am honestly just wondering.

metals/src/main/scala/scala/meta/internal/builds/WorkspaceReloadStatus.scala
ckipp01: > `Reload` seems to suggest that it has only something to do with the `workspace/reload` method

Yup, this does actually. Previously this was only used in `BloopInstall` so the wording was really bloop specific. Now we use this for the status of how `BloopInstall` went or the status of a `workspace/reload`. I went with the reload since it was closer to the spec naming. I wouldn't really go towards `WorkspaceConnectionStatus`, since it's not really signifying the connect as much as it is the result of the _install_ or _reload_.

metals/src/main/resources/MetalsSbtBsp.scala
```diff
@@ -0,0 +1,64 @@
...
+import sbt._
+import sbt.Keys._
+import Project.inConfig
+import sbt.internal.inc.ScalaInstance
+
+object MetalsPlugin extends AutoPlugin {
+  override def requires = plugins.JvmPlugin
+  override def trigger = allRequirements
+
+  val semanticdbVersion = "4.3.21"
```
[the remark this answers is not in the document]
ckipp01: Yup true, I'll tie this into that value.

metals/src/main/resources/MetalsSbtBsp.scala
ckipp01: To be honest, not really. Some of this was more or less of a shameless copy pasta from here https://github.com/sbt/sbt/blob/85a0d08a4eb424f88332bcb1f80c1a69344f1e30/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala.

I'll play around with this a bit more, see what is included already, and see what the diff is vs me including them here.

metals/src/main/resources/MetalsSbtBsp.scala
ckipp01: Ach, duh, yea that should actually work fine. I'll work on migrating this over as a plugin in this pr.

metals/src/main/resources/MetalsSbtBsp.scala
tgodzik: I think you shouldn't bother. If it does work, it should be perfectly fine to leave as is. That setting does exist, so most likely we should use it. I was just curious.

metals/src/main/scala/scala/meta/internal/builds/WorkspaceReloadStatus.scala
tgodzik: `WorkspaceLoadStatus` ? But, it's really just a nitpick, we can leave as is.

metals/src/main/resources/MetalsSbtBsp.scala
ckipp01: Playing around with this a bit, should we include both plugins in the `metals.sbt` file? I'm thinking through the best way to do this, and I guess we could do it a few different ways:

1. Two different files (Probably not)
2. One file that includes both the `sbt-metals` and `sbt-bloop` plugin
3. One file that dynamically gets the `sbt-metals` or `sbt-bloop` plugin placed in

I'm leaning a bit towards 2 just for the easiness of it, however, then we also have to worry about users that potentially use the `bloopSbtAlreadyInstalled` user config. However, we could just change the logic in that case to only include `sbt-metals` then.


What do you think?

olafurpg:
> Looking at the history, it seemed that the embedded launcher jar was used because of some problems with the launcher in windows back in 2018

@eatkins The primary reason to use the embedded launcher is to avoid a dependency on the `$PATH` environment variable. For example,  Metals gets different environment variables when VS Code is launched with the `code` binary or by double-clicking on `Visual Studio Code.app` in macOS. The embedded launcher avoids this problem by having a guaranteed way to launch sbt regardless of `$PATH`. This problem originally manifested in our Windows CI, but it's relevant for all operating systems. I would caution against having OS-specific ways of launching sbt.

metals/src/main/resources/MetalsSbtBsp.scala
tgodzik: I think it would be best if we used 3. Adding the sbt plugin will add semanticdb to the build definition, which will not be neccessary if we used Bloop.  Let's only add the new metals plugin is anyone invokes generateBspConfig, especially now as we still do not recommend using sbt.

eatkins:
> @eatkins The primary reason to use the embedded launcher is to avoid a dependency on the $PATH environment variable. 

Yes, I assumed that this was the reason for using the embedded launcher. There are a number of reasons though that I think we should at least consider changing the default. The `sbt --client` experience is noticeably better for the metals use case. Whenever I run the metals.build-import task, it is very tedious to wait for the new sbt server to launch to run the `bloopInstall` task (in the sbt project with coc.nvim, I have to do this fairly often because I had to blow away the .metals and .bloop directories because something got in a bad state). It also can cause two instances of sbt to run concurrently which can be a problem in some cases (like during project load when the two instances may be contending for the ivy lock).

The out of the box behavior can be enhanced if `sbt --client` is used instead of the embedded launcher by default. My assumption, which could be wrong of course, is that the vast majority of metals users have sbt installed on their systems so the PATH issue is unlikely to be a problem. But I agree that it's not completely uncommon that sbt wouldn't be on the PATH. We have had problems in the past with a dependency of sbt's assuming that windows users had `powershell` on their PATH which wasn't always true.

Two ways to workaround the possibility of sbt not being on the PATH, independent of os, are:
1. Try running `sbt` with the assumption that it exists on the path. If not, the process will throw an `IOException` and metals could fall back on the embedded jar.
2. Scan the paths in the PATH environment variable for the sbt script. Fall back to the embedded launcher if it isn't found.

If either of these approaches were taken, a metals setting could be added to force the use of the embedded launcher, much like there is already the sbtScript setting to override the default. For the windows powershell issue above, we took option 2 and fell back to looking in the default system locations if powershell was nowhere on the path https://github.com/dirs-dev/directories-jvm/pull/40. `SbtBuildTool` already has the ability to check the `build.properties` for the version, so I don't see why it couldn't also add some logic to find/cache the sbt launcher script location if the version meet or exceeds 1.4.1.

Apologies for hijacking this PR thread. I do think it's relevant though that in a pure sbt metals bsp project, where I think it's fair to assume the user has sbt installed, the UX of this work could be enhanced by using `sbt --client` when possible. If `sbt --client` (or `sbtn`) isn't the default, I doubt many people will use it. That all being said, I understand the argument for not messing with something that has been working fine for years.

tgodzik:
> Apologies for hijacking this PR thread. I do think it's relevant though that in a pure sbt metals bsp project, where I think it's fair to assume the user has sbt installed, the UX of this work could be enhanced by using sbt --client when possible. If sbt --client (or sbtn) isn't the default, I doubt many people will use it. That all being said, I understand the argument for not messing with something that has been working fine for years.

Actually, when using sbt with BSP we will need to run sbt only once. Each next time we will just reload the workspace on changes. `-client` might indeed be useful when running `bloopInstall` and if someone is running sbt concurrently. That's why I think this is specifically out of scope for this PR. Created a new issue, where we can focus the conversation: https://github.com/scalameta/metals/issues/2160

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -313,7 +328,7 @@ object BuildServerConnection {
         if (isCI)
           initializeResult.get(20, TimeUnit.SECONDS)
         else
-          initializeResult.get(5, TimeUnit.SECONDS)
+          initializeResult.get(205, TimeUnit.SECONDS)
```
tgodzik, line 331: 205 might be too much and should only be needed in case of sbt. Also, if it might take a longer time we could show the progress?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1389,8 +1408,72 @@ class MetalsLanguageServer(
           indexWorkspaceSources()
         }.asJavaObject
       case ServerCommands.RestartBuildServer() =>
-        bloopServers.shutdownServer()
+        bspSession.foreach { session =>
```
tgodzik, line 1411: Could we move the logic to a separate method?

tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
tgodzik: We should create an issue for it though, otherwise this might be easily missed.

tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
```diff
@@ -0,0 +1,138 @@
...
+ * include these here and mark this as a TODO.
+ */
+class SbtServerSuite
+    extends BaseImportSuite("sbt-server")
+    with ScriptsAssertions {
+
+  val preBspVersion = "1.3.13"
+  // TODO after we go past 1.4.1 in Metals, this can just be pulled from
+  // V.sbtVersion
+  val bspVersion = "1.4.1"
```
tgodzik, line 35: I will try to update to sbt 1.4.1. It's pretty annoying that we are blocked by docs.

tgodzik (commented):
Left some more comments after going through it in detail. Again, brilliant work!

metals/src/main/resources/MetalsSbtBsp.scala
ckipp01: Alright, sounds good. I went with 3 and it seems to be working fine 👍

ckipp01: Alright, sounds good. I went with 3 and it seems to be working fine 👍

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
ckipp01: Addressed now that it's an actual plugin.

metals/src/main/scala/scala/meta/internal/builds/WorkspaceReloadStatus.scala
ckipp01: Haha, no it's a fair point. I ended up landing on `WorkspaceLoadedStatus`.

metals/src/main/scala/scala/meta/internal/builds/BuildServerProvider.scala
```diff
@@ -0,0 +1,38 @@
...
+  def generateBspConfig(
+      workspace: AbsolutePath,
+      languageClient: MetalsLanguageClient,
+      systemProcess: List[String] => Future[BspConfigGenerationStatus]
+  ): Future[BspConfigGenerationStatus] =
+    if (workspaceSupportsBsp(workspace)) {
+      systemProcess(createBspFileArgs(workspace))
+    } else {
+      Future.successful(
+        Failed(Right(Messages.NoSbtBspSupport.toString()))
```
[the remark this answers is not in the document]
ckipp01: True, changed.

metals/src/main/scala/scala/meta/internal/builds/BuildTool.scala
```diff
@@ -57,4 +57,13 @@ object BuildTool {
     Files.copy(embeddedFile, outFile, StandardCopyOption.REPLACE_EXISTING)
     outFile
   }
+
+  def copyFromResource(
```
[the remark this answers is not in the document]
ckipp01: Correct. Removed now.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -68,6 +71,58 @@ case class SbtBuildTool(
...
+          JvmOpts.fromWorkspace(workspace),
+          jarArgs,
+          sbtArgs
+        ).flatten
+    }
+    allArgs
+  }
+
+  def workspaceSupportsBsp(workspace: AbsolutePath): Boolean = {
+    import SbtBuildTool.loadVersion
```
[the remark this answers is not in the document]
ckipp01: Ah, yea changed.

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -68,6 +71,58 @@ case class SbtBuildTool(
   override val minimumVersion: String = "0.13.17"
   override val recommendedVersion: String = BuildInfo.sbtVersion
 
+  override def createBspFileArgs(workspace: AbsolutePath): List[String] = {
+    val sbtArgs = List[String](
+      "bspConfig"
+    )
+    val allArgs = userConfig().sbtScript match {
```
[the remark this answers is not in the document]
ckipp01: True, I've refactored this out into a shared method.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
ckipp01: True. I changed around the logic a bit and if we are trying to connect to sbt, I'll give it a minute. We can make that a bit larger if need be, but probably good to start. Then if it's not sbt, I got rid of the CI check and I just default to 30 seconds.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
ckipp01: Yea for sure, I moved this out now.

tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
ckipp01: Awesome, thanks for taking care of this. Rebased on the latest changes and included this from our BuildInfo now.

metals/src/main/scala/scala/meta/internal/bsp/BspConfigGenerator.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Given multiple build tools that are all BuildServerProviders, allow the
+   * choose the desired build server and then connect to it.
+   */
+  def chooseAndGenerate(
+      buildTools: List[BuildTool with BuildServerProvider]
+  ): Future[(BuildTool, BspConfigGenerationStatus)] = {
+    for {
+      Some(buildTool) <- chooseBuildServerProvider(buildTools)
```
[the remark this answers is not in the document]
ckipp01: 🤔 yea true. I'll have to play with this a bit. This will only _ever_ be hit if there are two bsp server available to generate apart from bloop. At the moment, that impossible, since we don't have two BuildTools that extend `BuildServerProvider`, but still good to check.

metals/src/main/scala/scala/meta/internal/metals/ScalaTarget.scala
```diff
@@ -26,7 +26,13 @@ case class ScalaTarget(
 
   def targetroot: AbsolutePath = scalac.targetroot(scalaVersion)
 
-  def baseDirectory: String = info.getBaseDirectory()
+  def baseDirectory: String = {
+    val baseDir = info.getBaseDirectory()
+    // In the BSP spec base directory doesn't have to be defined, so we do a
+    // null check here and just give it a "" if it's null. Mainly this just
+    // impacts the order of the build targets in the Doctor and tree view.
+    if (baseDir != null) baseDir else ""
```
ckipp01, line 34: So this isn't related to sbt, but it seemed relevant so I added it in. I discovered this when I actually wanted to try out a workspace with multiple `.bsp/entries` and I tested the new Mill BSP support, and they had null values for some targets causing this to throw when we tried to sort by it in the Doctor.

tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
ckipp01: Sounds good. I've created an issue here for this https://github.com/scalameta/metals/issues/2163. We can discuss a bit, see what the best way to move forward is, and then dive in.

metals/src/main/scala/scala/meta/internal/metals/ScalaTarget.scala
ckipp01: On second thought, seeing that this isn't related to these changes, I'll just include this in another pr.

https://github.com/scalameta/metals/pull/2164

tgodzik:
I noticed that if you first connected to a BSP server it is no longer possible to generate Bloop config. `Import build` does not seem to run. Any idea what that is about?

ckipp01:
> I noticed that if you first connected to a BSP server it is no longer possible to generate Bloop config. `Import build` does not seem to run. Any idea what that is about?

So part of this is expected. If you ignore the first prompt to import, the `.bloop` directory is never created. Then after you run `generate-bsp-config`, your choice of sbt has been saved. So at that point if you do a `build-import`, you're actually just triggering a workspace reload. So you'd need to reset the build server choice, which then when you do a `build-import` it will trigger you to import the build with Bloop.

However, I do notice that if `.bloop` doesn't exist and you try to do a `bsp-switch`, it won't allow you to switch back to Bloop. I should probably have it so that even in that scenario if Bloop has never been used and the user tries a `bsp-switch`, they still get an option for Bloop even though a `.bloop` doesn't' exist. wdyt?

ckipp01:
This is what I explained above.
- Build import doesn't trigger a `bloopInstall` but just reloads
- `bsp-switch` warns that there is no other build server 
- `reset-choice` resets the build server choice
- Build import now does a `bloopInstall`

tgodzik:
> So part of this is expected. If you ignore the first prompt to import, the .bloop directory is never created. Then after you run generate-bsp-config, your choice of sbt has been saved. So at that point if you do a build-import, you're actually just triggering a workspace reload. So you'd need to reset the build server choice, which then when you do a build-import it will trigger you to import the build with Bloop.

Oh, it makes sense. I was experimenting with a workspace switching the server back and forth - that's probably not something most people do, so it's a pretty niche use case.

> However, I do notice that if `.bloop` doesn't exist and you try to do a `bsp-switch`, it won't allow you to switch back to Bloop. I should probably have it so that even in that scenario if Bloop has never been used and the user tries a `bsp-switch`, they still get an option for Bloop even though a `.bloop` doesn't' exist. wdyt?

Sounds good! I did run `build import` since that seemed analogous to generating config and then switching. I imagine it could also be a bit confusing to some users if it was difficult to switch back.

ckipp01:
> Sounds good! I did run `build import` since that seemed analogous to generating config and then switching. I imagine it could also be a bit confusing to some users if it was difficult to switch back.

Sounds good. I'll special case this then and then a user can no matter what always do a `bsp-switch` and get back to using Bloop.

tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
```diff
@@ -0,0 +1,135 @@
...
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.ServerCommands
+import scala.meta.internal.metals.{BuildInfo => V}
+import scala.meta.io.AbsolutePath
+
+import tests.BaseImportSuite
+import tests.ScriptsAssertions
+
+/**
+ * TODO make a TestingServer powered by sbt
```
tgodzik, line 17: We can remove the TODO or let's link the issue here.

tgodzik (commented):
Just one small comment aside from the one before. We should be almost ready to merge. Great work! :1st_place_medal:

tgodzik (approved):
LGTM! I think this is ready to let users test it out!

ckipp01:
Here we go!

Issue #1502, opened 2020-03-14
Importing java.util.UUID is incorrect


ctataryn:
**Describe the bug**

When using the "Quick Fix" option to import `java.util.UUID`, the actual import that occurs is:

`import java.{util => ju}`

And only the one reference that your cursor was under when you invoked the Quick Fix gets updated to:

`ju.UUID`

**To Reproduce** Steps to reproduce the behavior:

[Video of the problem](https://i.imgur.com/lmCvRa7.gifv)

**Expected behavior**

Either `java.util.UUID` should be imported, or otherwise all instances of `UUID` should be deferenced as `ju.UUID`

**Screenshots**

See above.

**Installation:**

- Operating system: Windows 10 v1909
- VSCode version: 

> Version: 1.43.0 (system setup)
> Commit: 78a4c91400152c0f27ba4d363eb56d2835f9903a
> Date: 2020-03-09T19:47:57.235Z
> Electron: 7.1.11
> Chrome: 78.0.3904.130
> Node.js: 12.8.1
> V8: 7.8.279.23-electron.0
> OS: Windows_NT x64 10.0.18363
> 

- VSCode extension version:

0.8.1

**Additional context**

I think this might have happened after the most recent update?

olafurpg:
Thank you for reporting! I agree the current solution is not ideae but I'm not sure what the expected behavior would be. Either Metals should not use the `util => ju` prefix or Metals should insert the `ju.` prefix for all "symbol not found" errors in the same document with the same name 🤔

ctataryn:
My personal preference is for `import java.util.UUID`. But perhaps giving the user the option to choose either solution from within the Quick Fix menu would be best (at least when talking from the perspective of VSode)

olafurpg:
I think the best solution is to limit the `util => ju` convention for symbols that have conflicting symbols in scope (`Map`, `Set`, `List`, ..). That would mean `UUID` would not trigger the renamed import similar to how most other auto-imports work.

ckipp01:
> I think the best solution is to limit the `util => ju` convention for symbols that have conflicting symbols in scope (`Map`, `Set`, `List`, ..). That would mean `UUID` would not trigger the renamed import similar to how most other auto-imports work.

[...]

PR #2155, opened 2020-10-23
Don't import {java.util => ju} when no conflicting symbols are available

tgodzik:
Fixes https://github.com/scalameta/metals/issues/1502

We only add the default prefixes if there is a conflicting symbol in scope. However, if the user intentionally modified the prefixes we always use the modified ones.

### DIFF
--- a/mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
+++ b/mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
@@ -33,6 +33,11 @@ public interface PresentationCompilerConfig {
     }
 
     /**
+     * Returns true if user didn't modify symbol prefixes
+     */
+    boolean isDefaultSymbolPrefixes();
+
+    /**
      * What text format to use for rendering `override def` labels for completion items.
      */
     OverrideDefFormat overrideDefFormat();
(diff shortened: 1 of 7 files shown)


mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
```diff
@@ -32,6 +32,11 @@
         return map;
     }
 
+    /**
+     * Returns true if user didn't modify symbol prefixes
+     */
+    boolean isDefaultSymbolPrefixes();
```
ckipp01, line 38: More of just a curiosity, but I've always meant to ask about the `symbolPrfixes` in the `PresentationCompilerConfig` interface. I know I don't actually expose this in `coc-metals`, and it looks like we don't in the VS Code extension either. Is this intentional or just an oversight that you originally never used this in the VS Code extension?

ckipp01 (approved):
This looks good! Glad to have this one tackled. LGMT

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
tgodzik: I don't think this feature was used too much at any point, it has been implemented early with the other completion features by @olafurpg , but not sure what was the intention.

PR #2157, opened 2020-10-25
Update sbt to 1.4.1

ckipp01:
This also syncs to the lastest sbt-extras and bumps sbt-dotty.

The issue we hit on in #2124 should be fixed with the sbt-dotty update to 0.4.4.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.3.13
+sbt.version=1.4.1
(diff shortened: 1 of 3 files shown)


ckipp01:
🤔 ok, maybe not... I assumed that this https://github.com/lampepfl/dotty/pull/9756, which is included in sbt-dotty 0.4.3 fixed it.

Issue #2156, opened 2020-10-24
DownloadDependencies crashes with ScalafmtDynamicError$ConfigMissingVersion


sideeffffect:
**Describe the bug**

Running `scala.meta.metals.DownloadDependencies` crashes with `org.scalafmt.dynamic.ScalafmtDynamicError$ConfigMissingVersion: Missing version`
```
#8 44.04 Exception in thread "main" org.scalafmt.dynamic.ScalafmtDynamicError$ConfigMissingVersion: Missing version
#8 44.04 	at org.scalafmt.dynamic.ScalafmtDynamic.$anonfun$readVersion$2(ScalafmtDynamic.scala:226)
#8 44.04 	at scala.util.Either$LeftProjection.flatMap(Either.scala:561)
#8 44.04 	at org.scalafmt.dynamic.ScalafmtDynamic.readVersion(ScalafmtDynamic.scala:222)
#8 44.04 	at org.scalafmt.dynamic.ScalafmtDynamic.resolveConfigWithScalafmt(ScalafmtDynamic.scala:133)
#8 44.05 	at org.scalafmt.dynamic.ScalafmtDynamic.$anonfun$resolveConfig$4(ScalafmtDynamic.scala:119)
#8 44.05 	at scala.util.Try$.apply(Try.scala:213)
...
```

Visible on the PR https://github.com/che-dockerfiles/che-sidecar-scala/pull/9 check here:
https://github.com/che-dockerfiles/che-sidecar-scala/pull/9/checks?check_run_id=1302331720#step:4:696

**To Reproduce**

```bash
METALS_VERSION="0.9.4"
coursier launch -r sonatype:snapshots org.scalameta:metals_2.12:$METALS_VERSION --main scala.meta.metals.DownloadDependencies
```

The line in Dockerfile
https://github.com/che-dockerfiles/che-sidecar-scala/blob/60bba4d43f85588c3bb8e2248a792d70ce994cfb/Dockerfile#L40

**Expected behavior**

Shouldn't crash, a default (latest?) Scalafmt version should be downloaded.

**Installation:**

- Operating system: GitHub Actions, `ubuntu-latest`
- Metals version: v0.9.4

ckipp01:
Thanks for the report @sideeffffect, I'm able to reproduce this. We'll have to take a closer look to see what's going on. At initial glance, this shouldn't happen since it should default to the scalafmt version we have defined in our build utilizing `BuildInfo`, so something's obviously going wrong.

[...]

PR #2158, opened 2020-10-27
Make sure that default version is picked up for Scalafmt provider

tgodzik:
Fixes https://github.com/scalameta/metals/issues/2156


ckipp01 (approved):
Simple fix. I like it. LGTM!

PR #2159, opened 2020-10-28
Mute InvalidProtocolException which might happen normally

tgodzik:
It seems some users were experiencing a lot of such messages as this situation might happen normally. File created or modified events will be sent also in cases that a file wasn't finished writing. In that case we should not show the error.


ckipp01 (approved):
Looks good. I've actually been seeing this a few time lately, whereas I haven't noticed it before.

tgodzik:
> Looks good. I've actually been seeing this a few time lately, whereas I haven't noticed it before.

I think the reason is the change to swoval, which seems to be reporting changes more often.

PR #2161, opened 2020-10-28
Update sbt to 1.4.1

tgodzik:


build.sbt
```diff
@@ -292,6 +292,8 @@ val mtagsSettings = List(
     baseDirectory.in(ThisBuild).value / "mtags",
     scalaVersion.value
   ),
+  // @note needed to deal with issues in https://github.com/scalameta/metals/pull/2157
+  sources in (Compile, doc) := Seq.empty,
```
tgodzik, line 296: Since Metals is not a library, I imagine this will not be needed.

tgodzik:
Follow up from https://github.com/scalameta/metals/pull/2157

build.sbt
ckipp01: Ahh, good point. I should have thought of this before. 👍

ckipp01 (approved):
LGTM!

PR #2164, opened 2020-10-29
Account for possible null value in ScalaTarget's baseDirectory

ckipp01:
I discovered this while playing around with the Mill BSP support since some of the build targets don't always have the `baseDirectory` set as it can be absent [according to the spec](https://build-server-protocol.github.io/docs/specification.html#build-target). This was throwing a null pointer exception when we then tried to order by the base dir in the Doctor. In the case of `null` I just set it to an empty `""` which only affects the ordering of the target in the Doctor or tree view.


tgodzik (approved):
LGTM

PR #2165, opened 2020-10-29
Change scalafix and scalafmt conf to absolute path 

tgodzik:
This changes `metals.scalafixConfigPath` and `metals.scalafmtConfigPath` to use absolute path instead of a relative one. I found the previous behaviour to be problematic because:
- to set global config user would need to set each workspace to a different relative path and it's much more often that people have one global config than that they have a different naming scheme in each workspace they work in
- for scalafmt we would search the defined relative path in each workspace folder, while we currently do not support multiple workspace folders, so it might be quite random which config we choose.

It's important to note it's still possible to specify the path as for example `project/.scalafix.conf`, which will work the same way as previously. @olafurpg not sure why it was a relative path in the beginning, but I feel like it's no longer needed this way. Let me know if this approach might cause some issues.

I also fixed the issue with scalafix config not being updated

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
@@ -218,10 +218,7 @@ final class FormattingProvider(
 
   private def scalafmtConf: AbsolutePath = {
     val configpath = userConfig().scalafmtConfigPath
-    (workspace :: workspaceFolders).iterator
-      .map(_.resolve(configpath))
-      .collectFirst { case path if path.isFile => path }
-      .getOrElse(workspace.resolve(configpath))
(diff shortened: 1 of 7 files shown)


gabro (approved):
I was the one originally implementing it this way, but honestly don't remember the reason. So LGTM 👍

ckipp01 (approved):
Looks good to me! LGTM!

olafurpg:
LGTM 👍 

The motivation for the original behavior is that users are not encouraged to maintain computer-global configuration files in places like `~/.scalafmt.conf`. The problem with computer-global configuration files is that they are not portable to other contributors who may also work on the codebase. 

However, there are valid use-cases to use computer-global configuration files for single-person projects. Also, the sbt plugin and IntelliJ integration don't require a relative path.

mlachkar (approved):

PR #2168, opened 2020-10-30
Preload scalafix to optimize first organize imports run

tgodzik:
This is done similarily to FormattingProvider.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
@@ -1,6 +1,8 @@
 package scala.meta.internal.metals
 
 import java.net.URLClassLoader
+import java.nio.file.Files
+import java.nio.file.Path
 import java.util.Collections
 
 import scala.collection.concurrent.TrieMap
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -109,6 +116,44 @@ case class ScalafixProvider(
...
   }
 
+  private def scalafixConf: Option[Path] = {
+    val scalafixConfPath = userConfig().scalafixConfigPath
+      .getOrElse(workspace.resolve(".scalafix.conf"))
+    if (scalafixConfPath.isFile) Some(scalafixConfPath.toNIO)
+    else None
+  }
+
+  private def evaluateScalafix(
```
ckipp01, line 126: Super nitpicky, but to me when I read this I think we are evaluating Scalafix. Can we instead do:
```suggestion
  private def scalafixEvaluate(
```

ckipp01 (approved):
Just one nit that you can feel free to ignore. LGTM!

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
tgodzik: Fixed!

mlachkar (approved):

Issue #2094, opened 2020-09-24
Valid Dotty code leads to syntax error coming from the scalameta parser when first opening a file


smarter:
I don't know what the scalameta parser is for, but until it supports Dotty it would be nice to have an option to turn it off.

**To Reproduce**

1. Open any file using Dotty indentation syntax
2. Observe syntax errors
3. The errors go away when you start typing (assuming the build import succeeded)

**Expected behavior**

No spurious syntax error.

**Screenshots**

![](https://gitlab.epfl.ch/lamp/cs210/-/raw/master/labs/images/syntax-error-bug.png)

**Installation:**

- Operating system: all of them
- Editor: Visual Studio Code
- Metals version: v0.9.4

**Additional context**

This is the second most confusing things about using metals for students (the most confusing thing is https://github.com/scalameta/metals/issues/1683)

tgodzik:
Thanks for reporting! That is indeed an issue, it's connected to https://github.com/scalameta/metals/issues/1653 and I planned on working on it at the same time. The problem is that before connecting to Bloop we are loading up the presentation compiler for Scala 2, which detects issues in the code as it is not Scala 2 syntax.

PR #2169, opened 2020-10-30
Make sure build server is connected or not available before parsing

tgodzik:
Fixes https://github.com/scalameta/metals/issues/2094

### DIFF
--- a/tests/unit/src/test/scala/tests/ReferenceLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/ReferenceLspSuite.scala
@@ -38,11 +38,11 @@ class ReferenceLspSuite extends BaseRangesSuite("reference") {
           |  val z: Int = y.a
           |  val param: String = a.A.param(arg = 2)
           |}
-          |""".stripMargin,
-        preInitialized = { () => server.didOpen("a/src/main/scala/a/A.scala") }
+          |""".stripMargin
       )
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):
This is a much smaller change than I thought it was going to take for this, but yea, this make total sense. LGTM!

ckipp01 (changes requested):
Actually the referenceLsp suite seems to not be happy.

tests/unit/src/test/scala/tests/ReferenceLspSuite.scala
```diff
@@ -38,11 +38,11 @@ class ReferenceLspSuite extends BaseRangesSuite("reference") {
           |  val z: Int = y.a
           |  val param: String = a.A.param(arg = 2)
           |}
-          |""".stripMargin,
-        preInitialized = { () => server.didOpen("a/src/main/scala/a/A.scala") }
```
tgodzik, line 42: So this would hang, because it's run before sending `initialized` message.

ckipp01: Ah gotcha, that makes sense.

ckipp01 (approved):
Looks good! LGTM!

tgodzik:
JFR is failing again, which makes no sense :/ I will take a look at it later on.

PR #2170, opened 2020-10-30
Fix resolver for sbt-metals snapshot

ckipp01:
Ach.... this was wrong, I had the one from bloop in instead of this one for our snapshots.


tgodzik (approved):
LGTM

PR #2172, opened 2020-10-31
Allow for reset in html doctor

### LINKED ISSUE #2171: No build server reset button in html view of doctor
**Describe the bug**

Not really a bug, but there is a missing reset button in the html view for the Doctor to reset the server choice. This still works if you do an interactive reset, but it would also be nice for VS Code users if the button was there to be consistent with the build tool choice reset button.

**To Reproduce**

1. Open a project and use the sbt BSP server
2. Open doctor
3. See there is not reset choice button

**Expected behavior**

There should be a reset choice button

**Screenshots**

Screenshot 2020-10-30 at 21 19 24

**Installation:**
- Editor: Visual Studio Code or any other editor utilizing the html doctor
- Metals version: Latest snapshot

**Search terms**
html doctor, reset build server choice

ckipp01:
This pr adds in an `(Reset)` option in the html view of the doctor to reset your build server choice. The button is only shown if you've _explicitly_ chosen a build server. So for a user that just connects and uses the default, they will see that they are using Bloop as their build server, but won't be able to reset it, because there is no choice made.

So after making a choice if they do a `bsp-switch`, they the flow would look like this if they wanted to use Doctor to switch back and forth.

Closes #2171

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
@@ -215,7 +215,7 @@ object SbtBuildTool {
    */
   private def metalsPluginDetails(version: String) =
     (
-      "This file enables semsantic information to be produced by sbt.",
+      "This file enables semantic information to be produced by sbt.",
       s""""org.scalameta" % "sbt-metals" % "$version""""
     )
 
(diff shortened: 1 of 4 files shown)


tgodzik (approved):
LGTM!

gabro:
How does this work in tools that don't use html, such as vim? There's another dedicated command and this is just a shortcut?

ckipp01:
> How does this work in tools that don't use html, such as vim? There's another dedicated command and this is just a shortcut?

Yup, exactly. So what the html view is doing is just adding the parm to `reset-choice` to be sent in. If there is no param, then you enter the "interactive" reset. So in Vim it would look like this:



This worked before, and it also works in VS Code if you just trigger it from the command palette.

PR #2182, opened 2020-11-01
Track and show progress to user about connecting to sbt

ckipp01:
Depending on the size of the project, when the decision to use sbt server is made, it may take a bit of time to start sbt before the bsp connection can be made. This shows the process in the status bar so the user has an idea of what's going one instead of not having a clue what's happening for that time.


tgodzik (approved):
LGTM!

Issue #2093, opened 2020-09-24
Build import is not reliable


smarter:
After onboarding ~270 students with metals my main observation is that the initial build import for a project is just not reliable, even on very simple projects: in many situations it will fail, usually with a timeout connecting to the BSP server. Once that happens, the user has to try various manual steps to get things working, and it's very hard to get students to do this correctly (even with step-by-step instructions: https://gitlab.epfl.ch/lamp/cs210/-/blob/master/labs/example-lab.md#troubleshooting). I don't know what needs to be done to fix this exactly (and things might be completely different when using sbt as a BSP server instead of bloop), but I think metals should try harder to get things working by whatever means it can: retrying the build import, restarting the BSP server, etc.

**Installation:**

- Operating system: All of them, most people seem to have issues on Windows but my sample might be biased.
- Editor: Visual Studio Code
- Metals version: v0.9.4

Attached are two metals.log files from two students as samples.
[metals.log](https://github.com/scalameta/metals/files/5276811/metals.log)
[metals-1.log](https://github.com/scalameta/metals/files/5276813/metals-1.log)

tgodzik:
Thanks for reporting! I do agree this step should work almost 100%, though we haven't had issues like that reported too often. 

One thing we could do is increase the timeout for connecting to Bloop and maybe do something like 5 retries connecting to bsp server similar to how VS Code does it.

@smarter are the cases here that eventually users were able to connect or did they never manage to do it?

`metals-1.log` does show things going quite awry, but `metals.log` seem only to indicate that Metals was shutdown while connecting to Bloop

smarter:
> @smarter are the cases here that eventually users were able to connect or did they never manage to do it?

Not sure as I don't get feedback from everyone, I think most people managed to get it working in the end.

tgodzik:
> > @smarter are the cases here that eventually users were able to connect or did they never manage to do it?
> 
> Not sure as I don't get feedback from everyone, I think most people managed to get it working in the end.

Ok, so it should be just the case of making it all more reliable.

ColOfAbRiX:
Adding my personal experience here as I think it might be related: since version 0.9.2 every time I open Metals on VSCode I receive he message "failed to connect to build server" and I'm forced to manually import the build.

tgodzik:
[...]

PR #2184, opened 2020-11-02
Add retry in case of timeout in build/initialize

tgodzik:
Connected to https://github.com/scalameta/metals/issues/2093

I am not a 100% sure about this solution, but this should help with flaky connection issues. It's also the way it's done with the language server in VS Code. Opinions? This should not change too much, it's more "just in case"

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
@@ -246,7 +246,8 @@ object BuildServerConnection {
       connect: () => Future[SocketConnection],
       reconnectNotification: DismissedNotifications#Notification,
       config: MetalsServerConfig,
-      serverName: String
+      serverName: String,
+      retry: Int = 5
   )(implicit
       ec: ExecutionContextExecutorService
   ): Future[BuildServerConnection] = {
@@ -264,10 +265,18 @@ object BuildServerConnection {
           .create()
         val listening = launcher.startListening()
         val server = launcher.getRemoteProxy
-        val result =
-          BuildServerConnection.initialize(workspace, server, serverName)
         val stopListening =
           Cancelable(() => listening.cancel(false))
+        val result =
+          try {
+            BuildServerConnection.initialize(workspace, server, serverName)
+          } catch {
+            case e: TimeoutException =>
+              conn.cancelables.foreach(_.cancel())
+              stopListening.cancel()
+              scribe.error("Timeout waiting for 'build/initialize' response")
+              throw e
+          }
         LauncherConnection(
           conn,
           server,
@@ -279,16 +288,34 @@ object BuildServerConnection {
       }
     }
 
-    setupServer().map { connection =>
-      new BuildServerConnection(
-        setupServer,
-        connection,
-        languageClient,
-        reconnectNotification,
-        config,
-        workspace
-      )
-    }
+    setupServer()
+      .map { connection =>
+        new BuildServerConnection(
+          setupServer,
+          connection,
+          languageClient,
+          reconnectNotification,
+          config,
+          workspace
+        )
+      }
+      .recoverWith { case e: TimeoutException =>
+        if (retry > 0) {
+          scribe.warn(s"Retrying connection to the build server $serverName")
+          fromSockets(
+            workspace,
+            localClient,
+            languageClient,
+            connect,
+            reconnectNotification,
+            config,
+            serverName,
+            retry - 1
+          )
+        } else {
+          Future.failed(e)
+        }
+      }
   }
 
   final case class BspExtraBuildParams(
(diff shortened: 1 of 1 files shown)


ckipp01:
> Attempt to fix #2093
> 
> I am not a 100% sure about this solution, but this should help with flaky connection issues. It's also the way it's done with the language server in VS Code. Opinions? This should not change too much, it's more "just in case"

To be honest, I wish we just knew more about what was going on. Looking back at #2093 and looking at the logs, it's still difficult to tell. One of the examples show the user closing Metals and starting it again 3-4 times all with the same `Waiting for the bsp connection to come up...`. I'm just skeptical that retrying the connection 5 times is going to make any difference in this scenario since that person seemed to close and open 5 times to no avail.

I personally have only seen that happen to me a couple times and have never been able to pin down the reason about what's actually happening, but it happens to infrequently to me that I also don't pay that close attention to it. I'm not against adding in a retry, I'm just skeptical that it will help.

tgodzik:
> To be honest, I wish we just knew more about what was going on. Looking back at #2093 and looking at the logs, it's still difficult to tell. One of the examples show the user closing Metals and starting it again 3-4 times all with the same `Waiting for the bsp connection to come up...`. I'm just skeptical that retrying the connection 5 times is going to make any difference in this scenario since that person seemed to close and open 5 times to no avail.
> 
> I personally have only seen that happen to me a couple times and have never been able to pin down the reason about what's actually happening, but it happens to infrequently to me that I also don't pay that close attention to it. I'm not against adding in a retry, I'm just skeptical that it will help.

In #2093 in one of the logs I think the first connection should have succeed, but maybe it was a slower machine and 5 seconds was not enough for `build/initialize`, which with the increased timeout should not be an issue. Most likely retry will not help in that scenario, however I feel like it's only sensible to retry if the server did not connect. I did have situations, where Metals didn't connect and a simple restart helped. 

This most likely doesn't solve all possible issue, but I think it's a good step. We still can leave the issue open and try to gather more information.

ckipp01:
> This most likely doesn't solve all possible issue, but I think it's a good step. We still can leave the issue open and try to gather more information.

For sure. That makes sense to me.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -327,17 +356,12 @@ object BuildServerConnection {
...
-          initializeResult.get(20, TimeUnit.SECONDS)
-        }
-      } catch {
-        case e: TimeoutException =>
-          scribe.error("Timeout waiting for 'build/initialize' response")
-          throw e
+      if (serverName == SbtBuildTool.name) {
+        initializeResult.get(60, TimeUnit.SECONDS)
+      } else {
+        initializeResult.get(20, TimeUnit.SECONDS)
```
ckipp01, line 362: This was 5 in the past as well before the sbt pr, so hopefully this being a bit longer will help as well.

ckipp01 (approved):
Hopefully this will help some 👍

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -279,16 +289,35 @@ object BuildServerConnection {
...
+        new BuildServerConnection(
+          setupServer,
+          connection,
+          languageClient,
+          reconnectNotification,
+          config,
+          workspace
+        )
+      }
+      .recoverWith { case NonFatal(e) =>
```
gabro, line 303: Minor, but while in `setupServer` we're explicitly hanlding `TimeoutException` (for then rethrowing it) here we're catching all non fatals. Should we limit ourselves to `TimeoutException` here too? Do you expect other errors happening here?

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -279,16 +289,35 @@ object BuildServerConnection {
...
+        scribe.warn("Could not connect to build server")
```
gabro, line 304: if we catch all `NonFatal` I would probably log this 

```suggestion
        scribe.warn("Could not connect to build server", e)
```

(I think scribe allows for passing `e` as a second parameter?)

gabro:
I should also mention that I'm 👍 for this. It may not solve the root cause of timeouts, but I don't see anything bad in re-trying a connection a few times since it's a relatively quick and safe operation to perform and it may address some hard-to-debug transient errors people may see in the wild.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
tgodzik: Right, `TimeoutException` should be actually the only thing that gets thrown here. Fixed!

gabro (approved):

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
tgodzik: Actually I removed the message, this message will be reported in `setupServer()`, so this would just be duplicate.

PR #2185, opened 2020-11-02
Remove condition for jfr, since it's now reliably available on CI

tgodzik:
So this error was actually introduced by me in the previous PR, since I readded `Testing.isEnabled` flag, which hasn't been used for a while.


ckipp01 (approved):
No complaints here 😆. Good catch. LGTM!

PR #2186, opened 2020-11-02
Add in timer methods to a new timerProvider.

ckipp01:
This doesn't add in any new functionality, but we have  a few timer
related methods floating around in `MetalsLanguageServer`, and I've
started to really notice how much stuff we have stuck in there. I'd like
to slowly start moving some stuff out, and these were easy to move out.
Plus some of other methods that use these I plan on moving out as well,
so it will be helpful to have these in their own class to make that
easier instead of passing them from `MetalsLanguageServer` into
other classes.

I ended up deleting the `Timed` class as well since it wasn't being used
at all.

Also this relates to some overdue work in listed in here https://github.com/scalameta/metals/issues/672


tgodzik (approved):
LGTM

PR #2191, opened 2020-11-04
Ensure scalacOptions for Test are correct for sbt BSP.

### LINKED ISSUE #2189: SemanticDB options on an sbt dotty project are wrong
**Describe the bug**

The sbt-metals plugin generates wrong SemanticDB options on the `Test` configuration of a Dotty project.

**To Reproduce**

1. Create an sbt dotty project
```scala
// project/plugin.sbt
addSbtPlugin("ch.epfl.lamp" % "sbt-dotty" % "0.4.4")
addSbtPlugin("org.scalameta" % "sbt-metals" % "0.9.5-SNAPSHOT")
```

```scala
// build.sbt
scalaVersion := "0.27.0-RC1"
```
2. Run `show Test / scalacOptions` in an sbt shell

```shell
sbt:example> show Test / scalacOptions
[info] * -Ysemanticdb
[info] * -semanticdb-target
[info] * /example/target/scala-0.27/meta
[info] * /example/target/scala-0.27/test-meta
```

3. Run `Test / compile` in an sbt shell

```shell
sbt:example> Test / compile
[info] compiling 4 Scala sources to /example/target/scala-0.27/test-classes ...
[error] expected file, received directory '/example/target/scala-0.27/test-meta'
[error] one error found
[error] (Test / compileIncremental) Compilation failed
[error] Total time: 0 s, completed Nov 3, 2020, 9:19:14 PM
```

**Expected behavior**

The `scalacOptions` of the Test configuration should be:

```shell
sbt:example> show Test / scalacOptions
[info] * -Ysemanticdb
[info] * -semanticdb-target
[info] * /example/target/scala-0.27/test-meta
```

**Installation:**

- Operating system: Linux
- Metals version: v0.9.5-SNAPSHOT

**Search terms**

SemanticDB, compiler options, dotty, scala 3, sbt-metals

ckipp01:
So in the same project, this is the diff I got from before these changes and after, which match the desired outcome of #2189 

```diff
sbt:dotty-test> show Test / scalacOptions
[info] * -bootclasspath
[info] * /Users/ckipp/.sdkman/candidates/java/8.0.242-librca/jre/lib/resources.jar:/Users/ckipp/.sdkman/candidates/java/8.0.242-librca/jre/lib/rt.jar:/Users/ckipp/.sdkman/candidates/java/8.0.242-librca/jre/lib/sunrsasign.jar:/Users/ckipp/.sdkman/candidates/java/8.0.242-librca/jre/lib/jsse.jar:/Users/ckipp/.sdkman/candidates/java/8.0.242-librca/jre/lib/jce.jar:/Users/ckipp/.sdkman/candidates/java/8.0.242-librca/jre/lib/charsets.jar:/Users/ckipp/.sdkman/candidates/java/8.0.242-librca/jre/lib/jfr.jar:/Users/ckipp/.sdkman/candidates/java/8.0.242-librca/jre/classes
[info] * -Ysemanticdb
[info] * -semanticdb-target
- [info] * /Users/ckipp/Documents/scala-workspace/dotty-test/target/scala-0.27/meta
[info] * /Users/ckipp/Documents/scala-workspace/dotty-test/target/scala-0.27/test-meta
```

Fixes #2189

### DIFF
--- a/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
+++ b/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
@@ -1,13 +1,14 @@
 package metals
 
-import sbt._
-import sbt.Keys._
-import Project.inConfig
-import sbt.internal.inc.ScalaInstance
 import scala.meta.internal.sbtmetals.BuildInfo
 
+import sbt._
+import sbt.Keys._
+import sbt.internal.inc.ScalaInstance
+import sbt.plugins.JvmPlugin
+
 object MetalsPlugin extends AutoPlugin {
-  override def requires = plugins.JvmPlugin
+  override def requires = JvmPlugin
   override def trigger = allRequirements
 
   val semanticdbVersion = BuildInfo.semanticdbVersion
@@ -49,7 +50,7 @@ object MetalsPlugin extends AutoPlugin {
       val targetRoot = semanticdbTargetRoot.value
       val versionOfScala = scalaVersion.value
       if (ScalaInstance.isDotty(versionOfScala))
-        List("-semanticdb-target", targetRoot.toString)
+        List.empty
       else
         List(
           s"-P:semanticdb:sourceroot:${baseDirectory.in(ThisBuild).value}",


ckipp01:
Actually hold on... this isn't working as expected. I'm now getting 
```
[error] not found: /Users/ckipp/Documents/scala-workspace/dotty-test/target/scala-0.27/test-meta
```

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
```diff
@@ -45,21 +46,38 @@ object MetalsPlugin extends AutoPlugin {
...
-        )
-    }
+    semanticdbOptions ++= createSemanticDbOptions(
+      scalaVersion.value,
+      semanticdbTargetRoot.value,
+      baseDirectory.in(ThisBuild).value
+    )
+  )
+  lazy val testSettings: Seq[Def.Setting[_]] = List(
+    // remove Compile targetRoot from Test config
```
tgodzik, line 56: Could you maybe comment more about why do we need to remove all semanticdb settings here? Not sure I understand why we need to remove all options here and how they got there in the first place.

tgodzik (approved):
Looks good! I also tested it locally and it is working alright. The only thing would that would be useful would be some more comments? It seems a bit counter intuitive that we need to remove anything from test settings and it's a bit hard to figure out what is going on.

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
ckipp01: To be 100% honest, I don't fully get what is going on. I wrestled with this for a while since I was assuming that this is what I wanted:

https://github.com/sbt/sbt/blob/c00196abf0c0b60d301e9516b8513b498359f987/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala#L75

However, that actually messed things up for me more. So instead I just adapted the other stuff that sbt has, and that ended up working for me. I plan on spending a bit more time on this to get a better understanding because I need to get what's happening here.

ckipp01: Actually thinking about this some more, I don't need to remove all of those, especially when it falls into the Scala 2 else. Let me adapt this a bit.

ckipp01: So it looks like without our plugin the `semanticdb-target` is already set correctly. So there is no need for us to set it I think.

ckipp01:
So I was way over thinking this. With Scala 3 the `semanticdb-target` is already correctly set. What I was doing was setting it again without realizing it duplicating it, and then in `Test` I needed to explicitly remove it because the extra setting I added in was still there. Instead, I just don't need to add it for Scala 3. Much simpler and fixes it.

tgodzik (approved):
LGTM! Looks much simpler now.

PR #2190, opened 2020-11-04
Add support for Scala3 3.0.0-M1

tgodzik:

PR #2194, opened 2020-11-07
Show warning about unsupported Scala version in worksheets

ckipp01:
This is a quick follow-up to my comment in #2190 about showing the message to the user when they are using a worksheet but their Scala version isn't supported. So now instead of just logging it, we both log it and show it to the user.

Another question that I _think_ I remember being discussed somewhere, but I find it a bit unexpected that if I'm using 0.27-RC1 we show the message about what version they _should_ be using, but then instead of defaulting to that version, we default to 2.12.12. I'm assuming because we already have everything we need for it, but would it be possible to instead of defaulting to 2.12.12, we detect that the user wants a Scala 3 version and default to the recommended Scala 3 instead? Basically make the rambo compiler closer to what they want.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
@@ -18,6 +18,18 @@ import org.eclipse.lsp4j.ShowMessageRequestParams
  */
 object Messages {
 
+  object Worksheets {
+    def unsupportedScalaVersion(
+        unsupportedVersion: String,
+        fallback: String,
+        recommended: String
+    ) = new MessageParams(
+      MessageType.Warning,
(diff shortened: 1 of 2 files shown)


tgodzik:
> Another question that I think I remember being discussed somewhere, but I find it a bit unexpected that if I'm using 0.27-RC1 we show the message about what version they should be using, but then instead of defaulting to that version, we default to 2.12.12. I'm assuming because we already have everything we need for it, but would it be possible to instead of defaulting to 2.12.12, we detect that the user wants a Scala 3 version and default to the recommended Scala 3 instead? Basically make the rambo compiler closer to what they want.

I need to work on that together with the standalone compiler, it's a similar case and it would be good to add an option to choose.

tgodzik (approved):
LGTM!

PR #2196, opened 2020-11-09
Rename user configuration options for decoration provider

tgodzik:
While writing the release notes it dawned on me that the previous names might not be too clear, especially the `annotation` part.

It's not yet release so changing the names should not be a problem, what do you think?


ckipp01 (approved):
I'm all for the clarity, and I think this is better. I don't think this will really affect too many people seeing that people would have needed to have these set to use them anyways, and the extension wasn't released. 👍

PR #2199, opened 2020-11-10
Update release docs

tgodzik:
Noticed that we are a bit behind :sweat_smile:


ckipp01 (approved):
Haha, good call. LGTM!

PR #2203, opened 2020-11-11
Fix scala3Dependency for various binary scala3/dotty versions

### LINKED ISSUE #2201: DownloadDependencies fails when trying to download scala3-library_0.27
```
https://repo1.maven.org/maven2/org/scala-lang/scala3-library_0.27/0.27.0-RC1/scala3-library_0.27-0.27.0-RC1.pom
```
doesn't exist.

In the PR https://github.com/che-dockerfiles/che-sidecar-scala/pull/9 , logs: https://github.com/che-dockerfiles/che-sidecar-scala/pull/9/checks?check_run_id=1381752614#step:4:747

The command that causes it
```bash
METALS_VERSION="0.9.5"
...
coursier launch -r sonatype:snapshots org.scalameta:metals_2.12:$METALS_VERSION --main scala.meta.metals.DownloadDependencies
```

ckipp01:
This is a quick fix for #2201. This was causing us to try and pull 0.27 from `org:scala-lang`,  which isn't where it lives. Also as a side now, we should add this as one of the checks we do on release to be sure this works. Better yet, we should test to ensure it's continually working.

Closes #2201

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
@@ -156,11 +156,19 @@ object Embedded {
   private def scala3Dependency(scalaVersion: String): Dependency = {
     val binaryVersion =
       ScalaVersions.scalaBinaryVersionFromFullVersion(scalaVersion)
-    Dependency.of(
-      "org.scala-lang",
-      s"scala3-library_$binaryVersion",
-      scalaVersion
-    )
+    if (binaryVersion.startsWith("3")) {
+      Dependency.of(
+        "org.scala-lang",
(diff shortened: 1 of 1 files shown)


kpbochenek (approved):
LGTM :+1:


For writing test, it would potentially have to be after actually doing a release?, more like a validation step as command in issue already requires released metals, and shorter version:
`sbt "metals/runMain scala.meta.metals.DownloadDependencies"` 
requires mtags to be published. For some tests mtags already need to be published so maybe not a problem after all?

tgodzik (approved):
LGTM

tgodzik:
We can add it as a validation step after doing the release. Let me take a look at it.

PR #2205, opened 2020-11-11
Fix release notes to mention emacs

tgodzik:
CC @kurnevsky - what do you think?

### DIFF
--- a/website/blog/2020-11-10-lithium.md
+++ b/website/blog/2020-11-10-lithium.md
@@ -147,10 +147,16 @@ Both new options are disabled by default, since they add a great number of
 information, which might not be necessary for all users. Full name with package
 for each of the additional types or values will be available on hover.
 
+These options can be set in Metals options in VS Code or in Emacs
+[(lsp-metals)](https://github.com/emacs-lsp/lsp-metals) using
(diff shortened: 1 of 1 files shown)


website/blog/2020-11-10-lithium.md
```diff
@@ -147,10 +147,19 @@ Both new options are disabled by default, since they add a great number of
 information, which might not be necessary for all users. Full name with package
 for each of the additional types or values will be available on hover.
 
+These options can be set in Metals options in VS Code or in Emacs
+([lsp-metals])(https://github.com/emacs-lsp/lsp-metals) using:
+
+```
+(customize-set-variable 'lsp-metals-show-implicit-arguments t)
```
kurnevsky, line 154: I'd probably just use `setq` like:

```elisp
(setq lsp-metals-show-implicit-arguments t)
```

Or we can avoid it at all, because emacs has its own UI to modify custom user variables that can be shown with `customize-variable`.

tgodzik: Changed to just mentioned variable names. Is this ok? I am totally lost when it comes to emacs :sweat:

kurnevsky (approved):

PR #2215, opened 2020-11-12
Add additional newline to fix problems with markdown

tgodzik:
It seems if any previous content ends with a list, `*` gets treated as an another item.

metals-1605202520026

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
@@ -226,7 +226,7 @@ class Compilers(
       ds.asScala.headOption match {
         case None =>
           diagnostics.onNoSyntaxError(path)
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -226,7 +226,7 @@ class Compilers(
       ds.asScala.headOption match {
         case None =>
           diagnostics.onNoSyntaxError(path)
-        case Some(diagnostic) =>
+        case Some(diagnostic) if !path.isInReadonlyDirectory(workspace) =>
```
tgodzik, line 229: I also added a small change to not publish syntax errors in library sources. It seems to happen sporadically for Scala 3 sources.

ckipp01 (approved):
Looks good. LGTM!

PR #2234, opened 2020-11-16
Reorganize docs and add in FAQ.

ckipp01:
This breaks up the side bar a bit. Up to this point we just had the three following sections:

- Text Editors
- Build Tools
- Contributing

Screenshot 2020-11-16 at 13 40 34

A whole bunch of stuff has found its way into Contributing that probably doesn't belong there, so I've tried to shift things around a bit. Now you'll see the following:

- Text Editors
- Build Tools
- Contributing
- LSP Extensions/DAP
- Integrating with Metals
- Troubleshooting

Screenshot 2020-11-16 at 13 41 10

I also added in a new faq page that I included some of the common questions that I've been seeing lately. There are definitely more, so feel free to add some as well if there are any that pop into mind or as you come across them.

Let me know what you think of the new structure, if you'd rather things shifted around, etc.


docs/troubleshooting/faq.md
```diff
@@ -0,0 +1,119 @@
+---
+id: faq
+title: Frequently asked questions
+---
+
+This is a collection of frequently asked questions. We do our best to ensure
+everything is included in the docs in the relevant places, but sometimes it's
+either hard to explain in the context of the docs or there just isn't a good
```
tgodzik, line 8:
```suggestion
either hard to explain in the context of the docs or there just isn't a good enough
```
?

docs/troubleshooting/faq.md
```diff
@@ -0,0 +1,119 @@
...
+  "source.organizeImports": true
+}
+```
+
+This happens because [Scalafix](https://scalacenter.github.io/scalafix/)(which
+is used for this feature) needs your code saved before it can organize your
+imports. There are some conversations going on about how to achieve this in the
+future, but for now you'll need to save your file before using this feature.
+
+For context: https://github.com/scalameta/metals/issues/2120
```
tgodzik, line 119:
```suggestion
For context [#2120](https://github.com/scalameta/metals/issues/2120)
```

website/i18n/en.json
```diff
@@ -83,6 +83,32 @@
...
+      "integrations/new-build-tool": {
+        "title": "Integrating a new build tool"
+      },
+      "integrations/new-editor": {
+        "title": "Integrating a new editor"
+      },
+      "integrations/remote-language-server": {
+        "title": "Remote Language Servers"
+      },
+      "lsp-extensions/debug-adapter-protocol": {
```
tgodzik, line 96: I think lsp-extensions and integrations are pretty similar. Integrating a new editor for example might need to implement some of the extensions. I would keep it all under `integrations`

tgodzik (changes requested):
Great work @ckipp01 ! Thanks for improving the docs. I just have a couple of questions.

olafurpg (approved):
The new sidebar structure looks great! Thank you @ckipp01  🙏

docs/troubleshooting/faq.md
ckipp01: Uh, both the original sentence and the suggestion basically means exactly the same thing. I'll add in enough to make it explicit.

tgodzik: Yeah, I just felt it was missing there :sweat_smile: You can fully ignore it.

website/i18n/en.json
ckipp01: Sounds good. I moved them all under integrations.

ckipp01:
> Great work @ckipp01 ! Thanks for improving the docs. I just have a couple of questions. (Edit: or really comments)

Sound good. Changed it around a bit per your suggestion.

Screenshot 2020-11-16 at 15 24 29

tgodzik (approved):
LGTM!

PR #2238, opened 2020-11-17
Add sbt and bloop directory as bsp directory

akilegaspi:
As of the moment, even if you do have sbt in your path, the config generator will still go to the tmp directory of the metals folder that has the sbt-launcher.

This commit allows us to detect the sbt-launcher that maybe BSP capable

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspServers.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspServers.scala
@@ -157,8 +157,10 @@ final class BspServers(
 
 object BspServers {
   def globalInstallDirectories: List[AbsolutePath] = {
-    val dirs = ProjectDirectories.fromPath("bsp")
-    List(dirs.dataLocalDir, dirs.dataDir).distinct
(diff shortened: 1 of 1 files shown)


ckipp01:
> As of the moment, even if you do have sbt in your path, the config generator will still go to the tmp directory of the metals folder that has the sbt-launcher.
> 
> This commit allows us to detect the sbt-launcher that maybe BSP capable

Thanks for the pr @akilegaspi. Can you expand a bit more on this change and how it will address the concern you brought up? The method that you changed actually just detects the `bsp` global install directories which we then later look in those to see if there is a `*.json` file in there to see what BSP servers may be available. So this really doesn't have anything to do with using sbt on your path instead of our embedded launcher. Plus, if possible we also want to make sure that this method stays within the BSP spec and where things should be defined. I noticed you also have some bloop logic in here, and Metals has pretty specific logic to detect Bloop on it's own, so this isn't needed here.

akilegaspi:
Sorry @ckipp01 I completely misunderstood what the global does and our error was actually more of in our project for this case.
i'm closing this PR now.

PR #2241, opened 2020-11-17
Update setup-gpg action

gabro:
setup-gpg seems to have issues as well since the new changes introduced by GitHub, so the release workflow is broken. Updating to v3 seems to have fixed the same issue in MUnit, so let's try here as well


gabro:
Ah, dependabot beat me to it https://github.com/scalameta/metals/pull/2240 😍

Issue #2204, opened 2020-11-11
Unable to use tree view with sbt BSP


ckipp01:
**Describe the bug**

When using sbt BSP, if you try to dive down into the tree view panel, compilation goes nuts. I'm not really sure what's happening yet and will need to dive into this, but wanted to submit this just in case someone else comes across it.

**To Reproduce**

1. Open a project
2. Switch to sbt BSP
3. Make sure project is compiled
4. Open the side panel and try to drill down into the package
5. Constant compilations happens

**Expected behavior**

Not constant compilation

**Screenshots**

![2020-11-11 11 07 07](https://user-images.githubusercontent.com/13974112/98799057-e8def500-240e-11eb-8799-85bfc53e1621.gif)

**Installation:**

- Operating system: macOS
- Editor: Tried on both VS Code and coc-metals
- Metals version: v0.9.5

**Search terms**

Tree view sbt BSP

tgodzik:
Thanks for finding this! I can confirm it happens also in my case. The problem is probably something around https://github.com/scalameta/metals/blob/48b20201f3357738ce472c40dcae47b0cccc56fb/metals/src/main/scala/scala/meta/internal/tvp/MetalsTreeViewProvider.scala#L69-L77

But it seems super weird.

ckipp01:
So I'm having a way harder time than expected tracking this one down.

Am I correct in understanding that the only reason the `doCompile(id)` in the snippet up above exists is to populate the tree if using Bloop and you have not yet triggered a compilation, but open the tree view? I'm still trying to figure out what exactly is causing the cyclical call to compile, but one of the first things I tried was just commenting the `doCompile(id)` out, and everything worked as expected with both Bloop and sbt _except_ in the situation where I opened up the `build.sbt` only, and then opened the treeview while using Bloop to find it empty until I navigated to a build target.

ckipp01:
Closed via #2242

PR #2242, opened 2020-11-17
Fix continuous compilation when opening tree view in sbt BSP

ckipp01:
So I've put more time into this than I care to admit. I'll be honest
that this isn't the fix that I really want, but sort of the fix that may
just makes the most sense at the moment.

This is an interesting one. So for a bit of background I'm assuming that
the `doCompile` was first added in the situation where a user opens up
their `build.sbt` while using Bloop, and then tries to open the tree
view. Without this compile request here, the tree view projects will be
empty. However, with it, it will trigger the necessary compilation and
populate the tree. However, subsequent compile requests or notifications
of compiles seem to cause no problems with Bloop. _Even_ though a
compile request _probably_ isn't needed here every time. However, with
sbt BSP, since loading the build up includes compilation, this isn't
needed. Just commenting it out when using sbt BSP works fine. It took me
a while to figure this out, but this is due to Bloop returning no-op
compilations and us marking the compilation as such. If we remove that
check from compilations and never mark any of Bloop's compilations as
no-op's then we get the same issue when using Bloop and there is a
continuous compile request cycle.

A quick fix for this is to just add in a check to see if the user is
using Bloop, and if so leave the `doCompile` to account for the cases it
is needed, but then if the user isn't using Bloop, then skip it.
Everything seems to work fine without it when not using Bloop.

Again, I think there may still be some logic in there that isn't right,
for example the `pendingProjectUpdates` in `MetalsTreeViewProvider`
never gets anything taken out of it, so it's always has an entry. That
may in part be the issue, but until someone takes the time and does a
deep dive into the TVP stuff to see what's going on, this fixes the
issue.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/tvp/MetalsTreeViewProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/tvp/MetalsTreeViewProvider.scala
@@ -26,7 +26,8 @@ class MetalsTreeViewProvider(
     definitionIndex: GlobalSymbolIndex,
     statistics: StatisticsConfig,
     doCompile: BuildTargetIdentifier => Unit,
-    sh: ScheduledExecutorService
+    sh: ScheduledExecutorService,
+    isBloop: () => Boolean
 ) extends TreeViewProvider {
   private val ticks =
     TrieMap.empty[String, ScheduledFuture[_]]
@@ -67,7 +68,7 @@ class MetalsTreeViewProvider(
       )
     },
     { (id, symbol) =>
-      doCompile(id)
+      if (isBloop()) doCompile(id)
       buildTargets.scalacOptions(id) match {
         case None =>
           Nil.iterator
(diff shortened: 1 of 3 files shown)


ckipp01:
As stated in the commit, I don't think this fixes the root issue, but this does fix _the issues we are seeing in the ticket_. Let me know what you think is best, to create a follow up ticket with the details I provided, or maybe I'm just completely missing something. I was going to continue to try and search and figure it out, but I went in so many circles with this that I'm honestly just a bit burnt out with it, and wanted to send this in to fix the immediate issue.

Closes #2204

metals/src/main/scala/scala/meta/internal/tvp/MetalsTreeViewProvider.scala
```diff
@@ -67,7 +68,7 @@ class MetalsTreeViewProvider(
       )
     },
     { (id, symbol) =>
-      doCompile(id)
+      if (isBloop()) doCompile(id)
```
tgodzik, line 71: Should be good for a quick fix for sure. I haven't been able to take a proper look here, but I think this is perfectly fine.

tgodzik (approved):
LGTM

PR #2243, opened 2020-11-17
Add source and line to debug output in case of stack traces

tgodzik:
Just found out this is actually pretty simple to do and should work in all clients.

metals-sample-1605631480731

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProxy.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProxy.scala
@@ -10,6 +10,8 @@ import scala.concurrent.Promise
 import scala.meta.internal.metals.Cancelable
 import scala.meta.internal.metals.Compilers
 import scala.meta.internal.metals.GlobalTrace
+import scala.meta.internal.metals.StacktraceAnalyzer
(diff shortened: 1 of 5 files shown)


metals/src/main/scala/scala/meta/internal/metals/debug/DebugProxy.scala
```diff
@@ -110,6 +114,20 @@ private[debug] final class DebugProxy(
       }
       response.setResult(args.toJson)
       client.consume(response)
+    case message @ ErrorOutputNotification(output) =>
+      Try {
+        stackTraceAnalyzer
+          .fileLocationFromLine(output.getOutput())
+      }.toOption.flatten
```
ckipp01, line 121: I may be totally missing something, but `fileLocationFromLine` here just returns an `Option` and `getOutput` a `String`. Why do we need to wrap in a `Try` here?

tgodzik: It could actually throw still from substring. Fixed it to not throw and left it with option. Good point!

ckipp01: Ahh true, missed the substring. 👍

ckipp01 (approved):
Looks good! LGTM!

PR #2246, opened 2020-11-19
Refresh synthetics when user settings change

tgodzik:
This enables us to quickly toggle the decorations in the clients:

metals-sample-1605780598336

@ckipp01 I think since the settings are managed by the clients, we need to do the actual toggle command in the client. This will be most useful in VS Code and not sure if it will be helpful in other editors. Especially without decorations it doesn't change anything.

### DIFF
--- a/docs/editors/overview.md
+++ b/docs/editors/overview.md
@@ -13,7 +13,6 @@ specific sections.
 
 ```
 
-
 ## Editor support
 
 Metals works with the following text editors with varying degree of
@@ -213,6 +212,15 @@ functionality.
      <td align="center">✅</td>
      <td align="center"></td>
   </tr>
+  <tr>
+     <td>Implicit decorations</td>
+     <td align="center">✅</td>
+     <td align="center">Shown in hover</td>
+     <td align="center">Shown in hover</td>
+     <td align="center">Shown in hover</td>
+     <td align="center">✅</td>
+     <td align="center">Shown in hover</td>
+  </tr>
 </tbody>
 </table>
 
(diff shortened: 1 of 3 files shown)


docs/editors/overview.md
```diff
@@ -418,6 +427,12 @@ Language Server Protocol (LSP). These extensions are not necessary for Metals to
 function but they improve the user experience. To learn more about Metals
 extensions, see [integrating a new editor](../integrations/new-editor.md).
 
+## Implicit decorations 
+
+**✅**: Additional information inferred from the code can be show within the code using additional virtual text.
```
ckipp01, line 432:
```suggestion
**✅**: Additional information inferred from the code can be show within the code using virtual text.
```
Super nitpicky, I think we can just remove the additional here.

docs/editors/overview.md
```diff
@@ -418,6 +427,12 @@ Language Server Protocol (LSP). These extensions are not necessary for Metals to
...
+
+**Shown in hover**: Additional information inferred from the code can be show when hovering over a specific line. That hover only shows the current line.
```
ckipp01, line 434:
```suggestion
**Shown in hover**: Additional information inferred from the code can be show when hovering over a symbol. That hover only shows information for the current line the symbol is one.
```
Not sure if you can actually just hover anywhere on the in other letters to trigger this or if it actually has to be on something. 🤔 take it or leave it.

ckipp01 (approved):
Just a couple nits on wording in the docs, but other than that, LGTM!

ckipp01:
> @ckipp01 I think since the settings are managed by the clients, we need to do the actual toggle command in the client. This will be most useful in VS Code and not sure if it will be helpful in other editors. Especially without decorations it doesn't change anything.

For sure, this looks fantastic in VS Code 😍

docs/editors/overview.md
tgodzik: Fixed!

docs/editors/overview.md
tgodzik: This is anywhere on the line, since it seemed natural as we show the whole line. I will leave it with `line`

ckipp01: 👍

PR #2247, opened 2020-11-19
Update scalameta to 4.4.0 and add support for Scala 2.13.4

tgodzik:


tests/cross/src/test/scala/tests/pc/CompletionDocSuite.scala
```diff
@@ -200,15 +210,25 @@ class CompletionDocSuite extends BaseCompletionSuite {
...
+        s"""
+           |${predefDocString(commonlyUsedTypesPre2134)}
+           |Predef scala
+           |""".stripMargin,
+      "2.13.3" ->
+        s"""
+           |${predefDocString(commonlyUsedTypesPre2134)}
+           |Predef scala
+           |""".stripMargin,
+      "2.13.4" ->
```
ckipp01, line 229: If "2.13.2", "2.13.3", and "2.13.4" are all the same can't this just stay "2.13" instead of separating them out?

ckipp01 (commented):
Jus

tests/cross/src/test/scala/tests/pc/CompletionDocSuite.scala
tgodzik: There are actually different, 2.13.4 has changed docs. Might not be that noticable with the naming :sweat_smile:

tests/unit/src/test/resources/definition/example/NamedArguments.scala
```diff
@@ -22,8 +22,8 @@ object NamedArguments/*NamedArguments.scala*/ {
 
   // anonymous classes
   @deprecated/*deprecated.scala*/(
-    message/*<no symbol>*/ = "a",
-    since/*<no symbol>*/ = susan/*NamedArguments.scala*/
+    message/*deprecated.scala fallback to scala.deprecated#*/ = "a",
```
tgodzik, line 25: So this is fixed, but the main issue I tried to fix needed to be reopnened unfortunately https://github.com/scalameta/scalameta/issues/2144

tests/unit/src/test/resources/foldingRange/expect/Blocks.scala
```diff
@@ -17,23 +17,23 @@ class A >>region>>{
 
   def longSingleLineBlock = { ???; ???; ???; }
 
-  def chain = >>region>>Seq(1).map>>region>>{
-    x =>
+  def chain = >>region>>Seq(1).map{
+    >>region>>x =>
```
tgodzik, line 21: So this will work better and is actually a fix. Who fixed this you ask? I have no idea :sweat_smile:

tgodzik:
Cross tests seem to be failing - probably due to updating 2.13.4 and dotty compat :/

tests/cross/src/test/scala/tests/pc/CompletionDocSuite.scala
tgodzik: `commonlyUsedTypesPre2134` and `commonlyUsedTypesPost2134` Post/Pre

Any idea on naming? This one is pretty bad I guess.

ckipp01: Honestly, I don't think it's that bad. I just _totally_ glossed over the `Post/Pre` part. Do you think it's clearer if you preface it with `post1234`/`pre1234` instead of the suffix? /shrug

tests/unit/src/test/resources/foldingRange/expect/Blocks.scala
ckipp01: Great? haha

kpbochenek (approved):
Yeah looks like you found a workaround with mtags,

tests pass ==> LGTM :+1:

tgodzik:
I plan to run the release tomorrow. Does anyone mind if I include https://github.com/scalameta/metals/pull/2246 ?

ckipp01:
> I plan to run the release tomorrow. Does anyone mind if I include #2246 ?

So will we do a release tomorrow and then again right away for 2.12.13? As for #2246, for sure, it looks good!

tgodzik:
> > I plan to run the release tomorrow. Does anyone mind if I include #2246 ?
> 
> So will we do a release tomorrow and then again right away for 2.12.13? As for #2246, for sure, it looks good!

Thanks for checking it out. I think we shouldn't wait for 2.12.13, it might be still a couple of day. It's not a problem to run a quick release in this case.

ckipp01 (approved):
> Thanks for checking it out! 🎉 I think we shouldn't wait for 2.12.13, it might be still a couple of days. It's not a problem to run a quick release in this case.

Sounds good! LGTM!

PR #2251, opened 2020-11-22
New directories namespace `io.github.soc` -> `dev.dirs`

ckipp01:
So it looks like this changed namespaces from their personal one to a `dev.dirs` namespace. I just randomly came across this as I was poking around at some of the related code. So this just updates from `io.github.soc` -> `dev.dirs` and bumps the version to the latest.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -424,7 +424,7 @@ lazy val metals = project
       // for producing SemanticDB from Java source files
       "com.thoughtworks.qdox" % "qdox" % "2.0.0",
       // for finding paths of global log/cache directories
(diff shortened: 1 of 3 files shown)


kpbochenek (approved):
LGTM :+1: 

Could you only change a commit message(or PR message? the one used for Release Notes) to make more sense when looking at future Release Notes?

ckipp01:
> Could you only change a commit message(or PR message? the one used for Release Notes) to make more sense when looking at future Release Notes?

Good point, updated both.

Issue #1941, opened 2020-07-23
Metals doesn't offer autofill scaladoc comment without underlying string of valid scala code


banzecrew:
When i try to insert scaladoc comment, metals offer do this automatically like at screenshot

<img width="538" alt="Screenshot 2020-07-16 at 13 29 59" src="https://user-images.githubusercontent.com/17661856/87660966-8def0700-c768-11ea-9b74-58c52dbcee7a.png">

As you can see, there is valid definition of function "repeatString" under the comment

But if i:
<ul>
<li>remove</li>
<li>comment</li>
<li>write comment after function definition</li>
</ul>

then metals will not offer me an autofilling

<img width="320" alt="Screenshot 2020-07-16 at 13 41 51" src="https://user-images.githubusercontent.com/17661856/87661942-25a12500-c76a-11ea-85cb-b1820aa8271c.png">
<img width="345" alt="Screenshot 2020-07-16 at 13 42 58" src="https://user-images.githubusercontent.com/17661856/87662028-4a959800-c76a-11ea-9053-85a028ae1e92.png">
<img width="166" alt="Screenshot 2020-07-16 at 13 43 49" src="https://user-images.githubusercontent.com/17661856/87662114-6731d000-c76a-11ea-904d-dd05f1b64707.png">

**Installation:**

- Operating system: macOS
- VSCode version: June 2020 (version  1.47.1)
- Metals version: v1.9.2

**Search terms**

scaladoc autofill

PR #2252, opened 2020-11-23
Autocomplete empty scaladoc if nothing is defined afterwards

tanishiking:
This enables us to autocomplete empty scaladoc even if nothing is defined after the `/**`, like:

scaladoc

fix https://github.com/scalameta/metals/issues/1941

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/ScaladocCompletions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/ScaladocCompletions.scala
@@ -21,7 +21,7 @@ trait ScaladocCompletions { this: MetalsGlobal =>
    */
   case class ScaladocCompletion(
       editRange: l.Range,
-      associatedDef: MemberDef,
+      associatedDef: Option[MemberDef],
       pos: Position,
       text: String
   ) extends CompletionPosition {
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/completions/ScaladocCompletions.scala
```diff
@@ -81,7 +90,9 @@ trait ScaladocCompletions { this: MetalsGlobal =>
             editRange,
             newText
           ),
-          completionsSymbol(associatedDef.name.toString()),
+          completionsSymbol(
+            associatedDef.map(d => d.name.toString()).getOrElse("Scaladoc")
```
tanishiking, line 94: As long as I know, [sym](https://github.com/scalameta/metals/blob/5d32828171f202402088265e90066dd34fa49567/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala#L39) is used only for ordering the completion members (based on its attributes) here: https://github.com/scalameta/metals/blob/5d32828171f202402088265e90066dd34fa49567/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala#L152

Since the auto-completion for scaladoc always returns only one member, it's fine to give an arbitrary symbol to `sym` for `TextEditMember`.

tgodzik (approved):
LGTM! Thanks for taking a look at it!

PR #2254, opened 2020-11-23
make launch.json doc simpler, add buildTarget,  fix #2253 

### LINKED ISSUE #2253: Improve docs about launch configuration
**Describe the bug**

We should better explain what is a buildTarget and when it should be specified. We can start with a minimal example and then go into details when anything additional might be needed.

**Search terms**

launch documentation

bjornregnell:
fix #2253

### DIFF
--- a/docs/editors/overview.md
+++ b/docs/editors/overview.md
@@ -13,7 +13,6 @@ specific sections.
 
 ```
 
-
 ## Editor support
 
 Metals works with the following text editors with varying degree of
@@ -213,6 +212,15 @@ functionality.
      <td align="center">✅</td>
      <td align="center"></td>
   </tr>
+  <tr>
+     <td>Implicit decorations</td>
+     <td align="center">✅</td>
+     <td align="center">Shown in hover</td>
+     <td align="center">Shown in hover</td>
+     <td align="center">Shown in hover</td>
+     <td align="center">✅</td>
+     <td align="center">Shown in hover</td>
+  </tr>
 </tbody>
 </table>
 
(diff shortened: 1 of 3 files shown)


docs/editors/vscode.md
```diff
@@ -177,33 +177,32 @@ guide them. In the end users should end up with something like this:
...
-      // optional build target name in case there more than one
-      // class with the same name
-      "buildTarget": "root"
+      "testClass": "com.example.Test"
     }
   ]
 }
 ```
 
+You can add an optional build target name in case there more than one class with the same name. Inside `"configurations":` add the key `buildTarget` with your target name, e.g. `root`:
```
tgodzik, line 201:
```suggestion
You can also add an optional build target name, which is needed in case there are more than one class with the same name or when launching a class from outside the project. Inside `"configurations":` add the key `buildTarget` with your target name, e.g. `root`:
```

We can also add a paragraph explaining what a build target is after the json:

```
Build target name corresponds to your project name. For example in sbt for `lazy val interfaces = project` the name of the build target will be `interfaces` for sources and `interfaces-test` for tests. To make sure what are the target names please run the command `Metals: Run Doctor`.
```

tgodzik (changes requested):
Thanks for raising this! I added some suggestions on how to explain build target and bit about when it's needed.

bjornregnell:
Yes, good changes!

bjornregnell:
I've clicked "Accept changes" here just now. Should I do something else here or in my fork? Or is it good to go now? @tgodzik

tgodzik:
Ach, I wanted to push and add one paragraph, but that flopped :sweat:  Sorry about that, I will follow up with a new PR

PR #2258, opened 2020-11-24
Don't add semanticDB plugin for Scala 2.10 when using sbt BSP

### LINKED ISSUE #2250: sbt-metals should not use semanticdb on scala 2.10 project
**Describe the bug**

I cannot import a project that is cross-built on Scala 2.10.7 because of the semanticdb plugin that is added by `sbt-metals`:
```
sbt.librarymanagement.ResolveException: Error downloading org.scalameta:semanticdb-scalac_2.10.7:4.3.24
```

**To Reproduce**

1. Create a Scala 2.10 project in sbt
2. Import in Metals using sbt

**Expected behavior**

sbt-metals should not try to activate the smenanticdb plugin when Scala < 2.11

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.9.5

**Search terms**

sbt, semanticdb, 2.10

ckipp01:
This addresses the problem outlined in #2250 where the semanticdb compiler plugin was being added to a 2.10 projects as well as semanticdb related settings in both `scalacOptions` and `semanticdbOptions`. Now this just adds in a check to ensure we don't do any of that for projects on 2.10.x.

Closes #2250

### DIFF
--- a/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
+++ b/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
@@ -12,6 +12,7 @@ object MetalsPlugin extends AutoPlugin {
   override def trigger = allRequirements
 
   val semanticdbVersion = BuildInfo.semanticdbVersion
+  val supportedScala2Versions = BuildInfo.supportedScala2Versions.toList
 
   override lazy val projectSettings: Seq[Def.Setting[_]] = Seq(
     semanticdbCompilerPlugin := {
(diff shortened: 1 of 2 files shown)


sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
```diff
@@ -13,33 +13,42 @@ object MetalsPlugin extends AutoPlugin {
 
   val semanticdbVersion = BuildInfo.semanticdbVersion
 
+  private def isDottyOr210(scalaVersion: String) = {
+    scalaVersion.startsWith("2.10") || ScalaInstance.isDotty(scalaVersion)
```
tgodzik, line 17: It might be better to use `supportedScalaVersions` like we do in ScalaVersions. Otherwise we might end up trying to add semanticdb for versions that do not have the support.

tgodzik (changes requested):
I would check if the specific scala version is supported to avoid any kind of trouble.

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
ckipp01: Yea, good idea. I've gone ahead and re-used the `supportedScala2` for this.

tgodzik (approved):
LGTM!

ckipp01:
Super random unrelated test failure, so I'm still going to just merge.
```
[error] Failed tests:
[error] 	tests.pc.CompletionSuite
```

PR #2259, opened 2020-11-24
Bump swoval to latest version

eatkins:
This new version of swoval supports Apple silicon using arm64.


ckipp01 (approved):
Thanks for sending this in @eatkins 👍 

LGTM!

PR #2260, opened 2020-11-25
Add support for Scala 3 3.0.0-M2

tgodzik:


build.sbt
```diff
@@ -214,7 +214,7 @@ lazy val V = new {
   def scala2Versions = nonDeprecatedScala2Versions ++ deprecatedScala2Versions
 
   // Scala 3
-  def nonDeprecatedScala3Versions = Seq(scala3, "0.27.0-RC1")
+  def nonDeprecatedScala3Versions = Seq(scala3, "3.0.0-M1", "0.27.0-RC1")
```
ckipp01, line 217: Just because I don't fully remember where we ended up on this. How long will we keep 0.27.0-RC1 around as non-deprecated?

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -182,9 +182,11 @@ object Embedded {
       scalaVersion: String,
       scalaBinaryVersion: String
   ): Dependency = {
-    // Mdoc support for 0.27.0-RC1 was dropped with "2.2.11"
+    // Mdoc support for 0.27.0-RC1 was dropped with "2.2.11" and for 3.0.0-M1 with "2.12.13"
     val mdocVersion =
-      if (scalaVersion == "0.27.0-RC1") "2.2.10" else BuildInfo.mdocVersion
+      if (scalaVersion == "0.27.0-RC1") "2.2.10"
+      else if (scalaVersion == "3.0.0-M1") "2.2.12"
+      else BuildInfo.mdocVersion
```
ckipp01, line 189: I honestly don't love that we are doing this. Was it just not possible to publish mdoc for both M1 and M2? It seems a bit extreme that have to have logic to include 1 of 3 different versions of a single dependency.

build.sbt
tgodzik: Probably under the end of the course at EPFL.

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
tgodzik: It seemed simpler to just put in the logic here. We need to double the code for each Scala 3 version currently and it's really only Metals doing it. I might have been too quick, but there was no real changes between all those changes that would impact worksheets.

ckipp01 (approved):
LGTM

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
tgodzik: After the changes in https://github.com/lampepfl/dotty/pull/10172 we should be able to just do https://github.com/scalameta/mdoc/commit/81e43c1d42dc54fbe14c32412ee0c2cc67544d03

PR #2262, opened 2020-11-26
Update release instructions

tgodzik:
After thinking about a bit it turns out we actual isntructions still do not reflect the actual process:
- release notes should be agreed upon before actually running the release - we might discover that something is missing at that point
- we did not mention that VS Code release should be run via the release page.
- other small fixes


ckipp01 (approved):
Nice! Thanks for doing this. LGTM!

PR #2264, opened 2020-11-26
Compare release candidates and milestones in addition to major.minor.patch

tgodzik:
This showed up as a warning when using 3.0.0-M1:

image

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
@@ -60,7 +60,7 @@ object ScalaVersions {
     latestBinaryVersionFor(scalaVersion)
       .map(latest =>
         latest != scalaVersion && SemVer
-          .isCompatibleVersion(latest, scalaVersion)
+          .isLaterVersion(latest, scalaVersion)
       )
       .getOrElse {
         val versions =
@@ -68,7 +68,7 @@ object ScalaVersions {
             isLatestScalaVersion.filter(isScala3Version)
           else
             isLatestScalaVersion.filter(!isScala3Version(_))
-        versions.forall(ver => SemVer.isCompatibleVersion(ver, scalaVersion))
+        versions.forall(ver => SemVer.isLaterVersion(ver, scalaVersion))
       }
   }
 
--- a/mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
+++ b/mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
@@ -1,15 +1,45 @@
 package scala.meta.internal.semver
 
+import scala.util.Try
+
 object SemVer {
 
-  case class Version(major: Int, minor: Int, patch: Int) {
+  case class Version(
+      major: Int,
+      minor: Int,
+      patch: Int,
+      releaseCandidate: Option[Int],
+      milestone: Option[Int]
+  ) {
     def >(that: Version): Boolean = {
       this.major > that.major ||
       (this.major == that.major && this.minor > that.minor) ||
-      (this.major == that.major && this.minor == that.minor && this.patch > that.patch)
+      (this.major == that.major && this.minor == that.minor && this.patch > that.patch) ||
+      // 3.0.0-RC1 > 3.0.0-M1
+      this.releaseCandidate.isDefined && that.milestone.isDefined ||
+      // 3.0.0 > 3.0.0-M2 and 3.0.0 > 3.0.0-RC1
+      (this.milestone.isEmpty && this.releaseCandidate.isEmpty && (that.milestone.isDefined || that.releaseCandidate.isDefined)) ||
+      // 3.0.0-RC2 > 3.0.0-RC1
+      comparePreRelease(that, (v: Version) => v.releaseCandidate) ||
+      // 3.0.0-M2 > 3.0.0-M1
+      comparePreRelease(that, (v: Version) => v.milestone)
+
     }
 
     def >=(that: Version): Boolean = this > that || this == that
+
+    private def comparePreRelease(
+        that: Version,
+        preRelease: Version => Option[Int]
+    ): Boolean = {
+      val thisPrerelease = preRelease(this)
+      val thatPrerelease = preRelease(that)
+      this.major == that.major && this.minor == that.minor && this.patch == that.patch &&
+      thisPrerelease.isDefined && thatPrerelease.isDefined && thisPrerelease
+        .zip(thatPrerelease)
+        .exists { case (a, b) => a > b }
+    }
+
   }
 
   object Version {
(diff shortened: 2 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
```diff
@@ -60,15 +60,15 @@ object ScalaVersions {
     latestBinaryVersionFor(scalaVersion)
       .map(latest =>
         latest != scalaVersion && SemVer
-          .isCompatibleVersion(latest, scalaVersion)
+          .isLaterVersion(latest, scalaVersion)
```
tgodzik, line 63: This was the main issue. `3.0.0-M1` was equal to `3.0.0-M2`, because we were comparing only major/minor/patch. 
First of all it should not say it's a future version if it's equal, second of all it would be good to support also milestones/release candidates now.

tests/unit/src/test/scala/tests/ScalaVersionsSuite.scala
```diff
@@ -143,4 +143,59 @@ class ScalaVersionsSuite extends BaseSuite {
...
+    assert(
+      SemVer.isCompatibleVersion("3.0.0-M1", "3.0.0-M2")
+    )
+  }
+
+  test("compare-milestones-M1<M2") {
+    assert(
+      !SemVer.isCompatibleVersion("3.0.0-M2", "3.0.0-M1")
+    )
+  }
```
ckipp01, line 157: These two tests will always be the same no? If one fails, the other should fail. If on passes the other should pass.

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
```diff
@@ -1,12 +1,22 @@
...
+
 object SemVer {
 
-  case class Version(major: Int, minor: Int, patch: Int) {
+  case class Version(
+      major: Int,
+      minor: Int,
+      patch: Int,
+      releaseCandidate: Int,
+      milestone: Int
```
ckipp01, line 12: Do you think it's best to model this as an `Int`? Would an `Option[Int]` not be more appropriate?

tgodzik: I tried with option but that became rather complex, since I had to include the scenario where we compared release candidate with milestone. Though this will not work currently, so I might need to rethink it and do it with option.

tests/unit/src/test/scala/tests/ScalaVersionsSuite.scala
tgodzik: I also switched the order of the parameters, so this are not strictly opposite tests. At least I have seen one fail and the other not 😅

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
tgodzik: 3.0.0 will be lower than 3.0.0-M1 currently, which I didn't think of before 😱

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
```diff
@@ -1,12 +1,22 @@
...
+  ) {
     def >(that: Version): Boolean = {
       this.major > that.major ||
       (this.major == that.major && this.minor > that.minor) ||
-      (this.major == that.major && this.minor == that.minor && this.patch > that.patch)
+      (this.major == that.major && this.minor == that.minor && this.patch > that.patch) ||
+      (this.major == that.major && this.minor == that.minor && this.patch == that.patch && this.releaseCandidate > that.releaseCandidate) ||
+      (this.major == that.major && this.minor == that.minor && this.patch == that.patch && this.releaseCandidate == that.releaseCandidate && this.milestone > that.milestone)
```
tanishiking, line 19: As you notice, we have to compare the Pre-Release version in a different way.

FYI, here's the logic of comparing semantic versions in scalafmt
https://github.com/scalameta/scalafmt/blob/f05136cdd5ceeeabdb6054658d315f7a0a3d025e/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtVersion.scala#L15-L18


Following the logic from scalafmt, these two lines would be something like this (if we use `0` for no pre-release version):

```scala
(this.major == that.major && this.minor == that.minor && this.patch == that.patch && isLaterPreRelease(this.releaseCandidate, that.releaseCandidate)) ||
(this.major == that.major && this.minor == that.minor && this.patch == that.patch && isLaterPreRelease(this.releaseCandidate, that.releaseCandidate)) && isLaterPreRelease(this.milestone, that.milestone)

// returns true if v1 is later than v2 (for pre-release version)
def isLaterPreRelease(v1: Int, v2: Int): Boolean {
  v1 == 0 || (v2 !== 0 && v1 > v2)
}
```

Comparing two semantic versions is always complexed 😅

tgodzik: Thanks for the hint! I went with the Option solution suggested by @ckipp01

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
tgodzik: Added option, aaaaand a lot more tests haha.

tests/unit/src/test/scala/tests/ScalaVersionsSuite.scala
tgodzik: Ach! I know how it's called. I am actually testing transitivity!

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
```diff
@@ -1,27 +1,71 @@
...
       this.major > that.major ||
       (this.major == that.major && this.minor > that.minor) ||
-      (this.major == that.major && this.minor == that.minor && this.patch > that.patch)
+      (this.major == that.major && this.minor == that.minor && this.patch > that.patch) ||
+      // 3.0.0 > 3.0.0-M2 and 3.0.0 > 3.0.0-RC1
+      (this.milestone.isEmpty && this.releaseCandidate.isEmpty && (that.milestone.isDefined || that.releaseCandidate.isDefined)) ||
+      // 3.0.0-RC2 > 3.0.0-RC1
+      comparePreRelease(that, (v: Version) => v.releaseCandidate) ||
+      // 3.0.0-M2 > 3.0.0-M1
+      comparePreRelease(that, (v: Version) => v.milestone)
```
tanishiking, line 23: Sorry, I think I gave a wrong hint...
How about comparing versions with different pre-release versions? The following test should pass, but I guess it fails now.
 
```
    assert(
      SemVer.isLaterScalaVersion("3.0.0-M1", "3.0.0-RC1")
    )
```

We have to compare `releaseCandidate` and `milestone` simultaneously.
Or store the label (like `RC` or `M` part) to `Version` and compare those labels by ASCII sort order (this would be more generic implementation).

```
  case class Version(
      major: Int,
      minor: Int,
      patch: Int,
      preRelease: Option[(String, Int)], // something like this?
  ) 
```

> Identifiers with letters or hyphens are compared lexically in ASCII sort order.
> https://semver.org/#spec-item-11

tgodzik: ach, righ!. Every time I think I got it right, there are still edge cases. I will add tests for that too and try figure it out. Good catch.

tgodzik: Added additional check:
```scala
      // 3.0.0-RC1 > 3.0.0-M1
      this.releaseCandidate.isDefined && that.milestone.isDefined ||
```

tanishiking (approved):
LGTM! :+1:

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
ckipp01: lol so many checks. Great job on figuring these out.

ckipp01 (approved):
Looks good, and thanks for adding in all the tests! LGTM!

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
tgodzik: Yeah, that's why I was not keen on options at first. But I think with the comments it's not that bad. And I added a lot of tests :sweat_smile:

PR #2267, opened 2020-11-30
Customization bloop server port with java properties

VictorMikhaylov:
Bloop server starts on 8212 port by default. We can't customize this port.
It prevents team work in remote developing.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
@@ -62,9 +62,6 @@ class BspConnector(
           scribe.info("No build server found")
           Future.successful(None)
         case ResolvedBloop =>
-          scribe.info(
-            s"Start new Bloop server on port ${userConfiguration.bloopNailgunPort.getOrElse("default port")}"
-          )
           bloopServers.newServer(workspace, userConfiguration).map(Some(_))
         case ResolvedBspOne(details)
             if details.getName() == SbtBuildTool.name =>
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
@@ -62,6 +62,7 @@ class BspConnector(
           scribe.info("No build server found")
           Future.successful(None)
         case ResolvedBloop =>
+          scribe.info(s"Start new Bloop server on port ${userConfiguration.bloopNailgunPort.getOrElse("default port")}")
           bloopServers.newServer(workspace, userConfiguration).map(Some(_))
         case ResolvedBspOne(details)
             if details.getName() == SbtBuildTool.name =>
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
@@ -62,7 +62,9 @@ class BspConnector(
           scribe.info("No build server found")
           Future.successful(None)
         case ResolvedBloop =>
-          scribe.info(s"Start new Bloop server on port ${userConfiguration.bloopNailgunPort.getOrElse("default port")}")
+          scribe.info(
+            s"Start new Bloop server on port ${userConfiguration.bloopNailgunPort.getOrElse("default port")}"
+          )
           bloopServers.newServer(workspace, userConfiguration).map(Some(_))
         case ResolvedBspOne(details)
             if details.getName() == SbtBuildTool.name =>
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
@@ -91,7 +91,9 @@ final case class MetalsServerConfig(
       "metals.allow-multiline-string-formatting",
       default = true
     ),
-    bloopPort: Option[Int] = None
+    bloopPort: Option[Int] = Option(System.getProperty("metals.bloop-port"))
+      .filter(_.forall(Character.isDigit(_)))
+      .map(_.toInt)
 ) {
   override def toString: String =
     List[String](
@@ -128,14 +130,9 @@ object MetalsServerConfig {
     System.getProperty("metals.client")
   )
 
-  val metalsBloopPort: Option[Int] = Option(
-    System.getProperty("metals.bloop-port")
-  ).filter(_.forall(Character.isDigit(_)))
-    .map(_.toInt)
-
   def base: MetalsServerConfig = MetalsServerConfig()
   def default: MetalsServerConfig = {
-    var msConfig = metalsClientType.getOrElse("default") match {
+    metalsClientType.getOrElse("default") match {
       case "vscode" =>
         base.copy(
           icons = Icons.vscode,
@@ -201,8 +198,5 @@ object MetalsServerConfig {
       case _ =>
         base
     }
-    if (metalsBloopPort.isDefined)
-      msConfig = msConfig.copy(bloopPort = metalsBloopPort)
-    msConfig
   }
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
@@ -106,7 +106,8 @@ final case class MetalsServerConfig(
       s"input-box=$isInputBoxEnabled",
       s"ask-to-reconnect=$askToReconnect",
       s"icons=$icons",
-      s"statistics=$statistics"
+      s"statistics=$statistics",
+      s"bloop-port=${bloopPort.map(_.toString()).getOrElse("default")}"
     ).mkString("MetalsServerConfig(\n  ", ",\n  ", "\n)")
 }
 object MetalsServerConfig {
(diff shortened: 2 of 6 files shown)


metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -62,6 +62,7 @@ class BspConnector(
           scribe.info("No build server found")
           Future.successful(None)
         case ResolvedBloop =>
+          scribe.info(s"Start new Bloop server on port ${userConfiguration.bloopNailgunPort.getOrElse("default port")}")
```
ckipp01, line 65: So I'm not actually 100% sure what happens with Bloop if the server is already running with the default port, and then Metals is started with this setting. @tgodzik how does Bloop respond here? 

- Does it just detect that Bloop is already running and therefore go right to the BSP connection like it normally would?
- Or does the launcher detect that a new port is desired, and then shutdown the other and start a new one?

ckipp01: Also related, depending on what happens in the above scenario, we may want to change this logging a bit to only log in the case that the port is defined, since falling into this case doesn't always mean that a _new_ server will be started.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -220,9 +221,23 @@ object UserConfiguration {
           |See https://scalameta.org/metals/docs/contributors/remote-language-server.html for
           |documentation on remote language servers.
           |""".stripMargin
+      ),
+      UserConfigurationOption(
+        "bloop-port",
+        """empty int ``.""",
+        "8212",
+        "Bloop Nailgun listen port",
```
ckipp01, line 229:
```suggestion
        "Bloop Nailgun port",
```

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -220,9 +221,23 @@ object UserConfiguration {
...
+        """|This port be used for the Bloop build tool plugin, for any supported build tool,
+           |while importing in Metals as well as for running the embedded server""".stripMargin
```
ckipp01, line 231:
```suggestion
        """|Instead of using the default port for the Bloop build server, use the port specified.""".stripMargin
```

ckipp01 (changes requested):

ckipp01:
Thanks for the pr @VictorMikhaylov! Just a few questions/nits.

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
VictorMikhaylov: > So I'm not actually 100% sure what happens with Bloop if the server is already running with the default port, and then Metals is started with this setting. @tgodzik how does Bloop respond here?
> 
>     * Does it just detect that Bloop is already running and therefore go right to the BSP connection like it normally would?
> 
>     * Or does the launcher detect that a new port is desired, and then shutdown the other and start a new one?

As I've understood, Metals can't detect Bloop on any different port. It just starts new Bloop on new port. We'll have 2 running Bloop.
Only if we start Bloop on default port, change IDE metals properties and restart Metals with IDE. In this case we'll have 1 running Bloop on new port.

ckipp01: > As I've understood, Metals can't detect Bloop on any different port. It just starts new Bloop on new port. We'll have 2 running Bloop.

I don't believe this is correct as there can only be one Bloop instance running at a time on a machine.

tgodzik: I am not sure how it's synchronized, but this might indeed be two Bloop server, I will check how it works in this PR to make sure.

ckipp01: > but this might indeed be two Bloop server

Really? I was always under the assumption that this was a limitation of Bloop since right in the docs it says

> There can only be one bloop server instance running per machine

https://scalacenter.github.io/bloop/docs/what-is-bloop#limitations

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -220,9 +221,22 @@ object UserConfiguration {
           |See https://scalameta.org/metals/docs/contributors/remote-language-server.html for
           |documentation on remote language servers.
           |""".stripMargin
+      ),
+      UserConfigurationOption(
```
tgodzik, line 225: I would not want to include it as a user config options, as most people will have no idea what this is about and will not need to change it. Could you add it as another option in `MetalsServerConfig` ?

tgodzik (changes requested):
Please change bloopPort to be a java property. This a specfic rare case scenario, so I wouldn't want to show it to most users.

I also wanted to ask what is exactly the use case here? I thought that the problem here is that you run Bloop on a separate machine remotely and that's what you need to change here. If you are using the same machine with multiple users at the same time you should not need to change anything. We only need one Bloop per machine.

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
tgodzik: You might be right, I am currently just wondering if we ever enforce it.

VictorMikhaylov: Nobody could prohibit me run several bloop servers:
`bloop server`
`bloop server 8214`
`bloop server 8215`
I've done it right now :)

It's another thing bloop cli can't detect server on port different from default
`bloop about` `bloop project` always start server on default port

VictorMikhaylov:
> Please change bloopPort to be a java property. This a specfic rare case scenario, so I wouldn't want to show it to most users.
> 
> I also wanted to ask what is exactly the use case here? I thought that the problem here is that you run Bloop on a separate machine remotely and that's what you need to change here. If you are using the same machine with multiple users at the same time you should not need to change anything. We only need one Bloop per machine.

User case is multiple users on the same machine at the same time.
Unfortunately, "one Bloom per machine" does not work. See logs
[metals.log](https://github.com/scalameta/metals/files/5627679/metals.log)

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
VictorMikhaylov: If this config option is in UserConfiguration user could change property in ~/.config/Code/User/settings.json (for example) oneself (or from IDE settings page - i think it's target solution), restart build server and continue work
If this config option is in MetalsServerConfig which setting file should be changed and who could change it?

tgodzik: You can add `-Dmetal-bloop-port=123` in `metals.serverProperties`, which will add the java property to the running server process.

tgodzik:
> > Please change bloopPort to be a java property. This a specfic rare case scenario, so I wouldn't want to show it to most users.
> > I also wanted to ask what is exactly the use case here? I thought that the problem here is that you run Bloop on a separate machine remotely and that's what you need to change here. If you are using the same machine with multiple users at the same time you should not need to change anything. We only need one Bloop per machine.
> 
> User case is multiple users on the same machine at the same time.
> Unfortunately, "one Bloom per machine" does not work. See logs
> [metals.log](https://github.com/scalameta/metals/files/5627679/metals.log)

Hmm... would it be possible to start Bloop separately under certain permissions, maybe an additional group? It would be more beneficial that everyone can use one instance of Bloop and not create another one that would drain system resources.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
VictorMikhaylov: Done

tgodzik (changes requested):
Thanks for the changes! There is still one small thing that we can fix. Did you maybe manage to try and launch Bloop under different permissions/group?

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -194,5 +201,8 @@ object MetalsServerConfig {
       case _ =>
         base
     }
+    if (metalsBloopPort.isDefined)
```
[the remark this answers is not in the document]
VictorMikhaylov: > This will already be available in base, so no need to add it here.

You're right
Fixed

VictorMikhaylov:
> Thanks for the changes! There is still one small thing that we can fix. Did you maybe manage to try and launch Bloop under different permissions/group?

Unfortunately, not. I try 

> Thanks for the changes! There is still one small thing that we can fix. Did you maybe manage to try and launch Bloop under different permissions/group?

Not yet
My team and I will continue looking for solution

tgodzik (approved):
I think it would be much better for your work env to reuse a single Bloop instance, but I think it will be useful to have this setting as an option. Any objections to this?

ckipp01:
> Any objections to this?

No objections from my end. However, do we want to have this mentioned at all in the docs?

tgodzik:
> > Any objections to this?
> 
> No objections from my end. However, do we want to have this mentioned at all in the docs?

I will follow up with an additional PR, I think we can merge this

ckipp01 (approved):
LGTM!

PR #2268, opened 2020-11-30
Add in Scala Sources to standalone worksheets.

### LINKED ISSUE #2266: Standalone worksheet goto definition not working on Scala sources
**Describe the bug**

I think we are hitting on the exact same thing we fixed for Java in https://github.com/scalameta/metals/pull/2133, but for Scala. I'm not really sure why I didn't add a test also testing this for Scala, seems silly looking back. However giving the following code:

```scala
val a: List[Int] = List.fi@@ll(2)(2)
```

**To Reproduce**

1. Create a standalone worksheet
2. Paste in the code above
3. Try to navigate to `fill`
4. See that no definition can be found

**Expected behavior**

Brings you to the definition of `fill` in `GenTraversableFactory.scala`

**Installation:**

- Operating system: macOS
- Editor: Vim
- Metals version: v0.9.

**Search terms**

standalone  worksheets

ckipp01:
Like we did to fix #2133, if the user is in a situation where they are
given a standalone compiler for worksheets, we then ensure that they not
only have Java sources available to them, but also Scala sources. We
default to the version of Scala in Metals itself.

I do have a check in there to make sure there is no "scala-library" in
sources, but that _probably_ isn't necessary as (in this scenario) the
only time that would happen would be if they have included it as an
import in their worksheet. However, just to be safe, I added it in.

I also moved the logic for this from the compilers to the
StandaloneSymbolSearch apply methods to ensure that we don't get hit on
this in the future if we use these elsewhere.

Closes #2266

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
@@ -88,7 +88,9 @@ class Compilers(
   // The "rambo" compiler is used for source files that don't belong to a build target.
   lazy val ramboCompiler: PresentationCompiler = createStandaloneCompiler(
     PackageIndex.scalaLibrary,
-    Try(StandaloneSymbolSearch(workspace, buffers, isExcludedPackage))
+    Try(
+      StandaloneSymbolSearch(workspace, buffers, isExcludedPackage, userConfig)
+    )
       .getOrElse(EmptySymbolSearch),
     "metals-default"
   )
@@ -429,11 +431,6 @@ class Compilers(
     }
 
     created.getOrElse {
-      val sourcesWithJdk = JdkSources(userConfig().javaHome) match {
-        case Some(absPath) => absPath.toNIO :: sources
-        case None => sources
-      }
-
       jworksheetsCache.put(
         path,
         createStandaloneCompiler(
@@ -441,9 +438,10 @@ class Compilers(
           StandaloneSymbolSearch(
             workspace,
             buffers,
-            sourcesWithJdk,
+            sources,
             classpath,
-            isExcludedPackage
+            isExcludedPackage,
+            userConfig
           ),
           path.toString()
         )
--- a/metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
@@ -95,12 +95,20 @@ object StandaloneSymbolSearch {
       buffers: Buffers,
       sources: Seq[Path],
       classpath: Seq[Path],
-      isExcludedPackage: String => Boolean
+      isExcludedPackage: String => Boolean,
+      userConfig: () => UserConfiguration
   ): StandaloneSymbolSearch = {
+    val (sourcesWithExtras, classpathWithExtras) =
+      addScalaAndJava(
+        sources.map(AbsolutePath(_)),
+        classpath.map(AbsolutePath(_)),
+        userConfig().javaHome
+      )
+
     new StandaloneSymbolSearch(
       workspace,
-      classpath.map(path => AbsolutePath(path)),
-      sources.map(path => AbsolutePath(path)),
+      classpathWithExtras,
+      sourcesWithExtras,
       buffers,
       isExcludedPackage
     )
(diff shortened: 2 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -429,9 +429,16 @@ class Compilers(
     }
 
     created.getOrElse {
-      val sourcesWithJdk = JdkSources(userConfig().javaHome) match {
-        case Some(absPath) => absPath.toNIO :: sources
-        case None => sources
+      val sourcesWithScala =
+        if (sources.filter(_.toString.contains("scala-library")).isEmpty)
```
tgodzik, line 433: Could we move the logic to `StandaloneSymbolSearch`? We already do download scala library there and it would be good to make sure that both `apply` work correctly.

tgodzik (changes requested):
Thanks for the fix! I just left one comment that could help us in the future I think.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
ckipp01: Sure, I can move it. Do you think I should do the same for the JDK sources as well then?

tgodzik: Makes sense! Just in case someone creates the StandaloneSymbolSearch in a different place we will not need to fix it all again.

ckipp01: Sounds good! Moved both.

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
```diff
@@ -95,36 +95,87 @@ object StandaloneSymbolSearch {
...
+  private def addScalaAndJava(
+      sources: Seq[AbsolutePath],
+      classpath: Seq[AbsolutePath],
+      javaHome: Option[String]
+  ): (Seq[AbsolutePath], Seq[AbsolutePath]) = {
+    val missingScala: Boolean =
+      sources.filter(_.toString.contains("scala-library")).isEmpty
+
+    (missingScala, JdkSources(javaHome)) match {
+      case (true, None) => getScala(BuildInfo.scalaCompilerVersion)
```
tgodzik, line 152: That will be `BuildInfo.scala212` in reality since Metals is compiled with 2.12

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
```diff
@@ -95,36 +95,87 @@ object StandaloneSymbolSearch {
...
+    }
+  }
+
+  /**
+   * Retrieve scala for the given version and partition sources.
+   * @param scalaVersion
+   * @return (scalaSources, scalaClasspath)
+   */
+  private def getScala(
+      scalaVersion: String
```
tgodzik, line 173: `scalaVersion` is actually never used.

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
```diff
@@ -95,36 +95,87 @@ object StandaloneSymbolSearch {
...
+  ): (Seq[AbsolutePath], Seq[AbsolutePath]) =
+    Embedded
+      .downloadScalaSources(BuildInfo.scalaCompilerVersion)
```
tgodzik, line 176:
```suggestion
      .downloadScalaSources(BuildInfo.scala212)
```

tgodzik (changes requested):
Just a few last comments. Great work!

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
ckipp01: Yea, I knew it'd be the same, but I figured I'd add in `scalaCompilerVersion` to not have to change it in the future if we change to 2.13 for example. I can change it back to `BuildInfo.scala212` if  you want?

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
ckipp01: Whoops, added this in the start and forgot to remove it.

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
tgodzik: We can also use the one from mtags. There is already `scala.meta.internal.mtags.BuildInfo` with `scalaCompilerVersion` which will correspond to the metals version.

It's also fine to add an additional field, either way is fine.

ckipp01: Alright, if you're fine with it, I'll leave this as is and still use `scalaCompilerVersion` here since again _hopefully_ soon if we bump to 2.13, we can then default stuff like this to that and not have to go around changing it.

> We can also use the one from mtags.

But we don't have access to the mtags BuildInfo here right?

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
ckipp01: Actually kept this in, but made sure to actually use it.

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
ckipp01: I'm not just using the `scalaVersion` we pass in here.

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
tgodzik: We do have access, since Metals itself is using the 2.12 jar to access all the classes defined in the mtags module.

ckipp01: Huh, I guess I've just never done that before and reached into the BuildInfo of another module. Anyways, I changed this to do that.  It now grabs the `scalaCompilerVersion` from mtags.

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
```diff
@@ -95,36 +95,88 @@ object StandaloneSymbolSearch {
...
+        getScala(scala.meta.internal.mtags.BuildInfo.scalaCompilerVersion)
+      case (true, Some(absPath)) =>
+        val (scalaSources, scalaClasspath) = getScala(
+          scala.meta.internal.mtags.BuildInfo.scalaCompilerVersion
+        )
+        (
+          (scalaSources :+ absPath) ++ sources,
+          classpath ++ scalaClasspath
+        )
+      case (false, None) => (Nil, Nil)
```
tgodzik, line 162:
```suggestion
      case (false, None) => (sources, classpath)
```

otherwise aren't we replacing a potentially full list that contains scala sources with and empty one?

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
```diff
@@ -95,36 +95,88 @@ object StandaloneSymbolSearch {
...
+      case (false, Some(absPath)) =>
+        (sources :+ absPath, Nil)
```
tgodzik, line 164:
```suggestion
        (sources :+ absPath, classpath)
```

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
```diff
@@ -95,36 +95,88 @@ object StandaloneSymbolSearch {
...
+      sources: Seq[AbsolutePath],
+      classpath: Seq[AbsolutePath],
+      javaHome: Option[String]
+  ): (Seq[AbsolutePath], Seq[AbsolutePath]) = {
+    val missingScala: Boolean =
+      sources.filter(_.toString.contains("scala-library")).isEmpty
+
+    (missingScala, JdkSources(javaHome)) match {
+      case (true, None) =>
+        getScala(scala.meta.internal.mtags.BuildInfo.scalaCompilerVersion)
```
tgodzik, line 153:
```suggestion
        val (scalaSources, scalaClasspath) = getScala(scala.meta.internal.mtags.BuildInfo.scalaCompilerVersion)
        (sources ++ scalaSource, classpath ++ scalaClasspath )
```
?

tgodzik (changes requested):
Thanks for the fix, though I think we still might have some problems with the current approach especially if the sources are already available.

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
ckipp01: Ach, yes... Sorry I have some major brain farts in here.

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
ckipp01: Fixed.

metals/src/main/scala/scala/meta/internal/metals/StandaloneSymbolSearch.scala
ckipp01: Fixed.

ckipp01 (commented):
Addressed your points. Thanks for the reviews!

ckipp01:
> Thanks for the fix, though I think we still might have some problems with the current approach especially if the sources are already available.

But if the Scala sources are included we aren't including any extra, so this should be fine no?

tgodzik:
> > Thanks for the fix, though I think we still might have some problems with the current approach especially if the sources are already available.
> 
> But if the Scala sources are included we aren't including any extra, so this should be fine no?

It will be fine, I just meant that we needed to fix the things in comments. Trying to work on too many things at once and might not be clear enough here.

tgodzik (approved):
LGTM!

PR #2290, opened 2020-12-05
Add in ability to export worksheet evaluations.

ckipp01:
~Atm this is just here to show how https://github.com/scalameta/mdoc/pull/427 is meant to be used from the Metals side. I still need to add in some tests here, but I wanted to at least have this up to help with the mdoc review.~

~So when everything is fine and there are no error diagnostics the user will be able to just copy the output of a worksheet with a command like so:~

If there is a diagnostic error, then you won't be able to export

You can see what is needed for a client to implement this [here in coc-metals](https://github.com/scalameta/coc-metals/pull/360/files)

## New Approach
EDIT: The gifs are now out of date as the output are no longer on the same line, but all of them are under. Also, the full implementation of this is now done in Metals, no longer in mdoc.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1620,8 +1620,7 @@ class MetalsLanguageServer(
         val args = params.getArguments.asScala
         val worksheet = args.lift(0).collect {
           case ws: JsonPrimitive if ws.isString =>
-            val a = new URI(ws.getAsString())
-            AbsolutePath.fromAbsoluteUri(a)
+            ws.getAsString().toAbsolutePath
         }
 
         val output = worksheet.flatMap(worksheetProvider.copyWorksheetOutput(_))
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1616,6 +1616,23 @@ class MetalsLanguageServer(
       case ServerCommands.NewScalaProject() =>
         newProjectProvider.createNewProjectFromTemplate().asJavaObject
 
+      case ServerCommands.CopyWorksheetOutput() =>
+        val args = params.getArguments.asScala
+        val worksheet = args.lift(0).collect {
+          case ws: JsonPrimitive if ws.isString =>
+            val a = new URI(ws.getAsString())
+            AbsolutePath.fromAbsoluteUri(a)
+        }
+
+        val output = worksheet.flatMap(worksheetProvider.copyWorksheetOutput(_))
+
+        if (output.nonEmpty) {
+          Future(output).asJavaObject
+        } else {
+          languageClient.showMessage(Messages.Worksheets.unableToExport)
+          Future.successful(()).asJavaObject
+        }
+
       case cmd =>
         scribe.error(s"Unknown command '$cmd'")
         Future.successful(()).asJavaObject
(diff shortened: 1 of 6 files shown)


ckipp01:
You can find the VS Code to test this out here: https://github.com/scalameta/metals-vscode/pull/466

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -138,6 +141,48 @@ class WorksheetProvider(
     publisher.hover(path, position)
   }
 
+  /**
+   * Check to see for a given path if there is an evaluated worksheet that
+   * matches thie exact input.
+   *
+   * @param path to the input used to search previous  evaluations.
```
tgodzik, line 148:
```suggestion
   * @param path to the input used to search previous evaluations.
```

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -274,22 +319,30 @@ class WorksheetProvider(
...
     val relativePath = path.toRelative(workspace)
-    val worksheet = mdoc.evaluateWorksheet(relativePath.toString(), input.value)
-    val classpath = worksheet.classpath().asScala.toList
+    val evaluatedWorksheet =
+      mdoc.evaluateWorksheet(relativePath.toString(), input.value)
+    val classpath = evaluatedWorksheet.classpath().asScala.toList
     val previousDigest = worksheetsDigests.getOrElse(path, "")
     val newDigest = calculateDigest(classpath)
 
+    exportableEvaluations.update(
```
tgodzik, line 328: Should we remove a key if a file was closed?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1616,6 +1616,23 @@ class MetalsLanguageServer(
       case ServerCommands.NewScalaProject() =>
         newProjectProvider.createNewProjectFromTemplate().asJavaObject
 
+      case ServerCommands.CopyWorksheetOutput() =>
+        val args = params.getArguments.asScala
+        val worksheet = args.lift(0).collect {
+          case ws: JsonPrimitive if ws.isString =>
+            val a = new URI(ws.getAsString())
+            AbsolutePath.fromAbsoluteUri(a)
```
tgodzik, line 1624: `we.getAsString.toAbsolutePath` ?

tgodzik (approved):
Great work! I added some comments, but nothing major, so please feel free to ignore. LGTM

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
ckipp01: My awesome mac keyboard is starting to leave double spaces all over the place... Fixed!

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
ckipp01: I _don't think_ it really matters. Mainly because if you close a worksheet, you won't be able to send in a request to get the evaluations for that URI until it's open again, and that opening would re-evaluate and update this anyways. So I don't think there is much there we need to worry about.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
ckipp01: Ah nice, yes, that's better. Changed.

PR #2292, opened 2020-12-07
Default to Bloop is no choice was made about Build Server.

### LINKED ISSUE #2291: Metals tries to connect to sbt server by default
**Describe the bug**

Metals gets stuck on "Connecting to sbt .." on clean sbt build.

**To Reproduce**

* Clean clone an sbt 1.4.3 build
* Start sbt with `sbt`
* Start Metals with `code .`
* Metals prompts to import the build (with Bloop), but Metals also tries to connect to sbt server

**Expected behavior**

I expected Metals to ignore `.bsp/sbt.json`. By default, I want Metals to only use Bloop to import sbt builds.

**Screenshots**

**Installation:**

- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.9.7

**Additional context**

**Search terms**

BSP, sbt, bloop

ckipp01:
Previously if there was no explicit choice made about the Build Server a
user wanted, then we would default to Bloop _if a `.bloop/`_ existed.
However, I didn't take into account if there was no `.bloop/` available,
but a `.bsp/<build-server>.json` available. Now, no matter what we
default to Bloop if there was no explicit choice made.

Closes #2291

Note, there really isn't tests for the `BspConnector`, and I realize that it can actually be tested in isolation. This fixes the issue right away, but I don't add in tests for the BspConnector. We can either merge this is now for a quick fix and I can add in a suite to start testing the `BspConnector` or we can leave this open for a bit until I add them.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
@@ -31,12 +31,7 @@ class BspConnector(
     statusBar: StatusBar
 )(implicit ec: ExecutionContext) {
 
-  def resolve(): BspResolvedResult = {
-    resolveExplicit().getOrElse {
-      if (buildTools.isBloop) ResolvedBloop
-      else bspServers.resolve()
-    }
-  }
+  def resolve(): BspResolvedResult = resolveExplicit().getOrElse(ResolvedBloop)
 
   private def resolveExplicit(): Option[BspResolvedResult] = {
     tables.buildServers.selectedServer().flatMap { sel =>
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
@@ -71,16 +71,6 @@ class BspConnector(
             .map(Some(_))
         case ResolvedBspOne(details) =>
           bspServers.newServer(workspace, details).map(Some(_))
-        case ResolvedMultiple(_, availableServers) =>
-          val query =
-            Messages.SelectBspServer.request(availableServers, None)
-          for {
-            Some(item) <- client
-              .showMessageRequest(query.params)
-              .asScala
-              .map(item => query.details.get(item.getTitle))
-            conn <- bspServers.newServer(workspace, item)
-          } yield Some(conn)
       }
     }
 
--- a/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
@@ -319,6 +319,7 @@ object Messages {
         params: ShowMessageRequestParams,
         details: Map[String, BspConnectionDetails]
     )
+    // TODO This is currently only being used in tests. This can probably be removed
     def message: String =
       "Multiple build servers detected, which one do you want to use?"
     def isSelectBspServer(params: ShowMessageRequestParams): Boolean =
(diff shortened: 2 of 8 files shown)


tgodzik (approved):
LGTM! I think that scenario is one that we never really thought of.

ckipp01:
Actually, I need to make some changes in the `BillSuite` for this as well. Since this does change a bit of stuff there as well. I'll look at this later today.

olafurpg (approved):
Thank you for the quick fix!

ckipp01:
So this changes things a bit since we go away from the ability to have true _BSP discovery_ like we did follow before. Now, even if there is a `.bsp/<build-server>.json` file, it doesn't matter unless it was explicitly chosen to be used before You'll see this change reflected in the `BillLspSuite` and also in places where I used the `ResolvedMultiple` before. The only time you'll really see a "Resolved Multiple" would be when you are executing a `bsp-switch` and there are multiple other servers found. To illustrate this, I have a recording of me in a project that has a Mill and an sbt build definition, and I use all three as build servers. It might be easier to picture this way.

[asciicast](https://asciinema.org/a/i2SmfKfWFgUmdrIfsRJzLGRxi)

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -76,16 +71,6 @@ class BspConnector(
             .map(Some(_))
         case ResolvedBspOne(details) =>
           bspServers.newServer(workspace, details).map(Some(_))
-        case ResolvedMultiple(_, availableServers) =>
```
ckipp01, line 79: This will now never happen. As I explained, if there is no longer an explicit choice resolved. we default to Bloop, which means it will only ever resolve one when doing the initial connection.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -319,6 +319,7 @@ object Messages {
         params: ShowMessageRequestParams,
         details: Map[String, BspConnectionDetails]
     )
+    // TODO This is currently only being used in tests. This can probably be removed
```
ckipp01, line 322: I'll need to do a bit more work here (in a follow-up pr if that's ok) since this is only being used in tests. I need to remove them, but I also want to replace them with stuff that is actually testing thing we want. I plan to expand the `bsp-switch` suite and also add in a `BspConnector` suite.

tests/unit/src/test/scala/tests/BillLspSuite.scala
```diff
@@ -203,19 +211,4 @@ class BillLspSuite extends BaseLspSuite("bill") {
...
-    testSelectServerDialogue()
-  }
-
-  test("mix") {
-    cleanWorkspace()
-    RecursivelyDelete(globalBsp)
-    Bill.installWorkspace(workspace.toNIO, "Bill")
-    Bill.installGlobal(globalBsp.toNIO, "Bob")
-    testSelectServerDialogue()
-  }
```
ckipp01, line 220: I removed these because again, this will never actually happen anymore. In this situation, it would instead just default to Bloop.

ckipp01 (commented):
This should fix the issue, _however_ there is one edge case that would screw this up. _If_ a user is using a Scala build tool that we don't offer a `BloopInstall` for, but it's also a Build Server. In that scenario, it might be the _only_ time we'd want it to actually skip the default to Bloop and to then actually follow BSP discovery and grap whatever is in the `.bsp/<build-server>.json` and connect to it automatically. Do you see that happening or a valid concern? If so, I can build in that functionality that it checks to see what the build tool for the workspace is first, checks to see if it's a `BloopInstall` build tool, and if no then auto-connect. However, that does complicate the logic a bit. Let me know what you think.

ckipp01:
These changes aren't all really needed. I have a much simpler solution in https://github.com/scalameta/metals/pull/2297.

PR #2294, opened 2020-12-07
Change lampepfl/dotty.g8 to lampepfl/scala3.g8

tgodzik:
The template got recently updated.

PR #2295, opened 2020-12-07
Scala 3 template was moved from lampepfl org

tgodzik:
Follow up from https://github.com/scalameta/metals/pull/2294

It changed again and I haven't really noticed :sweat:

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/NewProjectProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/NewProjectProvider.scala
@@ -342,8 +342,8 @@ object NewProjectProvider {
         description = "Scala Native"
       ),
       MetalsQuickPickItem(
-        id = "lampepfl/scala3.g8",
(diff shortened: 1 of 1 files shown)


ckipp01 (approved):
lol teaches us to change things without actually testing that they work 😬 LGTM!

ckipp01:
Just tested this out, and it works aok. I'm going to go ahead and merge it right away.

kpbochenek:
@ckipp01 even if you tested that it would simply work. When you move project from one workspace to another github actually creates a symlink in previous place to link to a new one.
You can try create new project even now from `lampepfl/scala3.g8` and it will still work.

ckipp01:
> @ckipp01 even if you tested that it would simply work. When you move project from one workspace to another github actually creates a symlink in previous place to link to a new one.
> You can try create new project even now from `lampepfl/scala3.g8` and it will still work.

ahh true, good point!

PR #2297, opened 2020-12-08
Fix auto-connect BSP server without explicit choice.

### LINKED ISSUE #2291: Metals tries to connect to sbt server by default

ckipp01:
Before this if a `.bsp/sbt.json` file existed and there was no `.bloop/`
yet then it would attempt to auto-connect to sbt. This behavior changes
this slightly now so that the behavior I just explains _only_ happens
if an explicit choice was made before, or if there is a `.bsp/<build-server>.json`
entry that is a Build Tool that Metals doesn't have Bloop support for.

This will supersede https://github.com/scalameta/metals/pull/2292 and closes #2291 

As always, my brain tried to over-complicate things when there was a simple solution. 😬

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
@@ -33,7 +33,8 @@ class BspConnector(
 
   def resolve(): BspResolvedResult = {
     resolveExplicit().getOrElse {
-      if (buildTools.isBloop) ResolvedBloop
+      if (buildTools.loadSupported().nonEmpty || buildTools.isBloop)
+        ResolvedBloop
       else bspServers.resolve()
     }
   }


metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -33,7 +33,7 @@ class BspConnector(
 
   def resolve(): BspResolvedResult = {
     resolveExplicit().getOrElse {
-      if (buildTools.isBloop) ResolvedBloop
+      if (buildTools.loadSupported().nonEmpty) ResolvedBloop
```
ckipp01, line 36: `loadSupported` in builTools does pretty much exactly what we want. It checks if it's an sbt workspace, a gradle workspace, a maven workspace, or a mill workspace. Since we can handle all of those with Bloop, we then default to Bloop for all of them if there is no explicit choice made.

tgodzik (approved):
LGTM! I love when a small PR fixes an issue :+1:

terpstra:
This change is horrible. It broke all of the new workspaces created at my company.

We now have to hand-hold developers through the process of opening the command palette and switching to the BSP implementation needed for our repositories. The BSP selection is hard to find and to a user who does not understand the BSP or how the scala-metals works, it's totally not obvious what is wrong; it just fails mysteriously now.

If a project has a .bsp/xyz.json, you should NOT USE BLOOP. The project essentially told you what is the right BSP to use for that project: xyz. scala-metals no reason to think bloop would work on that project.

In our scenario, we use neither bloop nor sbt. Any attempt to use a tool other than the BSP implementation referenced in .bsp/xyz.json will fail. The default BSP for a freshly opened workspace should definitely be the BSP referenced by that project.

ckipp01:
> This change is horrible. It broke all of the new workspaces created at my company.

While we appreciate your feedback, I'll ask you to check your tone.

I'll address your concern in your bug report.

terpstra:
Yeah, sorry about that. I was just super frustrated when I finally realized what had happened.

Issue #2306, opened 2020-12-13
Misleading warning when wrong organizeImports has an incorrect setting


ckipp01:
**Describe the bug**

While using the built-in Organize Imports actions with the following rules:

```
OrganizeImports {
  groups = [
    "javax?\\.",
    "scala.",
    "*"
  ]
  removeUnused = true
  groupedImports = AggressiveMerge
}
```

With the current version of Metals, 0.9.7, if you attempt to ogranize your imports you'll just get the following warning:

```
 Organize import action not found.
```

However, this isn't fully true, and when you look in the logs you see the real issue:

```
SEVERE: Internal error: scala.meta.internal.metals.ScalafixProvider$ScalafixRunException: Unknown input 'AggressiveMerge'. Expected one of: Merge, Explode, Keep.
java.util.concurrent.CompletionException: scala.meta.internal.metals.ScalafixProvider$ScalafixRunException: Unknown input 'AggressiveMerge'. Expected one of: Merge, Explode, Keep.
        at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292)
        at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:308)
        at java.util.concurrent.CompletableFuture.uniAccept(CompletableFuture.java:661)
        at java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:646)
        at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:488)
        at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1990)
        at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1(CancelTokens.scala:40)
        at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1$adapted(CancelTokens.scala:38)
        at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
Caused by: scala.meta.internal.metals.ScalafixProvider$ScalafixRunException: Unknown input 'AggressiveMerge'. Expected one of: Merge, Explode, Keep.
[...]

PR #2307, opened 2020-12-13
Bump version of organizeImportRule to latest

ckipp01:
Without this being the latest users may be using certain new
settings that have been added which will cause a warning that
the rule can't be found, but it's just because the setting
doesn't yet exist with this version.

I found this when I found #2306, but this doesn't fix the the
underlying issue.


gabro (approved):
LGTM!

Issue #2285, opened 2020-12-01
Metals Doctor recommend using "semanticdb:sourceroot"


asakaev:
Metals Doctor recommend:
Add the compiler option "-P:semanticdb:sourceroot:/project" to ensure code navigation works.

It looks like code navigation works, but I am not sure if it a false recommendation.

I have added `Global / semanticdbEnabled := true` to ~/.sbt/1.0/global.sbt
`sbt -Dsbt.semanticdb=true` also has no effect.

- Operating system: macOS 10.15.7
- Editor: Visual Studio Code 1.51.1
- Metals version: v0.9.7
- Metals extension v1.9.9
- sbt 1.4.4 using BSP
- scala 3.0.0-M2

Related:
https://github.com/scalameta/metals/issues/2189
https://github.com/scalameta/metals/issues/2049

ckipp01:
Thanks for the question @asakaev, trying this out I'm unable to get the same recommendation in Doctor. Are you able to tell me what the out of `sbt semanticdbOptions` is for you in this project?

ckipp01:
Following up on this, I was able to take a closer look at this today and there is indeed a problem. From what I can tell, the problem is twofold. First off all, in a fresh 3.0.0-m2 project, `semanticdbEnabled` is never being set to true. So it should be easy enough if the `sbt-metals` plugin to enable that for Scala 3 projects. _After_ that is done, then indeed I also see the same message that you do, and I'm unable to get it to go away. By fiddling a bit I was actually unable to get that value added correctly in the plugin for some reason, even though it seemed fairly straightforward. We'll have to take a closer look.

PR #2326, opened 2020-12-15
Ensure semanticdb is enabled with sbt BSP and Scala 3.

ckipp01:
This addresses the first part of #2285, but there is still more work
to be done to get the sourceroot issue figured out.


tgodzik (approved):
LGTM!

PR #2327, opened 2020-12-15
Update scalameta to 4.4.2 and fix folding range suite

tgodzik:
Turns out that last time this was really a regression and also popped up as an error in scalafmt. This PR reverts it to the state before 4.4.0.


ckipp01 (approved):
Looks good! LGTM!

PR #2330, opened 2020-12-17
Add support for Scala 3.0.0-M3

tgodzik:
In order not to deprecate 0.27.0-RC1 and 3.0.0-M1 we needed to use an older munit version for those, as we needed to drop support for them in munit. There was also small needed fix, since typeRef was no longer available, but paramRef does invoke it underneath. 

Another issues that we faced was the fact the mtags no longer compiled on the 0.4.5 dotty plugin version and we needed to update both it and sbt, which caused some tests to fail and I ignored them for now until we can fix it in Bloop.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -189,15 +189,15 @@ lazy val V = new {
   val scalameta = "4.4.2"
   val semanticdb = scalameta
   val bsp = "2.0.0-M13"
-  val bloop = "1.4.5-28-e52cd3ad"
-  val scala3 = "3.0.0-M2"
+  val bloop = "1.4.6-15-209c2a5c"
+  val scala3 = "3.0.0-M3"
   val bloopNightly = bloop
   val sbtBloop = bloop
   val gradleBloop = bloop
   val mavenBloop = bloop
-  val mdoc = "2.2.13"
+  val mdoc = "2.2.14"
   val scalafmt = "2.7.4"
-  val munit = "0.7.19"
+  val munit = "0.7.20"
   val scalafix = "0.9.24"
   val lsp4jV = "0.10.0"
   // List of supported Scala versions in SemanticDB. Needs to be manually updated
@@ -217,8 +217,9 @@ lazy val V = new {
   def scala2Versions = nonDeprecatedScala2Versions ++ deprecatedScala2Versions
 
   // Scala 3
-  def nonDeprecatedScala3Versions = Seq(scala3, "3.0.0-M1", "0.27.0-RC1")
-  def deprecatedScala3Versions = Seq("0.26.0")
+  def nonDeprecatedScala3Versions =
+    Seq(scala3, "3.0.0-M2", "3.0.0-M1", "0.27.0-RC1")
+  def deprecatedScala3Versions = Seq()
   def scala3Versions = nonDeprecatedScala3Versions ++ deprecatedScala3Versions
 
   def supportedScalaVersions = scala2Versions ++ scala3Versions
@@ -305,7 +306,7 @@ val mtagsSettings = List(
     baseDirectory.in(ThisBuild).value / "mtags",
     scalaVersion.value
   ),
-  // @note needed to deal with issues in https://github.com/scalameta/metals/pull/2157
+  // @note needed to deal with issues with dottyDoc
   sources in (Compile, doc) := Seq.empty,
   libraryDependencies ++= crossSetting(
     scalaVersion.value,
@@ -581,7 +582,13 @@ lazy val mtest = project
     testSettings,
     sharedSettings,
     libraryDependencies ++= List(
-      "org.scalameta" %% "munit" % V.munit,
+      // munit had to drop support for 3.0.0-M1 and 0.27.0-RC1
+      if (
+        scalaVersion.value == "0.27.0-RC1" || scalaVersion.value == "3.0.0-M1"
+      )
+        "org.scalameta" %% "munit" % "0.7.19"
+      else
+        "org.scalameta" %% "munit" % V.munit,
       "io.get-coursier" % "interface" % V.coursierInterfaces
     ),
     buildInfoPackage := "tests",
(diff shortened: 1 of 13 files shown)


tests/slow/src/test/scala/tests/sbt/SbtBloopLspSuite.scala
```diff
@@ -439,7 +439,8 @@ class SbtBloopLspSuite
     } yield ()
   }
 
-  test("definition-deps") {
+  // TODO latest bloop-sbt plugin does not properly export sources
```
tgodzik, line 442: I am looking currently at the sbt-bloop issue. I will create anyway an issue to track this.

ckipp01:
> In order not to deprecate 0.27.0-RC1 and 3.0.0-M1 we needed to use an older munit version for those,

Are we also able to just drop 0.27? I've been seeing a lot of other libraries doing that. Or is it needed for the course still?

tgodzik:
> > In order not to deprecate 0.27.0-RC1 and 3.0.0-M1 we needed to use an older munit version for those,
> 
> Are we also able to just drop 0.27? I've been seeing a lot of other libraries doing that. Or is it needed for the course still?

Apparently they might still work on it in January, so I will leave it for now and avoid any confusion.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -119,7 +119,7 @@ case class ScalaPresentationCompiler(
       val sourceFile = CompilerInterfaces.toSource(params.uri, params.text)
       driver.run(uri, sourceFile)
 
-      given ctx as Context = driver.currentCtx
+      val ctx = driver.currentCtx
```
ckipp01, line 122: 🤔 This isn't 100% clear to me. How come this doesn't need to be a `given` anymore?

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.4.2
+sbt.version=1.4.5
```
ckipp01, line 1: Do we really want to do this with the test issues?

ckipp01: Ah... I suppose we have to if we want to bump sbt-dotty.

tests/slow/src/test/scala/tests/sbt/SbtBloopLspSuite.scala
ckipp01: Sounds good. I have create this in sbt as well https://github.com/sbt/sbt/issues/6203

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
tgodzik: This is a workaround not to use givens, because they changed the syntax and I would have duplicate the whole file. By doing it all explicitely it seems to work fine.

project/build.properties
tgodzik: Yeah, I need to finally look into the issue here

build.sbt
```diff
@@ -132,8 +132,8 @@ inThisBuild(
     resolvers += Resolver.sonatypeRepo("public"),
     dependencyOverrides += V.guava,
     // faster publishLocal:
-    publishArtifact.in(packageDoc) := sys.env.contains("CI"),
-    publishArtifact.in(packageSrc) := sys.env.contains("CI"),
+    publishArtifact.in(Compile / packageDoc) := sys.env.contains("CI"),
```
tgodzik, line 135: Funily enough, without this change we would try to publish Tests and it would fail currently.

mtags/src/main/scala-3/scala/meta/internal/pc/Scala3CompilerWrapper.scala
```diff
@@ -12,8 +12,8 @@ class Scala3CompilerWrapper(driver: InteractiveDriver)
   override def compiler(): InteractiveDriver = driver
 
   override def resetReporter(): Unit = {
-    given ctx as Context = driver.currentCtx
-    ctx.reporter.removeBufferedMessages
+    val ctx = driver.currentCtx
+    ctx.reporter.removeBufferedMessages(using ctx)
```
gabro, line 16: Out of sheer curiosity, was this change needed or is it just a style preference?

tgodzik: It's so that we can cross compile 3.0.0-M2 and 3.0.0-M3 - given syntax changed between those two versions.

gabro: Aha, smart!

tgodzik:
We need to wait for https://github.com/scalacenter/bloop/pull/1432, since -Ysemanticdb no longer works due to a bug reported here https://github.com/lampepfl/dotty/issues/10844 :/

romanowski:
sbt-dotty `1.5.1` has been just released that should fix the problem with publish in `mtags/test`

tgodzik:
> sbt-dotty `1.5.1` has been just released that should fix the problem with publish in `mtags/test`

Cool! Thanks! I need to fix it in Bloop still unfortunately.

tests/unit/src/main/scala/tests/QuickBuild.scala
```diff
@@ -118,6 +118,7 @@ case class QuickBuild(
           s"org.scala-lang:scala-reflect:$scalaVersion"
         )
       }
+    println(scalaDependencies.toList)
```
ckipp01, line 121: 😬 forgotten println

ckipp01 (approved):
Other than the random pritnln, LGTM! Great job hunting down all these issues and getting them either fixed or figuring out workarounds for all of them 🚀

tests/unit/src/main/scala/tests/QuickBuild.scala
tgodzik: Ach, damn! Fixing it now!

gabro (approved):
LGTM!

PR #2331, opened 2020-12-18
Bump Ammonite to the 2.13.4 published version.

### LINKED ISSUE #2248: Update ammonite for 2.13.4
Currently we test Ammonite  support on 2.13.3, but as soon as it's available for 2.13.4, we should switch.

ckipp01:
This will close #2248

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -233,7 +233,7 @@ lazy val V = new {
   def dap4j = "org.eclipse.lsp4j" % "org.eclipse.lsp4j.debug" % lsp4jV
   val coursierInterfaces = "1.0.2"
   val coursier = "2.0.7"
-  val ammonite = "2.3.8"
+  val ammonite = "2.3.8-4-88785969"
   val mill = "0.9.3"
   val organizeImportRule = "0.4.4"
 }
@@ -571,8 +571,7 @@ def publishBinaryMtags =
     .in(interfaces)
     .dependsOn(
       publishAllMtags(
-        // TODO https://github.com/scalameta/metals/issues/2248 remove 2.13.3 once resolved
-        List(V.scala211, V.sbtScala, V.scala212, "2.13.3", V.scala213, V.scala3)
+        List(V.scala211, V.sbtScala, V.scala212, V.scala213, V.scala3)
       )
     )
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -634,8 +633,6 @@ lazy val unit = project
       "ch.epfl.scala" %% "bloop-config" % V.bloop,
       "org.scalameta" %% "munit" % V.munit,
       // The dependencies listed below are only listed so Scala Steward
-      // will pick them up and update them. They aren't actually used.
-      "com.lihaoyi" %% "ammonite-util" % V.ammonite intransitive (),
```
ckipp01, line 638: I just removed this... util is no longer published for 12, so this won't work with the new version. We also can't put the repl-api here which is what would be accurate because it needs to be exact. I was annoyed a bit by this mind gymnastics of figuring out the right thing to put here, so I think we should just remove it. Each month or so if we do a new release we always look to see if the new Ammonite is published for that version anyways, so we can just update it that way.

ckipp01:
:/ actually now, the repl also seems to not be published for 2.12. need to take a closer look

ckipp01:
So it looks like the publish may have failed for 2.12.12.... https://github.com/lihaoyi/Ammonite/runs/1570558987#step:5:4696

ckipp01:
The Ammonite release job was fixed and now we have the desired versions to bump this.

build.sbt
```diff
@@ -641,7 +640,6 @@ lazy val unit = project
       "ch.epfl.scala" %% "bloop-config" % V.bloop,
       "org.scalameta" %% "munit" % V.munit,
       // The dependencies listed below are only listed so Scala Steward
-      // will pick them up and update them. They aren't actually used.
```
tgodzik, line 644: I think this might have been removed by mistake?

tgodzik (changes requested):

tgodzik:
Thanks for looking into it @ckipp01 !

build.sbt
ckipp01: Ah, yes, fixed now.

tgodzik (approved):
LGTM!

PR #2333, opened 2020-12-18
Add in better message for user if trying to fetch invalid Ammonite.

ckipp01:
I came across this today in an Ammonite script. I wanted to try
something out with 2.13 and I without really thinking put the following
at the top of my Ammonite script.

```scala
// scala 2.13.4
```

And it just showed me:
```
Error importing /Users/ckipp/Documents/scala-workspace/worksheets/tester.sc. See
 the logs for more details.
```

When in reality, we have a much better message to show the user, since
we know the reason it failed is that the version of Ammonite we are
using isn't available for that Scala Version. So this change better
captures that exception, and actually shows it to the user. So now in
this same situation you'd instead see:

```
Error fetching Ammonite 2.3.8 for scala 2.13.4
```

Part of this is also because I'm seeing a pattern in the survey results
so far of people having something go wrong, and not fully understanding
what the issue is. So hopefully small changes like this will help with
that, and moving forward I'd like to do a better job at showing info the
users that we have and know is either more helpful or actionable then a
generic error message.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ammonite/Ammonite.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ammonite/Ammonite.scala
@@ -22,6 +22,7 @@ import scala.meta.internal.bsp.BuildChange
 import scala.meta.internal.builds.BuildTools
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals._
+import scala.meta.internal.metals.ammonite.Ammonite.AmmoniteMetalsException
 import scala.meta.io.AbsolutePath
 
 import ammrunner.AmmoniteFetcher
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/ammonite/Ammonite.scala
```diff
@@ -239,35 +240,41 @@ final class Ammonite(
           if (docs.isEmpty) "No Ammonite script is opened"
           else "No open document is not an Ammonite script"
         scribe.error(msg)
-        Future.failed(new Exception(msg))
+        Future.failed(new AmmoniteMetalsException(msg))
```
ckipp01, line 243: Also, previously we created this `msg` here, bu this would never be shown to the user. Now if this ever happens, it is caught in the `recoverWith` down below which was added.

tgodzik (approved):
LGTM!

ckipp01:
An unrelated test failure:

```
[error] 	tests.pc.CompletionOverrideConfigSuite
```
So I'll go ahead and merge.

PR #2334, opened 2020-12-19
Update Deprecated Usage Of `setDeprecated` from `lsp4j`

isomarcte:
In a very amusing situation, `setDeprecated` from `org.eclipse.lsp4j.CompletionItem` is now itself deprecated.

This commit updates the usage to the new form, using `setTags` to add a `deprecated` tag on the `CompletionItem`.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -451,7 +451,7 @@ case class ScalaPresentationCompiler(
     }
 
     if (completion.symbols.forall(_.isDeprecated)) {
-      item.setTags((Option(item.getTags).fold(Set.empty[CompletionItemTag])(_.asScala.toSet) ++ Set(CompletionItemTag.Deprecated)).toList.asJava)
+      item.setTags(List(CompletionItemTag.Deprecated).asJava)
     }
 
     completion.symbols.headOption
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -451,7 +451,7 @@ case class ScalaPresentationCompiler(
     }
 
     if (completion.symbols.forall(_.isDeprecated)) {
-      item.setTags((item.getTags.asScala.toSet ++ Set(CompletionItemTag.Deprecated)).toList.asJava)
+      item.setTags((Option(item.getTags).fold(Set.empty[CompletionItemTag])(_.asScala.toSet) ++ Set(CompletionItemTag.Deprecated)).toList.asJava)
     }
 
     completion.symbols.headOption
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -83,8 +83,6 @@ case class ScalaPresentationCompiler(
     workspace: Option[Path] = None
 ) extends PresentationCompiler {
 
-  import ScalaPresentationCompiler._
-
   def this() = this(Nil, Nil)
 
   import InteractiveDriver._
@@ -448,13 +446,10 @@ case class ScalaPresentationCompiler(
       doc <- ParsedComment.docOf(sym)
     } yield doc
 
-    if (documentation.nonEmpty) {
-      item.setDocumentation(hoverContent(None, None, documentation))
+    if (completion.symbols.forall(_.isDeprecated)) {
+      item.setTags((item.getTags.asScala.toSet ++ Set(CompletionItemTag.Deprecated)).toList.asJava)
     }
 
-    if (completion.symbols.forall(_.isDeprecated)) {
-      setCompletionItemDeprecated(item)
-    }
     completion.symbols.headOption
       .foreach(s => item.setKind(completionItemKind(s)))
     item
@@ -545,13 +540,4 @@ case class ScalaPresentationCompiler(
       )
     }
   }
-}
-
-object ScalaPresentationCompiler {
-  private[this] val deprecatedTagList: List[CompletionItemTag] =
-    List(CompletionItemTag.Deprecated)
-
-  /** Add the `Deprecated` tag to the completion item. */
-  private def setCompletionItemDeprecated(value: CompletionItem): Unit =
-    value.setTags(deprecatedTagList.asJava)
 }
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -20,6 +20,7 @@ import org.eclipse.lsp4j.Diagnostic
 import org.eclipse.lsp4j.Location
 import org.eclipse.lsp4j.CompletionItem
 import org.eclipse.lsp4j.CompletionItemKind
+import org.eclipse.lsp4j.CompletionItemTag
 import org.eclipse.lsp4j.CompletionList
 import org.eclipse.lsp4j.Position
 import org.eclipse.lsp4j.Range
@@ -81,6 +82,8 @@ case class ScalaPresentationCompiler(
     config: PresentationCompilerConfig = PresentationCompilerConfigImpl(),
     workspace: Option[Path] = None
 ) extends PresentationCompiler {
+
+  import ScalaPresentationCompiler._
 
   def this() = this(Nil, Nil)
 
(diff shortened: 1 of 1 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -541,3 +546,12 @@ case class ScalaPresentationCompiler(
...
   }
 }
+
+object ScalaPresentationCompiler {
+  private[this] val deprecatedTagList: List[CompletionItemTag] =
+    List(CompletionItemTag.Deprecated)
+
+  /** Add the `Deprecated` tag to the completion item. */
+  private def setCompletionItemDeprecated(value: CompletionItem): Unit =
+    value.setTags(deprecatedTagList.asJava)
```
ckipp01, line 556: A couple thoughts on this. We also use tags to tag deprecated in the `CompletionProvider` as we can see here https://github.com/scalameta/metals/blob/06f59123c61eb6124f6cb98762de3748d4e7a04e/mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala#L227-L229.

So if we were to stick with having a utility like this, it might be best to use it both places. On the other hand, if we ever decide to use more tags for whatever reason, this sort of falls apart since this doesn't append, just replaces. So I guess I'd prefer one of the following:

- Change this to instead of just setting tags to deprecated, take in ca completion item, check if there are any other tags, concat and set them, and then also use it here and in the `CompletionProvider`
- Simply not worry about having this here now, and inline it like we do in `CompletionProvider`. This is the easier option, and probably the route I'd go.

What do you think?

ckipp01 (changes requested):
Thanks for the contribution @isomarcte! Let me know what you think!

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
gabro: I'm 👍 on inlining this too. Despite being a little verbose, I don't think it's worth the indirection

isomarcte: I didn't consider that it would be overwriting other tags, that's a valid point. 

It may be worth noting that as of now there is only a single possible tag, `Deprecated`. Though I assume the reason lsp4j is changing this API is that they anticipate adding more soon.

With that in mind, I think something like this would be more appropriate,

```scala
      (item.getTags.asScala.toSet ++ Set(CompletionItemTag.Deprecated)).toList.asJava
```

This assumes that having more than one tag of the same type is semantically the same, e.g. `List(Deprecated)` means the same as `List(Deprecated, Deprecated)`. Whether or not that assumption holds true depends on what other tags lsp4j adds, but I think it is _likely_ it will hold true.

For my personal tastes, that is quite a lot of Java<->Scala back and forth for inlining, but if that is preferred route to go I'll update the PR to that.

isomarcte: Updated

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -445,11 +446,10 @@ case class ScalaPresentationCompiler(
       doc <- ParsedComment.docOf(sym)
     } yield doc
 
-    if (documentation.nonEmpty) {
-      item.setDocumentation(hoverContent(None, None, documentation))
```
gabro, line 449: I may have missed the discussion, but why is this gone?

isomarcte: @gabro _That_ looks like a typo on my part...Sorry about that.

isomarcte: @gabro updated.

gabro (approved):
LGTM, thanks for the contribution!

tgodzik (approved):
LGTM!

ckipp01 (approved):
Looks good! Thanks for taking care of this!

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -449,7 +450,10 @@ case class ScalaPresentationCompiler(
       item.setDocumentation(hoverContent(None, None, documentation))
     }
 
-    item.setDeprecated(completion.symbols.forall(_.isDeprecated))
+    if (completion.symbols.forall(_.isDeprecated)) {
+      item.setTags((item.getTags.asScala.toSet ++ Set(CompletionItemTag.Deprecated)).toList.asJava)
```
tgodzik, line 454: I think `tags` can be null here and that is failing in the tests. Weirdly enough it causes the whole suite to hang, which should not happen, especially that each tests is max 10 minutes :thinking: 

Since we create the item in this method we can just do:
```scala
  item.setTags(List(CompletionItemTag.Deprecated).asJava) 
```

tgodzik (changes requested):
Thanks again for the contribution! Test failures are actually correct:

```
[info] tests.pc.CompletionSuite.adt1_3.0.0-M1 started
Exception in thread "pool-36-thread-1" java.lang.NullPointerException
	at scala.meta.internal.pc.ScalaPresentationCompiler.completionItem(ScalaPresentationCompiler.scala:454)
	at scala.meta.internal.pc.ScalaPresentationCompiler.complete$$anonfun$2$$anonfun$1(ScalaPresentationCompiler.scala:132)
	at scala.collection.immutable.List.map(List.scala:250)
	at scala.meta.internal.pc.ScalaPresentationCompiler.complete$$anonfun$1(ScalaPresentationCompiler.scala:132)
	at scala.meta.internal.pc.CompilerAccess.withSharedCompiler(CompilerAccess.scala:137)
	at scala.meta.internal.pc.CompilerAccess.$anonfun$1(CompilerAccess.scala:87)
	at scala.meta.internal.pc.CompilerAccess.onCompilerJobQueue$$anonfun$1(CompilerAccess.scala:197)
	at dotty.runtime.function.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
	at scala.meta.internal.pc.CompilerJobQueue$Job.run(CompilerJobQueue.scala:103)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

Let's maybe do it the same way as in Scala 2.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
isomarcte: @tgodzik would you be opposed to this?

```
item.setTags((Option(item.getTags).map(_.asScala.toSet).getOrElse(Set.empty) ++ Set(CompletionItemTag.Deprecated)).toList.asJava)
```

I think `  item.setTags(List(CompletionItemTag.Deprecated).asJava) ` is fine for today, but since I'm assuming lsp4j is changing this API to allow for more types of tags, I can foresee a future bug where we set the tags in one place, then overwrite them in another.

isomarcte: @tgodzik I've updated to check for `null`. Let me know if you strongly prefer

```
 item.setTags(List(CompletionItemTag.Deprecated).asJava) 
 ```
 
 If so, I'll change it to that.

Comment (tgodzik) [reply mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala]: Item is really created in the same method just above, so I don't think we benefit much by doing so. If anyone wants to move it at some point they will have to take a look at what's going on anyway.

I think it's really impossible to future-proof it and this might actually never change.

File: mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala [hunk 4]
```diff
@@ -449,7 +450,10 @@ case class ScalaPresentationCompiler(
       item.setDocumentation(hoverContent(None, None, documentation))
     }
 
-    item.setDeprecated(completion.symbols.forall(_.isDeprecated))
+    if (completion.symbols.forall(_.isDeprecated)) {
+      item.setTags((Option(item.getTags).fold(Set.empty[CompletionItemTag])(_.asScala.toSet) ++ Set(CompletionItemTag.Deprecated)).toList.asJava)
```
tgodzik, line 454: Honestly, we would be sacrificing readability for the sake of possible small issue in the future. It looks a bit scary really.

isomarcte: This is why I initially had them in the companion object.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
isomarcte: I'll update it to the overwriting version you suggested.

isomarcte:
@tgodzik updated

tgodzik (approved):
LGTM! Thanks for the change!

isomarcte:
@tgodzik I see one failure in CI, but I don't _think_ it is related to this PR.

tgodzik:
> @tgodzik I see one failure in CI, but I don't _think_ it is related to this PR.

It doesn't seem to be, looks like flaky failure. Thanks again for the fix!

Issue #2167, opened 2020-10-29
Emacs Metals complains about using scala 2.10.7 when the project is not using it


ashkan-leo:
**Describe the bug**
I have a Scala 2.11.12 project. I use Metals with Emacs to do my work. Recently, I updated Emacs and Metals. When I do build-import, I don't have any auto completion and the log says "Unsupported Scala 2.10.7". The lsp-metals-doctor-run shows the following:

```
Metals Doctor

Build server currently used: Bloop.

These are the installed build targets for this workspace. One build target corresponds to one classpath. For example, normally one sbt project maps to two build targets: main and test.

  Build target  Scala  Diagnostics  Goto definition  Completions  Find references  Recommendation    
  root  2.11.12  ✅   ✅   ✅   ✅   This Scala version might not be supported in upcoming versions of Metals, please upgrade to Scala 2.12.12.    
  root-test  2.11.12  ✅   ✅   ✅   ✅   This Scala version might not be supported in upcoming versions of Metals, please upgrade to Scala 2.12.12.    
  meng-build  2.10.7  ✅   ⚠️   ⚠️   ⚠️   Code navigation is not supported for this compiler version, change to Scala 2.13.3 or 2.12.12 and run 'Build import' to enable code navigation.    
  meng-build-build  2.10.7  ✅   ⚠️   ⚠️   ⚠️   Code navigation is not supported for this compiler version, change to Scala 2.13.3 or 2.12.12 and run 'Build import' to enable code navigation.    

```

**To Reproduce**

1. Open a Scala project in Emacs.
2. do lsp-metals-build-import
3. do lsp-metals-doctor-run
4. observe the additional projects that use Scala 2.10 are added to the list.

**Expected behavior**
Metals should infer the versions correctly.

**Screenshots**

**Installation:**

- Operating system: macOS
- Editor: Emacs
- Metals version: v0.9.4 and 0.9.4+104-85ff9dc0-SNAPSHOT

**Additional context**

**Search terms**

ckipp01:
[...]

PR #2339, opened 2020-12-23
Refactor doctor and make sure issues with sbt are properly reported

tgodzik:
The main goal of this refactor is to make sure we have just one category of problems that will be reported both in the message request and the doctor. Previously, both would need to be updated independently and testing for those was scarce. Thanks to having a single category of problems we can now properly unit test them to make sure every message is sensible. It's also easier now to see what each problem is caused by and what message will be displayed to the user.

Also, this should fix https://github.com/scalameta/metals/issues/2167 - sbt issues will now be reported separately, so that the users are not confused.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
@@ -4,19 +4,18 @@ import java.net.URLEncoder
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
 
-import scala.collection.mutable.ListBuffer
 import scala.concurrent.ExecutionContext
 
 import scala.meta.internal.bsp.BspResolvedResult
+import scala.meta.internal.bsp.BspSession
 import scala.meta.internal.bsp.ResolvedBloop
 import scala.meta.internal.bsp.ResolvedBspOne
 import scala.meta.internal.bsp.ResolvedMultiple
 import scala.meta.internal.bsp.ResolvedNone
 import scala.meta.internal.metals.Messages.CheckDoctor
 import scala.meta.internal.metals.MetalsEnrichments._
-import scala.meta.internal.metals.ScalaVersions._
 import scala.meta.internal.metals.config.DoctorFormat
-import scala.meta.internal.semver.SemVer
+import scala.meta.internal.troubleshoot.ProblemResolver
 import scala.meta.io.AbsolutePath
 
 import org.eclipse.lsp4j.ExecuteCommandParams
@@ -30,22 +29,19 @@ final class Doctor(
     workspace: AbsolutePath,
     buildTargets: BuildTargets,
     languageClient: MetalsLanguageClient,
-    currentBuildServer: () => Option[String],
+    currentBuildServer: () => Option[BspSession],
     calculateNewBuildServer: () => BspResolvedResult,
     httpServer: () => Option[MetalsHttpServer],
     tables: Tables,
     clientConfig: ClientConfiguration
 )(implicit ec: ExecutionContext) {
   private val hasProblems = new AtomicBoolean(false)
-  private var bspServerName: Option[String] = None
-  private var bspServerVersion: Option[String] = None
-
-  def isUnsupportedBloopVersion(serverVersion: String): Boolean = {
-    bspServerName.contains("Bloop") && !SemVer.isCompatibleVersion(
-      BuildInfo.bloopVersion,
-      serverVersion
+  private val problemResolver =
+    new ProblemResolver(
+      workspace,
+      currentBuildServer,
+      clientConfig.isCommandInHtmlSupported()
     )
-  }
 
   /**
    * Returns a full HTML page for the HTTP client.
@@ -70,6 +66,9 @@ final class Doctor(
       }
     )
   }
+
+  def isUnsupportedBloopVersion(): Boolean =
+    problemResolver.isUnsupportedBloopVersion()
 
   /**
    * Executes the "Reload doctor" server command.
@@ -123,10 +122,8 @@ final class Doctor(
   /**
    * Checks if there are any potential problems and if any, notifies the user.
    */
-  def check(serverName: String, serverVersion: String): Unit = {
-    bspServerName = Some(serverName)
-    bspServerVersion = Some(serverVersion)
-    val summary = problemSummary
+  def check(): Unit = {
+    val summary = problemResolver.problemMessage(allTargets())
     executeReloadDoctor(summary)
     summary match {
       case Some(problem) =>
@@ -146,126 +143,6 @@ final class Doctor(
         }
       case None =>
         () // All OK.
-    }
-  }
-
-  private def recommendation(
-      scalaVersion: String,
-      isSemanticdbEnabled: Boolean,
-      scala: ScalaTarget
-  ): String = {
-    def isMaven: Boolean = workspace.resolve("pom.xml").isFile
-    def hint() =
-      if (isMaven) {
-        val website = clientConfig.doctorFormat match {
-          case DoctorFormat.Json =>
-            "Metals Website - https://scalameta.org/metals/docs/build-tools/maven.html"
-          case DoctorFormat.Html =>
-            "<a href=https://scalameta.org/metals/docs/build-tools/maven.html>Metals website</a>"
-        }
-        "enable SemanticDB following instructions on the " + website
-      } else s"run 'Build import' to enable code navigation."
-
-    if (!isSemanticdbEnabled) {
-      if (bspServerVersion.exists(isUnsupportedBloopVersion)) {
-        s"""|The installed Bloop server version is ${bspServerVersion.get} while Metals requires at least Bloop version ${BuildInfo.bloopVersion},
-            |To fix this problem please update your Bloop server.""".stripMargin
-      } else if (
-        isSupportedScalaVersion(
-          scalaVersion
-        )
-      ) {
-        hint.capitalize
-      } else if (isSupportedScalaBinaryVersion(scalaVersion)) {
-        val recommended = recommendedVersion(scalaVersion)
-        val isRecommenedVersionNewer =
-          SemVer.isCompatibleVersion(scalaVersion, recommended)
-        if (isRecommenedVersionNewer) {
-          s"Upgrade to Scala $recommended and " + hint
-        } else {
-          s"Scala $scalaVersion is not yet supported"
-        }
-
-      } else {
-        val versionToUpgradeTo =
-          if (ScalaVersions.isScala3Version(scalaVersion)) {
-            s"Scala ${BuildInfo.scala3}"
-          } else {
-            s"Scala ${BuildInfo.scala213} or ${BuildInfo.scala212}"
-          }
-        s"Code navigation is not supported for this compiler version, change to " + versionToUpgradeTo + " and " + hint
-      }
-    } else {
-      val messages = ListBuffer.empty[String]
-      if (ScalaVersions.isDeprecatedScalaVersion(scalaVersion)) {
-        messages += s"This Scala version might not be supported in upcoming versions of Metals, " +
-          s"please upgrade to Scala ${recommendedVersion(scalaVersion)}."
-      } else if (!isLatestScalaVersion(scalaVersion)) {
-        messages += s"Upgrade to Scala ${recommendedVersion(scalaVersion)} to enjoy the latest compiler improvements."
-      }
-      if (!scala.isSourcerootDeclared) {
-        messages += s"Add the compiler option ${workspace.sourcerootOption} to ensure code navigation works."
-      }
-      messages.toList match {
-        case Nil => ""
-        case head :: Nil => head
-        case _ =>
-          val html = new HtmlBuilder()
-          html.unorderedList(messages)(html.text).toString
-      }
-    }
-  }
-
-  private def problemSummary: Option[String] = {
-
-    def message(
-        filter: String => Boolean,
-        apply: Iterable[String] => String
-    ): Option[String] = {
-      val versions = (for {
-        target <- allTargets.toIterator
-        if filter(target.scalaVersion)
-      } yield target.scalaVersion).toSet
-
-      if (versions.nonEmpty) {
-        Some(apply(versions))
-      } else {
-        None
-      }
-    }
-
-    message(
-      ScalaVersions.isFutureVersion,
-      Messages.FutureScalaVersion.message
-    ).orElse {
-      message(
-        ver => !ScalaVersions.isSupportedScalaVersion(ver),
-        Messages.UnsupportedScalaVersion.message
-      )
-    }.orElse {
-      possiblyMissingSemanticDB
-    }.orElse {
-      message(
-        ScalaVersions.isDeprecatedScalaVersion,
-        Messages.DeprecatedScalaVersion.message
-      )
-    }
-  }
-
-  private def possiblyMissingSemanticDB: Option[String] = {
-    val targets = allTargets()
-    val isMissingSemanticdb = targets.filter(!_.isSemanticdbEnabled)
-    val count = isMissingSemanticdb.length
-    val isAllProjects = count == targets.size
-    if (isMissingSemanticdb.isEmpty) {
-      None
-    } else if (isAllProjects) {
-      Some(CheckDoctor.allProjectsMisconfigured)
-    } else if (count == 1) {
-      val name = isMissingSemanticdb.head.displayName
-      Some(CheckDoctor.singleMisconfiguredProject(name))
-    } else {
-      Some(CheckDoctor.multipleMisconfiguredProjects(count))
     }
   }
 
@@ -489,8 +366,7 @@ final class Doctor(
       } else {
         Icons.unicode.check
       }
-    val recommenedFix =
-      recommendation(scalaVersion, target.isSemanticdbEnabled, target)
+    val recommenedFix = problemResolver.recommendation(target)
     DoctorTargetInfo(
       target.displayName,
       scalaVersion,
--- a/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
@@ -548,7 +548,7 @@ object Messages {
 
   object DeprecatedScalaVersion {
     def message(
-        usingNow: Iterable[String]
+        usingNow: Set[String]
     ): String = {
       val using = "legacy " + usingString(usingNow)
       val recommended = recommendationString(usingNow)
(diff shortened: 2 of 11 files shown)


metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -39,6 +38,13 @@ final class Doctor(
   private val hasProblems = new AtomicBoolean(false)
   private var bspServerName: Option[String] = None
   private var bspServerVersion: Option[String] = None
+  private val problemResolver =
+    new ProblemResolver(
+      workspace,
+      () => bspServerVersion.exists(isUnsupportedBloopVersion),
+      () => bspServerVersion,
```
ckipp01, line 45: I think the logic is a bit mixed in here and in `ProblemResolver.scala` and it makes it a _bit_ hard to follow.

Do you think it's necessary to actually have `isUnsupportedBloopVersion` _and_ `bloopVersion` be parameters in `ProblemResolver`? What do think of instead just move all of this logic into `ProblemResolver` and handle it all there. We'd move the `isUnsupportedBloopVersion` method right below this into there as well. Then there would be no need to have both of these parameter, just the `BloopVersion`, and then we could do the logic of determining if it's unsupported in the `ProblemResolver`. wdyt?

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
```diff
@@ -0,0 +1,161 @@
...
+import scala.meta.internal.metals.Messages
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.ScalaTarget
+import scala.meta.internal.metals.ScalaVersions
+import scala.meta.io.AbsolutePath
+
+class ProblemResolver(
+    workspace: AbsolutePath,
+    isUnsupportedBloopVersion: () => Boolean,
+    bloopVersion: () => Option[String],
```
ckipp01, line 14: Mentioned this in `Doctor.scala`, but I think it might be cleaner to only have `bloopVersion` here and then move the `isUnsupportedBloopVersion` method from `Doctor.scala` into here and have the logic to determine whether or not it's unsupported in here since a lot of the other logic for finding various problems is in here.

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
```diff
@@ -0,0 +1,75 @@
+package scala.meta.internal.troubleshoot
+
+import scala.meta.internal.metals.BuildInfo
+import scala.meta.internal.metals.ScalaVersions
+
+sealed abstract class ScalaProblem {
```
ckipp01, line 6: More of just a curiosity, but I notice in a lot of Scalameta projects we prefer `abstract class` over `trait`. What's the main reason we do that?

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
```diff
@@ -0,0 +1,75 @@
...
+  def message: String
+
+  protected val hint = "run 'Build import' to enable code navigation."
+
+}
+
+// scala version issues
```
ckipp01, line 13: I think we can probably just remove this comment or include proper scaladocs for these.

Also related, if possible can you add a short Scaladoc to `ScalaProblem`. I find that when all of the classes have short descriptions it makes it much easier for someone new to the codebase to get a quick idea of what a class is meant for then.

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
```diff
@@ -0,0 +1,75 @@
...
+      val recommended = ScalaVersions.recommendedVersion(version)
+      s"Upgrade to Scala $recommended and " + hint
+    } else {
+      val versionToUpgradeTo =
+        if (ScalaVersions.isScala3Version(version)) {
+          s"Scala ${BuildInfo.scala3}"
+        } else {
+          s"Scala ${BuildInfo.scala213} or ${BuildInfo.scala212}"
+        }
+      s"Code navigation is not supported for this compiler version, change to " + versionToUpgradeTo + " and " + hint
```
ckipp01, line 26:
```suggestion
      s"Code navigation is not supported for this compiler version, please change to " + versionToUpgradeTo + " and " + hint
```

ckipp01: Make sure Metals is polite 😄

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
```diff
@@ -0,0 +1,75 @@
...
+}
+
+case class DeprecatedScalaVersion(version: String) extends ScalaProblem {
+  override def message: String =
+    s"Scala $version might not be supported in upcoming versions of Metals, " +
+      s"please upgrade to Scala ${ScalaVersions.recommendedVersion(version)}."
+}
+
+case class FutureScalaVersion(version: String) extends ScalaProblem {
+  override def message: String = s"Scala $version is not yet supported"
```
ckipp01, line 38: Should we also include more info to downgrade to the latest supported version? Similar to what we have in `DeprecatedScalaVersion`?
```suggestion
  override def message: String = s"Scala $version is not yet supported, " +
    s"please downgrade to Scala ${ScalaVersions.recommendedVersion(version)}."
```

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// semanticdb problems
+case class SemanticDBDisabled(
+    scalaVersion: String,
+    bloopVersion: String,
+    isBloopVersionSupported: Boolean
+) extends ScalaProblem {
+  override def message: String = {
+    if (isBloopVersionSupported) {
```
ckipp01, line 48: This confuses me a bit? I'd expect that `isBloopVersionSupported` is true, then we'd have no issue. But this message is saying that it's not supported and Metals requires at least Bloop x. Is this correct?

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
```diff
@@ -0,0 +1,75 @@
...
+      s"""|The installed Bloop server version is $bloopVersion while Metals requires at least Bloop version ${BuildInfo.bloopVersion},
+          |To fix this problem please update your Bloop server.""".stripMargin
+    } else if (ScalaVersions.isSupportedScalaVersion(scalaVersion)) {
+      hint.capitalize
+    } else {
+      "Enable semanticdb plugin for the project."
```
ckipp01, line 54: I'm not 100% sure on the situation on when this will be hit, but I think the more info we can give here the better, since for a lot of people this may not be clear. They might want to why this is being shown. Maybe this will just be combined with other messages, and if so ignore this, but I think having a bit more info, something like "Semanticdb is required for code navigation to work correctly in your project, however the semanticdb plugin doesn't seem to be enabled for your project. Please enable the semanticdb plugin for this project in order for code navigation to work correctly." wdyt?

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
```diff
@@ -0,0 +1,75 @@
...
+}
+case class MissingSourceRoot(sourcerootOption: String) extends ScalaProblem {
+  override def message: String =
+    s"Add the compiler option $sourcerootOption to ensure code navigation works."
+}
+
+// sbt problems
+case object UnsupportedSbtVersion extends ScalaProblem {
+  override def message: String =
+    "Code navigation is not supported for this sbt version, please upgrade to 1.3.2 and reimport the build"
```
ckipp01, line 66:
```suggestion
    "Code navigation is not supported for this sbt version, please upgrade to at least 1.3.2 and reimport the build"
```

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
```diff
@@ -0,0 +1,75 @@
...
+}
+case object DeprecatedSbtVersion extends ScalaProblem {
+  override def message: String =
+    "Code navigation might not be supported in the future for this sbt version, please upgrade to 1.3.2 and reimport the build"
```
ckipp01, line 70:
```suggestion
    "Code navigation might not be supported in the future for this sbt version, please upgrade to at least 1.3.2 and reimport the build"
```

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
```diff
@@ -0,0 +1,161 @@
...
+      issue <- findProblem(target)
+    } yield {
+      issue match {
+        case UnsupportedScalaVersion(version) => unsupportedVersions += version
+        case DeprecatedScalaVersion(version) => deprecatedVersions += version
+        case FutureScalaVersion(version) => futureVersions += version
+        case _: SemanticDBDisabled => misconfiguredProjects += 1
+        case _: MissingSourceRoot => misconfiguredProjects += 1
+        case UnsupportedSbtVersion => unsupportedSbt = true
+        case DeprecatedSbtVersion => deprecatedSbt = true
```
ckipp01, line 45: For some of these, it seems like we have an extra middle step that isn't really necessary. For example for these, we find the issue, and if it's one of these two, we set it to true, then down below we check if it's true and create a Message. Couldn't we just skip the middle step and instead here instead of returning a Boolean return `Some(message)` or None right away and then there is no need to check for it down below?

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
```diff
@@ -0,0 +1,161 @@
...
+      case single :: Nil => Some(single)
+      case Nil => None
+      case messages if messages == scalaVersionsMessages =>
+        Some(
+          s"Your build definition contains multiple unsupported and deprecated Scala versions."
+        )
+      case _ =>
+        Some(
+          s"Multiple problems detected in your build."
+        )
```
ckipp01, line 121: So in this scenario, we do all the work up above to potentially get multiple messages, but then we never show them to the user, we just show `Multiple problems detected in your build`? Wouldn't it be better to actually show them to the user? I'd assume that `problemSummary` would include the summary of issues, not just a message saying there is issues.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -126,7 +132,7 @@ final class Doctor(
   def check(serverName: String, serverVersion: String): Unit = {
     bspServerName = Some(serverName)
     bspServerVersion = Some(serverVersion)
-    val summary = problemSummary
+    val summary = problemResolver.problemSummary(allTargets())
```
ckipp01, line 135: Mentioned this at the definition site, but I think this is a bit misleading since `summary` here isn't really a summary, but just a message about issues. What do you think of actually including all of those messages if there are more than one here?

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -587,6 +587,29 @@ object Messages {
     }
   }
 
+  object DeprecatedSbtVersion {
+    def message: String = {
+      val recommended = "1.3.2"
+      s"You are using an old sbt version, navigation for which might not be supported in the future versions of Metals. " +
+        s"Please upgrade at least to sbt $recommended."
```
ckipp01, line 594:
```suggestion
        s"Please upgrade to at least sbt $recommended."
```
Just a nit. I'm not sure why, but I think this sounds more natural.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -587,6 +587,29 @@ object Messages {
...
+    }
+  }
+
+  object UnsupportedSbtVersion {
+    def message: String = {
+      val recommended = "1.3.2"
+      s"You are using an old sbt version, navigation for which is not supported in this version of Metals. " +
+        s"Please upgrade at least to sbt $recommended."
```
ckipp01, line 602:
```suggestion
        s"Please upgrade to at least sbt $recommended."
```

ckipp01 (changes requested):
As always, great job on this. I have a few questions/comments. Let me know what you think.

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
tgodzik: Haha, no idea, I didn't even realize I created an abstract class. :sweat_smile:

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
tgodzik: So problem summary will be shown as a message to the user and if we get too much problems then it will be quite bloated. 

`Multiple problems detected in your build` will be accompanied by `More information`, so people can easily take a look. Though I have never honestly seen more than one issue at the same time, so it's highly improbable scenario.

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
tgodzik: I think it's easier to understand the way it is. First we check what problems we have and only then translate to the messages. Otherwise we will have to mix those two steps, but I do understand what you mean and I can change it if you feel like it will be better.

ckipp01: I get what you mean. For consistency then feel free to just leave it as is.

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
ckipp01: Ah ok I totally missed the `More information` part. If that's there, then nevermind.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: Good idea! I moved everything to `ProblemResolver` I also changed to only use `() => Option[BspSession]` which replaces 3 parameters in the doctor and we no longer need the vars.

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
tgodzik: I removed the smaller comments and add scaladoc to `ScalaProblem`

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
tgodzik: Fixed! :sweat_smile:

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
tgodzik: Good idea, fixed!

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
tgodzik: Och yeah, that was actually badly named. Renamed to `unsupportedBloopVersion`

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
tgodzik: That should almost never happen, but your message is a much better one for sure.

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
tgodzik: Fixed!

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
tgodzik: Fixed!

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: I renamed it to `problemMessage`, but I think we should stay with a single message, since it avoids spamming the user and the `More information` option should be visible enough for people to easily check what is going on.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
tgodzik: Yeah, sounds better! Changed!

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
tgodzik: Changed!

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
ckipp01: win, win, win 👍🏼

metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
ckipp01: haha I'm glad I caught this then.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
ckipp01: yea sounds good 👍🏼

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -30,22 +29,19 @@ final class Doctor(
...
-    currentBuildServer: () => Option[String],
+    currentBuildServer: () => Option[BspSession],
     calculateNewBuildServer: () => BspResolvedResult,
     httpServer: () => Option[MetalsHttpServer],
     tables: Tables,
     clientConfig: ClientConfiguration
 )(implicit ec: ExecutionContext) {
   private val hasProblems = new AtomicBoolean(false)
-  private var bspServerName: Option[String] = None
-  private var bspServerVersion: Option[String] = None
```
ckipp01, line 41: Nice to see these are needed anymore 🦾

ckipp01 (approved):
This looks good @tgodzik! Hopefully this will help people diagnose what's going on! LGTM!

Issue #2340, opened 2020-12-24
InvalidPathException on file URLs


nilskp:
Seems that NIO file system is used for file URLs, e.g. `file://`, which crashes on Windows.

```
[Error - 2:06:00 PM] Request textDocument/hover failed.
  Message: Internal error.
  Code: -32603 
java.util.concurrent.CompletionException: java.nio.file.InvalidPathException: Illegal char <:> at index 4: file:///C:/Users/nils/git/delta/.metals/readonly/scala/Option.scala
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:331)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:346)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:704)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088)
	at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1(CancelTokens.scala:40)
	at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1$adapted(CancelTokens.scala:38)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.nio.file.InvalidPathException: Illegal char <:> at index 4: file:///C:/Users/nils/git/delta/.metals/readonly/scala/Option.scala
	at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
	at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:229)
	at java.base/java.nio.file.Path.resolve(Path.java:515)
[...]

PR #2345, opened 2020-12-28
Calculate semanticdb for standalone files

tgodzik:
This solves all issues connected to semanticdb in standalone files. It's now possible to find references, rename etc in standalone files. This was also helpful for testing out and fixing an issue with URIs being wongly produced for semanticdb files produced in the compiler.

Fixes https://github.com/scalameta/metals/issues/2340

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
@@ -66,7 +66,7 @@ final class SyntheticsDecorationProvider(
       val textDocument = currentDocument(path)
       textDocument match {
         case Some(doc) =>
-          publishSyntheticDecorations(path, doc)
+          Future(publishSyntheticDecorations(path, doc))
         case _ =>
           Future.successful(())
       }
@@ -79,7 +79,9 @@ final class SyntheticsDecorationProvider(
     for {
       focused <- focusedDocument()
       filePath = AbsolutePath(path)
-      sourcePath <- SemanticdbClasspath.toScala(workspace, filePath)
+      sourcePath <-
+        if (!filePath.isScala) SemanticdbClasspath.toScala(workspace, filePath)
+        else Some(filePath)
       if sourcePath == focused || !clientConfig.isDidFocusProvider()
       textDoc <- textDocument.documents.headOption
       source <- fingerprints.loadLastValid(sourcePath, textDoc.md5, charset)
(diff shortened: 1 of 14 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
```diff
@@ -27,6 +29,12 @@ class SemanticdbTextDocumentProvider(val compiler: MetalsGlobal) {
     )
     import semanticdbOps._
     val document = unit.toTextDocument
-    document.withUri(filename)
+    val fileUri = Paths.get(new URI(filename))
+    compiler.workspace
+      .map { workspacePath =>
+        val relativeUri = workspacePath.relativize(fileUri).toString()
+        document.withUri(relativeUri)
```
tgodzik, line 36: We expect this to be a relative URI (relative to sourceroot)

tests/unit/src/test/scala/tests/decorations/SyntheticDecorationsLspSuite.scala
```diff
@@ -286,4 +286,45 @@ class SyntheticDecorationsLspSuite extends BaseLspSuite("implicits") {
...
+
+  test("readonly-files") {
+    for {
+      _ <- server.initialize(
+        s"""|/metals.json
+            |{
+            |  "a": {}
+            |}
+            |
+            |/standalone/Main.scala
```
tgodzik, line 298: It was easier to test out this issue with standalone files, because readonly files need to be opened from a build target to be associated to a compiler instance.

tgodzik:
Damn, seems I touched a hornets nest, so I will need to rework this a bit.

mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
```diff
@@ -15,6 +18,9 @@ class SemanticdbTextDocumentProvider(val compiler: MetalsGlobal) {
       cursor = None
     )
     typeCheck(unit)
+    import semanticdbOps._
+    // This cache is never update in semanticdb and will contain the old source
+    gSourceFileInputCache.remove(unit.source)
```
tgodzik, line 23: This seems to be caused by the assumption that the plugin would run only in scalac.

tgodzik (commented):
Ok, I think it should be ready. The main issue was that semanticdb path for documents generated in the compiler would not be relative and that would cause issues on windows.

To test that I decided to add support for standalone files, which needed the interactive semanticdb to be updated on changes, which was itself not a big issue, but it showed two more bugs that we never saw before:

- worksheet dependencies were never linked to any build target or worksheet, so functionalities depending on semanticdb in dependency sources would not work.
- even if we linked dependency source in worksheets to specific build targets, it could miss some dynamically added dependencies, which is why I needed to link those sources to a specific worksheet

So all in all, it took a bit longer, but this PR should now fix 3 bugs and add a new functionality :sweat_smile:

metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
```diff
@@ -79,7 +79,9 @@ final class SyntheticsDecorationProvider(
     for {
       focused <- focusedDocument()
       filePath = AbsolutePath(path)
-      sourcePath <- SemanticdbClasspath.toScala(workspace, filePath)
+      sourcePath <-
+        if (!filePath.isScala) SemanticdbClasspath.toScala(workspace, filePath)
+        else Some(filePath)
```
ckipp01, line 84: Out of curiosity, was this causing issues before? In reality, what other types of files would actually be hitting this? Java files?

mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
```diff
@@ -15,6 +18,9 @@ class SemanticdbTextDocumentProvider(val compiler: MetalsGlobal) {
       cursor = None
     )
     typeCheck(unit)
+    import semanticdbOps._
+    // This cache is never update in semanticdb and will contain the old source
```
ckipp01, line 22:
```suggestion
    // This cache is never updated in semanticdb and will contain the old source
```

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -893,10 +894,12 @@ class MetalsLanguageServer(
       .map(new ApplyWorkspaceEditParams(_))
       .foreach(languageClient.applyEdit)
 
+    // trigger compilation in preparation definition requests for dependency sources and standalone files
```
ckipp01, line 897:
```suggestion
    // trigger compilation in preparation for definition requests for dependency sources and standalone files
```

ckipp01 (approved):
Great job on hunting all of this down @tgodzik. Just a few small nits, but LGTM!

metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
tgodzik: Ach, so this was normally only getting semanticdb files (which would correspond to some scala file), but now that we can produce them from standalone files, then that's the path that is being sent.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: Fixed

mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
tgodzik: Fixed

PR #2350, opened 2020-12-30
Fix incompatibility between currently used Bloop and sbt versions < 1.3.0

tgodzik:
On 1.2.8 we are currently getting:
`[error] java.lang.NoSuchMethodError: sbt.Keys$.reresolveSbtArtifacts()Lsbt/SettingKey;`

I think it's fine to stop supporting 1.2.x in Bloop, since users can always use the older version and not much changes in the integration plugin. But inside Metals this might be problematic, because we force the Bloop version on users.

I am also not 100% how we could fix it in Bloop especially that we compile for 1.3.0 in reality.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
@@ -69,7 +69,8 @@ class BspConnector(
           SbtBuildTool.writeSingleSbtMetalsPlugin(
             workspace.resolve("project"),
             userConfig,
-            isBloop = false
+            isBloop = false,
+            details.getVersion()
           )
           val connectionF = bspServers.newServer(workspace, details)
           statusBar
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):
This makes sense. Not a blocking comment, but what do you think about issuing a warning when a user hits this? I feel like it might makes sense to show a warning that sbt < 1.3.x is deprecated and could be removed in the past, and it's recommended to upgrade. What do  you think? Other than that, LGTM

tgodzik:
It added a warning for older SBT versions  in the other pr refactoring the doctor. I think that should be enough of a warning I think. Although, that one is connected to the Scala version used by SBT, but does correspond to a similar SBT version.

Otherwise we would most likely get two warnings at the same time. What do you think?

ckipp01:
> Otherwise we would most likely get two warnings at the same time. What do you think?

I think we should avoiding showing 2 at the same time, so that's fine then. I haven't had time yet to take a look at that pr, so I'll try to look at it soon.

PR #2364, opened 2021-01-04
Remove Atom from the docs.

ckipp01:
While it's sort of sad to do this, it's probably time to remove
Atom from the supported editors list in the docs. Since the plugin
is no longer maintained, Atom usage is lowering every year, and
the most recent survey had the lowest support shown for Atom, it's
probably as good as time as any.

Good times Atom.


README.md
```diff
@@ -47,8 +47,6 @@ case classes remain.
 - [IntelliJ IDEA](https://www.jetbrains.com/help/idea/discover-intellij-idea-for-scala.html):
   the most widely used IDE for Scala using a re-implementation of the Scala
   typechecker.
-- [Scala IDE](http://scala-ide.org/): Eclipse-based IDE using the Scala
```
ckipp01, line 50: I also just went ahead and removed this. I really don't think it's a recommended alternative anymore. I saw a tweet this weekend about how mentioning this in various places and tutorial often cause confusion for newcomers to Scala, and I thought it was a valid point. I think we should remove this.

website/i18n/en.json
```diff
@@ -104,6 +101,17 @@
       "integrations/tree-view-protocol": {
         "title": "Tree View Protocol"
       },
+      "lsp-extensions/debug-adapter-protocol": {
```
tgodzik, line 104: These changes do not seem to be related, why is it needed?

tgodzik (approved):
LGTM! It's a pity to see an editor go from the list :sweat: but well it makes sense. I just added one comment, which I am unsure about.

gabro (approved):

website/i18n/en.json
ckipp01: So it's not _needed_ per se, but when you run docusaurus locally it generates all the i18n stuff for the various pages. Last time when I was moving stuff around I must have not ran it again for these last two, so they were never created. So I didn't do anything to create these apart from just building the website.

PR #2367, opened 2021-01-05
Update Vim mappings and add reference to nvim-metals.

ckipp01:
This just updates the recommended mappings to the latest from coc.nvim.
This also adds in reference to the new nvim-metals plugin since it's at
a point where it supports pretty much everything except code lenses, debugging,
and the tree view protocol.


tgodzik (approved):
LGTM! Though, I know nothing about the specifics here, but I trust you know what you're doing :+1:

PR #2368, opened 2021-01-05
Add in CopyWorksheetOutput to all so it's in the docs.

ckipp01:
Before this it wasn't being shown under the server commands in the docs.
I also ordered them to make it easier to dig through them since the order
matters in how they appear.


tgodzik (approved):
LGTM!

Issue #2348, opened 2020-12-28
metals occasionally hangs and reports stale errors though sbt compiles fine


nkgm:
**Is your feature request related to a problem? Please describe.**
Autogenerated factory methods with long parameter lists with default values, where one has to invoke by use of named arguments.
Especially useful in scalajs facades where we tend to have long lists of properties with default values of `js.undefined`.

**Describe the solution you'd like**
When autocomplete is invoked within a method call's arguments, the named parameters should be listed. It that's too confusing to list along the other valid completions, maybe list them when the user types`=` as the first character of the argument if possible.

**Describe alternatives you've considered**
The only alternative at this moment is to depend on VSCode's generic text completion, which is pretty much useless. Another alternative is to look for the parameter in the method info popup, which for long lists with default arguments would require going through a long listing of unhighlighted text.

**Search terms**:
`autocomplete` `named arguments` `default parameter values`

tanishiking:
I tried to reproduce in the following environment, but it works fine for me. Indeed, completion for args doesn't work if the args starts with `$` :) 
@nkgm Is my reproduction environment the same as yours? If not, is it possible to give us a minimum reproducible environment? 

<details>
<summary>Click here to show my environment</summary>

metals version: 0.9.8

```scala
// project/build.properties
sbt.version=1.4.4
```

```scala
// project/plugins.sbt
addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.3.1")
```

```scala
// build.sbt
scalaVersion := "2.13.3"
name := "scalajs-playground"
organization := "com.github.tanishiking"
version := "1.0"

enablePlugins(ScalaJSPlugin)
scalaJSUseMainModuleInitializer := true
```

```scala
// src/main/scala/Main.scala
import scala.scalajs.js

object Main {
  def main(args: Array[String]): Unit = {

  }

  def createPerson(
      name: js.UndefOr[String] = js.undefined,
      age: js.UndefOr[Int] = js.undefined
  ): Person = {
    Person(
      name.toOption,
      age.toOption
    )
  }

  case class Person(
      name: Option[String],
      age: Option[Int]
  )
}
```

</details>

![scalajs](https://user-images.githubusercontent.com/9353584/103442878-36ecd800-4c9d-11eb-8cb0-b84021f8f91d.gif)

nkgm:
Thanks so much for taking the time to check and provide the gif capture. 
[...]

PR #2369, opened 2021-01-05
Bare minimum arg completion for scala3

tanishiking:
Related https://github.com/scalameta/metals-feature-requests/issues/112

Though there're still some things to do, I'm submitting a draft PR because I want to ask for advice around the failing tests :pray:, or want to at least have this as a reference implementation for people who want to implement arg completion in the future.

scala3-arg-minimum

I think it's ideal to fix all tests implemented in `cross`, but it might be ok to ship it because it's better than nothing 😅

---

### arg / arg9
```scala
object Main {
  1.until(@@)
}
```

We get empty completion because `methodSym` in `ArgCompletion` always `NoSymbol`.
I guess this is because `assert(@@)` and `1.until(@@)` have multiple overloaded methods like `until(end: Long)` and `until(start: Long, end: Long)`, and that's why we can't retrieve a specific symbol.

```scala
val method = apply.fun
val methodSym = method.symbol
```

It looks like scala2 implementation somehow deals with this case (as `arg9`), but I have no idea how to deal with it in scala3 🤔 
(How to retrieve at least one possible symbol for the method).

### arg4
```scala
case class User(
    name: String = "John",
    age: Int = 42,
    address: String = "",
    followers: Int = 0
)
object Main {
  User("", @@, address = "")
}
```

- in scala2 implementation, the completion suggests `age` and `followers`
- in this impl, `age`, `address`, and `followers`
  - because `args` in ArgCompletion is `List(Literal(Constant("")), Select(Ident(User),$lessinit$greater$default$3), Select(Ident(User),$lessinit$greater$default$4))`, and the nemed arg `address` looks disappeared 🤔 :

### arg6
```scala
case class User(
    name: String = "John",
    age: Int = 42,
    address: String = "",
    followers: Int = 0
)
object Main {
  User("", @@ "")
}
```

- in scala2 implementation, the completion suggests `age`, `followers`, and `followers`
- in this impl `address` and `followers`
  - because `args` in ArgCompletion is `List(Literal(Constant("")), Literal(Constant("")), Select(Ident(User),$lessinit$greater$default$3), Select(Ident(User),$lessinit$greater$default$4))` and the second argument is still there, that's why `name` and `age` are dealed with as filled args.

## Things we're missing that implemented for scala2 (not to do in this PR)
- filtering and sorting completions
- `findPossibleDefaults` from scope
- `fillAllFields`

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/Completions.scala
@@ -1,0 +1,104 @@
+package scala.meta.internal.pc
+
+import dotty.tools.dotc.ast.tpd.{Ident, Apply, Tree, NamedArg, Template, Block, Literal, Select}
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.util.SourcePosition
+import dotty.tools.dotc.core.Constants.Constant
+import dotty.tools.dotc.interactive.Completion
+import dotty.tools.dotc.core.Symbols.Symbol
+import dotty.tools.dotc.core.Names.Name
+
+trait Completions {
+  def completionPosition(
+    pos: SourcePosition,
+    path: List[Tree],
+  )(using ctx: Context): List[Completion] = {
+    path match {
+      case (ident: Ident) :: (app: Apply) :: _ => // fun(arg@@)
+        ArgCompletion(Some(ident), app, pos).contribute
+      case (app: Apply) :: _ => // fun(@@)
+        ArgCompletion(None, app, pos).contribute
+      case _ =>
+        Nil
+    }
+  }
+}
+
+case class ArgCompletion(
+  ident: Option[Ident],
+  apply: Apply,
+  pos: SourcePosition,
+)(using ctx: Context) {
+  val method = apply.fun
+  val methodSym = method.symbol
+
+  // paramSymss contains both type params and value params
+  val vparamss = methodSym.paramSymss.filter(
+    params => params.forall(p => p.isTerm)
+  )
+  val argss = collectArgss(apply)
+
+  // get params and args we are interested in
+  // e.g.
+  // in the following case, the interesting args and params are 
+  // - params: [apple, banana]
+  // - args: [apple, b]
+  // ```
+  // def curry(x; Int)(apple: String, banana: String) = ???
+  // curry(1)(apple = "test", b@@)
+  // ```
+  val (baseParams, baseArgs) = vparamss.zip(argss).lastOption.getOrElse((Nil, Nil))
+
+  val args = ident
+    .map(i => baseArgs.filterNot(_ == i)).getOrElse(baseArgs)
+    .filterNot(isUselessLiteral)
+
+  val isNamed: Set[Name] = args.iterator
+    .zip(baseParams.iterator)
+    // filter out synthesized default args
+    .filterNot {
+      case (Ident(name), _) => name.toString.contains("$")
+      case (Select(Ident(_), name), _) => name.toString.contains("$") // apply method of case class
+      case _ => false
+    }
+    .map {
+      case (NamedArg(name, _), _) => name
+      case (_, param) => param.name
+    }.toSet
+
+  val allParams: List[Symbol] = {
+    baseParams.filterNot(param =>
+      isNamed(param.name) ||
+      param.name.toString.contains("$") // filter out synthesized param, like evidence
+    )
+  }
+
+  val prefix = ident.map(_.name.toString).getOrElse("")
+  val params: List[Symbol] = allParams.filter(param => param.name.startsWith(prefix))
+
+  def contribute: List[Completion] = {
+    val printer = SymbolPrinter()
+    params.map(p => {
+      Completion(
+        s"${p.name.toString} = ",
+        printer.typeString(p.info),
+        List(p)
+      )
+    })
+  }
+
+  private def isUselessLiteral(arg: Tree): Boolean = {
+    arg match {
+      case Literal(Constant(())) => true // unitLiteral
+      case Literal(Constant(null)) => true // nullLiteral
+      case _ => false
+    }
+  }
+
+  private def collectArgss(a: Apply): List[List[Tree]] = {
+    a.fun match {
+      case app: Apply => collectArgss(app) :+ a.args
+      case _ => List(a.args)
+    }
+  }
+}
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/Completions.scala
@@ -7,6 +7,8 @@ import dotty.tools.dotc.core.Constants.Constant
 import dotty.tools.dotc.interactive.Completion
 import dotty.tools.dotc.core.Symbols.Symbol
 import dotty.tools.dotc.core.Names.Name
+import dotty.tools.dotc.core.NameKinds.DefaultGetterName
+import dotty.tools.dotc.core.Flags
 
 trait Completions {
   def completionPosition(
@@ -55,10 +57,11 @@ case class ArgCompletion(
 
   val isNamed: Set[Name] = args.iterator
     .zip(baseParams.iterator)
-    // filter out synthesized default args
+    // filter out synthesized args and default arg getters
     .filterNot {
-      case (Ident(name), _) => name.toString.contains("$")
-      case (Select(Ident(_), name), _) => name.toString.contains("$") // apply method of case class
+      case (arg, _) if arg.symbol.denot.is(Flags.Synthetic) => true
+      case (Ident(name), _) => name.is(DefaultGetterName) // default args
+      case (Select(Ident(_), name), _) => name.is(DefaultGetterName) // default args for apply method
       case _ => false
     }
     .map {
@@ -69,7 +72,7 @@ case class ArgCompletion(
   val allParams: List[Symbol] = {
     baseParams.filterNot(param =>
       isNamed(param.name) ||
-      param.name.toString.contains("$") // filter out synthesized param, like evidence
+      param.denot.is(Flags.Synthetic) // filter out synthesized param, like evidence
     )
   }
 
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -81,7 +81,7 @@ case class ScalaPresentationCompiler(
     sh: Option[ScheduledExecutorService] = None,
     config: PresentationCompilerConfig = PresentationCompilerConfigImpl(),
     workspace: Option[Path] = None
-) extends PresentationCompiler {
+) extends PresentationCompiler with Completions {
 
   def this() = this(Nil, Nil)
 
@@ -124,7 +124,10 @@ case class ScalaPresentationCompiler(
       val pos = sourcePosition(driver, params, uri)
       val items = driver.compilationUnits.get(uri) match {
         case Some(unit) =>
-          Completion.completions(pos)(using ctx.fresh.setCompilationUnit(unit))._2
+          val path = Interactive.pathTo(driver.openedTrees(uri), pos)(using ctx)
+          val completions = Completion.completions(pos)(using ctx.fresh.setCompilationUnit(unit))._2
+          val completions2 = completionPosition(pos, path)(using ctx)
+          completions ++ completions2
         case None => Nil
       }
       new CompletionList(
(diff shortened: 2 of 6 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/Completions.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  val args = ident
+    .map(i => baseArgs.filterNot(_ == i)).getOrElse(baseArgs)
+    .filterNot(isUselessLiteral)
+
+  val isNamed: Set[Name] = args.iterator
+    .zip(baseParams.iterator)
+    // filter out synthesized default args
+    .filterNot {
+      case (Ident(name), _) => name.toString.contains("$")
```
tgodzik, line 60: I wonder if we should indeed do this? Maybe Scala 3 works differently here? This will also cause people to not be able to use named params that contain `$`, which I think is a bug and is reported under https://github.com/scalameta/metals/issues/2348

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -124,7 +124,10 @@ case class ScalaPresentationCompiler(
       val pos = sourcePosition(driver, params, uri)
       val items = driver.compilationUnits.get(uri) match {
         case Some(unit) =>
-          Completion.completions(pos)(using ctx.fresh.setCompilationUnit(unit))._2
+          val path = Interactive.pathTo(driver.openedTrees(uri), pos)(using ctx)
+          val completions = Completion.completions(pos)(using ctx.fresh.setCompilationUnit(unit))._2
+          val completions2 = completionPosition(pos, path)(using ctx)
```
tgodzik, line 129:
```suggestion
          val metalsCompletions = completionPosition(pos, path)(using ctx)
```
?

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -114,7 +143,14 @@ class CompletionArgSuite extends BaseCompletionSuite {
        |age = : Int
        |followers = : Int
        |""".stripMargin,
-    topLines = Option(3)
+    topLines = Option(3),
+    compat = Map(
+      "3.0" ->
+        """|address = : String
```
tgodzik, line 149: It looks the same as the default, so maybe compat is not needed? I think I noticed it in a couple of other places

tgodzik (changes requested):
Looks really good! I left some comments and as for the failing tests:

> arg9

This might be good to report in Dotty, I don't think anything  should return `NoSymbol`. This might be something that was fixed in Scala 2 a while ago.

> arg4

I think this might also be problem to fix in Dotty itself. Or at least discussed with the compiler team.

> arg6

I wonder if it might be connected to the fact that in Scala 2 we add the artificial `CURSOR` to the interpreted source code. I don't really remember why it was needed, but it might be the reason.

tgodzik:
Anyway, I am fine ignoring the tests that are not working currently (it is still improvement over it not working currently :sweat_smile: ) - we should just follow up with issues in Metals and maybe Dotty in some cases.

mtags/src/main/scala-3/scala/meta/internal/pc/Completions.scala
tanishiking: Yeah, in e745df8 tried no to use `$` for filtering out synthetic params and arg getters.

- we could filter out synthesized param like `evidence$1` for something like `def foo[T: Ordering](...)`.
  - https://github.com/scalameta/metals/pull/2369/commits/e745df83da8341f420fd8b176d226ca4509393b9#diff-020d74d58479b87b103a4b80f6ac22e6dd510390fa221baad3305cba2b321649R75
- and we could filter out default args by using `DefaultGetterName` instead of just checking name contains `$` or not.
  - https://github.com/lampepfl/dotty/blob/a5d3ba96b4fd9bcefe3d6d1e7ef69453fe1c0fee/compiler/src/dotty/tools/dotc/core/NameKinds.scala#L328

Maybe we could filter out synthesized params by using `Symbol#isStynthetic` in scala2 :)

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
tanishiking: Ah, that's correct, fixed in  https://github.com/scalameta/metals/pull/2369/commits/8d2a018299384c70dff7e64dc64edcf16d2b3a52

tanishiking:
Thank you so much for the review!
Added some commits for fixing your comments, and skipped failing tests in https://github.com/scalameta/metals/pull/2369/commits/4331986b0ba6acb8490f725dd1a09094161578a2

Now it's ready to review, and I'm also fine with shipping this leaving falling tests (because they require much more effort to fix, working on dotty compiler maybe).

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -44,7 +51,12 @@ class CompletionArgSuite extends BaseCompletionSuite {
        |Main arg2
        |:: scala.collection.immutable
        |""".stripMargin,
-    topLines = Option(3)
+    topLines = Option(3),
+    compat = Map( // findPossibleDefaults and fillAllFields not yet supported
```
tgodzik, line 55: This is not an issue in this test case I think so maybe remove the comment? Same above and below.

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -114,7 +145,8 @@ class CompletionArgSuite extends BaseCompletionSuite {
        |age = : Int
        |followers = : Int
        |""".stripMargin,
-    topLines = Option(3)
+    topLines = Option(3),
+    assumeCond = Some(!isScala3)
```
tgodzik, line 149: Let's add a comment here, why it's not working.

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -227,7 +280,8 @@ class CompletionArgSuite extends BaseCompletionSuite {
        |argument = : Int
        |argument = argument : Int
        |""".stripMargin,
-    topLines = Some(3)
+    topLines = Some(3),
```
tgodzik, line 283: Should it not be just:
```scala
"""|argument: Int
    |argument = : Int
""".stripMargin
```

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -247,7 +301,13 @@ class CompletionArgSuite extends BaseCompletionSuite {
...
        |argument = number8 : Int
        |""".stripMargin,
-    topLines = Some(5)
+    topLines = Some(5),
+    compat = Map(
+      "3.0" ->
+        """|argument = : Int
+           |""".stripMargin
+    ),
+    assumeCond = Some(!isScala3)
```
tgodzik, line 310: Why ignore it? Shouldn't it work currently?

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -261,7 +321,8 @@ class CompletionArgSuite extends BaseCompletionSuite {
         |}
         |""".stripMargin,
     "foo(rele@@)",
-    "foo(relevant)"
+    "foo(relevant)",
+    assumeCond = Some(!isScala3)
```
tgodzik, line 325: This should work currently normally.

tgodzik (changes requested):
Left some comments, thanks for the great work! We could refine it more later. Scala 3 is still not stable, so it's fine to be a bit experimental.

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
tanishiking: Unfortunately, no it doesn't work in scala3.
Because `relevant` is searched from the scope, but scala3 doesn't support it yet.

tanishiking:
@tgodzik Thanks for the review!
I will delve into the dotty internal and file an issue later :)

tgodzik:
@tanishiking is it ready to rereview?

tanishiking:
@tgodzik Oops, sorry yes  it's ready for review :)

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -33,10 +34,18 @@ abstract class BasePCSuite extends BaseSuite {
   val executorService: ScheduledExecutorService =
     Executors.newSingleThreadScheduledExecutor()
   val scalaVersion = BuildInfoVersions.scalaVersion
+  val isScala3: Boolean = isScala3Version(scalaVersion)
```
tgodzik, line 37: This is no longer used.

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -33,10 +34,18 @@ abstract class BasePCSuite extends BaseSuite {
...
   protected val index = new DelegatingGlobalSymbolIndex()
   protected val workspace = new TestingWorkspaceSearch
   val tmp: AbsolutePath = AbsolutePath(Files.createTempDirectory("metals"))
 
+  implicit class RichTestOptions(options: TestOptions) {
```
tgodzik, line 42: And I think this also?

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -31,7 +31,12 @@ class CompletionArgSuite extends BaseCompletionSuite {
        |Main arg1
        |:: scala.collection.immutable
        |""".stripMargin,
-    topLines = Option(3)
+    topLines = Option(3),
+    compat = Map(
+      "3.0" ->
```
tgodzik, line 36: This can be removed since we ignore the test.

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -238,6 +247,9 @@ abstract class BasePCSuite extends BaseSuite {
     }
   }
 
+  object IgnoreScala3Version
```
tgodzik, line 250:
```suggestion
  object IgnoreScala3
```
since it ignores it all?

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -99,11 +118,17 @@ class CompletionArgSuite extends BaseCompletionSuite {
        |Main arg5
        |User arg5
        |""".stripMargin,
-    topLines = Option(4)
+    topLines = Option(4),
+    compat = Map(
```
tgodzik, line 122: Compat section can be removed here also.

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -212,7 +252,12 @@ class CompletionArgSuite extends BaseCompletionSuite {
         |""".stripMargin,
     """|isResourceFile = : Boolean
        |isResourceFile = isLargeBanana : Boolean
-       |""".stripMargin
+       |""".stripMargin,
+    compat = Map(
```
tgodzik, line 256: If the test case is ignored we can remove the compat section.

tgodzik (changes requested):
Looks good! We should be able to merge it, but I would remove some of a leftover code.

Thanks again for the great contribution!

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
tanishiking: Actually, this `compat` is for test `arg1` which is not ignored.
(I'm also confused because GitHub folded the code  😅 )

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
tanishiking: It's for `arg5` not `arg4`

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
tanishiking: This is for `arg14` not for `arg9`

tanishiking:
@tgodzik Thank you so much for reviewing!
Removed leftover code as your feedback (sry, I left a lot of unnecessary code 😅 ), renamed `s/IgnoreScala3Version/IgnoreScala3/`.

I left some comments around `compat`, please take a look :pray:

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
tgodzik: Ach! Sorry about that!

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
tgodzik: My brain filtered out the line numbers :sweat:

tgodzik (approved):
Thanks for removing the leftover code and sorry about those wrong comments. I totally missed that!

tgodzik:
`An error occurred while performing auto-merge.` - this is great :sweat_smile:

PR #2372, opened 2021-01-08
Make sure that worksheet decorations are shown again when the focused

tgodzik:
Previously, if the user switched an editor window the worksheet decorations would disappear and even when they went back, it would not appear until another evaluation was triggered. Now, it will just republish existing evalautions.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -932,6 +932,7 @@ class MetalsLanguageServer(
       CompletableFuture.completedFuture(DidFocusResult.RecentlyActive)
     } else {
       syntheticsDecorator.publishSynthetics(path)
+      worksheetProvider.onDidFocus(path)
       buildTargets.inverseSources(path) match {
         case Some(target) =>
           val isAffectedByCurrentCompilation =
--- a/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
@@ -110,6 +110,17 @@ class WorksheetProvider(
     reset()
   }
 
+  def onDidFocus(path: AbsolutePath): Future[Unit] = Future {
+    if (path.isWorksheet) {
+      val input = path.toInputFromBuffers(buffers)
+      exportableEvaluations.get(input) match {
+        case Some(evaluatedWorksheet) =>
+          publisher.publish(languageClient, path, evaluatedWorksheet)
+        case None =>
+      }
+    }
+  }
+
   def evaluateAndPublish(
       path: AbsolutePath,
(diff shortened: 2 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -882,6 +882,7 @@ class MetalsLanguageServer(
     focusedDocument = Some(path)
     openedFiles.add(path)
     syntheticsDecorator.publishSynthetics(path)
+    worksheetProvider.onDidFocus(path)
```
ckipp01, line 885: How come this one is necessary? Upon leaving and entering, the `didFocusTextDocument` triggers, which you have below and that makes sense. However, when you first open a document, that one is already working. So how come this change?

ckipp01 (commented):
I could have sworn this worked before 🤔. Is there any chance that this somehow regressed? I really feel like this worked in the past. If not, then the change that you have in `didFocusTextDocument` makes sense, but I'm not really sure why it's needed in `didOpen` as it seems that it already works without it?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: I think having it in `didOpen` might be needed in case of editors that don't support focusedDocument I think?

tgodzik:
> I could have sworn this worked before . Is there any chance that this somehow regressed? 

I also think this worked, but I have no idea why it stopped. Though most likely before it would unnecessarily evaluate the worksheet again, so I think this solution is still better.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
ckipp01: So I just tested it by turning off `didFocusTextDocument` in nvim-metals and then opening a worksheet. It still evaluates as expected. From my understanding this works currently and having this in here may actually trigger two evaluations on a `didOpen`. Thinking about this more I'm almost positive this used to work, so I'm very curious what happened. The whole point of the `didFocusTextDocument` was because we wanted some way to still trigger this when it was open before and then regained focus. So afaik, I don't think this is needed in `didOpen` since the worksheet will be evaluated no matter what on the first open without it.

tgodzik: You're right, I removed it from `didOpen`. I think I misunderstood how `didOpen` might work in other editors. Thanks for pointing this out! I am also not sure why it stopped working, tried to look around, but nothing recent broke this behaviour.

ckipp01: ha, it's fresh on my mind since I recently had to think about it for adding worksheet support in nvim-metals.

ckipp01 (approved):
Looks good. LGTM!

PR #2374, opened 2021-01-12
Use scalameta parser dialect Scala3 for Scala 3 sources

tgodzik:
Previously, we would not use scalameta parser for Scala 3 source, which would make some features to not work.

Now, we make sure that we use a proper dialect for each source with a fallback of Scala213.

Features specific to Scala 3 will still require some work, but since this PR is already large I decided to do it separately.


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -175,7 +180,14 @@ class MetalsLanguageServer(
   val parseTrees = new BatchedFunction[AbsolutePath, Unit](paths =>
     CancelableFuture(
       buildServerPromise.future
-        .flatMap(_ => Future.sequence(paths.distinct.map(compilers.didChange)))
+        .flatMap(_ =>
+          Future {
```
gabro, line 184: minor, but since you're not using Future.sequence anymore this can just be map, without creating a new Future here

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
```diff
@@ -0,0 +1,102 @@
...
+    for {
+      text <- buffers.get(path)
+    } yield {
+      val dial = dialect(path)
+      val input = Input.VirtualFile(path.toString(), text)
+      (dial(input).parse[Source], dial)
+    }
+  }
+
+  private def dialect(path: AbsolutePath): Dialect = {
```
ckipp01, line 85:
```suggestion
  private def getDialect(path: AbsolutePath): Dialect = {
```

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
```diff
@@ -0,0 +1,102 @@
...
+    buffers: Buffers,
+    defaultDialect: Dialect = dialects.Scala213
+) {
+
+  private val trees = TrieMap.empty[AbsolutePath, Tree]
+
+  def get(path: AbsolutePath): Option[Tree] =
+    trees.get(path).orElse {
+      // Fallback to parse without caching result.
+      parse(path).flatMap { _._1.toOption }
```
ckipp01, line 32:
```suggestion
      parse(path).flatMap { case (parsedSource, _) => parsedSource.toOption }
```
Just to avoid the `_._1`

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
```diff
@@ -0,0 +1,102 @@
...
+            )
+          case Parsed.Success(tree) =>
+            trees(path) = tree
+            List()
+        }
+      case None =>
+        () // Unknown extension, do nothing.
+        List()
+    }
+  }
```
ckipp01, line 71: So unless I'm missing where, it seems that the only place this is used is in `MetalsLanguageServer` in `buildServerPromise`, and we ignore the value. So I'm curious about this signature it it returning a `List[Diagnostic` when it seems that those are never actually used anywhere?

ckipp01: Also as a total aside, this may benefit from a scaladoc. I feel like a lot of this may not be entirely clear from the outside what this method actually does when you just see `path -> List[Diagnostic]` as the signature.

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  private def parse(
+      path: AbsolutePath
+  ): Option[(Parsed[Source], Dialect)] = {
+    for {
+      text <- buffers.get(path)
+    } yield {
+      val dial = dialect(path)
+      val input = Input.VirtualFile(path.toString(), text)
+      (dial(input).parse[Source], dial)
```
ckipp01, line 81: The `dial` here doesn't seem to be used when we call this in `get` up above. We could just simplify this and only return the parsed source and skip return the dial. Unless you think it will be used in something else you're about to do. wdyt?

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -13,13 +14,14 @@ final class CodeActionProvider(
...
-    scalafixProvider: ScalafixProvider
+    scalafixProvider: ScalafixProvider,
+    trees: Trees
 )(implicit ec: ExecutionContext) {
   private val allActions: List[CodeAction] = List(
     new ImplementAbstractMembers(compilers),
     new ImportMissingSymbol(compilers),
     new CreateNewSymbol(),
-    new StringActions(buffers),
+    new StringActions(buffers, trees),
```
ckipp01, line 24: Good idea to just pass this in 👍🏼

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -994,6 +1014,7 @@ class MetalsLanguageServer(
     val path = params.getTextDocument.getUri.toAbsolutePath
     buffers.remove(path)
     compilers.didClose(path)
+    trees.didClose(path)
```
ckipp01, line 1017: So I honestly don't know, that's why I'm asking, but will there be a penalty paid at all if a user closes a file, and then re-opens it? In that case, we removed the tree from `trees` in `Trees.scala`. It's probably not a huge deal, but am more just curious if it's necessary.

metals/src/main/scala/scala/meta/internal/metals/ScalaTarget.scala
```diff
@@ -18,6 +20,17 @@ case class ScalaTarget(
     autoImports: Option[Seq[String]]
 ) {
 
+  def dialect: Dialect = {
+    scalaBinaryVersion match {
+      case _ if info.getDataKind() == "sbt" => Sbt
+      case "2.11" => Scala211
+      case "2.12" => Scala212
+      case "2.13" => Scala213
+      case version if version.startsWith("3.") => Scala3
```
ckipp01, line 29: Do we also need to check for `startsWith("0")` still?

metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
```diff
@@ -31,44 +31,42 @@ private[debug] final class SetBreakpointsRequestHandler(
...
+            (breakpoint, sym)
+          }
       }
 
-      val allUris = partitions.map(_.getSource().getPath()).toSet
-      val removed = previousBreakpointUris.diff(allUris)
-      previousBreakpointUris = allUris
+    val groups = symbols.groupBy(_._2)
+    val partitions = groups.flatMap {
+      case (Some(fqcn), breakpoints) =>
```
ckipp01, line 52: I have no idea what `fqcn` stands for.

metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
```diff
@@ -31,44 +31,42 @@ private[debug] final class SetBreakpointsRequestHandler(
...
+            val sym = classFinder.findClass(path, pos)
+            (breakpoint, sym)
+          }
       }
 
-      val allUris = partitions.map(_.getSource().getPath()).toSet
-      val removed = previousBreakpointUris.diff(allUris)
-      previousBreakpointUris = allUris
+    val groups = symbols.groupBy(_._2)
+    val partitions = groups.flatMap {
```
ckipp01, line 51: Also, this may be me just not knowing a whole log about DAP stuff, but why are these called `partitions` when these are `Iterable[SetBreakPointsArguments]`?

metals/src/main/scala/scala/meta/internal/parsing/TokenEditDistance.scala
```diff
@@ -312,11 +312,15 @@ object TokenEditDistance {
     } else if (originalInput.value.isEmpty() || revisedInput.value.isEmpty()) {
       noMatch
     } else {
+
+      // Tokenizing should work without issues in all Scala sources
+      val defaultDialect = scala.meta.dialects.Scala213
```
ckipp01, line 317: Just an observation, we seem to define `defaultDialect` quite a few places around the codebase. I wonder if there is a way we can simplify this and set the default in one place?

ckipp01 (changes requested):
Great job on this @tgodzik! Much of this is stuff I've never touched before, so keep that in mind 😬. I have a few questions and comments though!

metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
```diff
@@ -75,6 +76,8 @@ case class MillBuildTool(userConfig: () => UserConfiguration)
 }
 
 object MillBuildTool {
+
+  def defaultDialect: Dialect = scala.meta.dialects.Scala213
```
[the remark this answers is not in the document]
tgodzik: It might a be a `val` really. I will change it.

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -157,7 +156,8 @@ object Digest {
   ): Boolean = {
     try {
       val input = file.toInput
-      val tokens = Trees.defaultDialect(input).tokenize.get
+      val defaultDialect = scala.meta.dialects.Scala213
```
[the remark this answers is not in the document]
tgodzik: Scala 2.13 parser will safely work also for 2.12. There are some additional things added there, so it's rather safe. Besides, it's really just tokens, so it might not matter at all even when it comes to dotty files.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: the penalty should rather be small, parsing is fast, so this is just needed to keep memory issues in check.

metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
tgodzik: fully qualified ??? name - I will make it a bit more explicit.

metals/src/main/scala/scala/meta/internal/parsing/TokenEditDistance.scala
tgodzik: I thinking of doing `defaultTokenizingDialect` for now, since tokenizing code works ok for all languages - it just might not give you real tokens like `KwGiven`, but honestly it's not needed in most places.

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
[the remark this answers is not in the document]
tgodzik: Will change it to `Trees.defaultTokenizerDialect` to make it explicit. We could try to get the right dialect everywhere, but honestly for digest tokenizing even wrongly is fine. As long as it's stable.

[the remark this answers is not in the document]
ckipp01: Sounds good!

metals/src/main/scala/scala/meta/internal/metals/ScalaTarget.scala
tgodzik: So actually the parser will not work with those since the syntax changed. I think we should drop `0.` in this PR actually. Let's do a release with `2.12.13`, merge this and that should give enought time for EPFL students to finish their semester and never look at 0.27 version ever gain :sweat_smile:

metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
tgodzik: We partition breakpoints into groups that pertain to a certain classfile. So for example if we have:
```scala
class Main{
   ...
}

object Main{
  ...
}
```

Breakpoints from `Main.scala` will be split into partitions for `Main$.class` and `Main.class`

tgodzik: Ach and each `SetBreakPointsArguments` is an actual parition.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -694,4 +696,29 @@ object MetalsEnrichments
       )
   }
 
+  implicit class XtensionTreeTokenStream(tree: m.Tree) {
+    import scala.meta._
+    def leadingTokens: Iterator[m.Token] =
+      tree.origin match {
+        case Origin.Parsed(input, dialect, pos) =>
+          val tokens = dialect(input).tokenize.get
```
tanishiking, line 704: [note] Previously, the tokenizer always use `Trees.defaultDialect`, now it uses an appropriate dialect for the input.

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
```diff
@@ -0,0 +1,101 @@
...
+      // Fallback to parse without caching result.
+      parse(path).flatMap { case (parsedSource, _) => parsedSource.toOption }
+    }
+
+  def didClose(fileUri: AbsolutePath): Unit = {
+    trees.remove(fileUri)
+  }
+
+  def didChange(path: AbsolutePath): List[Diagnostic] = {
+    parse(path) match {
```
tanishiking, line 40: Maybe we can skip parsing scala3 code here, for now?

```scala
val dial = dialect(path)
if (dial.allowSignificantIndentation) {
  List()
} else {
  parse(path) match {
  // ...
```

tanishiking (commented):
I left a nit comment but looks good so far :tada:
Thanks to this change, we don't have to duplicate these logics in the scala3 presentation compiler, and that's really awesome 😄

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: Change it a bit to fix some issues.

metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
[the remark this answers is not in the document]
tgodzik: Removed it altogether.

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
tgodzik: Fxied

metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
tgodzik: Fixed!

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
tgodzik: It's still useful to parse it as a lot of the functionalities will start to work even without any additional changes.

metals/src/main/scala/scala/meta/internal/parsing/TokenEditDistance.scala
tgodzik: I changed msot places to use Trees.defaultTokenizerDialect. Any leftovers will be migrated once I manage to fix Mtags to use a correct versions, but I did not want to do it all in this one PR.

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
tgodzik: Changed it around so that dialect is not part of the return value.

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
tgodzik: Ok, so actually this was a bug. We should have been doing something with the return value, I fixed it and added a scaladoc.

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
tgodzik: Fixed

build.sbt
```diff
@@ -9,26 +9,19 @@ def isScala211(v: Option[(Long, Long)]): Boolean = v.contains((2, 11))
 def isScala212(v: Option[(Long, Long)]): Boolean = v.contains((2, 12))
 def isScala213(v: Option[(Long, Long)]): Boolean = v.contains((2, 13))
 def isScala2(v: Option[(Long, Long)]): Boolean = v.exists(_._1 == 2)
-def isScala3(v: Option[(Long, Long)]): Boolean =
-  v.exists(_._1 == 3) || v.exists(_._1 == 0)
-def isScala30(v: Option[(Long, Long)]): Boolean = v.exists(_._1 == 3)
-def isScala3Prerelease(v: Option[(Long, Long)]): Boolean = v.exists(_._1 == 0)
+def isScala3(v: Option[(Long, Long)]): Boolean = v.exists(_._1 == 3)
```
ckipp01, line 12: It makes me happy to be simplifying in this :)

metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
```diff
@@ -77,6 +77,16 @@ final class Diagnostics(
     compileTimer.remove(target)
   }
 
+  def onSyntaxError(path: AbsolutePath, diags: List[Diagnostic]): Unit = {
+    diags.headOption match {
+      case Some(diagnostic) if !path.isInReadonlyDirectory(workspace) =>
+        syntaxError(path) = diagnostic
+        publishDiagnostics(path)
```
ckipp01, line 84: This confuses me a bit. If we have the list of diagnostics already, how come we check to see if it's nonEmpty, and if so, we do `publishDiagnostics(path)`, which then does a lookup to get the diagnostics for that path. But we already have the diagnostics here. Couldn't we just pass them?

metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
```diff
@@ -39,7 +38,8 @@ final class Diagnostics(
     buffers: Buffers,
     languageClient: LanguageClient,
     statistics: StatisticsConfig,
-    config: () => UserConfiguration
+    config: () => UserConfiguration,
+    workspace: AbsolutePath
```
ckipp01, line 42: Not a big deal now, but just digitally thinking out loud here. I know we were thinking about trying to avoid having to pass the workspace all around, which delays the creation of some classes. I see we use it in here only to see if a file is a read only file. If you see an easy way to do this without including the workspace, it may be preferable. If not, no worries.

ckipp01 (changes requested):
Again, great job @tgodzik! Just a couple questions.

metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
tgodzik: This might have made sense at some point, but might have lost it due to refactoring. I will try to simplify it!

tgodzik: Ok, so the difference here is that we treat syntax errors differently to the errors coming from the build server. At each publish we combine both to make sure there are not duplicates. That's why we save the syntax error here, then run publishDiagnostics that will actually combine them both.

We could rework this a bit for sure as it is not clear at all, but I don't think it should be done in this PR as it's already complex. Maybe create a follow up issue to simplify diagnostics?

metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
tgodzik: Changed it to option, so it should work fine in both case.

metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
ckipp01: Ahhh I totally didn't know that. Thanks for the explanation! As for the re-work, yea, add it to the long list.

ckipp01 (approved):
Great job @tgodzik! LGTM 🚀

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -247,6 +257,21 @@ class MetalsLanguageServer(
...
+          // diagnostics from the compiler are still needed for Scala 3
+          compilers.didChange(path).map { diags =>
+            diagnostics.onSyntaxError(path, diags ++ trees.didChange(path))
+          }
+        } else {
+          Future.successful(())
+        }
+      }
+    }
+  }
```
gabro, line 273: not enough traverse 😁

```suggestion
    Future.traverse(paths.distinct) { path =>
      if (path.isScalaFilename) {
        // diagnostics from the compiler are still needed for Scala 3
        compilers.didChange(path).map { diags =>
          diagnostics.onSyntaxError(path, diags ++ trees.didChange(path))
        }
      } else {
        Future.successful(())
      }
    }
  }
```

gabro (approved):
I had a cursory look, but it looks very good!

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: It was `traverse` all along!

PR #2377, opened 2021-01-13
Fix wrong name being displayed for a selected server

tgodzik:


ckipp01 (approved):
Ah, yea good catch! LGTM!

PR #2383, opened 2021-01-14
Add support for Scala 2.12.13

tgodzik:
Running it for now with a SNAPSHOT to make sure it makes sense to release a new semanticdb version.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -186,7 +186,7 @@ lazy val V = new {
   val sbtScala = "2.12.10"
   val scala212 = "2.12.12"
   val scala213 = "2.13.4"
-  val scalameta = "4.4.4"
+  val scalameta = "4.4.6"
   val semanticdb = scalameta
   val bsp = "2.0.0-M13"
   val bloop = "1.4.6-21-464e4ec4"
@@ -198,7 +198,7 @@ lazy val V = new {
   val mdoc = "2.2.14"
   val scalafmt = "2.7.4"
   val munit = "0.7.20"
-  val scalafix = "0.9.24"
+  val scalafix = "0.9.25"
   val lsp4jV = "0.10.0"
   // List of supported Scala versions in SemanticDB. Needs to be manually updated
   // for every SemanticDB upgrade.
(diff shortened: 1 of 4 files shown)


tgodzik:
Doesn't look like there are any issues, so:

https://github.com/scalameta/scalameta/actions/runs/485614408

tgodzik:
Now we are stuck on https://github.com/scalameta/mdoc/pull/445

build.sbt
```diff
@@ -213,7 +213,7 @@ lazy val V = new {
   def deprecatedScala2Versions =
     Seq(scala211, "2.12.8", "2.12.9", "2.13.0", "2.13.1")
   def nonDeprecatedScala2Versions =
-    Seq(scala213, scala212, "2.12.11", "2.12.10", "2.13.2", "2.13.3")
+    Seq(scala213, scala212, "2.12.13", "2.12.11", "2.12.10", "2.13.2", "2.13.3")
```
tgodzik, line 216: I decided to add support for 2.12.13, but keep Metals on 2.12.12 until we can just calmly update. I should have done it in the beginning, but it seemed I would be able to just do one more thing and they just kept poping up.

tests/unit/src/main/scala/tests/QuickBuild.scala
```diff
@@ -304,7 +305,13 @@ object QuickBuild {
 
     val repositories =
       Repository.defaults().asScala ++
-        List(Repository.central(), Repository.ivy2Local())
+        List(
+          Repository.central(),
+          Repository.ivy2Local(),
+          MavenRepository.of(
+            "https://oss.sonatype.org/content/repositories/public"
```
tgodzik, line 312: This will be useful to test out semanticdb snapshot versions in tests.

build.sbt
ckipp01: No worries at all. We can easily just do it after.

ckipp01 (approved):
If nothing is breaking, looks good to me!

Issue #2384, opened 2021-01-14
NPE thrown when `rootUri` in `initialize` is null.


ckipp01:
### Taken from discussions

So this seems to be a problem:
`workspace = AbsolutePath(Paths.get(URI.create(params.getRootUri))).dealias`

and `params.getRootUri` seems to be null?

_Originally posted by @tgodzik in https://github.com/scalameta/metals/discussions/2379#discussioncomment-281741_

### Description

So looking into this a bit more, I think we obviously need to address the NPE, but also think a bit about what should actually happen in the situation where `null` is given here. With Metals, there is a lot that rides on that being set, since it's what we use to define the `workspace` in the server. Then that's passed all around, the providers are created based of it, etc. In reality, if that's not set, then there are quite a few providers that may not make a bunch of sense to even make. For example, the `WorkspaceReload` wouldn't make any sense.

Grabbing some relevant stuff from the spec:

```
/**
	 * The rootPath of the workspace. Is null
	 * if no folder is open.
	 *
	 * @deprecated in favour of `rootUri`.
	 */
	rootPath?: string | null;

	/**
	 * The rootUri of the workspace. Is null if no
	 * folder is open. If both `rootPath` and `rootUri` are set
	 * `rootUri` wins.
	 *
	 * @deprecated in favour of `workspaceFolders`
	 */
	rootUri: DocumentUri | null;

	/**
	 * The workspace folders configured in the client when the server starts.
	 * This property is only available if the client supports workspace folders.
	 * It can be `null` if the client supports workspace folders but none are
	 * configured.
	 *
	 * @since 3.6.0
	 */
	workspaceFolders?: WorkspaceFolder[] | null;
``` 

Ultimately it looks like `workspaceFolders` is the only non-deprecated way to do this, but we should possible look for all 3. However, this begs the question, in what scenario should Metals really be triggered to start with all 3 of these being `null`? Even if we open a single Scala file, that can still be give as the root. I need to look a bit closer at the thread to see what was going on with Sublime, but I'm still a bit confused why Metals was even being triggered if nothing at all was open. If something was open, then the issue may actually lie with how Sublime is setting these values. Again, that doesn't excuse the NPE.

My main concern is that I don't want to do a large refactoring to account for `null` here if that literally _should_ never be the case. If it truly shouldn't be the case, we can even just do nothing and issue a warning.

ckipp01:
[...]

PR #2391, opened 2021-01-15
Avoid NPE when there is no `rootUri`.

ckipp01:
This is mainly to just get a conversation going about the best way to handle this. I've been digging into #2384 a bit, and I'm really unsure of how to handle that situation. This is an attempt to just not throw a NPE and handle it a bit more gracefully. The idea is that for some reason if a user opens a standalone `*.scala` file that isn't in any type of a workspace, and no `rootPath` or `rootUri` is set, then we show them an error and the functionality that would be provided by Metals is very minimal. However minimal, this would be better than forcing an exit like we saw in #2384 which leaves the user with no way to tell what's going on.

Part of the trouble is that a lot of extensions/clients handle this differently.

- In VS Code, https://github.com/scalameta/metals-vscode/blob/d7a76c3f8e61dab4911565f3117cd2cad23cdae5/src/extension.ts#L181-L186, we check to see if `workspaceFolders` is empty, and if so we don't even start Metals. This actually means that in VS Code, you can't just open a Scala file alone and get Metals to work. You can get it to work if you're in a valid workspace and then open a Scala file from another place that is standalone. I'd say this is actually confusing as our docs make it sound like this should work.
- In `coc-metals` coc.nvim will automatically always have this set. If it doesn't match any of the root patters, then it just defaults to the parent directory as the `rootUri`.
- In `nvim-metals` it actually explodes on the neovim side if no root is given, so you have to have some sort of root. So I did the same thing that coc.nvim did and just default to the parent directory if no root patterns are found https://github.com/scalameta/nvim-metals/pull/67.
- In Sublime, if you open a single Scala file, you're not blocked, but both the `rootPath` and the `rootUri` are empty.

In this pr I don't actually check the `workspaceFolders` primarily because Metals doesn't say that it supports `workspaceFolders`, so I don't think it even makes sense to start checking it until we do and say that in `serverCabailities`.

So really the only change of this pr would be in the Sublime case where the user should now see a warning that they need to open up a valid workspace. Ideally, instead it'd be great if they are just opening up a single file, they give the parent as the workspace. I'm not sure how hard that would be @ayoub-benali from the extension.

Finally, in general, it's hard to get rid of the idea of having a workspace, if no impossible with Metals. So much relies on it, and it's tied into everything. So making changes to still have functionality working correctly with no workspace being set would be a major overhaul from what I can tell. One that I'm not sure is worth it at all.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
@@ -18,6 +18,12 @@ import org.eclipse.lsp4j.ShowMessageRequestParams
  */
 object Messages {
 
+  val noRoot = new MessageParams(
+    MessageType.Error,
+    """|No rootUri or rootPath detected.
+       |Metals will not function correctly without either of these set since a workspace is needed.
+       |Try opening your project at the workspace root.""".stripMargin
+  )
   object Worksheets {
     def unsupportedScalaVersion(
         unsupportedVersion: String,
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -426,12 +426,6 @@ class MetalsLanguageServer(
       clientConfig,
       statusBar,
       clientConfig.icons,
-      Option(params.getWorkspaceFolders) match {
-        case Some(folders) =>
-          folders.asScala.map(_.getUri.toAbsolutePath).toList
-        case _ =>
-          Nil
-      },
       tables
     )
     newFileProvider = new NewFileProvider(
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -273,384 +273,397 @@ class MetalsLanguageServer(
   }
 
   private def updateWorkspaceDirectory(params: InitializeParams): Unit = {
-    workspace = AbsolutePath(Paths.get(URI.create(params.getRootUri))).dealias
-    MetalsLogger.setupLspLogger(workspace, redirectSystemOut)
-    val clientInfo = MetalsServerConfig.metalsClientType match {
-      case Some(clientType) => s"for client ${clientType}"
-      case None => ""
-    }
-    scribe.info(
-      s"Started: Metals version ${BuildInfo.metalsVersion} in workspace '$workspace' $clientInfo."
-    )
 
-    clientConfig.experimentalCapabilities =
-      ClientExperimentalCapabilities.from(params.getCapabilities)
-    clientConfig.initializationOptions = InitializationOptions.from(params)
+    // NOTE: we purposefully don't check workspaceFolders here
+    // since Metals technically doesn't support it. Once we implement
+    // https://github.com/scalameta/metals-feature-requests/issues/87 we'll
+    // have to change this.
+    val root =
+      Option(params.getRootUri()).orElse(Option(params.getRootPath()))
 
-    buildTargets.setWorkspaceDirectory(workspace)
-    tables = register(new Tables(workspace, time, clientConfig))
-    buildTargets.setTables(tables)
-    workspaceReload = new WorkspaceReload(
-      workspace,
-      languageClient,
-      tables
-    )
-    buildTools = new BuildTools(
-      workspace,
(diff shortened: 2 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -273,390 +273,397 @@ class MetalsLanguageServer(
...
+      Option(params.getRootUri()).orElse(Option(params.getRootPath()))
 
-    supermethods = new Supermethods(
-      languageClient,
-      definitionProvider,
-      implementationProvider
-    )
+    root match {
+      case None =>
+        languageClient.showMessage(Messages.noRoot)
```
ckipp01, line 286: This pr looks huge because of the shift, but really this is the only difference. If there is no root, we just show the message, and then don't replace the providers.

gabro:
@ckipp01 I think this is definitely better than exploding like we used to do. With this change in place, I think it makes sense for the VSCode extension to skip the check and just rely on the server message.

ckipp01:
> @ckipp01 I think this is definitely better than exploding like we used to do. With this change in place, I think it makes sense for the VSCode extension to skip the check and just rely on the server message.

However, in that scenario then, the advertised support for standalone Scala files won't work in VS Code. Do you think it makes sense to replace the check that I linked there, and if it's emtpy, then give the parent dir as the `rootUri`? Then it will work as expected?

gabro:
Right, that should be fixed.

More in general: the standalone file support means running the rambo compiler on it, right? Why do we need a `rootUri` for that? Conceptually we should be able to get away without it, although I suspect it may require a non-trivial refactor

ckipp01:
> Right, that should be fixed.
> 
> More in general: the standalone file support means running the rambo compiler on it, right? Why do we need a `rootUri` for that? Conceptually we should be able to get away without it, although I suspect it may require a non-trivial refactor

In theory, yea it should work. But like you mention, I don't think the refactor will be trivial. There is also a bunch of small things then we'd also need to consider. Like do we still provide formatting with a standalone file then? If we don't know the workspace, then we don't know where to look or stick the `.scalafmt` file. Ok, that's fine, then we can just make logic to look at what file then are in and place it next to it in the same parent dir.... As you can see, I think there are a bunch of cases like this with the various providers that we'd need to figure out, without a workspace do we still offer them, add in extra logic to handle not having a workspace, etc. 😆 I don't envision that being a fun refactor.

Thinking about it, there is probably a lot of overlap with offering multiple roots in the workspace. So we _could_ save that refactor for when we ever tackle https://github.com/scalameta/metals-feature-requests/issues/87.

gabro:
@ckipp01 I suspected that it wouldn't be trivial. I think you're right in saying that a good time to refactor this would be for the support of multi-folder workspaces, so I think this PR is already enough to fix the crash on the clients that do not perform specific checks. 👍

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
gabro: image

This helps a lot with the review

gabro (approved):
LGTM

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
ckipp01: ahhhh TIL! I've never used this before!

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -18,6 +18,12 @@ import org.eclipse.lsp4j.ShowMessageRequestParams
  */
 object Messages {
 
+  val noRoot = new MessageParams(
+    MessageType.Error,
+    """|No rootUri or rootPath detected.
```
tgodzik, line 23: `No workspace folder detected` ?

`rootUri` and `rootPath` seem internal to LSP, so might be confusing to users

ckipp01: So I guess I was purposefully avoiding that term since `workspaceFolder` is part of the LSP spec that we don't support. So I was trying to avoid using the term. Do you still think it may be clearer to use it?

tgodzik: Right, it might be fine then with the current message. I was mostly thinking aloud here.

tgodzik (approved):
LGTM!

PR #2393, opened 2021-01-16
Clean up --version output.

ckipp01:
So _maybe_ not the most important pr, but I was working on a nice way in `nvim-metals` to show the users what exact version of Metals they are using and what Scala version are supported. `metals --version` actually works pretty nicely for this, but with the current output, if you screen is split or anything the display isn't that nice. With this change it clearly differentiates between the Scala 2.x, and 3 versions and displays them a bit differently.

### Old

Screenshot 2021-01-16 at 14 27 09

### New

Screenshot 2021-01-16 at 14 21 40

### DIFF
--- a/metals/src/main/scala/scala/meta/metals/Main.scala
+++ b/metals/src/main/scala/scala/meta/metals/Main.scala
@@ -11,20 +11,33 @@ import scala.meta.internal.metals.GlobalTrace
 import scala.meta.internal.metals.MetalsLanguageClient
 import scala.meta.internal.metals.MetalsLanguageServer
 import scala.meta.internal.metals.MetalsServerConfig
+import scala.meta.internal.metals.ScalaVersions
 
 import org.eclipse.lsp4j.jsonrpc.Launcher
 
 object Main {
   def main(args: Array[String]): Unit = {
     if (args.exists(Set("-v", "--version", "-version"))) {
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/metals/Main.scala
```diff
@@ -17,14 +17,26 @@ import org.eclipse.lsp4j.jsonrpc.Launcher
 object Main {
   def main(args: Array[String]): Unit = {
     if (args.exists(Set("-v", "--version", "-version"))) {
-      val supportedScalaVersions =
-        BuildInfo.supportedScalaVersions.sorted.mkString(", ")
+      val supportedScala2Versions =
+        BuildInfo.supportedScala2Versions
+          .groupBy(_.substring(0, 4))
```
tgodzik, line 22:
```suggestion
          .groupBy(ScalaVersions.scalaBinaryVersionFromFullVersion)
```
?

tgodzik (approved):
Looks good! Just a minor nitpick, it should be a bit safer, though probably 99% not neccessary.

metals/src/main/scala/scala/meta/metals/Main.scala
ckipp01: Ah of course, I totally forgot that I could use that. Switched.

PR #2397, opened 2021-01-17
Revert "Add release notes for Metals 0.9.9"

tgodzik:
This reverts commit fcc99abecdfcab73e5e96f2f4c7b0dc44bb2349f.

Unfortunately, it seems we need to wait until we can upgrade to 2.12.13, otherwise a number of things would need to change.


ckipp01 (approved):
:( but understood. LGTM

PR #2399, opened 2021-01-18
Update sbt-launch.jar.

ckipp01:
We were still in the 1.3.x range, so this just updates to the latest which we'll
need to start playing around with feeding in `--client` and such to it.


docs/contributors/getting-started.md
```diff
@@ -227,3 +227,15 @@ This property will make your server run in debug mode on port 5005 without
...
+
+The easiest way to update the sbt-launcher is with the following coursier
+command:
+
+```sh
+cs bootstrap org.scala-sbt:sbt-launch:<version> --assembly --preamble=false -o sbt-launch.jar
+```
+
+This will allow you to not have to do some of the manual steps with the launcher
+properties file listed [here](https://github.com/sbt/launcher).
```
ckipp01, line 241: This took me longer than I care to admit to figure out, so I included it in here. I'm still not 100% sure this is the exact way that we did this in the past, but it was the only way I could easily do it without messing around with all this stuff: https://github.com/sbt/launcher#rebundling, which I really wanted to avoid. Everything seems to still be working, but please check this to ensure that something sneaky is happening that I may not realize.

ckipp01:
O, also as an aside, feel free to just leave this until after the release. That's probably best anyways.

tgodzik (approved):
Let's merge it after the release just in case :sweat_smile:

docs/contributors/getting-started.md
alexarchambault: I think `cp "$(cs fetch org.scala-sbt:sbt-launch:<version>)" sbt-launch.jar` just works. `org.scala-sbt:sbt-launch` has no dependencies, so its JAR doesn't need to go through `cs bootstrap`, it can be used as is as an assembly.

ckipp01: Ah, of course. Yea that does seem to work as well, and simplify it. Thanks @alexarchambault!

PR #2402, opened 2021-01-18
Look for clientInfo in intializeParams instead of serverConfig.

ckipp01:
The last couple versions of the LSP spec have included the clientInfo object
in the initializationParams. I think it's better if we just rely on this, especially
seeing that some clients won't have the client set via flags anymore, since it's
no longer necessary due to everything being configurable via intializationOptions.
Plus, doing it this way will also give us the version. All of the modern clients that
I looked into have these filled.

### Before (with the example of Nvim, since it's used without `-Dmetals.client`)
Screenshot 2021-01-18 at 13 56 31

### After
Screenshot 2021-01-18 at 13 54 37

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -287,10 +287,13 @@ class MetalsLanguageServer(
       case Some(path) =>
         workspace = AbsolutePath(Paths.get(URI.create(path))).dealias
         MetalsLogger.setupLspLogger(workspace, redirectSystemOut)
-        val clientInfo = MetalsServerConfig.metalsClientType match {
-          case Some(clientType) => s"for client ${clientType}"
+
+        val clientInfo = Option(params.getClientInfo()) match {
+          case Some(info) =>
+            s"for client ${info.getName()} ${Option(info.getVersion).getOrElse("")}"
           case None => ""
         }
+
         scribe.info(
           s"Started: Metals version ${BuildInfo.metalsVersion} in workspace '$workspace' $clientInfo."
         )


tgodzik (approved):
LGTM!

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -287,10 +287,13 @@ class MetalsLanguageServer(
       case Some(path) =>
         workspace = AbsolutePath(Paths.get(URI.create(path))).dealias
         MetalsLogger.setupLspLogger(workspace, redirectSystemOut)
-        val clientInfo = MetalsServerConfig.metalsClientType match {
-          case Some(clientType) => s"for client ${clientType}"
+
+        val clientInfo = Option(params.getClientInfo()) match {
```
gabro, line 291: do all clients set this? This is potentially breaking for clients that used to rely on the flag and are not yet sending this info via LSP

gabro: Sorry, just saw the comment in the PR description. It's cool then!

Do we still need `metalsClientType` in the config or can we remove it?

gabro: At a quick glance, it seems like this is the only spot where we used `metalsClientType` directly. Other than that it's used to initialize the default config in https://github.com/scalameta/metals/blob/a004fd0a7c0153415243f735a54070abe178924e/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala#L134 so it would be better to align the source of truth and use LSP for that too, wdyt?

ckipp01: As far as I can tell, yes. A few examples of the lesser popular clients
- [emacs-lsp](https://github.com/emacs-lsp/lsp-mode/blob/a0c41e489c656bc69a3cdd0d04ee549f45e42a59/lsp-mode.el#L6656)
- [Neovim](https://github.com/neovim/neovim/commit/1a4d380b5abfff2cf5e46a1b00e98f4381b7e5b0)

And of course the big dogs like VS Code for sure set this.

> potentially breaking 

Also, this is only being used to show the client in the logs, everything else regarind the `-Dmetals.client` stuff is still the same and will still work as expected, it just won't show the client name in the logs.

The one that I did find that doesn't set this is `vim-lsc`. However, I'll send in a pr to fix this for them  since this has been part of the spec for a while now.

ckipp01: https://github.com/natebosch/vim-lsc/pull/373

ckipp01: >  so it would be better to align the source of truth and use LSP for that too, wdyt?

Sorry I was answering your other questions, and totally missed this one. I think in theory, yea, it'd be awesome to just fully get away from the `MetalsServerConfig` altogether, but I'm honest just not sure when. We could do a check on `clientInfo.name` and try to somewhat match on that, but that also may be a bit eh. I'm sort of in the mindset for now to still just leave this in here for a bit since it's not really harming anything. Especially since all the values are looked at last in here: https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/ClientConfiguration.scala

ayoub-benali:
FYI sublime LSP also sends sets clientInfo https://github.com/sublimelsp/LSP/blob/master/plugin/core/sessions.py#L136 but I am not sure if the format matches

ckipp01:
> FYI sublime LSP also sends sets clientInfo [https://github.com/sublimelsp/LSP/blob/master/plugin/core/sessions.py#L136](https://github.com/sublimelsp/LSP/blob/master/plugin/core/sessions.py?rgh-link-date=2021-01-18T13%3A45%3A31Z#L136) but I am not sure if the format matches

Yup, :) I looked there as well to make sure. It's the same format, so it's all good.

Issue #2403, opened 2021-01-18
Current snapshots are throwing exceptions and not working.


ckipp01:
**Describe the bug**

When using the current snapshot of Metals, you'll see the following exception in the logs.

```
Exception in thread "pool-7-thread-1" java.lang.NoSuchMethodError: 'scala.tools.nsc.reporters.Reporter scala.meta.internal.pc.MetalsGlobal.reporter()'
        at scala.meta.internal.pc.ScalaCompilerWrapper.resetReporter(ScalaCompilerAccess.scala:17)
        at scala.meta.internal.pc.CompilerAccess.loadCompiler(CompilerAccess.scala:39)
        at scala.meta.internal.pc.CompilerAccess.withSharedCompiler(CompilerAccess.scala:137)
        at scala.meta.internal.pc.CompilerAccess.$anonfun$withNonInterruptableCompiler$1(CompilerAccess.scala:125)
        at scala.meta.internal.pc.CompilerAccess.$anonfun$onCompilerJobQueue$1(CompilerAccess.scala:197)
        at scala.meta.internal.pc.CompilerJobQueue$Job.run(CompilerJobQueue.scala:103)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:834)
```

**To Reproduce**

1. Set the server version to the latest snapshot.
2. Open a build
3. Look at the logs and see the exception and see that tons of stuff doesn't work

**Expected behavior**

That it works

**Installation:**

- Operating system: All
- Editor: All
- Metals version: `0.9.9+9-372f5c01-SNAPSHOT` and all of them since https://github.com/scalameta/metals/pull/2383

**Additional context**

The issue is that we have the `2.12.13` compiler as a dependency, when our project is on `2.12.12`. So all things related to the presentation compiler are going crazy.

```
❯ cs resolve org.scalameta:metals_2.12:0.9.9+9-372f5c01-SNAPSHOT -r bintray:scalacenter/releases -r sonatype:snapshots | grep compiler
org.scala-lang:scala-compiler:2.12.13:default
```

I'm just adding this in here in case someone searches for it, they will see that we are aware of it.

tgodzik:
Ah yes. that is the issue that caused me to revert the vscode release. Working on it!

PR #2401, opened 2021-01-18
Ensure semanticdbVersion can be overwritten.

ckipp01:
This ensures that if a user is going to set the `semanticdbVersion`
locally with something like:

```
ThisBuild / semanticdbVersion := "4.4.6"
```

That they are still able to.

Closes #2395

PR #2404, opened 2021-01-18
Update Metals Scala 2.12 version to 2.12.13

tgodzik:
Fixes https://github.com/scalameta/metals/issues/2403

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -184,7 +184,7 @@ lazy val V = new {
   val scala210 = "2.10.7"
   val scala211 = "2.11.12"
   val sbtScala = "2.12.10"
-  val scala212 = "2.12.12"
+  val scala212 = "2.12.13"
   val scala213 = "2.13.4"
   val scalameta = "4.4.6"
   val semanticdb = scalameta
@@ -213,7 +213,7 @@ lazy val V = new {
   def deprecatedScala2Versions =
     Seq(scala211, "2.12.8", "2.12.9", "2.13.0", "2.13.1")
   def nonDeprecatedScala2Versions =
-    Seq(scala213, scala212, "2.12.13", "2.12.11", "2.12.10", "2.13.2", "2.13.3")
+    Seq(scala213, scala212, "2.12.12", "2.12.11", "2.12.10", "2.13.2", "2.13.3")
   def scala2Versions = nonDeprecatedScala2Versions ++ deprecatedScala2Versions
 
   // Scala 3
@@ -571,7 +571,15 @@ def publishBinaryMtags =
     .in(interfaces)
     .dependsOn(
       publishAllMtags(
-        List(V.scala211, V.sbtScala, V.scala212, V.scala213, V.scala3)
+        List(
+          V.scala211,
+          V.sbtScala,
+          V.scala212,
+          // TODO https://github.com/scalameta/metals/issues/2392
+          "2.12.12",
+          V.scala213,
+          V.scala3
+        )
       )
     )
 
--- a/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
+++ b/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
@@ -13,14 +13,16 @@ object MetalsPlugin extends AutoPlugin {
 
   val supportedScala2Versions = BuildInfo.supportedScala2Versions.toList
 
-  override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
-    semanticdbVersion := BuildInfo.semanticdbVersion
-  )
+  val semanticdbVersion = BuildInfo.semanticdbVersion
 
   override lazy val projectSettings: Seq[Def.Setting[_]] = Seq(
     semanticdbCompilerPlugin := {
-      ("org.scalameta" % "semanticdb-scalac" % semanticdbVersion.value)
-        .cross(CrossVersion.full)
+      if (supportedScala2Versions.contains(scalaVersion.value))
+        ("org.scalameta" % "semanticdb-scalac" % semanticdbVersion)
+          .cross(CrossVersion.full)
+      else {
+        semanticdbCompilerPlugin.value
+      }
     },
     allDependencies ++= {
       val versionOfScala = scalaVersion.value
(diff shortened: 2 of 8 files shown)


build.sbt
```diff
@@ -571,7 +571,15 @@ def publishBinaryMtags =
     .in(interfaces)
     .dependsOn(
       publishAllMtags(
-        List(V.scala211, V.sbtScala, V.scala212, V.scala213, V.scala3)
+        List(
+          V.scala211,
+          V.sbtScala,
+          V.scala212,
+          // TODO https://github.com/scalameta/metals/issues/2392
```
ckipp01, line 578: Can we just bump the version of Ammonite right away to not have to worry about this? It looks like it's available for 2.12.13 -> `2.3.8-32-64308dc3`. Or you you purposefully not including that?

tgodzik: So turns out there was a regression, but once https://github.com/lihaoyi/Ammonite/pull/1146 is released we can upgrade.

ckipp01: https://media.giphy.com/media/H3qWY13eGZUomKxOs6/giphy.mp4

Regressions this week...

ckipp01 (approved):
Great job tackling all the stuff needed for this @tgodzik! LGTM!

tgodzik:
It seems the change to semanticdb versions does not allow for Metals to specify the semanticdb definition now :sigh:

ckipp01:
> It seems the change to semanticdb versions does not allow for Metals to specify the semanticdb definition now :sigh:

So is this https://github.com/scalameta/metals/pull/2401 causing our 4.4.6 to be overwritten by the `SemanticdbPlugin` value, 4.4.0?

tgodzik:
> > It seems the change to semanticdb versions does not allow for Metals to specify the semanticdb definition now :sigh:
> 
> So is this #2401 causing our 4.4.6 to be overwritten by the `SemanticdbPlugin` value, 4.4.0?

Yeah, I think we cannot differentiate between the default and the a version set by the user. What about selecting the newest version? Similar to other dependencies?

ckipp01:
> > > It seems the change to semanticdb versions does not allow for Metals to specify the semanticdb definition now :sigh:
> > 
> > 
> > So is this #2401 causing our 4.4.6 to be overwritten by the `SemanticdbPlugin` value, 4.4.0?
> 
> Yeah, I think we cannot differentiate between the default and the a version set by the user. What about selecting the newest version? Similar to other dependencies?

Yea we can do that, but if we select the newest we sort of have the inverse problem, where if someone is trying to set `4.4.0` they won't be able to since we'll just choose the one we provide since it's newer? Is that ok?

tgodzik:
> > > > It seems the change to semanticdb versions does not allow for Metals to specify the semanticdb definition now :sigh:
> > > 
> > > 
> > > So is this #2401 causing our 4.4.6 to be overwritten by the `SemanticdbPlugin` value, 4.4.0?
> > 
> > 
> > Yeah, I think we cannot differentiate between the default and the a version set by the user. What about selecting the newest version? Similar to other dependencies?
> 
> Yea we can do that, but if we select the newest we sort of have the inverse problem, where if someone is trying to set `4.4.0` they won't be able to since we'll just choose the one we provide since it's newer? Is that ok?

I think so, we wouldn't want the users to use an older version, which might have some issues still. Although, I am actually unable to fix the error. It always tries to download 4.4.0/

ckipp01:
> > > > > It seems the change to semanticdb versions does not allow for Metals to specify the semanticdb definition now :sigh:
> > > > 
> > > > 
> > > > So is this #2401 causing our 4.4.6 to be overwritten by the `SemanticdbPlugin` value, 4.4.0?
> > > 
> > > 
> > > Yeah, I think we cannot differentiate between the default and the a version set by the user. What about selecting the newest version? Similar to other dependencies?
> > 
> > 
> > Yea we can do that, but if we select the newest we sort of have the inverse problem, where if someone is trying to set `4.4.0` they won't be able to since we'll just choose the one we provide since it's newer? Is that ok?
> 
> I think so, we wouldn't want the users to use an older version, which might have some issues still. Although, I am actually unable to fix the error. It always tries to download 4.4.0/

Alright, https://github.com/scalameta/metals/pull/2406 seems to fix it.

ckipp01:
Alright, the change I added on top in here https://github.com/scalameta/metals/pull/2407 seems to do what we want.

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
```diff
@@ -13,14 +13,16 @@ object MetalsPlugin extends AutoPlugin {
...
-  override lazy val globalSettings: Seq[Def.Setting[_]] = Seq(
-    semanticdbVersion := BuildInfo.semanticdbVersion
-  )
+  val semanticdbVersion = BuildInfo.semanticdbVersion
 
   override lazy val projectSettings: Seq[Def.Setting[_]] = Seq(
     semanticdbCompilerPlugin := {
-      ("org.scalameta" % "semanticdb-scalac" % semanticdbVersion.value)
-        .cross(CrossVersion.full)
+      if (supportedScala2Versions.contains(scalaVersion.value))
```
tgodzik, line 20: @ckipp01 what about this? I think this might have been the actual problem, we should not set `semanticdbCompilerPlugin` if we don't currently support the given scala version. If a user had the Metals plugin and `semanticdbEnabled := true` then we would have replaced the original without actually needing to do it. So it would break user's builds even though MEtals does not support the given version.

ckipp01: Yea, that makes sense 👍🏼 . So if it's not in our supported versions, we just default to what they want. The only problem (and I don't know how to get around) is that for some reason if someone wants to bump the semanticdb version while using a Scala version that we do support, they can't.

tgodzik: I think it's fine, we update the semanticdb version often enough. There should be no actual need to do it manually. Besides, if the user modifies the version we cannot know if it actually supports the given Scala version. Metals should be a source of truth here the same as with Bloop.

PR #2409, opened 2021-01-19
Bump mdoc to 2.2.15

tgodzik:
Starting from 2.12.13 mdoc is no longer binary compatible for the earlier Scala versions. Legacy 2.12.12 versions is released alongside to remedy this.

This is just be on the safe side, since everything seemed to work ok without this change. After merging this I will take care of the release.


ckipp01 (approved):
Sounds good. LGTM!

tgodzik:
Actually, since mdoc brings in it's own Scala compiler versions, this is not needed. I will just bump mdoc version then.

ckipp01:
> Actually, since mdoc brings in it's own Scala compiler versions, this is not needed. I will just bump mdoc version then.

Even better then 👍🏼

PR #2411, opened 2021-01-20
Fix default version of Scala given to Ammonite.

ckipp01:
We are pulling the version of Ammonite from buildInfo and the version
of Ammonite we are using isn't available for 2.12.13 yet. We did fix
this is in the Ammonite tests, but we forgot to actually change it where
we get Ammonite.

Also I did check and there still isn't a new version published for for Ammonite and 2.12.13.

This was originally reported by @soronpo on [Gitter](https://gitter.im/scalameta/metals?at=6007952bdfdbc1437f9d42a5)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -184,6 +184,8 @@ lazy val V = new {
   val scala210 = "2.10.7"
   val scala211 = "2.11.12"
   val sbtScala = "2.12.10"
+  // TODO https://github.com/scalameta/metals/issues/2392
+  val ammonite212Version = "2.12.12"
   val scala212 = "2.12.13"
   val scala213 = "2.13.4"
   val scalameta = "4.4.6"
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -184,6 +184,8 @@ lazy val V = new {
   val scala210 = "2.10.7"
   val scala211 = "2.11.12"
   val sbtScala = "2.12.10"
+  // TODO https://github.com/scalameta/metals/issues/2392
+  val legacy212 = "2.12.12"
```
tgodzik, line 188:
```suggestion
  val ammonite212Version = "2.12.12"
```
We can keep it as is and when all is fixed change it to:
`  val ammonite212Version = scala212`

Since this happens frequently.

tgodzik (changes requested):

build.sbt
ckipp01: Ah, good idea 👍🏼 changed.

tgodzik (approved):
LGTM!

PR #2412, opened 2021-01-20
Add rename and references for sbt files

### LINKED ISSUE #2398: Support for references/rename in sbt files
Thanks to https://github.com/scalameta/metals/pull/1865 we now have most of the needed features for supporting sbt files. 

One of the missing features is references and features that use it internally like `rename` or `go to implementation`. We most likely will not be able to use Bloop in this case, we should maybe try creating semanticdb files on demand (I think we do have a way to do it currently.) This is problematic especially in case of multiple source files, since we would need to update all manually.

Maybe we could also try using the presentation compiler here - it would be useful to implement a fallback `find references` there.

Find references working correctly in sbt files.

Searching for them textually

**Search terms**:
sbt rename references

dos65:
Fixes #2398.
Enables `references`/`rename` features.


metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
```diff
@@ -149,16 +148,9 @@ final class InteractiveSemanticdbs(
     }
   }
 
-  def getBuildTarget(
```
tgodzik, line 152: This is still needed - when looking into files in the `./metals/readonly` we try to infer the build target (and the right presentation compiler) based on where the file comes from. For example if we opened a file in project A and go to definition opened another file for example `Predef.scala` we will associate `Predef.scala` with project A.

I think that actually we can move it to `BuildTargets` class and create`inferBuildTarget` method. This we could use in the `Compilers` class instead of always doing`.orElse(interactiveSemanticdbs.flatMap(_.getBuildTarget(path)))` 

We would need to also move `toFileOnDisk`, `readonlyToSource`, `didDefinition`, and `inferBuildTarget` but that should not be an issue.

mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
```diff
@@ -31,7 +35,10 @@ class SemanticdbTextDocumentProvider(val compiler: MetalsGlobal) {
       compiler.reporter,
       SemanticdbConfig.default
     )
-    val document = unit.toTextDocument
+
+    // TODO: worksheets support
```
tgodzik, line 39: Best not to leave TODO - let's create a feature request for worksheets.

tgodzik (changes requested):
Thanks for the amazing work! I left some comments on the PR.

Could you also try to keep the PR name something like:
`Add rename and references for sbt files`
and add `Fixes #2398` in the PR description?

It makes it easier to generate release notes later on and it's more user friendly.

metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
```diff
@@ -170,34 +162,79 @@ final class InteractiveSemanticdbs(
...
       .get(
         clientConfig.initialConfig.compilers.timeoutDelay,
         clientConfig.initialConfig.compilers.timeoutUnit
       )
-    val textDocument = TextDocument.parseFrom(bytes)
-    textDocument
-
+    val textDocument = s.TextDocument.parseFrom(bytes)
+    if (prependedLinesSize > 0)
+      cleanupAutoImports(textDocument, text, prependedLinesSize)
```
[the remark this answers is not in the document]
dos65: In the beginning, I thought to implement it as you suggest.

However, I found that `references` and `rename` are based on `DefinitionProvider.positionOccurence`.
It internally uses `TokenEditDistance` between buffer and TextDocument. So for sbt files, it means that TextDocument has to be cleaned from sbt-specific imports before it will be passed into `TokenEditDistance`. Otherwise, it's method `toOriginal` fails.
It might be handled by adding sbt/adjust-specific logic into `DefinitionProvider` but I don't think that it's the right way to go.

So, I decided that it would be better just to leave these hacks in `IntercativeSemanticdb` because in this case adjustment is applied on TextDocument, not on LSP positions.

metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
dos65: I deleted it because `BuildTargets` already has `inferBuildTarget` method. 
It's implemented a little bit different way but looks similar. I tested that `references` in `./metals/readonly` still works in this PR

tgodzik: Och, yeah there is also `inferBuildTarget` method but it is much more taxing (creates new class loader each time :O). However, it seems that `inferBuildTarget` here does not make much sense. We do save everything to database and it will stay longer than the in memory `readonlyToSource`. I think it's a good idea to remove it, it's probably a leftover.

We should also remove then `readonlyToSource`, which does not make any sense anymore and `toFileOnDisk`, which we can just use the normal underlying `toFileOnDisk` in `WorkspaceSymbolProvider`.

It would also be good to cache the `URLClassLoader` that gets created there, it may actually create a memory leak, but it can be done in a separate PR.

metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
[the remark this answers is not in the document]
tgodzik: I think that should not be an issue. DefinitionProvider would still use `InteractiveSemanticdb`, so there would be no need to add sbt specific logic anywhere. However, I see now that it's a bigger refactor here, so maybe I can take a look a bit later while taking a look at how we infer build targets (especially for worksheets, which I guess might not work perfectly now) and how we could improve it.

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
dos65: > We should also remove then readonlyToSource

:+1:  Done

> It would also be good to cache the URLClassLoader that gets created there, it may actually create a memory leak, but it can be done in a separate PR.

Agree about doing this in separate PR. Will do.

tgodzik (approved):
LGTM!

PR #2417, opened 2021-01-21
Add configuration option for default Scala version

### LINKED ISSUE #1653: Improve standalone presentation compiler integration
Current situation with presentation compiler is as follows:

1. Get a file uri.
2. Check if it belongs to a build target.
  a. If it does, check if the Scala version is supported.
      I. If it does, create a presentation compiler with the build target classpath and it's Scala version
      II. Otherwise, fallback to b. 
  b. If it doesn't, use a "rambo" 2.12 compiler without a classpath except Scala library.

Proposal:

1. Get a file uri.
2. Check if it belongs to a build target.
  a. If it does, check if the Scala version is supported.
    I. If it is, create a presentation compiler with the build target classpath and it's Scala version
    II. Otherwise, use ScalaVersion.recommenedVersion() and create a presentation compiler with that recommended version and build target's options + classpath.
  b. If it doesn't, use a "rambo" compiler without a classpath except Scala library, although first ask for a default Scala version for the workspace if it is not already set.

For 2.a.II I think we shouldn't ask for the Scala Version, since multiple targets could have different unsupported Scala version. Alternatively, we could use the recommended Scala version  as a default suggestion for the workspace.
 
This will allow us to work specifically with Dotty snapshot versions and better support older Scala versions.

dos65:
Closes #1653

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
@@ -13,9 +13,9 @@ import scala.util.Try
 import scala.meta.inputs.Input
 import scala.meta.inputs.Position
 import scala.meta.internal.builds.SbtBuildTool
+import scala.meta.internal.metals.Compilers.PresentationCompilerKey
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.ammonite.Ammonite
-import scala.meta.internal.mtags
 import scala.meta.internal.pc.EmptySymbolSearch
 import scala.meta.internal.pc.LogMessages
 import scala.meta.internal.pc.ScalaPresentationCompiler
@@ -59,7 +59,8 @@ class Compilers(
     sh: ScheduledExecutorService,
     initializeParams: Option[InitializeParams],
     diagnostics: Diagnostics,
-    isExcludedPackage: String => Boolean
+    isExcludedPackage: String => Boolean,
+    scalaVersionSelector: ScalaVersionSelector
 )(implicit ec: ExecutionContextExecutorService)
     extends Cancelable {
   val plugins = new CompilerPlugins()
@@ -67,9 +68,9 @@ class Compilers(
   // Not a TrieMap because we want to avoid loading duplicate compilers for the same build target.
   // Not a `j.u.c.ConcurrentHashMap` because it can deadlock in `computeIfAbsent` when the absent
   // function is expensive, which is the case here.
-  val jcache: ju.Map[BuildTargetIdentifier, PresentationCompiler] =
+  val jcache: ju.Map[PresentationCompilerKey, PresentationCompiler] =
     Collections.synchronizedMap(
-      new java.util.HashMap[BuildTargetIdentifier, PresentationCompiler]
+      new java.util.HashMap[PresentationCompilerKey, PresentationCompiler]
     )
   private val jworksheetsCache: ju.Map[AbsolutePath, PresentationCompiler] =
     Collections.synchronizedMap(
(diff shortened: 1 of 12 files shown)


ckipp01 (changes requested):
Great job diving into this @dos65! From my perspective, I think it's quite important that if we're introducing this setting, that it's used for standalone Scala files, standalone worksheets, and standalone Ammonite scripts. I have some more comments in the relevant place about it.

Excited about these changes 🚀

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -220,6 +221,15 @@ object UserConfiguration {
...
           |""".stripMargin
+      ),
+      UserConfigurationOption(
+        "default-scala-version",
+        BuildInfo.scala212,
+        BuildInfo.scala212,
+        "Should display the default Scala version",
+        """|Scala-compiler version that is used as default or fallback in case if file
+           |doesn't belong to any build target or specified Scala version isn't supoorted by Metals
+           |""".stripMargin
```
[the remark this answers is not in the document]
dos65: Thanks for the correction and pointing that we should apply the same logic in Ammonite scripts!

>If we do go ahead and ensure that it's for all standalone files, another thing we need to think about is the Ammonite situation where a user may have default-scala-version set to 2.12.12, but then at the top of the Ammonite script the have // scala 2.13.3. Which one wins?

I've slightly updated PR. Now ammonite uses this setting too.
`// scala 2.13.3` comment from the script is prioritized more.

> what about just naming it `fallback-scala-version`? wdyt?

Changed on `fallback-scala-version`

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -72,57 +72,90 @@ class Compilers(
...
-  )
-
   private def createStandaloneCompiler(
+      scalaVersion: String,
       classpath: Seq[Path],
       standaloneSearch: SymbolSearch,
       name: String
   ): PresentationCompiler = {
     scribe.info(
       "no build target: using presentation compiler with only scala-library"
```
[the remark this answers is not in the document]
dos65: Added it to log

dos65:
@ckipp01 @tgodzik 
I've updated the scala version selection logic

Now it's the following:
- If `fallback-scala-version` is configured, select it
- If it isn't then try to find the latest version from `BuildTarget`
- Or finally fallback onto `scala212`

Additional rules:
- For Ammonite - use `scala213` in case the steps above returned `scala3`
- Correct resulting scala version by `ScalaVersion.recommendedVersion` if it isn't supported

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -72,57 +72,90 @@ class Compilers(
...
+      List.empty,
+      classpath,
+      standaloneSearch,
+      name
+    )
+  }
+
+  private def defaultScalaVersion: String = {
+    userConfig().defaultScalaVersion match {
+      case Some(v) if ScalaVersions.isSupportedScalaVersion(v) => v
```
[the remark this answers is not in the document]
dos65: Good point! Updated this thing

ckipp01:
It looks like there are still some worksheet issues/tests to work out @dos65.

dos65:
@ckipp01 @tgodzik 
Finally, I managed to fix issues with the worksheets.

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -283,4 +280,14 @@ object Embedded {
     new URLClassLoader(urls, classLoader)
   }
 
+  /**
+   * Provides scala3-library dependency for stadalone PC and mdoc instances
+   */
+  def standaloneDependencies(scalaVersion: String): List[Path] = {
```
tgodzik, line 286:
```suggestion
  def scalaLibrary(scalaVersion: String): List[Path] = {
```

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -283,4 +280,14 @@ object Embedded {
...
+    if (ScalaVersions.isScala3Version(scalaVersion))
+      downloadDependency(scala3Dependency(scalaVersion), scalaVersion)
+    else
+      List.empty
```
tgodzik, line 290: We might still need to download Scala 2 ones if it's for example `2.13.4`

metals/src/main/scala/scala/meta/internal/metals/ScalaVersionSelector.scala
```diff
@@ -0,0 +1,27 @@
...
+          .getOrElse(BuildInfo.scala212)
+    }
+
+    // ammonite doesn't support Scala3 yet
+    if (!allowScala3 && ScalaVersions.isScala3Version(selected))
+      BuildInfo.scala213
+    else if (ScalaVersions.isSupportedScalaVersion(selected))
+      selected
+    else
+      ScalaVersions.recommendedVersion(selected)
```
tgodzik, line 25: We should show message in this case, that the selected version is not supported.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -220,6 +221,16 @@ object UserConfiguration {
           |See https://scalameta.org/metals/docs/contributors/remote-language-server.html for
           |documentation on remote language servers.
           |""".stripMargin
+      ),
+      UserConfigurationOption(
+        "fallback-scala-version",
```
tgodzik, line 226: We should also create a PR to metals-vscode and add the new configuration option to `package.json`, we should also only allow to select from a number of values (I think that should be supported)

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -80,23 +84,42 @@ class WorksheetProvider(
...
   private val exportableEvaluations =
     new TrieMap[VirtualFile, EvaluatedWorksheet]()
-  private val currentScalaVersion =
-    scala.meta.internal.mtags.BuildInfo.scalaCompilerVersion
-  private val currentBinaryVersion =
-    ScalaVersions.scalaBinaryVersionFromFullVersion(currentScalaVersion)
-  private lazy val ramboMdoc =
-    embedded.mdoc(currentScalaVersion, currentBinaryVersion)
+
+  private def ramboMdoc: Mdoc = {
```
tgodzik, line 92:
```suggestion
  private def fallbackMdoc: Mdoc = {
```
?

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -450,7 +485,9 @@ class Compilers(
   )(fn: (PresentationCompiler, Position, AdjustLspData) => T): T = {
     val path = params.getTextDocument.getUri.toAbsolutePath
     val compiler =
-      loadCompiler(path, interactiveSemanticdbs).getOrElse(ramboCompiler)
+      loadCompiler(path, interactiveSemanticdbs).getOrElse({
```
tgodzik, line 488:
```suggestion
      loadCompiler(path, interactiveSemanticdbs).getOrElse(fallbackCompiler)
```

tgodzik (changes requested):
Looks pretty good! I left some comments, but I think nothing major. You should also update metals-vscode plugin. Let me know if you have any trouble working on it!

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
dos65: Yep, I have a branch with these changes. Will publish them soon.
Decided to delay this PR creation because I expected there might be some changes in option name and it's description)

metals/src/main/scala/scala/meta/internal/metals/ScalaVersionSelector.scala
dos65: Added warning about that on `workspace/didChangeConfiguration`

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
dos65: For some reason, everything works well without explicit scala-library in classpath on scala 2 versions.
But nor PC, nor Mdoc doesn't work without it on scala 3.

Anyway, I agree that is better to cover all branches here because it shouldn't affect the PC/Mdoc startup speed. 
Fixed.

ckipp01 (approved):
This looks great @dos65! Thanks for you work on this! It's a LGTM from me, but can you squash some of your "review fixes" commits. We try to keep each commit meaningful and able to stand on its own.

dos65:
@ckipp01 Sure! Done.

tgodzik (changes requested):
Just a small nitpick from my side, but otherwise it looks good!

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1105,6 +1113,20 @@ class MetalsLanguageServer(
             excludedPackageHandler.update(userConfig.excludedPackages)
             workspaceSymbols.indexClasspath()
           }
+
+          userConfig.fallbackScalaVersion.foreach { version =>
+            if (!ScalaVersions.isSupportedScalaVersion(version)) {
+              val message =
```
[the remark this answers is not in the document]
dos65: Done

tgodzik (approved):
LGTM! Thanks!

PR #2421, opened 2021-01-22
Provide a more specific error message when scalafix fails organizing imports

mlachkar:
There are two levels of possible failure on scalafix:
- The first one: Scalafix completely fails running because the interface is not called with the correct arguments, or the provided file to fix doesn't exit. So the possible errors here are : commandLine error, Nofile to fix. This is not supposed to happen with Metals. 
- The second level: Scalafix can fix multiple files. For each file, we store if scalafix run successfully or not. In Metals, we only run on one file, and the possible errors are ParseError (if the file cannot be parsed), StaleSemanticdbError, MissingSemanticdbError. Those errors may happen and we want to display them.

So this PR improves the errors displayed. thanks for your feedbacks

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
@@ -94,12 +94,11 @@ case class ScalafixProvider(
               exception
             )
             Future.failed(exception)
-          case Success(results) if !results.isSuccessful =>
-            val scalafixError = results.getErrorMessage().asScala
-            val message = scalafixError.getOrElse(defaultErrorMessage)
-            val exception = ScalafixRunException(message)
+          case Success(results) if !scalafixSucceded(results) =>
+            val scalafixError = getMessageErrorFromScalafix(results)
+            val exception = ScalafixRunException(scalafixError)
             reportScalafixError(
-              message,
+              scalafixError,
               exception
             )
             Future.failed(exception)
@@ -114,6 +113,22 @@ case class ScalafixProvider(
         }
       }
     }
+  }
+  private def scalafixSucceded(evaluation: ScalafixEvaluation): Boolean =
+    evaluation.isSuccessful && evaluation
+      .getFileEvaluations()
+      .forall(_.isSuccessful)
+
+  private def getMessageErrorFromScalafix(
+      evaluation: ScalafixEvaluation
+  ): String = {
+    (if (!evaluation.isSuccessful)
+       evaluation.getErrorMessage().asScala
+     else
+       evaluation
+         .getFileEvaluations()
+         .headOption
+         .flatMap(_.getErrorMessage.asScala)).getOrElse(defaultErrorMessage)
   }
 
   private def scalafixConf: Option[Path] = {


metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -115,6 +115,23 @@ case class ScalafixProvider(
       }
     }
   }
+  private def scalafixSucceded(evaluation: ScalafixEvaluation): Boolean =
+    evaluation.isSuccessful && evaluation
+      .getFileEvaluations()
+      .headOption
```
ckipp01, line 121: I think you can just skip this `.headOption` since whether or not it's empty won't effect the `.forAll(_.isSuccessful)` below it.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -94,8 +94,8 @@ case class ScalafixProvider(
               exception
             )
             Future.failed(exception)
-          case Success(results) if !results.isSuccessful =>
-            val scalafixError = results.getErrorMessage().asScala
+          case Success(results) if !scalafixSucceded(results) =>
+            val scalafixError = getMessageErrorFromScalafix(results)
             val message = scalafixError.getOrElse(defaultErrorMessage)
```
ckipp01, line 99: Is this default actually still needed? Will `scalafixError` actually ever be `None` here? If I'm understanding the pr description correctly one of the possible errors that a file evaluation would fail would be 

> StaleSemanticdbError, MissingSemanticdbError.

And this is more or less what `defaultErrorMessage` is. So since we get the error messages for each file down below in `getMessageErrorFromScalafix` is this default still needed?

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -115,6 +115,23 @@ case class ScalafixProvider(
...
+  private def getMessageErrorFromScalafix(
+      evaluation: ScalafixEvaluation
+  ): Option[String] = {
+    if (!evaluation.isSuccessful)
+      evaluation.getErrorMessage().asScala
+    else
+      evaluation
+        .getFileEvaluations()
+        .headOption
+        .flatMap(_.getErrorMessage.asScala)
```
ckipp01, line 133: If what I have above is correct and we can rely on this always returning something, then we can simplify and just do
```suggestion
      evaluation
        .getFileEvaluations()
        .map(_.getErrorMessage.asScala)
```

But again, only if in reality if a file fails there will always be a message.

ckipp01 (changes requested):

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
mlachkar: No the defaut il not needed anymore.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
mlachkar: Yes. it's true. I m removing the default message

tgodzik: Honestly, I would keep it all as options, since we cannot assume that it's always correct are that there will never be a regression. 

Having the default message does not cost us anything and just keeps us safe.

mlachkar: I agree, it's safer to keep this option. As you like.

tgodzik: @ckipp01 what do you think?

ckipp01: Yea it's fine. Originally it was just a question on my end.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -115,6 +114,23 @@ case class ScalafixProvider(
...
+  ): String = {
+    if (!evaluation.isSuccessful)
+      evaluation.getErrorMessage().get
+    else
+      evaluation
+        .getFileEvaluations()
+        .head
+        .getErrorMessage
+        .get
+  }
```
ckipp01, line 133: Sorry if my review was confusing before, but yea, maybe to just avoid any potential issues in the future and be crystal we can probably just keep the default string as a backup, and also avoid any calls that can throw. Something along the lines of:

```suggestion
  private def getMessageErrorFromScalafix(
      evaluation: ScalafixEvaluation
  ): String = {
    (if (!evaluation.isSuccessful) {
      evaluation.getErrorMessage().asScala
    } else {
      val evaluations = evaluation
        .getFileEvaluations()
        .flatMap(_.getErrorMessage.asScala)
      if (evaluations.nonEmpty) Some(evaluations.mkString("\n"))
      else None
    }).getOrElse(default)
  }
```

Don't use that exact code 😆 just using it to show an example, but I think you get the idea. Just try to avoid eht `.head` and `.get`, still return a string, and still fallback to fallback if something goes wonky.

mlachkar: ok

ckipp01 (approved):
Looks great. Thanks a lot @mlachkar! LGTM

tgodzik (approved):
LGTM!

Issue #2425, opened 2021-01-22
"Debug test" cannot hit a break point defined in main source file


shaofengjiang:
**Describe the bug**

When I set a breakpoint in the main source file and run the unit test in the debugging mode (clicking "debug test"), the program only pauses in the unit test file and I cannot step into the source file. 

**To Reproduce** 

Attached bug.zip is a very simple project that can reproduce the issue, and the concrete steps go as follows.
[bug.zip](https://github.com/scalameta/metals-vscode/files/5858876/bug.zip)

1. Set a breakpoint at line 4 of myObject.scala.
2. Click "debug test" in myObject.test.scala.
3. The debugger pauses in myObject.test.scala instead of myObject.scala (where the breakpoint is set). Furthermore, in the call stack, myObject.doSomething(Seq) method is grayed, and it is marked as "Unknown Source" - this could be the reason that I cannot step into it.

**Expected behavior**

The debugger is expected to pause at the main source file where I set the breakpoint.

**Screenshots**

![image](https://user-images.githubusercontent.com/73434442/105553460-1c42d780-5d0e-11eb-9ce5-3b11b38f05b0.png)

**Installation:**

- Operating system: Windows 10 Pro 20H2
- VSCode version: 1.52.1
- VSCode extension version: 1.9.13
- Metals version: (found in VSCode settings, under `metals.serverVersion`) 0.9.10

**Additional context**

**Search terms**

debug, vscode, scala, breakpoint, cannot step into, unknown source

shaofengjiang:
Thanks for the prompt reply.

However, I’m afraid the issue is only partially resolved – the latest snapshot version works on the bug.zip project, but it still doesn’t work if I add the main source in a package.

Please check the following slightly modified version of the project that can reproduce the bug.
[bug1.zip](https://github.com/scalameta/metals/files/5860899/bug1.zip)

To reproduce the bug, simply set a breakpoint at line 4 of myObject.scala and "debug test" myObject.test.scala.
![image](https://user-images.githubusercontent.com/73434442/105611803-34732f00-5dc0-11eb-94f7-48d9be330f64.png)

tgodzik:
Ach, that is a related bug, fixing it here: https://github.com/scalameta/metals/pull/2427

shaofengjiang:
Thank you very much! The latest snapshot version indeed solves the problem in my example.

[...]

PR #2426, opened 2021-01-23
Prepend `_empty_/` when searching for symbol without a package

tgodzik:
This fixes https://github.com/scalameta/metals/issues/2425 which was caused by a very specific scenario trying to debug a class from test that has an empty package.

I tried to reproduce it in the tests, however it proved to be a bit difficult, since `QuickBuild` doesn't have `test` and main separate targets and to reproduce this we would need to rework the infrastructure.


ckipp01 (approved):
Sneaky one. LGTM!

PR #2427, opened 2021-01-24
Search also transitive sources when falling back from symbol search

tgodzik:
It seems in cases where we have `package object myObject {}` inside `myObject.scala` file the path given to us will be `myObject/myObject`, but the actual symbol is `myObject/package.`. This normally works with a standard  naming where package objects are put into `package.scala` file, so the path would be `myObject/package`.

As a fallback for symbol search we also check workspace sources for the file, but we haven't done it transitively, so files in dependency project would not be found. This fixes it so that the fallback works in those cases too.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathProvider.scala
@@ -28,7 +28,7 @@ final class SourcePathProvider(
 
     val symbolBase = if (base.contains("/")) base else "_empty_/" + base
     val symbols = for {
-      symbol <- Set(symbolBase + ".", symbolBase + "#")
+      symbol <- Set(symbolBase + ".", symbolBase + "#").toIterator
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathProvider.scala
```diff
@@ -28,7 +28,7 @@ final class SourcePathProvider(
 
     val symbolBase = if (base.contains("/")) base else "_empty_/" + base
     val symbols = for {
-      symbol <- Set(symbolBase + ".", symbolBase + "#")
+      symbol <- Set(symbolBase + ".", symbolBase + "#").toIterator
```
tgodzik, line 31: This is just so that we don't search for class if an object was already found.

ckipp01 (approved):
Looks good 👍🏼.

Total aside, I've never actually seen a package object used not in a `package.scala` before, so it makes sense we wouldn't have thought of this right away.

LGTM!

PR #2428, opened 2021-01-24
Ensure Bloop isn't offered as a server in an unsupported workspace.

### LINKED ISSUE #2424: `bsp-switch` offers Bloop as a choice even in a workspace that doesn't support Bloop
**Describe the bug**

When using the `bsp-switch` command it looks like Bloop is offered as a choice even in a workspace that doesn't support Bloop.

**To Reproduce**

1. Have a workspace with a build-tool that Bloop doesn't support
2. Also note that there would be another Build server in this situation that has an entry in `.bsp/xyz.json`
3. Execute a `bsp-switch`
4. See Bloop as an option

**Expected behavior**

In this situation where a user is using another BSP server _and_ the workspace is a workspace of a build tool that Metals doesn't know about, then Bloop shouldn't be an option.

**Additional context**

I'm pretty sure the issue is right here:

https://github.com/scalameta/metals/blob/94a8c9c39f5f675cac279051d561c829776ac3b4/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala#L127-L158

We should just add another check in here to see if the workspace is a build-tool that Bloop knows about. If it's not, then we shouldn't just automatically list Bloop in the set of options.

This was mentioned in #2420 

**Search terms**

BSP bloop unsupported build tool

ckipp01:
Previously if a user would issue a `bsp-switch` command and there was an
entry in `.bsp/` for a server, then no matter what, Bloop was offered as
a choice. However, there may be a situation where a workspace is a
workspace that does have a `.bsp/` entry, but the build tools is one
that Bloop doesn't work with. In this scenario, Bloop shouldn't be
offered since it doesn't know how to import from this build tool.

An example. Let's say we have build tool `xyz` and server `xyz` with a
corresponding `.bsp/xyz.json` file. If a user open Metals in this
workspace, it will auto-connect to `xyz`. If the user issues a
`bsp-switch`, they will just see a warning now that says:

> Unable to switch build server since there is only one supported build
> server 'xyz' detected for this workspace.

Fixes #2424

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
@@ -2,6 +2,7 @@ package scala.meta.internal.bsp
 
 import java.nio.file.Files
 
+import scala.collection.immutable.Nil
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 
@@ -125,35 +126,20 @@ class BspConnector(
   }
 
   private def askUser(
-      bspServerConnections: List[BspConnectionDetails],
+      availableBspConnections: List[BspConnectionDetails],
       currentBsp: Option[String]
   ): Future[BspResolvedResult] = {
-    val bloop = new BspConnectionDetails(
-      "Bloop",
-      ImmutableList.of(),
-      userConfig().currentBloopVersion,
-      "",
-      ImmutableList.of()
-    )
-
     val query = Messages.SelectBspServer.request(
-      bloop :: bspServerConnections,
+      availableBspConnections,
       currentBsp
     )
     for {
       item <- client.showMessageRequest(query.params).asScala
     } yield {
       val chosenMaybe = Option(item).flatMap(i => query.details.get(i.getTitle))
-      val result = chosenMaybe
-        .map { chosen =>
-          if (chosen == bloop) {
-            ResolvedBloop
-          } else {
-            ResolvedBspOne(chosen)
-          }
-        }
+      chosenMaybe
+        .map(BspResolvedResult.fromDetails)
         .getOrElse(ResolvedNone)
-      result
     }
   }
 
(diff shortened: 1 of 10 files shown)


metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -165,18 +151,37 @@ class BspConnector(
...
+              "",
+              ImmutableList.of()
+            ) :: multipleServers
+          else multipleServers
+        askUser(allPossibleConnections, currentSelectedServer).map {
           case ResolvedBloop
-              if currentBsp.contains(BspConnector.BLOOP_SELECTED) =>
+              if currentSelectedServer.contains(
+                BspConnector.BLOOP_SELECTED
+              ) =>
```
dos65, line 184: This pattern-match might be a little bit simplified by adding bloop option into the available server list before matching.
Smth like:
```scala
val availableServers = {
  val found = bspServers.findAvailableServers
  if (bloopIsAnOption) {
     new BspConnectionDetails(
          "Bloop",
           ImmutableList.of(),
           userConfig().currentBloopVersion,
           "",
           ImmutableList.of()
       ) :: found
  } else found
}
availableServers match {
   case Nil =>
        client.showMessage(BspSwitch.noInstalledServer)
        Future.successful(false)
    case singleServer :: Nil  =>
        client.showMessage(
          BspSwitch.onlyOneServer(name = singleServer.getName())
        )
        Future.successful(false)
     case multipleServers =>
        val currentSelectedServer = tables.buildServers.selectedServer()
        askUser(multipleServers, currentSelectedServer)
         ...
}
```

ckipp01: ah, great suggestion! Fixed and simplified.

metals/src/main/scala/scala/meta/internal/bsp/BspResolvedResult.scala
```diff
@@ -14,3 +14,11 @@ case class ResolvedBspOne(details: BspConnectionDetails)
     extends BspResolvedResult
 case class ResolvedMultiple(md5: String, details: List[BspConnectionDetails])
     extends BspResolvedResult
+
+object BspResolvedResult {
+  def fromDetails(details: BspConnectionDetails): BspResolvedResult =
+    details.getName().toLowerCase() match {
+      case "bloop" => ResolvedBloop
```
tgodzik, line 21:
```suggestion
      case BloopServers.name => ResolvedBloop
```
might be better to have a single place where this is defined.

tgodzik (changes requested):
Thanks a lot  for taking care of it! Left some smaller comments.

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -2,6 +2,7 @@ package scala.meta.internal.bsp
 
 import java.nio.file.Files
 
+import scala.collection.immutable.Nil
```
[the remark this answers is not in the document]
ckipp01: Huh interesting. I imported it earlier to use it, but then ended up not needing it. However, I'm surprised this wasn't removed with Scalafix. I wonder if it's because we use `Nil` in the pattern match?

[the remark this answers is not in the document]
tgodzik: I think the symbol is used, but also is in Predef, so usually it's just implictely imported.

metals/src/main/scala/scala/meta/internal/bsp/BspResolvedResult.scala
ckipp01: Yea good idea. I added this in and replaced the usage of "Bloop" with it in various places.

tgodzik (approved):
LGTM!

PR #2433, opened 2021-01-25
Update editor overview doc (sublime)

ayoub-benali:
Just like VSCode organize imports it is available under LSP -> source action


ckipp01 (approved):
Looks good. Thanks @ayoub-benali! LGTM!

PR #2434, opened 2021-01-25
Fix issue with breakpoints in multiple files

tgodzik:
Breakpoints need to be registered for each class file separately, so Metals takes care of translating source file locations to separate class file names displayed as for example `dap-fqcn:packageName.objectName$`.

Breakpoints for each file are send from the DAP client (like for example VS Code) separately and because of that we would remove breakpoints for the previously send files upon receiving another breakpoint request. Now, it's changed, so that specific breakpoints for classfiles are remembered separately for each file. This helps to to recognize when an empty request needs to be sent for a certain classfile.

This one was a bit tougher, but thanks for reporting @shaofengjiang!


dos65 (approved):
LGTM

ckipp01 (approved):
Good job tackling all these debug issues! LGTM!

shaofengjiang:
Yes, I tested, and it indeed fixes the issue! Wonderful work!!

PR #2436, opened 2021-01-26
Remove MunitReportPlugin

tgodzik:
It seems we are hitting on:
`com.google.cloud.storage.StorageException: The project to be billed is associated with a closed billing account.`
so porbably some issue with the storage for the reports. We can remove it for now I guess and fix/setup something later.

@olafurpg any idea what is going on?

### DIFF
--- a/docs/contributors/tests.md
+++ b/docs/contributors/tests.md
@@ -5,6 +5,7 @@ title: Tests overview
 
 The following table is an overview of the test suites in the Metals codebase.
 
+<!-- 
 ```scala mdoc:munit
 
-```
+``` -->
--- a/build.sbt
+++ b/build.sbt
@@ -700,4 +700,4 @@ lazy val docs = project
     )
   )
   .dependsOn(metals)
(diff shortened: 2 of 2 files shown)


docs/contributors/tests.md
```diff
@@ -5,6 +5,7 @@ title: Tests overview
 
 The following table is an overview of the test suites in the Metals codebase.
```
olafurpg, line 6: May want to remove this sentence, or the page altogether.

olafurpg (approved):
LGTM 👍 I got some emails about an expired GCP trial, which I'm guessing is related. You can fix this by creating new credentials following the instructions here https://scalameta.org/munit/docs/reports.html#installation

tgodzik:
> LGTM  I got some emails about an expired GCP trial, which I'm guessing is related. You can fix this by creating new credentials following the instructions here https://scalameta.org/munit/docs/reports.html#installation

I started a trial in google cloud and setup the json. Let's see if it fixes it. I can take a look later and see how to make it more pernament.

PR #2438, opened 2021-01-26
Minor fix: remove usused parameter

dos65:
A small addition to #2412


ckipp01 (approved):
Thanks for the follow-up! LGTM

PR #2440, opened 2021-01-27
Change bucket name to the non default one

tgodzik:


olafurpg (approved):
LGTM 👍

build.sbt
```diff
@@ -693,6 +693,7 @@ lazy val docs = project
     sharedSettings,
     skip.in(publish) := true,
     moduleName := "metals-docs",
+    munitBucketName := Some("scala-metals-test-reports"),
```
olafurpg, line 696: You may want to leave a comment on what Google account owns this bucket to help the next person investigate the failing CI 😅

tgodzik: I think I should be able to give more people access to it. I will try to do that later if it works, but I added a comment for now.

PR #2443, opened 2021-01-27
Make `munitBucketName` a global setting

tgodzik:
So it seems with the previous change, the setting only changed for docs, which caused other projects to have the default name. This should finally fix it.


tgodzik:
I will go ahead and merge to see if this fixes the issue.

PR #2444, opened 2021-01-27
Update docs on system-properties and inline decorations.

### LINKED ISSUE #2439: Misleading docs about system properties
**Describe the bug**

It used to be that all of the new configuration added to metals also had a system property option. You can see them all on [the website here](https://scalameta.org/metals/docs/integrations/new-editor.html#metals-server-properties). However, new settings that are being added in `UserConfiguration` is still being pulled in and server property documentation for it is being made and published even though those options aren't valid. This can cause confusion. Sorry @ayoub-benali .

The issue is here:

https://github.com/scalameta/metals/blob/eed9660b66d58285106d51316a94fc8f0114902a/metals-docs/src/main/scala/docs/UserConfigurationModifier.scala#L19-L31

We should instead pull those from `MetalsServerConfig`.

**Search terms**

system properties

ckipp01:
This removes the generated system-properties fields for new user configuration
settings. The reason for this is that new user configuration settings aren't
actually available as system properties anymore.

This also adds in a small section about enabling inline decorations to the
decoration protocol page.

Fixes #2439

### DIFF
--- a/metals-docs/src/main/scala/docs/UserConfigurationModifier.scala
+++ b/metals-docs/src/main/scala/docs/UserConfigurationModifier.scala
@@ -15,20 +15,6 @@ class UserConfigurationModifier extends StringModifier {
       reporter: Reporter
   ): String = {
     info match {
-      case "system-property" =>
-        UserConfiguration.options
-          .map { option =>
-            s"""
-               |### `-Dmetals.${option.key}`
-               |
-               |${option.description}
-               |
-               |**Default**: ${option.default}
-               |
-               |**Note**: this property can also be defined as user configuration option [${option.title}](#${option.headerID}).
-               |""".stripMargin
-          }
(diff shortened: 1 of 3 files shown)


metals-docs/src/main/scala/docs/UserConfigurationModifier.scala
```diff
@@ -15,20 +15,6 @@ class UserConfigurationModifier extends StringModifier {
       reporter: Reporter
   ): String = {
     info match {
-      case "system-property" =>
```
ckipp01, line 18: Note that the system properties that are still valid are still in the docs, they were hardcoded there. So this will only be removing the `UserConfiguration` generated system properties which are for almost all the cases not valid.

tgodzik (approved):
LGTM!

PR #2446, opened 2021-01-28
Add support for type decorations in for comprehensions and patterns

tgodzik:
Totally forgot about adding those and turns out it's not that hard to do. `CaseGenerator` is coming from Dotty and will not work for a while, but should start once we get symbol information inside semanticdb generated by the Scala 3 compiler.

If anyone has any ideas, what we might be missing still, let me know!


ckipp01 (approved):
Ah man, this would be great. Anxiously waiting for Neovim to hurry up and get inline virtual text 😆 

This looks good. LGTM!

PR #2447, opened 2021-01-29
Minor adjustements to type decorations

tgodzik:
-  added type decorations for function params
-  remove type decoration from bound patterns, which can cause duplication:

`a @ T.Comment()` would show `a: T.Comment @ T.Comment()`


tests/unit/src/test/scala/tests/decorations/SyntheticDecorationsLspSuite.scala
```diff
@@ -408,7 +410,7 @@ class SyntheticDecorationsLspSuite extends BaseLspSuite("implicits") {
            |  val List(l1: Int, l2: Int) = List(12, 13)
            |  println("Hello!")
            |  val abc: Int = 123
-           |  val tupleBound: (String, String) @ (one: String, two: String) = ("1", "2")
```
tgodzik, line 411: This is really duplicated information, so I think it's not beneficial to show it. In most cases more details can be seen on the right.

dos65 (approved):
Looks good!

PR #2448, opened 2021-01-29
Add support for new Scala 3 syntax in document outline.

dos65:


metals/src/main/scala/scala/meta/internal/parsing/DocumentSymbolProvider.scala
```diff
@@ -250,6 +250,73 @@ class DocumentSymbolProvider(trees: Trees) {
...
+                    (inits.map(_.tpe.syntax).mkString(" with "), inits.head.pos)
+                }
+              case name => (name.value, name.pos)
+            }
+          addChild(
+            name,
+            SymbolKind.Class,
+            t.pos,
+            pos,
+            ""
```
tgodzik, line 317: Let's put the type here always as details if name is specified? Or always just keep it here and have name empty in case of anonymous givens?

tests/input3/src/main/scala/example/Extension.scala
```diff
@@ -0,0 +1,4 @@
+package example
+
+extension (i: Int)
+  def asString: String = i.toString
```
tgodzik, line 4: You can also add a test for extension group with multiple params using braces.

build.sbt
```diff
@@ -508,6 +508,15 @@ lazy val input = project
   )
   .disablePlugins(ScalafixPlugin)
 
+lazy val input3 = project
+  .in(file("tests/input3"))
+  .settings(
+    sharedSettings,
+    scalaVersion := V.scala3,
```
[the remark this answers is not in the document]
tgodzik: If it's too complicated it's okay to leave it separate.

tgodzik (changes requested):
Looks pretty neat! I left some suggestions to maybe make a bit nicer in the outline.

> mutiline extension
>  transparent traits (fixed in scalameta/scalameta#2232)

These don't need to be included here, they do impact only the parser not the AST itself.

build.sbt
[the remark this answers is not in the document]
dos65: Yep, that will be hard. It requires to do smth like in `crossPublishLocal`.

Also, there is an option not to use `input` as a project.
The only advantage it provides is that before running `unit` it checks that `input` compiles.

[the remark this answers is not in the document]
tgodzik: Ok, I think it's fine to leave it as is then. It looks pretty sensible currently.

metals/src/main/scala/scala/meta/internal/parsing/DocumentSymbolProvider.scala
```diff
@@ -250,6 +250,73 @@ class DocumentSymbolProvider(trees: Trees) {
...
+            t.eparam.decltpe.fold(("", t.pos))(tpe => (tpe.syntax, tpe.pos))
+          addChild(
+            s"extension $name",
+            SymbolKind.Class,
+            t.pos,
+            pos,
+            ""
+          )
+          newOwner()
+        case t: Defn.GivenAlias =>
```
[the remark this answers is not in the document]
dos65: It's handled below

tests/input3/src/main/scala/example/Extension.scala
dos65: Done

metals/src/main/scala/scala/meta/internal/parsing/DocumentSymbolProvider.scala
dos65: Added details. I don't think that there is a need to keep an empty name for anonymous instances.
Attached screenshot of how it looks now:
Screenshot from 2021-02-05 20-11-57

tgodzik: It looks a bit weerd to have `Ord[String] Ord[String` - we should not duplicate information unless absolutely needed.

dos65: That's fun because I've added it to show that it isn't weird at all =)
Ok, changed!

metals/src/main/scala/scala/meta/internal/parsing/DocumentSymbolProvider.scala
```diff
@@ -250,6 +250,75 @@ class DocumentSymbolProvider(trees: Trees) {
...
+          val (tpeName, initPos) =
+            t.templ.inits match {
+              case Nil => ("(anonymous)", t.pos)
+              case inits =>
+                (inits.map(_.tpe.syntax).mkString(" with "), inits.head.pos)
+            }
+
+          val (name, pos) =
+            t.name match {
+              case Name.Anonymous() => (" ", initPos)
```
dos65, line 311: I've tried to pass an empty string as a name but it breaks smth.

There are no exceptions in log and vscode displays as `No symbols found in document $filename`

metals/src/main/scala/scala/meta/internal/parsing/DocumentSymbolProvider.scala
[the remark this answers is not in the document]
tgodzik: That's `Defn.Given` not `Decl.Given` - it's a bit confusing, but one is for given definitions and the other for abstract givens.

metals/src/main/scala/scala/meta/internal/parsing/DocumentSymbolProvider.scala
tgodzik: Maybe then just put `tpeName` into name and leave the detail empty?

tgodzik (changes requested):
Some related tests seem to be failing in DocumentSymbolScala3Suite.

dos65:
Fixed

tgodzik (approved):

PR #2449, opened 2021-01-29
Fix printed inferred type for complex types

tgodzik:
While browsing Metals with the new feature I came upon some issues in non-obvious cases. I try to fix them all here:

- existential types are not that useful to show in the type decorations, so I changed to show the base type only. It happens in a situation like:
```scala
   val job = ec.submit(new Runnable {
      override def run(): Unit = {
        launcher.runLauncher(
          bloopVersion,
          skipBspConnection = false,
          Nil
        )
        finished.success(())
      }
    })
```
aside from existential type in the above example we also had an issue with a locally defined type. Instead of `local123` we now just print `_`, which is also returned by normal hover.

- structural types now show with `...` only if there is something contained in them
- with types like `new A with B {}` no longer show `AnyRef` as first type

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
+++ b/metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
@@ -10,7 +10,8 @@ class SemanticdbTreePrinter(
     isHover: Boolean,
     printSymbol: String => String,
     createSymtab: => PrinterSymtab,
-    rightArrow: String
+    rightArrow: String,
+    ellipsis: String
 ) {
 
   lazy val symtab = createSymtab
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
```diff
@@ -48,20 +56,27 @@ class SemanticdbTreePrinter(
...
+        if (isHover) {
           Print.tpe(Format.Detailed, t, symtab)
-        else s" ${printType(tpe)} { ... }"
+        } else {
+          s"${printType(tpe)} {${printScope(scope)}}"
+        }
     }
 
+  def printScope(scope: Option[s.Scope]): String = {
+    if (scope.exists(_.hardlinks.nonEmpty)) "..." else ""
```
ckipp01, line 77: Just a nit, but did you want a `...` or `…` here for this?

ckipp01 (approved):
Apart from the small comment (which is up to you) LGTM!

tests/unit/src/test/scala/tests/decorations/SyntheticDecorationsLspSuite.scala
```diff
@@ -443,6 +443,66 @@ class SyntheticDecorationsLspSuite extends BaseLspSuite("implicits") {
...
+           |
+           |class Main(implicit ec: ExecutionContextExecutorService){
+           |  // structural types
+           |  val anon1 = new A {}
+           |  val anon2 = new A { def a = 123 }
+           |  val anon3 = new A with B {}
+           |  // existential type
+           |  val job = ec.submit(new Runnable {
+           |     override def run(): Unit = {}
+           |  })
```
dos65, line 469: A minor addition.

I've tried the following sample:
```scala
val runnable = new Runnable {
  override def run(): Unit = {}
}
```
And got printed type:
```scala
val runnable: Object with Runnable {} = new Runnable {
   override def run(): Unit = {}
}
```

Probably in this case, `Object` should be dropped in the same way as it's done for `AnyRef` in `WithType`

dos65 (approved):

tests/unit/src/test/scala/tests/decorations/SyntheticDecorationsLspSuite.scala
tgodzik: Good catch, fixed it also!

metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
tgodzik: I actually intended to change it :sweat_smile:  Changed to using ellpsis when applicable.

Issue #2233, opened 2020-11-16
Step in and step over does not work reliably


tgodzik:
**Describe the bug**

It seems often reported by users that step in/out feature is not working currently 100% of time. The debugger is implemented in bloop and is using the microsoft debug server. Not sure how much is a problem inside Bloop and how much is coming from the debugger library.

Reported originally on gitter by @bjornregnell 

**To Reproduce**

Some user reports:
```
Sometimes (but I'm not sure how to reproduce it consistently) the step-in did not work and the stack trace did not show anything if breakpoints are set e.g. in the println in many. I'll see if I can find a way to consistently reproduce during the weekend. Is a link to a video ok in an issue? It is difficult to report GUI-related stuff in pure text without the live action :) ? @tgodzik

I made a video for my students on how to get the debugger with metals going for 2.13.3 and I ran into all sorts of issues such as debugger going crazy in stepping etc. The video is unfortunately in Swedish but I think maybe the failures are still obvious as you can see where I click and what does not happen. Some of the issues are probably not easy to consistently repeat, but maybe this is still valuable to the community as a kind of interactive user experience test. This was run on an x86 box with Ubuntu 18.04 and a rather fast CPU i7-4790K with 8 threads and 4GHz, using VS Code 1.50.1, Scala 2.13.3, sbt 1.4.2, Metals 1.9.6. Here is the link to the video with some strange Metals/bloop behavior (in Swedish): https://www.youtube.com/watch?v=UKLvd_V36bU&t=256s
Let me know if I can contribute with more details of what happened when the debugger step-into etc did not work.
Especially here at time stamp 32:30 the debugging fails when stepping in and out: https://www.youtube.com/watch?v=UKLvd_V36bU&t=1950s

```

based on code sample:

```scala
object Main {
  def many(s: String, n: Int) = {
    println("I'm now in many")
    s * n
  }

  def greet(who: String, n: Int = 42) = 
    many(s"Hello $who! \n", n)

  def main(args: Array[String]): Unit = {
    val mrPresidentElect = "Joe"
    val msg = greet(mrPresidentElect)
    println(msg)
  }
}
```

**Expected behavior**

Step in/out feature works reliably.

**Screenshots**

https://www.youtube.com/watch?v=UKLvd_V36bU&t=1950s

**Installation:**

- Operating system: macOS, but mostly liekly other systems too
- Editor: Visual Studio Code
- Metals version: v0.9.5

**Search terms**

debugging step in out

[...]

PR #2450, opened 2021-01-30
Wait for compilation to finish before starting debug session #2233

pvid:
Finally following up on  #2233.

I was not able to reproduce the issue but I agree with @tgodzik that it might be a race condition, that this MR wants to solve

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
@@ -47,15 +47,16 @@ final class Compilations(
   def wasPreviouslyCompiled(buildTarget: b.BuildTargetIdentifier): Boolean =
     lastCompile.contains(buildTarget)
 
-  def compilationFinished(
-      source: AbsolutePath
-  ): Future[Unit] = {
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
```diff
@@ -94,14 +95,17 @@ final class Compilations(
     } yield ()
   }
 
-  def cascadeCompileFiles(paths: Seq[AbsolutePath]): Future[Unit] = {
-    val targets =
-      expand(paths).flatMap(buildTargets.inverseDependencyLeaves).distinct
+  def cascadeCompile(targets: Seq[BuildTargetIdentifier]): Future[Unit] = {
+    val inverseDependencyLeaves =
```
pvid, line 99: I am not really sure about this - should it not compile all inverse dependencies and not just leaves?

tgodzik: Leaves will trigger also the compilation of their dependencies, so we should be good.

tgodzik (approved):
LGTM! Thanks for following  up on this!

PR #2464, opened 2021-02-01
Remove dependency on io.methvin:directory-watcher.

### LINKED ISSUE #2365: Remove directory-watcher
**Describe the bug**

We are no longer actively using directory watcher, but instead we are using `swoval` and just reusing some of the existing classes. We should just get rid of the dependency since we can easily just create those classes inside Metals. 

**Additional context**

`swoval` was added here https://github.com/scalameta/metals/pull/2014

we are still getting update PRs, which only contribute to noise - https://github.com/scalameta/metals/pull/2356

**Search terms**

swoval file watching directory watcher

ckipp01:
As of https://github.com/scalameta/metals/pull/2014 we replaced the
directory watching functionality with that of Swoval. However, we do
re-use some of the functionality of directory-watcher, mainly the
DirectoryChangeEvent and some hashing utils that it utilizes. This
change removes the full directory-watcher dependency but copies over the
few necessary classes from that library into a
scala.meta.internal.watcher package.

Closes #2365

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -393,7 +393,6 @@ lazy val metals = project
       // for measuring memory footprint
       "org.openjdk.jol" % "jol-core" % "0.14",
       // for file watching
-      "io.methvin" % "directory-watcher" % "0.10.1",
       "com.swoval" % "file-tree-views" % "2.1.6",
       // for http client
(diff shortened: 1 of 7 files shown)


tgodzik (approved):
LGTM! Thanks for taking care of this!

PR #2465, opened 2021-02-01
Pass URI to `semanticdbTextDocument` method and fix possible issues

tgodzik:
Previously we would use string for `PresentationCompiler.semanticdbTextDocument` method, which is not neccessary as we need to turn it back into URI anyway. This seems to also cause possibly cause the issues with:
```
Feb 01, 2021 11:32:01 AM scala.meta.internal.pc.CompilerAccess handleError
SEVERE: null
java.nio.file.ProviderMismatchException
        at sun.nio.fs.UnixPath.toUnixPath(UnixPath.java:200)
        at sun.nio.fs.UnixPath.relativize(UnixPath.java:410)
        at sun.nio.fs.UnixPath.relativize(UnixPath.java:43)
        at scala.meta.internal.pc.SemanticdbTextDocumentProvider.$anonfun$textDocument$2(SemanticdbTextDocumentProvider.scala:38)
        at scala.Option.map(Option.scala:230)
        at scala.meta.internal.pc.SemanticdbTextDocumentProvider.textDocument(SemanticdbTextDocumentProvider.scala:37)
```
I also added a Try there just to be on the safe side and additional check when calculating semanticdb in InteractiveSemanticdb.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
@@ -54,15 +54,19 @@ final class InteractiveSemanticdbs(
 
   override def textDocument(source: AbsolutePath): TextDocumentLookup = {
 
+    def doesNotBelongToBuildTarget = buildTargets.inverseSources(source).isEmpty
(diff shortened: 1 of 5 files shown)


ckipp01 (approved):
Looks good! LGTM!

PR #2467, opened 2021-02-01
Fix issues with references in standalone files

tgodzik:
This fixes a number of issues:
- we would throw an exception when interactive semanticdb passed a Scala file path to ReferencesProvider
- no updates on sbt files would get indexed

Now instead of throwing an exception we just forward the scala path and additionally the current workflow was changed to:
| standalone file change -> create a text document in InteractiveSemanticdb -> if succesfull pass information to semanticdbIndexer -> pass information to all the need text documents

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
@@ -1,7 +1,6 @@
 package scala.meta.internal.decorations
 
 import java.nio.charset.Charset
-import java.nio.file.Path
 import java.util.concurrent.atomic.AtomicReference
 
 import scala.collection.mutable
@@ -18,7 +17,6 @@ import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.UserConfiguration
 import scala.meta.internal.metap.PrinterSymtab
 import scala.meta.internal.mtags.Md5Fingerprints
-import scala.meta.internal.mtags.SemanticdbClasspath
 import scala.meta.internal.mtags.Semanticdbs
 import scala.meta.internal.parsing.TokenEditDistance
 import scala.meta.internal.parsing.Trees
@@ -85,21 +83,16 @@ final class SyntheticsDecorationProvider(
       Future.successful(())
   }
 
-  def onChange(textDocument: TextDocuments, path: Path): Unit = {
+  def onChange(textDocument: TextDocuments, path: AbsolutePath): Unit = {
     for {
       focused <- focusedDocument()
-      filePath = AbsolutePath(path)
-      sourcePath <-
-        if (!filePath.isScalaFilename)
-          SemanticdbClasspath.toScala(workspace, filePath)
-        else Some(filePath)
-      if sourcePath == focused || !clientConfig.isDidFocusProvider()
+      if path == focused || !clientConfig.isDidFocusProvider()
       textDoc <- textDocument.documents.headOption
-      source <- fingerprints.loadLastValid(sourcePath, textDoc.md5, charset)
+      source <- fingerprints.loadLastValid(path, textDoc.md5, charset)
     } {
       val docWithText = textDoc.withText(source)
       Document.set(docWithText)
-      publishSyntheticDecorations(sourcePath, docWithText)
+      publishSyntheticDecorations(path, docWithText)
     }
   }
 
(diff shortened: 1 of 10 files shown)


metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
```diff
@@ -89,10 +89,7 @@ final class SyntheticsDecorationProvider(
     for {
       focused <- focusedDocument()
       filePath = AbsolutePath(path)
-      sourcePath <-
-        if (!filePath.isScalaFilename)
-          SemanticdbClasspath.toScala(workspace, filePath)
-        else Some(filePath)
+      sourcePath <- SemanticdbClasspath.toScala(workspace, filePath)
```
tgodzik, line 92: This is now handled inside `toScala` - we don't have `.semanticdb` file for the generated text documents and I think  it's not a good idea to create an artificial one.

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
```diff
@@ -81,20 +80,12 @@ class SemanticdbIndexer(
     }
   }
 
-  def onStandaloneFilesChange(path: AbsolutePath): Unit = {
-    if (!path.isWorksheet && buildTargets.inverseSources(path).isEmpty) {
```
tgodzik, line 85: This was duplicated with the logic in InteractiveSemanticdb

tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala
```diff
@@ -143,9 +143,8 @@ class WorkspaceSymbolLspSuite extends BaseLspSuite("workspace-symbol") {
       )
       _ <- server.didOpen("a/src/main/scala/a/A.scala")
       _ = server.workspaceSymbol("scala.None")
-      option = ".metals/readonly/scala/Option.scala"
-      _ <- server.didOpen(option)
-      references <- server.references(option, "object None")
+      _ <- server.didOpen(".metals/readonly/scala/Option.scala")
+      references <- server.references("a/src/main/scala/a/A.scala", " None")
```
tgodzik, line 147: So actually, I think this was testing the wrong thing. We should not find references in readonly files when looking in files in the workspace, but it should find them in the file itself.

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -196,7 +196,7 @@ final class ReferenceProvider(
       includeSynthetics: Synthetic => Boolean
   ): Seq[Location] = {
     val isSymbol = alternatives + occ.symbol
-    if (occ.symbol.isLocal) {
+    if (occ.symbol.isLocal || source.isDependencySource(workspace)) {
```
tgodzik, line 199: This will make sure that when searching for references in dependency sources it will search the current file despite those sources not being indexed.

tests/unit/src/test/scala/tests/decorations/SyntheticDecorationsLspSuite.scala
```diff
@@ -297,28 +297,30 @@ class SyntheticDecorationsLspSuite extends BaseLspSuite("implicits") {
...
            |""".stripMargin
       )
       _ <- server.didOpen("standalone/Main.scala")
       _ = assertNoDiagnostics()
       _ = assertNoDiff(
+        // currently interactive semanticdb doesn't produce symbol signatures
         client.workspaceDecorations,
         """|object Main{
-           |  augmentString("asd.").stripSuffix(".")
+           |  val value = augmentString("asd.").stripSuffix(".")
```
tgodzik, line 317: I discovered that symbol signatures are not produced by interactive semanticdb :/

ckipp01 (approved):
Looks good! LGTM!

metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
dos65: I've just realized that this logic is a little bit clumsy =)

This method might receive two different types of paths: source file or sematicdb document.
As I see, both call sites are located in `SemanticdbIndexer`: 
- [onChange](https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala#L100) (passes semanticdb)
- [onStandaloneFileChange](https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala#L100) (passes source)

I think it might be better to change `SemanticdbIndexer.onChange` to make it pass the source location and do not change `SemanticdbClasspath.toScala`.
wdyt?

tgodzik: Good idea, changed!

dos65 (approved):
:rocket:

PR #2470, opened 2021-02-02
Add code lens to easily discover copy worksheet command

tgodzik:
Previously, only way to discover the copy worksheet command would be to run it manually, which is always an additional barier. Now, it's possible to use the code lense that is shown at the top of the file.

This is how it looks in VS Code:

metals-sample-1612289058641

What worries me is that we rely on the name of the command in the client to be `metals` + the actual command id. Otherwise, the server command might be invoked without handling it properly. It's kind of implicit requirement for the clients and there is no way to discover if a client actually contributes it. I have no idea currently how to work around it :/

### DIFF
--- a/docs/integrations/new-editor.md
+++ b/docs/integrations/new-editor.md
@@ -112,6 +112,7 @@ The currently available settings for `InitializationOptions` are listed below.
         "parameterHintsCommand": string,
         "snippetAutoIndent": boolean,
       }
+      "copyWorksheetOutputProvider": boolean,
       "debuggingProvider": boolean,
       "decorationProvider": boolean,
       "didFocusProvider": boolean,
@@ -124,12 +125,12 @@ The currently available settings for `InitializationOptions` are listed below.
       "isExitOnShutdown" : boolean,
       "isHttpEnabled": boolean,
       "openFilesOnRenameProvider": boolean,
+      "openNewWindowProvider": boolean,
       "quickPickProvider": boolean,
       "renameFileThreshold": number,
       "slowTaskProvider": boolean,
       "statusBarProvider": "on" | "off" | "show-message" | "log-message",
       "treeViewProvider": boolean
-      "openNewWindowProvider": boolean
     }
 ```
 
@@ -222,6 +223,14 @@ Possible values:
   case for VS Code, Sublime, and coc.nvim.
 - `off`: the client does not add any indentation when receiving a multi-line
   textEdit
+
+##### `copyWorksheetOutputProvider`
+
+Boolean value signifying whether or not the client supports running
+CopyWorksheetOutput server command and copying it's results into the local
+buffer.
+
+Default value: `false`
 
 ##### `debuggingProvider`
 
@@ -334,6 +343,14 @@ rename.
 
 Default: `false`
 
+##### `openNewWindowProvider`
+
+Boolean value signifying whether or not the client supports opening up a new
+window with the newly created project. Used in conjunction with the New Project
+Provider.
+
+Default value: `false`
+
 ##### `quickPickProvider`
 
 Boolean value to signify whether or not the client implements the
@@ -373,14 +390,6 @@ Possible values:
 
 Boolean value signifying whether or not the client supports the
 [Tree View Protocol](../integrations/tree-view-protocol.md).
-
-Default value: `false`
-
-##### `openNewWindowProvider`
-
-Boolean value signifying whether or not the client supports opening up a new
-window with the newly created project. Used in conjunction with the New Project
-Provider.
 
 Default value: `false`
 
--- a/metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
@@ -193,6 +193,19 @@ object ClientCommands {
        |""".stripMargin
   )
 
+  val CopyWorksheetOutput = new Command(
+    "metals.copy-worksheet-output",
+    "Copy Worksheet Output",
+    s"""|Copy the contents of a worksheet to your local buffer.
+        |
+        |Note: This command should execute the ${ServerCommands.CopyWorksheetOutput.id} 
+        |      server command to get the output to copy into the buffer.
+        |
+        |Server will attempt to create code lens with this command if `copyWorksheetOutputProvider` option is set.
+        |""".stripMargin,
+    "[uri], the uri of the worksheet that you'd like to copy the contents of."
+  )
+
   def all: List[Command] =
     List(
       OpenFolder,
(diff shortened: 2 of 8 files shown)


ckipp01:
> What worries me is that we rely on the name of the command in the client to be metals + the actual command id. Otherwise, the server command might be invoked without handling it properly.

🤔 I'm not sure I understand. What do you mean by it might be invoked without handling it properly?


> It's kind of implicit requirement for the clients and there is no way to discover if a client actually contributes it. I have no idea currently how to work around it :/

So it's implicit that it's always done, but technically not needed right? When the command comes in, we just strip it out anyways.

😬 sorry I'm totally not following what the problem is

tgodzik:
> > What worries me is that we rely on the name of the command in the client to be metals + the actual command id. Otherwise, the server command might be invoked without handling it properly.
> 
>  I'm not sure I understand. What do you mean by it might be invoked without handling it properly?
> 

So we register the command in VS Code as `metals.copy-worksheet-output`, while the actual name of the command is `copy-worksheet-output`. This is why it's handled on the client side, because we do have an actual client command that is prefixed with `metals.`.  If we didn't add the prefix, the command would be just forwarded to the server. So in this case it's actually a client command, not a server one. We rely on the knowledge that there are client commands that are named the same as the server ones, but prefix with `metals.`  There really might not be a way to do it differently, so maybe I am just overthinking here :thinking:  

> > It's kind of implicit requirement for the clients and there is no way to discover if a client actually contributes it. I have no idea currently how to work around it :/
> 
> So it's implicit that it's always done, but technically not needed right? When the command comes in, we just strip it out anyways.
> 
>  sorry I'm totally not following what the problem is

This might not actually be a problem. I forgot we strip `metals.` from the command anyway.

I need to think about this all a bit, but this is not a problem of this particular PR in reality.

ckipp01 (changes requested):

ckipp01:
> If we didn't add the prefix, the command would be just forwarded to the server. So in this case it's actually a client command, not a server one.

Isn't this a server command no matter what though? Like no matter how it's triggered isn't it just a `workspace/executeCommand` being sent to Metals from the client?

metals/src/main/scala/scala/meta/internal/metals/codelenses/WorksheetCodeLense.scala
```diff
@@ -0,0 +1,29 @@
+package scala.meta.internal.metals.codelenses
+
+import scala.meta.internal.metals.MetalsEnrichments._
+import org.eclipse.{lsp4j => l}
+import scala.meta.internal.metals.ServerCommands.CopyWorksheetOutput
+import scala.meta.io.AbsolutePath
+
+object WorksheetCodeLense extends CodeLens {
+
+  override def isEnabled: Boolean = true
```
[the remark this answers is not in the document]
ckipp01: I really don't think it's necessary to. Maybe if someone complains we can consider it, but I'd rather just check if the client support code lenses, and if so, enable it.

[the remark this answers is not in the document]
tgodzik: If a client does not support code lenses it will not send the `textDocument/codeLens` request, so we are safe on this side.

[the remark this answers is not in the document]
ckipp01: ahh duh... of course...

tgodzik:
> > If we didn't add the prefix, the command would be just forwarded to the server. So in this case it's actually a client command, not a server one.
> 
> Isn't this a server command no matter what though? Like no matter how it's triggered isn't it just a `workspace/executeCommand` being sent to Metals from the client?

It's not actually, this might also be a client command that is not implemented in Metals at all. And in this case we need a client command to copy things into the buffer. I am thinking that ideally we should have an option for it if we add it in code lense, otherwise people will just get code lense that will do nothing. What do you think?

ckipp01:
> > > If we didn't add the prefix, the command would be just forwarded to the server. So in this case it's actually a client command, not a server one.
> > 
> > 
> > Isn't this a server command no matter what though? Like no matter how it's triggered isn't it just a `workspace/executeCommand` being sent to Metals from the client?
> 
> It's not actually, this might also be a client command that is not implemented in Metals at all. And in this case we need a client command to copy things into the buffer. I am thinking that ideally we should have an option for it if we add it in code lense, otherwise people will just get code lense that will do nothing. What do you think?

Ahh, yea you're totally right. Hmmm, so basically what do we do with a client that supports code lenses, but may not have the `metals.copy-worksheet-output` implemented. 🤔 since yea, they will get the code lens, click it, and it will do nothing.

tgodzik:
> Ahh, yea you're totally right. Hmmm, so basically what do we do with a client that supports code lenses, but may not have the `metals.copy-worksheet-output` implemented.  since yea, they will get the code lens, click it, and it will do nothing.

Ok, I think it should be fine now, but I will take a look at it with a fresh head tomorrow and also make sure it's all documented.

It was way more complicated than I anticipated haha.

tgodzik:
Added a corresponding PR here: https://github.com/scalameta/metals-languageclient/pull/236

docs/integrations/new-editor.md
```diff
@@ -128,8 +128,9 @@ The currently available settings for `InitializationOptions` are listed below.
       "renameFileThreshold": number,
       "slowTaskProvider": boolean,
       "statusBarProvider": "on" | "off" | "show-message" | "log-message",
-      "treeViewProvider": boolean
-      "openNewWindowProvider": boolean
+      "treeViewProvider": boolean,
+      "openNewWindowProvider": boolean,
+      "copyWorksheetOutputProvider": boolean
```
ckipp01, line 133: _Extremely nit picky_ 😬 but I put can we move the new value up to be in alphabetical order? It helps when easily scanning all the init options in the docs.

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -193,6 +193,19 @@ object ClientCommands {
        |""".stripMargin
   )
 
+  val CopyWorksheetOutput = new Command(
+    "metals.copy-worksheet-output",
+    "Copy Worksheet Output",
+    s"""|Copy the contents of a worksheet to your local buffer.
+        |
+        |Note: This command should execute ${ServerCommands.CopyWorksheetOutput.id}
```
ckipp01, line 201:
```suggestion
        |Note: This command should execute the ${ServerCommands.CopyWorksheetOutput.id} 
```

ckipp01 (approved):
Just two nits. Other than that, LGTM!

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
tgodzik: Fixed!

docs/integrations/new-editor.md
tgodzik: Not that nitpicky really :sweat_smile: Moved it up and also fixed the order for `openNewWindowProvider`

ckipp01: 🙏🏼

Issue #2408, opened 2021-01-19
Using scalafmt from the editor can make Metals hang


kubukoz:
**Describe the bug**

I'm seeing ~600% CPU utilization from Metals after I format a large file a couple times. The exact amount is yet to be determined, but it's still the case even after disabling "format on save".

**To Reproduce**

- todo
- 

**Expected behavior**

The editor doesn't hang

**Screenshots**

![image](https://user-images.githubusercontent.com/894884/104972562-6ec08300-59f2-11eb-845f-fd663cdd8320.png)

![image](https://user-images.githubusercontent.com/894884/104972568-754efa80-59f2-11eb-8187-a6944148bbdb.png)

![image](https://user-images.githubusercontent.com/894884/104972583-7ed86280-59f2-11eb-9c7f-1c57b0ce3524.png)

**Installation:**

- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.9.8

**Additional context**

My metals server flags: `-Xmx1G`

**Search terms**

scalafmt, hangs, cpu, formatting, formatter

ckipp01:
Thanks for the report @kubukoz! So I'm playing around with this a bit and I have a file (in Metals), which is 2500 lines long. I spammed format 30 times in a row, and then just waited.

<img width="1015" alt="Screenshot 2021-01-19 at 09 21 48" src="https://user-images.githubusercontent.com/13974112/105007075-d0571080-5a37-11eb-81c9-9d692430d9af.png">

However, a lot of the code in this file probably isn't that hard to parse compared some other code bases, so I'm curious in this file what you make heavy use of? Any specific language feature or library? Or if you've ever hit on this in any typelevel project or anything that have a huge file, feel free to point me in that direction. If not I'll start digging around for something.

Either way, there's no reason to see the numbers you're seeing. We need to get to the bottom of what's happening here.

gabro:
It would be interesting to see whether the same issue is reproducible outside of the editor (i.e. with the Scalafmt CLI). IIRC there were performance issues in heavily nested code

hedefalk:
I think I'm having the same problem. Seeing about 500-600% CPU

![image](https://user-images.githubusercontent.com/116520/106586021-44b9a580-6548-11eb-9f72-e62969c7ccde.png)


and since I have format-on-save turned on, the file save action seems to be intercepted and not saved to disk so sbt can't see the change which is very disruptive for me.

This notif at least lets me cancel the format and the save is done:

[...]

PR #2476, opened 2021-02-04
Try to fix memory leak in presentation compiler

tgodzik:
It seems that the presentation compiler thread might have not been shutdown after replacing it with the Metals one. ~~As we are actually not doing any work on the presentation compiler thread, I shut it down at the start.~~

~~Also it seems that we are not actually able to shut down just the default thread, unless we also shutdown the whole compiler, so the workaround here was not working.~~

~~Having a separate compiler thread implementation was more important at the time of:
https://github.com/scalameta/metals/commit/42869b83115eddca3f0e0dd900947a31dff2b7b9#diff-595864b0c8640c58607d7984e12eba675a293acedb5950339d4f129e69598638R58~~

~~it should not be that important now and all features work with it shutdown.~~

Actually, we still do use `.interrupt`, so I followed @mlachkar advice from https://github.com/scalacenter/scalafix/pull/1272 

This should help with https://github.com/scalameta/metals/issues/2408 as we no longer keep any leftover threads that would hold MetalsGlobal in memory.

@olafurpg would love it if you could take a look, I think once the CompilerJobQueue was added this became obsolete.

### DIFF
--- a/mtags/src/main/scala-2/scala/tools/nsc/interactive/GlobalProxy.scala
+++ b/mtags/src/main/scala-2/scala/tools/nsc/interactive/GlobalProxy.scala
@@ -2,6 +2,7 @@ package scala.tools.nsc.interactive
 
 import java.util.logging.Level
 
+import scala.tools.nsc.util.WorkScheduler
 import scala.util.control.NonFatal
 
 import scala.meta.internal.pc.MetalsGlobal
@@ -38,8 +39,8 @@ trait GlobalProxy { this: MetalsGlobal =>
   private def newRunnerThread(): Thread = {
     if (compileRunner.isAlive) {
       try {
-        val re = askForResponse(() => throw ShutdownReq)
-        re.get
+        this.askShutdown()
+        while (compileRunner.isAlive) Thread.sleep(0)
       } catch {
         case NonFatal(e) =>
           logger.log(
@@ -49,6 +50,7 @@ trait GlobalProxy { this: MetalsGlobal =>
           )
       }
     }
+    this.scheduler = new WorkScheduler
     compileRunner = new MetalsGlobalThread(this, "Metals")
     compileRunner.setDaemon(true)
     compileRunner.start()


mtags/src/main/scala-2/scala/tools/nsc/interactive/GlobalProxy.scala
```diff
@@ -32,26 +27,4 @@ trait GlobalProxy { this: MetalsGlobal =>
     }
   }
 
-  /**
-   * Shuts down the default presentation compiler thread and replaces it with a custom implementation.
-   */
-  private def newRunnerThread(): Thread = {
-    if (compileRunner.isAlive) {
-      try {
-        val re = askForResponse(() => throw ShutdownReq)
```
tgodzik, line 41: So this exception would just be caught and nothing would happen.

mtags/src/main/scala-2/scala/tools/nsc/interactive/MetalsGlobalThread.scala
```diff
@@ -1,67 +0,0 @@
...
-        compiler.log.close()
-
-        // make sure we don't keep around stale instances
-        compiler = null
-      case ex: Throwable =>
-        compiler.log.flush()
-
-        ex match {
-          // + scalac deviation
-          case InterruptException() =>
```
tgodzik, line 49: We no longer do interrupt the presentation compiler thread itself, so this should no longer be needed. Handling `ThreadDeath` would probably be still useful, but it doesn't really help us to have it here as the presentation compiler thread is still running anyway.

olafurpg (approved):
Big 👍 to remove the custom thread if it's no longer needed.

pvid:
🎉 this is great! Just tried it out manually. I think think it would be worth it to cut a new version just for this - it will improve the user experience by quite a bit

tgodzik:
>  this is great! Just tried it out manually. I think it would be worth it to cut a new version just for this - it will improve the user experience by quite a bit

Yeah, I think it actually fixes a number of issues :sweat_smile: but we might want to wait a bit to test it out on the snpashots.

mtags/src/main/scala-2/scala/tools/nsc/interactive/GlobalProxy.scala
```diff
@@ -49,6 +50,7 @@ trait GlobalProxy { this: MetalsGlobal =>
           )
       }
     }
+    this.scheduler = new WorkScheduler
```
ckipp01, line 53: So I guess I don't fully understand this part. What exactly does create and assigning a new WorkScheduler do? Mainly how does it help this situation?

tgodzik: This is a bit of a hack. When we shutdown the compiler the scheduler gets replaced with something like `NoWorkScheduler` which will not accept any work. We replace it back with the default one so that the modified thread can work with it.

ckipp01: 🤔 ok, just to make sure I understand, after we shutdown, and then right below when we create a new `MetalsGlobalThread`, don't we also get a new compiler, which would already have a new WorkSchedule?

tgodzik: Whenever Global is created it comes with it's own thread and WorkScheduler, so we can't create a new one. Here, we shutdown the compiler, which shuts down the thread and replaces the scheduler. After that, we replace the scheduler with a working one and add the new thread, in this way hijack this particular Global object.

ckipp01: Alright, I think I understand. Thanks a lot for explaining.

ckipp01 (approved):
LGTM 🚀

mtags/src/main/scala-2/scala/tools/nsc/interactive/GlobalProxy.scala
tgodzik: I did dig into that for a while before @mlachkar suggestion, so I do understand your confusion here. It's not perfect, it would probably be much better to just use the defaults, but I don't want to change too much and the reason for the changed compiler thread might still exist.

ckipp01: Yea understandable, just trying to wrap my head around what's fully going on here.

kubukoz:
Trying out this beauty!

kubukoz:
It worked. It's awesome.

PR #2477, opened 2021-02-04
Fix TreeViewLspSuite

tgodzik:
Tt seems like animal-sniffer-annotations.jar disappeared from the libraries, which unexpected as we keep the same versions here, but it should not be caused by an issues on Metals side.


ckipp01 (approved):
This is weird, I don't get why this is changing, but LGTM if it's fixing the tests.

PR #2481, opened 2021-02-06
A bit of refactoring around BSP connection and Build Import.

### LINKED ISSUE #2347: Metals will not connect to sbt if compilation fails
**Describe the bug**

If we try to connect to the sbt BSP server on a project that has a compile error, it will not succeed

**To Reproduce**

1. Create a new project
2. Introduce an error
3. Try to switch to sbt bsp

**Expected behavior**

sbt just reports the error.

**Installation:**

- Metals version: v0.9.8

**Additional context**

This was raised on gitter

**Search terms**

sbt BSP compile error

ckipp01:
So there is no user facing change in this, it's mainly just moving stuff
around a bit, and then trying to improve some logging. This is in
response to #2347 and should close it, unless we really want to do some
more refactoring. Mainly in that ticket there was a situation where we
were unable to get sources for a build target, and therefore the
connection to the build server didn't work, and the message was that the
connection failed, and check the logs for more details. However, the
logs aren't helpful at all:

```
2021.01.05 16:37:03 ERROR Failed to connect with build server, no functionality will work.
org.eclipse.lsp4j.jsonrpc.ResponseErrorException:
        at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleResponse(RemoteEndpoint.java:209)
        at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:193)
        at org.eclipse.lsp4j.jsonrpc.TracingMessageConsumer.consume(TracingMessageConsumer.java:114)
        at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:194)
        at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:94)
        at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:113)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
```

So in order to at least improve the logging in this scenario, I added
another case in the recover in our register Method that we wrap all the
BSP calls in to catch the class of what we are trying to get, and
instead of returning an exception like the above when we _do_ have info
about what may have failed replace it with a more useful exception like
the one below:

```
scala.meta.internal.metals.MetalsBspException: BSP connection failed in the attempt to get: SourcesResult
        at scala.meta.internal.metals.BuildServerConnection$$anonfun$1.applyOrElse(BuildServerConnection.scala:230)
        at scala.meta.internal.metals.BuildServerConnection$$anonfun$1.applyOrElse(BuildServerConnection.scala:223)
        at scala.concurrent.Future.$anonfun$recoverWith$1(Future.scala:417)
        at scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:41)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.lang.Thread.run(Thread.java:834)
```

Having this would have made figuring out the issue in #2347 _way_
easier. An argument _could_ be made to still maybe log out the original,
so let me know if you think I should do that.

The other change is just moving the `importedBuild` method out of the
`MetalsLanguageServer` companion object and into the companion object of
`ImportedBuild`. I think it makes a bit more sense there and also gets
rid of another thing in `MetalsLanguageServer`.

Closes #2347

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ImportedBuild.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ImportedBuild.scala
@@ -2,7 +2,10 @@ package scala.meta.internal.metals
 
 import java.{util => ju}
 
-import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+import scala.meta.internal.metals.MetalsEnrichments._
 
 import ch.epfl.scala.bsp4j._
 
@@ -50,4 +53,26 @@ object ImportedBuild {
       new SourcesResult(ju.Collections.emptyList()),
       new DependencySourcesResult(ju.Collections.emptyList())
     )
+
+  def fromConnection(
+      conn: BuildServerConnection
+  )(implicit ec: ExecutionContext): Future[ImportedBuild] =
+    for {
+      workspaceBuildTargets <- conn.workspaceBuildTargets()
+      ids = workspaceBuildTargets.getTargets.map(_.getId)
+      scalacOptions <- conn.buildTargetScalacOptions(
+        new ScalacOptionsParams(ids)
+      )
+      sources <- conn.buildTargetSources(new SourcesParams(ids))
+      dependencySources <- conn.buildTargetDependencySources(
+        new DependencySourcesParams(ids)
+      )
+    } yield {
+      ImportedBuild(
+        workspaceBuildTargets,
+        scalacOptions,
+        sources,
+        dependencySources
+      )
+    }
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -17,7 +17,6 @@ import scala.collection.immutable.Nil
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 import scala.concurrent.Await
-import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContextExecutorService
 import scala.concurrent.Future
 import scala.concurrent.Promise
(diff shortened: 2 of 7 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1939,40 +1938,40 @@ class MetalsLanguageServer(
...
+        val message =
+          "Failed to connect with build server, no functionality will work."
+        val details = " See logs for more details."
+        languageClient.showMessage(
+          new MessageParams(MessageType.Error, message + details)
+        )
+        scribe.error(message, e)
+        BuildChange.Failed
+      }
+      .flatMap(compileAllOpenFiles)
```
ckipp01, line 1974: I hesitated to completely redo this, but wanted to instead just focus on the task at hand. However, I think we can actually improve this here. Mainly, in the first part we are actually disconnecting and then re-connecting. Then we are running the full build import, and then initiating the tree view. However, if at any point during that process something goes wrong, we `disconnectOldBuildServer()` when that's probably not necessary at all depending on the situation. It may be preferable to instead break those three things out from one another and handle a failure in differently. An argument could be made that if something goes wonky during the build import that _maybe_ could be fixed easily, do we really want to disconnect fully? There may be partial support still very much available. Just something to think about.

metals/src/main/scala/scala/meta/internal/metals/ImportedBuild.scala
```diff
@@ -3,8 +3,11 @@ package scala.meta.internal.metals
 import java.{util => ju}
 
 import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
 
 import ch.epfl.scala.bsp4j._
+import ch.epfl.scala.{bsp4j => b}
```
tgodzik, line 10: We already import `import ch.epfl.scala.bsp4j._`, so this should not be needed.

metals/src/main/scala/scala/meta/internal/metals/ImportedBuild.scala
```diff
@@ -43,11 +46,35 @@ case class ImportedBuild(
 }
 
 object ImportedBuild {
+  import scala.meta.internal.metals.MetalsEnrichments._
```
tgodzik, line 49: Can we move it up to the other imports? That's usually how it's done in other places.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: We should not fail on tree view for sure, we used to fail on compile too, but I moved it out because of that. It's fine to do it here, but also can be a separate PR.

tgodzik (approved):
Looks good! Just some nitpicks on my side.

metals/src/main/scala/scala/meta/internal/metals/ImportedBuild.scala
ckipp01: Ah duh, fixed.

metals/src/main/scala/scala/meta/internal/metals/ImportedBuild.scala
ckipp01: Well I purposefully put it here since in the actual class we do a lot of `.asScala` in various places, and when we put the `MetalsEnrichments` in scope of those, everything breaks. So this was an easy way to not have to touch any of that code and only scope it to what is using it here. Is that alright, or do you want me to also refactor the `++` and `targetUris` in `ImportedBuild`?

ckipp01: Actually nevermind. It was just conflicting with one of the other imports. I can just remove it and all is good.

PR #2483, opened 2021-02-07
Introduce a `metals/listBuildTargets` LSP extension.

ckipp01:
You can find more of my mindset behind this in this discussion:
https://github.com/scalameta/metals/discussions/2479, but I'd like to
propose that we expose this info. One concrete thing that I'm planning
on using this for is having `nvim-metals` be able to quickly get the
build targets for a workspace, which then the user can just quickly
choose the one they want, and that is how the debug process or run
process starts without the need to manually add the build target or rely
on code lenses.

After playing around with this a bit, I'm also quite excited about other
possibilities that this opens up. There is _so much_ information in here
that client extensions can use. Everything from simple things like
`javaHome` or Scala versions to more complex things like the full
classpath for a build target can be accessed by extensions to do all
sorts of things. I'm excited to play around with this more.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1708,6 +1708,13 @@ class MetalsLanguageServer(
     }
   }
 
+  @JsonRequest("metals/listBuildTargets")
+  def listBuildTargets(): CompletableFuture[ju.List[ScalaTarget]] = {
+    Future {
+      buildTargets.all.toList.asJava
+    }.asJava
+  }
+
   @JsonRequest("metals/treeViewChildren")
   def treeViewChildren(
       params: TreeViewChildrenParams
--- a/tests/unit/src/test/scala/tests/BuildTargetsLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/BuildTargetsLspSuite.scala
@@ -1,29 +1,32 @@
 package tests
 
+import scala.meta.internal.metals.MetalsEnrichments._
+
 class BuildTargetsLspSuite
     extends BaseLspSuite("build-targets")
     with TestHovers {
 
   override def munitIgnore: Boolean = isWindows
 
+  val workspaceBuildTargets: String =
+    s"""/metals.json
+       |{
+       |  "a": {
+       |    "scalaVersion": "2.10.6",
+       |    "libraryDependencies": ["com.lihaoyi::sourcecode:0.1.7"],
+       |    "additionalSources": [ "shared/Main.scala" ]
+       |  },
+       |  "b": {
+       |    "scalaVersion": "${BuildInfo.scalaVersion}",
+       |    "libraryDependencies": ["com.lihaoyi::sourcecode:0.1.7"],
+       |    "additionalSources": [ "shared/Main.scala" ]
+       |  }
+       |}
+        """.stripMargin
+
   test("scala-priority") {
     for {
-      _ <- server.initialize(
-        s"""/metals.json
-           |{
-           |  "a": {
-           |    "scalaVersion": "2.10.6",
-           |    "libraryDependencies": ["com.lihaoyi::sourcecode:0.1.7"],
-           |    "additionalSources": [ "shared/Main.scala" ]
-           |  },
-           |  "b": {
-           |    "scalaVersion": "${BuildInfo.scalaVersion}",
-           |    "libraryDependencies": ["com.lihaoyi::sourcecode:0.1.7"],
-           |    "additionalSources": [ "shared/Main.scala" ]
-           |  }
-           |}
-        """.stripMargin
-      )
+      _ <- server.initialize(workspaceBuildTargets)
       // Assert that a supported Scala version target is picked over 2.10.
       _ <- server.assertHover(
         "shared/Main.scala",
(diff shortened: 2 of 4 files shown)


tests/unit/src/test/scala/tests/BuildTargetsLspSuite.scala
```diff
@@ -35,4 +38,22 @@ class BuildTargetsLspSuite
       )
     } yield ()
   }
+  test("metals/buildTargets") {
```
ckipp01, line 41: Since `metals/buildTargets` is mainly just copying a copy of the internal state of build targets that Metals has, I wasn't sure exactly what would actually be useful to test. Mainly I have this here just as a sanity check to ensure that we are indeed getting back what we'd sort of expect without having to include all of the data that is actually coming when a user would call this.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -152,23 +152,23 @@ class MetalsLanguageServer(
   val excludedPackageHandler: ExcludedPackagesHandler =
     new ExcludedPackagesHandler(userConfig.excludedPackages)
   var ammonite: Ammonite = _
-  val buildTargets: BuildTargets = BuildTargets.withAmmonite(() => ammonite)
+  val buildTargetz: BuildTargets = BuildTargets.withAmmonite(() => ammonite)
```
ckipp01, line 155: If you have a better name for this, I'm all ears. It conflicts with the new `buildTargets` method down below so I needed to change it.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1708,6 +1708,13 @@ class MetalsLanguageServer(
     }
   }
 
+  @JsonRequest("metals/buildTargets")
```
tgodzik, line 1711: Is there any other use case than specifying it when running/debugging? From what I've seen idea of the `target` actually confuses users and besides it would be needed in some very niche use cases.

Also, it might be easier to name it `listBuildTargets` and leave `buildTagets` name for the current purpose.

ckipp01: Sure switching it to `listBuildTargets` shouldn't be an issue. Plus, related to the first point, this isn't really something that would be directly shown to a user without a bit more context to ensure it's clear.

> Is there any other use case than specifying it when running/debugging?

Not really any _concrete_ things at the moment, but I'm playing around with a few things related to displaying a build to a user in a way that is agnostic to the build tool. Mainly I just think there is a ton of information contained in `buildTargets` about your build. All of this _is_ available through your build tool, but I think it's quite powerful to have a way to access this is a way that may be easier than through the build tool, and also agnostic to the actual build tool. I like the idea that if a user is curious about what compiler options they have whether they are using mill or sbt, it doesn't matter, we have that information in a that could just be returned to a client, and a client could show it the same regardless of the build tool.

We don't really have any endpoints like this atm that just give the client a bunch of info that they can start to play around with and come up with useful ways to either display it or help the user with their build. _Maybe_ the closes thing would be `workspace/sybmol` where a huge list of symbols can just be returned to the client to do whatever it wants with.

More so than any use case I currently have for this, I more curious about just exposing this and seeing what comes of it. At worse, it's just not used. At best, we get some cool client features that can maybe do think we didn't really think about before.

ckipp01: Also for a bit more context, one thing I was thinking about was right now when you send in a `debug-adapter-start` a main class is required. While that is somewhat easy most of the time to find, it also may not be obvious especially in projects that have multiple modules. So the flow that I was thinking of is that in a situation with no code lenses a user could do maybe a `:MetalsFindMainsMethods` or something like that. Under the hood that would do a `metals/listBuildTargets` to which they could choose `app` or `foo` or whatever they want, then that would send in a `metals/findMainMethodsForTarget` or something like that, which would then return the main methods for that target. So in theory a user could go from opening a project to getting the main method to running right away without having to manually do anything.

However, _maybe_ an easier way to accomplish that is to allow for `debug-adapter-start` to also only allow for `buildTarget` to be set, and then do a look up for the main method in that target.

tgodzik: I think allowing to have just a build target set would make it more general across the clients. We could even allow have nothing specified at all and use quick pick to select the main class to run. 

There is a lot of possibilities really 😅 and we could hugely improve the user experience here. 

We do actually also expose build target information via the tree view. We could add a lot more information there like scala options etc.

ckipp01: > We could even allow have nothing specified at all and use quick pick to select the main class to run.

Yea this is an interesting idea. I've been looking a lot at the jdtls extensions, and `nvim-jdtls` utilizes the `vscode.java.resolveMainClass` when they setup DAP and it just returns the main classes for this to do what they want with, which is nice.

https://github.com/mfussenegger/nvim-jdtls/blob/a35ce80f0e618da0eceb0d5f6874608d58d3ec26/lua/jdtls.lua#L983

> We do actually also expose build target information via the tree view. We could add a lot more information there like scala options etc.

Indeed we do, but this also forces a user to access it through the tree view, which a certain client may not have.

Anyways since I'm working on DAP stuff currently I may try to play around with this a bit and just try to provide an option that if `debug-adapter-start` is triggered with nothing, we give them a quick pick like said. From poking around, I don't think that would actually be that hard.

tgodzik:
I don't mind merging this in, but I think in most cases the client wants to expose some specific information, which is easier to work with on the server side and only show what is needed at the time. This also makes it easier for the clients to implement, since they would just add the front-end part.

TLDR; if it's really needed let's merge it, but otherwise I would wait for a concrete use case.

ckipp01:
> I don't mind merging this in, but I think in most cases the client wants to expose some specific information, which is easier to work with on the server side and only show what is needed at the time. This also makes it easier for the clients to implement, since they would just add the front-end part.
> 
> TLDR; if it's really needed let's merge it, but otherwise I would wait for a concrete use case.

Let me work on `debug-adapter-start` a bit first. If I can get an easy way for a user to just execute a command, pick what the run and then run it it without needing to manually enter anything or rely on code lenses, then I'll just close this. For now I'll just mark it as a draft.

ckipp01:
Close in favor of https://github.com/scalameta/metals/pull/2532

Issue #2480, opened 2021-02-05
NPEs in recent snapshot


kubukoz:
**Describe the bug**

From metals.log:

```scala
SEVERE: null
java.lang.NullPointerException
        at scala.meta.internal.pc.completions.OverrideCompletions$OverrideCompletion.<init>(OverrideCompletions.scala:46)
        at scala.meta.internal.pc.completions.Completions.completionPositionUnsafe(Completions.scala:500)
        at scala.meta.internal.pc.completions.Completions.completionPositionUnsafe$(Completions.scala:409)
        at scala.meta.internal.pc.MetalsGlobal.completionPositionUnsafe(MetalsGlobal.scala:29)
        at scala.meta.internal.pc.completions.Completions.completionPosition(Completions.scala:401)
        at scala.meta.internal.pc.completions.Completions.completionPosition$(Completions.scala:386)
        at scala.meta.internal.pc.MetalsGlobal.completionPosition(MetalsGlobal.scala:29)
        at scala.meta.internal.pc.CompletionProvider.safeCompletionsAt(CompletionProvider.scala:446)
        at scala.meta.internal.pc.CompletionProvider.completions(CompletionProvider.scala:57)
        at scala.meta.internal.pc.ScalaPresentationCompiler.$anonfun$complete$1(ScalaPresentationCompiler.scala:124)
        at scala.meta.internal.pc.CompilerAccess.withSharedCompiler(CompilerAccess.scala:137)
        at scala.meta.internal.pc.CompilerAccess.$anonfun$withInterruptableCompiler$1(CompilerAccess.scala:87)
        at scala.meta.internal.pc.CompilerAccess.$anonfun$onCompilerJobQueue$1(CompilerAccess.scala:197)
        at scala.meta.internal.pc.CompilerJobQueue$Job.run(CompilerJobQueue.scala:103)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:834)
```

And another one:

```scala
Feb 03, 2021 4:38:21 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint fallbackResponseError
SEVERE: Internal error: java.lang.NullPointerException
java.util.concurrent.CompletionException: java.lang.NullPointerException
[...]

PR #2486, opened 2021-02-08
Remove unused val that may cause null pointer exceptions

tgodzik:
Fixes https://github.com/scalameta/metals/issues/2480


ckipp01 (approved):
Literally the worst. Unused code that is throwing exceptions 😆 

Let's get this thing merged!

PR #2487, opened 2021-02-08
[Scala 3] Add filtering and sorting for completions

sellophane:
Adding completions filtering and ordering based on Scala 2.* `CompletionProvider` algorithm


tests/cross/src/test/scala/tests/pc/CompletionExtensionSuite.scala
```diff
@@ -0,0 +1,28 @@
...
+import tests.BuildInfoVersions
+
+class CompletionExtensionSuite extends BaseCompletionSuite {
+
+  override def requiresJdkSources: Boolean = true
+
+  override def excludedScalaVersions: Set[String] =
+    BuildInfoVersions.scala3Versions.dropWhile(_ != "3.0.0-M2").toSet
+
+  check(
```
tgodzik, line 13: I think this is unrelated to the PR, could you maybe add the test in a separate one?

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -0,0 +1,169 @@
...
+    ctx: Context
+) {
+    implicit val context: Context = ctx
+      
+    def completions() : List[Completion]
+        = Completion.completions(pos)._2.filterInteresting.sorted(completionOrdering(ctx, pos))
+
+    implicit class CompletionExtension(val c: Completion) {
+      // completionItem method either way takes only head symbol, so why can safely ignore the rest
+      val sym: Symbol = c.symbols.head
```
tgodzik, line 30: We can use extension methods instead of implicit classes in Scala 3.

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    implicit class CompletionListExtension(val l: List[Completion]) {
+      val filterInteresting: List[Completion] = {
+        val isSeen = mutable.Set.empty[String]
+        val buf = List.newBuilder[Completion]
+        implicit class SrcPosExtension(val s: SrcPos) {
+          def isAfter(s1: SrcPos) = s.sourcePos.point > s1.sourcePos.point
+        }
+        def visit(head: Completion): Boolean = {
```
tgodzik, line 40:
```suggestion
        def isUninteresting(completion: Completion): Boolean = {
```

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -0,0 +1,169 @@
...
+      // completionItem method either way takes only head symbol, so why can safely ignore the rest
+      val sym: Symbol = c.symbols.head
+    }
+
+    implicit class CompletionListExtension(val l: List[Completion]) {
+      val filterInteresting: List[Completion] = {
+        val isSeen = mutable.Set.empty[String]
+        val buf = List.newBuilder[Completion]
+        implicit class SrcPosExtension(val s: SrcPos) {
+          def isAfter(s1: SrcPos) = s.sourcePos.point > s1.sourcePos.point
```
tgodzik, line 38: It's actually used in one place, maybe we can inline it?

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -0,0 +1,169 @@
...
+import dotty.tools.dotc.util.{NameTransformer, NoSourcePosition, SourcePosition}
+
+class CompletionProvider(
+    pos: SourcePosition,
+    ctx: Context
+) {
+    implicit val context: Context = ctx
+      
+    def completions() : List[Completion]
+        = Completion.completions(pos)._2.filterInteresting.sorted(completionOrdering(ctx, pos))
```
tgodzik, line 26:
```suggestion
        = {
        val (_, completions) = Completion.completions(pos) 
        completions.filterInteresting.sorted(completionOrdering(ctx, pos))
      }
```
I think looks a bit better.

tgodzik (commented):
Thanks for the contribution! :tada: 

Can we also maybe reuse some tests in CompletionSuite that deal with sorting? Would also give us more assurance that we are on par with what was achieved for Scala 2.

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -0,0 +1,169 @@
...
+      ctx: Context,
+      position: SourcePosition
+  ): Ordering[Completion]
+    = new Ordering[Completion] {
+      implicit val context: Context = ctx
+      def compareLocalSymbols(o1: Completion, o2: Completion): Int = {
+        val s1 = o1.sym
+        val s2 = o2.sym
+        if (s1.isLocal && s2.isLocal) {
+          if (s1.line > s2.line) -1
```
tanishiking, line 132: nitpick
```suggestion
          if (s1.sourcePos.isAfter(s2.srcPos)) -1
```

?

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -0,0 +1,169 @@
...
+      override def compare(o1: Completion, o2: Completion): Int = {
+        val byLocalSymbol = compareLocalSymbols(o1, o2)
+          if (byLocalSymbol != 0) byLocalSymbol
+          else {
+            val byRelevance = Integer.compare(
+              computeRelevancePenalty(o1.sym),
+              computeRelevancePenalty(o2.sym)
+            )
+            if (byRelevance != 0) byRelevance
+            else {
```
tanishiking, line 147: it would be awesome if we have `byFuzzy` comparison between `byRelevance` and `byIdentifier` as scala2 does
https://github.com/scalameta/metals/blob/4da9f0e1c68380bd9b47b01609ee30b0446995b0/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala#L173-L182

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -0,0 +1,169 @@
...
+              val byIdentifier = IdentifierComparator.compare(o1.sym.name.mangledString, o2.sym.name.mangledString)
```
tanishiking, line 148: just out of curiosity, why using `mangledString` instead of `toString`?

tanishiking (commented):
On top of the @tgodzik feedback, took a quick look and left some nit comments, and it looks good so far :tada:

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
sellophane: Actually, it's now used in one more place

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
sellophane: Agree.

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
sellophane: Actually, there was no good reason...

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
sellophane: Yes, I skipped that one because I previously failed to obtain query for completion, but now it's back

tgodzik (approved):
LGTM!

Issue #2484, opened 2020-01-02
Refactor: Insert inferred result type


gabro:
**Is your feature request related to a problem? Please describe.**
Annotating result types of public members is considered best practice in many code-bases, but doing so consistently by hand can be tedious.

Also, sometimes it's not immediately obvious which type to insert and/or which imports are needed in order to bring the inferred type into scope.

**Describe the solution you'd like**
I would like to see:

- a code action that can insert the inferred result type of a member, including relevant imports:

```scala
// before
def someMethod(a: List[String], f: String => Int) =
  a.map(f).toBuffer

// after
import scala.collection.mutable

def someMethod(a: List[String], f: String => Int): mutable.Buffer[Int] =
  a.map(f).toBuffer
```

- (possibly) a code action (paired with a command) that does the same thing for the entire file

Given the potential complexity of this refactoring and the number of edge cases, it think it would be good to re-use the `ExplicitResultTypes` rule of Scalafix (which would require solving scalameta/metals#1675 in the process)

**Describe alternatives you've considered**
- Inserting type by hands is what I do now, but it's tedious and easy to forget
- Scalafix has a CLI interface that can run the refactor in batch, but having an editor integration would allow for much more granular control.

**Additional context**

**Search terms**:
inferred result type, return type, scalafix

PR #2490, opened 2021-02-09
Add insert inferred type code action

tgodzik:
Inferred type code action is responsible for adding type annotations for any place that might need it such as:
- val/var declaration
- method declaration
- pattern matches
- lambdas
- for comprehensions

The code action itself has two steps:

1. Using the scalameta parser we check if there is a name at a given position and if it belongs to a tree node that doesn't have the type declared. If both is true we return code action with the `insert-inferred-type` command.

2. When the `insert-inferred-type` code action is run we use the presentation compiler and the new provider `InferredTypeProvider` to calculate the needed edits that actually add the type. This also includes any imports that are needed.

To that purpose there are two suites that test this behaviour:
- InsertInferredTypeSuite - checks the presentation compiler implementation
- InsertInferredTypeLspSuite - checks the whole workflow from the client perspective

Demo on the scalafmt's Router class:

insert-type-scalafmt

Fixes https://github.com/scalameta/metals/issues/2484

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Command.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Command.scala
@@ -2,6 +2,7 @@ package scala.meta.internal.metals
 
 import scala.meta.internal.metals.MetalsEnrichments._
 
+import com.google.gson.JsonPrimitive
 import org.eclipse.{lsp4j => l}
 
 case class Command(
@@ -14,4 +15,22 @@ case class Command(
 
   def toLSP(arguments: List[AnyRef]): l.Command =
     new l.Command(title, id, arguments.asJava)
+}
+
+object Argument {
+
+  def getAsString(obj: AnyRef): Option[String] = {
+    obj match {
+      case p: JsonPrimitive if p.isString => Option(p.getAsString())
+      case _ => None
+    }
+  }
+
+  def getAsInt(obj: AnyRef): Option[Int] = {
+    obj match {
+      case p: JsonPrimitive if p.isNumber => Option(p.getAsInt())
+      case _ => None
+    }
+  }
+
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1564,9 +1564,7 @@ class MetalsLanguageServer(
           for {
             args <- Option(params.getArguments())
             argObject <- args.asScala.headOption
-            arg = argObject.asInstanceOf[JsonPrimitive]
-            if arg.isString()
-            symbol = arg.getAsString()
+            symbol <- Argument.getAsString(argObject)
             location <- definitionProvider.fromSymbol(symbol).asScala.headOption
           } {
             languageClient.metalsExecuteClientCommand(
@@ -1662,18 +1660,10 @@ class MetalsLanguageServer(
 
       case ServerCommands.NewScalaFile() =>
         val args = params.getArguments.asScala
-        val directoryURI = args.lift(0).collect {
-          case directory: JsonPrimitive if directory.isString =>
-            new URI(directory.getAsString())
-        }
-        val name = args.lift(1).collect {
-          case name: JsonPrimitive if name.isString =>
-            name.getAsString()
-        }
-        val fileType = args.lift(2).collect {
-          case fileType: JsonPrimitive if fileType.isString =>
-            fileType.getAsString()
-        }
+        val directoryURI =
+          args.lift(0).flatMap(Argument.getAsString).map(new URI(_))
+        val name = args.lift(1).flatMap(Argument.getAsString)
+        val fileType = args.lift(2).flatMap(Argument.getAsString)
         newFileProvider
           .handleFileCreation(directoryURI, name, fileType)
           .asJavaObject
@@ -1701,6 +1691,35 @@ class MetalsLanguageServer(
           Future.successful(()).asJavaObject
         }
 
+      case ServerCommands.InsertInferredType() =>
+        CancelTokens.future { token =>
+          val args = params.getArguments().asScala
+          val futureOpt = for {
+            arg0 <- args.lift(0)
+            uri <- Argument.getAsString(arg0)
+            arg1 <- args.lift(1)
+            line <- Argument.getAsInt(arg1)
+            arg2 <- args.lift(2)
+            character <- Argument.getAsInt(arg2)
+            pos = new l.Position(line, character)
+            textDoc = new l.TextDocumentIdentifier(uri)
+            params = new TextDocumentPositionParams(textDoc, pos)
+          } yield {
+            for {
+              edits <- compilers.insertInferredType(params, token)
+              if (!edits.isEmpty())
+              workspaceEdit = new l.WorkspaceEdit(Map(uri -> edits).asJava)
+              _ <- languageClient
+                .applyEdit(new ApplyWorkspaceEditParams(workspaceEdit))
+                .asScala
+            } yield ()
+          }
+
+          futureOpt.getOrElse {
+            languageClient.showMessage(Messages.InsertInferredTypeFailed)
+            Future.unit
+          }.withObjectValue
+        }
       case cmd =>
         scribe.error(s"Unknown command '$cmd'")
         Future.successful(()).asJavaObject
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -304,6 +304,17 @@ object ServerCommands {
     "[uri], the uri of the worksheet that you'd like to copy the contents of."
   )
 
+  val InsertInferredType = new Command(
+    "insert-inferred-type",
+    "Insert inferred type of a value",
+    """|Whenever a user chooses code action to insert the inferred type this command is later ran to 
+       |calculate the type and insert it in the correct location.
+       |""".stripMargin,
+    """|[uri, line, character], uri to the document that the command needs to be invoked on 
+       |together with line number and character/column.
+       |""".stripMargin
+  )
+
   /**
    * Open the browser at the given url.
    */
@@ -390,6 +401,7 @@ object ServerCommands {
       GotoSuperMethod,
       GotoSymbol,
       ImportBuild,
+      InsertInferredType,
       NewScalaFile,
       NewScalaProject,
       ResetChoicePopup,
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala
@@ -1,0 +1,97 @@
+package scala.meta.internal.metals.codeactions
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+import scala.meta.Defn
+import scala.meta.Enumerator
+import scala.meta.Pat
+import scala.meta.Term
+import scala.meta.Tree
+import scala.meta.inputs.Position
+import scala.meta.internal.metals.CodeAction
+import scala.meta.internal.metals.Compilers
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.ServerCommands
+import scala.meta.internal.parsing.Trees
+import scala.meta.pc.CancelToken
+
+import org.eclipse.{lsp4j => l}
+
+class InsertInferredType(trees: Trees, compilers: Compilers)
+    extends CodeAction {
+
+  import InsertInferredType._
+  override def kind: String = l.CodeActionKind.QuickFix
+
+  override def contribute(
+      params: l.CodeActionParams,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = Future {
+
+    def insertInferTypeAction(title: String): l.CodeAction = {
+      val codeAction = new l.CodeAction()
+      codeAction.setTitle(title)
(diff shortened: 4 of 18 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -0,0 +1,185 @@
...
+        val typeNameEdit =
+          new TextEdit(insertPos.toLSP, ": " + prettyType(tpt.tpe))
+
+        typeNameEdit :: additionalImports
+
+      /* `case t =>`
+       *  turns into
+       * `case t: Int =>`
+       */
+      case bind @ Bind(name, body) =>
```
gabro, line 145: is this a safe/desired refactor? 

Turning 

```scala
def a: Int = ???
a match {
  case n if n > 2 => ??? 
  case n          => ???
}
```

into

```scala
def a: Int = ???
a match {
  case n if n > 2 => ??? 
  case n: Int     => ???
}
```

it's not a type annotation, but it's actually inserting a runtime check. I'm not sure I would put it in the same basket as the other refactors this PR covers.

gabro (commented):
Really cool work @tgodzik, I'm excited to see this functionality landing in Metals!

I've left one question line, and I was also wondering whether it could be possible to write a test to check the inferred type selected for insertion is always the same we provide as decoration.

I think it would be surprising for a user if we did otherwise, so it may be worth testing it explicitly

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
tgodzik: I think that should be safe. It's the actual type of a, so the check is basically checking if Int is Int. If anyone changes the type of a the compiler will issue a warning.

We could name it differently similar to how Intellij does it `add type annotation to pattern definition`.

tgodzik:
> I've left one question line, and I was also wondering whether it could be possible to write a test to check the inferred type selected for insertion is always the same we provide as decoration.
> 
> I think it would be surprising for a user if we did otherwise, so it may be worth testing it explicitly

So it's not always exactly the same, especially in case of renames etc. There is an example in the gif where it is not the same, although in most simple cases it will be. I don't think it's currently possible to have it always 100% the same. We base the decorations on semanticdb, which doesn't have additional information about the imports.

I don't think we need the check specifically, since all the cases we created are rather obvious (and adding the check might be complicated :sweat_smile: ). Rather, I would check some specific situations if needed.

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -0,0 +1,185 @@
...
+
+      /* `.map(a => a + a)`
+       *     turns into
+       * `.map((a: Int) => a + a)`
+       */
+      case vl @ ValDef(_, name, tpt, _) if vl.symbol.isParameter =>
+        val nameEnd = vl.pos.start + name.length()
+        val namePos = tpt.pos.withEnd(nameEnd).withStart(nameEnd).toLSP
+        def typeNameEdit(rightParen: String) =
+          new TextEdit(namePos, ": " + prettyType(tpt.tpe) + rightParen)
```
ckipp01, line 85: This is sort of a nit, but to better represent this, can we pass in an `Option[String]` here instead so that when we are using it down below and don't need a paren we can just pass in `None` instead of `""` and just handle that here in the method. Or even have the param be `rightNeeded: Boolean` or something and just add it or not based off that. Mainly, I just don't love `rightParen = ""`.

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -0,0 +1,185 @@
...
+
+        def paramStart = vl.pos.withEnd(vl.pos.start).toLSP
+        def startParamEdit = new TextEdit(paramStart, "(")
+
+        val needsParens = lastVisitedParentTrees match {
```
ckipp01, line 90: So I think we'll need to maybe add another check here for this situation:

```scala
myList.map((a) => a * 2)
```
Because for whatever reason if someone has this, currently you are given:

```scala
myList.map(((a: Int)) => a * 2)
```

It's safe probably, but just an annoyance so it could be nice to catch it.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -304,6 +304,17 @@ object ServerCommands {
     "[uri], the uri of the worksheet that you'd like to copy the contents of."
   )
 
+  val InsertInferredType = new Command(
+    "insert-inferred-type",
+    "Insert inferred type of a value",
+    """|Whenever user chooses code action to insert inferred type this command is later run to 
+       |calculate the type and insert it in a correct location.
```
ckipp01, line 311:
```suggestion
    """|Whenever a user chooses code action to insert the inferred type this command is later ran to 
       |calculate the type and insert it in the correct location.
```

ckipp01 (changes requested):
Really fantastic job @tgodzik! I'm excited to be able to use this! Just a few nits and questions.

Also an example with another editor if you're curious:

https://user-images.githubusercontent.com/13974112/107418143-99e05300-6b16-11eb-9657-0cdc6306c90b.mov

metals/src/main/scala/scala/meta/internal/metals/Command.scala
```diff
@@ -15,3 +16,21 @@ case class Command(
   def toLSP(arguments: List[AnyRef]): l.Command =
     new l.Command(title, id, arguments.asJava)
 }
+
+object Argument {
+
+  def getAsString(obj: AnyRef): Option[String] = {
+    obj match {
+      case p: JsonPrimitive if p.isString => Option(p.getAsString())
```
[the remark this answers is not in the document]
gabro: The Option constructor doesn't catch exceptions anyway, it only turns null into None, so I too tend to use it when dealing with Java API. You never know where a null may hide 👻

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
gabro: 👍 on naming it like that

metals/src/main/scala/scala/meta/internal/metals/Command.scala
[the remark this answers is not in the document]
ckipp01: Spooky. 😨

gabro:
@tgodzik ok, sounds good, I suspected they may differ and I suppose it's not too bad

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -0,0 +1,185 @@
...
+    def prettyType(tpe: Type) =
+      metalsToLongString(tpe.widen.finalResultType, history)
+
+    typedTree match {
+      /* `val a = 1` or `var b = 2`
+       *     turns into
+       * `val a: Int = 1` or `var b: Int = 2`
+       */
+      case vl @ ValDef(_, name, tpt, _) if !vl.symbol.isParameter =>
+        // it seems name might sometime contain the space afterwards
```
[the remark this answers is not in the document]
dos65: It isn't =)
Local names have space at the end and there is `dropLocal` method for `Name` that removes it

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1701,6 +1691,37 @@ class MetalsLanguageServer(
...
+                .asScala
+            } yield ()
+          }
+
+          futureOpt
+            .getOrElse {
+              languageClient.showMessage(Messages.InsertInferredTypeFailed)
+              Future.unit
+            }
+            .asInstanceOf[Future[Object]]
```
dos65, line 1723: A nitpick triggered by `asInstanceOf`
```suggestion
          val paramsOpt = for {
            arg0 <- args.lift(0)
            uri <- Argument.getAsString(arg0)
            arg1 <- args.lift(1)
            line <- Argument.getAsInt(arg1)
            arg2 <- args.lift(2)
            character <- Argument.getAsInt(arg2)
            pos = new l.Position(line, character)
            textDoc = new l.TextDocumentIdentifier(uri)
            params = new TextDocumentPositionParams(textDoc, pos)
          } yield {
            for {
              edits <- compilers.insertInferredType(params, token)
              if (!edits.isEmpty())
              workspaceEdit = new l.WorkspaceEdit(Map(uri -> edits).asJava)
              _ <- languageClient
                .applyEdit(new ApplyWorkspaceEditParams(workspaceEdit))
                .asScala
            } yield ()
          }
          
          paramsOpt.fold(Future.unit){ params =>
              for {
                 edits <- compilers.insertInferredType(params, token)
                 if (!edits.isEmpty())
                 workspaceEdit = new l.WorkspaceEdit(Map(uri -> edits).asJava)
                 _ <- languageClient
                   .applyEdit(new ApplyWorkspaceEditParams(workspaceEdit))
                   .asScala
                } yield ()
          }.asJavaObject
```

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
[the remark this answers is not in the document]
ckipp01: Ah ok cool. Didn't know that. Thanks!

metals/src/main/scala/scala/meta/internal/metals/Command.scala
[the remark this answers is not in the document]
tgodzik: Haha, yes. I am always afraid of that 😂

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
tgodzik: Changed!

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: `asJavaObject` would need `.asScala`, so wrapping it even more. And anyway asInstanceOf is hapening in `asJavaObject`

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
```diff
@@ -83,6 +83,11 @@
      */
     public abstract CompletableFuture<List<TextEdit>> implementAbstractMembers(OffsetParams params);
 
+/**
+     * Return the missing implements and imports for the symbol at the given position.
+     */
```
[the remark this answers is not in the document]
tgodzik: Fixed!

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
tgodzik: Changed it a bit not to require passing it as string.

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
tgodzik: Added a check and a test. Good catch!

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
tgodzik: Fun fact, I think it's the most common English error for people from Slavic languages. We never really need to insert anything like `the`/`a`/`an` and it's so embedded that it's often hard to see the error even.

metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala
```diff
@@ -0,0 +1,96 @@
...
+      val codeAction = new l.CodeAction()
+      codeAction.setTitle(InsertInferredType.title)
+      codeAction.setKind(l.CodeActionKind.RefactorRewrite)
+      val range = params.getRange().getStart()
+      codeAction.setCommand(
+        ServerCommands.InsertInferredType.toLSP(
+          List(
+            params.getTextDocument().getUri(),
+            range.getLine(): java.lang.Integer,
+            range.getCharacter(): java.lang.Integer
```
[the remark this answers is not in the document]
tgodzik: Actually no, it's needed to pass it as AnyRef. It's automatically converted to `java.lang.Integer`, but the compiler requires the annotation.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
dos65: Yes, but it the first time when it will be used directly in this file. Other actions are using `asJavaObject`. I think it would be better to preserve this kind of consistency.

dos65 (approved):

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: Added a method `unitToJavaObject` that just does the cast. What do you think?

tgodzik: Or actually I think `withObjectValue` makes more sense.

dos65: Looks good!

ckipp01 (approved):
Really excited about this one! LGTM! 🚀

PR #2497, opened 2021-02-11
Improve short names when importing

tgodzik:
A while ago there were some changes to the scalafix implementation of printing types, which is actually based off Metals. While investigating some issues with importing, I decide to backport those fixes to Metals. That backported code is mainly the `MetalsGlobal.shortType`, `ShortenedNames.nameResolvesToSymbol` and `ShortenedNames.tryShortenName` methods.

Besides that:
- changed the location of where we get context to the actual place of the cursor, while keeping a separate context for the imports
- made sure that stale symbols are to looked up in relevant situations, which is anything that inserts code
- fixed the tests to adhere to these changes, where you can see most of the improvements/changes

This also fixes https://github.com/scalameta/metals/issues/1253 and some issues with insert type.

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
@@ -25,7 +25,7 @@ trait AutoImports { this: MetalsGlobal =>
 
   def doLocateImportContext(
       pos: Position,
-      autoImport: Option[AutoImportPosition]
+      autoImport: Option[AutoImportPosition] = None
   ): Context = {
     try doLocateContext(
       autoImport.fold(pos)(i => pos.focus.withPoint(i.offset))
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
@@ -17,6 +17,7 @@ import scala.tools.nsc.interactive.GlobalProxy
 import scala.tools.nsc.interactive.InteractiveAnalyzer
 import scala.tools.nsc.reporters.Reporter
 import scala.util.control.NonFatal
+import scala.{meta => m}
 
 import scala.meta.internal.mtags.MtagsEnrichments._
 import scala.meta.internal.semanticdb.scalac.SemanticdbOps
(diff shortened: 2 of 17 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -25,7 +25,7 @@ trait AutoImports { this: MetalsGlobal =>
 
   def doLocateImportContext(
       pos: Position,
-      autoImport: Option[AutoImportPosition]
+      autoImport: Option[AutoImportPosition] = None
```
tgodzik, line 28: This is needed so that we only use auto import position when inserting actual imports:
```scala
package a
import java.nio.Paths
<<import_position>>

object A{
  class Path
  val a<<actual_position>>  = Paths.get(".")
}

```
view from the `actual_position` differs from `import_position`

mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
```diff
@@ -194,80 +195,126 @@ class MetalsGlobal(
       isVisited += key
       val result = tpe match {
         case TypeRef(pre, sym, args) =>
-          val ownerSymbol = pre.termSymbol
```
tgodzik, line 197: Changes in this class come from https://github.com/scalacenter/scalafix/blob/69c0369ab49c41b0473ca68fe0a8ade396d3157e/scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala#L166

mtags/src/main/scala-2/scala/meta/internal/pc/Signatures.scala
```diff
@@ -138,52 +132,90 @@ trait Signatures { this: MetalsGlobal =>
       nameResolvesToSymbol(top.name.toTermName, top)
     }
 
-    def nameResolvesToSymbol(sym: Symbol): Boolean = {
-      nameResolvesToSymbol(sym.name, sym)
+    def isSymbolInScope(sym: Symbol, prefix: Type = NoPrefix): Boolean = {
```
tgodzik, line 135: Changes here and the two methods below come from:
https://github.com/scalacenter/scalafix/blob/69c0369ab49c41b0473ca68fe0a8ade396d3157e/scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala#L382

mtags/src/main/scala-2/scala/meta/internal/pc/Signatures.scala
```diff
@@ -138,52 +132,90 @@ trait Signatures { this: MetalsGlobal =>
...
-    def nameResolvesToSymbol(name: Name, sym: Symbol): Boolean = {
+    def nameResolvesToSymbol(
+        name: Name,
+        sym: Symbol,
+        prefix: Type = NoPrefix
+    ): Boolean = {
       lookupSymbol(name) match {
-        case Nil => true
-        case lookup => lookup.exists(_.symbol.isKindaTheSameAs(sym))
+        case Nil => false
```
tgodzik, line 144: This is a change I did here, since if lookup did not find anything then it should be false. Indeed a lot of tests seem to have unneccesarily long names.

mtags/src/main/scala-2/scala/meta/internal/pc/Signatures.scala
```diff
@@ -226,6 +258,12 @@ trait Signatures { this: MetalsGlobal =>
     }
   }
 
+  implicit class XtensionNameMetals(name: Name) {
+    def otherName: Name =
+      if (name.isTermName) name.toTypeName
```
tgodzik, line 263: Here we try to find type name or term name with the same name.

mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
```diff
@@ -188,7 +188,9 @@ trait OverrideCompletions { this: MetalsGlobal =>
         }
 
       val history = new ShortenedNames(
-        lookupSymbol = { name => context.lookupSymbol(name, _ => true) :: Nil },
+        lookupSymbol = { name =>
+          context.lookupSymbol(name, sym => !sym.isStale) :: Nil
```
tgodzik, line 192: Without isStale, we would pick up classes that are no longer here. This actually showed up in `CompletionMatchSuite.stale2`, which seems to now have worked correctly.

tests/cross/src/test/scala/tests/pc/CompletionIssueSuite.scala
```diff
@@ -56,18 +56,16 @@ class CompletionIssueSuite extends BaseCompletionSuite {
...
-      |  NestedLeaf
-      |}""".stripMargin
+    """|package a
+       |object A {
+       |  object Nested{
+       |    object NestedLeaf
+       |  }
+       |}
+       |object B {
+       |  A.Nested.NestedLeaf
```
tgodzik, line 66: This is caused by the change to minimize the length of anything imported.

ckipp01 (approved):
👍🏼 some cool improvements in here. Definitely best for someone else that is probably more familiar with these parts than me to take a look, but from what I can see, LGTM

tgodzik:
> 👍🏼 some cool improvements in here. Definitely best for someone else that is probably more familiar with these parts than me to take a look, but from what I can see, LGTM

This was essentially already reviewed :sweat_smile: , but for sure it would be better if @olafurpg could confirm that it makes sense to port those. I think it does at least from what I see.

olafurpg:
It makes a lot of sense to port the implementation from Scalafix. The Scalafix implementation started as a fork of the Metals implementation but then fixed a long tail of bugs when I was running ExplicitResultTypes against Twitter’s monorepo.

tgodzik:
@tanishiking this is connected to be the short names that you started to implement. We don't have symbol search in Scala 3 part yet, so it should most likely be simpler.

mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
```diff
@@ -194,80 +195,126 @@ class MetalsGlobal(
...
+              case _ =>
+                history.renames.get(sym) match {
+                  case Some(rename)
+                      if history.nameResolvesToSymbol(rename, sym) =>
+                    TypeRef(
+                      NoPrefix,
+                      sym.newErrorSymbol(rename),
+                      args.map(arg => loop(arg, None))
+                    )
+                  case _ if history.isSymbolInScope(sym, pre) =>
```
tanishiking, line 229: Is there any pattern to reach this case? I guess this pattern will be captured by https://github.com/scalacenter/scalafix/blob/69c0369ab49c41b0473ca68fe0a8ade396d3157e/scalafix-rules/src/main/scala/scala/meta/internal/pc/ScalafixGlobal.scala#L177 in the first place.

mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
```diff
@@ -194,80 +195,126 @@ class MetalsGlobal(
       isVisited += key
       val result = tpe match {
         case TypeRef(pre, sym, args) =>
-          val ownerSymbol = pre.termSymbol
-          def hasConflictingMembersInScope =
-            history.lookupSymbol(sym.name).exists {
-              case _: LookupSucceeded => true
-              case _ => false
-            }
+          if (history.isSymbolInScope(sym, pre)) {
```
tanishiking, line 198: [note]
- https://github.com/scalacenter/scalafix/commit/f2a8bcda692ae04b37553a7ba16b08d8e7bc7a69
- https://github.com/scalacenter/scalafix/commit/6af6cfc6dcb1c3fdd573b9b612b5f72378a7a0db

> Previously, the rewrite would insert names like `Value` in the position
of a path dependent type such as `Day.Value` even if the `Value` in
scope had a conflicting type `Bool.Value`. Now, when querying the
scope we take into account the prefix from where the symbols in scope
origin.

> The implementation in this PR is probably incorrect for some cases. I
had to use a hacky workaround for the custom `PrettyType` type. Either
way, this diff makes the test suite pass and allow us to move on.

👍

tanishiking (commented):
@tgodzik
Since I gonna start off with shortening names only for hover (which is used for showing to the users and not used for a rewrite or imports and that's why not so strict for conflicting types) it'll be simple in the first step.
However, when we implement features like override completions or auto imports, indeed we need to do kind the same thing in scala3. Thank you for the mention!

Btw, I took a quick look and left a nitpick, otherwise look good, and happy to see https://github.com/scalameta/metals/issues/1253 is fixed :)

mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
tgodzik: Good catch! Removed it!

PR #2503, opened 2021-02-13
Fix issues where created files should use backticks

### LINKED ISSUE #2375: Creating new scala file doesn't escape keywords in package statement
**Describe the bug**
When creating a new Scala file, if the folder in the package path contains a reserved keyword, such as `type`, the package statement won't escape that keyword.

**To Reproduce**

1. Open Scala project in VS Code
2. Create new folder at any level inside either `src/main` or `src/test`, with the name **type**
3. Create any Scala file inside that folder
4. Package declaration includes keyword without backticks.

**Expected behavior**

Package declaration should escape reserved keywords. E.g. if the new Scala file is created inside the folder path `src/main/com/example/type` then the package declaration should be ```package com.example.`type` ```

**Screenshots**

Screenshot 2021-01-12 at 19 26 34

**Installation:**

- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.9.8

tgodzik:
Previously, we would not any backticks when generating files from the `New Scala File` command or when a file was created with a package name inserted automatically. Now, we wrap each part in backticks if needed.

Fixes #2375

### DIFF
--- a/tests/unit/src/test/scala/tests/AddPackageLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/AddPackageLspSuite.scala
@@ -1,11 +1,12 @@
 package tests
 
 import java.nio.file.Files
+import java.nio.file.Paths
 
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.RecursivelyDelete
+
 import munit.TestOptions
-import java.nio.file.Paths
 
 class AddPackageLspSuite extends BaseLspSuite("add-package") {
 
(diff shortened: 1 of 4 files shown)


tests/unit/src/test/scala/tests/AddPackageLspSuite.scala
```diff
@@ -1,180 +1,105 @@
...
+import java.nio.file.Paths
 
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.RecursivelyDelete
 
+import munit.TestOptions
+
 class AddPackageLspSuite extends BaseLspSuite("add-package") {
 
-  test("single-level") {
```
tgodzik, line 10: I refactored a bit the tests here. That's done in a separate PR

ckipp01 (approved):
Looks good! LGTM!

Issue #2495, opened 2020-04-30
Better names for types in hover/completions in Scala 3


tgodzik:
**Is your feature request related to a problem? Please describe.**
Currently, Scala 3 shows the  fully qualified name for each type, which sometimes causes it not to show everything that is needed.

"com.something.very.long.ButUniqueName" should show most likely just "ButUniqueName"

**Describe the solution you'd like**
This should work as it worked in Scala 2

**Search terms**:
hover completions scala3

tgodzik:
First part of the functionality was merged in https://github.com/scalameta/metals/pull/2504

We still need to add auto imports for the signatures to always be shorter irrelevant of whether the symbol is in scope or not. We should also make types shorter for completions.

tanishiking:
I finally started taking a look at prettifying names for type in completions.
I'll work on it this month 💪 

Here are rough TODOs.

- [ ] Pretty print method symbol
  - [ ] shorten signature and print it 
    - https://github.com/scalameta/metals/blob/45eaa4151ac5f54cabddce73de3d888d5157b1f3/mtags/src/main/scala-2/scala/meta/internal/pc/Signatures.scala#L329
  - [ ] hide implicit evidence param
  - [ ] print context bounds 
    - https://github.com/scalameta/metals/blob/45eaa4151ac5f54cabddce73de3d888d5157b1f3/mtags/src/main/scala-2/scala/meta/internal/pc/Signatures.scala#L390-L396
  - [ ] (not needed for completion) print default value 
    - https://github.com/scalameta/metals/blob/45eaa4151ac5f54cabddce73de3d888d5157b1f3/mtags/src/main/scala-2/scala/meta/internal/pc/Signatures.scala#L399
- [ ] Pretty print other symbols
  - dealias type alias
    - https://github.com/scalameta/metals/blob/45eaa4151ac5f54cabddce73de3d888d5157b1f3/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala#L234
- [ ] Compute type parameters based on the qualifier.	
  - https://github.com/scalameta/metals/blob/45eaa4151ac5f54cabddce73de3d888d5157b1f3/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala#L258-L262

tgodzik:
Ok cool, thanks for the info!

tgodzik:
@tanishiking do you think we can close this? We have better signatures currently and I would open new issues if we find any problematic cases.

[...]

PR #2504, opened 2021-02-14
Basic shorter signature support for hover in Scala3

tanishiking:
https://github.com/scalameta/metals/issues/2495
Previously, Metals on Scala 3 shows the fully qualified type names on hover, which is difficult to read. This PR enables Metals to shorten the signatures as long as the name can be resolved from the specific context.

This implementation would not work perfectly for other usages than hover like `auto-import` and `override def completions`, but this would be a good starting point for shortening names, and at least works well for prettying hover texts.

- How do we look up the symbols by name?
  - using [Typer#findRef](https://github.com/lampepfl/dotty/blob/64aa7cfb0a0a245261c490e268052880a07f5c00/compiler/src/dotty/tools/dotc/typer/Typer.scala#L135-L142), and I guess this is a decent choice because `findRef` is used for [typedIdent](https://github.com/lampepfl/dotty/blob/64aa7cfb0a0a245261c490e268052880a07f5c00/compiler/src/dotty/tools/dotc/typer/Typer.scala#L472-L492) in Scala3 typer, as [Context#lookupSymbol](https://github.com/scala/scala/blob/f262a1cc0fa1de97a91529b62408e91db1279e70/src/compiler/scala/tools/nsc/typechecker/Typers.scala#L5466) does in scala2.
    - `findRef` looks try to look up a type that corresponds to the given name, from the starting context to the root context, and from imports
- How to determine the starting context from the given position?
  - use `Interactive.contextOfPath` which is used for [calculating the completions](https://github.com/lampepfl/dotty/blob/64aa7cfb0a0a245261c490e268052880a07f5c00/compiler/src/dotty/tools/dotc/interactive/Completion.scala#L48) too.

#### before

Screen Shot 2021-02-13 at 14 08 13

#### after
Screen Shot 2021-02-14 at 17 25 06

- SrcPos was not shortened because the name `SrcPos` cannot be resolved from the scope (since it's not imported).
- `using` clause renamed to `implicit`, which is better to fix.

#### before

Screen Shot 2021-02-13 at 14 08 32

#### after
Screen Shot 2021-02-14 at 17 25 41

- `dotty.tools.dotc.ast.tpd.Tree` is shortened to `tpd.Tree`, because `Tree` can not be resolved from the given context, but `tpd` can be resolved.

### references
- [Scala 3/Internals/Type System](https://dotty.epfl.ch/docs/internals/type-system.html)


mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -490,4 +500,135 @@ case class ScalaPresentationCompiler(
...
+    def loop(tpe: Type, name: Option[ShortName]): Type = {
+      val key = tpe -> name
+      // NOTE: Prevent infinite recursion, see https://github.com/scalameta/metals/issues/749
+      if (isVisited(key)) return cached.getOrDefault(key, tpe)
+      isVisited += key
+      val result = tpe match {
+        case TypeRef(prefix, designator) =>
+          // designator is not necessarily an instance of `Symbol` and it's an instance of `Name`
+          // this can be seen, for example, when we are shortening the signature of 3rd party APIs.
+          val sym =
```
tgodzik, line 580: Can you extract it to a def ?

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -490,4 +500,135 @@ case class ScalaPresentationCompiler(
...
+            else tpe.typeSymbol
+          val short = ShortName(sym)
+          TypeRef(loop(prefix, Some(short)), sym)
+
+        case TermRef(prefix, designator) =>
+          val sym =
+            if (designator.isInstanceOf[Symbol]) designator.asInstanceOf[Symbol]
+            else tpe.termSymbol
+          val short = ShortName(sym)
+          if (sym.isOneOf(Flags.Package | Flags.Module) && history.tryShortenName(name)) NoPrefix
```
tgodzik, line 591: Why `sym.isOneOf(Flags.Package | Flags.Module) ` is needed? Do you have examples of this in the tests?

tgodzik (changes requested):
Great work @tanishiking ! I left just a few small comments.

tgodzik:
Hey @tanishiking it would be awesome to include your PR in the next release, is it ok if I update it myself? I want to work on the release tomorrow.

tanishiking:
@tgodzik Ok! actually, I'm, bit busy this week and difficult to find the time until next weekend :/ so it'd nice if you finish up the PR instead, and include it in the next release 👍

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
tanishiking: Aw, realized actually we don't need it!
Besides, we don't need to pass `Some(short)` to `loop` in the else clause (since we already know the `sym` cannot be resolved from the scope in the else clause).

```diff
diff --git a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
index cfb3e1ca..2298a0ff 100644
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -588,8 +588,8 @@ case class ScalaPresentationCompiler(
             if (designator.isInstanceOf[Symbol]) designator.asInstanceOf[Symbol]
             else tpe.termSymbol
           val short = ShortName(sym)
-          if (sym.isOneOf(Flags.Package | Flags.Module) && history.tryShortenName(name)) NoPrefix
-          else TermRef(loop(prefix, Some(short)), sym)
+          if (history.tryShortenName(name)) NoPrefix
+          else TermRef(loop(prefix, None), sym)
 
         case ThisType(tyref) =>
           if (history.tryShortenName(name)) NoPrefix
```

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -247,7 +248,16 @@ case class ScalaPresentationCompiler(
                 case _: ImportType =>
                   printer.typeString(symbol.paramRef(using ctx))
                 case _ =>
-                  printer.typeString(tpw)
+                  val shortendType = driver.compilationUnits.get(uri) match {
+                    case Some(unit) =>
+                      implicit val newctx = ctx.fresh.setCompilationUnit(unit)
+                      val path = Interactive.pathTo(newctx.compilationUnit.tpdTree, pos.span)
+                      val context = Interactive.contextOfPath(path)
```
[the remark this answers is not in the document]
tgodzik: Doesn't seem to work with that.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
tgodzik: Ach no, that's a bit different. NVM

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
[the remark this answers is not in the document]
tanishiking: I'm not 100% sure, but I guess it doesn't work because the `path` defined in line 226, is based on an untyped tree whose denotation doesn't have a corresponding symbol yet.
Therefore, here we recalculate the path using a typed tree (referring https://github.com/tanishiking/dotty/blob/64aa7cfb0a0a245261c490e268052880a07f5c00/compiler/src/dotty/tools/dotc/interactive/Completion.scala#L47)

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
tgodzik: Looks like it's take from the previous implementation, we can leave that in.

tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+
+import tests.pc.BaseHoverSuite
+import tests.BuildInfoVersions
+
+import java.nio.file.Files
+
+class HoverScala3TypeSuite extends BaseHoverSuite {
+
+  override protected def excludedScalaVersions: Set[String] =
+    Set(
```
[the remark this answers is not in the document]
tgodzik: Fixed.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
tanishiking: > Looks like it's take from the previous implementation, we can leave that in.

Actually, yes I took it from the previous implementation and didn't realize we don't need it 🤦
I think it's better to remove it at least now so that we won't be confused with this (like why we need this condition?) in the future.

tgodzik: Ok removed that, thanks!

mtags/src/main/scala-3.0.0-M3/scala/meta/internal/mtags/VersionSpecificEnrichments.scala
```diff
@@ -0,0 +1,23 @@
...
+
+trait VersionSpecificEnrichments {
+
+  extension(context: Context) {
+    def findRef(name: Name): Type = {
+      context.typer.findRef(
+        name,
+        defn(using context).AnyType,
+        EmptyFlags,
+        EmptyFlags,
```
tgodzik, line 18: There is one extra parameter here, which is missing in M1 and M2

tanishiking: Oops, didn't realize it, thanks for fixing it 👍 
it looks good! (and good to see the use of the `extension` keyword!)

tgodzik (approved):
Thanks for @tanishiking ! Going to merge it and start preparing the release.

PR #2515, opened 2021-02-15
Update scalameta and fix tests that include Scala 3 givens

tgodzik:
Previously there was a bug in scalameta that would not include the given keyword in the actual given related AST. We fixed it in 4.4.8, which is why we also need to fix it here.

The changes were done by running `sbt save-expect`


ckipp01 (approved):
Looks good. LGTM!

tgodzik:
Aaaa, that was supposed to be auto-merge. I need another coffee :sweat:

PR #2516, opened 2021-02-15
add explanation on how to annotate stack traces for Emacs

ag91:
From your last blog I got you were missing an explanation on how to annotate stack traces. I did not know about this feature, and now I am using it! Also found there is likely a bug in lsp-mode.
Anyway any feedback for this documentation would be useful!

### DIFF
--- a/docs/editors/emacs.md
+++ b/docs/editors/emacs.md
@@ -125,6 +125,47 @@ usage with Emacs.
 
 ```
 
+## LSP Tips
+
+### Show navigable stack trace
+
+You can annotate your stack trace with code lenses (which requires the
+following bit of configuration mentioned earlier: `(lsp-mode . lsp-lens-mode)`). 
+These allow you to run actions from your code.
+
+One of these actions allow you to navigate your stack trace.
+
+You can annotate any stack trace that lives in your `lsp` project by
+marking it with your region and using `M-x lsp-metals-analyze-stacktrace`. 
+
+This will open a new Scala buffer that has code lenses annotations:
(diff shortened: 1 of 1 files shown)


docs/editors/emacs.md
```diff
@@ -125,6 +125,47 @@ usage with Emacs.
...
+### Show navigable stack trace
+
+You can annotate your stack trace with code lenses (which requires the
+following bit of configuration mentioned earlier: `(lsp-mode . lsp-lens-mode)`). 
+These allow you to run actions from your code.
+
+One of these actions allow you to navigate your stack trace.
+
+You can annotate any stack trace that lives in your `lsp` project by
+marking it with your region and using `M-x lsp-metals-analyze-stacktrace`.
```
ckipp01, line 139: So this part may be a bit confusing since one might wonder what it means to "live" in your lsp project. I'm not 100% sure how it works in emacs, but I know in the other clients as long as you have the stacktrace copied in your local buffer, when you execute the `analyze-stacktrace` command, it pulls it from the buffer and then gives you the new one with code lenses. So it might be better to clarify that here if that is the case.

ckipp01 (changes requested):
Thanks for the contribution! Just a quick note.

docs/editors/emacs.md
ag91: thanks for the feedback, I have tried to make that clearer

ckipp01 (approved):
Awesome, thanks for the update! LGTM!

PR #2518, opened 2021-02-16
Revert https://github.com/scalameta/metals/commit/519c6414797edcc35d5d714b220c7d7d140c4b4

ckipp01:
@tgodzik and I talked a bunch off line about this but the jist of the issue is that with this change if you open `test.worksheet.sc` directly now it won't evaluate. Even after that if you directly open another `second.worksheet.sc` it still won't evaluate since focused document isn't changed on the `didOpen` after the listed commit.

We need to rethink this a bit since the original issue with VS Code still exists.


tgodzik (approved):
:cry:

ckipp01:
> 😢

haha yes, sad.

PR #2519, opened 2021-02-16
Revert to recently focused document if the current one was immedietally closed

tgodzik:
Some clients when using peek functionality might send didOpen and didClose notifications immediately one after the other. In this case we don't actually want to change the focused document so we revert to the one that was most recently focused.


ckipp01 (approved):
Yea this seems to help 👍🏼 LGTM

PR #2520, opened 2021-02-16
Handle Java-only build targets

Arthurm1:
The idea here is to make Metals aware of Java-only (i.e. non-Scala) build targets, which it currently discards.

This would mean that Java targets would appear in the Metals Tree View and the Doctor, and would be included in the `clean compile workspace` command.

Bloop currently doesn't support `buildTarget/JavacOptions` endpoint which is needed for this.
There is a [Pull Request here](https://github.com/scalacenter/bloop/pull/1397), but I'm not sure what the state of it is.
For now I'm faking `buildTarget/JavacOptions` in `scala.meta.internal.metals.ImportedBuild` by translating them from the scalacOptions.  This can be dropped and the javacOptions endpoint could be used if/when Bloop supports it.
I see Intellij BSP attempts to ask Bloop for `buildTarget/JavacOptions` and gets an error.

I was hoping for some feedback on whether this is the right approach to take and what Metals is trying to do in the 10 places I've commented with `TODO` so that I'll know whether to handle java-only targets there as I'm unfamiliar with the majority of the codebase.

There's not much to the PR.  Most of it is splitting `BuildTargets#allScala` into `BuildTargets#allScala`, `BuildTargets#allJava`, `BuildTargets#allCommon` and the fallout from that.


metals/src/main/scala/scala/meta/internal/metals/FileSystemSemanticdbs.scala
```diff
@@ -22,6 +22,7 @@ final class FileSystemSemanticdbs(
 ) extends Semanticdbs {
 
   override def textDocument(file: AbsolutePath): TextDocumentLookup = {
+    // TODO need to handle java targets?
```
tgodzik, line 25: This should be possible via plugin recently released that can produce semanticdb for Java files: https://github.com/sourcegraph/lsif-java

We would probably need to merge it all together for it to make sense.

metals/src/main/scala/scala/meta/internal/metals/FileWatcher.scala
```diff
@@ -111,6 +111,7 @@ final class FileWatcher(
     // Watch the source directories for "goto definition" index.
     buildTargets.sourceRoots.foreach(watch(_, isSource = true))
     buildTargets.sourceItems.foreach(watch(_, isSource = true))
+    // TODO - do java target have to be handled as well?
```
tgodzik, line 114: Once we introduce semanticdb plugin for java then we would need to handle it too.

metals/src/main/scala/scala/meta/internal/metals/ImportedBuild.scala
```diff
@@ -68,9 +75,37 @@ object ImportedBuild {
         new DependencySourcesParams(ids)
       )
     } yield {
+      // waiting on Bloop buildTarget/javacOptions support
```
tgodzik, line 78: I can take a look at that at some point and work towards merging it, this would probably be a better way.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1075,6 +1075,7 @@ class MetalsLanguageServer(
   }
 
   private def didCompileTarget(report: CompileReport): Unit = {
+    // TODO need to handle java targets?
```
tgodzik, line 1078: only once semanticdb is produced via the plugin

metals/src/main/scala/scala/meta/internal/metals/debug/BuildTargetClasses.scala
```diff
@@ -122,6 +122,7 @@ final class BuildTargetClasses(
   private def descriptorsForMainClasses(
       buildTarget: b.BuildTargetIdentifier
   ): List[String => Descriptor] = {
+    // TODO how to handle Java Target here. "List(Descriptor.Term)", "List(Descriptor.Type)", "List(Descriptor.Term, Descriptor.Type)"
```
tgodzik, line 125: `Descriptor.Type` I think since java doesn't have objects

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
```diff
@@ -60,6 +70,7 @@ class SemanticdbIndexer(
    * and re-index all of its `*.semanticdb` children.
    */
   def onOverflow(): Unit = {
+    // TODO do java targets have to be handled as well?
```
tgodzik, line 73: They should - it's the same case as previously. We need to add the plugin.

tests/unit/src/main/scala/tests/MetalsTestEnrichments.scala
```diff
@@ -97,6 +97,7 @@ object MetalsTestEnrichments {
         libraries.flatMap(_.classpath.entries).map(_.toURI.toString).asJava,
         ""
       )
+      // TODO test javacOptions?
```
tgodzik, line 100: would probably be useful once we start testing

tgodzik (commented):
Thanks for the contribution @Arthurm1 ! Sorry it took me so long to get to it.

We ignore java build targets mostly due to the fact that we can't do anything with them currently. We still haven't figured out how to properly interoperate with a Java Language Server. There is related issue https://github.com/scalameta/metals-feature-requests/issues/5 that we need to handle in order to properly support java projects.

Although, recently @olafurpg worked on a java compiler plugin that can easily produce semanticdb files: 

https://github.com/sourcegraph/lsif-java

This would enable us to support references, renames, implementations and better go to definition. To include that plugin in Metals we would most likely make the changes that you made in this PR, so this is I think a good start. I will need to spend some time working on Bloop next couple of weeks, so I can also take a look at the javacOptions endpoint. With that in place we could add the sourcegraph plugin and at least enable the previously mentioned features.

We will still need to figure out how to provide completions, hovers, signature help etc. We could probably use JDT, but I am not sure how to do it properly yet. Alternatively, we could try to start a java language server alongside metals, but I haven't done any research yet in that area.

metals/src/main/scala/scala/meta/internal/metals/debug/BuildTargetClasses.scala
olafurpg: I think Descriptor.Type since Java main methods are classes.

Arthurm1:
Thanks for reviewing.

In my mind this was an early stage of a lot more stages for Java support.  Those next stages are probably beyond my knowledge.  The reason I wanted this is that I had a Bloop PR for some level of Gradle Android support.  Without Bloop's support for javacOptions I couldn't see the targets in Intellij.  This was one way to see the targets in a UI.

Didn't know there was a SemanticDB plugin coming along for Java - that's great news - go Olafur!

I'd eventually like to see more done with build targets e.g. highlighting of source/resource directories in VSCode file explorer, context menus in the file explorer for running tests, finer control over clean/compile per target, target info pages showing dependencies/classpath,  new-scala-file (and new java file) restricted to source dirs.  But those are all nice-to-haves.

I'll keep the PR as draft until javacOptions is supported.  Feel free to say if you think Java support needs to go in a completely different direction.

Arthurm1:
@tgodzik

I've converted to using JavacOptions (rather than faking them) and failing gracefully if BSP doesn't support it.

I was going to have a look at Bloop to see if it's easy enough to get JavacOptions supported there unless you've already started?

tgodzik:
> @tgodzik
> 
> I've converted to using JavacOptions (rather than faking them) and failing gracefully if BSP doesn't support it.
> 
> I was going to have a look at Bloop to see if it's easy enough to get JavacOptions supported there unless you've already started?

I haven't looked into it yet. I was mostly focused on zinc migration.

Arthurm1:
I've taken the existing [Bloop JavacOptions PR](https://github.com/scalacenter/bloop/pull/1397), deployed locally and tested against this PR.  All Java-only projects get recognised as build targets. They appear in the "Doctor", get compiled on "Clean compile workspace" and are navigable in the Metals Tree View.

I've also tested against Intellij+BSP.  Intellij was fine before but now it doesn't complain that Bloop doesn't support JavacOptions and it doesn't have any issues now that Bloop does support JavacOptions

The Bloop PR looks complete.  I didn't need to change any code - it just needed a rebase which only has a single line conflict.  I think it's ready to merge post-rebase.  I'm not sure what the protocol is for taking another's PR - should I create my own PR with their commit?  I asked the author back in Jan about the status and got no response - but I will try again.

tgodzik:
@Arthurm1 Since it's more important now I will contact the author today and try to get it merged. Maybe release `1.4.9` before the zinc changes.

tgodzik:
The Bloop PR is merged, you can use `1.4.8-15-10dc7269` Bloop version in this PR.

Arthurm1:
@tgodzik
Is the intention to eventually add a `javaSemanticDBVersion` entry to `bloop.settings.json` to allow Metals to specify the lsif-java semanticdb plugin in the same way that it can specify the scala plugin, and then Bloop would auto-add the necessary options to the javac options?

tgodzik:
> @tgodzik
> Is the intention to eventually add a `javaSemanticDBVersion` entry to `bloop.settings.json` to allow Metals to specify the lsif-java semanticdb plugin in the same way that it can specify the scala plugin, and then Bloop would auto-add the necessary options to the javac options?

That is is exactly the plan :sweat_smile:

Arthurm1:
Has anyone started work on that part already? I thought I might give it a go.

Arthurm1:
@tgodzik  So I went ahead and implemented the `javaSemanticDBVersion` in `bloop.settings.json` and it successfully generates semanticdb files.

The problem is that I'm getting weird highlighting in the Java files when I click on a class/interface name...

image

Note that I clicked on `AutoImportsResult` but VSCode has highlighted the chars partially to the left of the class - by exactly the length of `interface` + a space char.

Is this something I have to handle in Metals or would this be the LSIF plugin supplying wrong position range info - I wasn't sure if I should raise an issue with the LSIF project or whether I need to delve into the Metals code

tgodzik:
> Has anyone started work on that part already? I thought I might give it a go.

We planned to include it as part of GSoC, but it's fine to implement it before. There will be some work required to also include JDT, so we can focus on that there.

> @tgodzik So I went ahead and implemented the `javaSemanticDBVersion` in `bloop.settings.json` and it successfully generates semanticdb files.
> 
> The problem is that I'm getting weird highlighting in the Java files when I click on a class/interface name...
> 
> image
> 
> Note that I clicked on `AutoImportsResult` but VSCode has highlighted the chars partially to the left of the class - by exactly the length of `interface` + a space char.
> 
> Is this something I have to handle in Metals or would this be the LSIF plugin supplying wrong position range info - I wasn't sure if I should raise an issue with the LSIF project or whether I need to delve into the Metals code

I think that might be the LSIF plugin. You can check the positions by using the `metap` tool `coursier install metap`, which can print out semanticdb file and we should be able to see what's going on there. 

Thanks for taking a look at that!

olafurpg:
@Arthurm1 That issue has been fixed in the latest SNAPSHOT of semanticdb-javac. We're planning to cut a new release this week so it should not be a problem soon :)

Don't hesitate to report other similar issues to https://github.com/sourcegraph/lsif-java/ We're actively working on the project.

olafurpg:
BTW, the command `lsif-java snapshot DIRECTORY_WITH_SEMANTICDB_FILES/` will generate human-readable files in the `generated/` directory that you can use to debug position issues like these. The `lsif-java` cli will soon be installable via `coursier install lsif-java`

Arthurm1:
Thanks @tgodzik 

I'll be stopping before JDT :-)  I just wanted my java targets to appear in Metals and the existence of LSIF plugin was an added bonus - although I'm not entirely sure what parts of functionality the plugin on its own delivers to Metals - references but not code completion?  Anything else?

@olafurpg Thanks for fixing - preemptively :-)  I'll report any more issues now I know where the Metals/LSIF line is drawn.

tgodzik:
> Thanks @tgodzik
> 
> I'll be stopping before JDT :-) I just wanted my java targets to appear in Metals and the existence of LSIF plugin was an added bonus - although I'm not entirely sure what parts of functionality the plugin on its own delivers to Metals - references but not code completion? Anything else?

References, implementation, anything that needs to operate on the entire codebase at once. We use semanticdb in those case for indexes.

Arthurm1:
Linked to scalacenter/bloop/pull/1534

@tgodzik
There's a couple of bits for me to tidy and tests to add but this gets Bloop to add the LSIF plugin and retrieves Java-only build targets.

I was wondering whether you're happy with the way I've split out `ScalaTarget` into its mini-hierarchy of `CommonTarget`, `ScalaOnlyTarget` `JavaTarget` etc. before I start looking at tests.

Also - I'm not sure where to go next.  I assume there has to be some kind of Java equivalent to `ScalaPresentationCompiler` or can the Java stuff be put in with the `ScalaPresentationCompiler`?  At what point can Metals delegate stuff to the Eclipse JDT library - would that library implement something akin to `JavaPresentationCompiler`?

tgodzik:
> I was wondering whether you're happy with the way I've split out `ScalaTarget` into its mini-hierarchy of `CommonTarget`, `ScalaOnlyTarget` `JavaTarget` etc. before I start looking at tests.

I will take a look at that, but most likely next week. From what I remember it looked good, but I last looked at it a while ago.

> 
> Also - I'm not sure where to go next. I assume there has to be some kind of Java equivalent to `ScalaPresentationCompiler` or can the Java stuff be put in with the `ScalaPresentationCompiler`? At what point can Metals delegate stuff to the Eclipse JDT library - would that library implement something akin to `JavaPresentationCompiler`?

I actually played a bit with it and it seems to use JDT we need to setup OSGI inside Metals, it should be possible, but I haven't managed to make it work properly. It also seems we need to create the entire project hierarchy, which also doesn't seem too straightforward. 

That said, JDT has interfaces for completions, formatting etc. which we could use in Metals to at least provide the basic functionalities, so this is for sure good place to start looking.

I was also thinking if we could actually use the Java language server somehow and redirect a subset of queries to it :thinking: - we would most likely need to generate eclipse files (that's what Java LSP uses) and maybe then start the other server inside Metals? This way we would at least have the same functionalities as the current server has and we wouldn't need to implement everything from scratch. We could even set it up as a separate server and use LSP to communicate with it (forward completions, fallback definition etc.) This might actually be more doable than using JDT explicitely, but I am not sure if it will all play nicely.

We could have separate JavaPresentationCompiler that would just start the Java Language Server underneath and we would run `sbt eclipse` or the gradle/maven equivalent before starting.

This is all just my ideas, we need to check if any of that makes sense. I am kind of tempted to use Java Language Server, because we would not need to reimplement everything ourselves this way, but it's still a lot of work.

metals/src/main/scala/scala/meta/internal/metals/JavaTarget.scala
```diff
@@ -0,0 +1,23 @@
...
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.io.AbsolutePath
+
+import ch.epfl.scala.bsp4j.JavacOptionsItem
+
+trait JavaTarget extends CommonTarget {
+
+  val javac: JavacOptionsItem
+
+  def isJavaSemanticdbEnabled: Boolean = javac.isSemanticdbEnabled
```
tgodzik, line 14:
```suggestion
  def semanticdbEnabled: Boolean = javac.isSemanticdbEnabled
```
if we put it in the main trait, we could just return `Target` or `CommonTarget`  and maybe only check a method `isJavaOnlyTarget` when needed.

tgodzik (commented):
I feel like there is too much classes here, could we reduce it to a main trait `Target` and two subtraits `ScalaTarget` and `JavaTarget`, where scala can contain also javac options? I would then use the main trait wherever possible, otherwise I feel like we are not leveraging inheritance enough.

We could even return `scalaVersion` for JavaTargets, which would be the `fallbackScalaVersion`, since any scala file in there will be a standalone file.

tgodzik:
Alternatively, I would just have a single `CommonTarget` class with some options like `Option[ScalacOptions]` etc. and merge them inside the `CommonTarget` object.

Arthurm1:
These classes are really just wrappers round the java BSP4J classes and designed to extract info from them.

I've shifted the extraction to `MetalsEnrichments` and `BuildTargets`

`CommonTarget`, `JavaTarget` and `ScalaTarget` are now stand alone classes and only created when multiple bits of info are needed from the build target.  I've removed the other classes.

tgodzik:
Looks pretty good, let's wait until the Bloop PR merge and test it out again.

When testing it were you able to navigate java files? I wonder if we need to do something on the client side to make it work, but maybe that's enough.

Arthurm1:
> When testing it were you able to navigate java files? I wonder if we need to do something on the client side to make it work, but maybe that's enough.

Yes - in a Java file in VSCode you can now:  `Ctrl + mouse hover` to see declaration, `Go to Definition`, `Find All References` and clicking on a symbol highlights any usage of that symbol throughout the same file.
All that only started working after https://github.com/scalameta/metals/pull/2520/commits/1f8c0c1cd49f191b14752a602d522d4ce18c1ba2 - so not the initial commit - I can squash these if you prefer.

Code completions don't work but I'm assuming that's because there's no presentation compiler.  Nor do formatting and re-factoring but just being able to navigate symbols makes using Java files a lot easier.

Error highlighting doesn't highlight the correct range (just highlights the start of the line) but that should be fixed when https://github.com/scalacenter/bloop/pull/1477 is merged

Arthurm1:
Added support for Java file formatting (range and full document but _not_ onType)

Uses `eclipse-formatter.xml` files

Arthurm1:
Added "New Java File" command

gravelld:
Would this allow a Java only project with code inside it that depended on a separate Scala only project (and defined as such using Gradle)? Because at the moment, that's what's keeping me from using Metals. I actually thought this was a Bloop problem.

Arthurm1:
Yes it would.  https://github.com/scalacenter/bloop/pull/1534 also needs to be merged.

This goes some way towards Java support (references, code navigation, formatting) but there's still a lot to do (code completions, refactoring...).  I imagine this PR will get merged as is and the rest will be follow-up work.

gravelld:
In this case, what would the user do about separate Java plugins? Would they need to disable whatever Java plugin they were using to avoid conflicting compilation etc?

Arthurm1:
@tgodzik what's your thinking on merging this - are you snowed under with other stuff?

tgodzik:
Not at all, if you think it's ready let's rebase the pr and o can tak a look 🥳

Arthurm1:
Stuff that works on java files...

- references
- goto
- new java file
- formatting (file or range)
- decorations (e.g. super)

Stuff that doesn't...
 - No presentation compiler so...
     - the above stuff that works, won't work when the java file doesn't compile
     - code completion
     - format on type
 - Dependency sources - currently find references/goto doesn't work with jar/src.zip java files in the readonly area.  It should since it's got all the semantic info.  I think that's because of the disconnect between jar: uri and the readonly file uri.  With virtual docs (which i've had no time to finish) this should be added as the uri is the uri.

In terms of adding Metals config options - this adds a few java formatting options for eclipse java file formatting which increases the number of settings Metals has but I think it's the minimum possible.

Error underlining will improve with https://github.com/scalacenter/bloop/pull/1477, otherwise java files just get the first char of the line underlined but that hasn't changed.  You've got the hardest job @tgodzik but this doesn't affect this PR - just the feel of using java files under Metals.

There are very few tests - not sure what scala tests should be replicated for java files - there's new file tests but that's it.
I fully expect this PR to raise new issues but hopefully only issues around this new functionality. Gotta pull the trigger on it sometime :shrug:

This needs https://github.com/scalameta/metals-vscode/pull/670 to work on vscode

Arthurm1:
Hmmm - not sure these are relevant - earlier test failures were but these don't look like they are - but I'm not sure how they'd be impacted by these changes

tgodzik:
> Hmmm - not sure these are relevant - earlier test failures were but these don't look like they are - but I'm not sure how they'd be impacted by these changes

It could be a classpath change adding some completion options. Might not be an issue if the completions still make sense.

I am looking slowly through the PR, great work! :tada:

tgodzik:
There are some weird issues when looking at the logs:

Looks like sourcegraph plugin has issues with JDK 17 @olafurpg any idea what's going on?
```
2021.11.29 08:23:18 ERROR java.lang.IllegalAccessError: class com.sourcegraph.semanticdb_javac.SemanticdbPlugin (in unnamed module @0x6468c45b) cannot access class com.sun.tools.javac.api.BasicJavacTask (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.api to unnamed module @0x6468c45b
```

metals/src/main/scala/scala/meta/internal/metals/JavaFormattingProvider.scala
```diff
@@ -0,0 +1,230 @@
...
+import scala.meta.io.AbsolutePath
+import scala.meta.{inputs => m}
+
+import org.eclipse.jdt.core.JavaCore
+import org.eclipse.jdt.core.ToolFactory
+import org.eclipse.jdt.core.formatter.CodeFormatter
+import org.eclipse.jdt.internal.formatter.DefaultCodeFormatterOptions
+import org.eclipse.jface.text.BadLocationException
+import org.eclipse.jface.text.Document
+import org.eclipse.text.edits.MalformedTreeException
```
tgodzik, line 23: This seems not to work with JDK 8 :fearful: 
```
'/home/runner/work/metals/metals/tests/unit/target/e2e/implementation/on-usage' .
Exception in thread "pool-18-thread-19" java.lang.UnsupportedClassVersionError: org/eclipse/text/edits/MalformedTreeException has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

tgodzik: Probably connected to https://github.com/redhat-developer/vscode-java/issues/1543

Maybe we need to use an earlier version of jdt core?

Arthurm1: Damn - that didn't even occur to me.  It looks as though the other option: https://github.com/google/google-java-format also requires jdk11.
I guess we could use an older version, but then it wouldn't format new language features.
Or disable it if the JDK Metals is running under is <11?
Or use `ShellRunner#runJava` with different versions of the formatter depending on JDK being used?
I'll have a think - shame as it was an easy win - or so I thought.

tgodzik: Looks like we can use `3.19.0` version for now maybe? We can decide on switching from JDK 8 later

metals/src/main/scala/scala/meta/internal/metals/JavaFormattingProvider.scala
```diff
@@ -0,0 +1,230 @@
...
+    workspace: AbsolutePath,
+    buffers: Buffers,
+    userConfig: () => UserConfiguration,
+    buildTargets: BuildTargets
+)(implicit
+    ec: ExecutionContext
+) {
+  // TODO cache formatting file and reload on change?
+  // TODO progress monitor?
+  // TODO onTypeFormatting - wait until presentation compiler is integrated
```
tgodzik, line 36: Could we sum this up in the Java support issue instead of having it here?

metals/src/main/scala/scala/meta/internal/metals/JavaFormattingProvider.scala
```diff
@@ -0,0 +1,230 @@
...
+    Future {
+      val options = params.getOptions
+      val path = params.getTextDocument.getUri.toAbsolutePath
+      val input = path.toInputFromBuffers(buffers)
+      runFormat(path, input, options, fromLSP(input)).asJava
+    }
+  }
+
+  private def fromLSP(input: Input, range: l.Range): m.Position.Range =
+    m.Position.Range(
```
tgodzik, line 116: There is already a `toMeta` extension  method on `l.Range`

metals/src/main/scala/scala/meta/internal/metals/JavaFormattingProvider.scala
```diff
@@ -0,0 +1,230 @@
...
+
+      val version = for {
+        targetID <- buildTargets.sourceBuildTargets(path)
+        java <- buildTargets.javaTarget(targetID)
+        sourceVersion <- java.sourceVersion
+        targetVersion <- java.targetVersion
+      } yield ((sourceVersion, targetVersion))
+
+      version
+        .take(1)
```
tgodzik, line 165:
```suggestion
        .headOption
```

So that it doesn't break if the target is not available?

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -74,7 +74,17 @@ object MetalsEnrichments
   implicit class XtensionBuildTarget(buildTarget: b.BuildTarget) {
 
     def isSbtBuild: Boolean =
-      buildTarget.getDataKind == "sbt"
+      dataKind == "sbt"
+
+    def baseDirectory: String = {
+      val baseDir = buildTarget.getBaseDirectory()
+      if (baseDir != null) baseDir else ""
+    }
```
tgodzik, line 82:
```suggestion
    def baseDirectory: String = Option(buildTarget.getBaseDirectory()).getOrElse("")
```

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -74,7 +74,17 @@ object MetalsEnrichments
...
+
+    def dataKind: String = {
+      val dataking = buildTarget.getDataKind()
+      if (dataking != null) dataking else ""
+    }
```
tgodzik, line 87:
```suggestion
    def dataKind: String = Option(buildTarget.getDataKind()).getOrElse("")
```

metals/src/main/scala/scala/meta/internal/metals/newScalaFile/NewFileTypes.scala
```diff
@@ -5,50 +5,83 @@ import scala.meta.internal.metals.MetalsQuickPickItem
 object NewFileTypes {
   sealed trait NewFileType {
     val id: String
+    val syntax: String
```
tgodzik, line 8:
```suggestion
    val syntax: Option[String]
```
to maybe avoid having `"NA"`

metals/src/main/scala/scala/meta/internal/metals/Warnings.scala
```diff
@@ -33,6 +33,7 @@ final class Warnings(
     }
     val isReported: Option[Unit] = for {
       buildTarget <- buildTargets.inverseSources(path)
+      // TODO(arthurm1) add javaTarget warnings (targetroot, sourceroot, semanticDB etc.)
```
tgodzik, line 36: Let's add it to the issue or maybe add in this PR if possible?

tgodzik (changes requested):
Couple of files still left to review, getting to the finish :sweat_smile:

metals/src/main/scala/scala/meta/internal/metals/JavaFormattingProvider.scala
Arthurm1: I think 3.25 might work - I'll try and go back further if not...
https://github.com/eclipse-mirrors/org.eclipse.jdt.core/commit/32abdfb9c5100e8c2ff7669768b7d7fcc83c7e23#diff-b2b8692cb3b6e958a605b1cb599cc6b55c0f7c1ae57961b0a105a31b8a9f74ef

tgodzik (changes requested):
Finished! Great work! I think the semanticdb issues with JDK 17 is the only blocking one

But it seems that we might need to add additional options for it:
https://github.com/sourcegraph/lsif-java/commit/a86db1e231bae733f121eb400411f26966a6ed93

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -2377,11 +2398,17 @@ class MetalsLanguageServer(
...
+      "indexed workspace Scala SemanticDBs",
       clientConfig.initialConfig.statistics.isIndex
     ) {
       semanticDBIndexer.onScalacOptions(i.scalacOptions)
     }
+    timerProvider.timedThunk(
+      "indexed workspace Java SemanticDBs",
+      clientConfig.initialConfig.statistics.isIndex
+    ) {
+      semanticDBIndexer.onJavacOptions(i.javacOptions)
```
Arthurm1, line 2410: @tgodzik It's likely that projects with Scala and Java files will have the same TargetRoot set under `scalacOptions` and `javacOptions`.  So calling `semanticDBIndexed.onScalacOptions` and `semanticDBIndexed.onJavacOptions` will be doing the same thing so `SemanticDBIndexer#onChangeDirectory` will be called twice for each target.  Is that a performance issue?

olafurpg:
I'm sorry for the slow response, GitHub notifications don't distinguish between mentions or non-mentions once you have been mentioned once in a PR. Don't hesitate to ping me directly via email, Discord or Twitter DM about semanticdb-javac questions.

The latest version of lsif-java works with Java 17 but must use these JVM options
```
--add-exports jdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED
--add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
--add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED
--add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
--add-exports jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED
```

To compile sources in sbt with Java 17 you need to enable forking and update javacOptions to include the same exports with the `-J` prefix, like here https://github.com/sourcegraph/lsif-java/blob/a86db1e231bae733f121eb400411f26966a6ed93/build.sbt#L316-L340

Arthurm1:
@tgodzik I think I've addressed everything except some test failures.

You're right about the formatting comments config option - it's also set in the eclipse formatter file so I've removed it as a Metals option.

On `CreateNewSymbolLspSuite` failures - can I just add `com.sourcegraph.shaded.com.google.protobuf.DescriptorProtos.SourceCodeInfo` to the list of expected imports?  Java plugins have to be on the classpath so I'd expect the sourcegraph classes to be an option for imports

On the "add-exports" requirement for jdk 17: I can make a PR for Bloop.  I think this is going to depend on the java version in `bloop.data.Project#compileJdkConfig.javaHome`.  If it exists and then fallback to the JVM that Bloop runs under.  Do you know any way of easily getting the java version of a `javac` binary other than running `javac -version` and parsing the output (bit yucky)?  And do you know if there is the equivalent of `Properties.isJavaAtLeast` for comparing 2 java versions rather than a Java version with the running jvm version.  I can just copy that code from `Properties.isJavaAtLeast` but seems wasteful.

Arthurm1:
@olafurpg I don't suppose there is a plan in Sourcegraph to write build tool plugins to allow projects to create and deploy the equivalent of `projectname-semanticdb.jar` to maven in the same way as they'd deploy `*-sources.jar` and `*-javadoc.jar` but filled with semanticdb files?  Then Metals could download these and index the dependencies in a much better way than current indexing

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: That might be, so it would be cool to avoid that if it's the same directory.

tgodzik:
> On CreateNewSymbolLspSuite failures - can I just add com.sourcegraph.shaded.com.google.protobuf.DescriptorProtos.SourceCodeInfo to the list of expected imports? Java plugins have to be on the classpath so I'd expect the sourcegraph classes to be an option for imports

Maybe we could add `com/sourcegraph/shaded` (maybe some other also from the plugin, though `com/sourcegraph` is too general) to `defaultExclusions` in ExcludedPackagesHandler ? I wouldn't want that to show up for users since that would be different from what the build tool suggests.

> On the "add-exports" requirement for jdk 17: I can make a PR for Bloop.

That would be awesome!

>  Do you know any way of easily getting the java version of a javac binary other than running javac -version and parsing the output (bit yucky)?

There seems to be a properties file called `release` in `javaHome` (I checked openjdk, zulu and graalvm) that we could parse and maybe if it doesn't exist show a warning and default ot current version in Bloop?

> And do you know if there is the equivalent of Properties.isJavaAtLeast for comparing 2 java versions rather than a Java version with the running jvm version. I can just copy that code from Properties.isJavaAtLeast but seems wasteful.

Maybe we could use JavaVersion from `com.google.gson.internal`?

Arthurm1:
@tgodzik I think that covers everything.

The java format settings are now grouped into a single class and also grouped in the VS-code settings.

The semanticdb indexing of targetroots should now only be done once regardless of whether it's Java and/or Scala target.

I've added `com/sourcegraph/shaded/` to the exclusions list.  `com.sourcegraph.semanticdb_javac` is the only other package in that jar which I've left.  I figure if users raise issues then it could be excluded as well.

On the Java 17 `--add-exports` failures, Do we need to depend on a Bloop nightly now?  Or do we need a proper release of Bloop to depend against in `build.sbt`?  Currently it's `val bloop = "1.4.11"`

Also - I've left these as individual commits so it's easier for you to review - I assume you can squash them yourself if you feel it's necessary?

Arthurm1:
rebased

build.sbt
```diff
@@ -264,6 +264,7 @@ lazy val V = new {
   val bloop = "1.4.11-17-f4ef8273"
```
tgodzik, line 264: Let's bump bloop to the newest snapshot to see if it's working. I should release 1.4.12 version for sure also for the Test Explorer.

Arthurm1:
@tgodzik 

I'm unsure how to fix these test failures.  I'm getting the following message...
```
Javac is running in 'local' mode. These flags have been removed:
-J--add-exports, -Jjdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED, -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED, -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED, -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED, -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
```

Looks like this is coming from Bloop (zinc) and it is not forking java but instead re-using the running JVM.  The code seems to be the same in the latest zinc [LocalJava](https://github.com/sbt/zinc/blob/a83a002d42d1474d2a9306efb804f5f88ee0d436/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/LocalJava.scala#L302)

So is there a way to force forking in the tests?  Or would this still cause an issue in deployment generally.  I don't really understand what makes Bloop (or zinc) decide when to use the current JVM and when to fork - does it decide to use `LocalJava` purely based on whether it's compiling with the same jvm version that bloop is running under?

How does Bloop normally handle the need for `-J` when forking and not when not forking?  Surely this can't be the first time the need for jvmargs being passed has come up but I can't see anything obvious in the code.  I thought it might be possible for zinc to strip the `-J` when it came to `-J--add-exports` but I'd have thought that if it was that simple then it would already be in the code.

tgodzik:
@Arthurm1 I think I know what is going on, I will take a look into how we can fix it in Bloop. We could work around it here by specifying the Bloop jvm settings before running Bloop, which would make the JDK 17 tests pass.

olafurpg:
> I don't suppose there is a plan in Sourcegraph to write build tool plugins to allow projects to create and deploy the equivalent of projectname-semanticdb.jar to maven in the same way as they'd deploy *-sources.jar and *-javadoc.jar but filled with semanticdb files? Then Metals could download these and index the dependencies in a much better way than current indexing

@Arthurm1 No plans for that. I don't think we can expect the community to adopt anything like that so we'll always need to handle the case when SemanticDB files are missing anyways. On the bright side, compiling with semanticdb-javac is pretty fast compared to Scala compilation. If Metals wants to SemanticDB index all jars on the classpath then it could have its own global cache somehow.

olafurpg:
> com.sourcegraph.semanticdb_javac is the only other package in that jar which I've left. I figure if users raise issues then it could be excluded as well.

Please don't exclude this package since that would mean you can't get completions when working in the lsif-java repo :)

Arthurm1:
@tgodzik Should I wait for Bloop fix? Or look to change the jvm options?  I don't really want it to cause issues with users running Bloop under jdk 17 but I'm not sure if Bloop properly works under 17 yet anyway?

tgodzik:
We can add the export options here: https://github.com/scalameta/metals/blob/main/bin/test.sh#L10

Though I am not 100% sure how to detect that we are running JDK 17 :thinking:

tgodzik:
Actually we could just add `JAVA_VERSION: ${{ matrix.java }}` here:

https://github.com/scalameta/metals/blob/main/.github/workflows/ci.yml#L46

And then check for that in the bash script mentioned before.

tgodzik (approved):
LGTM! There are a view things we can improve on such as adding code lense and renames, but I would rather not have it as part of this PR.

Awesome work! I think we'll be aiming for a release early next year, so we should have time to test everything out.

tgodzik:
It seems that the eclipse is bringing in jars compiled with JDK 11 :/ We might need to go further back.

And there is an issue with sbt tests, I will take a look at those.

Arthurm1:
The problem seems to be that all transitive dependencies in eclipse jars are defined using ranges. e.g.
```
<dependency>
  <groupId>org.eclipse.platform</groupId>
  <artifactId>org.eclipse.text</artifactId>
  <version>[3.6.0,4.0.0)</version>
</dependency>
```

Version 3.9 was the last one compiled under jdk1.8 and then 3.10 was compiled under jdk11.  Version 4.0.0 doesn't even exist yet. But because of this range use, the valid jdk1.8 jar is ignored for a later version.  So I can't go back to an earlier eclipse version as the transitive dependency will still be jdk11.

As far as I can see - there's no way round this but having to specify the exact version of every transitive dependency that's declared using ranges.
e.g. `dependencyOverrides += "org.eclipse.platform" % "org.eclipse.text" % "3.9.0"`

Unless you know a better way?

I can see the same issue in the move from 11 -> 17 sometime in the future

lefou:
An explicit version is like an right open range starting at that version. You can try to specify an rage ending at the last version that still supports Java 8.

Arthurm1:
I'd still have to specify a version for every transitive dependency wouldn't I?

I've only got 1 dependency in the build: `"org.eclipse.jdt" % "org.eclipse.jdt.core" % "3.25.0"` (3.25.0 is the last jdk8 version), but its pom has...
```
<dependencies>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.core.resources</artifactId>
      <version>[3.14.0,4.0.0)</version>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.core.runtime</artifactId>
      <version>[3.13.0,4.0.0)</version>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.core.filesystem</artifactId>
      <version>[1.7.0,2.0.0)</version>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.text</artifactId>
      <version>[3.6.0,4.0.0)</version>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.team.core</artifactId>
      <version>[3.1.0,4.0.0)</version>
      <optional>true</optional>
    </dependency>
  </dependencies>
```

Sometime in those versions there's a switch from compiling using jdk8 to jdk11.
e.g. for `org.eclipse.text` the last jdk8 was version 3.9.0
for `org.eclipse.core.filesystem` the last jdk8 was version 1.7.500

So I'd have to work out what the last version compiled for jdk8 was for all of these jars, specify it and then look at all their dependencies and so on until no more ranges.  Or am I missing something here?  I've not used ranges before.

lefou:
As far a I know, Eclipse 4.16 was the last release which was built for Java 8. The probably easiest way to figure out which versions belong to it, is to download a Eclipse 4.16 release zip and look what jars are into it. But yes, I think there is no other easy way to select compatible versions, as this fact is not a property of the dependency artifacts.

lefou:
I once compiled a list of artifacts which belong to Eclipse 4.16 to use in some other closed projects.Don't know if it is of any help to you: https://gist.github.com/lefou/71b574af75c1b42653b1be8440561a8b#file-build_eclipse_bundles-sc

build.sbt
```diff
@@ -112,6 +112,23 @@ inThisBuild(
     resolvers += Resolver.sonatypeRepo("public"),
     resolvers += Resolver.sonatypeRepo("snapshot"),
     dependencyOverrides += V.guava,
+    dependencyOverrides += "org.eclipse.platform" % "org.eclipse.ant.core" % "3.5.500",
```
tgodzik, line 115: Uff, this is a lot, thanks for compiling the list!

Arthurm1: I've had more fun afternoons

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -160,10 +159,34 @@ class BuildServerConnection private (
     register(server => server.workspaceBuildTargets()).asScala
   }
 
+  def buildTargetJavacOptions(
+      params: JavacOptionsParams
+  ): Future[JavacOptionsResult] = {
+    val resultOnJavacOptionsUnsupported = new JavacOptionsResult(
+      List.empty[JavacOptionsItem].asJava
+    )
+    if (isSbt) Future.successful(resultOnJavacOptionsUnsupported)
```
tgodzik, line 168: @Arthurm1 I added this check since it seems that sbt shuts down when unexpected request is made.

tgodzik:
@Arthurm1 It's all green! feel free to merge, we can improve things further later on.

Arthurm1:
What's the standard - should I squash or ...?

build.sbt
ckipp01: If possible can we add a note here as to why we are doing this? I can see later on someone jumping in here and looking at this and wondering what in the world we are doing here with all these overrides.

tgodzik: I was thinking of moving it later to a separate file and commenting. This PR is pretty huge and I don't want to add anything more right now.

tgodzik: Also, I think you did another comment and I can't find it :sweat_smile:  which is a good indicator that this got way to big. What do you think? Is it ok to do same follow ups later?

ckipp01: :) of course, a follow-up is totally fine.

For the other comment 😆 I deleted it. A couple weeks ago I started reviewing this and only got like half way through and didn't realize I had draft comments saved, so when I added this comment it also posted those. They aren't relevant anymore, so I just removed them. Sorry for the noise!

ckipp01 (approved):
Incredible work @Arthurm1 🚀

kpodsiad:
Wow, this is magnificent @Arthurm1!

Issue #2493, opened 2021-02-10
Go to definition not working for Scala 3 source


tgodzik:
**Describe the bug**

Currently, it's impossible to go to definition inside Scala 3 sources.

**To Reproduce**

1. Open `mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala`
2. Try to go to definition on any compiler classes.

**Expected behavior**

Metals goes to the exact sources location.

**Additional context**

The main issue here is that we are still using the 2.13 dialect for parsing library sources. We should look for all the palces where it's used and make sure that a proper dialect is used based on the build target the `go to definition` call originated.

**Search terms**

goto definition dependencies libraries

tgodzik:
That's actually already fixed.

mpilquist:
I'm running in to this same issue with Metals 1.10.5 and Scala 3.0.0 final. Go to definition works for Java dependencies but not any Scala 3 dependencies. ScalaDoc previews work fine though. Is this a known limitation/issue?

tgodzik:
> I'm running in to this same issue with Metals 1.10.5 and Scala 3.0.0 final. Go to definition works for Java dependencies but not any Scala 3 dependencies. ScalaDoc previews work fine though. Is this a known limitation/issue?

I think that might be a regression. I will take a look at it.

tgodzik:
Ok, I seemed to have the same issue, but only because the file was not compiled. After reimporting the build and compiling I started to get correct go to definition for the compiler sources.

@mpilquist do you have a specific instance/library that this doesn't work for?

mpilquist:
I'll experiment a bit and report back. I ran in to this on a closed source project but I'll try to reproduce on OSS. I definitely had the project imported and compiled and there are no concerning messages in Metals Output tab.

tgodzik:
It might be possible that we inferred a wrong dialect for some sources. I think maybe we should add more debug diags in those cases to turn on when needed.

mpilquist:
Everything is working now, both on closed source project and open source -- I think the root cause was me running `sbt bloopInstall` manually and without `-Dbloop.export-jar-classifiers=sources`. I had only done that as a workaround for import build failing. Turns out I had a bad SBT script path in my Metals setting.

BTW, I did have to run `rm -rf {.bsp,.bloop}` on scodec, followed by a fresh import build.

tgodzik:
[...]

PR #2527, opened 2021-02-19
Fallback to Scala3 dialects when parsing library sources

tgodzik:
To enable finding symbols in library sources we use the parser to find the exact symbol that we need. Previously it would not work for Scala 3 due to using the Scala213 dialect. Now, we fallback to Scala3 dialect in case of issues, which is a temporary workaround until https://github.com/scalameta/metals/issues/2493 is fixed.

I tried to do it here, but it turns out it's a much more compelx issue and would not want to merge it before the release.


ckipp01 (approved):
Makes sense. LGTM!

Issue #2392, opened 2021-01-15
Update ammonite for 2.13.5


tgodzik:
Currently we test Ammonite support on 2.12.12, but as soon as it's available for 2.12.13, we should switch.

Once we do it we should:
- move Ammonite212 suite back to unit tests and set V.scala212 version
- remove `2.12.12` from `publishAllMtags` invocation in `build.sbt`

alexarchambault:
Ammonite for `2.12.13` has just been [published](https://repo1.maven.org/maven2/com/lihaoyi/ammonite-interp_2.12.13/2.3.8-32-64308dc3) (it should have been published earlier, something went wrong in its release jobs…).

tgodzik:
Awesome, thanks! I will update the PR then.

tgodzik:
2.12.13 is now supported, but we now are missing 2.13.5, which is not yet released.

PR #2530, opened 2021-02-19
Update Ammonite to support Scala 2.12.13

tgodzik:
Fixes https://github.com/scalameta/metals/issues/2392

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -179,9 +179,8 @@ lazy val V = new {
   val scala210 = "2.10.7"
   val scala211 = "2.11.12"
   val sbtScala = "2.12.10"
-  // TODO https://github.com/scalameta/metals/issues/2392
-  val ammonite212Version = "2.12.12"
   val scala212 = "2.12.13"
+  val ammonite212Version = scala212
   val scala213 = "2.13.4"
   val scalameta = "4.4.9"
   val semanticdb = scalameta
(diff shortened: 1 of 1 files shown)


ckipp01 (approved):
~Glad to have this fixed. LGTM!~

ckipp01 (changes requested):
Actually sorry looking closer I totally missed that the Ammonite LSP tests were failing. We'll need to figure out what's going on with these releases...

tgodzik:
> Actually sorry looking closer I totally missed that the Ammonite LSP tests were failing. We'll need to figure out what's going on with these releases...

Yeah, I noticed that something seems off. I will investigate next week.

alexarchambault:
I think the issue lies in ammonite-runner. It needs to pull amm-compiler along the other Ammonite dependencies it fetches, because of https://github.com/lihaoyi/Ammonite/pull/1144.

tgodzik:
> I think the issue lies in ammonite-runner. It needs to pull amm-compiler along the other Ammonite dependencies it fetches, because of [com-lihaoyi/Ammonite#1144](https://github.com/com-lihaoyi/Ammonite/pull/1144).

@alexarchambault do you need any help with that?

alexarchambault:
That should be fine, thanks! ammonite-runner `0.3.1` should be on its way to Maven Central. I didn't try, but it ought to fix the errors we're seeing here.

PR #2531, opened 2021-02-20
Fix standalone Scala 3 worksheets

tgodzik:
In one of the previous PRs we changed the parser to take into account the real dialect of the code to parse based on the build target the file is located in. Recently, we also added the possibility to change the default Scala version, which allowed users to use standalone Scala 3 worksheets. We haven't however changed the parser dialect that is used for those worksheets, which results in Scala 3 specific code to be marked as an error. In this PR I am making the parser use the correct parser dialect.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -293,6 +293,7 @@ def multiScalaDirectories(root: File, scalaVersion: String) = {
   if (isScala3(partialVersion)) {
     result += base / "scala-3"
   }
+  result += base / s"scala-$scalaVersion"
   result.toList
 }
 
--- a/tests/slow/src/test/scala/tests/feature/WorksheetCrossLspSuite.scala
+++ b/tests/slow/src/test/scala/tests/feature/WorksheetCrossLspSuite.scala
@@ -3,7 +3,10 @@ package tests.feature
 import scala.meta.internal.metals.{BuildInfo => V}
 
 class Worksheet211LspSuite extends tests.BaseWorksheetLspSuite(V.scala211)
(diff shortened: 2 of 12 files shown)


build.sbt
```diff
@@ -293,6 +293,7 @@ def multiScalaDirectories(root: File, scalaVersion: String) = {
   if (isScala3(partialVersion)) {
     result += base / "scala-3"
   }
+  result += base / s"scala-$scalaVersion"
```
tgodzik, line 296: There were issues in the build because mtags3 were lacking the version specific directory.

tests/slow/src/test/scala/tests/feature/WorksheetCrossLspSuite.scala
```diff
@@ -3,7 +3,10 @@ package tests.feature
 import scala.meta.internal.metals.{BuildInfo => V}
 
 class Worksheet211LspSuite extends tests.BaseWorksheetLspSuite(V.scala211)
-class Worksheet3LspSuite extends tests.BaseWorksheetLspSuite(V.scala3)
+class Worksheet3LspSuite extends tests.BaseWorksheetLspSuite(V.scala3) {
+  override def versionSpecificCodeToValidate: String =
```
tgodzik, line 7: Worksheets would show errors on a version specific code.

build.sbt
ckipp01: Ironic I just hit on this when I cleaned and tried to build everything locally. It was on my list to look at after reviewing this, ha.

ckipp01 (approved):
Nice job catching this before the release. LGTM!

PR #2532, opened 2021-02-20
Smoother main and test discovery

### LINKED ISSUE #2491: Smoother main discovery when no `mainClass` is given.
**Feature request and way to do it**

Currently there is some manual work required to run a main class even if there is only one in your application, and your client doesn't have code lenses. We also handle the situation where we give a `mainClass`, but not the `buildTarget`. I'd like to be able to just give a `buildTarget` for example and have it also work. Even more so, I'd like to just give it nothing, and if there is only one main, it runs it for example, or if there are multiple then I'd like the server to show me and let me choose.

We can handle both of these already. I've been poking around and we already pretty much have the logic to do all this, but it all needs to be glued together. I've started to work on this already since I'd like to have a smoother experience for clients that may not have code lenses or don't want to use code lenses, but still don't want to manually have to enter the values. I also need this in `nvim-metals` for integration with `nvim-dap` 😬 .

The same concept should also be able to be applied to testing.

_Sorry for breaking our own rules and creating a feature request here, but I'm already working on it..._

**Search terms**:
running and debugging without code lens

ckipp01:
## What this is
This pr introduces some new functionality around run/test/debug. My main goal with this is to enable clients to quickly be able to run or test based off of where they are, and that's it. This is especially useful for users without code lenses or useful in the case where a user doesn't want to go back to the file with the main in to hit run, or scroll all the way back up to the top of a file to hit `test`, or even if they want an easy way to test an entire build target.

## How does a client use this
This is accomplished by introducing a new params object for the `debug-adapter-start` where all a client has to send in is the current file it's in, and the `runType` which can be `run`, `testFile`, or `testTarget`. Env, args, and options are all still possible, but not necessary. Ideally a client will have these three different run types mapped to a command. You can see a very rough example of this in [nvim-metals](https://github.com/scalameta/nvim-metals/pull/103/files#diff-443712fdd0616dd1a70fa546aa9e4105b39f837eb426425a7c74706adecd8930) since that's how I was testing this.

## Some examples

In the case where there is just a single main method (in the build target the file belongs to), it will just be ran. You don't even have to be in that file, just the build target. If there are no mains in your build target, you'll be warned.

In the case of multiple mains in a single build target, you'll be prompted to choose which you'd like to run.

The `testFile` option will check to see if you're currently in a test suite. If so, it runs it, if not you will get a warning about no tests being found in your file. If you use the `testTarget` it will run all the suites in that target. If there are no tests in that target, you'll get a warning.

## What's left

In this pr I'd still like to get a couple things done, but before I do I want to make sure you're on board with this direction. After that I need to make sure

- [x] Debug docs are updated

Closes #2491


tests/unit/src/test/scala/tests/DebugDiscoverySuite.scala
```diff
@@ -0,0 +1,241 @@
...
+          "testFile"
+        ).toJson
+      )
+      _ <- debugger.initialize
+      _ <- debugger.launch
+      _ <- debugger.configurationDone
+      _ <- debugger.shutdown
+      output <- debugger.allOutput
+    } yield assert(output.contains("All tests in a.Foo passed"))
+  }
```
ckipp01, line 174: So I'm curious if you have any input here, I'm having a really hard time figuring out how to get this test to pass. Locally with this same exact layout everything works as expected, but when testing here the testClasses never actually get cached, so when I look for them, it just returns an empty list and therefore no tests are found. I had the same issue with testing `run`, but was able to get around it by just including the `didOpen` on the file, which seemed to ensure all the build target classes were fetched and main classes stored. However, that's not working for tests here and I'm unsure how to do this. Any ideas?

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -480,6 +499,13 @@ object Messages {
         s"Class '$cls' not found."
       )
     }
+
+    def show(msg: String): MessageParams = {
```
tgodzik, line 503:
```suggestion
    def errorMessageParams(msg: String): MessageParams = {
```
?

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -434,3 +434,12 @@ case class DebugUnresolvedAttachRemoteParams(
     port: Int,
     buildTarget: String
 )
+
+case class DebugFullyUnresolvedParams(
+    path: String,
```
tgodzik, line 439: Why not jsut add nullable/optional fields to DebugUnresolvedMainClassParams  and DebugUnresolvedTestClassParams ?

metals/src/main/scala/scala/meta/internal/metals/config/RunType.scala
```diff
@@ -0,0 +1,20 @@
...
+object RunType {
+  sealed trait RunType
+  case object Run extends RunType
+  case object TestFile extends RunType
+  case object TestTarget extends RunType
+
+  def fromString(string: String): Option[RunType] = {
+    string match {
+      case "run" => Some(Run)
+      case "testFile" => Some(TestFile)
```
tgodzik, line 12: we can have just test and if a target is specified and not file it would mean it's supposed to test all test in target?

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -130,6 +140,180 @@ class DebugProvider(
     }
   }
 
+  /**
+   * Given a BuildTargetIdentifier either get the displayName of that build
+   * target or defaul to the full URI to display to the user.
```
tgodzik, line 145:
```suggestion
   * target or default to the full URI to display to the user.
```

metals/src/main/scala/scala/meta/internal/metals/config/RunType.scala
ckipp01: But then where would the target come from? A user would have to include it? This way, there is no need need, a user can either bind a `testFile` or a `testTarget` to whatever they want, without any need to have to provide a target.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
ckipp01: Yea true I'll try doing that.

metals/src/main/scala/scala/meta/internal/metals/config/RunType.scala
tgodzik: Hmm.. that is a tough one. Your solution might be best here if we want to provide the functionality to run all tests in the current target.

One improvement could be that if a user doesn't specify file and type is `testFile` or `testTarget` we could use focusedDocument (or focusedTarget which I think exists.)

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
ckipp01: Ah yea, this was left over from me wanting a "generic" pass-through way to just show an error message where the full message was provided. I've gone ahead and moved this up a level and converted it to a generic `showError` method.

metals/src/main/scala/scala/meta/internal/metals/config/RunType.scala
ckipp01: Yea it's a good point. I'm just not convinced that on the client side if we are using `testFile` or `testTarget`, there shouldn't really be a reason why we can't send along uri since it should be quite easy to. Do you think this might be an issue from VS Code if someone is trying to trigger from like the side panel or something? I don't think it's unreasonable to just expect a uri.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
ckipp01: Alright, so I've been playing around with this a bit, and I'm getting pretty convinced that just adding some more `@Nullable` fields here isn't the answer. It adds way more complexity in trying to handle all of the various cases, and I'm finding it way easier to manage to just have another case class for it. Just to give a small example:

### Current
```
case class DebugUnresolvedTestClassParams(
    testClass: String,
    @Nullable buildTarget: String = null
)
```

### Possible new
```
case class DebugUnresolvedTestClassParams(
    @Nullable testClass: String,
	@Nullable path: String,
    @Nullable testType: String,
    @Nullable buildTarget: String = null
)
```
Apart from not loving having every field nullable, you get into all sorts of interesting things that we have to handle server side. What if `testType` is set to `testTarget` and you are provided with a `testClass` that doesn't fit that target. What if for type you pass in `testTarget`, but then you pass in a `path` in one target and the `buildTarget` for another target. There are just so many possiblities that you end up making really strict rules on how they are all handled, but then you lose all that info since it's all contained in a class with all nullable fields. I think having a new class for the discover `DebugFullyUnresolvedParams` for this just makes it way cleaner on the server side. I'm open to maybe trying to figure this out more, but I'm not seeing the benefit of having to change the current functionality to account for the new nullables, and also account for all the crazy cases with the class having every field nullable vs just having a separate class for it. wdyt?

tgodzik: It it all becomes much more complex because of the change then sure, I agree it might be beneficial to have a separate class. I just thought it might work better for the existing client implementations, but I haven't dug much into it.

metals/src/main/scala/scala/meta/internal/metals/config/RunType.scala
tgodzik: We should be able to always send a URI - especially if we start from a command. The only exception would be if we wanted to have a `launch.json` configuration that would just run anything in the current file. This would allow to just use the existing shortcut for debug (F5), but on the other hand we can always add a separate command for that and `launch.json` might be a hack. So maybe leave it out for now and I will see how best to implement it on the vs code side.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -18,6 +19,13 @@ import org.eclipse.lsp4j.ShowMessageRequestParams
  */
 object Messages {
 
+  def showError(msg: String): MessageParams = {
```
tgodzik, line 22:
```suggestion
  def errorMessageParams(msg: String): MessageParams = {
```
? Since the method does not show anything and the name would indicate a side effect happening.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -130,6 +140,180 @@ class DebugProvider(
...
+    }
+  }
+
+  /**
+   * Given fully unresolved params this figures out the runType that was passed
+   * in and then discovers either the main methods for the build target the
+   * path belongs to or finds the tests for the current file or build target
+   */
+  def debugDiscovery(
+      params: DebugFullyUnresolvedParams
```
tgodzik, line 250:
```suggestion
      params: DebugDiscoveryParams
```
maybe?

tests/unit/src/test/scala/tests/DebugDiscoverySuite.scala
tgodzik: I think we need to have the files fully compiled, so maybe add `didSave` after did open? I think it should actually wait for compilation to finish.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
ckipp01: ahhhh I get what you were saying before. Sorry I misunderstood at first. I'll change this.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
ckipp01: Yea I like that.

tests/unit/src/test/scala/tests/DebugDiscoverySuite.scala
ckipp01: Yea even with the `didSave` it still doesn't fine them 🤔

tgodzik: I started looking into how we made it work in other tests and it seems we didn't! We don't actually have any tests for testing. There is also an issue I encountered in Bloop which would make the tests endpoint timeout pretty often and is the number one flaky test currently. I think it could all be connected :/

ckipp01: Hmm, bummer. I'll play around a bit more with them to see if I can figure out what's going on.

tgodzik: We can skip testing it currently or await some time for the tests classes information to show up. I will try to investigate the Bloop side.

ckipp01:
Alright, I think this should now be ready to review. I ended up keeping the separate params mainly because I think it is way cleaner than accounting for everything being nullable. I tried that and it got messy pretty quickly. Plus, this allows for clients to start to use this without worrying at all about any existing implementation or things breaking as the others work fully the same.

I figured out the tests, and it was just a timing thing, and I need to wait a bit to ensure that the classes are cached and then the tests work. _However_, for windows I had major issues getting the URI right. Finding a clean way to always send in the correct URI proved to be way harder than anticipated and I kept getting a mixture of `file:///D:\blah\blah/blah/blah.scala` with mixed `/`. I couldn't figure out a clean way to do that, so if I am missing an obvious way, please let me know. For now I just marked them as `.flakyWindows`. Not ideal, but I was sick of wrestling with it.

I think this is a good foundation to start actually using the discovery and then fine tuning. Let me know what you think.

tests/unit/src/test/scala/tests/DebugDiscoverySuite.scala
```diff
@@ -0,0 +1,253 @@
...
+import scala.meta.internal.metals.debug.WorkspaceErrorsException
+
+// note(@tgodzik) all test have `System.exit(0)` added to avoid occasional issue due to:
+// https://stackoverflow.com/questions/2225737/error-jdwp-unable-to-get-jni-1-2-environment
+class DebugDiscoverySuite extends BaseDapSuite("debug-discovery") {
+  private val mainPath = "a/src/main/scala/a/Main.scala"
+  private val fooPath = "a/src/main/scala/a/Foo.scala"
+  private val barPath = "a/src/main/scala/a/Bar.scala"
+
+  test("run".tag(FlakyWindows)) {
```
tgodzik, line 18: What is the issue on windows? Most other tests work fine there and we haven't had the need to ignore much on Windows. It seems kind of harsh to make the entire suite flaky. We will never seen any issues pop up there.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -473,6 +473,8 @@ final class TestingServer(
     server.executeCommand(new ExecuteCommandParams(command, args)).asScala
   }
 
+  def waitFor(sec: Long): Future[Unit] = Future { Thread.sleep(sec) }
```
tgodzik, line 476: Could we maybe wait on the indexingPromise in MetalsLanguageServer? We can make it public and it should work the same.

tgodzik (changes requested):
Thanks for digging into the failures! That was a lot of debugging from what I've seen in the emails :fearful:

docs/integrations/debug-adapter-protocol.md
```diff
@@ -88,10 +88,30 @@ a.b.key8=value 8   # will be ignored
 a-b-key9=value 9   # will be ignored
 ```
 
+- For Metals discovery
+
+This options works a bit different than the other two param shapes as you don't
+specific a test or main class, but rather a `runType` of either `"run"`,
+`"testFile"`, or `"testFile"` and a file URI representing your current location.
```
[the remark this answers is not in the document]
ckipp01: Ach, yea thanks.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -130,6 +140,180 @@ class DebugProvider(
...
+          NoTestsFoundException("build target", displayName(target))
+        )
+      case (Some(TestFile), Some(target)) =>
+        textDocumentO
+          .fold[Future[Seq[BuildTargetClasses.ClassName]]] {
+            Future.failed(new Exception("Semanticdb not found for path."))
+          } { textDocument =>
+            Future {
+              for {
+                occurence <- textDocument.occurrences
```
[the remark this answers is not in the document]
ckipp01: Ah ok cool that works perfectly then.

tests/unit/src/main/scala/tests/TestingServer.scala
ckipp01: 🤔 so I tried adding in 
```
      _ <- server.server.indexingPromise.future
```
Which I assumed would wait for it, but it seems to not work. This is what you had in mind right?

tests/unit/src/test/scala/tests/DebugDiscoverySuite.scala
ckipp01: So there isn't any other test suites that actually "pretend" to send in a URI the same way that a client would send it in. The trick part here is that the `${workspace}` is always nice and done in the OS specific way. However, the path that we put together, like in this example, `mainPath`, is done in unix fashion with `a/src/main/scala/a/Main.scala`. Then the testing server takes that apart and makes sure that it is created correctly for the OS. However, how do you then get that exact created path back to pretend to pass it into the server for it to match that it's indeed there. Everything I tried resulted in a screwed up `file:///D:\a\metals\metals\tests\unit\target\e2e\debug-discovery\run/a/src/main/scala/a/Main.scala` like syntax. I even tried a gross `replaceAll` on the `mainPath` variable when `isWindows` was detected. At that point it was getting pretty hacky and I just ignored windows. Any idea how to accurately get this file URI in the OS specific way in the test to match exactly how a client would send it? I just pushed this up with the `flakyWindows` removed from the `run` test just to show you.

tests/unit/src/main/scala/tests/TestingServer.scala
tgodzik: It should wait until the workspace is indexed, so maybe we should also wait on `didSave` (it will wait on the compilation)?  The test classes are still not available by then?

tests/unit/src/test/scala/tests/DebugDiscoverySuite.scala
```diff
@@ -0,0 +1,253 @@
...
+           |    System.exit(0)
+           |  }
+           |}
+           |""".stripMargin
+      )
+      _ <- server.didOpen(mainPath)
+      _ <- server.waitFor(TimeUnit.SECONDS.toMillis(10))
+      debugger <- server.startDebuggingUnresolved(
+        new DebugDiscoveryParams(
+          s"file:///${server.toPath(mainPath)}",
```
tgodzik, line 39: Why not do `server.toPath(mainPath).toURI.toString` ?

ckipp01: ugh... out of all the things I tried... how did I not try this. This worked. Thank you 🙏🏼

tests/unit/src/main/scala/tests/TestingServer.scala
ckipp01: Yea but there already is a did save in the others that I tried this on. For example in `testTarget` we have the `didOpen`, `didSave`, and then the `indexingPromise.future`, and that still doesn't have the test classes. I have no idea why that takes more time.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -434,3 +434,12 @@ case class DebugUnresolvedAttachRemoteParams(
     port: Int,
     buildTarget: String
 )
+
+case class DebugDiscoveryParams(
```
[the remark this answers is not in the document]
ckipp01: Yea for sure, I think once we can get this across the finish line, use it a bit, then we can start to fine tune what makes the most sense, and add more on top of it. That's my hope anyway.

tests/unit/src/main/scala/tests/TestingServer.scala
tgodzik: Damn, it might have something to do with Bloop. Alternative would be to use SBT bsp here, but it's fine to leave it as is.

ckipp01:
Woo, all ✅ on the window test!

tests/unit/src/main/scala/tests/TestingServer.scala
ckipp01: 🤔 I'm thinking for now to leave it as I played around with this a bit more and still can't get it to work as expected.

tgodzik (approved):
LGTM!

PR #2536, opened 2021-02-21
Turn Dotty import suggestions into code actions

gabro:
A first attempt at https://twitter.com/tomekgodzik/status/1363447488653451265?s=21

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
@@ -30,4 +30,22 @@ object ScalacDiagnostic {
         case _ => None
       }
   }
+
+  object SymbolNotFoundWithImportSuggestions {
+    // https://github.com/lampepfl/dotty/blob/3b741d67f8631487aa553c52e03ac21157e68563/compiler/src/dotty/tools/dotc/typer/ImportSuggestions.scala#L311-L344
+    private val regexes = for {
+      fix <- List("One of the following imports", "The following import")
+      help <- List("fix ", "make progress towards fixing")
+    } yield s"""(?s)$fix imports might $help the problem:(.*)""".r
+    def unapply(d: l.Diagnostic): Option[List[String]] =
+      regexes.collectFirst {
+        case regex if regex.findFirstIn(d.getMessage()).isDefined =>
+          val suggestedImports = regex
+            .findFirstMatchIn(d.getMessage())
+            .map(
+              _.group(1).linesIterator.filterNot(_.isEmpty()).map(_.trim).toList
+            )
+          suggestedImports.getOrElse(Nil)
+      }
+  }
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingGiven.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingGiven.scala
@@ -1,0 +1,86 @@
+package scala.meta.internal.metals.codeactions
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals._
+import scala.meta.pc.CancelToken
+
+import org.eclipse.{lsp4j => l}
+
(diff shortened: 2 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingGiven.scala
```diff
@@ -0,0 +1,86 @@
...
+        diagnostic: l.Diagnostic,
+        packageName: String,
+        name: String
+    ): Future[Seq[l.CodeAction]] = {
+      val textDocumentPositionParams = new l.TextDocumentPositionParams(
+        params.getTextDocument(),
+        diagnostic.getRange.getEnd()
+      )
+      compilers
+        .autoImports(textDocumentPositionParams, name, token)
```
tgodzik, line 37: `autoImports` aren't yet implemented for Scala 3, so we would have to make do without it.

gabro: ah, I was just thinking about that. The only difficult thing would be to determine import position

metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
```diff
@@ -30,4 +30,22 @@ object ScalacDiagnostic {
         case _ => None
       }
   }
+
+  object SymbolNotFoundWithImportSuggestions {
```
tgodzik, line 34: I also wonder if we could get this from the compiler itself and maybe use completions?

metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
```diff
@@ -30,4 +30,22 @@ object ScalacDiagnostic {
...
+    // https://github.com/lampepfl/dotty/blob/3b741d67f8631487aa553c52e03ac21157e68563/compiler/src/dotty/tools/dotc/typer/ImportSuggestions.scala#L311-L344
+    private val regexes = for {
+      fix <- List("One of the following imports", "The following import")
+      help <- List("fix ", "make progress towards fixing")
+    } yield s"""(?s)$fix imports might $help the problem:(.*)""".r
```
tanishiking, line 39:
```suggestion
    } yield s"""(?s)$fix might $help the problem:(.*)""".r
```

?

smarter, line 39: The exact format of the compiler message strings is not stable and may change at any time, please do not rely on it. Instead, I encoruage you to contribute upstream a way to expose this information semantically in our Reporting API (we also implement scalafix-like rewrites which could be displayed as quick fix actions in vscode).

smarter (changes requested):

gabro:
Closing this for a few reasons:

- too much infrastructure missing for auto-imports with Scala 3
- the regex approach was a quick and dirty attempt to get it done, but I agree it's unreliable and it would be better to contribute an API in the dotty's presentation compiler.
- that said, I won't realistically have time to put down the amount of work required to implement this feature properly

Thanks everyone for the feedback

tgodzik:
Should we turn it into a feature request though? We might be able to get that data from the compiler anyway, so it might still be doable, but will need a greater amount of time.

gabro:
https://github.com/scalameta/metals-feature-requests/issues/192

metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
ckipp01: Revisiting long after the fact. @smarter has any thought gone into what this would look like, exposing this information semantically? I've been thinking about this often lately and sort of year for an error index like [rust has](https://doc.rust-lang.org/error-index.html) and more structured errors that we can use for stuff like this. I see some convos that touch on this a bit like [this old convo](https://contributors.scala-lang.org/t/towards-better-error-messages-in-scalac/1470/32), but has there been any design though or conversation around what this could look like in Dotty?

smarter: > has there been any design though or conversation around what this could look like in Dotty?

No, this is still waiting for a motivated individual to step up :).

PR #2538, opened 2021-02-22
[Scala 3] Add an ability to generate on demand semanticdb

tgodzik:
Generating on demand semanticdb is used for features such as references or renames in case of:
- standalone scala files
- dependency sources

Previously, it was not possible for Scala 3. Now, it turned out not to be that hard to reuse the existing Dotty implementation to provide the fucntionality.


metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
```diff
@@ -176,7 +176,11 @@ final class InteractiveSemanticdbs(
         clientConfig.initialConfig.compilers.timeoutDelay,
         clientConfig.initialConfig.compilers.timeoutUnit
       )
-    val textDocument = s.TextDocument.parseFrom(bytes)
+    val textDocument = {
+      val doc = s.TextDocument.parseFrom(bytes)
+      if (doc.text.isEmpty()) doc.withText(text)
```
dos65, line 181: Is this check required?
It looks like `SemanticdbTextDocumentProvider` sets up `text` value correctly.

dos65 (approved):

metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
tgodzik: It's actually not serialized to bytes, which I think might be the default way to write it to file.

PR #2539, opened 2021-02-22
Fix rename/references for worksheets

tgodzik:
Thanks to the recent work on generating semanticdb for sbt, it's also possible to generate semanticdb for worksheets and in this way allow:
- references
- rename
- document highlight

The only thing missing is `Go to implementations` which doesn't work for anything that is not indexed globally. We don't want to index worksheets globally as their symbols are local to the file they exist in.

Fixes https://github.com/scalameta/metals/issues/2540

### DIFF
--- a/docs/editors/overview.md
+++ b/docs/editors/overview.md
@@ -464,8 +464,8 @@ scripts.
   </tr>
   <tr>
     <td>Find references</td>
-    <td align="center"></td>
-    <td align="center"></td>
+    <td align="center">✅</td>
+    <td align="center">✅</td>
     <td align="center">✅</td>
     <td align="center">✅</td>
   </tr>
@@ -478,15 +478,15 @@ scripts.
   </tr>
   <tr>
     <td>Find implementations</td>
-    <td align="center"></td>
+    <td align="center">✅</td>
     <td align="center"></td>
     <td align="center">✅</td>
     <td align="center">✅</td>
   </tr>
   <tr>
     <td>Rename symbol</td>
-    <td align="center"></td>
-    <td align="center"></td>
+    <td align="center">✅</td>
+    <td align="center">✅</td>
     <td align="center">✅</td>
     <td align="center">✅</td>
   </tr>
@@ -506,8 +506,8 @@ scripts.
   </tr>
   <tr>
     <td>Workspace symbols</td>
-    <td align="center"></td>
-    <td align="center"></td>
+    <td align="center">✅</td>
+    <td align="center">All symbols are local</td>
     <td align="center">✅</td>
     <td align="center">✅</td>
   </tr>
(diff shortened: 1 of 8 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
```diff
@@ -12,12 +12,26 @@ import scala.meta.io.AbsolutePath
 class SemanticdbTextDocumentProvider(val compiler: MetalsGlobal) {
   import compiler._
 
+  private val magicImportsRegex = """import\s+($ivy|$repo|$dep|$scalac)\..*""".r
```
dos65, line 15: It needs to escape `$` characters.
```suggestion
  private val magicImportsRegex = """import\s+(\$ivy|\$repo|\$dep|\$scalac)\..*""".r
```

dos65 (changes requested):

docs/editors/overview.md
```diff
@@ -478,15 +478,15 @@ scripts.
   </tr>
   <tr>
     <td>Find implementations</td>
-    <td align="center"></td>
+    <td align="center">✅</td>
     <td align="center"></td>
```
tgodzik, line 482: `implementations` are not yet supported - added an issue here: https://github.com/scalameta/metals-feature-requests/issues/190

metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
```diff
@@ -78,8 +79,8 @@ final class InteractiveSemanticdbs(
           if (existingDoc == null || existingDoc.md5 != sha) {
             val compiled = compile(path, text)
             Option(compiled).foreach(doc =>
-              // don't index dependency source files
-              if (!source.isDependencySource(workspace))
+              // don't index dependency source files or worksheets, since their definitions are local
+              if (!source.isDependencySource(workspace) && !source.isWorksheet)
```
tgodzik, line 83: We don't want to index worksheets globally at least until we can distinguish different build targets and unrelated symbols with the same name. Also, worksheet definitions are only available inside the worksheet itself.

mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
tgodzik: Fixed, also move it to a separate file to also use in the Scala 3 worksheets.

tgodzik: And also added a test.

dos65 (approved):
LGTM!

PR #2544, opened 2021-02-23
Add support for Scala 2.13.5

tgodzik:
Issues encountered during the update:
- Scala presentation compiler will no longer complete type symbols on `a.b.Name` in positions where type is not allowed, which is actually a bugix, but we needed to change some test
- Ammonite does not have support for 2.13.5 yet and even if it has we can't currently update. In order to make the experience smoother for the users I added the logic in VersionSelector that will fallback to the last Ammonite supported Scala version instead of just failing
- there was a fix in scalameta for indented regions that changed their ends, which needed to be updated in the expect tests
- one test started to fail with 2.13.5 - is unclear if the behaviour was ever fixed in Metals or whether we relied on the compiler #2546

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -181,9 +181,10 @@ lazy val V = new {
   val sbtScala = "2.12.10"
   // TODO https://github.com/scalameta/metals/issues/2392
   val ammonite212Version = "2.12.12"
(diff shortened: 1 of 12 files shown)


ckipp01 (approved):
Looks good! LGTM!

PR #2552, opened 2021-02-25
Ensure token edit distance is always calculated with correct dialect.

### LINKED ISSUE #2550: Compilation error on Scala 3 macro not reported
**Describe the bug**

1. Create a Scala 3 project with sbt
2. Open it with Metals using sbt BSP server
3. Create a macro definition with an error in it:

EDITED:
```scala
import scala.quoted._

inline def printCode(inline expr: Any): Unit = ${printCodeImpl('expr)}

def printCodeImpl(expr: Expr[Any])(using ctx: Quotes): Expr[Unit] =
  import quotes.reflect.Position
  '{ println("") }

def foo: String = xxx
```

**Expected behavior**

The error is correctly sent by sbt to Metals but it is not reported by Metals on the source file.

```
// bsp.trace.json

[Trace - 09:45:23 PM] Received notification 'build/publishDiagnostics'
Params: {
  "textDocument": {
    "uri": "file:///home/piquerez/demo/src/test/scala/demo/Foo.scala"
  },
  "buildTarget": {
    "uri": "file:/home/piquerez/demo/#demo/Test"
  },
  "diagnostics": [
    {
      "range": {
        "start": {
          "line": 7,
          "character": 2
        },
        "end": {
          "line": 7,
          "character": 3
        }
      },
      "severity": 1,
      "source": "sbt",
      "message": "Not found: xxx"
    }
  ],
    "reset": true
}


[Trace - 09:45:23 PM] Received notification 'build/taskFinish'
Params: {
  "taskId": {
    "id": "79",
    "parents": []
  },
  "eventTime": 1614199523154,
  "message": "Compiled demo-test",
  "status": 2,
  "dataKind": "compile-report",
  "data": {
    "target": {
      "uri": "file:/home/piquerez/demo/#demo/Test"
    },
    "errors": 1,
    "warnings": 0,
    "time": 205
  }
}
```

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.10.0

ckipp01:
This fixes #2550 where a diagnostic from Scala 3 may be reported, but the `tokenEditDistance` is unable to be calculated since it's attempting to use the 2.13 dialect. This change ensures that any time we attempt to calculate the `tokenEditDistance` that we access the correct dialect first.

I marked this as a draft because I'm still playing around a bit with this. I'm not sure how I feel about literally passing `trees` _everywhere_ almost. However it does work.


metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
```diff
@@ -114,6 +114,6 @@ object Trees {
   /* Tokenizing works perfectly fine with 212 dialect as long as we are only
    * interested in having stable results. This is not the case for parsing.
    */
-  val defaultTokenizerDialect = scala.meta.dialects.Scala213
+  val defaultTokenizerDialect: Dialect = scala.meta.dialects.Scala213
```
tgodzik, line 117: We should probably remove it if possible.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -1213,42 +1224,43 @@ final class TestingServer(
...
-    Trees.defaultTokenizerDialect(input).tokenize.get.foreach { token =>
-      val params = token.toPositionParams(identifier)
-      val definition = server
-        .definitionOrReferences(params, definitionOnly = true)
-        .asJava
-        .get()
-      definition.definition.foreach { path =>
-        if (path.isDependencySource(workspace)) {
-          readonlySources(path.toNIO.getFileName.toString) = path
+    trees.getDialect(AbsolutePath(input.path))(input).tokenize.get.foreach {
```
tgodzik, line 1227: We could add something like:

```suggestion
    trees.tokenize(AbsolutePath(input.path)).foreach {
```
tokenize would return Option[Array[Token]]

tgodzik (commented):
Passing trees around is probably necessary :/ I wanted to avoid it, but unfortunately I don't see a way around it.

It would probably awesome if we had and adaptable parser available that would say what dialect it has, but parse everything anyway.  Alternatively we can have a Scala 213 dialect that enables all Dotty features, but we couldn't use that for diagnostics anymore. 

Buuut maybe we could use it in ScalaMtags? `Tree.EverythingDialect` :D

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
ckipp01: Yea, I was thinking the same thing. Lemme see if I can tackle the rest of the places that use this as well.

tests/unit/src/main/scala/tests/TestingServer.scala
ckipp01: Ah good idea.

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
tgodzik: Or really `scala.meta.dialects.Scala3` would probably work? We can make it a custom dialect, but might be simpler than in case of ScalaMtags

tests/unit/src/main/scala/tests/TestingServer.scala
ckipp01: Actually, looking at this I can't this to work right. Where are the implicit tokenize and dialect coming from?

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
ckipp01: 🤔 so this is proving to be a bit harder than I thought since all the digests would need this as well. So it's not ideal to have to pass around `Trees` to them. I think it might be better to fix this here, and then to create a follow-up to figure out how to fully get rid of them.

tgodzik: We don't really need Trees for digests. Sbt will stay on Scala 2 for a long while, I think the same with Mill.

tests/unit/src/main/scala/tests/TestingServer.scala
tgodzik: I think `scala.meta._`

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
ckipp01: Ah then honestly the main other things that are using it then are bench tests and digests. So it shouldb be good then.

tests/unit/src/main/scala/tests/TestingServer.scala
ckipp01: 🤔 I see we use it in `InferredTypeProvider` and indeed that works on `<string>.tokenize`, but it doesn't seem to work here on trees unless I'm totally doing something wrong.

tgodzik: Ach no, I just meant we need to add that method to Trees. Sorry if I was not clear enough.

ckipp01: 😆 honestly, that's what I first thought, when trying to do it I did a goto definition on `tokenize` where I was using it, and it took me to the scala meta tokenize extension method. Then I thought "ooooo, he must be trying to tell me to use this".... communication 😆

tgodzik: Yeah, I forgot there is so much magic with those extension methods.

ckipp01: haha, alright, fixed!

ckipp01:
So this should be good to have another look at. I ended up not fully getting rid of the default since we use it in a couple places that removing it didn't really justify all the changes at the moment (stuff like digest and benchmarks). So this will at least fix the linked issue and ensure it doesn't happen at all when calculating token edit distance anywhere.

tgodzik (approved):
LGTM!

adpi2:
Thanks @ckipp01 and @tgodzik for your reactivity!

Issue #2556, opened 2021-02-26
Metals logs out the entire open file on start up


ckipp01:
**Describe the bug**

When starting up Metals if you tail the logs, you'll see that right after a connection is made, the entire file is printed out.

**To Reproduce**

1. Just open a file and tail the logs.

**Expected behavior**

Don't do this.

**Logs**

<details>
<summary>logs</summary>
<br>

```
2021.02.26 11:03:02 INFO  Started: Metals version 0.10.0 in workspace '/Users/ckipp/Documents/scala-workspace/tester' for client Neovim 0.5.0.
2021.02.26 11:03:03 INFO  time: initialize in 1.24s
2021.02.26 11:03:03 INFO  Attempting to connect to the build server...
Starting the bsp launcher for bloop...
Opening a bsp server connection with 'bsp --protocol local --socket /var/folders/fq/nx_jsnyd6550xp03czx898d40000gn/T/bsp-launcher15758168778861240003/bsp.socket'...
Waiting for the bsp connection to come up...
2021.02.26 11:03:03 WARN  no build target for: /Users/ckipp/Documents/scala-workspace/tester/build.sbt
[D] Loading workspace settings from bloop.settings.json
[D] Waiting for a connection at local:///var/folders/fq/nx_jsnyd6550xp03czx898d40000gn/T/bsp-launcher15758168778861240003/bsp.socket...
The server is listening for incoming connections at local:///var/folders/fq/nx_jsnyd6550xp03czx898d40000gn/T/bsp-launcher15758168778861240003/bsp.socket...
2021.02.26 11:03:03 INFO  skipping build import with status 'Installed'
Starting thread that pumps stdin and redirects it to the bsp server...
Starting thread that pumps server stdout and redirects it to the client stdout...
2021.02.26 11:03:04 INFO  tracing is enabled: /Users/ckipp/Library/Caches/org.scalameta.metals/bsp.trace.json
2021.02.26 11:03:04 INFO  Attempting to connect to the build server...
Starting the bsp launcher for bloop...
2021.02.26 11:03:04Opening a bsp server connection with 'bsp --protocol local --socket /var/folders/fq/nx_jsnyd6550xp03czx898d40000gn/T/bsp-launcher16437070823763113898/bsp.socket'...
INFO  Attempting to connect to the build server...
Waiting for the bsp connection to come up...
Starting the bsp launcher for bloop...
[...]

PR #2557, opened 2021-02-26
Remove println in SemanticdbTextDocumentProvider

tgodzik:
Fixes https://github.com/scalameta/metals/issues/2556

I am so so very sorry about that. Tried to squeeze in too much into one release.


ckipp01 (approved):
ah what the heck, I even searched for println.... how did I miss this!

LGTM!

PR #2560, opened 2021-02-26
Show an error to the user if they are using a bad dep in a worksheet.

ckipp01:
I ran across this today and I knew to look in the logs right away to see
why my worksheet wasn't working, but I imagine in the situation where
someone is trying to use a dependency in their worksheet and it's not
the correct dep, it fails silently without any indication that something
is going wrong. So this change just captures if it's a resolution issue,
grabs the first line from coursier and actually shows it to the user.


tgodzik (approved):
LGTM!

Issue #2537, opened 2021-02-22
Rename refactoring does not work in VS Code


abzif:
**Describe the bug**

Rename refactoring does not work in VS Code

**To Reproduce**

Let's assume we have a simple scala code:
val a = 5
val b = a + 7

Put cursor on variable 'a' and press (Ctrl + Shift + R) - VS Code shows "No refactorings available"
or alternatively
Put cursor on variable 'a' and press F2 (shortcut for rename refactoring). It is possible to rename the variable but reference in the next line is not updated.

**Expected behavior**

Symbol name (variable name, function name, class name) and its references should be changed across the project - in the same file and other files referencing the symbol in the project.

**Installation:**

- Operating system: Windows 10
- Editor: Visual Studio Code version 1.53.2
- Extension Scala Metals version 1.9.13
- Extension Scala syntax version 0.5.1 (dependency of Scala Metals)
- Metals version: 0.9.10

tgodzik:
Thanks for reporting! I think `No refactorings available` is actually shown for code actions and rename is not part of those.

As for F2, could you provide a more complete example? Is this file in a project? Is it just a single file? There are some cases that rename will not work, which will be fixed by the next PR.

tgodzik:
Reported here: https://github.com/lampepfl/dotty/issues/11498

abzif:
[scala-example.zip](https://github.com/scalameta/metals/files/6021589/scala-example.zip)

Please look at the attached example. It is a very simple sbt project.
Look at comments in **Base** source file regarding variable, function and trait renaming

It seems that there are two problems here:
- Renaming seems to work properly if variable is referenced in a different file. However local references are not updated
- Subsequent try to rename fails with message that element cannot be renamed

tgodzik:
Can't seem to reproduce your issue:

![testing](https://user-images.githubusercontent.com/3807253/108704474-5c16fd80-750c-11eb-9521-04f879e26cde.gif)

do you get any errors or warning in `.metals/metals.log` ?

abzif:
Indeed, in the metals.log I got something like:

lut 22, 2021 1:08:59 PM scala.meta.internal.pc.CompilerAccess handleError
SEVERE: null
java.nio.file.ProviderMismatchException
	at sun.nio.fs.WindowsPath.toWindowsPath(WindowsPath.java:375)
	at sun.nio.fs.WindowsPath.relativize(WindowsPath.java:382)
[...]

PR #2574, opened 2021-03-01
Fix issues with renames on Windows

tgodzik:
Fixes https://github.com/scalameta/metals/issues/2537

Previously, renames would not work, because we would group locations by URI string, which might sometimes, especially on windows, be different:

`"file:///c%3A/Users/Tomasz/Documents/Work/scala-example/src/main/scala/Base.scala"`

vs

`"file:///C:/Users/Tomasz/Documents/Work/scala-example/src/main/scala/Base.scala"`

This fixes the issue by grouping via AbsoluteFile, which is the same for both cases.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -139,12 +139,13 @@ final class RenameProvider(
         }
 
         val allChanges = for {
-          (uri, locs) <- allReferences.toList.distinct.groupBy(_.getUri())
+          (path, locs) <- allReferences.toList.distinct
+            .groupBy(_.getUri().toAbsolutePath)
         } yield {
           val textEdits = for (loc <- locs) yield {
             textEdit(isOccurrence, loc, newName)
           }
-          Seq(uri.toAbsolutePath -> textEdits.toList)
+          Seq(path -> textEdits.toList)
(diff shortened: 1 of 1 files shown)


ckipp01:
It looks like we don't actually have any of the rename suites under cross, although we do have one named `RenameCrossLspSuite` under slow. Do you think we should have some type of rename suite under cross to make sure this doesn't resurface? That may have caught this as well.

tgodzik:
> It looks like we don't actually have any of the rename suites under cross, although we do have one named `RenameCrossLspSuite` under slow. Do you think we should have some type of rename suite under cross to make sure this doesn't resurface? That may have caught this as well.

We don't use presentation compiler for renames, so the cross test would not make sense here. Testing it will be actually pretty hard :/ as I would need to emulate the specific way VS Code encodes the URI.

ckipp01:
> > It looks like we don't actually have any of the rename suites under cross, although we do have one named `RenameCrossLspSuite` under slow. Do you think we should have some type of rename suite under cross to make sure this doesn't resurface? That may have caught this as well.
> 
> We don't use presentation compiler for renames, so the cross test would not make sense here. Testing it will be actually pretty hard :/ as I would need to emulate the specific way VS Code encodes the URI.

Ah of course. Makes sense. LGTM!

ckipp01 (approved):

PR #2578, opened 2021-03-03
Update metals doctor hint to match exact 'Import build' instead of 'Build import'

er1c:
This was slightly confusing to see:

image

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
+++ b/metals/src/main/scala/scala/meta/internal/troubleshoot/ScalaProblem.scala
@@ -15,7 +15,7 @@ sealed abstract class ScalaProblem {
    * Comprehensive message to be presented to the user.
    */
   def message: String
-  protected val hint = "run 'Build import' to enable code navigation."
+  protected val hint = "run 'Import build' to enable code navigation."
 }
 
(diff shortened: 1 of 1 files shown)


ckipp01:
Thanks for sendingthis in @er1c! So I think this makes sense. The only tricky part is that it seems VS Code is really the only one that uses "Import Build" instead of "Build Import". For example in both coc-metals and nvim-metals I went the "Build Import" route. We actually had a short convo a while back about completely renaming this https://github.com/scalameta/metals/discussions/2344.

Either way it's probably best to align. I don't know how other clients word it either. Wdyt @tgodzik?

er1c:
since metals is the reference implementation, it still probably makes sense IMHO, and probably isn't any more specific-confusing for other LSP clients

tgodzik (approved):
I think it's fine. Both are close enough that user should understand what's happening. We were thinking about maybe using "sync" but maybe it's a bit to late to change that.

ckipp01:
Not a huge deal at all, but I think the original wording in doctor was modeled after the command being `metals.build-import`. This will for sure align the hint with VS Code, but it now introduces a mismatch for other clients, such as the Vim clients. I'm not opposed to it, it's just a bummer that now Vim users (or any other client that modeled the UI off the actual command) may now have the same confusion  that @er1c had in VS Code.

PR #2579, opened 2021-03-04
Update millw scripts to the latest released ones

tgodzik:


ckipp01 (approved):
We should think of a way to remember to do this more often, ha. LGTM!

PR #2580, opened 2021-03-04
Add code action to rename import

tgodzik:
Created a quick WiP to gather some feedback. I found it often that I had to manually add braces and move the cursor to a correct place although it is pretty easy to add this code action.

rename-import

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/RenameImport.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/RenameImport.scala
@@ -1,0 +1,117 @@
+package scala.meta.internal.metals.codeactions
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+import scala.meta.Importee
+import scala.meta.Name
+import scala.meta.Tree
+import scala.meta.inputs.Position
+import scala.meta.internal.metals.CodeAction
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.parsing.Trees
+import scala.meta.pc.CancelToken
+
+import org.eclipse.lsp4j.TextEdit
+import org.eclipse.{lsp4j => l}
+import scala.meta.internal.metals.ServerCommands
+import scala.meta.internal.trees.Origin.Parsed
+import scala.meta.internal.mtags.Semanticdbs
+
+class RenameImport(trees: Trees, semanticdbs: Semanticdbs) extends CodeAction {
+
+  import RenameImport._
+  override def kind: String = l.CodeActionKind.Refactor
+
+  override def contribute(
+      params: l.CodeActionParams,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = Future {
+
+    def findLastEnclosingAtPos(tree: Tree, pos: Position): Option[Name] = {
+      tree.children.find { child =>
+        child.pos.start <= pos.start && pos.start <= child.pos.end
+      } match {
+        case None =>
+          tree match {
+            case name: Name => Some(name)
+            case _ => None
+          }
+        case Some(value) => findLastEnclosingAtPos(value, pos)
+      }
+    }
+
+    def renameImportCodeAction(name: Name): Option[l.CodeAction] =
+      name.parent.flatMap {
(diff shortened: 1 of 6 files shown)


dos65 (approved):
That's cool! :rocket:

metals/src/main/scala/scala/meta/internal/metals/codeactions/RenameImport.scala
```diff
@@ -0,0 +1,92 @@
...
+          }
+        case Some(value) => findLastEnclosingAtPos(value, pos)
+      }
+    }
+
+    def renameImportCodeAction(name: Name): Option[l.CodeAction] =
+      name.parent.flatMap {
+        case nm @ Importee.Name(_) =>
+          val range = nm.pos.toLSP
+          val text = s"{ $name => }"
```
ckipp01, line 46: So I'm not fully sure how this works with cursor position. For example I tried it in both nvim-metals and coc-metals just to see how two different clients would handle the codeAction edit and the cursor seems to stay in place for both of them.


### nvim-metals




### coc-metals

ckipp01 (commented):
Really neat idea! I think we need to figure out the cursor positioning though because on both of the clients I tried it on, the cursor ended up staying in place.

tgodzik:
> Really neat idea! I think we need to figure out the cursor positioning though because on both of the clients I tried it on, the cursor ended up staying in place.

Yeah, I managed to get it working by exposing the command in the extension, but I wanted to get some feedback before trying to solve this.

I wonder if there is any specific way in LSP to make sure that the cursor ends up at a particular place 🤔

ckipp01:
> > Really neat idea! I think we need to figure out the cursor positioning though because on both of the clients I tried it on, the cursor ended up staying in place.
> 
> Yeah, I managed to get it working by exposing the command in the extension, but I wanted to get some feedback before trying to solve this.
> 
> I wonder if there is any specific way in LSP to make sure that the cursor ends up at a particular place 🤔

I don't know if it's possible to do https://github.com/microsoft/language-server-protocol/issues/724

tgodzik:
> > > Really neat idea! I think we need to figure out the cursor positioning though because on both of the clients I tried it on, the cursor ended up staying in place.
> > 
> > 
> > Yeah, I managed to get it working by exposing the command in the extension, but I wanted to get some feedback before trying to solve this.
> > I wonder if there is any specific way in LSP to make sure that the cursor ends up at a particular place thinking
> 
> I don't know if it's possible to do [microsoft/language-server-protocol#724](https://github.com/microsoft/language-server-protocol/issues/724)

We might need to use the command then. Is `GotoLocation` command exposed in the other clients? From what I checked, edit is applied first and then the command executed. So in theory it should work unless nvim or coc do not support running client commands? We could do a workaround:
- code action has an attached command `metals.goto-location`
- serwer gets that command and invokes ClientCommands.GoToLocation
- client executes that command, which should already be supported

What do you think?

ckipp01:
> > > > Really neat idea! I think we need to figure out the cursor positioning though because on both of the clients I tried it on, the cursor ended up staying in place.
> > > 
> > > 
> > > Yeah, I managed to get it working by exposing the command in the extension, but I wanted to get some feedback before trying to solve this.
> > > I wonder if there is any specific way in LSP to make sure that the cursor ends up at a particular place thinking
> > 
> > 
> > I don't know if it's possible to do [microsoft/language-server-protocol#724](https://github.com/microsoft/language-server-protocol/issues/724)
> 
> We might need to use the command then. Is `GotoLocation` command exposed in the other clients? From what I checked, edit is applied first and then the command executed. So in theory it should work unless nvim or coc do not support running client commands? We could do a workaround:
> 
>     * code action has an attached command `metals.goto-location`
> 
>     * serwer gets that command and invokes ClientCommands.GoToLocation
> 
>     * client executes that command, which should already be supported
> 
> 
> What do you think?

Yea in theory I think that should work.

tgodzik:
Some further experiments:

metals-sample-1615827812535

This time I am first adding rename with some random new name, then running the command to save and rename that new symbol. Looks to be working fine aside from the wating time in case of saving the file. We should be able to work around that, but not yet sure how.

tgodzik:
I might get back to this at some point, but closing for now. Not entirely satisfied with what can be done.

PR #2581, opened 2021-03-07
Try to capture a better message on BSP exceptions.

ckipp01:
So in some cases if it was the reload that would fail, `T` here would
just be `Object`, which isn't very useful to show that it failed in
attempting to get `T`. Also, we are sort of swallowing the actual
message, so this just adds in a check to see if it's `Object`, then just
skip it and return the `Throwable` like normal, and if `T` is something
useful like `CompileResult` then we capture that _and_ also the original
`Throwable` message to pass on and later log.

This is for you @er1c 🎁 
Screenshot 2021-03-07 at 10 20 50


tgodzik (approved):
LGTM!

PR #2585, opened 2021-03-09
Change project directory to lower case when using New Scala Project

tgodzik:
Giter8 templates make the directory names lower case by default, which was not handled in Metals and we would open a new empty directory.


ckipp01 (approved):
Looks good! LGTM!

PR #2586, opened 2021-03-09
Fix renaming renamed imports

tgodzik:
Previously, when renaming a renamed import such as `import java.util.{List => JList}` we would try to rename the occurrences of List instead of the local `JList`. This would not work for a symbol from dependencies, but would rename anything locally defined in the workspace. This PR does two thing:
- we no longer rename the renamed symbol if rename was triggered from the renamed symbol
- we now rename all the local occurences of that renamed import


dos65 (approved):
Nice work! LGTM

ckipp01 (approved):
👍🏼 LGTM!

PR #2588, opened 2021-03-10
Update cache-action to v6

alexarchambault:
~Just checking how this will work out.~ `v6` should work better out-of-the-box with multiple jobs and matrices (better isolation between the jobs and the matrix entries).


alexarchambault:
Seems it worked fine… Caches are now really per job and matrix entry. Feel free to merge, if you'd like!

tgodzik (approved):
Will do thanks!

Issue #1695, opened 2020-05-05
Scala3 found issues


kpbochenek:
In repo: https://github.com/kpbochenek/dotty-complex

---

Wrong hint?(mouse is on `Red`):
![2020-05-05-171336_541x117_scrot](https://user-images.githubusercontent.com/10478402/81083151-81426d00-8ef4-11ea-866e-e345e95e4814.png)

I think it should be something like: 
`case Red: com.kpbochenek.example.dotty.SimpleEnum.Red`

---

Wrong hint(mouse is on `Green`):
![2020-05-05-172315_740x99_scrot](https://user-images.githubusercontent.com/10478402/81083771-44c34100-8ef5-11ea-9295-a26a422f4109.png)

Funny enough hints are okay for other cases(except first).

---

Selecting `color` highlights whole (...) block and doesn't highlight all occurrences.
Also searching for references of `color` doesn't work.
![2020-05-05-173108_284x88_scrot](https://user-images.githubusercontent.com/10478402/81085348-68878680-8ef7-11ea-9b2f-1c8a07126b50.png)

Selecting a place where `color` is used correctly highlights everything(even declaration place)

---

Jumping to definition of String works most of time except this case, where it finds 3 definitions(because misinterprets it as `makeColor` function ??
![2020-05-05-180036_353x177_scrot](https://user-images.githubusercontent.com/10478402/81087721-59560800-8efa-11ea-8a5d-2a2fd23ab30d.png)

---

Highlight ends in half of next word :)
Also searching for references being on `Cancelable` yields no results, but searching on any other `Cancelable` yields 5 results
![2020-05-05-180802_351x66_scrot](https://user-images.githubusercontent.com/10478402/81088559-63c4d180-8efb-11ea-82be-84c0f698ed9c.png)

tgodzik:
Thanks for checking it out, I think the first one I am already fixing in https://github.com/scalameta/metals/pull/1701

The rest is I think connected to how we generate semanticDB in Dotty and should be fixed there.

kpbochenek:
just checked after #1701 merge and `IdentType` is fixed :+1: 
If rest of problems must be solved directly in dotty you can close this ticket

tgodzik:
@kpbochenek  Maybe, let's keep this open or open a separate issue in Dotty? It's most likely still on our heads to fix it.

tgodzik:
I will keep this open so that the errors are not lost. We might need to do the work in Dotty, but better to leave it here.

PR #2589, opened 2021-03-10
Add tests for Scala 3 semanticdb and additional ones for hovers

tgodzik:
Fixes https://github.com/scalameta/metals/issues/1695

A bunch of problems from the issue were actually fixed, but a number of other issues surfaced, which I will follow up on in Dotty.


tests/unit/src/test/resources/semanticdb-scala3/example/Extension.scala
```diff
@@ -0,0 +1,9 @@
+package example
+
+/*example.Extension$package.*/extension (i/*example.Extension$package.asString().(i)*/: Int/*scala.Int#*/)
+  def asString/*example.Extension$package.asString().*/: String/*scala.Predef.String#*/ = i/*example.Extension$package.asString().(i)*/.toString/*scala.Any#toString().*/
+
+extension (s/*example.Extension$package.asInt().(s)*//*example.Extension$package.double().(s)*/: String/*scala.Predef.String#*/) {
```
tgodzik, line 6: https://github.com/lampepfl/dotty/issues/11690

tests/unit/src/test/resources/semanticdb-scala3/example/ColorEnum.scala
```diff
@@ -0,0 +1,6 @@
+package example
+
+enum Color/*example.Color#*/(val rgb/*example.Color#rgb.*/: Int/*scala.Int#*/):
+   case Red/*example.Color.Red.*/   extends Color/*example.Color#*/(0xFF0000)/*scala.runtime.EnumValue#*/
```
tgodzik, line 4: https://github.com/lampepfl/dotty/issues/11689

tests/unit/src/test/resources/semanticdb-scala3/example/GivenAlias.scala
```diff
@@ -0,0 +1,4 @@
+package example
+
+/*example.GivenAlias$package.*/given intValue/*example.GivenAlias$package.intValue.*/: Int/*scala.Int#*/ = 4
+given String/*scala.Predef.String#*/ = "st/*example.GivenAlias$package.given_String.*/r"
```
tgodzik, line 4: https://github.com/lampepfl/dotty/issues/11692

tests/unit/src/test/resources/semanticdb-scala3/example/Ord.scala
```diff
@@ -0,0 +1,12 @@
+package example
+
+trait Ord/*example.Ord#*/[T/*example.Ord#[T]*/]:
+   def compare/*example.Ord#compare().*/(x/*example.Ord#compare().(x)*/: T/*example.Ord#[T]*/, y/*example.Ord#compare().(y)*/: T/*example.Ord#[T]*/): Int/*scala.Int#*/
+
+/*example.Ord$package.*/given intOrd/*example.Ord$package.intOrd.*/: Ord/*example.Ord#*/[Int/*scala.Int#*/] with
```
tgodzik, line 6: https://github.com/lampepfl/dotty/issues/11693

mtags/src/main/scala/scala/meta/internal/mtags/SymbolOccurrenceOrdering.scala
```diff
@@ -19,18 +19,18 @@ object SymbolOccurrenceOrdering {
           if (byLine != 0) {
             byLine
           } else {
-            val byCharacter = Integer.compare(
-              a.startCharacter,
-              b.startCharacter
+            val byEnd = Integer.compare(
```
tgodzik, line 22: We add the decorations comments at the end of a range, which is why we should first sort by end. We don't actually use the start character at all in SemanticdbSuite, where this is used.

tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
```diff
@@ -42,6 +42,35 @@ class HoverScala3TypeSuite extends BaseHoverSuite {
...
+  // https://github.com/lampepfl/dotty/issues/11683
+  check(
+    "enums",
+    """|
+       |object SimpleEnum:
+       |  enum Color:
+       |   case <<Re@@d>>, Green, Blue
+       |
+       |""".stripMargin,
+    """|val Red: enums.SimpleEnum.Color.Red
```
tanishiking, line 55: I didn't realize that, thanks for adding this test and creating an issue on dotty :+1:

tests/unit/src/test/resources/semanticdb-scala3/example/ColorEnum.scala
tanishiking: What do you think about adding a link to this issue on a line comment in `ColorEnum.scala`, so we can see what's wrong with the test later?

tanishiking (approved):
LGTM! Great work for investigating the issues on semanticdb extractor on dotty 🚀 

One thing, what do you think about adding a comment that links to dotty issues on test cases?

tests/unit/src/test/resources/semanticdb-scala3/example/ColorEnum.scala
tgodzik: The problem is that it will show up for any other tests too. I will put it into the test suite itself.

tests/unit/src/test/scala/tests/SemanticdbSuite.scala
```diff
@@ -35,3 +41,16 @@ class SemanticdbSuite extends DirectoryExpectSuite("semanticdb") {
     }
   }
 }
+
+class SemanticdbScala2Suite
+    extends SemanticdbSuite(InputProperties.scala2(), "semanticdb")
+
+/**
+ * There is a number of issues observed in Scala 3 that can be seen in the tests:
+ * - extension methods https://github.com/lampepfl/dotty/issues/11690
```
tgodzik, line 50: Added the links here, otherwise the comments would get duplicated in the different expect suites

ckipp01 (approved):
Nice detective skills on these 😆 LGTM!

Issue #1086, opened 2019-11-27
Rename symbol not working on named parameters in constructor


ckipp01:
**Describe the bug**
Given the following code snippet, if I do a rename symbol on `isRenameAwesome` parameter, it doesn't seem to correctly rename the named parameter as I would expect.

```scala
package example

object Example extends App {

  class Test(isRenameAwesome: Boolean)
  val x = new Test(isRenameAwesome = true)

}
```

**To Reproduce**
1. Use the code above
2. Rename the `isRenameAwesome` parameter
3. See if the named parameter where x is created is renamed

**Installation:**
- Operating system: macOS
- Editor: Visual Vim
- Metals version: Latest Snapshot

**Additional context**
I'm not 100% sure if this is limited by this https://github.com/scalameta/scalameta/issues/1169 or if it is a different issue.

**Search terms**
rename symbol, named parameters

ckipp01:
So I've been looking into this a bit, and I see that you have the semanticdb label added to this @tgodzik. Is that meaning the fix for this needs to happen in scalameta? I basically took the example
```scala
package example

object Main {

  class Foo(isRenameAwesome2: Boolean)
  new Foo(isRenameAwesome = true)
}
```
And only the definition of `isRenameAwesome` shows up in `SymbolOccurrence`s.
```scala
        SymbolOccurrence(Some(Range(5, 9, 5, 9)), "example/Main.Foo#`<init>`().", REFERENCE),
        SymbolOccurrence(Some(Range(4, 12, 4, 27)), "example/Main.Foo#isRenameAwesome.", DEFINITION),
        SymbolOccurrence(Some(Range(5, 6, 5, 9)), "example/Main.Foo#", REFERENCE),
        SymbolOccurrence(Some(Range(4, 8, 4, 11)), "example/Main.Foo#", DEFINITION),
        SymbolOccurrence(Some(Range(0, 8, 0, 15)), "example/", DEFINITION),
        SymbolOccurrence(Some(Range(4, 11, 4, 11)), "example/Main.Foo#`<init>`().", DEFINITION),
        SymbolOccurrence(Some(Range(2, 7, 2, 11)), "example/Main.", DEFINITION),
        SymbolOccurrence(Some(Range(4, 29, 4, 36)), "scala/Boolean#", REFERENCE)
```
Is the idea for this to be captures via semantic db and therefore fixed in scalameta, or where you thinking of taking a different approach for this to capture that reference?

tgodzik:
[...]

PR #2596, opened 2021-03-11
Fix renaming named parameters in constructors

tgodzik:
Fixes https://github.com/scalameta/metals/issues/1086

This however will not work the other way. Main parameter will not be renamed when rename is started from the named parameter. That needs to be fixed in https://github.com/scalameta/metals/issues/1553

At first I tried to change the occurrence of the named parameter inside scemanticdb, that however proved to be more difficult and I was not 100% sure if the current status is actually wrong. Since it involves quite a bit of changes for the other tools I decided against going that way.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
@@ -125,6 +125,21 @@ final class ReferenceProvider(
           info.symbol.owner,
           Descriptor.Type(info.displayName)
         )
+    // Returns true if `info` is a named parameter of the primary constructor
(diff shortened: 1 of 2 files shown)


ckipp01:
> I was not 100% sure if the current status is actually wrong

Do you mean you you're unsure if the current semanticdb output for this is wrong?

ckipp01 (approved):

tgodzik:
> > I was not 100% sure if the current status is actually wrong
> 
> Do you mean you you're unsure if the current semanticdb output for this is wrong? Either way, this seems to do the trick 👍🏼

Yes, currently it outputs the information that it's a parameter of the primary constructor, which is true. It's also the name of a val/var however, so I was wondering whether to set it always as such.

Issue #2078, opened 2020-09-18
semanticdb indexing can fail on Windows with an exception trying to access a file opened by another process


smarter:
**Describe the bug**
Metals failed to properly import a simple project on Windows, here are the contents of metals.log:
```
> 2020.09.16 16:15:35 INFO  started: Metals version 0.9.3+82-189ba64a-SNAPSHOT in workspace 'C:\Users\bapti\cs210-recfun' for client vscode.
2020.09.16 16:15:36 INFO  time: initialize in 0.79s
2020.09.16 16:21:23 INFO  no build target: using presentation compiler with only scala-library
2020.09.16 16:21:23 WARN  no build target for: C:\Users\bapti\cs210-recfun\src\main\scala\recfun\RecFun.scala
sept. 16, 2020 4:21:23 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint handleCancellation
AVERTISSEMENT: Unmatched cancel notification for request id 3
sept. 16, 2020 4:21:23 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint handleCancellation
AVERTISSEMENT: Unmatched cancel notification for request id 1
2020.09.16 16:21:23 WARN  no build target for: C:\Users\bapti\cs210-recfun\src\main\scala\recfun\RecFunInterface.scala
2020.09.16 16:21:23 WARN  no build target for: C:\Users\bapti\cs210-recfun\src\main\scala\recfun\RecFun.scala
2020.09.16 16:21:38 WARN  no build target for: C:\Users\bapti\cs210-recfun\src\main\scala\recfun\RecFun.scala
2020.09.16 16:21:40 WARN  no build target for: C:\Users\bapti\cs210-recfun\src\main\scala\recfun\RecFun.scala
2020.09.16 16:22:31 WARN  no build target for: C:\Users\bapti\cs210-recfun\src\main\scala\recfun\RecFunInterface.scala
2020.09.16 16:22:37 WARN  no build target for: C:\Users\bapti\cs210-recfun\src\main\scala\recfun\RecFun.scala
2020.09.16 16:22:38 WARN  no build target for: C:\Users\bapti\cs210-recfun\src\main\scala\recfun\RecFun.scala
2020.09.16 16:22:38 WARN  no build target for: C:\Users\bapti\cs210-recfun\src\main\scala\recfun\RecFun.scala
[...]

PR #2599, opened 2021-03-12
Handle changed but locked files on Windows

Arthurm1:
This occurs when Metals attempts to read the file while Bloop is still writing to it here...
https://github.com/scalameta/metals/blob/dc0da653ead60396832d8b783e152acae368b38f/metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala#L92

Linux/Mac seem to allow reading of the file but, as it's only partially written, `com.google.protobuf.InvalidProtocolBufferException` is thrown and handled by Metals

Windows doesn't allow access to a locked file so throws a different exception: `java.nio.file.FileSystemException` but this isn't handled.

I think this is just a case of having to handle `FileSystemException` in the same way as `InvalidProtocolBufferException`.

I've tried this PR locally and the exception no longer appears.  I guess there's a worry that handling this will mask other errors but I've not seen this Exception thrown for any other reason on Windows or at all on WSL2.

Fixes: #2078

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
@@ -1,5 +1,6 @@
 package scala.meta.internal.metals
 
+import java.nio.file.FileSystemException
 import java.nio.file.Files
 import java.nio.file.Path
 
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM! Thanks for taking a look at it. I think it's fine to ignore it, we will get another notification once the file is written fully from what I experimented before.

tgodzik:
You might need to run `sbt scalafixAll` to make all the checks pass.

PR #2600, opened 2021-03-12
[Scala3] Support new Scala3 syntax in symbol index

dos65:
Summary:
 -  Indexer uses the correct dialect
 -  Added `Scala3ToplevelMtags`
 -  At least, it fixes go to toplevel `def`/`val`/`given`/`extension method`


mtags/src/main/scala-2/scala/meta/internal/mtags/ScalaMtags.scala
```diff
@@ -160,20 +202,146 @@ class ScalaMtags(val input: Input.VirtualFile)
...
+          withScala3Owner {
+            enterPatterns(t.pats, Kind.METHOD, Property.VAR.value); stop()
+          }
         case t: Decl.Var =>
-          enterPatterns(t.pats, Kind.METHOD, Property.VAR.value); stop()
+          withScala3Owner {
+            enterPatterns(t.pats, Kind.METHOD, Property.VAR.value); stop()
+          }
+        case t: Defn.Def =>
+          if (currentOwner.endsWith("/")) {
```
tgodzik, line 228: Maybe add an additional method?
```suggestion
          if (isFileOwner) {
```

mtags/src/main/scala-2/scala/meta/internal/mtags/ScalaMtags.scala
```diff
@@ -160,20 +202,146 @@ class ScalaMtags(val input: Input.VirtualFile)
...
       }
     }
+
+  /**
+   * Use sorce level package object as an owner for toplevel defition
+   */
+  private def withScala3Owner[A](f: => A): A =
+    withOwner(scala3Owner)(f)
+
+  private def scala3Owner: String =
```
tgodzik, line 333:
```suggestion
  private def fileOwner: String =
```
?

mtags/src/main/scala-2/scala/meta/internal/mtags/ScalaMtags.scala
```diff
@@ -160,20 +202,146 @@ class ScalaMtags(val input: Input.VirtualFile)
...
+            }
+          }
         case _ => stop()
       }
     }
+
+  /**
+   * Use sorce level package object as an owner for toplevel defition
+   */
+  private def withScala3Owner[A](f: => A): A =
```
tgodzik, line 330:
```suggestion
  private def withFileOwner[A](f: => A): A =
```

mtags/src/main/scala-2/scala/meta/internal/mtags/ScalaMtags.scala
```diff
@@ -160,20 +202,146 @@ class ScalaMtags(val input: Input.VirtualFile)
...
-          enterPatterns(t.pats, Kind.METHOD, Property.VAR.value); stop()
+          withScala3Owner {
+            enterPatterns(t.pats, Kind.METHOD, Property.VAR.value); stop()
+          }
         case t: Decl.Var =>
-          enterPatterns(t.pats, Kind.METHOD, Property.VAR.value); stop()
+          withScala3Owner {
+            enterPatterns(t.pats, Kind.METHOD, Property.VAR.value); stop()
+          }
+        case t: Defn.Def =>
```
tgodzik, line 227: Can this be removed from the template case above?

mtags/src/main/scala-2/scala/meta/internal/mtags/ScalaMtags.scala
```diff
@@ -160,20 +202,146 @@ class ScalaMtags(val input: Input.VirtualFile)
...
+            enterTermParameters(t.ctor.paramss, isPrimaryCtor = true)
+          }
+          term(t.name.value, t.name.pos, Kind.OBJECT, 0)
+          continue()
+        case t: Defn.RepeatedEnumCase =>
+          t.cases.foreach(c => withOwner(currentOwner)(term(c, Kind.OBJECT, 0)))
+        case t: Defn.EnumCase =>
+          term(t.name, Kind.OBJECT, 0)
+        case t: Defn.ExtensionGroup =>
+          // t.params are ignored - don't know which symbol/owner they should have
```
tgodzik, line 249: Let's also add an issue in Metals, so that we don't forget about this case.

mtags/src/main/scala-2/scala/meta/internal/mtags/ScalaMtags.scala
```diff
@@ -160,20 +202,146 @@ class ScalaMtags(val input: Input.VirtualFile)
...
+          // need to wait for https://github.com/lampepfl/dotty/issues/11690
+          val (owner, overloads) =
+            if (currentOwner.endsWith("/"))
+              (s"$currentOwner$toplevelSourceOwner", toplevelOverloads)
```
tgodzik, line 253:
```suggestion
              (scala3Owner, toplevelOverloads)
```
?

tests/unit/src/test/resources/definition-scala3/example/GivenAlias.scala
```diff
@@ -0,0 +1,4 @@
+package example
+
+given intValue/*GivenAlias.scala*/: Int/*Int.scala*/ = 4
+given String/*Predef.scala*/ = "str"
```
tgodzik, line 4: Can we also add a usage of the anonymous given? Also for non toplevel.

tests/unit/src/test/scala/tests/DefinitionSuite.scala
```diff
@@ -135,13 +148,42 @@ class DefinitionSuite extends DirectoryExpectSuite("definition") {
...
+      )
+    )
+
+class DefinitionScala3Suite
+    extends DefinitionSuiteBase(
+      inputProps = InputProperties.scala3(),
+      directory = "definition-scala3",
+      dialect = dialects.Scala3,
+      badFileNames = List(
+        "Extension.scala"
```
tgodzik, line 187: why is this excluded?

tests/unit/src/test/scala/tests/DefinitionSuite.scala
```diff
@@ -23,29 +24,41 @@ import scala.meta.internal.{semanticdb => s}
...
  * to separate test suites to assert that the range positions are accurate.
  */
-class DefinitionSuite extends DirectoryExpectSuite("definition") {
+
+abstract class DefinitionSuiteBase(
+    inputProps: InputProperties,
+    directory: String,
+    dialect: Dialect,
+    badFileNames: List[String] = List.empty,
+    filter: AbsolutePath => Boolean = (_: AbsolutePath) => true
```
tgodzik, line 33: Looks like the filter is not used

tgodzik (changes requested):
Thanks for the great work! It's a bit of a complex subject so it's great that you managed to figure it all out. I left some comments, but I don't think there is anything major.

build.sbt
```diff
@@ -500,9 +500,6 @@ lazy val input = project
   .settings(
     sharedSettings,
     skip.in(publish) := true,
-    scalacOptions ++= List(
-      "-P:semanticdb:synthetics:on"
```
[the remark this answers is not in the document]
dos65: This setting is duplicated on L508: https://github.com/scalameta/metals/blob/8f78cf55fc46e8b6fcdb7061b57b8dc4900a94df/build.sbt#L508

mtags/src/main/scala-2/scala/meta/internal/metals/SemanticdbDefinition.scala
```diff
@@ -42,7 +43,7 @@ object SemanticdbDefinition {
   ): Unit = {
     input.toLanguage match {
       case Language.SCALA =>
-        val mtags = new ScalaToplevelMtags(input, true) {
+        val mtags = new ScalaToplevelMtags(input, true, dialects.Scala213) {
```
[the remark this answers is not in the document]
dos65: Yep, but this method is used in search. 
I will create an issue because it would be better to do it in a separate PR. Getting the right dialect at call-sites of this method isn't trivial now.

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
```diff
@@ -194,14 +207,24 @@ final class OnDemandSymbolIndex(
...
       val toplevel = symbol.toplevel
       toplevels.get(toplevel.value) match {
         case Some(file) =>
           addMtagsSourceFile(file)
         case _ =>
-          loadFromSourceJars(trivialPaths(toplevel))
+          val scala3Toplevel =
+            if (toplevel.value.endsWith("$package.")) {
+              val path = toplevel.value.stripSuffix("$package.") + ".scala"
+              loadFromSourceDirectories(path).orElse(
```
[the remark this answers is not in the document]
dos65: After the proper implementation of `ScalaToplevelMtags`, there will be no need for `sourceDirectories` here.
That's a temporal workaround that was easy to implement.

Actually, I can remove this hack from this PR and fix toplevels in the next or push an [additional commit](https://github.com/dos65/metals/commit/c107ddb16a48bd04109370da189ac4a3124f975f) into this one .
What do you think would be the best option?

mtags/src/main/scala-2/scala/meta/internal/mtags/ScalaMtags.scala
dos65: It might be but requires to extract `OverloadsDisambiguator` into a variable.

tests/unit/src/test/scala/tests/DefinitionSuite.scala
dos65: To suppress errors because #2627

mtags/src/main/scala-2/scala/meta/internal/metals/SemanticdbDefinition.scala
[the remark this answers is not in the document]
tgodzik: Sure, makes sense

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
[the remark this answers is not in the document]
tgodzik: If you have it ready then let's include it

mtags/src/main/scala-2/scala/meta/internal/mtags/Mtags.scala
```diff
@@ -22,7 +24,11 @@ final class Mtags {
       List(toplevelClass)
     } else if (language.isScala) {
       addLines(language, input.text)
-      new ScalaToplevelMtags(input, includeInnerClasses = false)
+      new ScalaToplevelMtags(
+        input,
+        includeInnerClasses = false,
+        dialects.Scala213
```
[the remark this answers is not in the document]
dos65: Fixed

mtags/src/main/scala-2/scala/meta/internal/mtags/ScalaMtags.scala
```diff
@@ -160,20 +202,146 @@ class ScalaMtags(val input: Input.VirtualFile)
...
+              enterTermParameters(t.paramss, isPrimaryCtor = false)
+            }
+          }
+        case t: Defn.Enum =>
+          withOwner(currentOwner) {
+            tpe(t.name, Kind.CLASS, 0)
+            enterTypeParameters(t.tparams)
+            enterTermParameters(t.ctor.paramss, isPrimaryCtor = true)
+          }
+          term(t.name.value, t.name.pos, Kind.OBJECT, 0)
```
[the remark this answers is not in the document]
dos65: Enum defines type and term at once.
It's similar to:
```scala
trait X
object X {
  case object Y extends X
}
```

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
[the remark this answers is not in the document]
dos65: Done. Now it works without this hack

mtags/src/main/scala-2/scala/meta/internal/mtags/ScalaMtags.scala
dos65: Added `isPackageOwner`

mtags/src/main/scala-2/scala/meta/internal/mtags/ScalaMtags.scala
dos65: Changed in a little bit different way

metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
```diff
@@ -97,4 +97,35 @@ object ScalaVersions {
...
+
+    scalaVersionRegex
+      .findAllMatchIn(dropEnding)
+      .toList
+      .flatMap { m =>
+        val weight = if (Option(m.group(1)).isDefined) 0 else 1
+        val major = m.group(2)
+        val minor = m.group(3)
+        val ending = Option(m.group(4)).map(s => s"$s").getOrElse("")
+        val version = s"$major.$minor$ending"
```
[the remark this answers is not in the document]
dos65: Added doc.

> Get the version suffix

Hypothetically there might be some artifacts which version matches on scala version.
Smth like `library-name_3.0.0-RC1-2.12` .
And it's anyway unclear where is the version suffix starts in case if artifact name has `-` symbols

dos65:
@tgodzik 
I've additionally included toplevel parser for Scala3. This PR has doubled in size - sorry =)

I did some tweaks in `MetalsBench` locally to run the benchmark and it's close to scala 2 implementation:
```
[info] Benchmark                         Mode  Cnt  Score   Error  Units
[info] MetalsBench.toplevelsScala3Index    ss   10  0.435 ± 0.056   s/op
[info] MetalsBench.toplevelsScalaIndex     ss   10  0.418 ± 0.031   s/op
```
So technically, we could replace the existing implementation with this new one. But honestly, I would prefer to do it later as a separate PR.

mtags/src/main/scala-2/scala/meta/internal/mtags/Scala3ToplevelMtags.scala
```diff
@@ -0,0 +1,408 @@
...
+          case _ =>
+            scanner.nextToken()
+            loop(indent, false)
+        }
+      } else indent
+    }
+    loop(0, true)
+  }
+
+  def acceptToStartSep(): Unit = {
```
tgodzik, line 290:
```suggestion
  def acceptToStatSep(): Unit = {
```

mtags/src/main/scala-2/scala/meta/internal/mtags/ScalaMtags.scala
```diff
@@ -45,9 +49,31 @@ class ScalaMtags(val input: Input.VirtualFile)
   override def indexRoot(): Unit = {
     root match {
       case Parsed.Success(tree) => apply(tree)
-      case _ => // do nothing in case of parse error
+      case err: Parsed.Error => // do nothing in case of parse error
+        println(err)
```
tgodzik, line 53: Remove the println?

metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
```diff
@@ -97,4 +97,45 @@ object ScalaVersions {
...
+   */
+  def scalaBinaryVersionFromJarName(filename: String): Option[String] = {
+    val dropEnding = filename
+      .stripSuffix(".jar")
+
+    scalaVersionRegex
+      .findAllMatchIn(dropEnding)
+      .toList
+      .flatMap { m =>
+        val weight = if (Option(m.group(1)).isDefined) 0 else 1
```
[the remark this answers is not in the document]
dos65: To provide a higher priority to a version that has `_` prefix.
The case is rare but what if there are artifacts like `smth-library_2.13-2.11` which has two parts that look like a valid scala version.

I will change it to smth like `hasUndescorePrefix: Boolean` to make it less unclear

mtags/src/main/scala-2/scala/meta/internal/mtags/Scala3ToplevelMtags.scala
```diff
@@ -0,0 +1,408 @@
...
+      case _ =>
+        require(isOk = false, "package name or package object")
+        false
+    }
+  }
+
+  /**
+   * Consume token stream like "a.b.c" and return List(a, b, c)
+   */
+  def parsePath(): List[Identifier] = {
```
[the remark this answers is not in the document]
dos65: Yep, I thought about that.

However, as this parser works well for Scala 2 too I planned to propose to left it as the only implementation after passing review/merge. So the work related to the extraction of common methods might be useless. wdyt?

tests/unit/src/test/resources/definition-scala3/example/Extension.scala
```diff
@@ -0,0 +1,15 @@
+package example
+
+extension/*<no symbol>*/ (i/*Extension.scala fallback to example.Extension$package.*/: Int/*Int.scala*/)
+  def asString/*Extension.scala fallback to example.Extension$package.*/: String/*Predef.scala*/ = i/*Extension.scala fallback to example.Extension$package.*/.toString
+
+extension/*<no symbol>*/ (s/*Extension.scala fallback to example.Extension$package.*/: String/*Predef.scala*/) {
+  def asInt/*Extension.scala fallback to example.Extension$package.*/: Int/*Int.scala*/ = s/*Extension.scala fallback to example.Extension$package.*/.toInt/*StringOps.scala*/
+  def double/*Extension.scala fallback to example.Extension$package.*/: String/*Predef.scala*/ = s/*Extension.scala fallback to example.Extension$package.*/ */*StringOps.scala*/ 2
```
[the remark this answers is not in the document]
dos65: I think it comes from `DefinitionAlternatives.methodOwner`

metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
```diff
@@ -97,4 +97,45 @@ object ScalaVersions {
       case _ => Scala213
     }
   }
+
+  private val scalaVersionRegex =
+    "(_?)(\\d)\\.(\\d{1,2})(\\.\\d(-(RC|M)\\d)?)?".r
+
+  /**
+   * Extract scala binary version from dependency jar name.
```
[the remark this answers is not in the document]
dos65: Unfortunately, we will anyway have to keep this thing. The new BSP method will be optional.
There will be an additional less creepy version of this method for it.

tgodzik (approved):
LGTM! Great work! Let's just rebase the changes and we can merge.

PR #2601, opened 2021-03-12
Find alternative symbol references when not searching from the definition file

tgodzik:
Previously, we would search the current file, when searching for references, for alternative references such as object reference of a case class. This caused us to get different results depending on where we invoked textDocument/references. Now we check those alternatives in the definition file.

Additionally, we now also:
- check for case classes if searching from an object, which is synthetic
- check for field if searching from constructor parameter

Fixes https://github.com/scalameta/metals/issues/1553

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
@@ -85,7 +85,7 @@ final class ReferenceProvider(
           definition.positionOccurrence(source, params.getPosition, doc)
         maybeOccurrence match {
           case Some(occurrence) =>
-            val alternatives = referenceAlternatives(doc, occurrence)
+            val alternatives = referenceAlternatives(occurrence.symbol, doc)
             val locations = references(
               source,
               params,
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -113,94 +113,62 @@ final class ReferenceProvider(
...
+      Some(referenceDoc)
+    } else {
+      for {
+        location <- definition.fromSymbol(symbol).asScala.headOption
+        source = location.getUri().toAbsolutePath
+        definitionDoc <- semanticdbs.textDocument(source).documentIncludingStale
+      } yield definitionDoc
+    }
+
+    definitionDoc match {
```
tgodzik, line 129: So previously, we would check any document that the cursor currently points to. This would cause us not to find synthetic objects, since their definition only exists in the file where the case class was also defined.

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -113,94 +113,62 @@ final class ReferenceProvider(
...
+      case Some(definitionDoc) =>
+        val name = symbol.desc.name.value
+        val alternatives = new SymbolAlternatives(symbol, name)
+
+        val candidates = for {
```
tgodzik, line 134: Here we search for any synthetic symbols that we might be also interested in.

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -113,94 +113,62 @@ final class ReferenceProvider(
...
+
+        val nonSyntheticSymbols = for {
+          occ <- definitionDoc.occurrences
+          if isCandidate(occ.symbol) || occ.symbol == symbol
+          if occ.role.isDefinition
+        } yield occ.symbol
+
+        def isSyntheticSymbol = !nonSyntheticSymbols.contains(symbol)
+
+        def additionalAlternativesForSynthetic = for {
```
tgodzik, line 155: This is an additional section that I actually added. We sometimes want to find non-synthetic symbols for synthetic symbols such as:
- case class for a synthetic object
- class field for a synthetic constructor param (they always synthetic when in `new A(param = "")`)

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -353,3 +321,102 @@ final class ReferenceProvider(
   }
 
 }
+
+class SymbolAlternatives(symbol: String, name: String) {
```
tgodzik, line 325: Almost everything is moved except for isCompanionClass and isFieldParam

dos65 (approved):
LGTM!

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -113,94 +113,62 @@ final class ReferenceProvider(
...
+      } yield definitionDoc
+    }
+
+    definitionDoc match {
+      case Some(definitionDoc) =>
+        val name = symbol.desc.name.value
+        val alternatives = new SymbolAlternatives(symbol, name)
+
+        val candidates = for {
+          info <- definitionDoc.symbols.iterator
```
ckipp01, line 135: Do we need the `iterator` here since `symbols` is already just a seq?

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -113,94 +113,62 @@ final class ReferenceProvider(
...
+        val nonSyntheticSymbols = for {
+          occ <- definitionDoc.occurrences
+          if isCandidate(occ.symbol) || occ.symbol == symbol
+          if occ.role.isDefinition
+        } yield occ.symbol
+
+        def isSyntheticSymbol = !nonSyntheticSymbols.contains(symbol)
+
+        def additionalAlternativesForSynthetic = for {
+          info <- definitionDoc.symbols.iterator
```
ckipp01, line 156: Same question as up above, do we need `iterator` here?

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
ckipp01: This could just be a case class no?

ckipp01 (commented):
Nice job with this @tgodzik! Just a couple questions.

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
tgodzik: I just copied it, so that might now be needed at all, I will check that out!

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
ckipp01: That's fine 👍🏼

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
tgodzik: Fixed

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
tgodzik: Fixed

ckipp01 (approved):
Looks good! LGTM!

PR #2602, opened 2021-03-13
Show better message if no semanticdb is found for testFile.

### LINKED ISSUE #2598: Nothing happens when `testFile` is triggered and no semanticdb exists for that path
**Describe the bug**

When using the new `testFile` functionality, there is logic that captures the various possible states and shows that to the user. For example if the workspace has errors, if there are no tests in that file etc. One that isn't handled well is if there is no semanticdb for a given path. @alexcardell pointed this out to me on Gitter that when this is the case, nothing really happens in the UI, but in the logs you see

```
SEVERE: Internal error: java.lang.Exception: Semanticdb not found for path.                                         
```

I'm curious what your thoughts are @tgodzik on what to show to the user in this case? Showing "No semanticdb found for this path" is probably not clear at all to the average user, so I'm a bit unsure of what to return. _Or_ do we just leave this case up to the client to handle? I ask because it's the only one that we don't handle here:

https://github.com/scalameta/metals/blob/dc0da653ead60396832d8b783e152acae368b38f/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala#L408-L447

**Expected behavior**

A clearer message for the user, but I'm not sure what it should be.

**Search terms**

semanticdb dap testFile

ckipp01:
Previously when there was no semanticdb found for the file you were
trying to test via `testFile` an exception would happen and the user
wouldn't be notified at all. Now they are shown a bit clearer message
similar to that of when navigation fails for a similar reason.

The second part of this adds in a check so when a user is using debug
functionality a check is made to ensure they have metals status enabled
before sending the status, and if not it handles messages and logs as
well.

Fixes #2598

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
@@ -19,6 +19,7 @@ import scala.util.Try
 import scala.meta.internal.metals.BuildServerConnection
 import scala.meta.internal.metals.BuildTargets
 import scala.meta.internal.metals.ClientCommands
+import scala.meta.internal.metals.ClientConfiguration
 import scala.meta.internal.metals.Compilations
 import scala.meta.internal.metals.DebugDiscoveryParams
 import scala.meta.internal.metals.DebugUnresolvedAttachRemoteParams
@@ -33,6 +34,7 @@ import scala.meta.internal.metals.Messages.UnresolvedDebugSessionParams
 import scala.meta.internal.metals.MetalsBuildClient
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.MetalsLanguageClient
+import scala.meta.internal.metals.MetalsStatusParams
 import scala.meta.internal.metals.StacktraceAnalyzer
 import scala.meta.internal.metals.StatusBar
 import scala.meta.internal.metals.config.RunType
@@ -67,7 +69,8 @@ class DebugProvider(
     index: OnDemandSymbolIndex,
     stacktraceAnalyzer: StacktraceAnalyzer,
     icons: Icons,
-    semanticdbs: Semanticdbs
+    semanticdbs: Semanticdbs,
+    clientConfig: ClientConfiguration
 ) {
 
   lazy val buildTargetClassesFinder = new BuildTargetClassesFinder(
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -407,13 +413,25 @@ class DebugProvider(
...
-          ClientCommands.FocusDiagnostics.id,
-          Nil.asJava
-        )
-      )
+      val msg =
+        "Cannot run or debug due to existing errors in the workspace. " +
+          "Please fix the errors and retry."
+      clientConfig.statusBarState() match {
+        case On =>
+          statusBar.addMessage(Messages.DebugErrorsPresent(icons, msg))
```
tgodzik, line 421: Could we actually handle this all in `statusBar`? This way even if someone forgets to properly check statusBar config, it would work. This would most likely deduplicate the code below also.

`FocusDiagnostics` should probably be handled by a different option.

tgodzik (changes requested):
Great work! That one was hard to reproduce. I left a comment that might be useful for the future and in case someone forgets about status bar to being supported everywhere (that was probably me :sweat_smile: ).

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
ckipp01: > Could we actually handle this all in statusBar? This way even if someone forgets to properly check statusBar config, it would work. This would most likely deduplicate the code below also.

Ah, actually it looks like I can just use this:

https://github.com/scalameta/metals/blob/9ac13b22569f56128527abe424e556725fb62f57/metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala#L35-L49

> FocusDiagnostics should probably be handled by a different option.

What do you mean?

tgodzik: I just meant that if we want to send the focus diagnostics command then we should check if that is handled by the client irrelevant of the status bar.

ckipp01: ahhh ok, gotcha yea. Agreed.

tgodzik (approved):
LGTM!

PR #2604, opened 2021-03-14
Fix renaming renamed imports when cursor is at the rename

tgodzik:
Previously, we fixed renaming to also work with renamed imports in a local scope, however we haven't check if it works at the actual rename position. Now, the test is added and the case is fixed.


ckipp01 (approved):
Looks good. LGTM!

dos65 (approved):
:+1:

PR #2605, opened 2021-03-14
Automatically backtick symbols when renaming

tgodzik:
In cases where a symbol name is not valid we should automatically wrap it in backticks

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -18,11 +18,13 @@ import scala.meta.internal.metals.MetalsLanguageClient
 import scala.meta.internal.metals.ReferenceProvider
 import scala.meta.internal.metals.TextEdits
 import scala.meta.internal.parsing.Trees
+import scala.meta.internal.pc.Identifier
 import scala.meta.internal.semanticdb.Scala._
 import scala.meta.internal.semanticdb.SelectTree
 import scala.meta.internal.semanticdb.SymbolOccurrence
 import scala.meta.internal.semanticdb.Synthetic
 import scala.meta.internal.semanticdb.TextDocument
+import scala.meta.internal.{semanticdb => s}
 import scala.meta.io.AbsolutePath
 import scala.meta.pc.CancelToken
 
@@ -30,7 +32,6 @@ import org.eclipse.lsp4j.Location
 import org.eclipse.lsp4j.MessageParams
 import org.eclipse.lsp4j.MessageType
 import org.eclipse.lsp4j.Position
-import org.eclipse.lsp4j.Range
 import org.eclipse.lsp4j.ReferenceContext
 import org.eclipse.lsp4j.ReferenceParams
 import org.eclipse.lsp4j.RenameFile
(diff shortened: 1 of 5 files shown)


metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -101,8 +102,9 @@ final class RenameProvider(
         val newName =
           if (suggestedName.charAt(0) == '`')
             suggestedName.substring(1, suggestedName.length() - 1)
-          else suggestedName
-
+          else {
+            Identifier.backtickWrap(suggestedName)
+          }
```
ckipp01, line 107: So it looks like this screws a bit with renamed imports as well. Since if you have `Java-List` already in backtics and you attempt to rename it, the `suggestedName` for it here is without ticks causing it to be wrapped in ticks and then you end up with double ticks. 

Also total nit, but can can we try to be consistent with blocks or not instead of mixing. So either

```suggestion
else Identifier.backtickWrap(suggestedName)
```
or add blocks to the if clause.

tests/unit/src/test/scala/tests/RenameLspSuite.scala
```diff
@@ -288,7 +288,7 @@ class RenameLspSuite extends BaseRenameLspSuite("rename") {
        |  "" <<::>> user
        |}
        |""".stripMargin,
-    newName = "method:"
+    newName = "+++:"
```
tgodzik, line 291: That was actually a wrongly named method - `method:` should be wrapped in backticks, which it started doing.

tests/unit/src/test/scala/tests/RenameLspSuite.scala
```diff
@@ -562,6 +562,50 @@ class RenameLspSuite extends BaseRenameLspSuite("rename") {
...
+       |}
+       |object Main2{
+       |  val toRename = Main.<<`to-R@@ename`>>
+       |}
+       |""".stripMargin,
+    newName = "`other-rename`"
+  )
+
+  renamed(
+    "backtick-not-needed",
```
tgodzik, line 596: We will also now unbacktick anything that doesn't need backticks anymore. It will not backtick anything that already had them like when used in pattern matches:
```scala
val a = 12
val b = 12
a match{
  `b` => println("They are equal!")
}
```

tgodzik: Actually, scratch that. We cannot unbacktick anything as it might still be needed in some cases like for example pattern matches.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
tgodzik: Ok it turns out this was harder than aniticipated :sweat_smile:  I added some more tests, made the existing ones pass. Now the current behaviour will add backticks if they needed, but are not yet added.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -241,7 +231,7 @@ final class RenameProvider(
       realName = rng.inString(semanticDb.text)
       if occ.symbol == occurence.symbol &&
         withoutBacktick(realName) == withoutBacktick(renameName)
-    } yield realLocationWithoutBackticks(realName, rng.toLSP)
+    } yield new Location(uri, rng.toLSP)
```
tgodzik, line 234: If new name is checked to have backticks then this is not needed.

metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
ckipp01: haha it always is.

ckipp01 (approved):
LGTM!

PR #2614, opened 2021-03-15
Revert cache-action update

ckipp01:


tgodzik (approved):
Rerunning https://github.com/scalameta/metals/pull/2609 worked, however we might be better to revert until this is handled.

ckipp01:
> Rerunning #2609 worked, however we might be better to revert until this is handled.

huh, if it did work, maybe let's try to re-run them all, and then see if that fixes it. It could also be that when Steward sent them in all at the same times there were issues going on.

alexarchambault:
What happened with the new version?

ckipp01:
> What happened with the new version?

https://github.com/scalameta/metals/pull/2608#issuecomment-799228719

alexarchambault:
Reverting should just make most jobs use the same cache key. Some jobs' artifacts may never be cached because of that (as they're reading another job's cache).

Maybe the action should just trap those errors, like in https://github.com/world-federation-of-advertisers/actions/pull/3.

I opened https://github.com/coursier/cache-action/issues/131 to track this.

ckipp01:
Thanks for creating that @alexarchambault. I'll go ahead and close this as re-running seems to have fixed it for everything.

ckipp01:
Alright, sorry to keep bouncing back and forth, but there are a bunch that are failing again. I'd rather just not deal with this and revert.

PR #2615, opened 2021-03-15
Add init option to strip color from DAP communication.

### LINKED ISSUE #2061: Is there a way to turn off color when running tests via `debug-adapter-start`?
Is there a way to easily disable colors in the bloop instance?
`^[[32m[` I see lots of escape characters in the output.
Thanks!

ckipp01:
This is meant to address the situations where a DAP client may not be
able to handle color codes in the DAP output. Instead of trying to build
this into both sbt and Bloop I think it's just easier to strip it here
and put it behind a flag. The normal behavior is unchanged, but now a
client can set `stripDAPOutputColor = true` and get the color codes
stripped.

Closes #2061

### Before

Screenshot 2021-03-15 at 15 44 53
Screenshot 2021-03-15 at 15 45 33

### After with `stripDAPOutputColor` set to true

Screenshot 2021-03-15 at 15 48 11
Screenshot 2021-03-15 at 15 48 41

This should also take care of:

- https://github.com/scalameta/nvim-metals/issues/107
- https://github.com/scalacenter/bloop/issues/1478

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
@@ -39,6 +39,8 @@ import org.eclipse.{lsp4j => l}
  * @param openNewWindowProvider if the client can open a new window after new project creation.
  * @param copyWorksheetOutputProvider if the client can execute server CopyWorksheet command and
  *                                    copy results to the local buffer.
+ * @param disableColorOutput in the situation where your DAP client may not handle color codes in
+ *                            the output, you can enable this to strip them.
  */
 final case class InitializationOptions(
     compilerOptions: CompilerInitializationOptions,
@@ -61,7 +63,8 @@ final case class InitializationOptions(
     statusBarProvider: Option[String],
     treeViewProvider: Option[Boolean],
     openNewWindowProvider: Option[Boolean],
-    copyWorksheetOutputProvider: Option[Boolean]
+    copyWorksheetOutputProvider: Option[Boolean],
+    disableColorOutput: Option[Boolean]
 ) {
   def doctorFormat: Option[DoctorFormat.DoctorFormat] =
     doctorProvider.flatMap(DoctorFormat.fromString)
@@ -75,6 +78,7 @@ object InitializationOptions {
 
   val Default: InitializationOptions = InitializationOptions(
     CompilerInitializationOptions.default,
+    None,
     None,
     None,
     None,
@@ -143,7 +147,8 @@ object InitializationOptions {
       treeViewProvider = jsonObj.getBooleanOption("treeViewProvider"),
       openNewWindowProvider = jsonObj.getBooleanOption("openNewWindowProvider"),
       copyWorksheetOutputProvider =
-        jsonObj.getBooleanOption("copyWorksheetOutputProvider")
+        jsonObj.getBooleanOption("copyWorksheetOutputProvider"),
+      disableColorOutput = jsonObj.getBooleanOption("disableColorOutput")
     )
   }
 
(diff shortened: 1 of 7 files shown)


metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
```diff
@@ -61,7 +63,8 @@ final case class InitializationOptions(
     statusBarProvider: Option[String],
     treeViewProvider: Option[Boolean],
     openNewWindowProvider: Option[Boolean],
-    copyWorksheetOutputProvider: Option[Boolean]
+    copyWorksheetOutputProvider: Option[Boolean],
+    stripDAPOutputColor: Option[Boolean]
```
tgodzik, line 67:
```suggestion
    disableColorOutput: Option[Boolean]
```
does a more general flag make sense?

tgodzik (approved):
Looks good, just one minor comment, but not important at all.

metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
ckipp01: Yea I went back and forth about whether it should include DAP or not. I went ahead and made it more generic. 👍🏼

adpi2:
Thanks @ckipp01!

Do you think it would be possible to reuse `stripDAPOutputColor` or something similar for BSP logs. My Metals output in vscode looks like this:

```
2021.03.16 21:18:35 ERROR [31m-- [E007] Type Mismatch Error: /home/piquerez/adpi2/pandora/src/main/scala/pandora/Product.scala:43:31 [0m
[31m43 |[0m    (Term(tags.size, factors), mapping)
   |                               [31m^^^^^^^[0m
   |   Found:    [1m[31m([0m[1m[31mmapping[0m[1m[31m [0m[1m[31m:[0m[1m[31m [0mList[List[List[([1m[31mAny[0m, [1m[31mAny[0m, Int, [1m[31mNothing[0m, [1m[31mAny[0m)]]][1m[31m)[0m
   |   Required: List[List[List[([1m[32mInt[0m, [1m[32mInt[0m, Int, [1m[32mInt[0m, [1m[32mInt[0m)]]]
```

ckipp01:
> Thanks @ckipp01!
> 
> Do you think it would be possible to reuse `stripDAPOutputColor` or something similar for BSP logs. My Metals output in vscode looks like this:
> 
> ```
> 2021.03.16 21:18:35 ERROR �[31m-- [E007] Type Mismatch Error: /home/piquerez/adpi2/pandora/src/main/scala/pandora/Product.scala:43:31 �[0m
> �[31m43 |�[0m    (Term(tags.size, factors), mapping)
>    |                               �[31m^^^^^^^�[0m
>    |   Found:    �[1m�[31m(�[0m�[1m�[31mmapping�[0m�[1m�[31m �[0m�[1m�[31m:�[0m�[1m�[31m �[0mList[List[List[(�[1m�[31mAny�[0m, �[1m�[31mAny�[0m, Int, �[1m�[31mNothing�[0m, �[1m�[31mAny�[0m)]]]�[1m�[31m)�[0m
>    |   Required: List[List[List[(�[1m�[32mInt�[0m, �[1m�[32mInt�[0m, Int, �[1m�[32mInt�[0m, �[1m�[32mInt�[0m)]]]
> ```

🤔 huh, so this only seems to happen when using Scala 3 but not 2. I'm actually surprised that VS Code doesn't handle these in the output panel, but it does in the debug panel?

In theory, yea we could re-use ~`stripDAPOutputColor`~ `disableColorOutput` (we renamed it), and it would strip the color here, but it would also strip the color in the debug console for the user which I'm assuming they may not want.

PR #2617, opened 2021-03-16
Run Scala 3 migrate on mtags sources

tgodzik:
While testing out https://github.com/scalacenter/scala3-migrate I decided to run it on mtags module to see if there is anything that we can improve.

This included:
- changes to scalac options for Scala 3
- scalafix rewrites that deals with 2.13 deprecations, which weren't sctrictly neccessary, but might be usful for the future if we decide to move some of the Scala 2 code
- I also added a test case for one of Scala 3 issues we had (hovers on extension methods)

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/Identifier.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/Identifier.scala
@@ -62,7 +62,7 @@ object Identifier {
   def backtickWrap(s: String): String = {
     if (s.isEmpty) "``"
     else if (s(0) == '`' && s.last == '`') s
-    else if (needsBacktick(s)) '`' + s + '`'
+    else if (needsBacktick(s)) "" + ('`') + s + '`'
     else s
   }
 
--- a/tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
+++ b/tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
@@ -71,6 +71,21 @@ class HoverScala3TypeSuite extends BaseHoverSuite {
        |""".stripMargin.hover
   )
 
+  // https://github.com/scalameta/metals/issues/1991
+  check(
+    "extension-methods",
+    """|
+       |object Foo:
+       |    extension (s: String):
+       |        def double = s + s
+       |        def double2 = s + s        
+       |    end extension
+       |    "".<<doub@@le2>>
+       |end Foo
+       |""".stripMargin,
+    ""
+  )
+
   // TODO: better printing for using
   // currently "def apply[T](a: T)(implicit x$2: Int): T"
   check(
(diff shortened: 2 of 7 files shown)


tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
```diff
@@ -71,6 +71,21 @@ class HoverScala3TypeSuite extends BaseHoverSuite {
        |""".stripMargin.hover
   )
 
+  // https://github.com/scalameta/metals/issues/1991
+  check(
+    "extension-methods",
```
dos65, line 76: This one was fixed in dotty?

dos65 (approved):

tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
tgodzik: Might be fixed in 3.0.0-RC2, so I wanted to add the test case to close the issue when updating.

mtags/src/main/scala-2/scala/meta/internal/pc/Identifier.scala
```diff
@@ -62,7 +62,7 @@ object Identifier {
   def backtickWrap(s: String): String = {
     if (s.isEmpty) "``"
     else if (s(0) == '`' && s.last == '`') s
-    else if (needsBacktick(s)) '`' + s + '`'
+    else if (needsBacktick(s)) "" + ('`') + s + '`'
```
ckipp01, line 65: I've seen this in a couple places where it prefaces with an empty string. What's the reasoning for that?

ckipp01 (approved):
Just a quick question, but other than that LGTM!

mtags/src/main/scala-2/scala/meta/internal/pc/Identifier.scala
tgodzik: I was also confused, but it seems the reason is https://scalacenter.github.io/scala-3-migration-guide/docs/incompatibilities/dropped-features.html#any2stringadd-conversion

ckipp01: huh, alright, thanks for linking that. I hope this doesn't become confusing for someone in the future that comes across this and wonders why there is a `""` prefacing it. I can see myself seeing that just removing it 😆

dos65: Hm. I've just realized that this line might look less weird if at least the first backtick was a type of `String`:
```scala
else if (needsBacktick(s)) "`"+ s + "`"`
```

tgodzik: ach yes, I didn't notice the comments when merging, sorry!

PR #2620, opened 2021-03-17
Update dead links

ajzawawi:
Found some old dead links that needed an update


ckipp01 (approved):
Thanks for finding and fixing these @ajzawawi! LGTM!

Issue #1704, opened 2020-05-06
Document highlight highlights different signatures at the same time


tgodzik:
**Describe the bug**
Document highlight is not behaving properly for Scala 3 in case of different signatures.

**To Reproduce**
1. Create a file in Scala 3 workspace:
```
class Foo {
  def hello() = ""
  def hello(a : Int) = ""
  def hello(a : Int, b : String) = ""
}
```
2. Click on second hello
3. First hello will also be highlighted

**Expected behavior**
Only second hello is highlighted

**Screenshots**

![image](https://user-images.githubusercontent.com/3807253/81205832-845c5c80-8fcb-11ea-9abe-acb901003ce6.png)

**Installation:**
- Operating system: macOS/Windows/Linux
- Editor: Visual Studio Code/Atom/Vim/Sublime/Emacs
- Metals version: v0.9.0

**Additional context**
This might need some fixes on the Dotty side.

**Search terms**
document highlight scala 3

ckipp01:
Just checked on this. Using 2.13.5 you get the same behavior. I don't believe this is just a Scala 3 issue.

tgodzik:
> Just checked on this. Using 2.13.5 you get the same behavior. I don't believe this is just a Scala 3 issue.

Good catch! I hadn't figured that it might actually be a more global bug. It was actually the first thing that I implemented so there might be a bunch of issues here :sweat_smile:

PR #2622, opened 2021-03-17
Fix highlighting overloaded methods

tgodzik:
PReviously, we would always highlight 0 parameter method when the cursor was on any other overlaoded method. Now, we only highlight the right occurrence.

Fixes https://github.com/scalameta/metals/issues/1704


dos65 (approved):
Looks good!

Issue #2525, opened 2021-02-18
Scala 3 + Akka streams + worksheet not working


Swoorup:
**Describe the bug**

Scala 3 Worksheet does not work when using Akka Streams. 

**To Reproduce**

Execute the following as worksheet file:

```scala
import akka.actor.ActorSystem
import akka._
import akka.stream.scaladsl._
import com.typesafe.config.ConfigFactory
import java.io.File
import akka.actor._
import scala.concurrent.Await
import scala.concurrent.duration.DurationInt

implicit val system: ActorSystem = ActorSystem("QuickStart")
val source: Source[Int, NotUsed] = Source(1 to 100)
val future = source.runWith(Sink.foreach(println))
Await.result(future, 3.seconds)
```

**Expected behavior**

Worksheet evaluates successfully.

**Screenshots**

![image](https://user-images.githubusercontent.com/3408009/108344156-e6bbdc00-7230-11eb-9037-d910006cb48c.png)

**Installation:**

- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.9.10
- Scala version: 3.0.0-M3

**Additional context**

Also attached a sample sbt solution to reproduce the issue.

[test.zip](https://github.com/scalameta/metals/files/6002106/test.zip)

tgodzik:
Thanks for reporting and the reproduction!

jvdp:
I think it's not related to Scala 3 nor Akka Streams. I was able to get a bit further by resolving the configuration manually:

```scala
import $dep.`com.typesafe.akka::akka-actor:2.6.13`
import akka.actor.Actor
import com.typesafe.config.ConfigFactory
import akka.actor.ActorSystem

val config = ConfigFactory
  .load()
  .withFallback(ConfigFactory.parseResources(Actor.getClass(), "/version.conf"))
  .withFallback(
    ConfigFactory.parseResources(Actor.getClass(), "/reference.conf")
  )
  .resolve()

implicit val system = ActorSystem("worksheet", config)
```

This results in another error:
```
java.lang.ClassNotFoundException: akka.event.DefaultLoggingFilter
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	at java.lang.Class.forName0(Native Method)
[...]

PR #2631, opened 2021-03-25
Use correct parent classloader for worksheets.

### LINKED ISSUE #2187: ClassDefNotFound error only in Worksheet
Trying to follow the Book of Doobie using a Worksheet, the following code produces a `ClassDefNotFound` exception when in a worksheet context. This same exact code works when input directly into ammonite's REPL or when compiled in a standard scala file via SBT.

**To Reproduce**

Create a worksheet (e.g. `book-of-doobie.worksheet.sc`) with the following code:

```scala
import $ivy.`org.tpolecat::doobie-core:0.9.0`
import $ivy.`org.tpolecat::doobie-postgres:0.9.0`
import $ivy.`org.tpolecat::doobie-specs2:0.9.0`

import cats._, cats.effect._, cats.data._, cats.implicits._
import doobie._
import doobie.implicits._

import doobie.util.ExecutionContexts
implicit val cs = IO.contextShift(ExecutionContexts.synchronous)

val xa = Transactor.fromDriverManager[IO](
  "org.postgresql.Driver",     // driver classname
  "jdbc:postgresql:world",     // connect URL (driver-specific)
  "postgres",                  // user
  "",                          // password
  Blocker.liftExecutionContext(ExecutionContexts.synchronous) // just for testing
)
```
```
[mdoc] [E] java.lang.NoClassDefFoundError: java/sql/Connection
  at doobie.util.transactor$Transactor$fromDriverManager$.apply(transactor.scala:377)                                                                          
  at repl.MdocSession$App.<init>(book-of-doobie.worksheet.sc:35)
  at repl.MdocSession$.app(book-of-doobie.worksheet.sc:3)
```

**Expected behavior**

Code should compile without error

**Installation:**

- Operating system: macOS
- Editor: nvim
- Metals version: v0.9.5 via coc-metals

ckipp01:
Instead of passing in `null` when we extend ClassLoader in our
MdocClassLoader this passes in the actual parent since on Java +9 when
using `null` that's the bootstrap classloader which doesn't contain
modules like java.sql.

Big thanks to @smarter pointing this out in https://github.com/lampepfl/dotty/pull/11658.

Fixes #2187

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetClassLoader.scala
+++ b/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetClassLoader.scala
@@ -8,7 +8,8 @@ package scala.meta.internal.worksheets
  * allowing the Metals server to call mdoc instances from different Scala
  * versions.
  */
-class MdocClassLoader(parent: ClassLoader) extends ClassLoader(null) {
+class MdocClassLoader(parent: ClassLoader)
+    extends ClassLoader(ClassLoader.getSystemClassLoader.getParent) {
   override def findClass(name: String): Class[_] = {
     val isShared =
       name.startsWith("mdoc.interfaces") || name.startsWith("coursierapi")
--- a/tests/unit/src/test/scala/tests/worksheets/WorksheetLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/worksheets/WorksheetLspSuite.scala
@@ -91,4 +91,27 @@ class WorksheetLspSuite extends tests.BaseWorksheetLspSuite(V.scala212) {
       )
     } yield ()
   }
+  // Ensure that on Java +9 that all modules are correctly loaded with the Mdoc
+  // classloader including things like the java.sql module.
+  // https://github.com/scalameta/metals/issues/2187
+  test("classloader") {
+    val path = "hi.worksheet.sc"
+    for {
+      _ <- server.initialize(
+        s"""
+           |/metals.json
+           |{
+           |  "a": {}
+           |}
+           |/${path}
+           |new java.sql.Date(100L)
+           |""".stripMargin
+      )
+      _ <- server.didOpen(path)
+      _ = assertNoDiff(
+        client.workspaceDecorations,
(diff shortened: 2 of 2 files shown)


metals/src/main/scala/scala/meta/internal/worksheets/WorksheetClassLoader.scala
```diff
@@ -8,7 +8,7 @@ package scala.meta.internal.worksheets
  * allowing the Metals server to call mdoc instances from different Scala
  * versions.
  */
-class MdocClassLoader(parent: ClassLoader) extends ClassLoader(null) {
+class MdocClassLoader(parent: ClassLoader) extends ClassLoader(parent) {
```
tgodzik, line 11: So I think this one was on purpose in order not to bring in the Scala 2.12 classes from Metals server. Maybe let's use exactly what was used in Dotty?
```suggestion
class MdocClassLoader(parent: ClassLoader) extends ClassLoader(ClassLoader.getSystemClassLoader.getParent) {
```

tgodzik (changes requested):
Thanks for looking into that! We are having some issues with Scala 2.13 and Scala 3 worksheets.

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetClassLoader.scala
ckipp01: Ahh yes, I was just looking at the failures now. Let me try that.

tgodzik:
Also, https://github.com/scalameta/metals/issues/2525 is probably related.

ckipp01:
> Also, #2525 is probably related.

Playing around with it a bit, this doesn't change that behavior at all, so I don't think it is.

tgodzik:
> > Also, #2525 is probably related.
> 
> Playing around with it a bit, this doesn't change that behavior at all, so I don't think it is.

The resources in akka are coming most likely from the classloader, which we change, so I thought it might be connected :thinking:  I will take a look at it.

tgodzik (approved):
LGTM!

PR #2635, opened 2021-03-27
Do not insert override twice

tgodzik:
Previously, if a symbol was already overriden, we would add the override flag twice. Now, we check if that flag is already containe din parent.


ckipp01 (approved):
Looks good. LGTM!

Issue #2558, opened 2021-02-26
"Insert type annotation" breaks compilation sometimes when higher-kinded types are involved


kubukoz:
**Describe the bug**

There are some cases where inserting the type annotation puts in something that doesn't quite compile. This is often related to kind-projector being used or some variance shenanigans.

**To Reproduce**

For now, I have a bunch of screenshots and links to the source code being manipulated:

---

- [x] Fixed?

![image](https://user-images.githubusercontent.com/894884/109310662-1d06e600-7845-11eb-8001-61e793d553a5.png)
![image](https://user-images.githubusercontent.com/894884/109310689-285a1180-7845-11eb-97ef-ba97af598acb.png)

https://github.com/pitgull/pitgull/blob/2e96c29483ba82b98a2a05a9842633e448e7f1a7/src/main/scala/io/pg/Main.scala#L72

---

- [x] Fixed?

This is probably the same case as above:

![image](https://user-images.githubusercontent.com/894884/109310810-4c1d5780-7845-11eb-947d-68fafcb4afce.png)

![image](https://user-images.githubusercontent.com/894884/109310828-52abcf00-7845-11eb-8097-c976b70bf492.png)

https://github.com/pitgull/pitgull/blob/2e96c29483ba82b98a2a05a9842633e448e7f1a7/src/main/scala/io/pg/Main.scala#L29 (you'll need to remove the existing type annotation)

---

- [x] Fixed?

![image](https://user-images.githubusercontent.com/894884/109310971-7bcc5f80-7845-11eb-94cd-1b988598df48.png)

![image](https://user-images.githubusercontent.com/894884/109310986-80911380-7845-11eb-91cb-6c022afe7809.png)

https://github.com/pitgull/pitgull/blob/2e96c29483ba82b98a2a05a9842633e448e7f1a7/src/main/scala/io/pg/Application.scala#L81

---

I'll be adding more issues if I find any in the future :)

**Expected behavior**

Newly inserted type annotations still compile

**Installation:**

- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.10.0

**Additional context**

none

PR #2636, opened 2021-03-27
Make sure higher kinded functions in signatures are inferred properly

tgodzik:
Previously for a higher kinded type like:

```scala
TypeRef(
  pre = <noprefix>,
  sym = class Resource,
  args = List(
    PolyType(
      typeParams = List(type x),
      resultType = TypeRef(
        pre = <noprefix>,
        sym = type F,
        args = List(TypeRef(pre = <noprefix>, sym = type x, args = List()))
      )
    ),
    TypeRef(pre = <noprefix>, sym = class Unit, args = List())
  )
)
```

we would print:

```scala
Resource[[x]F[x],Unit]
```

Which is nothing that can actually work in the code. Instead, it should just be `Resource[F, Unit]` as the type constructors can't really appear in signatures (I think)

Fixes https://github.com/scalameta/metals/issues/2558

I am not 100% sure of this solution, but I this seems to work now. I anyone has a better idea about `PolyType` or higher kinded types do let me know. As it stands, this should not break anything, but can fix things.

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
@@ -325,8 +325,15 @@ class MetalsGlobal(
             quantified.map(sym => sym.setInfo(loop(sym.info, None))),
             loop(underlying, None)
           )
-        case PolyType(tparams, resultType) =>
-          PolyType(tparams, resultType.map(t => loop(t, None)))
+        case PolyType(typeParams, resultType) =>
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):
Honestly I'm not really sure if there is a better way to do this or if this misses something etc, but this looks good to me from what I can tell 👍🏼 LGTM!

dos65 (approved):
LGTM :+1:

tanishiking (approved):
LGTM! :tada:

kubukoz:
🥰

PR #2638, opened 2021-03-29
Make sure that universal types deco are printed properly

tgodzik:
Follow up to https://github.com/scalameta/metals/pull/2636

This make type decorations consistent with what can be expected to pop up in the code

### DIFF
--- a/tests/unit/src/test/scala/tests/decorations/SyntheticDecorationsLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/decorations/SyntheticDecorationsLspSuite.scala
@@ -575,4 +575,85 @@ class SyntheticDecorationsLspSuite extends BaseLspSuite("implicits") {
       )
     } yield ()
   }
+
+  test("inferred-type-hkt") {
+    for {
+      _ <- server.initialize(
+        s"""
+           |/metals.json
+           |{
+           |  "a": {
+           |    "libraryDependencies": [
+           |        "org.typelevel::cats-effect:2.4.0"
+           |    ]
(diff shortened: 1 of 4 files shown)


tests/unit/src/test/scala/tests/decorations/SyntheticDecorationsLspSuite.scala
```diff
@@ -575,4 +575,85 @@ class SyntheticDecorationsLspSuite extends BaseLspSuite("implicits") {
...
+           |import cats.effect.Timer
+           |
+           |object Main2 extends IOApp {
+           |
+           |  trait Logger[T[_]]
+           |
+           |  def mkLogger[F[_]: ConcurrentEffect: Timer: ContextShift]: Resource[F, Logger[F]] = ???
+           |
+           |  // The actual tested method:
+           |  def serve[F[_]: ConcurrentEffect: ContextShift: Timer: Parallel]: Resource[F, Unit]() =
```
dos65, line 648: Hmm. Here is something wrong with the position of type decoration.
The type signature here should go after `()`

tgodzik: Damn, I haven't noticed that :/

tgodzik: Fixed! The problem is that we were looking first for `]` then for `)`, it should have been the other way around.

dos65 (approved):

PR #2639, opened 2021-03-29
Add support for Scala 3.0.0-RC2

tgodzik:


ckipp01 (approved):
This looks great @tgodzik. LGTM 🚀

tgodzik:
It turns out I can work around the description issue, so that we still get sensible completions. The issue was that overloaded symbols are all returned in a single completion, which doesn't really make sense from the Metals perspective. We need to somehow indicate the different signatures that are possible.

tgodzik:
@tanishiking this might cause some conflict with your PR, I will take at look later on.

tanishiking:
@tgodzik Thank you for letting me know, I'll also try to rebase  https://github.com/scalameta/metals/pull/2637 onto `main` later :)

PR #2642, opened 2021-03-31
A few changes to docs including a User Configuration section.

ckipp01:
This makes a few changes to the docs. Noticeably it adds two things:
  1. It adds a section for user configuration. I've gotten the request a
     few times to add something like this and have gotten comments in
     the past that it's not always clear what configuration options are
     available. We did have these in the site, but they were buried
     under integrating a new editor. I didn't remove them from there as
     they are still useful to someone creating a new extension, but made
     a more user-focused intro and added them to the forefront.
  2. Just add in an example of scalac imports.

Let me know what you think of this.

### DIFF
--- a/metals-docs/src/main/scala/docs/UserConfigurationModifier.scala
+++ b/metals-docs/src/main/scala/docs/UserConfigurationModifier.scala
@@ -14,50 +14,25 @@ class UserConfigurationModifier extends StringModifier {
       code: Input,
       reporter: Reporter
   ): String = {
-    info match {
-      case "lsp-config-default" =>
-        UserConfiguration.options
-          .map { option =>
-            s"""
-               |### ${option.title}
-               |
-               |${option.description}
-               |
-               |**Default**: ${option.default}
-               |
-               |**Example**:
-               |```json
-               |{
-               |  "metals": {
-               |    "${option.key}": ${option.example}
-               |  }
-               |}
-               |```
-               |""".stripMargin
-          }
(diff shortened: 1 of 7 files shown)


metals-docs/src/main/scala/docs/UserConfigurationModifier.scala
```diff
@@ -14,50 +14,25 @@ class UserConfigurationModifier extends StringModifier {
...
-               |  "metals.${option.camelCaseKey}": ${option.example}
-               |}
-               |```
-               |""".stripMargin
-          }
-          .mkString("\n")
-      case els =>
-        reporter.error(s"unknown user-config '$els'")
-        ""
-    }
```
ckipp01, line 61: So I wasn't fully sure why we actually were displying everything kebab case since looking at the various extensions, everything sends it in via camelcase. So I think it might be better to just align and show them that way.

tgodzik (approved):
LGTM!

metals-docs/src/main/scala/docs/UserConfigurationModifier.scala
gabro: IIRC it used to accept both syntaxes since we originally parsed these from different sources, such as system properties.

ckipp01: Totally, it actually still does. I didn't realize before making this change, but you can still pass in all the user configurations as properties.

Issue #1196, opened 2019-12-17
OnDemandSymbolIndex doesn't distinguish between build targets


mzarnowski:
Currently, despite having the same symbol (let's say class `foo.bar.Baz`) defined in two modules, there would be only one mapping `foo.bar.Baz -> SourceFile` in the OnDemandSymbolIndex and it would be the one being indexed last, so some race conditions could apply making some of the other features nondeterministic.

kubukoz:
I'm not sure if this is 100% the same problem, but I noticed I see references to a symbol defined in module A in module B (if there's the same symbol in B), even if they don't share a classpath.

Renames also get applied on all the references across modules, which can be quite a problem for multi-module builds.

tgodzik:
@kubukoz we assume that a symbol is unique within a workspace so that is kind of related problem yeah. In case of references we should most likely filter out symbols coming from unrelated build targets and that will fix renames as well.

tgodzik:
Followed up with https://github.com/scalameta/metals/issues/2265

tgodzik:
Fixed in https://github.com/scalameta/metals/pull/2668

PR #2668, opened 2021-04-01
SymbolIndex: multiply symbol locations

### LINKED ISSUE #2265: Found references for symbols in unrelated build targets.
**Describe the bug**

In case we have two unrelated build targets with the same symbol name, references will find symbols from both despite those being essentially separate symbols. It will also rename those symbols.

**To Reproduce**

1. Create two modules A and B
2. Create class `example.First` in A and B.
3. Find references to A.
4. See references from both modules

**Expected behavior**

Only references to A or B are shown.

**Installation:**

- Metals version: v0.9.6

**Additional context**

We might filter out files based on the build target. If we are in A we should go over all related build targets (upstream and downstream) and with that least filter out any source belonging to a target not in set.

**Search terms**

rename references multi module

dos65:
Previously, the index has been storing only one Path per symbol.
It didn't work correctly in cases if there were more than one symbol definition across the project that affected all features that are based on querying index like `go-to-definition` and  `references`.

In this PR this thing was reworked - now index stores all locations for symbol and in most places correct location is selected by filtering them by build targets. In particular, it fixes #2265.

Also, the way how dependency sources are extracted into `.readonly` was changed. This fixes navigation when there are several versions of the same library across build targets.
Now, if the source comes from source-jar it will be saved in: `.metals/readonly/dependencies/${source-jar-name}`.
For example for symbol `scala.Predef`:
- old: `.metals/readonly/scala/Predef.scala`
- now: `.metals/readonly/dependencies/scala-library_2.13.3-sources.jar/scala/Predef.scala`

Notes: the only features where wrong navigation is still possible are `StackTraceAnalyser` and `debugging`. 
They using LSP json parameters and they are slightly different from other places where `definitionProvider` is used. I think it would be ok to create issues and fix them later.


metals/src/main/scala/scala/meta/internal/metals/JarTopLevels.scala
```diff
@@ -29,10 +27,10 @@ final class JarTopLevels(conn: () => Connection) {
    */
   def getTopLevels(
       path: AbsolutePath
-  ): Option[TrieMap[String, AbsolutePath]] =
+  ): Option[List[(String, AbsolutePath)]] =
     try {
       val fs = PlatformFileIO.newJarFileSystem(path, create = false)
-      val toplevels = TrieMap[String, AbsolutePath]()
+      val toplevels = List.newBuilder[(String, AbsolutePath)]
```
ckipp01, line 33: Out of curiosity, why the change from the `TrieMap` to `List`? Normally I prefer to have a Map for this over a list of tuples unless there is a good reason not to.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -275,15 +275,40 @@ object MetalsEnrichments
...
+        )
+      } else {
+        None
+      }
+    }
+
     /**
      * Writes zip file contents to disk under $workspace/.metals/readonly.
+     *
+     * In case if path is jar then target directory is $workspace/.metals/readonly/$jarName
```
ckipp01, line 299: So I'm a bit confused by this. So what is the difference between:

```
metals/readonly/dependencies/scala-library_2.13.3-sources.jar/scala/Predef.scala
```
and
```
metals/readonly/scala-library_2.13.3-sources.jar/scala/Predef.scala
```

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -2300,12 +2305,15 @@ class MetalsLanguageServer(
       item <- dependencySources.getItems.asScala
       scalaTarget <- buildTargets.scalaTarget(item.getTarget)
       sourceUri <- Option(item.getSources).toList.flatMap(_.asScala)
+      path = sourceUri.toAbsolutePath
+      _ = buildTargets.addDependencySource(path, item.getTarget)
+      _ = jdkSources.foreach(jdk =>
+        buildTargets.addDependencySource(jdk, item.getTarget)
```
ckipp01, line 2311: nit
```suggestion
      _ = jdkSources.foreach(source =>
        buildTargets.addDependencySource(source, item.getTarget)
```
Calling that `jdk` may be a bit confusing.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -319,20 +322,37 @@ final class BuildTargets(
...
+          .toRelative(workspace.resolve(Directories.dependencies))
+          .toNIO
+          .getName(0)
+          .filename
+
+      inverseDependencySources
+        .find { case (path, _) =>
+          path.filename == jarName
+        }
+        .flatMap { case (_, ids) => ids.headOption }
```
ckipp01, line 343: I must be losing my mind or something but I swear I already commented this, but now it's gone 🤔 anways, probably just style choice, but you could also do

```scala
        .collectFirst {
          case (path, ids) if path.filename == jarName && ids.nonEmpty =>
            ids.head
        }
```
To avoid iterating twice.

I wish we had a `nonEmptySet` we could use here since realistically `ids` will never be empty.

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -254,22 +265,66 @@ class DestinationProvider(
...
+
+    val definitions = index.definitions(Symbol(symbol)).filter(_.path.exists)
+    allowedBuildTargets match {
+      case None => definitions.headOption
+      case Some(targets) =>
+        val asSet = targets.toSet
+        val matched = definitions.find { defn =>
+          sourceBuildTargets(defn.path).exists(id => asSet.contains(id))
+        }
+        // Fallback on any definition - it needs for worksheets
```
ckipp01, line 286:
```suggestion
        // Fallback to any definition - it's needed for worksheets
```
?

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -254,22 +265,66 @@ class DestinationProvider(
...
+  ): Option[SymbolDefinition] = {
+    val allowedBuildTargets =
+      for {
+        src <- source
+        id <- buildTargets.inverseSources(src)
+      } yield buildTargets.buildTargetTransitiveDependencies(id)
+
+    val definitions = index.definitions(Symbol(symbol)).filter(_.path.exists)
+    allowedBuildTargets match {
+      case None => definitions.headOption
```
ckipp01, line 280: I see this `headOption` in a lot of places. How do we ensure that the `headOption` is actually the desired one?

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -254,22 +265,66 @@ class DestinationProvider(
...
+  private def sourceBuildTargets(
+      source: AbsolutePath
+  ): List[BuildTargetIdentifier] = {
+    def trivialSource: Option[List[BuildTargetIdentifier]] =
+      Option(buildTargets.sourceBuildTargets(source).toList).filter(_.nonEmpty)
+
+    def dependencySource: Option[List[BuildTargetIdentifier]] = {
+      source.jarPath
+        .map(buildTargets.inverseDependencySource)
+        .filter(_.nonEmpty)
```
ckipp01, line 302: Nit, Map + Filter = Collect

metals/src/main/scala/scala/meta/internal/metals/JarTopLevels.scala
ckipp01, line 33: Any reason to change from using a Map to a list? Typically I prefer Maps over List of tuples for this type of thing.

tests/slow/src/test/scala/tests/sbt/SbtBloopLspSuite.scala
```diff
@@ -453,7 +453,7 @@ class SbtBloopLspSuite
       _ <- assertDefinitionAtLocation(
         "build.sbt",
         "sc@@alaVersion := \"2.12.11\"",
-        ".metals/readonly/sbt/Keys.scala",
+        "sbt/Keys.scala",
```
ckipp01, line 456: Shouldn't this still be in `.metals/readonly`?

tests/unit/src/main/scala/tests/BaseAmmoniteSuite.scala
```diff
@@ -69,21 +62,21 @@ abstract class BaseAmmoniteSuite(scalaVersion: String)
       _ <- assertDefinitionAtLocation(
         "main.sc",
         "foo.as@@Json.noSpaces",
-        ".metals/readonly/io/circe/syntax/package.scala"
+        "io/circe/syntax/package.scala"
```
ckipp01, line 65: Again are all these right no longer being in `.metals/readonly`? Or are they all now relative? If it's easy to switch, having the full path here might be clearer.

ckipp01 (changes requested):
Great job on this @dos65! I have a couple questions and a couple nits.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -319,20 +322,37 @@ final class BuildTargets(
     }
   }
 
+  /**
+   *  Match build target by source jar name if source was saved in `.metals/readonly/dependecies/${source-jar-name}`
+   */
+  private def inferByJarDirectoryName(
+      source: AbsolutePath
+  ): Option[BuildTargetIdentifier] = {
+    if (source.isInDependenciesDirectory(workspace)) {
```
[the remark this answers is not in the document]
dos65: I reworked this a little bit to avoid double-checking the similar clause.

ckipp01:
If I'm not mistaken, this should also fix https://github.com/scalameta/metals/issues/1196 right?

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -50,7 +49,7 @@ final class BuildTargets(
   private val buildTargetSources =
     TrieMap.empty[BuildTargetIdentifier, util.Set[AbsolutePath]]
   private val inverseDependencySources =
-    TrieMap.empty[AbsolutePath, BuildTargetIdentifier]
+    TrieMap.empty[AbsolutePath, Set[BuildTargetIdentifier]]
```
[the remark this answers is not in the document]
dos65: I wish the same but as it used in the index too it's required to implement it without additional mem-overhead.
I've seen such implementation in [cats](https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/data/NonEmptySet.scala) but I don't know if it worths to copy-paste it for these two cases.

dos65:
> If I'm not mistaken, this should also fix #1196 right?
Yep!

metals/src/main/scala/scala/meta/internal/metals/JarTopLevels.scala
dos65: `scala.collection.concurrent.TrieMap` is too complex type for this method. It was used because `OnDemandSymbolIndex.toplevels` had this signature and there was this tricky method: https://github.com/scalameta/metals/blob/03704c17b5397bd8feb4771d0bedf9a6dfe0a660/mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala#L96-L104

But as `toplevels` signature was changed to `TrieMap[String, Set[AbsolutePath]]` it became inconvenient to use `addSourceJarTopLevels`  as they were so I changed this part and replaced `TrieMap` here on smth more suitable.

It might be `Map` but it anyway used only as a `List`.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
dos65: I forgot to add `/dependencies` in the comment. Thx!

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
dos65: Thx! Replaced by `collectFirst`

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
dos65: It might be non desired one but at least it returns smth and works in a similar way to the current implementation.

There were two places where I wasn't sure how to pass `allowedBuildTargets` properly into this method so I decided to fix them later.

tests/slow/src/test/scala/tests/sbt/SbtBloopLspSuite.scala
dos65: They still are in `.metals/readonly` but the path became much longer - `.metals/readonly/dependencies/sbt-${version}-sources.jar/sbt/Keys.scala`.
Honestly, it was easier to fix tests by leaving these paths without full prefix than investigating the right one per every test. Anyway, there isn't a lot of meaning in seeing them.

Also, I think it might be annoying to update jar names in these paths in case if some dependency version will be changed in tests

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -461,13 +471,14 @@ final class BuildTargets(
...
+    val acc = inverseDependencySources.getOrElse(sourcesJar, Set.empty)
+    inverseDependencySources(sourcesJar) = acc + target
   }
 
   def inverseDependencySource(
       sourceJar: AbsolutePath
-  ): Option[BuildTargetIdentifier] = {
-    inverseDependencySources.get(sourceJar)
+  ): List[BuildTargetIdentifier] = {
+    inverseDependencySources.get(sourceJar).map(_.toList).getOrElse(List.empty)
```
ckipp01, line 481: Not a huge deal, but we could just keep a `Set` here.

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
ckipp01: Sounds good. For those two places, make sure to have TODOs or issues created for them then so we don't forget about them.

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -254,22 +265,66 @@ class DestinationProvider(
...
+        // aren't registered in buildTargets
+        matched.orElse(definitions.headOption)
+    }
+  }
+
+  private def sourceBuildTargets(
+      source: AbsolutePath
+  ): List[BuildTargetIdentifier] = {
+    def trivialSource: Option[List[BuildTargetIdentifier]] =
+      Option(buildTargets.sourceBuildTargets(source).toList).filter(_.nonEmpty)
```
ckipp01, line 297: Couldn't this be simplified to just `List`? Whenever I see `Option[List]` I always think that unless there is a real difference between `None` and `Some(List.empty)`.

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -254,22 +265,66 @@ class DestinationProvider(
...
+
+    def dependencySource: Option[List[BuildTargetIdentifier]] = {
```
ckipp01, line 299: Same question as above, is the `Option[List` necessary?

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -275,15 +276,41 @@ object MetalsEnrichments
...
 
+    def toRelativeInside(prefix: AbsolutePath): Option[RelativePath] = {
+      val relative = path.toRelative(prefix)
+      if (relative.toNIO.getName(0).filename != "..") Some(relative)
+      else None
+    }
+
+    def jarPath: Option[AbsolutePath] = {
+      val filesystem = path.toNIO.getFileSystem()
+      if (filesystem.provider().getScheme().equals("jar")) {
```
ckipp01, line 287: Do we also need to consider srcjars here? https://github.com/scalameta/metals/pull/2710

ckipp01 (changes requested):
Sorry this took me a while to get back to! Just a couple nits/questions. But apart from that I'll use this snapshot today to ensure nothing wonky happens and we should be able to merge this soon!

metals/src/main/scala/scala/meta/internal/metals/JarTopLevels.scala
ckipp01: Alright, sounds good and thanks for the explanation.

tests/slow/src/test/scala/tests/sbt/SbtBloopLspSuite.scala
ckipp01: Yea good point about changing these in the future. 👍🏼

tests/unit/src/main/scala/tests/BaseAmmoniteSuite.scala
ckipp01: Seen your reason up above. Just ignore this.

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
dos65: Yep, I agree that in most cases `Option[List[A]]` should be reduced to `List[A]`.

However, here it's used only in inner functions just to take benefits from `orElse` combinator in usage below - https://github.com/scalameta/metals/blob/a82a14240bd55284ba220650fe5a70c4198b5cde/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala#L305-L308
Otherwise, I had to inекщвгсу smth like `orElseIfNil` for `List` for this small part of code.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
dos65: Good catch. Thx!

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
ckipp01: Alright sounds good 👍🏼

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
ckipp01: Ah ok, I wasn't 100% sure, that's why I brought it up.

ckipp01 (approved):
Really great job @dos65. 🚀 

This is a LGTM from me, but I'm not sure if @tgodzik wants to take a quick look over first.

tgodzik:
I will take a look tomorrow! I am going slowly through my backlog of reviews 😅

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
```diff
@@ -51,90 +51,127 @@ final class OnDemandSymbolIndex(
...
   ): Unit = {
     if (source.isAmmoniteScript || !isTrivialToplevelSymbol(path, toplevel)) {
-      toplevels(toplevel) = source
+      val acc = toplevels.getOrElse(toplevel, Set.empty)
+      toplevels(toplevel) = acc + source
     }
   }
 
-  private def tryRun(thunk: => Unit): Unit =
+  private def addToplevelSymbol0(symbol: String, source: AbsolutePath): Unit = {
```
tgodzik, line 167: Can we inline this if it's only used in one palce?

tests/unit/src/test/scala/tests/DefinitionLspSuite.scala
```diff
@@ -326,4 +326,67 @@ class DefinitionLspSuite extends BaseLspSuite("definition") {
     } yield ()
   }
 
+  test("clashing-references") {
```
tgodzik, line 329: You should also uncomment https://github.com/scalameta/metals/blob/03704c17b5397bd8feb4771d0bedf9a6dfe0a660/tests/unit/src/test/scala/tests/debug/BreakpointDapSuite.scala#L633
to see if it's working correctly now.

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
```diff
@@ -51,90 +51,127 @@ final class OnDemandSymbolIndex(
...
-          onError(InvalidJarException(jar, e))
       }
-    }
+    )
 
   // Used to add cached toplevel symbols to index
-  def addSourceJarTopLevels(
+  def addIndexedSourceJar(
       jar: AbsolutePath,
-      loadTopLevels: () => TrieMap[String, AbsolutePath]
```
tgodzik, line 98: I think this was done so that `loadTopLevels` is not calculated again if a jar was already added. Doesn't your change make it evaluate always? We could maybe make it `=> symbols: List[(String, AbsolutePath)]` and make it lazy val at invocation at the `None` case? It would serve the same purpose.

tgodzik (changes requested):
That is a lot of great changes and I calculate that we can solve 3 issues with this PR. I left some comments, I don't think anything major popped up. Good job!

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -275,15 +276,41 @@ object MetalsEnrichments
         workspace.resolve(Directories.readonly).toNIO
       )
 
+    def toRelativeInside(prefix: AbsolutePath): Option[RelativePath] = {
+      val relative = path.toRelative(prefix)
+      if (relative.toNIO.getName(0).filename != "..") Some(relative)
```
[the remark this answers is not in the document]
dos65: Honestly, I'm not sure about Windows.
In Linux, it can't.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -2300,12 +2305,15 @@ class MetalsLanguageServer(
       item <- dependencySources.getItems.asScala
       scalaTarget <- buildTargets.scalaTarget(item.getTarget)
       sourceUri <- Option(item.getSources).toList.flatMap(_.asScala)
+      path = sourceUri.toAbsolutePath
+      _ = buildTargets.addDependencySource(path, item.getTarget)
```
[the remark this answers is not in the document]
dos65: It was needed to correctly call `buildTargets.addDependencySource(path, item.getTarget)`.
Otherwise, it registers jar only once for the first build target where it appeared.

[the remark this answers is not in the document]
tgodzik: Ach alright. This should be fine.

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
```diff
@@ -51,90 +51,127 @@ final class OnDemandSymbolIndex(
   private val onErrorOption = onError.andThen(_ => None)
 
   override def definition(symbol: Symbol): Option[SymbolDefinition] = {
-    try findSymbolDefinition(symbol, symbol)
+    try findSymbolDefinition(symbol, symbol).headOption
```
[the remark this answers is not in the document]
dos65: Unfortunately, there are still a lot of places where it's used.

Initially, I thought that this fix will take less time than I already spend on it.
So, I would prefer to turn it into an issue - a kind of the second interaction of fixing wrong definitions. 

I think I will return to it after completing scala 3 related things.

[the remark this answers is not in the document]
tgodzik: Sure! Let's create a follow up to make sure that all places are properly handled.

mtags/src/main/scala-2/scala/meta/internal/mtags/OnDemandSymbolIndex.scala
dos65: It slightly changes behavior but jar-indexing definitely happens only once as it should
The difference in case if a jar was already added:
 - previously: absolutely nothing happens
 - now: for the first jar occurrence during the import it will request toplevels once from H2 and they will be ignored in this function.

I think it's how it should work.
Also, I've just found that we don't recreate index during build reloading that might lead to stale symbols so we can't rely on `DefinitionIndex` state.

tgodzik: Makes sense! Let's leave it as is.

metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathProvider.scala
```diff
@@ -29,7 +29,7 @@ final class SourcePathProvider(
     val symbolBase = if (base.contains("/")) base else "_empty_/" + base
     val symbols = for {
       symbol <- Set(symbolBase + ".", symbolBase + "#").toIterator
-      definition <- definitionProvider.fromSymbol(symbol).asScala
+      definition <- definitionProvider.fromSymbol(symbol, None).asScala
```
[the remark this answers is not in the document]
dos65: There is not enough data currently - only stacktrace as a string. I will mention it in the issue for step 2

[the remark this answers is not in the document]
dos65: Fixed

metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
```diff
@@ -68,7 +68,7 @@ class StacktraceAnalyzer(
 
   def fileLocationFromLine(line: String): Option[l.Location] = {
     def findLocationForSymbol(s: String): Option[Location] =
-      definitionProvider.fromSymbol(s).asScala.headOption
+      definitionProvider.fromSymbol(s, None).asScala.headOption
```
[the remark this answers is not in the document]
dos65: There is not enough data currently - only stacktrace as a string. I will mention it in the issue for step 2

tests/unit/src/test/scala/tests/DefinitionLspSuite.scala
dos65: Thx! Now it works.

tgodzik (approved):

PR #2669, opened 2021-04-02
Update org.eclipse.lsp4j to 0.12.0

tgodzik:
CompletionItem.textEdit is no longer just `TextEdit`, but now is `Either[TextEdit, InsertReplaceEdit]` - this was the main issue with migration.

Supersedes https://github.com/scalameta/metals/pull/2657 

Actually `0.12.0` just got released, so updating to that.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -208,7 +208,7 @@ lazy val V = new {
   val scalafmt = "2.7.4"
   val munit = "0.7.23"
   val scalafix = "0.9.27"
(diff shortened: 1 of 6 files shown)


dos65 (approved):

ckipp01 (approved):
Looks good. Sometimes I wish they made it super clear in the LSP docs about when a specific one is preferred. Either way, I think we are fine staying as is. LGTM!

Issue #2641, opened 2021-03-31
company-lsp no longer available


tjarvstrand:
**Describe the bug**

Initialization error when starting Emacs after following the instructions here: https://scalameta.org/metals/docs/editors/emacs.html:

```
Error (use-package): Failed to install company-lsp: Package ‘company-lsp-’ is unavailable
```

It seems the package has been removed from MELPA: https://github.com/melpa/melpa/issues/7461

I realize it may not be within the scope of this project to specify what intellisense solution people should use but it would be helpful. In any case, the reference to `company-lsp` should probably be removed from the documentation.

**To Reproduce**

Follow the instructions mentioned above.

**Search terms**

emacs, melpa, company, lsp

dos65:
Thanks for reporting!

Recently, @strobe added a [comment](https://github.com/strobe/metals/blame/9e4a2697565d5bef8c2dba67a303ac168e88d4ee/docs/editors/emacs.md#L96-L97) to this package with a link on https://github.com/emacs-lsp/lsp-mode/pull/1983.

I don't know if it would be correct to remove it completely from doc so I have questions to emacs users:
- Is possible that someone still uses an older version of `lsp-mode` that requires `company-lsp`?
- I see that in  https://github.com/emacs-lsp/lsp-mode/pull/1983 it's recommended to use `company-capf`. Does it require any additional configuration that should be mention in documentation?

tototoshi:
I haven't tried `company-lsp`, but from the discussion on MELPA, it doesn't seem to work anymore.

If company-mode is installed, `company-capf` is automatically selected, but it would be better to set the variable `lsp-completion-provider` to use `company-capf` explicitly.

```el
(use-package company
  :hook (scala-mode . company-mode)
  ;; Optionally enable completion-as-you-type behavior.
  :config
  (setq lsp-completion-provider :capf))
```

https://github.com/emacs-lsp/lsp-mode/blob/57dccd67c289d62323b0656edffd10c2caba4b29/lsp-completion.el#L30-L36

dos65:
Closed by #2672

PR #2672, opened 2021-04-02
Update the documentation to recommend using company-capf

tototoshi:
https://github.com/scalameta/metals/issues/2641


tgodzik (approved):
Looks great! Thanks for the contribution! Can someone experienced in emacs can confirm?

SergiyKolesnikov:
I've had the same problem as described in #2641 and ended up with basically the same solution as proposed in this PR. It works as expected for me.

ckipp01:
> I've had the same problem as described in #2641 and ended up with basically the same solution as proposed in this PR. It works as expected for me.

Cool, thanks for verifying! I'll go ahead and merge this in then.

PR #2674, opened 2021-04-02
Fix issues with old typesafe config in the worksheets

tgodzik:
It seems that somehow when using akka in worksheets the actual typesafe config lib that is being loaded is coming from mdoc, not from akka. Which means that akka will not work and throw ClassNotFound exceptions since mdoc has 1.2.1 version on the classpath. We should also fix it in mdoc, but I have currently no idea how to fix it (we should be using the correct classpath currently), so I decided to make the classpath a bit thiner as worksheets do not need everything that mdoc brings. Especially, as there are no config options, so no need for the typesafe config.

Fixes https://github.com/scalameta/metals/issues/2525 hopefully

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
@@ -120,7 +120,21 @@ object Embedded {
     val jars = downloadMdoc(scalaVersion, scalaBinaryVersion, resolutionParams)
 
     val parent = new MdocClassLoader(this.getClass.getClassLoader)
-    val urls = jars.iterator.map(_.toUri().toURL()).toArray
+
+    // Full mdoc classpath seems to be causing some issue with akka
+    // We want to keep a minimal set of jars needed here
+    val runtimeClasspath = jars.filter { path =>
+      val pathString = path.toString
+      pathString.contains("scala-lang") ||
+      pathString.contains("fansi") ||
+      pathString.contains("pprint") ||
+      pathString.contains("sourcecode") ||
+      pathString.contains("mdoc") ||
+      pathString.contains("scalameta") ||
+      pathString.contains("metaconfig") ||
+      pathString.contains("diffutils")
+    }
+    val urls = runtimeClasspath.iterator.map(_.toUri().toURL()).toArray
     new URLClassLoader(urls, parent)
   }
 
--- a/tests/unit/src/test/scala/tests/worksheets/WorksheetLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/worksheets/WorksheetLspSuite.scala
@@ -114,4 +114,56 @@ class WorksheetLspSuite extends tests.BaseWorksheetLspSuite(V.scala212) {
       )
     } yield ()
   }
+
+  test("akka") {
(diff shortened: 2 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -120,7 +120,21 @@ object Embedded {
...
+      pathString.contains("scala-lang") ||
+      pathString.contains("fansi") ||
+      pathString.contains("pprint") ||
+      pathString.contains("sourcecode") ||
+      pathString.contains("mdoc") ||
+      pathString.contains("scalameta") ||
+      pathString.contains("metaconfig") ||
+      pathString.contains("diffutils")
+    }
+    val urls = runtimeClasspath.iterator.map(_.toUri().toURL()).toArray
```
ckipp01, line 137: So, I tested this and it seems fine, however I'm worried about his in the long term. Mainly because it seems that we are both cherry picking stuff from the classpath in mdoc and also in here. I feel it makes it quite difficult to understand what ultimately ends up there, and where something may have been filtered off. Just trying to understand this.

We download mdoc in this file. That's what `jars` are here for the runtime classpath. I guess I don't get why we would filter that seeing that I was _assuming_ that the evaluation that mdoc is doing has it's own separate classpath. I get even further confused because looking at the `MarkdownBuilder` I also see:

https://github.com/scalameta/mdoc/blob/e0455f655f0c29a035a4ef703c21ad2d24902f89/mdoc/src/main/scala/mdoc/internal/markdown/MarkdownBuilder.scala#L80-L98

So It looks that when we are actually creating that we are already doing some filtering 🤔 

I just don't get why this is happening and that is bleeding in here. I'm probably speaking to the choir. I'll also try to spend a little time on the mdoc side and figure this out because it baffles me.

Either way for now, LGTM! But ultimately I think we should remove this in the future when the linked ticket you have attached in mdoc gets closed.

ckipp01 (approved):

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
tgodzik: This is indeed a bit hacky workaround, because I was unable to figure out why the classloader would load up something from a wrong jar. Especially that now with the recent changes, everything should work alright. I think this has something to do with how classloaders operate. So if we classload a class from a certain jar using a certain classloaders then everything else connected to it will use the same classloader. So I am thinking this is why we are actually using mdoc's classloader to load up typesafe config. But I haven't yet dug deep into how they work, this is more like my impression.

This change makes it impossible to load up that jar, which is why it's working I think. 

> I just don't get why this is happening and that is bleeding in here. I'm probably speaking to the choir. I'll also try to spend a little time on the mdoc side and figure this out because it baffles me.

Haha, yes I am as confused as you are :sweat_smile: But this seems to work and we shouldn't get similar issues anymore.

tgodzik: I think the main issue stems from the fact that the worksheet classloader has the main mdoc classpath as parent, so it is not as nicely isolated as it should.

PR #2677, opened 2021-04-04
Get behind unified slash syntax.

ckipp01:
Seeing that the old style is now deprecated in sbt moving forward, I
figured it'd be best to just align everything behind the unified slash
syntax right away. I was playing around with the scalafix rule for this
that was created, and thought Metals was a good test.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -139,8 +139,8 @@ inThisBuild(
     resolvers += Resolver.sonatypeRepo("public"),
     dependencyOverrides += V.guava,
     // faster publishLocal:
-    publishArtifact.in(packageDoc) := sys.env.contains("CI"),
-    publishArtifact.in(packageSrc) := sys.env.contains("CI"),
+    packageDoc / publishArtifact := sys.env.contains("CI"),
+    packageSrc / publishArtifact := sys.env.contains("CI"),
     // forking options
     javaOptions += {
       import scala.collection.JavaConverters._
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -282,7 +282,7 @@ lazy val interfaces = project
       V.lsp4j
     ),
     crossVersion := CrossVersion.disabled,
-    javacOptions in (Compile / doc) ++= List(
+    ((Compile / doc) / javacOptions) ++= List(
```
gabro, line 285: just out of curiosity but are the double parens here needed? Or is it the result of the automatic rewrite?

gabro (approved):
Nice! 🚀

build.sbt
ckipp01: So I actually had the same question about pretty much all of the `()` and I'm pretty sure it's done for consistency across the board. Even the simple ones like:

```scala
(Compile / unmanagedSourceDirectories) ++= thing
```

Could just be:

```scala
Compile / unmanagedSourceDirectories ++ thing
```

So it seems that everything is just wrapped to avoid confusion. That's my assumption anyway. It was done this way with the rewrite.

ckipp01: So I was wrong with my assumption. Do we have a preference on this? Do we want things wrapped, or would we prefer them not? 1, 2, 3, what do you prefer?

1.
```scala
((Compile / doc) / scalacOptions)
```

2.
```scala
(Compile / doc) / scalacOptions
```

3.
```scala
Compile / doc / scalacOptions
```

ckipp01: I'm inclined to go style 3 so I'll push that up and wait to see if anyone complains. Thanks for pointing this out 👍🏼

tgodzik (approved):
LGTM! I did not even realise it was deprecated :O

PR #2678, opened 2021-04-04
Mtags: fix Scala3 parsing for Windows.

dos65:
`Paths.get("jar:file:...")` doesn't work on Windows.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/mtags/CommonMtagsEnrichments.scala
+++ b/mtags/src/main/scala/scala/meta/internal/mtags/CommonMtagsEnrichments.scala
@@ -1,5 +1,6 @@
 package scala.meta.internal.mtags
 
+import java.net.URI
 import java.nio.charset.StandardCharsets
 import java.nio.file.Files
 import java.nio.file.Path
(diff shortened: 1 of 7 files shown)


mtags/src/main/scala/scala/meta/internal/mtags/CommonMtagsEnrichments.scala
```diff
@@ -269,6 +271,17 @@ trait CommonMtagsEnrichments {
...
   }
 
+  implicit class XtensionInputVirtual(input: Input.VirtualFile) {
+    def filename: String = {
+      Try {
+        val uri = URI.create(input.path)
+        Paths.get(uri).filename
+      }.getOrElse {
+        Paths.get(input.path).filename
+      }
```
dos65, line 281: The problem is that there is no strict format of what might be in `path` value.
Possible values :
- URI
- absolute path
- relative path

That causes failures on Windows:
```scala
Paths.get("jar:file:...") // <- fails on Windows, should be replaced by `Paths.get(uri: URI)`
URI.create("win\path\separator") // <- fails on Windows

```
The correct fix here would be to enforce a single format to `path` value across the codebase. However, that's challenging.
I've tried to align it to be URI but got much more failing tests. 

I will try to fix this thing properly later.

tgodzik (approved):
LGTM!

PR #2682, opened 2021-04-06
Add code action to add/remove string interpolation

### LINKED ISSUE #2666: Code action to convert strings to interpolation strings
Often, it would be nice to just be able to quickly change to an interpolation string, which is doable using completions, but not that intuitive. 

We might also consider the reverse code action if there are no interpolations in the string maybe we could offer a code action of type "refactor" to remove the useless interpolator.

A new code action/actions addressing the above points. 

You can always use completions to create an interpolated string.

**Additional context**
Feature based on the discussion here:
https://github.com/scalameta/metals/pull/1723

**Search terms**:
code action interpolation string

Giggiux:
This pull request should close #2666, by adding two code actions: 

1. Quickly interpolate a string or a multiline string
2. Quickly remove an interpolated string, if the string doesn't contains "Splice" parts.

Fixed failing tests due to new code actions and also added some for the new features.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/StringActions.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/StringActions.scala
@@ -32,8 +32,8 @@ class StringActions(buffers: Buffers, trees: Trees) extends CodeAction {
             Trees.defaultTokenizerDialect(source).tokenize.toOption
           )
         tokenized match {
-          case Some(tokens) =>
-            tokens
+          case Some(tokens) => {
+            val stripMarginActions = tokens
               .filter(t =>
                 t.pos.startLine == range.getStart.getLine
                   && t.pos.endLine == range.getEnd.getLine
@@ -68,6 +68,33 @@ class StringActions(buffers: Buffers, trees: Trees) extends CodeAction {
                 Nil
             }
 
+            val interpolationActions = tokens.collect {
+              case token: Token.Constant.String
+                  if token.pos.toLSP.overlapsWith(range) =>
+                interpolateAction(uri, token.pos.toLSP)
+            }.toList
+
+            val removeInterpolationActions = tokens.zipWithIndex
+              .flatMap {
+                case (start: Token.Interpolation.Start, i: Int)
+                    if (i + 2 < tokens.length
+                      && tokens(i + 2).is[Token.Interpolation.End]) =>
+                  def overlaps = List(start, tokens(i + 1), tokens(i + 2))
+                    .exists(_.pos.toLSP.overlapsWith(range))
+                  if (overlaps) {
+                    val lspRange = start.pos.toLSP
+                    val editRange =
+                      new l.Range(lspRange.getStart, lspRange.getEnd)
+                    Some(removeInterpolationAction(uri, editRange))
+                  } else {
+                    None
+                  }
+                case _ => None
+              }
+
+            stripMarginActions ++ interpolationActions ++ removeInterpolationActions
+          }
+
           case None => Nil
         }
       }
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/codeactions/StringActions.scala
```diff
@@ -68,6 +68,80 @@ class StringActions(buffers: Buffers, trees: Trees) extends CodeAction {
...
+                }
+              }
+
+              val removeInterpolationToken =
+                tokens.zipWithIndex
+                  .collect {
+                    case (start: Token.Interpolation.Start, i: Int) =>
+                      if (
+                        start.pos.toLSP.overlapsWith(range)
+                        && isPart(i + 1) && isEnd(i + 2)
```
tgodzik, line 111: There might be more parts than one I think here. When encountering Start you could do something similar to 

https://github.com/scalameta/metals/blob/8c24e5d2b909dda88f39312c66089195b5e36e9c/metals/src/main/scala/scala/meta/internal/metals/MultilineStringFormattingProvider.scala#L208

So it would probably make sense to use index instead of doing collect. But otherwise something like this should work:

```scala
case (start: Token.Interpolation.Start, i: Int) if tokens(i+2).is[Token.Interpolation.End] =>
   token
```
Shortest possible interpolation string would be Start, Part, End

metals/src/main/scala/scala/meta/internal/metals/codeactions/StringActions.scala
```diff
@@ -87,23 +161,59 @@ class StringActions(buffers: Buffers, trees: Trees) extends CodeAction {
...
+
+    codeAction
+  }
 }
 
 object StringActions {
-  def title: String = "Convert to multiline string"
+  def multilineTitle: String = "Convert to multiline string"
+  def interpolationTitle: String = "Convert to interpolation string"
+  def removeInterpolationTitle: String = "Remove interpolation string"
```
tgodzik, line 218:
```suggestion
  def removeInterpolationTitle: String = "Remove interpolation"
```
should be enough,.

tests/unit/src/test/scala/tests/codeactions/StringActionsLspSuite.scala
```diff
@@ -143,6 +265,29 @@ class StringActionsLspSuite extends BaseCodeActionLspSuite("stringActions") {
...
+        |${StringActions.interpolationTitle}""".stripMargin,
+    """|package a
+       |
+       |object A {
+       |  val e = s"hello ${} "
+       |  val c = s"this is a string"
+       |  val d = s"hello ${} "
+       |}
+       |""".stripMargin,
+    1
```
tgodzik, line 288: Could you use named parameter?
```suggestion
    selectedActionIndex = 1
```
best ot use it in case of primitive params, makes it clearer for the reviewers.

tgodzik (changes requested):
Thanks for the contribution! Looks good, but I think we can simplify it a bit.

metals/src/main/scala/scala/meta/internal/metals/codeactions/StringActions.scala
Giggiux: The motivation behind this solution is due to which part overlaps with `range` when considering the case of multiline interpolated strings: it's one of Start, Part, or End, based on if the cursor is in on of the following positions: 
```
s"<<">>"my awesome <<stri>>ng"<<">>"
```  

On non multiline strings the overlap is given on all three parts, and the collect returns three time the Start token, that's also why I do `.distinct` in the end.

Also, not sure it's needed, but the `isPart`, `isStart`, and `isEnd` functions also check if the index of the token exists before getting it and checking it for the correct token type, If not needed, `isInstanceOf[]` makes everything cleaner.

In general, can you help me providing an example when an interpolated strings without Splices can have more Parts in it?

tgodzik: I think without splices it can only be max 3. Start, Part, End.

Why not then:
```scala
.flatMap{
  case (start: Token.Interpolation.Start, i: Int) if tokens(i+2).is[Token.Interpolation.End] =>
    def overlaps = List(start, tokens(i + 1), tokens(i +2) ).exists(_.pos.toLSP.overlapsWith(range))
    if( i +2 < tokens.length && overlaps) {
      val lspRange = start.pos.toLSP
      val editRange = new l.Range(lspRange.getStart, lspRange.getEnd)
      Some( removeInterpolationAction(uri, editRange))
    } else {
     None
    }
    
  case _ => None
}

```

Giggiux: I've implemented the suggested changes in the latest commit
I just changed where to check for `i +2 < tokens.length`, by doing it before using `tokens(i+2)` in the following line:
```scala
 case (start: Token.Interpolation.Start, i: Int) if tokens(i+2).is[Token.Interpolation.End] =>
 ```
 That became
 ```scala
  case (start: Token.Interpolation.Start, i: Int) if ( i+2 < tokens.length && tokens(i+2).is[Token.Interpolation.End]) =>
  ```

ckipp01:
So from testing this locally it seems that this works great when there actually is no interpolation going on, but if there is, then it doesn't offer it. For example:



Whereas if I remove the `$` then it works as expected:

tgodzik:
I mean, that is the intended behaviour. We don't want to add interpolation if it's already there. I am not sure I understand the problem

ckipp01:
> I mean, that is the intended behaviour. We don't want to add interpolation if it's already there. I am not sure I understand the problem

ok ok, just wanted to double check. I thought that might be the case :)

Looks good to me 👍🏼

ckipp01 (approved):

tgodzik:
> > I mean, that is the intended behaviour. We don't want to add interpolation if it's already there. I am not sure I understand the problem
> 
> ok ok, just wanted to double check. I thought that might be the case :)
> 
> Looks good to me 👍🏼 

Thanks for checking it out then! I might now have described clearly the issue's intention 😅

PR #2696, opened 2021-04-09
Print exception stack trace in DAP tests

tgodzik:


ckipp01:
So many test issues lately. What in the world is going on.

ckipp01 (approved):
This should help 👍🏼

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -497,7 +497,7 @@ class DebugProvider(
     val socket = new Socket()
 
     val address = new InetSocketAddress(uri.getHost, uri.getPort)
-    val timeout = TimeUnit.SECONDS.toMillis(10).toInt
+    val timeout = TimeUnit.SECONDS.toMillis(20).toInt
```
tgodzik, line 500: Bumping up this one, maybe it's help - most the issue seems to be tests?

tests/unit/src/main/scala/tests/BaseLspSuite.scala
```diff
@@ -51,6 +53,24 @@ abstract class BaseLspSuite(suiteName: String) extends BaseSuite {
...
+    super.munitTestTransforms :+
+      new TestTransform(
+        "Print exception Stack",
+        { test =>
+          test.withBody(() =>
+            test
+              .body()
+              .andThen {
+                case Failure(exception) =>
+                  exception.printStackTrace()
```
tgodzik, line 66: Without this, munit will not print the actual exception.

olafurpg: The `-F` option should print the full original stack trace https://scalameta.org/munit/docs/troubleshooting.html#print-full-stack-traces

It might make sense to enable this by default in CI

olafurpg: I would avoid using this workaround here because I think it will make all test failures very verbose

ckipp01: ooo I didn't know about the `-F` option. I think that's a good idea to use for this.

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -497,7 +497,9 @@ class DebugProvider(
     val socket = new Socket()
 
     val address = new InetSocketAddress(uri.getHost, uri.getPort)
-    val timeout = TimeUnit.SECONDS.toMillis(10).toInt
+    def isCI: Boolean = "true" == System.getenv("CI")
+    val timeoutDuration = if (isCI) 60 else 10
```
tgodzik, line 501: Even 60 seconds do not seem enough :/ the java process doesn't seem to be able to start, or is it jsut sockets?

tests/unit/src/main/scala/tests/BaseLspSuite.scala
olafurpg: The MUnit docs are missing a reference table that documents all of the flag options. Most of the flags in the table here should work https://github.com/sbt/junit-interface 

The `-F` flag name is the same as for ScalaTest.

I regret using cryptic single-character flags instead of more descriptive names 😢

olafurpg: The only time I've needed the `-F` flag was to debug an issue in MUnit internals. If the stack trace is hidden it indicates the MUnit framework is failing with an unexpected exception

olafurpg (commented):
It’s possible to add the F flag in build.sbt for CI only. I think we already have some settings to use the sbt logger.

tgodzik:
This got a much simpler PR :sweat_smile:  Merging this one in, since it might be useful to get as much information as possible on CI.

Issue #2686, opened 2021-04-07
Unexpected import when using insert inferred type and the import has been renamed


ckipp01:
**Describe the bug**

When using the new insert inferred type code action and you are importing something that is already in scope, but the import was renamed, it will insert the old un-renamed import package.

**To Reproduce**

Take a look in the Metals code base for example in `InsertInferredType.scala` and try to insert the inferred type where the `@@` is down below.

```scala
import org.eclipse.{lsp4j => l}

class InsertInferredType(trees: Trees, compilers: Compilers)
    extends CodeAction {

  import InsertInferredType._
  override def kind: String = l.CodeActionKind.QuickFix

  override def contribute(
      params: l.CodeActionParams,
      token: CancelToken
  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = Future {

    def insertInferTypeAction(title: String): l.CodeAction = {
      val code<<@@>>Action = new l.CodeAction()
```

I'd expect for this to give me:

```scala
      val codeAction: l.CodeAction = new l.CodeAction()
```

But what I really get is:

```scala
      val codeAction: lsp4j.CodeAction = new l.CodeAction()
```

1. Go to the file in Metals
2. Insert the inferred type
3. See result

**Installation:**
- Metals version: v0.10.1

**Search terms**

insert inferred type with renamed package

tgodzik:
Thanks for finding that one! Seems it might be an issue with auto imports. Does it work with `implement all abstract members` action?

ckipp01:
> Does it work with implement all abstract members action?

This seems to work as expected. Given:

```scala
import java.time.{Instant => I}

trait Main {
  def a: I
  def b: I
}

class Thing extends Main {
}
```
When I trigger the code action I get:

```scala
import java.time.{Instant => I}

trait Main {
  def a: I
  def b: I
}

class Thing extends Main {

  override def a: I = ???

  override def b: I = ???

}
```
So that looks good.

PR #2699, opened 2021-04-10
Make insert type work with renamed imports

tgodzik:
Fixes https://github.com/scalameta/metals/issues/2686

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
@@ -287,22 +287,36 @@ class MetalsGlobal(
             }
           }
         case ThisType(sym) =>
-          if (history.tryShortenName(name)) NoPrefix
+          val owners = sym.ownerChain
+          // to make sure we always use renamed package
+          // what is saved in renames is actually companion module of a package
+          val renamedOwnerIndex =
+            owners.indexWhere(s => history.renames.contains(s.companionModule))
+          if (renamedOwnerIndex < 0 && history.tryShortenName(name)) NoPrefix
           else {
-            val owners = sym.ownerChain
-            val prefix = owners.indexWhere { owner =>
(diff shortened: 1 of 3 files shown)


ckipp01:
So I'm not 100% sure what's going on, but testing it on the example where I found this in `InsertInferredType.scala`, I still have the same issue. In a simple example like you have in your test, it seems to work fine, but if you try to insert the inferred type on `codeAction` here, you still end up with `lsp4j.CodeAction`.

https://github.com/scalameta/metals/blob/13dd814f50f009ce6e748c4d4faf1fd92c7ff256/metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala#L34

mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
```diff
@@ -287,22 +287,36 @@ class MetalsGlobal(
             }
           }
         case ThisType(sym) =>
-          if (history.tryShortenName(name)) NoPrefix
+          val owners = sym.ownerChain
+          // to make sure we always use renamed package
+          // what is saved in renames is actually companion module of a package
+          val renamedOwnerIndex =
```
tgodzik, line 293: So this turned out harder than anticipated. We would not properly rename packages before, which would also pop up in signatures.

ckipp01: huh, interesting. Great job finding this!

ckipp01 (approved):
Awesome seeing this working! LGTM!

mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
tgodzik: I have no idea actually why the package type and module type is different, but I don't think this will actually break anything. If it will, I will probably need to figure it out :S

PR #2702, opened 2021-04-12
Don't override testSettings in unit tests

tgodzik:
I haven't noticed this one in the previous PR, because the failed test came from slow tests :/


ckipp01 (approved):
LGTM

PR #2703, opened 2021-04-12
Bump sbt to 1.5.0

ckipp01:


tgodzik (approved):
LGTM!

build.sbt
```diff
@@ -313,8 +311,6 @@ val mtagsSettings = List(
     (ThisBuild / baseDirectory).value / "mtags",
     scalaVersion.value
   ),
-  // @note needed to deal with issues with dottyDoc
-  Compile / doc / sources := Seq.empty,
```
ckipp01, line 317: If I'm not mistaken this should work now right? Or is it silly to remove since we don't use the docs anyways?

tgodzik: That was removed mostly due to issues with the Dotty plugin, it's fine to remove if everything publishes fine for the different Scala 3 versions. Might be useful to run `+mtags/publishLocal` to test that out.

ckipp01: Ok perfect. I did a full `+publishLocal` to test everything so this should be good then 👍🏼

ckipp01: What the heck... breaks in CI.

tgodzik: Hmm... This might not be related, it seems we are failing to resolve from bintray, which is shutting down I think? I am not sure though  why it's not available on maven central then.

tgodzik:
```
We replaced the isDottyCompat extension method by the static for3Use2_13 object whose usage is:

(“a” % “b” % “1.0”).cross(CrossVersion.for3Use2_13)

Symmetrically we added the for2_13Use3 that has no counterpart in sbt-dotty.

For now, the doc task is made available on Scala 3 by setting:

thisBuild / resolvers += Resolver.JCenterRepository

This is due to scala3doc depending on Dokka, whose artifacts are not available in the default resolvers. This dependency is planned to be removed.
```

from https://contributors.scala-lang.org/t/scala-center-q4-2020-update/4873

tgodzik: That is not related, but this is https://github.com/sbt/sbt/issues/6294#issuecomment-817836965 :/

tgodzik: Yeah, I think this might not be fixed for M1, we would need to drop it.

ckipp01: > Yeah, I think this might not be fixed for M1, we would need to drop it.

Sounds good. For now I kept it in, and then when we drop M1 we can remove it. So nice to see all ✅ again.

ckipp01:
Something is gone wrong and I didn't catch it. Everything passed, but all the unit tests didn't actually run..... I'm looking into it.

ckipp01:
I'm totally stumped here with what's going on. @adpi2 do you by chance have any idea why the `test` would behave differently with an update to 1.5.0? Before this pr `unit/test` for example ran fine, and now after this pr `unit/test` won't test anything.

PR #2705, opened 2021-04-13
Add testframeworks back in to ensure tests run.

ckipp01:
I removed this in https://github.com/scalameta/metals/pull/2703 and shouldn't have. It's causing no test to be detected and ran.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -529,6 +529,7 @@ lazy val testSettings: Seq[Def.Setting[_]] = List(
   Test / parallelExecution := false,
   publish / skip := true,
   fork := true,
+  testFrameworks := List(TestFrameworks.MUnit),
   Test / testOptions ++= {
     if (isCI) {
       // Enable verbose logging using sbt loggers in CI.


tgodzik (approved):
Och damn, I did the same in another PR, but thought we needed to update to 1.5.0 before it can work. Any idea why it doesn't work in Metals with 1.5.0?

ckipp01:
> Och damn, I did the same in another PR, but thought we needed to update to 1.5.0 before it can work. Any idea why it doesn't work in Metals with 1.5.0?

Yea normally it wouldn't be needed, but we explicitly set `testFrameworks := List()` inside of `inThisBuild`, so it needs to be added back in. Totally missed this before!

ckipp01:
Not super sure what's going on with the sbt, maven, and gradle tests. Just got done running them locally and they seem to be fine, but in CI when it hits the debugging related things we have issues. I'm going to merge this though just to actually get tests back running.

PR #2706, opened 2021-04-13
Always print stack traces when testing

tgodzik:
Previously, we would get:
```
==> X tests.codeactions.ExtractClassLspSuite.only-sourceFile  1.892s munit.ComparisonFailException: /home/tgodzik/Documents/metals/tests/unit/src/test/scala/tests/codeactions/ExtractClassLspSuite.scala:19 Obtained empty output!
18:
19:  check(
20:    "only-sourceFile",
```
it is impossible to say where the exception occurs in. Now it should be:
```
==> X tests.codeactions.ExtractClassLspSuite.only-sourceFile  1.889s munit.ComparisonFailException: /home/tgodzik/Documents/metals/tests/unit/src/test/scala/tests/codeactions/ExtractClassLspSuite.scala:19 Obtained empty output!
18:
19:  check(
20:    "only-sourceFile",
    at munit.Assertions.failComparison(Assertions.scala:275)
    ....
    at tests.Assertions$.assertNoDiff(Assertions.scala:73)
    at tests.TestingServer.$anonfun$assertCodeAction$2(TestingServer.scala:956)
    at scala.util.Success.$anonfun$map$1(Try.scala:255)
    at scala.util.Success.map(Try.scala:213)
    at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
    at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
    at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -535,7 +535,7 @@ lazy val testSettings: Seq[Def.Setting[_]] = List(
       // Enable verbose logging using sbt loggers in CI.
       List(Tests.Argument(TestFrameworks.MUnit, "+l", "--verbose", "-F"))
     } else {
-      Nil
+      List(Tests.Argument(TestFrameworks.MUnit, "-F"))
     }
   }
 )


ckipp01 (approved):
I suppose locally this is helpful as well 👍🏼

build.sbt
```diff
@@ -535,7 +535,7 @@ lazy val testSettings: Seq[Def.Setting[_]] = List(
       // Enable verbose logging using sbt loggers in CI.
       List(Tests.Argument(TestFrameworks.MUnit, "+l", "--verbose", "-F"))
     } else {
-      Nil
+      List(Tests.Argument(TestFrameworks.MUnit, "-F"))
```
olafurpg, line 538: I would advise against this. Full stack traces are HUGE most of the time making it much harder to make sense of normal errors. The only time I have needed a full stack trace was to fix a bug in the scalacheck-munit integration. If there's a bug in how traces are trimmed then it would be best to fix them upstream https://github.com/scalameta/munit/blob/35d868403aca0e71b31b8a794472ddcbee370334/munit/shared/src/main/scala/munit/internal/console/StackTraces.scala

Alternatively, it could be that the exception is happening inside a custom test/value transform 🤔

tgodzik:
Let's try to fix this one in munit then. We can do a workaround with `-- -F` for the time being.

PR #2708, opened 2021-04-14
Add hint to welcome messages to use '-F' flag

tgodzik:
Connected to https://github.com/scalameta/munit/issues/348

So this is kind of a middle ground until we fix it in munit.

### DIFF
--- a/project/Welcome.scala
+++ b/project/Welcome.scala
@@ -24,6 +24,11 @@ object Welcome {
     ),
     UsefulTask(
       "",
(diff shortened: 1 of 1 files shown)


tgodzik:
Fun, so it seems it is all getting worse now :| Every single test that was run on ubuntu failed.

ckipp01 (approved):
LGTM.

ckipp01:
> Fun, so it seems it is all getting worse now :| Every single test DAP that was run on ubuntu failed.

Yea that's no bueno :/

PR #2709, opened 2021-04-14
Fix issues with DAP tests

tgodzik:
Change 0.0.0.0 to 127.0.0.1 to avoid socket connection timeouts

Previously, tests would often fail for DAP and the only way to fix it seems to force 127.0.0.1, which seems a much safer option.

Thanks to @dos65 for the hint!


dos65 (approved):
LGTM!

ckipp01 (approved):
So nice to see all those ✅ again finally 😍 .

Great job chasing this down!

PR #2710, opened 2021-04-14
Treat .srcjar as jars

darl:
I'm currently working on bloop config generation from bazel files.
And found difficulties with generated sources (scalapb for example).
Currently i'm attaching them as external dependencies with jars generated by bazel.
It's common in bazel world to place generated files inside `.srcjar` files but they are ignored by metals.

It's a bit hacky but i hope it will be accepted 😄


tgodzik (approved):
Thanks for the contribution! That looks like a resonable fix. I will just update the branch to run on most recent master to make sure everything is passing. We did some fixes today that should help.

PR #2723, opened 2021-04-15
Code action to move classes to their own files if requirements are met

### LINKED ISSUE #2691: "Move Class to file" refactoring
Given: I have 3 small case classes and a trait in 1 file

When: The case classes grow in size and should be extracted to separate files

Then: I should be able to invoke "Move class to file" refactoring which moves the class out of its current file and into its new one, changing references if the new file is of a new package, auto-imports needed packages in the destination file and deletes unneeded imports in the original file.

Giggiux:
This pull request should close #2691 
The "Extract Class" code action will appear on classes if:
- There is more than one class definition in the file. Inner classes are not counted
- The class doesn't extends Traits/classes defined as `sealed`  
- The class itself isn't `sealed`

The new file will have the name of the Class, if doesn't exists already (otherwise the file creation will fail and every other TextEdit action should not happen)
In the new file there will be all the package definition and imports definition that appear before the class definition in the source file, in the same order.

Another code action is added, as suggested in the issue: "Rename file as class".
This code action appears when there is a single class definition in the file, that has a different name than the file itself, and doesn't extends a Trait with the same filename.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
@@ -17,11 +17,15 @@ final class CodeActionProvider(
     scalafixProvider: ScalafixProvider,
     trees: Trees
 )(implicit ec: ExecutionContext) {
+
+  private val extractMemberAction = new ExtractRenameMember(buffers, trees)
+
   private val allActions: List[CodeAction] = List(
     new ImplementAbstractMembers(compilers),
     new ImportMissingSymbol(compilers),
     new CreateNewSymbol(),
     new StringActions(buffers, trees),
+    extractMemberAction,
     new OrganizeImports(scalafixProvider, buildTargets),
     new InsertInferredType(trees, compilers)
   )
@@ -46,4 +50,14 @@ final class CodeActionProvider(
     Future.sequence(actions).map(_.flatten)
   }
 
+  def executeCommands(
+      codeActionCommandData: CodeActionCommandData,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[CodeActionCommandResult] = {
+    codeActionCommandData match {
+      case data: ExtractMemberDefinitionData =>
+        extractMemberAction.executeCommand(data)
+      case data => Future.failed(new IllegalArgumentException(data.toString))
+    }
+  }
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
@@ -53,7 +53,7 @@ final class CodeActionProvider(
   def executeCommands(
       codeActionCommandData: CodeActionCommandData,
       token: CancelToken
-  )(implicit ec: ExecutionContext): Future[CodeActionCommandResult] = {
+  ): Future[CodeActionCommandResult] = {
     codeActionCommandData match {
       case data: ExtractMemberDefinitionData =>
         extractMemberAction.executeCommand(data)
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -49,6 +49,7 @@ import scala.meta.internal.metals.Messages.AmmoniteJvmParametersChange
 import scala.meta.internal.metals.Messages.IncompatibleBloopVersion
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.ammonite.Ammonite
+import scala.meta.internal.metals.codeactions.ExtractMemberDefinitionData
 import scala.meta.internal.metals.codelenses.RunTestCodeLens
 import scala.meta.internal.metals.codelenses.SuperMethodCodeLens
 import scala.meta.internal.metals.codelenses.WorksheetCodeLens
@@ -82,6 +83,7 @@ import ch.epfl.scala.{bsp4j => b}
 import com.google.gson.JsonElement
 import com.google.gson.JsonPrimitive
 import io.undertow.server.HttpServerExchange
+import org.eclipse.lsp4j.ExecuteCommandParams
 import org.eclipse.lsp4j._
 import org.eclipse.lsp4j.jsonrpc.messages.{Either => JEither}
 import org.eclipse.lsp4j.jsonrpc.services.JsonNotification
@@ -1541,6 +1543,22 @@ class MetalsLanguageServer(
   def executeCommand(
       params: ExecuteCommandParams
   ): CompletableFuture[Object] = {
+    def textDocumentPosition(
+        args: mutable.Buffer[AnyRef]
+    ): Option[(TextDocumentPositionParams, String)] = {
+      for {
+        arg0 <- args.lift(0)
+        uri <- Argument.getAsString(arg0)
+        arg1 <- args.lift(1)
+        line <- Argument.getAsInt(arg1)
+        arg2 <- args.lift(2)
+        character <- Argument.getAsInt(arg2)
+        pos = new l.Position(line, character)
+        textDoc = new l.TextDocumentIdentifier(uri)
+        params = new TextDocumentPositionParams(textDoc, pos)
+      } yield (params, uri)
+    }
+
     val command = Option(params.getCommand).getOrElse("")
     command.stripPrefix("metals.") match {
       case ServerCommands.ScanWorkspaceSources() =>
@@ -1744,17 +1762,7 @@ class MetalsLanguageServer(
       case ServerCommands.InsertInferredType() =>
         CancelTokens.future { token =>
           val args = params.getArguments().asScala
-          val futureOpt = for {
-            arg0 <- args.lift(0)
-            uri <- Argument.getAsString(arg0)
-            arg1 <- args.lift(1)
-            line <- Argument.getAsInt(arg1)
-            arg2 <- args.lift(2)
-            character <- Argument.getAsInt(arg2)
-            pos = new l.Position(line, character)
-            textDoc = new l.TextDocumentIdentifier(uri)
-            params = new TextDocumentPositionParams(textDoc, pos)
-          } yield {
+          val futureOpt = textDocumentPosition(args).map { case (params, uri) =>
             for {
               edits <- compilers.insertInferredType(params, token)
               if (!edits.isEmpty())
@@ -1768,6 +1776,37 @@ class MetalsLanguageServer(
           futureOpt.getOrElse {
             languageClient.showMessage(Messages.InsertInferredTypeFailed)
             Future.unit
+          }.withObjectValue
+        }
+      case ServerCommands.ExtractMemberDefinition() =>
+        CancelTokens.future { token =>
+          val args = params.getArguments().asScala
+
+          val futureOpt = for {
+            (params, uri) <- textDocumentPosition(args)
+          } yield {
+            val data = ExtractMemberDefinitionData(uri, params)
+            for {
+              result <- codeActionProvider.executeCommands(data, token)
+              future <- languageClient.applyEdit(result.edits).asScala
+            } yield {
+              result.goToLocation.foreach { location =>
+                languageClient.metalsExecuteClientCommand(
+                  new ExecuteCommandParams(
+                    ClientCommands.GotoLocation.id,
+                    List(location: Object).asJava
+                  )
+                )
+              }
+            }
+          }
+
+          futureOpt.getOrElse {
+            Future(
+              languageClient.showMessage(
+                Messages.ExtractMemberDefinitionFailed
+              )
+            )
           }.withObjectValue
         }
       case cmd =>
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/CodeActionCommandData.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/CodeActionCommandData.scala
@@ -1,0 +1,21 @@
+package scala.meta.internal.metals.codeactions
+
+import org.eclipse.lsp4j.ApplyWorkspaceEditParams
+import org.eclipse.lsp4j.Location
+import org.eclipse.lsp4j.TextDocumentPositionParams
+
+trait CodeActionCommandData {
+  val uri: String
+  val actionType: String
+}
+
+case class ExtractMemberDefinitionData(
+    uri: String,
+    params: TextDocumentPositionParams,
+    actionType: String = ExtractRenameMember.extractDefCommandDataType
+) extends CodeActionCommandData
+
+case class CodeActionCommandResult(
+    edits: ApplyWorkspaceEditParams,
+    goToLocation: Option[Location]
+)
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
@@ -120,9 +120,28 @@ class ExtractRenameMember(
     case _ => false
   }
 
-  private def sealedNames(tree: Tree): List[String] = tree.collect {
-    case node: Defn.Trait if isSealed(node) => node.name.value
-    case node: Defn.Class if isSealed(node) => node.name.value
+  private def sealedNames(tree: Tree): List[String] = {
+    def completeName(node: Member): String = {
+      def completePreName(node: Tree): List[String] = {
+        node.parent match {
+          case Some(t) =>
+            t match {
+              case o: Defn.Object => o.name.value :: completePreName(o)
+              case po: Pkg.Object => po.name.value :: completePreName(po)
+              case _: Source => Nil
+              case _ => completePreName(t)
+            }
+          case None => Nil
+        }
+      }
+
+      (node.name.value :: completePreName(node)).reverse.mkString(".")
+    }
+
+    tree.collect {
+      case node: Defn.Trait if isSealed(node) => completeName(node)
+      case node: Defn.Class if isSealed(node) => completeName(node)
+    }
   }
 
   private def notSealed(
@@ -135,12 +154,19 @@ class ExtractRenameMember(
     !memberExtendsSealed && !isSealed(member)
   }
 
+  private def names(t: Term): List[Term.Name] = {
+    t match {
+      case s: Term.Select => names(s.qual) :+ s.name
+      case n: Term.Name => n :: Nil
+    }
+  }
+
   private def newFileContent(
       tree: Tree,
       range: l.Range,
       member: Member,
       companion: Option[Member]
-  ): String = {
+  ): (String, Int) = {
     // List of sequential packages or imports before the member definition
     val packages: ListBuffer[Pkg] = ListBuffer()
     val imports: ListBuffer[Import] = ListBuffer()
@@ -161,22 +187,18 @@ class ExtractRenameMember(
 
     traverser(tree)
 
-    def names(t: Term): List[Term.Name] = {
-      t match {
-        case s: Term.Select => names(s.qual) :+ s.name
-        case n: Term.Name => n :: Nil
+    def mergeNames(ns: List[Term.Name]): Option[Term.Ref] = {
+      def merge(n1: Term.Ref, n2: Term.Name): Term.Select = n1 match {
+        case s: Term.Select => Term.Select(qual = s, name = n2)
+        case n: Term.Name => Term.Select(qual = n, name = n2)
       }
-    }
 
-    def merge(n1: Term.Ref, n2: Term.Name): Term.Select = n1 match {
-      case s: Term.Select => Term.Select(qual = s, name = n2)
-      case n: Term.Name => Term.Select(qual = n, name = n2)
-    }
-
-    def mergeNames(ns: List[Term.Name]): Option[Term.Ref] = ns match {
-      case Nil => None
-      case head :: Nil => Some(head)
-      case head :: second :: xs => Some(xs.foldLeft(merge(head, second))(merge))
+      ns match {
+        case Nil => None
+        case head :: Nil => Some(head)
+        case head :: second :: xs =>
+          Some(xs.foldLeft(merge(head, second))(merge))
+      }
     }
 
     val termNames = packages
@@ -191,9 +213,17 @@ class ExtractRenameMember(
       member.toString ::
       companion.map(_.toString).getOrElse("") :: Nil
 
-    structure
-      .filter(_.nonEmpty)
-      .mkString("\n\n")
+    val preDefinitionLines = pkg.toList.length + imports.length
+    val defnLine =
+      if (preDefinitionLines == 0) 0
+      else preDefinitionLines + 2 // empty line + defn line
+
+    (
+      structure
+        .filter(_.nonEmpty)
+        .mkString("\n\n"),
+      defnLine
+    )
   }
 
   private def parents(member: Member): List[String] = {
@@ -202,6 +232,10 @@ class ExtractRenameMember(
       t.inits.flatMap {
         _.tpe match {
           case Type.Name(value) => Some(value)
+          case t: Type.Select =>
+            Some(
+              (names(t.qual) :+ t.name).mkString(".")
+            )
           case _ => None
         }
       }
@@ -289,7 +323,7 @@ class ExtractRenameMember(
       definitions = membersDefinitions(tree)
       memberDefn <- definitions.find(_.name.pos.toLSP.overlapsWith(range))
       companion = definitions.find(isCompanion(memberDefn))
-      fileContent = newFileContent(
+      (fileContent, defnLine) = newFileContent(
         tree,
         range,
         memberDefn,
@@ -307,6 +341,9 @@ class ExtractRenameMember(
         companion
       )
       val newFileMemberRange = new l.Range()
+      val pos = new l.Position(defnLine, 0)
+      newFileMemberRange.setStart(pos)
+      newFileMemberRange.setEnd(pos)
       val workspaceEdit = new WorkspaceEdit(Map(uri -> edits.asJava).asJava)
       CodeActionCommandResult(
         new ApplyWorkspaceEditParams(workspaceEdit),
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
@@ -129,7 +129,7 @@ class ExtractRenameMember(
               case o: Defn.Object => o.name.value :: completePreName(o)
               case po: Pkg.Object => po.name.value :: completePreName(po)
               case _: Source => Nil
-              case _ => completePreName(t)
+              case _ => Nil
             }
           case None => Nil
         }
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
@@ -128,6 +128,7 @@ class ExtractRenameMember(
             t match {
               case o: Defn.Object => o.name.value :: completePreName(o)
               case po: Pkg.Object => po.name.value :: completePreName(po)
+              case tmpl: Template => completePreName(tmpl)
               case _: Source => Nil
               case _ => Nil
             }
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
@@ -1,0 +1,382 @@
+package scala.meta.internal.metals.codeactions
+
+import scala.collection.mutable.ListBuffer
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+import scala.meta.Defn
+import scala.meta.Import
+import scala.meta.Member
+import scala.meta.Mod
+import scala.meta.Pkg
+import scala.meta.Source
+import scala.meta.Template
+import scala.meta.Term
+import scala.meta.Tree
+import scala.meta.Type
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals._
+import scala.meta.internal.metals.codeactions.ExtractRenameMember.CodeActionCommandNotFoundException
+import scala.meta.internal.metals.codeactions.ExtractRenameMember.getMemberType
+import scala.meta.internal.parsing.Trees
+import scala.meta.io.AbsolutePath
+import scala.meta.pc.CancelToken
+import scala.meta.transversers.SimpleTraverser
+
+import org.eclipse.lsp4j.ApplyWorkspaceEditParams
+import org.eclipse.lsp4j.Location
+import org.eclipse.lsp4j.WorkspaceEdit
+import org.eclipse.{lsp4j => l}
+
+class ExtractRenameMember(
+    buffers: Buffers,
+    trees: Trees
+)(implicit ec: ExecutionContext)
+    extends CodeAction {
+
+  override def contribute(params: l.CodeActionParams, token: CancelToken)(
+      implicit ec: ExecutionContext
+  ): Future[Seq[l.CodeAction]] = Future {
+    val uri = params.getTextDocument.getUri
+    val path = uri.toAbsolutePath
+    val range = params.getRange
+
+    trees.get(path) match {
+      case Some(tree) =>
+        val fileName = uri.toAbsolutePath.filename.replaceAll("\\.scala$", "")
+
+        val definitions = membersDefinitions(tree)
+        val sldNames = sealedNames(tree)
+        val defnAtCursor =
+          definitions.find(_.name.pos.toLSP.overlapsWith(range))
+
+        def canRenameDefn(defn: Member): Boolean = {
+          val differentNames = defn.name.value != fileName
+          val newFileUri = newPathFromClass(uri, defn)
+
+          differentNames && !newFileUri.exists && defnAtCursor.exists(
+            _.equals(defn)
+          )
+        }
+
+        def canExtractDefn(defn: Member): Boolean = {
+          val differentNames = defn.name.value != fileName
+          val notExtendsSealedOrSealed = notSealed(defn, sldNames)
+          val companion = definitions.find(c => {
+            !c.equals(defn) && c.name.value.equals(defn.name.value)
+          })
+          val companionNotSealed =
+            companion.exists(notSealed(_, sldNames)) || companion.isEmpty
+
+          val newFileUri = newPathFromClass(uri, defn)
+
+          differentNames && notExtendsSealedOrSealed && companionNotSealed && !newFileUri.exists
+        }
+
+        definitions match {
+          case Nil => Nil
+          case head :: Nil if canRenameDefn(head) =>
+            Seq(renameFileAsMemberAction(uri, head))
+          case _ =>
+            val codeActionOpt = for {
+              defn <- defnAtCursor
+              if canExtractDefn(defn)
+              memberType <- getMemberType(defn)
+              title = ExtractRenameMember.title(memberType, defn.name.value)
+            } yield extractClassAction(uri, defn, title)
+
+            codeActionOpt.toList
+        }
+
+      case _ => Nil
+    }
+
+  }
+
+  private def membersDefinitions(tree: Tree): List[Member] = {
+    val nodes: ListBuffer[Member] = ListBuffer()
+
+    val traverser = new SimpleTraverser {
+      override def apply(tree: Tree): Unit = tree match {
+        case p: Pkg =>
+          super.apply(p)
+        case c: Defn.Class => nodes += c
+        case t: Defn.Trait => nodes += t
+        case o: Defn.Object => nodes += o
+        case e: Defn.Enum => nodes += e
+        case s: Source =>
+          super.apply(s)
+        case _ =>
+      }
+    }
+    traverser(tree)
+
+    nodes.toList
+  }
+
+  private def isSealed(t: Tree): Boolean = t match {
+    case node: Defn.Trait => node.mods.exists(_.isInstanceOf[Mod.Sealed])
+    case node: Defn.Class => node.mods.exists(_.isInstanceOf[Mod.Sealed])
+    case _ => false
+  }
+
+  private def sealedNames(tree: Tree): List[String] = tree.collect {
+    case node: Defn.Trait if isSealed(node) => node.name.value
+    case node: Defn.Class if isSealed(node) => node.name.value
+  }
+
+  private def notSealed(
+      member: Member,
+      sealedNames: List[String]
+  ): Boolean = {
+    val memberExtendsSealed: Boolean =
+      parents(member).exists(sealedNames.contains(_))
+
+    !memberExtendsSealed && !isSealed(member)
+  }
+
+  private def newFileContent(
+      tree: Tree,
+      range: l.Range,
+      member: Member,
+      companion: Option[Member]
+  ): String = {
+    // List of sequential packages or imports before the member definition
+    val packages: ListBuffer[Pkg] = ListBuffer()
+    val imports: ListBuffer[Import] = ListBuffer()
+
+    // Using a custom traverser to avoid hitting inner classes by stopping the recursion on the chosen members
+    object traverser extends SimpleTraverser {
+      override def apply(tree: Tree): Unit = tree match {
+        case p: Pkg if p.pos.toLSP.overlapsWith(range) =>
+          packages += Pkg(ref = p.ref, stats = Nil)
+          super.apply(p)
+        case i: Import =>
+          imports += i
+        case s: Source =>
+          super.apply(s)
+        case _ =>
+      }
+    }
+
+    traverser(tree)
+
+    def names(t: Term): List[Term.Name] = {
+      t match {
+        case s: Term.Select => names(s.qual) :+ s.name
+        case n: Term.Name => n :: Nil
+      }
+    }
+
+    def merge(n1: Term.Ref, n2: Term.Name): Term.Select = n1 match {
+      case s: Term.Select => Term.Select(qual = s, name = n2)
+      case n: Term.Name => Term.Select(qual = n, name = n2)
+    }
+
+    def mergeNames(ns: List[Term.Name]): Option[Term.Ref] = ns match {
+      case Nil => None
+      case head :: Nil => Some(head)
+      case head :: second :: xs => Some(xs.foldLeft(merge(head, second))(merge))
+    }
+
+    val termNames = packages
+      .flatMap(p => names(p.ref))
+
+    val mergedTermsOpt = mergeNames(termNames.toList)
+
+    val pkg: Option[Pkg] = mergedTermsOpt.map(t => Pkg(ref = t, stats = Nil))
+
+    val structure = pkg.toList.mkString("\n") ::
+      imports.mkString("\n") ::
+      member.toString ::
+      companion.map(_.toString).getOrElse("") :: Nil
+
+    structure
+      .filter(_.nonEmpty)
+      .mkString("\n\n")
+  }
+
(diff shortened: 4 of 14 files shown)


metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala
```diff
@@ -0,0 +1,250 @@
...
+
+    // Using a custom traverser to avoid hitting inner classes by stopping the recursion on Defn.Class
+    object traverser extends SimpleTraverser {
+      override def apply(tree: Tree): Unit = tree match {
+        case p: Pkg if p.pos.toLSP.overlapsWith(range) =>
+          structure += Pkg(ref = p.ref, stats = Nil)
+          super.apply(p)
+        case i: Import =>
+          structure += i
+        case classNode: Defn.Class
```
tgodzik, line 104: If we get class calculated previously we can just get use it and only add imports

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala
```diff
@@ -0,0 +1,250 @@
...
+    codeAction
+  }
+
+  private def extractClassAction(
+      uri: String,
+      tree: Tree,
+      classDefn: Defn.Class
+  ): l.CodeAction = {
+
+    def versionedTextDocumentIdentifier(
```
tgodzik, line 168: This should not be needed currently we don't use those anywhere

tests/unit/src/main/scala/tests/TestingClient.scala
```diff
@@ -101,7 +103,26 @@ final class TestingClient(workspace: AbsolutePath, buffers: Buffers)
   }
 
   private def applyWorkspaceEdit(edit: WorkspaceEdit): Unit = {
-    edit.getChanges().forEach(applyEdits)
+    val changes = edit.getChanges()
```
tgodzik, line 106:
```suggestion
    val changes = Option(edit.getChanges())
```

tests/unit/src/main/scala/tests/TestingClient.scala
```diff
@@ -101,7 +103,26 @@ final class TestingClient(workspace: AbsolutePath, buffers: Buffers)
...
+    val documentChanges = edit.getDocumentChanges()
```
tgodzik, line 107:
```suggestion
    val documentChanges = Option(edit.getDocumentChanges())
```

tests/unit/src/main/scala/tests/TestingClient.scala
```diff
@@ -148,14 +169,32 @@ final class TestingClient(workspace: AbsolutePath, buffers: Buffers)
...
   }
+
   def pathDiagnostics(filename: String): String = {
     pathDiagnostics(toPath(filename))
   }
+
+  private def toPath(filename: String): AbsolutePath =
+    TestingServer.toPath(workspace, filename)
+
+  def workspaceDiagnosticsCount: String = {
```
tgodzik, line 184: Why is this moved? Maybe that is not neccessary like some other change sin this file.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -138,7 +138,7 @@ final class TestingServer(
     isReliableFileWatcher = Properties.isLinux
   )
   server.connectToLanguageClient(client)
-
+  val Docstring: Regex = " *\\/?\\*.*".r
```
tgodzik, line 141: This file has a lot of unrelated changes, can you revert them? It will be easier to review them.

tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
```diff
@@ -33,7 +33,8 @@ abstract class BaseCodeActionLspSuite(suiteName: String)
       scalafixConf: String = "",
       scalacOptions: List[String] = Nil,
       configuration: => Option[String] = None,
-      scalaVersion: String = scalaVersion
+      scalaVersion: String = scalaVersion,
+      extraOperations: Option[() => Unit] = None
```
tgodzik, line 37: Why is this needed?

tests/unit/src/test/scala/tests/codeactions/ExtractClassLspSuite.scala
```diff
@@ -0,0 +1,498 @@
...
+      expectedActions: String,
+      newFileName: String,
+      selectedActionIndex: Int = 0,
+      expectNoDiagnostics: Boolean = true,
+      kind: List[String] = Nil,
+      scalafixConf: String = "",
+      scalacOptions: List[String] = Nil,
+      scalaVersion: String = scalaVersion
+  )(implicit loc: Location): Unit = {
+    val fileName: String = "A.scala"
```
tgodzik, line 460: Would you be able to reuse  the `check` method here? You should be able modify it by adding an additional parameter.

tgodzik (changes requested):
Thanks for the great work! I think we can further improve this, let me know if something is not clear enough.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  private def getClassDefinitions(tree: Tree): List[Defn.Class] = {
+    val nodes: ListBuffer[Defn.Class] = ListBuffer()
+
+    val traverser = new SimpleTraverser {
+      override def apply(tree: Tree): Unit = tree match {
+        case p: Pkg =>
+          super.apply(p)
+        case classNode: Defn.Class =>
```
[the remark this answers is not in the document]
Giggiux: Traits, Objects and Enums can be imported in the new file, if we copy them we have to: 
or delete and import them from the previous file, 
or ending up with duplicate code

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala
```diff
@@ -0,0 +1,250 @@
...
+            List(renameFileAsClassAction(uri, classDefinitions.head))
+          } else Nil
+
+        case _ => Nil
+      }
+
+    }
+  }
+
+  private def getClassDefinitions(tree: Tree): List[Defn.Class] = {
```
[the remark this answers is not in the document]
Giggiux: The method only extracts `Defn.Class`, not other elements of the Tree. I will definitely change the method name, but the result should not be changed imho.

tests/unit/src/main/scala/tests/ResourceOperations.scala
```diff
@@ -0,0 +1,74 @@
...
+import org.eclipse.lsp4j.{CreateFile, DeleteFile, RenameFile, ResourceOperation}
+
+import java.nio.channels.FileChannel
+import java.nio.file.{Files, Path, StandardCopyOption, StandardOpenOption}
+import scala.meta.internal.metals.MetalsEnrichments._
+
+/**
+ * Client implementation of how to interpret `ResourceOperation` from LSP, used for testing purposes.
+ */
+object ResourceOperations {
```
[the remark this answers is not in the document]
Giggiux: Yup! I didn't realize there was still the one in mtest 😳

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala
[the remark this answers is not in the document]
tgodzik: What do you mean? I just thinking that we should be able to extract also unrelated objects etc. to a separate file. Anything that resembles a class.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala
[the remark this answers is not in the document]
tgodzik: We could also have a trait that has already the same name as a file, so I am thinking we need to treat all the same. In that case we will need type to just be Tree (or Defn ?).

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala
```diff
@@ -0,0 +1,250 @@
...
+    nodes.toList
+  }
+
+  private def getNewFileTreeStructure(
+      tree: Tree,
+      range: l.Range,
+      currFileName: String
+  ): Option[(Tree, Defn.Class)] = {
+
+    val sealedDefn = tree.collect {
```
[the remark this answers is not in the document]
Giggiux: I'm extracting the names of the classes and traits that are sealed from the entire file tree: doesn't match check the structure of the entire tree?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala
```diff
@@ -0,0 +1,250 @@
...
+      case node: Defn.Trait if node.mods.exists(_.isInstanceOf[Mod.Sealed]) =>
+        node.name.value
+      case node: Defn.Class if node.mods.exists(_.isInstanceOf[Mod.Sealed]) =>
+        node.name.value
+    }
+
+    def classExtendsSealed(classDefn: Defn.Class): Boolean =
+      getExtendsNames(classDefn).exists(sealedDefn.contains(_))
```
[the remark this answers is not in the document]
Giggiux: As we tested, `C` can extend `B` from another file if `B` is not sealed.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala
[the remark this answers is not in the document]
Giggiux: Ah ok! My bad: I thought you wanted to extract during the "Extract Class" action of a specific class, all the Objects and Traits and Enums in the source file!

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala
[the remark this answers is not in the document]
tgodzik: Ok, maybe let's leave it as is then.

tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
Giggiux: I didn't want to create a new `check` function with duplicated code just to add extra checks at the end in the same workspace, so I added `extraOperations` that are used as following:
```
extraOperations = Some(() => {
        val (fileName, content) = newFile
        val absolutePath = workspace.resolve(getPath(fileName))
        assert(
          absolutePath.exists,
          s"File $absolutePath should have been created"
        )
        assertNoDiff(absolutePath.readText, content)
      })
      ```

File: metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala (hunk 2 above)
Comment (Giggiux) [reply metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala:184]: The problem is that `TextDocumentEdit` requires a `VersionedTextDocumentIdentifier`, and the `WorkspaceEdit` requires a list of `List[Either[l.TextDocumentEdit, l.ResourceOperation]]`, so I was obliged to use `VersionedTextDocumentIdentifier`. 
But if there is another way, I'm very happy to change it! :)

File: metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala [hunk 14]
```diff
@@ -0,0 +1,266 @@
...
+    }
+    traverser(tree)
+
+    nodes.toList
+  }
+
+  private def sealedNames(tree: Tree) = tree.collect {
+    case node: Defn.Trait if node.mods.exists(_.isInstanceOf[Mod.Sealed]) =>
+      node.name.value
+    case node: Defn.Class if node.mods.exists(_.isInstanceOf[Mod.Sealed]) =>
```
tgodzik, line 95: Can enum be extended and sealed?

tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
dos65: Its type might be simplified to
```scala
extraOperations: => Unit = ()
```

dos65 (commented):
Nice work!
Just left a couple of minor comments

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala
```diff
@@ -0,0 +1,266 @@
...
+          val canExtractClass = differentNames && definitions.size > 1 &&
+            notExtendsSealedOrSealed
+
+          // This check is here only to avoid doing more computations in case of a single class in the file
+          if (canExtractClass) {
+            val classDefn = classAtCursor.get
+            val newTree =
+              getNewFileTreeStructure(tree, range, classDefn, fileName)
+            extractClassAction(uri, newTree, classDefn) :: Nil
+          } else if (canRenameClass) {
```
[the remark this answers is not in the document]
Giggiux: I believe the code is way cleaner as suggested! I will rewrite it this way!

tests/unit/src/main/scala/tests/ResourceOperations.scala
```diff
@@ -0,0 +1,80 @@
...
+   * there is no guarantee that is compliant with the LSP specs
+   * @param operation CreateFile operation
+   */
+  def createFile(operation: CreateFile): Unit = {
+    val uri = operation.getUri
+    val options = Option(operation.getOptions)
+
+    val overwrite: Boolean =
+      options
+        .map(opt => opt.getOverwrite: Boolean)
```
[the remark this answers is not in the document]
Giggiux: Without it, since the `.getOverwrite` method returns a java.lang.Boolean, the return type of the whole `overwrite` is `Any`. With the `: Boolean` there, there is some sort of implicit conversion that makes the `overwrite` a scala Boolean. Not so sure what's causing the error, but it seems to work.
Happy to change it in something more "readable"!

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractClass.scala
Giggiux: Scala 2 enums can be sealed being just objects with Enumeration, but Scala 3 enums cannot. 
The latest commit should consider this.

tests/unit/src/main/scala/tests/ResourceOperations.scala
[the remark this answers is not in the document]
dos65: Thx for the explanation!

I think that's ok - I have not idea how to make it better. The only alternative here is to call map as `.map[Boolean](..` but I don't see much difference here

dos65 (changes requested):
@Giggiux I've tried this branch locally and there is an issue with the newly created file. It duplicates a package.

I've tried it on this file: `mtags/src/main/scala/scala/meta/internal/mtags/GlobalSymbolIndex.scala`

It has the following class: 
```scala
...
case class SymbolDefinition(
    querySymbol: Symbol,
    definitionSymbol: Symbol,
    path: AbsolutePath
)
```

When I apply `extract class` it creates a file with the following content:
```scala
package scala.meta.internal.mtags

package scala.meta.internal.mtags // <- package stm is duplicated
import scala.meta.Dialect // <- no newline
import scala.meta.internal.mtags
import scala.meta.io.AbsolutePath // <- no newline
case class SymbolDefinition(querySymbol: Symbol, definitionSymbol: Symbol, path: AbsolutePath)
```

Giggiux:
@dos65 the spacing is intended, because when I copy the `Tree` elements it doesn't copy also the indentation lines,  while the double package definition, obviously should not happen.

Is it possible that it's conflicting with some other metals/vscode functionality that automatically puts the package definition on the first line in new files?
I'm thinking this may be the issue because I've tried the exact same file and doing `ctrl + z` on it, first removes the `package` definition and then reverts the code action, while in my tests it only reverted the code action.

dos65:
@Giggiux yep - there is a thing that automatically appends package value into new file.
I'm not sure but probably it happens in `textDocument/didOpen` + `packageProvider.workspaceEdit`.

Giggiux:
I believe what happens is that when the file is created by the WorkspaceEdit using the ExtractClass code action, it automatically opens it and then it copies the new content inside. 
In that little time between opening and adding the content, the `textDocument/didOpen` request is performed and here the `packageProvider` checks if the size of the file is 0 (`if (path.isScala && path.toFile.length() == 0)`) and, being done in kinda parallel to the ExtractClass workspace edit it checks true and also applies the workspace edit to add the package statement as if the file was empty.

The issue is that I'm not 100% sure it happens always in the same order, being a race condition it also generated something like this to me:

```scala
package scala.meta.internal.mtags
import scala.meta.Dialect
import scala.meta.internal.mtags
case class SymbolDefinition(querySymbol: Symbol, definitionSymbol: Symbol, path: AbsolutePath)package scala.meta.internal.mtags
```

I would just remove the first package definition that we find on extracting, but I'm not sure is the right behavior in this case.

tgodzik:
Does it happen always? Maybe we can just leave out the package definition in that  case if it matches the directory structure? Only add anything that does not match it?

tgodzik:
Also, there is a way to implement the `packageProvider` mechanism better with the new `DidCreateFile`, which might be helpful here.

dos65:
I'm wondering if we might solve this problem by creating  `CodeAction` in a different way without using `CreateFile`.
Would it work with one `TextEdit` for the current file and custom `Command` that creates a new file with predefined content?

Giggiux:
> Would it work with one `TextEdit` for the current file and custom `Command` that creates a new file with predefined content?

Yes it would work and would essentially solve the issue, but thinking about it: isn't the automatic package provision on creating an empty scala file a little bit invasive? 
As far as I can tell, when one creates a new Scala file it's has the option to choose which kind of file wants, and it seems to automatically use a template that has already the correct package definition. 
This means that the current functionality on `didOpen` works only with new scala files that the user voluntarily chose to create without template (no class, no case class, no trait, etc...).  (Side note: testing the functionality appears that if an empty file is created with the same name of a previous file, in the `didOpen` it will not add the package, and I believe it's due to not being removed onDelete from the `ignorePackageFiles`). 
Wouldn't be better to use something else to solve this particular situation, so that the user can actually create an empty file without the package definition?

dos65:
@Giggiux 
I think there are a lot of people who get used to this feature with automatic package addition and we had to keep it as is.
`NewScalaFile` is different from UX viewpoint.

Giggiux:
@dos65 with the latest changes the double package error should be fixed! 
I've implemented it with a new ServerCommand, so it requires a new revision.

There is just one thing I prefer about the previous method:  UX.
Using `WorkspaceEdit`s with `ResourceOperation`s to create a new file and editing it to add the content made it so that Visual Studio Code (and I believe any other client) would automatically open the new file for the user.
With the current method it doesn't happen. Is there any way to "enforce" the opening of the file after the creation?

dos65:
@Giggiux 
> Is there any way to "enforce" the opening of the file after the creation?

I haven't seen anything like that in the current codebase.
But I checked lsp-spec and there is [Show Document Request](https://microsoft.github.io/language-server-protocol/specification#window_showDocument) that should fit for this purpose.

Giggiux:
> New in version 3.16.0
🥲 I would've loved to use also `codeAction/resolve` instead of the ServerCommand, but 3.16 is not yet supported by all clients, so @tgodzik suggested to not use its features yet

dos65:
@Giggiux Just checked it locally - it works! Cool!

The last thing I think is required is having newlines after packageDef and import block.
Otherwise, it would be annoying to use this action. It offers automatic extraction into a new file but you anyway have to edit it manually. I believe it isn't hard to implement.

tgodzik:
> With the current method it doesn't happen. Is there any way to "enforce" the opening of the file after the creation?

Take a look at the NewFileProvider, we invoke go to location command there that will open up the new document.

ckipp01:
Really cool @Giggiux! I'll throw a few things here as I'm trying this out tonight. One things I noticed which may be unexpected to the user or cause issues is the following:

```scala
package example

object Main extends App {
  println("hi")
}

case class Foo(a: Int)

object Foo {
  def a = 3
}
```

If I trigger the code action on `Foo` right now it will either bring only the case class or the object over to a new file. Since it doesn't automatically open it up atm, this will actually break compilation since now there will be a `Foo` in here and also a `Foo` in `Foo.scala` in the same package. Would it be doable to determine if there is a companion object and bring them both over?

tests/unit/src/test/scala/tests/codeactions/ExtractRenameMemberLspSuite.scala
```diff
@@ -0,0 +1,568 @@
...
+  checkExtractedMember(
+    "extract-class-inner-subpackage",
+    """|package a
+       |
+       |case class A()
+       |
+       |package b {
+       |  case class B()
+       |  package c {
+       |case class <<C>>()
```
tgodzik, line 376:
```suggestion
       |    case class <<C>>()
```

tests/unit/src/test/scala/tests/codeactions/ExtractRenameMemberLspSuite.scala
```diff
@@ -0,0 +1,568 @@
...
+       |package b {
+       |  case class B()
+       |  package c {
+       |
+       |  }
+       |}
+       |""".stripMargin,
+    newFile = (
+      "C.scala",
+      s"""|package a
```
tgodzik, line 394: Can we and should we change it to `a.b.c` ?

tests/unit/src/test/scala/tests/codeactions/ExtractRenameMemberLspSuite.scala
```diff
@@ -0,0 +1,568 @@
...
+  )
+
+  checkExtractedMember(
+    "extract-class-inner-package-no-other-packages",
+    """|package a
+       |
+       |case class A()
+       |
+       |package b {
+       |case class <<B>>()
```
tgodzik, line 410:
```suggestion
       |  case class <<B>>()
```

tests/unit/src/test/scala/tests/codeactions/ExtractRenameMemberLspSuite.scala
```diff
@@ -0,0 +1,568 @@
...
+          |package b
+          |package c
+          |
+          |case class C()
+          |""".stripMargin
+    )
+  )
+
+  checkExtractedMember(
+    "extract-class-inner-package-no-other-packages",
```
tgodzik, line 404: Also, do we need this test? It seems to duplicate much of the other tests. I would honestly minimize the number of LSP tests as previously mentioned and just test the CodeAction itself for unit tests/

tgodzik (changes requested):
The PR already looks pretty cool, but we can improve some things, simplify it a little.

I would also remove some of the tests, they aren't really testing different branches and I would opt for a more Unit style tests if you want to include everything. They run faster (we already take a while to run all tests) and are much easier to write.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
```diff
@@ -0,0 +1,338 @@
...
+          super.apply(s)
+        case _ =>
+      }
+    }
+    traverser(tree)
+
+    nodes.toList
+  }
+
+  private def sealedNames(tree: Tree) = tree.collect {
```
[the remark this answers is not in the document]
Giggiux: The issue here is that `collect` signature is:
```scala
def collect[T](fn: PartialFunction[Tree, T]): List[T]
```

By using the suggested version changing it using a PF we'd have a `List[Boolean]`, that in the `notSealed` function  would not work to get the `memberExtendsSealed`, because it uses the sealed nodes' `name.value` to compare it with the `inits` (extends and with) of the candidate `Member`
But the `isSealed` function would definitely work to avoid code repetition here and in the next part of the code

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
```diff
@@ -0,0 +1,338 @@
...
+        definitions = membersDefinitions(tree)
+        memberDefn <- definitions.find(_.name.pos.toLSP.overlapsWith(range))
+        companion = definitions.find(isCompanion(memberDefn))
+        newFileContent = getNewFileContent(
+          tree,
+          range,
+          memberDefn,
+          companion
+        )
+        newFileUri = newURIFromClass(uri, memberDefn)
```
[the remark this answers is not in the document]
Giggiux: Actually, I moved that check even before the codeAction is proposed to the user.
Should I also testi it there anyway, due to the asynchronous nature of the command?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
```diff
@@ -0,0 +1,338 @@
...
+  }
+}
+
+object ExtractRenameMember {
+  def getMemberType(member: Member): String = member match {
+    case _: Defn.Class => "class"
+    case _: Defn.Enum => "enum"
+    case _: Defn.Trait => "trait"
+    case _: Defn.Object => "object"
+    case _ => "unknown type"
```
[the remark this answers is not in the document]
Giggiux: This function is used essentially only to get the right type in the CodeAction Title, and the Code Action itself cannot be created if the `Member` is not one from `class`/`enum`/`trait`/`object`, this "match map" is there just to avoid having "Extract member X from file Y.scala`, so even using an option at a certain point we would need to get it out of it and return some other string to "fill the space", or not showing anything in the CodeAction Title.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1763,13 +1775,74 @@ class MetalsLanguageServer(
...
+            if edits.nonEmpty
+            workspaceEdit = new l.WorkspaceEdit(
+              Map(uri -> edits.asJava).asJava
+            )
+            applyWorkSpaceEditParams = new ApplyWorkspaceEditParams(
+              workspaceEdit
+            )
+          } yield languageClient
+            .applyEdit(applyWorkSpaceEditParams)
+            .asScala
```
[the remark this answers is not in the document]
Giggiux: It could, but the next map uses `newFileUri` and `range` from `ExtractMemberDefinitionResult`, and instead of returning an Option with those data, having to then move again form Option and Futures, I believe it can be better to just map into the new future that returns unit, but any suggestion how to avoid having to pass around `newFileUri` and `range` is welcome

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -46,4 +47,19 @@ final class CodeActionProvider(
...
 
+  def executeCommands(
+      codeActionCommandData: CodeActionCommandData,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[Option[CodeActionCommandResult]] = {
+    val basePF: PartialFunction[CodeActionCommandData, Future[
+      Option[CodeActionCommandResult]
+    ]] =
+      PartialFunction.empty
+    val pf =
```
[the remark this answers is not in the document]
Giggiux: do you mean `pf` or `basePF`?
In case of `pf` it uses the CancelToken that passes to the CodeActions' executeCommand, so I don't think it can be a private value, rather a private definition, but that would not change anything
`basePF` makes sense to have it as a private value, so that is not recreated every time

tests/unit/src/test/scala/tests/codeactions/ExtractRenameMemberLspSuite.scala
```diff
@@ -0,0 +1,568 @@
...
+
+  check(
+    "extract-class-only-source-file",
+    """|package a
+       |
+       |case class A()
+       |case class <<B>>()
+       |""".stripMargin,
+    s"${ExtractRenameMember.title("class", "B")}",
+    """|package a
```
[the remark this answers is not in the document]
Giggiux: Exactly, I'll just leave the more complete one!

tests/unit/src/test/scala/tests/codeactions/ExtractRenameMemberLspSuite.scala
```diff
@@ -0,0 +1,568 @@
...
+       |package b {
+       |case class <<B>>()
+       |}
+       |""".stripMargin,
+    s"""|${ExtractRenameMember.title("class", "B")}""".stripMargin,
+    """|package a
+       |
+       |case class A()
+       |
+       |package b {
```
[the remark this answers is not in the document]
Giggiux: I'll do it in the next commit!

tests/unit/src/test/scala/tests/codeactions/ExtractRenameMemberLspSuite.scala
Giggiux: Next commit will have this! :)

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
```diff
@@ -0,0 +1,399 @@
...
+        definitions match {
+          case Nil => Nil
+          case head :: Nil if canRenameDefn(head) =>
+            definitions.map(renameFileAsMemberAction(uri, _))
+          case _ =>
+            defnAtCursor
+              .filter(defn => canExtractDefn(defn))
+              .map { classDefn =>
+                extractClassAction(uri, classDefn)
+              }
```
ckipp01, line 81: Nit you can just use `collect` here.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1763,13 +1775,62 @@ class MetalsLanguageServer(
...
             Future.unit
           }.withObjectValue
         }
+      case ServerCommands.ExtractMemberDefinition() =>
+        def applyEdits(
+            uri: String,
+            data: CodeActionCommandResult
+        ): Future[Unit] = {
+          data match {
+            case ExtractMemberDefinitionResult(edits, newFileUri, range)
```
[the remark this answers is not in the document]
tgodzik: We should have just one result, no need for more really. I think this was actually changed based on my previous comments, but not pushed.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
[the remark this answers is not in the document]
tgodzik: I would just not show the code action at all if cannot create a title. In that case it means something went really wrong.

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -46,4 +51,14 @@ final class CodeActionProvider(
...
+  def executeCommands(
+      codeActionCommandData: CodeActionCommandData,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Option[Future[CodeActionCommandResult]] = {
+    allActions
+      .foldRight(basePF)((codeAction, pf) =>
+        codeAction.executeCommand(token) orElse pf
+      )
+      .lift(codeActionCommandData)
+  }
```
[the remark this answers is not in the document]
tgodzik: That was kind of my suggestion, but might be overenginnering. We could just replace it by a single match, something along the lines of:
```scala

codeActionCommandData match{
  case data: ExtractMemberDefinition =>
    extractMemberAction.runCommand(data)
}
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/CodeActionCommandData.scala
```diff
@@ -0,0 +1,26 @@
...
+import org.eclipse.lsp4j.Range
+import org.eclipse.lsp4j.TextDocumentPositionParams
+import org.eclipse.lsp4j.TextEdit
+
+trait CodeActionCommandData {
+  val uri: String
+  val actionType: String
+}
+
+trait CodeActionCommandResult {
```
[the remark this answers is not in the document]
Giggiux: should be fixed in latest commit

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
```diff
@@ -0,0 +1,387 @@
...
+      newFilePath,
+      fileContent,
+      memberDefn,
+      companion,
+      newFileMemberRange
+    )
+
+    // from checks already done when proposing the CodeAction
+    // we know that the data in the Option will exist
+    val (
```
tgodzik, line 309: Can you not do this all in the yield block?

```scala
val opt = for {
      tree <- trees.get(path)
      definitions = membersDefinitions(tree)
      memberDefn <- definitions.find(_.name.pos.toLSP.overlapsWith(range))
      companion = definitions.find(isCompanion(memberDefn))
      fileContent = newFileContent(
        tree,
        range,
        memberDefn,
        companion
      )
      newFilePath = newPathFromClass(uri, memberDefn)
      if !newFilePath.exists
      newFileMemberRange = new l.Range()
    } yield {
      val newFileUri = newFilePath.toURI.toString

      extractClassCommand(
        newFileUri,
        fileContent,
        memberDefn,
        companion
      )
    }
```

you can later do `opt.toIterable.flatten`

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
```diff
@@ -0,0 +1,387 @@
...
+      uri: String,
+      member: Member,
+      title: String
+  ): l.CodeAction = {
+
+    val range = member.name.pos.toLSP
+
+    val codeAction = new l.CodeAction()
+    codeAction.setTitle(title)
+    codeAction.setKind(kind)
```
tgodzik, line 250:
```suggestion
    codeAction.setKind(l.CodeActionKind.RefactorExtract)
```
maybe inline it since we have two different action produced in the class.

tests/unit/src/test/scala/tests/codeactions/ExtractRenameMemberLspSuite.scala
```diff
@@ -0,0 +1,474 @@
...
+        codeActions <- codeActionFut
+        _ = {
+          if (codeActionExpected) assert(codeActions.nonEmpty)
+          else assert(codeActions.isEmpty)
+        }
+      } yield ()
+    }
+
+  // Same structure of path as in function `check`
+  private def getPath(name: String) = s"a/src/main/scala/a/$name"
```
tgodzik, line 473:
```suggestion
  private def testedFilePath(name: String) = s"a/src/main/scala/a/$name"
```

tests/unit/src/test/scala/tests/codeactions/ExtractRenameMemberLspSuite.scala
```diff
@@ -0,0 +1,474 @@
...
+        assert(
+          !oldAbsolutePath.exists,
+          s"File $oldAbsolutePath should have been renamed"
+        )
+      }
+    )
+
+  }
+
+  def checkNoLSP(
```
tgodzik, line 423:
```suggestion
  def checkActionProduced(
```
maybe?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1763,13 +1773,42 @@ class MetalsLanguageServer(
...
           }.withObjectValue
         }
+      case ServerCommands.ExtractMemberDefinition() =>
+        CancelTokens.future { token =>
+          val args = params.getArguments().asScala
+
+          val futureOpt = for {
+            (params, uri) <- textDocumentPosition(args)
+            data = ExtractMemberDefinitionData(uri, params)
+            result <- codeActionProvider.executeCommands(data, token)
```
tgodzik, line 1790: Let's return a future here, but make it `Future.failed()` instead of None. We should almost never have this situation so it's best to have it fail explicitely.

This way we could do:
```scala
val futureOpt = for {
  (params, uri) <- textDocumentPosition(args)
} yield {
  val data = ExtractMemberDefinitionData(uri, params)
  // future for
  for {
    result <- codeActionProvider.executeCommands(data, token)
    future <- languageClient.applyEdit(result.edits)
  } yield {
    result.goToLocation.foreach { location =>
      languageClient.metalsExecuteClientCommand(
        new ExecuteCommandParams(
          ClientCommands.GotoLocation.id,
          List(location: Object).asJava
        )
      )
    }
  }
}

```

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
```diff
@@ -0,0 +1,387 @@
...
+        )
+      )
+    )
+
+    codeAction
+  }
+
+  def executeCommand(
+      data: ExtractMemberDefinitionData
+  ): CodeActionCommandResult = {
```
tgodzik, line 266:
```suggestion
  ): Future[CodeActionCommandResult] = Future {
```
There is a bit of things happening here so let's not run it on the main thread.

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -46,4 +50,14 @@ final class CodeActionProvider(
     Future.sequence(actions).map(_.flatten)
   }
 
+  def executeCommands(
+      codeActionCommandData: CodeActionCommandData,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Option[CodeActionCommandResult] = {
```
tgodzik, line 56:
```suggestion
  )(implicit ec: ExecutionContext): Future[CodeActionCommandResult] = {
```

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -46,4 +50,14 @@ final class CodeActionProvider(
...
+    codeActionCommandData match {
+      case data: ExtractMemberDefinitionData =>
+        Option(extractMemberAction.executeCommand(data))
```
tgodzik, line 59:
```suggestion
        extractMemberAction.executeCommand(data)
```

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -46,4 +50,14 @@ final class CodeActionProvider(
...
+      case _ => None
```
tgodzik, line 60:
```suggestion
      case data => Future.failed(new CodeActionCommandNotFound(data.title))
```

we can create an additional exception. This should never happen, but if ever we do something wrong we can easily figure out what's happening instead of having it just skipped.

tgodzik (changes requested):
We're almost there! There are only a couple of smaller things I would fix.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
```diff
@@ -0,0 +1,387 @@
...
+        val fileName = uri.toAbsolutePath.filename.replaceAll("\\.scala$", "")
+
+        val definitions = membersDefinitions(tree)
+        val sldNames = sealedNames(tree)
+        val defnAtCursor =
+          definitions.find(_.name.pos.toLSP.overlapsWith(range))
+
+        def canRenameDefn(defn: Member): Boolean = {
+          val differentNames = defn.name.value != fileName
+          val notGetsExtendedByFileName =
```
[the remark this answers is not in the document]
Giggiux: We should also probably check if the file already exists, even if it the rename would fail anyway I can see that it's better to not even give the code action option!
This check was introduced because when we were checking only classes, and not also traits/objects/enums, there could be a trait that had the same name as the file extended by a class. But now the `canRenameDefn` check is made only if there is exactly one member in the file, so all those checks are indeed useless. I will change them with the file exists! :)

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
Giggiux: I need to extract also the `newFileUri`, would something like this be acceptable?

```scala
yield {
      val newFileUri = newFilePath.toURI.toString
      (extractClassCommand(
        newFileUri,
        fileContent,
        memberDefn,
        companion
      ), newFileUri)
    }

    val edits = opt.map(_._1).toList.flatten
    val newFileUri = opt.map(_._2).get
    ```

Comment (tgodzik) [reply metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala:309]: I would put everything into the yield then and return CodeActionCommandResult from it. We can the treat None in this case Future.failed

File: metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala [hunk 42]
```diff
@@ -46,4 +50,14 @@ final class CodeActionProvider(
...
   }
 
+  def executeCommands(
+      codeActionCommandData: CodeActionCommandData,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[CodeActionCommandResult] = {
+    codeActionCommandData match {
+      case data: ExtractMemberDefinitionData =>
+        extractMemberAction.executeCommand(data)
+      case _ => Future.failed(new IllegalArgumentException)
```
tgodzik, line 60:
```suggestion
      case data => Future.failed(new IllegalArgumentException(data.toString))
```
?

Otherwise we'll need to dig in into what was sent.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
```diff
@@ -0,0 +1,380 @@
...
+      val newFileMemberRange = new l.Range()
+      val workspaceEdit = new WorkspaceEdit(Map(uri -> edits.asJava).asJava)
+      CodeActionCommandResult(
+        new ApplyWorkspaceEditParams(workspaceEdit),
+        Option(new Location(newFileUri, newFileMemberRange))
+      )
+    }
+
+    opt.getOrElse(
+      throw CodeActionCommandNotFoundException("CodeAction Command not found")
```
tgodzik, line 318:
```suggestion
      throw CodeActionCommandNotFoundException("Could not execute command ${data.actionType}")
```

tgodzik (changes requested):
Just two minor changes from me. @ckipp01 should we merge the PR?

tgodzik (approved):
LGTM!

dos65 (approved):
:rocket: :rocket: :rocket:

ckipp01 (approved):
:shipit: 

Great job @Giggiux!

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -46,4 +50,14 @@ final class CodeActionProvider(
     Future.sequence(actions).map(_.flatten)
   }
 
+  def executeCommands(
+      codeActionCommandData: CodeActionCommandData,
+      token: CancelToken
+  )(implicit ec: ExecutionContext): Future[CodeActionCommandResult] = {
```
tgodzik, line 56:
```suggestion
  ): Future[CodeActionCommandResult] = {
```

nipick, this can be removed.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractRenameMember.scala
```diff
@@ -0,0 +1,419 @@
...
+  private def sealedNames(tree: Tree): List[String] = {
+    def completeName(node: Member): String = {
+      def completePreName(node: Tree): List[String] = {
+        node.parent match {
+          case Some(t) =>
+            t match {
+              case o: Defn.Object => o.name.value :: completePreName(o)
+              case po: Pkg.Object => po.name.value :: completePreName(po)
+              case _: Source => Nil
+              case _ => completePreName(t)
```
tgodzik, line 132: That would amount to:
```suggestion
              case _ => Nil
```
no?

tgodzik (approved):

gabro:
Nice!! 👏

PR #2730, opened 2021-04-21
Remove test reports

tgodzik:
It seems that test overview isn't working even after my recent changes and I don't feel like report plugin is mature enough to use. It would be cool to figure out where to keep those reports, since Google Cloud currently needs payment.


ckipp01 (approved):
I don't really ever view them, so I've find with removing them. 👍🏼

dos65 (approved):
To be honest, I did not notice that there is such a thing in build

PR #2731, opened 2021-04-21
Remove munit sbt plugin to fix CI

tgodzik:
I previously removed a bunch of settings, but I forgot about the munit-sbt plugin, which now fails on CI.

Most recent failures: https://github.com/scalameta/metals/actions/runs/771330085


ckipp01 (approved):
Ah good catch. LGTM!

PR #2734, opened 2021-04-22
Add in the parentUri when offering the newSymbol code action.

ckipp01:
Most clients don't have an issue with this, but I just hit on a case
where the server providing `null` here is causing issues when the client
then tries to encode the value to send back and it chokes on it. In this
specific case, we do this step later if it's passed in as null, but
since we already have what we need to fill in the parentUri where the
new file will be created, we just do that right away instead of passing
in `null`.


tgodzik (approved):
LGTM!

PR #2735, opened 2021-04-22
Support 3.16.0 lsp specification with compatibility for non `array`/`…

Giggiux:
…object` notification parameters

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -988,52 +988,78 @@ class MetalsLanguageServer(
   }
 
   @JsonNotification("metals/didFocusTextDocument")
-  def didFocus(uri: String): CompletableFuture[DidFocusResult.Value] = {
-    val path = uri.toAbsolutePath
-    focusedDocument = Some(path)
-    buildTargets
-      .inverseSources(path)
-      .foreach(focusedDocumentBuildTarget.set)
+  def didFocus(
+      params: AnyRef
+  ): CompletableFuture[DidFocusResult.Value] = {
 
-    // unpublish diagnostic for dependencies
-    interactiveSemanticdbs.didFocus(path)
-    // Don't trigger compilation on didFocus events under cascade compilation
-    // because save events already trigger compile in inverse dependencies.
-    if (path.isDependencySource(workspace)) {
-      CompletableFuture.completedFuture(DidFocusResult.NoBuildTarget)
-    } else if (openedFiles.isRecentlyActive(path)) {
-      CompletableFuture.completedFuture(DidFocusResult.RecentlyActive)
-    } else {
-      syntheticsDecorator.publishSynthetics(path)
-      worksheetProvider.onDidFocus(path)
-      buildTargets.inverseSources(path) match {
-        case Some(target) =>
-          val isAffectedByCurrentCompilation =
-            path.isWorksheet ||
-              buildTargets.isInverseDependency(
-                target,
-                compilations.currentlyCompiling.toList
-              )
-          def isAffectedByLastCompilation: Boolean =
-            !compilations.wasPreviouslyCompiled(target) &&
-              buildTargets.isInverseDependency(
-                target,
-                compilations.previouslyCompiled.toList
-              )
-          val needsCompile =
-            isAffectedByCurrentCompilation || isAffectedByLastCompilation
-          if (needsCompile) {
-            compilations
-              .compileFile(path)
-              .map(_ => DidFocusResult.Compiled)
-              .asJava
-          } else {
-            CompletableFuture.completedFuture(DidFocusResult.AlreadyCompiled)
-          }
-        case None =>
-          CompletableFuture.completedFuture(DidFocusResult.NoBuildTarget)
-      }
+    val uriOpt: Option[String] = params match {
+      case string: String =>
+        Option(string)
+      case (h: String) :: Nil =>
+        Option(h)
+      case _ =>
+        scribe.warn(
+          s"Unexpected notification params received for didFocusTextDocument: $params"
+        )
+        None
     }
+
+    uriOpt match {
+      case Some(uri) => {
+        val path = uri.toAbsolutePath
+        focusedDocument = Some(path)
+        buildTargets
+          .inverseSources(path)
+          .foreach(focusedDocumentBuildTarget.set)
+
+        // unpublish diagnostic for dependencies
+        interactiveSemanticdbs.didFocus(path)
+        // Don't trigger compilation on didFocus events under cascade compilation
+        // because save events already trigger compile in inverse dependencies.
+        if (path.isDependencySource(workspace)) {
+          CompletableFuture.completedFuture(DidFocusResult.NoBuildTarget)
+        } else if (openedFiles.isRecentlyActive(path)) {
+          CompletableFuture.completedFuture(DidFocusResult.RecentlyActive)
+        } else {
+          syntheticsDecorator.publishSynthetics(path)
+          worksheetProvider.onDidFocus(path)
+          buildTargets.inverseSources(path) match {
+            case Some(target) =>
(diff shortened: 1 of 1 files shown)


tgodzik:
> Thanks for the pr @Giggiux. Could you expand a bit more on the purpose for this?
> 
> This is changing the expected param of the `metals/didFocusTextDocument` notification, which is a custom Metals LSP extension, so the change to 3.16.0 shouldn't really matter here. What is the use case from the client perspective where a list of URIs would be sent for `metals/didFocusTextDocument` instead of just the URI that is being focused? If we were to make this change we'd also need to actually change the [docs](https://scalameta.org/metals/docs/integrations/new-editor.html#metalsdidfocustextdocument) to reflect it, but it'd help to get some clarity on the use case first.

A bit of context behind this. With the update to LSP 3.16.0 in VS Code, we can no longer send simple string to the server, it will be packed into array automatically:

https://github.com/microsoft/vscode-languageserver-node/blob/d58c00bbf8837b9fd0144924db5e7b1c543d839e/jsonrpc/src/common/connection.ts#L1071

And that is actually consistent with the specification. You can only send an object or an array:

https://microsoft.github.io/language-server-protocol/specifications/specification-current/#notificationMessage

And Metals will throw an exception here, which is why I reverted the change in metals-vscode. But! Thanks to this hack, we should be able to accept both String and an Array. So other clients could just send a string as previously done.

I do agree that we should change the specification (Array of string?), because the current one is actually incompatible with LSP.

Giggiux:
Yes, absolutely! Sorry @ckipp01  about the non descriptive PR, It has been a long day trying to debug and fix this, and I've been lazy on the description 🤦‍♂️ 

So the general idea is to update to 3.16.0 to start supporting [code action resolve requests](https://microsoft.github.io/language-server-protocol/specification#codeAction_resolve), this was supposed to be done by [upgrading to vscode-languageserver-protocol to v.3.16.0](https://github.com/scalameta/metals-languageclient/pull/264) in scalameta/metals-languageclient, and [updating dependencies](https://github.com/scalameta/metals-vscode/pull/541) in scalameta/metals-vscode. 
Those two PR were merged, but then @tgodzik realized there was an error in the `metals/didFocusTextDocument` request (and reverted the changes in scalameta/metals-vscode@371dbe55cb2a13796c40448e7c135eb64cb80898 ):

```
Apr 22, 2021 6:48:07 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint logIssues
WARNING: Issue found in NotificationMessage: Message could not be parsed.
com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected a string but was BEGIN_ARRAY at line 1 column 67 path $.params
	at com.google.gson.Gson.fromJson(Gson.java:944)
	at org.eclipse.lsp4j.jsonrpc.json.adapters.MessageTypeAdapter.fromJson(MessageTypeAdapter.java:329)
	at org.eclipse.lsp4j.jsonrpc.json.adapters.MessageTypeAdapter.parseParams(MessageTypeAdapter.java:249)
	at org.eclipse.lsp4j.jsonrpc.json.adapters.MessageTypeAdapter.read(MessageTypeAdapter.java:119)
	at org.eclipse.lsp4j.jsonrpc.json.adapters.MessageTypeAdapter.read(MessageTypeAdapter.java:55)
	at com.google.gson.Gson.fromJson(Gson.java:932)
	at org.eclipse.lsp4j.jsonrpc.json.MessageJsonHandler.parseMessage(MessageJsonHandler.java:119)
	at org.eclipse.lsp4j.jsonrpc.json.MessageJsonHandler.parseMessage(MessageJsonHandler.java:114)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:193)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:94)
	at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:113)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.IllegalStateException: Expected a string but was BEGIN_ARRAY at line 1 column 67 path $.params
	at com.google.gson.stream.JsonReader.nextString(JsonReader.java:826)
	at com.google.gson.internal.bind.TypeAdapters$16.read(TypeAdapters.java:402)
	at com.google.gson.internal.bind.TypeAdapters$16.read(TypeAdapters.java:390)
	at com.google.gson.Gson.fromJson(Gson.java:932)
	... 15 more
```

This error was thrown when [this request happened](https://github.com/scalameta/metals-vscode/blob/63bcf5e8eb056f12cf14c36dc68b8179790736b7/src/extension.ts#L807).

```ts
client.sendNotification(
            MetalsDidFocus.type,
            editor.document.uri.toString()
          );
```

The notification sent is of type `string`.
According to [NotificationMessage's LSP specification](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#notificationMessage) the message can either be an `array` or an `object`, but since version 3.16.0 of lsp in vscode ( microsoft/vscode-languageserver-node@a93e8b5ee26b7fb28245797adbdb0da585c0fb56 ) the specification was never enforced.
From version 3.16.0, the single parameter that is not an `array` or `object` is automatically converted in an `array`, as [from source code](https://github.com/microsoft/vscode-languageserver-node/blob/faa68d90a976f158a8403a0f5cce54117ce61dbb/jsonrpc/src/common/connection.ts#L1071)

So this PR is to support the latest LSP on metals side, without losing "retro compatibility" of just `string` notifications, that probably still works for other clients.

I hope that's clear enough, it's a lot to summary.

ckipp01:
Ahhhh, gotcha! Thank you for the detailed explanation! 🙏🏼 I'm actually surprised we haven't hit on this before.

> So this PR is to support the latest LSP on metals side, without losing "retro compatibility" of just string notifications, that probably still works for other clients.

Honestly this probably isn't that big of a deal. It is a nice to have, but at the same time it should be pretty easy to fix on the client side, and then it forces others to be inline with the spec.

Ha, sorry this update to 3.16.0 has been so much trouble 😆

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -988,52 +988,75 @@ class MetalsLanguageServer(
...
+      params: AnyRef
+  ): CompletableFuture[DidFocusResult.Value] = {
+
+    val uriOpt: Option[String] = params match {
+      case string: String =>
+        Option(string)
+      case (h: String) :: Nil =>
+        Option(h)
+      case _ =>
+        None
```
ckipp01, line 1001: Only suggestion is _maybe_ to log something out here if we actually hit this. Since like you said an `Object` is actually valid as a notification. Actually even more then that if this is `nil` by chance for some reason from the client it might be nice to have something in here like:
```suggestion
      scribe.warn(s"Unexpected notification params received for didFocusTextDocument: $parms")
```
wdyt?

ckipp01 (approved):
Just a super small suggestion. Not a blocker. LGTM! And thanks again for tracking all these down!

Giggiux:
I believe that a good solution to be complaint would be to use directly a class definition for the request params, essentially switching from a string to an object, instead of using the arrays. As it's already done in metals with the `WindowStateDidChangeParams`.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
Giggiux: Added!

PR #2737, opened 2021-04-23
Add extra check before auto-connecting to server.

### LINKED ISSUE #2341: Bloop warning being shown when no build import was done but `.bsp/sbt.json` exists.
## Describe the bug

Metals complains about a missing bloop build after starting up, see **Exhibit A**.

It can also be provoked independently of the import question by telling it to connect to a running build server, see **Exhibit B**.

No connection to the build server is established and Metals does thus not provide any apparent functionality that would require a build definition.

### metals.log

```text
2020.12.25 23:27:47 INFO  Started: Metals version 0.9.8 in workspace '/Users/phi/Documents/workspace/Scala/MetalsBug' for client vscode.
2020.12.25 23:27:49 INFO  time: initialize in 1.5s
2020.12.25 23:27:49 INFO  Attempting to connect to the build server...
Starting the bsp launcher for bloop...
Opening a bsp server connection with 'bsp --protocol local --socket /var/folders/jw/6060chfn35bb7_4l_1w3ss0h0000gn/T/bsp-launcher14851760122142146247/bsp.socket'...
Waiting for the bsp connection to come up...
2020.12.25 23:27:49 WARN  no build target for: /Users/phi/Documents/workspace/Scala/MetalsBug/build.sbt
[0m[32m[D][0m Waiting for a connection at local:///var/folders/jw/6060chfn35bb7_4l_1w3ss0h0000gn/T/bsp-launcher14851760122142146247/bsp.socket...
The server is listening for incoming connections at local:///var/folders/jw/6060chfn35bb7_4l_1w3ss0h0000gn/T/bsp-launcher14851760122142146247/bsp.socket...
Starting thread that pumps stdin and redirects it to the bsp server...
Starting thread that pumps server stdout and redirects it to the client stdout...
2020.12.25 23:27:50 INFO  tracing is enabled: /Users/phi/Library/Caches/org.scalameta.metals/bsp.trace.json
2020.12.25 23:27:50 INFO  time: connected to build server in 0.27s
2020.12.25 23:27:50 INFO  Connected to Build server v1.4.6-15-209c2a5c
2020.12.25 23:27:49 ERROR Empty build targets. Expected at least one build target identifier.
2020.12.25 23:27:49 ERROR Empty build targets. Expected at least one build target identifier.
2020.12.25 23:27:49 ERROR Empty build targets. Expected at least one build target identifier.
2020.12.25 23:27:53 INFO  time: indexed workspace in 3.5s
2020.12.25 23:27:53 WARN  no build target for: /Users/phi/Documents/workspace/Scala/MetalsBug/build.sbt
2020.12.25 23:27:53 INFO  no build target: using presentation compiler with only scala-library
```

## To Reproduce

1. Visual Studio Code is closed.
2. Run this in a terminal outside of Visual Studio Code in the directory of your project:

```zsh
foreach s (.bsp .metals .bloop .vscode .DS_Store target metals.sbt) echo "Deleting $s…"; fd --hidden $s --exec rm -rf {}; end
```

3. If you want to be extra thorough:

```zsh
rm -rfv ~/.sbt/boot ~/.sbt/preloaded # Very optional
```

4. fwiw I happen to have no global sbt settings. In particular, I do not have this set:

```
Global / semanticdbEnabled := true
```

I don't know if it affects anything but felt it necessary to point out since sbt recommends enabling the setting:

```
[warn] Metals requires the semanticdb compiler plugin
[warn] consider setting 'Global / semanticdbEnabled := true' in your global sbt settings ($HOME/.sbt/1.0)
```

but https://scalameta.org/metals/blog/2020/11/06/sbt-BSP-support makes no mention of it.

5. start `sbt`:
```
[info] welcome to sbt 1.4.5 (N/A Java 15.0.1)
[info] loading settings for project global-plugins from sbt-errors-summary.sbt,sbt-updates.sbt ...
[info] loading global plugins from /Users/phi/.sbt/1.0/plugins
[info] loading project definition from /Users/phi/Documents/workspace/Scala/MetalsBug/project
[info] loading settings for project metalsbug from build.sbt ...
[info] set current project to scratchpad (in build file:/Users/phi/Documents/workspace/Scala/MetalsBug/)
[info] sbt server started at local:///Users/phi/.sbt/1.0/server/473b095c179e4dd7e355/sock
[info] started sbt server
sbt:scratchpad>
```

6. Start Visual Studio Code and open this directory. I have done so via `code .`
7. Metals will offer you to import the build, along with an error, shown in **Exhibit A**.

## Expected behavior

It should not show an error.

## Installation

- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: 0.9.8
- Extension version: v1.9.10
- sbt version: 1.4.5 (N/A Java 15.0.1)

## Further information

If, despite the error, Metals is instructed to attempt to generate a bsp configuration for the build tool, after a few moments, a warning is shown as can be seen in **Exhibit C**, indicating that navigation would not be working.

This is the generated `sbt.json` (after formatting):

```json
{
  "name": "sbt",
  "version": "1.4.5",
  "bspVersion": "2.0.0-M5",
  "languages": [
    "scala"
  ],
  "argv": [
    "/usr/local/Cellar/openjdk/15.0.1/libexec/openjdk.jdk/Contents/Home/bin/java",
    "-Xms100m",
    "-Xmx100m",
    "-classpath",
    "/var/folders/jw/6060chfn35bb7_4l_1w3ss0h0000gn/T/metals1547539892474703038/sbt-launch.jar",
    "xsbt.boot.Boot",
    "-bsp"
  ]
}
```

### Exhibit A
image

### Exhibit B
image

### Exhibit C
image

### Directory Structure
```
tree -a
.
├── build.sbt
├── project
│  ├── build.properties
│  └── project
└── src
   └── main
      └── scala
         └── Scratchpad.scala
```

### build.sbt

```
name := "scratchpad"
ThisBuild / organization := "io.doerfler"
ThisBuild / version := "1-SNAPSHOT"
ThisBuild / scalaVersion := "2.13.4"
```

### Scratchpad.scala

```
object Scratchpad extends App {
  println("hello")
}
```

### project/build.properties

```
sbt.version=1.4.5
```

ckipp01:
This fixes #2341, which causes a user to get an error message from Bloop
if the user attempts to use Metals for the first time on an sbt
workspace that already contains a `.bsp/sbt.json` file. The issue is
that Metals would think that since there is a `.bsp/<entry>` already
there, it's auto-connectable. So even though it realizes it should use
Bloop, instead of waiting for the users choice via the prompt, it will
try auto-connecting to Bloop right away causing the error outlined in
the issue.

We know add another check to ensure that if there is a `.bsp/sbt.json`
file in the workspace the first time Metals is opened that we don't
attempt to auto-connect _unless_ there was a specific decision made
(which will be never if it truly is the first time a user is opening
this workspace). Then after that initial open there will either be a
`.bloop/` present making it auto-connectable or a user will use the
`bsp-swich` command,  which saves the choice, and also makes it
auto-connectable which ensure the behavior stays the same for
everything else.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
@@ -24,10 +24,20 @@ import scala.meta.io.AbsolutePath
 final class BuildTools(
     workspace: AbsolutePath,
     bspGlobalDirectories: List[AbsolutePath],
-    userConfig: () => UserConfiguration
+    userConfig: () => UserConfiguration,
+    explicitChoiceMade: Boolean
 ) {
-  def isAutoConnectable: Boolean =
-    isBloop || isBsp
+  // NOTE: We do a couple extra check here before we say a workspace with a
+  // `.bsp` is auto-connectable, and we ensure that a user has explicity chosen
+  // to use another build server besides Bloop or it's a BSP server for a build
+  // tool we don't support. If this isn't done, it causes unexpected warnings
+  // since if a `.bsp/<something>.json` exists before a `.bloop` one does in a
+  // workspace with a build tool we support, we will attempt to autoconnect to
+  // Bloop since Metals thinks it's in state that's auto-connectable before the
+  // user is even prompted.
+  def isAutoConnectable: Boolean = {
+    isBloop || (isBsp && all.isEmpty) || (isBsp && explicitChoiceMade)
+  }
   def isBloop: Boolean = {
     hasJsonFile(workspace.resolve(".bloop"))
   }
@@ -117,6 +127,7 @@ object BuildTools {
     new BuildTools(
       workspace,
       Nil,
-      () => UserConfiguration()
+      () => UserConfiguration(),
+      explicitChoiceMade = false
     )
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -346,7 +346,8 @@ class MetalsLanguageServer(
         buildTools = new BuildTools(
           workspace,
           bspGlobalDirectories,
-          () => userConfig
+          () => userConfig,
+          tables.buildServers.selectedServer().nonEmpty
         )
         fileSystemSemanticdbs = new FileSystemSemanticdbs(
           buildTargets,


tgodzik (approved):
Looks sensible, LGTM!

metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
```diff
@@ -117,6 +125,7 @@ object BuildTools {
     new BuildTools(
       workspace,
       Nil,
-      () => UserConfiguration()
+      () => UserConfiguration(),
+      false
```
ckipp01, line 129: Actually I'll need to change this to `true` in order for the tests to pass. In reality the way we treat Bill is a bit not like we actually treat BSP servers in that we expect to be able to auto-connect to Bill right away without any choice been made. So I'll change this to `true` to "mimic" the actual behaviour that would happen in real life if we expected the ability to auto-connect to a new `.bsp/entry`.

ckipp01: Actually this was incorrect as well. Instead I just needed to add another check. So basically there are 3 things that make the BSP connection auto-connectable:

1. There is already a `.bloop/`
2. There is a `.bsp/<entry>.json` and an explicit choice was made already to use it
3. There is a `.bsp/<entry>.json` and it's a build tool we don' have support for (Like Bill).

Issue #2736, opened 2021-04-22
Wrong import for Akka's Address class


tgodzik:
**Describe the bug**

Decribed by @EveryonePrint-dev:
"
(I see this mostly since either 0.10.1, or 0.10.2)

The code actions show the right completions (imports).
But choosing a completion makes the wrong code change.

I.e.:

This class which not yet has an import: 
```
Address("http", "Test", TestNodeHostName, 1234)
```
shows the right code action as "import" -> "akka.actor.Address".

But when executed the code action the code is changed inline as:

```
Command.akka.actor.Address("http", "Test", TestNodeHostName, 1234)
```
And no import is being made at all.

In other cases a code completion makes an import of a parent package and and replaces the unknown class with a package prefix like:

```
Foo(1, 2, 3)
```
->

```
import some.parent

subparent.Foo(1, 2, 3)
```

Although there is no reason to not just import `import some.parent.subparent.Foo` and be done.
"

**Installation:**

- Operating system: -
- Editor: -
- Metals version: v0.10.2

**Search terms**

completion akka import auto

EveryonePrint-dev:
This is not only for the `Address` class but for all classes of Akka.

The other case I mentioned with the subpackages, this happens also when there are no import clashes.

tgodzik:
> This is not only for the `Address` class but for all classes of Akka.
> 
> The other case I mentioned with the subpackages, this happens also when there are no import clashes.

Do you have an example where that happens?

tgodzik:
Also, it seems that I can't reproduce this behaviour on a clean akka project from https://github.com/akka/akka-quickstart-scala.g8.

![bug](https://user-images.githubusercontent.com/3807253/115843211-df76a180-a41e-11eb-8243-932c5c3b0ced.gif)

Does it happen in that project also? You can run `New Scala Project` and choose the project from the drop down to test that out quickly.

mdbergmann:
I'll try the quickstart.

Got another example here. Actually it seems to be unrelated to Akka.

The text in the Emacs completion popup shows ok, see 'label' below.
But the replacement is completely off, see 'newText'.

```
[server-reply] (id:116) Fri Apr 23 10:38:11 2021:
(:jsonrpc "2.0" :id 116 :result
          (:isIncomplete t :items
                         [(:label "AuthUserAction - action.authentication" :kind 9 :detail " action.authentication" :preselect t :sortText "00000" :filterText "AuthUserAction" :insertTextFormat 2 :textEdit
[...]

PR #2739, opened 2021-04-25
Fix wrong import when working with akka

tgodzik:
Previously, we would not check if the file renames were all working with an actual symbol. This was not a problem previously, but recently we added a check that would always try to use renamed imports, which would break in case of a NoSymbol. Now, we jsut check it before adding to the renames map.

Fixes https://github.com/scalameta/metals/issues/2736


dos65 (approved):
:rocket:

PR #2745, opened 2021-04-27
Toplevel Mtags: use single indexer implementation

### LINKED ISSUE #2742: Incorrect symbol information
**Describe the bug**

In VsCode, listing all symbols gives wrong information when there are multiple objects inside top object.
I suspect that related issue is that such symbols are not suggested for import in quick fix list.

**To Reproduce**

This is the repository where the bug is reproduced.
https://github.com/AurimasAnskaitis/metalsBug
After project import, press Ctrl+T and enter "C2" (C2 is a case class). Observe the path shown for this symbol.
It should be A.B2.C2 but is quite different.

**Expected behavior**

**Screenshots**

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: current stable version shipped with vscode, but I notice this bug for a long time, and only now it was reproduced in minimal project.

dos65:
Technically it's a replacement `ScalaToplevelMtags` by `Scala3ToplevelMtags` with minor changes to skip indentations processing for scala2 dialect.

In addition, it fixes #2742


tests/slow/src/test/scala/tests/feature/WorkspaceSymbolRegressionSuite.scala
```diff
@@ -7,7 +7,7 @@ import tests.BaseWorkspaceSymbolSuite
 
 class WorkspaceSymbolRegressionSuite extends BaseWorkspaceSymbolSuite {
   def workspace: AbsolutePath = Corpus.akka()
-  check("Actor", "1025 results")
+  check("Actor", "1009 results")
```
dos65, line 10: I checked all affected tests that were changed except this one.

Actually, this change just spotted the same problem as described in #2742.

tgodzik (approved):
LGTM!

PR #2751, opened 2021-04-30
Apply new Scala 3 rewrites

tgodzik:
Let's try to experiment with the new scalafmt features! Should we embrace Optional Braces? Or should we disable them altogether?

Opinons?

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,8 +1,9 @@
-version = "3.0.0-RC1"
+version = "3.0.0-RC2"
 project.git = true
 align.preset = none
 align.stripMargin = true
-docstrings = Asterisk
+docstrings.style = Asterisk
+docstrings.wrap = no
 assumeStandardLibraryStripMargin = true
 binPack.literalsExclude = ["Term.Name"]
 onTestFailure = """
@@ -20,5 +21,9 @@ project.excludeFilters = [
 fileOverride {
   "glob:**/scala-3*/**" {
     runner.dialect = scala3
+    rewrite.scala3.convertToNewSyntax = true
+    rewrite.rules = [RedundantBraces]
+    rewrite.scala3.removeOptionalBraces = yes
+
   }
 }
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConfigGenerationStatus.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConfigGenerationStatus.scala
@@ -7,15 +7,16 @@ object BspConfigGenerationStatus {
   type Message = String
   sealed trait BspConfigGenerationStatus
   case object Generated extends BspConfigGenerationStatus
-  case object Cancelled extends BspConfigGenerationStatus
+  case object Cancelled
+      extends BspConfigGenerationStatus
 
-  /**
-   * This is special cased as an Either to account for a build tool trying to
-   * generate the bsp config and failing, thus returning an exit code, but also
-   * used in the case where we can see that the build tool is maybe too old,
-   * and therefore we can just fail right away and return a Failed with the
-   * message that it's too old.
-   */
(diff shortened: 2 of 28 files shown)


.scalafmt.conf
```diff
@@ -20,5 +21,9 @@ project.excludeFilters = [
 fileOverride {
   "glob:**/scala-3*/**" {
     runner.dialect = scala3
+    rewrite.scala3.convertToNewSyntax = true
```
tgodzik, line 24: I would for sure try to keep these too rules. Maybe redundant braces also outside the scala 3 codebase?

metals/src/main/scala/scala/meta/internal/bsp/BspConfigGenerationStatus.scala
```diff
@@ -7,15 +7,16 @@ object BspConfigGenerationStatus {
   type Message = String
   sealed trait BspConfigGenerationStatus
   case object Generated extends BspConfigGenerationStatus
-  case object Cancelled extends BspConfigGenerationStatus
+  case object Cancelled
```
tgodzik, line 10: This is a bug that is already fixed, so we might need to update after the final release.

.scalafmt.conf
```diff
@@ -20,5 +21,9 @@ project.excludeFilters = [
...
+    rewrite.rules = [RedundantBraces]
+    rewrite.scala3.removeOptionalBraces = yes
```
tgodzik, line 26: To keep it in line with the rest of the codebase we might even need to disable optional braces, but not sure.

tgodzik (commented):
Everything seems to work nicely! Looks great! CC @kitbellew

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -39,30 +38,24 @@ class CompletionProvider(
...
             head.sym.is(Param)
 
-        if (
-          !isSeen(id) &&
-          !isUninterestingSymbol(head.sym) &&
-          isNotLocalForwardReference
-        ) {
+        if !isSeen(id) &&
+        !isUninterestingSymbol(head.sym) &&
+        isNotLocalForwardReference then
```
dos65, line 53: It seems like the indentation is missing  one second and third line
```scala
if !isSeen(id) &&
   !isUninterestingSymbol(head.sym) &&
   isNotLocalForwardReference then
```

kitbellew: actually, it is needed, @dos65 is right. (not because of significant indentation, of course, but because multiline infix needs to be indented if not within `{}`.) the only question is: should we break before `then`? i.e.

```
if !isSeen(id) &&
  !isUninterestingSymbol(head.sym) &&
  isNotLocalForwardReference then
   isSeen += id
...
```
or
```
if !isSeen(id) &&
  !isUninterestingSymbol(head.sym) &&
  isNotLocalForwardReference
then
   isSeen += id
```
i find the second one more readable. @tgodzik @dos65 what do you think?

tgodzik: Definitely, the second looks more readable. 🤔

kitbellew: is it valid scala3? :) before i implement this rule...

dos65: @kitbellew yep, it's valid )

kitbellew: scalameta/scalafmt#2491

PR #2765, opened 2021-05-04
Update Scalameta to 4.4.15

tgodzik:
The changes in handling cases are caused most likely https://github.com/scalameta/scalameta/pull/2286

Succeeds https://github.com/scalameta/metals/pull/2762


tests/unit/src/test/resources/foldingRange/expect/PartialFunctionCases.scala
```diff
@@ -8,6 +8,7 @@ class A >>region>>{
         println()
         println()
         println()
+        println()
```
tgodzik, line 11: The threshold is 3, so we should have 4 statements for the fold region to show up.

dos65 (approved):
LGTM!

PR #2767, opened 2021-05-06
Fix failing tests: SbtBloopSuite + FoldingRangeSuite (Windows)

dos65:
The previous test logic was based on downloading artifact from bintray that doesn't work anymore

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/parsing/FoldingRangeExtractor.scala
+++ b/metals/src/main/scala/scala/meta/internal/parsing/FoldingRangeExtractor.scala
@@ -30,10 +30,12 @@ final class FoldingRangeExtractor(
   }
 
   def extractFrom(tree: Tree, enclosing: Position): Unit = {
+    val tokens = tree.tokens
     if (span(tree.pos) > spanThreshold) {
       val newEnclosing = tree match {
         case Foldable(pos) if span(enclosing) - span(pos) > spanThreshold =>
-          distance.toRevised(pos.toLSP) match {
+          val adjusted = adjustEnd(pos, tokens)
+          distance.toRevised(adjusted.toLSP) match {
             case Some(revisedPos) =>
               val range = createRange(revisedPos)
               ranges.add(Region, range)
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):
Looks good. LGTM!

dos65:
Hmm. It seems that `tests.FoldingRangeScala2Suite` is broken on Windows.

I'm going to look at it because I'm unable to merge PR if some checks were not successful

metals/src/main/scala/scala/meta/internal/parsing/FoldingRangeExtractor.scala
```diff
@@ -106,6 +108,22 @@ final class FoldingRangeExtractor(
     foldingRange
   }
 
+  /**
+   * Due to logic thats forcibly shifts position in `Foldable`
+   * `end` might point at `\r>>range.end<<\n`
+   * that causes problems on Windows (at least in tests)
```
ckipp01, line 114: I wonder why this is popping up now?

ckipp01 (approved):
Looks good, LGTM!

metals/src/main/scala/scala/meta/internal/parsing/FoldingRangeExtractor.scala
tgodzik: This might have been an issue before, but only popped up recently due to my changes.

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/parsing/FoldingRangeExtractor.scala
tgodzik: Too late, I will take a look at it.

dos65: Opps, sorry)

Yep, I thought that it would be better to check how it works in the editor but I was too lazy to install Windows VM.
However, to be sure we have to test it on several editor.

tgodzik: I have a Windows machine set up and will test it out. Shouldn't take long to look at it. Thanks for fixing this btw!

PR #2774, opened 2021-05-10
Properly adjust folding range end on windows

tgodzik:
Previously, I didn't take into account windows line endings.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/parsing/FoldingRangeExtractor.scala
+++ b/metals/src/main/scala/scala/meta/internal/parsing/FoldingRangeExtractor.scala
@@ -30,12 +30,10 @@ final class FoldingRangeExtractor(
   }
 
   def extractFrom(tree: Tree, enclosing: Position): Unit = {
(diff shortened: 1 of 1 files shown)


dos65 (approved):
This one is more precise! LGTM

metals/src/main/scala/scala/meta/internal/parsing/FoldingRangeExtractor.scala
```diff
@@ -183,7 +165,16 @@ final class FoldingRangeExtractor(
 
         case c: Case =>
           val startingPoint = c.cond.getOrElse(c.pat)
-          val bodyEnd = c.body.pos.end + 1
+          val bodyEnd = {
+            c.body.pos match {
+              case Position.None => c.body.pos.end
+              case Position.Range(input, _, end) =>
+                val nextChar = input.chars(end + 1)
```
dos65, line 172: Oh. I missed test symbol idx changes.
It's needed to tests `end -1` because `Position.end` isn't inclusive

PR #2779, opened 2021-05-14
Add support for Scala 3.0.0

tgodzik:


metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
```diff
@@ -84,7 +84,7 @@ object ScalaVersions {
     if (isScala3Milestone(scalaVersion))
       scalaVersion
     else
-      scalaVersion.split('.').take(2).mkString(".")
+      scalaVersion.split('.').take(2).takeWhile(_ != "0").mkString(".")
```
tgodzik, line 87: 3.0.0 needs to translate to 3

bin/test-release.sh
```diff
@@ -6,12 +6,10 @@ argumentsRest=${@:2}
 suffix=${argumentsRest:-}
 
 coursier fetch \
-  org.scalameta:mtags_3.0.0-M1:$version \
-  org.scalameta:mtags_3.0.0-M2:$version \
-  org.scalameta:mtags_3.0.0-M3:$version \
   org.scalameta:mtags_3.0.0-RC1:$version \
   org.scalameta:mtags_3.0.0-RC2:$version \
   org.scalameta:mtags_3.0.0-RC3:$version \
+  org.scalameta:mtags_3.0.0:$version \
```
ckipp01, line 12: Sort of random, but are Scala 3 artifacts supposed to be `_3` or `_3.0.0`? This morning I hit on this in coursier and we had to correct to `_3` https://github.com/coursier/apps/pull/98

tgodzik: Mtags always resolve with the full scala version similar to anything that depend on the compiler. Just like the other mtags versions. This is finally an end to the very confusing time 😅

ckipp01: ahhh of course, duh. sorry wasn't thinking at all there.

build.sbt
```diff
@@ -226,8 +226,9 @@ lazy val V = new {
   def scala2Versions = nonDeprecatedScala2Versions ++ deprecatedScala2Versions
 
   // Scala 3
-  def nonDeprecatedScala3Versions = Seq(scala3, "3.0.0-RC2", "3.0.0-RC1")
-  def deprecatedScala3Versions = Seq("3.0.0-M3", "3.0.0-M2", "3.0.0-M1")
+  def nonDeprecatedScala3Versions =
+    Seq(scala3, "3.0.0-RC3", "3.0.0-RC2", "3.0.0-RC1")
```
tgodzik, line 230: Those are still used for courses at EPFL, so I will hold off dropping them until end of the semester most likely.

metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
gabro: this will break with 3.1.0. I realize it's far away, but maybe we can be a bit more explicit here:

```scala
if (isScala3Milestone(scalaVersion))
  scalaVersion
else if (scalaVersion.startsWith("3"))
  "3"
else
  scalaVersion.split(".").take(2).mkString(".")
```

?

gabro (commented):
🚀 one minor suggestion

metals/src/main/scala/scala/meta/internal/metals/ScalaVersions.scala
tgodzik: Why would it break for `3.1.0` ? It should return `3.1`, which I think would be expected.

gabro: what's this version used for? If 3.0.0 returns 3, I think 3.1.0 should return 3 as well

tgodzik: That's the binary version, so I think we will break binary compatibility for 3.1, or will it be always "3" ?

gabro: my understanding is that it will always be "3" due to the new binary compatibility guarantees in Scala 3. Minor versions of the language won't break binary compatibility (concretely we will be able to use libraries compiled for 3.0 from 3.1)

tgodzik: You are right, that's pretty neat! Will change it then.

tgodzik: Fixed!

gabro (approved):
🎊

tgodzik:
The failure is unrelated. Let's merge this!

Issue #2629, opened 2021-03-23
Symbol search - use correct dialect for Scala3 sources


dos65:
Currently, sources are indexing using a hardcoded `Scala213` dialect and parser.
After #2600 will be merged it needs to check all places where `Mtags.index` happens and provide a correct dialect.

PR #2780, opened 2021-05-14
SemanticdbDefinition - use correct dialect

dos65:
Allows querying workspace symbol from Scala3 sources (that can't be parsed by 2.13 dialect).

Closes: #2629  

Explanation:
The most interesting part of this changeset that now `SymbolDefinition` [carries dialect value](https://github.com/scalameta/metals/pull/2780/files#diff-64541b8a7e3f17eb6260aae5641ea55d7fa3cb8338e3600ee55157ae3fbf7e08R102-R106).
It was added to not introduce additional overhead in [`WorkspaceSymbolProvider.search`](https://github.com/scalameta/metals/pull/2780/files#diff-3038c34369f390eb83a7830e574b6eb0ed545173089182d69e95e57bcf06dfe5R129-R131) on inferring dialect by the path of source.


mtags/src/main/scala/scala/meta/internal/mtags/SymbolIndexBucket.scala
```diff
@@ -0,0 +1,279 @@
...
+  }
+  // similar as addSourceFile except indexes all global symbols instead of
+  // only non-trivial toplevel symbols.
+  private def addMtagsSourceFile(file: AbsolutePath): Unit = {
+    val docs: s.TextDocuments = PathIO.extension(file.toNIO) match {
+      case "scala" | "java" | "sc" =>
+        val language = file.toLanguage
+        val toIndexSource0 = toIndexSource(file).getOrElse(file)
+        val input = toIndexSource0.toInput
+        val document = mtags.index(language, input)
```
tgodzik, line 187: We should also pass the dialect here to avoid the fallback in https://github.com/scalameta/metals/blob/12db95cb14b1fefdc982243f0e4a1a6311914304/mtags/src/main/scala/scala/meta/internal/mtags/ScalaMtags.scala#L27

tgodzik (approved):
LGTM! Thanks for the follow up!

mtags/src/main/scala/scala/meta/internal/mtags/SymbolIndexBucket.scala
tgodzik: I will take a look at it

Issue #2795, opened 2021-05-15
Do not backtick soft keywords always


tgodzik:
**Describe the bug**

Currently, we are always backticking `extension` and `inline `soft keywords, however in most cases that is not neccessary. They are only needed to be backtickedt only if they are method calls. 

**To Reproduce**

1. Create an object `Test` in package `extension`
2. Try to auto import it from another package
3. `extension` will be automatically backticked.

**Expected behavior**

`extension` doesn't need to be backticked here

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.10.2-SNAPSHOT

**Additional context**

We could maybe try to fix it by checking if it's a standalone indentifier and in those cases only backtick it?

**Search terms**

soft keywords backticked

PR #2796, opened 2021-05-15
Remove most soft keywords from being backticked.

tgodzik:
Most of the soft keyword do not need to be backticked. Follow issue for the more problematic ones  https://github.com/scalameta/metals/issues/2795


dos65 (approved):
LGTM!

PR #2799, opened 2021-05-15
Forward dialect to ScalaMtags to avoid parsing fallback

tgodzik:


mtags/src/main/scala-2/scala/meta/internal/metals/ScaladocIndexer.scala
```diff
@@ -19,7 +20,7 @@ import scala.meta.pc.SymbolDocumentation
 class ScaladocIndexer(
     input: Input.VirtualFile,
     fn: SymbolDocumentation => Unit
-) extends ScalaMtags(input) {
+) extends ScalaMtags(input, Scala213) {
```
dos65, line 23: It seems that here we can also pass the correct dialect too.
It's used only in one place and dialect might be taken from `SymbolDefinition`: https://github.com/scalameta/metals/blob/e2518701d992316d5866c5839709e02a6ea5b969/mtags/src/main/scala-2/scala/meta/internal/metals/Docstrings.scala#L82-L90 

But this is optional,  I don't mind to fix it later

dos65 (approved):
Awesome!
Thanks for completing these dialect-related issues.

mtags/src/main/scala-2/scala/meta/internal/metals/ScaladocIndexer.scala
tgodzik: good catch, thanks! Fixed!

PR #2803, opened 2021-05-15
Fix field ordering in completions

tgodzik:
Previously, methods would be shown before field. Now, class fields are first.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
@@ -148,13 +148,21 @@ class CompletionProvider(
     import MemberOrdering._
     var relevance = 0
     val sym = completion.value.sym
+
+    def hasGetter = try {
+      // isField returns true for some classes
+      sym.isField || sym.getter != NoSymbol
+    } catch {
+      case _ => false
+    }
+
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -148,13 +148,21 @@ class CompletionProvider(
     import MemberOrdering._
     var relevance = 0
     val sym = completion.value.sym
+
+    def hasGetter = try {
+      // isField returns true for some classes
+      sym.isField || sym.getter != NoSymbol
```
tgodzik, line 154: `sym.isField` returns true for StringIndexOutOfBoundsException which is causing some of the other ordering issues. I will raise a PR in Dotty.

dos65: Does it return true for `StringIndexOutOfBoundsException` class or object?

There is a weird thing that classpath search might produce object symbols that don't actually exist.
I've just got an idea how to fix it and will try that.

tgodzik: Och, i haven't actually checked. Maybe that's the reason?

dos65 (approved):

PR #2804, opened 2021-05-16
[Scala3] Completions - fix `Renamed` and `specifyOwner` Autoimports application

dos65:
Fixes the following cases related to  `renameConfig` 
- incorrect owner edit in completion
  ```scala
  object Main {
    StringBuilder@@ // StringBuilder - scala.collection.mutable
  }
  // --apply--
  object Main {
  mutable.StringStringBuilder
  }
  ```
- owner/renamed owner import duplication
  ```scala
  import scala.collection.mutable
  object Main {
     StringBuilder@@ // StringBuilder - scala.collection.mutable
  }
  // --apply--
  import scala.collection.mutable
  import scala.collection.mutable
  object Main {
     mutable.StringBuilder
  } 
  ```
 
~~Unfortunately, tests on these cases are ignored because it's needed to fix our `TextEdit.applyEdits` implementation. 
I've tried to fix it but it seems like a non-trivial task. I'm going to look at it but want to merge this PR to not include these bugs in the release.~~

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
@@ -17,6 +17,7 @@ import org.eclipse.{lsp4j => l}
 
 import scala.annotation.tailrec
 import dotty.tools.dotc.util.Spans
+import scala.meta.internal.pc.AutoImports.AutoImportEdits
 
 object AutoImports {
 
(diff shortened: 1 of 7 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -111,12 +111,17 @@ object AutoImports {
       }
     }
 
-    private def specifyOwnerEdit(symbol: Symbol, owner: String): l.TextEdit = {
-      val name = symbol.showName
-      new l.TextEdit(nameEditRange(name), s"$owner.$name")
+    private def specifyOwnerEdit(owner: String): l.TextEdit = {
```
tgodzik, line 114: Could you use `ShortenedNames` here?

dos65: I'm not sure. Probably not, at least in its existing state.

dos65: Anyway, even plain imports currently are ignoring `ShortenedNames`. I will look at it. It seems it will require tweaking `AutoImports` implementation.

tests/cross/src/test/scala/tests/pc/CompletionWorkspaceSuite.scala
```diff
@@ -610,4 +610,35 @@ class CompletionWorkspaceSuite extends BaseCompletionSuite {
        |""".stripMargin,
     filter = _ == "Implicits - scala.concurrent.ExecutionContext"
   )
+
+  checkEdit(
+    // it's needed to rework TextEdits.applyEdits
+    "specify-owner".tag(IgnoreScala2).ignore,
```
tgodzik, line 616:
```suggestion
    "specify-owner".ignore,
```
should it not work the same for Scala 2?

Same for the one below

tgodzik (changes requested):
Thanks for the fix!

tgodzik (approved):
LGTM!

PR #2806, opened 2021-05-18
Group Scala versions in docs to make it easier to see which are supported

tgodzik:
image


dos65 (approved):
Looks much better!

gabro (commented):
Nice improvement! I wonder whether we should also separate the stable from the deprecate ones

tgodzik:
> Nice improvement! I wonder whether we should also separate the stable from the deprecate ones

Not sure,  I was thinking of maybe adding a visual indicator like cursive or crossed out text, but I fear that would just add confusion. I added sorting also, which should be useful.

Issue #2549, opened 2021-02-24
Duplicate diagnostics getting reported from Dotty and Bloop


ckipp01:
**Describe the bug**

Given a small code snippet with an error while using Scala 3 and Bloop..
```scala
object Main {
  val x =
}
```

Duplicated diagnostics are reported
```
[Trace - 07:39:54 PM] Sending notification 'textDocument/publishDiagnostics'
Params: {
  "uri": "file:///Users/ckipp/Documents/scala-workspace/dotty-tester/src/main/scala/Main.scala",
  "diagnostics": [
    {
      "range": {
        "start": {
          "line": 2,
          "character": 9
        },
        "end": {
          "line": 3,
          "character": 0
        }
      },
      "severity": 1,
      "source": "bloop",
      "message": "expression expected but \u0027}\u0027 found"
    },
    {
      "range": {
        "start": {
          "line": 2,
          "character": 9
        },
        "end": {
          "line": 2,
          "character": 9
        }
      },
      "severity": 1,
      "code": "18",
      "source": "dotty",
      "message": "expression expected but \u0027}\u0027 found"
    }
  ]
}
```

This causes editors to see this as two separate diagnostics displaying both when in reality they are one in the same.
**To Reproduce**

1. Fresh Scala 3 project
2. Copy code
3. Save and look at trace file

**Expected behavior**

I'd expect this to be reported once

**Screenshots**

<img width="500" alt="Screenshot 2021-02-24 at 19 42 27" src="https://user-images.githubusercontent.com/13974112/109049337-756ca500-76d8-11eb-85fb-77a4b9c9c355.png">

**Installation:**

- Operating system: macOS
- Editor: Vim
- Metals version: v0.10.0

**Search terms**

Scala 3 duplicate diagnostics

tgodzik:
Thanks for finding this, we do try to deduplicate the errors coming from the parser and the compiler, but it might not work perfectly in case of the Dotty parser itself.

tgodzik:
It's actually being done here: https://github.com/scalameta/metals/blob/8c24e5d2b909dda88f39312c66089195b5e36e9c/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala#L194
We might add additional logic to handle this case.

ckipp01:
> It's actually being done here:
> 
[...]

PR #2807, opened 2021-05-18
Publish scalameta parser diagnostics for Scala 3

tgodzik:
Previously, when the build was not imported and the fallback version set to Scala 3.0.0, we would get parsing errors in sbt files. Now, we added an explicit check to avoid it.

Fixes https://github.com/scalameta/metals/issues/2549

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -395,16 +395,8 @@ case class ScalaPresentationCompiler(
 
   override def didChange(
       params: VirtualFileParams
-  ): CompletableFuture[ju.List[Diagnostic]] = {
-    compilerAccess.withNonInterruptableCompiler(
-      Nil.asJava,
-      params.token
-    ) { access =>
-      val driver = access.compiler()
-      val uri = params.uri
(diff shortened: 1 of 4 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -402,7 +402,11 @@ case class ScalaPresentationCompiler(
     ) { access =>
       val driver = access.compiler()
       val uri = params.uri
-      CompilerInterfaces.parseErrors(driver, uri, params.text)
+      val isNotSbt = !uri.getPath.endsWith(".sbt")
+      if (isNotSbt)
+        CompilerInterfaces.parseErrors(driver, uri, params.text)
+      else
+        Nil.asJava
```
dos65, line 409: I've checked scala2 PC and this method has an empty implementation and does nothing.

I'm wondering why we use it?

tgodzik: Scala 3 parser diagnostics seem to be nicer, but maybe we can just remove it altogether?

tgodzik: I removed them, it should also fix some other issues

dos65 (approved):
LGTM!

PR #2809, opened 2021-05-19
Fix duplicate toplevels object definition when using enums

tgodzik:
This was done to also emit the object when enums being accessed as `Color.Red`, but since we don't do that for case classes, I think we shouldn't do it here.


mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
```diff
@@ -253,7 +253,6 @@ class ScalaToplevelMtags(
         withOwner(owner) {
           tpe(name.name, name.pos, Kind.CLASS, 0)
         }
-        term(name.name, name.pos, Kind.OBJECT, 0)
```
dos65, line 256: I did the same thing in [`ScalaMtags`](https://github.com/scalameta/metals/blob/80b0dde6ec45659445a0359b2f76c66e835200b8/mtags/src/main/scala/scala/meta/internal/mtags/ScalaMtags.scala#L196).

tgodzik: Fixed it there too!

dos65 (approved):
Cool! Thx for fixing this!

PR #2814, opened 2021-05-21
[Scalafmt] Rewrite `.scalafmt.conf` if sources require non-default dialect

dos65:
With these changes, Metals tries to detect if it's needed to upgrade scalafmt version and specify `runner.dialect` in `.scalafmt.conf` and automatically rewrites conf file if users allow that.


build.sbt
```diff
@@ -203,6 +203,7 @@ lazy val V = new {
   val mavenBloop = bloop
   val mdoc = "2.2.21"
   val scalafmt = "2.7.5"
+  val scalafmt3 = "3.0.0-RC3"
```
tgodzik, line 206: We should just update scalafmt, there is no need for a separate version

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -542,6 +542,39 @@ object Messages {
     }
   }
 
+  object UpdateScalafmtConf {
+
+    def letUpdate = new MessageActionItem("Let Metals update .scalafmt.conf")
+    def createMessage(dialect: ScalafmtDialect): String = {
+      s"Some source can't be formatted by scalafmt because they requires `runner.dialect = ${dialect.value}` setting."
```
ckipp01, line 549:
```suggestion
    def createMessage(dialect: ScalafmtDialect): String = {
      s"Some sources can't be formatted by scalafmt because they require `runner.dialect = ${dialect.value}` setting."
```

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
```diff
@@ -215,6 +216,101 @@ final class FormattingProvider(
...
+          } else if (item == Messages.dontShowAgain) {
+            tables.dismissedNotifications.UpdateScalaftmConf.dismissForever()
+          } else ()
+        }
+      case _ => Future.unit
+    }
+  }
+
+  def validateWorkspace(): Future[Unit] = {
+    if (scalafmtConf.exists) {
```
ckipp01, line 300: So I know when we do an actual format we do a check to ensure that if the user has manually set the `scalafmtConfigPath` it actually exists. here we also call `scalafmtConf`, but don't do that check. Do you think we should also do a check to ensure something wonky doesn't happen here, or even move that check into `scalafmtConf` to ensure that not only have they set the configPath, but it's also valid and there?

metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
```diff
@@ -0,0 +1,148 @@
...
+    runnerDialect.fold(applyVersion)(withUpdatedDialect(applyVersion, _))
+  }
+
+  def parse(path: AbsolutePath): Either[SyntaxError, ConfigValues] =
+    parse(ConfigFactory.parseFile(path.toFile))
+
+  def parse(text: String): Either[SyntaxError, ConfigValues] =
+    parse(ConfigFactory.parseString(text))
+
+  def parse(config: Config): Either[SyntaxError, ConfigValues] = {
```
ckipp01, line 97: Just a question. Does Scalafmt expose this functionality that we could just utilize instead of adding it all in here? This is a lot of logic to parse the scalafmt file, and then all that logic will probably be re-done in Scalafmt again later on. It'd be great to just utilize that.

tgodzik: Scalafmt doesn't normally modify the file, so that will not be available in the API. 

Alternatively,  we could just append file override to the file as you can have multiple and they should be merged in HOCON (I think). As for version and dialect, we could use regexes. 

However, if we already have this logic here then maybe I would leave it.

tgodzik: For parsing config, we would need to depend on Scalafmt and metaconfig artifacts themselves, so I think it's better to avoid it.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
dos65: As always, thx for grammar check!
I reworked this message a little bit could you check it one more time?

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -542,6 +543,48 @@ object Messages {
...
+  object UpdateScalafmtConf {
+
+    def letUpdate = new MessageActionItem("Let Metals update .scalafmt.conf")
+    def createMessage(
+        sourceItems: List[RelativePath],
+        dialect: ScalafmtDialect
+    ): String = {
+      val directories = sourceItems.mkString("[", ",\n", "]")
+      s"The following source directories can't be formatted by scalafmt " +
+        s"because they require `runner.dialect = ${dialect.value}` setting:" +
```
ckipp01, line 555:
```suggestion
        s"because they require the `runner.dialect = ${dialect.value}` setting:" +
```

Small nit that I actually missed last time.

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
dos65: `scalafmtConf` is this function that checks `userConfig`
https://github.com/scalameta/metals/blob/270804287a8abe18bdd072ec7cb0608fc77b1e2c/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala#L362-L365

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
dos65: Thx!

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
ckipp01: Sure, but it doesn't validate anything right? So if `userConfig().scalafmtConfigPath` returns `/my/wrong/path.conf`, that is still returned here and give to you even if it doesn't exists.

dos65: oh. now I got what you mean. Will fix that

dos65: Did two changes around that:
- in case if the conf file doesn't exist or has syntax error now message shows the correct file path in the text
- in case if upgrade is required Metals will suggest only to fix if it's the default one - (`.scalafmt.conf`).
  For a file that is specified in `userConfig` it will only warn about that as it might be shared across several projects.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -585,6 +587,17 @@ object Messages {
     }
   }
 
+  object InvalidScalaftmConfUserConfiguration {
+    def params(path: AbsolutePath): MessageParams = {
+      val params = new MessageParams()
+      params.setMessage(
+        s"The setting: `Scalafmt Config Path` is invalid: $path doesn't exists. Please check it to be able format sources"
```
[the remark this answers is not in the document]
dos65: Oh. I forgot to remove this message.

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
```diff
@@ -215,11 +218,157 @@ final class FormattingProvider(
...
+      val text = scalafmtConf.toInputFromBuffers(buffers).text
+      ScalafmtConfig.parse(text) match {
+        case Failure(e) =>
+          scribe.error(s"Failed to parse ${scalafmtConf}", e)
+          client.showMessage(
```
tgodzik, line 348: Let's not show it to the user, just log it. It's not critical, so it's fine to just have it there and it's highly possible that our parsing solution is not a 100 % valid. Especially if anyone has a more complex configuration there should be able to figure it out. Otherwise, they will keep on getting the errors every time.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -522,14 +523,16 @@ object Messages {
...
       )
+    private val messageStart = "Unable to format since"
     def isCreateScalafmtConf(params: ShowMessageRequestParams): Boolean =
-      params.getMessage == createScalafmtConfMessage
-    def createScalafmtConfMessage: String =
-      s"Unable to format since this workspace has no .scalafmt.conf file. " +
-        s"To fix this problem, create an empty .scalafmt.conf and try again."
-    def params(path: AbsolutePath): ShowMessageRequestParams = {
+      params.getMessage.startsWith(messageStart)
+    def message(path: String): String = {
```
tgodzik, line 529: I think we can leave this message. What was the reason for a change?

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -542,6 +545,50 @@ object Messages {
...
   }
 
+  object UpdateScalafmtConf {
+
+    def letUpdate = new MessageActionItem("Let Metals update .scalafmt.conf")
+    def createMessage(
+        sourceItems: List[RelativePath],
+        dialect: ScalafmtDialect
+    ): String = {
+      val directories = sourceItems.mkString("[", ",\n", "]")
```
tgodzik, line 555: Can we just print the number, but log everything and just mention that? It might actually be a lot of directories sometimes.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -542,6 +545,50 @@ object Messages {
...
+          if (canUpdate) Some(letUpdate) else None,
+          if (canUpdate) Some(notNow) else None,
+          Some(dontShowAgain)
+        ).flatten.asJava
+      )
+      params
+    }
+  }
+
+  object ErrorsInScalafmtConf {
```
tgodzik, line 580: Let's remove this. It's not really good to show it.

metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    final class Regex(regex: String) extends PathMatcher {
+      private val pattern = java.util.regex.Pattern.compile(regex)
+      def matches(path: AbsolutePath): Boolean =
+        pattern.matcher(path.toString).find()
+    }
+  }
+
+  case class ConfigValues(
```
tgodzik, line 43: Can this just be `ScalafmtConfig` class ?

metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
```diff
@@ -0,0 +1,206 @@
...
+      runnerDialect: Option[ScalafmtDialect],
+      fileOverrides: Map[String, ScalafmtDialect],
+      includeFilters: List[PathMatcher],
+      excludeFilters: List[PathMatcher]
+  ) {
+
+    def overrideFor(path: AbsolutePath): Option[ScalafmtDialect] = {
+      val fs = FileSystems.getDefault
+      fileOverrides
+        .find { case (pm, _) =>
```
tgodzik, line 54:
```suggestion
        .collect { case (pm, dialect) if matchesPath(pm) => dialect
```
and
```scala
def matchesPath(pm: PathMatcher) = fs.getPathMatcher(pm).matches(path.toNIO)
```

metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
```diff
@@ -0,0 +1,206 @@
...
+  case class ConfigValues(
+      version: Option[String],
+      runnerDialect: Option[ScalafmtDialect],
+      fileOverrides: Map[String, ScalafmtDialect],
+      includeFilters: List[PathMatcher],
+      excludeFilters: List[PathMatcher]
+  ) {
+
+    def overrideFor(path: AbsolutePath): Option[ScalafmtDialect] = {
+      val fs = FileSystems.getDefault
```
tgodzik, line 52:
```suggestion
      val fileSystem = FileSystems.getDefault
```
it's easier to understand. Reviewers need to think about what fs stands for. This is obviously a nitpick, but I am kind of inspired by the recent keynote at Scalacon about reading code.

metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
```diff
@@ -0,0 +1,206 @@
...
+      }
+    }
+
+    def withUpdatedDialect(content: String, d: ScalafmtDialect): String = {
+      val doc = docFrom(content)
+      if (doc.hasPath("runner.dialect"))
+        doc.withValueText("runner.dialect", d.value).render
+      else {
+        // append to the end
+        val sb = new StringBuilder
```
tgodzik, line 99: Could we extract that part to a common function in order no to repeat writing to config?

metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
```diff
@@ -0,0 +1,206 @@
...
+
+  def parse(path: AbsolutePath): Try[ConfigValues] =
+    Try(ConfigFactory.parseFile(path.toFile)).flatMap(parse)
+
+  def parse(text: String): Try[ConfigValues] =
+    Try(ConfigFactory.parseString(text)).flatMap(parse)
+
+  def parse(config: Config): Try[ConfigValues] = {
+
+    def getVersion(conf: Config): Try[Option[String]] =
```
tgodzik, line 123:
```suggestion
    def getVersion(conf: Config): Option[String] =
```
and use `toOption` ?

metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
```diff
@@ -0,0 +1,206 @@
...
+
+    def getRunnerDialect(
+        conf: Config
+    ): Try[Option[ScalafmtDialect]] = {
+      Try(getRunnerDialectRaw(conf))
+    }
+
+    def getFileOverrides(
+        conf: Config
+    ): Try[Map[String, ScalafmtDialect]] = {
```
tgodzik, line 145:
```suggestion
    ): Map[String, ScalafmtDialect] = {
```
and just default to empty map?

metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
```diff
@@ -0,0 +1,206 @@
...
+      if (conf.hasPath("runner.dialect")) {
+        val v = conf.getString("runner.dialect")
+        ScalafmtDialect.fromString(v)
+      } else
+        None
+    }
+
+    def getRunnerDialect(
+        conf: Config
+    ): Try[Option[ScalafmtDialect]] = {
```
tgodzik, line 139:
```suggestion
    ): Option[ScalafmtDialect] = {
```
toOption?

tgodzik (changes requested):
That turned out to be much more work than anticipated. Thanks for the great work! I think the general approach is solid, but I do have a few comments.

metals/src/main/scala/scala/meta/internal/metals/ScalafmtDialect.scala
```diff
@@ -0,0 +1,30 @@
+package scala.meta.internal.metals
+
+sealed abstract class ScalafmtDialect(val value: String)
+    extends Product
```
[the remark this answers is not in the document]
dos65: I might remove it but it's a common thing.
At one moment I had code where I needed it. The following won't compile without this trick: `List(ScalafmtDialect.Scala3, ScalafmtDialect.213).sorted`

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
```diff
@@ -215,11 +218,157 @@ final class FormattingProvider(
...
+    buildTargets.all.toList
+      .map(_.fmtDialect)
+      .sorted
+      .lastOption
+
+  private def initialConfig(): String = {
+    val runnerDialect = highestDialect.filter(_ != ScalafmtDialect.Scala213)
+
+    val versionLine = s"""version = "${BuildInfo.scalafmtVersion}""""
+    val runnerLine = runnerDialect.map(dialect => s"runner.dialect = $dialect")
```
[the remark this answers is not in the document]
dos65: I thought about that but it's hard to do absolutely correctly because of `globs`/`regex` - they have a free from.
Let's say there is the following config:
```
version = "2.7.5"
maxColumn = 100

fileOverride {
  'glob:$some-matcher' {
    maxColumn = 120
  }
}
```
In this case, we can't modify `fileOverride` as they might intersect each other.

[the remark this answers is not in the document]
tgodzik: ach damn, so maybe in case of 3. we should just show a warning with a link to file-override configuration setting?

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
dos65: It was changed to show the correct path to the config file that Metals tried to use.

A user might set a non-standard path for it in configuration and forget about that.
Then it will receive this notification and if we will decide to create `.scalafmt.conf` by hand nothing will be changed

metals/src/main/scala/scala/meta/internal/metals/ScalafmtDialect.scala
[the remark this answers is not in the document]
tgodzik: So maybe let's make it a case class and the exact dialects can be `val Scala3` etc. ?

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
dos65: I think it's better to show these directories because there are several options for solving this warning.
In addition to automatic rewrite, you might take these paths and do `fileOverride` by hand or add them in `excludePaths`.

Also, there is yet another one - create an issue that Metals wrongly detects dialect :smile:

metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
dos65: There is a difference:
- for `dialect` it is `append`
- for `version` it is `prepend` (version usually goes first)

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
tgodzik: I would rather change the message depending on whether we have a default scalafmt or a custom path. I would keep the current message for the current situation and change it for the latter to just `Custom scalafmt configuration at $path does not exist`.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
tgodzik: How will the message look in case of a really large project? In case you have 20 modules with test and main sources it will print 40 directories. You can say `look in the logs to find out which.` And you can even use the same logs link that we use when importing the build.

metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
tgodzik: Does it make a difference where the version is put?

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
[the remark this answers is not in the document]
dos65: Hmm. My first thought was that in most cases everything should be fine just to set global dialect but then I tried to do that for metals repo.
I got some non-significant rewrites related to newlines and 4 errors:
 - 3 of them about `inline`, `using`, `export` usage
 - `do while` syntax
 -  XML literals (will be fixed in the next scalafmt release?)
 
So, I agree that for the mixed project it would be better to use `fileOverride`.
The logic will be following:
- set the dialect using `fileOverride` per every non scala213 source directory if it `fileOverride` doesn't exist.
- if it exists - just make a warning. It will be hard to rewrite it because of issues with HOCON library +  I believe there aren't a lot of people who know about this setting.

metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
dos65: I believe that there is an unspoken rule about putting the version at the beginning of file

metals/src/main/scala/scala/meta/internal/metals/ScalafmtConfig.scala
dos65: It's better to combine these parsing functions over `Try` than `Option`.
In case of error, it will provide a certain error that will be printed in logs + if there any issues with parsing logic it's better just to fail this op than returning non-valid parsed value.

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
```diff
@@ -215,6 +218,147 @@ final class FormattingProvider(
...
+      ScalafmtConfig.parse(text) match {
+        case Failure(e) =>
+          scribe.error(s"Failed to parse ${scalafmtConf}", e)
+          client.showMessage(Messages.ErrorsInScalafmtConf.params)
+          Future.unit
+        case Success(values) =>
+          checkIfDialectUpgradeRequired(values)
+      }
+    } else {
+      handleMissingFile(scalafmtConf).map(_ => ())
```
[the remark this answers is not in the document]
dos65: Ok. Reverted all changes that were triggered by having `handleMissingFile` in this function.

metals/src/main/scala/scala/meta/internal/metals/ScalafmtDialect.scala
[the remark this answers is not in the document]
dos65: Removed `extends Product with Serailized`.

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
```diff
@@ -215,11 +218,157 @@ final class FormattingProvider(
...
+      if (buildTargetIds.nonEmpty && scalaFiles.nonEmpty) {
+        val required =
+          buildTargetIds
+            .flatMap(buildTargets.scalaTarget)
+            .map(_.fmtDialect)
+            .sorted
+            .headOption
+            .getOrElse(ScalafmtDialect.Scala213)
+
+        if (ord.gt(required, default)) {
```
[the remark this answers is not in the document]
dos65: It's needed to compare them using ordering.  required` might be less than `default`.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
dos65: Reverted this change

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
dos65: Moved `directories` list to logs

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
```diff
@@ -215,6 +223,155 @@ final class FormattingProvider(
...
+      config.fileOverrides.nonEmpty &&
+      files.forall(p => config.overrideFor(p).exists(d => ord.gteq(dialect, d)))
+    }
+
+    def inferDialectForSourceItem(
+        p: AbsolutePath,
+        buildTargetIds: List[BuildTargetIdentifier],
+        default: ScalafmtDialect
+    ): Option[ScalafmtDialect] = {
+      val scalaFiles =
```
tgodzik, line 258: I would defer calculating `scalaFiles` until `hasCorrectDialectOverride`:

```scala
    def hasCorrectDialectOverride(
        files: List[AbsolutePath],
        dialect: ScalafmtDialect
    ): Boolean = {
     lazy val scalaFiles = ... // calculate them
      config.fileOverrides.nonEmpty &&
      scalaFiles.nonEmpty &&
      scalaFiles.forall(p => config.overrideFor(p).exists(d => ord.gteq(dialect, d)))
    }
```

Otherwise, I am afraid we are adding a whole lot of effort whenever we index the workspace. And this way if everything is set ok, we will not need to calculate anything.

tgodzik (changes requested):
I left some more comments, the main blocker for me would be to not recursively list all the files if it's not needed.

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
```diff
@@ -299,10 +456,75 @@ final class FormattingProvider(
...
+  ) {
+
+    def canUpdate: Boolean =
+      rewriteType match {
+        case _: RewriteType.Auto => true
+        case RewriteType.Manual => false
+      }
+
+    def rewrite(text: String): String = {
+      val updVersion = versionUpgrade
```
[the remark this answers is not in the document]
dos65: To avoid triggering non-required reformatting that might happen due to version upgrade.

tests/unit/src/test/scala/tests/ScalafmtConfigSuite.scala
```diff
@@ -0,0 +1,176 @@
...
+      config: String,
+      expected: String,
+      updateVersion: Option[String] = None,
+      updateRunnerDialect: Option[ScalafmtDialect] = None,
+      updateFileOverride: Map[String, ScalafmtDialect] = Map.empty
+  ): Unit = {
+    test(options) {
+      val confPath =
+        AbsolutePath(BuildInfo.targetDirectory)
+          .resolve("scalaftm-config-tests")
```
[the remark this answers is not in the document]
dos65: Ha, this `confPath` wasn't needed at all. Removed it.

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
```diff
@@ -215,6 +223,155 @@ final class FormattingProvider(
...
+  }
+
+  private def checkIfDialectUpgradeRequired(
+      config: ScalafmtConfig
+  ): Future[Unit] = {
+
+    if (tables.dismissedNotifications.UpdateScalafmtConf.isDismissed)
+      Future.unit
+    else {
+      inspectDialectRewrite(config) match {
```
tgodzik, line 329: One last think I would do is maybe put this into a Future, so that we don't block on the check.

tgodzik:
```suggestion
      Future(inspectDialectRewrite(config)).flatMap{
```

tgodzik (approved):
LGTM from me, @ckipp01 what do you think?

ckipp01 (approved):
> LGTM from me, @ckipp01 what do you think?

Same for me LGTM!

PR #2815, opened 2021-05-24
Intend on paste

Giggiux:
This is a WIP, in particular for where the file where the code is situated, and because it's not properly tested.

The changes will work better if integrated with the changes in PR scalameta/metals-vscode#572. 

This PR, merged with the previously linked one to `metals-vscode` should close scalameta/metals-feature-requests#169, by adding automatic on paste indentation and automatic auto increase and decrease of indentation for optional braces.

The indent on paste works on "`rangeFormatting`", it assumes that the first line pasted is correctly indented, and then checks and computes, based on that line indentation, the correct indentation of the following pasted lines.

The main issue I find in the current solution is that we don't have any indentation configuration, so we don't know if the user is using tabs or spaces, and how many for each indent. What we do, is try to check in a simple way the current file indentation, otherwise will go to a default "two spaces" indentation.  Anyway, this solution should not break the user's code even if the indentation is not exactly the same in the whole file. And Scalafmt can always be run to fix any wrong indentation.
Also, the code is currently situated in the `MultilineStringFormattingProvider`, but that was just for simplicity of implementation, I do not expect to leave it there.

indent on paste


tgodzik:
We would welcome all comments/suggestions, so that we can be sure that the suggested solution is widely accepted. This heavily influences the UX and the overall experience of writing in Scala so it is important to get it right.

We are using here OnTypeFormatting again, but I see no better solution for this in Metals. Python Vs Code extension does it via separate paste command, but that I think would not be that visible to the users.

ckipp01:
> The indent on paste works on "`rangeFormatting`", it assumes that the first line pasted is correctly indented, and then checks and computes, based on that line indentation, the correct indentation of the following pasted lines.

I barely have any experience with indentation based syntax, but I'll really interested to see how this works. From playing around a bit with it, this sounds easy, but is actually a bit tricky unless the lines you are copying from has the exact indentation you already want, the first line won't be right. I don't know how to get around this.

> The main issue I find in the current solution is that we don't have any indentation configuration, so we don't know if the user is using tabs or spaces, and how many for each indent.

Actually this information should be contained in the `FormattingOptions`. It's required to be sent it and has both `tabSize` and `insertSpaces`. https://microsoft.github.io/language-server-protocol/specification#textDocument_formatting

> Also, the code is currently situated in the `MultilineStringFormattingProvider`, but that was just for simplicity of implementation, I do not expect to leave it there.

👍🏼 big thumbs up on this. Actually this may be a great time to get a better pattern in place to make it easier to add other features via `onTypeFormatting`. Right now it's pretty tightly coupled with multi-line strings, but there are other tickets that would be pretty easy to tackle if it was more decoupled, like https://github.com/scalameta/metals-feature-requests/issues/99

Giggiux:
> is actually a bit tricky unless the lines you are copying from has the exact indentation you already want, the first line won't be right

Essentially that's why it assumes that the first line pasted is correctly indented and it will work best (in vscode) with the changes in PR scalameta/metals-vscode#572: it has support for indentation in scala3. 

> Actually this information should be contained in the `FormattingOptions`.

The only issue I can find is that the `FormattingOptions` are sent for each `textDocument/formatting` request, so in order to use it with `rangeFormatting`, assuming that those options are unique in the whole project, we would have to save them from a `formatting` request. To do so, we could force/fake a formatting request on startup to save the FormattingOptions somewhere to reuse them, but this would be very unresponsive, and also just a hackish workaround.   
If you have any different approaches in mind, let me know!

In the meantime the current approach is to search of the correct indentation in a very simple way: checking in the current file where there should be already an indent and using what's found (if tabs or spaces, and if spaces, how many).

ckipp01:
> The only issue I can find is that the FormattingOptions are sent for each textDocument/formatting request, so in order to use it with rangeFormatting, assuming that those options are unique in the whole project, we would have to save them from a formatting request. 

Ahh, I was assuming that it'd be coming in during the `onTypeFormatting` request.

Giggiux:
Maybe you're right: not on `onTypeFormatting` but on `rangeFormatting` we also have the `FormattingOptions`: the `DocumentRangeFormattingParams` also have `options: FormattingOptions`. 
So it may actually work fine! 

I'll try and write a solution using these options!

Giggiux:
@ckipp01 with the latest commit I added a very basic abstraction of the `onTypeFormatting` and `rangeFormatting` provider, not having everything anymore in multilineStringFormattingProvider, that essentially doesn't exist anymore as a provider but has been repurposed as an `OnTypeRangeFormatter`. 
I tried to use a similar structure to the one we use in `CodeActionProvider`, but maybe there is a better solution to which I'm open to implementing! 


I still need to add tests, will do it in the next commits!

laughedelic:
> but is actually a bit tricky unless the lines you are copying from has the exact indentation you already want, the first line won't be right. 

I agree with this. as a vim(-mode) user, I often copy a set of lines (in visual-line mode) and when they are pasted, it doesn't matter where the cursor stands, insert will start on the next line with whatever indentation it originally had. I haven't tried this yet, but based on the description, I think to be able to use this functionality, I would need to switch to insert mode and place the cursor in the right position before pasting. It might sound vim-specific, but I'm sure other editors have similar behavior when copying several lines as a whole (e.g. by selecting via gutter).

I've written Haskell before and some editor plugin had this functionality (long time ago, nothing fancy). As far as I remember, the logic there was similar, but based indentation on the previous line, so if you have an open indentation block, it would paste inside that block, if the previous line doesn't open a block, it would past on the same indentation level. Just an idea to consider.

Giggiux:
I should've added support for the vim(-mode) users: now the indentation happens also if the first line is not correctly indented. 
Changing this also added some weird behavior that I had to fix, but now it should work.

Anyone that can help me find unsupported edge cases that I did not got right, is welcome!
An example of such edge case is the when parsing after a multi-line string: 

```scala
object Main:
  val myMultilineString: String = """| My
                                     | string""".stripMargin
  //@@ pasting here
end Main
```

the correct indentation found to indent the first line should not be considered the one of the piped line (`       | string`) but the one of the actual string declaration (`val myMultilineString ...`)

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/IndentOnPaste.scala
```diff
@@ -0,0 +1,168 @@
...
+    val pastedRange = new Range(rangeStart, endPos.toLSP.getEnd)
+    val startLine = startPos.toLSP.getStart.getLine
+    val endLine = endPos.toLSP.getEnd.getLine
+
+    val splitLinesWithIndex = splitLines.zipWithIndex
+    val inRangeLines = splitLinesWithIndex.slice(startLine, endLine + 1)
+    val pastedLines = inRangeLines.map(_._1)
+    val pastedLinesWithIndex = pastedLines.zipWithIndex
+
+    val increaseIndentPatternRegex =
```
tgodzik, line 34: We can put them toplevel in order to not recalculate them all the time.

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/IndentOnPaste.scala
```diff
@@ -0,0 +1,168 @@
+package scala.meta.internal.metals.onTypeRangeFormatters
+import scala.meta.internal.mtags.MtagsEnrichments._
+import scala.meta.tokens.Tokens
+
+import org.eclipse.lsp4j
+import org.eclipse.lsp4j.FormattingOptions
+import org.eclipse.lsp4j.Range
+import org.eclipse.lsp4j.TextEdit
+
+case class IndentOnPaste() extends OnTypeRangeFormatter {
```
tgodzik, line 10:
```suggestion
object IndentOnPaste extends OnTypeRangeFormatter {
```

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/MultilineString.scala
```diff
@@ -1,31 +1,48 @@
...
 import org.eclipse.lsp4j.Range
-import org.eclipse.lsp4j.TextDocumentIdentifier
 import org.eclipse.lsp4j.TextEdit
 
-class MultilineStringFormattingProvider(
-    buffers: Buffers,
-    trees: Trees,
-    userConfig: () => UserConfiguration
-) {
+case class MultilineString() extends OnTypeRangeFormatter {
```
tgodzik, line 13:
```suggestion
object MultilineString() extends OnTypeRangeFormatter {
```

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/MultilineString.scala
```diff
@@ -44,10 +61,19 @@ class MultilineStringFormattingProvider(
     }
   }
 
-  private def isMultilineString(token: Token): Boolean = {
```
tgodzik, line 47: Please, don't move the methods around. It's almost impossible to see what changed in this class.

tests/unit/src/test/scala/tests/rangeFormatting/MultilineStringRangeFormattingWhenPastingSuite.scala
```diff
@@ -319,10 +327,9 @@ class RangeFormattingWhenPastingSuite extends BaseLspSuite("rangeFormatting") {
        |object Main {
        |  val str = s'''
        |               |ok'''.stripMargin
-       |$indent
        |  val other = '''
-       |              |  some text
-       |              |'''.stripMargin
+       |  |  some text
```
tgodzik, line 331: The indentation should not have changed here I think

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -890,7 +897,8 @@ final class TestingServer(
       filename: String,
       original: String,
       paste: String,
-      root: AbsolutePath
+      root: AbsolutePath,
+      formattingOptions: Option[FormattingOptions]
```
tgodzik, line 901:
```suggestion
      formattingOptions: Option[FormattingOptions] = None
```

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -913,8 +922,10 @@ final class TestingServer(
     rangeFromString(filename, original, root) {
       case (text, textId, rangeSelection) =>
         val params = new DocumentRangeFormattingParams()
+        val formattingOptions = new FormattingOptions(2, true)
```
tgodzik, line 925: Let's add FormattingOptions parameter here also with a default None

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  def onTypeContribute(
+      sourceText: String,
+      range: Range,
+      splitLines: Array[String],
+      startPos: StartPosition,
+      endPos: EndPosition,
+      triggerChar: String,
+      position: Position,
+      enableStripMargin: Boolean,
```
tgodzik, line 32:
```suggestion
```
This is needed for a single class, I would just have it as a fucntion parameter of `MultilineString`

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
```diff
@@ -0,0 +1,36 @@
...
+
+trait OnTypeRangeFormatter {
+  type StartPosition = meta.Position
+  type EndPosition = meta.Position
+
+  def onRangeContribute(
+      sourceText: String,
+      range: Range,
+      splitLines: Array[String],
+      startPos: StartPosition,
```
tgodzik, line 18: How is `startPos` and `endPos` different from range?

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
```diff
@@ -0,0 +1,36 @@
...
+      endPos: EndPosition,
+      formattingOptions: FormattingOptions,
+      tokens: Option[Tokens]
+  ): Option[List[TextEdit]] = None
+
+  def onTypeContribute(
+      sourceText: String,
+      range: Range,
```
tgodzik, line 26: Ontype shouldn't need range, there should be really only a single position, no?

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
```diff
@@ -0,0 +1,36 @@
...
+import org.eclipse.lsp4j.TextEdit
+
+trait OnTypeRangeFormatter {
+  type StartPosition = meta.Position
+  type EndPosition = meta.Position
+
+  def onRangeContribute(
+      sourceText: String,
+      range: Range,
+      splitLines: Array[String],
```
tgodzik, line 17: Do we need to pass `splitLines`? That's an implementation detail of a particular OnTypeRangeFormatter

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
```diff
@@ -0,0 +1,36 @@
+package scala.meta.internal.metals.onTypeRangeFormatters
+
+import scala.meta.tokens.Tokens
+
+import org.eclipse.lsp4j.FormattingOptions
+import org.eclipse.lsp4j.Position
+import org.eclipse.lsp4j.Range
+import org.eclipse.lsp4j.TextEdit
+
+trait OnTypeRangeFormatter {
```
tgodzik, line 10: Could we split both into different traits? A method can be both, but sometimes we might want to implement only one method.

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/IndentOnPaste.scala
```diff
@@ -0,0 +1,168 @@
...
+      headLine = convertedLines(headIdx)
+      indentTailLines = increaseIndentation(headLine)
+      block = if (indentTailLines) 1 else 0
+      blockIndent = block * tabSize
+    } yield for {
+      line <- convertedLines.drop(headIdx + 1)
+      pastedLineIndentation <- codeStartPosition(line)
+    } yield {
+      val diff = currentIndentationLevel + blockIndent - pastedIndentation
+      if (diff != 0)
```
tgodzik, line 122:
```suggestion
```

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/IndentOnPaste.scala
```diff
@@ -0,0 +1,168 @@
...
+        if (diff < 0) {
+          if (pastedLineIndentation < -diff) {
+            stringRepeat(blank, currentIndentationLevel + blockIndent) ++ line
+              .slice(pastedLineIndentation, line.length)
+          } else line.slice(-diff, line.length)
+        } else stringRepeat(blank, diff) ++ line
```
tgodzik, line 128:
```suggestion
        } else if (diff > 0) stringRepeat(blank, diff) ++ line
        else line
```

metals/src/main/scala/scala/meta/internal/metals/OnTypeRangeFormattingProvider.scala
```diff
@@ -0,0 +1,137 @@
...
+import scala.meta.io.AbsolutePath
+import scala.meta.tokens.Tokens
+
+import org.eclipse.lsp4j.DocumentOnTypeFormattingParams
+import org.eclipse.lsp4j.DocumentRangeFormattingParams
+import org.eclipse.lsp4j.Range
+import org.eclipse.lsp4j.TextDocumentIdentifier
+import org.eclipse.lsp4j.TextEdit
+
+class OnTypeRangeFormattingProvider(
```
tgodzik, line 18: Can we split this too?

tgodzik (changes requested):
Thanks for the great work! I have some preliminary comments, but haven't dug into `IndentOnPaste` too much.

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
Giggiux: TBF, they were already implemented like this in the previous `MultilineStringFormattingProvider` implementation. 
As far as I can understand the `startPos` and `endPos` returned from the `withToken` method are the starting and ending position in the virtual file used in the `withToken` method. I'm not sure they actually change.

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
Giggiux: Since `splitLines` are used both in the `MultilineString` and in the `IndentOnPaste` formatters, I thought it would be more "efficient" to avoid having to repeat the same splitting operation of the entire sourceCode for both formatters.

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
Giggiux: It was already possible to implement only one method, just by using the default implementation of the correct `contribute`, but I've split them anyway!

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
Giggiux: I tried to add it as a `MultilineString` parameters, but that means it's gonna be a `case class` and not anymore an `object`, as you suggested.

tests/unit/src/test/scala/tests/rangeFormatting/MultilineStringRangeFormattingWhenPastingSuite.scala
Giggiux: This case is not handled by the MultilineStringFormatter, or at least it seems it's not, so the `rangeFormatter` falls back on the `IndentOnPaste` formatter that tries to indent it as normal code, and not multiline strings.

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
tgodzik: It's fine, it can be the case class then.

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
tgodzik: I get that, but since we are doing an API for more formatters, it might be a time to tidy it up. So if possible let's reduce the number of parameters, we can create an additional case class to maybe contain them if it's not possible to throw some out.

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
tgodzik: ok sure, might be needed then. Alternatively we can make it a lazy val in a case class that would contain more of the parameters, this way it will only be used if needed.

tests/unit/src/test/scala/tests/rangeFormatting/MultilineStringRangeFormattingWhenPastingSuite.scala
tgodzik: Hmm... but if the code is already indented in the paste snippet, shouldn't we keep the indentation? The same as with any block that has parts of it indented.

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/IndentOnPaste.scala
```diff
@@ -0,0 +1,173 @@
...
+    val (blank, tabSize) =
+      if (insertSpaces) (" ", originalTabSize) else ("\t", 1)
+
+    // These are the lines from the first pasted line, going above
+    val prePastedLines = splitLinesWithIndex.take(startLine).reverse
+
+    val currentIndentationLevel = (for {
+      (line, _) <- prePastedLines.find(t => {
+        val trimmed = t._1.trim()
+        trimmed.nonEmpty && !trimmed.startsWith("|")
```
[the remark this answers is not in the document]
Giggiux: Essentially it's to support this case:

> Anyone that can help me find unsupported edge cases that I did not got right, is welcome!
> An example of such edge case is the when parsing after a multi-line string:
> 
> ```scala
> object Main:
>   val myMultilineString: String = """| My
>                                      | string""".stripMargin
>   //@@ pasting here
> end Main
> ```
> 
> the correct indentation found to indent the first line should not be considered the one of the piped line (` | string`) but the one of the actual string declaration (`val myMultilineString ...`)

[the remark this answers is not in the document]
tgodzik: Ach yeah! After thinking a bit I realized the reason for it. I wonder though if we would be able to cover all possible cases. Since you can also have for example (I think that is only in Scala 3):
```scala
  val intNum = 123 
    + 123
```

Not sure what other possibilities we can have.

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/IndentOnPaste.scala
```diff
@@ -0,0 +1,173 @@
...
+      }) // Find first line non empty (aka code) that is not a piped multi-string
+
+      indentation <- codeStartPosition(line) // get indentation spaces
+      nextIncrease = increaseIndentation(
```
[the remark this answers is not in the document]
Giggiux: This part of the code checks if the first pasted line needs to have increase indentation compared to the non-pasted line.
Suppose you're pasting right after an if:
```scala
def function():
  if (1 < 2):
    @@ pasting here
```
Since we're not assuming anymore that we have the first line indented correctly due to @laughedelic suggestion,  we need to compute the first line correct indentation based on the previous one.
The rest of the code, essentially uses the idea you propose.

Also related to this, @dos65 tested that vim-mode for VSCode doesn't actually trigger `rangeFormatting` when pasting, but maybe using `coc-metals` it does?

tests/unit/src/test/scala/tests/rangeFormatting/MultilineStringRangeFormattingWhenPastingSuite.scala
Giggiux: I haven't actually checked the code but if the `MultilineString` formatter doesn't return an empty list, but a `None`, the formatting "passes" to the `IndentOnPaste` that doesn't consider "correct" the multiline-string formatting, or well, in this particular cases behave like this:

1. finds the correct indentation for the first line
2. finds the correct indentation for the second line based on the first one
3. computes all other lines according to the difference between the pasted indentation and the second line

It works like this because, as already discussed, the first line can either be pasted from the beginning of the line having references the copied indentation, or be pasted already in the correct indentation level not having those references. So the computation for the difference in indentation between the pasted lines and the current file is done comparing the second line to the others.

tgodzik: However, in this case ` val` is indented. We could tweak the algorithm in this case to adjust the intentation on the next lines starting with the second:
```
newIndentation = secondLineIndentation - firstLineIndentation + caluclatedIndetationForFirstline
```
Otherwise, if the second line is indented, we will always get it wrong.

Or, we could special case `|` and other operators to be always indented, which would work in most cases I think

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
Giggiux: Which parameters do you think are the best to keep in the formatting API, + the new case class?

tgodzik: If we need everything then it can probably be all the existing parameters.

Giggiux: I tried to do implement some sort of abstraction, it may need another iteration or two to be good enough, tho.

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
```diff
@@ -0,0 +1,55 @@
...
+case class TokensParams(
+    startPos: meta.Position,
+    endPos: meta.Position,
+    tokens: Option[Tokens]
+)
+
+trait FormatterParams {
+  val sourceText: String
+  val range: Range
+  lazy val splitLines: Array[String] = sourceText.split('\n')
```
dos65, line 40:
```suggestion
  lazy val splitLines: Array[String] = sourceText.split(System.lineSeparator)
```

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/IndentOnPaste.scala
```diff
@@ -0,0 +1,170 @@
...
+
+    lazy val singleLineOption =
+      if (singleCodeLinePasted) indentedHead.map(List(_)) else None
+
+    newLines
+      .orElse(singleLineOption)
+      .map(lines =>
+        new TextEdit(
+          pastedRange,
+          lines.mkString("\n")
```
dos65, line 166:
```suggestion
          lines.mkString(System.lineSeparator)
```

dos65 (changes requested):
@Giggiux I checked what is wrong with windows tests - that was the line separator. 
Suggested changes should fix them.

metals/src/main/scala/scala/meta/internal/metals/onTypeRangeFormatters/OnTypeRangeFormatter.scala
dos65: This suggestion wasn't correct and leaded to `OnTypeFormattingSuite` failure.
Let's try this one. I checked that `tests.rangeFomatiing.*` + `tess.OnFomattingSuite` pass locally with the following:
```scala
lazy val splitLines: Array[String] = sourceText.split("\\r?\\n")
```

Giggiux: What about the `mkString` in file `IndentOnPaste`, should we change it back to `lines.mkString('\n')` to be consistent with the rest of the project?

dos65: @Giggiux no. We should keep `System.lineSeparator`  there. Otherwise the resulting text will have different sep from expected text on Windows

tgodzik:
I went ahead and simplified the code a bit, I didn't change any implementation details, but removed some classes and methods that were adding to the complexity.

I also moved all the new classes to the `formatting` package. I will follow up later and add the remaining things connected to formatting there.

metals/src/main/scala/scala/meta/internal/metals/formatting/OnTypeFormattingProvider.scala
```diff
@@ -0,0 +1,74 @@
...
+      params: DocumentOnTypeFormattingParams
+  ): List[TextEdit] = {
+    val uri = params.getTextDocument.getUri.toAbsolutePath
+    val range = new Range(params.getPosition, params.getPosition)
+    val triggerChar = params.getCh
+    val position = params.getPosition
+    buffers
+      .get(uri)
+      .map { sourceText =>
+        val virtualFile = Input.VirtualFile(sourceText.toString(), sourceText)
```
tgodzik, line 55: Instead of using `contribute` method I just inlined everything and removed the parent class. This was just adding to the cognitive overload.

tgodzik (approved):
Just tested it out after the changes and everything seems to work ok. We can still improve it a bit, but we can do it in the successive PRs

PR #2817, opened 2021-05-24
Fix issues when importing in an empty package

tgodzik:
Previously, the new import might end up in the middle of the comment. Now, if no package or import is defined, we put it into the first line.


dos65 (commented):
LGTM!

PR #2818, opened 2021-05-24
Update pprint to newest version

tgodzik:
Newest pprint version now support 2.11, 2.12, 2.13 and 3.0.0. This also means that it will not work on the RC versions, but since we plan to remove those in the near future and CI tests do not add pprint, I think we are safe to simplify adding it.


ckipp01 (approved):
haha so nice to not have to have 3 separate cases. LGTM!

dos65 (approved):
:rocket:

Issue #2776, opened 2021-05-11
Problems moving to scala 3.0.0-RC3 with sbt as build processor for metals


DavidGoodenough:
taking an existing scala-2 project and moving to scala-3 if I select sbt as the build processor it timesout.  Rebuilding the bspConfig helps but it still says that the project is misconfigured and when I ask for more details it shows the doctor with everything ticked but the following recommendation:-

Add the compiler option "-P:semanticdb:sourceroot:/home/david/src/OpenBusiness/UblBase" to ensure code navigation works.

DavidGoodenough:
The problem however persists.  When I choose sbt as BSP it times out, when I choose Bloop it works.  The message was the only diagnostic I could see.  If there is more information in a log somewhere please let me know what to look at.

DavidGoodenough:
The problem still seems to exist in the current metals snapshot.  It is still trying to connect after 5 minutes, and in the .metals/metals.log is just says that it timed out and is retrying.  Is there anything I can do to get more information?

DavidGoodenough:
Eventually it gave up and returned:-
```
2021.05.17 12:01:25 ERROR Timeout waiting for 'build/initialize' response
2021.05.17 12:01:25 ERROR Failed to connect with build server, no functionality will work.
java.util.concurrent.TimeoutException
	at java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1886)
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2021)
	at scala.meta.internal.metals.BuildServerConnection$.initialize(BuildServerConnection.scala:369)
	at scala.meta.internal.metals.BuildServerConnection$.$anonfun$fromSockets$1(BuildServerConnection.scala:283)
	at scala.util.Success.$anonfun$map$1(Try.scala:255)
	at scala.util.Success.map(Try.scala:213)
	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
	at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
	at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[...]

PR #2819, opened 2021-05-25
Don't warn about sourceroot for Scala 3

### LINKED ISSUE #2285: Metals Doctor recommend using "semanticdb:sourceroot"

ckipp01:
This has been causing some confusion, and we shouldn't warn about this. From what I can tell looking at the semanticdb related stuff in Scala3, this pretty much always seems to just be generated from the target and then made relative. From playing around, while this warning appears, it's often a red herring since everything is fine, but people assume something may be wrong or incorrectly blame the sourcepath not being set.

It doesn't mention anything about it [here in the docs](https://docs.scala-lang.org/scala3/guides/migration/tutorial-prerequisites.html#semanticdb) and the few places I see where it's used, it's often put together:

https://github.com/lampepfl/dotty/blob/b7d2a122555a6aa44cc7590852a80f12512c535e/compiler/src/dotty/tools/dotc/semanticdb/ExtractSemanticDB.scala#L613

 https://github.com/lampepfl/dotty/blob/af0de81a9d005ddbab172421b15f39432a45ff86/compiler/test/dotty/tools/dotc/semanticdb/SemanticdbTests.scala#L66

I don't think we need to warn about this even though it is a valid setting. An alternative would be to just set it here: https://github.com/scalameta/metals/blob/195e88162d040462a7e51c603ce59881f940b9fa/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala#L69-L89, but since we aren't setting the targetroot, and going with the default, I think we should do the same for the sourceroot.

Relates to #2776 
Closes #2285

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
+++ b/metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
@@ -162,7 +162,9 @@ class ProblemResolver(
             isUnsupportedBloopVersion()
           )
         )
-      case _ if !scalaTarget.isSourcerootDeclared =>
+      case _
+          if !scalaTarget.isSourcerootDeclared && !ScalaVersions
+            .isScala3Version(scalaTarget.scalaVersion) =>
         Some(MissingSourceRoot(workspace.sourcerootOption))
       case version
           if ScalaVersions.isDeprecatedScalaVersion(


tgodzik (approved):
LGTM! I don't think I ever seen this being useful.

PR #2821, opened 2021-05-25
Print better types on hover

tgodzik:
This improves a couple of things:
- show shorter type if it's importable (also needed further on)
- when hovering on package it's full path is shown (the change was needed due to the previous one)
- show final modifier
- show enum and case instead of final object and final val

The first change was needed also for properly writing types when doing inferred type action or overrides. The other changes stemmed from that. I think we can still further improve the hover information later on, but I don't want to do everything in one PR

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
@@ -23,16 +23,34 @@ class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx) {
   }
 
   def fullDefinition(sym: Symbol, tpe: Type): String = {
-    val isImplicit = sym.is(Flags.Implicit)
-    val name = nameString(sym)
-    val implicitKeyword = if (isImplicit) "implicit " else ""
+    def name = nameString(sym)
     keyString(sym) match {
       case "" =>
+        val isImplicit = sym.is(Flags.Implicit)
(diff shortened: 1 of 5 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
```diff
@@ -23,16 +23,34 @@ class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx) {
...
+        val implicitKeyword = if (isImplicit) "implicit " else ""
         s"$implicitKeyword$name: "
       case key if sym.is(Flags.Method) =>
         s"$key $name"
+      // package
+      case key if sym.is(Flags.Package) =>
+        val owners = for {
+          sym <- sym.ownersIterator
+          name = super.nameString(sym.name)
+          if name != "<root>" && name != "empty"
```
dos65, line 39:
```suggestion
          sym <- sym.ownersIterator
           if !(sym.isRoot || sym.isEmptyPackage)
          name = super.nameString(sym.name)
         
```

dos65 (approved):
LGTM!

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
tgodzik: Thanks! I couldn't find a way to do it without string. Fixed

PR #2822, opened 2021-05-26
Don't offer completions inside scaladoc and comments

### LINKED ISSUE #2775: Don't offer completions when inside of a scaladoc
**Describe the bug**

After using the insert scaladoc completion, I don't want completions given to me by Metals while still in the completion. For example in the following snippet:

```scala

object Main {

  /** @@ // At this point you'll get completions from Metals
    *@param a
    *@param b
    *@return
    */
  def doThing(a: Int, b: Int): Int = a + b + 1
}
```

For example here is what I see when I start typing `t`.

Screenshot 2021-05-10 at 14 20 22

**To Reproduce**

1. Make a simple method
2. Add the docs via Metals
3. Type `t`
4. See completions

**Expected behavior**

I would not expect Metals to return completions here.

**Installation:**

- Operating system: macOS
- Editor: NeoVim
- Metals version: v0.10.2+44-791e0eb9-SNAPSHOT

**Search terms**

scaladoc completions

Giggiux:
This PR should close #2775, removing the completions when inside comment lines, either multiline like scaladoc or single line `//`.

I am not sure the solution is an optimized one, but I added a similar check to the `isScaladocCompletion` to check whether the cursor is currently in a comment line. In this case it will return a `CommentCompletion` created equally to the `NoneCompletion`, the need to differentiate occurs when checking for possible Keywords completions: if it's a `CommentCompletion` it shouldn't propose the keywords either.

~~+: I need to fix something wrong with Keywords, then should be ready~~

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/Keywords.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/Keywords.scala
@@ -2,6 +2,9 @@ package scala.meta.internal.pc
 
 import scala.tools.nsc.reporters.StoreReporter
 
+import scala.meta.XtensionTokenizeInputLike
+import scala.meta.tokens.Token
+
 import org.eclipse.{lsp4j => l}
 
 trait Keywords { this: MetalsGlobal =>
@@ -10,8 +13,22 @@ trait Keywords { this: MetalsGlobal =>
       pos: Position,
       editRange: l.Range,
       latestEnclosing: List[Tree],
-      completion: CompletionPosition
+      completion: CompletionPosition,
+      text: String
   ): List[Member] = {
+    val start = pos.start
+    val end = pos.end
+    val tokens = text.tokenize.toOption
+    val notInComment = tokens
+      .flatMap(t =>
+        t.find {
+          case t: Token.Comment if t.pos.start < start && t.pos.end >= end =>
+            true
+          case _ => false
+        }
+      )
+      .isEmpty
+
     getIdentifierName(latestEnclosing, pos) match {
       case None =>
         completion match {
@@ -22,12 +39,13 @@ trait Keywords { this: MetalsGlobal =>
           // has already typed /* then they are going for the scaladoc, not the
           // other stuff.
           case _: ScaladocCompletion => List.empty
-          case _ =>
+          case _ if notInComment =>
             Keyword.all.collect {
               case kw if kw.isPackage => mkTextEditMember(kw, editRange)
             }
+          case _ => List.empty
         }
-      case Some(name) =>
+      case Some(name) if notInComment =>
         val isExpression = this.isExpression(latestEnclosing)
         val isBlock = this.isBlock(latestEnclosing)
         val isDefinition = this.isDefinition(latestEnclosing, name, pos)
@@ -47,6 +65,7 @@ trait Keywords { this: MetalsGlobal =>
               ) =>
             mkTextEditMember(kw, editRange)
         }
+      case _ => List.empty
     }
   }
 
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/Keywords.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/Keywords.scala
@@ -2,7 +2,7 @@ package scala.meta.internal.pc
 
 import scala.tools.nsc.reporters.StoreReporter
 
-import scala.meta.XtensionTokenizeInputLike
+import scala.meta._
 import scala.meta.tokens.Token
 
 import org.eclipse.{lsp4j => l}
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -409,6 +409,90 @@ trait Completions { this: MetalsGlobal =>
         NoneCompletion
     }
   }
+
+  private def isComment(str: String, pos: Position): Boolean = {
```
tgodzik, line 413: So we discussed this with @Giggiux and there seems to be no better way of finding a comment and it looks to be on the heavy side. The compiler doesn't seem to over a sensible way. I am thinking that we should leave it as is and close the issue as `wond't fix` - it's a small enough issue that shouldn't bother users in most cases. What do you think @ckipp01 ?

ckipp01: If we don't want to tackle it now, that's fine, but i wouldn't close it with `won't fix`. I think it's a reasonable expectation that in this situation you wouldn't get completions offered to you. For example I just checked IntelliJ and it doesn't offer completions like this in comments, so I think a popup that offers completion options like `trait` etc in a comment is really off putting.

tgodzik: Hmm.. in VS Code this doesn't seem to be a big problem, but I wonder whether it is only the keyword completions? Are you getting something else too? I think we can optimize it a bit in keywords :thinking:

mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -409,6 +409,90 @@ trait Completions { this: MetalsGlobal =>
...
+    object CommentDelimiter extends Enumeration {
+      type CommentDelimiter = Value
+      val Init, End = Value
+    }
+    // check if the position is between /** and */
+    // or if line contains before char's position //
+    val lineIdx = pos.line - 1
```
tgodzik, line 420: What if we do it differently. We could use scalameta tokenizer to tokenize part of the file that is in the last tree range? 
We would need to get the text, then call `.tokenize` and find the current at the current offset. That might be shorter and faster.

Additionally I would add this check at the end of completions in Keywords after `matchesPosition` :

```
if kw.matchesPosition(
   ...
) && isNotComment
```

mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
ckipp01: I think the reason I notice it so much is that every time use the scaladoc completion, after the completion finishes the next thing I see is this:

Screenshot 2021-05-30 at 10 05 21

It's only these keywords that are every offered.

mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
Giggiux: I tried to do it as you suggested!
I added a couple of tests and it seems to work in all scala versions (tested using ` +cross/testOnly tests.pc.CompletionKeywordSuite` locally).

I force pushed in this branch, removing the previous changes that are available in [Giggiux:old-fix-scaladoc-no-completions ](https://github.com/Giggiux/metals/tree/old-fix-scaladoc-no-completions)

mtags/src/main/scala-2/scala/meta/internal/pc/Keywords.scala
```diff
@@ -10,8 +13,22 @@ trait Keywords { this: MetalsGlobal =>
       pos: Position,
       editRange: l.Range,
       latestEnclosing: List[Tree],
-      completion: CompletionPosition
+      completion: CompletionPosition,
+      text: String
   ): List[Member] = {
+    val start = pos.start
```
tgodzik, line 19: Could you put it into a def? And later w can do
`lazy val isNotInComment = checkIfNotInComment()`

mtags/src/main/scala-2/scala/meta/internal/pc/Keywords.scala
```diff
@@ -2,6 +2,9 @@ package scala.meta.internal.pc
 
 import scala.tools.nsc.reporters.StoreReporter
 
+import scala.meta.XtensionTokenizeInputLike
```
tgodzik, line 5:
```suggestion
```
This should not be needed. You might do `import scala.meta._` instead (if there are conflicts you can import inside the def)

tgodzik (changes requested):
Thanks! This seems much simpler!

tgodzik (approved):
LGTM!

PR #2824, opened 2021-05-27
Fix pprint dependency for Scala3 RC versions

dos65:
Unfortunately, pprint `0.6.6` isn't published for RC* versions

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -235,8 +235,9 @@ lazy val V = new {
   def scala2Versions = nonDeprecatedScala2Versions ++ deprecatedScala2Versions
 
   // Scala 3
+  def scala3RCVersions = Seq("3.0.0-RC3", "3.0.0-RC2", "3.0.0-RC1")
   def nonDeprecatedScala3Versions =
-    Seq(scala3, "3.0.0-RC3", "3.0.0-RC2", "3.0.0-RC1")
+    scala3 +: scala3RCVersions
   def deprecatedScala3Versions = Seq()
   def scala3Versions = nonDeprecatedScala3Versions ++ deprecatedScala3Versions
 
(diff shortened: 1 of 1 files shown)


ckipp01:
Spring classes should be done at EPFL no? Do we still need to even support RC versions in the next release?

tgodzik:
> Spring classes should be done at EPFL no? Do we still need to even support RC versions in the next release?

Classes finish in a bit `June 4, 2021`. We can instead deprecate RC versions and keep them until exams end on 10th July.

build.sbt
```diff
@@ -350,8 +350,17 @@ val mtagsSettings = List(
     if (isCI) Nil
     // NOTE(olafur) pprint is indispensable for me while developing, I can't
     // use println anymore for debugging because pprint.log is 100 times better.
-    else
-      List("com.lihaoyi" %% "pprint" % "0.6.6")
+    else {
+      val scalaV = scalaVersion.value
+      val isScala3RC = scalaV.stripPrefix("3.0.0-").startsWith("RC")
```
tgodzik, line 355: Do we need it at all in `3.0.0-RC*` ? That's just a utility useful while testing, let's instead deprecate RC versions

dos65: It's hard to say ))
Maybe let's wait for the end of exams and deprecate RC after them?
Just to avoid bothering students by `You are using legacy ... Please upgrade` notifications while they are preparing for exams.

build.sbt
```diff
@@ -350,8 +350,17 @@ val mtagsSettings = List(
...
+      if (isScala3RC) {
```
tgodzik, line 356: Let's maybe do then:
```suggestion
      if (scala3RCVersions.contains(scalaVersion.value)) Nil
      else List("com.lihaoyi" %% "pprint" % "0.6.6")
```

pprint will not work too great with `for3Use2_13` from what I remember. And we can put `scala3RCVersions` into a separate collection.

tgodzik (approved):
LGTM!

PR #2825, opened 2021-05-27
[Scala 3] Enable inferred type code action

tgodzik:
This was actually easier to do for some parts, however there are some issues still:
- tuples are not shown correctly when using the presentation compiler, I am working around it by using scalameta to parse the tuple Val and match the type. This doesn't seem to be an issue for more advanced patterns.
- ~~import renames are not correctly picked up, we might need to still fix it in auto imports/signatures~~ this currently works
- type names in case of conflicts do not currently work, I will look into that later on 

Most other things seem to be working fine for now and I believe this is a good step forward.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
@@ -15,6 +15,10 @@ import dotty.tools.dotc.interactive.InteractiveDriver
 import dotty.tools.dotc.util.SourcePosition
 import dotty.tools.dotc.util.Spans
 import org.eclipse.{lsp4j => l}
+import dotty.tools.dotc.interactive.Interactive
+import scala.meta.pc.OffsetParams
+
+import java.net.URI
 
 object MtagsEnrichments
     extends CommonMtagsEnrichments
@@ -28,6 +32,23 @@ object MtagsEnrichments
       val source = driver.openedFiles(uri)
       val p = Spans.Span(offset)
       new SourcePosition(source, p)
+
+    def localContext(params: OffsetParams): Context = {
+      val unit = driver.currentCtx.run.units.head
+      val tree = unit.tpdTree
+      val pos = driver.sourcePosition(params)
+      val path =
+        Interactive.pathTo(driver.openedTrees(params.uri), pos)(using
+          driver.currentCtx
+        )
+
+      val newctx = driver.currentCtx.fresh.setCompilationUnit(unit)
+      val tpdPath =
+        Interactive.pathTo(newctx.compilationUnit.tpdTree, pos.span)(using
+          newctx
+        )
+      Interactive.contextOfPath(tpdPath)(using newctx)
+    }
 
   extension (pos: SourcePosition)
     def toLSP: l.Range = {
--- a/mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
@@ -11,14 +11,11 @@ import dotty.tools.dotc.core.Contexts._
 import dotty.tools.dotc.core.NameOps._
 import dotty.tools.dotc.core.Names._
 import dotty.tools.dotc.core.Symbols._
+import dotty.tools.dotc.interactive.Interactive
 import dotty.tools.dotc.interactive.InteractiveDriver
 import dotty.tools.dotc.util.SourcePosition
 import dotty.tools.dotc.util.Spans
 import org.eclipse.{lsp4j => l}
-import dotty.tools.dotc.interactive.Interactive
-import scala.meta.pc.OffsetParams
-
-import java.net.URI
 
 object MtagsEnrichments
     extends CommonMtagsEnrichments
@@ -34,6 +31,10 @@ object MtagsEnrichments
       new SourcePosition(source, p)
 
     def localContext(params: OffsetParams): Context = {
+      if (driver.currentCtx.run.units.isEmpty)
+        throw new RuntimeException(
+          "No source files were passed to the Scala 3 presentation compiler"
+        )
       val unit = driver.currentCtx.run.units.head
       val tree = unit.tpdTree
       val pos = driver.sourcePosition(params)
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
@@ -1,25 +1,27 @@
 package scala.meta.internal.pc
 
+import java.net.URI
+import java.nio.file.Paths
+
+import scala.meta as m
+
+import scala.meta.internal.mtags.MtagsEnrichments._
+import scala.meta.internal.pc.AutoImports.AutoImportsGenerator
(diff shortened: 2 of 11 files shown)


mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
```diff
@@ -28,6 +29,23 @@ object MtagsEnrichments
       val p = Spans.Span(offset)
       new SourcePosition(source, p)
 
+    def localContext(params: OffsetParams): Context = {
+      val unit = driver.currentCtx.run.units.head
```
ckipp01, line 33: Do we need to ever worry about the compilation units being empty here and throwing with `head`?

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -0,0 +1,261 @@
...
+    driver: InteractiveDriver,
+    config: PresentationCompilerConfig
+) {
+
+  def inferredTypeEdits(): List[TextEdit] = {
+    val uri = params.uri
+    val filePath = Paths.get(uri)
+    val source = SourceFile.virtual(filePath.toString, params.text)
+    driver.run(uri, source)
+    val unit = driver.currentCtx.run.units.head
```
ckipp01, line 52: Same as the other comment, I'm assuming it's fine, but we don't need to worry about the `head` call here?

ckipp01 (commented):
Really great work @tgodzik. Nothing really jumps out at me, but I'm also pretty unfamiliar with this part of Metals. Just a couple questions on the `head` usage though.

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
tgodzik: We shouldn't need to worry, that head is the just parsed source file. We can add a check, but this is not the only place we are doing it and there seems to be no issues around it.

ckipp01: No need to add a check then. I was just curious :)

mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
tgodzik: In this case, we should probably throw with a more specific message. I will add it.

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
tgodzik: And this is basically produced by `driver.run(uri, source)`

tests/cross/src/test/scala/tests/pc/InsertInferredTypeSuite.scala
```diff
@@ -267,56 +295,47 @@ class InsertInferredTypeSuite extends BaseCodeActionSuite {
...
+        |      logger <- mkLogger[F]
+        |    } yield ()
+        |
+        |}
+        |""".stripMargin
   )
 
   checkEdit(
-    "path",
+    "path".tag(IgnoreScala3),
```
tgodzik, line 338: This is the last problematic case, it's depended on fixing the `shortType` method in case of conflicts, which doesn't work perfectly currently.

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -0,0 +1,247 @@
...
+    def printType(tpe: Type): String = {
+      val short = shortType(tpe, shortenedNames)
+      symbolPrinter.typeString(short)
+    }
+    /*
+     * Get the exact position in ValDef pattern for val (a, b) = (1, 2)
+     * Suprisingly, val ((a, c), b) = ((1, 3), 2) will be covered by Bind
+     * https://github.com/lampepfl/dotty/issues/12627
+     */
+    def findTuplePart(
```
dos65, line 81:
```suggestion
    def editForTupleUnapply(
```

dos65 (commented):
Looks good!
I left only minor comments

mtags/src/main/scala-3/scala/meta/internal/pc/NamesInScope.scala
```diff
@@ -89,32 +92,39 @@ object NamesInScope {
...
-        }
+    val fromImports = ListBuffer.empty[Symbol]
+    val renames = ListBuffer.empty[(SimpleName, String)]
+    ctx.outersIterator.toList.foreach { ctx =>
+      Option(ctx.importInfo) match {
+        case Some(imp) =>
+          imp.selectors.foreach { sel =>
+            sel.renamed match {
+              case Ident(name) if name.toString != "_" =>
+                renames += ((sel.imported.name.toSimpleName, name.toString))
```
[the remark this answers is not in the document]
tgodzik: Och, do you mean that imports such as:
```
object A{
  import a.b.C as D
}
```
?

mtags/src/main/scala-3/scala/meta/internal/pc/Signatures.scala
```diff
@@ -64,6 +84,12 @@ object ShortName {
     ShortName(sym.name, sym)
 }
 
+case class PrettyType(name: String) extends Type {
+  def hash: Int = 0
+  def computeHash(bind: Binders) = hash
+  override def toString = name
+}
```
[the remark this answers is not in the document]
tgodzik: It would be best to probably just print the types on the go, but that is rather complex. Anyway, looks like a good idea, I will see if I can use it here.

mtags/src/main/scala-3/scala/meta/internal/pc/NamesInScope.scala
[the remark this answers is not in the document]
dos65: no, I mean that import renames are processed in `inspectImports` too and probably should be merged with renames that comes from `ctx.importInfo`: https://github.com/scalameta/metals/blob/7318b90e9782efc12632779119aee7f389515d56/mtags/src/main/scala-3/scala/meta/internal/pc/NamesInScope.scala#L170-L173

mtags/src/main/scala-3/scala/meta/internal/pc/Signatures.scala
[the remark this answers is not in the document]
dos65: Cool but anyway - feel free to ignore that. It's more a thing of taste and as it works for scala2 it shouldn't introduce any significant problem.

mtags/src/main/scala-3/scala/meta/internal/pc/NamesInScope.scala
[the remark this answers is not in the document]
dos65: This isn't actual anymore. I found the original problem and will delete this method soon

dos65 (approved):
I think this one is ready. It's needed just to merge with changes from main

mtags/src/main/scala-3/scala/meta/internal/pc/Signatures.scala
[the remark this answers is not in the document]
tgodzik: Ok, let's leave it for now, we might need to work on it anyway in the future.

PR #2826, opened 2021-05-28
Add support for 2.12.14

dos65:


build.sbt
```diff
@@ -225,6 +226,7 @@ lazy val V = new {
     Seq(
       scala213,
       scala212,
+      "2.12.14",
```
tgodzik, line 229: You should change scala212 to be `2.12.14`, we try to keep on the latest version in Metals, and add `2.12.13` here. You will most likely need to set ammonite212Version to `2.12.14` if it's not yet published.

Also, please update `bin/test-release.sh`

dos65: scalafix isn't published for 2.12.14 yet and fails with:
```
[error] error: The ExplicitResultTypes rule only supports the exact Scala versions 
'List(2.12.13, 2.12.11, 2.12.12)' for this binary version.
```

tgodzik: ach, ok, we can update it later.

dos65: and the same for ammonite

tgodzik (approved):
LGTM!

build.sbt
```diff
@@ -134,6 +134,7 @@ inThisBuild(
     ),
     testFrameworks := List(),
     resolvers += Resolver.sonatypeRepo("public"),
+    resolvers += Resolver.sonatypeRepo("snapshot"),
```
ckipp01, line 137: I know this is already merged, but does the snapshot resolver need to stay in here?

tgodzik: It's useful for testing out snapshot releases. We can remove it for sure, but it shouldn't matter too much.

PR #2850, opened 2021-06-01
Revert Bloop version upgrade

tgodzik:
~~We recently updated Bloop to use the scala-debug-adapter library, which seems to work correctly when terminated is set to false. Bloop always seemed to terminate the process irrelevant of the flag.~~

Ok, since it's failing in almost every PR and my change is not helping I will take a look at it later on.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -196,7 +196,7 @@ lazy val V = new {
   val scalameta = "4.4.20"
   val semanticdb = scalameta
   val bsp = "2.0.0-M13"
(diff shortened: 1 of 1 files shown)


tgodzik:
So this seems to fix things locally, but on CI it still randomly breaks :/

dos65:
There is an issue in NuProcess about this - https://github.com/brettwooldridge/NuProcess/issues/128

dos65 (approved):

PR #2851, opened 2021-06-01
Update amonite to 2.3.8-124-2da846d2 and add support for 2.13.6

tgodzik:
succeeds https://github.com/scalameta/metals/pull/2837


tgodzik:
This is just a follow up from the ammonite update, I will go ahead and merge.

PR #2852, opened 2021-06-02
Add support for Scala 3.0.1-RC1

tgodzik:
This also includes a bunch of fixes that were kind of a chain reaction due to some tests failing for 3.0.1-RC1
- print outer type instead of singleton type for objects
- make sure to take method signature into account when filtering symbols

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
@@ -89,13 +89,34 @@ class CompletionProvider(
       s.sourcePos.exists && s1.sourcePos.exists && s.sourcePos.point > s1.sourcePos.point
   }
 
+  extension (sym: Symbol) {
+    def detailString: String = {
+      if (sym.is(Method)) {
+        val sig = sym.signature
+        val sigString =
+          if (sig.paramsSig.isEmpty) "()"
+          else
+            sig.paramsSig
+              .map(p => p.toString)
+              .mkString("(", ",", ")")
+        sym.showFullName + sigString
+      } else {
+        sym.showFullName
+      }
+    }
+  }
   extension (l: List[CompletionValue]) {
     def filterInteresting(): (List[CompletionValue], SymbolSearch.Result) = {
       val isSeen = mutable.Set.empty[String]
       val buf = List.newBuilder[CompletionValue]
       def visit(head: CompletionValue): Boolean = {
         val sym = head.value.sym
-        val id = sym.showFullName
+        val id = head match {
+          case _: CompletionValue.NamedArg =>
+            sym.detailString + "="
+          case _ =>
+            sym.detailString
+        }
         def isNotLocalForwardReference: Boolean =
           !sym.isLocalToBlock ||
             !sym.srcPos.isAfter(pos) ||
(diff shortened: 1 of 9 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -151,7 +172,9 @@ class CompletionProvider(
 
     def hasGetter = try {
       // isField returns true for some classes
-      sym.isField || sym.getter != NoSymbol
+      (sym.isField && !sym.is(JavaDefined)
```
tgodzik, line 175: sym.isField was returning true for a bunch of java classes

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -259,7 +282,7 @@ class CompletionProvider(
                     )
                     if (byParamCount != 0) byParamCount
                     else {
-                      s1.show.compareTo(s2.show)
+                      s1.detailString.compareTo(s2.detailString)
```
tgodzik, line 285: We should compare the full signature, show for symbol was mostly just name, which wasn't an issue prior to 3.0.1, since before the completions with the same symbol name were grouped together.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -293,34 +293,33 @@ case class ScalaPresentationCompiler(
...
-              symbols.flatMap(ParsedComment.docOf(_)(using ctx))
-            val keywordName = symbols.headOption.map { symbol =>
-              printer.fullDefinition(
-                symbol,
-                tpw
-              )
-            }
-            val typeString = symbols.headOption.map { symbol =>
+              symbols.flatMap(ParsedComment.docOf(_))
+            val hoverString = symbols.headOption.map { symbol =>
```
tgodzik, line 313: Improved it to return the full hover string here, full separation was actually problematic to maintain.

tests/cross/src/test/scala/tests/hover/HoverDefnSuite.scala
```diff
@@ -147,7 +147,7 @@ class HoverDefnSuite extends BaseHoverSuite {
       |""".stripMargin,
     "",
     compat = Map(
-      "3.0" -> "object MyObject: MyObject".hover
+      "3.0" -> "object MyObject: object".hover
```
tgodzik, line 150: `object` is the name of the package, same in all the other tests.

tests/cross/src/test/scala/tests/hover/HoverTermSuite.scala
```diff
@@ -147,7 +147,7 @@ class HoverTermSuite extends BaseHoverSuite {
        |""".stripMargin.hover,
     compat = Map(
       // https://github.com/lampepfl/dotty/issues/8835
-      "3.0" -> "object num: Xtension#num".hover
+      "3.0" -> "object num: interpolator-unapply.a$.Xtension".hover
```
tgodzik, line 150: This is wrong anyway, we can fix it in a separate PR.

tgodzik:
Integration tests seem to be failing unrelated to the changes. I need to take a look at it separately, I think it's one of the test that randomly start failing :(

dos65 (approved):
LGTM!

tgodzik:
The weird test failure disappeared :thinking: not sure what it was caused by.

PR #2853, opened 2021-06-04
Try to fix ClasspathSymbolRegressionSuite

tgodzik:


tgodzik:
Well, this seems to be a problem only on my PR :(

Issue #2856, opened 2020-05-11
Scalafix support for Scala 3


tgodzik:
**Is your feature request related to a problem? Please describe.**
Currently we are working on properly supporting Scalafix inside Metals, however there is also a need for such an integration for Scala 3. This work will be done concurrently, since first what is required is the Scalameta parser working for Scala 3 https://github.com/scalameta/scalameta/issues/2022 , then work on Scalafix itself in the Scalafix repository and finally after https://github.com/scalameta/metals/issues/1675 we should be able to 
introduce it to Metals.

**Describe the solution you'd like**
Scalafix in Metals for Scala 3

**Additional context**
This is part of the ongoing work on the Scala 3 support in Metals

**Search terms**:
scalafix scala3

PR #2857, opened 2021-06-05
Enable organize imports for Scala 3

tgodzik:
Since 0.9.29 it's possible to use most of the existing rules on Scala 3 including organize-imports. This still have 2 caveats:
- unused imports will not be removed since we don't get that information from the compiler
- we will need to run 2.13 scalafix for Scala 3 sources until Scalafix is released for Scala 3. This is only a problem for ExplicitResultTypes, which anyway wouldn't work in Metals.

Fixes https://github.com/scalameta/metals/issues/2856


dos65 (approved):
:rocket:

PR #2858, opened 2021-06-05
Run organize imports on Scala 3 sources

tgodzik:
It's not yet possible to properly enable it for CI, since some of the rules will not work and we don't have a similar thing to scalafmt where we can enable a certain config for a subset of files a.k.a. `fileOverrides`


ckipp01 (approved):
LGTM!

PR #2860, opened 2021-06-05
Enable organize-imports for mtags3

dos65:


build.sbt
```diff
@@ -171,6 +171,7 @@ Global / onLoad ~= { old =>
   old
 }
 Global / cancelable := true
+Global / excludeLintKeys += scalafixConfig
```
tgodzik, line 174: What does this do?

tgodzik (approved):
Och cool, I didn't think about it.

build.sbt
dos65: sbt wrongly warns that `scalafixConfig` setting isn't used

bjaglin: I just opened https://github.com/scalacenter/sbt-scalafix/pull/233 to avoid that

PR #2862, opened 2021-06-07
Add support for `textDocument/selectionRange`.

### LINKED ISSUE #2861: Implement LSP textDocument/selectionRange
In the upcoming LSP v3.15 specification there is a new `textDocument/selectionRange` https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#textDocument_selectionRange

If I understand correctly, this feature corresponds to IntelliJ's "expand selection" feature which is one of the biggest features that I'm missing with Metals.

ckipp01:
While I still have a couple things yet to figure out with this, I wanted to through it up here to get any thoughts. Mainly because I'm hitting on (what I would call) unclarity in the spec about how this is exactly supposed to work. The interesting part is that according to the [spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_selectionRange), the client can send in multiple positions in the `selectionRangeParams`. From this point, things get interesting. I'll outline the two editors that I was testing in because they behave pretty differently even though the result is pretty much the same.

## VS Code

The steps it follows:
- Grabs the position your cursor is in, and sends that in as a single position
- The selection range is returned and each range has the parent filled in. So you get the entire tree
- When you continue the selection range, no further requests are made to the server. This is actually all done with one request, and it seems that if the full tree is there, then it just skips making any additional requests

This works pretty well and is actually pretty great that no other requests are necessary. However, it then begs the question of when should VS Code ever send in more than a single position for this request? I actually have no clue. Even selecting a range and triggering it from the range has VS Code sending in the start of the range, not the entire range or multiple positions.

So contrast this with coc-metals utilizing coc.nvim

## coc-metals

So while this looks very similar, it's radically different in the LSP communication.

The steps it follows:
- Grabs the position your cursor is in, and sends that in as a single position
- The selection range is returned and each range has the parent filled in. So you get the entire tree
- When you continue the selection range an entire new request is sent, this time with first returned range split. Meaning the first position in the array was the start position of the returned range, and the second positions is the end position in the returned range
- Now there are two different selection ranges returned. In some scenarios (most scenarios from what I can tell) the actually return the same tree since the closes enclosing tree is the same whether you approach it from the start or the end. - It somehow intelligently skips the first enclosing tree if it's the same that was highlighted before and then continues on.
- Again, the further you expand, new requests are sent every time, but never containing more than 2 positions.

## Some thoughts

If you watch closely, you'll notice a very slight difference in the VS Code gif where it has one more selection range that goes to the beginning of the line. That's actually not part of the selection range returned, so I have no idea why it does that. It's almost seems like they are doing a bit more behind the scenes than the spec outlines, but still mainly following the full tree that is returned.

It doesn't bother me that much since this works pretty well in both editors, but it matters in how we test this. I'm a bit stuck at the moment how to test the step after the initial expansion since there never is a next step in VS Code, and in coc.nvim it also seems to be doing some stuff behind the scenes to choose the next range correctly. I've a snippet of trace code in the test file if you're curious what it looks like. I see a couple options.

1. We can only do an initial call, get the full selection range tree, and then just ensure that all the ranges are the logical ones that we expect.
2. Somehow create another request after the initial one with the returned position split similar to how coc.nvim is doing it. Although then we'll have to make some opinionated decisions about how to handle the multiple returned selection ranges.

Any opinions or thoughts on this?

From looking at [eclipse.jdt.ls](https://github.com/eclipse/eclipse.jdt.ls/blob/64b15c5a9e5b11f62ceb5163ceb6930d5dea7129/org.eclipse.jdt.ls.tests/src/org/eclipse/jdt/ls/core/internal/handlers/SelectionRangeHandlerTest.java) it looks like they just pass in a single position and then ensures all the positions in the selection range are as expected. I'm tempted to take this approach as well.

Closes #2861

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -76,6 +76,7 @@ import org.eclipse.lsp4j.MarkupContent
 import org.eclipse.lsp4j.ParameterInformation
 import org.eclipse.lsp4j.Position
 import org.eclipse.lsp4j.Range
+import org.eclipse.lsp4j.SelectionRange
 import org.eclipse.lsp4j.SignatureHelp
 import org.eclipse.lsp4j.SignatureInformation
 import org.eclipse.lsp4j.TextEdit
@@ -282,6 +283,14 @@ case class ScalaPresentationCompiler(
         .inferredTypeEdits()
         .asJava
     }
+  }
+
+  override def selectionRange(
+      params: ju.List[OffsetParams]
+  ): CompletableFuture[ju.List[SelectionRange]] = {
+    CompletableFuture.completedFuture(
+      List.empty[SelectionRange].asJava
+    )
   }
 
   def hover(params: OffsetParams): CompletableFuture[ju.Optional[Hover]] =
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -288,9 +288,15 @@ case class ScalaPresentationCompiler(
   override def selectionRange(
       params: ju.List[OffsetParams]
   ): CompletableFuture[ju.List[SelectionRange]] = {
-    CompletableFuture.completedFuture(
-      List.empty[SelectionRange].asJava
-    )
+    CompletableFuture.completedFuture {
+      compilerAccess.withSharedCompiler(List.empty[SelectionRange].asJava) {
+        pc =>
+          new SelectionRangeProvider(
+            pc.compiler(),
+            params
+          ).selectionRange().asJava
+      }
+    }
   }
 
   def hover(params: OffsetParams): CompletableFuture[ju.Optional[Hover]] =
--- a/tests/unit/src/main/scala/tests/TestingServer.scala
+++ b/tests/unit/src/main/scala/tests/TestingServer.scala
@@ -12,6 +12,7 @@ import java.nio.file.attribute.BasicFileAttributes
 import java.util
 import java.util.Collections
 import java.util.concurrent.ScheduledExecutorService
+import java.{util => ju}
 
 import scala.collection.concurrent.TrieMap
 import scala.collection.mutable
@@ -640,7 +641,7 @@ final class TestingServer(
     val params = new FoldingRangeRequestParams(new TextDocumentIdentifier(uri))
     for {
       ranges <- server.foldingRange(params).asScala
-      textEdits = FoldingRangesTextEdits(ranges)
+      textEdits = RangesTextEdits.fromFoldingRanges(ranges)
     } yield TextEdits.applyEdits(textContents(filename), textEdits)
   }
 
@@ -651,6 +652,36 @@ final class TestingServer(
       folded <- foldingRange(filename)
       _ = Assertions.assertNoDiff(folded, expected)
     } yield ()
+
+  def retrieveRanges(
+      filename: String,
+      expected: String
+  ): Future[ju.List[l.SelectionRange]] = {
+    val path = toPath(filename)
+    val input = path.toInputFromBuffers(buffers)
+    val offset = expected.indexOf("@@")
+    if (offset < 0) sys.error("missing @@")
+    val start = input.text.indexOf(expected.replace("@@", ""))
+    val point = start + offset
+    val pos = m.Position.Range(input, point, point)
+    val params =
+      new l.SelectionRangeParams(
+        path.toTextDocumentIdentifier,
+        List(pos.toLSP.getStart).asJava
+      )
+
+    server.selectionRange(params).asScala
+  }
+
+  def assertSelectionRanges(
+      filename: String,
+      expected: String,
+      ranges: List[l.SelectionRange]
+  ): Unit = {
+    val edits = RangesTextEdits.fromSelectionRanges(ranges)
+    val edited = TextEdits.applyEdits(textContents(filename), edits)
+    Assertions.assertNoDiff(edited, expected)
+  }
 
   def onTypeFormatting(
       filename: String,
--- a/tests/unit/src/main/scala/tests/TestingServer.scala
+++ b/tests/unit/src/main/scala/tests/TestingServer.scala
@@ -675,12 +675,15 @@ final class TestingServer(
 
   def assertSelectionRanges(
       filename: String,
-      expected: String,
-      ranges: List[l.SelectionRange]
+      ranges: List[l.SelectionRange],
+      expected: List[String]
   ): Unit = {
-    val edits = RangesTextEdits.fromSelectionRanges(ranges)
-    val edited = TextEdits.applyEdits(textContents(filename), edits)
-    Assertions.assertNoDiff(edited, expected)
+    expected.headOption.foreach { expectedRange =>
+      val edits = RangesTextEdits.fromSelectionRanges(ranges)
+      val edited = TextEdits.applyEdits(textContents(filename), edits)
+      Assertions.assertNoDiff(edited, expectedRange)
+      assertSelectionRanges(filename, ranges.map(_.getParent()), expected.tail)
+    }
   }
 
   def onTypeFormatting(
--- a/tests/unit/src/test/scala/tests/SelectionRangeLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/SelectionRangeLspSuite.scala
@@ -42,35 +42,27 @@ class SelectionRangeLspSuite extends BaseLspSuite("selectionRange") {
       )
       _ = server.assertSelectionRanges(
         "a/src/main/scala/a/Main.scala",
-        """|object Main extends App {
-           |  >>region>>def double(a: Int) = {
-           |    a * 2
-           |  }<<region<<
-           |}
-           |""".stripMargin,
-        ranges.asScala.toList
-      )
-      parents = ranges.asScala.toList.map(_.getParent())
-      _ = server.assertSelectionRanges(
-        "a/src/main/scala/a/Main.scala",
-        """|object Main >>region>>extends App {
-           |  def double(a: Int) = {
-           |    a * 2
-           |  }
-           |}<<region<<
-           |""".stripMargin,
-        parents
-      )
-      parantsParents = parents.map(_.getParent())
-      _ = server.assertSelectionRanges(
-        "a/src/main/scala/a/Main.scala",
-        """|>>region>>object Main extends App {
-           |  def double(a: Int) = {
-           |    a * 2
-           |  }
-           |}<<region<<
-           |""".stripMargin,
-        parantsParents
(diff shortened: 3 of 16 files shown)


tests/unit/src/test/scala/tests/SelectionRangeLspSuite.scala
```diff
@@ -0,0 +1,150 @@
...
+ *  since the spec isn't the clearest on the way this is done and different
+ *  LSP clients seems to impliment this differently. Mainly, we mimic what VS
+ *  Code does and what other servers like jdtls do and we send in a single
+ *  position. Once that positions is recieved we check the Selection range and
+ *  all the parents in that range to ensure it contains the positions we'd
+ *  expect it to contain. In the case of VS Code, they never make a second
+ *  request, and instead they just rely on the tree to continue selecting. So
+ *  we mimic that here in the tests.
+ */
+class SelectionRangeLspSuite extends BaseLspSuite("selectionRange") {
```
tgodzik, line 15: We could add some test in `cross` since this is implemented using the compiler? We could reuse them later for Scala 3 and we can create a lot of smaller cases.

For example:
```scala
object Main extends App {
  val name: Option[String] = Option("Tom")
  name match {
    case Some(name) => println("Hello! " + name)
    case None =>
  }
}
```
This one seem particulary weird, since we get range on `case Some(name) => println("Hello! " + name)` and then with whitespaces before.

Is that maybe a tree node that we can skip?

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -652,6 +653,36 @@ final class TestingServer(
...
+    val params =
+      new l.SelectionRangeParams(
+        path.toTextDocumentIdentifier,
+        List(pos.toLSP.getStart).asJava
+      )
+
+    server.selectionRange(params).asScala
+  }
+
+  def assertSelectionRanges(
```
tgodzik, line 676: What about a recursive function:
```scala
  def assertSelectionRanges(     
      filename: String,
      ranges: List[l.SelectionRange],
      expected: *String,
  ): Unit = { expected.headOption.foreach { expectedFile =>
        val edits = RangesTextEdits.fromSelectionRanges(ranges)
        val edited = TextEdits.applyEdits(textContents(filename), edits)
       Assertions.assertNoDiff(edited, expectedFile)
      assertSelectionRanges(filename, ranges.map(_.getParent), expected.tail)
    }
  }

```
I think would look nicer in the test.

tgodzik (changes requested):
Thanks for the great work! The provider itself seems simple, which I was worried it would not be, so hurray!

I left some comments with Scala 3 issue a bit blocking. I wonder if we could just implement it using:

```scala
          val path =
            Interactive.pathTo(driver.openedTrees(uri), pos)(using ctx)
```
for now and optimize later to skip typing. What do you think? I don't want to force it on you to implement, I can take a look later, but at the same time I don't want to worsen Scala 3 experience.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -282,6 +283,14 @@ case class ScalaPresentationCompiler(
     )
   }
 
+  override def selectionRange(
+      params: ju.List[OffsetParams]
+  ): CompletableFuture[ju.List[SelectionRange]] = {
+    CompletableFuture.completedFuture(
+      List.empty[SelectionRange].asJava
```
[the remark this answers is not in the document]
tgodzik: The default selections ranges in VS Code I mean

tgodzik:
```
package scala.meta.internal.pc

import java.{util => ju}

import org.eclipse.lsp4j.SelectionRange
import dotty.tools.dotc.interactive.InteractiveDriver
import dotty.tools.dotc.core.Contexts.Context
import dotty.tools.dotc.interactive.Interactive
import scala.meta.pc.OffsetParams
import scala.collection.JavaConverters._
import java.nio.file.Paths
import dotty.tools.dotc.util.SourceFile
import scala.meta.internal.mtags.MtagsEnrichments._

/**
 * Provides the functionality necessary for the `textDocument/selectionRange` request.
 *
 * @param compiler Metals Global presentation compiler wrapper.
 * @param params offset params converted from the selectionRange params.
 */
class SelectionRangeProvider(
    driver: InteractiveDriver,
    params: ju.List[OffsetParams]
) {

  /**
   * Get the seletion ranges for the provider params
   *
   * @return selection ranges
   */
  def selectionRange(): List[SelectionRange] = {
    given ctx: Context = driver.currentCtx

    val selectionRanges = params.asScala.toList.map { param =>

      val uri = param.uri
      val filePath = Paths.get(uri)
      val source = SourceFile.virtual(filePath.toString, param.text)
      driver.run(uri, source)
      val pos = driver.sourcePosition(param)
      val path =
        Interactive.pathTo(driver.openedTrees(uri), pos)(using ctx)

      val bareRanges = path
        .map { tree =>
          val selectionRange = new SelectionRange()
          selectionRange.setRange(tree.sourcePos.toLSP)
          selectionRange
        }

      bareRanges.reduceRight(setParent)
    }

    selectionRanges
  }

  private def setParent(
      child: SelectionRange,
      parent: SelectionRange
  ): SelectionRange = {
    child.setParent(parent)
    child
  }

}

```
This seems to work, though not perfect. I think wes till need to add the cross tests and figure out edge cases like for comprehensions or match clauses.

dos65:
Hmm. I'm wondering if there is a need to implement this thing on PC side?
It seems that it should be possible to implement using scalameta parser. That would allow having a single implementation for both scala2 and scal3 versions.

tgodzik:
> Hmm. I'm wondering if there is a need to implement this thing on PC side?
> It seems that it should be possible to implement using scalameta parser. That would allow having a single implementation for both scala2 and scal3 versions.

The compiler parser works even in case of errors, which is important especially when manipulating code around.

tests/unit/src/test/scala/tests/SelectionRangeLspSuite.scala
tgodzik: Or maybe the whitespaces selection is caused by VS Code? Anyway, I would for sure add cross tests with some edge cases as mentioned below.

tests/unit/src/main/scala/tests/TestingServer.scala
tgodzik: Or even we could put all ranges in a single file the same as the FoldingRangeSuite does it

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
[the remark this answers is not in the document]
ckipp01: Ahh I didn't even think about this screwing up the defaults in VS Code for Scala 3. That's a huge bummer. No worries, I'll work on getting Scala 3 support in here as well then.

tests/unit/src/main/scala/tests/TestingServer.scala
ckipp01: Ahh I like the recursive function idea. That will make it much cleaner. Let me play around with this a bit.

ckipp01:
> Hmm. I'm wondering if there is a need to implement this thing on PC side?
It seems that it should be possible to implement using scalameta parser. That would allow having a single implementation for both scala2 and scal3 versions.

I think doing it with the PC is beneficial for the exact reason @tgodzik mentioned with being able to recover from errors in a file. I think the more features we can have that don't require a full compilation the better.

> This seems to work, though not perfect. I think wes till need to add the cross tests and figure out edge cases like for comprehensions or match clauses.

Ahh this is super helpful. I'll dig into this.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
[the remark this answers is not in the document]
ckipp01: Total aside, but testing this even though the selection range comes back empty for me in Scala 3 (in VS Code) the defaults expand selection still works for me with VS Code. When you were testing, what happened?

tests/unit/src/test/scala/tests/SelectionRangeLspSuite.scala
ckipp01: I _believe_ the whitespace all on that single line is VS Code either adding another range or something. I witnessed a lot of odd stuff when testing with VS Code that I can't reproduce when using coc-metals.

 


Either way, I'll for sure add some cross tests.

tgodzik: I was afraid it was because of VS Code. They are probably trying to be smart and add another selection from the start of the line.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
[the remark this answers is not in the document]
tgodzik: Och, just checked again and you are right. I might have checked optional braces, which don't work anyway. So you are fine to ignore Scala 3 at this point, I can add it later on.

ckipp01:
So I added in the framework to do cross tests on this and also added in the `SelectionRangeProvider` for Scala 3. I need to dig in a bit deeper as there are a few things I want to either confirm are bugs, my mistakes, or something else:

- [ ]  I'm getting some duplicates back in the ranges, especially on the Scala 3 side. Sometimes 3 identical ranges in a row, which seems odd that three nodes would be present in the tree with the same ranges, but who knows.
- [ ]  I also noticed some stuff in VS Code that I believe you already identified @tgodzik where it seems it tries to be "smart" and if there isn't a range that extends to the beginning of the line it like adds a new one in without making any call to the server.
- [ ] Some of the discrepancies between 2 and 3  that I'd like to figure out, especially with the endings. You can see this in the following example

```scala
// scala 2
object Main >>region>>extends App {
  val total = for {
    a <- Some(1)
    b <- Some(2)
  } yield a + b
}<<region<<
```
```scala
// scala 3
object Main extends >>region>>App {
  val total = for {
    a <- Some(1)
    b <- Some(2)
  } yield a + b<<region<<
}
```

tgodzik:
> [ ] I also noticed some stuff in VS Code that I believe you already identified @tgodzik where it seems it tries to be "smart" and if there isn't a range that extends to the beginning of the line it like adds a new one in without making any call to the server.

I don't think we can go around this unfortunately :/

>  I'm getting some duplicates back in the ranges, especially on the Scala 3 side. Sometimes 3 identical ranges in a row, which seems odd that three nodes would be present in the tree with the same ranges, but who knows.

I think we can just modify the algorithm to skip a node if the range is the same, this might be some internal Scala 3 stuff popping up.

>  Some of the discrepancies between 2 and 3 that I'd like to figure out, especially with the endings. You can see this in the following example

The `extends` is something I think no one can agree on where it should belong to 😅  I think if the cases are small it's fine to leave as is and can later try to fix it in the compiler.

ckipp01:
Alright, I think this should be good for another look over. If there are any specific tests that you think I should add let me know!

tgodzik (approved):
LGTM! Great work :clap:

PR #2863, opened 2021-06-07
WorkspaceSearchVisitior - make search more predictable.

dos65:
Fixes flaky `ClasspathSymbolRegressionSuite.Map#Entry` test.

Testing classpath is messed by netty artifacts that spotted [this problem](https://github.com/scalameta/metals/pull/2814/checks?check_run_id=2763907639#step:5:1187):
- `io:netty.netty-all:4.0.43.Final` - `io.netty.util.collection.IntObjectMap#Entry`
- `io.netty.netty-common:4.1.17.Final` - `io.netty.util.collection.IntObjectMap#PrimitiveEntry`

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/WorkspaceSearchVisitor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/WorkspaceSearchVisitor.scala
@@ -86,10 +86,14 @@ class WorkspaceSearchVisitor(
   ): Option[SymbolDefinition] = {
     val nme = Classfile.name(filename)
     val tpe = Symbol(Symbols.Global(pkg, Descriptor.Type(nme)))
-    index.definition(tpe).orElse {
+
+    val forTpe = index.definitions(tpe)
+    val defs = if (forTpe.isEmpty) {
       val term = Symbol(Symbols.Global(pkg, Descriptor.Term(nme)))
-      index.definition(term)
-    }
+      index.definitions(term)
+    } else forTpe
+
+    defs.sortBy(_.path.toURI.toString).headOption
   }
   override def shouldVisitPackage(pkg: String): Boolean = true
   override def visitWorkspaceSymbol(


metals/src/main/scala/scala/meta/internal/metals/WorkspaceSearchVisitor.scala
```diff
@@ -86,10 +86,14 @@ class WorkspaceSearchVisitor(
...
+
+    val forTpe = index.definitions(tpe)
+    val defs = if (forTpe.isEmpty) {
       val term = Symbol(Symbols.Global(pkg, Descriptor.Term(nme)))
-      index.definition(term)
-    }
+      index.definitions(term)
+    } else forTpe
+
+    defs.sortBy(_.path.toURI.toString).headOption
```
ckipp01, line 96: Just to make sure I understand this. In your example with:

```
io:netty.netty-all:4.0.43.Final - io.netty.util.collection.IntObjectMap#Entry
io.netty.netty-common:4.1.17.Final - io.netty.util.collection.IntObjectMap#PrimitiveEntry
```
This is to ensure that we are always getting the same one back right? But aren't these technically two different symbols? How come `index.definition(tpe)` was returning a random one if they are unique? I'm not super familiar with the GlobalSymbolIndex so I may just totally be missing something.

metals/src/main/scala/scala/meta/internal/metals/WorkspaceSearchVisitor.scala
```diff
@@ -86,10 +86,14 @@ class WorkspaceSearchVisitor(
   ): Option[SymbolDefinition] = {
     val nme = Classfile.name(filename)
     val tpe = Symbol(Symbols.Global(pkg, Descriptor.Type(nme)))
-    index.definition(tpe).orElse {
+
+    val forTpe = index.definitions(tpe)
```
ckipp01, line 90: Does `forTpe` actually mean something here that I'm unaware of? Would it be more clear to just call this `symbolDefenitions` or something?

metals/src/main/scala/scala/meta/internal/metals/WorkspaceSearchVisitor.scala
dos65: They are different but here `index.definitions` is performed only to find the location of a top-level class.

In this test  classpath search found these two class files wich has name that fit query:
  - `io/netty/util/colletion/IntObjectMap$Entry.class`
  - `io/netty/util/colletion/IntObjectMap$PrimitiveEntry.class`

Then it calls `index.definition` to find toplevel symbols `io.netty.util.collection.IntObjectMap` to perform actual matching against semantic defitions extracted from source file.

metals/src/main/scala/scala/meta/internal/metals/WorkspaceSearchVisitor.scala
dos65: It means that it tries to find a definition for typeSymbol.
If it's empty it also tries to find termSymbol.
Unfortunately, there is no equivalent to `orElse` combinator for empty List so I had to extract it into val

metals/src/main/scala/scala/meta/internal/metals/WorkspaceSearchVisitor.scala
ckipp01: Ahh ok, thanks for explaining 👍🏼

ckipp01 (approved):
Looks good to me. LGTM!

Issue #2163, opened 2020-10-29
How to approach the `TestingServer` as we continue to get more BSP servers


ckipp01:
## Discussion

Up to this point Metals has heavily relied (and will continue to rely) on Bloop for our testing server. Pretty much all of our suites that involve any type of `initialize` in the LSP sense rely on a build server connection. During the `initialize` we currently do a `bloopInstall` on the workspace and use Bloop as a build server. You can see this in the `initialize` method here:

https://github.com/scalameta/metals/blob/17ed475be308ef3f914a806af00d32e8fc39cff2/tests/unit/src/main/scala/tests/TestingServer.scala#L381-L389

With the new BSP support in sbt, especially the integrated support coming in https://github.com/scalameta/metals/pull/2154, and also seeing that Mill has also introduced BSP support, I think it's a good idea to discuss the future of how we'd like to test. While we do have tests that cover BSP discovery, we also have a lot of tests the rely on a build server being present. A couple things I've been thinking about:

1. The possibility of introducing multiple different "test servers" to use in the various tests that are specific to the build server (Bloop, sbt, Mill, maybe more?)
2. Refactoring the `TestingServer` to basically be able to "swap out" the build server
3. Do we duplicate all the LSP related tests with each server? Or do we simply make sure the connection work correctly, and focus on one build server for all the heavy LSP related stuff?
4. Other concerns I haven't yet thought about...

I'd love to get some more input and discussion on this topic to ensure that whatever additional support we continue to add for other build servers is properly tested and as stable as it currently is with Bloop.

**Search terms**

BSP servers, BSP testing

olafurpg:
Thank you for bringing this up! I agree the current testing strategy does not scale well with the increasing number of new build servers. It would be lovely if we could somehow run existing tests against different build servers.

One possible approach would be to extend `QuickBuild` to generate sbt/mill builds. That would allow us to continue to declare the build structure via `metals.json` but have the tests run against mill/sbt instead of Bloop.

tgodzik:
I think it would make sense to run tests with different BSP server in the slow tests. We could add there a couple of basic tests to check if everything is working. I don't think we should duplicate all the tests as we are already running a lot of them and I doubt I will be able to optimize anything further.

adpi2:
I would like to add more tests on the integration with sbt server so I am currently trying to do something like 1. and 2. as in:

[...]

PR #2868, opened 2021-06-09
Lazily get explicit choice.

### LINKED ISSUE #2866: Does not connect after GenerateBspConfig
**Describe the bug**

Metals does not connect to sbt server after running `GenerateBspConfig` in a freshly cloned project

**To Reproduce**

1. Clone new project (or run `git clean -dfx`)
2. Open VS code and wait for Metals to start 
3. Don't clik `Import build` in the "New sbt workspace" notification
4. Run `GenerateBspConfig`
5. Metals does not connect to sbt server
6. It does not connect either after running `BspSwitch` or `ConnectBuildServer`

**Expected behavior**

Metals should connect and import the build after running `GenerateBspConfig` or `BspSwitch`

**Screenshots**

Logs are
```text
...
2021.06.09 16:17:01 WARN  Build server is not auto-connectable.
2021.06.09 16:17:27 INFO  sbt 1.5.2 found for workspace.
2021.06.09 16:17:30 INFO  [info] welcome to sbt 1.5.2 (AdoptOpenJDK Java 1.8.0_275)
...
2021.06.09 16:17:41 INFO  [success] Total time: 0 s, completed Jun 9, 2021 4:17:42 PM
2021.06.09 16:17:41 INFO  sbt bspConfig exit: 0
2021.06.09 16:17:42 INFO  time: ran 'sbt bspConfig' in 15s
2021.06.09 16:17:42 WARN  Build server is not auto-connectable.
```

**Installation:**

- Metals version: v0.10.4

**Additional context**

https://github.com/scalameta/metals/blob/ab10007589114d6cf29764f3c8d8dd7486160446/metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala#L38-L40

The root cause is probably that `explicitChoiceMade` is false. Maybe it should be mutated to true when running `GenerateBspConfig` or `BspSwitch`.

**Search terms**

sbt server bsp switch

ckipp01:
This fixes #2866 where with a very fresh project if you don't do a
BloopInstall and then you do a `generate-bsp-config` the explicit choice
gets inserted but the value we have is still the old one. This will
lazily get it to ensure we have the latest.

Note I didn't add an explicit test for this since we do a `bloopInstall`
right away in the build initialize in the Testing Server so it's hard to
actually test without making some large changes to that (which I'd like
to do in a separate pr).

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
@@ -25,7 +25,7 @@ final class BuildTools(
     workspace: AbsolutePath,
     bspGlobalDirectories: List[AbsolutePath],
     userConfig: () => UserConfiguration,
-    explicitChoiceMade: Boolean
+    explicitChoiceMade: () => Boolean
 ) {
   // NOTE: We do a couple extra check here before we say a workspace with a
   // `.bsp` is auto-connectable, and we ensure that a user has explicity chosen
@@ -36,7 +36,7 @@ final class BuildTools(
   // Bloop since Metals thinks it's in state that's auto-connectable before the
   // user is even prompted.
   def isAutoConnectable: Boolean = {
-    isBloop || (isBsp && all.isEmpty) || (isBsp && explicitChoiceMade)
+    isBloop || (isBsp && all.isEmpty) || (isBsp && explicitChoiceMade())
   }
   def isBloop: Boolean = {
     hasJsonFile(workspace.resolve(".bloop"))
@@ -128,6 +128,6 @@ object BuildTools {
       workspace,
       Nil,
       () => UserConfiguration(),
-      explicitChoiceMade = false
+      explicitChoiceMade = () => false
     )
 }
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM!

adpi2:
Thanks @ckipp01 for the fix!

> Note I didn't add an explicit test for this since we do a bloopInstall
right away in the build initialize in the Testing Server so it's hard to
actually test without making some large changes to that (which I'd like
to do in a separate pr).

It would be great to have that! Let me know if you need help.

ckipp01:
> It would be great to have that! Let me know if you need help.

It's basically this https://github.com/scalameta/metals/issues/2163. The way we initialize the testing server assumes a `bloopInstall` happens. My thought was to just allow for something like an `Option[BuildTool with BuildServerProvider]` to be passed in and if there is `None` than we do the `bloopInstall` like before and if not, then we'd add in the functionality to handle the install from other servers. This would help with all sorts of testing since we can essentially test almost all of the functionality simply by passing in another build server. The thing we may hit on though is the `bloopInstall` is crazy fast in that initialize, and the sbt process might be quite a bit slower to start up and load up the build.

Issue #2869, opened 2021-06-10
"save as" duplicates packages


odersky:
**Describe the bug**

An open buffer in `dotty/tools/dotc/C.scala` starting with:
```
package dotty.tools.dotc

class C ...
```
Command: "Save as", argument: `C1.scala`

This gives a buffer
```
package dotty.tools.dotc

package dotty.tools.dotc

class C ...
```

**Expected behavior**

If there is a package clause given, that package clause should be either removed, or no new clause should be inserted. In any case, the package clause should not be duplicated.

dos65:
Thanks for reporting!

I can't reproduce but it seems like a possible bug as we have seen the similar one in original PR #2723.

@Giggiux I found that we replaced `lsp.ResourceOperaion` by command usage only for `Extract` action. I'm afraid there are no other options except of doing the same for `Rename`

tgodzik:
@dos65 I managed to reproduce it, but I am not sure how to fix it. Tried it with `DidCreateFile` notification, but it seems that when the new file is first opened it's empty, so we add the package, but it's not really empty. The main problem seems to be that the file write is not atomic, VS Code first creates an empty file, opens it and then copies over the content.

I think we might need to leave it as is for now :/ The workaround is just using Ctrl+Z (undo) in those situations.

dos65:
@tgodzik It shouldn't be hard to fix. The only we need is to perform an actual rename operation on metals side instead of relying on editor implementation.

tgodzik:
Th problem is that it's not rename, `Save as` copies the file as a new one, there is no way around it.

gabro:
We already have a check in place to avoid inserting the package statement into a file that contains content, but it's currently checking `path.toFile` here:

https://github.com/scalameta/metals/blob/48157e811b9a039174f5eebd95d371ef1488bfe1/metals/src/main/scala/scala/meta/internal/metals/PackageProvider.scala#L51-L51

What we could try instead is to check the content of the params of `didOpen`, specifically `params.getTextDocument().getText()`.

It's an educated guess for now, but maybe that text document is populated when the "Save as" happens whereas `path.toFile` isn't?

ckipp01:
To follow up on your comment in #2871  @tgodzik 

[...]

PR #2871, opened 2021-06-10
Use workspace/didCreateFile for automatically filling package name

tgodzik:
Since 3.16 (I think) there is an ability to register for FileCreated notifications, which we can use for PackageProvider. This is a better way of doing it, because we now get that information from the editor and not assume a file is new if it's empty.

For any client that doesn't support these notifications we keep the old way of adding the package.

Connected to https://github.com/scalameta/metals/issues/2869, but doesn't fix it unfortunately.


tgodzik:
I am thinking we might need to go another way and only add package when files are created via `new Scala file` command, we would need an ability to create an empty file though to make it possible for people to just start with a clean slate. We are unable currently to recognize when a file is really created empty, which can cause a bunch of unrelated errors and contribute to the `lack of polish` feel in Metals.

PR #2876, opened 2021-06-13
Ensure a valid URI is returned for non-html analyze stacktrace.

ckipp01:
`coc-metals` was crazy lenient with this. Nvim core LSP, not so much. Just ensure that the path in the location we are returning is a valid URI.


tgodzik (approved):
LGTM!

PR #2889, opened 2021-06-17
Get rid of warning about setting -Xsemanticdb twice.

ckipp01:
This is already set in the shared settings for Scala 3, so we don't
need it here as well


dos65 (approved):
:+1:

Issue #2820, opened 2021-05-25
Metals process doesn't exit after closing VS Code


dos65:
1. Open Vscode and wait when Metals fully start
2. Do Reload Window (CTRL +R ) / Close editor
3. Check processes: ` ps ax | grep metals` - there will be handed one

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.10.3

**Additional context**

The interesting thing that this situation doesn't happen in the case of the usual exit from VSCode.
It might be related to `PC` shutdown. I noticed that the hanging process has one running `PC` thread.

tgodzik:
Thanks for reporting! I think I found it recently also, but this might be more to do with the metals-vscode plugin? Or alternatively we can add a timeout in the shutdown method to make sure it's always shutdown?

dos65:
__Update__: I just found that the same happens even in case if the editor was closed in the usual manner.
Will change the title.

@tgodzik Are you about `ScalaPresentationCompiler.shutdown` method?
According to `thread dump`, PC thread stucks at [`waitForMoreWork`](https://github.com/scalameta/metals/blob/main/mtags/src/main/scala-2/scala/tools/nsc/interactive/MetalsGlobalThread.scala#L24)

tgodzik:
Maybe shutdown is never run? I meant the shutdown method in the MetalsLanguageServer, it should cancel everything and exit.

dos65:
I will try to investigate it deeper.

I've already tried to fix it but without success and `cancel` were called correctly.

MRandl:
Is this the same as #2722 ? If so this is a known feature

dos65:
@MRandl it's a different one. In this case, I can confirm that a dangling process is Metals itself

PR #2890, opened 2021-06-18
Fix CompilerJobQueue concurrent access/shutdown

dos65:
Previously, it was possible that `CompilerJobQueue` might start a new `ThreadPoolExecutor` after `shutdown` call that lead to thread hanging and issue with server shutdown.

Fixes: #2820


tgodzik (approved):
LGTM, just tested it and it seems to help a lot!

Issue #2893, opened 2021-06-21
[Scala3] Does not stop on some breakpoints


adpi2:
**Describe the bug**

On some Scala 3 source files, some breakpoints cannot be set or the debugger does not stop on them.

**To Reproduce**

On this piece of code:
![image](https://user-images.githubusercontent.com/13123162/122773578-c36d7000-d2a8-11eb-9fa8-20d5c8a12ee2.png)

**Expected behavior**

The debug server receives 3 `setBreakpoints` requests:
```json
{
  "type": "request",
  "seq": 3,
  "command": "setBreakpoints",
  "arguments": {
    "source": {
      "name": "Hello.scala",
      "path": "dap-fqcn:example.Hello$package"
    },
    "breakpoints": [{"line": 4},{"line": 8}],
    "lines": [4,8],
    "sourceModified": false
  }
}
```
```json
{
  "type": "request",
  "seq": 4,
  "command": "setBreakpoints",
  "arguments": {
    "source": {
      "name": "Hello.scala",
      "path": "dap-fqcn:example"
    },
    "breakpoints": [{"line": 5}],
    "lines": [5],
    "sourceModified": false
  }
}
```
```json
{
  "type": "request",
  "seq": 5,
  "command": "setBreakpoints",
  "arguments": {
    "source": {
      "name": "Hello.scala",
      "path": "dap-fqcn:"
    },
    "breakpoints": [{"line": 9}],
    "lines": [9],
    "sourceModified": false
  }
}
```
The first path is correct: `"dap-fqcn:example.Hello$package"`
But the next one seems wrong: `"dap-fqcn:example"`
And the last one is empty: `"dap-fqcn:"`

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.10.4

**Additional context**

Origin of the issue: https://github.com/scalacenter/scala-debug-adapter/issues/47

**Search terms**

debug, debugger, breakpoint, break point, main annotation, scala 3

PR #2894, opened 2021-06-21
Fix off by one error while setting DAP breakpoints

tgodzik:
Fixes https://github.com/scalameta/metals/issues/2893

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -778,7 +778,7 @@ object MetalsEnrichments
       breakpoint: l.debug.SourceBreakpoint
   ) {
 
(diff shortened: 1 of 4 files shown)


adpi2 (approved):
Thanks!

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -774,6 +774,14 @@ object MetalsEnrichments
       trailingTokens.find(predicate)
   }
 
+  implicit class XtensionSourceBreakpoint(
+      breakpoint: l.debug.SourceBreakpoint
+  ) {
+
+    // LSP Postion is 0-based, while breakpoints are 1-based
```
dos65, line 781:
```suggestion
    // LSP Position is 0-based, while breakpoints are 1-based
```

PR #2897, opened 2021-06-23
Always enable semanticdb for users in sbt (when using sbt BSP).

ckipp01:
Note that we still check the users version of sbt before we
get to this point so it shoulnd't be a risk to just simplify
this and always turn it on.

### DIFF
--- a/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
+++ b/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
@@ -56,11 +56,7 @@ object MetalsPlugin extends AutoPlugin {
            } else None)).distinct
       }
     },
-    semanticdbEnabled := {
-      val old = semanticdbEnabled.value
-      if (ScalaInstance.isDotty(scalaVersion.value)) true
-      else old
-    },
+    semanticdbEnabled := true,
     semanticdbTargetRoot := {
       val in = semanticdbIncludeInJar.value
       if (in) classDirectory.value


sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
```diff
@@ -56,11 +56,7 @@ object MetalsPlugin extends AutoPlugin {
            } else None)).distinct
       }
     },
-    semanticdbEnabled := {
-      val old = semanticdbEnabled.value
-      if (ScalaInstance.isDotty(scalaVersion.value)) true
-      else old
```
ckipp01, line 62: Am I totally missing something with this? I don't remember when I first added this why I did this. Thinking about it now, this should always just be enabled no? We do check the sbt version before we ever even get to this point, so there isn't a risk that we are enabling this in a version of sbt that doesn't support it. Also, I'm unsure why we were setting it to `true` for Dotty, but not in Scala 2. I believe this should just be enabled all the time when a user is in a valid sbt version.

dos65: :thinking: It seems that there is some magical thing in sbt-BSP that anyway enables semanticdb for scala2 projects)
Otherwise, I have no idea how it worked before as people rarely enable this setting explicitly.

tgodzik: Wasn't this due to the fact that sbt can have a wrong version of semanticdb? We are setting the compiler plugin separately, so `semanticdbEnabled` should not be needed at all in case of Scala 2. For Scala 3 it's safe as it just adds the -Xsemanticdb flag.

tgodzik (changes requested):

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
ckipp01: ahhhhhh you're right. I totally forgot about this. 👍🏼 I'll close.

tgodzik: Maybe we should add a comment though :P I also took me a while to remember this.

ckipp01: 👍🏼 I'm making some changes to ensure the sbt plugin gets added to the metabuild when the sbt support lands, so I'll add it in that pr.

tgodzik: Thanks!

PR #2898, opened 2021-06-23
Add support for Scala 3.0.1-RC2

tgodzik:


build.sbt
```diff
@@ -200,7 +200,7 @@ lazy val V = new {
   val bsp = "2.0.0-M13"
   val bloop = "1.4.8-19-4d9f966b"
   val scala3 = "3.0.0"
-  val nextScala3RC = "3.0.1-RC1"
+  val nextScala3RC = "3.0.1-RC2"
```
tgodzik, line 203: I don't really want to support all RC versions, I think if users want to check them out then they can just us the latest and that should be fine.

dos65 (approved):

ckipp01 (approved):

PR #2900, opened 2021-06-25
[Scala3] Use symbol search for PC `definitions`.

dos65:
Also, fixes `PCDefitionSuite`. Previsously the checking of non-local location wasn't total.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
@@ -113,11 +113,8 @@ class PcDefinitionProvider(
       // For a named arg, find the target `DefDef` and jump to the param
       case NamedArg(name, _) :: Apply(fn, _) :: _ =>
         val funSym = fn.symbol
-        if (
-          funSym.name == StdNames.nme.copy
-          && funSym.is(Synthetic)
-          && funSym.owner.is(CaseClass)
-        ) List(funSym.owner.info.member(name).symbol)
+        if (funSym.is(Synthetic) && funSym.owner.is(CaseClass))
+          List(funSym.owner.info.member(name).symbol)
         else {
           val classTree = funSym.topLevelClass.asClass.rootTree
           val paramSymbol =
@@ -163,7 +160,7 @@ class PcDefinitionProvider(
 
     def recoverDenot(d: Denotation): List[Symbol] = d match {
       case multi: MultiPreDenotation =>
-        List(multi.last, multi.first)
+        List(multi.first, multi.last)
           .map(_.symbol)
           .filter(_ != NoSymbol)
       case _ => List(d.symbol).filter(_ != NoSymbol)
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
@@ -19,6 +19,7 @@ import dotty.tools.dotc.core.Contexts.Context
 import dotty.tools.dotc.core.Denotations
 import dotty.tools.dotc.core.Denotations.Denotation
 import dotty.tools.dotc.core.Denotations.MultiPreDenotation
+import dotty.tools.dotc.core.Denotations.PreDenotation
 import dotty.tools.dotc.core.Flags._
 import dotty.tools.dotc.core.NameOps._
 import dotty.tools.dotc.core.Names
@@ -158,17 +159,19 @@ class PcDefinitionProvider(
   ): List[Symbol] = {
     import indexed.ctx
 
-    def recoverDenot(d: Denotation): List[Symbol] = d match {
-      case multi: MultiPreDenotation =>
-        List(multi.first, multi.last)
-          .map(_.symbol)
-          .filter(_ != NoSymbol)
-      case _ => List(d.symbol).filter(_ != NoSymbol)
+    def extractSymbols(d: PreDenotation): List[Symbol] = {
+      d match {
+        case multi: MultiPreDenotation =>
+          extractSymbols(multi.denot1) ++ extractSymbols(multi.denot2)
+        case d: Denotation => List(d.symbol)
+        case _ => List.empty
+      }
     }
 
     tree match {
       case select: Select =>
-        recoverDenot(select.qualifier.typeOpt.member(select.name))
+        extractSymbols(select.qualifier.typeOpt.member(select.name))
+          .filter(_ != NoSymbol)
       case ident: Ident => indexed.findSymbol(ident.name).toList.flatten
       case _ => Nil
     }
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
@@ -1,0 +1,197 @@
+package scala.meta.internal.pc
+
+import java.nio.file.Paths
+import java.util.ArrayList
+
+import scala.annotation.tailrec
+import scala.concurrent.Future
+import scala.jdk.CollectionConverters._
+
+import scala.meta.internal.mtags.MtagsEnrichments._
+import scala.meta.pc.DefinitionResult
+import scala.meta.pc.OffsetParams
+import scala.meta.pc.SymbolSearch
+
+import dotty.tools.dotc.ast.tpd
+import dotty.tools.dotc.ast.tpd._
+import dotty.tools.dotc.ast.untpd
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.Denotations
+import dotty.tools.dotc.core.Denotations.Denotation
+import dotty.tools.dotc.core.Denotations.MultiPreDenotation
+import dotty.tools.dotc.core.Flags._
+import dotty.tools.dotc.core.NameOps._
+import dotty.tools.dotc.core.Names
+import dotty.tools.dotc.core.Names._
+import dotty.tools.dotc.core.StdNames
+import dotty.tools.dotc.core.Symbols._
+import dotty.tools.dotc.core.Types.NamedType
+import dotty.tools.dotc.interactive.Interactive
+import dotty.tools.dotc.interactive.InteractiveDriver
+import dotty.tools.dotc.interactive.SourceTree
+import dotty.tools.dotc.transform.SymUtils._
+import dotty.tools.dotc.util.NoSourcePosition
+import dotty.tools.dotc.util.SourceFile
+import dotty.tools.dotc.util.SourcePosition
+import org.eclipse.lsp4j.Location
+
+class PcDefinitionProvider(
+    driver: InteractiveDriver,
+    params: OffsetParams,
+    search: SymbolSearch
+) {
+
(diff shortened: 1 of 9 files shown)


tests/cross/src/test/scala/tests/pc/PcDefinitionSuite.scala
```diff
@@ -161,7 +199,15 @@ class PcDefinitionSuite extends BasePcDefinitionSuite {
        |import scala.concurrent./*scala/concurrent/Future. Future.scala*/@@Future
        |object Main {
        |}
-       |""".stripMargin
+       |""".stripMargin,
+    compat = Map(
+      "3.0" ->
+        """|
+           |import scala.concurrent./*scala/concurrent/Future. Future.scala*//*scala/concurrent/Future# Future.scala*/@@Future
```
tgodzik, line 206: Why do we get both object and class here? Or more persaisly, why we didn't get it in Scala 2?

tests/cross/src/test/scala/tests/pc/PcDefinitionSuite.scala
```diff
@@ -211,7 +257,15 @@ class PcDefinitionSuite extends BasePcDefinitionSuite {
        |object Main {
        |  assert(/*scala/Predef.assert(). Predef.scala*/@@assertion = true)
        |}
-       |""".stripMargin
+       |""".stripMargin,
+    compat = Map(
+      "3.0" ->
+        """|
+           |object Main {
+           |  assert(/*scala/Predef.assert(+1). Predef.scala*/@@assertion = true)
```
tgodzik, line 265: Is it a bug in Scala 2 implementation? It points to the first assert instead of the second

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
```diff
@@ -0,0 +1,202 @@
...
+          .map(_.symbol)
+          .filter(_ != NoSymbol)
+      case Some(denot) => List(denot.symbol).filter(_ != NoSymbol)
+      case None => List.empty
+    }
+  }
+
+  def symOrdering(using Context): Ordering[Symbol] =
+    new Ordering[Symbol] {
+      def compare(first: Symbol, second: Symbol): Int = {
```
tgodzik, line 194: We might compare based on semanticdb symbols, I think that would influence the `assert` ordering. Or alternatively use the one in Completions, where I added the method signature.

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
```diff
@@ -0,0 +1,202 @@
...
+        }
+      case Nil => Nil
+    }
+  }
+
+  private def recoverError(tree: Tree)(using ctx: Context): List[Symbol] = {
+    val recoveredDenot = tree match {
+      case select: Select => Some(select.qualifier.typeOpt.member(select.name))
+      case ident: Ident =>
+        val ref = ctx.typer.findRef(
```
tgodzik, line 168: Would we want to use the `IndexedContext.findSymbol` ?

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
```diff
@@ -0,0 +1,202 @@
...
+            SemanticdbSymbols.symbolName(sym),
+            res
+          )
+        }
+      case Nil => DefinitionResultImpl.empty
+    }
+  }
+  assert(true)
+
+  private def location(p: SourcePosition): Option[Location] = {
```
tgodzik, line 106: We already have `def location` in the ScalaPresentationCompiler, could you move it to extension and move in both places?

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
```diff
@@ -0,0 +1,202 @@
...
+
+  private def findDefinitions(
+      tree: Tree,
+      path: List[Tree],
+      pos: SourcePosition,
+      driver: InteractiveDriver
+  )(using ctx: Context): DefinitionResult = {
+    enclosingSymbols(path, pos) match {
+      case symbols @ (sym :: other) =>
+        val semanticSym = SemanticdbSymbols.symbolName(sym)
```
tgodzik, line 69: Could we move it down just above invocation? So that it's only calculated once if needed

tgodzik (changes requested):
Great work! I have some questions and suggestions, but nothing major.

tests/cross/src/test/scala/tests/pc/PcDefinitionSuite.scala
```diff
@@ -116,7 +116,13 @@ class PcDefinitionSuite extends BasePcDefinitionSuite {
       "2.13" ->
         """|
            |object Main {
-           |  /*scala/collection/immutable/List.apply(). List.scala*//*scala/collection/IterableFactory#apply(). Factory.scala*/List(1)
```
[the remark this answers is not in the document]
dos65: Previously, the test logic around checking non-local definitions was partially incorrect.
For `2.13` and `3.0` PC returns only `/*scala/collection/IterableFactory#apply(). Factory.scala*/`.

`/*scala/collection/immutable/List.apply(). List.scala*/` was needed here only to obtain the same code for assertion.

tests/cross/src/test/scala/tests/pc/PcDefinitionSuite.scala
dos65: I thought that is bug in Scala3 but everything is correct because there are stdLibPatches in scala3 https://github.com/lampepfl/dotty/issues/12942#issuecomment-868668468

The actual location should be [`scala/runtime/stdLibPatches/Predef.assert(+1)`](https://github.com/lampepfl/dotty/blob/master/library/src/scala/runtime/stdLibPatches/Predef.scala#L9) but it think it would be better to fix in separate PR. Probably we might overwrite such symbols somewhere in `DefitionProvider`

tgodzik: Let's follow up with an issue then not to forget and add it to the comment here.

tests/cross/src/test/scala/tests/pc/PcDefinitionSuite.scala
dos65: I thought to check if we can do the same for scala2 but decided to save time and keep PR only about scala3.
I don't think that's critical and it works as it was initially implemented in `Interactive`

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
dos65: moved to extension

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
dos65: Completions doesn't fit for that. Changed this part a little bit

tests/cross/src/main/scala/tests/pc/BasePcDefinitionSuite.scala
```diff
@@ -52,19 +57,25 @@ abstract class BasePcDefinitionSuite extends BasePCSuite {
...
+      val expected = getExpected(original, compat, scalaVersion).removePos
       assertNoDiff(
         obtained,
-        getExpected(expected, compat, scalaVersion)
+        expected
       )
     }
   }
+
+  implicit class CodeCleanup(s: String) {
```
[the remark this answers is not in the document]
dos65: Move to `CrossTestEnrichments`

mtags/src/main/scala-3/scala/meta/internal/pc/SemanticdbSymbols.scala
```diff
@@ -75,4 +80,66 @@ object SemanticdbSymbols {
       case NonFatal(e) => Nil
     }
   }
+
+  /** The semanticdb name of the given symbol */
```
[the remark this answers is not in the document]
dos65: Added that in comment bellow

tests/cross/src/test/scala/tests/pc/PcDefinitionSuite.scala
```diff
@@ -128,7 +134,32 @@ class PcDefinitionSuite extends BasePcDefinitionSuite {
        |object Main {
        |  /*scala/Predef.assert(+1). Predef.scala*//*scala/Predef.assert(). Predef.scala*/@@assert
        |}
-       |""".stripMargin
+       |""".stripMargin,
+    compat = Map(
+      "3.0" ->
+        """|
+           |object Main {
+           |  /*scala/Predef.assert(). Predef.scala*//*scala/Predef.assert(+1). Predef.scala*/assert
```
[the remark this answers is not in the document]
dos65: Added symbol sorting to Scala2 PC too

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
```diff
@@ -0,0 +1,197 @@
...
+      pos: SourcePosition,
+      indexed: IndexedContext
+  ): List[Symbol] = {
+    import indexed.ctx
+    path match {
+      // For a named arg, find the target `DefDef` and jump to the param
+      case NamedArg(name, _) :: Apply(fn, _) :: _ =>
+        val funSym = fn.symbol
+        if (
+          funSym.name == StdNames.nme.copy
```
tgodzik, line 117:
```suggestion
          funSym.name == StdNames.nme.copy
```
what about `apply` ? Wouldn't the two next conditions be enough?

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
```diff
@@ -0,0 +1,197 @@
...
+
+  private def recoverError(
+      tree: Tree,
+      indexed: IndexedContext
+  ): List[Symbol] = {
+    import indexed.ctx
+
+    def recoverDenot(d: Denotation): List[Symbol] = d match {
+      case multi: MultiPreDenotation =>
+        List(multi.last, multi.first)
```
tgodzik, line 166: Why `last` and `first` ? Could you add a comment explaining it here?

tgodzik (changes requested):
Just two more comments I haven't noticed before. Looks amazing already and should be really nice quality of life improvement.

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
dos65: Good catch! Updated and added tests for that.
Honestly, as it's a it was initially copypasted from `Interactive` I just ignored these lines :smile:

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
dos65: Initially, I missed that `MultiDenotation` might contain more than two denotations.
Reworked this part

tgodzik (approved):
LGTM!

Issue #2907, opened 2021-06-30
Could not optimize @tailrec annotated method factorial


sbushmanov:
When I do in scala worksheet:

```scala
  import scala.annotation.tailrec

  @tailrec
>>def factorial(n: BigInt, acc: BigInt): BigInt = {
    if (n <=1) acc
    else factorial(n-1, n*acc)
  }

  factorial(3,1)

```
I'm getting an error:
```
[mdoc] [E] could not optimize @tailrec annotated method factorial: it is
   neither private nor final so can be overridden
```

Expected behavior:
`factorial` should be properly evaluated without an error.

Setup:
- vim version: NVIM v0.5.0-dev+1444-gc1120ad          
- node version: v15.14.0                                
- coc.nvim version: 0.0.80-131f6b267f 
- same error for both bloop and sbt

tgodzik:
Thanks for reporting! We already were fixing it upstream, but forgot to update the fix in Metals. Should work after #2908 is released

sbushmanov:
Perhaps not related, but may I ask why for a problem diagnostic to be identified I need to reload the build?

tgodzik:
You should not need to reload anything to get diagnostics, in case of worksheets it should work on save. However, if there is an internal error you might need to reload the worksheet or force another save.

sbushmanov:
In `*.scala` files I'm only getting diagnostics on save.  In `*.worksheet.sc` I need to reload build.
Am I missing something? 
Or should I open an issue?

tgodzik:
We only publish diagnostics on save aside from syntax errors, so that is expected.

For worksheets only saving is also needed. How do you reloaf the build? What's the snippet that you are using? Probably best to raise a separate issue with a reproduction

Issue #2764, opened 2021-05-02
Value class inside an object gives error: "value classes may not be a member of another class - mdoc"


njlbenn:
**Describe the bug**

In a worksheet I encounter the error `value classes may not be a member of another class - mdoc` when I try to define a Value Class inside an object.

**To Reproduce**

Try to insert code like the following into a worksheet:

```
object Foo {
  case class Bar(b: Int) extends AnyVal
}
```

The worksheet in this case fails to compile and the error displayed is "value classes may not be a member of another class - mdoc"

**Expected behavior**

That this compiles as it does, say, in the vanilla Scala REPL

```
scala> :paste
// Entering paste mode (ctrl-D to finish)

object Foo {
  case class Bar(b: Int) extends AnyVal
}

// Exiting paste mode, now interpreting.

defined object Foo
```

**Installation:**

- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.10.2

**Additional context**

The worksheet is in a standalone context -- i.e. it is not part of a Scala project -- and I have explicitly set the `fallbackScalaVersion` to be 2.13.5

tgodzik:
Thanks for reporting! This is an issue in mdoc, looking at it here https://github.com/scalameta/mdoc/pull/492

For now the best workaround is to have the value class inside normal source and just import it into a workshett.

olafurpg:
FWIW, you can try to not use `AnyVal`. I would be surprised if you notice a performance difference when evaluating your worksheet.

njlbenn:
> FWIW, you can try to not use `AnyVal`. I would be surprised if you notice a performance difference when evaluating your worksheet.

@olafurpg ah, this was just some toy code to demonstrate the use of `AnyVal` - the use wasn't motivated by performance. So the bugfix isn't critical for me. Just thought I would report it

tgodzik:
We need to fix it differently or document it as unsupported in the worksheets.

PR #2908, opened 2021-06-30
Wrap worksheets in object instead of a class

tgodzik:
Previously, we would use the default mdoc modifier for worksheets, which uses a class to wrap statements. Now, we use `reset-object`, which wraps statements in an object and allows us to use some of the features not available in classes.

Fixes https://github.com/scalameta/metals/issues/2764 and fixes https://github.com/scalameta/metals/issues/2907

### DIFF
--- a/tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
+++ b/tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
@@ -179,7 +179,7 @@ abstract class BaseWorksheetLspSuite(scalaVersion: String)
              |```
              |1.to(10).toVector
              |```scala
-             |res1: Vector[Int] = Vector(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
+             |res2: Vector[Int] = Vector(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
              |```
              |val List(a, b) = List(42, 10)
              |```scala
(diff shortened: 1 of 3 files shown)


tests/unit/src/test/scala/tests/worksheets/WorksheetLspSuite.scala
```diff
@@ -115,6 +115,53 @@ class WorksheetLspSuite extends tests.BaseWorksheetLspSuite(V.scala212) {
     } yield ()
   }
 
+  // We use Mdoc with `reset-object` modifier which makes sure that objec is used to wrap statements
```
ckipp01, line 118:
```suggestion
  // We use Mdoc with `reset-object` modifier which makes sure that object is used to wrap statements
```

ckipp01 (approved):
I had totally forgot that we fixed this in mdoc. LGTM!

dos65 (approved):
:+1:

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
```diff
@@ -277,18 +277,36 @@ abstract class BaseWorksheetLspSuite(scalaVersion: String)
       _ = assertNoDiff(
         client.workspaceDiagnostics,
         getExpected(
-          """|a/src/main/scala/Main.worksheet.sc:2:1: error: java.lang.RuntimeException: boom
-             |	at repl.MdocSession$App.<init>(Main.worksheet.sc:11)
+          """|a/src/main/scala/Main.worksheet.sc:2:1: error: java.lang.ExceptionInInitializerError
```
tgodzik, line 280: Because we changed to an object, the exception stack traces changed, but they are shown in the same place. We could possibly improve them on the mdoc side, but they seem ok for now.

PR #2917, opened 2021-07-01
[Scala2] PcDefinition fix

dos65:
At least fixes the following problem that happens in 2.12:
```scala
// Prevously retuned location was /*List.::*/
Option(1).@@toInt :: List(1,2,3)
```

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/PcDefinitionProvider.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/PcDefinitionProvider.scala
@@ -20,6 +20,7 @@ class PcDefinitionProvider(val compiler: MetalsGlobal, params: OffsetParams) {
         params.uri().toString(),
         None
       )
+      typeCheck(unit)
       val pos = unit.position(params.offset())
       val tree = definitionTypedTreeAt(pos)
       if (
@@ -95,7 +96,7 @@ class PcDefinitionProvider(val compiler: MetalsGlobal, params: OffsetParams) {
         case _ => tree
       }
     }
-    val typedTree = typedTreeAt(pos)
+    val typedTree = locateTree(pos)
     val tree0 = typedTree match {
       case sel @ Select(qual, _) if sel.tpe == ErrorType => qual
       case Import(expr, _) => expr
@@ -133,4 +134,5 @@ class PcDefinitionProvider(val compiler: MetalsGlobal, params: OffsetParams) {
         loop(tree)
     }
   }
+
 }
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
@@ -610,6 +610,20 @@ trait Completions { this: MetalsGlobal =>
         case _ => EmptyTree
       }
     }
+    private def addToPath(path: List[Tree], t: Tree): List[Tree] = {
+      def goesAfter(prev: Tree): Boolean =
+        // in most cases prev tree should include next one
+        // however it doesn't work for some synthetic trees
+        // in this case check if both tree have different position ranges
+        prev.pos.includes(t.pos) || !t.pos.includes(prev.pos)
+
+      path match {
+        case Nil => List(t)
+        case head :: tl if goesAfter(head) => t :: head :: tl
+        case head :: tl => head :: addToPath(tl, t)
+      }
+    }
+
     protected def isEligible(t: Tree): Boolean = !t.pos.isTransparent
     override def traverse(t: Tree): Unit = {
       t match {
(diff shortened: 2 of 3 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/PcDefinitionProvider.scala
```diff
@@ -131,4 +131,58 @@ class PcDefinitionProvider(val compiler: MetalsGlobal, params: OffsetParams) {
...
+    }
+
+    override def traverse(t: Tree) {
+      t match {
+        case tt: TypeTree
+            if tt.original != null && (tt.pos includes tt.original.pos) =>
+          traverse(tt.original)
+        case _ =>
+          if (t.pos includes pos) {
+            if (isEligible(t) && isMoreNarrow(t.pos))
```
dos65, line 169: The difference with original `typedTreeAt` and `Locator` is in this line. I added `isMoreNarrow` check.

##### Explanation

Having the following source
```scala
class Main {
  val a = 42
  val all = 42 :: List(1, 2)
}
```
The left side tree for `val all` is:
```
Block(
  List(
    ValDef(.. x$1, ... , Select(This(Main), a)) // 1 <- synthetic 
    Apply(...) // 2 call `::` in reversed order using synthetic - `List(1,2) :: x$1`
  )
)
```

The problem that is in scala2.12  `Locator` first receives 1(synthetic) and then 2(call `::`).
Both of them includes definition pos so 2(call `::`) tree overwrites the previous one that leads to having wrongly located tree.

mtags/src/main/scala-2/scala/meta/internal/pc/PcDefinitionProvider.scala
```diff
@@ -131,4 +131,58 @@ class PcDefinitionProvider(val compiler: MetalsGlobal, params: OffsetParams) {
         loop(tree)
     }
   }
+
+  /** A fully attributed tree located at position `pos` */
+  private def typedTreeAt(pos: Position, unit: RichCompilationUnit): Tree = {
+    typeCheck(unit)
+    new Locator(pos).locateIn(unit.body)
```
tgodzik, line 138:
```suggestion
    new MetalsLocator(pos).locateIn(unit.body)
```

I think the new condition is already included there, so we should be good to use it or maybe adjust it if I am mistaken? You could drop the new Locator below.

tgodzik (changes requested):
Thanks for the fix! That's a tricky one, but maybe we could reuse some of the existing code?

mtags/src/main/scala-2/scala/meta/internal/pc/PcDefinitionProvider.scala
ckipp01: Yea that was my thought as well, I think we should be able to use some of the existing logic in `MetalsLocator` instead of creating a new one down below.

dos65: Thanks!
I didn't know that we already have our own patched version of `Locator`

dos65: Hm. I've tried to update existing `MetalsLocator` but this breaks [`CompletionInterpolationSuite.token-error`](https://github.com/scalameta/metals/blob/c2380821a14ee76fc982cfece217ae22a901288c/tests/cross/src/test/scala/tests/pc/CompletionInterpolatorSuite.scala#L467) because of having hack with `CURSOR`.

@tgodzik @ckipp01 Don't you mind to leave it as it is? It isn't clear how to properly fix it

tgodzik: Which hack is that? Maybe we could put that behind a boolean and not use here?

dos65: Upd. The problem wasn't related to `CURSOR`. The assumption that all next trees in traverser must be more narrow wasn't correct. At least might not work for `StringContext` trees as in mentioned tests-case.
I moved this logic into exiting `MetalsLocator` but changed it a little bit.

tgodzik (approved):
LGTM, just with one small question

mtags/src/main/scala-2/scala/meta/internal/pc/PcDefinitionProvider.scala
```diff
@@ -20,6 +20,7 @@ class PcDefinitionProvider(val compiler: MetalsGlobal, params: OffsetParams) {
         params.uri().toString(),
         None
       )
+      typeCheck(unit)
```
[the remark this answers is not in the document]
dos65: Yes, it's needed.
Previously, `typeCheck` was called inside `typeTreeAt` at this line: https://github.com/scalameta/metals/blob/751670ecfce1f2d01dd9f301cbaeca74fd9b551c/mtags/src/main/scala-2/scala/meta/internal/pc/PcDefinitionProvider.scala#L98
Now it replaced by `locateTree` which doesn't preform typecheck before running traverser.

PR #2920, opened 2021-07-01
Search for symbol references in workspace

Z1kkurat:
Resolves https://github.com/scalameta/metals-feature-requests/issues/211
Tried locally in VSCode and it works, though it shows local file results (in the same library file) before workspace ones despite me returning workspace references in a list specifically before local ones (and in tests they are returned in the correct order). Not sure if that's OK and how to fix it otherwise, may need some advice.


tgodzik (approved):
LGTM!

tgodzik:
That's true, but we are working on that currently in https://github.com/scalameta/metals/pull/2520

PR #2921, opened 2021-07-01
Don't offer to organize imports when there is an error in the file

ckipp01:
I need to figure out a good way to test this as the current framework for code actions don't really have a good way to test stuff while the file isn't compiling, _but_ I wanted to throw this up here right away to get any feedback. The point of this pr is to try to avoid a user seeing an error message all the time about state semanticdb when they refactoring a file. The reason this happens is because they at one point in time had a compiling file which produced semanticdb, and then when they start to refactor and come across unused imports, it tries to calculate the scalafix diff and errors out on stale semanticdb. This makes two changes:

1. The first change has to do when seeing if organize imports can be contributed or not. When a user has `-Ywarn-unused:imports` on and then go over a context that has an unused diagnostic it will now add an extra check to ensure that the file doesn't contain an error diagnostic. If it does, it simply doesn't offer it up as a code action and moves on. Again, this is only done _if_ it's in a valid context with an unused import there.
2. The second has to do when a user explicitly triggers the organize imports action from their editor. In this scenario it also does the same check and if the file has an error, it warns them via a message and doesn't try to do anything else.

Playing around with this I first actually put this check inside of the `ScalafixProvider`, but seeing that it's always called through `OrganizeImports` atm _and_ I wanted different behavior depending on whether it was explicitly called or it was just trying to generate the diff to contribute to the code action, I thought it made more sense to instead put it where I did. If you don't think that's a good idea, lemme know.

In some ways this is a temporaryish fix because ultimately I'd like to just be able to pass in the `TextDocument` we generate as a fallback or something to Scalafix, but seeing that the api doesn't allow for that atm, we'll go this route.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
@@ -35,7 +35,6 @@ import org.eclipse.{lsp4j => l}
  * goto definition in stale buffers.
  */
 final class Diagnostics(
-    buildTargets: BuildTargets, // TODO this can be removed
     buffers: Buffers,
     languageClient: LanguageClient,
     statistics: StatisticsConfig,
--- a/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
@@ -35,7 +35,7 @@ import org.eclipse.{lsp4j => l}
  * goto definition in stale buffers.
  */
 final class Diagnostics(
-    buildTargets: BuildTargets,
+    buildTargets: BuildTargets, // TODO this can be removed
     buffers: Buffers,
     languageClient: LanguageClient,
     statistics: StatisticsConfig,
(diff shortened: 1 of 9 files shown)


metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
```diff
@@ -169,6 +169,19 @@ final class Diagnostics(
   def hasSyntaxError(path: AbsolutePath): Boolean =
     syntaxError.contains(path)
 
+  def hasDiagnosticError(path: AbsolutePath): Boolean = {
```
tgodzik, line 172: Should we not check the build target errors? It will not work even if the file is ok, but some others failed.

ckipp01: I went back and forth on this. If we want to be 100% sure they can actually organize the imports, then yea, we should check build target errors. However, there may be situations where they want to clean up a file and organize the imports in there and the semanticdb for that file is still valid since nothing changed in it. I could be convinced either way. What do you think?

tgodzik: We could catch the stale exception and check afterwards if the build target contains an error. This would just mean changing the message in case of the stale semanticdb error.

ckipp01: Ah yea that's a good idea. I'll add that extra check.

mlachkar: I didn't know that we produce a semanticdb file if it's not compiling.

mlachkar:
Looks good to me!

ckipp01:
Alright so this changes the flow a bit. The behavior is now as follows:

- Everything compiling -> same as before
- File can't compile -> No code action given. If you explicitly try you get a warning that you need to fix your file
- Build target isn't compiling -> If the semanticdbt isn't stale, then all is good. However, if it comes back stale we treat it differently than a different failure and warn the user that it's due to something not compiling.

tgodzik (approved):
LGTM

PR #2924, opened 2021-07-02
Do not use $TMPDIR to store sbt-launcher.jar

kpodsiad:
Fixes https://github.com/scalameta/metals/issues/2776

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
@@ -1,6 +1,7 @@
 package scala.meta.internal.builds
 import java.nio.charset.StandardCharsets
 import java.nio.file.Files
+import java.nio.file.Path
 import java.util.Properties
 
 import scala.meta.inputs.Input
@@ -28,8 +29,10 @@ case class SbtBuildTool(
    * can't rely on `sbt` resolving correctly when using system processes, at
    * least it failed on Windows when I tried it.
    */
-  lazy val embeddedSbtLauncher: AbsolutePath = {
-    val out = BuildTool.copyFromResource(tempDir, "sbt-launch.jar")
+  def embeddedSbtLauncher(
+      outDir: Path
+  ): AbsolutePath = {
+    val out = BuildTool.copyFromResource(outDir, "sbt-launch.jar")
     AbsolutePath(out)
   }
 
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
```diff
@@ -28,8 +28,11 @@ case class SbtBuildTool(
    * can't rely on `sbt` resolving correctly when using system processes, at
    * least it failed on Windows when I tried it.
    */
-  lazy val embeddedSbtLauncher: AbsolutePath = {
-    val out = BuildTool.copyFromResource(tempDir, "sbt-launch.jar")
+  def embeddedSbtLauncher(workspace: AbsolutePath): AbsolutePath = {
+    val out = BuildTool.copyFromResource(
+      workspace.resolve(".bsp").toNIO,
```
tgodzik, line 33: Should this only happen if we run `generateBspConfig` ? Or does it get always created when sbt is run?

I was thinking that we can still use tmp in case of `bloopInstall`, but if it also generates the .bsp directory then I guess it would be good to keep it there at all times.

ckipp01: Yea I agree, we should probably only do this in the scenario that we are the ones creating the `.bsp/sbt.json`. Mainly because there might be situations where a user has the setting that creates the `.bsp/sbt.json` file turned off since they don't want it. Then it will end up being confusing if we create that dir to store our `sbt-launch.jar` in. Storing in in /tmp for the bloopInstall situation won't matter because it's fine if that's cleared, we can just create another whereas if it's hardcoded in the `.bsp/sbt.json` to the /tmp one, that's when we have issues.

ckipp01 (changes requested):

ckipp01:
O, and of course, thanks for the pr @kpodsiad!

metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
kpodsiad: Thanks for the suggestions, `bloopInstall` works as before and now changes affect only `generateBspConfig`.

ckipp01 (approved):
This looks good to me. You'll just need to run `sbt scalafixAll` and then you should be good to go. 👍🏼

tgodzik (approved):

PR #2927, opened 2021-07-03
Update to Docusaurus 2.0

ckipp01:
So there are a bunch of changes here and I still need to update a last few things.

- [x] Clean up dead links (*NOTE*: That when clicking around on the example site I had to get rid of the `/metals/` root in order for it to deploy right on Vercel, so if you randomly click on something that is hardcoded to `metals/docs` for example it will 404 just because on my published version it's just `docs/`. I've been able to locally though confirm that all the links are valid. It warns you now when you build if there are any.)
- [x] Make sure full blog is there on build

Just to give a brief overview, taken from the Docusaurus website:

> Docusaurus v2 is beta but already quite stable and widely used.

> We highly encourage you to use Docusaurus v2 over Docusaurus v1, as Docusaurus v1 will be deprecated soon.

> A lot of users are already using Docusaurus v2 (trends).

> Use Docusaurus v2 if:

> ✅ You want a modern Jamstack documentation site
> ✅ You want a single-page application (SPA) with client-side routing
> ✅ You want the full power of React and MDX
> ✅ You do not need support for IE11
> Use Docusaurus v1 if:

> ❌ You don't want a single-page application (SPA)
> ❌ You need support for IE11

I also did this migration partially with their [migration-helper](https://docusaurus.io/docs/migration/automated), but there ended up being a pretty large chunk that needed to be done manually.

I published a temp version of the site at [https://metals-test-ckipp01.vercel.app/](https://metals-test-ckipp01.vercel.app/) so feel free to poke around and let me know what you think.


docs/editors/overview.md
```diff
@@ -13,7 +14,6 @@ specific sections.
 
 ```
 
-
```
tgodzik, line 16: This break the title below, we need to keep this newline.

docs/editors/vscode.md
```diff
@@ -1,13 +1,11 @@
 ---
 id: vscode
+sidebar_label: VS Code
 title: Visual Studio Code
 ---
 
 ![Goto Definition](https://user-images.githubusercontent.com/1408093/48776422-1f764f00-ecd0-11e8-96d1-170f2354d50e.gif)
 
-```scala mdoc:requirements
```
tgodzik, line 8: Why is this removed? We should keep it, it's the https://scalameta.org/metals/docs/editors/vscode.html#requirements section.

tgodzik (changes requested):
Thanks for working on this! Looks pretty neat. I have some doubts about:

image

O we able to modify it to some other picture or just text? It's kind of weird to have these emoticons there.

ckipp01 (commented):
> Thanks for working on this! Looks pretty neat. I have some doubts about:
> O we able to modify it to some other picture or just text? It's kind of weird to have these emoticons there.

So by default that's what they use. I think it's pretty standard as well since when you look around for examples of light/dark mode switches you often see them with a sun and moon icon to notify the user which they are on. Is there something specific about them that you don't like or that worries you?

EDIT:

It looks like you can switch the icons, but I'm unsure what we'd switch them to. https://docusaurus.io/docs/api/themes/configuration#color-mode---dark-mode

tgodzik:
> So by default that's what they use. I think it's pretty standard as well since when you look around for examples of light/dark mode switches you often see them with a sun and moon icon to notify the user which they are on. Is there something specific about them that you don't like or that worries you?

That's I guess personal preference, it kind of looks like an old communicator icon, so I might be biased because of that.  It's just weird to see emoticons, where they are not needed. Even plain sun/moon would be much better.

image

This is the one.

ckipp01:
> Even plain sun/moon would be much better.

Let me see if I can find two plain icons.

ckipp01:
What do you think of this?

tgodzik:
Looks good!

tgodzik (approved):
LGTM from my side! Tested it locall and everything seems to work.

ckipp01:
> LGTM from my side! Tested it locall and everything seems to work.

Cool, I'll wait for another day to see if the others have any input, if now I'll merge away!

ckipp01:
Publish failed. Looks like a command isn't found, digging into it now.

PR #2933, opened 2021-07-05
Replace the usage of nuprocess

dos65:


metals/src/main/scala/scala/meta/internal/ansi/LineListener.scala
```diff
@@ -46,6 +46,8 @@ class LineListener(onLine: String => Unit) {
     )
   }
 
+  def appendLine(text: String): this.type = appendString(text + "\n")
```
ckipp01, line 49: If you look at the scaladoc of this class you'll see

>  * Converts a stream of strings (with potential ANSI codes and newlines) into a callback for indvidual lines where each individual line has no newline \n characters.

It's sort of confuses me now a bit since we say that there is no newline characters, but then we have a method that adds a newline character.

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
```diff
@@ -1,30 +1,80 @@
...
+  ): SystemProcess = {
+    def readOutput(stream: InputStream, f: String => Unit): Thread = {
+      val lineListeneter = new LineListener(f)
+      val thread = new Thread {
+        override def run(): Unit = {
+          // use scala.sys.process implementation
+          try BasicIO.processFully(line => lineListeneter.appendLine(line))(
+            stream
+          )
+          catch { case NonFatal(_) => }
```
ckipp01, line 36: If I'm understanding this correctly this would probably only ever be an `IOException` right? If that happens do we want to log this out or anything? If this ever fails it might be hard to track down if we do nothing here.

ckipp01 (commented):
Hopefully this will improve things! Just a couple questions. I'll actually publish this locally and use it today to see if I hit on anything weird.

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
dos65: Thx for pointing at this! I checked it again. 
I added this `try``catch` by mistake. `IOException` handling is done inside `BasicIO`.

The actual problem was that I caught an exception while I was debugging test on windows but now I see that it came from `ps.start()`.Reworked this thing.

metals/src/main/scala/scala/meta/internal/ansi/LineListener.scala
dos65: Honestly,  I wanted to avoid touching it :smile: But, yea, you right - it's confusing.

I reduced `LineListener` to `AnsiFilter` and dropped existing logic around handling newlines. It was used only for process handling and now it's readers split output by LF automatically.

`AnsiFilter` might look a little bit weird because of having an internal variable but I took into account [these notes](https://github.com/scalameta/metals/blob/16806aa5674c0656b7c4e634dcf882dbddffe00a/metals/src/main/scala/scala/meta/internal/ansi/LineListener.scala#L50-L59) about it's performance and tried to leave it comparable to the previous implementation.

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
dos65: All right, it might throw exception. Slightly changed this part.
The funny thing that this comment directed me to cover an additional case)

metals/src/main/scala/scala/meta/internal/ansi/LineListener.scala
ckipp01: > AnsiFilter might look a little bit weird because of having an internal variable but I took into account these notes about it's performance and tried to leave it close to the previous implementation.

👍🏼 makes sense.

ckipp01 (approved):
From what I can tell, this looks good. 👍🏼

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
```diff
@@ -1,30 +1,109 @@
...
+      cwd: AbsolutePath,
+      redirectErrorOutput: Boolean,
+      env: Map[String, String]
+  ): SystemProcess = {
+
+    try {
+      val builder = new ProcessBuilder(cmd.asJava)
+      builder.directory(cwd.toNIO.toFile)
+      val envMap = builder.environment()
+      envMap.putAll(env.asJava)
```
tgodzik, line 34: We should also add Java Home

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
```diff
@@ -1,30 +1,109 @@
...
+              stream
+            )
+          } catch {
+            case _: IOException => // that's ok, happens on cancel
+            case NonFatal(e) =>
+              scribe.error("Unexcepted error in reading out", e)
+          }
+        }
+      }
+      thread.setDaemon(true)
```
tgodzik, line 67: Do we need it to be a deamon?

metals/src/main/scala/scala/meta/internal/ansi/AnsiStateMachine.scala
```diff
@@ -4,24 +4,15 @@ package scala.meta.internal.ansi
...
 case class AnsiStateMachine(apply: Int => AnsiStateMachine)
 object AnsiStateMachine {
   // Helpful docs on ANSI escape sequences: http://ascii-table.com/ansi-escape-sequences.php
   val Start: AnsiStateMachine = AnsiStateMachine {
     case 27 => AnsiEscape
-    case '\r' => CarriageReturn
     case _ => Print
   }
 
-  val CarriageReturn: AnsiStateMachine = AnsiStateMachine {
```
tgodzik, line 20: Why dropping `\r` is no longer needed?

tgodzik (changes requested):
Looks good! I just left a couple of smaller comments and questions.

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
dos65: thx! Good catch. Added it in `ShellRunner`

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
dos65: I think yes. Just to avoid a possibility of having dangling thread that keeps JVM waiting its completion on shutdown.

metals/src/main/scala/scala/meta/internal/ansi/AnsiStateMachine.scala
dos65: Process's output is processed using `BufferedReader.readLine` internally so there is no more need to split it by lines by `LineListener`.

tgodzik: That's what I was suspecting. Thanks for the explanation!

tgodzik (approved):
LGTM! I checked the PR out and everything seems to work alright.

I also looked at nuprocess use cases and I think running a single process doesn't really fit it. The real improvements are seen when running a lot of processes, so I think we are ok to merge.

Issue #2923, opened 2021-07-01
VScode - Organize Import as CodeAction doesn't work


dos65:
Triggering code action on the line with `Unused Import` warning returns `No code action available.`
However, there are trace with `textDocument/codeAction` repsonse:

<details>
  <summary>Trace</summary>
  <pre>
  [Trace - 11:31:17 PM] Received request 'textDocument/codeAction - (24)'
Params: {
  "textDocument": {
    "uri": "file:///home/dos65/projects/metals/metals/tests/unit/src/test/scala/tests/BspSwitchLspSuite.scala"
  },
  "range": {
    "start": {
      "line": 6,
      "character": 18
    },
    "end": {
      "line": 6,
      "character": 18
    }
  },
  "context": {
    "diagnostics": [
      {
        "range": {
          "start": {
            "line": 6,
            "character": 0
          },
          "end": {
            "line": 6,
            "character": 34
          }
        },
        "severity": 2,
        "source": "bloop",
        "message": "Unused import"
      }
    ]
  }
}

[Trace - 11:31:17 PM] Sending response 'textDocument/codeAction - (24)'. Processing request took 75ms
Result: [
  {
    "title": "Organize imports",
    "kind": "source.organizeImports",
    "edit": {
      "changes": {
        "file:///home/dos65/projects/metals/metals/tests/unit/src/test/scala/tests/BspSwitchLspSuite.scala": [
          {
            "range": {
              "start": {
                "line": 0,
                "character": 0
              },
              "end": {
                "line": 56,
                "character": 0
              }
            },
[...]

PR #2935, opened 2021-07-06
Organize imports quickfix action

kpodsiad:
Previously, our organize imports action supported only source.organizeImports kind (implemented in https://github.com/scalameta/metals/pull/1971) because code actions were filtered based on kind parameter (CodeActionProvider:50)
In order to show quickfix action vscode expects that kind of response is quickifx (hence in https://github.com/scalameta/metals/issues/2923 there was no action available although there was one in the response from the server but it was filtered out)

I extracted common logic to the abstract class and create a separate class per kind - IMO this is a good balance between DRY, single responsibility and code readability.   
Another option is to change `kind` in `CodeAction` to Set[String] and support multiple actions simultaneously but I'm not sure about this approach.  

I'll fix failing tests and implement new ones after receiving some feedback.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
@@ -18,7 +18,7 @@ import scala.meta.pc.CancelToken
 import org.eclipse.lsp4j.CodeActionParams
 import org.eclipse.{lsp4j => l}
 
-final class OrganizeImports(
+sealed abstract class OrganizeImports(
     scalafixProvider: ScalafixProvider,
     buildTargets: BuildTargets,
     diagnostics: Diagnostics,
@@ -26,19 +26,18 @@ final class OrganizeImports(
 )(implicit ec: ExecutionContext)
     extends CodeAction {
 
-  override def kind: String = OrganizeImports.kind
-
+  protected def title: String
+  protected def isCallAllowed(
+      file: AbsolutePath,
+      params: CodeActionParams
+  ): Boolean
   override def contribute(params: CodeActionParams, token: CancelToken)(implicit
       ec: ExecutionContext
   ): Future[Seq[l.CodeAction]] = {
     val uri = params.getTextDocument.getUri
     val file = uri.toAbsolutePath
 
-    if (
-      explicitlyCalledAndAllowed(file, params)
-      || hasUnusedImportAndAllowed(file, params)
-    ) {
-
+    if (isCallAllowed(file, params)) {
       val scalaTarget = for {
         buildId <- buildTargets.inverseSources(file)
         target <- buildTargets.scalaTarget(buildId)
@@ -49,7 +48,6 @@ final class OrganizeImports(
         case _ => Future.successful(Seq())
       }
     } else Future.successful(Seq())
-
   }
 
   private def organizeImportsEdits(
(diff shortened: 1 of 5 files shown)


gabro (approved):
I personally like the approach and I think changing CodeAction to support multiple kinds would have been worse, since this case seems fairly unique.

tgodzik:
I think it's a good approach, :+1: from me.

ckipp01:
> Another option is to change kind in CodeAction to Set[String] and support multiple actions simultaneously but I'm not sure about this approach.

It's probably pretty rare that the same action would have multiple types. I'd maybe not do this unless you can think of another example. Either way,  no strong preference.

I like this being able to be offered as a QF action as well, so 👍🏼 from me.

dos65:
IMO, it would be easier to keep the single instance of `OrganizeImports` and set its kind based on checking if the `CodeActionParams` has `"only": quickfix`.

Also, I'm wondering for which case we need this action with `source.organizeImports` kind ?

ckipp01:
> Also, I'm wondering for which case we need this action with `source.organizeImports` kind ?

Most editors have an action built-in that a user can trigger a `Organize Imports` command that automatically calls the code action with the `source.organizeImports`. We should be sure to keep this in or else that functionality would break.

dos65:
> Most editors have an action built-in that a user can trigger a Organize Imports command that automatically calls the code action with the source.organizeImports. We should be sure to keep this in or else that functionality would break.

Oh. Now I got what's going on. Then :+1: from me too

kpodsiad:
It took me a while but I finally managed to write a few new tests. I'm not sure if `quickfix-available-on-hover` checks exactly what I wanted so please take a look at it

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
```diff
@@ -170,9 +187,61 @@ class OrganizeImportsLspSuite
...
+  check(
+    "quickfix-available-on-hover",
+    """
+      |import scala.@@util.Try
+      |
+      |object A {
+      |  <<val action = "quick should be available">>
+      |}
+      |""".stripMargin,
+    s"${OrganizeImportsQuickFix.title}",
```
kpodsiad, line 226: Turned out that the test is failing 😅. Is there a way to mimic hover behavior in tests?

tgodzik: Quick fixes are unrelated to hovers, so no need to test them. I think the current test is `"on-unused"` is enough.

kpodsiad: Are you sure? In vscode when I hover unused there is quick fix action possible

metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
```diff
@@ -111,21 +158,9 @@ final class OrganizeImports(
...
 }
-object OrganizeImports {
-  def title: String = "Organize imports"
-  def kind: String = l.CodeActionKind.SourceOrganizeImports
 
-  type ScalaBinaryVersion = String
-  type ScalaVersion = String
+object OrganizeImportsQuickFix {
+  final val kind: String = l.CodeActionKind.QuickFix
+  final val title: String = "Fix organize imports"
```
tgodzik, line 165:
```suggestion
  final val title: String = "Fix imports"
```

tgodzik (changes requested):
Thanks for the contribution! Looks good, but I would change the name slightly, what do you think?

kpodsiad:
> Thanks for the contribution! Looks good, but I would change the name slightly, what do you think?

Yeah, I agree with that.  
BTW, I committed this suggestion from a browser, what a cool feature from Github!

tgodzik (approved):
LGTM!

PR #2937, opened 2021-07-06
Fix dead links and add in serve command

ckipp01:


ckipp01:
I'm sort of at a loss at what's best here... and it's tricky to track it down. The `docusaurus serve`, `vercel` and `github pages` all seems to have different behavior. I can't get the `toExtensions: ["html"]` to work locally or on vercel to test to see if that would actually work to force ".html" all over the place. Not having the redirect plugin at all works perfectly when using the serve functionality since both the `.html#query` works and the `#query` works... I'm sort of just shooting in the dark here and trying to align everything to not use `.html` and then hoping the redirects actually work as they are supposed to.

dos65 (approved):
LGTM

I noticed two things about website but they aren't related to this PR:

- There is `stylesheets` setting https://github.com/scalameta/metals/blob/6fc4020f320c65ae40b93032b8342bc1d08225a7/website/docusaurus.config.js#L9 but there is no `custom.css` (Found 404 error in browser console)

- I don't if it was working previously but navigation for opening a link with `#$query` doesn't work as it should. It works only after clicking on it after page was loaded. Do you know if possible to fix?

ckipp01:
> LGTM
> 
> I noticed two things about website but they aren't related to this PR:
> 
> * There is `stylesheets` setting https://github.com/scalameta/metals/blob/6fc4020f320c65ae40b93032b8342bc1d08225a7/website/docusaurus.config.js#L9

>    but there is no `custom.css` (Found 404 error in browser console)

Ah good catch, I'll fix this and just include it in this pr

> * I don't if it was working previously but navigation for opening a link with `#$query` doesn't work as it should. It works only after clicking on it after page was loaded. Do you know if possible to fix?

I'm trying to figure out the best redirects for this to work, since they are now all being down client side. I'm having some issues though. For example

https://scalameta.org/metals/docs/editors/vim#getting-started-with-worksheets <- this will open up just fine
https://scalameta.org/metals/docs/editors/vim.html#getting-started-with-worksheets <- this will open, but redirect you to https://scalameta.org/metals/docs/editors/vim

Even worse, atm if you use the search bar to search for vim worksheets, it should lead you to https://scalameta.org/metals/docs/editors/vim.html#getting-started-with-worksheets, but it will 404.

There are a few different settings to control this, but it seems to behave differently in different contexts. For example they have a demo app up here https://codesandbox.io/s/docusaurus?file=/docusaurus.config.js:907-909 with no redirect listed, and magically both `.html#query` and `#query` both work. No idea how. I asked on their discord how that works because when it try leaving out the rewrite rule to drop `.html`, I get a 404 when I hit a `.html` page instead of being redirected.

PR #2938, opened 2021-07-07
[Scala 3] Properly show enum on hover

tgodzik:
This was actually done badly* previously and not actual tests were present, which is why this issue sneaked trough

*Done badly by me :sweat_smile:

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
@@ -40,12 +40,12 @@ class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx) {
           name = super.nameString(sym.name)
         } yield name
         "package " + owners.toList.reverse.mkString(".")
-      // enum
-      case _ if sym.companionClass.is(Flags.Enum) =>
-        s"enum $name: "
       // enum case
       case _ if sym.is(Flags.EnumVal) =>
         s"case $name: "
+      // enum
+      case _ if sym.is(Flags.Enum) || sym.companionClass.is(Flags.Enum) =>
+        s"enum $name: "
       // default
       case key =>
         // no need to add final on object, since they are all final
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
```diff
@@ -40,12 +40,12 @@ class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx) {
...
         } yield name
         "package " + owners.toList.reverse.mkString(".")
-      // enum
-      case _ if sym.companionClass.is(Flags.Enum) =>
-        s"enum $name: "
       // enum case
       case _ if sym.is(Flags.EnumVal) =>
         s"case $name: "
+      // enum
+      case _ if sym.is(Flags.Enum) =>
```
tgodzik, line 47: `EnumVal` is also an enum, which is why I switched the order around.

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
```diff
@@ -73,7 +73,12 @@ class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx) {
     val typeSymbol = info.typeSymbol
     sym match {
       case p if p.is(Flags.Package) => fullDef
-      case p if typeSymbol.is(Flags.Module) =>
+      /* Type cannot be shown on the right since it is already a type
+       * let's instead use that space to show the full path.
+       */
+      case p
+          if typeSymbol.isOneOf(Flags.Module) || // object
```
dos65, line 80:
```suggestion
          if typeSymbol.is(Flags.Module) || // object
```

dos65 (approved):
Cool!
I've seen this issue but thought that the reason should be more close to compiler :smile:

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
tgodzik: Fixed!

PR #2947, opened 2021-07-09
Add Scala 3.0.1 support

dos65:
Also, dropped support for all previous 3.0.x-RC versions.


build.sbt
```diff
@@ -243,9 +242,11 @@ lazy val V = new {
   def scala2Versions = nonDeprecatedScala2Versions ++ deprecatedScala2Versions
 
   // Scala 3
-  def scala3RCVersions = Seq("3.0.0-RC3", "3.0.0-RC2", "3.0.0-RC1")
```
tgodzik, line 246: @smarter is it ok to drop them? I can't find the date that you mentioned previously.

build.sbt
```diff
@@ -243,9 +242,11 @@ lazy val V = new {
...
   def nonDeprecatedScala3Versions =
-    scala3 +: scala3RCVersions :+ nextScala3RC
+    Seq(
```
tgodzik, line 246: Can it be oneline? Just change it if you do any further fixes, otherwise I would leave it.

tgodzik (approved):
Thanks for remembering about it! As soon as it is green and @smarter has approved it we should be good to merge.

tgodzik:
`tests.ScalaToplevelLibrarySuite.scala3-compiler_3-3.0.1-sources.jar ` failure seems legit :/

build.sbt
smarter: Could you wait until next week to drop RC1? We had to move the final exam for our class to tomorrow (Saturday) and are still using 3.0.0-RC1 there.

smarter: In fact if you can wait until next week for the release in general this ensures that we won't run into any new exciting issue tomorrow :sweat_smile:

dos65: @smarter Sure. It's a preparation step for release. We aren't going to do release until Monday.

dos65:
@tgodzik yep. I'm looking at it - it's a parser issue

build.sbt
Swoorup: can we release on nightly for bleeding edge users 😄

tgodzik (approved):
LGTM!

build.sbt
dos65: @swoorup Unfortunately, that's won't work as we can't publish artifacts per every nightly version.

tgodzik: it should be possible to setup an infrastructure for it, but that would require building mtags on the dotty side

PR #2959, opened 2021-07-14
Use scala-debug-adapter 2.x to enable code evaluation

adpi2:
# Introducing `MetalsDebugAdapter2x`

scala-debug-adapter 2.x now does the mapping from source files to class files by itself, because it needs it for expression evaluation.

As a consequence Metals `ProxyDebugServer` job is a simplified:
- Metals can send breakpoint request on a source file directly (without converting it to a class name)
- Metals receives stack frames with source path instead of class name
- ...

However the debug server does not know about the `.metals/readonly/dependencies` folder and it expects the dependencies source paths to be inside source jars. We introduce the `MetalsDebugAdapter2x` to do this conversion back and forth:
`jar:file://some-path/cats_3-2.0.0.jar!/cats/core/Map.scala` <->
`file://some-path/example/.metals/readonly/dependencies/cats_3-2.0.0.jar/cats/core/Map.scala`

The old behavior is still available in `MetalsDebugAdapter1x` for those who used  a version of Bloop that does not yet support expression evaluation.

The code that does the switch is:
```scala
if (buildServer.hasExpressionEvaluation) {
  MetalsDebugAdapter.`2.x`(
    buildTargets,
    targets
  )
} else {
  MetalsDebugAdapter.`1.x`(
    definitionProvider,
    buildTargets,
    classFinder,
    scalaVersionSelector,
    targets
  )
}
```

Where `buildServer.hasExpressionEvaluation` is:
```scala
  def hasExpressionEvaluation: Boolean =
    isSbt || (isBloop && SemVer.isCompatibleVersion("1.4.10", version))
```

Once we have all moved to Bloop 1.4.10, `MetalsDebugAdapter1x`, `ClassFinder` and all related code can be removed. 

# Expression evaluation with sbt as the build server

Metals injects `sbt-debug-adapter %% "2.0.2"` in `project/metals.sbt` so that expression evaluation is available.
This is tested in `SbtEvaluationDapSuite` by evaluating the basic "1 + 1" expression. I don't think this needs more tests because expression evaluation is tested more extensively in the scala-debug-adapter [`ExpressionEvaluatorSpec`](https://github.com/scalacenter/scala-debug-adapter/blob/main/core/src/test/scala/ch/epfl/scala/debugadapter/ExpressionEvaluatorSpec.scala)

# Expression evaluation with Bloop as the build server

Expression evaluation is available in Bloop since `1.4.9-14-67e8541e`. It will become stable in Bloop 1.4.10.
It is tested in `EvaluationDapSuite`.


tests/unit/src/test/scala/tests/debug/StackFrameDapSuite.scala
```diff
@@ -62,4 +62,29 @@ class StackFrameDapSuite
       )
     )
   )
+
+  assertStackFrame("overridden-toString")(
```
tgodzik, line 66: We can use Bloop's newest snapshot version and there will be no need to separate this out.

metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathAdapter.scala
```diff
@@ -0,0 +1,66 @@
...
+
+import ch.epfl.scala.bsp4j.BuildTargetIdentifier
+
+private[debug] final class SourcePathAdapter(
+    workspace: AbsolutePath,
+    sources: Set[AbsolutePath],
+    buildTargets: BuildTargets
+) {
+  private val dependencies = workspace.resolve(Directories.dependencies)
+  def toBspURI(sourcePath: AbsolutePath): Option[URI] = {
```
tgodzik, line 19:
```suggestion
  def toDapURI(sourcePath: AbsolutePath): Option[URI] = {
```
wouldn't make more sense?

tgodzik (changes requested):
Awesome work! I think we could add Bloop support already in this PR by bumping to the latest snapshot. It will be easier for people to test it out.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -305,11 +307,8 @@ final class BuildTargets(
         names match {
           case Directories.dependenciesName :: jarName :: _ =>
             // match build target by source jar name
-            inverseDependencySources
```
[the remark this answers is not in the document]
adpi2: In term of memory this is nothing because a build often have less than 100 jar files. Even if it contains 10.000 jar files the Map is still very small.

I needed to create the `sourceJarFile` method for the `SourcePathAdapter` and this is the simplest implementation I could think of.

metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathAdapter.scala
adpi2: IDK which is better. My idea was that the paths of the source jars come from the response of the `buildTarget/dependencySources` BSP request.

tests/unit/src/test/scala/tests/debug/StackFrameDapSuite.scala
adpi2: That's a good idea! I am trying that now.

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -69,6 +70,8 @@ class BuildServerConnection private (
   // hasDebug is not yet available in BSP capabilities
   // https://github.com/build-server-protocol/build-server-protocol/pull/161
   def hasDebug: Boolean = isBloop || isSbt
+  def hasExpressionEvaluation: Boolean =
+    isSbt || (isBloop && SemVer.isCompatibleVersion("1.4.10", version))
```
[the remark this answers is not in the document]
adpi2: :+1:

metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathAdapter.scala
tgodzik: `toDapURI` would make sense in this context I think since we are in debug package.

metals/src/main/scala/scala/meta/internal/metals/debug/MetalsDebugAdapter.scala
```diff
@@ -0,0 +1,105 @@
...
+ */
+private[debug] class MetalsDebugAdapter2x(sourcePathAdapter: SourcePathAdapter)
+    extends MetalsDebugAdapter {
+
+  override def adaptSetBreakpointsRequest(
+      sourcePath: AbsolutePath,
+      request: SetBreakpointsArguments
+  ): Iterable[SetBreakpointsArguments] = {
+    // try to find a BSP uri corresponding to the source path or don't send the request
+    sourcePathAdapter.toBspURI(sourcePath).map { sourceUri =>
```
tgodzik, line 93:
```suggestion
    sourcePathAdapter.toDapURI(sourcePath).map { sourceUri =>
```

The jobs are failing on this one. Did you use Metals and it failed to ranem :fearful:

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/metals/debug/MetalsDebugAdapter.scala
adpi2: oupsy

ckipp01:
First try with nvim-dap. This is amazing @adpi2

ckipp01:
btw from a pure usability standpoint, I sort of expected this to work:
Screenshot 2021-09-23 at 18 33 52

I wouldn't have expected that knowing `y` would have mattered here yet since I wasn't invoking anything to do with `y`, but rather just trying to evaluate `3 * 4`. Is this expected?

tgodzik:
Looks like ` tests.debug.StepDapSuite.step-into-java-lib` and ` tests.debug.StepDapSuite.step-into-scala-lib` are failing currently. Any idea what is going on? 

`java.lang.Exception: Obtained [null, 404], expected [D:\a\metals\metals\tests\unit\target\e2e\debug-step\step-into-scala-lib\.metals\readonly\dependencies\scala-library-2.12.15-sources.jar\scala\Predef.scala, 404]` <- looks like we are finding the line, but the uri gets wrongly translated to null.

> I wouldn't have expected that knowing y would have mattered here yet since I wasn't invoking anything to do with y, but rather just trying to evaluate 3 * 4. Is this expected?

@ckipp01  Does it happen if this is the only evaluation that you did? @tdudzik do you know what might be going on?

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -69,6 +70,11 @@ class BuildServerConnection private (
   // hasDebug is not yet available in BSP capabilities
   // https://github.com/build-server-protocol/build-server-protocol/pull/161
   def hasDebug: Boolean = isBloop || isSbt
```
ckipp01, line 72: I know you didn't touch this _but_ seeing we are talking about relating things. I'm assuming that both of these set that the are `runProvider` and `debugProvider` in the `buildServerCapabilities`. Couldn't we just grab that instead of explicitly checking for `isBloop` or `isSbt`?

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -69,6 +70,11 @@ class BuildServerConnection private (
...
+  def hasExpressionEvaluation: Boolean =
```
ckipp01, line 73: Do you think there is any chance of baking this into the BSP spec so that a server could just declare this during initialize?

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
adpi2: Oh yes indeed. However sbt does not set the `debugProvider` field. So this should be fixed in sbt beforehand, but the `buildServerCapabilities` in sbt may not be overriden easily...

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
adpi2: This method is temporary, to manage the transition between scala-debug-adapter 1.x and scala-debug-adapter 2.x. Once Bloop `1.4.10` is released we won't need it anymore.

I will give it a better name like`usesScalaDebugAdapter2x` and add a comment.

adpi2:
@ckipp01 

This is a bug in [scalacenter/scala-debug-adapter](https://github.com/scalacenter/scala-debug-adapter), so I invite you to report it there. We know some other cases are not working and some other bugs will be found. It should not impact the code in Metals because we will simply need to bump the `scala-debug-adapter` version after each bug fix.

tgodzik (changes requested):
Looks like the JDK 17 test are failing with:
`2021.09.24 09:44:33 ERROR java.lang.UnsupportedOperationException: Records requires ASM8`

I found a related issue:
https://github.com/johnrengelman/shadow/issues/679

This might be caused by an outdated dependency in scala-debug-adapter maybe?

We could turn off the tests and debugging for JDK 17 for now and work on it later? We should for sure add testing for JDK-17 in scala-debug-adapter.

adpi2:
JDK 17 is not supported in the scala-debug-adapter because we use an old version of asm. So yes I think we should turn off the tests and debugging for JDK 17. @tgodzik  Do you have any pointer on how to do that?

tgodzik:
> JDK 17 is not supported in the scala-debug-adapter because we use an old version of asm. So yes I think we should turn off the tests and debugging for JDK 17. @tgodzik Do you have any pointer on how to do that?

There is a method in `BaseSuite.scala` which has:
```
  def isJava8: Boolean =
    !Properties.isJavaAtLeast("9")
```

we could add:
```
  def isHigherThan11: Boolean =
    Properties.isJavaAtLeast("12")
```
and use that.

Btw. is there an issue in scala-debug-adapter about JDK 17 or can we create it? How much effort it would be to swithc to never version of asm?

adpi2:
@tgodzik I have added support for JDK 17 by bumping the version of scala-debug-adapter to 2.0.3

tgodzik (approved):
LGTM!

This is really exciting!

PR #2971, opened 2021-07-15
Mtags3 - exclude scala2 depedencies.

### LINKED ISSUE #2800: Imports from other build targets suggested
**Describe the bug**

In Metals, when working on Scala 3 it seems we are also suggested classes from the Scala 2 compiler.

**To Reproduce**

1. ScalaPresentationCompiler for Scala 3
2. Write tree...
3. You will be suggested treeChecker from `scala.tools.nsc.Global.treeChecker`
4. Code will not compiler

**Expected behavior**

`treeChecker` is not suggested

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.10.2-SNAPSHOT

**Additional context**

It seems that we are suggesting too many things and we should check if they are available in the build target

**Search terms**

Scala3 imports auto

dos65:
They come transitively from scalameta.
Fixes #2800

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -356,6 +356,8 @@ val mtagsSettings = List(
       "org.scala-lang" %% "scala3-compiler" % scalaVersion.value,
       ("org.scalameta" %% "scalameta" % V.scalameta)
         .cross(CrossVersion.for3Use2_13)
+        .exclude("org.scala-lang", "scala-reflect")
+        .exclude("org.scala-lang", "scala-compiler")
     )
   ),
   libraryDependencies ++= List("org.lz4" % "lz4-java" % "1.8.0"),


build.sbt
```diff
@@ -1,3 +1,4 @@
+import org.apache.ivy.core.module.descriptor.ExcludeRule
```
tgodzik, line 1: I think you can remove this one.

tgodzik (changes requested):
Thanks for figuring this out, I hadn't realised that we had the dependency. Just one small nitpick from my side.

build.sbt
dos65: :smile: Thx, fixed

tgodzik (approved):
LGTM!

Issue #2973, opened 2021-07-16
In one Ammonite script file cannot recognize another sc file's variable and methods when attempt use code completed.


LoranceChen:
**Describe the bug**

**To Reproduce** Steps to reproduce the behavior:

1. create one sc file: `MyScript.sc`:
```
// Ammonite 2.4.0, Scala 2.13.6 

println("Hello World!!")

val myCount= 12345

def myFunc = {
  "hello world!"
}
```
2. create second sc file: `MyScript2.sc`, the problem is there is not proper tips for `myCount` or `myFunc` when input `MyScript.`
```
// Ammonite 2.4.0, Scala 2.13.6 

import $file.MyScript

MyScript.
```

**Expected behavior**
give tips for `myCount` or `myFunc` when input `MyScript.`

**Screenshots**
![image](https://user-images.githubusercontent.com/8775828/125950282-39f7901c-2010-49d7-b692-e76a6dc3331f.png)

**Installation:**
- Operating system: windows10
- VSCode version: 1.58.2
- VSCode extension version: 
- Metals version: (found in VSCode settings, under `metals.serverVersion`) metals: 1.10.7

**Additional context**

**Search terms**
ammonite script metals scala

ckipp01:
Btw this is probably the same issue as in https://github.com/scalameta/metals/issues/2122. It's not limited to getting completions from another script, the issue actually exists in the same file if you try to complete from user defined types.

tgodzik:
> Btw this is probably the same issue as in #2122. It's not limited to getting completions from another script, the issue actually exists in the same file if you try to complete from user defined types.

Thanks Chris, this prompted me to recheck what's going on and I think I got it fixed. And it was all my fault! :sweat_smile:

ckipp01:
> Thanks Chris, this prompted me to recheck what's going on and I think I got it fixed. And it was all my fault! 😅

Ha, just as much our fault as well for not catching it. Glad you figured it out!

Issue #2122, opened 2020-10-04
No completions given for user defined types in an Ammonite script


ckipp01:
**Describe the bug**

It seems that Ammonite isn't providing any completion items for members of self defined types.

**To Reproduce**
Given the following:

```scala
case class Test(a: Int, b: String)

class Test2 {
  def a(a: Int): Int = a
}

object Test3 {
  val a = "a"
}

val test = Test(1, "one")

test.@@ // neither a or b completions offered here.

val test2 = new Test2

test2.@@ // no a completion given here

Test3.@@ // no a completion given here
```

**Expected behavior**

I would expect member completions for all of the given examples

**Installation:**

- Operating system: macOS
- Editor: Vim
- Metals version: `0.9.4+43-6a98caf6-SNAPSHOT`

**Search terms**

Ammonite completions

tgodzik:
Thanks for reporting! Maybe we are not giving the presentation compiler the right path to the compiled artifacts?

AcrossForest:
Same issue for VSCode on win10.

thinkiny:
same issume for emacs

PR #2974, opened 2021-07-19
Fix ammonite type completions

tgodzik:
Due to https://github.com/scalameta/metals/pull/1846 we were actually filtering out a lot of ammonite completions, these were all the scope completions. Now we should properly filer out the wrong compiler completions after $file, because we are using a much more specific condition and the same that we use for AmmoniteFileCompletions.

Fixes https://github.com/scalameta/metals/issues/2973 and fixes https://github.com/scalameta/metals/issues/2122


ckipp01 (approved):
Ahhh good catch! Just need to run scalafix, but other than that LGTM!

PR #2978, opened 2021-07-21
[Scala 3] Support extension methods in hover

tgodzik:
Previously, hover would show extension methods as a simple method. Now, we try to show the extension signature next to the actual extension method signature. This isn't yet perfect as it will not work if a method has a using param as the first one. This is not a big issue though as parameter in both places will work exactly the same.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
@@ -2,6 +2,8 @@ package scala.meta.internal.pc
 
 import scala.collection.mutable.ListBuffer
 
+import scala.meta.internal.mtags.MtagsEnrichments._
+
 import dotty.tools.dotc.core.Contexts.Context
 import dotty.tools.dotc.core.Flags
 import dotty.tools.dotc.core.NameKinds.EvidenceParamName
@@ -10,6 +12,7 @@ import dotty.tools.dotc.core.Names._
 import dotty.tools.dotc.core.Symbols._
 import dotty.tools.dotc.core.Types._
 import dotty.tools.dotc.printing.RefinedPrinter
+import org.eclipse.lsp4j.CompletionItemKind
 
 class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx) {
 
@@ -23,38 +26,6 @@ class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx) {
     toText(tpw).mkString(defaultWidth, false)
   }
 
-  def fullDefinition(sym: Symbol, tpe: Type): String = {
-    def name = nameString(sym)
-    keyString(sym) match {
-      case "" =>
-        val isImplicit = sym.is(Flags.Implicit)
-        val implicitKeyword = if (isImplicit) "implicit " else ""
-        s"$implicitKeyword$name: "
-      case key if sym.is(Flags.Method) =>
-        s"$key $name"
-      // package
-      case key if sym.is(Flags.Package) =>
-        val owners = for {
-          sym <- sym.ownersIterator
-          if !(sym.isRoot || sym.isEmptyPackage)
-          name = super.nameString(sym.name)
-        } yield name
-        "package " + owners.toList.reverse.mkString(".")
-      // enum case
-      case _ if sym.is(Flags.EnumVal) =>
-        s"case $name: "
-      // enum
-      case _ if sym.is(Flags.Enum) || sym.companionClass.is(Flags.Enum) =>
-        s"enum $name: "
-      // default
-      case key =>
-        // no need to add final on object, since they are all final
-        val finalKeyword =
-          if (sym.is(Flags.Final) && !sym.is(Flags.Module)) "final " else ""
-        s"$finalKeyword$key $name: "
-    }
-  }
-
   /**
    * for
    * - method: method signature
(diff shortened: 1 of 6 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
```diff
@@ -117,31 +144,100 @@ class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx) {
...
+  ): (List[List[Symbol]], List[List[Symbol]]) = {
+    def isUsingClause(params: List[Symbol]): Boolean = params match {
+      case sym :: _ => sym.is(Flags.Given)
+      case _ => false
+    }
+
+    def isTypeParamClause(params: List[Symbol]): Boolean = params match {
+      case sym :: _ => sym.is(Flags.TypeParam)
+      case _ => false
+    }
```
dos65, line 203:
```suggestion
    def headHasFlag(params: List[Symbol], flag: Flag): Boolean = params match {
      case sym :: _ => sym.is(flag)
      case _ => false
    }
       
    def isUsingClause(params: List[Symbol]): Boolean  =
       headHasFlag(params, Flags.Given)

    def isTypeParamClause(params: List[Symbol]): Boolean =
       headHasFlag(params, Flags.TypeParam)
```

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
```diff
@@ -117,31 +144,100 @@ class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx) {
...
+        if (gsym.name.isRightAssocOperatorName) {
+          val (leadingTyParamss, rest1) = paramss.span(isTypeParamClause)
+          val (leadingUsing, rest2) = rest1.span(isUsingClause)
+          val (rightTyParamss, rest3) = rest2.span(isTypeParamClause)
+          val (rightParamss, rest4) = rest3.splitAt(1)
+          val (leftParamss, rest5) = rest4.splitAt(1)
+          val (trailingUsing, rest6) = rest5.span(isUsingClause)
+          if (leftParamss.nonEmpty)
+            leadingTyParamss ::: leadingUsing ::: leftParamss ::: trailingUsing ::: rightTyParamss ::: rightParamss ::: rest6
+          else paramss // it wasn't a binary operator, after all.
```
dos65, line 221: Is there a test case on right accoc extension method?

dos65 (commented):
Cool! It was annoying for me seeing `extension (...) def ..` in completions. Thx for fixing that!

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
```diff
@@ -62,13 +62,12 @@ class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx) {
...
-  def hoverDetails(
+  def hoverDetailString(
       sym: Symbol,
       history: ShortenedNames,
-      info: Type,
-      addFullDef: Boolean = true
+      info: Type
   )(using Context): String = {
-    val fullDef = if (addFullDef) fullDefinition(sym, info) else ""
+    val fullDef = fullDefinition(sym, info)
```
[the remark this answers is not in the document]
tgodzik: Yeah, I was thinking about it, but did not want to touch too much in this PR. I guess I already did that :sweat_smile: 

Overall, I think it's useful to separate these for completions and hovers, I always had issues figuring it out how to make both work.

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
tgodzik: Not currently, I will add that.

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
```diff
@@ -117,31 +144,100 @@ class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx) {
...
     }.iterator
+    val paramLabelss = label(methodParams)
+    val extLabelss = label(extParams)
+
+    val returnType = typeString(shortType(gtpe.finalResultType, shortenedNames))
+    val extensionSignature = paramssString(extLabelss, extParams)
+
+    val extensionSignatureString =
+      if (extParams.nonEmpty) extensionSignature.mkString("extension ", "", " ")
+      else ""
```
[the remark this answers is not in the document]
tgodzik: Actually, not sure if I understood that.

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
tgodzik: Added!

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
tgodzik: Thanks, fixed.

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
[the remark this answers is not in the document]
tgodzik: Ok, I merged the two methods and reworked it all a bit. Let me know what you think!

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
```diff
@@ -23,71 +24,92 @@ class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx) {
...
+      typeSymbol.owner.fullName.stripModuleClassSuffix.toString
+
+    def name: String = nameString(sym)
+
+    def parentPackagesString: String = {
+      val owners = for {
+        sym <- sym.ownersIterator
+        if !(sym.isRoot || sym.isEmptyPackage)
+      } yield super.nameString(sym.name)
+      owners.toList.reverse.mkString(".")
```
dos65, line 56: wdty about moving it into MtagsEnrichments?
There is a similar method: `fullNameBackticked` so the common part of `take to the root or empry` might be reused

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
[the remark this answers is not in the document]
dos65: This is pretty minor - `extensionSignatureString` is used only if `onlyMethodParams = false`.
Now it's at line 177

dos65 (approved):
LGTM!
Other comments are minor - feel free to ignore

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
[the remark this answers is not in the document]
dos65: Cool! IMO, now it's easier to understand rendering rules)

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
tgodzik: Actually, we can just use `fullNameBackticked` :sweat_smile: Removed the method altogether.

mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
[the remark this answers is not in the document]
tgodzik: Changed it into a def

Issue #2977, opened 2021-07-21
"-print-lines" breaks auto-completion


yuriy-yarosh:
**Describe the bug**

"-print-lines" dotty arg breaks auto-completion.

**To Reproduce**

Use these scalacOptions 

```scala
ThisBuild / scalacOptions ++=
  "-encoding" :: "UTF-8" ::
    "-unchecked" ::
    "-explain" ::
    "-rewrite" ::
    "-new-syntax" ::
    "-pagewidth" :: "180" ::
    "-print-lines" ::
    "-Yexplicit-nulls" ::
    "-Ycheck:all" ::
    "-Xmax-inlined-trees" :: "4000000" ::
    "-Xmax-inlines" :: "128" ::
    Nil
```

Results in

```scala
import cats.effect.`                       |Clock`
```

**Expected behavior**

We shouldn't get proper autocompletion 
```scala
import cats.effect.Clock
```

**Screenshots**

![screenshot](https://i.imgur.com/yk3YUj7.png)

**Installation:**

- Operating system: Arch
- Editor: Visual Studio Code
- VSCode Metals plugin version: v1.10.7
- sbt-bloop 1.4.8-19-4d9f966b

tgodzik:
Thanks for reporting! Not sure why it's happening, but we can for sure just filter those options out. They are not needed by teh presentation compiler.

Fixing it in https://github.com/scalameta/metals/pull/2979

yuriy-yarosh:
NP, Tomasz.
Thanks for the followup and the fix.

PR #2979, opened 2021-07-21
[Scala 3] Filter out -print-lines and -print-tasty options

tgodzik:
They are not needed for the presentation compiler and break completions.

Fixes https://github.com/scalameta/metals/issues/2977


dos65 (approved):
LGTM

PR #2980, opened 2021-07-21
References - filter out wrong results from unrelated build targets

### LINKED ISSUE #2901: Rename refactoring affects unrelated targets.
I created a minimal project to reproduce the issue. Follow the `README.md` here: https://github.com/agilesteel/rename-issue

dos65:
Fixes #2901

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
@@ -22,6 +22,7 @@ import scala.meta.internal.semanticdb.TextDocuments
 import scala.meta.internal.{semanticdb => s}
 import scala.meta.io.AbsolutePath
 
+import ch.epfl.scala.bsp4j.BuildTargetIdentifier
 import com.google.common.hash.BloomFilter
 import com.google.common.hash.Funnels
 import org.eclipse.lsp4j.Location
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -174,6 +175,59 @@ final class ReferenceProvider(
...
+    buildTargets.inverseSources(source) match {
+      case None => Seq.empty
+      case Some(id) =>
+        val allowedBuildTargets = buildTargets.allInverseDependencies(id)
+        val visited = scala.collection.mutable.Set.empty[AbsolutePath]
+        val result = for {
+          (path, bloom) <- index.iterator
+          if isSymbol.exists(bloom.mightContain)
+          scalaPath = AbsolutePath(path)
+          if buildTargets.inverseSources(scalaPath).exists(allowedBuildTargets)
```
ckipp01, line 194: So I sort of wish we had some benchmarks for where we are before and after this. Not saying we have to add them now, but we should probably think about getting some fresh ones. Do we have any performance concerns in situations where the bloom actually contains a _lot_ of references and we're therefore doing this query for every single one?

Out of total curiosity I'm curious if it' save us anything to actually store the build target in the index alongside the path. I'm not convinced that it'd really save us much since we'd still have to look that up every time an `onChange` happened and we updated it. 

Either way, am I worrying about this for no reason, or could this potentially slow down finding refernces?

dos65: Hm.. Honestly, I haven't checked that `index` is just a map :smiley: 
I was thinking that it's probably some complicated structure like we have in `ClasspathSearch`. 

I changed it to keep build target id in index entry to not recompute for every reference search. Not it should be more effective for case you described.

I don't think that there is need in benchmark for this now.

ckipp01 (approved):
👍🏼 Looks good to me LGTM!

tgodzik (approved):
LGTM!

Issue #2982, opened 2021-07-22
"Insert type annotation" available in vim ?


armeltd:
Hello !

I can't figure out how to have the "Insert type" action in vim. It doesn't appear in the choices.
Is it expected ?

![image](https://user-images.githubusercontent.com/29709300/125683925-aa978488-44c8-4499-8974-53fbc6f7c8de.png)

OS: Linux
vim version: NVIM v0.5.0
node version: v12.22.0
coc.nvim version: 0.0.80-aa24232066
coc-metals: 1.0.7
scala: 2.13.5

_Originally posted by @armeltd in https://github.com/scalameta/metals/issues/2875#issuecomment-880171397_

tgodzik:
Thanks for reporting! It looks like the position sent to Metals is that of `"tot"` as the code action suggests the `Convert to interpolation string` code action. Any idea what's going here @ckipp01  ?

ckipp01:
@armeltd are you able to post your mapping that you're using to trigger the code action?

armeltd:
Issue fixed indeed it was a binding issue.
Here's what I had :
```
" Remap for do codeAction of current line
xmap <leader>a  <Plug>(coc-codeaction-line)
nmap <leader>a  <Plug>(coc-codeaction-line)
```
I think I copy pasted them a couple of months ago from the [metals website](https://scalameta.org/metals/docs/editors/vim#recommended-cocnvim-mappings). 

the default bindings have changed now (line  => selected) and it works when I get the new ones :
```
" Applying codeAction to the selected region.
" Example: `<leader>aap` for current paragraph
xmap <leader>a  <Plug>(coc-codeaction-selected)
nmap <leader>a  <Plug>(coc-codeaction-selected)
```
![image](https://user-images.githubusercontent.com/29709300/126761777-36a21fb3-ee93-4f91-ac9f-89528de5196f.png)

Thanks for all what your doing it is really awesome. :+1:

ckipp01:
Glad it's fixed. If you are able to, please do shoot in a PR to update the bindings, which can be found here: https://github.com/scalameta/metals/blob/main/docs/editors/vim.md#recommended-cocnvim-mappings. I'll go ahead and leave this open a bit until we get the mappings updated. If you don't have time I should be able to quickly do it later today.

armeltd:
The default mappings on the site now are working so they don't need any update. it is just the old ones at the time I copy pasted that were not. It is just to "old users" to remind to update their conf sometimes. 
[...]

PR #2985, opened 2021-07-23
doc(): update default action binding for coc.nvim

armeltd:


ckipp01:
Thanks for the pr! So tbh, this is an opinionated change. Since by default the coc.nvim example in their README does include 

```
" Applying codeAction to the selected region.
" Example: `<leader>aap` for current paragraph
xmap <leader>a  <Plug>(coc-codeaction-selected)
nmap <leader>a  <Plug>(coc-codeaction-selected)
```
So they are meant to be used when the code has been selected. Unless this is out of date in their own docs. Whereas the ones you're included are meant to be used without a selection, but rather on top of the symbol. Part of me doesn't like having a default configuration on the website since it really is a coc.nvim detail that needs to be set, but not really specific to coc-metals.

Anyways, I'm not sure how I feel about changing these since they are the defaults even that coc.nvim recommends. The ones you are changing them to can be included, but as an additional mapping, not a replacement. wdyt?

armeltd:
Mmm in fact I was blindly doing it after @otto-dev 's advice : https://github.com/scalameta/metals/issues/2982#issuecomment-885574942
I let you guys decide if it is needed or not. 
From what I have seen there is no issue with what the current docs says.

ckipp01:
@armeltd I'm actually going to go ahead and close this. I believe I'm going to instead just point to the coc.nvim docs to ensure users are able to pull their default mappings from there, and then only include the coc-metals specific ones in our docs. Thanks again for sending this in and bringing up the topic.

PR #2987, opened 2021-07-23
Correctly paste extension groups

tgodzik:
Previously, extension groups would not be picked up when pasting and the indentation would be wrong. Now they are handled in the regex. I also split the original regex to make it more readable.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
@@ -8,8 +8,13 @@ import org.eclipse.lsp4j.TextEdit
 
 object IndentOnPaste extends RangeFormatter {
 
+  private val noEndClause = raw"((<!\bend\b\s*?)\b(if|while|for|match|try))"
+  private val ifThenClause = raw"(\bif\s+(?!.*?\bthen\b.*?$$)[^\s]*?)"
+  private val keywordClause = raw"\b(then|else|do|catch|finally|yield|case)"
+  private val openBraceClause = raw"""(^.*\{[^}"']*$$)"""
+  private val extensionClause = raw"""extension\s*((\(|\[).*(\)|\]))+"""
   val increaseIndentPatternRegex: Regex =
-    raw"""(((<!\bend\b\s*?)\b(if|while|for|match|try))|(\bif\s+(?!.*?\bthen\b.*?$$)[^\s]*?)|(\b(then|else|do|catch|finally|yield|case))|=|=>|<-|=>>|:)\s*?$$|(^.*\{[^}"']*$$)""".r
+    raw"""($noEndClause|$ifThenClause|$keywordClause|$extensionClause|=|=>|<-|=>>|:)\s*?$$|$openBraceClause""".r
   val indentRegex: Regex = raw"\S".r
 
   private def codeStartPosition(line: String): Option[Int] =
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
```diff
@@ -8,8 +8,13 @@ import org.eclipse.lsp4j.TextEdit
 
 object IndentOnPaste extends RangeFormatter {
 
+  private val noEndClause = raw"((<!\bend\b\s*?)\b(if|while|for|match|try))"
+  private val ifThenClause = raw"(\bif\s+(?!.*?\bthen\b.*?$$)[^\s]*?)"
+  private val keywordClause = raw"\b(then|else|do|catch|finally|yield|case)"
+  private val openBraceClause = raw"""(^.*\{[^}"']*$$"""
```
dos65, line 14: Should it start from `(` or should it have `)` at the end?
Originally it was: `(^.*\{[^}"']*$$)`

Giggiux: It should have the complete grouping, so 
```scala
private val openBraceClause = raw"""(^.*\{[^}"']*$$)"""
```

And then the full `increaseIndentPatternRegex` becomes 

```scala
raw"""($noEndClause|$ifThenClause|$keywordClause|$extensionClause|=|=>|<-|=>>|:)\s*?$$|$openBraceClause""".r
```

I believe that the confusion arose from the initial opening bracket that is closed after the arrows matching, with a `:)`, not at the pure end of the regex.

tgodzik: Fixed! The only problem with the extension regex is that it will not work with mutiline extension definition, but I think it's fine for now.

metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
```diff
@@ -8,8 +8,13 @@ import org.eclipse.lsp4j.TextEdit
 
 object IndentOnPaste extends RangeFormatter {
 
+  private val noEndClause = raw"((<!\bend\b\s*?)\b(if|while|for|match|try))"
+  private val ifThenClause = raw"(\bif\s+(?!.*?\bthen\b.*?$$)[^\s]*?)"
+  private val keywordClause = raw"\b(then|else|do|catch|finally|yield|case)"
+  private val openBraceClause = raw"""(^.*\{[^}"']*$$)"""
+  private val extensionClause = raw"""extension\s*((\(|\[).*(\)|\]))+"""
```
Giggiux, line 15: This allows to have an increase pattern also in the following case:
```scala
extension [T] \\ here it will increase the Intendation
  (using b: B)(s: String) 
```

will this be a valid syntax?

And I'm not sure that other regexes will handle the indentation after the second line, but as @tgodzik said, it doesn't handle multiple lines.

tgodzik: Och, I think this should not increase the indentation, that's msot likely not a valid syntax.

tgodzik: Actually, the syntax works, but it was not the intention :O

tgodzik: Scratch that! Funnily enough, we should indent after anything because:
```scala
extension [T] \\ here it will increase the Intendation
  (using b: B)(s: String) 
```

is valid while:
```scala
extension [T] \\ here it will increase the Intendation
(using b: B)(s: String) 
```
is not.

dos65 (approved):
:+1:

Giggiux:
👍

PR #2988, opened 2021-07-23
Point to coc.nvim for defaults rather than have them in here.

ckipp01:
Since these can easily get stale, I think we should instead just point
to coc.nvim to get the default mappings for coc.nvim users. This now
only includes the `coc-metals` specific mappings.

This is a follow-up to the convo in https://github.com/scalameta/metals/pull/2985


tgodzik (approved):
LGTM!

PR #2989, opened 2021-07-25
Include parens on lambda in Scala 3 when inserting type in params with brace.

ckipp01:
In Scala 2 when given the following piece of code and using the insert
inferred type functionality to insert the type of `x`, there is no issue
not including the parens.

This:
```scala
List(1, 2, 3).map{<<x>> => x * 2}
```

Becomes this:
```scala
List(1, 2, 3).map{x: Int => x * 2}
```

However in Scala 3 the compiler will error on this:

```
Diagnostics:
1. parentheses are required around the parameter of a lambda
   This construct can be rewritten automatically under -rewrite -source 3.0-migration.
```

This change makes sure that in this scenario in Scala 3 that the parens
are added so that it becomes:

```scala
List(1, 2, 3).map{(x: Int) => x * 2}
```

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
@@ -76,6 +76,7 @@ final class InferredTypeProvider(
       val printer = ScopeAwareTypePrinter(indexedCtx)
       printer.typeString(short)
     }
+
     /*
      * Get the exact position in ValDef pattern for val (a, b) = (1, 2)
      * Suprisingly, val ((a, c), b) = ((1, 3), 2) will be covered by Bind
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -133,23 +134,42 @@ final class InferredTypeProvider(
...
-              else
-                baseEdit(withParens = false) :: Nil
-
-            case _ => baseEdit(withParens = false) :: Nil
+              checkForParensAndEdit(appl.fun.endPos.end, '(', block.startPos)
+
+            // labda `map{a => ???}` apply
+            // Ensures that this becomes {(a: Int) => ???} since parentheses
+            // are required around the parameter of a lambda in Scala 3
+            case valDef :: defDef :: (block: untpd.Block) :: (_: untpd.Block) :: (appl: untpd.Apply) :: _
```
ckipp01, line 167: I corrected the comments in here since I _think_ the one above was wrong?

For
```scala
map{a => ???}
```
It won't be caught by the case up above, so it shouldn't be in that comment right? Since this gets translated to:
```scala
map({a => ???})
```
That's where there is an extra block in the path before the apply right?

tgodzik: Yeah, actually the comments were wrong. For `{}` we just annotated it with the type in `baseEdit`

tgodzik (approved):
Thanks for fixing this! I have't realized there was a change here.

PR #2993, opened 2021-07-27
Mention new run/test commands in the VS Code editor page

tgodzik:
Just realized that we haven't actually put those anywhere in the docs.

Came up in https://github.com/scalameta/metals/discussions/2992

### DIFF
--- a/docs/editors/vscode.md
+++ b/docs/editors/vscode.md
@@ -49,9 +49,9 @@ variables.
 ## Configure Java version
 
 The VS Code plugin uses by default the `JAVA_HOME` environment variable (via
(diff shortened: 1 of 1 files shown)


dos65 (approved):
:+1:

ckipp01 (approved):
Good call. They are actually mentioned in here https://scalameta.org/metals/docs/integrations/debug-adapter-protocol, but I'm assuming a lot of people aren't looking there.

tgodzik:
> Good call. They are actually mentioned in here https://scalameta.org/metals/docs/integrations/debug-adapter-protocol, but I'm assuming a lot of people aren't looking there.

Yeah, I realized that I wanted to mention all the possible way to run a test, but there is no single point in the docs that has everything.

PR #2998, opened 2021-07-28
[Scala3] Additional completion pos fix for `Select` tree

dos65:


ckipp01 (approved):
I never realized this is what was causing the 
```
Exception in thread "pool-9-thread-1" java.lang.StringIndexOutOfBoundsException: begin 105, end 97, length 131
```
I've seen this before and didn't have time to dig into it. TIL. LGTM for me, testing it out locally gets rid of the issue for me.

tgodzik (approved):
LGTM!

PR #2999, opened 2021-07-28
Add more tests on using sbt as the build server

adpi2:
## Customizable response of show-message requests in `TestingClient`

https://github.com/scalameta/metals/pull/2999/commits/5cf9bbcf53621884f330a98bf92c63ff41d5e142 makes the response of the show-message requests customizable on a test basis.

For instance, in a test one can write:
```scala
// do not import the build
client.importBuild = ImportBuild.notNow

// import the build
client.importBuild = ImportBuild.yes

// select the "Bob" bsp server
client.selectBspServer = { actions => actions.find(_.getTitle("Bob")).get } 
```

The initial motivation was to stop running `sbt bloopInstall` before connecting to sbt server in the `SbtServerSuite`. I then decided to make every response customizable so to make clear which test depends on what message/response with the client.

## Add the `BuildServerInitializer` axis for testing

In https://github.com/scalameta/metals/pull/2999/commits/5fba671e3952ac04aba075ec5fbde1be6c22ca9a

 The `BuildServerInitializer` is an abstraction trait on how to initialize the build server.
It has 3 implementations:
- `QuickBuildInitializer`: Read a metals.json to generate the bloop configuration files
- `BloopImportInitializer`: Connect to Bloop after running bloop install (works with sbt, maven, gradle or mill)
- `SbtServerInitializer`: Generate the .bsp/sbt.json configuration file and switch to sbt build server

Any existing test class can now be abstracted to run on different `BuildServerInitializer`s.
Currently the tests in `unit` use the fastest `QuickBuildInitializer`, the sbt, maven, gradle and mill tests use the `BloopImportInitializer` except the `SbtServerSuite` which uses the `SbtServerInitializer`.

## Add the `BuildToolLayout` axis for testing

In https://github.com/scalameta/metals/pull/2999/commits/138245c32bab5fd05ab4f55da2303743c485f019

The `BuildToolLayout` is an abstraction trait on the build tool configuration files layout. It currently has 2 implementations:
- `QuickBuildLayout`: Creates a metals.json file
- `SbtBuildLayout`: Creates a project/build.properties and a build.sbt file
We could also define the `MillBuildLayout`, the `MavenBuildLayout`...

Using a `BuildToolLayout` one can write a test whose layout is independent of the build tool.

Example:
```scala
test("example") {
 val layout = buildToolLayout(
    """|/a/src/main/scala/a/Main.scala
        |package a
        |
        |object Main {
        |  class Preceding
        |
        |  def main(args: Array[String]): Unit = {
        |    println()
        |    System.exit(0)
        |  }
        |}
        |""".stripMargin,
   scalaVersion =  "2.12.14"
 )

  for {
    _ <- initialize(layout)
   ....
  }
}
```

A `BuildToolLayout` can be combined with a `BuildServerInitializer`, as in:
```scala
class A extends QuickBuildInitializer with QuickBuildLayout
class B extends BloopImportInitializer with SbtBuildLayout
class C extends SbtServerInitialize with SbtBuildLayout
```

## Run DAP suites with sbt server

In https://github.com/scalameta/metals/pull/2999/commits/e2c6697f0ce5669ca769cf58ad59e637ed187271

The `CrossDebugSuite`, `BreakpointDapSuite`, `StackFrameDapSuite` and `StepDapSuite` are abstracted over the `BuildServerInitalizer` and the  `BuildToolLayout` so that they can still run with Bloop (quick build) but also with sbt server. 

example:
```scala
class BreakpointDapSuite 
    extends BaseBreakpointDapSuite("debug-breakpoint")
    with QuickBuildInitializer
    with QuickBuildLayout

class SbtBreakpointDapSuite
    extends BaseBreakpointDapSuite("sbt-debug-breakpoint")
    with SbtServerInitializer
    with SbtBuildLayout
```


tests/unit/src/main/scala/tests/BuildServerInitializer.scala
```diff
@@ -0,0 +1,110 @@
...
+  def initialize(layout: String, expectError: Boolean = false): Future[Unit]
+}
+
+trait QuickBuildInitializer extends BuildServerInitializer {
+  this: BaseLspSuite =>
+  override def initialize(
+      layout: String,
+      expectError: Boolean = false
+  ): Future[Unit] = {
+    Debug.printEnclosing()
```
ckipp01, line 26: I'm not 100% sure, but this was always called before in the `initialize` Do we only want to call this here, or do we also want to for the other initializers?

tests/unit/src/main/scala/tests/BuildServerInitializer.scala
```diff
@@ -0,0 +1,110 @@
...
+trait BloopImportInitializer extends BuildServerInitializer {
+  this: BaseLspSuite =>
+  override def initialize(
+      layout: String,
+      expectError: Boolean = false
+  ): Future[Unit] = {
+    writeLayout(layout)
+    for {
+      _ <- server.initialize()
+      // Import build using Bloop
```
ckipp01, line 49: It's probably somewhat obvious for people that are familiar with the process, but for someone not maybe we should instead just add some Scaladocs to these initializers. Something like

```
/**
 * Set up your workspace using QuickBuild as your build tool. This will take your `metals.json` file and quickly
 * produce `.bloop/` files from it.
 * /
QuickBuildInitializer
```

```
/**
 * Set up your workspace by responding to a Import Build request which will
 * run Bloop Install via the build tool being used.
 * /
BloopImportInitializer
 ```

```
/**
 * Assumes sbt is being used as a build tool and also for your BSP server. This sets
 * up the necessary sbt files and also generates the corresponding bsp config file.
 * /
SbtServerInitializer
 ```
 
Something like that more or less so it's really clear what is happening and suppose to happen with these.

tests/unit/src/main/scala/tests/BuildServerInitializer.scala
```diff
@@ -0,0 +1,110 @@
...
+      layout: String,
+      expectError: Boolean = false
+  ): Future[Unit] = {
+    writeLayout(layout)
+    generateBspConig()
+    for {
+      _ <- server.initialize()
+      _ <- server.initialized()
+      // choose sbt as the Bsp Server
+      _ = client.selectBspServer = { _ => new MessageActionItem("sbt") }
```
ckipp01, line 72: Since we just chose sbt here, do we still need the line below this where we do a `bsp-switch` and choose sbt again?

tests/slow/src/test/scala/tests/gradle/GradleLspSuite.scala
```diff
@@ -62,19 +62,16 @@ class GradleLspSuite extends BaseImportSuite("gradle-import") {
     } yield {
       assertNoDiff(
         client.workspaceMessageRequests,
-        List(
-          // Project has .bloop directory so user is asked to "re-import project"
-          importBuildChangesMessage,
-          progressMessage
```
ckipp01, line 68: How come in this situation and the one down below the `progressMessage` is gone?

tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
```diff
@@ -13,43 +12,48 @@ import scala.meta.io.AbsolutePath
...
 class SbtServerSuite
     extends BaseImportSuite("sbt-server")
-    with ScriptsAssertions {
+    with SbtServerInitializer {
 
   val preBspVersion = "1.3.13"
   val supportedBspVersion = V.sbtVersion
   val scalaVersion = V.scala212
   val buildTool: SbtBuildTool = SbtBuildTool(None, () => userConfig)
+  val sbtCommonSettings: String =
```
ckipp01, line 28: Not a huge deal at all, but these are also defined in the `SbtBuildLayout`. Do you think it's worth somehow putting them in the same place so we can utilize it wherever we want the common settings?

tests/unit/src/test/scala/tests/NewFileLspSuite.scala
```diff
@@ -402,13 +402,14 @@ class NewFileLspSuite extends BaseLspSuite("new-file") {
...
-                                  |}
-                                  |$existingFiles
-                                  |""".stripMargin)
+        _ <- initialize(
+          s"""/metals.json
+             |{
+             |  "a": { }
+             |}
+             |$existingFiles
+          """.stripMargin
```
ckipp01, line 411: Super random, but it's odd to me that the formatting changes differently here, vs in other files. For example in the NewProjectLspSuite:

```scala
       _ <- initialize(s"""
                           |/metals.json
                           |{
                           |  "a": { }
                           |}
                           |""".stripMargin)
```
It's like sometimes the entire string is shifted down and in, but in others it's not. Unless I'm totally missing the pattern.

ckipp01 (changes requested):
Really fantastic job @adpi2. Impressive work. I think this will really improve the clarity around BSP related stuff in testing and definitely allow for things to be easily extended in the future. I just left a few comments/suggestions/questions, but nothing major. Also, thanks for the great description of everything.

tests/slow/src/test/scala/tests/gradle/GradleLspSuite.scala
ckipp01: I noticed this in a lot of places actually.

tests/unit/src/main/scala/tests/BuildServerInitializer.scala
adpi2: We definitely want them everywhere.

tests/unit/src/main/scala/tests/BuildServerInitializer.scala
adpi2: Thanks!

tests/unit/src/main/scala/tests/BuildServerInitializer.scala
adpi2: This line means that the client will respond with "sbt" if prompted to choose a bsp server. The next line triggers the switch, which sends the `SelectBspServer` message to the client. So both lines are useful.

Yet I think the "sbt" parameter in `server.executeCommand(ServerCommands.BspSwitch.id, "sbt")` is useless because the `BspSwitch` command doesn't expect a parameter:
https://github.com/scalameta/metals/blob/3c83447ec658f87fdccbfb3f0a39fca1cec4ef6e/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1601

tests/slow/src/test/scala/tests/gradle/GradleLspSuite.scala
adpi2: That's because the default response of the `importBuildChanges` is "notNow" (it was "yes" before).
I think it is better to write it explicitly when we want to test a build re-import :
```diff
for {
    ...
+     _ = client.importBuildChanges = ImportBuildChanges.yes
      _ <- server.didSave("build.sc")(identity)
    } yield {
      assertNoDiff(
        client.workspaceMessageRequests,
+       List(
          // Project has .bloop directory so user is asked to "re-import project"
          importBuildChangesMessage,
+         progressMessage
+        ).mkString("\n")
      )
    }
```

Testing that once for each build tool is enough and the rest of the time we just want to test that the user is prompted.
Yet I probably removed too much of them so I am adding some back.

tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
adpi2: yes thanks

tests/unit/src/test/scala/tests/NewFileLspSuite.scala
adpi2: I think it depends on whether there is a line break between the `(` and the `s""""`.
I prefer with a line break because that makes formatting more stable.

tests/unit/src/main/scala/tests/BuildServerInitializer.scala
ckipp01: Ahhh gotcha gotcha. That make sense 👍🏼

tests/slow/src/test/scala/tests/gradle/GradleLspSuite.scala
ckipp01: Sounds good 👍🏼

ckipp01 (approved):
Looks good. Again, great job on this! LGTM 🚀

tests/slow/src/test/scala/tests/feature/CompletionCrossLspSuite.scala
```diff
@@ -26,7 +26,7 @@ class CompletionCrossLspSuite
   test("serializable-2.13") {
     cleanWorkspace()
     for {
-      _ <- server.initialize(
```
tgodzik, line 29: Could we instead of adding the hierachy via `initialize` method, just do the class hierachy inside TestServer?
```
trait TestServer { def initialize(...) : Future[Unit] }
class SbtServer extends TestServer { def initialze(...) = ...}
class QuickBuildServer extends TestServer { def initialze(...) = ...}
class BloopServer extends TestServer { def initialze(...) = ...}
```
this would avoid a lot of the changes and reduce the size of the PR.

tgodzik: We could also just keep BaseLSPSuite and jut add a constructor parameter on what to use, with default set to quickbuild

adpi2: I don't think that would be any better, because I think it is clearer to keep things separated:
- `TestingServer` is about the Metals language server: what request or command we send and what response we are waiting from it
- `TestingClient` is about the user interaction with the language client: how the user responds to the show message requests
- `workspace` is about writing some files into the filesystem

Whenever we need to coordinate the workspace, the server and the client we should do in the suite level.

If we mix everything inside `TestingServer` then it makes things confusing. (In your proposition `SbtServer` is not really the sbt server, it is the Metals language server in an sbt repository connected to sbt build server. But `BloopServer` could also be in an sbt repository but connected to Bloop and `QuickBuildServer` is also connected to Bloop). So IMO it is simpler to keep things out of `TestingServer`.

So yes that bring a lot of changes but I really think it makes the code easier to navigate and more extensible.

tgodzik: > TestingServer is about the Metals language server: what request or command we send and what response we are waiting from it

I don't believe we need to be super strict about it, it can be both language server and build server, why would that be a problem? We can easily add a new server if we need to that. 

> So yes that bring a lot of changes but I really think it makes the code easier to navigate and more extensible.

Is it though? Adding more traits add more things to navigate through and mixing in traits is not really somethings that is easier to reason about. It's much easier to navigate methods than class hierarchy.  Also, I would really keep changes minimal, since it makes reviews quicker. It's a standard in scalafmt reviews and I think it works really good.

But anyway, what do others think about it?

tgodzik: I will try to make a full review next week.

adpi2: Okay I did an alternative version https://github.com/scalameta/metals/pull/3006 where `BuildServerInitializer` and `BuildToolLayout` are passed as parameters rather than being mixed in as traits.

adpi2:
Closing in favor of #3006

tests/slow/src/test/scala/tests/feature/CompletionCrossLspSuite.scala
ckipp01: I mean if the goal is size, then there really isn't much of a difference at all.

Regarding styles, we barely use mixins at all in Metals, which is fine. It typically personally don't like the heavily use them at all. So when I first saw them in here I was skeptical, but I do find the current approach readable and easy-to-navigate. Plus there are places where we maybe avoid class hierarchy too much and end up with some pretty massive files because of it. I don't feel strong on this. This approach is fine to me, but if others strongly prefer the other in #3006, then I have no issue with that either.

PR #3004, opened 2021-07-29
Minimize file rewrites in `.metals/readonly/dependencies`

dos65:
Previously, `toFileOnDisk` was always updating dependency source files in
order to keep them up to date.

After introducing `.metals/readonly/dependencies/$jar-name` directories, it is possible to check if the source file is up to date and avoid non-necessary rewrites.
Now, this function also write `jar.meta` file that contains original jar path and modified time that is used for this purpose.

#### Additional context

It should fix the flaky `WorkspaceLspSuite.dependencies.` [test](https://github.com/scalameta/metals/pull/2998/checks?check_run_id=3180754434).
I'm not totally sure, but it seems that the reason of failure was related to the concurrent file rewriting. It might happen in references search.

I think that there is chance to get `$path.exists == false` if the operation `Files.copy($jarPath, $path, REPLACE_EXISTING)` happens at same moment even if `$path` exist. This the only that might explain the problem with this test.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -1,5 +1,6 @@
 package scala.meta.internal.metals
 
+import java.io.IOException
 import java.net.URI
 import java.nio.charset.StandardCharsets
 import java.nio.file.FileAlreadyExistsException
@@ -31,6 +32,7 @@ import scala.util.control.NonFatal
 import scala.{meta => m}
 
 import scala.meta.inputs.Input
+import scala.meta.internal.io.FileIO
 import scala.meta.internal.mtags.MtagsEnrichments
 import scala.meta.internal.parsing.EmptyResult
 import scala.meta.internal.semanticdb.Scala.Descriptor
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -352,6 +380,40 @@ object MetalsEnrichments
...
+              withJarDirLock(jarDir) {
+                if (updateMeta) {
+                  val prevFiles = FileIO
+                    .listAllFilesRecursively(jarDir)
+                    .filter(_.filename != ".lock")
+                  prevFiles.foreach(_.delete())
+                  Files.write(jarMetaFile.toNIO, jarMeta.getBytes)
+                }
+                copyFile(path, out)
+              }(retry)
```
ckipp01, line 408: Probably not a real issues, but could this ever cause a loop? If there is an `IOException`, then if we wait a bit and try again it should be fine, but is there any situation where that exception could keep happening and fall into an infinite loop?

ckipp01 (commented):
Good job hunting this down. From what i can tell this looks good. Just one question though.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -352,6 +380,40 @@ object MetalsEnrichments
...
+            val jarDir =
+              workspace.resolve(Directories.dependencies).resolve(jar.filename)
+            val out = jarDir.resolveZipPath(path.toNIO)
+            val jarMetaFile = jarDir.resolve("jar.meta")
+
+            lazy val currentJarMeta = readJarMeta(jarMetaFile)
+            lazy val jarMeta = toJarMeta(jar)
+
+            val updateMeta = !jarDir.exists || !currentJarMeta.contains(jarMeta)
+            if (!out.exists || updateMeta) {
```
tgodzik, line 398: Wouldn't it be enough to just check if the file exists and then not copy it? Since, we have directory based on the jar name and jar names contain the version, it should never really change, only if someone creates snapshot jars locally, but in that case they will also look at the original sources.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
dos65: Good point. I think there non-zero chance to fall into it due to some extra obstacles like unexpected workspace directory ownership changes  /not enough space on disk etc.
Added an exit clause to avoid infinity loop.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
dos65: I agree that it feels a bit of redundantly but this mechanic with jar-name is far from being perfect. 

There might be a lot of rare cases when jar name won't guarantee that it's the same artifact.
In addition to local snapshots it might be a snapshot jar from repository. Or it might be a forked artifact with the same version (like: `org.scalameta.metals-..` and `com.virtuslab.metals-..`).

I think that even in these rare cases, it is worth preventing the use of the wrong sources, if possible. 
Otherwise someone might be confused/spend time investigating why the sources is different from a real execution before he realize that it's outdated and only manual cleanup will help.

tgodzik: Makes sense!

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -352,6 +386,43 @@ object MetalsEnrichments
...
+        throw new Exception(s"Unable to save $path in workspace")
+      } else if (path.toNIO.getFileSystem == workspace.toNIO.getFileSystem) {
+        path
+      } else {
+        path.jarPath match {
+          case Some(jar) =>
+            val jarDir =
+              workspace.resolve(Directories.dependencies).resolve(jar.filename)
+            val out = jarDir.resolveZipPath(path.toNIO)
+            val jarMetaFile = jarDir.resolve("jar.meta")
```
tgodzik, line 401: Maybe let's add a dot so that it's a hidden file in some systems?
```suggestion
            val jarMetaFile = jarDir.resolve(".jar.meta")
```

tgodzik (approved):
LGTM, just a small suggestion.

PR #3005, opened 2021-07-30
Add in sbt-sourcegraph to upload LSIF to Sourcegraph.

ckipp01:
So I wanted to throw this up here to get others thoughts on this. I've noticed myself a lot more recently relying on Sourcegraph to quickly browse through code online at times especially if looking at a pr. Currently when you're looking at the Metals codebase all the results you're seeing are search-base:

Screenshot 2021-07-30 at 12 38 40

While this is already super helpful, it'd be amazing to have even [better code intelligence](https://docs.sourcegraph.com/code_intelligence/explanations/precise_code_intelligence) when on GitHub. If I'm understanding everything correctly, we can do this by utilizing sbt-sourcegraph to turn the semanticDB created into LSIF for the project, and then upload it to Sourcegraph. @olafurpg am I understanding this correctly?

The necessary stuff is pretty minimal, and I did the actual command in a way that doesn't change our build definition at all, but the sbt command in the `sourcegraph.yml` takes care of everything. The only change would be adding the plugin.

The only thing I'm not fully sure of is if any org can just upload LSIF? Looking at the plugin I saw that it defaults to the main Sourcegraph, but am unsure how it's managed to ensure someone doesn't just randomly upload LSIF for a repo. Do we need an account or anything. Just throwing this up here to see if there is interest in it, since I know I'd use it.

### DIFF
--- a/.github/workflows/sourcegraph.yml
+++ b/.github/workflows/sourcegraph.yml
@@ -1,0 +1,28 @@
+name: Sourcegraph
+on:
+  push:
+    branches:
+      - main
+    pull_request:
+
+jobs:
+  lsif:
+    runs-on: ubuntu-latest
+    name: "Upload LSIF"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: olafurpg/setup-scala@v12
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "1.15.6"
+      - run: |
+          mkdir -p bin
+          curl -L https://sourcegraph.com/.api/src-cli/src_linux_amd64 -o bin/src
+          chmod +x bin/src
+          export PATH="$PATH:$PWD/bin"
+          sbt \
+              'set every semanticdbEnabled := true' \
+              'set every semanticdbVersion := "4.4.24"' \
+              sourcegraphUpload
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -4,6 +4,7 @@ addSbtPlugin("com.geirsson" % "sbt-ci-release" % "1.5.7")
 addSbtPlugin("org.scalameta" % "sbt-mdoc" % "2.2.22")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.10.0")
 addSbtPlugin("com.github.reibitto" % "sbt-welcome" % "0.2.1")
+addSbtPlugin("com.sourcegraph" % "sbt-sourcegraph" % "0.2.0")
 
 libraryDependencies += "org.scala-sbt" %% "scripted-plugin" % sbtVersion.value
 


tgodzik (approved):
LGTM from me, might be useful to start using sourcegraph, however I have no idea if this will work as far as enabling it goes.

dos65:
@ckipp01 Cool!
I'm wondering if we can check if it works before merging.
Could we move this PR on the branch in the main repository and run this action against this branch?

ckipp01:
> @ckipp01 Cool!
> I'm wondering if we can check if it works before merging.
> Could we move this PR on the branch in the main repository and run this action against this branch?

Yea sure. I'll wait to maybe hear from @olafurpg to ensure I did this right and I'm not missing anything. Then I can move it to a metals branch and we can give it a go.

olafurpg (approved):
LGTM 👍 The changes look good. Normally, you wouldn't need `set every semanticdbEnabled := true` if it's already enabled in the build but Metals uses `addCompilerPlugin(semanticdb-scalac)`, which isn't understood by sbt-sourcegraph yet.

FWIW, we have two exciting upcoming improvements that should make this integration even more worthwhile:

- Soon, the sources of library dependencies will be automatically navigateable on Sourcegraph. Currently, the semantic navigation is limited to the repo itself. I'm actively working right now on the part to auto-index Scala libraries (it's already working for Java libraries).
- Soon, the hover messages will render method/class signatures using Scala syntax instead of Java syntax.

ckipp01:
One more question actually. Do you think it's worth enabling this for both prs and main? @olafurpg I'm assuming if we do that they are kept separately?  Then you'd be able to benefit from it as well while reviewing prs?

Strum355:
> Do you think it's worth enabling this for both prs and main?

@ckipp01 personally, this would be the ideal way to go. When reviewing PRs, youd want semantic results for the parts that have changed, which you wouldnt get if LSIF uploads dont exist for the commits in the PR :slightly_smiling_face:

olafurpg:
> Do you think it's worth enabling this for both prs and main? @olafurpg I'm assuming if we do that they are kept separately? Then you'd be able to benefit from it as well while reviewing prs?

@ckipp01 I think it's worth enabling uploads for PRs as well. If it turns out to be too slow then we can look into how to optimize the upload step. The current PR adds a separate CI job for the LSIF upload but it's possible to reuse the existing Scalafix check job (which already compiles with SemanticDB) so that you do only minimal additional work (upload a few megabytes) to enable precise intel on Sourcegraph.

build.sbt
```diff
@@ -265,12 +268,7 @@ lazy val V = new {
 
 val sharedSettings = List(
   libraryDependencies ++= crossSetting(
-    scalaVersion.value,
-    if2 = List(
-      compilerPlugin(
-        "org.scalameta" % "semanticdb-scalac" % V.scalameta cross CrossVersion.full
-      )
-    )
```
ckipp01, line 273: So do we have to include this like this, or can we do like I did up above and instead set `semanticdbEnabled := true` for everything? We are still able to set the desired `semanticdbVersion` in that way. If that's ok it sort of makes adding sourcegraph in here much easier since if not we'll need to add some extra settings since the sourcegraph upload will fail if `semanticdbEnabled` isn't set to true, which is isn't for us. Also, even if it was it will look for the semanticdb where `semanticdbTargetRoot` is set to, but the way we have this setup, we are always putting the semanticdb in the class directory, where Scalafix defaults to. This change (and the stuff up in `inThisBuild`) will instead let sbt handle the enabled, we set the correct version, and then set `semanticdbIncludeInJar := false`. Thoughts on this?

ckipp01:
> The current PR adds a separate CI job for the LSIF upload but it's possible to reuse the existing Scalafix check job (which already compiles with SemanticDB) so that you do only minimal additional work (upload a few megabytes) to enable precise intel on Sourcegraph.

I don't think we're able to easily re-use the scalafix check job since it's part of a matrix. So we need to add in Go and setup the sourcegraph binary, which isn't easy to do in the matrix without some pretty large changes. So I'd rather just keep that clean the way it is and add in a seperate job for this.

build.sbt
```diff
@@ -150,7 +150,10 @@ inThisBuild(
         .map { configKey => s"-D$configKey=${props.getProperty(configKey)}" }
         .mkString(" ")
     },
-    resolvers += Resolver.bintrayRepo("scalacenter", "releases")
+    resolvers += Resolver.bintrayRepo("scalacenter", "releases"),
+    semanticdbEnabled := true,
```
tgodzik, line 154: Weirdly enough tests started to fail due to a missing semanticdb :thinking: Might it be due to target root not set/read properly?

build.sbt
tgodzik: crossSetting(scalaVersion.value) doesn't seem to do anything now.

ckipp01: 🤔 that's odd. I'm not sure. Let me dig in more.

ckipp01: Ah yea true. I'll fix this but still can't get it to work. I'll need to dig in some more.

ckipp01: Scratch this, this won't work this way since we need this for the properties files that are created in tests...

ckipp01:
Ok so I'm fairly convinced that the best way to do this is to completely keep it separate from our build definition. If not, it will become really confusing and we'll have a bunch of settings only relevant to sourcegraph in our build definition. I tried a whole bunch of things, but due to the way we include the semanticdb plugin we'd need to add some extra things in order to make this work that I just don't think are worth it. We'd basically need to ensure `sourcegraphSemanticdbDirectories` is set correctly for every module and due to the very different types of modules we have, in sbt this isn't straight forward at all, unless I'm totally missing something obvious. Even _if_ we set that when running `sourcegraphLsif` I still get errors about `semanticdbEnabled` not being true. So to just avoid all of this, even though we're redoing work in CI to make this work, I'd suggest we just keep it completely separate.

PR #3006, opened 2021-07-30
Alternative version of #2999: Add more tests on using sbt as the build server

adpi2:
Alternative version of #2999 where `BuildServerInitializer` and `BuildToolLayout` are passed as parameters rather than being mixed as traits.

@tgodzik @ckipp01 do you think this one is better than the previous one?


tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
```diff
@@ -33,23 +32,24 @@ class SbtServerSuite
...
       _ = assertNoDiff(
         client.workspaceMessageRequests,
         List(
-          importBuildMessage,
-          progressMessage
+          importBuildMessage
         ).mkString("\n")
       )
       _ = client.messageRequests.clear()
-      _ = assertStatus(_.isInstalled)
```
tgodzik, line 52: Why remove this check?

tests/unit/src/main/scala/tests/BuildServerInitializer.scala
```diff
@@ -0,0 +1,135 @@
...
+      _ = client.selectBspServer = { _ => new MessageActionItem("sbt") }
+      _ <- server.executeCommand(ServerCommands.BspSwitch.id)
+    } yield {
+      if (!expectError) {
+        server.assertBuildServerConnection()
+      }
+    }
+  }
+
+  private def generateBspConig(workspace: AbsolutePath): Unit = {
```
tgodzik, line 106:
```suggestion
  private def generateBspConfig(workspace: AbsolutePath): Unit = {
```

tests/unit/src/main/scala/tests/BuildServerInitializer.scala
```diff
@@ -0,0 +1,135 @@
...
+    val bspFolder = workspace.resolve(".bsp")
+    val sbtJson = bspFolder.resolve("sbt.json")
+    // don't overwrite existing BSP config
+    if (!sbtJson.isFile) {
+      // we create bsp/sbt.json file manually because `sbt bspConfig` takes too long
+      val sbtLaunchJar =
```
tgodzik, line 112: Are you able to just use `SbtBuildTool` and run it?

tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
```diff
@@ -33,23 +32,24 @@ class SbtServerSuite
...
-            |scalaVersion := "${V.scala212}"
-            |""".stripMargin
-      )
+      _ <- server.initialize()
+      _ <- server.initialized()
       _ = assertNoDiff(
         client.workspaceMessageRequests,
         List(
-          importBuildMessage,
-          progressMessage
```
tgodzik, line 48: Why did the progressMessage disappear ? It's better, I am just asking out of curiosity.

tgodzik (approved):
LGTM! I left just some minor comments. Thanks a lot for the follow up!

tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
adpi2: In `TestingClient` the default response of the `importBuildMessage` is:
```scala
var importBuild: MessageActionItem = ImportBuild.notNow
```

So Metals does not run `bloopInstall` and that's why there is no `progressMessage` anymore. Instead it switches to sbt as the build server.

tests/unit/src/main/scala/tests/BuildServerInitializer.scala
adpi2: That's tested in `SbtServerSuite.generate`. I am not using `SbtBuildTool` here to save some time by not running `sbt bspConfig`.

tests/unit/src/main/scala/tests/BuildServerInitializer.scala
adpi2: :+1:

tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
adpi2: All tests used to run `bloopInstall` but that takes too long and it is not always useful. Here we want to test sbt server as the build server so we don't need to run `bloopInstall`, it is already tested in `SbtBloopLspSuite`.

PR #3016, opened 2021-08-02
Just to trigger CI

ckipp01:


ckipp01:
🤔 so sending in a pr to this branch seems to not trigger the lsif job like I thought it would. Does sending in a pr to a non-default branch still trigger the default CI rather than the CI tied to that branch?

tgodzik:
Is it correctly formatted though? `env:` seems off in the last PR. Essentially if you add a new job for pull_requests it should run in that PR unless the formatting/logic is broken.

ckipp01:
> Is it correctly formatted though? `env:` seems off in the last PR. Essentially if you add a new job for pull_requests it should run in that PR unless the formatting/logic is broken.

Normally it will fail if the formatting in the yaml is wrong though no?

tgodzik:
I think YAML is not an exact science :D

ckipp01:
Ah, actually I think you're right.

tgodzik:
Yeah, it's most likely causing the yaml file to be ignored, I also thought it might be due to it not being in main, but that was actually never the case. Which I think makes sense, since you would want to test out any changes.

ckipp01:
Yup yup, just changed the indentation and now it's working. Freaking yaml.

PR #3017, opened 2021-08-02
Merge sourcegraph branch into main.

ckipp01:
I tried sending in a pr against the sourcegraph branch to try to get the LSIF job to run, but it doesn't. I'm pretty sure this needs to actually be merged into main in order to pick it up.

### DIFF
--- a/.github/workflows/sourcegraph.yml
+++ b/.github/workflows/sourcegraph.yml
@@ -1,0 +1,28 @@
+name: Sourcegraph
+on:
+  push:
+    branches:
+      - main
+    pull_request:
+
+jobs:
+  lsif:
+    runs-on: ubuntu-latest
+    name: "Upload LSIF"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: olafurpg/setup-scala@v12
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "1.15.6"
+      - run: |
+          mkdir -p bin
+          curl -L https://sourcegraph.com/.api/src-cli/src_linux_amd64 -o bin/src
+          chmod +x bin/src
+          export PATH="$PATH:$PWD/bin"
+          sbt \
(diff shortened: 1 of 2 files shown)


ckipp01:
\o/ well at least the upload _seems_ to have worked.

ckipp01:
> Now, how do you actually use sourcegraph? :D

lol that's what I'm currently trying to figure out. I was _sort_ of expecting to see some info here about he uploads:

Screenshot 2021-08-02 at 13 57 23

But as you can see it says only a site admin can. So I'm unsure if maybe we just can't see it, or if there is none. @olafurpg at this stage, how can one verify that the upload did anything? For example looking at source in this branch still shows that it's search-based. Is there a time period that it takes to sync? Or is there a way to verify on the sourcegraph side that an upload worked as expected?

olafurpg:
The upload succeeded but I'm not able to see that the navigation is precise. 

image

I'll look into this!

olafurpg:
I'm not sure why it wasn't for the `sourcegraph` branch but precise navigation started working on the main branch after the merge. You can try it here https://sourcegraph.com/github.com/scalameta/metals@d10c66d88db9a1ef92927bce44e97baf9ae2c7e1/-/blob/tests/mtest/src/main/scala/tests/BaseSuite.scala?L38:16#tab=references

image

Navigation to the JDK is working correctly but not library dependencies (yet). Once I've added a small fix to sbt-sourcegraph and we merge the PR https://github.com/sourcegraph/sourcegraph/pull/22984 then navigatin should work for library dependencies as well :)

ckipp01:
> Navigation to the JDK is working correctly but not library dependencies (yet). Once I've added a small fix to sbt-sourcegraph and we merge the PR [sourcegraph/sourcegraph#22984](https://github.com/sourcegraph/sourcegraph/pull/22984) then navigatin should work for library dependencies as well :)

Whoa, this is super cool!

PR #3018, opened 2021-08-03
Make the SemanticDB compiler options used by mtags configurable.

olafurpg:
Previously, mtags used a hardcoded list of SemanticDB compiler options
to produce SemanticDB text documents from source. This commit makes the
options configurable via PresentationCompilerConfig. The motivation for
this change is that we recently started using mtags for LSIF indexing
Scala projects. To render Scala method/class signatures we need SemanticDB
`SymbolInformation`, which are not emitted by default in mtags since
they are not used by Metals.

### DIFF
--- a/mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
+++ b/mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
@@ -1,6 +1,8 @@
 package scala.meta.pc;
 
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


tgodzik:
Actually, that explains why we weren't getting type decorations in sbt build files :O 

We might want to add it as the default instead, what do you think @olafurpg  ?

tgodzik (approved):
Anyway, this looks good! I might follow up with the change to generate symbol information.

olafurpg:
Symbols were disable by default for performance reasons but I doubt the overhead is noticeable for users so it’s probably a good idea to enable symbols by default.

olafurpg:
This PR is yet another anecdote supporting the philosophy of not doing premature optimizations.

PR #3019, opened 2021-08-03
Generate symbol information for standalone files, worksheets and sbt files

tgodzik:
Previously, generated semanticdb would not contain symbol information. Now, it's enabled and allows for type decorations and implementations in worksheets, standalone files, dependencies and sbt files.

Should help with https://github.com/scalameta/metals-feature-requests/issues/190

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
@@ -83,7 +83,7 @@ final class Diagnostics(
       case Some(diagnostic) if !workspace.exists(path.isInReadonlyDirectory) =>
         syntaxError(path) = diagnostic
         publishDiagnostics(path)
-      case None =>
+      case _ =>
         onNoSyntaxError(path)
     }
   }
--- a/metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
@@ -87,8 +87,7 @@ final class InteractiveSemanticdbs(
           if (existingDoc == null || existingDoc.md5 != sha) {
             val compiled = compile(path, text)
             Option(compiled).foreach(doc =>
-              // don't index dependency source files or worksheets, since their definitions are local
-              if (!source.isDependencySource(workspace) && !source.isWorksheet)
+              if (!source.isDependencySource(workspace))
                 semanticdbIndexer().onChange(source, doc)
             )
             compiled
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -992,6 +992,7 @@ class MetalsLanguageServer(
     val loadInteractive = Future {
       interactiveSemanticdbs.textDocument(path)
     }
+    parseTrees(path).flatMap(_ => syntheticsDecorator.publishSynthetics(path))
     if (path.isDependencySource(workspace)) {
       CancelTokens { _ =>
         // publish diagnostics
@@ -1008,9 +1009,6 @@ class MetalsLanguageServer(
         .sequence(
           List(
             loadInteractive,
-            parseTrees(path).flatMap(_ =>
-              syntheticsDecorator.publishSynthetics(path)
-            ),
             loadFuture,
             compileFuture
           )
@@ -1050,6 +1048,7 @@ class MetalsLanguageServer(
         // Don't trigger compilation on didFocus events under cascade compilation
         // because save events already trigger compile in inverse dependencies.
         if (path.isDependencySource(workspace)) {
+          syntheticsDecorator.publishSynthetics(path)
           CompletableFuture.completedFuture(DidFocusResult.NoBuildTarget)
         } else if (openedFiles.isRecentlyActive(path)) {
           CompletableFuture.completedFuture(DidFocusResult.RecentlyActive)
--- a/mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
+++ b/mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompilerConfig.java
@@ -110,7 +110,6 @@ public interface PresentationCompilerConfig {
     static List<String> defaultSemanticdbCompilerOptions() {
         return Arrays.asList(
             "-P:semanticdb:synthetics:on",
-            "-P:semanticdb:symbols:none",
             "-P:semanticdb:text:on"
         );
     }
--- a/tests/cross/src/test/scala/tests/pc/PcSemanticdbSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/PcSemanticdbSuite.scala
@@ -39,11 +39,12 @@ class PcSemanticdbSuite extends BasePCSuite {
        |  val a = 123
        |  val b = a + 1
        |}""".stripMargin,
-    """|import $ivy.`org.kohsuke:github-api:1.114`
+    // local0 comes most likely from the script object use to wrap ScriptSource
+    """|/*local0*/import $ivy.`org.kohsuke:github-api:1.114`
        |
-       |object O/*local0*/ {
-       |  val a/*local1*/ = 123
-       |  val b/*local2*/ = a/*local1*/ +/*scala.Int#`+`(+4).*/ 1
+       |object O/*local1*/ {
+       |  val a/*local2*/ = 123
+       |  val b/*local3*/ = a/*local2*/ +/*scala.Int#`+`(+4).*/ 1
        |}
        |""".stripMargin,
     filename = "A.worksheet.sc",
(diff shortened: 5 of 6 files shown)


ckipp01:
Seems to be affecting the references suite with some valid failures 🤔

tgodzik:
> Seems to be affecting the references suite with some valid failures

yeah, it seems that each fix breaks another things. And I thought it would be an easy fix :sweat_smile:

tanishiking:
Oh, when I was using a worksheet, I found some IDE features didn't work, and this's why! I really appreciate this change 😄 

---

So I dug the problem:

- With `filename = if (uriString.endsWith(".sc")) uriString + ".semanticdb" else uriString,`, `trait AA` in the worksheet has global symbol `empty/AA#`, and without the change we have local symbol for them.
  - I guess this is because of https://github.com/scalameta/metals/blob/fe7e96fe67e765be279ad1ce61bc5eb4acac4cb8/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala#L490-L493 but **I'm not sure why `ScriptSource` generates only `local` symbols** 🤔 
- When we invoke `reference`, `ReferenceProvider` calculates the result based on
  - snapshot semanticdb (generated by `AggregatedSemanticDBs.textDocument`) for **local symbols**
    - https://github.com/scalameta/metals/blob/0ca74b38ee6bdd203cadbb1bd797028d845e101d/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala#L189-L198
  - Search for the symbol from bloom filter for **global symbol**, which are indexed by `ReferenceProvider.onChange`.
    - https://github.com/scalameta/metals/blob/0ca74b38ee6bdd203cadbb1bd797028d845e101d/metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala#L214-L217
- However, the worksheet's symbols are never indexed to the bloom filter
  - because https://github.com/scalameta/metals/blob/0ca74b38ee6bdd203cadbb1bd797028d845e101d/metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala#L82-L84
  - **That's why `reference` stopped working with this change** (the symbols in a worksheet have global symbols but they are never indexed to the bloom filter).

So I think we can fix the problem by

```diff
diff --git a/metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala b/metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
index d641e6eb..d18fb852 100644
--- a/metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
@@ -79,8 +79,8 @@ final class InteractiveSemanticdbs(
           if (existingDoc == null || existingDoc.md5 != sha) {
             val compiled = compile(path, text)
             Option(compiled).foreach(doc =>
-              // don't index dependency source files or worksheets, since their definitions are local
-              if (!source.isDependencySource(workspace) && !source.isWorksheet)
+              // don't index dependency source files since their definitions are local
+              if (!source.isDependencySource(workspace))
                 semanticdbIndexer().onChange(source, doc)
             )
             compiled
```

---

Also, I propose the following diff, because

- Renaming the worksheet file to `worksheet.sc.semanticdb` (even though it's not a semanticdb file) seems wired for me.
- We can remove `filename.isScalaScript` from `addCompilationUnit` instead, and it works same.

```diff
diff --git a/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala b/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
index dcfb5983..36518111 100644
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
@@ -488,7 +488,7 @@ class MetalsGlobal(
     val unit = newCompilationUnit(codeWithCursor, filename)

     val source =
-      if (filename.isScalaScript || filename.isSbt)
+      if (filename.isSbt)
         ScriptSourceFile(unit.source.file, unit.source.content)
       else unit.source
     val richUnit = new RichCompilationUnit(source)
diff --git a/mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala b/mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
index f1625081..4dded88d 100644
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
@@ -25,9 +25,7 @@ class SemanticdbTextDocumentProvider(
     val uriString = uri.toString()
     val unit = addCompilationUnit(
       code = validCode,
-      // it seems that otherwise we use a cached version with magic imports still in place
-      filename =
-        if (uriString.endsWith(".sc")) uriString + ".semanticdb" else uriString,
+      filename = uriString,
       cursor = None
     )
     typeCheck(unit)
```

tgodzik:
Thanks @tanishiking ! I need to get back to this one, I will check out what you suggest and see if we don't regress too much with the performance. I remember some issues that I had, but I think it was not related to the performance in the end.

metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
```diff
@@ -87,8 +87,7 @@ final class InteractiveSemanticdbs(
           if (existingDoc == null || existingDoc.md5 != sha) {
             val compiled = compile(path, text)
             Option(compiled).foreach(doc =>
-              // don't index dependency source files or worksheets, since their definitions are local
-              if (!source.isDependencySource(workspace) && !source.isWorksheet)
+              if (!source.isDependencySource(workspace))
```
tgodzik, line 90: I removed the condition to start indexing worksheet files, since they can contain symbols from the workspace.

metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
```diff
@@ -83,7 +83,7 @@ final class Diagnostics(
       case Some(diagnostic) if !workspace.exists(path.isInReadonlyDirectory) =>
         syntaxError(path) = diagnostic
         publishDiagnostics(path)
-      case None =>
+      case _ =>
```
tgodzik, line 86: This could actually fail if we modified a dependency file and it had an error.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -992,6 +992,7 @@ class MetalsLanguageServer(
     val loadInteractive = Future {
       interactiveSemanticdbs.textDocument(path)
     }
+    parseTrees(path).flatMap(_ => syntheticsDecorator.publishSynthetics(path))
```
tgodzik, line 995: We want to show synthetics in dependencies now.

tgodzik (commented):
@tanishiking thanks for the hints!

I actually kept the ScriptSource, since it is needed for the presentation compiler to work correctly with toplevel statements, but it seems there is no need to change it. The `local0` must be coming from the wrapping object in the compiler itself.

tgodzik:
@ckipp01 @dos65 This should be good to review now, it took much longer than expected for me to get back to it.

metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
ckipp01: Just to make sure I'm following. If you now are using a defined type _in_ a worksheet and search for references of that type, this would mean that the result from the worksheet would also appear right?

ckipp01 (commented):
Maybe i'm missing it, but with the changes here, this will only change the behavior for worksheets, not sbt files right?

metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
tgodzik: Yes exactly, though for Scala 2 everything defined in the worksheet is local, so in reality these will not pop up in references.
However, this change causes the refreences to workspace symbols, that are used in worksheets, to show up, which I think is super useful.

tgodzik:
> Maybe i'm missing it, but with the changes here, this will only change the behavior for worksheets, not sbt files right?

We were never filtering out sbt files, so we are good here.

ckipp01:
> We were never filtering out sbt files, so we are good here.

Ahh ok, that's my mistake. The PR title make it sound like this was changing the behavior to now generate symbol information for sbt files, when I was under the impression that we indeed did do that before

ckipp01 (approved):
LGTM!

tanishiking (approved):
LGTM too! 👍

tgodzik:
> > We were never filtering out sbt files, so we are good here.
> 
> Ahh ok, that's my mistake. The PR title make it sound like this was changing the behavior to now generate symbol information for sbt files, when I was under the impression that we indeed did do that before

So we did generate occurences before and that was responsible for refrences and renames, now we also are able to print type information.

Issue #2961, opened 2021-07-14
FileWatcher not working for a large project on macOS Big Sur


pvid:
**Describe the bug**

Recently I started to work on a large SBT project (metals reports 125 build targets + several hundred library dependencies) with a  newly set up Intel MacBook Pro (running Big Sur).

I noticed "rename" and "find all references" did not work most of the time. The strange thing is that sometimes they worked, sometimes they worked for only some projects, the next time they worked for a different subset of projects. It was all a bit non-deterministic.

I have done some debugging and the issue seems to be that `FileWatcher` does not report events for some or most of the projects, which causes the semanticdb files to not be indexed by the `ReferenceProvider`. There is nothing useful in the `metals.log`. I even tried to turn on the `swoval` debug logging and I did not see any errors. However, no events updates were triggered when recompiling subprojects where "find all references" did not work.

I tried to look into the macOS settings and tried to increase the `maxfiles` setting, which was set relatively low. That seems to help a bit, but the problem still occurs. Unfortunately, I don't know much about how file watching on macOS works. Maybe @eatkins of swoval fame could offer some insight - is my computer just terribly misconfigured? Could there be an issue in how swoval and macOS APIs interact when watching a large number of directories?

**To Reproduce**

That is going to be quite hard... I don't know of a open source project with so many subprojects and the behaviour is non-deterministic.

**Expected behavior**

File watching to work reliably or report a clear error in the logs.

**Installation:**

- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.10.4, v0.10.5

**Search terms**

macos, big sur, file watcher

eatkins:
This kind of thing is very difficult to debug. In the past, I've seen similar behavior when there are too many [file system event streams](https://developer.apple.com/documentation/coreservices/1443980-fseventstreamcreate) created on the machine. There are some libraries that try to interact with the osx file system watcher as though it were inotify and create a stream for every directory in the file system hierarchy. Since inotify handles are _not_ recursive, libraries might naively try to register an event stream for each directory in the file system hierarchy which is what you are required to do on Linux. Unfortunately, this works very badly with OSX and you quickly exhaust what appears to be a system wide limit on file event streams. The swoval library tries to be smart and whenever the application registers a new directory for watching, it checks if the new directory is already in the hierarchy of a directory under monitoring or if there are previously registered directories that are beneath the new directory in the hierarchy. In either case, it consolidates the streams into a single string and closes the redundant streams.

[...]

PR #3021, opened 2021-08-03
Fix file watcher in large projects on macOS

pvid:
Fixes https://github.com/scalameta/metals/issues/2961

I also did some maintenance:
- removed copy-pasted remnants of directory-watcher. Murmu3 hash is actually a part of the scala standard library and I recreated the rest of the functionality in scala
- removed the OVERFLOW event and related callbacks. That was dead code since the refactor that introduced swoval
- removed creation of empty directories. This has been talked about in https://github.com/scalameta/metals/pull/2014#issuecomment-678560544 but did not land in master

I have been daily driving this version for the past 1.5 weeks on a large project (75k+ files) without a hitch. One optimization that I needed to introduce was to hash hash only files which are actually used for indexing. That significantly improved file watcher startup time when watching the whole project root directory (from ~3 minutes to ~2 seconds).

I did not have a change to test the changes on a linux or a windows machine though.


pvid:
the `isCoveredBy` method is a bit rushed and contains a bug. I'll update the PR with some tests for it and a correct implementation

tgodzik:
@pvid looks like the tests are all passing, should we mark it ready to review?

pvid:
@tgodzik I just added some last touches and rebased the PR. I also left a TODO comment concerning file watching and creation of empty directories and an issue that I encountered and could not resolve. However, it does not have to block this PR - I just replicate the creation of directories of the old implementation.

@eatkins, any ideas about https://github.com/scalameta/metals/pull/3021/files#diff-314e34520d936536cd836f653612edc262cf2afb106ae5072025978256a1fa8cR144? I tried to replicate the issue I saw as a swoval test case, but did not succeed 😅 I'll probably try it again in a few days with fresh eyes.

BTW, I tried to use the same macOS approach (watching only the project root directory and filter events) on linux too, but it can easily hit inotify limits in larger projects.

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
```diff
@@ -0,0 +1,222 @@
...
+        callback
+      )
+      repo.register(workspace, Int.MaxValue)
+      () => repo.close()
+    } else {
+      // Other OSes register all the files and directories individually
+      val repo = initFileTreeRepository(watchFilter, callback)
+
+      // TODO(@pvid) swoval's FileTreeRepository should be able to create watch
+      // for files/directories that do not exist yet. However, there is an issue
```
eatkins, line 145: If I'm reading this correctly, you're saying that on linux or windows that the semanticdb file changes are not detected if the swoval watch is created before the semantic db file parent directories exist?

I'm not sure what is going on since there are test cases in swoval for the use case of watching a file when its parent doesn't exist, but I believe you that it is not working.

eatkins: I scanned the PR at a high level and it does look pretty good to me overall.

pvid: unfortunately no - semanticdb file changes are not detected if the _source_ directories are not created before registering the watches. It makes no sense to me. 

There might be some subtle bug when registering multiple non-existing files, but it eludes me - I did not manage to come up with a failing swoval test case

eatkins: Huh, that is bizarre. But unfortunately it has definitely been my experience that some file watching bugs are very difficult to reproduce (which your inability to create a test case sadly confirms).

pvid:
@tgodzik the PR is ready for review :) to resolve the issue described in the TODO comment is a fight for another day

tgodzik (approved):
LGTM! Thanks a lot for making this happen! This should be really helpful for anyone on MacOS

PR #3022, opened 2021-08-03
Upgrade to the latest sbt-sourcegraph

olafurpg:
This version includes a fix so that cross-repo navigation should work
after https://github.com/sourcegraph/sourcegraph/pull/22984 is merged.


ckipp01 (approved):
Cool, thanks Olaf!

PR #3003, opened 2021-07-29
Update Bloop to the newest nightly

tgodzik:

PR #3028, opened 2021-08-06
Run tests on JDK 17

tgodzik:


tgodzik:
Blocked on https://github.com/scalameta/metals/pull/3003

.github/workflows/ci.yml
```diff
@@ -12,22 +12,21 @@ jobs:
...
             shard: 1
-          - os: windows-latest
-            java: 8
+          - os: ubuntu-latest
+            java: 'adopt@1.8'
             shard: 2
-          - os: macOS-latest
-            java: 8
+          - os: ubuntu-latest
+            java: 'zulu@1.17.0-0'
```
sideeffffect, line 25: It's probably better to just use `17` here instead of the `zulu...`

tgodzik:
Finally the JDK 17 tests seems to be passing :tada:

ckipp01 (approved):
This is great that this is now working. We'll have to remember to update the requirements in the docs when we release to reflect this:  https://scalameta.org/metals/docs/editors/vscode#requirements

tgodzik:
> This is great that this is now working. We'll have to remember to update the requirements in the docs when we release to reflect this: https://scalameta.org/metals/docs/editors/vscode#requirements

Updated the docs!

PR #3032, opened 2021-08-10
[Scala3] Replace `InteractiveDriver` by our own wrapper to dotc.Compiler

dos65:
`InteractiveDriver` was designed for dotty-language-server and we use it
improperly. For Metals we don't need to keep a buch of files in memory.
Also it might fix some problems(memory leaks + some pc tests) that might be caused of having a single
context that is passed trough all received sources.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
@@ -29,27 +29,6 @@ object MtagsEnrichments extends CommonMtagsEnrichments {
       val p = Spans.Span(offset)
       new SourcePosition(source, p)
 
-    def localContext(params: OffsetParams): Context = {
-      if (driver.currentCtx.run.units.isEmpty)
-        throw new RuntimeException(
-          "No source files were passed to the Scala 3 presentation compiler"
-        )
-      val unit = driver.currentCtx.run.units.head
-      val tree = unit.tpdTree
-      val pos = driver.sourcePosition(params)
-      val path =
-        Interactive.pathTo(driver.openedTrees(params.uri), pos)(using
-          driver.currentCtx
-        )
-
-      val newctx = driver.currentCtx.fresh.setCompilationUnit(unit)
-      val tpdPath =
-        Interactive.pathTo(newctx.compilationUnit.tpdTree, pos.span)(using
-          newctx
-        )
-      MetalsInteractive.contextOfPath(tpdPath)(using newctx)
-    }
-
   extension (pos: SourcePosition)
     def toLSP: l.Range = {
       new l.Range(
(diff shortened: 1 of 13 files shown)


ckipp01:
> InteractiveDriver was designed for dotty-language-server and we use it
improperly

Can you explain a bit more of what you mean by this? From my understanding the `InteractiveDriver` is _specifically_ designed to be used by IDEs so it seems a bit fishy to me that we want to migrate away from it.

tgodzik:
Should we close this since the issue was fixed in another PR?

dos65:
@tgodzik Oh, I forgot about this one.

I'll look at it one more time.
I don't know why by this PR fixed [one test](https://github.com/scalameta/metals/pull/3032/files#diff-9ab9c5f7b86393976b3619ea86ddef73737390ca43e2a3d8ce27d66aa1f576e7R1081)

tgodzik:
> @tgodzik Oh, I forgot about this one.
> 
> I'll look at it one more time.
> I don't know why by this PR fixed [one test](https://github.com/scalameta/metals/pull/3032/files#diff-9ab9c5f7b86393976b3619ea86ddef73737390ca43e2a3d8ce27d66aa1f576e7R1081)

The ordering in those tests seems a bit random, so not sure if it's not just a coincidence.

smarter:
> From my understanding the InteractiveDriver is specifically designed to be used by IDEs

True, but it was designed with the usecase of the dotty language server in mind, if that's not general enough then we would accept PRs in the dotty repo to improve InteractiveDriver (you could also completely stop using it like this PR is proposing, but that ties metals a bit more to implementation details of the compiler).

ckipp01:
> > From my understanding the InteractiveDriver is specifically designed to be used by IDEs
> 
> True, but it was designed with the usecase of the dotty language server in mind, if that's not general enough then we would accept PRs in the dotty repo to improve InteractiveDriver (you could also completely stop using it like this PR is proposing, but that ties metals a bit more to implementation details of the compiler).

Yea that makes sense. I guess I really like the idea of having the `InteractiveDriver`. So if we do have specific issues with it, as you said, I really do think we should work on upstreaming them rather than moving away from it. I think that would help the ecosystem as a whole.

dos65:
@ckipp01 It depends on a case. I agree that approach fixing in upstream sounds more right and solid but it won't work at least for this initial period while we have a lot of unimplemented things. Upstream fixes takes more time and they aren't in priority and might not merged.
Also, in our codebase we already have a lot of test-cases that might spot some issues related to fix earlier than we found them after new compiler release.

For now, I would do fixes or backport them to compiler only in case if dotty codebase takes benefit from it. Like `Completions`, if there will be any fixes for them, it's definitely better to do in compiler because they are used in REPL.

ckipp01:
Sure that all makes sense. However I guess I still have a few clarification questions.

> we use it improperly

It's not clear to me how we are using it improperly. Could you clarify?

Also, what other issues is this pr fixing. Like what are the identified issues that we have with the `InteractiveDriver` that we are aiming to fix with this. Is it memory usage? A bug with completions? Something else?

dos65:
@ckipp01 I'll close this pr at least for while. It was just an attempt to fix memory issue. I thought maybe removing not used source trees and do some changes around Context will help with memory usage

ckipp01:
> @ckipp01 I'll close this pr at least for while. It was just an attempt to fix memory issue. I thought maybe removing not used source trees and do some changes around Context will help with memory usage

👍🏼 ha, I didn't mean to have you close it. We can totally go this route, I was just trying to understand what the exact goal with it was 😆

Issue #2956, opened 2021-07-13
Metals locks up and becomes unresponsive with Scala 3


ckipp01:
**Describe the bug**

I've been working on a very small project in Scala 3 the last few nights and I keep hitting on Metals just completely freezing and locking up. CPU is locked at like 900%, VisualVM won't open for me to see what's going on, and even `jstack` wasn't working. I'm just completely unable to kill the process and have to restart my computer. I'm currently trying to figure out what's going on, but figured I'd throw what I have up here in case someone else spots it.

**To Reproduce**
It's a bit hard to reproduce as it doesn't happen _all the time_, but just sort of randomly for me. However tonight when it happened after failing a few times to get one I was finally able to get jstack to output something.

<details>
<summary>jstack dump</summary>
<br>

```
2021-07-13 22:02:16
Full thread dump OpenJDK 64-Bit Server VM (11.0.11+8-jvmci-21.1-b05 mixed mode, sharing):

Threads class SMR info:
_java_thread_list=0x00007ff113d69610, length=37, elements={
0x00007ff114009000, 0x00007ff12480a800, 0x00007ff124013000, 0x00007ff124013800,
0x00007ff12480b800, 0x00007ff114815000, 0x00007ff124014800, 0x00007ff11400e000,
0x00007ff12480f800, 0x00007ff114fd0000, 0x00007ff0d417d000, 0x00007ff114fe1000,
0x00007ff0c482b000, 0x00007ff0d41a3800, 0x00007ff0d41c5800, 0x00007ff0f56be000,
0x00007ff124ec8800, 0x00007ff124ec2000, 0x00007ff0f403b000, 0x00007ff0c400b800,
0x00007ff104279000, 0x00007ff0d41db000, 0x00007ff0d41d5800, 0x00007ff0f51ce000,
0x00007ff1140d4000, 0x00007ff124f54800, 0x00007ff124f3d000, 0x00007ff1140d5000,
0x00007ff0c4035000, 0x00007ff086115800, 0x00007ff0f53af000, 0x00007ff0f544c800,
0x00007ff124ea0800, 0x00007ff124cc6800, 0x00007ff08747b800, 0x00007ff0e4199000,
0x00007ff084f0d800
}

"main" #1 prio=5 os_prio=31 cpu=921.92ms elapsed=1803.39s allocated=60245K defined_classes=1871 tid=0x00007ff114009000 nid=0x2803 waiting on condition  [0x0000700008ff9000]
[...]

PR #3043, opened 2021-08-16
[Scala3] Fix memory usage spike on empty completions

dos65:
The previous incorrect `scopeSymbols` implementation was the main contributor in #2956 issue.
Calling an empty completion leaded to that all chain of `IndexedContext` were starting initialization of `lazy val scopeSymbols` where `scopeSymbols` has been storing a duplicated subsequences of symbols on several layers.

For example, one of the last outer context imports `java.lang._`(~900 entries).
It means that each inner context was keeping a list of 900+ elements: 
`scopeSymbols = ${current ctx symbols} ++ .. ++ ${java.lang._ symbols}` .
So a lot of space were used by `scala.collection.immutable::`

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/IndexedContext.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/IndexedContext.scala
@@ -15,6 +15,7 @@ import dotty.tools.dotc.typer.ImportInfo
 sealed trait IndexedContext {
   given ctx: Context
   def scopeSymbols: List[Symbol]
+  def names: IndexedContext.Names
   def rename(name: SimpleName): Option[String]
   def outer: IndexedContext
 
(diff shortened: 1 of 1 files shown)


ckipp01:
Oooooo this looks spot on @dos65. I was recently trying to dig into this more and kept seeing literally millions of `::`. So your conclusion totally looks correct.

https://cdn.discordapp.com/attachments/662615990043017236/875359189839536219/Screenshot_2021-08-12_at_14.42.09.png

dos65:
@ckipp01 I use [mat](https://www.eclipse.org/mat/) for analyzing heap-dumps.  Using it, you can look at `dominator tree` which shows the biggest objects. To find out where these objects located exactly you can call `Path to GC Roots`/ `Merge Shortests Paths to GCRoot` from context menu.

However, what about this issue - it's tricky. I had a lot of heap dumps after OOM but no one from them didn't had `IndexedContext` in heap. I guess, it's because it there is no strong reference on it, as this allocation happens only per operation and after catching OOM gc anyway might remove it from hep before producing heap dump.

I spot this issue by an accident by looking at memory chart in visualvm while I was investigating scala3 compiler leaks and was fast enough to perform heap dump.

tgodzik (approved):
This is awesome! Thanks for figuring this out!

Issue #1645, opened 2020-04-19
Go to parent code lens use too much CPU


dos65:
**Describe the bug**
Opening some (all?) files from scala/scala leads to hight CPU loading (up to 100%) for long periods (at least takes more than 2 minutes).

**To Reproduce**
1. Clone https://github.com/scala/scala and open `src/compiler/scala/tools/nsc/Global.scala` or `src/compiler/scala/tools/nsc/SubComponent.scala`
2. Check the CPU loading / notice the noise from the cooling system

**Installation:**
- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.8.4

**Additional context**
Not sure if it's close to problem but probably this might be related to  super/implementation references finding.
`jstack` dosn't show anything suspicious except thread with trace that contain `scala.meta.internal.implementation.SuperMethodProvider` / `scala.meta.internal.implementation.ImplementationProvider` 
<details><summary>Stacktrace sample</summary>
<pre><code>
"pool-1-thread-34" #67 prio=5 os_prio=0 cpu=83225,05ms elapsed=132,72s tid=0x00007f400c032800 nid=0x5bd2 runnable  [0x00007f3fb2bf7000]
   java.lang.Thread.State: RUNNABLE
	at sun.nio.fs.UnixNativeDispatcher.realpath0(java.base@11.0.5/Native Method)
	at sun.nio.fs.UnixNativeDispatcher.realpath(java.base@11.0.5/UnixNativeDispatcher.java:262)
	at sun.nio.fs.UnixPath.toRealPath(java.base@11.0.5/UnixPath.java:857)
	at scala.meta.internal.metals.MetalsEnrichments$XtensionAbsolutePathBuffers.dealias(MetalsEnrichments.scala:328)
	at scala.meta.internal.metals.MetalsEnrichments$XtensionString.toAbsolutePath(MetalsEnrichments.scala:419)
	at scala.meta.internal.metals.MetalsEnrichments$XtensionScalacOptions.$anonfun$targetroot$2(MetalsEnrichments.scala:510)
	at scala.meta.internal.metals.MetalsEnrichments$XtensionScalacOptions$$Lambda$583/0x00000008404a2440.apply(Unknown Source)
	at scala.Option.getOrElse(Option.scala:189)
	at scala.meta.internal.metals.MetalsEnrichments$XtensionScalacOptions.targetroot(MetalsEnrichments.scala:510)
[...]

PR #3045, opened 2021-08-17
Simplify super method lens using overriddenSymbols

tanishiking:
Potentially fix: https://github.com/scalameta/metals/issues/1645
This PR simplifies the implementation of the `super method lens` feature using `overriddenSymbols introduced in https://github.com/scalameta/scalameta/pull/2233
(`overriddenSymbols` in Scala3 is coming soon https://github.com/lampepfl/dotty/pull/13295 )


metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -1,295 +1,27 @@
...
-        methodSymbolInformation,
-        methodSignature,
-        documentWithPath,
-        cache,
-        findSymbol
-      )
-    } yield superSymbol
+    getSuperMethodHierarchy(
+      methodSymbolInformation
+    ).headOption
```
tanishiking, line 12: One concern is that the order of `overriddenSymbol` is semanticdb extractor implementation-dependent.
At least Scala2 (and Scala3 will) returns symbol in order of direct parent -> grand-parent ..., but to be robust to the change in Scala2 and Scala3, we may should check the direct super class of the `SymbolInformation` and pick the symbol from the parent class from `overriddenSymbol`.

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
```diff
@@ -1,295 +1,27 @@
...
-
-  final val stopSymbols: Set[String] = Set(
-    "scala/AnyRef#",
-    "scala/Serializable#",
-    "java/io/Serializable#",
-    "scala/AnyVal#",
-    "scala/Any#"
+  private final val stopSymbols: Set[String] = Set(
+    "scala/AnyRef#", "scala/Serializable#", "java/io/Serializable#",
+    "java/lang/Object#", "scala/AnyVal#", "scala/Any#"
```
tanishiking, line 25: Added `"java/lang/Object#"` because `toString` method of case classes overrides `java/lang/Object#toString()`

tanishiking:
Also, to confirm this PR fixes https://github.com/scalameta/metals/issues/1645 we need to check the profiler as https://github.com/scalameta/metals/issues/1645#issuecomment-618985927

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
dos65, line 12: A nitpick: it seems that it should be enough only to take a headOption and check that it doesn't belong to stopSymbols.
Smth like: `methodSymbolInformation.overriddenSymbols.headOption.filter(isNonStopSymbol)`

dos65 (changes requested):
Thanks for this fix!
LGTM overall, there is only one minor comment

metals/src/main/scala/scala/meta/internal/implementation/SuperMethodProvider.scala
tanishiking: Actually, this check is done inside of `getSuperMethodHierarchy` :)

dos65: Yep, I've seen that.
The only difference that it will be just slightly effective as there is no need to filter a full hierarchy chain in this case

tanishiking: Ah, I got it!
addressed in https://github.com/scalameta/metals/pull/3045/commits/0dbbabac20bd796c823abcdca597647299a4540e

tgodzik (approved):
LGTM! Glad to get rid of the complicated logic that I implemented :sweat_smile:

tgodzik:
FYI we can still simplify `MethodImplementation` to not use the asSeenFromMap, which is only needed for generics and since we can have an exact symbol now, there is no need for it. We can improve it later on.

dos65 (approved):
:+1:

PR #3046, opened 2021-08-17
Add documentation about GitHub Codespaces and GitHub.dev support

gabro:
This is in reference to https://github.com/scalameta/metals-feature-requests/issues/225

### DIFF
--- a/docs/editors/vscode.md
+++ b/docs/editors/vscode.md
@@ -10,7 +10,6 @@ title: Visual Studio Code
 
 ```
 
-
 ## Installation
 
 Install the Metals extension from the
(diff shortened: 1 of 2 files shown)


docs/editors/vscode.md
```diff
@@ -271,3 +270,14 @@ extension to use default IntelliJ shortcuts with VS Code.
 | Type info        | Show hover                |
 | Expand           | Fold                      |
 | Extend Selection | Expand selection          |
+
+## GitHub Codespaces and GitHub.dev support
+
+First, some definitions:
+
+- [GitHub Codespaces] are remote dev containers you can connect to using VS Code Desktop or VS Code Web.
```
dos65, line 278:
```suggestion
- [GitHub Codespaces](https://github.com/features/codespaces) are remote dev containers you can connect to using VS Code Desktop or VS Code Web.
```

dos65 (approved):
Thx for clarification! I noticed the conversation in the original feature request and this thing is really confusing.

ckipp01 (commented):
👍🏼 good to clarify this. One thing that comes to mind though is that we are talking about codespaces here, but then we also have an online IDEs section as well [here](https://scalameta.org/metals/docs/editors/online-ides#gitpod) where we talk about gitpod for example. Do you think we should try to centralize or link these?

docs/editors/vscode.md
gabro: woops, thanks :) Integrating it

gabro:
@ckipp01 that's a great idea! I've moved the documentation there and added a pointer from the vscode-specific one

docs/editors/vscode.md
```diff
@@ -271,3 +270,7 @@ extension to use default IntelliJ shortcuts with VS Code.
 | Type info        | Show hover                |
 | Expand           | Fold                      |
 | Extend Selection | Expand selection          |
+
+## GitHub Codespaces and GitHub.dev support
+
+See https://scalameta.org/metals/docs/editors/online-ides#github-codespaces-and-githubdev
```
gabro, line 276: the domain is hardcoded, since we copy-paste this file to the VS Code README during the release process

ckipp01 (approved):
Looks good! Thanks @gabro!

gabro:
Forcing the merge, since the CI failure is unrelated

PR #3049, opened 2021-08-17
Add tests explicitly to tests shard

tgodzik:
Previously, first shard would execute much longer than the second one because all new tests were run on it. Now, I added all the missing tests to project/TestGroups.scala and also added a warning that will inform a contributor that they should add it if possible.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -640,11 +640,18 @@ lazy val cross = project
   )
   .dependsOn(mtest, mtags)
 
-def isInTestShard(name: String) = {
+def isInTestShard(name: String, logger: Logger): Boolean = {
+  val groupIndex = TestGroups.testGroups.indexWhere(group => group(name))
+  if (groupIndex == -1) {
+    logger.warn(
+      s"""|Test is not contained in a shard: $name
+          |It will be executed by default in the first shard.
+          |Please add it to "project/TestGroups.scala". """.stripMargin
+    )
+  }
   if (!isCI) {
     true
   } else {
-    val groupIndex = TestGroups.testGroups.indexWhere(group => group(name))
     val groupId = Math.max(0, groupIndex) + 1
     System.getenv("TEST_SHARD").toInt == groupId
   }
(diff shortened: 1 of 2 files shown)


tgodzik:
The tests look much more evenly distributed now.

build.sbt
```diff
@@ -640,11 +640,18 @@ lazy val cross = project
   )
   .dependsOn(mtest, mtags)
 
-def isInTestShard(name: String) = {
+def isInTestShard(name: String, logger: Logger): Boolean = {
+  val groupIndex = TestGroups.testGroups.indexWhere(group => group(name))
+  if (groupIndex == -1) {
+    logger.warn(
```
ckipp01, line 646: 👍🏼 good idea to put a warning here.

project/TestGroups.scala
```diff
@@ -56,7 +61,27 @@ object TestGroups {
...
+      "tests.codeactions.InsertInferredTypeLspSuite",
+      "tests.FoldingRangeScala2Suite", "tests.SystemProcessSuite",
+      "tests.formatting.OnTypeFormattingSuite", "tests.DefinitionScala3Suite",
+      "tests.AnsiFilterSuite", "tests.DocumentSymbolScala3Suite",
+      "tests.ClassBreakpointSuite", "tests.FoldingRangeScala3Suite",
+      "tests.ScalafmtConfigSuite", "tests.DocumentSymbolScala2Suite",
+      "tests.StacktraceParseSuite", "tests.WorksheetDependencySourcesSuite",
+      "tests.DefinitionScala2Suite", "tests.AnalyzeStacktraceLspSuite",
+      "tests.codeactions.ExtractRenameMemberLspSuite",
+      "tests.debug.DotEnvFileParserSuite", "tests.SemanticdbScala3Suite",
```
ckipp01, line 83: 😆 totally didn't realize we were missing this many.

ckipp01 (approved):
Looks good! LGTM!

project/TestGroups.scala
tgodzik: It's been a while since I added it, so a lot of tests popped up :sweat_smile: 

It should be better with the warning.

project/TestGroups.scala
```diff
@@ -56,7 +61,27 @@ object TestGroups {
       "tests.MessagesSuite", "tests.pants.PantsGlobsSuite",
       "tests.SelectBspServerSuite", "tests.InverseDependenciesSuite",
       "tests.pantsbuild.PantsProjectNameSuite", "tests.TrigramSubstringsSuite",
```
dos65, line 63: It seems that some of tests were removed. 
Like: `tests.pantsbuild.PantsProjectNameSuite`

tgodzik: Yeah, I need try also a warning for tests that were removed, but not sure how to do it aside from manual.

It doesn't actually cause any issues though.

Issue #2120, opened 2020-10-02
Run organize imports on files that are not compiled


tgodzik:
Currently it's not possible to run organize imports if you don't first save the file and compile it. If a user doesn't do it a message is shown informing them to do so.

There are two approaches we could take:
- create semanticdb file bu using the presentation compiler and later saving it for scalafix to use
- handling the compilation in scalafix itself

Either way, it would improve the UI of organize imports.

@mlachkar is already experimenting with different approaches

PR #3055, opened 2021-08-20
Allow organize imports on unsaved code

tgodzik:
Previously, if a file was not saved it wouldn't be possible to run the organize imports rule.

Now, we try to produce a new semanticdb, save it to disk and point scalafix to it. It might be better to transfer this directly, but scalafix is currently lacking an API for it. We can change once it's supported.

Fixes https://github.com/scalameta/metals/issues/2120

metals-sample-1629736205938

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
@@ -14,6 +14,8 @@ import scala.util.Try
 
 import scala.meta._
 import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.mtags.SemanticdbClasspath
+import scala.meta.internal.semanticdb.TextDocuments
 import scala.meta.io.AbsolutePath
 
 import org.eclipse.lsp4j.MessageParams
@@ -33,7 +35,8 @@ case class ScalafixProvider(
     icons: Icons,
     languageClient: MetalsLanguageClient,
     buildTargets: BuildTargets,
-    buildClient: MetalsBuildClient
+    buildClient: MetalsBuildClient,
+    interactive: InteractiveSemanticdbs
 )(implicit ec: ExecutionContext) {
   import ScalafixProvider._
   private val scalafixCache = TrieMap.empty[ScalaBinaryVersion, Scalafix]
@@ -48,9 +51,10 @@ case class ScalafixProvider(
           case (_, targets) => targets.headOption
         }
         val tmp = AbsolutePath(Files.createTempFile("metals", ".scala"))
-        tmp.writeText("object Main{}\n")
+        val contents = "object Main{}\n"
+        tmp.writeText(contents)
         for (target <- targets)
-          scalafixEvaluate(tmp, target)
+          scalafixEvaluate(tmp, target, contents, produceSemanticdb = true)
 
         tmp.delete()
       } catch {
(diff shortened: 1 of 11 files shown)


metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -64,64 +67,87 @@ case class ScalafixProvider(
 
   def organizeImports(
       file: AbsolutePath,
-      scalaTarget: ScalaTarget
+      scalaTarget: ScalaTarget,
+      retried: Boolean = false
   ): Future[List[l.TextEdit]] = {
     val fromDisk = file.toInput
     val inBuffers = file.toInputFromBuffers(buffers)
-    if (isUnsaved(inBuffers.text, fromDisk.text)) {
```
tgodzik, line 71: This is mostly just moved, I recommend turning on the option to ignore whitespace.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -64,64 +67,87 @@ case class ScalafixProvider(
...
-              edits.getOrElse(Nil)
-            }
-        }
+          }
+        case Success(results) =>
+          Future.successful {
+            val edits = for {
+              fileEvaluation <- results.getFileEvaluations().headOption
+              patches <- fileEvaluation.previewPatches().asScala
+            } yield textEditsFrom(patches, fromDisk)
```
ckipp01, line 128: Won't this cause issues in the situation where the buffer doesn't match what was last saved on disk? For example if the user has an unsaved buffer and they trigger this, it will indeed now run the organize imports, but it also then applies those patches to the last saved file on disk resulting in reverting their code back. For example:



Notice how it removed the `val u = word.toUpperCase()` that I just added.

tgodzik: Ach, good catch, this should use the value from buffers. Thanks!

mlachkar (approved):
Nice!

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
```diff
@@ -59,6 +59,36 @@ class OrganizeImportsLspSuite
     scalacOptions = scalacOption
   )
 
+  check(
+    "basic-unsaved",
+    """
+      |package a
+      |import scala.concurrent.duration._
+      |import scala.concurrent.Future<<>>
+      |import scala.concurrent.ExecutionContext.global
```
dos65, line 68: @tgodzik Are you sure that this in this test scalafix is performed over semanticdb file from interactive?

I might be wrong, but I thought that diagnostic is always empty in files from InteractiveSemanticdb. So I think it shouldn't delete `import scala.concurrent.ExecutionContext.global` from imports.

tgodzik: yeah, some weird things are happening :/

tgodzik:
Actually, this will not work due to:

https://github.com/scalacenter/scalafix/blob/bf0df227c9b72fd6a9466f3a6fe11a8ce5f14efd/scalafix-cli/src/main/scala/scalafix/internal/v1/MainOps.scala#L224

tests/unit/src/main/scala/tests/ResourceOperations.scala
```diff
@@ -72,6 +74,12 @@ object ResourceOperations {
     val newPath = newUri.toAbsolutePath
     val fileExists = newPath.exists
 
+    val oldContents = buffers.get(oldPath)
+    oldContents.foreach { old =>
+      buffers.remove(oldPath)
+      buffers.put(newPath, old)
```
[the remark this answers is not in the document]
tgodzik: We are renaming the file, so we need to remove the old path and save under the new path.

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
tgodzik: Actually, you were right that it was using the wrong semanticdb, but the semanticdb produced in Metals does contain the right diagnostics if we specify them for the module.

I worked around the issue by creating the unsaved file in tmp so that it can be used by Scalafix. It's not super pretty, but until https://github.com/scalacenter/scalafix/issues/1236 is figured out this should make do. 

There is the mention of a symbol table that might be needed, but I think it's not necessary for organize imports itself.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
tgodzik: That should work correctly now, this was also hiding the issue that Scalafix actually reads the file, so this could not have worked at all. I worked around it by saving both semanticdb file and the source file contained in the buffers to tmp.

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
dos65: > Actually, you were right that it was using the wrong semanticdb, but the semanticdb produced in Metals does contain the right diagnostics if we specify them for the target.

That's fun. At least I'm glad that the my wrong assumption spotted an issue)
Could you also add a test for a file that has compile-error?

What about the whole approach how we interact with organize-import rule, I thought that at some moment it would be better to extract some part of it's API to be able to execute them without semanticdb file at all. It would allow to make [auto-imports be organized too](https://github.com/scalameta/metals-feature-requests/issues/221).

tgodzik: That might make sense, but it's a bigger effort I think.

dos65: >  but it's a bigger effort I think.

Yep, I agree. I mentioned it only to express that probably this issue in scalafix isn't required for this use-case with organize-imports.

tgodzik: > Could you also add a test for a file that has compile-error?

I added a test case for that and also added a check for error diagnostics. We will now fail if the semanticdb was not produced without errors.

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
```diff
@@ -59,6 +59,82 @@ class OrganizeImportsLspSuite
...
+      |  val d = Duration(10, MICROSECONDS)
+      |  val k = Future.successful(1)
+      |}
+      |""".stripMargin,
+    s"${SourceOrganizeImports.title}",
+    """|package a
+       |import scala.concurrent.duration._
+       |import scala.concurrent.Future
+       |import scala.concurrent.ExecutionContext.global
+       |import java.nio.file.ClassDoNotExist
```
dos65, line 116: It looks like organize wasn't performed. The ordering should be:
```scala
import java.nio.file.ClassDoNotExist

import scala.concurrent.ExecutionContext.global
import scala.concurrent.Future
import scala.concurrent.duration._
```

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -287,10 +354,14 @@ case class ScalafixProvider(
...
-    fromBuffers.linesIterator.zip(fromFile.linesIterator).exists {
-      case (line1, line2) => line1 != line2
-    }
+  private def isUnsaved(fromBuffers: String, fromFile: String): Boolean = {
+    // zipAll will extend the shorter collection, which is needed for accurate comparison
+    fromBuffers.linesIterator
+      .zipAll(fromFile.linesIterator, null, null)
+      .exists { case (line1, line2) =>
+        line1 != line2
+      }
```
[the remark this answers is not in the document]
tgodzik: That's what I thought! But it seems it breaks on Windows with `\n` and `\r\n`

tests/unit/src/test/scala/tests/codeactions/OrganizeImportsLspSuite.scala
tgodzik: That's right, it's because the file didn't compile and I added a check for that.

dos65: Oh, yep - found that.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -64,64 +68,92 @@ case class ScalafixProvider(
...
+      file: AbsolutePath,
+      contents: String
+  ) = {
+    interactive
+      .textDocument(file, Some(contents))
+      .documentIncludingStale
+      .flatMap { semanticdb =>
+        if (semanticdb.diagnostics.exists(_.severity.isError)) {
+          None
+        } else {
```
dos65, line 144: Here we can cleanup diagnostic in case of error before saving it to a file.
That would allow to perform imports reordering in this case.
Smth like:
```scala
val toSave = 
  if (semanticdb.diagnostics.exists(_.severity.isError))
    semanticdb.copy(diagnostic = Seq.empty)
  else
    semanticdb
```

tgodzik: I don't think we want to remove the diagnostics (also it does work with them), because if the file doesn't compile the results of OrganizeImports might be unexpected. Essentially, what would happen here is that `ClassDoNotExist` will be removed while it might have just been a typo.

dos65: The result is expected - without diagnostic it reorders imports only.
I've checked the suggested change locally and it works well.

Sometimes, when I add import manually and file has compile errors, I would like to perform organize imports right after that to put in in correct place.

dos65 (approved):
I'm really glad to see this change! 
LGTM!

PR #3058, opened 2021-08-21
Handle Scala 3 style blocks of code without braces

### LINKED ISSUE #3029: [Scala3] Missing folding range for indented Def.Def
If method body is wrapped in braces and long enough it has a folding range on the same line where it starts.
However, the same written in optional-braces syntax can't be folded.
Examples:
```scala
  // case 1 - with braces
>>>def foo(): Unit = {
     println()
     println()
     println()
   }<<<

  // case 2 - without braces - no folding ranges
  def foo(): Unit =
    println()
    println()
    println()
  
  // case 3 - larger body - folding range starts in body
  def foo(): Unit =
 >>>println()
    println()
    println()
    println()
    println()<<<
```

Expectations: `case 2` should have the same ranges as in `case 1`

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.10.5

tpasternak:
Blocks with braces are always interpreted as longer ones. For example a block:
```
def foo =
  ???
  ???
```
is two lines long, while the same one, but with braces:
```
def foo = {
  ???
  ???
}
```
is four lines long.

Apart from this, the first line of the second block is the one containing `def`, while in the first
one it's the one with `???`.

Due to this, the line folding markers were incorrectly placed.

Fixes #3029


tgodzik (approved):
LGTM!

PR #3060, opened 2021-08-21
Type annotation on code selection

### LINKED ISSUE #3059: Type annotation on code selection
It would be nice to be able to inspect the type of a code selection (not necessarily only a single word like it is now).

Something similar to what Intellij does when hitting Ctrl+Shift+p on selected code. For example, in a snippet like
```scala
final def flatMap[C](f: B => Read[A, C]): Read[A, C] = 
    Read.instance(a => read(a).flatMap(f(_).read(a)))
```
selecting in the editor the text
```scala
f(_).read(a)
```
it would be nice to be able to query for its type.

The only alternative I know about is using Intellij

**Search terms**:
snippet type annotation selection

KacperFKorban:
closes #3059
Peek 2021-08-21 21-36

We are limited by frontend capabilities, so to display the information properly we have to use the type annotation as hover.

Things that may be controversial:
- Hover for literals (Only for ranges now)
- Expression type for every hover (Only for ranges now)
- ~~key binding [here](https://github.com/scalameta/metals-vscode/pull/652)~~

vscode plugin changes: https://github.com/scalameta/metals-vscode/pull/652


tests/cross/src/test/scala/tests/hover/HoverLiteralSuite.scala
```diff
@@ -0,0 +1,74 @@
+package tests.hover
+
+import tests.pc.BaseHoverSuite
+
+class HoverLiteralSuite extends BaseHoverSuite {
+
+  check(
+    "literal-int",
```
tgodzik, line 8: @olafurpg Was there are reason for not adding hover over literals? This PR adds it but it can be reverted.

olafurpg: Literals were intentionally excluded because they are verbose. The hover messages take up a lot of visual real estate on the screen and the hypothesis was that it’s mentally easy to infer the type of literals.

ckipp01 (commented):
So this is really rad @KacperFKorban, I will use this a ton!

Using `nvim-metals`, this seems to just work as expected out of the box.

https://user-images.githubusercontent.com/13974112/130366198-c8c3b0e7-f424-4ef8-86b3-3dced5a9d802.mov

With that in mind and looking at the code fully recognising I may have missed something, it seems that the biggest thing we are still looking for is the offset coming off of the end of the range. In my video above, that's all that is being sent in, the position at the end of the visual selection. Do we actually need the full range? Won't the type at the end point _always_ be what we are after? If that's possible, then we don't actually need to break away from the LSP spec for `hover` by using our own class instead of the `TextDocumentPositionParams`.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1319,7 +1319,7 @@ class MetalsLanguageServer(
     }
 
   @JsonRequest("textDocument/hover")
-  def hover(params: TextDocumentPositionParams): CompletableFuture[Hover] =
+  def hover(params: HoverExtParams): CompletableFuture[Hover] = {
```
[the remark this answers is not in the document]
tgodzik: Weirdly enough this seems do be done by both Rust and Haskell LSP, currently there is no way to send in range, so we need to replace it ourselves. This is not perfect, but is compatible with the normal LSP spec.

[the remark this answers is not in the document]
ckipp01: Cool, didn't know they did this as well. Again, totally not against it, just think we should be careful whenever we break away from the spec.

metals/src/main/scala/scala/meta/internal/metals/HoverExtension.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import javax.annotation.Nullable
+
+import org.eclipse.lsp4j.Position
+import org.eclipse.lsp4j.Range
+import org.eclipse.lsp4j.TextDocumentIdentifier
+
+case class HoverExtParams(
+    textDocument: TextDocumentIdentifier,
+    @Nullable position: Position = null,
```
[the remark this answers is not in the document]
KacperFKorban: To have the guarantee of `Position`, we can change the `vscode`/`nvim` part to always send `Position`. So for a range request, the `Position` would be the `startPosition` of the `Range`.
This way only the range would be nullable with a small drawback of sometimes sending the same information twice.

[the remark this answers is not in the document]
ckipp01: It's probably not necessary to send both. This is just me being overly cautious 😆

KacperFKorban:
> So this is really rad @KacperFKorban, I will use this a ton!
> 
> Using `nvim-metals`, this seems to just work as expected out of the box.
> 
>  Screen.Recording.2021-08-22.at.15.56.00.mov 
> With that in mind and looking at the code fully recognising I may have missed something, it seems that the biggest thing we are still looking for is the offset coming off of the end of the range. In my video above, that's all that is being sent in, the position at the end of the visual selection. Do we actually need the full range? Won't the type at the end point _always_ be what we are after? If that's possible, then we don't actually need to break away from the LSP spec for `hover` by using our own class instead of the `TextDocumentPositionParams`.

Hmmm, can it be the case that nvim sends different position?

In vscode triggering the hover on the end of an expression doesn't work. Range on a term name seems to include the position before, but not the position after. Like here:
Peek 2021-08-22 20-56

Technically `endOffset` - 1 can work for ranges, but it can bring problems with position requests.

ckipp01:
> Hmmm, can it be the case that nvim sends different position?

> in vscode triggering the hover on the end of an expression doesn't work. Range on a term name seems to include the position before, but not the position after. Like here:

Gotcha. I think it's partly because when nvim sends in the position from the selection range, it's still just a single point that often isn't actually after the symbol, since unlike VS Code, you can't really extend past the text like in your image.

Either way, great work on this 👍🏼 I'll for sure play around with this a bit more to see if anything will even need to be added for this to work the same way in nvim-metals as it does in VS Code.

KacperFKorban:
@ckipp01 I thought of an example that won't work with just the end offset- method calls in infix notation.
e.g. for:
```scala
List("1", "2", "3").filter(<<_ contains "2">>)
```
the result will be Boolean.
and for:
```scala
List("1", "2", "3").filter(_ contains "2@@")
```
the result will be String.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -312,14 +312,25 @@ class Compilers(
...
+          .map(_.asScala.map { hover => adjust.adjustHoverResp(hover) })
+      }
+    } else {
+      val param: TextDocumentPositionParams =
+        new TextDocumentPositionParams(params.textDocument, params.position)
+      withPCAndAdjustLsp(param) { (pc, pos, adjust) =>
+        pc.hover(CompilerOffsetParams.fromPos(pos, token))
+          .asScala
+          .map(_.asScala.map { hover => adjust.adjustHoverResp(hover) })
+      }
```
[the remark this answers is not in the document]
KacperFKorban: There is another difference in `withPCAndAdjustLsp` methods- the second version creates `scala.meta.inputs.Position` from range. This is a bit harder to deduplicate, since `toMeta` requires local value- `input`.
Deduplicated the adjustment part, see if it's ok :smiley:

mtags/src/main/scala-2/scala/meta/internal/pc/HoverProvider.scala
```diff
@@ -175,8 +204,18 @@ class HoverProvider(val compiler: MetalsGlobal, params: OffsetParams) {
       range: Position
   ): Option[Hover] = {
     if (tpe == null || tpe.isErroneous || tpe == NoType) None
-    else if (symbol == null || symbol == NoSymbol || symbol.isErroneous) None
-    else if (tpe.typeSymbol.isAnonymousClass) None
+    else if (symbol == null || symbol == NoSymbol || symbol.isErroneous) {
+      val context = doLocateContext(pos)
```
tgodzik, line 208: Is this change needed for `type on selection` ? If not I would move it to a separate PR

mtags/src/main/scala-2/scala/meta/internal/pc/HoverProvider.scala
```diff
@@ -13,89 +15,115 @@ import org.eclipse.lsp4j.MarkupKind
...
+        val symbol = tree.symbol
+        toHover(
+          symbol = symbol,
+          keyword = "",
+          seenFrom = symbol.info,
+          tpe = symbol.info,
+          pos = pos,
+          range = pos
+        )
+      case _: Literal =>
```
tgodzik, line 114:
```suggestion
      case _: Literal if params.isInstanceOf[RangeParams] =>
```
or something along these lines in order not to change the current behaviour.

mtags/src/main/scala/scala/meta/internal/metals/CompilerOffsetParams.scala
```diff
@@ -37,3 +38,42 @@ object CompilerOffsetParams {
...
+      offset,
+      token
+    )
+}
+
+object CompilerRangeParams {
+
+  def fromPos(pos: Position, token: CancelToken): CompilerRangeParams = {
+    val syntax = pos.input.syntax
+    val uri =
```
tgodzik, line 62: Could you move this to a method? It's also repeated above.

tests/cross/src/test/scala/tests/hover/HoverDefnSuite.scala
```diff
@@ -50,8 +66,12 @@ class HoverDefnSuite extends BaseHoverSuite {
       |  <<def @@empty[T] = Option.empty[T]>>
       |}
       |""".stripMargin,
-    """|def empty[T]: Option[T]
-       |""".stripMargin.hover
+    """|Option[T]
```
tgodzik, line 69: Why are these tests actually changed? There is no selection here.

mtags/src/main/scala/scala/meta/internal/pc/HoverMarkup.scala
```diff
@@ -16,17 +16,14 @@ object HoverMarkup {
       docstring: String
   ): String = {
     val markdown = new StringBuilder()
-    val needsExpressionType = !symbolSignature.endsWith(expressionType)
```
tgodzik, line 19: Is this change causing the changes to all the tests? Could we instead take an additional parameter that will be set if range is present?

tgodzik (changes requested):
Thanks for the great work! I have a few comments, overall it would be awesome to limit the changes to only the code selection if possible.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -311,15 +311,29 @@ class Compilers(
     }
   }
 
+  def pcWithAdjustForHover[T](
+      params: HoverExtParams
+  ): (((PresentationCompiler, Position, AdjustLspData) => T) => T) = {
+    (fn: (PresentationCompiler, Position, AdjustLspData) => T) =>
+      if (params.range != null) {
+        withPCAndAdjustLsp(params)(fn)
```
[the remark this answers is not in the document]
KacperFKorban: sure, done

mtags/src/main/scala-2/scala/meta/internal/pc/HoverProvider.scala
KacperFKorban: ok

mtags/src/main/scala/scala/meta/internal/metals/CompilerOffsetParams.scala
KacperFKorban: sure

mtags/src/main/scala-2/scala/meta/internal/pc/HoverProvider.scala
KacperFKorban: not necessary, removed

mtags/src/main/scala/scala/meta/internal/pc/HoverMarkup.scala
KacperFKorban: good idea, done

tests/cross/src/test/scala/tests/hover/HoverLiteralSuite.scala
KacperFKorban: As suggested, added it just for ranges.

ayoub-benali:
> > Really cool feature ! What needs to be done for none VS code clients, say Sublime Text ? :)
> 
> You would need to send a range instead of position, similar to [scalameta/metals-vscode#652](https://github.com/scalameta/metals-vscode/pull/652) - this was actually discussed in the LSP github repo, but not sure if anything will change on that front. This is already being done for Rust and I think Haskell.

@tgodzik Do you mind posting the link so I get the context ?

tgodzik:
The discussion is happening here: https://github.com/microsoft/language-server-protocol/issues/377

bloznelis:
Sorry for necrobumping, but a question for @ckipp01, did you manage to make it work properly in nvim-metals?

PR #3061, opened 2021-08-21
Add action to replace () with {} in functions and vice versa

tgodzik:
Previously, users would need to manually edit parens to braces if they wanted to include multiple expressions. Now, it's possible to rewrite it automatically.

It's also possible to rewrite braces to parens in case of single statement.

metals-sample-1629574444266


dos65 (approved):
LGTM!

metals/src/main/scala/scala/meta/internal/metals/codeactions/RewriteBracesParensCodeAction.scala
```diff
@@ -0,0 +1,95 @@
...
+              codeAction
+          }
+      }
+      .flatten
+      .toSeq
+  }
+}
+
+object RewriteBracesParensCodeAction {
+  val toParens = "Rewrite to parenthesis"
```
kpodsiad, line 93:
```suggestion
  val toParens = "Rewrite to parentheses"
```

kpodsiad (approved):
Maybe it's worth to add negative test like 
```scala
List(1,2,3).map { <<x>> => 
  println(x)
  x + 2
}
```
?

metals/src/main/scala/scala/meta/internal/metals/codeactions/RewriteBracesParensCodeAction.scala
ckipp01: Actually `parenthesis` is correct here.

ckipp01 (approved):
LGTM!

tgodzik:
> Maybe it's worth to add negative test like
> 
> ```scala
> List(1,2,3).map { <<x>> => 
>   println(x)
>   x + 2
> }
> ```
> 
> ?

Added!

metals/src/main/scala/scala/meta/internal/metals/codeactions/RewriteBracesParensCodeAction.scala
kpodsiad: Isn't the `parenthesis` singular?

Issue #3062, opened 2021-08-10
Support `.tasty` files


nicolasstucki:
**Is your feature request related to a problem? Please describe.**

Inability, in VSCode, to open and inspect contents of `.tasty` files.

<img width="997" alt="Screenshot 2021-08-10 at 15 26 02" src="https://user-images.githubusercontent.com/3648029/128875129-6ab2a34d-d197-4043-a978-d176e505b44e.png">

When we click on _open it anyway_ we get that standard binary contents

 <img width="997" alt="Screenshot 2021-08-10 at 15 24 24" src="https://user-images.githubusercontent.com/3648029/128875189-78cb8020-2b7d-4069-92f6-a02c52306123.png">

**Describe the solution you'd like**

Add part of the support we had in the old dotty VsCode plugin to display tasty files.
This was added back in https://github.com/lampepfl/dotty/pull/5513. For this request, we can ignore the decompilation part and only focus on printing the raw tasty format.

![image](https://user-images.githubusercontent.com/4102586/48970828-64242200-f011-11e8-9b9b-08fd456e38eb.png)

Note: Back then we had to click on _open it anyway_  to see the formatted tasty contents. That was a limitation at the time that might be gone by now.

**Additional context**

* Original implementation: https://github.com/lampepfl/dotty/pull/5513
* Tasty HTML printer from from bytes of `.tasty` file: https://github.com/lampepfl/dotty/pull/5513/files#diff-dbc503236ce6915a00eb22ba1ad5671e57d291c8f0be3bb5cf72d083c9eeb65aR12
* To print the tasty file we do not need an instance of the compiler as the original implementation needed. Can take inspiration from https://github.com/lampepfl/dotty/pull/13276/files#diff-38ccef034ef127da7aaece8757397dc262615726a30546433f0f98e5751821ffR45-R47.

PR #3063, opened 2021-08-24
Show TASTy for open file command

kpodsiad:
Closes https://github.com/scalameta/metals/issues/3062

From now, it's possible to open the TASTy equivalent for the `.scala` file via command palette and the `Show TASTy` command.

I'm not convicted about names. `decompile` was created with future reuse for bytecode decompilation in mind. However, AFAIK TASTy isn't decompiled, it's parsed so `decompile` in this context can be misleading. Waiting for your opinion!

Needs https://github.com/scalameta/metals-vscode/pull/655 to work.


ckipp01:
So right now this will only work with an editor that has a web view. Do you think it'd be possible to add in a check when the request comes in to look at `ClientConfiguration.isHttpEnabled`. If that settings is `false`, then this can just stay the same, but if this is `true` it'd be nice to then just use the http server metals provides and still show this for those clients in the browser.

I haven't played around with TASTy much at all, but looking there seems to be a generic `TastyPrinter` that we could also use for clients that don't have a web view, but also want to just display the text locally. I don't think we need to really do this know, but I know for myself in NeoVim that's something I'll probably aim at doing later on. So it might be good to at least keep it in mind so that there is possibly a setup in place that we can handle this similar to how we do the doctor with different formats based on the client.

kpodsiad:
> So right now this will only work with an editor that has a web view. Do you think it'd be possible to add in a check when the request comes in to look at `ClientConfiguration.isHttpEnabled`. If that settings is `false`, then this can just stay the same, but if this is `true` it'd be nice to then just use the http server metals provides and still show this for those clients in the browser.
> 
> I haven't played around with TASTy much at all, but looking there seems to be a generic `TastyPrinter` that we could also use for clients that don't have a web view, but also want to just display the text locally. I don't think we need to really do this know, but I know for myself in NeoVim that's something I'll probably aim at doing later on. So it might be good to at least keep it in mind so that there is possibly a setup in place that we can handle this similar to how we do the doctor with different formats based on the client.

I even wasn't aware of that option and possibility, thanks for pointing that out! I'd check how doctor command is handled and try to do the same with TASTy.

metals/src/main/scala/scala/meta/internal/metals/Decompiler.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import scala.meta.internal.metals.MetalsEnrichments._
+
+import com.google.gson.JsonPrimitive
+import org.eclipse.{lsp4j => l}
+import java.{util => ju}
+
+class Decompiler(
+    compilers: Compilers
+) {
```
tgodzik, line 16: Do we need a separate class for this? We could include this in `Compilers`

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -371,13 +377,50 @@ class Compilers(
     }
   }
 
+  def decompile(
+      uri: URI
+  ): Future[ju.Optional[ExecuteCommandParams]] = {
+    def buildTargetToPcAndFileUri(
+        target: ScalaTarget
+    ): Option[(PresentationCompiler, URI)] = {
+      val filePath = AbsolutePath.fromAbsoluteUri(uri)
```
tgodzik, line 386: We should check if the file is already a tasty file.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -190,6 +190,14 @@ object ServerCommands {
     "[string], where the string is a stacktrace."
   )
 
+  val Decompile = new Command(
+    "decompile",
+    "Show TASTy",
+    """|Show tasty file and show its content.
```
tgodzik, line 196:
```suggestion
    """|If the file is a Scala 3 source  this command will try to find the relevant tasty file for it, read it and display in a human readable form. If the argument already points to a Scala file, it will be read directly.
```

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -190,6 +190,14 @@ object ServerCommands {
...
+       |""".stripMargin,
+    "[string], where the string is a path to tasty file."
```
tgodzik, line 198:
```suggestion
    "[string], where the string is a path to a Scala 3 source or a tasty file."
```

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
```diff
@@ -73,6 +74,12 @@
      */
     public abstract CompletableFuture<DefinitionResult> definition(OffsetParams params);
 
+    /**
+     * Decompile .tasty or .class file.
+
+     */
+    public abstract CompletableFuture<Optional<ExecuteCommandParams>> decompile(URI targetUri);
```
tgodzik, line 81: Maybe we could return string or some data structure that could also be used for json file for other editors.

```suggestion
    public abstract CompletableFuture<Optional<String>> decompile(URI targetUri);
```

This can be easily translated later on to invoke a command.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1717,6 +1719,15 @@ class MetalsLanguageServer(
           scribe.debug(s"Executing AnalyzeStacktrace ${command}")
         }.asJavaObject
 
+      case ServerCommands.Decompile() =>
+        val command = decompiler.decompile(params)
+        command.map { commandOpt =>
+          commandOpt.asScala.foreach(
+            languageClient.metalsExecuteClientCommand
```
tgodzik, line 1726: This should be based on what kind of support a client has (can it display html or json etc.) simialr to how the Doctor works.

tgodzik (changes requested):
This looks promising! Thanks for the great work. I left some suggestions and overall ideas about the code.

Alos, you should run `sbt scalafixAll` to sort the imports and some tests seem to be failing still.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
kpodsiad: Assuming that client knows what it supports execute the show client command should be enough - client can handle this request differently, depending on the configuration. 
The response format is taken into account (isHtmlSupported flag)

metals/src/main/scala/scala/meta/internal/metals/Decompiler.scala
kpodsiad: I find separate file more readable because`Compilers` already have ~700loc. Also, separate file plays nicely with single responsibility - tasty stuff is extracted to a tasty file. Any chance that I convinced you? ;)

kpodsiad:
I've refactored few things and that's how things look now.
Untitled Diagram (1)

I didn't implement show tasty support via with HTTP server - I didn't know how to implement the async endpoint in the undertow and 20minutes of googling wasn't enough so I gave up.

Also, some tests are failing and honestly, I don't know why. I'll have to look closer

mtags/src/main/scala-3/scala/meta/internal/pc/TastyUtils.scala
```diff
@@ -0,0 +1,73 @@
...
+    HtmlBuilder()
+      .page(title, htmlStyles :: headElems, bodyAttributes) { builder =>
+        builder
+          .element("pre")(_.raw(tastyHtml))
+      }
+      .render
+  }
+
+  private def tastyHtmlPageTitle(file: URI) = {
+    val filename = PathIO.basename(AbsolutePath.fromAbsoluteUri(file).toString)
```
dos65, line 56:
```suggestion
    import MtagsEnrichments._ 
    val filename = AbsolutePath.fromAbsoluteUri(file).filename
```

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -257,6 +257,9 @@ final class TestingServer(
     }
   }
 
+  def assertShowTasty(filePath: String): Future[Option[String]] =
```
dos65, line 260: As there is no asserts maybe?
```suggestion
  def showTasty(filePath: String): Future[Option[String]] =
```

dos65:
> I didn't implement show tasty support via with HTTP server - I didn't know how to implement the async endpoint in the undertow and 20minutes of googling wasn't enough so I gave up.

@kpodsiad I've never worked with undertow but looking at to `HttpHandler` signature it seems that smth like the this should work well:
```scala
 new HttpHandler {
      override def handleRequest(exchange: HttpServerExchange): Unit = {
          $tastyAsyncOp.onComplete {
             case Success(value) =>
                 exchange.getResponseSender.send(..) 
              case Failure(e) =>
                  // set error http code and complete
          }
 }
```

metals/src/main/scala/scala/meta/internal/metals/TastyHandler.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  def executeShowTastyCommand(
+      params: l.ExecuteCommandParams
+  ): Future[Unit] = {
+    val tastyResponseOpt = getTasty(params)
+    tastyResponseOpt.map { tastyOpt =>
+      val commandOpt = tastyOpt.map { tasty =>
+        new l.ExecuteCommandParams(
+          "metals-show-tasty",
+          List[Object](tasty).asJava
```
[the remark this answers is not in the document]
kpodsiad: Without type params code will not compile - `Note: String <: Object, but Java-defined trait List is invariant in type `

kpodsiad:
> @kpodsiad I've never worked with undertow but looking at to `HttpHandler` signature it seems that smth like the this should work well:
> 
> ```scala
>  new HttpHandler {
>       override def handleRequest(exchange: HttpServerExchange): Unit = {
>           $tastyAsyncOp.onComplete {
>              case Success(value) =>
>                  exchange.getResponseSender.send(..) 
>               case Failure(e) =>
>                   // set error http code and complete
>           }
>  }
> ```

Unfortunately, it doesn't work this way :(

kpodsiad:
I also wonder what would be the best way to tell users that this feature works only with scala 3 and only for .scala or .tasty files. Currently, there is no message nor notification executing show tasty seems like it doesn't do anything.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -190,6 +190,15 @@ object ServerCommands {
     "[string], where the string is a stacktrace."
   )
 
+  val ShowTasty = new Command(
+    "show-tasty",
+    "Show TASTy",
+    """|If the file is a Scala 3 source  this command will try to find the relevant tasty file for it, 
+       |read it and display in a human readable form. If the argument already points to a Scala file, 
+       |it will be read directly.""".stripMargin,
```
ckipp01, line 198:
```suggestion
    """|If the file is a Scala 3 source this command will try to find the relevant tasty file for it, 
       |read it and display it in a human readable format. If the argument already points to a Scala file, 
       |it will be read directly.""".stripMargin,
```

metals/src/main/scala/scala/meta/internal/metals/TastyHandler.scala
```diff
@@ -0,0 +1,151 @@
...
+      tastyFileURI <- getTastyFileURI(
+        uri,
+        buildTarget.scalac.getClassDirectory.toAbsolutePath
+      )
+    } yield (pc, tastyFileURI)
+
+    pcAndTargetUriOpt match {
+      case Some((pc, tastyUri)) =>
+        pc.getTasty(tastyUri, isHtmlSupported, isHttpEnabled).asScala
+      case _ => Future.successful("Error, .tasty file doesn't exist")
```
ckipp01, line 122: Do you think it's possible to maybe change this up a bit and not return just a string as an argument, but rather something more structured? Right now if this happens (which it did to me when testing), this will need to be handled differently by the client (especially when they aren't using html) than the actual tasty string that is returned. The only way to detect this currently would is to basically do a string.startsWith("Error") type of check. 
Even something simple like:

```ts
interface TastyReponse {
  error?: String
  body?: String
}
```
Then it'd be possible to check for an error. Either that, or don't return `successful` in this case. wdyt?

mtags/src/main/scala-2/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -153,6 +153,15 @@ case class ScalaPresentationCompiler(
       new AutoImportsProvider(pc.compiler(), name, params).autoImports().asJava
     }
 
+  override def getTasty(
+      targetUri: URI,
+      isHtmlSupported: Boolean,
+      isHttpEnabled: Boolean
+  ): CompletableFuture[String] =
+    CompletableFuture.completedFuture {
+      "Currently, there is no support for Show TASTy in Scala 2."
```
ckipp01, line 162:
```suggestion
      """Currently, there is no support for the "Show TASTy" feature in Scala 2."""
```

metals/src/main/scala/scala/meta/internal/metals/TastyHandler.scala
```diff
@@ -0,0 +1,151 @@
...
+    val absolutePathOpt = Try(AbsolutePath.fromAbsoluteUri(uri)) match {
+      case Success(value) => Some(value)
+      case Failure(exception) =>
+        scribe.error(exception.toString())
+        None
+    }
+    val pcAndTargetUriOpt = for {
+      absolutePath <- absolutePathOpt
+      buildTargetId <- buildTargets.inverseSources(absolutePath)
+      buildTarget <- buildTargets.scalaTarget(buildTargetId)
```
ckipp01, line 111: I saw you asked a question about letting the user know that it's not supported in Scala 3. I think it'd be nice to just warn them in that situation with a message. We can check that here once we have the `buildTarget`.

ckipp01 (commented):
This is really rad @kpodsiad!

First up a little demo of how it looks with nvim-metals:


I just left a few suggestions. Mainly around making it easier for a client to determine if something is going wrong or not and I do think we should just warn the user if they try to use it on Scala 3 sources.

As you see at the end of my gif, it is also seeming to not find the tasty file for my `Main.scala` consistently for some reason. I haven't been able to spot why that is.

metals/src/main/scala/scala/meta/internal/metals/TastyHandler.scala
kpodsiad: Yeah, I cannot agree more. 
I even tried to create discriminated union type but I don't think it's possible due to how command serialization works :( so I ended with sth like you proposed.

metals/src/main/scala/scala/meta/internal/metals/TastyHandler.scala
kpodsiad: I added here the check, but IMO differentiating between warning and error might add unnecessary complexity to the code since it would require a third possible result. Let's call it the da. 
If users would start to complain about this, then we could alter it. Seems fine?

metals/src/main/scala/scala/meta/internal/metals/TastyHandler.scala
```diff
@@ -0,0 +1,184 @@
...
+
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.io.AbsolutePath
+
+import com.google.gson.JsonPrimitive
+import org.eclipse.{lsp4j => l}
+
+// response which is send to the lsp client. Because of java serialization we cannot use
+// sealed hierarchy to model union type of success and error.
+// Moreover, we canot also use Option, so instead every field is nullable
```
ckipp01, line 19: Maybe let's switch this to valid scaladocs. Acutally same for the `TastyHandler`. Whenever we add new handlers I always try to remember to add scaladocs to them since it's fresh in our mind now, but for a newcomer later on it's really nice to have that info there.

ckipp01:
```suggestion
// response which is sent to the lsp client. Because of java serialization we cannot use
// a sealed hierarchy to model union type of success and error.
// Moreover, we cannot use Option, so instead every field is nullable
```

metals/src/main/scala/scala/meta/internal/metals/TastyHandler.scala
```diff
@@ -0,0 +1,184 @@
...
+      case js: JsonPrimitive if js.isString => js
+    }
+  } yield arg.getAsString()
+
+  /**
+   * For given uri (could be both .scala and .tasty file) try to find:
+   * - build target
+   * - uri to existing tasty file
+   * Iff they both exists get decoded tasty file content (pc.getTasty takes into account if client supports html)
+   */
```
ckipp01, line 111:
```suggestion
   * For a given uri (could be both .scala and .tasty file) try to find:
   * - build target
   * - uri to existing tasty file
   * If they both exists get decoded tasty file content (pc.getTasty takes into account if client supports html)
   */
```

ckipp01 (approved):
Just a couple small nits on wording, but other than that, LGTM from me! Great job on this!

Much simpler now to show the error to the user:



And much easier now to display the correct file name when it comes back 😆

metals/src/main/scala/scala/meta/internal/metals/TastyHandler.scala
ckipp01: 👍🏼 sounds good to me.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -190,6 +190,15 @@ object ServerCommands {
     "[string], where the string is a stacktrace."
   )
 
+  val ShowTasty = new Command(
+    "show-tasty",
+    "Show TASTy",
+    """|If the file is a Scala 3 source, this command will try to find the relevant tasty file for it, 
+       |read it and display it in a human readable format. If the argument already points to a Scala file, 
+       |it will be read directly.""".stripMargin,
```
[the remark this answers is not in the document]
kpodsiad: It will certainly decrease the need to jumping across the sources, I like this idea.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -190,6 +191,25 @@ object ServerCommands {
...
+      requestedUri: URI,
+      tasty: String,
+      error: String
+  )
+
+  val ShowTasty = new Command(
+    "show-tasty",
+    "Show TASTy",
+    """|If the file is a Scala 3 source, this command will try to find the relevant tasty file for it, 
+       |read it and display it in a human readable format. If the argument already points to a Scala file,
```
tgodzik, line 208:
```suggestion
       |read it and display it in a human readable format. If the argument already points to a Tasty file, 
```

tgodzik (approved):
Just one small nitpick, but otherwise LGTM

ckipp01:
Again, great job on this @kpodsiad!

nicolasstucki:
Thank you @kpodsiad this will save me a lot of time.

PR #3070, opened 2021-08-25
Fix comments erasing all later variables in .env file

tgodzik:
Previously, regex used for .env files would cover all characters after any comments, so effectively until end of file and miss all the later variables. Now, we only match comment without any newlines.


dos65 (approved):
LGMT!

PR #3077, opened 2021-08-30
Display build target info

Arthurm1:
Displays detailed target info.

This depends on some of #2520 because of the target refactor and javacoptions additions

This is my first change to the VSCode extension and I don't really know where the line is drawn in expanding the API by adding commands and where the split in functionality between the extension and Metals should lie especially due to none-vscode clients.

I've commandeered the Doctor webview because of the navigation between it and the targets.  I was going to use the [Virtual Document](https://code.visualstudio.com/api/extension-guides/virtual-documents) API as it seems a lot simpler, more lightweight and usable across non-vscode clients but I don't see that navigation is possible with it.

It doesn't seem possible to reveal a directory/folder in the VSCode file explorer which I'd like to navigate to when clicking on the various classes/classpath/source folders as the VSCode file explorer doesn't seem to have much of a public API.  Has anyone seen a way to do this?

It also looks as though Metals doesn't request `resources` directories from BSP - I assume because it doesn't need them - or is this stored somewhere outside `scala.meta.internal.metals.BuildTargets`?

The motivation for this was not having to keep navigating and perusing the json files in the .bloop folder for build setup and also viewing the additional javac/scalac options that Bloop adds (e.g. semanticdb)

Here's an example...

https://user-images.githubusercontent.com/5612295/131269946-3ec903ca-bcb2-41b1-b4db-892f94b2fdd5.mp4

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -197,6 +197,7 @@ lazy val V = new {
   val ammonite213Version = scala213
   val scalameta = "4.4.27"
   val semanticdb = scalameta
+  val javaSemanticdb = "0.6.8"
   val bsp = "2.0.0-M13"
   val bloop = "1.4.8-114-e47368ed"
   val scala3 = "3.0.2"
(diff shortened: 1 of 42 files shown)


tgodzik:
That's super cool! Sorry, I am so late with all the reviews, I managed to bump Bloop in Metals now, so I plan to finish some remaining issues for 1.4.9 and release it.

Arthurm1:
replaced by #3380

PR #3078, opened 2021-08-30
Use new `lsif-java index` command to upload Sourcegraph index

olafurpg:
Previously, the CI step uploaded LSIF through the sbt-sourcegraph
plugin. Now, we use the built-in `lsif-java index` command that handles
everything for us we no longer need the sbt-sourcegraph plugin
dependency. The benefit of this change is that Java sources will also
get indexed!


olafurpg:
I'm working on fixing the error by creating a lsif-java-action to do all of the steps

ckipp01:
Close in favor of https://github.com/scalameta/metals/pull/3152

PR #3092, opened 2021-09-01
Github actions issue for reporting

tgodzik:


tgodzik:
It looks like old actions are being reproted despite not being run at all, reported it here: https://github.com/actions/runner/issues/1303

Let's see if it's cached somehow maybe

omsmith:
Hi, saw your issue in the actions/runner repo.

Your old jobs are still showing up in the status checks because they are _Required_ as per your Branch Protection settings (hence why they're _Expected_ and not failed or the like). Update your Branch Protection settings for _main_ to use your new job names and remove the old ones.

tgodzik:
@omsmith Awesome, thanks! I think it was not there when I enabled them.

PR #3093, opened 2021-09-01
Implement 'find in JAR files' LSP extension

Z1kkurat:
This is an attempt to implement this functionality: https://github.com/scalameta/metals-feature-requests/issues/196  
The implementation is very naive and just scans through all workspace jars, reading files by mask and searching for a full content match. As was discussed on Discord with @olafurpg, this can be tolerated as the initial implementation.

To test this I implemented client-side functionality in `metals-vscode` using a separate custom tree view just to showcase that the functionality itself works. I think I will attempt to make a PR with it once I clean it up (the code is very bad and dirty right now).  

out


metals/src/main/scala/scala/meta/internal/metals/FindTextInFiles.scala
```diff
@@ -0,0 +1,130 @@
...
+              attrs: BasicFileAttributes
+          ): FileVisitResult = {
+            val absPath = AbsolutePath(file)
+            if (isSuitableFile(absPath, mask)) {
+              val fileRanges: List[Range] = visitFileInsideJar(absPath, content)
+
+              val fileLocations: List[Location] =
+                if (fileRanges.nonEmpty) {
+                  val result = absPath.toFileOnDisk(workspace())
+                  fileRanges.map(range => new Location(result.toString, range))
```
dos65, line 79: Location.uri takes a string but it's actually should be an uri encoded string
```suggestion
                  fileRanges.map(range => new Location(result.toURI.toString, range))
```

tests/unit/src/test/scala/tests/FindTextInFilesSuite.scala
```diff
@@ -0,0 +1,38 @@
...
+           |{
+           |  "a": {
+           |    "scalaVersion": "2.12.4",
+           |    "libraryDependencies": ["com.typesafe.akka::akka-actor-typed:2.6.16"]
+           |  }
+           |}
+        """.stripMargin
+      )
+      locations <- server.findTextInFiles(".conf", "jvm-shutdown-hooks")
+      _ = assertEquals(locations, expectedLocations)
```
dos65, line 35: In metals tests, usually received and expected results are rendered into String representation that is more readable.
Could you do assert in a similar way to how it's done in `WorkspaceSymbolLspSuite`? https://github.com/scalameta/metals/blob/55161a500b77946e4052a91e38479ad42dd285ff/tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala#L160-L198

dos65 (changes requested):
Demonstration looks awesome! Thanks for this contribution!

The approach in general looks good.
The only thing is that I'm not sure if it should be named `Find in non-source files`.
Maybe `Search in dependency sources`? Anyway let's wait what others will say.

I left some suggestions to the code:

metals/src/main/scala/scala/meta/internal/metals/FindTextInFiles.scala
```diff
@@ -0,0 +1,130 @@
...
+import org.eclipse.lsp4j.Position
+import org.eclipse.lsp4j.Range
+
+class FindTextInFiles(
+    buildTargets: BuildTargets,
+    workspace: () => AbsolutePath
+) {
+  def find(mask: String, content: String): List[Location] = {
+    if (mask == null || content == null) {
+      List.empty[Location]
```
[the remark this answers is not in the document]
tgodzik: Probably an annotation would work there?

metals/src/main/scala/scala/meta/internal/metals/FindTextInFiles.scala
```diff
@@ -0,0 +1,130 @@
...
+    } else {
+      val allLocations: ArrayBuffer[Location] = new ArrayBuffer[Location]()
+
+      buildTargets.allWorkspaceJars.foreach { classpathEntry =>
```
tgodzik, line 32: I think we could cache text file paths when indexing, which would make it all much faster and wouldn't take up much more memory.

metals/src/main/scala/scala/meta/internal/metals/FindNonSourceTextRequest.scala
```diff
@@ -0,0 +1,6 @@
+package scala.meta.internal.metals
+
+case class FindNonSourceTextRequest(
+    mask: java.lang.String,
```
tgodzik, line 4: Do we need a mask really? Jars will contain `.conf` files and `.properties` most likely, so not much possible values there.

tgodzik (commented):
Looks great!

I wonder though if it wouldn't be a better UX if we had just one `QuickPick` displayed that would contain all text files (we could cache them), we can than easily filter those in the quick pick window by writing `.conf`.  That would mean we would need to search for a specific string later on, so maybe that wouldn't be better :thinking:

tgodzik:
> To test this I implemented client-side functionality in metals-vscode using a separate custom tree view just to showcase that the functionality itself works. I think I will attempt to make a PR with it once I clean it up (the code is very bad and dirty right now).

What if we unpack all the matched files to `.metals/dependencies` and run search in VS Code on those files? That would mean we wouldn't need a new view.

metals/src/main/scala/scala/meta/internal/metals/FindNonSourceTextRequest.scala
dos65: I think there are more plain-text extensions that might be presented in jars and we will never cover them all.
Also the mask might be used not only for filtering by extension but also just by full name like: `sample.json`.

metals/src/main/scala/scala/meta/internal/metals/FindTextInFiles.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  def visitFileInsideJar(path: AbsolutePath, content: String): List[Range] = {
+    var reader: BufferedReader = null
+    val positions: ArrayBuffer[Int] = new ArrayBuffer[Int]()
+    val results: ArrayBuffer[Range] = new ArrayBuffer[Range]()
+    val contentLength: Int = content.length()
+
+    try {
+      reader = new BufferedReader(
+        new InputStreamReader(Files.newInputStream(path.toNIO))
```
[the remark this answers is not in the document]
Z1kkurat: But this would be rather memory-inefficient, no? Right now I'm reading the file line by line so GC can immediately collect previously read lines. And in the case of reading the whole file as a String, it's allocating the string with the size of a file, splitting it into a list of lines, and only after that - iterating through the lines, looking for a match.  
Since this code already does a lot of work naively scanning all jars and all files inside, I wanted to try to make this as fast and efficient as possible.

metals/src/main/scala/scala/meta/internal/metals/FindNonSourceTextRequest.scala
Z1kkurat: I did it by advice from @dos65 and I guess he has a point - some people put xml/json/txt files in jars, so who knows what a user may need to search for

Z1kkurat:
>What if we unpack all the matched files to .metals/dependencies and run search in VS Code on those files? That would mean we wouldn't need a new view.

Hmm, I never thought of that, a good point! But here we need to estimate the performance impact, I guess? Copying (potentially) many files to dependencies directory (and no guarantees that we will find required content there) versus reading files in-place. I have nothing against both solutions, but I guess they are different in their nature of the interaction with the editor - copying is a side-effect which returns nothing and requires executing the editor command later and in-place search is a plain function which returns a list of locations so editor client can update some internal view representation.  
Do you think copying would be a better approach?

>I wonder though if it wouldn't be a better UX if we had just one QuickPick displayed that would contain all text files (we could cache them), we can than easily filter those in the quick pick window by writing .conf. That would mean we would need to search for a specific string later on, so maybe that wouldn't be better 🤔

Initially, I wanted to provide a 'peek' view, similar to the one used in 'find references', but VSCode API said no 😄 - it required peeking from some specific file location. TreeView looks okay I think - it gives both the file name and a preview of found entries. But I also don't mind coming to a consensus for the best solution. And other editors will be able to implement it the way they see fit for themselves anyway 😄

metals/src/main/scala/scala/meta/internal/metals/FindTextInFiles.scala
Z1kkurat: So essentially an index of `jar path -> list of files inside a jar`? Would that not be huge (since there are lots of classfiles etc)?

metals/src/main/scala/scala/meta/internal/metals/FindTextInFiles.scala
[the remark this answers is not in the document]
Z1kkurat: You mean `javax.annotation.Nonnull`? I didn't see its usages in metals, will it work?

metals/src/main/scala/scala/meta/internal/metals/FindNonSourceTextRequest.scala
```diff
@@ -0,0 +1,6 @@
...
+    content: java.lang.String
```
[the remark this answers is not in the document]
Z1kkurat: done

metals/src/main/scala/scala/meta/internal/metals/FindTextInFiles.scala
```diff
@@ -0,0 +1,130 @@
...
+              jarLocations ++= fileLocations
+            }
+
+            FileVisitResult.CONTINUE
+          }
+        }
+      )
+    }
+
+    jarLocations.toList
```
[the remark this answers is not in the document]
Z1kkurat: done

metals/src/main/scala/scala/meta/internal/metals/FindTextInFiles.scala
Z1kkurat: done

tests/unit/src/test/scala/tests/FindTextInFilesSuite.scala
Z1kkurat: Not yet done, moved file so this change is not really outdated, will fix this

Z1kkurat:
A task list for myself:

- [ ] fix equality assertion in suite
- [ ] (possibly?) index for jar contents per Tomasz's suggestion - need to research this
- [ ] discuss the possibility of copying files with opening editor native search later instead of scanning and reading files

metals/src/main/scala/scala/meta/internal/metals/FindTextInFiles.scala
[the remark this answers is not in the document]
dos65: Sounds reasonable.
Even if it's weird, I think it's a possible case that somebody has large datasets in resources.
Let's keep it as it is.

dos65:
It's too late, but I realized that we need to check if vscode is able to search in virtual file systems. 
I took a look at vscode sources and probably it can do that.

This feature is anyway requires support from the editor side so maybe we can do that using std vscode api.

Z1kkurat:
> It's too late, but I realized that we need to check if vscode is able to search in virtual file systems.
> I took a look at vscode sources and probably it can do that.
> 
> This feature is anyway requires support from the editor side so maybe we can do that using std vscode api.

In my showcase implementation I used only standard vscode API for opening files and showing them, nothing fancy, so I guess it _should_ work?

Z1kkurat:
By the way, here it is - very unpolished and not yet ready for PR to the main repo, so I made a PR to the main branch of my fork so changes can be easily seen:
Branch: https://github.com/Z1kkurat/metals-vscode/tree/find-files-in-dependency-jars
PR: https://github.com/Z1kkurat/metals-vscode/pull/1

dos65:
@Z1kkurat I'm about a different thing. There is `registerFileSystemProvider` function in [workspace api](https://code.visualstudio.com/api/references/vscode-api#workspace). I'm wondering if we can make a custom one for dependencies and vscode will be able to search trough it.

Z1kkurat:
@dos65 can you please elaborate? I'm not very familiar with vscode API, so I guess we will need to index such dependency files to present them as a filesystem tree, since my current implementation scans files on demand?

dos65:
I checked vscode api. There are plans to expose interfaces to make custom search providers but it still in [proposed status](https://github.com/microsoft/vscode/issues/59921)

I think the solution in this pr is quite good and cover this missing feature for some users who are switching from IDEA to Metals.
I don't think that there is a need in indexing for this search as it should be fast enough plus I don't think that it will be used often. 
The only issue might happen with large files but I think we can add a limit on filesize. 

@tgodzik @ckipp01 wdyt? Could we agree on introducing this new endpoint and push this feature further?

metals/src/main/scala/scala/meta/internal/metals/FindTextInDependencyJarsRequest.scala
```diff
@@ -0,0 +1,6 @@
+package scala.meta.internal.metals
+
+case class FindTextInDependencyJarsRequest(
```
ckipp01, line 3: Can we add some scaladocs to this. From the outside it might not be clear what exactly `mask` is supposed to be here or even content really.

ckipp01 (commented):
Great job on this @Z1kkurat, as @dos65 said I think this is a big feature to help those transitioning from IntelliJ. I know I would use this feature.

> @tgodzik @ckipp01 wdyt? Could we agree on introducing this new endpoint and push this feature further?

👍🏼 from me

olafurpg:
> I checked vscode api. There are plans to expose interfaces to make custom search providers but it still in proposed status

Can we confirm that the current extension will be usable with the new proposed text search API? It would be a shame if we have to change the JSON-RPC endpoints to make it work with the new text search API once it’s available.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1891,6 +1894,46 @@ class MetalsLanguageServer(
         .orNull
     }.asJava
 
+  @JsonRequest("metals/findTextInDependencyJars")
+  def findTextInDependencyJars(
+      params: FindTextInDependencyJarsRequest
+  ): CompletableFuture[util.List[Location]] = {
+    def readMask: Future[String] = Option(params.mask) match {
```
tgodzik, line 1901: Could we move most of the logic to FindTextInDependencyJars and just invoke it here with options?
```
findTextInJars.find(Option(params.mask), Option(params.content))
```

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1891,6 +1894,46 @@ class MetalsLanguageServer(
...
+      case Some(mask) =>
+        Future.successful(mask)
+      case None =>
+        languageClient
+          .metalsInputBox(
+            MetalsInputBoxParams(value = ".conf", prompt = "Enter file mask")
+          )
+          .asScala
+          .map(result => result.value)
```
tgodzik, line 1910:
```suggestion
          .map{
            case name if !name.cancelled && name.value.nonEmpty =>
               Some(name.value)
            case _ =>
               None
          }
```

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1891,6 +1894,46 @@ class MetalsLanguageServer(
...
+        Future.successful(content)
+      case None =>
+        languageClient
+          .metalsInputBox(
+            MetalsInputBoxParams(
+              prompt = "Enter content to search for"
+            )
+          )
+          .asScala
+          .map(result => result.value)
```
tgodzik, line 1924:
```suggestion          
          .map{
            case name if !name.cancelled && name.value.nonEmpty =>
               Some(name.value)
            case _ =>
               None
          }
```

Or you could also use a function like:
```scala
def checkResult(result: MetalsInputBoxResult) = result match {
  case name if !name.cancelled && name.value.nonEmpty =>
      Some(name.value)
  case _ =>
      None
}
```

and use it for both params.  You will not have to check for `mask == null || content == null)` you will just be able to pass an Option. I think especially `mask` can be empty here.

metals/src/main/scala/scala/meta/internal/metals/FindTextInDependencyJarsRequest.scala
```diff
@@ -0,0 +1,6 @@
...
+    mask: String,
```
tgodzik, line 4: Should we instead have `include` and `exclude` the same as normal search so that it's compatible with new proposed text search API? We can use include for now for the mask.

Also instead of content we could model it on the VS Code API from https://github.com/Microsoft/vscode/blob/f35b6ce8d8a24f2dcf756d8bd6e2069dd4f25cc9/src/vs/vscode.proposed.d.ts#L21

And essentially change everything to
```scala

case class FindTextInDependencyJarsRequest(
    @Nullable
    options: FindTextInFilesOptions,
    query: TextSearchQuery
)

case class TextSearchQuery(
   pattern: String,
   @Nullable
   isRegExp: java.lang.Boolean,
   @Nullable
   isCaseSensitive:  java.lang.Boolean,
   @Nullable
   isWordMatch:  java.lang.Boolean;
)

case class FindTextInFilesOptions(
 @Nullable
  include: String,
  @Nullable
  exclude: String
)
```

Almost everything here is nullable, so we could implement it later, but it would be good to have an API we can use later on.

metals/src/main/scala/scala/meta/internal/metals/FindTextInDependencyJars.scala
```diff
@@ -0,0 +1,110 @@
...
+            e
+          )
+      }
+    }
+
+    allLocations.toList
+  }
+
+  private def isSuitableFile(path: AbsolutePath, mask: String): Boolean = {
+    path.isFile && path.filename.contains(mask)
```
tgodzik, line 46: This should probably be a glob pattern instead of simply checking `contains`. 
```suggestion
    path.isFile && ScalafmtConfig.PathMatcher.Nio(mask).matches(path)
```

it coud make sense to move `ScalafmtConfig.PathMatcher` to just PathMatcher.

tgodzik: With include/exclude we could do:
```scala
path.isFile && ScalafmtConfig.PathMatcher.Nio(include).matches(path) && !ScalafmtConfig.PathMatcher.Nio(exclude).matches(path)
```

metals/src/main/scala/scala/meta/internal/metals/FindTextInDependencyJarsRequest.scala
```diff
@@ -0,0 +1,6 @@
+package scala.meta.internal.metals
```
tgodzik, line 1:
```suggestion
package scala.meta.internal.metals.findfiles
```
maybe keep this and the provider in a separate package similar to how treeview is kept?

tgodzik (changes requested):
Thanks for the great work! I finally managed to dig into the PR and think about it. I left a bunch of comments, which might be a bit overwhelming, so if you need help let me know I would be happy to work on this a bit! We don't have to implement everything in this PR for sure,

Z1kkurat:
Wow, that's a lot of feedback, thank you very much @tgodzik ! 😄 
I am planning to come back to it somewhere during this week, but I'm not sure I will have enough time to address all of the comments, so I would definitely appreciate your help on that after I address as many of them as I can.

metals/src/main/scala/scala/meta/internal/metals/FindTextInDependencyJarsRequest.scala
Z1kkurat: done!

Z1kkurat:
@tgodzik I think that I addressed most of the comments. Please let me know if there's anything else on your mind regarding this PR or if you notice that I have missed something!

ckipp01:
🚀 working in `nvim-metals`.

PR #3095, opened 2021-09-01
Add support for Scala 3.0.2

tgodzik:


build.sbt
```diff
@@ -243,7 +242,7 @@ lazy val V = new {
   def scala2Versions = nonDeprecatedScala2Versions ++ deprecatedScala2Versions
 
   // Scala 3
-  def nonDeprecatedScala3Versions = Seq(nextScala3RC, scala3, "3.0.0")
+  def nonDeprecatedScala3Versions = Seq(scala3, "3.0.1", "3.0.0")
```
gabro, line 245: probably not worth doing it now, but is there a plan to deprecate older patch versions of the compiler? Scala 3 versions should be really easy to upgrade, so Metals could be a bit more aggressive in deprecating older versions which don't benefit from many compiler fixes

tgodzik: Maybe we should start depreacating with the next major version? `3.1.0` is next, so I would deprecate `3.0.0` then

gabro: 👍 the cool thing about Scala 3 is that minor versions are binary compatible, so I think for all intents and purposes Metals could treat all 3.x.y versions the same, and maintain compatibility for the most recent X (3? 4?) ones.

This is because, AFAIK, moving from 3.0.1 to 3.0.2 or from 3.0.2 to 3.1.0 should be same from a user perspective.

dos65 (approved):
LGTM! 
Failed tests are flaky

build.sbt
smarter: FWIW, the FP course at EPFL this semester will use Scala 3.0.2, so I'd appreciate it if it didn't get deprecated in the middle of the semester.

Issue #3098, opened 2021-09-02
Check for trace files also in `.metals/` directory


kpodsiad:
### Discussed in https://github.com/scalameta/metals/discussions/3074

Long story short:
<div type='discussions-op-text'>
<sup>Originally posted by **tgodzik** August 28, 2021</sup>

I think it's a good idea, though I would maybe add it as a new option? So if you create `~/.metals/lsp.trace.json` it will only log there and not to the global one. What do you think?
</div>

PR #3103, opened 2021-09-06
Take into account trace files in .metals directory

kpodsiad:
Previously, in order to log JSON-RPC traces, one has to create proper files in the cache directory. What's important, those files are global for every metals server.
Now, there is an additional option available. One can create those files in the `.metals` directory and opt-in/out logging per project. No more reloading specific metals instance in order to see logs from it!

closes https://github.com/scalameta/metals/issues/3098

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
@@ -280,9 +280,9 @@ object BuildServerConnection {
 
     def setupServer(): Future[LauncherConnection] = {
       connect().map { case conn @ SocketConnection(_, output, input, _, _) =>
-        val tracePrinter = GlobalTrace.setupTracePrinter("BSP")
+        val tracePrinter = Trace.setupTracePrinter("BSP", workspace)
         val launcher = new Launcher.Builder[MetalsBuildServer]()
-          .traceMessages(tracePrinter)
+          .traceMessages(tracePrinter.orNull)
           .setOutput(output)
           .setInput(input)
           .setLocalService(localClient)
--- a/metals/src/main/scala/scala/meta/internal/metals/Trace.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Trace.scala
@@ -16,23 +16,29 @@ import dev.dirs.ProjectDirectories
  */
 object Trace {
 
-  /**
-   * Returns a printer to trace JSON messages if the user opts into it.
-   */
-  def setup(protocolName: String): PrintWriter = {
+  def protocolTracePath(
+      protocolName: String,
+      workspace: AbsolutePath = PathIO.workingDirectory
+  ): AbsolutePath = {
+    val traceFilename = s"${protocolName.toLowerCase}.trace.json"
+    globalDirectory.resolve(".metals").resolve(traceFilename)
+  }
+
+  def setup(
+      protocolName: String,
+      workspace: AbsolutePath = PathIO.workingDirectory
+  ): Option[PrintWriter] = {
     MetalsLogger.redirectSystemOut(globalLog)
-    setupTracePrinter(protocolName)
+    setupTracePrinter(protocolName, workspace)
   }
 
   def globalLog: AbsolutePath = globalDirectory.resolve("global.log")
 
-  def protocolTracePath(protocolName: String): AbsolutePath = {
-    val traceFilename = s"${protocolName.toLowerCase}.trace.json"
-    globalDirectory.resolve(traceFilename)
-  }
-
-  def setupTracePrinter(protocolName: String): PrintWriter = {
-    val tracePath = protocolTracePath(protocolName)
+  def setupTracePrinter(
+      protocolName: String,
+      workspace: AbsolutePath
+  ): Option[PrintWriter] = {
+    val tracePath = protocolTracePath(protocolName, workspace)
     val path = tracePath.toString()
     if (tracePath.isFile) {
       scribe.info(s"tracing is enabled: $path")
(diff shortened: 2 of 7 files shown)


metals/src/main/scala/scala/meta/internal/metals/Trace.scala
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * Returns a printer to trace JSON messages if the user opts into it.
+   */
+  def setup(protocolName: String): Option[PrintWriter] = {
+    MetalsLogger.redirectSystemOut(globalLog)
+    setupTracePrinter(protocolName)
+  }
+
+  def globalLog: AbsolutePath =
+    globalDirectoryOpt.getOrElse(localDirectory).resolve("global.log")
```
kpodsiad, line 30: I wonder if `globalLog` is necessary, it doesn't contain any useful information (only a few logs from the startup)

kpodsiad:
After code review I'll change also [documentation](https://scalameta.org/metals/docs/contributors/getting-started/#json-rpc-trace)

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -141,7 +143,7 @@ class MetalsLanguageServer(
 
   private val fingerprints = new MutableMd5Fingerprints
   private val mtags = new Mtags
-  var workspace: AbsolutePath = _
+  var workspace: Workspace = _
```
tgodzik, line 146: Do we need to change AbsolutePath to Workspace? That causes a bunch of unrelated changes

kpodsiad: I was thinking about adding and passing the boolean flag `isMetaWorkspace` through e.g. `BspConnector@connect`, `BuildServerConnection@fromSockets` etc. but that information is strictly coupled with the workspace path and separating them may lead to invalid states such as path pointing to the meta workspace and `isMetaWorkspace` being set to false. That's why I decided to do what I did. I agree that it causes a lot of changes, but for me, it seemed as a better option. What do you think?
Why that knowledge is needed at all? It's needed because metals server establishes one connection for project workspace and may establish more meta-connections for sbt meta-workspaces. Those meta connections are created later than workspace one and bsp trace file contains logs only from the last created connection. IMO only logs from the project workspace are valuable and some information about workspace is needed to decide whether setup tracing or not. https://github.com/scalameta/metals/blob/5de8af9708a173215babd55c89f04e4d9b4a6144/metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala#L268

metals/src/main/scala/scala/meta/internal/metals/Trace.scala
tgodzik: I think we could remove it, I never used it

metals/src/main/scala/scala/meta/metals/Main.scala
```diff
@@ -66,6 +66,7 @@ object Main {
         .create()
       val clientProxy = launcher.getRemoteProxy
       server.connectToLanguageClient(clientProxy)
+      scribe.info(s"Metals server started and connected with client")
```
tgodzik, line 69:
```suggestion
      scribe.info(s"Metals server started and connected with the client")
```

metals/src/main/scala/scala/meta/internal/metals/BuildServerConnection.scala
```diff
@@ -265,9 +266,11 @@ object BuildServerConnection {
 
     def setupServer(): Future[LauncherConnection] = {
       connect().map { case conn @ SocketConnection(_, output, input, _, _) =>
-        val tracePrinter = GlobalTrace.setupTracePrinter("BSP")
+        val tracePrinter =
+          if (workspace.isMetaWorkspace) None
+          else Trace.setupTracePrinter("BSP")
```
tgodzik, line 271: Could we instead of adding the new `Workspace` entity, add a second entry here that would be the path. We could then write to `.metals` directory in sbt project if needed. This way we will not override the default BSP connection trace and we can also debug the other connections if we want.

metals/src/main/scala/scala/meta/internal/metals/Trace.scala
```diff
@@ -0,0 +1,107 @@
...
+
+import scala.meta.internal.io.PathIO
+import scala.meta.io.AbsolutePath
+
+import dev.dirs.ProjectDirectories
+
+/**
+ * Manages JSON-RPC tracing of incoming/outgoing messages via BSP and LSP.
+ */
+object Trace {
```
tgodzik, line 19: Could you rename the file using git (I think https://git-scm.com/docs/git-mv should work) so that it's possible to see what is the actual change.

tgodzik (changes requested):
Let's reduce the scope of the changes. I don't think there is a need for new workspace entity.

metals/src/main/scala/scala/meta/internal/metals/Trace.scala
```diff
@@ -14,25 +14,31 @@ import dev.dirs.ProjectDirectories
...
+  }
+
+  def setup(
+      protocolName: String,
+      workspace: AbsolutePath = PathIO.workingDirectory
+  ): Option[PrintWriter] = {
     MetalsLogger.redirectSystemOut(globalLog)
-    setupTracePrinter(protocolName)
+    setupTracePrinter(protocolName, workspace)
   }
```
kpodsiad, line 33: Apparently, this method is needed in order to start the server. Otherwise, it hangs out at accepting socket connection. I don't know why redirecting output stream is so important, but if it works I don't wanna change it :P

tgodzik: Maybe we could redirect it to `.metals/metals.log` ?

kpodsiad: `.metals/metals.log` seems like a good replacement for `global.log` 👍

kpodsiad:
@tgodzik If the code is okay, I'll tweak documentation.

tgodzik (approved):
LGTM!

tgodzik:
If you tweak the docs, do mention that this is working differently from the next version or in current snapshot.

PR #3105, opened 2021-09-06
Fix issue with tmp directory popping up in random places

tgodzik:
Previously, we would try to relatavize a path even if it isn't contained in the workspace. This might cause tmp folder with the created semanticdb popping up in weird places. The reason for that is that we would get a path like `../../../../tmp/metals1945019806267779000.scala` and would resolve it from tmp directory. This sometimes would make the semanticdb file generated for warming up scalafix to show in workspace directory.

Now, we only try to resolve files for organize imports for workspace files and also create the temporary warmup files in `.metals` directory

This is blocking for the next release.


ckipp01 (approved):
LGTM!

dos65 (approved):
:+1:

PR #3107, opened 2021-09-08
add javap/semanticdb file viewer

Arthurm1:
Shows class files and semanticdb files in readable format.

This uses VSCode's Virtual Documents rather than the WebView as https://github.com/scalameta/metals/pull/3063 does for Tasty files.

There are pros and cons for both methods and I'm not sure what's best in this case.

Virtual Documents allow Metals to pass back only the plain text that should appear for the output from javap and metap in response to a command with a URI parameter.
This means no hyperlinks and no formatting, both of which the webview will allow but it does make it easily extendable to handle more file viewers (e.g. scalap?) without managing additional WebViews
Highlighting is done using textmate grammars which are pretty long-winded.  But I'm guessing this method is more portable across non-vscode clients?  Also - I don't see that javap or metap offer an HTML output

[jvm-bytecode-viewer](https://github.com/mnxn/vscode-jvm-bytecode-viewer) has some nice javap grammars which maybe could be used.  For now I've only spent spent 5 mins creating an example one.

https://user-images.githubusercontent.com/5612295/132593299-88fe69d0-73e5-48f1-b5e5-7c61ee093146.mp4

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
@@ -1,0 +1,372 @@
+package scala.meta.internal.metals
+
+import java.io.ByteArrayOutputStream
+import java.io.PrintStream
+import java.net.URI
+import java.net.URLDecoder
+import java.nio.charset.StandardCharsets
+import java.nio.file.Paths
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.util.Properties
+import scala.util.Try
+import scala.util.control.NonFatal
+
+import scala.meta.cli.Reporter
+import scala.meta.internal.builds.ShellRunner
+import scala.meta.internal.io.FileIO
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metap.Main
+import scala.meta.internal.mtags.SemanticdbClasspath
+import scala.meta.io.AbsolutePath
+import scala.meta.metap.Format
+import scala.meta.metap.Settings
+
+import ch.epfl.scala.bsp4j.BuildTargetIdentifier
+
+final case class DecoderResponse(
+    requestedUri: String,
+    value: String,
+    error: String
+)
+final class FileDecoderProvider(
+    workspace: AbsolutePath,
+    compilers: Compilers,
+    buildTargets: BuildTargets,
+    userConfig: () => UserConfiguration,
+    shellRunner: ShellRunner,
+    fileSystemSemanticdbs: FileSystemSemanticdbs
+)(implicit ec: ExecutionContext) {
+
+  private case class PathInfo(
+      targetId: Option[BuildTargetIdentifier],
+      path: AbsolutePath
+  )
+  private case class Finder(findFile: String => Option[PathInfo])
+  private case class Decoder(decode: PathInfo => Future[Option[String]])
+
+  /**
+   * URI format...
+   * metalsDecode:/fileUrl?decoder=decoderName&keyValueArguments
+   * jar:file:///jarPath/jar-sources.jar!/packagedir/file.java
+   *
+   * Examples...
+   * metalsDecode:file:///somePath/someFile.java.javap-verbose?decoder=javap&verbose=true
+   * metalsDecode:file:///somePath/someFile.scala.javap-verbose?decoder=javap&verbose=true
+   * metalsDecode:file:///somePath/someFile.class.javap-verbose?decoder=javap&verbose=true
+   *
+   * metalsDecode:file:///somePath/someFile.java.javap?decoder=java
+   * metalsDecode:file:///somePath/someFile.scala.javap?decoder=javap
+   * metalsDecode:file:///somePath/someFile.class.javap?decoder=javap
+   *
+   * metalsDecode:file:///somePath/someFile.java.semanticdb-compact?decoder=semanticdb&format=compact
+   * metalsDecode:file:///somePath/someFile.java.semanticdb-detailed?decoder=semanticdb
+   *
+   * metalsDecode:file:///somePath/someFile.scala.semanticdb-compact?decoder=semanticdb&format=compact
+   * metalsDecode:file:///somePath/someFile.scala.semanticdb-detailed?decoder=semanticdb
+   *
+   * metalsDecode:file:///somePath/someFile.java.semanticdb.semanticdb-compact?decoder=semanticdb&format=compact
+   * metalsDecode:file:///somePath/someFile.java.semanticdb.semanticdb-detailed?decoder=semanticdb
+   *
+   * metalsDecode:file:///somePath/someFile.scala.semanticdb.semanticdb-compact?decoder=semanticdb&format=compact
+   * metalsDecode:file:///somePath/someFile.scala.semanticdb.semanticdb-detailed?decoder=semanticdb
+   *
+   * metalsDecode:file:///somePath/someFile.scala.tasty-detailed?decoder=tasty
+   * metalsDecode:file:///somePath/someFile.tasty.tasty-detailed?decoder=tasty
+   *
+   * jar:file:///somePath/someFile-sources.jar!/somePackage/someFile.java
+   */
+  def decodedFileContents(uriAsStr: String): Future[DecoderResponse] = {
+    for {
+      check <- Future { getDecodeInfo(uriAsStr) }
+      output <- check match {
+        case None => Future.successful(None)
+        case Some((decoder, path)) => decoder.decode(path)
+      }
+      checkedOutput = output match {
+        case None => errorReponse(uriAsStr)
+        case Some(success) => DecoderResponse(uriAsStr, success, null)
+      }
+    } yield checkedOutput
+  }
+
+  def errorReponse(uriAsStr: String): DecoderResponse =
+    DecoderResponse(uriAsStr, null, errorMessage(uriAsStr))
+
+  private def errorMessage(input: String): String =
+    s"""|$input
+        |
+        |Unexpected uri, Metals accepts ones such as:
+        |
+        |metalsDecode:file:///somedir/someFile.scala.javap-verbose?decoder=javap
+        |
+        |Take a look at scala/meta/internal/metals/FileDecoderProvider for more information.
+        |
+        |Or wait for indexing/compiling to finish and re-try
+        |""".stripMargin
+
+  private def getDecodeInfo(
+      uriAsStr: String
+  ): Option[(Decoder, PathInfo)] = {
+    val uri = Try(URI.create(uriAsStr)).toOption
+    uri.flatMap(uri =>
+      uri.getScheme() match {
+        case "jar" => decodeJar(uri)
+        case "file" => decodeMetalsFile(uri)
+        case "metalsDecode" => getDecodeInfo(uri.getSchemeSpecificPart())
+        case _ => None
+      }
+    )
+  }
+
+  private def decodeJar(uri: URI): Option[(Decoder, PathInfo)] = {
+    Try {
+      // jar file system cannot cope with a heavily encoded uri
+      // hence the roundabout way of creating an AbsolutePath
+      // must have "jar:file:"" instead of "jar:file%3A"
+      val decodedUriStr = URLDecoder.decode(uri.toString(), "UTF-8")
+      val decodedUri = URI.create(decodedUriStr)
+      AbsolutePath(Paths.get(decodedUri))
+    }.toOption
+      .map(path =>
+        (
+          Decoder(path =>
+            Future {
+              Try(FileIO.slurp(path.path, StandardCharsets.UTF_8)).toOption
+            }
+          ),
+          PathInfo(None, path)
+        )
+      )
+  }
+
+  private def decodeMetalsFile(
+      uri: URI
+  ): Option[(Decoder, PathInfo)] = {
+    for {
+      query <- Option(uri.getQuery())
+      paramArr = query.split("&").map(_.split("=", 2))
+      if (paramArr.forall(_.length == 2))
+      params <- Try {
+        paramArr
+          .map(f => f.map(URLDecoder.decode(_, "UTF-8")))
+          .map(f => f(0) -> f(1))
+          .toMap
+      }.toOption
+      if (params.contains("decoder"))
+      (finder, decoder) <- getDecoder(params)
+      fileToDecode <- finder.findFile(uri.getPath())
+    } yield ((decoder, fileToDecode))
+  }
+
+  private def getDecoder(
+      params: Map[String, String]
+  ): Option[(Finder, Decoder)] = {
+    params("decoder") match {
+      case "javap" => Some(getJavapDecoder(params))
+      case "semanticdb" => Some(getSemanticdbDecoder(params))
+      case "tasty" => Some(getTastyDecoder(params))
+      case _ => None
+    }
+  }
+
+  private def boolParam(params: Map[String, String], name: String): Boolean =
+    params.get(name).map(_.toBoolean).getOrElse(false)
+
+  private def toFile(
+      uriPath: String,
+      suffixToRemove: String
+  ): Option[AbsolutePath] = {
+    Try {
+      s"file://${uriPath}".stripSuffix(suffixToRemove).toAbsolutePath
+    }.toOption.filter(_.exists)
+  }
+
+  private def getJavapDecoder(
+      params: Map[String, String]
+  ): (Finder, Decoder) = {
+    val verbose = boolParam(params, "verbose")
+    val suffix = if (verbose) ".javap-verbose" else ".javap"
+    val finder = Finder(uriPath =>
+      for {
+        path <- toFile(uriPath, suffix)
+        classesPath <-
+          if (path.isScalaOrJava) findClassesDirFileFromSource(path, "class")
+          else Some(PathInfo(None, path))
+      } yield classesPath
+    )
+    val decoder = Decoder(decodeJavapFromClassFile(_, verbose))
+    (finder, decoder)
+  }
+
+  private def getSemanticdbDecoder(
+      params: Map[String, String]
+  ): (Finder, Decoder) = {
+    val format = params
+      .get("format")
+      .map(_.toUpperCase match {
+        case "DETAILED" => Format.Detailed
+        case "COMPACT" => Format.Compact
+        case "PROTO" => Format.Proto
+      })
+      .getOrElse(Format.Detailed)
+    val suffix = format match {
+      case Format.Detailed => ".semanticdb-detailed"
+      case Format.Compact => ".semanticdb-compact"
+      case Format.Proto => ".semanticdb-proto"
+    }
+    val finder = Finder(uriPath =>
+      for {
+        path <- toFile(uriPath, suffix)
+        semanticPath <-
+          if (path.isScalaOrJava) findSemanticDBFileFromSource(path)
+          else Some(PathInfo(None, path))
+      } yield semanticPath
+    )
+    val decoder = Decoder(decodeFromSemanticDBFile(_, format))
+    (finder, decoder)
+  }
+
+  private def getTastyDecoder(
+      params: Map[String, String]
+  ): (Finder, Decoder) = {
+    val finder = Finder(uriPath =>
+      for {
+        path <- toFile(uriPath, ".tasty-detailed")
+        classesPath <-
+          if (path.isScalaOrJava) findClassesDirFileFromSource(path, "tasty")
+          else findPathInfoFromClassesPath(path)
+      } yield classesPath
+    )
+    val decoder = Decoder(decodeFromTastyFile(_))
+    (finder, decoder)
+  }
+
+  private def findPathInfoFromClassesPath(
+      path: AbsolutePath
+  ): Option[PathInfo] = {
+    val pathInfos = for {
+      scalaTarget <- buildTargets.all
+      classPath = scalaTarget.classDirectory.toAbsolutePath
+      if (path.isInside(classPath))
+    } yield PathInfo(Some(scalaTarget.id), path)
+    pathInfos.toList.headOption
+  }
+
+  private def findClassesDirFileFromSource(
+      sourceFile: AbsolutePath,
+      newExtension: String
+  ): Option[PathInfo] = {
+    for {
+      targetId <- buildTargets.sourceBuildTargets(sourceFile).headOption
+      target <- buildTargets.scalaTarget(targetId)
+      sourceRoot <- buildTargets.inverseSourceItem(sourceFile)
+      classDir = target.classDirectory.toAbsolutePath
+      oldExtension = sourceFile.extension
+      relativePath = sourceFile
+        .toRelative(sourceRoot)
+        .resolveSibling(_.stripSuffix(oldExtension) + newExtension)
+    } yield PathInfo(Some(targetId), classDir.resolve(relativePath))
+  }
+  private def findSemanticDBFileFromSource(
+      sourceFile: AbsolutePath
+  ): Option[PathInfo] = {
+    for {
+      targetId <- buildTargets.inverseSources(sourceFile)
+      target <- buildTargets.scalaTarget(targetId)
+      sourceRoot <- buildTargets.workspaceDirectory(targetId)
+      targetRoot = target.targetroot
+      relativePath = SemanticdbClasspath.fromScala(
+        sourceFile.toRelative(sourceRoot.dealias)
+      )
+      foundSemanticDbPath <- fileSystemSemanticdbs.findSemanticDb(
+        relativePath,
+        targetRoot,
+        sourceFile,
+        workspace
+      )
+    } yield PathInfo(Some(targetId), foundSemanticDbPath.path)
+  }
+
+  private def decodeJavapFromClassFile(
+      pathInfo: PathInfo,
+      verbose: Boolean
+  ): Future[Option[String]] = {
+    try {
+      val args = if (verbose) List("-verbose") else Nil
+      val sb = new StringBuilder()
+      shellRunner
+        .run(
+          "Decode using javap",
+          JavaBinary(userConfig().javaHome, "javap") :: args ::: List(
+            pathInfo.path.filename
+          ),
+          pathInfo.path.parent,
+          redirectErrorOutput = true,
+          Map.empty,
+          s => {
+            sb.append(s)
+            sb.append(Properties.lineSeparator)
+          },
+          s => (),
+          propagateError = true,
+          logInfo = false
+        )
+        .map(_ => Some(sb.toString))
+    } catch {
+      case NonFatal(_) => Future.successful(None)
+    }
+  }
+
+  private def decodeFromSemanticDBFile(
+      pathInfo: PathInfo,
+      format: Format
+  ): Future[Option[String]] = {
+    Future {
+      Try {
+        val out = new ByteArrayOutputStream()
+        val err = new ByteArrayOutputStream()
+        val psOut = new PrintStream(out)
+        val psErr = new PrintStream(err)
+        try {
+          val reporter =
+            Reporter().withOut(psOut).withErr(psErr)
+          val settings =
+            Settings()
+              .withPaths(List(pathInfo.path.toNIO))
+              .withFormat(format)
+          val main = new Main(settings, reporter)
+          main.process()
+          val output = new String(out.toByteArray);
+          val error = new String(err.toByteArray);
+          if (error.isEmpty)
+            output
+          else
+            error
+        } finally {
+          psOut.close()
+          psErr.close()
+        }
+      }.toOption
+    }
+  }
+
+  private def decodeFromTastyFile(
+      pathInfo: PathInfo
+  ): Future[Option[String]] = {
+    for {
+      pc <- Future {
+        for {
+          targetId <- pathInfo.targetId
+          pc <- compilers.loadCompiler(targetId)
+        } yield pc
+      }
+      output <- pc match {
+        case None => Future.successful(None)
+        case Some(pc) =>
+          pc.getTasty(pathInfo.path.toURI, false, false).asScala.map(Some(_))
+      }
+    } yield output
+  }
+}
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -251,6 +251,7 @@ class MetalsLanguageServer(
   private var symbolSearch: MetalsSymbolSearch = _
   private var compilers: Compilers = _
   private var scalafixProvider: ScalafixProvider = _
+  private var fileDecoderProvider: FileDecoderProvider = _
   private var workspaceReload: WorkspaceReload = _
   private var buildToolSelector: BuildToolSelector = _
   def loadedPresentationCompilerCount(): Int =
@@ -673,6 +674,14 @@ class MetalsLanguageServer(
           () => httpServer,
           tables,
           clientConfig
+        )
+        fileDecoderProvider = new FileDecoderProvider(
+          workspace,
+          compilers,
+          buildTargets,
+          () => userConfig,
+          shellRunner,
+          fileSystemSemanticdbs
         )
         popupChoiceReset = new PopupChoiceReset(
           workspace,
@@ -1614,6 +1623,15 @@ class MetalsLanguageServer(
         quickConnectToBuildServer().asJavaObject
       case ServerCommands.DisconnectBuildServer() =>
         disconnectOldBuildServer().asJavaObject
+      case ServerCommands.DecodeFile() =>
+        val argsMaybe = Option(params.getArguments())
+        (argsMaybe.flatMap(_.asScala.headOption) match {
+          case Some(arg: JsonPrimitive) =>
+            val uri = arg.getAsString()
+            fileDecoderProvider.decodedFileContents(uri)
+          case _ =>
+            Future { fileDecoderProvider.errorReponse(s"$params") }
+        }).asJavaObject
       case ServerCommands.RunDoctor() =>
         Future {
           doctor.executeRunDoctor()
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -57,6 +57,20 @@ object ServerCommands {
        |""".stripMargin
   )
 
+  /** Decode a file e.g. javap, semanticdb */
+  val DecodeFile = new Command(
+    "file-decode",
+    "Decode file",
+    """|Decode a file into a human readable format.
+       |
+       |Compilation involves various binary files that can't be read directly
+       |in a text editor so they need to be decoded into a human readable format.
+       |Examples include `.class` and `.semanticdb`.
+       |""".stripMargin,
+    """|[uri], uri of the file with any parameters required for decoding.
+       |""".stripMargin
+  )
+
   val RunDoctor = new Command(
     "doctor-run",
     "Run doctor",
@@ -440,6 +454,7 @@ object ServerCommands {
       ResetChoicePopup,
       RestartBuildServer,
       RunDoctor,
+      DecodeFile,
       ScanWorkspaceSources,
       StartAmmoniteBuildServer,
       StartDebugAdapter,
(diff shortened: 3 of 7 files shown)


tgodzik (commented):
Great idea! I added some comments and will comment some more on the VS Code plugin.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,263 @@
...
+    params.get(name).map(_.toBoolean).getOrElse(false)
+  }
+
+  private def decodeFileContents(
+      filePath: AbsolutePath,
+      params: Map[String, String]
+  ): Future[String] = {
+    params("decoder") match {
+      case "javap" =>
+        decodeJavap(filePath, getBoolParam("verbose", params))
```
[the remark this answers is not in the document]
Arthurm1: I'll remove non-verbose

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,263 @@
...
+      case "semanticdb" =>
+        val format = params
+          .get("format")
+          .map(_.toUpperCase)
+          .collect {
+            case "DETAILED" => Format.Detailed
+            case "COMPACT" => Format.Compact
```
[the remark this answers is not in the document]
Arthurm1: Since it's fairly simple to add more options I'll take the compact out.  The VSCode file-explorer context menu is already getting busy.  Maybe spin off a Metals sub-menu at one point?

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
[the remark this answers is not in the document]
kpodsiad: I'm not sure about removing the non-verbose option. I took a look at [jvm-bytecode-viewer](https://github.com/mnxn/vscode-jvm-bytecode-viewer) to see the verbose output and it's significantly larger than normal mode. If someone isn't master in bytecode reading it could be overwhelming at the beginning.
Maybe non-verbose options could be commands and in the file-explorer context menu will have only verbose ones?

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+  ) extends ErrorWithException(s"Error decoding file `$path`", ex)
+
+  private case class Decoder(
+      findFile: AbsolutePath => Option[AbsolutePath],
+      decode: AbsolutePath => Future[Either[Error, String]]
+  )
+
+  /**
+   * URI format...
+   * metalsDecodeFile:/fileUrl?decoder=decoderName&keyValueArguments&originalURI=originalFileURI
```
kpodsiad, line 79: If verbose option is gone are `keyValueArguments` still necessary?

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+      target <- buildTargets.scalaTarget(targetId)
+      sourceRoot <- buildTargets.inverseSourceItem(sourceFile)
+      classDir = target.classDirectory.toAbsolutePath
+      oldExtension = sourceFile.extension
+      relativePath = sourceFile
+        .toRelative(sourceRoot)
+        .resolveSibling(_.stripSuffix(oldExtension) + newExtension)
+    } yield classDir.resolve(relativePath)
+  }
+  private def findSemanticDBFileFromSource(
```
tanishiking, line 207: I'm wondering maybe we can re-use the `FileSystemSemanticdbs` to find the corresponding semanticdb file from the given file path. It looks like there's some duplication with the implementation in `FileSystemSemanticdbs`, and I think it would be great if we could use the same procedure.

https://github.com/scalameta/metals/blob/ef9cf6462867c1a56d06eb813f2889a864263e14/metals/src/main/scala/scala/meta/internal/metals/FileSystemSemanticdbs.scala#L57

But, this can be done in a separate issue/PR :)

tanishiking (commented):
Awesome! This feature looks useful especially for tooling developers 😍  I just left a nit comment.

kpodsiad (commented):
Outstanding work!

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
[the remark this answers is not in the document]
tgodzik: You might be right @kpodsiad ! I will leave it up to you @Arthurm1

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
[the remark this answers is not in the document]
tgodzik: Sub menu would be cool probably, we could also just put some as separate commands

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+      shellRunner
+        .run(
+          "Decode using javap",
+          List(
+            JavaBinary(userConfig().javaHome, "javap"),
+            "-verbose",
+            filePath.filename
+          ),
+          filePath.parent,
+          true,
```
tgodzik, line 235: Could you use named arguments for the booleans?

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+              |${sw.toString}""".stripMargin
+        case Left(e) =>
+          scribe.error(e.message)
+          e.message
+        case Right(success) => success
+      }
+    } yield checkedOutput
+  }
+
+  private def getDecodeInfo(
```
tgodzik, line 111: Could we simplify this a bit and could we instead of having a separate error for each, just show one error at the end if anything is amiss.

It could show something like:
```
Unexpected uri, Metals accepts ones such as:
metalsDecodeFile:/somedir/someFile.java.javap.verbose?decoder=javap&originalURI=file:///somedir/someFile.java
Take a look at scala/meta/internal/metals/FileDecoderProvider for more information.
```
This will only show up for people creating the plugins, so that should be alright.

Alternatively, we can link to the section about the command in https://scalameta.org/metals/docs/integrations/new-editor

We  could simplify it a bit in the for comprehension, added some examples for that.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+      uriAsStr: String
+  ): Either[Error, (Decoder, AbsolutePath)] = {
+    for {
+      uri <- Try(URI.create(uriAsStr)) match {
+        case Failure(ex) => Left(InvalidURI(uriAsStr, ex))
+        case Success(uri) => Right(uri)
+      }
+      _ <-
+        if (uri.getScheme != "metalsDecodeFile") Left(InvalidScheme(uri))
+        else Right(uri)
```
tgodzik, line 121:
```suggestion
      if uri.getScheme == "metalsDecodeFile"
```

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  private def getDecodeInfo(
+      uriAsStr: String
+  ): Either[Error, (Decoder, AbsolutePath)] = {
+    for {
+      uri <- Try(URI.create(uriAsStr)) match {
+        case Failure(ex) => Left(InvalidURI(uriAsStr, ex))
+        case Success(uri) => Right(uri)
+      }
```
tgodzik, line 118:
```suggestion
      uri <- Try(URI.create(uriAsStr)).toEither
```

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+      _ <-
+        if (uri.getScheme != "metalsDecodeFile") Left(InvalidScheme(uri))
+        else Right(uri)
+      paramArr <- {
+        val paramArr = uri.getQuery.split("&").map(_.split("=", 2))
+        if (paramArr.exists(_.length != 2)) Left(InvalidQuery(uri))
+        else Right(paramArr)
+      }
```
tgodzik, line 126:
```suggestion
      paramArr  = uri.getQuery.split("&").map(_.split("=", 2))
      if paramArr.exists(_.length != 2)
```

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+      params <- Try {
+        paramArr
+          .map(f => f.map(URLDecoder.decode(_, StandardCharsets.UTF_8)))
+          .map(f => f(0) -> f(1))
+          .toMap
+      } match {
+        case Failure(ex) => Left(InvalidQueryParams(paramArr, ex))
+        case Success(params) => Right(params)
+      }
```
tgodzik, line 135:
```suggestion
      }.toEither
```

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,285 @@
...
+
+  /**
+   * URI format...
+   * metalsDecodeFile:/fileUrl?decoder=decoderName&keyValueArguments&originalURI=originalFileURI
+   *
+   * Examples...
+   * metalsDecodeFile:/somedir/someFile.java.javap.verbose?decoder=javap&originalURI=file:///somedir/someFile.java
+   * metalsDecodeFile:/somedir/someFile.class.javap.verbose?decoder=javap&originalURI=file:///somedir/someFile.class
+   * metalsDecodeFile:/somedir/someFile.scala?decoder=semanticdb&originalURI=file:///somedir/someFile.scala
+   * metalsDecodeFile:/somedir/someFile.scala.semanticdb?decoder=semanticdb&originalURI=file:///somedir/someFile.scala.semanticdb
```
tgodzik, line 85: Do we need originalURI?
We could just get the relevant parts from the existing uri. Set query to none and scheme to file.

tgodzik (changes requested):
Looks pretty cool! I think we can simplify it a bit.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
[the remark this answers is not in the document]
Arthurm1: I'll add non-verbose back in but only as a "show active file javap" command.  The context menu is getting big.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
Arthurm1: I was wondering about the potential for a file in a jar.  isn't the uri for that something like "jar:file:/path/file.class"?  And I saw some chatter on Discord about using read-only files which VirtualDocuments provides so maybe this could be extended to navigate to jar sources and decompile classes in jars that don't have sources?

tgodzik: We never show class files, so currently we would never send a jar URI. But potentially we could try returning it in the future when sources are not available and decompile it, but in that case we would don't really need the source file, since we can assume class files can be decompiled with javap

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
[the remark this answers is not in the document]
Arthurm1: Actually it doesn't matter now we have a sub-menu so I'll keep both on the sub-menu

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
Arthurm1: Verbose is now making a comeback

Arthurm1:
Hopefully this addresses everything:
1. Simplified error handling to replace with single error message
2. Re-instated verbose as there's a submenu now so it's not too busy
3. Used `FileSystemSemanticdbs#findSemanticDb` - there's not much overlap but I guess it copes with symlinks
4. Got rid of `originalURI` param.  Decoders can work this transform out on their own.
5. Named boolean params.

Arthurm1:
I've extended this to cope with jars.

Opening any standard jar url in vscode e.g. `jar:file:///jarPath/jar-sources.jar!/packagedir/file.java` will cause a VirtualDocument of the contents to open in VSCode...

https://user-images.githubusercontent.com/5612295/132927082-a74e203d-a58c-4e48-9d80-585552e53d88.mp4

As a test - in the example above I've changed `DefinitionProvider#fromSymbol` to just return the URI of the jarPath rather than extracting the contents to `.metals/readonly/dependencies/jarname/file.java` and returning that URI.

I've not committed that change as I think it will break existing non VSCode clients.

I'm thinking it might be good for Metals to migrate away from using the `readonly` area to Virtual Documents but I don't know the implications.

gabro:
@Arthurm1 that looks great!
I think a viable solution might be to declare a custom capability, so that clients can specify whether they support virtual documents, and Metals can fallback to using the current .readonly trick in case they don't

tgodzik (approved):
LGTM! This is a lot great work!

tgodzik:
Let me know when you want to merge it or should it already not be a draft?

Arthurm1:
I'm just adding custom config so that clients can specify whether they support Virtual Documents and also a slight change to the way the scheme is specified so hold off for a bit please.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1606,6 +1614,17 @@ class MetalsLanguageServer(
         quickConnectToBuildServer().asJavaObject
       case ServerCommands.DisconnectBuildServer() =>
         disconnectOldBuildServer().asJavaObject
+      case ServerCommands.DecodeFile() =>
+        val argsMaybe = Option(params.getArguments())
+        argsMaybe.flatMap(_.asScala.headOption) match {
+          case Some(arg: JsonPrimitive) =>
+            val uri = arg.getAsString()
```
ckipp01, line 1621: Out of curiosity, what made you choose to go with a `uri` here? From a pure client implementation perspective I find this a bit more confusing that having this just be a json object that we pass in. We also don't do this `uri` format in any other place as far as I'm aware so we're sort of breaking pattern here. I'm not against this, more just curious why you chose this and if there is an advantage that I'm not seeing.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1606,6 +1614,17 @@ class MetalsLanguageServer(
...
+            fileDecoderProvider.decodedFileContents(uri).asJavaObject
+          case _ =>
+            val error = s"Invalid DecodeFile arguments $params"
+            scribe.error(error)
+            Future { error }.asJavaObject
```
ckipp01, line 1626: I'd recommend handling this a little different. Since no matter what this seems to just be returning a string, for a client there won't be a good way to determine if they are getting back a valid string, or just this string which is really an error. Maybe instead of just returning the error string here we could show a message to the user letting them know this and don't return the error string as a valid response like we are doing now.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -56,6 +56,20 @@ object ServerCommands {
...
   )
 
+  /** Decode a file e.g. javap, semanticdb */
+  val DecodeFile = new Command(
+    "file-decode",
+    "Decode file",
+    """|Decode a file into a human readable format.
+       |
+       |Compilation involves various binary files that can't be read directly
+       |in a text editor so need to be decoded into a human readable format.
```
ckipp01, line 66:
```suggestion
       |in a text editor so they need to be decoded into a human readable format.
```

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,277 @@
...
+   *
+   * Examples...
+   * metalsDecodeFile:/somedir/someFile.java.javap.verbose?decoder=javap
+   * metalsDecodeFile:/somedir/someFile.class.javap?decoder=javap&verbose=false
+   * metalsDecodeFile:/somedir/someFile.scala?decoder=semanticdb
+   * metalsDecodeFile:/somedir/someFile.scala.semanticdb?decoder=semanticdb
+   *
+   * errors get logged and sent back as the text contents
+   */
+  def decodedFileContents(uriAsStr: String): Future[String] = {
```
ckipp01, line 49: I mentioned this in the TASTy pr as well, but I don't love just returning a `Future[String]` here as it makes it more difficult on the client side to know if you are truly getting the string you want, or just some other random error string. It ends up forcing the client to do odd things with string matching to see what they are getting back. I'd much rather this be a more structured return type so it's easy for clients to immediately know if what they are getting back is valid or not instead of no matter what getting just a `String`.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,277 @@
...
+      case "javap" => Some(getJavapDecoder(params))
+      case "semanticdb" => Some(getSemanticdbDecoder(params))
+      case _ => None
+    }
+  }
+
+  private def boolParam(params: Map[String, String], name: String): Boolean =
+    params.get(name).map(_.toBoolean).getOrElse(false)
+
+  private def toFile(
```
ckipp01, line 126: I don't want to push on this too much seeing that you're as far along as you are on this, but again I wonder about the choice to use a URI here instead of a json object where we could easily specify the decoder type etc and then just pass in the file uri. Then we wouldn't need to worry about doing these types of things like creating a file uri out of the uri that we already passed in. Again, I may totally be missing the advantage of using a uri over an object here.

ckipp01 (commented):
This will be such a great addition @Arthurm1, great job on this 🚀 

I do have just a couple questions on your usage of `uri` over passing in an object and a couple other small notes that I think may help clients better handle the cases where an error string is being returned instead of valid output. Again, great job.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
Arthurm1: You mean instead of passing URI as a String, you'd want something like this?...
```
{
  "uri" : "someURI"
}
```

I can change the format - the URI is the only data you have to hand when implementing a Virtual Document so I just passed it on

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
Arthurm1: I could destructure the data so instead of...

`metalsDecodeFile:file:///somePath/someFile.java.javap.verbose?decoder=javap&verbose=true`

you'd have

```
{
  "file": "file:///somePath/someFile.java.javap.verbose",
  "decoder": "javap",
  "verbose": "true"
}
```

But the only way I can trigger a VirtualDocument is by creating this URI on VSCode in the first place.  I don't know how other clients implement this kind of functionality but in VSCode it can only be done in response to a URI and all the data required has to be in that URI.
I guess if other clients handled VirtualDocuments differently then it might make more sense to move the URI interpretation code from Metals into the VSCode extension.

I saw [this NeoVim code](https://github.com/neovim/nvim-lspconfig/blob/59bc16aa2c9cfc6f8c2a67ae01635eeea1d43943/lua/lspconfig/denols.lua) and it seems as though VIM also responds to a URI to trigger a Virtual Document but my knowledge of VI extends as far as `:wq` so I don't really know what this code does :-)

I guess I can't really make a decision on this without knowing more about other clients - I'd certainly change it if it made it easier for non-VSCode clients
I also wonder if I'll need to implement a version of this PR that writes the files to the `.metals/readonly` area to cope with other clients

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
Arthurm1: Sure.  Something like below?
```
{
  "output": "Some file text"
}
```
and
```
{
  "error": "Some error message"
}
```
Or do you have an example already that I can work off?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
ckipp01: > the URI is the only data you have to hand when implementing a Virtual Document so I just passed it on

Ahh ok, this explains it then. I didn't know this. I'd say don't change it then. I was thinking more like the example you showed in the other comment down below:

```{
  "file": "file:///somePath/someFile.java.javap.verbose",
  "decoder": "javap",
  "verbose": "true"
}
```

However, if that makes it more difficult for VS Code and virtual documents, then for sure don't change it. 

I typically default to do what is best for VS Code since it has the highest numbers, but do it in a way that still works in other clients. This will still work fine in other clients, was more just curious about the choice. 👍🏼 I'd say leave this as is.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
ckipp01: Yea it can be pretty simple in the case:

```
interface GoodName {
  error?: String
  body?: String
}
```

Or is someone else has a better interface. But yea, I think this will be nicer for clients to be able to differentiate between these cases.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
ckipp01: > But the only way I can trigger a VirtualDocument is by creating this URI on VSCode in the first place

Yea to reiterate what I said above, I wasn't aware of this, so don't worry about it.

> I also wonder if I'll need to implement a version of this PR that writes the files to the .metals/readonly area to cope with other clients

I actually don't think this is necessary unless you want to persist or cache the output, which I don't think is that necessary. I have a POC working for this locally with `nvim-metals` and I can push it up tomorrow to give you an idea of how another client will handle it. However, it's pretty simple, I just take the long string that is returned, split it into lines and then fill a scratch buffer with it. 



> I saw this NeoVim code and it seems as though VIM also responds to a URI to trigger a Virtual Document but my knowledge of VI extends as far as :wq so I don't really know what this code does :-)

Ah so this is a great example, what they do here:
https://github.com/neovim/nvim-lspconfig/blob/59bc16aa2c9cfc6f8c2a67ae01635eeea1d43943/lua/lspconfig/denols.lua#L50-L54 is more or less exactly how I handle it. It's getting a "virtual document", but just treating it as a scratch buffer. Ha, this makes me feel like I'm doing in the right way seeing it here like this.

Thanks a lot for explaining the reasoning and such. I'd say carry on with the URI.

Arthurm1: Looks great.

I might have to change the uri syntax slightly (prefixing: `metalsDecodeFile:file:///` rather than replacing: `file:///` with `metalsDecodeFile:///`

Are you using a special highlighter/syntax/grammar?  Or is that the `scala` one applied to the semanticdb output.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -0,0 +1,277 @@
...
+  private case class Decoder(decode: AbsolutePath => Future[Option[String]])
+
+  /**
+   * URI format...
+   * metalsDecodeFile:/fileUrl?decoder=decoderName&keyValueArguments
+   * jar:file:///jarPath/jar-sources.jar!/packagedir/file.java
+   *
+   * Examples...
+   * metalsDecodeFile:/somedir/someFile.java.javap.verbose?decoder=javap
+   * metalsDecodeFile:/somedir/someFile.class.javap?decoder=javap&verbose=false
```
ckipp01, line 43: I'm not fully sure what's happening with these, but I'm unable to get this to work. I keep getting:

```
metalsDecodeFile:/Users/ckipp/Documents/sanity/src/main/scala/Main.java.javap.verbose?decoder=javap
Unexpected uri, Metals accepts ones such as:
metalsDecodeFile:/somedir/someFile.java.javap.verbose?decoder=javap
Take a look at scala/meta/internal/metals/FileDecoderProvider for more information.
```
Just to verify, how come when we do semanticdb we do `someFile.scala?decoder=semanticdb` and then here we do `someFile.java.javap?decoder=javap`. I would have assumed this should have been `someFile.scala?decoder-javap`.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
ckipp01: > Are you using a special highlighter/syntax/grammar? Or is that the scala one applied to the semanticdb output.

Yea I just quickly set the syntax to `scala`.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
Arthurm1: The URIs are from me mucking about with Java files.

If you have a scala source then use `someFile.scala.javap`

It's also possible to run semanticdb and javap decoders directly on the semanticdb and class file URIs in the `.bloop/**` dirs

ckipp01: 🤔 So I assumed that it was `.scala` at first and no matter what I try I keep getting:

```
Error: class not found: Simple.class
```

No matter what file I'm trying it in. What I end up sending in is:

```
[Trace - 08:37:51 AM] Received request 'workspace/executeCommand - (13)'
Params: {
  "command": "metals.file-decode",
  "arguments": [
    "metalsDecodeFile:///Users/ckipp/Documents/sanity/src/main/scala/Simple.scala.javap?decoder\u003djavap\u0026verbose\u003dfalse"
  ]
}
```

Which should be fine no?

Arthurm1: It should be.  Is there a class file present? I imagine there should be if the semanticdb one is present.

Something like: `./bloop/some bsp classes path/sanity/Simple.class`?

You should be able to run javap directly on that class file as well as the source file e.g.
`metalsDecodeFile:///Users/ckipp/Documents/sanity/.bloop/XXXX/Simple.class.javap?decoder\u003djavap\u0026verbose\u003dfalse`
If that works then for some reason it's not getting the build target correctly and then working out where the class file should be

I'm surprised you're getting that error - I thought all errors were intercepted to just return invalid URI info.

ckipp01: Yea it's odd. It works for me when I view semanticdb, but not javap.



> You should be able to run javap directly on that class file as well as the source file e.g.
metalsDecodeFile:///Users/ckipp/Documents/sanity/.bloop/XXXX/Simple.class.javap?decoder\u003djavap\u0026verbose\u003dfalse
If that works then for some reason it's not getting the build target correctly and then working out where the class file should be

This is much trickier in other clients, and I haven't actually looked into this yet. `nvim-metals` works by detecting the file type and then deciding whether or not it should "attach" the LSP server to that buffer. So in this situation `.class` files aren't something that are recognized, so if you open them Metals isn't attatched. It's something I'll need to figure out.

Arthurm1: Ah - I think this is actually the output of `javap`.  can you try javap from the CLI on that class file?

ckipp01: Screenshot 2021-09-14 at 14 24 51

Works fine from the CLI

Arthurm1: It may be that I need to specify the classpath in the CLI even though it runs from that folder.
Could you try `javap -cp . -verbose Simple.class`

ckipp01: > It may be that I need to specify the classpath in the CLI even though it runs from that folder.
Could you try javap -cp . -verbose Simple.class

This shows everything

```
~/Documents/scala-workspace/sanity/.bloop/foo/bloop-bsp-clients-classes/classes-Metals-oOUyR9noRRmJGfmsWpXaXg==/sanity
❯ javap -cp . -verbose Simple.class
Classfile /Users/ckipp/Documents/scala-workspace/sanity/.bloop/foo/bloop-bsp-clients-classes/classes-Metals-oOUyR9noRRmJGfmsWpXaXg==/sanity/Simple.class
  Last modified 14 Sep 2021; size 599 bytes
  MD5 checksum 270ecda1de731925fae6153cc6190c3e
  Compiled from "Simple.scala"
public final class sanity.Simple
  minor version: 0
  major version: 52
  flags: (0x0031) ACC_PUBLIC, ACC_FINAL, ACC_SUPER
  this_class: #2                          // sanity/Simple
  super_class: #4                         // java/lang/Object
  interfaces: 0, fields: 0, methods: 1, attributes: 3
Constant pool:
   #1 = Utf8               sanity/Simple
   #2 = Class              #1             // sanity/Simple
   #3 = Utf8               java/lang/Object
   #4 = Class              #3             // java/lang/Object
   #5 = Utf8               Simple.scala
   #6 = Utf8               Lscala/reflect/ScalaSignature;
   #7 = Utf8               bytes
   #8 = Utf8               \u0006\u0001y9Q!\u0002\u0004\t\u0002%1Qa\u0003\u0004\t\u00021AQaE\u0001\u0005\u0002QAq!F\u0001C\u0002\u0013\u0005a\u0003\u0003\u0004\u001e\u0003\u0001\u0006IaF\u0001\u0007\'&l\u0007\u000f\\3\u000b\u0003\u001d\taa]1oSRL8\u0001\u0001\t\u0003\u0015\u0005i\u0011A\u0002\u0002\u0007\'&l\u0007\u000f\\3\u0014\u0005\u0005i\u0001C\u0001\b\u0012\u001b\u0005y!\"\u0001\t\u0002\u000bM\u001c\u0017\r\\1\n\u0005Iy!AB!osJ+g-\u0001\u0004=S:LGO\u0010\u000b\u0002\u0013\u0005)Ao\u001c;bYV\tq\u0003E\u0002\u000f1iI!!G\b\u0003\r=\u0003H/[8o!\tq1$\u0003\u0002\u001d\u001f\t\u0019\u0011J\u001c;\u0002\rQ|G/\u00197!\u0001
   #9 = Utf8               total
  #10 = Utf8               ()Lscala/Option;
  #11 = Utf8               ()Lscala/Option<Ljava/lang/Object;>;
  #12 = Utf8               sanity/Simple$
  #13 = Class              #12            // sanity/Simple$
  #14 = Utf8               MODULE$
  #15 = Utf8               Lsanity/Simple$;
  #16 = NameAndType        #14:#15        // MODULE$:Lsanity/Simple$;
  #17 = Fieldref           #13.#16        // sanity/Simple$.MODULE$:Lsanity/Simple$;
  #18 = NameAndType        #9:#10         // total:()Lscala/Option;
  #19 = Methodref          #13.#18        // sanity/Simple$.total:()Lscala/Option;
  #20 = Utf8               Code
  #21 = Utf8               Signature
  #22 = Utf8               SourceFile
  #23 = Utf8               RuntimeVisibleAnnotations
  #24 = Utf8               ScalaSig
{
  public static scala.Option<java.lang.Object> total();
    descriptor: ()Lscala/Option;
    flags: (0x0009) ACC_PUBLIC, ACC_STATIC
    Code:
      stack=1, locals=0, args_size=0
         0: getstatic     #17                 // Field sanity/Simple$.MODULE$:Lsanity/Simple$;
         3: invokevirtual #19                 // Method sanity/Simple$.total:()Lscala/Option;
         6: areturn
    Signature: #11                          // ()Lscala/Option<Ljava/lang/Object;>;
}
SourceFile: "Simple.scala"
RuntimeVisibleAnnotations:
  0: #6(#7=s#8)
    scala.reflect.ScalaSignature(
      bytes="\u0006\u0001y9Q!\u0002\u0004\t\u0002%1Qa\u0003\u0004\t\u00021AQaE\u0001\u0005\u0002QAq!F\u0001C\u0002\u0013\u0005a\u0003\u0003\u0004\u001e\u0003\u0001\u0006IaF\u0001\u0007\'&l\u0007\u000f\\3\u000b\u0003\u001d\taa]1oSRL8\u0001\u0001\t\u0003\u0015\u0005i\u0011A\u0002\u0002\u0007\'&l\u0007\u000f\\3\u0014\u0005\u0005i\u0001C\u0001\b\u0012\u001b\u0005y!\"\u0001\t\u0002\u000bM\u001c\u0017\r\\1\n\u0005Iy!AB!osJ+g-\u0001\u0004=S:LGO\u0010\u000b\u0002\u0013\u0005)Ao\u001c;bYV\tq\u0003E\u0002\u000f1iI!!G\b\u0003\r=\u0003H/[8o!\tq1$\u0003\u0002\u001d\u001f\t\u0019\u0011J\u001c;\u0002\rQ|G/\u00197!\u0001"
    )
  ScalaSig: length = 0x3 (unknown attribute)
   05 00 00
``

Comment (Arthurm1) [reply metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala:43]: I'm pretty sure the `Error: class not found: Main.class` is the output of `javap`.

Could you log what is going on in `FileDecoderProvider#decodeJavapFromClassFile`
Specifically what the `filePath.filename` and `filePath.parent` is.
I wonder if this is a symlink thing.  I'm on Windows + WSL2 but maybe it's not handling something on the Mac

**Comment by Arthurm1** [comment]
I've changed the URIs to follow the same method as `jar:` so instead of replacing `file:/` with `metalsDecodeFile:` I now prefix the URI with `metalsDecode`: `metalsDecode:File:///`

The suffixes have also changed from `javap.verbose` to `javap-verbose` so VSCode can enable menus based on the full suffix.

There's a full list of examples in `FileDecoderProvider`

Originally I was going to switch over source jar navigation to use virtual documents.  This works well but there are a lot of places in the code that transform URI<->AbsolutePath and they need to be changed slightly so I'll do that in a separate PR.

Metals now returns the message and error separately as JSON `{ "value": "some text", "error": "" }`.  Hopefully this is how you want it @ckipp01

**Review by tgodzik — APPROVED** [review]
Looks good! Pretty exciting to have this functionality especially from the tooling perspective.

**Comment by Arthurm1** [comment]
Added a Tasty file decode option via URI.  It's not linked to any commands in VSCode as the WebView option looks better unless there's a textmate grammar for Tasty.  Thought it might help non-VSCode clients

File: metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala [hunk 18]
```diff
@@ -0,0 +1,362 @@
...
+            sb.append(s)
+            sb.append(Properties.lineSeparator)
+          },
+          s => (),
+          propagateError = true,
+          logInfo = false
+        )
+        .map(_ => Some(sb.toString))
+    } catch {
+      case _: Throwable => Future.successful(None)
```
kpodsiad, line 308: More Scala way to catch exceptions (because of fatal errors and `ControlTrowable`)
```suggestion
      case NonFatal(_) => Future.successful(None)
```

kpodsiad, line 308: More Scala way to catch exceptions (because of fatal errors and `ControlTrowable`)
```suggestion
      case NonFatal(_) => Future.successful(None)
```

kpodsiad (approved):
What a wonderful work, @Arthurm1!

Arthurm1:
Sorry (reviewers) to change this PR again.

I hadn't realised that the JSON encoding is handled for you until I saw @kpodsiad PR (thanks)

I've changed the output to pass a `DecoderResponse` instead of JSON...
```
export interface DecoderResponse {
  requestedUri?: string;
  value?: string;
  error?: string;
}
```

tgodzik (approved):
Looks good to merge! Should we go ahead with it @Arthurm1 ?

Arthurm1:
@tgodzik yes please!  :-)

Issue #3108, opened 2021-09-09
Something is off with the 0.10.6 indent on paste


spangaer:
6 Images say 6000 words right? ;)

![metals_v0 10 6_indent-bug](https://user-images.githubusercontent.com/677147/132642057-d56dfebf-cb0e-4a3c-aa46-0f90928f59bb.gif)

tgodzik:
Thanks for finding this! We will try to fix it as soon as possible. If it's problematic you can disable `Formatting on paste` option in the settings.

spangaer:
Had read that in the release notes indeed, but thanks for reminding.

The only thing I could think though: "So hate optional braces"  🤷‍♂️

tgodzik:
Optional braces are being highly disputed, but this feature should help out also when printing in normal blocks (unless it breaks :fearful: )

dos65:
Hmm. @tgodzik did you reproduce that? I didn't

tgodzik:
Yeah, you need to paste something after a def.

spangaer:
Input source
![image](https://user-images.githubusercontent.com/677147/132648556-a90d34b2-b05a-4a7e-8023-e3c3c4b416da.png)

And Scala 2.12.10, should those make any difference.

dos65:
Still without success even after upgrading vscode to the latest one.
Maybe it somehow depends on platform.
Ok, I'll wait a fix from @tgodzik

PR #3109, opened 2021-09-09
Fix overindenting when the indentation is more than expected

tgodzik:
Fixes https://github.com/scalameta/metals/issues/3108

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
@@ -12,10 +12,11 @@ object IndentOnPaste extends RangeFormatter {
   private val noEndClause = raw"((<!\bend\b\s*?)\b(if|while|for|match|try))"
   private val ifThenClause = raw"(\bif\s+(?!.*?\bthen\b.*?$$)[^\s]*?)"
   private val keywordClause = raw"\b(then|else|do|catch|finally|yield|case)"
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
```diff
@@ -209,7 +209,7 @@ object IndentOnPaste extends RangeFormatter {
       ): String = {
         val identToStart = codeStartPosition(full).getOrElse(0)
 
-        if (identToStart != expectedIdent) {
+        if (identToStart < expectedIdent) {
```
dos65, line 212: That works but I'm thinking maybe it would be more reliable if it won't do anything with indentation if paste changes only one non-empty line?

dos65: Also, we have't covered all cases for `expectedIndent` calculation
In added test-case it's 2 while it should 4 because goes after `(`

tgodzik: > That works but I'm thinking maybe it would be more reliable if it won't do anything with indentation if paste changes only one non-empty line?

Added additional logic for checking that

> Also, we have't covered all cases for expectedIndent calculation
In added test-case it's 2 while it should 4 because goes after (

Good catch, updated the regex and added tests.

dos65 (approved):
Looks good! Thx for fixing this

PR #3112, opened 2021-09-10
Fix process exiting

dos65:
Use `sys.exit(0)` to stop all threads to avoid hanged process.

In some cases (if metals was closed just right after build import/
during import) there is some thread from unknown thread pool that keep
process alive.

I've checked all places in our codebase but haven't found anything.
Probably it starts somwehere in dependency so `sys.exit(0)` is the only
solution here.


tgodzik (approved):
LGTM

PR #3114, opened 2021-09-13
Fix message about unsupported version for RC Scala 3 versions

tgodzik:


ckipp01 (commented):
Looks like there are some valid issues in the Warning suite.

tgodzik:
> Looks like there are some valid issues in the Warning suite.

Looks like the leftover logic from the Scala milestone days :sweat: 

Should be fixed now, let's see

PR #3115, opened 2021-09-13
[Scala3] Improve completions ordering

dos65:
Align completion items ordering with Scala2 implementation

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
@@ -199,7 +199,10 @@ trait Completions { this: MetalsGlobal =>
               if (byFuzzy != 0) byFuzzy
               else {
                 val byIdentifier =
-                  IdentifierComparator.compare(o1.sym.name, o2.sym.name)
+                  IdentifierComparator.compare(
+                    o1.sym.name.decode,
+                    o2.sym.name.decode
+                  )
                 if (byIdentifier != 0) byIdentifier
                 else {
                   val byOwner =
--- a/tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
@@ -16,9 +16,8 @@ class CompletionArgSuite extends BaseCompletionSuite {
         |""".stripMargin,
     """|assertion = : Boolean
        |Main arg
-       |:: scala.collection.immutable
        |""".stripMargin,
-    topLines = Option(3)
+    topLines = Option(2)
   )
 
   check(
@@ -185,9 +184,8 @@ class CompletionArgSuite extends BaseCompletionSuite {
         |""".stripMargin,
     """|end = : Int
        |Main arg9
-       |:: scala.collection.immutable
        |""".stripMargin,
-    topLines = Option(3)
+    topLines = Option(2)
   )
 
   check(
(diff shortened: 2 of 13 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -199,7 +199,10 @@ trait Completions { this: MetalsGlobal =>
               if (byFuzzy != 0) byFuzzy
               else {
                 val byIdentifier =
-                  IdentifierComparator.compare(o1.sym.name, o2.sym.name)
+                  IdentifierComparator.compare(
+                    o1.sym.name.decode,
+                    o2.sym.name.decode
+                  )
```
dos65, line 205: Some changes in test data for Scala2 were triggered by this change.
For example `::` was compared as `$colon$colon` that resulted to the difference in ordering between 2 and 3

dos65:
Now it's finally ready to review.

What was changed:
- named arg less relevevant than other available values
- workspace symbols (that requires auto import) less relevant than scope symbols
- methods are sorted by amount of parameters
- fixed implicit conversion detection

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -1289,13 +1235,80 @@ class CompletionSuite extends BaseCompletionSuite {
...
+        |""".stripMargin,
+    """|fooB: String
+       |fooA: String
+       |""".stripMargin,
+    topLines = Some(2)
+  )
+
+  // issues with scala 3 https://github.com/lampepfl/dotty/pull/13515
+  check(
+    "ordering-4".tag(IgnoreScala3),
```
tgodzik, line 1296: Could we ignore the current Scala 3 versions up to 3.1.0? So that we automatically check it for 3.1.1 ?

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -1289,13 +1235,80 @@ class CompletionSuite extends BaseCompletionSuite {
        |""".stripMargin
   )
 
-  checkItems(
-    "scope-completions-empty-query",
```
tgodzik, line 1293: We remove this test?

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -16,9 +16,8 @@ class CompletionArgSuite extends BaseCompletionSuite {
         |""".stripMargin,
     """|assertion = : Boolean
        |Main arg
-       |:: scala.collection.immutable
        |""".stripMargin,
-    topLines = Option(3)
+    topLines = Option(2)
```
tgodzik, line 20: Why change this?

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -185,9 +184,8 @@ class CompletionArgSuite extends BaseCompletionSuite {
         |""".stripMargin,
     """|end = : Int
        |Main arg9
-       |:: scala.collection.immutable
        |""".stripMargin,
-    topLines = Option(3)
+    topLines = Option(2)
```
tgodzik, line 188: And this?

tgodzik (commented):
Looks good, I just have a couple of questions.

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
dos65: after changing the ordering for scala 2 - `name.decoded` the third symbol come either `::` or `#::` - depends on scala version and that's correct due to collections rework.

I decided that decreasing the number here is better than adding compat section because other values aren't about `ArgSuite`

tgodzik: Ok 👌

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
dos65: I've modified in locally in attempt to fix #3140 but it wasn't a case for quick fix.
Then I realized that this test isn't really valuable as it doesn't check result.

I might return as it is. Or it will be added later with a fix for #3140 and more strict result validation

tgodzik (approved):
LGTM!

PR #3117, opened 2021-09-14
bug report template - downgrade version

dos65:
The latest one is `0.10.6` :sweat_smile:


kpodsiad (approved):
Hah, thanks for catching it!😅

PR #3125, opened 2021-09-15
Separate indentation on paste to a command instead of automatic

tgodzik:
Previously, we will automatically adjust indentation whenever user pasted anything, which might have caused unexpected results. Now, we have a separate command that can be used whenever the user wants to explicitely adjust indentation (for example when pasting from another location with different indentation level). This is done the same in the Python LSP extension.

I will follow up soon with a VS Code PR, but wanted to get this into the next release if possible.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
@@ -105,6 +105,11 @@ object Messages {
     "Could not extract the given definition, please check the logs for more details or report an issue."
   )
 
+  val PasteWithIndentationFailed = new MessageParams(
+    MessageType.Error,
+    "Could not paste with indentation, please check the metals.log for more details or report an issue."
+  )
+
   val ReloadProjectFailed = new MessageParams(
     MessageType.Error,
     "Reloading your project failed, no functionality will work. See the log for more details"
--- a/metals/src/main/scala/scala/meta/internal/metals/formatting/RangeFormattingProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/formatting/RangeFormattingProvider.scala
@@ -10,6 +10,7 @@ import scala.meta.tokens.Tokens
 import org.eclipse.lsp4j.DocumentRangeFormattingParams
 import org.eclipse.lsp4j.FormattingOptions
 import org.eclipse.lsp4j.Range
+import org.eclipse.lsp4j.TextDocumentIdentifier
 import org.eclipse.lsp4j.TextEdit
 
 case class RangeFormatterParams(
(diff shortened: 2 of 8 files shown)


metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -105,6 +105,11 @@ object Messages {
     "Could not extract the given definition, please check the logs for more details or report an issue."
   )
 
+  val PasteWithIndentationFailed = new MessageParams(
+    MessageType.Error,
+    "Could not paste with indentation, please check the logs for more details or report an issue."
```
kpodsiad, line 110: Maybe we can be more specific and tell which file is relevant
```suggestion
    "Could not paste with indentation, please check the metals.log for more details or report an issue."
```

metals/src/main/scala/scala/meta/internal/metals/formatting/RangeFormattingProvider.scala
```diff
@@ -67,4 +69,17 @@ class RangeFormattingProvider(
       }
       .getOrElse(Nil)
   }
+
+  def fromatIdentOnPaste(
```
kpodsiad, line 73:
```suggestion
  def formatIdentOnPaste(
```

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -698,6 +698,35 @@ final class TestingServer(
...
   }
 
+  def indentOnPaste(
+      filename: String,
+      query: String,
+      expected: String,
+      paste: String,
+      root: AbsolutePath,
+      formattingOptions: Option[FormattingOptions]
+  )(implicit loc: munit.Location): Future[Unit] = {
```
kpodsiad, line 708: You can delete brackets which surround `for` and `yield`

kpodsiad: Why `loc` is needed here?

dos65:
@tgodzik Will it work differently from user experience? Or there would be a way just to pefrom it automatically after paste?

kpodsiad (approved):
Seems good, waiting for follow-up changes on vscode client.

tests/unit/src/main/scala/tests/TestingServer.scala
tgodzik: It might not be needed actually, will fix.

dos65:
> I will add a separate paste command that will work with indentation.

I really missed this automatic feature in Metals and I wold like to have it as it is.
Is there any significant reason/samples where it won't work/corrupt the code that we can't fix?

If it doesn't work well now maybe it would be easier to add a setting for enabling/disabling it for some period until it become stable?

tgodzik:
> > I will add a separate paste command that will work with indentation.
> 
> I really missed this automatic feature in Metals and I wold like to have it as it is.
> Is there any significant reason/samples where it won't work/corrupt the code that we can't fix?
> 

```
def method1 = 
  val c = a + b
  c
@@
```

In this case if you want to paste another method it will get indented the same as the previous one. We could add an option that if we have 2 newlines it reverts to the next indentation, but I am afraid there will be always issues with it and there might be people (see gitter about my discussion with Martin) that it will be problematic for.

> If it doesn't work well now maybe it would be easier to add a setting for enabling/disabling it for some period until it become stable?

You can always bind it as the default paste, which would make it work as previously.  I don't want to add another setting here, unless people feel like this is ok for this kind of functionality?

dos65:
@tgodzik I read the conversation in gitter and I think we can handle these problems but we need time. 
The only problem with this feature that we haven't tested it well on optional braces syntax.

I don't think that moving it into a separate command makes sense as it will significantly reduce amount of people who use it. 
That will lead to less bug reports/behavior changes requests.

Could you reconsider the option with a setting one more time?) This is the similar situation as with super-method-lenses.
I think it the best we can do as this feature was released/announced and probably works quite well for scala2 syntax.

tgodzik:
@dos65 Alternative approach https://github.com/scalameta/metals/pull/3127

tgodzik:
This is now under option and easy to turn off, so no need for this PR

PR #3127, opened 2021-09-15
Use the current indentation as reference when pasting

tgodzik:
Previously, we would try to guess the current indentation for the user, but since we don't know exactly what is their intention, it was easy guess wrong. Now we use the current indentation to see what user intends.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
@@ -9,42 +9,36 @@ import org.eclipse.lsp4j.TextEdit
 
 object IndentOnPaste extends RangeFormatter {
 
-  private val noEndClause = raw"((<!\bend\b\s*?)\b(if|while|for|match|try))"
-  private val ifThenClause = raw"(\bif\s+(?!.*?\bthen\b.*?$$)[^\s]*?)"
-  private val keywordClause = raw"\b(then|else|do|catch|finally|yield|case)"
-  private val openBraceParenBracketClause =
-    raw"""(^.*(\{[^}"']*|\([^)"']*|\[[^]"']*)$$)"""
-  private val extensionClause = raw"""extension\s*((\(|\[).*(\)|\]))+"""
-  val increaseIndentPatternRegex: Regex =
-    raw"""($noEndClause|$ifThenClause|$keywordClause|$extensionClause|=|=>|<-|=>>|:)\s*?$$|$openBraceParenBracketClause""".r
-  val indentRegex: Regex = raw"\S".r
+  private val indentRegex: Regex = raw"\S".r
 
   private def codeStartPosition(line: String): Option[Int] =
     indentRegex.findFirstMatchIn(line).map(_.start)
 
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
```diff
@@ -171,17 +153,7 @@ object IndentOnPaste extends RangeFormatter {
...
-    // drop leading/trailing empty lines
-    val lastIdx = idented.length - 1
-    val range = 0 to lastIdx
-    val trimmedStart =
-      range.dropWhile(converted(_).isEmpty).headOption.getOrElse(0)
-    val trimmedEnd = range.reverse
-      .dropWhile(converted(_).isEmpty)
-      .headOption
-      .getOrElse(lastIdx) + 1
-    idented.slice(trimmedStart, trimmedEnd)
```
dos65, line 184: There were some issues with this?

dos65 (approved):
Looks good! Now it's just an indentation corrector without any non-reliable logic.

metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
tgodzik: There is no need to trim the pasted snippet, it's also unexpected behaviour.

metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
```diff
@@ -77,20 +71,14 @@ object IndentOnPaste extends RangeFormatter {
...
-    val prePastedLines = splitLines.take(startLine).reverse
+    val pastedLines = splitLines.slice(startLine, endLine + 1)
 
     // Do not adjust indentation if we pasted into an existing line content
     def pastedIntoNonEmptyLine = {
       pastedLines match {
-        case Array(singleLine) =>
-          val originalLine = singleLine.substring(0, originalStart) +
-            singleLine.substring(endPos.endColumn)
+        case array if array.length > 0 =>
```
tgodzik, line 79: I changed the logic here to not reformat if we are printing into any line that already contains code.

tests/unit/src/test/scala/tests/rangeFormatting/MultilineStringRangeFormattingWhenSelectingSuite.scala
```diff
@@ -110,23 +110,23 @@ class MultilineStringRangeFormattingWhenSelectingSuite
     s"""
        |object Main {
```
tgodzik, line 111: The change here is due to the new way of calculating expected indent, I think we should at some point detect whether we are pasting or selecting, but not sure exactly how to do that.

metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
dos65: Ok. I don't remember exactly what was the reason. Probably, I added this just to have less changes in test data :smile:

dos65 (approved):
:rocket:

PR #3130, opened 2021-09-16
Support more scenarios with 'replace () with {}' code action

kpodsiad:
Aims to close https://github.com/scalameta/metals-feature-requests/issues/233.

I've found a scenario when rewrite parens action didn't work. Then I found an issue created by @agilesteel so I fed two birds with one seed :D

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/RewriteBracesParensCodeAction.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/RewriteBracesParensCodeAction.scala
@@ -15,10 +15,23 @@ import scala.meta.tokens.Token
 import org.eclipse.lsp4j.CodeActionParams
 import org.eclipse.{lsp4j => l}
 
+/**
+ * Rewrite parens to brackets and vice versa.
+ * It's a transformation between Term.Apply(_, List(_: Term)) and Term.Apply(_, List(Term.Block(List(_: Term))))
+ * Term.Block is equivalent to "surrounded by braces"
+ *
+ * Parens to brackets scenarios:
+ * 1: def foo(n: Int) = ???
+ *    foo(5)           ->   foo{5}
+ * 2: x.map(a => a)    ->   x.map{a => a}
+ * 3: x.map(_ match {       x.map{_ match {
+ *      case _ => 0    ->     case _ => 0
+ *    })                    }}
+ * Brackets to parens scenarios are the opposite of the ones above.
+ */
 class RewriteBracesParensCodeAction(
     trees: Trees
 ) extends CodeAction {
-
   override def kind: String = l.CodeActionKind.RefactorRewrite
 
   override def contribute(params: CodeActionParams, token: CancelToken)(implicit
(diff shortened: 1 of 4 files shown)


kpodsiad:
Now there is code action for
```scala
  List(1,2).map { a =>
    <<>>println(a)
    a
  }
```
which effects in `println{a}` which is correct. There is no code action to rewrite braces for `map`

Although I think it really through, I'll be very grateful for more negative scenarios or cases when it's not working correctly. Especially braces to parens conversion

metals/src/main/scala/scala/meta/internal/metals/codeactions/RewriteBracesParensCodeAction.scala
```diff
@@ -27,8 +41,13 @@ class RewriteBracesParensCodeAction(
     val path = params.getTextDocument().getUri().toAbsolutePath
     val range = params.getRange()
     val applyWithSingleFunction: Term.Apply => Boolean = {
-      case Term.Apply(_, List(_: Term.Function)) => true
-      case Term.Apply(_, List(Term.Block(List(_: Term.Function)))) => true
+      // exclude case when body has more than one line (is a Block) because it cannot be rewritten to parens
+      case Term.Apply(_, List(Term.Block(List(func: Term.Function))))
+          if func.body.isInstanceOf[Term.Block] =>
```
tgodzik, line 46:
```suggestion
      case Term.Apply(_, List(Term.Block(List(Term.Function(_, _: Term.Block))))) =>
```

tgodzik (changes requested):
Thanks a lot for the fixes! The tests seem to be failing unfortunately.

tests/unit/src/test/scala/tests/codeactions/ImportMissingSymbolLspSuite.scala
```diff
@@ -17,6 +18,7 @@ class ImportMissingSymbolLspSuite
     s"""|${ImportMissingSymbol.title("Future", "scala.concurrent")}
         |${ImportMissingSymbol.title("Future", "java.util.concurrent")}
         |${CreateNewSymbol.title("Future")}
+        |${RewriteBracesParensCodeAction.toBraces}
```
tgodzik, line 21: Why does the action pop up here? It shouldn't. Also in the other places in this test suite.

tgodzik (changes requested):

tests/unit/src/test/scala/tests/codeactions/ImportMissingSymbolLspSuite.scala
kpodsiad: I think it should. Look what is selected: part of the `Term.Apply`, to be specific you selected following
```json
                    "type": "Term.Apply",
                    "fun": {
                      "type": "Term.Select", 
                      "qual": {
                        "type": "Term.Name",
                        "value": "Future" <<< here
                      },
                      "name": {
                        "type": "Term.Name",
                        "value": "successful"
                      }
```
I think here it's a valid usage, tbh I'm more concerned about more ambiguous cases. Maybe, because of that, we should contribute this code action only when the selection range is equal to zero (just cursor, no selection at all)?

metals/src/main/scala/scala/meta/internal/metals/codeactions/RewriteBracesParensCodeAction.scala
```diff
@@ -15,56 +15,71 @@ import scala.meta.tokens.Token
...
-      case _ => false
-    }
-    trees
-      .findLastEnclosingAt[Term.Apply](
-        path,
-        range.getStart(),
-        applyWithSingleFunction
-      )
+    val applyTree =
+      if (range.getStart == range.getEnd)
```
kpodsiad, line 44: @tgodzik what do you think about this condition? IMO it reduces the ambiguity of scenarios where the user selects multiple lines and sees "rewrite to braces" code action

metals/src/main/scala/scala/meta/internal/metals/codeactions/RewriteBracesParensCodeAction.scala
```diff
@@ -92,4 +107,15 @@ class RewriteBracesParensCodeAction(
 object RewriteBracesParensCodeAction {
   val toParens = "Rewrite to parenthesis"
   val toBraces = "Rewrite to braces"
+  private val applyWithSingleFunction: Term.Apply => Boolean = {
```
tgodzik, line 110: Can it be just a private method in the class? Any particular reason for moving it to the object?

tests/unit/src/test/scala/tests/codeactions/ConvertPatternMatchLspSuite.scala
```diff
@@ -12,7 +13,7 @@ class PatternMatchRefactorLspSuite
        |  List(1,2).map { 
        |    _ match {
        |      case 1 => {
-       |        x = <<1>>
+       |        x = <<>>1
```
tgodzik, line 16: Is this change needed?

tgodzik (changes requested):
Just some minor comments left, good job!

metals/src/main/scala/scala/meta/internal/metals/codeactions/RewriteBracesParensCodeAction.scala
kpodsiad: Sound great. Rewrite braces, convert pattern match, is that all which could benefit from that distinction?

tgodzik: string interpolation actions also, really anything that has a single title

tgodzik (approved):
LGTM!

PR #3134, opened 2021-09-16
Set the default fallbackScalaVersion to scala 3 instead of 2.12

smarter:
The scala 3 compiler is more likely to be able to compile your scala 2
code than the reverse, so it makes sense to use scala 3 as the fallback.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalaVersionSelector.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalaVersionSelector.scala
@@ -18,7 +18,7 @@ class ScalaVersionSelector(
           .map(_.scalaInfo.getScalaVersion)
           .sorted
           .lastOption
-          .getOrElse(BuildInfo.scala212)
+          .getOrElse(BuildInfo.scala3)
     }
 
     val binary = ScalaVersions.scalaBinaryVersionFromFullVersion(selected)
--- a/tests/unit/src/test/scala/tests/FormattingLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/FormattingLspSuite.scala
@@ -19,6 +19,8 @@ class FormattingLspSuite extends BaseLspSuite("formatting") {
         s"""|/.scalafmt.conf
             |maxColumn = 100
             |version=${V.scalafmtVersion}
+            |/metals.json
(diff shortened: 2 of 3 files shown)


gabro (approved):
This seems like a good idea to me, it's worth it even just for the parser (I'm assuming this also controls the Scalameta parser)

smarter:
> [error] ==> X tests.FormattingLspSuite.basic  3.514s coursierapi.error.SimpleResolutionError$1: Error downloading org.scalameta:mtags_3.0.2:0.0.0+1-d8ebc291+20210916-1636-SNAPSHOT

Don't know what that means, maybe the tests need to manually publish mtags for scala 3 before running?

tgodzik:
> > [error] ==> X tests.FormattingLspSuite.basic  3.514s coursierapi.error.SimpleResolutionError$1: Error downloading org.scalameta:mtags_3.0.2:0.0.0+1-d8ebc291+20210916-1636-SNAPSHOT
> 
> Don't know what that means, maybe the tests need to manually publish mtags for scala 3 before running?

I can take a look tomorrow, most likely some tests do not have a version explicitely set and since the tests are run using Scala 2.12, only module for 2.12 is published.

dos65:
@smarter  you need configure `unit` project in `build.sbt` by adding:
```scala
Test/test := (Test/test).dependsOn(crossPublish(V.scala3))
Test/testOnly := (Test/test).dependsOn(crossPublish(V.scala3))
```

tgodzik:
> @smarter you need to configure `unit` project in `build.sbt` by adding:
> 
> ```scala
> Test / test := (Test / test).dependsOn(crossPublish(V.scala3))
> Test / testOnly := (Test / test).dependsOn(crossPublish(V.scala3))
> ```

This will cause the unit tests to always compile Scala 3 and 2.12 before running, which we reserve slow tests for. It should be easy to fix it the other way.

dos65:
@tgodzik 
Alternatively we can set back it to scala212 by default in test in `initialize` function: https://github.com/scalameta/metals/blob/ebe838cf1be5c2998779c8ad3e1ab2068f3405c9/tests/unit/src/main/scala/tests/BaseLspSuite.scala#L63-L66

tgodzik:
I just pushed a fix for the failing tests.

PR #3139, opened 2021-09-17
Add docs for Metals custom search command

gabro:


tgodzik (approved):
Thanks! I was thinking that we should most likely add some docs for the recent features, which we haven't done

kubukoz (approved):

ckipp01 (approved):

Issue #2200, opened 2020-11-10
Show BSP server version next to name in Doctor


ckipp01:
Just a short task so I don't forget. I'd be useful to show the BSP server version in the Doctor next to the server name. For some context, https://gitter.im/scalameta/metals?at=5faaf06cb86f640704424ee0. I think in this situation having the 1.4.xx displayed next to Bloop in the Doctor could have saved some time since right away the issue would have been apparent.

tgodzik:
Good idea! We do have that data available.

tgodzik:
Though coming to think of it, when Metals couldn't connect to Bloop, which happens with `3.0.0-M1` then we will be unable to restart the server. Not sure how to get the Bloop version in that case and it might be impossible to restart Bloop in that case.

ckipp01:
Ahhh, yea you're right. In that situation Doctor _would_ have still picked up that it was Bloop even though there was no active connection, but that's it. It wouldn't have had any connection details even to show the version...

https://github.com/scalameta/metals/blob/10ef69ce6e0f5823880dcabf9b0c6f2e0ccad6a1/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala#L314-L315

🤔 maybe this won't be as useful as I thought.

tgodzik:
I wonder if we should try to run bloopgun via ShellRunner to stop Bloop in case we were not able to connect to the build server?

There is a jar dependency for it: https://mvnrepository.com/artifact/ch.epfl.scala/bloopgun_2.12/1.4.5

ckipp01:
:( and now looking back at my code snippet up above I realize my typo..... 🙄 

> I wonder if we should try to run bloopgun via ShellRunner to stop Bloop in case we were not able to connect to the build server?

Huh, that's an interesting idea.

tgodzik:
Based on https://github.com/scalacenter/bloop/issues/1155 I think we can run via coursier:

`coursier launch ch.epfl.scala:bloopgun_2.12:1.4.5 --main-class bloop.bloopgun.Bloopgun -- exit`

so it should be the same via ShellRunner class.

tgodzik:
Checked it locally and seems to work well. We might need to run it in a temp directory though so that Bloop doesn't try to load the project which is failing.

tgodzik:
Also, we might want to do this automatically in the retry if the server is not able to connect. The exceptions is probably different than timeout in that case.

ckipp01:
[...]

PR #3141, opened 2021-09-18
Add build server version to the doctor view

kpodsiad:
Added version to the doctor as discussed in https://github.com/scalameta/metals/issues/2200


tgodzik (approved):
LGTM!

PR #3143, opened 2021-09-19
View source jar files as virtual docs

Arthurm1:
Currently when the client wants to browse files in source jars, the files are extracted and saved to `.metals/readonly/dependencies`
e.g.
`file:///workspacePath/.metals/readonly/dependencies/src.zip/java.base/java/nio/file/Path.java`
`file:///workspacePath/.metals/readonly/dependencies/trees_2.12-4.4.28-sources.jar/scala/meta/io/AbsolutePath.scala`

Current downsides:  The extracted source files aren't read only and they have to be written and cleared up.
Virtual doc downsides: These files won't be re-opened when VSCode is shutdown and re-opened

This PR means jar/jdk sources are read directly and passed back to the client as text.  This method uses the original URIs...
e.g.
`jar:file:///usr/lib/jvm/adoptopenjdk-11-hotspot-amd64/src.zip!/java.base/java/nio/file/Path.java`
`jar:file:///userPath/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scalameta/trees_2.12/4.4.28/trees_2.12-4.4.28-sources.jar!/scala/meta/io/AbsolutePath.scala`

This setting can be set per client in `MetalsServerConfig#isVirtualDocumentSupported`

There are a number of places that URIs are passed around as text and transformed into `AbsolutePath`.  That's fine for  `/somepath` but not `jar:file:///jarpath!package/file.java` so these transforms have to go through URI.  I may have missed some of these as I don't know all the code paths from VSCode->Metals.

Also - I'm working around https://github.com/coursier/coursier/issues/2186 (although if this issue is correct then I don't need to work around) but I think the workaround in `MetalsEnrichments#XtensionString#toAbsolutePath` can be left as is.

This PR depends on #3107

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
@@ -53,7 +53,8 @@ final class DefinitionProvider(
     remote: RemoteLanguageServer,
     trees: Trees,
     buildTargets: BuildTargets,
-    scalaVersionSelector: ScalaVersionSelector
+    scalaVersionSelector: ScalaVersionSelector,
+    saveDefFileToDisk: Boolean
 )(implicit ec: ExecutionContext) {
 
   val destinationProvider = new DestinationProvider(
@@ -63,7 +64,8 @@ final class DefinitionProvider(
     workspace,
     Some(semanticdbs),
     trees,
-    buildTargets
+    buildTargets,
+    saveDefFileToDisk
   )
 
   def definition(
@@ -270,7 +272,8 @@ class DestinationProvider(
     workspace: AbsolutePath,
     semanticdbsFallback: Option[Semanticdbs],
     trees: Trees,
-    buildTargets: BuildTargets
+    buildTargets: BuildTargets,
+    saveSymbolFileToDisk: Boolean
 ) {
 
   private def bestTextDocument(
@@ -350,7 +353,9 @@ class DestinationProvider(
   ): Option[DefinitionDestination] = {
     definition(symbol, allowedBuildTargets).map { defn =>
       val destinationDoc = bestTextDocument(defn)
-      val destinationPath = defn.path.toFileOnDisk(workspace)
+      val destinationPath =
+        if (saveSymbolFileToDisk) defn.path.toFileOnDisk(workspace)
+        else defn.path
       val destinationDistance =
         buffers.tokenEditDistance(destinationPath, destinationDoc.text, trees)
       DefinitionDestination(
--- a/metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
@@ -33,20 +33,32 @@ class JavaInteractiveSemanticdb(
   private val readonly = workspace.resolve(Directories.readonly)
 
   def textDocument(source: AbsolutePath, text: String): s.TextDocument = {
-    val sourceRoot = source.parent
     val workDir = AbsolutePath(
       Files.createTempDirectory("metals-javac-semanticdb")
     )
     val targetRoot = workDir.resolve("target")
     Files.createDirectory(targetRoot.toNIO)
 
+    val localSource =
+      if (source.isLocalFileSystem(workspace)) {
+        source
+      } else {
+        val sourceRoot = workDir.resolve("source")
+        Files.createDirectory(sourceRoot.toNIO)
+        val localSource = sourceRoot.resolve(source.filename)
+        Files.write(localSource.toNIO, text.getBytes)
+        localSource
+      }
+
+    val sourceRoot = localSource.parent
     val targetClasspath = buildTargets
-      .inferBuildTarget(sourceRoot)
+      .inferBuildTarget(source)
       .flatMap(buildTargets.targetJarClasspath)
       .getOrElse(Nil)
       .map(_.toString)
 
-    val jigsawOptions = addExportsFlags ++ patchModuleFlags(source, sourceRoot)
+    val jigsawOptions =
+      addExportsFlags ++ patchModuleFlags(localSource, sourceRoot, source)
     val mainOptions =
       List(
         javac.toString,
@@ -58,7 +70,7 @@ class JavaInteractiveSemanticdb(
     val pluginOption =
       s"-Xplugin:semanticdb -sourceroot:${sourceRoot} -targetroot:${targetRoot}"
     val cmd =
-      mainOptions ::: jigsawOptions ::: pluginOption :: source.toString :: Nil
+      mainOptions ::: jigsawOptions ::: pluginOption :: localSource.toString :: Nil
 
     val stdout = List.newBuilder[String]
     val ps = SystemProcess.run(
@@ -71,7 +83,7 @@ class JavaInteractiveSemanticdb(
     )
 
     val future = ps.complete.recover { case NonFatal(e) =>
-      scribe.error(s"Running javac-semanticdb failed for $source", e)
+      scribe.error(s"Running javac-semanticdb failed for $localSource", e)
       1
     }
 
@@ -80,7 +92,7 @@ class JavaInteractiveSemanticdb(
       targetRoot
         .resolve("META-INF")
         .resolve("semanticdb")
-        .resolve(s"${source.filename}.semanticdb")
+        .resolve(s"${localSource.filename}.semanticdb")
 
     val doc = if (semanticdbFile.exists) {
       FileIO
@@ -91,7 +103,7 @@ class JavaInteractiveSemanticdb(
       val log = stdout.result()
       if (exitCode != 0 || log.nonEmpty)
         scribe.warn(
-          s"Running javac-semanticdb failed for $source. Output:\n${log.mkString("\n")}"
+          s"Running javac-semanticdb failed for ${source.toURI}. Output:\n${log.mkString("\n")}"
         )
       s.TextDocument()
     }
@@ -108,7 +120,8 @@ class JavaInteractiveSemanticdb(
 
   private def patchModuleFlags(
       source: AbsolutePath,
-      sourceRoot: AbsolutePath
+      sourceRoot: AbsolutePath,
+      originalSource: AbsolutePath
   ): List[String] = {
     // Jigsaw doesn't allow compiling source with package
     // that is declared in some existing module.
@@ -127,7 +140,22 @@ class JavaInteractiveSemanticdb(
             case _ =>
               Nil
           }
-        case None => Nil
+        case None =>
+          if (
+            originalSource.jarPath.exists(_.filename == JdkSources.zipFileName)
+          ) {
+            originalSource.toNIO
+              .iterator()
+              .asScala
+              .headOption
+              .map(_.filename)
+              .map(moduleName =>
+                List("--patch-module", s"$moduleName=$sourceRoot")
+              )
+              .getOrElse(Nil)
+          } else {
+            Nil
+          }
       }
     } else Nil
   }
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -27,7 +27,9 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.Promise
 import scala.concurrent.duration.FiniteDuration
+import scala.util.Failure
 import scala.util.Properties
+import scala.util.Success
 import scala.util.Try
 import scala.util.control.NonFatal
 import scala.{meta => m}
@@ -289,7 +291,7 @@ object MetalsEnrichments
     }
     def isDependencySource(workspace: AbsolutePath): Boolean = {
       (isLocalFileSystem(workspace) &&
-      isInReadonlyDirectory(workspace))
+        isInReadonlyDirectory(workspace)) || isJarFileSystem
     }
 
     def isWorkspaceSource(workspace: AbsolutePath): Boolean =
@@ -300,7 +302,8 @@ object MetalsEnrichments
     def isLocalFileSystem(workspace: AbsolutePath): Boolean =
       workspace.toNIO.getFileSystem == path.toNIO.getFileSystem
 
-    def isJarFileSystem: Boolean = path.toURI.getScheme() == "jar"
+    def isJarFileSystem: Boolean =
+      path.toNIO.getFileSystem().provider().getScheme().equals("jar")
 
     def isInReadonlyDirectory(workspace: AbsolutePath): Boolean =
       path.toNIO.startsWith(
@@ -395,7 +398,7 @@ object MetalsEnrichments
         AbsolutePath(out)
       }
 
-      // prevent inifinity loop
+      // prevent infinity loop
       if (retryCount > 5) {
         throw new Exception(s"Unable to save $path in workspace")
       } else if (path.toNIO.getFileSystem == workspace.toNIO.getFileSystem) {
@@ -530,6 +533,32 @@ object MetalsEnrichments
 
   }
 
+  implicit class XtensionURI(value: URI) {
+    def toAbsolutePath: AbsolutePath = toAbsolutePath(followSymlink = true)
+    def toAbsolutePath(followSymlink: Boolean): AbsolutePath = {
+      val path =
+        if (value.getScheme() == "jar")
+          Try {
+            AbsolutePath(Paths.get(value))
+          } match {
+            case Success(path) => path
+            case Failure(_) =>
+              // don't close - put up with the resource staying open so all AbsolutePath methods don't have to be wrapped
+              m.internal.io.PlatformFileIO.newFileSystem(
+                value,
+                new java.util.HashMap[String, String]()
+              )
+              AbsolutePath(Paths.get(value))
+          }
+        else
+          AbsolutePath(Paths.get(value))
+      if (followSymlink)
+        path.dealias
+      else
+        path
+    }
+  }
+
   implicit class XtensionString(value: String) {
 
     /**
@@ -588,17 +617,12 @@ object MetalsEnrichments
     def toAbsolutePath(followSymlink: Boolean): AbsolutePath = {
       // jar schemes must have "jar:file:"" instead of "jar:file%3A" or jar file system won't recognise the URI.
       // but don't overdecode as URIs may not be recognised e.g. "com-microsoft-java-debug-core-0.32.0%2B1.jar" is correct
-      val decodedUriStr =
-        if (value.toUpperCase.startsWith("JAR:FILE%3A"))
-          URLDecoder.decode(value, "UTF-8")
-        else value
-      val path = AbsolutePath(
-        Paths.get(URI.create(decodedUriStr.stripPrefix("metals:")))
-      )
-      if (followSymlink)
-        path.dealias
+      if (value.toUpperCase.startsWith("JAR%3AFILE"))
+        URLDecoder.decode(value, "UTF-8").toAbsolutePath(followSymlink)
+      else if (value.toUpperCase.startsWith("JAR:FILE%3A"))
+        URLDecoder.decode(value, "UTF-8").toAbsolutePath(followSymlink)
       else
-        path
+        URI.create(value.stripPrefix("metals:")).toAbsolutePath(followSymlink)
     }
 
     def indexToLspPosition(index: Int): l.Position = {
@@ -737,7 +761,7 @@ object MetalsEnrichments
   implicit class XtensionClasspath(classpath: List[String]) {
     def toAbsoluteClasspath: Iterator[AbsolutePath] = {
       classpath.iterator
-        .map(uri => AbsolutePath(Paths.get(URI.create(uri))))
+        .map(_.toAbsolutePath)
         .filter(p => Files.exists(p.toNIO))
     }
   }
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -2,6 +2,7 @@ package scala.meta.internal.metals
 
 import java.io.IOException
 import java.net.URI
+import java.net.URLDecoder
 import java.nio.charset.StandardCharsets
 import java.nio.file.FileAlreadyExistsException
 import java.nio.file.Files
@@ -264,7 +265,7 @@ object MetalsEnrichments
 
   implicit class XtensionPath(path: Path) {
     def toUriInput: Input.VirtualFile = {
-      val uri = path.toUri.toString
+      val uri = path.toAbsolutePath.toUri.toString
       val text = new String(Files.readAllBytes(path), StandardCharsets.UTF_8)
       Input.VirtualFile(uri, text)
     }
@@ -286,9 +287,10 @@ object MetalsEnrichments
         accum.resolve(filename.toString)
       }
     }
-    def isDependencySource(workspace: AbsolutePath): Boolean =
-      isLocalFileSystem(workspace) &&
-        isInReadonlyDirectory(workspace)
+    def isDependencySource(workspace: AbsolutePath): Boolean = {
+      (isLocalFileSystem(workspace) &&
+      isInReadonlyDirectory(workspace))
+    }
 
     def isWorkspaceSource(workspace: AbsolutePath): Boolean =
       isLocalFileSystem(workspace) &&
@@ -297,6 +299,8 @@ object MetalsEnrichments
 
     def isLocalFileSystem(workspace: AbsolutePath): Boolean =
       workspace.toNIO.getFileSystem == path.toNIO.getFileSystem
+
+    def isJarFileSystem: Boolean = path.toURI.getScheme() == "jar"
 
     def isInReadonlyDirectory(workspace: AbsolutePath): Boolean =
       path.toNIO.startsWith(
@@ -448,7 +452,7 @@ object MetalsEnrichments
      */
     def toInputFromBuffers(buffers: Buffers): m.Input.VirtualFile = {
       buffers.get(path) match {
-        case Some(text) => Input.VirtualFile(path.toString(), text)
+        case Some(text) => Input.VirtualFile(path.toURI.toString(), text)
         case None => path.toInput
       }
     }
@@ -582,8 +586,14 @@ object MetalsEnrichments
 
     def toAbsolutePath: AbsolutePath = toAbsolutePath(followSymlink = true)
     def toAbsolutePath(followSymlink: Boolean): AbsolutePath = {
+      // jar schemes must have "jar:file:"" instead of "jar:file%3A" or jar file system won't recognise the URI.
+      // but don't overdecode as URIs may not be recognised e.g. "com-microsoft-java-debug-core-0.32.0%2B1.jar" is correct
+      val decodedUriStr =
+        if (value.toUpperCase.startsWith("JAR:FILE%3A"))
+          URLDecoder.decode(value, "UTF-8")
+        else value
       val path = AbsolutePath(
-        Paths.get(URI.create(value.stripPrefix("metals:")))
+        Paths.get(URI.create(decodedUriStr.stripPrefix("metals:")))
       )
       if (followSymlink)
         path.dealias
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -311,9 +311,14 @@ object MetalsEnrichments
       )
 
     def toRelativeInside(prefix: AbsolutePath): Option[RelativePath] = {
-      val relative = path.toRelative(prefix)
-      if (relative.toNIO.getName(0).filename != "..") Some(relative)
-      else None
+      // windows throws an exception on toRelative when on different drives
+      if (path.toNIO.getRoot() != prefix.toNIO.getRoot())
+        None
+      else {
+        val relative = path.toRelative(prefix)
+        if (relative.toNIO.getName(0).filename != "..") Some(relative)
+        else None
+      }
     }
 
     def isInside(prefix: AbsolutePath): Boolean =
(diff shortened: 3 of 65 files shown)


ckipp01:
> This setting can be set per client in MetalsServerConfig#isVirtualDocumentSupported

I'd actually recommend only supporting and documenting this via `InitializationOptions`. This sort of seems like the best place to support it (I see you already do), so I'm not sure if there is any benefit to also adding it to `MetalsServerConfig`. That was sort of the very initial place settings used to be added before we started using `ClientExperimentalCapablities` or `InitializationOptions`. Now everything that is client specific should be added via `InitialziationOptions`.

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -334,7 +337,9 @@ class DestinationProvider(
   ): Option[DefinitionDestination] = {
     definition(symbol, allowedBuildTargets).map { defn =>
       val destinationDoc = bestTextDocument(defn)
-      val destinationPath = defn.path.toFileOnDisk(workspace)
+      val destinationPath =
+        if (saveSymbolFileToDisk) defn.path.toFileOnDisk(workspace)
```
tgodzik, line 341: This makes sure that we don't save the file to disk, but I can understand how the file is being sent back to the user. Do we send the path with the jar and VS Code is responsible for that?

tgodzik (changes requested):
I left some minor comments, but looks good. Could you take a look at the failing tests? Some of them are flaky, but it seems that some of the failures are real.

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
Arthurm1: Yes - the wonder of Virtual documents.  Metals tells VSCode (or any client) to open a jar URI e.g. `jar:file://path/jarname.jar!source/source.java` instead of the URI of the read-only file that we've extracted from the jar and saved on the file system.  VSCode intercepts any URI with the `jar` scheme and asks the `MetalsFileProvider` to provide the contents of that document.  `MetalsFileProvider` sends a `DecodeFile` command to Metals with this same URI. Metals hands this off to the `FileDecoderProvider` which extracts the file's contents from the jar and sends it back to VSCode.  No persistence to the filesystem occurs.  It's quite a nice way (I think) of handling any data that we don't have saved on the file system but refer to by URI.  Then you can use standard jar style URIs to navigate round the source jars or still create read-only files and still navigate round source jars using `file` scheme URIs.  If all clients supported this then Metals could dump the readonly folder entirely (I think)

tgodzik: Thanks for explanation!

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -434,12 +437,18 @@ object MetalsEnrichments
       filename.endsWith(".jar") || filename.endsWith(".srcjar")
     }
 
+    def isSourceJar: Boolean =
+      path.jarPath.exists(_.filename.endsWith("-sources.jar"))
+
+    def isJDKSource: Boolean =
+      path.jarPath.exists(_.filename == "src.zip")
```
dos65, line 444: There functions are correct in 99% percents but I believe it might cause some exceptionally rare issues that would be hard to investigate.

There is non zero possibility that there might a be a dependency jar without a version that ends with `-sources` in build - some local unmanaged jar.

The same for `src.zip` - probably there might be some `src.zip` that aren't about Jdk at all. Also, I'm not sure if all jdk/os distros put sources in `src.zip`

It would be more reliable always check it if jar is a source in `BuildTargets` as it's anyway quite fast.

Arthurm1:
I think this is finished now.  The Windows test failure is a timeout in `DebugDiscoverySuite`

Paired with https://github.com/scalameta/metals-vscode/pull/695

kpodsiad (approved):

tgodzik (changes requested):
There seem to be some issues with Tree View, which I am taking a look at here:
https://github.com/scalameta/metals/compare/main...tgodzik:source_jars_virtual_documents?expand=1

I will update the PR with the fixes later on.

Arthurm1:
I'm off on holiday for a week so I can look when I get back - I don't think there's a rush for this feature as it doesn't add any functionality

tgodzik:
> I'm off on holiday for a week so I can look when I get back - I don't think there's a rush for this feature as it doesn't add any functionality

I will take a quick look then and maybe focus on some missing features in Java support instead.

Arthurm1:
There's an issue with calculating interactive semanticdb files on jar classes now that they are virtual documents rather than files in the readonly dir.

For Java, I'm not sure it's possible without saving to file first as I don't think you can call javac on a single file in a jar.  For Scala files - I haven't looked into yet.

One thing I didn't understand though: In `InteractiveSemanticdbs` there is this code...
```scala
 def shouldTryCalculateInteractiveSemanticdb = {
      source.isLocalFileSystem(workspace) && (
        unsavedContents.isDefined ||
          source.isInReadonlyDirectory(workspace) || // dependencies
          source.isSbt || // sbt files
          source.isWorksheet || // worksheets
          doesNotBelongToBuildTarget // standalone files
      )
    }
```
It seems as though semanticdb info would never be calculated on dependencies unless they've been unzipped into the readonly directory.  Is interactive semanticdb info only required if the user has the file open?  e.g. Metals doesn't need semanticdb info on `java.time.LocalDate` unless it's being viewed?  It wouldn't need the info if the user is just hovering over a `LocalDate` reference or using `Goto definition` on it?

dos65:
@Arthurm1 

> For Java, I'm not sure it's possible without saving to file first as I don't think you can call javac on a single file in a jar. For Scala files - I haven't looked into yet.

You right. For Java it will be required to dump into a file but I don't think that it's a problem.
For Scala there is no need in doing this as we directly pass virtual source to Presentation Compiler.

> It seems as though semanticdb info would never be calculated on dependencies unless they've been unzipped into the readonly directory. Is interactive semanticdb info only required if the user has the file open? e.g. Metals doesn't need semanticdb info on java.time.LocalDate unless it's being viewed? It wouldn't need the info if the user is just hovering over a LocalDate reference or using Goto definition on it?

Yep, for dependencies semanticdb is required only if this file is opened in the editor.
Actually, it's the same for other types of source files in most cases except of `Find References`/`Rename` features that work differently or shouldn't work for dependency sources.

For example, if you use `java.time.LocalDate` in source `A.scala` then there will be occurrence of `java.time.LocalDate`  in `A.scala.semanticdb` and that's is enough information to perform goto-definition from `A.scala` to `j/t/LocalDate.java`.

olafurpg:
This is exciting! I just want to leave a short note that VS Code’s FileSystemProvider is preferable over TextDocumentContentProvider because it gives you breadcrumbs in the UI and the ability to list files in a directory (making it easy to explore the entire jar file instead of only a single file). Implementing the file system API may require slightly differently shaped messages in the JSON-RPC layer.

Arthurm1:
@olafurpg I did have a brief look at it but it needs this PR done first as Metals doesn't understand `jar:` URIs everywhere yet.

The way VSCode handles FileSystemProvider is as an additional workspace.  Metals doesn't understand multiple workspaces yet but it's not that difficult to handle this specific feature.  It does then mean breadcrumbs work, but I'm not sure there's any benefit over the current MetalsTreeView which already allows the user to explore the jar.

Maybe it would mean we could drop the `libraries` section of the the TreeView and make the support easier across non-vscode clients, although VIM does support the treeview.  I can't see why we'd keep both FileSystemProvider and TreeView:Libraries.

olafurpg:
With FileSystemProvider you still need to implement your own tree views (vscode doesn’t provide one automatically), so the libraries view can stay unchanged. The main selling point is the breadcrumbs, which is a big deal IMO because it makes it much easier to see what file you have open (especially if the URI has a human readable way to identify the library name and version). The vscode proposed api has new methods to integrate with the native file tree so there may be other benefits in the future.

olafurpg:
I’ve been working with the vscode file system APIs lately and just wanted to share these findings. No need to act on it, having virtual documents alone would be a nice huge improvement already :)

Arthurm1:
That makes sense.  I do think breadcrumbs are handy - I just thought I was missing something else.

Arthurm1:
I'm struggling to get this working in all scenarios.  I can either get `ImplementationLspSuite` to pass or `JavaDefinitionVirtualDocSuite` but not both.

The difference in code to get them working is `InteractiveSemanticdbs`.  I either have the current...
```scala
    def shouldTryCalculateInteractiveSemanticdb = {
      source.isLocalFileSystem(workspace) && (
        unsavedContents.isDefined ||
          source.isInReadonlyDirectory(workspace) || // dependencies
          source.isSbt || // sbt files
          source.isWorksheet || // worksheets
          doesNotBelongToBuildTarget // standalone files
      )
    }
```

or the new...
```scala
    def shouldTryCalculateInteractiveSemanticdb = {
      source.isLocalFileSystem(workspace) && (
        unsavedContents.isDefined ||
          source.isInReadonlyDirectory(workspace) || // dependencies
          source.isSbt || // sbt files
          source.isWorksheet || // worksheets
          doesNotBelongToBuildTarget // standalone files
      ) || source.isJarFileSystem // VIRTUAL DOC DEPENDENCIES
    }
```
The last line is the difference.  It's needed to get the [JavaDefinitionSuite#jdk-String-patch-module](https://github.com/scalameta/metals/blob/bc2fa7a60be8369e073ef0b9aab363d09e658d31/tests/unit/src/test/scala/tests/JavaDefinitionSuite.scala#L36) test to pass, as `java.lang.String` file must have semanticdb generated for it.  But then semanticdb gets generated for `java.lang.Exception` in the [ImplementationLspSuite#java-classes](https://github.com/scalameta/metals/blob/bc2fa7a60be8369e073ef0b9aab363d09e658d31/tests/unit/src/test/scala/tests/ImplementationLspSuite.scala#L253) test which causes `ImplementationProvider` to fail to find everything.

I can see 2 ways of fixing this but I'm not sure the correct route.  Currently it works because in the above `InteractiveSemanticdbs` code, it will only generate semanticdb info if the file exists in the readonly area.  For `JavaDefinitionSuite#jdk-String-patch-module`, the test code calls `MetalsLanguageServer#workspaceSymbol` which saves `java.lang.String` to readonly area so it has semanticdb run on it.  In `ImplementationLspSuite#java-classes` the `java.lang.Exception` class is not saved to disk so no semanticdb is generated.

With virtual docs, this info about whether a file has or hasn't been saved in the readonly area is lost.  Every dependency uri is `jar:file//XXXX.jar` and is never saved.  So there is no info at the point in `InteractiveSemanticdbs` to decide whether we should or shouldn't generate semanticdb info. So we always have to on a `.scala` or `.java` file.

You'd think that this wouldn't be an issue - if we always generate it then the worst thing that can happen is a drop in performance but in [ImplementationProvider#implementations](https://github.com/scalameta/metals/blob/bc2fa7a60be8369e073ef0b9aab363d09e658d31/metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala#L131) we can see different paths taken depending on if semanticdb info exists.  Either it uses `globalTable#globalContextFor` if there is no semantidb info or `InheritanceContext#fromDefinitions` if there is.  But the issue is that `InheritanceContext#fromDefinitions` isn't aware of all the other inheritance info, e.g. that `RuntimeException` inherits from `Exception`, so doesn't return all of the implementations - only some of them.

So should `ImplementationProvider#implementations` be changed to check `InheritanceContext` _and_ `globalTable`?  Or should `Semanticdbs#textDocument` be changed to take more information than just a `path` e.g. `generateIfMissing: Boolean`.  Or should `InheritanceContext` have had the info on `RuntimeException` and how would it get that info? 

I just don't know which is supposed to be correct as I'm not sure what info should be stored where.

tgodzik:
@Arthurm1 I think this might actually be a bug, I did a quick fix for your use case here:

https://github.com/tgodzik/metals/commit/0d48e8fe27ec6bc31f40ecc93d75a26a749e1d0f

But we should not calculate semanticdb for ImplementationProvider, I will check if this is an issue in main also.

tgodzik:
Ok, interestingly enough, this can also happen in the current main if the user goes to definition on an Exception. I will look into that.

I noticed that there is an exception popping up:
```
2022.01.18 14:10:01 ERROR text document: file:///Users/runner/work/metals/metals/tests/unit/target/e2e/implementation/java-classes/jar%253Afile%253A%252F%252F%252FUsers%252Frunner%252FLibrary%252FCaches%252FCoursier%252Fv1%252Fhttps%252Frepo1.maven.org%252Fmaven2%252Forg%252Fscala-lang%252Fscala-library%252F2.12.15%252Fscala-library-2.12.15-sources.jar%2521%252Fscala%252Fpackage.scala
java.nio.file.NoSuchFileException: /Users/runner/work/metals/metals/tests/unit/target/e2e/implementation/java-classes/jar%3Afile%3A%2F%2F%2FUsers%2Frunner%2FLibrary%2FCaches%2FCoursier%2Fv1%2Fhttps%2Frepo1.maven.org%2Fmaven2%2Forg%2Fscala-lang%2Fscala-library%2F2.12.15%2Fscala-library-2.12.15-sources.jar%21%2Fscala%2Fpackage.scala
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
	at sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:219)
	at java.nio.file.Files.newByteChannel(Files.java:371)
	at java.nio.file.Files.newByteChannel(Files.java:422)
	at java.nio.file.Files.readAllBytes(Files.java:3206)
	at scala.meta.internal.io.PlatformFileIO$.slurp(PlatformFileIO.scala:45)
	at scala.meta.internal.io.FileIO$.slurp(FileIO.scala:24)
	at scala.meta.internal.metals.InteractiveSemanticdbs.$anonfun$textDocument$2(InteractiveSemanticdbs.scala:90)
	at scala.Option.getOrElse(Option.scala:189)
...
```
this might be related.

Arthurm1:
There's a couple of those where `TextDocument#uri = workspace.resolve(path)` is called on a jar.

`uri` is a `String` so pretty much anything can be put in there.  Sometimes a real `jar:file` uri comes in and then you get an exception as we try to resolve that against the workspace.  Most of the time `path` is a relative location in the workspace. e.g. `.metals/readonly/someunzippedjarfile.scala`.

I didn't realise this was happening on main.

Ideally I'd like `TextDocument#uri` to always be a proper uri e.g. 'jar:file///xxx' or 'file:/xxx' or 'metalsdecode:/XXX' but I don't know if this affects other code

tgodzik:
> There's a couple of those where `TextDocument#uri = workspace.resolve(path)` is called on a jar.
> 
> `uri` is a `String` so pretty much anything can be put in there. Sometimes a real `jar:file` uri comes in and then you get an exception as we try to resolve that against the workspace. Most of the time `path` is a relative location in the workspace. e.g. `.metals/readonly/someunzippedjarfile.scala`.
> 

I think that if we could check if it's a jar and then use the relative position in the jar. Or something along these lines?

> I didn't realise this was happening on main.

Actually I checked again and that seems to not be an issue as we don't direct to readonly directory in ImplementationProvider, it's always the jar, which is why it's popping up here. But the change I suggested here should be ok.

Arthurm1:
Not sure why the Windows test fails - flaky?

I think this is complete.  There's a number of tests that save jar classes to the `readonly` area.  I've subsclassed all of those so they test virtual docs and saveToDisk.  Hopefully I've caught all of them.

I've repurposed `readonlySources` in TestingServer.  It wasn't being used, so it now records jar names.  The issue with `TestingServer` is that it can easily find jar classes when they're saved to disk as they're all in the `dependencies` area.  When they're just accessed virtually they aren't saved anywhere so this `TrieMap` records that.

Jars are a pain to work with as the FileSystem has to be opened to just define an `AbsolutePath` let alone use methods on it.  Because of that I've left the FileSystem open for any jar that's defined using `URI.toAbsolutePath`.  Otherwise every `AbsolutePath` method would have to be wrapped in `PlatformFileIO.newFileSystem`.  It looks like Jars are already left open in `JarTopLevels#getTopLevels`.

In `ClientConfigurationAdapter` I've passed a `URI` when it should be a path but then there's no such thing for jars is there - it's always a URI?  I'm not sure why we don't use `InitializeRequestArgumentsPathFormat.URI` as the format but passing a jar URI works in both cases.

I've tested References, Goto, TreeView sync, debug stacktrace navigation, analyse stacktrace navigation.  Is there anything else?

tgodzik:
> Not sure why the Windows test fails - flaky?

I am rerunning them, but it's the first time I've seen them running out of memory :thinking: 

> I've tested References, Goto, TreeView sync, debug stacktrace navigation, analyse stacktrace navigation. Is there anything else?

Nothing currently coming to mind. I will have a try before merging it, but otherwise we can test it out a bit more on the newest snapshots, so fortunately we can iron it out before the next release (which this could be a main feature of :rocket: )

Arthurm1:
@tgodzik Cool.  Wouldn't mind if you also merged the `Display Build Info` functionality 😉

tgodzik:
> Cool. Wouldn't mind if you also merged the Display Build Info functionality

Sure! I can look at it first, looks shorter to review :S

tgodzik:
The windows failures look legit:
```
java.lang.Exception: Obtained [file:///D:/a/metals/metals/tests/unit/target/e2e/debug-step-save-to-disk/step-out/a/src/main/scala/Main.scala, 5], expected [D:\a\metals\metals\tests\unit\target\e2e\debug-step-save-to-disk\step-out\a\src\main\scala\Main.scala, 5]
```

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -527,6 +534,32 @@ object MetalsEnrichments
...
+        if (value.getScheme() == "jar")
+          Try {
+            AbsolutePath(Paths.get(value))
+          } match {
+            case Success(path) => path
+            case Failure(_) =>
+              // don't close - put up with the resource staying open so all AbsolutePath methods don't have to be wrapped
+              m.internal.io.PlatformFileIO.newFileSystem(
+                value,
+                new java.util.HashMap[String, String]()
```
tgodzik, line 550:
```suggestion
                create = false
```
? 

Not 100% sure what create does, but we sue it in a lot of places. I asume it means it will not throw an exception if the jar system is already created?

tests/slow/src/test/scala/tests/feature/WorksheetCrossLspSuite.scala
```diff
@@ -2,12 +2,45 @@ package tests.feature
...
-class Worksheet3LspSuite extends tests.BaseWorksheetLspSuite(V.scala3) {
+abstract class Worksheet211LspSuite(
+    useVirtualDocuments: Boolean,
+    suiteNameSuffix: String
+) extends tests.BaseWorksheetLspSuite(
+      V.scala211,
+      useVirtualDocuments,
+      suiteNameSuffix
+    )
+class Worksheet211LspSaveToDiskSuite
```
tgodzik, line 13: We can run `virtual-docs` as default and let's just add a single test suite (could be case really) that runs with `save-to-disk`. I don't think we need to replicate it for every Scala versions and does take a while to run it. We can even remove the `2.11` suite I think, since we did depracte it.

tests/slow/src/test/scala/tests/sbt/SbtStepDapSuite.scala
```diff
@@ -4,9 +4,18 @@ import tests.SbtBuildLayout
...
+    useVirtualDocuments: Boolean,
+    suiteNameSuffix: String
+) extends BaseStepDapSuite(
+      useVirtualDocuments,
+      s"sbt-debug-step-$suiteNameSuffix",
       SbtServerInitializer,
       SbtBuildLayout
     )
+
+class SbtStepDapSuiteSaveToDiskSuite
```
tgodzik, line 17: I would avoid running the tests double and just run a single test case with changed useVirtualDocuments option. These tests do take a while and I think is is also partly the reason why Windows tests run out of memory at some point.

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
```diff
@@ -8,10 +8,19 @@ import scala.meta.internal.metals.UserConfiguration
...
-    extends BaseLspSuite("worksheet") {
-  override def initializationOptions: Option[InitializationOptions] =
-    Some(InitializationOptions.Default.copy(decorationProvider = Some(true)))
+abstract class BaseWorksheetLspSuite(
+    scalaVersion: String,
+    useVirtualDocuments: Boolean,
+    suiteNameSuffix: String
+) extends BaseLspSuite(s"worksheet-$suiteNameSuffix") {
+
+  override protected def initializationOptions: Option[InitializationOptions] =
```
tgodzik, line 17: We could make `initializationOptions` a var and change `useVirtualDocuments` for a single case instead.

tests/unit/src/test/scala/tests/RenameLspSuite.scala
```diff
@@ -1,6 +1,21 @@
 package tests
 
-class RenameLspSuite extends BaseRenameLspSuite("rename") {
+import scala.meta.internal.metals.InitializationOptions
+
+abstract class RenameLspSuite(
```
tgodzik, line 5: RenameLsp suite take a lot of time, so this would be really bad for out CI to run it twice.

tgodzik (changes requested):
Thanks for the great work! This is a lot of changes, far more than I expected it to need.

Please avoid running the suites double as it may add a lot of running time to our CI and the windows jobs are already at an edge (I think the CI failed previously because of that). Could we instead have a single case everywhere that checks the specific logic when using readonly directory? I would assume the virtual docs to be the default for all the other cases.

We can do it 2 ways:
- make init options a var and replace it for a single test (find one that was failing for that suite when changed to virtual docs)
- have a separate test case for each suite that runs only when `.readonly` files option is ticket and everything else in other case

I need to rewrite some of the suites and move a lot the logic testing to some unit tests while leaving only real integration tests there.

Again, thanks for the great work and it already looks amazing. Sorry for the additional comments. If you unable to find time to do it, I can do it next week.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
Arthurm1: I think `create` is there if you want to create a jar from scratch.  The jars here should all exist.  Also - that param isn't available on the method I'm calling as I'm passing a URI, not an AbsolutePath

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
```diff
@@ -33,20 +33,32 @@ class JavaInteractiveSemanticdb(
...
 
   def textDocument(source: AbsolutePath, text: String): s.TextDocument = {
-    val sourceRoot = source.parent
     val workDir = AbsolutePath(
       Files.createTempDirectory("metals-javac-semanticdb")
     )
     val targetRoot = workDir.resolve("target")
     Files.createDirectory(targetRoot.toNIO)
 
+    val localSource =
```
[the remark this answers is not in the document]
Arthurm1: You mean the `workDir`?  I.e. wrap everything in a `try..finally`?

Arthurm1:
@tgodzik I'm not sure exactly what you want with altering the tests.

You mean just preserve 1 test within each suite to test the `save-to-disk` version?

I'll have to look through all the tests to find out which individual test actually writes to disk - before I just checked if the testsuite wrote to disk.

I get your point about not wanting to increase the testing time - I had to double up on 16 testsuites which is a lot.

tgodzik:
> You mean just preserve 1 test within each suite to test the save-to-disk version?

That's what I meant, I think that should be enough to test out all the issues that might pop up. It's not perfect for sure, but until we manage to get the test more effective it will have to do.

tgodzik:
> I'll have to look through all the tests to find out which individual test actually writes to disk - before I just checked if the testsuite wrote to disk.

I will take a look at it.

tgodzik:
I added a PR here: https://github.com/Arthurm1/metals/pull/1

That should work and not add too much testing time.

Arthurm1:
@tgodzik I've tried out your testsuite fix but it's a bit temperamental.

I think this is because `newServer` adjusts the `initializationOptions` depending on the value of `useVirtualDocs` but `newServer` is called in `beforeEach` and the `virtualDocs = true/false` is set in the test body which would be _after_ the `newServer` has been setup.

The only ways I can think of a way of influencing the `newServer` call are...
1. Go back to subclassing the suites which we're trying to avoid to reduce test times
2. Shift `newServer` into the `test` body - but then it would no longer be aware of the test name
3. Add a `VirtualDoc` test tag.  Which is what I've done here.  What do you think?  I don't really like it - it feels wrong - but it could be temporary if we eventually drop `savetodisk`.


Also - I see you've added `test("definition".only` into `BaseWorksheetLspSuite` but it fails on CI with the following error message that I don't understand...
```
[error] ==> X tests.worksheets.WorksheetLspSuite.definition-readonly  0.003s munit.FailException: /home/runner/work/metals/metals/tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala:494 'Only' tag is not allowed when `isCI=true`
```

tests/unit/src/main/scala/tests/BaseLspSuite.scala
```diff
@@ -77,6 +85,21 @@ abstract class BaseLspSuite(
...
+        testOpts
+          .withName(s"${testOpts.name}-virtualdoc")
+          .withTags(Set(virtualDocTag))
+      ) { fn }
+    } else {
+      test(testOpts)(fn)
+    }
+  }
+
   def newServer(workspaceName: String): Unit = {
```
tgodzik, line 103: We could add an additional option to `newServer` 
```suggestion
  def newServer(workspaceName: String, useVirtualDocs): Unit = {
```
and check in beforeEach if the tag is set, which would probably be better than the var I proposed.

tests/unit/src/main/scala/tests/BaseLspSuite.scala
```diff
@@ -49,6 +51,12 @@ abstract class BaseLspSuite(
 
   protected def initializationOptions: Option[InitializationOptions] = None
 
+  private val virtualDocTag = new Tag("UseVirtualDocs")
```
tgodzik, line 54: I think the tag approach is ok with a small modification for removing the var.

tests/unit/src/main/scala/tests/BaseWorksheetLspSuite.scala
```diff
@@ -489,6 +491,33 @@ abstract class BaseWorksheetLspSuite(scalaVersion: String)
     } yield ()
   }
 
+  test("definition".only, withoutVirtualDocs = true) {
```
tgodzik, line 494: This is my bad, `.only` is one of the default tags that will run only the test case tagged with it and it breaks on CI in case someone (such as me here) forgets to remove it :sweat:

tgodzik (changes requested):
> I've tried out your testsuite fix but it's a bit temperamental.

>I think this is because newServer adjusts the initializationOptions depending on the value of useVirtualDocs but newServer is called in beforeEach and the virtualDocs = true/false is set in the test body which would be after the newServer has been setup.

Ach you are right, I somehow thought it was working, but maybe I tested it out too quickly. Your approach with the tags is ok I think.

As for the `.only` issue that's also my bad, added an explanation in the comment. Sorry!

tests/unit/src/main/scala/tests/BaseLspSuite.scala
Arthurm1: Do you mean removing the `val virtualDocTag`?  If so - where would you put it - in a companion object?  Or do you mean remove the `val useVirtualsDoc` var?

tests/unit/src/main/scala/tests/BaseLspSuite.scala
Arthurm1: The var is still needed as the tests themselves need to know whether they are using virtualdocs, so need to be able to call `BaseLspSuite#useVirtualDocuments`.  And it can't just interrogate `initializationOptions` as this doesn't contain the virtualdoc info as the `initializationOptions` are added to in `newServer`.  We could add the additional option to `newServer` but it would just be duplicating the `BaseLspSuite#useVirtualDocs` var

tgodzik: Ok, makes sense!

tests/unit/src/main/scala/tests/BaseLspSuite.scala
tgodzik: I meant the var, but if we still need it then it's ok.

Arthurm1:
@tgodzik I think everything is addressed now and ready to merge unless you wanted me to wrap the `JavaInteractiveSemanticdb#textDocument` in a `Try` but the `javac` command itself is wrapped in a `Future` and the error handled.

tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala
```diff
@@ -13,6 +14,16 @@ import tests.MetalsTestEnrichments._
 
 class WorkspaceSymbolLspSuite extends BaseLspSuite("workspace-symbol") {
 
+  override protected def initializationOptions: Option[InitializationOptions] =
+    Some(
+      InitializationOptions.Default.copy(
```
tgodzik, line 19: I wonder if we could make something line `TestDefault` in the `BaseLspSuite` ? So that we don't repeat the init options everywhere.

tests/unit/src/test/scala/tests/debug/StepDapSuite.scala
```diff
@@ -5,7 +5,7 @@ import tests.QuickBuildLayout
 
 class StepDapSuite
     extends BaseStepDapSuite(
-      "debug-step",
+      s"debug-step",
```
tgodzik, line 8:
```suggestion
      "debug-step",
```

tgodzik (changes requested):
I think we are almost good to merge, I left some minor comments.

Arthurm1:
@tgodzik I think I've addressed everything

I've added `cleanWorkspace` to all the `WorksheetLspSuite` tests.  They were failing for me because I had an old H2 database around in the test area and it kept throwing errors.

tgodzik (approved):
LGTM!

PR #3144, opened 2021-09-20
Only set sbt dialect when file has indeed .sbt extension

tgodzik:
Previously, it would fail for scala files in /project.


ckipp01 (approved):
Ahhhh of course. Good catch! LGTM!

tgodzik:
> Ahhhh of course. Good catch! LGTM!

It took me a while to understand what is going on while working on Bloop :sweat_smile:

PR #3145, opened 2021-09-20
Reconnect to BSP server upon buildTarget/didChange

alexarchambault:


alexarchambault:
(Sorry, should have opened that as a draft, I'd like to add a test.)

metals/src/main/scala/scala/meta/internal/bsp/MetalsBspException.scala
```diff
@@ -1,6 +1,10 @@
 package scala.meta.internal.metals
 
-case class MetalsBspException(tryingToGet: String, msg: String)
-    extends Exception(
-      s"BSP connection failed in the attempt to get: $tryingToGet.  $msg"
+class MetalsBspException(
+    tryingToGet: String,
+    msg: String,
+    cause: Throwable = null
```
ckipp01, line 6: Why the change from case class to class? Also if we are storing both the `msg` and the `cause` couldn't we just store the `cause` and pull the `msg` from it when needed? I think we should also avoid the `null` default parameter unless there is a good reason to use it.

tests/unit/src/main/scala/tests/BuildServerInitializer.scala
```diff
@@ -133,3 +146,178 @@ object SbtServerInitializer extends BuildServerInitializer {
...
+      server: TestingServer,
+      client: TestingClient,
+      layout: String,
+      expectError: Boolean
+  )(implicit ec: ExecutionContext): Future[Unit] = {
+    generateBspConfig(workspace)
+    for {
+      _ <- server.initialize()
+      _ <- server.initialized()
+      // choose sbt as the Bsp Server
```
ckipp01, line 168:
```suggestion
      // choose scala-cli as the Bsp Server
```

tests/unit/src/test/scala/tests/BspBuildChangedSuite.scala
```diff
@@ -0,0 +1,74 @@
+package tests
+
+import scala.meta.internal.metals.ServerCommands
+
+import scala.util.Failure
+import scala.util.Success
+
+class BspBuildChangedSuite
+    extends BaseLspSuite("bsp-build-changed", ScalaCliBuildInitializer) {
```
ckipp01, line 9: Out of curiosity, why are we using `ScalaCliBuildInitialiazer` for this? Are we indirectly starting to add tests for Scala CLI as a way to test this change, or would this also just be testable with another existing BSP server? If the latter is true, then i'd rather do that and leave this addition to another PR that is about Scala CLI.

ckipp01 (commented):
Thanks for this @alexarchambault!

Just a couple clarifying questions right away.

metals/src/main/scala/scala/meta/internal/bsp/MetalsBspException.scala
alexarchambault: > Why the change from case class to class?

Exceptions shouldn't be case classes, because of their attached stack trace. The case class `equals` method doesn't take the stack trace into account, and wrongfully assumes different stack traces to be equal.

If `MetalsBspException` is a case class, we have
```scala
val ex1 = MetalsBspException("a", "b") // line 1
val ex2 = MetalsBspException("a", "b") // line2
ex1.getStackTrace.toSeq == ex2.getStackTrace.toSeq // false, ex1 has line 1, ex2 has line 2
ex1 == ex2 // wrongfully true
```
(`ex1` and `ex2` have different stack traces, but are made equals nonetheless.)

alexarchambault: > Also if we are storing both the `msg` and the `cause` couldn't we just store the `cause` and pull the `msg` from it when needed?

Good point.

alexarchambault: > I think we should also avoid the null default parameter unless there is a good reason to use it.

Just removed the default value, it wasn't necessary.

tests/unit/src/main/scala/tests/BuildServerInitializer.scala
alexarchambault: Fixed.

tests/unit/src/test/scala/tests/BspBuildChangedSuite.scala
alexarchambault: I needed a BSP server emitting `buildTarget/didChange` notifications. AFAIK, Bloop doesn't emit those. And quickbuild doesn't either it seems. (Not sure about sbt, but it's more heavyweight.)

The test downloads a `0.0.1` launcher, so not a nightly. And as soon as `0.0.2` is cut, we can download the class path of Scala CLI from Maven Central, so that it's even less likely to break in the future.

metals/src/main/scala/scala/meta/internal/bsp/MetalsBspException.scala
ckipp01: > Exceptions shouldn't be case classes, because of their attached stack trace. The case class equals method doesn't take the stack trace into account, and wrongfully assumes different stack traces to be equal.

Ahhh of course. 👍🏼 thanks for the explanation.

tests/unit/src/test/scala/tests/BspBuildChangedSuite.scala
alexarchambault: If there's a reasonably straightforward way of adding `buildTarget/didChange` support to quick build, I'd rather do that actually.

metals/src/main/scala/scala/meta/internal/bsp/MetalsBspException.scala
alexarchambault: (Also, I wanted to mention: I changed this so that I can detect an `InterruptedException` in the cause in the added test.)

alexarchambault:
Another thing I wanted to mention: I'm not 100% I understand what's going on with `client.workspaceDiagnostics` in the added test.

In the test, I write a file, check the diagnostics, then overwrite the file and check the diagnostics again. If I revert the order of the changes (so current final version, then current initial version), I don't get the diagnostics I expect after the overwrite (they remain empty).

But manual tests with Scala CLI just work as expected.

tests/unit/src/test/scala/tests/BspBuildChangedSuite.scala
ckipp01: > I needed a BSP server emitting buildTarget/didChange notifications. AFAIK, Bloop doesn't emit those. And quickbuild doesn't either it seems. (Not sure about sbt, but it's more heavyweight.)

No Bloop for sure doesn't. Actually thinking about this more I thought for sure sbt would, but I see during the `build/initialize` with sbt, it doesn't declare it under `capabilities`.

> If there's a reasonably straightforward way of adding buildTarget/didChange support to quickbuild, I'd rather do that actually.

Does `buildTarget/didChange` make sense actually when the build server writes to disk? I think the main reason that Bloop doesn't support it is because of that, since there is no concept of a reload or anything to trigger after a `didChange`, but rather it does a fresh export each time. Looking at `bill`, it might be somewhat straight forward to try to implement this there.

ckipp01:
Also taking a quick step back, sorry I should have asked this first. When a `buildTarget/didChange` even is recieved, why are we reconnecting to the BSP server rather than forcing a [reload](https://build-server-protocol.github.io/docs/specification.html#reload-request)? I sort of assumed that in this scenario where a change was detected in a build target that the `reload` would do basically what we want?

> The reload request is sent from the client to instruct the build server to reload the build configuration. This request should be supported by build tools that keep their state in memory. If the reload request returns with an error, it's expected that other requests respond with the previously known "good" state.

Which sort of leads me to another question. I know behind the scenes scala-cli is offloading to Bloop if I understand it correctly, so maybe reload doesn't make sense then. Is scala-cli keeping the state in memory, or just utilizing the `.bloop/` files?

tests/unit/src/test/scala/tests/BspBuildChangedSuite.scala
tgodzik: We use Bill to test out most of the build server specific stuff.

alexarchambault: Bill doesn't support dependencies AFAIK… That means, dependency support has to be added to Bill just to test that feature. I'm not sure it's worth the hassle…

alexarchambault:
Both Ammonite and Scala CLI have `buildTarget/didChange` support, but no reload support (they always just pick the new configuration, so they always reload in some sense).

But I think it's a different problem. The changes here are about making Metals aware of the new / updated build targets, not about asking the build tool to update how it sees them.

alexarchambault:
Scala CLI always makes sure the `.bloop/*.json` file it writes is in sync with its in-memory representation. (I mean, it does so right before asking Bloop to compile.) Maybe it should ask Bloop to reload actually 🤔  (although I never saw Bloop not reloading freshly updated JSON files).

alexarchambault:
So I scraped the Scala CLI stuff from the tests. (The test is now more "unitary", even though I prefer the former one, which was testing a more realistic scenario.)

I believe the CI job failure is unrelated to the changes of this PR.

ckipp01:
> Maybe it should ask Bloop to reload actually 🤔 (although I never saw Bloop not reloading freshly updated JSON files on its own).

Off the top of my head, I don't think Bloop is actually a `reload` provider.

ckipp01 (approved):
LGTM 👍🏼

alexarchambault:
(Just rebased on the latest `main`)

tgodzik (approved):
LGTM!

Issue #3135, opened 2021-09-16
Show tasty doesn't work for tasty files


tgodzik:
**Describe the bug**

When clicking on the tasty file it's not possible to currently read it.

**To Reproduce**

1. Classes directory inside `.bloop` directory
2. Click on any tasty file.
3. Run `Show tasty` command
4. See error about the file not belonging to build target.

**Expected behavior**

We should seach build targets for a class directory that belongs to it or as a fallback use the highest Scala 3 version available. I think we should always be able to read it with the highest versio? @kordyjan ?

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.10.7

**Search terms**

scala 3 tasty show

tgodzik:
We should also see if when opening tasty files in VS Code we could automatically display tasty instead. Currently it ask to open it and only if you try to open it with txt editor you will be able to run the command.

tgodzik:
Fixed by https://github.com/scalameta/metals/pull/3107

tgodzik:
OR  maybe, let's keep it open, the actual Show tasty command doesn't work yet. We should reuse the same logic as in #3107

kpodsiad:
> We should also see if when opening tasty files in VS Code we could automatically display tasty instead. Currently it ask to open it and only if you try to open it with txt editor you will be able to run the command.

Any chance that you know how to do this?

tgodzik:
> > We should also see if when opening tasty files in VS Code we could automatically display tasty instead. Currently it ask to open it and only if you try to open it with txt editor you will be able to run the command.
> 
> Any chance that you know how to do this?

Looks like we can use https://code.visualstudio.com/api/extension-guides/custom-editors

Example: https://github.com/Microsoft/vscode-extension-samples/tree/main/custom-editor-sample

kpodsiad:
Thanks @tgodzik, I'll look at them!

Issue #3136, opened 2021-09-16
Show tasty works only for classes named the same as the file


tgodzik:
**Describe the bug**

If we are in a file with multiple classes and none of them is named as the file (or we have not classes at all), Metals will not find the proper tasty file.

**To Reproduce**

1. Create a file name `Main.scala`
2. Add example contents:

```scala
class NotMain
class AlsoNotMain
```
3. Run `Show tasty` command
4. See error about not being able to find the tasty file.

**Expected behavior**

Proper file is shown based on the cursor positio.

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.10.7

**Additional context**

There are two way we can solve it:
- try to get that info by using the `ClassFinder` class, but it doesn't work perfectly for all inner classes or anonymous one
- use the same mechanism that was implemented for scala debug adapter

**Search terms**

tasty scala 3

keynmol:
After discussion on discord, it seems like it's sensitive to package vs. file structure

    Operating system: MacOS
    Editor: Neovim
    Metals version: v0.10.7


https://user-images.githubusercontent.com/1052965/133847054-23774bd8-33d3-4190-8c1d-6c4823db7058.mov

PR #3148, opened 2021-09-21
Add position to show tasty command

kpodsiad:
Closes https://github.com/scalameta/metals/issues/3135

Handling of 'show TASTy' command was merged into `FileDecodeProvider`.
It also requires cosmetic changes at metals-vscode (due to reused `DecoderResponse`)

TODO:
- [x] Fix https://github.com/scalameta/metals/issues/3135 (although opening as a text document is still necessary)
- [x] Take into account position when opening scala file (error handling is missing, but this could be easily solved by idea proposed below)
- [x] changes in metals-vscode (https://github.com/scalameta/metals-vscode/pull/707)

https://github.com/scalameta/metals/issues/3136 (and mirror issues about javap decompile) could be very easily solved by following:
- check if the given class contains multiple classes
- if yes, ask the user which class should be showed (could be done using quickpick). User picks class, there is no more ambiguity
- if not (there is only one option), show this only option

Flow is pretty straightforward and can play well with a position (if the position doesn't indicate any class, ask the user...), but I want to reduce the scope of this PR and do it in another one.


metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -19,24 +22,36 @@ import scala.meta.internal.io.FileIO
...
 
+/* Response which is sent to the lsp client. Because of java serialization we cannot use
+ * sealed hierarchy to model union type of success and error.
+ * Moreover, we cannot use Option to indicate optional values, so instead every field is nullable.
+ * */
 final case class DecoderResponse(
-    requestedUri: String,
-    value: String,
-    error: String
+    @Nullable requestedUri: String,
```
ckipp01, line 39: The `requestedUri` should always be there right? I don't believe this one needs to be `@Nullable`.

ckipp01: Also, 🙏🏼 for aligning this ha. I was working on the `nvim-metals` implementation tonight of the semanticdb/javap stuff and realized that we have a slightly different structure is returned and I was actually going to shoot in a pr making this change for the tasty stuff to just use this.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -256,19 +271,30 @@ final class FileDecoderProvider(
...
+      // we use a classFinder here because file structure can differ than actual packages
+      pathToClass <- classFinder
+        .findClass(sourceFile, position)
+        .map { className =>
+          val dir = className.replace('.', '/')
+          // only .class files contains $ in name
+          val stripped =
+            if (newExtension == "tasty") dir.stripSuffix("$")
+            else dir
+          s"$stripped.$newExtension"
```
ckipp01, line 290: So I think stuff still gets screwed up in here a bit. For example given a structure like:

```
❯ tree -L 4
.

├── src
│  └── main
│     └── scala
│        ├── Main.scala
│        └── Thing.scala
```

And inside of `Thing.scala` let's say you have:

```
package sanity

object Thing {
  val x = 3
}
```
The `sourceroot` ends up being `../src/main/scala`. So when it looks for the `className`, `findClass` here is actually returning `sanity`.  So when we try to put this together here we end up looking for `sanity.class`. So it's really close, but it really needs to instead look for `/sanity/Thing.class`.

ckipp01: Also the same thing happens if it's in a `sanity` dir.

tgodzik: Yeah, that's the same issue basically as https://github.com/scalameta/metals/issues/3136 we need to fix it in another pr I think

kpodsiad: Yeah, I've realized too that current implementation breaks on few scenarios :/ 
The decode file command had the same issues as show tasty, so I thought that this will help. But `ClassFinder`, in addition to sourceFile, takes also a position in the file. I thought that fallback to the `Position(1,1)` will work, but it turns out that it doesn't - that's why it's not working for your examples. For now, `ClassFinder` should be used only when a position is available.

Unfortunately, there are more issues :/ Now "show tasty" is very sensitive to the cursor position. ClassFinder for the following code will return just an empty string
```scala
package a.b
def foo = ()
>>
```
when for this one it will return the correct top-level name `a.b.<filename>$package` which then can be easily converted into the directory by replacing dots with slashes
```scala
package a.b
def foo = ()
>>
def foo2 = ()
```
For now, a simple "fallback to the filename" strategy (former behavior) should be sufficient if `ClassFinder` can't figure anything out.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -354,19 +380,106 @@ final class FileDecoderProvider(
...
+   */
+  def executeShowTastyCommand(uri: URI, position: l.Position): Future[Unit] =
+    if (
+      clientConfig.isExecuteClientCommandProvider() && !clientConfig
+        .isHttpEnabled()
+    ) {
+      val response = getTastyForURI(uri).map { result =>
+        DecoderResponse(
+          uri.toString(),
+          result.fold(_ => null, identity),
```
tgodzik, line 416: Will `result.orNull` work here?

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -354,19 +380,106 @@ final class FileDecoderProvider(
...
+            )
+          }
       }
-    } yield output
+
+  def getTastyForURI(
+      uri: URI,
+      cursorPosition: Option[l.Position] = None
+  ): Future[Either[String, String]] = {
+    val position = cursorPosition.getOrElse(new l.Position(1, 1))
```
tgodzik, line 446: This will most likely not work if packages are in weird places, we should instead just check the package inside the tree and create a class file name based on that.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1755,7 +1751,22 @@ class MetalsLanguageServer(
         }.asJavaObject
 
       case ServerCommands.ShowTasty() =>
-        tastyHandler.executeShowTastyCommand(params).asJavaObject
+        val parsedArgs = for {
+          args <- Option(params.getArguments.asScala)
```
tgodzik, line 1755: You can use TextDocumentPositionParams and parse it from params, or you can wait for https://github.com/scalameta/metals/pull/3149#pullrequestreview-760642240 to get merged, might improve the UX here.

Though I wonder if it will parse correctly without the position and set it to null?

Arthurm1:
As you're working on this 😉  - you could get rid of the webview and HTTP server stuff to simplify the Tasty code and client interaction.

The only benefit the webview seems to deliver is colouring and it's so simple that a textmate grammar would (probably) do the job.  I guess the downside would be that it would have to be kept in sync with Tasty printing changes :shrug:

tests/unit/src/test/scala/tests/ClassNameResolverSuite.scala
```diff
@@ -0,0 +1,203 @@
...
+import scala.meta.internal.metals.{BuildInfo => V}
+import scala.meta.internal.mtags.MtagsEnrichments._
+import scala.meta.internal.parsing.ClassFinder
+import scala.meta.internal.parsing.Trees
+import scala.meta.io.AbsolutePath
+
+import munit.FunSuite
+import munit.TestOptions
+
+class ClassNameResolverSuite extends FunSuite {
```
kpodsiad, line 19: @tgodzik @ckipp01 what do you think about how class name resolving works? It's cursor aware, but if the cursor isn't pointing to `Defn`, then it tries to fallback to the first defined class/object/definition, etc.

metals/src/main/scala/scala/meta/internal/parsing/ClassFinder.scala
```diff
@@ -82,4 +83,136 @@ class ClassFinder(trees: Trees) {
...
+
+    @tailrec
+    def loop(tree: Tree, acc: TraverseResult): TraverseResult = {
+      tree match {
+        case Source(List(pkg: m.Pkg)) => loop(pkg, acc)
+        case Source(stats) => findStat(stats, acc, acc)
+
+        case Pkg(select, (p: Pkg) :: _) =>
+          val pkg = extractPkgName(select)
+          loop(p, acc.extendPackage(pkg))
```
tgodzik, line 159: You also need to loop over objects/classes since they can be embedded in one another. We shouldn't have a totally separate logic than `findClass` and seems overall more complex than needed.

Why not just have something along the lines of
```scala
// get possible class names
def resolveClassName(
      path: AbsolutePath,
      pos: l.Position,
      newExtension: String
): List[String] = {
  findClass(path, pos) match {
    case Some(cls) => Some(cls)
    case None => findSingleClassInFile().orElse { if (isScala3) Some(s"$filename$$package.$extension") else None } 
  } 
}
```

`findSingleClassInFile` would return Some if we have only one class in file.

tests/slow/src/test/scala/tests/feature/ShowTastyLspSuite.scala
```diff
@@ -18,14 +19,62 @@ class ShowTastyLspSuite extends BaseLspSuite("showTasty") {
...
+            |}
+            |/app/src/main/scala/Main.scala
+            |package foo.bar.example
+            |class Foo
+            |class Bar
+            |""".stripMargin
+      )
+      _ <- server.didOpen("app/src/main/scala/Main.scala")
+      result <- server.showTasty(
+        s"$workspace/app/src/main/scala/Main.scala",
```
tgodzik, line 76: The test doesn't really show what should be opened in this case.

metals/src/main/scala/scala/meta/internal/parsing/ClassFinder.scala
tgodzik: Or does Tasty not produce the same structure of files as class files?

metals/src/main/scala/scala/meta/internal/parsing/ClassFinder.scala
```diff
@@ -82,4 +83,136 @@ class ClassFinder(trees: Trees) {
...
+    }
+
+    loop(tree, TraverseResult.empty)
+  }
+
+}
+
+private sealed trait Definition
+private object Definition {
+  case class Cls(name: String) extends Definition
```
tgodzik, line 197: Why add the whole class hierarchy here instead of returning a string?

tests/unit/src/test/scala/tests/ClassNameResolverSuite.scala
tgodzik: I don't think we should have any kind of fallback here, there should be an error in that case unless there is only single class in the file. It will be suprising to fallback to the first one, why not the closest one or maybe the one that is named the same as the file, or the last one? There is no 100% safe fallback here.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -256,19 +271,24 @@ final class FileDecoderProvider(
...
       target <- buildTargets.scalaTarget(targetId)
       sourceRoot <- buildTargets.inverseSourceItem(sourceFile)
-      classDir = target.classDirectory.toAbsolutePath
-      oldExtension = sourceFile.extension
-      relativePath = sourceFile
-        .toRelative(sourceRoot)
-        .resolveSibling(_.stripSuffix(oldExtension) + newExtension)
-    } yield PathInfo(Some(targetId), classDir.resolve(relativePath))
-  }
+      pathToClass <- classFinder.resolveClassName(
```
tgodzik, line 281: Shouldn't this be a different logic for tasty files and normal classfiles? We can fix tasty first and work on the other on a different PR.

kpodsiad: But resolving logic is almost the same, it only differs when it comes to inner classes.

metals/src/main/scala/scala/meta/internal/parsing/ClassFinder.scala
kpodsiad: It's leftover and it's unnecessary.

tests/slow/src/test/scala/tests/feature/ShowTastyLspSuite.scala
kpodsiad: Yeah, but previously it would fail. I'm still thinking what will be the best (the fastest) way to tests all of decompile features

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
tgodzik: Sure, but is there any way we check if it's tasty or classfile?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
ckipp01: 👍🏼 on the https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams if possible. It will make implementing this way simpler for clients.

tests/unit/src/test/scala/tests/ClassNameResolverSuite.scala
ckipp01: Yea I would also find it maybe a bit odd to fallback to the first one. I think I like the idea of the closest if that's possible.

metals/src/main/scala/scala/meta/internal/metals/Command.scala
```diff
@@ -18,6 +18,12 @@ case class Command(
 }
 
 object Argument {
+  def getParamsAsString(
```
tgodzik, line 21: You can now change this one to use the default command parsing as https://github.com/scalameta/metals/pull/3149#pullrequestreview-761777873 is merged.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
kpodsiad: `TextDocumentPositionParams` seems good, will use it!

tests/unit/src/test/scala/tests/ClassNameResolverSuite.scala
kpodsiad: Together with @tgodzik, we came to the following solution: if there is any ambiguity about which class/def metals can show, we should present all options to the user and let him pick one of them. I think it's a great idea and an easy solution to this problem

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
kpodsiad: For now, tasty and class files have different logic. Maybe asking the user to pick which class he wants to decompile would solve this problem at all.

kpodsiad:
> As you're working on this 😉 - you could get rid of the webview and HTTP server stuff to simplify the Tasty code and client interaction.
> 
> The only benefit the webview seems to deliver is colouring and it's so simple that a textmate grammar would (probably) do the job. I guess the downside would be that it would have to be kept in sync with Tasty printing changes 🤷

I agree with the webview, but I beg to differ with htpp. Metals support multiple clients and some of them (I think eclipse, sublime and emacs) use http server. @tgodzik wdyt?

tgodzik:
> > As you're working on this wink - you could get rid of the webview and HTTP server stuff to simplify the Tasty code and client interaction.
> > The only benefit the webview seems to deliver is colouring and it's so simple that a textmate grammar would (probably) do the job. I guess the downside would be that it would have to be kept in sync with Tasty printing changes shrug
> 
> I agree with the webview, but I beg to differ with htpp. Metals support multiple clients and some of them (I think eclipse, sublime and emacs) use http server. @tgodzik wdyt?

Yeah I agree, it's more reasonable to have http server support this just in case.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
kpodsiad: > The `requestedUri` should always be there right? I don't believe this one needs to be `@Nullable`.

You're right,`requestedUri`  shouldn't be `@Nullable`

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
kpodsiad: It would if `result` was `Option`, but it's an `Either` so unfortunately no.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -354,19 +387,113 @@ final class FileDecoderProvider(
...
+    val pcAndTargetUri =
+      for {
+        path <- Try(AbsolutePath.fromAbsoluteUri(uri)) match {
+          case Success(path) if !path.isFile => Left(s"$uri doesn't exist")
+          case Success(path) if path.isFile => Right(path)
+          case Failure(_) => Left(s"$uri has to be absolute")
+        }
+        pathInfo <-
+          if (path.isScala)
+            cursorPosition
```
tgodzik, line 468: So for `cursorPosition = None` we show nothing? Could we maybe use a default value for now that would work if a file is named the same as the tasty file? We could have a separate method that just finds a single or main class inside a file.

metals/src/main/scala/scala/meta/internal/parsing/ClassFinder.scala
```diff
@@ -1,8 +1,8 @@
 package scala.meta.internal.parsing
 
-import scala.meta.Defn
-import scala.meta.Pkg
-import scala.meta.Tree
+import scala.annotation.tailrec
+
+import scala.meta._
```
tgodzik, line 5: Could we keep separate imports instead of a wildcard one?

tests/slow/src/test/scala/tests/feature/ShowTastyLspSuite.scala
```diff
@@ -18,14 +19,62 @@ class ShowTastyLspSuite extends BaseLspSuite("showTasty") {
...
             |""".stripMargin
       )
       _ <- server.didOpen("app/src/main/scala/a/b/Main.scala")
       result <- server.showTasty(
-        s"$workspace/app/src/main/scala/a/b/Main.scala"
+        s"$workspace/app/src/main/scala/a/b/Main.scala",
+        new l.Position(1, 1)
+      )
+    } yield
+      if (result.isLeft) Assertions.fail("Command shouldn't failed")
```
tgodzik, line 31: Could we have an asserit if it's Right ? Also in the other ones?

tgodzik (changes requested):
Great work! This is shaping up quite nice.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -354,19 +387,113 @@ final class FileDecoderProvider(
...
+          Future.successful {
+            scribe.warn(
+              "Unable to run show tasty. Make sure `isHttpEnabled` is set to `true`."
+            )
+          }
       }
-    } yield output
+  }
+
+  def getTastyForURI(
```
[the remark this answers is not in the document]
kpodsiad: It's used by http server and `TestingServer`.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
kpodsiad: There is already `.orElse(findClassesDirFileFromSource(path, "tasty"))` which do what you want ;)
Current PR is just a small improvement in the terms of cases when a position is supplied. Otherwise, the behavior is the same as before.

tgodzik: Ach, right! Thanks!

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -251,7 +251,7 @@ class MetalsLanguageServer(
   private var symbolSearch: MetalsSymbolSearch = _
   private var compilers: Compilers = _
   private var scalafixProvider: ScalafixProvider = _
-  private var fileDecoderProvider: FileDecoderProvider = _
+  var fileDecoderProvider: FileDecoderProvider = _
```
tgodzik, line 254: Do we need to make it non private? We could just send a command instead.

tests/slow/src/test/scala/tests/feature/ShowTastyLspSuite.scala
```diff
@@ -18,14 +19,62 @@ class ShowTastyLspSuite extends BaseLspSuite("showTasty") {
...
       )
       _ <- server.didOpen("app/src/main/scala/a/b/Main.scala")
       result <- server.showTasty(
-        s"$workspace/app/src/main/scala/a/b/Main.scala"
+        s"$workspace/app/src/main/scala/a/b/Main.scala",
+        new l.Position(1, 1)
+      )
+    } yield
+      if (result.isRight) ()
+      else Assertions.fail("Command shouldn't failed")
```
tgodzik, line 32: What I meant here is to use:
```suggestion
      assertEquals(
        result,
        Right(
        """|actual tasty content
            |""".stripMArgin
       )
     )            
```
so that we can see that it's actually working

tgodzik (changes requested):
Just two more comments, I promise these are the last!

tests/slow/src/test/scala/tests/feature/ShowTastyLspSuite.scala
kpodsiad: Can we wait with this change until we switch from webview to virtual documents?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
kpodsiad: Same as below, could be wait with it until next PR? I have some idea how to improve it, but I would rather do it after bugfixes/improvements.

tests/slow/src/test/scala/tests/feature/ShowTastyLspSuite.scala
```diff
@@ -1,84 +1,285 @@
...
+        |class Foo
+        |class Bar
+        |""".stripMargin,
+    "app/src/main/scala/Main.scala",
+    new l.Position(1, 1),
+    Right(ShowTastyLspSuite.MultipleClasses)
+  )
+
+  def check(
+      testName: String,
```
kpodsiad, line 49:
```suggestion
      testName: TestOptions,
```

kpodsiad, line 49:
```suggestion
      testName: TestOptions,
```

tgodzik (approved):
LGTM!

PR #3149, opened 2021-09-21
Make commands parametrized and easier to use

tgodzik:
Previously, we would need to parse everything by hand when creating new commands. Now, the commands unapply methods contain the neccessary parsed arguments corresponding to thier definition.

I created a draft PR to see if anyone objects to the idea. It highly irritated me that command definitions are not able to help out in this case.


metals/src/main/scala/scala/meta/internal/metals/Command.scala
```diff
@@ -1,20 +1,111 @@
...
+        case parser.Jsonized(t1) =>
+          Some(t1)
+        case _ => None
+      }
+    }
+  }
+
+}
+
+case class ParametrizedCommand3[T1: ClassTag, T2: ClassTag, T3: ClassTag](
```
tgodzik, line 82: Not sure how to solve this one better, we should have used objects in all the cases, which would allow to just have `ParametrizedCommand1`

metals/src/main/scala/scala/meta/internal/metals/Command.scala
```diff
@@ -1,20 +1,82 @@
...
 
+import scala.reflect.ClassTag
+import scala.util.matching.Regex
+
 import scala.meta.internal.metals.MetalsEnrichments._
 
 import com.google.gson.JsonPrimitive
 import org.eclipse.{lsp4j => l}
 
+trait BaseCommand {
```
kpodsiad, line 10:
```suggestion
sealed trait BaseCommand {
```

metals/src/main/scala/scala/meta/internal/metals/Command.scala
```diff
@@ -1,20 +1,82 @@
...
+  def id: String
+  def title: String
+  def description: String
+  def arguments: String
```
kpodsiad, line 14: Plain `arguments` look strange to me, perhaps?
```suggestion
  def commandArguments: String
```

metals/src/main/scala/scala/meta/internal/metals/Command.scala
```diff
@@ -1,20 +1,82 @@
...
+
+  def toLSP(arguments: List[AnyRef]): l.Command =
+    new l.Command(title, id, arguments.asJava)
+
+  protected def isApplicableCommand(params: l.ExecuteCommandParams): Boolean = {
+    val command = Option(params.getCommand).getOrElse("")
```
kpodsiad, line 20: Is Option in `Option(params.getCommand)` needed if `getCommand` has `@Pure @NonNull` annotations?

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -338,18 +339,19 @@ object ServerCommands {
...
+      """|Whenever a user chooses a code action to extract a definition of a Class/Trait/Object/Enum this
+         |command is later ran to extract the code and create a new file with it
+         |""".stripMargin,
+      """|[uri, line, character], uri of the document that the command needs to be invoked on
+         |together with line number and character/column where the definition is.
+         |""".stripMargin
+    )
 
-  val InsertInferredType = new Command(
+  val InsertInferredType = new ParametrizedCommand[TextDocumentPositionParams](
```
tgodzik, line 354: I changed this one and `ExtractMemberDefinition` since this only used by the code actions (and shouldn't be done by the editor command), so we should be safe here.

metals/src/main/scala/scala/meta/internal/metals/codeactions/CodeActionCommandData.scala
```diff
@@ -5,15 +5,17 @@ import org.eclipse.lsp4j.Location
 import org.eclipse.lsp4j.TextDocumentPositionParams
 
 trait CodeActionCommandData {
-  val uri: String
+  def uri: String
   val actionType: String
```
kpodsiad, line 9: Maybe let's change this to `def` too
```suggestion
  def actionType: String
```

metals/src/main/scala/scala/meta/internal/metals/Command.scala
```diff
@@ -1,36 +1,102 @@
...
-  def getAsInt(obj: AnyRef): Option[Int] = {
-    obj match {
-      case p: JsonPrimitive if p.isNumber => Option(p.getAsInt())
-      case _ => None
+case class ParametrizedCommand[T: ClassTag](
+    id: String,
+    title: String,
+    description: String,
+    arguments: String
+) extends BaseCommand {
```
kpodsiad, line 63: I love the idea of `ParametrizedCommand`! 
Maybe at some point in time, we will be able to replace `gson` with some scala library that supports scala language constructs such as `Option`, sealed hierarchies and etc. This will give us great flexibility in the terms of how the command can look like.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -432,7 +442,7 @@ object ServerCommands {
     "Stop Ammonite build server"
   )
 
-  def all: List[Command] =
+  def all: List[BaseCommand] =
```
kpodsiad, line 445: I'd turn `all` into `Map[String, BaseCommand]`, preferably with `String` replaced by some new type like `CommandId`

kpodsiad (approved):
I've been recently thinking about how cumbersome is dealing with Command and their parameters. Now it's so much easier, awesome work on this!

metals/src/main/scala/scala/meta/internal/metals/Command.scala
tgodzik: ach, it might not be needed then, I just copied it over.

ckipp01 (approved):
I'm all for this. This is so much nicer 👍🏼

metals/src/main/scala/scala/meta/internal/metals/Command.scala
tgodzik: It's arguments in the command trait/classes, so I think this is fine.

metals/src/main/scala/scala/meta/internal/metals/Command.scala
tgodzik: Unfortunately, we would need to replace LSP4J in that case, so for now I would rather leave it as is.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
tgodzik: I think I will leave as is, I don't see a clear benefit for it and that would be a few changes to make it work.

metals/src/main/scala/scala/meta/internal/metals/codeactions/CodeActionCommandData.scala
tgodzik: Changed

metals/src/main/scala/scala/meta/internal/metals/Command.scala
tgodzik: Simplified it to not use option.

metals/src/main/scala/scala/meta/internal/metals/Command.scala
tgodzik: Added

PR #3152, opened 2021-09-22
Update Sourcegraph workflow to use `lsif-java index` command.

olafurpg:
The workflow was previously disabled due to 404s that I'm unable to
reproduce now.


olafurpg:
I tested this job manually here and the upload succeeded so I'm not sure what caused the original 404 error https://github.com/scalameta/metals/runs/3673008721?check_suite_focus=true

tgodzik (approved):
LGTM! Thanks for fixing this!

PR #3154, opened 2021-09-23
Fix go to parent code lenses for local symbols

tgodzik:
Previously, we would point only to symbols in the commands for parent code lenses, which cannot work with the local symbols. Now, in case of local symbols we use the command to go to an exact position.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodCodeLens.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodCodeLens.scala
@@ -11,7 +11,11 @@ import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.ServerCommands
 import scala.meta.internal.metals.UserConfiguration
 import scala.meta.internal.parsing.Trees
+import scala.meta.internal.semanticdb.Scala._
 import scala.meta.internal.semanticdb.SymbolInformation
+import scala.meta.internal.semanticdb.SymbolOccurrence
+import scala.meta.internal.semanticdb.TextDocument
+import scala.meta.io.AbsolutePath
 
 import org.eclipse.{lsp4j => l}
 
@@ -43,7 +47,9 @@ final class SuperMethodCodeLens(
       symbol = occurrence.symbol
       gotoSuperMethod <- createSuperMethodCommand(
         symbol,
-        search
+        search,
+        textDocument,
+        path
       ).toIterable
       range <-
         occurrence.range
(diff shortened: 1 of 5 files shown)


tgodzik:
Follow up from https://github.com/scalameta/metals/pull/3019 towards https://github.com/scalameta/metals-feature-requests/issues/190

tests/unit/src/test/scala/tests/CodeLensLspSuite.scala
```diff
@@ -203,51 +203,52 @@ class CodeLensLspSuite extends BaseCodeLensLspSuite("codeLenses") {
...
-      |
-    """.stripMargin
+  check("go-to-super-method-lenses-anonymous-class", printCommand = true)(
+    """|package a
+       |
+       |object X {
+       |  def main = {
+       |    class A { def afx(): Unit = ??? } 
+       |    val t = new A {
+       |<< afx goto-position[file:///home/tgodzik/Documents/metals/tests/unit/target/e2e/codeLenses/go-to-super-method-lenses-anonymous-class/a/src/main/scala/a/Foo.scala:4:18]>>
```
kpodsiad, line 246: local path

metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodCodeLens.scala
```diff
@@ -54,28 +60,51 @@ final class SuperMethodCodeLens(
...
-  ): l.Command = {
-    new l.Command(
-      s"${clientConfig.icons.findsuper} ${name}",
-      ServerCommands.GotoSymbol.id,
-      singletonList(symbol)
-    )
+      name: String,
+      textDocument: TextDocument,
+      path: AbsolutePath
+  ): Option[l.Command] = {
```
kpodsiad, line 86: One pair of braces is unnecessary ;)

kpodsiad (approved):
LGTM

tests/unit/src/main/scala/tests/BaseCodeLensLspSuite.scala
```diff
@@ -10,13 +10,14 @@ class BaseCodeLensLspSuite(name: String) extends BaseLspSuite(name) {
...
-      scalaVersion: Option[String] = None
+      scalaVersion: Option[String] = None,
+      printCommand: Boolean = false
   )(
       expected: String
   )(implicit loc: Location): Unit = {
     test(name) {
       cleanWorkspace()
-      val original = expected.replaceAll("<<.*>>[^a-zA-Z0-9@]+", "")
+      val original = expected.replaceAll("<<.*>>[^a-zA-Z0-9@]+$", "")
```
dos65, line 20: I was wondering why [this test](https://github.com/scalameta/metals/pull/3154/files#r716428172) passes on and found that input isn't cleaned from lenses because of wrong regex.

It should be:
```suggestion
      val original = expected.replaceAll(s"<<.*>>${System.lineSeparator()}", "")
```

tgodzik: `$` is already used for end of line, so this should be fine, no?

tgodzik: Ach, actually it is no, which is super weird.

tgodzik: Ach, silly me `$` matches end of string not end of line. Should be fixed now!

metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodCodeLens.scala
tgodzik: Removed the ones after if

tests/unit/src/test/scala/tests/CodeLensLspSuite.scala
tgodzik: Good catch! Fixed

dos65 (approved):

PR #3160, opened 2021-09-27
Tasty quickpick

kpodsiad:
Follow up of https://github.com/scalameta/metals/pull/3148
Fixes https://github.com/scalameta/metals/issues/3136
Need https://github.com/scalameta/metals-vscode/pull/710 to work.

What has been done:
- [x] uses quick-pick to fix issues with multiple classes. 
- [x] handles correctly packages
- [x] show quick-pick only when there is unambiguity about which `.tasty` file has to be shown. If there's only one possibility then skip the quick-pick step.

Next steps (not in this PR):
- use virtual document instead of webview
- add quick-pick logic to other decompile requests
- get rid of custom, "show tasty" logic in `FileDecodeProvider`


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1726,8 +1723,26 @@ class MetalsLanguageServer(
           scribe.debug(s"Executing AnalyzeStacktrace ${command}")
         }.asJavaObject
 
-      case ServerCommands.ShowTasty(uri) =>
-        tastyHandler.executeShowTastyCommand(uri).asJavaObject
+      case ServerCommands.ShowTasty(positionParams) =>
+        fileDecoderProvider.showTasty(positionParams).asJavaObject
+
+      case ServerCommands.ShowAvailableClasses(params) =>
```
ckipp01, line 1729: I wonder, is it possible to do this functionality without adding another endpoint? Couldn't we instead do this logic right in the show tasty function? Meaning if the request to show tasty comes in, if there are multiple classes for the given URI, then we send a `metals/quickPick` to get which one they want, and then show it?

tgodzik: I think we should go that way, it's always better to add less since that means less work for the clients.

kpodsiad: It's worth adding, that I want to remove the show tasty command in favor of the `DecodeFile` one and I'm looking for a solution that will suit both cases.

I have some doubts about only one endpoint approach. IMO, from the client's point of view, having two separate commands is more flexible than just one. `decompile` action consists of two separate steps: find a given resource (tasty/class/semanticDB) URI and then translate that resource into human-readable text (the true `decompile` part). 

Especially, I don't know how to handle VirtualDocuments or CustomEditors in vs code with just one command. They both require a provider, which will then provide text content for a given uri. From this point of view, having separate commands seems reasonable.

tgodzik: We can do it without the virtual document for now or not differentiate the URLs (show all the classes from one files under one URL for now.), so the quick pick will work the same in all clients.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -67,29 +67,27 @@ final class FileDecoderProvider(
    * jar:file:///jarPath/jar-sources.jar!/packagedir/file.java
    *
    * Examples...
-   * metalsDecode:file:///somePath/someFile.java.javap-verbose?decoder=javap&verbose=true
```
tgodzik, line 70: We should still be able to run on a java file.

tgodzik (commented):
Overall after the discussion in Discord I see 3 possibilities:
- add an additional command for asking about classes, but do not require to have it run before decompiling
- do not add an additional command, just in case of virtual document redirect user using `goto-location` to the new URI that can be decoded without asking the user again (.class/.tasty file)
- show all classes decompiled from a file in the same editor window one after another

I think I like the 3 idea the most since it will be easier for user to get a clear picture of the full file. This is also similar to how metap work, where pointing it to a directory will show all the semanticdb documents in there.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
kpodsiad: What about `metalsDecode:file:///somePath/someFile.java.semanticdb-compact`, is this correct? I thought that semanticDB could be produced only from scala files.

kpodsiad: I went with the first option because it requires only a few minor changes in implementation. I like the idea of `goto-location`, but this would require more changes in metals-vscode and scope of this and vscode related PR are big enough for now.

tests/slow/src/test/scala/tests/feature/FileDecoderProviderLspSuite.scala
```diff
@@ -0,0 +1,683 @@
...
+        |public final class foo.bar.example.Main$$package {
+        |  public static void foo();
+        |}
+        |""".stripMargin
+
+  private[this] val format = new SimpleDateFormat("d MMM yyy;")
+  private def javapVerbose(workspace: String) =
+    s"""|Classfile $workspace/foo/bar/example/Foo.class
+        |  Last modified ${format.format(new Date())} size 312 bytes
+        |  SHA-256 checksum fae10a1f13dd3fbccf35abc0dec9ec5189b302782a75868a5afed87f6abb1926
```
kpodsiad, line 574: Date format and algorithm used for hashing differ among JDK's :/ Perhaps it should be stripped in tests

tests/unit/src/test/scala/tests/classFinder/FindAllClasses.scala
```diff
@@ -0,0 +1,83 @@
...
+
+import java.nio.file.Paths
+
+import scala.meta.internal.metals.{BuildInfo => V}
+import scala.meta.io.AbsolutePath
+
+import munit.TestOptions
+
+class FindAllClasses extends BaseClassFinderSuite {
+
```
kpodsiad, line 12: These tests are failing on windows, I guess it's `\r\n` vs `\n` :/

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1605,6 +1604,13 @@ class MetalsLanguageServer(
         disconnectOldBuildServer().asJavaObject
       case ServerCommands.DecodeFile(uri) =>
         fileDecoderProvider.decodedFileContents(uri).asJavaObject
+      case ServerCommands.ChooseClass(params) =>
+        fileDecoderProvider
+          .chooseClassFromFile(
+            params.textDocument.getUri().toAbsolutePath,
+            params.includeInnerClasses
```
tgodzik, line 1611: Having to declare this in params here leaks the logic of understanding tasty files to the editor, which we should avoid.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -204,20 +205,22 @@ object ServerCommands {
...
-    """|If the file is a Scala 3 source, this command will try to find the relevant tasty file for it, 
-       |read it and display it in a human readable format. If the argument already points to a TASTy file, 
-       |it will be read directly.""".stripMargin,
-    """|
-       |Object with `document` and `position`, where the document is a path to a Scala 3 source or a tasty file.
+  final case class ChooseClassRequest(
+      textDocument: TextDocumentIdentifier,
+      includeInnerClasses: Boolean
+  )
+  val ChooseClass = new ParametrizedCommand[ChooseClassRequest](
```
tgodzik, line 212: Could we instead of using an additional command have a `redirect` option as a result?

metals/src/main/scala/scala/meta/internal/parsing/ClassFinder.scala
```diff
@@ -4,32 +4,74 @@ import scala.annotation.tailrec
...
+final case class ClassWithPos(
+    name: String,
+    pos: Position
+)
+
 class ClassFinder(trees: Trees) {
 
   def findClass(path: AbsolutePath, pos: l.Position): Option[String] =
-    findClass(path, pos, true)
+    findClass(path, pos, checkInnerClasses = true).filter(_.nonEmpty)
```
tgodzik, line 22: Why do you need to check for `nonEmpty` ?

metals/src/main/scala/scala/meta/internal/parsing/ClassFinder.scala
```diff
@@ -4,32 +4,74 @@ import scala.annotation.tailrec
...
 
   def findTasty(path: AbsolutePath, pos: l.Position): Option[String] =
-    findClass(path, pos, false).filter(_.nonEmpty).map(_.stripSuffix("$"))
+    findClass(path, pos, checkInnerClasses = false)
+      .filter(_.nonEmpty)
+      .map(_.stripSuffix("$"))
+
+  def findAllClasses(
```
tgodzik, line 29: Maybe:
```scala

  def findAllClasses(
      path: AbsolutePath,
      checkInnerClasses: Boolean
  ): Option[List[ClassWithPos]] = {
    for {
      tree <- trees.get(path)
    } yield {
      val definitions = new scala.collection.mutable.ArrayBuffer[ClassWithPos]()

      def name(tree: Tree) = tree match {
        case cls: Defn.Class => s"Class ${cls.name.value}"
        case cls: Defn.Object => s"Object ${cls.name.value}"
        case cls: Defn.Trait => s"Trait ${cls.name.value}"
        case cls: Defn.Enum => s"Enum ${cls.name.value}"
      }
      def loop(tree: Tree, isToplevel: Boolean = true): Unit = tree match {
        case _: Pkg | _: Pkg.Object =>
          tree.children.foreach(loop(_))
        case _: Defn.Class | _: Defn.Trait | _: Defn.Object | _: Defn.Enum =>
          definitions.append(ClassWithPos(name(tree), tree.pos))
          if (checkInnerClasses)
            tree.children.foreach(loop(_, isToplevel = false))
        case dfn: Defn.Def if isToplevel =>
          definitions.append(ClassWithPos(s"Toplevel package", dfn.pos))
          tree.children.foreach(loop(_, isToplevel = false))
        case _ =>
          tree.children.foreach(loop(_))
      }
      definitions.toList
    }
  }
```

This avoids the additional var and excessive checking in definitions.

Also, you forgot about enum.

tgodzik (changes requested):
Honestly, I feel like the quickpick approach is not that user friendly. These commands will be used by power users, so I think just concatenating the outputs would be much better in that case. It's easy to search the file for the info you want and you wouldn't need to invoke it a lot of times. 

Concatenating outputs allows us not to add an additional command, simplifies the client implementation  and removes the need for quickpick, which would futher reduce the code. 

Unless anyone believes that quickpick is the way to go I would just remove it. What do you think @dos65 @ckipp01 ?

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
ckipp01: > I thought that semanticDB could be produced only from scala files.

There were some recent changes in Bloop utilizing `lsif-java` to produce semanticDB for java files as well. https://github.com/scalacenter/bloop/pull/1534

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -285,42 +272,104 @@ final class FileDecoderProvider(
...
+      .filter(_.nonEmpty)
+    availableClasses match {
+      case Some(classes) =>
+        val pickedPosition: Future[Option[l.Position]] =
+          if (classes.size > 1) {
+            val positions =
+              classes.map(c => (c.name, c.pos.toLSP.getStart)).toMap
+            val quickPickParams = MetalsQuickPickParams(
+              classes.map(c => MetalsQuickPickItem(c.name, c.name)).asJava,
+              placeHolder = "Pick class you want to decode"
```
ckipp01, line 309:
```suggestion
              placeHolder = "Pick the class you want to decode"
```

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -285,42 +272,104 @@ final class FileDecoderProvider(
...
+   */
+  private def findPathInfoForScalaFile(
+      path: AbsolutePath,
+      includeInnerClasses: Boolean
+  ): Future[Either[String, PathInfo]] = {
+    val availableClasses = classFinder
+      .findAllClasses(path, includeInnerClasses)
+      .filter(_.nonEmpty)
+    availableClasses match {
+      case Some(classes) =>
```
ckipp01, line 302: Nit, and maybe just style preference so ignore this if it is, but it might be better to add a guard inside the match rather than `filter(_.nonEmpty)` here. The only reason I say this is that in the future someone may change this without realizing we don't guard against it in the `match`. So I'd rather leave that to the match than here. wdyt?

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -285,42 +272,104 @@ final class FileDecoderProvider(
...
+            } yield {
+              val suffix = if (!includeInnerClasses) ".tasty" else ".class"
+              val pathToTasty = className.replace('.', '/') + suffix
+              val classDir = target.classDirectory.toAbsolutePath
+              val pathToResource =
+                classDir.resolve(pathToTasty)
+              PathInfo(Some(targetId), pathToResource)
+            }
+            pathInfoOpt.toRight(
+              s"Cannot find build target for ${path.toURI.toString}"
```
ckipp01, line 335:
```suggestion
              s"Cannot find a build target for ${path.toURI.toString}"
```

ckipp01 (commented):
Nice job on this @kpodsiad!

> Honestly, I feel like the quickpick approach is not that user friendly. These commands will be used by power users, so I think just concatenating the outputs would be much better in that case. It's easy to search the file for the info you want and you wouldn't need to invoke it a lot of times.

> Concatenating outputs allows us not to add an additional command, simplifies the client implementation and removes the need for quickpick, which would futher reduce the code.

> Unless anyone believes that quickpick is the way to go I would just remove it. What do you think @dos65 @ckipp01 ?

I guess I don't have super strong feelings on it, but I also still don't understand the need for the extra command. This is probably just partly due to me not fully understanding the VS Code integration, so I digress.

One other alternative that could have been used if we don't want to fully concatenate everything would be to just find the enclosing class of the cursor and return that.

Again, all in all, there is a lot of work done here, so I'm open to having the quick pick, but again, no strong feelings on it. If we choose to remove it and just return the concatenated result, that's fine too. We could do that and see if people use the feature and then request that further fine grained support of a specific class in a file.

Note I haven't made the changes to test this yet in `nvim-metals`, but I'll do that soon and shoot back any more comments on usability from there.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -67,34 +66,38 @@ final class FileDecoderProvider(
...
-   * metalsDecode:file:///somePath/someFile.scala.semanticdb.semanticdb-compact?decoder=semanticdb&format=compact
-   * metalsDecode:file:///somePath/someFile.scala.semanticdb.semanticdb-detailed?decoder=semanticdb
+   * metalsDecode:file:///somePath/someFile.scala.semanticdb.semanticdb-compact
+   * metalsDecode:file:///somePath/someFile.scala.semanticdb.semanticdb-detailed
    *
-   * metalsDecode:file:///somePath/someFile.scala.tasty-detailed?decoder=tasty
-   * metalsDecode:file:///somePath/someFile.tasty.tasty-detailed?decoder=tasty
+   * tasty:
+   * metalsDecode:file:///somePath/someFile.scala.tasty-detailed
+   * metalsDecode:file:///somePath/someFile.tasty.tasty-detailed
```
ckipp01, line 93: One other things, we have all the details of how these uri's should be created in here, but can we either move this to where the command is defined or copy it there as well? Mainly just so that it appears [here](https://scalameta.org/metals/docs/integrations/new-editor/#decode-file) because right now it's impossible for clients to know how to integrate this without looking into the code.

ckipp01: Also, we have `tasty-detailed`, but there isn't a non detailed version right? Do we need to have the `-detailed` part?

ckipp01:
Seeing that the `quickPick` is in place and does work, I'd probably keep it. This is pretty nice to use.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
ckipp01: Also other small nit, for `javap` if we want compact we just send in `.javap`, whereas for semanticdb we send in `.semanticdb.compact`. What do you think about aligning all of these so from the outside, they are all the same meaning that instead of having to ensure for javap you use `verbose`, but for `semanticdb` you use `detailed` you'd just use the same value. So then no matter the decoder you'd choose `compact` or `detailed`? (or `verbose` if you like that better).

dos65:
@tgodzik What about having quickpick - I think it's better than input concatenation. In case of concatenation the output format will be changed that might be not obvious for users as it requires to extend its format to mark sections.

The only I would suggest is to change the approach for items scanning and rendering.
It's not fully correct for tasty case.

For example for the following source:
```scala
class Foo
object Foo
```
There is only one tasty file but quickpick suggest to choose either `Class Foo` or `Object Foo`.

I think it would be more clear if items will be shown as all tasty/ class files that belongs to this source.

Example:
```
// source
class Foo
object Foo
class Bar

// current impl for tasty and javap
Class Foo
Object Foo
Class Bar

// -- proposed

// tasty decode items
Foo.tasty
Bar.tasty

// javap decode items (description is optional)
// I think the user who is calling got used to having `$` in filenames
Foo.class    |  class Foo
Foo$.class   |  object Foo
Bar.class    |  class Bar
```

metals/src/main/scala/scala/meta/internal/parsing/ClassFinder.scala
kpodsiad: If you pass a wrong position then you'll get just an empty string. I'll move this filter to the `private findClass`

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
kpodsiad: It makes sense, will do

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
kpodsiad: Yes, there is only one way to decode `.tasty`, so the additional extension isn't necessary and can be omitted.
About aligning the rest of the commands, extensions come from options available in `javap` and `metap` commands so both keeping them as they are now and aligning them makes sense to me. I'll just leave it up to you :D

kpodsiad: Ok, I decided not to change extensions for javap and semanticdb. Now they nicely correspond to the options in these commands.
Also, although there is only one way to decode tasty file it's good to have an additional extension. Maybe `tasty-decoded` will be better than `tasty-detailed`?

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
kpodsiad: Like we agreed, we could do that later. I'll add some comment to the additional command that it'll be removed in the future.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
ckipp01: Yea I think either is probably fine.

kpodsiad:
@dos65 wdyt about this? I find `a.b.c.Foo.tasty` more readable than `a/b/c/Foo.tasty` because first one is suggesting both package and resource (tasty/class) filename at the same time. Displaying package is important for cases when there is more than 1 class with some name (e.g. it's inner class)
Screenshot 2021-10-06 at 13 27 23
(For now we don't allow to pick class which is defined inside method, for example on the picture below it would be `Bar$InnerMethod$1.class` )
Screenshot 2021-10-06 at 13 34 22

dos65:
@kpodsiad Cool! It's better now!

However, by some reason seeing `Class`, `Object`, `Enum` prefixes anyway makes me feel that smth is not as usual.
Then after a while, I realized that probably the issue is that in symbols search we use icons instead of prefixes.

Maybe it would look better if it would be written as `$(symbol-class) Foo`? ([full list of icon suffixes](https://github.com/dos65/metals-vscode/blob/33f51d10049043f382bdac4639ecdbbb3ba99de4/src/openSymbolSearch.ts#L63-L118))

The only problem with this thing is that I don't know if it would work well in non VSCode editors.

PR #3161, opened 2021-09-28
Don't search for all symbols in go-to-parent code lenses

tgodzik:
Previously, we would check all the symbols in a file if they don't have a parent, which is highly inefficient. Now, we only search in the local file, which contains all the definitions we are interested in. If there is no definition, we will never display those code lens.

This might have still been causing some issue for users. @kpodsiad reported an issue like that, this is probably the reason.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodCodeLens.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodCodeLens.scala
@@ -31,10 +31,8 @@ final class SuperMethodCodeLens(
     val textDocument = textDocumentWithPath.textDocument
     val path = textDocumentWithPath.filePath
 
-    val search = implementationProvider.defaultSymbolSearchMemoize(
-      path,
-      textDocument
-    )
+    def search(query: String) = textDocument.symbols.find(_.symbol == query)
+
     val distance = buffers.tokenEditDistance(path, textDocument.text, trees)
 
     for {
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodCodeLens.scala
```diff
@@ -31,10 +31,8 @@ final class SuperMethodCodeLens(
     val textDocument = textDocumentWithPath.textDocument
     val path = textDocumentWithPath.filePath
 
-    val search = implementationProvider.defaultSymbolSearchMemoize(
-      path,
-      textDocument
-    )
+    def search(query: String) = textDocument.symbols.find(_.symbol == query)
```
tanishiking, line 34: <del> I'm just wondering if we do this, we can't find the parent symbol who comes from another file (semanticdb), or class files 🤔 </del>
↑ Nevermind, `search` method is used just for finding the symbol on the cursor and finding overriddenSymbol is done by
https://github.com/scalameta/metals/blob/1404b26ae810f387ab784feea8b492d03c19de9a/metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodCodeLens.scala#L59

tanishiking (approved):
I could verify it works fine in my local environment, LGTM 👍

tanishiking:
> This might have still been causing some issue for users. @kpodsiad reported an issue like that, this is probably the reason.

Could you share the reported issue or the concern?

tgodzik:
> > This might have still been causing some issue for users. @kpodsiad reported an issue like that, this is probably the reason.
> 
> Could you share the reported issue or the concern?

It was just Metals taking up a lot of CPU with some of it in the SuperMethodCodeLens class, which fits what would happen wihtout this change.

PR #3162, opened 2021-09-28
Don't close source jars

tgodzik:
Previously, when running/debugging we would close previously opened source jars, which would cause all the existing paths to inside of that jar to stop working. Now we don't close them.


adpi2 (approved):
LGTM!

PR #3163, opened 2021-09-29
Try to fix flaky tests

tgodzik:


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1147,7 +1147,7 @@ class MetalsLanguageServer(
   }
 
   private def didCompileTarget(report: CompileReport): Unit = {
-    if (!isReliableFileWatcher) {
+    if (isReliableFileWatcher) {
```
dos65, line 1150: Missing not after revertion
```suggestion
    if (!isReliableFileWatcher) {
```

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -135,8 +135,7 @@ final class TestingServer(
     bspGlobalDirectories = bspGlobalDirectories,
     sh = sh,
     time = time,
-    // relying on the macOS/Windows file watchers causes flaky test failures.
-    isReliableFileWatcher = Properties.isLinux
+    isReliableFileWatcher = System.getenv("CI") == "true"
```
dos65, line 138:
```suggestion
    isReliableFileWatcher = System.getenv("CI") != "true"
```

tgodzik: :man_facepalming: Thanks!

dos65 (approved):

tgodzik:
They should be a bit less flaky, though there are some more tests on my radar that I need to take a look into.

PR #3164, opened 2021-09-29
Make sure generate-sources phase is always run for Maven

tgodzik:
Related to https://github.com/scalacenter/bloop/issues/1573

I haven't added a test for it, because protobuf mentioned in the issue needs to be installed locally.


metals/src/main/scala/scala/meta/internal/builds/MavenBuildTool.scala
```diff
@@ -22,6 +22,7 @@ case class MavenBuildTool(userConfig: () => UserConfiguration)
   def bloopInstallArgs(workspace: AbsolutePath): List[String] = {
     def command(versionToUse: String) =
       List(
+        "generate-sources",
```
dos65, line 25: I see that there is also `generate-test-sources` phase in [Lifecycle reference](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#lifecycle-reference). Probably we need to include it too.

dos65 (approved):

PR #3165, opened 2021-09-29
Updage scalafmt + use optional braces syntax for scala3

dos65:
As there is no open PR related to scala3 code, I think it's a right moment to switch it on a new syntax.


tgodzik:
I was wondering about it, but didn't think about switching it since we will then have a inconsistent code style across the codebase.

I am ok with merging it, but not sure if it's super neccessary.

dos65:
@tgodzik I understand your point about inconsistency, but I think that having an ability to work with a new syntax is more important for us. We will be able to spot some issues earlier + test indent on paste

ckipp01:
I'll be honest, and this is opinionated, but I don't like optional braces. Even after using it for a while, I find it harder to read and less clear with almost no benefit. With that being said, I don't feel strongly enough about it to have any issue at all if that's what others want. However, if we do go that route, can we use `end` in the relevant places to help out with the readability a bit.

> but didn't think about switching it since we will then have a inconsistent code style across the codebase.

I echo this sentiment as well.

Again, this is just my opinion... for an opinionated topic to begin with.

dos65:
@ckipp01 I added a setting for inserting end markers

What about the syntax in general. I found it much better but the main point for switching on it is to do dogfooding. Without it we are missing some issues/ point for improvements that might be spotted easily if you work with this syntax.

An additional thing.
This syntax will become the default one anyway - all materials from scala-center are written in this style.
Since this is a controversial topic, I think it's important to use it in publicly visible and scala-related projects to help scala-center promote it.

tgodzik:
> @ckipp01 I added a setting for inserting end markers
> 
> What about the syntax in general. I found it much better but the main point for switching on it is to do dogfooding. Without it we are missing some issues/ point for improvements that might be spotted easily if you work with this syntax.

I think we might need to change to optional braces due to this point. It's hard to discover problems with it if we never write code with the new syntax.

> An additional thing. This syntax will become the default one anyway - all materials from scala-center are written in this style. Since this is a controversial topic, I think it's important to use it in publicly visible and scala-related projects to help scala-center promote it.

Good point, but do WE want to promote it? :P I am not too keen on it myself, but well that's the direction we're kind of going with,

.scalafmt.conf
```diff
@@ -21,5 +22,8 @@ project.excludeFilters = [
 fileOverride {
   "glob:**/scala-3*/**" {
     runner.dialect = scala3
+    rewrite.scala3.convertToNewSyntax = yes
+    rewrite.scala3.removeOptionalBraces = yes
+    rewrite.scala3.insertEndMarkerMinLines = 30
```
tgodzik, line 27:
```suggestion
    rewrite.scala3.insertEndMarkerMinLines = 10
```
maybe a smaller number?

tgodzik (approved):
LGTM We should most likely merge it to try using the new style and test out the Metals feature. As an added bonus it will be easy to spot when we are in a Scala 3 code.

I would wait for more approval though to make sure we are on the same page.

.scalafmt.conf
dos65: I don't mind. Could you check locally if having it 10 it's ok for you?
Initially I set 20 but that there were much more markers than I would put manually.

dos65:
@tgodzik 
>Good point, but do WE want to promote it? :P I am not too keen on it myself, but well that's the direction we're kind of going with,

Wrong wording :smile: .
I mean that having more presence of new syntax usage + not having precedents where people reject it it's a kind of small contribution to its adoption in community.

.scalafmt.conf
keynmol: From personal experience, I copypaste the same config to all of my Scala 3 projects:

```
fileOverride {
  "glob:**/*.scala" {
    align.preset = more
    runner.dialect = scala3
    rewrite.scala3.insertEndMarkerMinLines = 15
    rewrite.scala3.removeOptionalBraces = true
    rewrite.scala3.convertToNewSyntax = true
  }
}
```

And fairly often I desire to have 10 lines (for super nested functions).

ckipp01 (approved):
> @ckipp01 I added a setting for inserting end markers

Awesome, thanks!

>  small contribution to its adoption in community.

😃 there is an assumption here that "adoption in the community" is the end goal of the community, but there are many that hope it's not.

However, not a hill I'd die on.

.scalafmt.conf
dos65: @keynmol thx for sharing! Changed to 15

PR #3167, opened 2021-09-29
Create dependency files on step if they don't exist

tgodzik:
Previously, we would only translate the URI to a local dependencies files. Now, we create the dependency file if it doesn't exist.


adpi2 (approved):
Nice!

PR #3180, opened 2021-10-04
Strip out [info] for stacktraces.

ckipp01:
This can happen at times when you are running tests and something explodes.
sbt will give you this information a la `[info]` instead of `[error]`. So
this allows the code lenses to work still without having to strip them out
manually


dos65 (approved):
LGTM!

PR #3192, opened 2021-10-10
Don't use `window/showMessage` to report generic Scalafix error

olafurpg:
Previously, generic Scalafix error messages were reported using
LSP `window/showMessage`. This was problematic because `showMessage`
should only be used for actionable high-signal error messages, whereas
Scalafix can report non-actionable error messages. This commit changes
the behavior to just log the message instead.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
@@ -111,10 +111,7 @@ case class ScalafixProvider(
         case Success(results) if !scalafixSucceded(results) =>
           val scalafixError = getMessageErrorFromScalafix(results)
           val exception = ScalafixRunException(scalafixError)
-          reportScalafixError(
-            scalafixError,
-            exception
-          )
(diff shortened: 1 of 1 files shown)


olafurpg:
Merging despite flaky failure
```
 [error] ==> X tests.CancelCompileLspSuite.basic  3.822s java.lang.AssertionError: assertion failed: expected didOpen future to fail with Cancellation Exception. If this happens frequently for unrelated changes, then this may be a flaky test that needs refactoring. If this assertion is flaky, feel free to remove it until it's refactored.
[error]     at scala.Predef$.assert(Predef.scala:223)
[error]     at tests.CancelCompileLspSuite.$anonfun$new$8(CancelCompileLspSuite.scala:41)
[error]     at tests.CancelCompileLspSuite.$anonfun$new$8$adapted(CancelCompileLspSuite.scala:33)
[error]     at scala.util.Success.$anonfun$map$1(Try.scala:255)
[error]     at scala.util.Success.map(Try.scala:213)
No more data in the client stdin, exiting...
```

PR #3194, opened 2021-10-11
[sbt server] Fix meta-build-target configuration

### LINKED ISSUE #3191: [sbt server] Navigation does not work in meta build target
**Describe the bug**

In sbt 1.5.5 we added support for the meta build target (the `SbtBuildTarget` in the [BSP specification](https://build-server-protocol.github.io/docs/extensions/sbt.html)).

Navigation does not work in Metals on this build target. That's probably because semanticdb is not enabled.

**To Reproduce**

1. In an sbt 1.5.5 or 1.6.0-M1 project
2. Import the build with sbt as the build server (Run `Metals: Switch build server` and choose `sbt`)

Metals warns that navigation will not work in `<project>-build` target.

One potential solution would be to add `semanticdbEnabled := true` in the generated `metals.sbt` file.

**Expected behavior**

Navigation should work.

**Screenshots**

image

image

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.10.7

**Search terms**

sbt server, navigation, meta build, semanticdb

adpi2:
Fixes #3191 

Add the `sbt-metals` plugin in the meta-meta-build (`project/project/metals.sbt`) to configure semanticdb in the meta-build:

```scala
// DO NOT EDIT! This file is auto-generated.

// This plugin enables semantic information to be produced by sbt.
// It also adds support for debugging using the Debug Adapter Protocol

addSbtPlugin("org.scalameta" % "sbt-metals" % "0.10.8-SNAPSHOT")

// This plugin makes sure that the JDI tools are in the sbt classpath.
// JDI tools are used by the debug adapter server.

addSbtPlugin("org.scala-debugger" % "sbt-jdi-tools" % "1.1.1")
```

Added a test to check that no mis-configuration message appears.

Tested locally:
- the semanticdb files of the `project/*.scala` files  are correctly produced. Metals can find them and use them for navigation, to find all references
- the semanticdb file of the `build.sbt` is not produced. That is because sbt compiles it with its own compiler (https://github.com/sbt/sbt/blob/develop/main-actions/src/main/scala/sbt/compiler/Eval.scala).

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/SbtBuildTool.scala
@@ -163,7 +163,7 @@ case class SbtBuildTool(
       val mainMeta = workspace.resolve("project")
       val metaMeta = workspace.resolve("project").resolve("project")
       sbtMetaDirs(mainMeta, Set(mainMeta, metaMeta)).foreach(dir =>
-        writeSinglePlugin(dir, plugin)
+        writePlugins(dir, plugin)
       )
     }
   }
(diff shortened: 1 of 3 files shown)


adpi2:
> Looks good, but it seems the CancelCompileLspSuite is failing, could it be caused by this PR?

I don't think so. Re-running all jobs to see if it goes back to normal.

dos65:
we need help :smile: 
@ckipp01 - please push the button!

PR #3200, opened 2021-10-14
Use SemanticdbPlugin in sbt-metals

### LINKED ISSUE #2990: Organize imports in test sources of subprojects does no longer work
**Describe the bug**

Organize imports in test sources of subprojects does no longer work. Build server is SBT (via BSP).

**To Reproduce**

1. have a (scala) subproject "example" where the test source directory (relative to sbt working dir) is like example/src/test/scala
2. have a source file there with at least one import statement
3. invoke the "organize imports" code action

**Expected behavior**

imports get organized :)

**Actual behavior**

tool tip "no organize import action is available shows"
error in metals log:
```
[Error - 10:45:05 AM] Request textDocument/codeAction failed.
  Message: Internal error.
  Code: -32603 
java.util.concurrent.CompletionException: scala.meta.internal.metals.ScalafixProvider$ScalafixRunException: SemanticDB not found: example/src/test/scala/Meep.scala
```

**Installation:**

- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.10.5

adpi2:
Fixes #2990

`sbt-metals` was not using the `SemanticdbPlugin` that is available by default in sbt.

I think the initial reason was that `SemanticdbPlugin` was not working so well, as shown by https://github.com/sbt/sbt/issues/5772, https://github.com/sbt/sbt/issues/5886, https://github.com/sbt/sbt/issues/5934. It is much more stable now so it is probably worth to rely entirely on it. Also it makes maintenance easier.

I added a scripted test in `sbt-metals`. The advantage of a scripted test is that if it fails, we know the problem is in sbt-metals or sbt but not in metals. Also we can easily test different project structures in the same build. So if a bug is found it is easy to reproduce.

Locally I run the scripted test on all sbt versions between (1.4.0 and 1.6.0-M1) and it passed. (for example: ```set `sbt-metals`/scriptedSbt := "1.4.0"; sbt-metals/scripted```). In the CI the scripted test runs on the current version of sbt in metals build.

What if a new bug is found in `SemanticdbPlugin`?:
- we can fix it and release a new version of sbt 
- the user can temporarily switch to Bloop as the build server

### DIFF
--- a/.github/workflows/sourcegraph.yml
+++ b/.github/workflows/sourcegraph.yml
@@ -19,6 +19,6 @@ jobs:
           mkdir -p bin
           curl -L https://sourcegraph.com/.api/src-cli/src_linux_amd64 -o bin/src
           chmod +x bin/src
-          ./bin/src lsif upload -github-token $GITHUB_TOKEN
+          ./bin/src lsif upload -trace=3 -ignore-upload-failure -github-token $GITHUB_TOKEN
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/build.sbt
+++ b/build.sbt
@@ -197,7 +197,7 @@ lazy val V = new {
   val scalameta = "4.4.28"
   val semanticdb = scalameta
   val bsp = "2.0.0-M13"
-  val bloop = "1.4.9-31-1df4194d"
+  val bloop = "1.4.9-33-c93326ba"
   val scala3 = "3.0.2"
   val nextScala3RC = "3.1.0-RC3"
   val bloopNightly = bloop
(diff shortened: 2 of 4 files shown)


adpi2:
> Actually the initial reason we didn't go that route was to have finer control over when it should be enabled

Basically in the `SemanticdbPlugin` you can enable or disable semanticdb at the project level, based on the `scalaVersion` or `bspEnabled` settings. I don't think we need a finer level of control.

> it basically boiled down to it being unsafe and/or a bad user experience for them to try it out in a project that doesn't support it

`SemanticdbPlugin` is available by default since at least sbt 1.4.0. So I don't think there is any build that support BSP but that does not support `semanticdbEnabled`. Maybe I am missing your point here...

> but we need to ensure this still protects against a user enabling this in a build where it won't work.

I added a scripted test. So if there is a build where it does not work we can easily reproduce the problem, fix it and make it not happen again.

Issue #3201, opened 2021-10-14
Snapshot build 0.10.7+94-7bd4dd41-SNAPSHOT appears to leak file handles


dickwall:
**Describe the bug**

I have been trying out version `0.10.7+94-7bd4dd41-SNAPSHOT` as it enables readable toStrings on the debugger (which works great by the way)!

However I noticed that repeated debugging sessions were failing to start with this version, and after some trial and error I noticed that the problem appeared to be `Too Many Open Files` in the Java process. This seems to be related to the SNAPSHOT build above (not sure if it is just related to the toString debug enhancement or something else, but I wanted to notify you before this version hits the mainstream).

Using standard version 0.10.7, after several restarts of the debugger, the two Java PIDs have the following file handle counts:

```
(base) dwall@L10441 ~ % lsof -p 28155 | wc -l
    2489
(base) dwall@L10441 ~ % lsof -p 28165 | wc -l
     753
```

However when running the snapshot version above, even after one debugging session, the number of file handles open in one of the processes was much higher:

```
(base) dwall@L10441 ~ % lsof -p 13335 | wc -l
    2535
(base) dwall@L10441 ~ % lsof -p 13344 | wc -l
   10749
```

This quickly exhausts any reasonable limit on the number of files per process and causes the debugger session (or any run session, including running tests) to fail.

**To Reproduce**

1. Enable the `0.10.7+94-7bd4dd41-SNAPSHOT` version in the vscode settings (Metals Server) and restart vscode/metals
2. Open a reasonably large, complex Scala project
3. Start the debugger
4. Check the number of files open from both java processes (find the java PIDs and `lsof -p <pid> | wc -l`)
5. Stop and restart the debugger a few times and watch the number growing. You may see the debugger fail to start once the number of open files is very high.

**Expected behavior**

With 0.10.7 (mainline) the number of open file handles is stable and reasonable (under 1000 even on a large project)

**Screenshots**

**Installation:**

- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.10.7 / v0.10.7+94-7bd4dd41-SNAPSHOT

**Additional context**

Making sure this is caught before it reaches v0.10.8

**Search terms**

Too many open files, debugger won't start

dos65:
I've tried to reproduce it on Metals repo but it seems that after several debug sessions the number of opened files doesn't change nor at Metals process nor at Bloop.

@dickwall Could you add main class name to processes you reported?

dickwall:
[...]

PR #3209, opened 2021-10-15
Bump Bloop and scala-debug-adapter

adpi2:
Fixes https://github.com/scalameta/metals/issues/3201

Also brings many bug fixes in the expression evaluator:
https://github.com/scalacenter/scala-debug-adapter/compare/v2.0.5...v2.0.6

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -197,7 +197,7 @@ lazy val V = new {
   val scalameta = "4.4.28"
   val semanticdb = scalameta
   val bsp = "2.0.0-M13"
-  val bloop = "1.4.9-31-1df4194d"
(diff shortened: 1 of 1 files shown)


dickwall:
Tried my original use case for the bug #3201 this morning with the `0.10.7+133-7b2fdabc-SNAPSHOT` and the file handle numbers are stable! Thanks for the quick fix.

adpi2:
@dickwall the fix is now available in `0.10.7+143-e42b771a-SNAPSHOT`
If you want to test it, make sure that it connects to the new version of Bloop which is: `1.4.9-33-c93326ba`

dickwall:
Will do. Thanks I'll let you know

PR #3210, opened 2021-10-15
Allow to generate run lenses for bsp servers which have debug capability

kpodsiad:
Previously, there was one global check if BSP server is an instance of bloop or sbt.
Now, thanks to the https://github.com/build-server-protocol/build-server-protocol/pull/161 we can check if the given buildTarget has debug capability and also remove previously used logic.


dos65:
@kpodsiad I have no idea but as understand the issue is that bloop doesn't fill these capability fields for build targets.

PR #3233, opened 2021-10-27
feature: Support Bazel as a build tool

kpodsiad:
Add support for Bazel. 
I created a simple [project](https://github.com/kpodsiad/bazel-scala) which uses Bazel and could be used for testing this integration.

Known todos:
- [ ] blocked by ~https://github.com/JetBrains/bazel-bsp/issues/105~ https://github.com/bazelbuild/rules_scala/issues/952 - without semanticDB metals capabilities are very limited. However, one can add to his BUILD scalac option which enables semanticdb plugin - `scalacopts = ["-Xplugin:<path_to_semanticdb_jar>", "-Yrangepos", "-P:semanticdb:targetroot:<path_to_generated_semanticdb>"]`
- [x] improve user experience for build servers that cannot generate bloop config (like bazel). Now, with fresh repo users have to execute `Attempt to generate bsp config for build tool` and then `switch build server`. This could be done automatically, I think. 
- [x] diagnostics aren't refreshed after compilation and they're displayed for wrong file - https://github.com/JetBrains/bazel-bsp/issues/111


tanishiking:
I added some changes so that

- Metals will automatically install `bazel-bsp 2.7.1` if `WORKSPACE` file is found && `.bsp/bazelbsp.json` isn't found.
- Metals reloads workspace if Bazel build configuration has changed (though we need changes in `metals-vscode` too, so it fires `textDocument/didSave` on change Bazel related build configurations).

tgodzik:
> I added some changes so that
> 
> * Metals will automatically install `bazel-bsp 2.7.1` if `WORKSPACE` file is found && `.bsp/bazelbsp.json` isn't found.
> * Metals reloads workspace if Bazel build configuration has changed (though we need changes in `metals-vscode` too, so it fires `textDocument/didSave` on change Bazel related build configurations).
> 
> have to fix tests

Do you need any help to finish? If you don't have time we can take over.

tanishiking:
Thanks @tgodzik maybe it's better someone picks up to take over this and finishing it up :)

## Scope
- [x] Metals should auto install `bazel-bsp` using `coursier` (`.bsp/bazelbsp.json` file should be written) for Bazel project.
  - Currently users have to manually install it https://github.com/JetBrains/bazel-bsp#easy-way-coursier
  - We should be able to detect Bazel project by checking there's a `WORKSPACE` file at project root.
  - Note that we shouldn't overwrite existing `.bsp/bazelbsp.json` because it may contain [project view setting](https://github.com/JetBrains/bazel-bsp/blob/master/executioncontext/projectview/README.md)
- [ ] Metals should show pop-up to prompt re-index the workspace when build configuration has changed.
  - I guess detecting change in `WORKSPACE`, `BUILD`, `BUILD.bazel`, `*.bzl` files are sufficient for now.
  - [ ] `metals-vscode` to fire `textDocument/didSave` on change above files: not 100% sure, but maybe we have to update https://github.com/scalameta/metals-vscode/blob/318862ece3628641b1799def7606167bbaf63068/packages/metals-vscode/package.json#L49-L57 ?
  - [x] server side, update workspace on `didSave` against above files.
- [ ] Fix tests: looks like `cs install` fails because there's no `cs` command on CI.

### Future work
- Documentation: how to use Metals on Bazel project.
- It would be great if we can bump `bazel-bsp`'s version by scala-steward or something.
  - Actually, bazel-bsp 2.7.2 is already out...

You should be able to test against a small Bazel project here: https://github.com/tanishiking/bazel-playground/tree/main/01-scala-tutorial By default, rules_scala will use Scala 2.13.6.

tanishiking:
@jkciesluk kindly took over the rest of work 💯

metals/src/main/scala/scala/meta/internal/metals/Configs.scala
```diff
@@ -37,6 +37,11 @@ object Configs {
           new FileSystemWatcher(
             Either.forLeft(s"$root/project/build.properties")
           ),
+          new FileSystemWatcher(Either.forLeft(s"$root/**/BUILD")),
+          new FileSystemWatcher(Either.forLeft(s"$root/**/BUILD.bazel")),
+          new FileSystemWatcher(Either.forLeft(s"$root/WORKSPACE")),
+          new FileSystemWatcher(Either.forLeft(s"$root/WORKSPACE.bazel")),
+          new FileSystemWatcher(Either.forLeft(s"$root/**/*.bzl")),
```
tanishiking, line 44: Ah, we have to register these settings so we make clients to fire `textDocuments/didChange`? nice catch :+1:

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
```diff
@@ -152,7 +152,8 @@ object FileWatcher {
         // However, the events are then filtered to receive only relevant events
 
         val trie = PathTrie(
-          pathsToWatch.files ++ pathsToWatch.directories
+          pathsToWatch.files ++ pathsToWatch.directories,
+          workspace,
```
tanishiking, line 156: Could you explain why/for what cases we need this change?

tanishiking (commented):
I haven't yet checked the tests and why the CI is failing, I'll continue checking tomorrow.

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -42,13 +45,16 @@ class BspConnector(
    */
   def resolve(): BspResolvedResult = {
     resolveExplicit().getOrElse {
+      val resolved = bspServers.resolve()
```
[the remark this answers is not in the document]
jkciesluk: Sorry, deleted it by mistake 😅

metals/src/main/scala/scala/meta/internal/builds/BazelBuildTool.scala
```diff
@@ -0,0 +1,146 @@
...
+  override def digest(workspace: AbsolutePath): Option[String] = {
+    BazelDigest.current(workspace)
+  }
+
+  def createBspFileArgs(workspace: AbsolutePath): Option[List[String]] =
+    Option.when(workspaceSupportsBsp(workspace))(composeArgs())
+
+  def workspaceSupportsBsp(workspace: AbsolutePath): Boolean = {
+    workspace.listRecursive.exists {
+      case file if file.filename == "WORKSPACE" => true
```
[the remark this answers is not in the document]
jkciesluk: Thanks, changed.

metals/src/main/scala/scala/meta/internal/builds/Digest.scala
```diff
@@ -97,13 +97,32 @@ object Digest {
...
+  def digestBazel(
+      file: AbsolutePath,
+      digest: MessageDigest,
+  ): Boolean = {
+    try {
+      Files
+        .readAllLines(file.toNIO)
+        .asScala
+        .mkString("\n")
+        .replaceAll("""#.*""", "") // replace any inline comment
```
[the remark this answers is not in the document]
jkciesluk: I changed so that we filter out lines that starts with `#`, this should be safer

metals/src/main/scala/scala/meta/internal/metals/Configs.scala
jkciesluk: Actually its `workspace/didChangeWatchedFiles`, ,but it works just as good

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
jkciesluk: This is not related only to bazel-bsp, but I noticed this bug while working on it. I don't remember the exact stack trace, but when bazel-bsp fails because of invalid config, the only build target uri we get is `bsp-workspace-root` (its not being a valid uri might cause some issues, but I haven't noticed any yet), and request for sources returns
```
"items": [
    {
      "target": {
        "uri": "bsp-workspace-root"
      },
      "sources": [],
      "roots": []
    }
  ]
  ```
  We then tried to create `PathTrie` with empty `paths`, which resulted in watching the whole file system.
  Basically `PathTrie` should be used as optimisation to not watch the whole workspace directory, so if the paths are empty, we fallback to watching workspace directory.

File: metals/src/main/scala/scala/meta/internal/builds/BazelDigest.scala [hunk 6]
```diff
@@ -0,0 +1,25 @@
...
+object BazelDigest extends Digestable {
+  override protected def digestWorkspace(
+      workspace: AbsolutePath,
+      digest: MessageDigest,
+  ): Boolean = {
+    workspace.listRecursive.forall {
+      // TODO: *.bzl also should be detected
+      // https://github.com/scalameta/metals/issues/5144
+      case file
+          if file.isBazelRelatedPath && !file.isInBazelBspDirectory(
```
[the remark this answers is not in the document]
jkciesluk: I think the only bazel related files in `bazel-*` directories are symbolic links to the files in our project, so there is no need

metals/src/main/scala/scala/meta/internal/metals/MetalsLspService.scala
```diff
@@ -1914,8 +1917,6 @@ class MetalsLspService(
         if (!buildTools.isAutoConnectable) {
           warnings.noBuildTool()
         }
-        // wait for a bsp file to show up
-        fileWatcher.start(Set(folder.resolve(".bsp")))
```
jkciesluk, line 1918: This was causing memory leak in tests.

If fileWatcher was started multiple times without canceling between them, only the last thread would be canceled.
Now we watch .bsp directory by default (see `Configs.scala`)

metals/src/main/scala/scala/meta/internal/builds/BazelDigest.scala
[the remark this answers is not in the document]
tanishiking: Ah, I didn't know `listRecursive` doesn't follow symbolic links
https://github.com/scalameta/metals/blob/bc0d3a0c2189d63846d704dc771afda729b14a23/mtags/src/main/scala/scala/meta/internal/mtags/ScalametaCommonEnrichments.scala#L272-L276 👍

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
tanishiking: Thanks for the explanation!
What do you think about adding these explanation to code?

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -1178,6 +1184,6 @@ object MetalsEnrichments
    * anything that is ESC (U+001B) plus [
    */
   def filterANSIColorCodes(str: String): String =
-    str.replaceAll("\u001B\\[[;\\d]*m", "")
+    str.replaceAll("\u001b\\[1A\u001b\\[K|\u001B\\[[;\\d]*m", "")
```
kasiaMarek, line 1187: I wonder if bazel it uses [cursor controls](https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797#cursor-controls) maybe we should try to cover all of them?

tests/slow/src/test/scala/tests/scalacli/ScalaCliSuite.scala
```diff
@@ -227,6 +228,18 @@ class ScalaCliSuite extends BaseScalaCliSuite(V.scala3) {
       _ <- server.initialized()
       _ = FileLayout.fromString(simpleFileLayout, workspace)
       _ = FileLayout.fromString(bspLayout, workspace)
+      _ <- server.fullServer
+        .didChangeWatchedFiles(
```
kasiaMarek, line 232: Is that needed? Shouldn't the file watcher detect the change?

kasiaMarek (commented):
A few generic comments:
 1. It would be great to have some docs about bazel suport.
 2. It seems we don't have `semanticdb` and so on for bazel, and this results in `multiple problems detected`, it'd be great to let the user know, that this is expected for bazel not actual problem with their build. So maybe a different message?
 3. When I tired changing the scala version I got:
 ```
 bazelbsp does not support `workspace/reload`, unable to reload
 ```
 in logs, and those changes weren't actually imported.  Am I missing something?

metals/src/main/scala/scala/meta/internal/metals/MetalsLspService.scala
```diff
@@ -1214,6 +1217,13 @@ class MetalsLspService(
   def didChangeWatchedFiles(
       events: List[FileEvent]
   ): Future[Unit] = {
+    def isNewBuildTool(path: AbsolutePath): Boolean = {
+      path.isInBspDirectory(folder) &&
+      path.extension == "json" &&
+      !tables.buildServers
+        .selectedServer()
+        .contains(path.filename.stripSuffix(".json"))
```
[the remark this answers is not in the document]
jkciesluk: I think it works good? If the currently chosen server is the same as the one that appeared in the `.bsp`, we don't want to call `quickConnectToBuildServer`, because it's most likely already been called in other place, but if some other build server appears we want to notify the user?

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
jkciesluk: Sounds like a good idea, but maybe we can do it in a follow up?

tests/slow/src/test/scala/tests/scalacli/ScalaCliSuite.scala
jkciesluk: File watcher detects changes from the other `didChangeWatchedFiles`, but this change is from lsp `workspace/didChangeWatchedFiles`. I'm not sure why we need both and which notifications should belong where

kasiaMarek: My understanding is that we need both, since not all editors do `workspace/didChangeWatchedFiles` (at least not reliably), so `workspace/didChangeWatchedFiles` is sort of "additional". @tgodzik, right?

metals/src/main/scala/scala/meta/internal/metals/MetalsLspService.scala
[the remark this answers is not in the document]
jkciesluk: So maybe if the newly added build tool is different from the one in `tables.buildServers`, we can clean this table and then call `quickConnectToBuildServer`?

metals/src/main/scala/scala/meta/internal/metals/MetalsLspService.scala
```diff
@@ -2011,6 +2023,18 @@ class MetalsLspService(
...
+            case (Some(_), _: BazelBuildTool) if !buildTools.isBazelBsp =>
+              BazelBuildTool
+                .maybeWriteBazelConfig(
+                  shellRunner,
+                  folder,
+                  languageClient,
+                  tables,
+                  forceImport,
+                )
+                .flatMap(_ => quickConnectToBuildServer())
```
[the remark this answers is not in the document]
jkciesluk: Won't this happen in `supportedBuildTool()`?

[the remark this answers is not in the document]
kasiaMarek: Just build tool selection not build server selection.

tests/slow/src/test/scala/tests/scalacli/ScalaCliSuite.scala
tgodzik: That's right, at least from what I remember. For build tool files we use the LSP since this should be fairly reliable.

metals/src/main/scala/scala/meta/internal/metals/MetalsLspService.scala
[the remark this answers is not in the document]
tgodzik: We can wait for someone invoking `Switch build server command` in that case I think. If they are already using a bsp server. This will be a rare enough scenario.

metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
```diff
@@ -47,7 +47,7 @@ final class BuildTools(
       .map(isBloop)
       .getOrElse(
         isBloop
-      ) || (isBsp && all.isEmpty) || (isBsp && explicitChoiceMade()) || (isBsp && isBazel)
```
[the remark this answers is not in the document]
jkciesluk: These are two different methods, first checks for toplevel `.bloop`, second looks for custom project root

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -343,7 +343,8 @@ object MetalsEnrichments
 
     def isDependencySource(workspace: AbsolutePath): Boolean = {
       (isLocalFileSystem(workspace) &&
-        isInReadonlyDirectory(workspace)) || isJarFileSystem
+        (isInReadonlyDirectory(workspace) || isInTmpDirectory(workspace))) ||
```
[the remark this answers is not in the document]
jkciesluk: In `ScalafixProvier.load()`, but looks like this check is not needed anymore, I will remove it.

metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
```diff
@@ -227,7 +229,9 @@ final class Compilations(
       connection: BuildServerConnection,
       targets: Seq[b.BuildTargetIdentifier],
   ): CancelableFuture[b.CompileResult] = {
+    val originId = "METALS-$" + UUID.randomUUID().toString
```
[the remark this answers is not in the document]
tgodzik: Ok, should be fine then.

tests/slow/src/test/scala/tests/bazel/BazelLspSuite.scala
```diff
@@ -0,0 +1,167 @@
...
+          "bazelbsp bspConfig",
+          bazelNavigationMessage,
+        ).mkString("\n"),
+      )
+      assert(bazelBspConfig.exists)
+      server.assertBuildServerConnection()
+    }
+  }
+
+  test("import-build") {
```
tgodzik, line 102: Could we add a check that the code actually compiles? I am unable to get it working locally actually, seems the dependencies are not seen. Also it's not possible to get information about a specific build target `Build target :main1 not found`

tgodzik (changes requested):

tgodzik:
It also looks like the presentation compiler is not started with the right classpath, I can't see any hover or completions from dependency module (like Hello in the sample).

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -253,6 +253,12 @@ object Messages {
     def multipleMisconfiguredProjects(count: Int): String =
       s"Code navigation will not work for $count build targets in this workspace due to mis-configuration. " + moreInfo
 
+    def multipleProblemsDetected: String =
+      s"Multiple problems detected in your build."
+
+    def bazelNavigation: String =
```
tgodzik, line 259: Could we suggest the flag to use to enable it?

jkciesluk: Currently there is no easy way to enable this, it would require changes in both metals and the users project. I think its best to work on it in a follow up.

tests/slow/src/test/scala/tests/bazel/BazelLspSuite.scala
tgodzik: What do you think about this? We can assert that there are not diagnsotics.

metals/src/main/scala/scala/meta/internal/builds/BazelBuildTool.scala
```diff
@@ -0,0 +1,184 @@
...
+  def workspaceSupportsBsp: Boolean = {
+    projectRoot.list.exists {
+      case file if file.filename == "WORKSPACE" => true
+      case _ => false
+    }
+  }
+
+  private def composeArgs(): List[String] = {
+    val classpathSeparator = java.io.File.pathSeparator
+    val classpath = Fetch
```
tgodzik, line 44: Could be a follow up, but it would be woth making sure that this gets downloaded in DownloadDependencies class.

metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
```diff
@@ -176,16 +176,30 @@ class ProblemResolver(
       futureMessage,
     ).flatten
 
+    def noSemanticDBMessages = {
```
tgodzik, line 179: Could we instead create bazel message instead of the basic semanticdb ones if the current build server is bazel?

tgodzik (approved):
Two minor things otherwise LGTM

metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
jkciesluk: If I understand correctly, we do (see `Messages.CheckDoctor.bazelNavigation`)

tgodzik: What I mean is that `noSemanticDBMessages` is used to figure out if there are no semanticdb messages, why do we need this as a fllback, instead of checking for semanticdb flags together with bazel?

jkciesluk: Ah right, I will try to change that

tgodzik (approved):
LGTM!

tgodzik:
Let's make sure we squash here, great work @jkciesluk ! :rocket:

PR #3247, opened 2021-10-31
Add CFR class file viewer

Arthurm1:
Adds a class file decompiler to the list of source analysis tools

@s5bug put me onto this while discussing an issue with Scala 3.  It was easier to understand what was going on than using `javap` so I thought it might help to integrate it into Metals

This includes a fix for using `javap` on java source files - it was looking in the wrong dir for the class file.  I haven't added any Java tests to `FileDecoderProviderLspSuite` because I don't think `didOpen` triggers a java file compilation with Metals yet, so it won't compile a single Java file and therefore can't test the analysis of the output - or am I wrong?

Other options besides [CFR](https://github.com/leibnitz27/cfr) are [fernflower](https://github.com/JetBrains/intellij-community/tree/master/plugins/java-decompiler/engine) and [procyon](https://github.com/mstrobel/procyon).  I figure if CFR isn't good enough then users can raise issues and other decompilers can be added/substituted.

I imagine this will also be handy for implementing https://github.com/scalameta/metals-feature-requests/issues/210

@ckipp01 Hopefully this is easy to add to VI.  In the VSCode PR I've also setup the `.cfr` file extension to use Java textmate grammar.  That's built into VSCode - hence I've just amended the language contribution point to use an additional extension.

License question: Do I have to include a reference to CFR in `NOTICE.md`?  I've added a dependency on CFR but not included source code.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
@@ -10,6 +10,7 @@ import java.util.Collection
 import java.{util => ju}
 import javax.annotation.Nullable
 
+import scala.annotation.tailrec
 import scala.collection.mutable.ListBuffer
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
@@ -56,11 +57,23 @@ object DecoderResponse {
   def failed(uri: URI, errorMsg: String): DecoderResponse =
     failed(uri.toString(), errorMsg)
 
+  private def getAllMessages(e: Throwable): String = {
+    @tailrec
+    def getAllMessages(e: Throwable, msgs: Vector[String]): Vector[String] = {
+      val cause = e.getCause
+      val newMsgs = msgs :+ e.getMessage
+      if (cause == e || cause == null)
+        newMsgs
+      else
+        getAllMessages(cause, newMsgs)
+    }
+    getAllMessages(e, Vector.empty[String]).mkString(Properties.lineSeparator)
+  }
   def failed(uri: String, e: Throwable): DecoderResponse =
-    failed(uri.toString(), e.getMessage)
+    failed(uri.toString(), getAllMessages(e))
 
   def failed(uri: URI, e: Throwable): DecoderResponse =
-    failed(uri, e.getMessage)
+    failed(uri, getAllMessages(e))
 
   def cancelled(uri: URI): DecoderResponse =
     DecoderResponse(uri.toString(), null, null)
@@ -399,7 +412,8 @@ final class FileDecoderProvider(
   )(path: AbsolutePath): Future[DecoderResponse] = {
     try {
       val args = if (verbose) List("-verbose") else Nil
-      val sb = new StringBuilder()
+      val sbOut = new StringBuilder()
+      val sbErr = new StringBuilder()
       shellRunner
         .run(
           "Decode using javap",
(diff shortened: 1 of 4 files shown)


Arthurm1:
Timeout on `DebugDiscoverySuite` on Windows only - spurious I assume?

s5bug:
FTR, the way I actually use CFR (I don't have it installed) is via http://www.javadecompilers.com/

Arthurm1:
I've added it into the "Metals Analyse Source" file context menu to go with the other file analysis tools.
It can be run on java, scala or class files.  Or run the "Metals: Show decompiled with CFR" command on the current java/scala file

image

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -406,6 +425,70 @@ final class FileDecoderProvider(
...
+                OutputSinkFactory.SinkClass.EXCEPTION_MESSAGE,
+                OutputSinkFactory.SinkClass.STRING
+              ).asJava
+
+            override def getSink[T](
+                sinkType: OutputSinkFactory.SinkType,
+                sinkClass: OutputSinkFactory.SinkClass
+            ): OutputSinkFactory.Sink[T] =
+              sinkType match {
+                case OutputSinkFactory.SinkType.JAVA => psOut.print(_)
```
dos65, line 455: Instead of having `ByteArrayOutputStream` and `PrintStream` there is an option just to append into list.
Smth like:
```scala
val out = List.newBuilder[String]
def appendToOut[A](a: A): Unit = 
  a match {
    case s: String => out += s
    case _ =>
  }

....
  sinkType match {
    case OutputSinkFactory.SinkType.JAVA  => appendToOut
  }
```

In this case there you can remove an additional `try/finally` block as there wouldn't be a need to close writers.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -406,6 +425,70 @@ final class FileDecoderProvider(
...
+                case OutputSinkFactory.SinkType.EXCEPTION =>
+                  _ match {
+                    case msg: SinkReturns.ExceptionMessage =>
+                      exceptions += msg.getThrownException()
+                    case f => psErr.print(f)
```
dos65, line 460: I'm wondering, in which case this branch is used?
Could it covered by a test?

dos65 (commented):
@Arthurm1 Thx for this new decoder!

Mostly looks good, I have a bit of minor comments/questions.

>  I haven't added any Java tests to FileDecoderProviderLspSuite because I don't think didOpen triggers a java file compilation with Metals yet, so it won't compile a single Java file and therefore can't test the analysis of the output - or am I wrong?

I've double-checked that - `didOpen` triggers compilation for java files.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
Arthurm1: I can't trigger it so I've removed it

Arthurm1:
@dos65 I was looking in the wrong area for class files.  I've now added tests for java files and for missing files leading to errors.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -362,50 +394,114 @@ final class FileDecoderProvider(
...
+            DecoderResponse.success(path.toURI, sbOut.toString)
+        })
     } catch {
       case NonFatal(e) =>
         scribe.error(e.toString())
         Future.successful(DecoderResponse.failed(path.toURI, e))
     }
   }
 
+  private def decodeCFRFromClassFile(
```
tgodzik, line 450: Could we run CFR using ShellRunner and download the dependency on demand? This would mean we wouldn't need to add the additional dependency to Metals.

I see there is a main that can be used https://github.com/leibnitz27/cfr/blob/9f5a97cf76d94d3ac695887abc6bbfb740c70a9c/src/org/benf/cfr/reader/Main.java

Arthurm1: Could do.  It's only 2MB but I guess it all adds up.  Is there someplace where this is already done - where I can nick the code?

tgodzik: Yeah, it all adds up and especially if it's a tool used by more advanced developers this might not be needed to download at all 

That should be for example in the NewProjectProvider, it uses gitter8 by downloading it and running with the current java home.

tgodzik: Here to be specific: 
 https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/builds/NewProjectProvider.scala#L107

Arthurm1: Done - `ShellRunner#runJava` made this very easy to change.

tgodzik (approved):
LGTM!

PR #3258, opened 2021-11-02
Add support for Scala 2.13.7

tgodzik:


ckipp01:
In case someone looks here and is wondering about the status, we are needing this fixed https://github.com/scalacenter/bloop/issues/1598

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -93,6 +93,7 @@ case class ScalaPresentationCompiler(
     workspace: Option[Path] = None
 ) extends PresentationCompiler:
 
+  //I should see name + age in autocomplete
```
ckipp01, line 96: Is this just leftover or can we put more context for what this is referring to?

ckipp01 (approved):
Just a small note on the comment, but other than that LGTM!

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
tgodzik: leftover, I was testing out the other issue and forgot to remove part of the tested code :sweat:

Issue #3262, opened 2021-11-02
"debug"/"debug test" not working on VS Code within Windows


marciocg:
**Describe the bug**

Debug session freezes and doesn't executes properly.
**To Reproduce**

1. Go to any .scala file
2. Click on 'debug' or 'debug test'
3. See error

Debug toolbar is inactive and keys F10/F11 can't be used on debug session. The line with breakpoint isn't attached for inspection.

Screenshot of the disabled debug toolbar:
![image](https://user-images.githubusercontent.com/4554906/139964452-1d406753-4e22-4680-bf82-4d65f1a255c9.png)

- Operating system: Windows
- Editor: Visual Studio Code
- Metals version: v0.10.8 and it's newest snapshots until 02/11

**Additional context**
There's an **IllegalArgumentException** on the .metals/metals.log 
[metals.log](https://github.com/scalameta/metals/files/7464485/metals.log)
:

dos65:
Thanks for reporting!
This one is already fixed in #3238 and will be released soon

tgodzik:
> Thanks for reporting! This one is already fixed in #3238 and will be released soon

Actually, just checked and it's a different error https://github.com/scalameta/metals/pull/3263

Should have fixed it together.

PR #3263, opened 2021-11-03
Fix issues when source path is using Windows format

tgodzik:
Fixes https://github.com/scalameta/metals/issues/3262

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathAdapter.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathAdapter.scala
@@ -44,7 +44,7 @@ private[debug] final class SourcePathAdapter(
     sourceUri.getScheme match {
       case "jar" =>
         Option(AbsolutePath(Paths.get(sourceUri)).toFileOnDisk(workspace))
-      case "file" => Some(sourcePath.toAbsolutePath)
+      case "file" => Some(AbsolutePath(Paths.get(sourceUri)))
       case _ => None
     }
   }


adpi2 (approved):
LGTM!

How is it possible that those bugs are not catched by:
```

CI / windows-latest jdk-adopt@1.11 unit tests 1 / 2 (pull_request)
CI / windows-latest jdk-adopt@1.11 unit tests 2 / 2 (pull_request)
```

tgodzik:
> LGTM!
> 
> How is it possible that those bugs are not catched by:
> 
> ```
> 
> CI / windows-latest jdk-adopt@1.11 unit tests 1 / 2 (pull_request)
> CI / windows-latest jdk-adopt@1.11 unit tests 2 / 2 (pull_request)
> ```

I think we operate on URIs and paths only in the tests :thinking: VS Code in Windows sends in the breakpoints as paths starting with `C:`

metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathAdapter.scala
```diff
@@ -44,7 +44,7 @@ private[debug] final class SourcePathAdapter(
     sourceUri.getScheme match {
       case "jar" =>
         Option(AbsolutePath(Paths.get(sourceUri)).toFileOnDisk(workspace))
-      case "file" => Some(sourcePath.toAbsolutePath)
+      case "file" => Some(AbsolutePath(Paths.get(sourceUri)))
```
dos65, line 47: Maybe extract `AbsolutePath(Paths.get(sourceUri))` into value as it's used in both cases?

Also, `toFileOnDisk`  might be used even for local files - in this case it won't create anything.

dos65 (approved):
LGTM

adpi2:
> I think we operate on URIs and paths only in the tests thinking VS Code in Windows sends in the breakpoints as paths starting with C:

What about those 2 methods? They are supposed to adapt the path of the client so that Metals only deals with uri. So maybe it should have been called somewhere.

https://github.com/scalameta/metals/blob/db5f9e235607b9c80d387031d683f5b6a5968414/metals/src/main/scala/scala/meta/internal/metals/debug/ClientConfigurationAdapter.scala#L45-L59

tgodzik:
> > I think we operate on URIs and paths only in the tests thinking VS Code in Windows sends in the breakpoints as paths starting with C:
> 
> What about those 2 methods? They are supposed to adapt the path of the client so that Metals only deals with uri. So maybe it should have been called somewhere.
> 

I think so, I will take a look next week to double and triple check everything and merge with the current status.

PR #3265, opened 2021-11-03
Add try in order not to break debugging session

tgodzik:
While testing on windows I encounter still some issues for classes within `src.zip` JDK sources.

It was throwing FileSytemNotFound exceptions and I was not able to figure out what was going on.

This is a quickfix for this release so that it's possible to use breakpoints properly on Windows, the only thing this will do is that it might sometimes be impossible to go to JDK sources.


adpi2 (approved):

tgodzik:
Raised the issue here: https://github.com/scalameta/metals/issues/3266

I need to properly look into all the issues around this, but that might wait.

PR #3267, opened 2021-11-03
SuperMethodCodeLens - do not show stale lenses

dos65:
Previosuly, lenses wasn't erased if document was changed in buffer only.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodCodeLens.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codelenses/SuperMethodCodeLens.scala
@@ -49,7 +49,7 @@ final class SuperMethodCodeLens(
       ).toIterable
       range <-
         occurrence.range
-          .flatMap(r => distance.toRevised(r.toLSP))
+          .flatMap(r => distance.toRevisedStrict(r).map(_.toLSP))
           .toList
     } yield new l.CodeLens(range, gotoSuperMethod, null)
   }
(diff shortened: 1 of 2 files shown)


tests/unit/src/test/scala/tests/SuperMethodLspSuite.scala
```diff
@@ -223,6 +228,77 @@ class SuperMethodLspSuite extends BaseLspSuite("gotosupermethod") {
...
+            |""".stripMargin
+      }
+      code2 <- server.codeLensDirect("a/src/main/scala/a/A.scala")
+      _ = assertNoDiff(
+        code2,
+        """|package a
+           |trait X {
+           |  def foo: Int
+           |}
+           |//case class Y(foo: Int) extends X
```
dos65, line 290: Before this change `<< foo>>`  was returned for this commented code

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -841,6 +841,20 @@ final class TestingServer(
     } yield TextEdits.applyEdits(textContents(filename), textEdits.toList)
   }
 
+  def codeLensDirect(filename: String): Future[String] = {
```
tgodzik, line 844: Why not use the existing `codeLenses` method ?

tgodzik (approved):
Looks good! Just one comment

tests/unit/src/main/scala/tests/TestingServer.scala
dos65: hmm, now I can return back to existing one
At the begging I wanted to test that there no lenses returned which leads to failure if existing one is used. But then changed test data.

tgodzik: that would be cool, since that class has already way more methods than needed.

tgodzik (approved):
LGTM!

PR #3271, opened 2021-11-04
Finer grained rules on build errors for debug discovery.

### LINKED ISSUE #3219: TestFile functionality should work if build target and the ones it depends on have no errors.
**Describe the bug**

Let's say we have 3 build targets.

```
A
B <-- This relies on A
C <-- This relies on B
```

If there are errors in C, but none in B and A and you try to trigger a test on B via something like this:

```
[Trace - 00:08:05 PM] Received request 'workspace/executeCommand - (2434)'
Params: {
  "command": "metals.debug-adapter-start",
  "arguments": [
    {
      "runType": "testFile",
      "path": "file:///ScalaSpecInB.scala"
    }
  ]
}
```
You won't be able to test it if there is an error in C. 

```
[Trace - 00:08:05 PM] Sending notification 'metals/status'
Params: {
  "text": " Errors in workspace",
  "show": true,
  "tooltip": "Cannot run or debug due to existing errors in the workspace. Please fix the errors and retry.",
  "command": "metals-diagnostics-focus"
}


[Trace - 00:08:05 PM] Sending notification 'metals/executeClientCommand'
Params: {
  "command": "metals-diagnostics-focus",
  "arguments": []
}
```

Right now we check the entire build for errors here:

https://github.com/scalameta/metals/blob/c34ab1aad6441fe0597f64678d2ef0395836e9bc/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala#L445-L448

I think if possible this check should be finer grained to not check for errors in the entire build client.

**Expected behavior**

I'd expect to be able to test in build target B here.

ckipp01:
This pr makes the check for errors in the build a bit more specific when
doing debug discovery. In the past if there was an error at all in the
build, we would error and notify the user they couldn't run/test the
file. However, instead of looking for errors in the entire build we can
instead use the build target and check that build target for errors
(which also checks the transitive build dependencies) which allows you
to run or test a file when there are unrelated errors.

Fixes #3219

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
@@ -294,7 +294,7 @@ class DebugProvider(
       buildTargetClasses.classesOf(bti).testClasses
 
     val result: Future[DebugSessionParams] = (runTypeO, buildTargetO) match {
-      case _ if buildClient.buildHasErrors =>
+      case (_, Some(buildTarget)) if buildClient.buildHasErrors(buildTarget) =>
         Future.failed(WorkspaceErrorsException)
       case (_, None) =>
         Future.failed(BuildTargetNotFoundForPathException(path))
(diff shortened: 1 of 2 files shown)


dos65 (approved):
LGTM!
Faced this problem recently by myself and glad to see that it's fixed.

PR #3273, opened 2021-11-05
Update organize-imports rule with a fix for givens

tgodzik:


ckipp01 (approved):
LGTM!

dos65 (commented):
@tgodzik I took a look how to make it work with scala3 a did a [PR](https://github.com/tgodzik/metals/pull/12) on your branch

.scalafix3.conf
```diff
@@ -4,6 +4,7 @@ rules = [
 
 OrganizeImports.groupedImports = Explode
 OrganizeImports.expandRelative = true
+OrganizeImports.removeUnused = false
```
ckipp01, line 7: This is because warn unused doesn't work with Scala 3 right?

dos65: yep + by default it's `true` that fails the internal validation for scala3 projects

dos65 (approved):

ckipp01 (approved):
LGTM!

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -202,8 +202,22 @@ case class ScalafixProvider(
       .getOrElse("Unexpected error while running Scalafix.")
   }
 
-  private def scalafixConf: Option[Path] = {
+  private lazy val scala3DefaultConfig = {
```
tgodzik, line 205: Turns out that Scala organize imports will fail with an empty confix for Scala3, so I added a temporary config file.

PR #3276, opened 2021-11-09
[Scala 3] Add selection range

tgodzik:
Previously, when selecting range in Scala 3 we would only get the usual Hover. Now, we should get the type of the selected expression.

This also includes expanding the type tree in Hover to find the exact generic type as it's done for Scala 2.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -328,12 +328,20 @@ case class ScalaPresentationCompiler(
       given ctx: Context = driver.currentCtx
       val pos = sourcePosition(driver, params, uri)
       val trees = driver.openedTrees(uri)
-      val path = Interactive.pathTo(trees, pos)
-      val enclosing = expandRangeToEnclosingApply(path, pos)
-      val tp = enclosing.head.tpe
-      val tpw = tp.widenTermRefExpr
 
-      if tp.isError || tpw == NoType || tpw.isError then ju.Optional.empty()
+      def typeFromPath(path: List[Tree]) =
+        if path.isEmpty then NoType else path.head.tpe
+
+      val path = Interactive.pathTo(trees, pos)
+      val tp = typeFromPath(path)
+      val tpw = tp.widenTermRefExpr
+      // For expression we need to find all enclosing applies to get the exact generic type
+      val enclosing = expandRangeToEnclosingApply(path, pos)
+      val exprTp = typeFromPath(enclosing)
+      val exprTpw = exprTp.widenTermRefExpr
+
+      if tp.isError || tpw == NoType || tpw.isError || path.isEmpty then
+        ju.Optional.empty()
       else
         Interactive.enclosingSourceSymbols(enclosing, pos) match
           case Nil =>
@@ -365,15 +373,12 @@ case class ScalaPresentationCompiler(
 
             val docString =
               docComments.map(_.renderAsMarkdown).mkString("\n")
-            val expressionType = printer.expressionTypeString(tpw, history)
-            val includeExpression = params match
-              case p: RangeParams => p.offset != p.endOffset
-              case _ => false
+            val expressionType = printer.expressionTypeString(exprTpw, history)
             val content = HoverMarkup(
               expressionType,
               hoverString,
               docString,
-              includeExpression
+              !pos.span.isZeroExtent
             )
             ju.Optional.of(new Hover(content.toMarkupContent))
 
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -31,6 +31,7 @@ import scala.meta.pc.DefinitionResult
 import scala.meta.pc.OffsetParams
 import scala.meta.pc.PresentationCompiler
 import scala.meta.pc.PresentationCompilerConfig
+import scala.meta.pc.RangeParams
 import scala.meta.pc.SymbolSearch
 import scala.meta.pc.VirtualFileParams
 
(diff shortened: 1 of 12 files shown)


mtags/src/main/scala/scala/meta/internal/pc/HoverMarkup.scala
```diff
@@ -17,9 +17,7 @@ object HoverMarkup {
       forceExpressionType: Boolean = false
   ): String = {
     val markdown = new StringBuilder()
-    val needsExpressionType =
```
tgodzik, line 20: This caused the changes in HoverDocSuite, but I don't think we should show expression in that case (it also causes some issues with Scala 3, which are currently hard to deal with). We should most likely fix the types in for `emptyList`.

tests/cross/src/test/scala/tests/hover/RangeHoverSuite.scala
```diff
@@ -96,7 +102,9 @@ class RangeHoverSuite extends BaseHoverSuite {
       "2.13" ->
         """|IndexedSeq[Int]
            |override def flatMap[B](f: Int => IterableOnce[B]): IndexedSeq[B]""".stripMargin.hoverRange,
-      "3" -> "x: Int".hoverRange
+      "3" ->
+        """|IndexedSeq[Int]
+           |def flatMap[B](f: A => IterableOnce[B]): IndexedSeq[Int]""".stripMargin.hoverRange
```
tgodzik, line 107: So in all of the cases the return type is alright, but I haven't managed to figure out how to change the actual function params. I think we can work on it further on.

mtags/src/main/scala/scala/meta/internal/pc/HoverMarkup.scala
tgodzik, line 20: This caused the changes in HoverDocSuite, but I don't think we should show expression in that case (it also causes some issues with Scala 3, which are currently hard to deal with). We should most likely fix the types in for `emptyList`.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -284,6 +285,35 @@ case class ScalaPresentationCompiler(
           ).selectionRange().asJava
       }
     }
+  end selectionRange
+
+  def expandRangeToEnclosingApply(
```
tgodzik, line 290: This was needed to get the actual type and not the generic type. I also caused the changes to hover (which are not 100% satifactory, but I think it's still better.)

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -299,51 +329,57 @@ case class ScalaPresentationCompiler(
...
-                      printer.hoverDetailString(symbol, history, tpw)(using
-                        context
-                      )
-                    case None =>
-                      val history = new ShortenedNames(IndexedContext(ctx))
-                      printer.hoverDetailString(symbol, history, tpw)
-            }
-            val content = hoverContent(hoverString, docComments)
-            ju.Optional.of(new Hover(content))
+                  printer.hoverDetailString(symbol, history, tpw)
```
dos65, line 363:
```suggestion
                  printer.hoverDetailString(symbol, history, symbol.info)
```
Here it's needed to put tpe of symbols instead of expression to render symbol signature correctly (without precise resulting type). And probably the changes in `HoverMarkup` won't be required.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -299,51 +329,57 @@ case class ScalaPresentationCompiler(
...
+            val expressionType = printer.expressionTypeString(tpw, history)
+            val includeExpression = params match
+              case p: RangeParams => p.offset != p.endOffset
+              case _ => false
+            val content = HoverMarkup(
+              expressionType,
+              hoverString,
+              docString,
+              includeExpression
+            )
```
dos65, line 377:
```suggestion
            val content = HoverMarkup(
              expressionType,
              hoverString,
              docString,
              !pos.span.isZeroExtent
            )
```

dos65 (changes requested):
Looks good except of changes in `Symbol signature` rendering.

I took a look at it and left a suggestion - should be easy to fix and make it work in the same way as for Scala2

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
tgodzik: This seems to break some other tests. I could just use type generated from `path` for signature, what do you think?

`HoverMarkup` change will still be needed as we would get a different typ for signature and expression.

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
tgodzik: Done!

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
tgodzik: I reverted the changes to the expressions, so this PR should only influence Scala 3 expression types.

dos65: I started looking what is wrong with just `symbol.info`.

The main difference was in this sample:
```scala
Lis@@t(1)
// Scala2 symbol signature: def apply[A](a: A*): List[A]
// Scala3 symbol signature: def apply[A](a: A*): CC[A]
```

Then I take a look at `HoverProvider` in Scala2 and realized that we haven't ported [`seenFromType`](https://github.com/scalameta/metals/blob/db3ae6638a8c16093aa3de5cb06c8e2e1a8b4f0d/mtags/src/main/scala-2/scala/meta/internal/pc/HoverProvider.scala#L135-L151) logic.

I added a kind of `seenFromType` + reverted the changes from HoverMarkup in [this wip commit](https://github.com/dos65/metals/commit/20cb4e805bb7cba2ce78ad42b24dcc70abd6e6e4) to validate if it's a key and got more close result to Scala2.

There are still failed tests but I think they are valid and we need to adjust object/class/trait .. rendering.
Could you look this commit? 
Honestly, it's in half-working state so I you want I can complete it a bit later.

mtags/src/main/scala/scala/meta/internal/pc/HoverMarkup.scala
dos65: Could we revert this and related changes in Scala2 tests and disable `needExpressionType` fallback only for Scala3?
Otherwise, we will have do it later.

I think the case in `HoverDocSuite` is valid, expression type is needed there. I agree that this logic is a bit weird but it works well. Or maybe we can move this clause to HoverProvider in Scala2 and pass `forceExpressionType = pos.start != po.end || !symbolSignature.endsWith(expressionType)` ?

dos65 (commented):
I think it's good to merge it as it is.

I have only one suggestion

tgodzik:
@dos65 I reverted the changes to Scala 2 as suggested, let me know if it's good to merge!

dos65 (approved):
Looks good!
Thx for your patience!

tgodzik:
> Looks good! Thx for your patience!

There was absolutely no hurry, so it was good to think over everything here, thanks!

PR #3277, opened 2021-11-09
Add test discovery endpoint

kpodsiad:
Adds endpoint which returns information about tests in targets in projects. This information can be consumed by LSP clients.

Needs:
- [metals-vscode](https://github.com/scalameta/metals-vscode/pull/755) - ready for review
- [scala debug adapter](https://github.com/scalacenter/scala-debug-adapter/pull/157) (sbt support) - merged
- [bloop](https://github.com/scalacenter/bloop/pull/1607) (bloop support) - merged

my-app-1637768302619

Tests are run using a DAP server - just like code lenses are. In addition, if it's possible (e.g. running tests for the whole project), tests are batched and run using a single DAP session because creating the DAP session takes some time. 

Information about text execution is passed from Build Server (for now it's only implemented for Bloop) to the DAP server, then the server forwards this event to the DAP client. The event contains sufficient information about the execution of both the whole test suite and particular test cases.

There are a few things which would be great to have:
- [x] better outputs for failed test classes. IMO I did my best on this one. The only source of information is [Event from `sbt-test-interface`](https://github.com/sbt/test-interface/blob/master/src/main/java/sbt/testing/Event.java) and there is no sufficient information for diff or other fancier stuff. For now, we'll have to live with plain `throwable.getMessage()`
Screenshot 2021-11-16 at 13 47 28

It turned out that this can be done even better, every test can have its own section in the test suite summary.
Screenshot 2021-12-02 at 19 52 51

- [x] finer granulation of test classes (grouped by packages, not only by project)
Screenshot 2021-11-24 at 16 34 01

- [ ] finer granulation of tests - now the whole test class is the smallest piece. Discovering a single test shouldn't be that hard, but there is an issue with running single tests. Namely, it's very hard at the moment without any changes in e.g. [bsp](https://github.com/scalameta/metals-feature-requests/issues/216). I have to think about it and I hope I'll come back with some proposition how to address this problem.


ckipp01:
This looks super cool @kpodsiad! I haven't looked super close yet, but what does come to mind right away is that I want to see if it's possible to use with with `nvim-metals` and `nvim-dap`. I'm surprised at how little code there is in here and taking a look at the testing API I'm assuming VS Code is doing a whole ton of stuff behind the scenes. Any chance you'd be willing to capture the `dap-client` and `dap-server` logs for one of these runs? It will help me understand what's happening under the hood and better help me figure out if this is possible for something non-vscode to utilize this without a ton of code.

Arthurm1:
Excellent stuff @kpodsiad.  I look forward to the day I don't have to scroll through log output to see why tests are failing.

The additional features you talk about would be great.  But even if it's not currently possible to run individual tests - I think viewing finer granulation of test results would be immensely helpful.  In the tree view - having each individual test result as a leaf node below its test suite node and then being able to select them and see the expected/obtained/diff would speed up the development cycle.  To be able to then re-run only that failing test would be icing on the cake.

kpodsiad:
> In the tree view - having each individual test result as a leaf node below its test suite node 

It isn't easy since the only information Metals got from the build server is a test's class name. Using the class name we should be able to determine to which test framework class belongs and then apply some logic to extract singular tests position and name. Fact that this extraction logic has to be implemented per each test framework worries me, but I don't know if there's a better approach

> In the tree view - having each individual test result as a leaf node below its test suite node 

It isn't easy since the only information Metals got from the build server is a test's class name. Using the class name we should be able to determine to which test framework class belongs and then apply some logic to extract singular tests position and name. Fact that this extraction logic has to be implemented per each test framework worries me, but I don't know if there's a better approach

> and then being able to select them and see the expected/obtained/diff would speed up the development cycle. 

Assuming that the former part is already done this should be relatively easy, at least partially. Thanks to changes in scala-debug-adapter and bloop, both sbt and bloop will send "detailed" info about each test. "detailed" because for errors it will only contain `error: string;` so we'll be able to display it next to the test. Currently, joined info about errors is displayed for the whole test suite (you can see it in one of the images).

> To be able to then re-run only that failing test would be icing on the cake.

Say no more! I have one idea of how to pursue this, but I don't know how what the outcome will be since probably it requires changes in a few places :/

Out of curiosity, what test framework do you use most frequently @Arthurm1 ?

Arthurm1:
@kpodsiad I use JUnit 5 for java code and Scalatest for scala code.  Sadly Bloop doesn't support Junit 5 [see here](https://github.com/scalacenter/bloop/issues/996) so I use Intellij.

It's probably fine to rely on the test framework's text output.  JUnit displays actual & expected.  MUnit does as well if I remember correctly.  The main thing I would like is the ability to click on a failed test and have the results of that specific test appear (whatever format that is) rather than having to scroll through logs to find out why that test failed.  I don't know about you but I find something like Github's way of displaying test results quite slow to drill into/navigate and that's what I would like to get away from here.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -799,6 +800,42 @@ final class TestingServer(
...
+            server.compilations.compileFile(path)
+          } else {
+            val error =
+              s"Could not fetch any test classes in $refreshCount tries"
+            testClasses.tryFailure(new NoSuchElementException(error))
+          }
+        }
+    }
+
+    Thread.sleep(300)
```
kpodsiad, line 827: Without those `Thread.sleep`s sometimes even 10 attempts was not enough and result was empty :/

tests/unit/src/test/scala/tests/testProvider/DiscoverTestClassesSuite.scala
```diff
@@ -0,0 +1,109 @@
...
+  def apply(discovery: TestDiscovery): TestDiscoveryWrapper =
+    TestDiscoveryWrapper(
+      discovery.targetName,
+      discovery.discovered.asScala
+        .map(x => TestDiscoveryResultWrapper(x))
+        .toList
+    )
+}
+
+sealed trait TestDiscoveryResultWrapper
```
kpodsiad, line 82: It's quite a lot of boilerplate, but asserting logic is easier. Maybe someone has a better proposition

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestDiscovery.scala
```diff
@@ -0,0 +1,46 @@
...
+      val fullyQualifiedName: String,
+      val className: String,
+      @Nullable val location: l.Location
+  ) extends TestDiscoveryResult
+
+  object TestClass {
+    def apply(pkg: String, testName: String, @Nullable location: l.Location) =
+      new TestClass("class", pkg, testName, location)
+  }
+}
```
kpodsiad, line 46: It's the best way to model discriminated union type which I discovered. These classes could be `case` classes but then the proper value of `kind` won't be enforced because of generated apply.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
```diff
@@ -0,0 +1,109 @@
...
+
+import scala.meta.internal.implementation.ImplementationProvider
+import scala.meta.internal.metals.BuildTargets
+import scala.meta.internal.metals.DefinitionProvider
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.debug.BuildTargetClasses
+
+import ch.epfl.scala.bsp4j.BuildTarget
+
+trait TestSuitesFinder {
```
ckipp01, line 11: More of just a curiosity, there is only one implementation of this. Are there plans for another? Or was was the reason to provide this interface over just a single concrete class?

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestDiscovery.scala
```diff
@@ -0,0 +1,45 @@
+package scala.meta.internal.metals.testProvider
+
+import javax.annotation.Nullable
+
+import scala.meta.internal.metals.MetalsEnrichments._
+
+import org.eclipse.{lsp4j => l}
+
+final case class TestDiscovery(
```
ckipp01, line 9: Nit

```suggestion
final case class TestDiscoveryResult(
```

wdyt? No strong feeling on my part, I think it makes it a bit clearer though.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Partitions the given testEntries depending on whether entry has nonempty package or not.
+   * Mapped those empty ones into TestClass, strip package for nonempty ones and repeat process for them.
+   */
+  private def groupTestsByPackageImpl(
+      testEntries: List[TestEntry]
+  ): List[TestDiscovery.Result] = {
+    val (withPackage, withoutPackage) =
+      testEntries.partition(e => e.packageParts.nonEmpty)
```
ckipp01, line 97:
```suggestion
      testEntries.partition(entry => entry.packageParts.nonEmpty)
```

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
```diff
@@ -0,0 +1,109 @@
...
+    val currentTestClasses = withoutPackage.map(_.testClass)
+    val testClassesInPackages: Iterable[TestDiscovery.Result] = withPackage
+      .groupBy(p => p.packageParts.head)
```
ckipp01, line 100:
```suggestion
      .groupBy(entry => entry.packageParts.head)
```

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
```diff
@@ -0,0 +1,109 @@
...
+      .mapValues(_.map(p => p.copy(packageParts = p.packageParts.drop(1))))
```
ckipp01, line 101:
```suggestion
      .mapValues(_.map(entry => entry.copy(packageParts = entry.packageParts.drop(1))))
```

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
```diff
@@ -0,0 +1,109 @@
...
+      location <- definition.toResult.flatMap(
+        _.locations.asScala.toList
+          .filter(_.getUri == definition.uri)
+          .headOption
+      )
+    } yield location
+    val fullyQualifiedName = className.split('.').toList
+    val testClass = TestDiscovery.TestSuite(
+      className,
+      fullyQualifiedName.takeRight(1).head,
```
ckipp01, line 66: Do we have a guarantee here that this will never be Nil?

tests/unit/src/main/scala/tests/TestingServer.scala
ckipp01: I had similar issues in the past with the sbt server suite and endup adding the `waitFor` method for this exact purpose, but instead used it in the for comprehension of the actual spec.

ckipp01 (approved):
Very cool @kpodsiad! 🚀 

Once I get a bit more time I'd like to see if it's possible for us to support this in `nvim-metals`. I love to have something like this.

Just a few nits, no blockers from my side.

kpodsiad:
> Once I get a bit more time I'd like to see if it's possible for us to support this in `nvim-metals`. I love to have something like this.

Even if it is not possible to add some structured view about I think that other editors can still benefit from this and following PR's. Namely, if Metals will be able to discover single tests inside the test suite and https://github.com/build-server-protocol/build-server-protocol/issues/249 will be added and implemented by build servers then Metals' clients can `run code lenses` per single test, which will be freaking awesome. I'm so psyched about this possibility.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
kpodsiad: There aren't plans for others implementations. 
Some time ago I played with this approach (trait + implementation) and I liked it. I find it very easy to reason about what code should do, because there is just a simple trait with all public API gathered -- without implementation overhead -- where you can quickly search & find methods. Sometimes, when there is so much code, quick look over API can tell more about class responsibilities than code. 

I know, it's a little bit strange to abstract single class, but I think it's worth to give this a try.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
kpodsiad: Yes, `fullyQualifiedName` contains package + class name, so the wort case scenario is when package is empty, but there's still class name and this is exactly what this piece of code is looking for. 
Some renaming will be very helpful, for example `val fullyQualifiedName = className` looks very suspicious for me now.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestDiscovery.scala
kpodsiad: Maybe even `TestSuitesDiscoveryResult` since it discovers whole suites?

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
ckipp01: 👍🏼 cool,  thanks for explaining.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestDiscovery.scala
ckipp01: Maybe sure, I think so it's more clear that this is the result I think it'd be helpful. No strong feelings on the exact name.

tests/unit/src/main/scala/tests/TestingServer.scala
kpodsiad: `server.waitFor(TimeUnit.SECONDS.toMillis(10))` seems working, we'll see if it'll work in the CI. Thanks!

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
kpodsiad: Now it should be clearer

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
```diff
@@ -341,7 +341,7 @@ final class ImplementationProvider(
     }
   }
 
-  private def findSymbolInformation(
+  def findSymbolInformation(
```
tgodzik, line 344: This should be private, or alternatively added to `Semanticdbs`.

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -119,6 +119,13 @@ final class DefinitionProvider(
     }
   }
 
+  def toPath(
```
tgodzik, line 122: Why `toPath` ? It's returning DefinitionDestination.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -42,7 +42,8 @@ case class UserConfiguration(
     enableStripMarginOnTypeFormatting: Boolean = true,
     enableIndentOnPaste: Boolean = false,
     excludedPackages: Option[List[String]] = None,
-    fallbackScalaVersion: Option[String] = None
+    fallbackScalaVersion: Option[String] = None,
+    disableTestCodeLenses: Boolean = false
```
tgodzik, line 46: Can we just have it as an enum
```suggestion
    testUserInterface: TestUserInterfaceType = TestUserInterfaceType.Lenses
```
?

`disableTestCodeLenses` doesn't really correspond to what is show on UI.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
tgodzik: Though, the drawback of this approach is that every time you use go to definition, you will later need to find implementations.

I am also not sure if it's less clear if a user looks at the implementation. It's exactly same line.

There is no benefit to seeing:
```scala
def findTestSuites(): Seq[TestSuiteDiscoveryResult]
```

separate from
```scala
override def findTestSuites(): Seq[TestSuiteDiscoveryResult] = {
    buildTargets.allBuildTargetIds.toList
    ....
```

I agree if there were to be a lot more methods, but with just one let's just have a class.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
```diff
@@ -0,0 +1,115 @@
...
+
+trait TestSuitesFinder {
+  def findTestSuites(): Seq[TestSuiteDiscoveryResult]
+}
+
+final class TestSuitesFinderImpl(
+    buildTargets: BuildTargets,
+    buildTargetClasses: BuildTargetClasses,
+    definitionProvider: DefinitionProvider,
+    implementationProvider: ImplementationProvider
```
tgodzik, line 19:
```suggestion
```
It's not used.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuiteDiscoveryResult.scala
```diff
@@ -0,0 +1,45 @@
...
+
+object TestSuiteDiscoveryResult {
+  sealed trait Discovered {
+    def nonEmpty: Boolean = this match {
+      case p: Package => p.children.asScala.nonEmpty
+      case _: TestSuite => true
+    }
+  }
+
+  final class Package private (
```
tgodzik, line 23: Why do we need package? Can't we have just a list of TestSuite ? This would leave us with just one class instead of hierarchy.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
```diff
@@ -0,0 +1,115 @@
...
+) extends TestSuitesFinder {
+
+  override def findTestSuites(): Seq[TestSuiteDiscoveryResult] = {
+    buildTargets.allBuildTargetIds.toList
+      .flatMap(buildTargets.info)
+      .filterNot(_.getDisplayName.endsWith("build"))
```
tgodzik, line 25: You should instead check if it's sbt build target.

Normal module can also use `build` in their name.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
```diff
@@ -0,0 +1,115 @@
...
+      .map(buildTarget => createTestDiscoveryFor(buildTarget))
+      // don't send empty results
+      .filter(_.discovered.asScala.exists(_.nonEmpty))
+  }
+
+  private def createTestDiscoveryFor(buildTarget: BuildTarget) = {
+    val classes = buildTargetClasses
+      .classesOf(buildTarget.getId)
+      .testClasses
+      // filter out symbols which don't represent classes
```
tgodzik, line 35: I think there can be tests which are objects, no ? Is there a need for this filtering?

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  private def createTestEntry(
+      buildTarget: BuildTarget,
+      symbol: String,
+      fullyQualifiedClassName: String
+  ): TestEntry = {
+    val location = for {
+      definition <- definitionProvider
+        .toPath(symbol, List(buildTarget.getId))
+      location <- definition.toResult.flatMap(
```
tgodzik, line 57: This should all be in `toPath` I think - that would justify the name.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
```diff
@@ -0,0 +1,115 @@
...
+      testEntries: List[TestEntry]
+  ): List[TestSuiteDiscoveryResult.Discovered] = groupTestsByPackageImpl(
+    testEntries
+  )
+
+  /**
+   * Partitions the given testEntries depending on whether entry has nonempty package or not.
+   * Mapped those empty ones into TestClass, strip package for nonempty ones and repeat process for them.
+   */
+  private def groupTestsByPackageImpl(
```
tgodzik, line 96: We could just send a list of test suites, which would simplify the logic. Client can decide later on how to represent it.

tests/unit/src/test/scala/tests/testProvider/DiscoverTestSuitesSuite.scala
```diff
@@ -0,0 +1,116 @@
...
+import scala.meta.internal.metals.testProvider.TestSuiteDiscoveryResult
+
+import tests.TestDiscoveryWrapper._
+
+class DiscoverTestSuitesSuite extends BaseLspSuite("discoverTestSuites") {
+
+  test("munit-test-suite") {
+    for {
+      _ <- Future.successful(cleanWorkspace())
+      _ <- initialize(s"""|/metals.json
```
tgodzik, line 18:
```suggestion
      _ <- initialize(
        s"""|/metals.json
```

tests/unit/src/test/scala/tests/testProvider/DiscoverTestSuitesSuite.scala
```diff
@@ -0,0 +1,116 @@
...
+                          |}
+                          |/app/src/main/scala/foo/bar/MyTestSuite.scala
+                          |package foo.bar
+                          |abstract class AbstractTestSuite1 extends munit.FunSuite 
+                          |
+                          |class MunitTestSuite extends AbstractTestSuite1 {
+                          |  test("foo") {}
+                          |}
+                          |""".stripMargin)
+      _ <- server.server.compilations.compilationFinished(
```
tgodzik, line 33:
```suggestion
      _ <- server.didSave("app/src/main/scala/foo/bar/MyTestSuite.scala")
```

This should be enough and will trigger compilation.

tests/unit/src/test/scala/tests/testProvider/DiscoverTestSuitesSuite.scala
```diff
@@ -0,0 +1,116 @@
...
+        workspace.resolve("app/src/main/scala/foo/bar/MyTestSuite.scala")
+      )
+      _ <- server.executeCommand(ServerCommands.CascadeCompile)
```
tgodzik, line 36: There is only one project so no need for this

tests/unit/src/test/scala/tests/testProvider/DiscoverTestSuitesSuite.scala
```diff
@@ -0,0 +1,116 @@
...
+      _ <- server.didOpen("app/src/main/scala/foo/bar/MyTestSuite.scala")
+      _ <- server.didSave("app/src/main/scala/foo/bar/MyTestSuite.scala")(
+        identity
+      )
+      _ <- server.waitFor(TimeUnit.SECONDS.toMillis(10))
```
tgodzik, line 41:
```suggestion
```

This should not be needed at all.

tests/unit/src/test/scala/tests/testProvider/DiscoverTestSuitesSuite.scala
```diff
@@ -0,0 +1,116 @@
...
+          )
+        )
+      )
+
+      assertEquals(expected, obtained)
+    }
+  }
+}
+
+final case class TestDiscoveryWrapper(
```
tgodzik, line 74: Why do you need to wrap it ? Can't you just use TestSuiteDiscoveryResult.TestSuite ? This seems to be duplciating the hierarchy again.

OR you can even have a json result as string to show what the client receives.

tgodzik (changes requested):
Great work @kpodsiad  

I left a bunch of comments, I think we can simplify a couple of things.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
kpodsiad: > I am also not sure if it's less clear if a user looks at the implementation. It's exactly same line.

There are also constructor arguments so there is still some cognitive overhead, but I agree that in this case it's negligible and placing `override def`s at the beginning of the class is good enough.

Though I already have a feature branch with an additional method and I guess there's potential for two more :P

tgodzik: Honestly, you can also just use `outline` in VS Code which allows a better view of what's going on in the class.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuiteDiscoveryResult.scala
kpodsiad: Could you provide an example? I think it's doable to get rid of hierarchy, but I don't see how to model following with just List, without Package at all.
```
TestSuite1
a.TestSuite2
a.b.TestSuite3
```

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
kpodsiad: Yeah, I know. I was just waiting for a goo advice how to handle that 😎

tests/unit/src/test/scala/tests/testProvider/DiscoverTestSuitesSuite.scala
kpodsiad: Because TestSuiteDiscoveryResults are just classes, without `case` which means that they don't have defined equals or toString method. In addition, they contain `java.util.List` because of GSON. 

I wrote that it's a lot of boilerplate, but it was only for quickly developing and testing things. I think that comparing strings should work well, I like this idea.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuiteDiscoveryResult.scala
kpodsiad: But then the client will have to map this flat structure into the hierarchy itself. I assumed that it's easier to do this once on the server-side, because the client can always flat this back into the list and grouping is slightly harder.

To sum up, I don't mind moving this logic to the client side.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
kpodsiad: I'll have to look at it once again, but from what I remember every suite was listed twice. Once as a class and te second time as an object :/

metals/src/main/scala/scala/meta/internal/implementation/ImplementationProvider.scala
kpodsiad: This will be needed in later PR's so I'll make it private again.

tests/unit/src/test/scala/tests/testProvider/DiscoverTestSuitesSuite.scala
tgodzik: It shouldn't be this way together with the refresh handler, which should actually solve that issue

kpodsiad: I'll move it to the refresh handler

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuiteDiscoveryResult.scala
kpodsiad: @ckipp01 @ayoub-benali How this looks like from the other editors' side? What would be a better format for nvim or sublime to process?

kpodsiad: Code shouldn't be vscode specific and should be easy to reuse by other editors as well.

tests/unit/src/test/scala/tests/testProvider/DiscoverTestSuitesSuite.scala
kpodsiad: `scalafmt` doesn't agree with your proposal

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
kpodsiad: Nah, it's something else. My problem was about `what is this class doing/what is public API` vs `how this class is doing it` and not about listing methods :/

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
kpodsiad: Ok, so here is how it looks like for most popular frameworks like scalatest, junit, munit and utest. There are two symbols, one for class and one for an object which corresponds to the same class name. However, only one of them has a location in code -- and that one will be kept while the second one will be filtered out.
```shell
"a/MunitTestSuite., a.MunitTestSuite"
"a/MunitTestSuite#, a.MunitTestSuite"
```

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuiteDiscoveryResult.scala
```diff
@@ -0,0 +1,35 @@
...
+    targetUri: String,
+    discovered: java.util.List[TestSuiteDiscoveryResult.Discovered]
+)
+
+object TestSuiteDiscoveryResult {
+  sealed abstract trait Discovered {
+    def nonEmpty: Boolean = this match {
+      case p: Package => p.children.asScala.nonEmpty
+      case _: TestSuite => true
+    }
```
tgodzik, line 18: Since we leave the hierarchy why not just implement the method in subclasses?
This look like a workaround for overloading.


```suggestion
    def nonEmpty: Boolean = true
```

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuiteDiscoveryResult.scala
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  final case class Package(
+      prefix: String,
+      children: java.util.List[Discovered]
+  ) extends Discovered {
+    val kind = "package"
```
tgodzik, line 25:
```suggestion
    val kind = "package"
    override def nonEmpty = children.asScala.nonEmpty
```

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuiteDiscoveryResult.scala
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  final case class TestSuite private (
+      fullyQualifiedName: String,
+      className: String,
+      location: l.Location
+  ) extends Discovered {
+    val kind = "suite",
```
tgodzik, line 33:
```suggestion
    val kind = "suite"
```

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
```diff
@@ -0,0 +1,103 @@
...
+          testClass
+        )
+    }
+  }
+
+  private def groupTestsByPackage(
+      testEntries: List[TestEntry]
+  ): List[TestSuiteDiscoveryResult.Discovered] = groupTestsByPackageImpl(
+    testEntries
+  )
```
tgodzik, line 70: This method does not look needed, you can just use the below one.
```suggestion
```

tests/unit/src/test/scala/tests/testProvider/TestSuitesFinderSuite.scala
```diff
@@ -0,0 +1,87 @@
...
+      _ <- initialize(s"""|/metals.json
+                          |{
+                          |  "app": {
+                          |    "libraryDependencies" : [ "org.scalameta::munit:0.7.29", "org.scalatest::scalatest:3.2.1" ],
+                          |    "scalaVersion": "2.13.6"
+                          |  }
+                          |}
+                          |/app/src/main/scala/foo/bar/MyTestSuite.scala
+                          |package foo.bar
+                          |abstract class AbstractTestSuite1 extends munit.FunSuite
```
tgodzik, line 26: Could you change the test to contain different packages and nesting? I think it's fine to adjust this one. We should also add no package suite

tgodzik (changes requested):
Just some small comments and we should probably add a bit more testing.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuiteDiscoveryResult.scala
kpodsiad: You're suggesting that pattern matching is worse than overloading? I think they're equal :D 
Anyway, it's thanks to your comment I improved what sealed hierarchy looks, now I don't have a `kind` field in subclasses which is great.

tgodzik (approved):
LGTM! Great work!

PR #3278, opened 2021-11-09
ScalaVersion - discover new mtags version by coursier

dos65:
Prevously, the specific Metals version was able to work with only
specified list of Scala versions.

With these changes, in case if new Scala versions was released, we will
be able to realese only mtags artifact instead of making a full Metals
realease.

This isn't really usefull without additional release workflows but I'll
make separate PR for that later.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MtagsResolver.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MtagsResolver.scala
@@ -7,76 +7,86 @@ import scala.util.control.NonFatal
 
 import coursierapi.error.SimpleResolutionError
 
-class MtagsResolver {
-
-  private val states =
-    new ConcurrentHashMap[String, State]()
-
-  def resolve(scalaVersion: String): Option[MtagsBinaries] = {
-    def logError(e: Throwable): Unit = {
-      val msg = s"Failed to fetch mtags for ${scalaVersion}"
-      e match {
-        case _: SimpleResolutionError =>
-          // no need to log traces for coursier error
-          // all explanation is in message
-          scribe.error(msg + "\n" + e.getMessage())
-        case _ =>
-          scribe.error(msg, e)
-      }
-    }
-
-    def fetch(tries: Int = 0): State = {
-      try {
-        val jars = Embedded.downloadMtags(scalaVersion)
-        State.Success(MtagsBinaries.Artifacts(scalaVersion, jars))
-      } catch {
-        case NonFatal(e) =>
-          logError(e)
-          State.Failure(System.currentTimeMillis(), tries)
-      }
-    }
-    def shouldResolveAgain(failure: State.Failure): Boolean = {
-      failure.tries < State.maxTriesInARow ||
-      (System.currentTimeMillis() - failure.lastTryMillis) > 5.minutes.toMillis
-    }
-
-    // The metals_2.12 artifact depends on mtags_2.12.x where "x" matches
-    // `mtags.BuildInfo.scalaCompilerVersion`. In the case when
-    // `info.getScalaVersion == mtags.BuildInfo.scalaCompilerVersion` then we
-    // skip fetching the mtags module from Maven.
-    if (MtagsBinaries.isBuildIn(scalaVersion)) {
-      Some(MtagsBinaries.BuildIn)
-    } else {
-      val computed = states.compute(
-        scalaVersion,
-        (_, value) => {
-          value match {
-            case null => fetch()
-            case succ: State.Success => succ
-            case failure: State.Failure =>
-              if (shouldResolveAgain(failure))
-                fetch(failure.tries + 1)
-              else {
-                scribe.info(s"No mtags for ${scalaVersion}.")
-                failure
-              }
-          }
-        }
-      )
-      computed match {
-        case State.Success(v) => Some(v)
-        case _: State.Failure => None
-      }
-    }
-  }
-
-  def isSupportedScalaVersion(version: String): Boolean =
+trait MtagsResolver {
+  def resolve(scalaVersion: String): Option[MtagsBinaries]
+  final def isSupportedScalaVersion(version: String): Boolean =
     resolve(version).isDefined
+}
 
-  sealed trait State
-  object State {
-    val maxTriesInARow: Int = 2
-    case class Success(v: MtagsBinaries.Artifacts) extends State
-    case class Failure(lastTryMillis: Long, tries: Int) extends State
+object MtagsResolver {
+
+  def default(): MtagsResolver = new Default
+
+  class Default extends MtagsResolver {
+
+    private val states =
+      new ConcurrentHashMap[String, State]()
+
+    def resolve(scalaVersion: String): Option[MtagsBinaries] = {
+      def logError(e: Throwable): Unit = {
+        val msg = s"Failed to fetch mtags for ${scalaVersion}"
+        e match {
+          case _: SimpleResolutionError =>
+            // no need to log traces for coursier error
+            // all explanation is in message
+            scribe.error(msg + "\n" + e.getMessage())
+          case _ =>
+            scribe.error(msg, e)
+        }
+      }
+
+      def fetch(tries: Int = 0): State = {
+        try {
+          val jars = Embedded.downloadMtags(scalaVersion)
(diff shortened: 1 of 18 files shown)


metals/src/main/scala/scala/meta/internal/metals/MtagsResolver.scala
```diff
@@ -0,0 +1,64 @@
...
+import java.util.concurrent.ConcurrentHashMap
+
+import scala.util.control.NonFatal
+
+import coursierapi.error.SimpleResolutionError
+
+class MtagsResolver {
+
+  private val states =
+    new ConcurrentHashMap[String, Either[Long, MtagsBinaries]]()
```
kpodsiad, line 12: I'll appreciate some comment, e.g. `Contains successfully resolved mtags binary or timestamp of last failed result`

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
```diff
@@ -141,19 +144,25 @@ class ProblemResolver(
   private def findProblem(
       scalaTarget: ScalaTarget
   ): Option[ScalaProblem] = {
+
+    // We don't publish all mtags for unit test
+    // So check only versions that comes from build-info
+    def isSupportedScalaVersion(version: String): Boolean =
+      if (!allSupportedMtagsRequired)
+        ScalaVersions.isSupportedAtReleaseMomentScalaVersion(version)
+      else mtagsResolver.isSupportedScalaVersion(version)
```
kpodsiad, line 153: Doesn't `requireAllSupportedMtags` sound better? Also, could you expand your comment? I'm afraid I don't get it.

metals/src/main/scala/scala/meta/internal/metals/MtagsResolver.scala
```diff
@@ -0,0 +1,64 @@
+package scala.meta.internal.metals
+
+import java.util.concurrent.ConcurrentHashMap
+
+import scala.util.control.NonFatal
+
+import coursierapi.error.SimpleResolutionError
+
+class MtagsResolver {
```
tgodzik, line 9: Could this be an object? Since we are anyway using `ConcurrentHashMap`

That would make sure that we are always using the same resolver.

metals/src/main/scala/scala/meta/internal/metals/MtagsResolver.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  private val states =
+    new ConcurrentHashMap[String, Either[Long, MtagsBinaries]]()
+
+  def resolve(scalaVersion: String): Option[MtagsBinaries] = {
+    def fetch(): Either[Long, MtagsBinaries] = {
+      try {
+        val jars = Embedded.downloadMtags(scalaVersion)
```
tgodzik, line 17: Maybe we could do a retry also just in case?

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
tgodzik: This doesn't seem to be needed I think? We only run unit tests for 2.12.x

tgodzik (changes requested):
Great work! This should help out with easier releases, thanks!

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
dos65: I wasn't happy about adding this thing but there are some tests in unit that uses scala 3 sources.
At least [this one](https://github.com/scalameta/metals/blob/main/tests/unit/src/test/scala/tests/NewFileLspSuite.scala#L330)

I can look at it one more time and modify tests to get rid of it.

metals/src/main/scala/scala/meta/internal/metals/MtagsResolver.scala
dos65: I don't think that it's good idea :smile: 
Then this state will be attached to JVM instance instead of LanguageServer.

tgodzik: Isn't de facto the case? `MtagsResolver` is only created once really in each JVM.

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
tgodzik: The change here shouldn't cause it to fail, it doesn't actually use the presentation compiler. Why did it start to fail?

We can for sure move it to slow if needed.

metals/src/main/scala/scala/meta/internal/metals/MtagsResolver.scala
dos65: I think it better to keep it per-server-instance rather that per-JVM-instance to avoid cache invalidation issue that is a classical programming challenge. 
I understand that it won't be an issue right now. However, I've seen the similar things too many times to get the strong feeling that it should be avoided.

For example. In case if we wouldn't have `fork := true` for tests, then it will be a problem if tests were run as `sbt unit/test cross/test` in a single JVM.

metals/src/main/scala/scala/meta/internal/metals/MtagsResolver.scala
dos65: Good point. Will change it a bit to retry it in case if was a networking exception or smth like that.

metals/src/main/scala/scala/meta/internal/metals/MtagsResolver.scala
```diff
@@ -0,0 +1,64 @@
...
+            case _: SimpleResolutionError =>
+            case _ =>
+              scribe.error(s"Failed to fetch mtags for ${scalaVersion}", e)
+          }
+          Left(System.currentTimeMillis())
+      }
+    }
+    def refetch(millis: Long): Boolean = {
+      val left = System.currentTimeMillis() - millis
+      left > 1000 * 60 * 5
```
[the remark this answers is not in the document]
dos65: thx!

kpodsiad (approved):

metals/src/main/scala/scala/meta/internal/metals/MtagsResolver.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  def default(): MtagsResolver = new Default
+
+  /**
+   * The default MtagsResolver always checks if mtags-artifacts for particular scala version exists.
+   * However this strict logic doesn't work for tests.
+   * For `unit` we don't publish mtags at all but there are some tests that trigger docker check.
+   * So for these cases, do fallback on previous mechanic by checking declared supported versions.
+   */
+  def forUnitTests(): MtagsResolver =
```
dos65, line 26: Anyway I had to add this for unit tests.
There is a suite that check if messages for support/unsupported versions are produced correctly and more other.

Moving it in slow won't solve the problem as we don't publish only specfic range of versions for these tests.

tests/unit/src/main/scala/tests/BaseLspSuite.scala
```diff
@@ -48,6 +49,9 @@ abstract class BaseLspSuite(
 
   protected def initializationOptions: Option[InitializationOptions] = None
 
+  protected def mtagsResolver: MtagsResolver =
+    MtagsResolver.forUnitTests()
```
tgodzik, line 53: We could have the actual implementation kept here.

tgodzik (changes requested):

tests/unit/src/main/scala/tests/BaseLspSuite.scala
dos65: Initially I defined it here but then I realized that it's needed to be used in ProblemSuite too. 
So I decided that it would be better to move it into MtagsResolver . Just to make it more clear that there is such instance.

tgodzik: Sure, though I would keep it in tests if possible

dos65: Ok, moved to tests

PR #3279, opened 2021-11-10
Adjust json trace docs

kpodsiad:
follow up of https://github.com/scalameta/metals/pull/3103


docs/contributors/getting-started.md
```diff
@@ -186,36 +186,38 @@ These logs contain information that may be relevant for regular users.
...
+type nul > $WORKSPACE/.metals/bsp.trace.json # build server
+type nul > $WORKSPACE/.metals/dap-server.trace.json # debug adapter
+type nul > $WORKSPACE/.metals/dap-client.trace.json # debug adapter
 ```
 
 Next when you start Metals, watch the logs with `tail -f`.
 
 ```sh
 # macOS
-tail -f ~/Library/Caches/org.scalameta.metals/lsp.trace.json
```
tgodzik, line 216: We could join both cases as they are exactly the same.

docs/contributors/getting-started.md
```diff
@@ -186,36 +186,38 @@ These logs contain information that may be relevant for regular users.
...
 
-To see trace of incoming/outgoing JSON communication with the text editor or
-build server, create empty files in your machine cache directory.
+**Important notice**  
+Since Metals 0.10.10 version log files should be placed in `$WORKSPACE` rather than your machine cache directory.
+
+To see trace of incoming/outgoing JSON communication with the text editor or build server, create empty files in `$WORKSPACE/.metals/` directory.
 
 ```sh
 # macOS
```
tgodzik, line 195: We could join MacOS and Linux cases. Maybe even remove Windows case, just mention that you can use `type nul > ` to create the empty trace file?

docs/contributors/getting-started.md
```diff
@@ -186,36 +186,38 @@ These logs contain information that may be relevant for regular users.
 
 ### JSON-RPC trace
 
-To see trace of incoming/outgoing JSON communication with the text editor or
-build server, create empty files in your machine cache directory.
+**Important notice**  
+Since Metals 0.10.10 version log files should be placed in `$WORKSPACE` rather than your machine cache directory.
```
tgodzik, line 190: We should mention what was `your machine cache directory` for each OS

tgodzik (changes requested):
Thanks for updating the docs, I think we could simplify this a bit now.

dos65:
btw, I missed the initial PR with this change. 
I thought that traces in cache would still working and writing into `$WORSPACE/.metals` is just an additional option.
Wdyt of returning traces in cache as a fallback if there is no trace file in `$WORKSPACE/.metals`?

For me it was easier to check traces there whenever I spot some strange behavior in any project.

kpodsiad:
> btw, I missed the initial PR with this change. I thought that traces in cache would still working and writing into `$WORSPACE/.metals` is just an additional option. Wdyt of returning traces in cache as a fallback if there is no trace file in `$WORKSPACE/.metals`?
> 
> For me it was easier to check traces there whenever I spot some strange behavior in any project.

Sure, I can add writing into the cache directory as a fallback.

docs/contributors/getting-started.md
ckipp01, line 190:
```suggestion
Since Metals v0.10.10, the log files should be placed in `$WORKSPACE` rather than your machine cache directory.
```

docs/contributors/getting-started.md
```diff
@@ -186,36 +186,38 @@ These logs contain information that may be relevant for regular users.
...
+
+To see trace of incoming/outgoing JSON communication with the text editor or build server, create empty files in `$WORKSPACE/.metals/` directory.
```
ckipp01, line 192:
```suggestion
To see the trace of incoming/outgoing JSON communication with the text editor or build server, create empty files in `$WORKSPACE/.metals/` directory.
```

docs/contributors/getting-started.md
kpodsiad: I joined macOS and Linux cases, but I didn't remove a window one. I think it's better to not delete it.

docs/contributors/getting-started.md
```diff
@@ -186,36 +186,30 @@ These logs contain information that may be relevant for regular users.
 
 ### JSON-RPC trace
 
-To see trace of incoming/outgoing JSON communication with the text editor or
-build server, create empty files in your machine cache directory.
+To see the trace of incoming/outgoing JSON communication with the text editor or build server, create empty files in `$WORKSPACE/.metals/` or your machine cache directory.
```
tgodzik, line 189: Could we mention that machine cache is not recommened, since multiple servers can override it? Also, we no longer mention when the cache is located, so we should probably add a section about it.

metals/src/main/scala/scala/meta/internal/metals/Trace.scala
```diff
@@ -4,45 +4,78 @@ import java.io.PrintWriter
...
-    } else {
-      scribe.info(
-        s"tracing is disabled for protocol $protocolName, to enable tracing of incoming " +
-          s"and outgoing JSON messages create an empty file at $path"
-      )
-      None
-    }
-  }
+    val metalsDir =
+      if (workspace.toNIO.endsWith(".metals")) workspace
```
tgodzik, line 52: Why would the workspace be the `.metals` directory?

docs/contributors/getting-started.md
```diff
@@ -186,36 +186,30 @@ These logs contain information that may be relevant for regular users.
 
 ### JSON-RPC trace
 
-To see trace of incoming/outgoing JSON communication with the text editor or
```
tgodzik, line 189: Could you run prettier on the file? IT should wrap the lines.

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/metals/Trace.scala
kpodsiad: That's because previously as a default I used `PathIO.workingDirectory.resolve(".metals/")` which makes no sense because it isn't a workspace dir. Now I default to the `PathIO.workingDirectory` and problem is gone

tgodzik (approved):

PR #3280, opened 2021-11-10
Support Scala3-NIGHTLY releases

dos65:
This introduces the following changes:
 - every release supports up to 5 of latest NIGHTLY Scala3 versions.
   Should be already valuable as every snapshot release will
   automatically catch new NIGHTLY releases.
 - `+cross/test` also run tests against the latest NIGHTLY release
  that should help with spotting issues earlier.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 on:
   push:
     branches:
-      - main 
+      - main
   pull_request:
 jobs:
   unit:
@@ -12,20 +12,20 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest, macOS-latest, ubuntu-latest]
-        java: ['adopt@1.11']
+        java: ["adopt@1.11"]
         shard: [1, 2]
         include:
           - os: ubuntu-latest
-            java: 'adopt@1.8'
+            java: "adopt@1.8"
             shard: 1
           - os: ubuntu-latest
-            java: 'adopt@1.8'
+            java: "adopt@1.8"
             shard: 2
           - os: ubuntu-latest
-            java: '17'
+            java: "17"
             shard: 1
           - os: ubuntu-latest
-            java: '17'
+            java: "17"
             shard: 2
           - os: ubuntu-latest
             CACHE_PATH: ~/.cache/coursier/v1
@@ -43,12 +43,9 @@ jobs:
         run: |
           bin/test.sh unit/test
         env:
-          TEST_SHARD:
-            ${{ matrix.shard }}
-          GOOGLE_APPLICATION_CREDENTIALS:
-            ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}
-          GOOGLE_APPLICATION_CREDENTIALS_JSON:
-            ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS_JSON }}
+          TEST_SHARD: ${{ matrix.shard }}
+          GOOGLE_APPLICATION_CREDENTIALS: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}
+          GOOGLE_APPLICATION_CREDENTIALS_JSON: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS_JSON }}
         shell: bash
   integration:
     name: ${{ matrix.name }}
@@ -97,6 +94,10 @@ jobs:
             command: sbt +cross/test
             name: Scala cross tests
             os: ubuntu-latest
+          - type: cross-latest-nightly
+            command: sbt cross-test-latest-nightly
+            name: Scala3 latest NIGHTLY cross test
+            os: ubuntu-latest
           - type: scalafix
             command: sbt scalafixCheck docs/docusaurusCreateSite
             name: Scalafix and docs
@@ -114,10 +115,8 @@ jobs:
       - name: ${{ matrix.command }}
         run: ${{ matrix.command }}
         env:
-          GOOGLE_APPLICATION_CREDENTIALS:
-            ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}
-          GOOGLE_APPLICATION_CREDENTIALS_JSON:
-            ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS_JSON }}
+          GOOGLE_APPLICATION_CREDENTIALS: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}
+          GOOGLE_APPLICATION_CREDENTIALS_JSON: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS_JSON }}
       - name: "test download dependencies"
         run: sbt downloadDependencies
         if: matrix.type == 'cross'
--- a/build.sbt
+++ b/build.sbt
@@ -181,10 +181,53 @@ addCommandAlias(
   "+publishLocal; metals/runMain scala.meta.metals.DownloadDependencies "
 )
 
-commands += Command.command("save-expect") { s =>
-  "unit/test:runMain tests.SaveExpect" ::
-    s
+def isNightliesEnabled: Boolean =
+  isCI || sys.env.get("NIGHTLIES").isDefined
+
+def configureMtagsScalaVersionDynamically(
+    state: State,
+    scalaV: String,
+    projectV: Option[String] = None
+): State = {
+  val scalaVersionSettings =
+    List(
+      mtest / scalaVersion := scalaV,
+      mtags / scalaVersion := scalaV,
+      cross / scalaVersion := scalaV
+    )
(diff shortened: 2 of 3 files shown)


tgodzik (commented):
I am a bit worried that this would pop up on the unrelated PRs.

Could we somehow only run it on master? Maybe by checking the current git branch ?

dos65:
@tgodzik Yep, I was thinking about that. 
For the case if there is an issue in NIGHTLY release, my plan was to check the reason: compiler/metals. In case if it's compiler then blacklist this release somewhere in this function.

I think seeing failed tests for NIGHTLY only after merging in master might be more annoying than on PR stage.

Maybe let's try as it is? If it will be more problematic than useful we are always free to remove it from `+cross/test`.

build.sbt
```diff
@@ -255,6 +255,8 @@ lazy val V = new {
   def nonDeprecatedScala3Versions = Seq(nextScala3RC, scala3, "3.0.2")
   def deprecatedScala3Versions = Seq("3.0.1", "3.0.0")
   def scala3Versions = nonDeprecatedScala3Versions ++ deprecatedScala3Versions
+  def nigthtlyScala3Versions =
```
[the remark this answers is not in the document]
dos65: Good idea, thx!

ckipp01:
> I am a bit worried that this would pop up on the unrelated PRs.

>> I think seeing failed tests for NIGHTLY only after merging in master might be more annoying than on PR stage.

Yea I sort of agree with this, it might be hard to spot if it's only happening when it hits main, since most of. the time if there is a failed build on merge I don't go back really and look at what caused it. We could do something similar to our sourcegraph job where if it fails for a pr, then it doesn't block the merge or fail the check, but rather notify us that something was wrong.

>  In case if it's compiler then blacklist this release somewhere in this function.

Also, total aside, but let's try to replace our usage of _blacklist_ with just _deny_ since it is a loaded term.

build.sbt
```diff
@@ -255,6 +255,8 @@ lazy val V = new {
   def nonDeprecatedScala3Versions = Seq(nextScala3RC, scala3, "3.0.2")
   def deprecatedScala3Versions = Seq("3.0.1", "3.0.0")
   def scala3Versions = nonDeprecatedScala3Versions ++ deprecatedScala3Versions
+  lazy val nigthtlyScala3Versions =
```
tgodzik, line 258:
```suggestion
  lazy val nightlyScala3Versions =
```

build.sbt
```diff
@@ -644,7 +648,8 @@ lazy val cross = project
   .settings(
     testSettings,
     sharedSettings,
-    crossScalaVersions := V.nonDeprecatedScalaVersions
+    crossScalaVersions :=
+      V.nonDeprecatedScalaVersions ++ V.nigthtlyScala3Versions.lastOption.toList
```
tgodzik, line 652: Could we only run this on a separate job and remove the check from required for a merge? I will probably need to do it manually later on, but let's maybe have this run separately. This could be done via special env variable maybe?
```suggestion
    crossScalaVersions := {
         if (System.getenv("TEST_NIGHTLY") != null)  V.nightlyScala3Versions.lastOption.toList 
         else V.nonDeprecatedScalaVersions
     }
```
and add that variable to a separate workflow. What do you think?

tgodzik (changes requested):

build.sbt
dos65: Good idea!
Will rework it a bit differently to call just a `++$LATEST_NIGHTLY cross/test` from a separate job

.github/workflows/ci.yml
```diff
@@ -2,7 +2,7 @@ name: CI
 on:
   push:
     branches:
-      - main 
+      - main
```
dos65, line 5: Some non-relevant changes were triggered by formatting in VScode but looks like they are valid.

build.sbt
```diff
@@ -255,6 +295,8 @@ lazy val V = new {
   def nonDeprecatedScala3Versions = Seq(nextScala3RC, scala3, "3.0.2")
   def deprecatedScala3Versions = Seq("3.0.1", "3.0.0")
   def scala3Versions = nonDeprecatedScala3Versions ++ deprecatedScala3Versions
+  lazy val nightlyScala3Versions =
+    Scala3NightlyVersions.nightlyReleasesAfter(nextScala3RC)
```
tgodzik, line 299:
```suggestion
  lazy val nightlyScala3Versions = { if (isCi)
    Scala3NightlyVersions.nightlyReleasesAfter(nextScala3RC)
    else Nil
  }
```

To avoid fetching the versions when starting sbt locally, which would add to load time. I would even add a separate variable so that this is only added when needed in the nightlies cross tests. Otherwise, each job will try to download the versions.

So maybe:
```suggestion
  lazy val nightlyScala3Versions = { if (System.getenv("NIGHTLIES") != null)
    Scala3NightlyVersions.nightlyReleasesAfter(nextScala3RC)
    else Nil
  }
```
and set `NIGHTLIES` for the single cross test and for the releases.

build.sbt
```diff
@@ -181,11 +181,51 @@ addCommandAlias(
...
+      .runTask(cross / Test / test, configured)
+  out
 }
 
+commands ++= Seq(
+  Command.command("save-expect") { s =>
+    "unit/test:runMain tests.SaveExpect" :: s
+  },
+  Command.command("cross-test-latest-nightly") { s =>
+    V.nightlyScala3Versions.lastOption match {
```
tgodzik, line 220: Would we want to test it for all last 5 versions?

build.sbt
```diff
@@ -181,11 +181,51 @@ addCommandAlias(
   "+publishLocal; metals/runMain scala.meta.metals.DownloadDependencies "
 )
 
-commands += Command.command("save-expect") { s =>
-  "unit/test:runMain tests.SaveExpect" ::
-    s
+def configureMtagsScalaVersionDynamically(
```
tgodzik, line 184: Wouldn't this be easier with the env var ?

build.sbt
```diff
@@ -333,7 +375,9 @@ def multiScalaDirectories(root: File, scalaVersion: String) = {
 }
 
 val mtagsSettings = List(
-  crossScalaVersions := V.supportedScalaVersions,
+  crossScalaVersions := {
+    V.supportedScalaVersions ++ V.nightlyScala3Versions
```
tgodzik, line 379: With this, why would we need `configureMtagsScalaVersionDynamically` ?

tgodzik (changes requested):

build.sbt
dos65: This function is anyway will be added in #3281 so I think it's better to use it as command.

build.sbt
dos65: I think having latest should be enough. Will save some minutes per ci run.

build.sbt
dos65: At the beginning I was thinking about additional costs from versions fetching but I was surprised by how fast it works an decided that probably it doesn't worth to prevent this call.

Just checked it more time: it takes less than 500ms.  Tested it with and without internet connection.
Maybe let's leave it as it is?

tgodzik: > it takes less than 500ms.

The problem is that at some point it all accumulates and sbt just loads super slow.
I would still change it to at least only run on CI, it's simple enough if anyone needs to make it work locally.

build.sbt
dos65: This is needed to publish mtags for nightly versions.
So at the release moment of for snapshot supported scala versions will be: all explicitly defined scala versions + up to 5 last nightly

kpodsiad (approved):
LGTM (includes Tomek's suggestions)

build.sbt
dos65: :smiling_face_with_tear: 
I understand your concerns but I think it will be confusing that you need to pass a special env to work with nightlies locally (test/debug/publish local). It also should be documented for contributors. 

Wdyt if I cover this thing by sbt cache? Probably it will be more faster.
The worst case will be that once a day coursier downloads this file(https://repo1.maven.org/maven2/org/scala-lang/scala3-compiler_3/maven-metadata.xml)   that I think should be about 500ms.

tgodzik: > I understand your concerns but I think it will be confusing that you need to pass a special env to work with nightlies locally (test/debug/publish local). It also should be documented for contributors.

To publish a nightly version all you need to do is `++3.0.1-RC1-bin-20210413-f3c1468-NIGHTLY` and `mtags/publishLocal`, so this wouldn't be a problem for releasing locally really. And with these changes when someone does `+publishLocal` they will have 5 more versions to publish each time even if everything was compiled last time.

I would really opt that it's all done in CI only.

dos65: Changed it - now it's needed  to set either `CI` or `NIGHTLIES` env var to enable this logic.

tgodzik (approved):
LGTM!

PR #3281, opened 2021-11-10
Release process: publish `mtags` for the latest Metals release

dos65:
The goal of this PR is to simplify our release process.
Currently, Metals has a list of supported compiler versions and for every new compiler version, we have to make a new release of Metals.

That is a bit problematic because a full release preparation requires more actions than just adding support for a new compiler version: we need to check/do last fixes for all new merged features, prepare release notes, some tests should be reworked while overall it isn't required for having basic functionality. So all these things increase the delay between compiler release and when it will be supported by Metals.
The additional problem is that this mechanic doesn't allow us to support scala3-NIGHTLY releases.

I spit the required changes on several PR's:

 - First one - [ScalaVersion - discover new mtags version by coursier](https://github.com/scalameta/metals/pull/3278). Having it Metals starts to check if there are `mtags` artifacts for the required Scala version instead of rejecting new version that wasn't supported at the release moment.
 
 - This PR is about the release process.
    Now we would be able to release a single mtags artifact for the new compiler version by pushing a tag with a specific format:
    `mtags_v${metalsVersion}_${scalaVersion}`. This will trigger the publishing process.
    
    Also, a added a new workflow:`mtags-auto-release.yml` to do that automatically.
    It might be triggered manually or automatically from CI process that publishes scala-3 artifacts.
    In case if `cross/test` pass that will create a tag that will trigger the publishing process.
    
    After validating this approach I will add documentation about that thing.
    
There is related [PR about NIGHTLY support](https://github.com/scalameta/metals/pull/3280) that fixes cross tests for such versions.

Also, the current set of changes wouldn't work for new Scala 2 versions as they require semanticdb version upgrade but I'm going do follow-up fixes for that later.

### DIFF
--- a/.github/workflows/mtags-auto-release.yml
+++ b/.github/workflows/mtags-auto-release.yml
@@ -1,0 +1,32 @@
+name: "Mtags auto release"
+on:
+  workflow_dispatch:
+    inputs:
+      scala_version:
+        description: "Scala Version"
+        required: true
+      metals_version:
+        description: "Metals Version"
+        required: true
+        default: "v0.10.9"
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.inputs.metals_version }}
+      - uses: olafurpg/setup-scala@v13
+      - uses: coursier/cache-action@v6.3
+      - name: "Test and push tag"
+        run: |
+          sbt test-mtags-dyn ${{ github.event.inputs.scala_version }}
+          if [ $? == 0]; then
+            git commit --allow-empty -m "Release mtags-${{ github.event.inputs.scala_version }} for ${{github.event.inputs.metals_version}}"
+            TAG_NAME="mtags_${{ github.event.inputs.metals_version }}_${{ github.event.inputs.scala_version }}"
+            git tag $TAG_NAME
+            git push origin $TAG_NAME 
+          fi
+        env:
+          GIT_USER: scalameta@scalameta.org
--- a/build.sbt
+++ b/build.sbt
@@ -196,6 +196,51 @@ commands ++= Seq(
         println("No nightly versions was found. Skipping cross/test")
         s
     }
+  },
+  Command.single("test-mtags-dyn") { (s, scalaV) =>
+    crossTestDyn(s, scalaV)
+  },
+  Command.args("publish-mtags-dyn", "<scala-version> <version>") {
+    (state, args) =>
+      val configured =
+        configureMtagsScalaVersionDynamically(state, args(0), Some(args(1)))
+      val (out, _) =
+        Project
+          .extract(configured)
+          .runTask(mtags / PgpKeys.publishSigned, configured)
+      out
+  },
+  /**
+   * In case if this command was called from tag that matches
+   * "mtags_v$metalsVersion_$scalaVersion" then publish only mtags with these parameters
+   * Otherwise, call std `ci-release` and publish docs
+   */
+  Command.command("gh-actions-release") { s =>
+    val isTag = Option(System.getenv("GITHUB_REF"))
+      .exists(_.startsWith("refs/tags"))
+
+    val metalsTagRegex = "^v\\d+\\.\\d+\\.\\d+.*".r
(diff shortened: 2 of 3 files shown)


.github/workflows/mtags-auto-release.yml
```diff
@@ -0,0 +1,31 @@
+name: "Mtags auto release"
+on:
+  workflow_dispatch:
+    inputs:
+      scala_version:
```
kpodsiad, line 5: So, according to the [docs](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#manual-events) this has to be dispatched manually, am I right?

.github/workflows/mtags-auto-release.yml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: olafurpg/setup-scala@v13
+      - uses: coursier/cache-action@v6.3
+      - name: Test Dyn
```
kpodsiad, line 21: Since workflow is oriented towards publishing, maybe
```suggestion
      - name: Publish Mtags
```

kpodsiad (approved):

.github/workflows/mtags-auto-release.yml
tgodzik: Maybe in that case it is not needed? I think the workflow for releasing manually would be:
1. Checkout latest Metals tag
2. Add new tag with the new Scala version
3. Push it or push a tag with some fixes.

As for releasing for nightlies, we could just run it in the release job?

build.sbt
```diff
@@ -181,11 +181,71 @@ addCommandAlias(
...
+   */
+  Command.command("gh-actions-release") { s =>
+    val isTag = Option(System.getenv("GITHUB_REF"))
+      .exists(_.startsWith("refs/tags"))
+
+    val mtagsTagRegex = "^mtags_v(\\d+\\.\\d+\\.\\d+)_(.+)".r
+
+    Option(System.getenv("GITHUB_REF_NAME")) match {
+      case Some(mtagsTagRegex(metalsVersion, scalaV)) if isTag =>
+        s"publish-mtags-dyn $scalaV $metalsVersion" :: s
```
tgodzik, line 242: Could we add some validation on the other clause too? So that we don't end up with broken release in case of a mistake?

.github/workflows/mtags-auto-release.yml
dos65: > As for releasing for nightlies, we could just run it in the release job?

It will be required to merge in master to trigger it so it might be unclear when nightly support appears.

Actually, I wanted to use it for releasing other versions too. Just trigger this job, if there are problems, you will get a report, if everything is well - no actions required :steam_locomotive::railway_car::railway_car:

build.sbt
dos65: Do you mean to add validation for `"ci-release" :: "docs/docusaurusPublishGhpages" :: s` ?
I've checked CIRelease plugin - there is no requirements. It checks internally if it's a tag or snapshot.

I see that the worst thing that might happen is that if there will be a typo in tag name for mtags release it will trigger usual release with a creepy version name.

.github/workflows/mtags-auto-release.yml
```diff
@@ -0,0 +1,31 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      scala_version:
+        description: "Scala Version"
+        required: true
+      metals_version:
+        description: "Metals Version"
+        required: true
+        default: "v0.10.9"
```
ckipp01, line 11: 😅 make sure to add bumping this to the next version during an actual full release to the checklist

ckipp01 (approved):
This whole thing is an interesting approach. Good job cooking this all up @dos65 🧑‍🍳 .

.github/workflows/mtags-auto-release.yml
dos65: Changed it a bit differently

build.sbt
dos65: added!

tgodzik (approved):
LGTM!

PR #3284, opened 2021-11-12
docs: update commands that now use `TextDocumentPositionParams`

ckipp01:
This was a small change that clients didn't account for in
https://github.com/scalameta/metals/commit/84dbb851ad753c6ac79a410a1b8aa272853db1f1
where the custom `GoToSuperMethodParams` were removed and just replaced
with `TextDocumentPositionParams`. This pr just updates the docs to
ensure that it's clear that the commands use this, not the old format.
Changes for this will need to happen in the client repos.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -295,18 +295,8 @@ object ServerCommands {
        |If symbol is a reference of a method it will jump to a definition.
        |If symbol under cursor is invalid or does not override anything then command is ignored.
        |
-       |Note: document in json argument must be absolute path.
        |""".stripMargin,
-    """|
-       |Object with `document` and `position`
-       |
-       |Example:
-       |```json
-       |{
-       |  document: "file:///home/dev/foo/Bar.scala",
-       |  position: {line: 5, character: 12}
-       |}
-       |```
+    """|This command should be sent in with the LSP [`TextDocumentPositionParams`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams)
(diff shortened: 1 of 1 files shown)


dos65:
@ckipp01 maybe it would be easier just to return these commands back to old parameters?
It seems that this change wasn't  applied on any client.

dos65:
Oops, I missed that you already did fixes for coc/nvim metals :smile:

ckipp01:
> @ckipp01 maybe it would be easier just to return these commands back to old parameters? It seems that this change wasn't applied on any client.

I think it's actually preferable to keep what we have now. Before we had a custom param that needed to be put together, whereas now we can just use an already existing part of the spec, https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams. I know for `nvim-metals`, it makes it much easier since they have methods built-in to just put this together for you. I see that the same exists in the VS Code extension with `getTextDocumentPositionParams`. I don't feel strongly about it, but I think if we ever have the choice to just use something that exists and is defined in the LSP spec, it's always preferable over our own alternative, and seeing that the feature still works with `TextDocumentPositionParams` I think it's preferable. wdyt?

tgodzik (approved):
Och damn! I totally missed it, I think I thought it's exactly the same :fearful: 

Let's try to update it in all the clients. I will try and see where this is needed.

tgodzik:
Fixing it in VS Code here: https://github.com/scalameta/metals-vscode/pull/757

tgodzik:
Sublime doesn't use it https://github.com/scalameta/metals-sublime

tgodzik:
And emacs should be fixed by https://github.com/emacs-lsp/lsp-metals/pull/61

Thanks again @ckipp01 for noticing it!

PR #3292, opened 2021-11-15
Update ammonite to 2.4.1

tgodzik:


tests/unit/src/main/scala/tests/BaseAmmoniteSuite.scala
```diff
@@ -355,7 +355,7 @@ abstract class BaseAmmoniteSuite(scalaVersion: String)
 
   test("simple errored script") {
     val expectedDiagnostics =
-      """main.sc:15:25: error: not found: type Fooz
+      """errored.sc:15:25: error: not found: type Fooz
```
tgodzik, line 358: It seems using `main` causes an additional warning because there is already `main` improted by ammonite. Not sure if it's really a bug though.

tgodzik:
I will go ahead and merge this.

ckipp01:
Ach, sorry I thought I had already approved this, but yes, glad we can finally update this!

PR #3293, opened 2021-11-16
Set rangeHoverProvider under experimental capabilities

ayoub-benali:
A workaround for https://github.com/microsoft/language-server-protocol/issues/377 until it is resolved
this would allow LSP clients to explicitly know that they can send a range with `textDocument/hover` instead of a position

If someone has a better name for the field I am happy to change it :)

@ckipp01 should this be added to `new-editor.md` ?

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
@@ -1,0 +1,11 @@
+package scala.meta.internal.metals
+
+/**
+ * Language Server Protocol extensions that are declared as "server
+ * capabilities" in the initialize response.
+ */
+
+object MetalsExperimental {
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):
Cool, seems to do the trick 👍🏼 
```
[Trace - 03:08:12 PM] Sending response 'initialize - (1)'. Processing request took 1445ms
Result: {
  "capabilities": {
    ...
    "experimental": {
      "rangeHoverProvider": true
    }
  },
  "serverInfo": {
    "name": "Metals",
    "version": "0.10.10-SNAPSHOT"
  }
}
```

As for the name, I think this is fine. Thanks for tackling this!

ckipp01:
> @ckipp01 should this be added to new-editor.md ?

No I think we're good. Under hover we already mention this:

https://scalameta.org/metals/docs/integrations/new-editor/#textdocumenthover

tgodzik (approved):

PR #3297, opened 2021-11-17
Add extract value code action

tgodzik:
This connected to https://github.com/scalameta/metals-feature-requests/issues/186

There is no sensible way yet to ask for additional input from the users (choose the exact expression or name)

Currently we will extract the last function parameter at the cursor.


tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -200,21 +200,21 @@ class CompletionMatchSuite extends BaseCompletionSuite {
...
-      |import java.nio.file.AccessMode.EXECUTE
-      |
-      |object Main {
-      |  (null: AccessMode) match {
-      |\tcase READ => $0
-      |\tcase WRITE =>
-      |\tcase EXECUTE =>
-      |}
-      |}""".stripMargin,
+    s"""
```
tgodzik, line 203: The change here makes sure that `\t` are translated to tabs. It doesn't work without `s` interpolator

mtags/src/main/scala/scala/meta/internal/metals/TextEdits.scala
```diff
@@ -22,10 +22,7 @@ object TextEdits {
       val out = new java.lang.StringBuilder()
       positions.foreach { case (edit, pos) =>
         out.append(text, curr, pos.start)
-        edit.getNewText().foreach {
```
tgodzik, line 25: This was causing some tests to look weird and almost impossible to test fails with tabs in them.

tgodzik:
Ok, I think this is ready to be reviewed. I looked at the current way this worked in Intellij and it seems to cover most of the scenarios I could find.

extract-value

It would be awesome to be able to specify the name, but I am still thinking of how to achieve that.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -0,0 +1,311 @@
...
+  private def indentationLength(text: String, lineStart: Int): Int = {
+    var i = lineStart
+    while (i < text.length() && (text(i) == '\t' || text(i) == ' ')) {
+      i += 1
+    }
+    i - lineStart
+  }
+
+  /**
+   * Check if it's possible to use braceless syntax and whther
```
kpodsiad, line 210:
```suggestion
   * Check if it's possible to use braceless syntax and whether
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -0,0 +1,311 @@
...
+   * it's the preffered style in the file.
```
kpodsiad, line 211:
```suggestion
   * it's the preferred style in the file.
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -0,0 +1,311 @@
...
+   */
+  private def canUseBracelessSyntax(stat: Tree, source: String) = {
+
+    def allowBracelessSyntax(tree: Tree) = tree.origin match {
+      case p: Parsed => p.dialect.allowSignificantIndentation
+      case _ => false
+    }
+
+    // Let's try to use the style of any existing parent.
+    def existsBracelessParent(tree: Tree): Boolean = {
```
kpodsiad, line 221: `@tailrec`

kpodsiad (approved):
Looks awesome!

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -779,6 +779,14 @@ object MetalsEnrichments
 
   }
 
+  implicit class XtensionChar(ch: Char) {
+    def stringRepeat(n: Int): String = {
+      if (n > 0)
+        ("%0" + n + "d").format(0).replace('0', ch)
```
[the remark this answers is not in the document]
tgodzik: I seemed to have some issues here, but your suggestion seems to work :thinking:

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -0,0 +1,311 @@
...
+      apply <- applyOpt
+      stats <- lastEnclosingStatsList(apply)
+      argument <- apply.args.find { arg => arg.pos.encloses(range) }
+      // avoid extracting lambdas (this needs actual type information)
+      if isNotLambda(argument)
+      stat <- stats.find(stat => stat.pos.encloses(apply.pos))
+      name = createNewName(stats)
+      source <- buffers.get(path)
+      blank =
+        if (source(stat.pos.start - stat.pos.startColumn) == '\t') '\t' else ' '
```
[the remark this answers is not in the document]
tgodzik: It's still available within the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#documentFormattingOptions), so I would try to accomodate it.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -0,0 +1,311 @@
...
+      indent = blank.stringRepeat(indentationLength(source, lineStart))
+      keyword = if (stat.isInstanceOf[Enumerator]) "" else "val "
+      // we will inset `val newValue = ???` before the existing statement containing apply
+      valueText = s"${indent}$keyword$name = ${argument.toString()}"
+      valueTextWithBraces = withBraces(stat, source, valueText, blank)
+      // we need to add additional () in case of  `apply{}`
+      replacementText =
+        if (argument.is[Term.Block] && !applyHasParens(apply)) s"($name)"
+        else name
+      replacedArgument = new l.TextEdit(argument.pos.toLSP, replacementText)
```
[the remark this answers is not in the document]
tgodzik: It looked nicer, but sure I can do that :P

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -0,0 +1,311 @@
...
+      val startBlockPos = equalsPos.toLSP
+      startBlockPos.setStart(startBlockPos.getEnd())
+      startBlockPos.setEnd(statStart.getStart())
+
+      val noIndentation = equalsPos.startLine == defn.body.pos.startLine
+
+      // Scala 3 optional braces
+      if (canUseBracelessSyntax(stat, source)) {
+        // we need to create a new indented region
+        if (noIndentation) {
```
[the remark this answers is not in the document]
tgodzik: I was thinking about different ways to make it clearer, but wasn't able to do  better. I don't think your suggestion will work unfortunately, you end up with:
```
(canUseBracelessSyntax(stat, source), noIndentation) match {
  case (true, false) =>
  case (true, true) =>
```
It's even less easy to reason about it this way.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -0,0 +1,311 @@
...
+      range.getStart(),
+      appl => appl.args.exists(_.pos.encloses(range))
+    )
+
+    val textEdits = for {
+      apply <- applyOpt
+      stats <- lastEnclosingStatsList(apply)
+      argument <- apply.args.find { arg => arg.pos.encloses(range) }
+      // avoid extracting lambdas (this needs actual type information)
+      if isNotLambda(argument)
```
ckipp01, line 48: Ahhh, haha this was actually the first thing I tried when testing it.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -0,0 +1,311 @@
...
+) extends CodeAction {
+  override def kind: String = l.CodeActionKind.RefactorExtract
+
+  override def contribute(params: CodeActionParams, token: CancelToken)(implicit
+      ec: ExecutionContext
+  ): Future[Seq[l.CodeAction]] = Future {
+    val path = params.getTextDocument().getUri().toAbsolutePath
+    val range = params.getRange()
+
+    val applyOpt = trees.findLastEnclosingAt[Term.Apply](
```
ckipp01, line 37: I'm not sure if it's right here or not that ends up being the issue, but I've found that at times it ends up not extracting what I'd expect. For example I tried this on various places in the Metals code base.

```scala
  private def newSymbolIndex(): OnDemandSymbolIndex = {
    OnDemandSymbolIndex.empty(
      onError = {
        case e @ (_: ParseException | _: TokenizeException) =>
          scribe.error(e.toString)
        case e: InvalidJarException =>
          scribe.warn(s"invalid jar: ${e.path}")
        case _: NoSuchFileException =>
        // only comes for badly configured jar with `/Users` path added.
        case NonFatal(e) =>
          scribe.error("unexpected error during source scanning", e)
      },
      toIndexSource = path => {
        if (path.is<<>>AmmoniteScript)
          for {
            target <- buildTargets.sourceBuildTargets(path).headOption
            toIndex <- ammonite.generatedScalaPath(target, path)
          } yield toIndex
        else
          None
      }
    )
  }
```

I sort of expected this to extract `path.isAmmoniteScript`, but instead it extracted:

```scala
    val newValue = toIndexSource = path => {
        if (path.isAmmoniteScript)
          for {
            target <- buildTargets.sourceBuildTargets(path).headOption
            toIndex <- ammonite.generatedScalaPath(target, path)
          } yield toIndex
        else
          None
      }
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -0,0 +1,311 @@
...
+      if isNotLambda(argument)
+      stat <- stats.find(stat => stat.pos.encloses(apply.pos))
+      name = createNewName(stats)
+      source <- buffers.get(path)
+      blank =
+        if (source(stat.pos.start - stat.pos.startColumn) == '\t') '\t' else ' '
+      lineStart = stat.pos.start - stat.pos.startColumn
+      indent = blank.stringRepeat(indentationLength(source, lineStart))
+      keyword = if (stat.isInstanceOf[Enumerator]) "" else "val "
+      // we will inset `val newValue = ???` before the existing statement containing apply
```
ckipp01, line 57:
```suggestion
      // we will insert `val newValue = ???` before the existing statement containing apply
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -0,0 +1,311 @@
...
+
+          Seq(
+            new l.TextEdit(
+              statStart,
+              s"""|$indentStat${valueString.trim()}
+                  |$innerIndentation""".stripMargin
+            )
+          )
+        }
+        // Scala 2
```
ckipp01, line 155: Just to make sure I'm tracking this is actually 
```suggestion
        // Scala 2 and Scala 3 non signification whitespace syntax
```
right?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -0,0 +1,311 @@
...
+      i += 1
+    }
+    i - lineStart
+  }
+
+  /**
+   * Check if it's possible to use braceless syntax and whther
+   * it's the preffered style in the file.
+   */
+  private def canUseBracelessSyntax(stat: Tree, source: String) = {
```
ckipp01, line 213: ☹️ there is a decent amount of code in here just trying to get the braceless syntax right. I honestly wish we didn't have to do this and just inserted code with braces if needed, and then let formatting figure it out.

ckipp01 (commented):
This is pretty awesome @tgodzik, good job on this! Just a couple comments/quetions.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
tgodzik: I need to do it via PresentationCompiler, doing some related work with inferring methods.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
tgodzik: I forgot about named arguments! :fearful: 

The current approach will try to extract a parameter, it doesn't work for if conditions or anything else. We could do that for sure, but maybe we could refactor it later?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
tgodzik: I know.. this actually the worst part to figure out and I spent way too much time on it :S

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
ckipp01: For sure!

dos65 (approved):
LGTM!

I left some minor suggestions but overall even with known limitations I think it's already good to merge and do adjustments later after collecting feedback from users.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
tgodzik: I fixed the issue with named parameters and added a test for it (modified existing one to use named params).

I have an idea of how to add support for Term.If etc, but what do you think we would want to extract also?

I think this can be:
- condition in while/if/doWhile
- interpolate
- return
- throw
- tuple
- matched expression
- new

Anything else comes to mind ?

tgodzik:
Thanks for the reviews! A lot of good suggestions! I applied most of them, beside the ones I explicitely commented on.

Do let me know if there is something more I should change in this PR

metals/src/main/scala/scala/meta/internal/metals/CodeActionProvider.scala
```diff
@@ -42,7 +42,8 @@ final class CodeActionProvider(
     ),
     new InsertInferredType(trees, compilers),
     new PatternMatchRefactor(trees),
-    new RewriteBracesParensCodeAction(trees)
+    new RewriteBracesParensCodeAction(trees),
```
kubukoz, line 45: y'all should just enable trailing commas in scalafmt :)

tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
```diff
@@ -118,6 +119,111 @@ class Scala3CodeActionLspSuite
        |""".stripMargin
   )
 
+  check(
+    "single-def",
+    """|object Main {
+       |  def method2(i: Int) = ???
+       |  def main = {
+       |    def inner(i : Int) = method2(i + 23 + <<123>>)
```
kubukoz, line 127: tbh in all these cases I'd expect 123 to be extracted, especially if it's fully highlighted. It's the last argument to `+`, isn't it?

tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala
```diff
@@ -0,0 +1,159 @@
...
+  check(
+    "for-comprehension",
+    """|object Main {
+       |  def method2(i: Int) = ???
+       |  
+       |  def main() = {
+       |    val opt = Option(1)
+       |    for {
+       |       i <- opt
+       |       res = method2(i + 23 + <<123>>)
```
kubukoz, line 39: How would it behave if this was `res <- method2(i + 23 + <<123>>)` and it was the first line in a `for`?

kubukoz:
> the last function parameter at the cursor

I think currently this doesn't count infix or symbol-only functions, is that correct?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
ckipp01: 🤔 I'm not really sure just off the top of my head. This is probably one of those things that I'd want to use and play around with, try to extract something, not be able to and then know that I want it 😆 I also think it's fine to start small and add as we go.

ckipp01 (approved):
Excited to use this. Nice job @tgodzik!

tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
tgodzik: I didn't include `Term.ApplyInfix`, since I was trying to guess best what the user might want to extract. But maybe we should do it similar to Intellij and just suggest a couple of different options for each apply argument? :thinking:

tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala
tgodzik: It will bug out! Good catch :sweat_smile:

tgodzik: Aaaand it's fixed!

tgodzik:
> I think currently this doesn't count infix or symbol-only functions, is that correct?

That's true for infix, not sure what you mean by `symbol-only functions` - I plan to work on it further in some separate PRs

kubukoz:
Yeah I just meant infix, thought there was some special handling of operators (didn't look at the impl code) :)

Issue #3300, opened 2021-11-18
Type Annotation on Code Selection doesn't work with Ammonite build server


ayoub-benali:
**Describe the bug**
`textDocument/hover` with position works fine with standalone ammonite scripts but when sending `range` metals sends an empty reponse.

**To Reproduce**
For the following `test.sc` file 
```
// ammonite 2.4.1

List(1, 2, 3).map(_.toFloat)

```
This is the LSP log when just a position is sent when hovering `toFloat`
```
:: --> metals textDocument/hover(37): {'position': {'line': 2, 'character': 24}, 'textDocument': {'uri': 'file:///home/ayoub/workspace/test.sc'}}
:: <<< metals 37: {'range': {'start': {'line': 2, 'character': 18}, 'end': {'line': 2, 'character': 27}}, 'contents': {'kind': 'markdown', 'value': '```scala\ndef toFloat: Float\n```'}}
```
And this is the LSP log when sending a range by selecting `toFloat`
```
:: --> metals textDocument/hover(40): {'textDocument': {'uri': 'file:///home/ayoub/workspace/test.sc'}, 'range': {'start': {'line': 2, 'character': 20}, 'end': {'line': 2, 'character': 27}}}
:: <<< metals 40: None
```

**Expected behavior**
`textDocument/hover` to work with range just like with position. 

**Installation:**

- Operating system: Linux
- Editor: Sublime
- Metals version: 0.10.9+46-30cc89e6-SNAPSHOT

**Additional context**
So far this behavior seems to happen only when using Ammonite as a build server. This isn't happening with bloop. 
I guess sbt remains to be tested.

**Search terms**
Ammonite, Type Annotation on Code Selection

PR #3303, opened 2021-11-19
Fix expression for selection range in Ammonite

tgodzik:
Ammonite generates additional file which we use in the presentation compiler for anything that requires types, so we neeed to adjust all request to actually point to a position in that generated file.

Previously, we would not adjust ranges when invoking hover and only positions. Now, we adjust both.

Also fixed the fallback version in case of Scala 3 for ammonite to be the last Ammonite suppored Scala 2.13 version instead of hte last Scala 2.13 version.

Fixes https://github.com/scalameta/metals/issues/3300


ckipp01 (approved):
Looks good, LGTM!

tgodzik:
FYI I haven't a test since the Ammonite hover test is currently set as flaky.

Issue #2188, opened 2020-11-03
[Task] - Migrate all BSP connection related methods to `BspConnector`


ckipp01:
As a follow-up to the work done in https://github.com/scalameta/metals/pull/2154, there are a few methods that are used throughout `MetalsLanguageServer` that should really be moved out. We started to identify some of these issues in https://github.com/scalameta/metals/issues/672, but now that we have a `BspConnector` class, it's the perfect place in my opinion for some of these methods:

- `slowConnectToBuildServer()`
- `slowConnectToBloopServer()`
- `quickConnectToBuildServer()`
- `autoConnectToBuildServer()`
- `disconnectOldBuildServer()`
- `connectToNewBuildServer()`

Between these, it's a couple hundred lines of code that could help get some of the code out of `MetalsLanguageServer`. I initially thought it'd be very simple to just move them into `BspConnector`, but pretty quickly I hit on some issues mainly due to the amount of `var`s that are in `MetalsLanguageServer` that we are relying on and mutating in these methods. So I want to put a bit more thought into it. Just to name a couple:

- `var bspSession: Option[BspSession] = None`
- `private var lastImportedBuilds = List.empty[ImportedBuild]`

So the point of this ticket is mainly for myself to outline what I'd like to do, and to have it on my radar explicitly. So my goals for this are:

- Get rid of some of the code that doesn't need to live next to the LSP request/response methods (already identified in #672)
- Clean up some of the vars like the two listed above
- Ensure that the `BspConnector` contains all the logic for helping facilitate a BSP connection

PR #3304, opened 2021-11-21
Simplify usage of sbt server when no .bsp/sbt.json exists

ckipp01:
This relates to the conversation we had in
https://github.com/scalameta/metals/discussions/3255 and essentially
gets rid of the need for a user to remember they need to run
`metals.generate-bsp-config` if no `.bsp/sbt.json` exists. Now, instead
a user is able to use the `metals.bsp-switch` command and it will detect
if the build tool can be a build server even if there is no
`.bsp/<entry>` that exists, shows it to them during the user choice, and
then if selected generates the entry and then connects to it.

The old behavior of the `metals.generate-bsp-config` is still the same
and can be used _if_ you want a fresh `.bsp/<entry>`.

_notice sbt is given as a choice even though no `.bsp/sbt.json` exists`_.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
@@ -18,6 +18,7 @@ import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.StatusBar
 import scala.meta.internal.metals.Tables
 import scala.meta.internal.metals.UserConfiguration
+import scala.meta.internal.semver.SemVer
 import scala.meta.io.AbsolutePath
 
 import ch.epfl.scala.bsp4j.BspConnectionDetails
@@ -60,7 +61,7 @@ class BspConnector(
 
   /**
    * Handles the connection to the build server. This assumes that all
-   * inforamtion that it needs is already in place by either having a
+   * information that it needs is already in place by either having a
    * workspace that can work with Bloop or a workspace that already has a bsp
    * entry. In the case that a user is switching build servers the generation
    * of the bsp entry has already happened at this point.
@@ -89,14 +90,34 @@ class BspConnector(
         case ResolvedBspOne(details) =>
           bspServers.newServer(workspace, details).map(Some(_))
         case ResolvedMultiple(_, availableServers) =>
+          val distinctServers = availableServers
+            .groupBy(_.getName())
+            .mapValues {
+              case singleVersion :: Nil => singleVersion
+              case multipleVersions =>
+                multipleVersions.reduceLeft[BspConnectionDetails] {
+                  case (a, b) =>
+                    if (
+                      SemVer.Version.fromString(a.getVersion()) > SemVer.Version
+                        .fromString(b.getVersion())
+                    ) a
+                    else b
+                }
+            }
+
           val query =
-            Messages.SelectBspServer.request(availableServers, None)
+            Messages.BspSwitch.chooseServerRequest(
+              distinctServers.keySet.toList,
+              None
+            )
           for {
             Some(item) <- client
               .showMessageRequest(query.params)
               .asScala
               .map(item =>
-                Option(item).flatMap(item => query.details.get(item.getTitle))
+                Option(item).map(item =>
+                  distinctServers(query.mapping(item.getTitle))
+                )
               )
             conn <- bspServers.newServer(workspace, item)
           } yield Some(conn)
@@ -152,7 +173,7 @@ class BspConnector(
       possibleServers: List[String],
       currentSelectedServer: Option[String]
   ): Future[Option[String]] = {
-    val params = Messages.ChooseBspServer.request(
+    val params = Messages.BspSwitch.chooseServerRequest(
       possibleServers,
       currentSelectedServer
     )
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
@@ -202,7 +202,7 @@ class BspConnector(
     // These are buildTools in the workspace that can serve as a build servers
     // and don't already have a .bsp entry
     val possibleServers: Map[String, Either[
-      BuildTool with BuildServerProvider,
+      BuildServerProvider,
       BspConnectionDetails
     ]] = buildTools
       .loadSupported()
@@ -212,17 +212,14 @@ class BspConnector(
               .exists(details =>
                 details.getName() == buildTool.executableName
               ) =>
-          buildTool
-      }
-      .map { possible =>
-        possible.executableName -> Left(possible)
+          buildTool.executableName -> Left(buildTool)
       }
       .toMap
 
     // These are build servers that already have a .bsp entry plus bloop if
     // it's an option.
     val availableServers: Map[String, Either[
-      BuildTool with BuildServerProvider,
+      BuildServerProvider,
       BspConnectionDetails
     ]] = {
       if (bloopPresent || buildTools.loadSupported().nonEmpty)
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
@@ -2,7 +2,6 @@ package scala.meta.internal.bsp
 
 import java.nio.file.Files
 
-import scala.collection.mutable
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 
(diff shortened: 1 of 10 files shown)


metals/src/main/scala/scala/meta/internal/builds/BuildServerProvider.scala
```diff
@@ -18,7 +17,6 @@ trait BuildServerProvider { this: BuildTool =>
    */
   def generateBspConfig(
       workspace: AbsolutePath,
-      languageClient: MetalsLanguageClient,
```
ckipp01, line 21: This was just cleanup since it wasn't being used in here.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -370,6 +370,38 @@ object Messages {
     }
   }
 
+  object ChooseBspServer {
+    case class Request(
+        params: ShowMessageRequestParams,
+        mapping: Map[String, String]
+    )
+
+    def request(
```
ckipp01, line 379: So this is very similiar to the request in `SelectBspServer` down below. The one down below however is only dealing with connection details, and it accounts for things like,  two different entries for the same build server that have the same name, but different version. I honestly don't think this would ever happen, but also didn't want to remove it since it is used in another place (only in the scenario when you open in a non Bloop workspace for the first time and 2 bsp entries are found). Since in this case we aren't guaranteed to have the connection details because the entry doesn't yet exist, we don't do that extra check. Do you think this is ok?

tgodzik: I don't think keeping the logic below is needed, I would instead warn if there are multiple servers with the same version and just suggest the newest. As you said, this scenario has never happened as each tool always replaces it's config and it would require some manual shenanigans for this to actually pop up.

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -119,84 +136,183 @@ class BspConnector(
...
-    val bloopPresent = buildTools.isBloop
 
-    val availableServers = {
-      val found = bspServers.findAvailableServers()
+    val foundServers = bspServers.findAvailableServers()
+    val bloopPresent: Boolean = buildTools.isBloop
+
+    // These are buildTools in the workspace that can serve as a build servers
+    // and don't already have a .bsp entry
+    val possibleServers: Map[String, Either[
```
tgodzik, line 183: Maybe instead of Either we could have a dedicated class:

```scala
case class BuildServerDetails private(connectionDetails: Option[BspConnectionDetails], bspBuildTool: Option[BuildTool with BuildServerProvider]){

  def this(connectionDetails: BspConnectionDetails) = this(Some(connectionDetails), None)
  def this(bspBuildTool: BuildTool with BuildServerProvider]) = this(None, Some(bspBuildTool))
  def displayName = connectionDetails.map(_.getName()).orElse(bspBuildTool.map(_.executableName))
}

```
This is just a suggestion, not sure if it will improve things.

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -119,84 +136,183 @@ class BspConnector(
...
+      BuildTool with BuildServerProvider,
```
tgodzik, line 184: Also, can't we just do 
```suggestion
      BuildServerProvider,
```
?

tgodzik (changes requested):
Thanks for working on this! It was way too difficult to change the servers previously.

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -119,84 +136,183 @@ class BspConnector(
...
+                )
+                .map(status => handleGenerationStatus(buildTool, status))
+            case Right(connectionDetails)
+                if connectionDetails.getName == BloopServers.name && currentSelectedServer
+                  .contains(
+                    BspConnector.BLOOP_SELECTED
+                  ) =>
+              Future.successful(false)
+            case Right(details) if details.getName == BloopServers.name =>
+              tables.buildServers.chooseServer(BspConnector.BLOOP_SELECTED)
```
dos65, line 274: Just a question - why Bloop is handled specifically?
Why it doesn't fall in case where 
```scala
case *  if currentSelected.contains(name) => 
   tables.buildServers.chooseServer(details.getName)
   Future.successful(true)
case * =>  
  Future.successful(false)   
```

Btw. I agree with @tgodzik  that probably having a specific type might be a case here.
Looking at this function I think that having smth like bellow would reduce the amount of branches that should be handled.
```scala
case class BspChoice(serverName: String, switch: () => Future[Unit])
object BspChoice {
   def bloop: BspChoice = ....
   def fromBspDetails(details: BspDetails): BspChoice = ...
   def fromBuildTool(buildTool: BuildTool): BspChoise = ...
}
```

ckipp01: > Just a question - why Bloop is handled specifically?

I think part of this is legacy due to the fact that we don't actually use BSP discovery with Bloop, but rather special case it through and through. In this specific situation (which is just carried over from how it was before) we do an extra check if it's bloop to see if there is already a `.bloop/` dir or not. Since we aren't creating a BSP entry for it, we instead have to call `createBloopAndConnect()`, which is actually `slowConnectToBuildServer()` whereas in the other cases we are using `quickConnectToBuildServer()`.

Part of me wants to think that this could be cleaned up some, but if we are going to attempt that I think we should fully tackle https://github.com/scalameta/metals/issues/2188. There is a bunch of connection stuff still in `MetalsLanguageServer`, but it's really intertwined 😅, but it would be a nice refactor/cleanup for the future.

As for the `BspChoice` type, I'll give that a try!

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
ckipp01: I thought so too, but then `runUnconditionally` compalins that it has a `BuildServerProvider` and wants a `BuildTool`.

tgodzik: Hmm... so maybe we can just do `trait BuildServerProvider extends BuildTool { ` ? Any reason for self type annotation actually?

ckipp01: Ah true, I didn't think of that. When I first made it I only included the self type since I modeled it after the `BloopInstallProvider` which also has a self type. So I just followed suit since they are used pretty much exactly the same.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
ckipp01: Sounds good, I got rid of this and added some logic that if there are two of the same name the newest is chose and there is a warning logged.

ckipp01:
So I played around with having a `BspChoice` case class instead of the `Map[String, Either[BuildServerProvider, BspConnectionDetails]]` but it didn't really clean it up, which basically lead to me just having a wrapper around that which didn't serve a purpose. I really liked your idea @dos65 with the `switch: () => Future[Unit]`, but it won't work without a larger refactor since we do special case Bloop like I explained in your comment and also need the boolean value that's returned and then based off that do some more stuff in `MetalsLanguageServer`.  I do think it'd be nice to have another round at #2188, but I'd like to get this done, get mill BSP in, and then maybe revisit the whole BSP connection thing. I do think it would be a large refactor. Let me know what you think!

ckipp01:
Ahh I now realize the SelectBspServerSuite is broke and isn't fully testing what I want. I need to do a bit more work there.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -370,82 +369,52 @@ object Messages {
...
-        } else {
-          val versionConflicts = candidates.count { c =>
-            c.getName == candidate.getName &&
-            c.getVersion == candidate.getVersion
-          }
-          if (versionConflicts < 2) {
-            s"${candidate.getName} v${candidate.getVersion}"
+      val mapping = mutable.Map.empty[String, String]
+      val messageActionItems =
+        possibleBuildServers.map { bs =>
```
tgodzik, line 387:
```suggestion
        possibleBuildServers.map { buildServer =>
```
Saying `bs` in my  head sounds a bit weird :sweat_smile:

tgodzik (approved):
Two very minor comments, but otherwise LGTM!

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
ckipp01: haha yea of course, switched.

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -119,84 +157,183 @@ class BspConnector(
...
+        case buildTool: BuildServerProvider
+            if !foundServers
+              .exists(details =>
+                details.getName() == buildTool.executableName
+              ) =>
+          buildTool
+      }
+      .map { possible =>
+        possible.executableName -> Left(possible)
+      }
```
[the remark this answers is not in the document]
ckipp01: 👍🏼 fixed

adpi2:
Thanks a lot @ckipp01 for improving this!

PR #3308, opened 2021-11-22
Better support for mill BSP

### LINKED ISSUE #3307: Integrated support for Mill BSP Server
[BSP is now a first class citizen in Mill](https://github.com/lihaoyi/mill/pull/969). It'd be great if Metals had the ability to generate the BSP config necessary for discovery and then auto-connect.

I'll also be keeping track of stuff that I find which will be important for Mill to implement on their side to ensure a good experience for Metals users.

- [x] https://github.com/lihaoyi/mill/issues/986
- [x] https://github.com/lihaoyi/mill/issues/987
- [x] https://github.com/lihaoyi/mill/issues/988
- [x] https://github.com/lihaoyi/mill/issues/989

Much of the work that is being done in https://github.com/scalameta/metals/pull/2154 will make this much smoother to do. I don't expect it to be too difficult.

Manually connect after creating the BSP config.

**Search terms**:
Mill BSP

ckipp01:
This adds the same type of functionality that we have with sbt where Mill can now also be a build server meaning that `metals.generate-bsp-config` and also `metals.bsp-switch` will apply to it. This also cleans up a few things that I did in #3304, fixes a bug in `Semver`, and also introduces a few small changes in `BuildServerProvider.scala` that are needed if a build tool and the name they use as a build server don't match.

Relates to https://github.com/com-lihaoyi/mill/issues/1546 and also closes #3307

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
@@ -7,7 +7,6 @@ import scala.concurrent.Future
 
 import scala.meta.internal.bsp.BspConfigGenerationStatus._
 import scala.meta.internal.builds.BuildServerProvider
-import scala.meta.internal.builds.BuildTool
 import scala.meta.internal.builds.BuildTools
 import scala.meta.internal.builds.SbtBuildTool
 import scala.meta.internal.metals.BloopServers
@@ -50,7 +49,7 @@ class BspConnector(
 
   private def resolveExplicit(): Option[BspResolvedResult] = {
     tables.buildServers.selectedServer().flatMap { sel =>
-      if (sel == BspConnector.BLOOP_SELECTED) Some(ResolvedBloop)
+      if (sel == BloopServers.name) Some(ResolvedBloop)
       else
         bspServers
           .findAvailableServers()
@@ -210,9 +209,9 @@ class BspConnector(
         case buildTool: BuildServerProvider
             if !foundServers
               .exists(details =>
-                details.getName() == buildTool.executableName
+                details.getName() == buildTool.getBuildServerName
               ) =>
-          buildTool.executableName -> Left(buildTool)
+          buildTool.getBuildServerName -> Left(buildTool)
       }
       .toMap
 
@@ -242,18 +241,18 @@ class BspConnector(
      * generate a bsp config has happened.
      */
     def handleGenerationStatus(
-        buildTool: BuildTool,
+        buildTool: BuildServerProvider,
         status: BspConfigGenerationStatus
     ): Boolean = status match {
       case BspConfigGenerationStatus.Generated =>
-        tables.buildServers.chooseServer(buildTool.executableName)
+        tables.buildServers.chooseServer(buildTool.getBuildServerName)
         true
       case Cancelled => false
       case Failed(exit) =>
         exit match {
           case Left(exitCode) =>
             scribe.error(
-              s"Creation of .bsp/${buildTool.executableName} failed with exit code: $exitCode"
+              s"Creation of .bsp/${buildTool.getBuildServerName} failed with exit code: $exitCode"
             )
             client.showMessage(
               Messages.BspProvider.genericUnableToCreateConfig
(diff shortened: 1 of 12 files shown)


metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -334,7 +328,3 @@ class BspConnector(
     }
   }
 }
-
-object BspConnector {
-  final val BLOOP_SELECTED = "BLOOP"
```
ckipp01, line 339: I realized we probably don't even need this, so I ripped it out of all the places and now just use `Bloop` from `BloopServers.name`. It simplifies it a bit.

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
```diff
@@ -282,14 +281,8 @@ class BspConnector(
                   args => bspConfigGenerator.runUnconditionally(buildTool, args)
                 )
                 .map(status => handleGenerationStatus(buildTool, status))
-            case Right(connectionDetails)
```
ckipp01, line 285: I think this _may_ have been what @dos65 was hinting at in my previous pr, but I realized this also isn't needed here.

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
```diff
@@ -12,17 +12,23 @@ object SemVer {
...
     def >(that: Version): Boolean = {
+      lazy val baseVersionEqual =
+        this.major == this.major && this.minor == that.minor && this.patch == that.patch
+
       this.major > that.major ||
       (this.major == that.major && this.minor > that.minor) ||
       (this.major == that.major && this.minor == that.minor && this.patch > that.patch) ||
       // 3.0.0-RC1 > 3.0.0-M1
-      this.releaseCandidate.isDefined && that.milestone.isDefined ||
+      baseVersionEqual && this.releaseCandidate.isDefined && that.milestone.isDefined ||
```
ckipp01, line 22: So I realized the logic here was missing a bit since previously it was showing that `SemVer.isCompatibleVersion("0.10.0-M4", "0.9.10")` was true since it wasn't correctly handling the cases where there was a major/minor/patch different plus one was a milestone or release candidate. You. can see what I mean more clearly in the tests I added for this.

tests/slow/src/test/scala/tests/mill/MillServerSuite.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import tests.BaseImportSuite
+import tests.MillBuildLayout
+import tests.MillServerInitializer
+
+/**
+ * Basic suite to ensure that a connection to sbt server can be made.
+ */
+class MillServerSuite
+    extends BaseImportSuite("sbt-server", MillServerInitializer) {
```
tgodzik, line 19:
```suggestion
    extends BaseImportSuite("mill-server", MillServerInitializer) {
```

tgodzik (approved):
LGTM! I have a very tiny, tiny comment - also thinking that we don't actually use those suite names and maybe we should remove them?

metals/src/main/scala/scala/meta/internal/bsp/BspConnector.scala
dos65: Yep, that's it!

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
```diff
@@ -12,17 +12,23 @@ object SemVer {
       milestone: Option[Int]
   ) {
     def >(that: Version): Boolean = {
+      lazy val baseVersionEqual =
+        this.major == this.major && this.minor == that.minor && this.patch == that.patch
+
```
dos65, line 17: Just a suggestion.

I've seen this part previously and was thinking to refactor it a bit if I will have to touch it.
Wdyt about replacing it on this impl?
```scala
  def <(that: Version): Boolean = {
      def toList(v: Version): List[Int] = {
        List(
          v.major,
          v.minor,
          v.patch,
          v.releaseCandidate.getOrElse(
            if (v.milestone.isEmpty) Int.MaxValue else 0
          ),
          v.milestone.getOrElse(
            if (v.releaseCandidate.isEmpty) Int.MaxValue else 0
          )
        )
      }

      val diff = toList(this)
        .zip(toList(that))
        .collectFirst {
          case (a, b) if a != b => a - b
        }
        .getOrElse(0)
      
      diff > 0
     }
```

dos65 (approved):
LGTM!

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
ckipp01: 🤔 I honestly think the other one might be easier to read. It's definitely more verbose, but it's clear to me. I don't have strong feelings on it if we really want it changed, but I don't mind how it is now.

dos65: np! feel free to ignore

Issue #3272, opened 2021-11-05
Align the documentation of how to do debugging with official docs, help beginners to get started with debugging


bjornregnell:
I'm trying to help my beginner programming students to grok how to use the debugger in vs code and one problem is that this page https://code.visualstudio.com/docs/editor/debugging does not mention Scala and this page https://scalameta.org/metals/docs/editors/vscode/#running-and-debugging-your-code does not really show how to do debugging in vs code and it is not really aligned with the official doc on the debugger and how to set it up. 

**Expected behavior**

Align the info and make it easier for a beginner programmer to get started with debugging with Metals. Would it even be possible for the official MS doc on vs code to point also to the Scala extension and for the Metals VS code page to align or link to the setup as in the official guide to reduce confusion?

bjornregnell:
Example of misalignemnt: On the official docs it says to press the "Launch program" button but in Scala Metals it is "Run and Debug" but the it starts a run config and that is more diffcult than just press "debug" above your @main. But that does not happen if you dont have a build.sbt etc.

bjornregnell:
If you follow the offical docs you risk ending up with 
![image](https://user-images.githubusercontent.com/3601817/140535803-b2cbd5cf-e9cf-47dc-bd12-1335cd32c312.png)

bjornregnell:
I guess Metals need to fix a launch config in order for the official docs to work. If we cannot impact the offical docs then we need to adapt so they work. Searching for "vs code debug" will land most beginners on the official docs...

tgodzik:
Related https://github.com/scalameta/metals-vscode/pull/767 <- however I don't think it will cause Metals to pop up on the main page, but should be more discoverable.

bjornregnell:
How about providing a launch config out-of-the-box so the standard "Launch Program" button "just works"?

tgodzik:
> How about providing a launch config out-of-the-box so the standard "Launch Program" button "just works"? Is this related to the label? Or what does it take to provide such a config to avoid the 'JSON with Comments' popup...

I am thinking we could try and run the existing command that will try to discover a main method in the current file instead of creating the config with the pop ups, which I never really found too useful. 

We could also link to the official docs from the Metals page as it's much more complete when it comes to the debugging explanations, what do you think?

bjornregnell:
Well, as long as the button that is given if you just click the Bug-tab on the left and the launch button just works then its fine (I thought that youperhaps had to have a launch config in order for the button to work)

[...]

PR #3311, opened 2021-11-26
Add option to run everything in the current file

tgodzik:
This will enable the possibility to run/debug without specifying the configuration in `launch.json`

Some further work can be done on the discovery so that we would be able to automatically feel new configurations for the user.

Related to https://github.com/scalameta/metals/issues/3272

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/config/RunType.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/config/RunType.scala
@@ -3,12 +3,14 @@ package scala.meta.internal.metals.config
 object RunType {
   sealed trait RunType
   case object Run extends RunType
+  case object RunOrTestFile extends RunType
   case object TestFile extends RunType
   case object TestTarget extends RunType
 
   def fromString(string: String): Option[RunType] = {
     string match {
       case "run" => Some(Run)
+      case "runOrTestFile" => Some(RunOrTestFile)
       case "testFile" => Some(TestFile)
       case "testTarget" => Some(TestTarget)
       case _ => None
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
@@ -9,6 +9,7 @@ import java.util.Collections.singletonList
 import java.util.concurrent.TimeUnit
 import java.{util => ju}
 
+import scala.collection.concurrent.TrieMap
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.Promise
@@ -275,6 +276,51 @@ class DebugProvider(
     }
   }
 
+  private def resolveInFile(
+      buildTarget: BuildTargetIdentifier,
+      classes: TrieMap[
+        BuildTargetClasses.Symbol,
+        ScalaMainClass
+      ],
+      testClasses: TrieMap[
+        BuildTargetClasses.Symbol,
+        BuildTargetClasses.ClassName
+      ],
+      params: DebugDiscoveryParams
+  )(implicit ec: ExecutionContext) = {
+    val path = params.path.toAbsolutePath
+    semanticdbs
+      .textDocument(path)
+      .documentIncludingStale
+      .fold[Future[DebugSessionParams]] {
+        Future.failed(SemanticDbNotFoundException)
+      } { textDocument =>
+        lazy val tests = for {
+          symbolInfo <- textDocument.symbols
+          symbol = symbolInfo.symbol
+          testClass <- testClasses.get(symbol)
+        } yield testClass
+        val mains = for {
+          symbolInfo <- textDocument.symbols
+          symbol = symbolInfo.symbol
+          mainClass <- classes.get(symbol)
+        } yield mainClass
+        if (mains.nonEmpty) {
+          verifyMain(buildTarget, mains.toList, params)
+        } else if (tests.nonEmpty) {
+          Future(
+            new b.DebugSessionParams(
+              singletonList(buildTarget),
+              b.DebugSessionParamsDataKind.SCALA_TEST_SUITES,
+              tests.asJava.toJson
+            )
+          )
+        } else {
+          Future.failed(NoRunOptionException)
+        }
+      }
+  }
+
   /**
    * Given fully unresolved params this figures out the runType that was passed
    * in and then discovers either the main methods for the build target the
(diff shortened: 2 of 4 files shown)


ayoub-benali:
Does this still requires DAP support from the client ?
In the case of sublime DAP would handled by a separate plugin https://github.com/daveleroy/sublime_debugger (similarly to vim ?). That plugin also seems to require a config file which is confusing in term of setup

tgodzik:
> Does this still requires DAP support from the client ? In the case of sublime DAP would handled by a separate plugin https://github.com/daveleroy/sublime_debugger (similarly to vim ?). That plugin also seems to require a config file which is confusing in term of setup

It does require some kind of DAP client, but the command should start the process and return the port to connect to. So some work on the editor side is required to wire the output of the command and starting of the session.

ayoub-benali:
Thanks for the clarification. Is this simplification only for running files ? do test cases for example require the "old" way (config file).

tgodzik:
> Thanks for the clarification. Is this simplification only for running files ? do test cases for example require the "old" way (config file).

Actually the aim of the PR is to provide an universal way of running things:
- if there is a main class in the file, run that
- if there are multiple main classes, show quickpick and run the chosen class
- if there are test classes, run all of them

metals/src/main/scala/scala/meta/internal/metals/config/RunType.scala
```diff
@@ -3,12 +3,14 @@ package scala.meta.internal.metals.config
 object RunType {
   sealed trait RunType
   case object Run extends RunType
+  case object RunFile extends RunType
```
tgodzik, line 6: This is just a proposal, but I think this would be useful in a lot of different clients.

metals/src/main/scala/scala/meta/internal/metals/config/RunType.scala
```diff
@@ -3,12 +3,14 @@ package scala.meta.internal.metals.config
...
   case object TestFile extends RunType
   case object TestTarget extends RunType
 
   def fromString(string: String): Option[RunType] = {
     string match {
       case "run" => Some(Run)
+      case "runFile" => Some(RunFile)
```
ckipp01, line 13: 🤔 the only thing that might be confusing about this is that depending on how the client sets it, it might totally not be clear. For example `nvim-metals` users will have this in their configuration if they use dap:

```
  dap.configurations.scala = {
    {
      type = "scala",
      request = "launch",
      name = "Run",
      metals = {
        runType = "run",
        --args = { "firstArg", "secondArg", "thirdArg" }, -- here just as an example
      },
    },
    {
      type = "scala",
      request = "launch",
      name = "Test File",
      metals = {
        runType = "testFile",
      },
    },
    {
      type = "scala",
      request = "launch",
      name = "Test Target",
      metals = {
        runType = "testTarget",
      },
    },
  }
```
So we'd just need to add another one that sends in `runFile`, but it's not really clear what the difference between `run` and `runFile` is without knowing how these work. Do you think maybe just having these documented super well is enough, or can you think of another name that would maybe be clearer? I'm sort of failing on coming up with anything better, so maybe it's fine. idn. wdyt?

tgodzik: `runOrTestFile` ? Or maybe `invokeFile`? Or just `debugFile` ? I wasn't perfectly sure about the name either.

ckipp01: Thinking about this a bit, maybe we went about this a bit wrong when we first did it. Maybe instead of testFile, testTarget, run, etc we could simplify it.

`runOrTestFile` -> executes the main or the tests in the file
`runOrTestTarget` -> runs the main or tests in the build target

Actually now that I'm thinking about it, this may not work in the scenario you have a main in your build target. Or maybe it would? We could detect that their are both and just ask the user if they want to run or test? Not sure.

I'd say if we keep the current way then maybe be explicit and just change it to `runOrTestFile`, but I don't feel strong about it.

if the other option would work maybe it'd be nicer for the user? Really not sure.

tgodzik: I think we could add `runOrTestFile` and then deprecate `run` and `testFile`?

I am not so sure about `runOrTestTarget` - I think `testTarget` is ok, since it's a different logic, user wants to run all tests and not just choose a test/main, so let's not add an additional step there. We could have a fallback for `runOrTestFile` that tries main or tests from other files, but it could be ok to not include it for now. `runOrTestTarget` would need to have to different purposes - run a single main or run all tests, which is kind of weird.

I think all of those should server a UX purpose:
- `runOrTestFile` - run whatever I am working on quickly
- `testTarget` - let's run all tests to see if I didn't break anything

tgodzik: Renamed it to `runOrTestFile`, let me know what you think!

metals/src/main/scala/scala/meta/internal/metals/config/RunType.scala
ckipp01: Do you think we should match the name with `RunOrTestFile`? No strong feeling though

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -629,6 +677,10 @@ case object WorkspaceErrorsException
     extends Exception(
       s"Cannot run class, since the workspace has errors."
     )
+case object NoRunOptionException
+    extends Exception(
+      s"There is nothing to run in the current file."
```
ckipp01, line 682:
```suggestion
      s"There is nothing to run or test in the current file."
```

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -275,6 +276,51 @@ class DebugProvider(
...
+        } else if (tests.nonEmpty) {
+          Future(
+            new b.DebugSessionParams(
+              singletonList(buildTarget),
+              b.DebugSessionParamsDataKind.SCALA_TEST_SUITES,
+              tests.asJava.toJson
+            )
+          )
+        } else {
+          Future.failed(NoRunOptionException)
```
ckipp01, line 319: _Total aside_, but on my TODO list to to refactor all this fails out. I'm not a huge fan of using failing futures for control flow. I think we can probably model all of these differently. One day 😆

ckipp01 (approved):
Couple nits, but other than that LGTM! I tried it out with `nvim-metals` and all seemed to work well!

metals/src/main/scala/scala/meta/internal/metals/config/RunType.scala
tgodzik: Forgot about it! Fixed!

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
kpodsiad: Are you talking about a dedicated error channel?

PR #3314, opened 2021-11-28
Add isCommandInHtmlSupported to InitializationOptions doc

ayoub-benali:
Noticed that it was missing while working on the sublime implementation of `analyze-stacktrace`


ckipp01 (approved):
Thanks for noticing and adding this in! LGTM!

PR #3315, opened 2021-11-28
Add missing commands to client commands documentation

ayoub-benali:
I initially noticed that `ShowStacktrace` was missing but I added the other missing ones as well.


ckipp01 (approved):
Thanks @ayoub-benali! LGTM!

PR #3316, opened 2021-11-28
Fix json rendering in documentation

ayoub-benali:
It seems that my previous PR https://github.com/scalameta/metals/pull/3315 broke the json rendering in the doc.
This is the current look
image


ckipp01 (approved):
ahh, thanks for fixing this and good catch. LGTM!

PR #3320, opened 2021-11-29
[Scala3] Hover - improve tpe selection for signature and expression

dos65:
Adapted the `seenFrom` logic from Scala2.
Now types in hover are close to be in pair in both implementations.


tgodzik (approved):
LGTM!

PR #3330, opened 2021-12-01
Improve the "(currently using)" message during bsp-switch.

ckipp01:
This is just a small change to improve the bsp-switch label that
includes the "(currently using)" message. Previously if you've never
made an explicit choice it wouldn't tell you which you were using. This
change defaults to the current connection if there is no explicit
choice made.


tgodzik (approved):
LGTM

PR #3332, opened 2021-12-01
Fix issues with non-latest Scala versions

tgodzik:
Previously, for mtags_2.12.8 we would download scala library 2.12.15. Now, we download the correct Scala version.

This popped up in https://github.com/scalameta/metals/pull/3278/files where we switched to downloadMtags method as default.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
@@ -102,7 +102,11 @@ final class Embedded(
      */
     resolutionParams.addExclusion("io.get-coursier", "interface")
     val jars =
-      Embedded.downloadMdoc(scalaVersion, scalaBinaryVersion, resolutionParams)
+      Embedded.downloadMdoc(
+        scalaVersion,
+        scalaBinaryVersion,
+        Some(resolutionParams)
+      )
 
     val parent = new MdocClassLoader(this.getClass.getClassLoader)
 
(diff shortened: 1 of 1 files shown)


dos65 (approved):
Thx for the fix!
LGTM

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -224,45 +231,48 @@ object Embedded {
...
   def downloadMdoc(
       scalaVersion: String,
       scalaBinaryVersion: String,
-      resolutionParams: ResolutionParams = ResolutionParams.create()
+      resolutionParams: Option[ResolutionParams] = None
   ): List[Path] =
     downloadDependency(
       mdocDependency(scalaVersion, scalaBinaryVersion),
-      scalaVersion,
+      scalaVersion = None,
```
tgodzik, line 275: We don't want to force scala version here since mdoc is released with binary compatibility

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -272,7 +282,7 @@ object Embedded {
       s"organize-imports_$scalaBinaryVersion",
       BuildInfo.organizeImportVersion
     )
-    downloadDependency(dep, scalaBinaryVersion)
+    downloadDependency(dep, scalaVersion = None)
```
tgodzik, line 285: Same here:  organize-imports is released with binary compatibility

tgodzik:
@dos65 I had to change it a bit to make it work, could you rereview?

dos65 (approved):
:rocket:

PR #3334, opened 2021-12-03
Remove sublime version from the doc

ayoub-benali:
After https://github.com/scalameta/metals-sublime/pull/57 only ST4 will be supported
Also technically the version number is dropped from the name now it is simply Sublime Text.


tgodzik (approved):
LGTM!

PR #3339, opened 2021-12-06
Cross tests - fix ignorance for `test-mtags-dyn`

dos65:
`test-mtags-dyn` might be called with a versions that isn't specified in
BuildInfo


tgodzik (approved):
LGTM!

PR #3340, opened 2021-12-06
Add information about source of exceptions

tgodzik:
Previously, users would report an issue, but we would not be able to point to any particular path or dependency. Now, we try to print much more information for the users in exceptional cases.

The issue popped up for example here: https://gitter.im/scalameta/metals?at=61a8d7a98853d31640283bf5

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -2687,8 +2687,12 @@ class MetalsLanguageServer(
       onError = {
         case e @ (_: ParseException | _: TokenizeException) =>
           scribe.error(e.toString)
-        case e: InvalidJarException =>
+        case e: IndexingExceptions.InvalidJarException =>
           scribe.warn(s"invalid jar: ${e.path}")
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -2687,8 +2687,12 @@ class MetalsLanguageServer(
       onError = {
         case e @ (_: ParseException | _: TokenizeException) =>
           scribe.error(e.toString)
-        case e: InvalidJarException =>
+        case e: IndexingExceptions.InvalidJarException =>
           scribe.warn(s"invalid jar: ${e.path}")
+        case e: IndexingExceptions.PathIndexingException =>
+          scribe.warn(s"issues while parsing: ${e.path}")
+        case e: IndexingExceptions.InvalidSymbolException =>
+          scribe.warn(s"searching for `${e.symbol}` failed")
```
kpodsiad, line 2695: Maybe it's worth to log `e.underlying` exception?

kpodsiad (approved):

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala

PR #3351, opened 2021-12-10
[Build] Add  `quick-publish-local` command

dos65:
Currently running `+publishLocal` takes a significant amount of time.
Added this command to speed up the local development workflow.


tgodzik (approved):
Good idea, thanks!

Issue #3352, opened 2021-12-10
Anything using quick pick is broken in the latest SNAPSHOT


ckipp01:
This commit introduced a regression where Gson isn't able to correctly handle the Scala Option. So when you try to create a new Scala File or a new Scala Project you see this in the logs and nothing happens:

```
SEVERE: Failed to invoke public scala.Option() with no args
java.lang.RuntimeException: Failed to invoke public scala.Option() with no args
        at com.google.gson.internal.ConstructorConstructor$3.construct(ConstructorConstructor.java:113)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:212)
        at com.google.gson.Gson.fromJson(Gson.java:963)
        at com.google.gson.Gson.fromJson(Gson.java:1034)
        at org.eclipse.lsp4j.jsonrpc.json.adapters.MessageTypeAdapter.fromJson(MessageTypeAdapter.java:339)
        at org.eclipse.lsp4j.jsonrpc.json.adapters.MessageTypeAdapter.parseResult(MessageTypeAdapter.java:221)
        at org.eclipse.lsp4j.jsonrpc.json.adapters.MessageTypeAdapter.read(MessageTypeAdapter.java:135)
        at org.eclipse.lsp4j.jsonrpc.json.adapters.MessageTypeAdapter.read(MessageTypeAdapter.java:55)
        at com.google.gson.Gson.fromJson(Gson.java:963)
        at org.eclipse.lsp4j.jsonrpc.json.MessageJsonHandler.parseMessage(MessageJsonHandler.java:119)
        at org.eclipse.lsp4j.jsonrpc.json.MessageJsonHandler.parseMessage(MessageJsonHandler.java:114)
        at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:193)
        at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:94)
        at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:113)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[...]

PR #3353, opened 2021-12-10
Fix quickpick

kpodsiad:
My idea was to hide implementation details and fact that we have to cope with nulls so there is private `rawMetalsQuickPick`, which is responsible for communication (and is gson-friendly). There is also public `metalsQuickPick` (so nothing changes at call sites) which gracefully transforms nullable results from a raw method into Option.

In addition to that, I grouped language clients in a `clients` package (maybe it should be a `clients.language` because there are also build ones?) and thanks to that those raw methods can be package private.

First commit contains changes in imports, the second one contains true changes.

@ckipp01 @tgodzik Let me know what do you think about this approach. If it's okay, I'll add unit tests to cover https://github.com/scalameta/metals/issues/3352.


ckipp01 (commented):
👍🏼 I like this approach, and I definitely think it's a good idea to start splitting stuff out, so also 👍🏼 from me on the clients package.

kpodsiad:
I'm afraid that #3352 cannot be easily tested, a real LSP client have to be run for that.

metals/src/main/scala/scala/meta/internal/metals/clients/language/MetalsLanguageClient.scala
```diff
@@ -65,9 +82,22 @@ trait MetalsLanguageClient
...
+  )(implicit
+      ec: ExecutionContext
+  ): CompletableFuture[Option[MetalsQuickPickResult]] =
+    rawMetalsQuickPick(params).asScala.map { result =>
+      if (result.cancelled != null && result.cancelled) {
+        None
+      } else {
+        Option(result.itemId).map(MetalsQuickPickResult(_))
+      }
+    }.asJava
```
dos65, line 100: Nitpick to conversions: java future -> scala future -> java future

Maybe replace `.asScala.map` with `thenApply` ?

dos65 (commented):
At the beginning I wanted to ask why there are a lot of file moves but I agree with @ckipp01 that it's a good direction

metals/src/main/scala/scala/meta/internal/metals/clients/language/MetalsLanguageClient.scala
kpodsiad: With `thenApply` execution context won't be needed, good point!

ckipp01 (approved):
Ha, thanks for splitting this into 2 commits, make it easier to review. Tested it all out and seems to fix the issues! Thanks for diving back into this! LGTM from me.

tgodzik (approved):
LGTM!

kpodsiad:
Probably I'll wait with merge until https://github.com/scalameta/metals/pull/3355 is merged. I don't want to potentially additional work to that PR.

kpodsiad:
@tgodzik could you merge this? There is a flaky test, but I don't want to re-run all jobs

Issue #3354, opened 2021-12-10
Implement LSP Code Lens Refresh Request


ayoub-benali:
Metals is using a custom command `metals-model-refresh` while there is an LSP command for the same purpose [Code Lens Refresh Request](https://microsoft.github.io/language-server-protocol/specification#codeLens_refresh)

This would avoid to the clients implementing custom commands.

**Expected behavior**
Metals using LSP commands when available. 

**Search terms**
LSP, code lens

PR #3355, opened 2021-12-11
Send Code lens refresh when supported by client

ayoub-benali:
Instead of the custom `metals-model-refresh`

Fix https://github.com/scalameta/metals/issues/3354

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
@@ -157,6 +157,8 @@ object ClientCommands {
     "Refresh model",
     """|Notifies the client that the model has been updated and it 
        |should be refreshed (e.g. by resending code lens request)
+       |
+       |Note: Metals will favor instead [Code Lens Refresh Request](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#codeLens_refresh) if the client has the capability.
        |""".stripMargin
   )
 
--- a/metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
@@ -155,10 +155,10 @@ object ClientCommands {
   val RefreshModel = new Command(
     "metals-model-refresh",
     "Refresh model",
-    """|Notifies the client that the model has been updated and it 
-       |should be refreshed (e.g. by resending code lens request)
+    """|**Note**: This request is deprecated and Metals will favor [Code Lens Refresh Request](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#codeLens_refresh) if supported by the client.
        |
-       |Note: Metals will favor instead [Code Lens Refresh Request](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#codeLens_refresh) if the client has the capability.
+       |Notifies the client that the model has been updated and it
+       |should be refreshed (e.g. by resending code lens request)
        |""".stripMargin
   )
 
--- a/metals/src/main/scala/scala/meta/internal/metals/ClientConfiguration.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ClientConfiguration.scala
@@ -19,7 +19,7 @@ case class ClientConfiguration(initialConfig: MetalsServerConfig) {
   private var initializationOptions = InitializationOptions.Default
   private var refreshSupport = false
 
-  def update(params: InitializeParams) = {
+  def update(params: InitializeParams): Unit = {
     experimentalCapabilities =
       ClientExperimentalCapabilities.from(params.getCapabilities)
     initializationOptions = InitializationOptions.from(params)
@@ -171,5 +171,5 @@ case class ClientConfiguration(initialConfig: MetalsServerConfig) {
 }
 
 object ClientConfiguration {
-  def Default() = ClientConfiguration(MetalsServerConfig())
+  def Default(): ClientConfiguration = ClientConfiguration(MetalsServerConfig())
 }
--- a/metals/src/main/scala/scala/meta/internal/metals/ClientConfiguration.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ClientConfiguration.scala
@@ -4,6 +4,7 @@ import scala.meta.internal.metals.Configs.GlobSyntaxConfig
 import scala.meta.internal.metals.config.DoctorFormat
 import scala.meta.internal.metals.config.StatusBarState
 
+import org.eclipse.lsp4j.ClientCapabilities
 import org.eclipse.lsp4j.InitializeParams
 
 /**
@@ -17,19 +18,14 @@ case class ClientConfiguration(initialConfig: MetalsServerConfig) {
 
   private var experimentalCapabilities = ClientExperimentalCapabilities.Default
   private var initializationOptions = InitializationOptions.Default
-  private var refreshSupport = false
+  private var clientCapabilities: Option[ClientCapabilities] = None
 
   def update(params: InitializeParams): Unit = {
-    experimentalCapabilities =
-      ClientExperimentalCapabilities.from(params.getCapabilities)
+    experimentalCapabilities = ClientExperimentalCapabilities.from(
+      params.getCapabilities
+    )
     initializationOptions = InitializationOptions.from(params)
-    val capabilities = params.getCapabilities()
-    val codeLenseRefreshSupport: Option[Boolean] = for {
-      workspace <- Option(capabilities.getWorkspace())
-      codeLens <- Option(workspace.getCodeLens())
-      refreshSupport <- Option(codeLens.getRefreshSupport())
-    } yield refreshSupport
-    refreshSupport = codeLenseRefreshSupport.getOrElse(false)
+    clientCapabilities = Some(params.getCapabilities())
   }
 
   def extract[T](primary: Option[T], secondary: Option[T], default: T): T = {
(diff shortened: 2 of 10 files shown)


metals/src/main/scala/scala/meta/internal/metals/Compilations.scala
```diff
@@ -218,7 +219,7 @@ final class Compilations(
         // See https://github.com/scalacenter/bloop/issues/1067
         classes.rebuildIndex(targets).foreach { _ =>
           if (targets.exists(isCurrentlyFocused)) {
-            languageClient.refreshModel()
+            languageClient.refreshModel(codeLenseRefreshSupport())
```
ayoub-benali, line 222: I am not sure if this should still run inside a `andThen` block :confused:

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageClient.scala
```diff
@@ -42,9 +42,20 @@ trait MetalsLanguageClient
...
+      metalsExecuteClientCommand(params)
+  }
+
+  final def refreshModel(
+      codeLensRefreshSupport: Boolean
+  ): CompletableFuture[Void] = {
+    if (codeLensRefreshSupport) this.refreshCodeLenses
+    else {
+      val params = ClientCommands.RefreshModel.toExecuteCommandParams()
+      CompletableFuture.runAsync(RunnableClientCommand(params))
```
ayoub-benali, line 57: I didn't find a better way to go from `Unit` to `CompletableFuture[Void]`

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageClient.scala
```diff
@@ -42,9 +42,18 @@ trait MetalsLanguageClient
...
-  final def refreshModel(): Unit = {
-    val params = ClientCommands.RefreshModel.toExecuteCommandParams()
-    metalsExecuteClientCommand(params)
+  final def refreshModel(
+      codeLensRefreshSupport: Boolean
+  ): CompletableFuture[Void] = {
+    if (codeLensRefreshSupport) this.refreshCodeLenses
+    else {
+      val params = ClientCommands.RefreshModel.toExecuteCommandParams()
+      CompletableFuture.runAsync(
```
ayoub-benali, line 51: I didn't find a better way to go from `Unit` to `CompletableFuture[Void]`

kpodsiad (commented):
Nice work! I left a few thoughts

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageClient.scala
ayoub-benali: That works but the future isn't really completed. There is still an command to send, no ?

kpodsiad: `metalsExecuteClientCommand` returns Unit so it seems more like a notification/fire & forget semantics. 
@tgodzik your attention will be needed here :D

metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
```diff
@@ -82,6 +82,15 @@ final class ConfiguredLanguageClient(
     }
   }
 
+  override def refreshModel(): CompletableFuture[Unit] = {
+    if (clientConfig.codeLenseRefreshSupport)
+      this.refreshCodeLenses.thenApply(_ => ())
+    else {
+      val params = ClientCommands.RefreshModel.toExecuteCommandParams()
+      CompletableFuture.completedFuture(metalsExecuteClientCommand(params))
```
kpodsiad, line 90: I think we should still use `underlying` instance here instead of `this`. 
As far as I remember:
- the most depth language client is created in `Main.scala` - `val clientProxy = launcher.getRemoteProxy`
- then it's passed to the `server.connectToLanguageClient(clientProxy)`
- `MetalsLanguaseServer` has already `DelegatingLanguageClient` created with NoopLC, but `connectToLanguageClient` changes that

So finally it should looks like:
clientProxy <- `ConfiguredLanguageClient` <- `DelegatingLanguageClient` <- `MetalsLanguageServer`

ayoub-benali: If we do `underlying.metalsExecuteClientCommand` we skip the check that is bellow (`clientConfig.isExecuteClientCommandProvider`). 

BTW shouldn't that check now be done in here ?

kpodsiad: I think we are both right and wrong at the same time hah :D
`refreshCodeLenses` should be called on underlying since it comes from `lsp.LanguageClient` and `metalsExecuteClientCommand` should be called on this, just as you wrote.

Yeah, since code lenses are implemented as a debug request, that check should be moved here

ayoub-benali: I moved the check now but `refreshCodeLenses` should be independent of debug provider.

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -157,6 +157,8 @@ object ClientCommands {
     "Refresh model",
     """|Notifies the client that the model has been updated and it 
        |should be refreshed (e.g. by resending code lens request)
+       |
+       |Note: Metals will favor instead [Code Lens Refresh Request](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#codeLens_refresh) if the client has the capability.
```
ckipp01, line 161: I'm not sure what others think of this, but I'd even be in favor of putting a not here that this is now deprecated and that instead you should use the code lens refresh if your editor supports it. You include the second part, but I think we can even word this more strongly to push people towards the actual refresh.

metals/src/main/scala/scala/meta/internal/metals/ClientConfiguration.scala
```diff
@@ -4,20 +4,33 @@ import scala.meta.internal.metals.Configs.GlobSyntaxConfig
...
+    experimentalCapabilities =
+      ClientExperimentalCapabilities.from(params.getCapabilities)
+    initializationOptions = InitializationOptions.from(params)
+    val capabilities = params.getCapabilities()
+    val codeLenseRefreshSupport: Option[Boolean] = for {
+      workspace <- Option(capabilities.getWorkspace())
+      codeLens <- Option(workspace.getCodeLens())
+      refreshSupport <- Option(codeLens.getRefreshSupport())
+    } yield refreshSupport
+    refreshSupport = codeLenseRefreshSupport.getOrElse(false)
```
ckipp01, line 32: Maybe style preference, but I don't love having the `private var refreshSupport` especially if we end up with more of these in the future.

What do you think instead having `private var capabilities`. Then have the `codeLenseRefreshSupport` method just pull it from there. That also would mean it'd need to do that each time, which isn't ideal, but it's odd to me that we have `refreshSupport` up there since it doesn't seem to fit along with the other `experimentalCababilities` and `initializationOptions` and the pattern of having the methods grab values out of them. Wdyt?

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
ayoub-benali: Something like this?

**Note**: This request is deprecated and Metals will favor [Code Lens Refresh Request](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#codeLens_refresh) if supported by the client. 
       
Notifies the client that the model has been updated and it
should be refreshed (e.g. by resending code lens request)

metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
```diff
@@ -82,19 +82,21 @@ final class ConfiguredLanguageClient(
...
+  override def refreshModel(): CompletableFuture[Unit] = {
+    if (clientConfig.codeLenseRefreshSupport)
+      underlying.refreshCodeLenses.thenApply(_ => ())
+    else if (
+      clientConfig.isExecuteClientCommandProvider && clientConfig.isDebuggingProvider
+    ) {
+      val params = ClientCommands.RefreshModel.toExecuteCommandParams()
+      CompletableFuture.completedFuture(metalsExecuteClientCommand(params))
+    } else CompletableFuture.completedFuture(())
+  }
```
kpodsiad, line 94: Ok, I realized that `refreshModel` in vscode is also used for `test explorer` so current logic breaks its behavior. 
Following should work in both cases:
```scala
    if(clientConfig.isExecuteClientCommandProvider && clientConfig.isDebuggingProvider) {
      val params = ClientCommands.RefreshModel.toExecuteCommandParams()
      metalsExecuteClientCommand(params)
    }

    if (clientConfig.isDebuggingProvider && clientConfig.codeLenseRefreshSupport)
      underlying.refreshCodeLenses.thenApply(_ => ())
    else CompletableFuture.completedFuture(())
  ```
I also added `clientConfig.isDebuggingProvider` to the condition since code lenses are implemented via the DAP server.

kpodsiad (approved):
Great work @ayoub-benali!

metals/src/main/scala/scala/meta/internal/metals/ConfiguredLanguageClient.scala
ayoub-benali: You mean `metals-model-refresh` shouldn't be deprecated and it still serves a different purpose than `workspace/codeLens/refresh` ?

kpodsiad (approved):
LGTM!

tgodzik (approved):
LGTM!

PR #3356, opened 2021-12-11
Use Mill `--import` to import bloop when supported

lolgab:
Mill `0.9.10+` and `0.10.0-M4+` have a new feature.
You can import plugins on the fly without modifying the build file, by passing `--import ivy:com.lihaoyi::mill-contrib-bloop:` where the missing version is automatically replaced by `$MILL_VERSION`.
This PR makes metals leverage this functionality in versions that support it.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
@@ -41,9 +41,7 @@ case class MillBuildTool(userConfig: () => UserConfiguration)
 
   override def redirectErrorOutput: Boolean = true
 
-  private def putTogetherArgs(cmd: List[String], workspace: AbsolutePath) = {
-    val millVersion = getMillVersion(workspace)
-
+  private def putTogetherArgs(cmd: List[String], millVersion: String) = {
     // In some environments (such as WSL or cygwin), mill must be run using interactive mode (-i)
     val fullcmd = if (Properties.isWin) "-i" :: cmd else cmd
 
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
```diff
@@ -57,10 +55,23 @@ case class MillBuildTool(userConfig: () => UserConfiguration)
 
   }
 
+  private def bloopImportArgs(millVersion: String) = {
+    val isImportSupported = SemVer.isCompatibleVersion(
+      "0.9.10",
+      millVersion
+    ) && (SemVer.isLaterVersion(millVersion, "0.10.0-M1") || SemVer
+      .isCompatibleVersion("0.10.0-M4", millVersion))
```
tgodzik, line 63: If it's supported then we can just have:
```suggestion
    val isImportSupported = SemVer.isCompatibleVersion(
      "0.9.10",
      millVersion
    )
```
I think?

tgodzik (commented):
This is great! Thank you very much! Just one question from me

metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
lolgab: This way we break the support from `0.10.0-M1` to `0.10.0-M3` since it was introduced in M4 and backported in `0.9.10`.
This might be ok since they are milestone versions, but I tried to not break anything.

tgodzik (approved):
LGTM!

Issue #3359, opened 2020-10-05
Add go to definition for implicit decorations


tgodzik:
**Is your feature request related to a problem? Please describe.**
We've added a feature that enables users to see implicit parameters and inferred types in their code either thanks to decoration protocol or in hover: https://github.com/scalameta/metals/pull/2103

We are still unable hover to go to definition of any of those implicits or types.

**Describe the solution you'd like**
There is a number of options here, but I think best one would be to add code lenses for any line that contains implicits. This solutions could work in all supported editors ans should be easy to switch off. Although visually it would take up most space.

**Describe alternatives you've considered**

- if cursor is just before the implicit or inferred type we could add that implicit's/type definition to the locations returned already by the compile - this seems very fragile

- add command to hover that would invoke a command to go to a definition - this will only work if an editor supports links, which in essence is only VS Code

**Additional context**

**Search terms**:
implicit decoration parameters

PR #3360, opened 2021-12-13
Add go to definition via links to synthetic decorations

tgodzik:
Previously, we would only display types of anything that was shown in the implicit decorations. Now, instead we show a shorter clickable link that effectively does `go to definition`

Fixes https://github.com/scalameta/metals/issues/3359

Demo!

definition-implicits

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
@@ -14,6 +14,7 @@ import scala.meta.internal.metals.Buffers
 import scala.meta.internal.metals.ClientConfiguration
 import scala.meta.internal.metals.HoverExtParams
 import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.ServerCommands
 import scala.meta.internal.metals.UserConfiguration
 import scala.meta.internal.metap.PrinterSymtab
 import scala.meta.internal.mtags.Md5Fingerprints
@@ -107,7 +108,7 @@ final class SyntheticsDecorationProvider(
           val edit = buffer.tokenEditDistance(path, textDocument.text, trees)
           val printer = new SemanticdbTreePrinter(
             isHover = true,
-            toHoverString(textDocument),
+            toHoverString(textDocument, params.textDocument.getUri()),
             PrinterSymtab.fromTextDocument(textDocument),
             clientConfig.icons().rightArrow,
             clientConfig.icons().ellipsis
@@ -174,11 +175,8 @@ final class SyntheticsDecorationProvider(
     if (interestingSynthetics.nonEmpty)
       addToHover(
         pcHover,
-        "**Synthetics**:\n"
-          +
-            HoverMarkup(
-              interestingSynthetics.mkString("\n")
-            )
+        "**Synthetics**:\n\n"
+          + interestingSynthetics.mkString("\n")
       )
     else None
   }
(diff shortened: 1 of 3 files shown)


tgodzik:
@ayoub-benali I don't remember where we finished with the sublime commands, but we could for sure change it to work also in Sublime Text

metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
```diff
@@ -269,23 +265,70 @@ final class SyntheticsDecorationProvider(
...
+  ): String = {
+    if (symbol.isLocal) localSymbolName(symbol, textDoc)
+    else {
+      symbol
+        .replace("/", ".")
+        .stripSuffix(".")
+        .stripSuffix("#")
+        .stripPrefix("_empty_.")
+        .replace("#", ".")
+        .replace("()", "")
```
dos65, line 287: it might be disambiguator with number like :`(+1)`

dos65 (commented):
Wow!Now I see why you asked to enable markdown on VSCode :smile: 

Looks good, I left just one minor comment

metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
tgodzik: Good catch, changed to .replaceAll(raw"\(\+?\d*\)", "")

ayoub-benali:
> @ayoub-benali I don't remember where we finished with the sublime commands, but we could for sure change it to work also in Sublime Text

@tgodzik I used a regex to fix the links ([here](https://github.com/scalameta/metals-sublime/blob/master/core/handle_execute_client.py#L51)) but that is very fragile. 

I also created this issue https://github.com/scalameta/metals-feature-requests/issues/242 to track the problem but with this PR adding links I think now Metals should allow the client to control the links format.

dos65 (approved):

tgodzik:
> > @ayoub-benali I don't remember where we finished with the sublime commands, but we could for sure change it to work also in Sublime Text
> 
> @tgodzik I used a regex to fix the links ([here](https://github.com/scalameta/metals-sublime/blob/master/core/handle_execute_client.py#L51)) but that is very fragile.
> 
> I also created this issue [scalameta/metals-feature-requests#242](https://github.com/scalameta/metals-feature-requests/issues/242) to track the problem but with this PR adding links I think now Metals should allow the client to control the links format.

Last question, is Sublime able to show the links in hovers?

ayoub-benali:
Yes it is possible to show links in the popup. This is for example native to LSP package
Peek 2021-12-13 16-36

tgodzik:
> Yes it is possible to show links in the popup. This is for example native to LSP package

Awesome! We should be able to provide the same experience then.

ckipp01 (approved):
Very cool! LGTM!

ayoub-benali:
I tried latest snapshot and the synthetics links shows up properly but the URL doesn't work because it is a command unknown to sublime. Same problem as https://github.com/scalameta/metals-feature-requests/issues/242
Peek 2021-12-14 00-27

tgodzik:
> I tried latest snapshot and the synthetics links shows up properly but the URL doesn't work because it is a command unknown to sublime. Same problem as [scalameta/metals-feature-requests#242](https://github.com/scalameta/metals-feature-requests/issues/242) Peek 2021-12-14 00-27

I will update it before the release, should not be too much issue.

PR #3374, opened 2021-12-15
Use full range when using go-to definition for synthetics

tgodzik:
Previously, we would only point to the first character, which would work differently to how normally go to definition works. Now, we point to the full range.


kpodsiad (approved):
LGTM!

dos65 (approved):

PR #3375, opened 2021-12-15
Enable to create file with just package

kpodsiad:
Closes https://github.com/scalameta/metals-feature-requests/issues/246

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/newScalaFile/NewFileTypes.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/newScalaFile/NewFileTypes.scala
@@ -9,6 +9,11 @@ object NewFileTypes {
     def toQuickPickItem: MetalsQuickPickItem = MetalsQuickPickItem(id, label)
   }
 
+  case object ScalaFile extends NewFileType {
+    override val id: String = "scala-file"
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/newScalaFile/NewFileTypes.scala
```diff
@@ -9,6 +9,11 @@ object NewFileTypes {
     def toQuickPickItem: MetalsQuickPickItem = MetalsQuickPickItem(id, label)
   }
 
+  case object JustPackage extends NewFileType {
+    override val id: String = "just-package"
+    override val label: String = "Empty file with package"
+  }
```
dos65, line 15: Maybe it would be ok to name it just as `ScalaFile` / `id = scala-file / `label = Scala file`?
It should work similarly no new file creation as it's currently works (package is added automatically) with only difference that you don't need to type `.scala`

dos65 (commented):
@kpodsiad you need to remove `.only` to fix tests.

Overall looks good, I have only one suggestion about naming.

ckipp01 (approved):
I also like @dos65's suggestion. Other than that LGTM!

dos65 (approved):
:rocket:

PR #3378, opened 2021-12-16
Support sublime command links

ayoub-benali:
Fix https://github.com/scalameta/metals-feature-requests/issues/242

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
@@ -203,15 +203,19 @@ sealed trait CommandHTMLFormat {
 }
 object CommandHTMLFormat {
   object Sublime extends CommandHTMLFormat {
+    val toEscape = Set('"', '<', '>', '&', '\'')
+    def escape(args: String): String = {
+      // The lib used to convert markdown to html in sublime doesn't properly
+      // recognize URL encoding so we have to use hexadecimal html encoding
+      args.flatMap {
+        case char if toEscape.contains(char) => s"&#x${char.toHexString};"
+        case char => char.toString
+      }
+    }
     override def createLink(
         commandId: String,
         arguments: List[String]
     ): String = {
-      def escape(args: String): String = {
-        // The lib used to convert markdown to html in sublime doesn't properly
-        // recognize URL encoding so we have to use hexadecimal html encoding
-        args.replaceAll("\"", "&#x22;")
-      }
       // sublime expect commands to follow the under_scores format
       val id = commandId.replaceAll("-", "_")
       val encodedArguments =
--- a/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
@@ -58,7 +58,6 @@ final case class InitializationOptions(
     inputBoxProvider: Option[Boolean],
     isExitOnShutdown: Option[Boolean],
     isHttpEnabled: Option[Boolean],
-    isCommandInHtmlSupported: Option[Boolean],
     commandInHtmlFormat: Option[CommandHTMLFormat],
     openFilesOnRenameProvider: Option[Boolean],
     quickPickProvider: Option[Boolean],
@@ -83,7 +82,6 @@ object InitializationOptions {
 
   val Default: InitializationOptions = InitializationOptions(
     CompilerInitializationOptions.default,
-    None,
     None,
     None,
     None,
@@ -143,8 +141,6 @@ object InitializationOptions {
       inputBoxProvider = jsonObj.getBooleanOption("inputBoxProvider"),
       isExitOnShutdown = jsonObj.getBooleanOption("isExitOnShutdown"),
       isHttpEnabled = jsonObj.getBooleanOption("isHttpEnabled"),
-      isCommandInHtmlSupported =
-        jsonObj.getBooleanOption("isCommandInHtmlSupported"),
       commandInHtmlFormat = jsonObj
         .getStringOption("commandInHtmlFormat")
         .flatMap(CommandHTMLFormat.fromString),
--- a/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
@@ -199,10 +199,13 @@ object InitializationOptions {
 }
 
 sealed trait CommandHTMLFormat {
+  val value: String
   def createLink(commandId: String, arguments: List[String]): String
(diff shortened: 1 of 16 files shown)


ayoub-benali:
@tgodzik I fixed the sublime format and tests. 
The links in analyse stacktrace response work now natively. 

But there is still an issue on the synthetics links, as you can see bellow this isn't valid markdown
```
'\n**Synthetics**:\n\n[augmentString](subl:lsp_metals_goto {"parameters": ["scala/Predef.augmentString()."]})', 'kind': 'markdown'}
```
Is it possible to send an HTML instead with `<a href='the_command'>` ?

Also I am not sure how the client are supposed to implement `goto` with just a string as argument. Shouldn't the server send an LPS location instead ?

tgodzik:
> @tgodzik I fixed the sublime format and tests. The links in analyse stacktrace response work now natively.
> 
> But there is still an issue on the synthetics links, as you can see bellow this isn't valid markdown
> 
> ```
> '\n**Synthetics**:\n\n[augmentString](subl:lsp_metals_goto {"parameters": ["scala/Predef.augmentString()."]})', 'kind': 'markdown'}
> ```
> 
> Is it possible to send an HTML instead with `<a href='the_command'>` ?

I think that would be totally fine, we could send <a ..> for sublime and maybe that will also work for VS Code? Not sure, but I can check later. Otherwise, we can send a different link for Sublime and VS Code.

> Also I am not sure how the client are supposed to implement `goto` with just a string as argument. Shouldn't the server send an LPS location instead ?

This is `goto` that has only symbol string and searches for that, `metals.goto` command will work with a single string parameter. So essentially it's kind of a lazy go to definition.

ayoub-benali:
> Otherwise, we can send a different link for Sublime and VS Code.

No need to. I added escaping for double quotes and it fixed the problem. What other ones you think we should escape ? I would prefer if we used a library for this rather than manually escaping things. 

> So essentially it's kind of a lazy go to definition.

You mean the client is supposed to forward the parameter to the server via execute command ?

metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
```diff
@@ -207,13 +207,18 @@ object CommandHTMLFormat {
         commandId: String,
         arguments: List[String]
     ): String = {
+      def escape(args: String): String = {
+        // The lib used to convert markdown to html in sublime doesn't properly
+        // recognize URL encoding so we have to use hexadecimal html encoding
+        args.replaceAll("\"", "&#x22;")
+      }
```
kpodsiad, line 214: >No need to. I added escaping for double quotes and it fixed the problem. What other ones you think we should escape ? I would prefer if we used a library for this rather than manually escaping things.

If what you did @ayoub-benali  works then it should be good enough for our needs. I found this [answer on SO](https://stackoverflow.com/a/40942504) about escaping, so maybe we could escape `& < > " '` ? Followig should work
```scala
val toEscape = Set('"', '<', '>', '&', '\'')
def escape(args: String): String =
  args.flatMap {
    case char if toEscape.contains(char) => s"&#x${char.toHexString}"
    case char => char.toString
  }
```

tgodzik:
> You mean the client is supposed to forward the parameter to the server via execute command ?

Yes, exactly.

ayoub-benali:
@tgodzik @kpodsiad I would need help for the failing test suite. I believe it is because this line 
https://github.com/scalameta/metals/blob/main/tests/unit/src/main/scala/tests/TestingServer.scala#L442
the value `commandInHtmlFormat` isn't properly extracted, I see `scala.meta.internal.metals.CommandHTMLFormat$VSCode$@200d81ac` in the logs which means in the lines bellow it isn't correctly converted to json and becomes an empty object 
`{"decorationProvider":true,"inlineDecorationProvider":true,"commandInHtmlFormat":{}}`

I am not sure what is going on there, does the `CommandHTMLFormat` trait should implement a method so that it value is extracted correctly ?

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -439,8 +439,9 @@ final class TestingServer(
           field.setAccessible(true)
           field.getName -> field.get(initOptions)
         }
-        .collect { case (key, Some(value)) =>
-          key -> value
+        .collect {
+          case (key, Some(value: Boolean)) => key -> value
```
ayoub-benali, line 443: So far in `InitializationOptions` only Boolean fields shouldn't be converted to String.

metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
```diff
@@ -141,6 +145,9 @@ object InitializationOptions {
       isHttpEnabled = jsonObj.getBooleanOption("isHttpEnabled"),
       isCommandInHtmlSupported =
         jsonObj.getBooleanOption("isCommandInHtmlSupported"),
+      commandInHtmlFormat = jsonObj
+        .getStringOption("commandInHtmlFormat")
+        .flatMap(CommandHTMLFormat.fromString),
```
kpodsiad, line 150: We should deprecate `isCommandInHtmlSupported` in favor of `commandInHtmlFormat`. 
Since you removed `isCommandInHtmlSupported` from `if`'s I wonder what we should do in other editors. For sure, vscode has to be reflect those changes and set `isCommandInHtmlSupported` instead of `isCommandInHtmlSupported`. 
@ckipp01 what about nvim?

If `isCommandInHtmlSupported` was set in vscode then maybe we should remove it right away? I can do this on vscode side, but I don't know if it won't break some other editor.

kpodsiad (commented):
LGTM. However, I have a question that requires others' opinion.

tests/unit/src/test/scala/tests/clients/CommandSuite.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import java.net.URLEncoder
+
+import scala.meta.internal.metals.CommandHTMLFormat
+import scala.meta.internal.metals.ServerCommands
+
+import org.eclipse.{lsp4j => l}
+import tests.BaseSuite
+
+class ScalaVersionsSuite extends BaseSuite {
```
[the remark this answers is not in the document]
tgodzik: Och, that might be my bad :sweat:

metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
ayoub-benali: `isCommandInHtmlSupported`'s value in initialization options completely ignored. Setting it has no effect.
I think we should simply removed it.

tgodzik (approved):
LGTM! I think it's fine to remove the old option as it was only used in VS Code as far as I can find. I will update that myself.

PR #3380, opened 2021-12-17
Display build target info

Arthurm1:
This is an alternative to https://github.com/scalameta/metals/pull/3077 which I think I'll ditch in favour of this.

Instead of webview it uses `metalsdecode:` and virtual docs.  The benefits are that it should be able to be supported in the same way across different clients (vscode, vim, sublime).  It also means that multiple build target views can be open at once.  The downside is that it loses the navigation to other build targets and to jars.

You can open a uri with the format `metalsDecode:file:///workspacePath/buildTargetName.metals-buildtarget` in the same way as other `metalsDecode` types work and Metals will hand you back the text describing that build target.
You can fire off `target-info-display` command and Metals will use the QuickPick functionality to offer a list of target names.

There's a textmate grammar file for highlighting the build target file.

@ckipp01 @ayoub-benali I've kind of assumed that since you have the `metalsDecode` functionality in Vim/Sublime that this will be easy to add?

This is rebased off the Java PR https://github.com/scalameta/metals/pull/2520 as it sets up all the javacoptions so there's no point reviewing til that's merged.

https://user-images.githubusercontent.com/5612295/146558462-82e64c9f-a8c2-4b1a-bc27-dcb23d668119.mp4


GavinRay97:
This is really cool + useful, thank you 👀

ayoub-benali:
Cool feature :tada: and yes I shouldn't be too difficult to add

ckipp01:
Same from me on the `nvim-metals` side. This should be easy to add 👍🏼

metals/src/main/scala/scala/meta/internal/tvp/ClasspathTreeView.scala
```diff
@@ -33,7 +34,8 @@ class ClasspathTreeView[Value, Key](
       viewId,
       rootUri,
       title + s" (${toplevels().size})",
-      collapseState = MetalsTreeItemCollapseState.collapsed
+      collapseState = MetalsTreeItemCollapseState.collapsed,
+      contextValue = "root"
```
tgodzik, line 38: Why do we need the new `contextValue` ? This should be documented in tree view protocol https://scalameta.org/metals/docs/integrations/tree-view-protocol if we add it here.

tgodzik (changes requested):
Great work! Finally I will not have to go manually into `.bloop` files :sweat_smile: 

And probably this would be much easier for sbt server than before. Left some questions and we should probably rebase.

ckipp01:
> @ckipp01 @ayoub-benali I've kind of assumed that since you have the metalsDecode functionality in Vim/Sublime that this will be easy to add?

Yup, the handling of just the raw text would be fine. Although I honestly think this would be a killer feature to return to the editor in a structured manner instead of just text. I know for myself this was actually what I proposed in https://github.com/scalameta/metals/discussions/2479 but kind of got shot down. If the client can have access to this info, it unlocks a _ton_ of stuff they can do with it. So the text is great, but I'd honestly love this as JSON.

metals/src/main/scala/scala/meta/internal/tvp/ClasspathTreeView.scala
Arthurm1: It's so VSCode can show a context menu only on projects, not libraries, in the treeview.  If you're ok with that then I'll add some docs

tgodzik: Would we be able to do it in VS Code only? I would rather avoid changing it if it's not needed for any other extensions. @ckipp01 what do you think?

Or maybe, we could make it general and add `command` filed instead something like:
```
"comand" :  { "id" : "show-target", args: [] }
```
what do you think?

metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala
```diff
@@ -0,0 +1,306 @@
...
+  }
+  private case class CommonTargetDetail(
+      id: BuildTargetIdentifier,
+      displayName: String,
+      baseDirectory: String,
+      tags: List[String],
+      languageIds: List[String],
+      dependencies: List[String],
+      dependentTargets: List[String],
+      capabilities: List[(String, Boolean)],
```
ckipp01, line 210: It'd be nicer if this could be a `List[BuildTargetCapability]`. Then we could also avoid the `._2` shenanigans up above.

metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala
```diff
@@ -0,0 +1,306 @@
...
+      info: ListBuffer[String],
+      classPath: List[Path]
+  ): Unit = {
+    def convertToName(path: Path): Path = {
+      if (path.toFile.isFile)
+        path.getFileName()
+      else
+        path
+    }
+    if (classPath.nonEmpty) {
```
ckipp01, line 165: I'm recommend breaking this up a bit. We have 3 nested ifs in here that can be tricky to follow. Breaking those out to functions might make this easier to read.

metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala
ckipp01: Actually thinking about this, there seems to be a lot of redundancy in here. Are all of these classes actually needed. For example we could also just use `BuildTargetCapabilities` from BSP for this couldn't we? Same with almost all of this stuff in here. This is very close to the actual `BuildTarget` interface.

metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala
```diff
@@ -0,0 +1,306 @@
...
+  }
+  private def extractCapabilities(
+      target: CommonTarget
+  ): List[(String, Boolean)] = {
+    var capabilities = List.empty[(String, Boolean)]
+    capabilities ::= ("Debug", target.capabilities.getCanDebug)
+    capabilities ::= ("Test", target.capabilities.getCanTest)
+    capabilities ::= ("Run", target.capabilities.getCanRun)
+    capabilities ::= ("Compile", target.capabilities.getCanCompile)
+    capabilities
```
ckipp01, line 264: Lets avoid the `var` here. We can just create a list and add these when we create the list like `one :: two :: three :: nil` and just return that. Although again as I stated above, we should avoid a tuple here, and even avoid the extra class if we can.

ckipp01 (commented):
I think we have a lot of duplication in here with already existing stuff that things like `BuildTarget` already gives is. Is it possible to maybe slim down all the extra classes that are added here and to instead use the existing ones?

metals/src/main/scala/scala/meta/internal/tvp/ClasspathTreeView.scala
tgodzik: I think that is a cool option to have it under right click, just wondering if we can make the API change less VS Code reliant. I am also worried that with this change we will need to update all the clients.

tgodzik: Actually, let me take a look if we can do it without the added parameter

ckipp01: If at all possible I think we should avoid baking in VS Code specific stuff to the actual protocol itself. The more we do that the harder it becomes to work around for all other clients. I don't really understand the full context behind this to be honest. I don't know what a context menu is.

Arthurm1: ContextValue is gone.  The context menu allows you to right-click on a project in the treeview and get a menu up to allow you to call Display the Build Target info directly on the project/target, instead of running the Display Build Target command and getting a list of project/targets up.  Just a different way of getting to view project info

metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala
```diff
@@ -0,0 +1,306 @@
...
+      .headOption
+      .flatMap(buildTargetDetail)
+      .getOrElse(s"Build target $targetName not found")
+  }
+
+  private def buildTargetDetail(
+      targetId: BuildTargetIdentifier
+  ): Option[String] = {
+    extractTargetDetail(targetId).map(detail => {
+      val info = ListBuffer[String]()
```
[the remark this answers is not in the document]
Arthurm1: With StringBuilder I think I'd need to add newlines everywhere

Arthurm1:
> > @ckipp01 @ayoub-benali I've kind of assumed that since you have the metalsDecode functionality in Vim/Sublime that this will be easy to add?
> 
> Yup, the handling of just the raw text would be fine. Although I honestly think this would be a killer feature to return to the editor in a structured manner instead of just text. I know for myself this was actually what I proposed in #2479 but kind of got shot down. If the client can have access to this info, it unlocks a _ton_ of stuff they can do with it. So the text is great, but I'd honestly love this as JSON.

From what I can see you want the client to be able to interact with the build server e.g. query targets or run a mainclass?

Instead of creating multiple commands for whatever the client wants from the underlying build server, how about creating a generic `pass this command to BSP` command and get Metals to send back the BSP json results.  Then the client can call whatever BSP stuff it wants without having to special case each requirement in Metals.

Arthurm1:
@tgodzik I don't understand the ScalaFix error here.  Can you help?

tgodzik:
> @tgodzik I don't understand the ScalaFix error here. Can you help?

It doesn't seem related. I will rerun.

Arthurm1:
@tgodzik `InsertInferredTypeLspSuite.match-bind` fails sometimes but has nothing to do with this.

It seems to fail because in `inferTypeTitle` a different match occurs...

Here's the code snippet...
```
    pattern.parent
          .collect {
            case Pat.Typed(_, _) => None
            case Pat.Bind(lhs, _) if lhs != pattern => Some(insertTypeToPattern)
            case _: Pat.Bind => None
            case vl: Defn.Val if vl.decltpe.isEmpty => Some(insertType)
            case vr: Defn.Var if vr.decltpe.isEmpty => Some(insertType)
            case _: Pat | _: Enumerator => Some(insertTypeToPattern)
          }
          .getOrElse(None)
```

Sometimes the test enters...
`case Pat.Bind(lhs, _) if lhs != pattern => Some(insertTypeToPattern)`
Sometimes it enters...
`case _: Pat.Bind => None`

One passes, the other fails.  I don't know why this randomly happens as I have no idea what this code is supposed to do.  It's fairly easy to reproduce the failure when you run the whole 
`InsertInferredTypeLspSuite` suite but if you just run `InsertInferredTypeLspSuite-match-bind` it always passes.

tgodzik:
I will take a look!

Arthurm1:
@tgodzik I think this is ready to merge.

tgodzik (approved):
LGTM!

PR #3382, opened 2021-12-17
Move to Scala 3 published artifacts where possible

ckipp01:
It looks like both the scala-java8-compat and geny are both published for Scala 3
so we can just use those instead of utilizing for3Use2_13.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -425,24 +425,21 @@ val mtagsSettings = List(
   ),
   // @note needed to deal with issues with dottyDoc
   Compile / doc / sources := Seq.empty,
-  libraryDependencies +=
+  libraryDependencies ++= Seq(
+    "com.lihaoyi" %% "geny" % V.genyVersion,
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM!

ckipp01:
Maybe this wasn't the right move. I see now in https://github.com/scalameta/metals/pull/3384 that we have both 2.13 and 3 in mtags 3 now because mtags3 is bringing in 3 and scalameta is bringing in 2.13.... however I don't get how this passed, but when I bump the sbt version it doesn't.

tgodzik:
> Maybe this wasn't the right move. I see now in #3384 that we have both 2.13 and 3 in mtags 3 now because mtags3 is bringing in 3 and scalameta is bringing in 2.13.... however I don't get how this passed, but when I bump the sbt version it doesn't. thinking

We could exclude geny dependency from scalameta, which should work alright.

PR #3384, opened 2021-12-17
Bump to sbt 1.6.0-RC2

ckipp01:


tests/slow/src/test/scala/tests/sbt/SbtServerSuite.scala
```diff
@@ -64,9 +64,8 @@ class SbtServerSuite
   test("generate") {
     def sbtLaunchJar = workspace.resolve(".bsp/sbt-launch.jar")
     def sbtBspConfig = workspace.resolve(".bsp/sbt.json")
-    def isBspConfigValid = sbtBspConfig.readText.contains(
-      s"--sbt-launch-jar=${sbtLaunchJar.toString()}"
-    )
```
ckipp01, line 69: So I'm not 100% sure what exactly we were meant to be testing here or if it was just a sanity check to ensure it has the launch jar listed @adpi2?

I just switching this around a bit to only check. to see if they launch jar is in there, and no longer look for the `--sbt-launch-jar` since that now looks to be `-Dsbt.script=` in 1.6.x and no longer actually points to the launch jar in that location.

build.sbt
```diff
@@ -445,6 +445,14 @@ val mtagsSettings = List(
...
         .exclude("org.scala-lang", "scala-reflect")
         .exclude("org.scala-lang", "scala-compiler")
+        .exclude(
+          "com.lihaoyi",
+          "geny_2.13"
+        ) // avoid 2.13 and 3 on the classpath since we rely on it directly
+        .exclude(
+          "com.lihaoyi",
+          "sourcecode_2.13"
+        ) // avoid 2.13 and 3 on the classpath since it comes in via pprint
```
ckipp01, line 455: These seem to causing issues with 1.6.x but not 1.5.x which is interesting. Either way in 1.6.x we get errors about multiple binary version on the classpath since scalameta is bringing in the 2.13 versions whereas we also directly rely on the 3 artifacts in here as well.

tgodzik (approved):
LGTM! Thanks for investigating and figuring this one out!

PR #3385, opened 2021-12-18
[UNTESTED] Fix Gradle detection, multimodule proj

GavinRay97:
I think something like this should be right.

The only guaranteed file to exist in a Gradle project I think is `settings.gradle(.kts)` in the root. Subprojects can be nested arbitrarily + named anything:

https://docs.gradle.org/current/userguide/declaring_dependencies_between_subprojects.html#sec:project_jar_dependencies

Currently, Metals fails to detect Gradle projects like the ones created by `gradle init -> "Application (Scala)"` because the directory layout is:

```
settings.gradle(.kts)
/app
  build.gradle(.kts)
  /src
```

image

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
@@ -63,10 +63,11 @@ final class BuildTools(
     }
   }
   def isMill: Boolean = workspace.resolve("build.sc").isFile
-  def isGradle: Boolean =
+  def isGradle: Boolean = {
     val defaultGradlePaths = List("settings.gradle", "settings.gradle.kts",
-      "build.gradle", "build.gradle.kts", "./app/build.gradle", "./app/build.gradle.kts")
+                                  "build.gradle", "build.gradle.kts")
     defaultGradlePaths.exists(workspace.resolve(_).isFile)
+  }
   def isMaven: Boolean = workspace.resolve("pom.xml").isFile
   def isPants: Boolean = workspace.resolve("pants.ini").isFile
   def isBazel: Boolean = workspace.resolve("WORKSPACE").isFile
--- a/metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
@@ -64,8 +64,12 @@ final class BuildTools(
   }
   def isMill: Boolean = workspace.resolve("build.sc").isFile
   def isGradle: Boolean = {
-    val defaultGradlePaths = List("settings.gradle", "settings.gradle.kts",
-                                  "build.gradle", "build.gradle.kts")
+    val defaultGradlePaths = List(
+      "settings.gradle",
+      "settings.gradle.kts",
+      "build.gradle",
+      "build.gradle.kts"
+    )
     defaultGradlePaths.exists(workspace.resolve(_).isFile)
   }
   def isMaven: Boolean = workspace.resolve("pom.xml").isFile
--- a/metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
@@ -64,9 +64,9 @@ final class BuildTools(
   }
   def isMill: Boolean = workspace.resolve("build.sc").isFile
   def isGradle: Boolean =
-    workspace.resolve("build.gradle").isFile || workspace
-      .resolve("build.gradle.kts")
-      .isFile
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
```diff
@@ -64,9 +64,9 @@ final class BuildTools(
   }
   def isMill: Boolean = workspace.resolve("build.sc").isFile
   def isGradle: Boolean =
-    workspace.resolve("build.gradle").isFile || workspace
-      .resolve("build.gradle.kts")
-      .isFile
+    val defaultGradlePaths = List("settings.gradle", "settings.gradle.kts",
```
ckipp01, line 67: We'll have to wrap this all in `{}`. However a couple thoughts on this. I think we should for sure add the extra couple root files, but I'm curious if the `./app/build.gradle` and `./app/build.gradle.kts` is necessary. In those situations will there be no gradle file at all in the root? If not, then there is a chance the editor will not be sending in a workspace root to match this. For example in some of the clients we have patterns we look for to detect a workspace root.

https://github.com/scalameta/nvim-metals/blob/main/lua/metals/config.lua#L253

So in the above scenario if there was only a `./app/build.gradle` then the client would send in `./build.gradle` as the root, meaning the `./app/build.gradle` would never be valid _unless_ there was also a build file at the root, and then that's be enough anyways. So a couple things.

1. Some of the clients will need to update their patterns as well probably. I'll do this in `nvim-metals`.
2. I don't think the `./app/build.*` is necessary and actually won't work. wdyt?

ckipp01 (changes requested):
Thanks a lot for the pr @GavinRay97! I've left a couple thoughts. Let me know what you think.

metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala
GavinRay97: > meaning the `./app/build.gradle` would never be valid unless there was also a build file at the root, and then that's be enough anyways

I think you're right here

I tried to Google around before sending this small patch and wasn't sure if there was a hard-set Gradle layout for multi-module projects

Just looked again -- I can't find a single example in the Gradle docs or in online comments that doesn't have a `settings.gradle(.kts)` at the project root.

So we should be good with just:

```scala
List("settings.gradle", "settings.gradle.kts", "build.gradle", "build.gradle.kts")
```

GavinRay97: Fixed in https://github.com/scalameta/metals/pull/3385/commits/64061e52b834008df9f4326e0f299f50818eb4fa 👍

kpodsiad:
@GavinRay97 run `./bin/scalafmt` in order to fix failed formatting job and PR should be good to merge ;)

GavinRay97:
> @GavinRay97 run `./bin/scalafmt` in order to fix failed formatting job and PR should be good to merge ;)

For some reason, after running `./bin/scalafmt`, it shows every file as being modified in `git status`, despite having `autoclrf = true` in my global Git config 🤔 

```sh
user@MSI:~/projects/metals$ git status
On branch patch-1
Your branch is up to date with 'origin/patch-1'.

user@MSI:~/projects/metals$ ./bin/scalafmt
Reformatting...
  100.0% [##########] 757 source files formatted

user@MSI:~/projects/metals$ git status
On branch patch-1
Your branch is up to date with 'origin/patch-1'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   bin/merged_prs.sc
        modified:   build.sbt
        modified:   metals-bench/src/main/scala/bench/ClasspathFuzzBench.scala
        modified:   metals-bench/src/main/scala/bench/ClasspathIndexingBench.scala
        modified:   metals-bench/src/main/scala/bench/ClasspathOnlySymbolSearch.scala
        modified:   metals-bench/src/main/scala/bench/ClasspathSymbolsBench.scala
        modified:   metals-bench/src/main/scala/bench/CompletionBench.scala

user@MSI:~/projects/metals$ git diff metals-bench/src/main/scala/bench/ClasspathFuzzBench.scala
warning: LF will be replaced by CRLF in metals-bench/src/main/scala/bench/ClasspathFuzzBench.scala.
The file will have its original line endings in your working directory
```

Any tips here?

GavinRay97:
I just pushed the single `metals/src/main/scala/scala/meta/internal/builds/BuildTools.scala` file, I guess the `modified`-on-everything issue can wait for some other time haha.

tgodzik (commented):
LGTM!

tgodzik (approved):

PR #3392, opened 2021-12-21
Use quick pick instead of message requests for debug discovery

tgodzik:
Previously, we would use message request when asking for user input about which main class to run. Now, we instead use quick pick to accomodate a larger number of main classes possible. This is also a much nicer UI in VS Code as there is no need to select anything with a mouse. Quick pick still falls back to messageRequests if the client doesn't support it.

https://user-images.githubusercontent.com/3807253/146952326-d398138c-1fda-401f-93d4-917681f5a37a.mp4


ckipp01 (approved):
Looks good, LGTM!

PR #3393, opened 2021-12-21
[Mtags releases] Send the latest semantidcb-plugin version to BSP

dos65:
This should be a final part #3281.
To avoid Metals release when new Scala2 version appears, we also need
to use the latest semanticdb-plugin version to correctly setup BSP
connection.

So, in case of new Scala2 version the release process will be:
- release scalameta
- release mtags for existing Metals version

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspExtra.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspExtra.scala
@@ -1,0 +1,120 @@
+package scala.meta.internal.bsp
+
+import scala.jdk.CollectionConverters._
+
+import scala.meta.internal.metals.BuildInfo
+import scala.meta.internal.semver.SemVer
+
+/**
+ * When Metals establish bsp-connection it sends
+ *  the required semanticdb version and supportedScalaVersions (only scala2) to bsp-server
+ *  to prepare compiler plugin setup on its side.
+ *
+ * To have the actual state of this metadata use `BspExtra.discoverNewReleases`.
+ * It checks if there were scalameta/semanticdb releases that support newer Scala2 versions
+ *  that were published after the Metals release.
+ */
+final case class BspExtra(
+    semanticdbVersion: String,
+    supportedScalaVersions: java.util.List[String]
+)
+
+object BspExtra {
+
+  def apply(
+      semanticdbVersion: String,
+      supportedScalaVersions: Seq[String]
+  ): BspExtra =
+    BspExtra(semanticdbVersion, supportedScalaVersions.asJava)
+
+  val knownAtReleaseMoment: BspExtra =
+    BspExtra(
+      BuildInfo.semanticdbVersion,
+      BuildInfo.supportedScala2Versions
+    )
+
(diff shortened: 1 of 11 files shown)


metals/src/main/scala/scala/meta/internal/bsp/BspExtra.scala
```diff
@@ -0,0 +1,120 @@
...
+        .complete()
+        .getCompletions()
+        .asScala
+
+      completions
+        .filter(_.startsWith(allowedSemanticdbVersionPrefix))
+        .map(SemVer.Version.fromString)
+        .filter(_.isStable)
+        .sortWith(_ > _)
+        .headOption
```
kpodsiad, line 61: Shouldn't it be wrapped in `Future`? It's blocking, it's doing an API call. Maybe it's better to explicitly wrap it into the future rather than expecting call site to do it?

kpodsiad (approved):
lgtm

metals/src/main/scala/scala/meta/internal/bsp/BspExtra.scala
dos65: > It's blocking

I know :smiling_face_with_tear:
Just decided to leave it as it is because it's how we use coursier api in other places. 

Probably, it's a bit late to care blocking stuff in Metals. Future in Metals codebase appears only when there is a real need for it. And I think that's ok.  In case of LSP server,  it doesn't have to serve thousands requests per seconds so we can allow us writing things ignoring these bloking/async matters

tgodzik:
I think this approach might be too difficult to maintain, we also use semanticdb in the presentation compiler and in sbt plugin, so we would also need to modify it there.

Instead I think we could:
1. Check out last release tag.
2. Bump semanticdb version.
3. Push as a new tag to release for the last tag.

This wouldn't be that difficult to do and I don't think there is a need to replace a simple manual step with a complex code.

What do you think?

dos65:
@tgodzik 
>  we also use semanticdb in the presentation compiler and in sbt plugin, so we would also need to modify it there
oops. I missed that but it shouldn't be hard to do the same thing.

I agree that yours approach is much more robust. On the other hand it still be a full release -release metals, release metals-vscode, publish release notes. I think the cost of these changes is less than amount of work that will be required for next N releases. In addition it won't be clear for users why some new Scala versions works without release and some doesn't.

tgodzik:
Wouldn't it be easier to introduce the same mechanism for releasing semanticdb for new versions? So that we can just release it for the last version? I can try to do it manually next time, but can introduce a workflow there, which would mean a lot more tools can skip upgrading scalameta version.

We could also extract it to a plugin or add to sbt-ci-release?

dos65:
> Wouldn't it be easier to introduce the same mechanism for releasing semanticdb for new versions? So that we can just release it for the last version? I can try to do it manually next time, but can introduce a workflow there, which would mean a lot more tools can skip upgrading scalameta version.

:thinking: oh, that's brilliant! I like this idea.

I'll look what we need to do in scalameta build for that.

PR #3394, opened 2021-12-22
[Scalafmt] Fixes for automatic dialect rewrite

dos65:
I've noticed that on snapshot version the wrong suggestion
about improper `.scalafmt.conf` appears in Metals repo.

Changes:
- support more dialects. Previsously any Scala2 target had `scala213` formatting dialect
- ignore sbt targets in rewrite `.scalafmt.conf` suggestions

### DIFF
--- a/tests/unit/src/test/scala/tests/FormattingLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/FormattingLspSuite.scala
@@ -333,7 +333,7 @@ class FormattingLspSuite extends BaseLspSuite("formatting") {
           server.textContents(".scalafmt.conf"),
           s"""|version = "${V.scalafmtVersion}"
               |maxColumn=40
-              |runner.dialect = scala213
+              |runner.dialect = scala212
               |""".stripMargin
         )
       }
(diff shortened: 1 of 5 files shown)


tests/unit/src/test/scala/tests/FormattingLspSuite.scala
```diff
@@ -415,6 +415,50 @@ class FormattingLspSuite extends BaseLspSuite("formatting") {
     } yield ()
   }
 
+  test("rewrite-dialect-ignore-sbt") {
+    cleanWorkspace()
+    client.showMessageRequestHandler = { params =>
+      if (
+        params.getMessage().startsWith(Messages.UpdateScalafmtConf.beginning)
```
tgodzik, line 422: Why do we need this if the `.scalafmt.conf` file doesn't change? Is it in order to make sure that the file will not change?

Could we instead throw an exception here if there are any params at all? No need for the new `beginning` value then.

tgodzik (approved):
Looks good, just one comment about the tests.

tests/unit/src/test/scala/tests/FormattingLspSuite.scala
dos65: Changed it a bit to trigger rewrite. 
Now it checks that applied will be `scala212` instead of `scala212source3` that might come from sbt project.

tgodzik: Could we not have just:
```
  params.getActions.asScala
          .find(_ == Messages.UpdateScalafmtConf.letUpdate)
```
without the `if` condition? We could get rid of `Messages.UpdateScalafmtConf.beginning`

tgodzik (approved):
LGTM!

PR #3395, opened 2021-12-22
File watcher - create semanticDB directories

kpodsiad:
The solution proposed by @uncle-betty + more logging.

@yaoqiang-se @gombosg could you try it and let us know if it works?

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1318,11 +1318,17 @@ class MetalsLanguageServer(
     val isScalaOrJava = path.isScalaOrJava
     event.eventType match {
       case EventType.Create =>
-        scribe.info(s"Created file: ${event.path}")
+        scribe.debug(
+          s"[File watcher] Detected creation of the file: ${event.path}"
+        )
       case EventType.Modify =>
-        scribe.info(s"Modified file: ${event.path}")
+        scribe.debug(
+          s"[File watcher] Detected modification of the file: ${event.path}"
+        )
       case EventType.Delete =>
-        scribe.info(s"Deleted file: ${event.path}")
+        scribe.debug(
+          s"[File watcher] Detected deletion of the file: ${event.path}"
+        )
     }
 
     if (isScalaOrJava && event.eventType == EventType.Delete) {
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1313,6 +1313,15 @@ class MetalsLanguageServer(
   ): CompletableFuture[Unit] = {
     val path = AbsolutePath(event.path)
     val isScalaOrJava = path.isScalaOrJava
+    event.eventType match {
+      case EventType.Create =>
+        scribe.info(s"Created file: ${event.path}")
```
tgodzik, line 1318: Could we have it under `debug` ? Or is it only for testing now?

kpodsiad: I was thinking about leaving it as it is. It's not that much I think and in case of any problems users don't have to enable any flags

tgodzik: I would change it to something like `[File watcher] Detected file creation` to be more exact. Just printing that the file was created is not that useful for the user.

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
```diff
@@ -146,9 +146,9 @@ object FileWatcher {
       // and run some tests with `-Dswoval.log.level=debug` to see which files are registered
       // and which file events are received.
       val directoriesToCreate =
-        filesToWatch.sourceDirectories ++ filesToWatch.sourceFiles.map(
-          _.getParent()
-        )
+        filesToWatch.sourceDirectories ++
```
tgodzik, line 149: This seems to break semanticdb on my machine altogether :thinking:

kpodsiad: That's a bummer :/

gombosg:
Hey! I tried it... it unfortunately doesn't work (still "no build target" error) and now VSCode compiles the project each time I change to a new file. 🤔

tgodzik:
We could probably close this one, no?

kpodsiad:
Superseded by https://github.com/scalameta/metals/pull/3405

PR #3400, opened 2021-12-23
Add run/debug code lense for Java main classes

tgodzik:


metals/src/main/scala/scala/meta/internal/metals/codelenses/RunTestCodeLens.scala
```diff
@@ -57,25 +63,89 @@ final class RunTestCodeLens(
...
+
+  private def javaLenses(
+      occurence: SymbolOccurrence,
+      textDocument: TextDocument,
+      target: BuildTargetIdentifier
+  ): Seq[l.Command] = {
+
+    if (occurence.symbol.endsWith("#main().")) {
+      textDocument.symbols.find(_.symbol == occurence.symbol).toSeq.flatMap {
+        a =>
```
kpodsiad, line 123: could it be formated like for example:
```suggestion
      textDocument.symbols.find(_.symbol == occurence.symbol).toSeq
      .flatMap { a =>
```
I mean, I just want `a =>` in the same line as the `flatMap` is, otherwise it looks strange to me.

metals/src/main/scala/scala/meta/internal/metals/codelenses/RunTestCodeLens.scala
```diff
@@ -57,25 +63,89 @@ final class RunTestCodeLens(
...
         val classes = buildTargetClasses.classesOf(buildTargetId)
-        codeLenses(textDocument, buildTargetId, classes, distance)
+        codeLenses(textDocument, buildTargetId, classes, distance, path)
       }
 
       lenses.getOrElse(Seq.empty)
     }
   }
 
+  private def isMainMethod(signature: Signature, textDocument: TextDocument) = {
```
kpodsiad, line 79: My proposition
```scala
  /**
    * Java main method: public static void main(String[] args)
    */
  private def isMainMethod(signature: Signature, textDocument: TextDocument) = {
    signature match {
      case MethodSignature(_, Seq(Scope(Seq(symlink), _)), returnType) =>
        def isVoid = returnType match {
          case TypeRef(_, symbol, _) => symbol == "scala/Unit#"
          case _ => false
        }

        def isStringArray = {
          textDocument.symbols.find(_.symbol == symlink) match {
            case Some(info) =>
              info.signature match {
                case ValueSignature(
                      TypeRef(
                        _,
                        "scala/Array#",
                        Vector(TypeRef(_, "java/lang/String#", _))
                      )
                    ) =>
                  true
              }
            case None => false
          }
        }

        isVoid && isStringArray
      case _ => false
    }
  }
```

tests/unit/src/test/scala/tests/CodeLensLspSuite.scala
```diff
@@ -104,6 +104,42 @@ class CodeLensLspSuite extends BaseCodeLensLspSuite("codeLenses") {
       )
     } yield ()
   }
+  test("run-many-main-java") {
```
kpodsiad, line 107: `run-many`? Maybe just `main-java`?

kpodsiad (approved):

metals/src/main/scala/scala/meta/internal/metals/codelenses/RunTestCodeLens.scala
kpodsiad: I find `Seq(Scope(Seq(symlink), _))` less hacky than `parameterLists.size == 1 && parameterLists.head.symlinks.size == 1`

Arthurm1:
Running this locally I get code lenses on `main` and I can use the new VSCode testing functionality with JUnit4 (which may or may not have already worked)

image

Brilliant stuff!  Thanks both of you.

metals/src/main/scala/scala/meta/internal/metals/codelenses/RunTestCodeLens.scala
tgodzik: Changed :+1:

tests/unit/src/test/scala/tests/CodeLensLspSuite.scala
tgodzik: changed

kpodsiad (approved):

PR #3401, opened 2021-12-24
Detect missing src.zip in doctor

### LINKED ISSUE #3397: Doctor doesn't report missing src.zip file
**Describe the bug**

When the `src.zip` file is missing from the java home directory, `textDocument/definition` would yield no results for java standard library classes like `java.time.LocalDate`.

Running metals doctor commands shows you that everything is fine. 

**To Reproduce**
1. Remove or rename `src.zip` 
2. Start Metals
3. Try go to definition in `java.time.LocalDate` 
4. Notice that nothing happens

**Expected behavior**

Metals doctor should show a warning with a possible fix like `sudo apt-get install openjdk-8-source` or equivalent.

**Installation:**

- Operating system: Linux
- Editor: Sublime
- Metals version: 0.10.9+213-69c1200a-SNAPSHOT

**Search terms**

java home, goto definiton, doctor

ayoub-benali:
Fix #3397

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
@@ -35,14 +35,16 @@ final class Doctor(
     httpServer: () => Option[MetalsHttpServer],
     tables: Tables,
     clientConfig: ClientConfiguration,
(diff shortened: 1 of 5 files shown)


tgodzik (approved):
LGTM!

Issue #3379, opened 2021-11-23
Find References not working


gombosg:
**Describe the bug**

**To Reproduce** Steps to reproduce the behavior:

1. Import project
2. Try a "find references" feature
3. No references found anywhere.

**Expected behavior**

Find references is working!

**Screenshots**

Run doctor says it's all right:
![image](https://user-images.githubusercontent.com/23086060/142980887-a413070f-a9b8-40b9-818e-0c3764d93038.png)
![image](https://user-images.githubusercontent.com/23086060/142981506-413cf14a-db43-481e-ba67-6d668048d696.png)

**Installation:**

- Operating system: Fedora 35
- VSCode version: 1.62.3 (bug is present with insiders, too)
- VSCode extension version: 1.10.15
- Metals version: 0.10.9

**Additional context**

[metals.log](https://github.com/scalameta/metals-vscode/files/7586189/metals.log)

...I can also send SemanticDB files if that helps.

Happens in VSCode main and insiders, tried re-importing, cleaning all data etc. it just won't work. I'd so much use Metals for work as it's blazing fast for code navigation, but this is kind of a deal breaker.
Other smaller projects are fine but I don't know what's the key to reproducing this issue in this project.

**Search terms**

find all references

dos65:
@tgodzik I think I've seen it at least two times but was unable to reproduce. After reloading window everything started working well.

kpodsiad:
> This is the culprit after git bisect: 2422494


@gombosg I'll add more logging points in changes coming from the commit you found and also an option that enables debug output. Then we should have more information about what's going on.

gombosg:
@kpodsiad  
Thanks! It's surely that commit because reverting it on top of `main` helps the issue for me.
You don't necessarily have to release a debugging patch, if that helps you can just add some debugging to a branch and I can compile that and get back with some logs.

kpodsiad:
@gombosg @yaoqiang-se I just merged https://github.com/scalameta/metals/pull/3388. Now 'no build target message' is more verbose and prints file for which presentation compiler isn't loaded. 
Could you use metals snapshot version published from that merge and send metals' output? Maybe it'll point us a direction what's not working.

kpodsiad:
Actually, you could add `-Dswoval.log.level=debug` to `Metals: Server Properties` and then watch `metals.log` using for example `tail -f -n 300 .metals/metals.log` (I don't know what will be the windows equivalent of this)

gombosg:
[...]

Issue #3398, opened 2021-12-23
Java support status placeholder


Arthurm1:
Supported features...
- [x] Shows Java-only build targets
- [x] Code navigation (Find References/Goto definition)
- [x] Formatting (file or selection)
- [x] New Java File
- [x] `super` code lenses
- [x] Document highlight
- [x] Error/warning - line number only
- [x] Run/Debug code lens (only main)
- [x] Find implementations
- [x] Folding ranges
- [x] Code navigation from source jar java files - depends on adding PC (PresentationCompiler)

Todo...
- [ ] Formatting on Type - depends on adding PC
- [ ] Error/Warning - exact range depends on https://github.com/scalacenter/bloop/pull/1477
- [ ] Code completion - depends on adding PC
- [ ] Code outline in VSCode
- [x] ~~Refactoring~~ Rename (@tgodzik I don't think we want to support more advanced refactorings than renames)
- [ ] Organize imports
- [ ] Hover (type at point)

ckipp01:
So, more of just a question, but in the TODO section, how many of these things do we actually want? Is the goal truly to cover them all? It's incredible that we now support java targets, but I'm also wondering at what point does Metals, the **Scala** language server venture too far into the Java land support? Realistically we'll never reach parity with something like IntelliJ for cross-JVM languages, but I also don't really think that's a goal. Adding all the features on this list is no small task, and also increases the scope of Metals. Is that what we want? I guess I just never assumed that something like formatting on type for Java sources is something that Metals would support or want to support.  Again, I don't mean this to take away from the fantastic achievement of the Java support we now have. https://github.com/scalameta/metals/pull/2520 is truly fantastic work! I'd just love to get a clearer idea of what the ultimate goal is with our Scala Java interop and Java support.

Arthurm1:
I'm pretty happy with the current set of functionality.  It makes working with combined Scala/Java codebases so much easier.

I think run/debug `main` and code folding should be simple to add and help ease of use and I will probably look to implement.

As for the others, I thought I'd fix whatever problems came up after release and then wait to see if anyone has a real need for anything else.  I doubt I'll be looking to add any of it anytime soon. (I don't actually need or want format-on-type)

[...]

PR #3404, opened 2021-12-26
Proposed workaround for #3379.

uncle-betty:
When registering paths, sort them by their longest existing prefix in descending order. See #3379 for more details.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
@@ -69,7 +69,7 @@ object FileWatcher {
   private case class FilesToWatch(
       sourceFiles: Set[Path],
       sourceDirectories: Set[Path],
-      semanticdDirectories: Set[Path]
+      semanticdbDirectories: Set[Path]
   )
 
   private def collectFilesToWatch(buildTargets: BuildTargets): FilesToWatch = {
@@ -124,7 +124,7 @@ object FileWatcher {
       // and also to hash only revelevant files when watching for changes
 
       val trie = PathTrie(
-        filesToWatch.sourceFiles ++ filesToWatch.sourceDirectories ++ filesToWatch.semanticdDirectories
+        filesToWatch.sourceFiles ++ filesToWatch.sourceDirectories ++ filesToWatch.semanticdbDirectories
       )
       val isWatched = trie.containsPrefixOf _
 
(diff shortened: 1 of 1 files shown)


gombosg:
Hey, thanks for this. No luck, this also yields me the same error, presentation compiler only etc. How does this fix the issue for you?
```
2021.12.26 11:18:58 INFO  no build target found for /home/gombosg/myproject/.metals/.tmp/Main-4611594563991175440.scala. Using presentation compiler with project's scala-library version: 2.13.4
```

gombosg:
Sorry! It's working! I just had to delete all `.metals` `.bloop` etc. directories and reimport the whole build. 
The question is, can we have a unit test or whatever to validate the theory that indeed this was the culprit and prevent regressions?
🎉

kpodsiad:
@gombosg `no build target found for ...` is not an issue, see https://github.com/scalameta/metals/issues/3398#issuecomment-1000512319.

gombosg:
@kpodsiad  see above. 🙂 I tested editing a file, it still works. What cases should be tested? Any logs that could be useful?

kpodsiad:
@gombosg Oh I see, we were writing at the same time :D

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
```diff
@@ -138,29 +138,27 @@ object FileWatcher {
...
+      // Workaround, see #3379 for details.
+      val sortedDirs = dirs.toSeq
+        .map(path => (path, prefLen(path)))
+        .sortBy({ case (_, len) => -len })
+        .map({ case (path, _) => path })
 
-      createdDirectories.toSeq.sortBy(_.toNIO).reverse.foreach { dir =>
-        if (dir.isEmptyDirectory) dir.delete()
-      }
+      sortedDirs.foreach(repo.register(_, Int.MaxValue))
```
kpodsiad, line 161: I think that the final shape of the solution will be 
- bump `file-tree-views` to 2.1.8
- remove directories creation
- and since in 2.1.8 we don't need sorting anymore also removing sorting
Seems ok?

kpodsiad (commented):
It looks very promising! I think we don't need any tests since there is already a test in swoval's repo (?)
@tgodzik @dos65 could you try this on your Linux machines and projects?

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
uncle-betty: I completely agree. See #3405, which upgrades `file-tree-views` and gets rid of directory creation.

Man, I really wasn't expecting upstream to move that fast!

uncle-betty:
I'm closing this in favor of #3405. Sorry for the back and forth. Upstream was just really fast in addressing this.

PR #3405, opened 2021-12-26
Upgrade file-tree-views to 2.1.8. See #3379.

uncle-betty:
Upstream was incredibly responsive and provided a fix in file-tree-views 2.1.8 in record time.

This supersedes the workaround in PR #3404 for issue #3379.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -262,7 +262,7 @@ commands ++= Seq(
         case _ =>
           val message =
             s"""|Skip release. Invalid tag name.
-                |It should be either: 
+                |It should be either:
                 | - "v$$num.$$num.$$num" - usual Metals release
                 | - "mtags_v$${existing-metals-release}_$${scala-version}" - mtags artifact release
                 |""".stripMargin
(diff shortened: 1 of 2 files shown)


uncle-betty:
@gombosg, it looks like upstream has made my workaround superfluous. We can now properly fix this by upgrading the `file-tree-views` library.

May I ask you to give this a spin one last time? It resolves the "find references" issue for me just like the workaround in #3404 did. Let's make sure that it also works for you.

gombosg:
It works great, I tested editing files, adding new files! tada

ckipp01 (approved):
@uncle-betty better thank you for your work in tracking this down, and also @eatkins thank you for addressing this so quickly!

I've tested this locally with no issues, so this is a 👍🏼 from me. I'll wait for one of the others to also approve and then I think we're good to merge. Thanks again for your work on this! 🚀

dos65 (approved):
Tested locally - works well.
@uncle-betty, @gombosg thx for intvestigation and fixes!

Issue #3402, opened 2021-12-24
Metals snapshot not starting on java 8


ayoub-benali:
**Describe the bug**

Metals doesn't start when using latest snapshot `0.10.9+226-e28384c3-SNAPSHOT` on Java 8. 

From the error bellow it seems to be related to https://github.com/scalameta/metals/pull/2520

```
 Started: Metals version 0.10.10-SNAPSHOT in workspace '/home/ayoub/workspace/testproject' for client Sublime Text LSP 1.14.0.
Exception in thread "pool-4-thread-2" java.lang.UnsupportedClassVersionError: org/eclipse/text/edits/MalformedTreeException has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:473)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	at scala.meta.internal.metals.MetalsLanguageServer.updateWorkspaceDirectory(MetalsLanguageServer.scala:509)
	at scala.meta.internal.metals.MetalsLanguageServer.$anonfun$initialize$2(MetalsLanguageServer.scala:799)
	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:659)
	at scala.util.Success.$anonfun$map$1(Try.scala:255)
	at scala.util.Success.map(Try.scala:213)
	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
	at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
	at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
[...]

PR #3406, opened 2021-12-27
Hardcode eclipse java formatter dependencies

Arthurm1:
In https://github.com/scalameta/metals/issues/3402 @ayoub-benali found that the snapshots weren't java 8 compatible

I was using `dependencyOverrides` to specify jdk8 versions of jars in `build.sbt` but they don't end up in the published `pom.xml`.

Now I've hardcoded all the dependencies and excluded transitive ones so the `pom.xml` has sections like...
```
        <dependency>
            <groupId>org.eclipse.platform</groupId>
            <artifactId>org.eclipse.osgi</artifactId>
            <version>3.15.0</version>
            <exclusions>
                <exclusion>
                    <groupId>*</groupId>
                    <artifactId>*</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

I'm not too knowledgeable about `pom.xml` but this works locally and makes logical sense.

I imagine `dependencyOverrides` worked for me with `publishLocal` before because the eclipse poms have ranges specified and if any valid version is present locally then it would be used.  And valid versions were all that were in my local cache as I'd run the build.sbt so they would have been downloaded then.


Arthurm1:
Errors all seem due to failing to install Adoptium jdk 11

dos65 (approved):
Checked it locally - works well! Thanks for the fix.

ayoub-benali:
Thanks for the fix @Arthurm1 :tada:

PR #3411, opened 2021-12-28
navigate to decompiled class if no source

Arthurm1:
Initial stab at https://github.com/scalameta/metals-feature-requests/issues/210

I've no idea whether this is done in the correct area of code but it works so 🤷.  If it is the correct area then I'll follow up with some tests.

One thing I've noticed is that `scala.meta.internal.mtags.SymbolDefinition` (and a number of other classes used for passing file locations around) use `AbsolutePath` to refer to file locations.  This is handy because `nio.Path` validates that the format is correct and there are lots of useful Path methods but it's not so handy when all that's really going on is URIs are being passed around.

Downsides of using AbsolutePath are that any `jar:file///somejar.jar!/someclass.java` style reference has to be opened in java `FileSystems#newFileSystem`, even if you're not accessing it, or it throws an exception.  Also - you can't use a virtual URI e.g. `metalsDecode:jar:file:///somejar.jar!/someclass.class.cfr` in AbsolutePath because it wraps `nio.Path` and I've not implemented a `metalsDecode` filesystem as I think that's over the top.

Is there a reason for using AbsolutePath rather than URI in all these places - or is it just historical?

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
@@ -1,5 +1,9 @@
 package scala.meta.internal.metals
 
+import java.net.URI
+import java.nio.file.FileSystemNotFoundException
+import java.nio.file.FileSystems
+import java.nio.file.Paths
 import java.util.Collections
 import java.{util => ju}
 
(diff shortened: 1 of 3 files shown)


tgodzik:
I think it was useful to have Absolute Path as it has more utility methods and also probably some validation. That might not be 100 percent needed in all places.

metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
```diff
@@ -310,19 +329,47 @@ class DestinationProvider(
...
+    val querySymbol: Symbol = Symbol(symbol)
+    val definitions = index.definitions(querySymbol).filter(_.path.exists)
+    val extraDefinitions = if (definitions.isEmpty) {
+      // search in all classpath, i.e. jars with no source
+      buildTargets
+        .inferBuildTarget(List(querySymbol))
+        .map(f => {
+          val uri = f.jar.toURI.toString()
+          val fullURI =
+            URI.create(s"jar:${uri}!/${symbol.stripSuffix("#")}.class")
```
tgodzik, line 341: This might not work for some more complex Scala symbols such as inner classes :thinking: 

In that case we might want to use ClasspathSymbols and find the class that contains the symbol.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -482,7 +491,18 @@ final class FileDecoderProvider(
...
-    val args = List("--analyseas", "CLASS", s"""${path.toNIO.toString}""")
+    val (parent, className) =
+      if (path.isJarFileSystem)
+        (workspace, path.toNIO.toString.stripPrefix("\\").stripPrefix("/"))
+      else (path.parent, path.toNIO.toString)
+    val extraClassPath = path.jarPath
+      .map(jarPath => List("--extraclasspath", jarPath.toString))
+      .getOrElse(List.empty)
+    val args = extraClassPath ::: List(
+      "--analyseas",
```
tgodzik, line 502: Typo?

tgodzik (commented):
Finally managed to take a look! I think it would be good to test it with some of the scalameta symbols that are generate from protobuf schema such as `SymbolInformation`

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
Arthurm1: No - although it does look it.  The options is meant to mean ANALYSE_AS.  Either `jar` or `class`

doofin:
Thanks for this nice feature! Does it also works for scala js?

Arthurm1:
@doofin I haven't used scala js. Can you give an example of what you mean?

tgodzik:
This is being done together with https://github.com/scalameta/metals/pull/3750, so closing

PR #3412, opened 2021-12-29
fix: ensure autoimports don't break scala-cli directives

### LINKED ISSUE #3399: auto-imports can break scala-cli scripts
**Describe the bug**

When using https://scala-cli.virtuslab.org/ if you have some code like

```
// using Scala 3.1.0
```

And then you trigger an auto-import, it will insert the import above the directive since it thinks it's just a comment, and then can lead to broken scripts.

**To Reproduce**

1. Copy the above
2. Auto import something
3. You now changed your scala version

**Expected behavior**

I'd expect for Metals to be able to know that `// using Scala 3.1.0` is a directive and not import above it.

ckipp01:
This just adds a check in the situation where a file has no existing import
and no package to look and see if there are `//` comments in the beginning
of the file and if so, skip over them before we add the import. The main reason
for this is to avoid braking directives in scala-cli.

We don't do a more advanced check for this for other types of comments mainly
because it'll be more difficult without more extensive parsing to know if a comment
is a header or a scaladoc for example. So for now we just do a more naive check
on `//`.

Closes #3399

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
@@ -1,5 +1,6 @@
 package scala.meta.internal.pc
 
+import scala.annotation.tailrec
 import scala.reflect.internal.FatalError
 
 import scala.meta.internal.mtags.MtagsEnrichments._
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
LGTM!

PR #3415, opened 2021-12-29
docs: add file ids to the docs for new file commands

ckipp01:
This pr adds the file ids and the label to the docs. I missed this in last PR where this changed from `worksheet` to `scala-worksheet` and this actually broke my implementation since I do have some commands that send in the filetype directly. Now these will be listed in the docs to be accessed without needing to dive into the code.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/newScalaFile/NewFileTypes.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/newScalaFile/NewFileTypes.scala
@@ -13,7 +13,7 @@ object NewFileTypes {
   case object ScalaFile extends NewFileType {
     override val id: String = "scala-file"
     override val syntax: Option[String] = None
-    override val label: String = "Scala file with automatically added package"
+    override val label: String = "Empty file"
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/newScalaFile/NewFileTypes.scala
```diff
@@ -13,7 +13,7 @@ object NewFileTypes {
   case object ScalaFile extends NewFileType {
     override val id: String = "scala-file"
     override val syntax: Option[String] = None
-    override val label: String = "Scala file with automatically added package"
+    override val label: String = "Empty file"
```
ckipp01, line 16: So this is more of a wdyt?

I personally fine the other one worded a bit oddly. When I first came across it I actually paused for a bit to make sure I understood what it meant. I honestly think we can just simplify this to `Empty file`. I think it gets the point across and does what most people would expect, just adding in the package. wdyt?

dos65 (approved):
LGTM!

PR #3418, opened 2021-12-30
refactor: small build cleanups

ckipp01:
This pr just changes a few small things in the build:

- Moves the list of developers to `project/Developers.scala`
- Takes deps that we use in multiple places and give them a `V.name` variable to use
- Alphabetizes the versions in `V`
- Removes duplicated deps declared in both mtags and metals

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 import scala.collection.mutable
 import scala.sys.process._
-import Tests._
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM! 

It should be fine to merge, the scalafix issue is solved in main.

PR #3420, opened 2021-12-30
refactor: take care of some deprecation warnings

ckipp01:
I through in `-deprecation` and took care of most of the deprecation warnings. There are still a couple left, but they are situations where we actually need to use the deprecated methods (since we don't support workspaceFolders for example) or other things that take more thought, like not using `thread.stop`.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
@@ -96,7 +96,6 @@ private[debug] final class SetBreakpointsRequestHandler(
     val partition = new SetBreakpointsArguments
     partition.setBreakpoints(Array.empty)
     partition.setSource(source)
-    partition.setLines(Array.empty)
     partition.setSourceModified(request.getSourceModified)
     partition
   }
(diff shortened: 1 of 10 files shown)


metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
```diff
@@ -109,14 +108,9 @@ private[debug] final class SetBreakpointsRequestHandler(
...
     source.setPath(s"dap-fqcn:$fqcn")
 
-    val lines = breakpoints
-      .map(_.getLine: Int)
-      .distinct
-
     val partition = new SetBreakpointsArguments
     partition.setBreakpoints(breakpoints)
     partition.setSource(source)
-    partition.setLines(lines)
```
ckipp01, line 119: I want to do some testing locally on these, but reading through it, it makes it seem that the lines should always be accessed through the breakpoints themselves, and that's why the `getLines` is also deprecated. So I don't believe this is even needed. I need to test a bit more locally to make sure this is still all working as expected.

ckipp01:
Weird with Bloop I have no compilation issues, but I do with sbt..

metals/src/main/scala/scala/meta/internal/metals/debug/SetBreakpointsRequestHandler.scala
ckipp01: Testing locally with `nvim-metals` this all seems to work fine, but can someone with VS Code test as well?

tgodzik: Looks to work ok! I think it was just added in case it was needed indeed.

tgodzik (approved):
LGTM!

Issue #3422, opened 2021-12-31
Hover and insert type annotation shows error


ckipp01:
**Describe the bug**

Given the following piece of code when in Scala 3:

```scala
final case class Dependency(
    org: String,
    name: Option[String],
    version: Option[String]
)

object Dependency {
  def apply(org: String) = Dependency(org, None, None)
  def apply(org: String, name: String) = Dependency(org, Some(name), None)
}
```

You'll get diagnostics on the apply that:

```
Overloaded or recursive method apply needs return type
```
Then when you try to use the insert type annotation code action on `apply` you end up with:

```scala
  def apply(org: String): <error unspecified error> = Dependency(org, None, None)
```
You can also see the `<error unspecified error>` in the hover on `apply` as well.

**Expected behavior**

This error shouldn't be shown to the user in the hover, and it for sure shouldn't be inserted as the type

**Installation:**

- Operating system: macOS
- Editor: nvim-metals
- Metals version: 0.10.9+249-88f3b27d-SNAPSHOT
- Scala version: 3.1.0

tgodzik:
Thanks Chris for noticing these issues, every time I've seen it I was busy with something else and forgot to file it.

Issue #3421, opened 2021-12-30
Odd completion given back


ckipp01:
**Describe the bug**

I'm only able to reproduce this while using `scala-cli` and Scala 3. Given the following piece of code:

```scala
// using scala 3.1.0

def thing() = ??? // same behavior happens if this is implemented

@main def run() = {

  val thing = ???

}
```

When you try to implement the `val thing` and start typing `thi` you'll see this as a completion option given:

```json
  {
    "label": "thing: \u003cerror Not found: thi\u003e",
    "kind": 5,
    "detail": "\u003cerror Not found: thi\u003e",
    "sortText": "00000",
    "filterText": "thing",
    "textEdit": {
      "range": {
        "start": {
          "line": 6,
          "character": 14
        },
        "end": {
          "line": 6,
          "character": 17
        }
      },
      "newText": "thing"
    },
    "additionalTextEdits": []
  }
```

**To Reproduce**

1. Create a scala-cli script like above
2. Try implementing the `val thing` with `thi`
3. See the funky `thing: <error Not found: thi>` completion label.

Note t
**Expected behavior**

I wouldn't expect `thing` to have this label here.

**Screenshots**

<img width="472" alt="Screenshot 2021-12-30 at 21 09 42" src="https://user-images.githubusercontent.com/13974112/147785314-e74dc059-3ab6-473e-a378-f810204a1bdb.png">

**Installation:**

- Operating system: macOS
- Editor: nvim-metals
- Metals version: 0.10.9+249-88f3b27d-SNAPSHOT

**Search terms**

ckipp01:
I'm also seeing this in other situations for example I just noticed this:
<img width="719" alt="Screenshot 2021-12-31 at 13 07 02" src="https://user-images.githubusercontent.com/13974112/147822539-199f32a6-02e2-40d7-b8e9-cb76ee84af99.png">

I made a spelling mistake here and tried to match on `deps` instead of `dep`, but for some reason I'm seeing that error message in my completion item.

tgodzik:
We can probably check if the type is an error, let me check that.

PR #3423, opened 2021-12-31
[Scala 3] Fix issues when type cannot be inferred by the compiler

tgodzik:
Includes:
- print Any if type cannot be inferred
- don't show error methods on hover

This fixes https://github.com/scalameta/metals/issues/3422 and https://github.com/scalameta/metals/issues/3421

### DIFF
--- a/tests/cross/src/test/scala/tests/hover/HoverErrorSuite.scala
+++ b/tests/cross/src/test/scala/tests/hover/HoverErrorSuite.scala
@@ -13,4 +13,20 @@ class HoverErrorSuite extends BaseHoverSuite {
        |""".stripMargin,
     ""
   )
+
+  check(
+    "error",
+    """|final case class Dependency(
+       |    org: String,
+       |    name: Option[String],
+       |    version: Option[String]
+       |)
+       |
+       |object Dependency {
+       |  def <<ap@@ply>>(org: String) = Dependency(org, None, None)
+       |}
+       |""".stripMargin,
+    "".stripMargin
+  )
+
 }
--- a/tests/cross/src/test/scala/tests/pc/CompletionErrorSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/CompletionErrorSuite.scala
@@ -1,0 +1,43 @@
+package tests.pc
+
+import tests.BaseCompletionSuite
+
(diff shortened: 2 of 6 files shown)


tests/cross/src/test/scala/tests/pc/CompletionErrorSuite.scala
```diff
@@ -0,0 +1,43 @@
...
+       |object InvalidDependency extends Dependency
+       |object Main{
+       |  def exists(dep: Dependency) = {
+       |    deps match {
+       |      case invalid @ InvalidDependency => println(inv@@)
+       |    }
+       |  }
+       |}
+       |""".stripMargin,
+    "invalid: InvalidDependency.type",
```
tgodzik, line 36: I tried to follow how we handle things in Scala 2, but in this case the Scala 2 compiler gives us a bit more info.

tests/cross/src/test/scala/tests/hover/HoverErrorSuite.scala
```diff
@@ -13,4 +13,21 @@ class HoverErrorSuite extends BaseHoverSuite {
...
+    "error",
+    """|final case class Dependency(
+       |    org: String,
+       |    name: Option[String],
+       |    version: Option[String]
+       |)
+       |
+       |object Dependency {
+       |  def <<ap@@ply>>(org: String) = Dependency(org, None, None)
+       |  def apply(org: String, name: String) = Dependency(org, Some(name), None)
```
ckipp01, line 27: I could have simplified this further, but just lazy copy pasted when I created the issue. The second apply can just be removed in all of these as well.

ckipp01 (approved):
Just tested and this seems to do the trick! Thanks for fixing this so quickly! Just one comment but not a blocker.

tgodzik:
> Just tested and this seems to do the trick! Thanks for fixing this so quickly! Just one comment but not a blocker.

Wanted to fix it quickly before I forget again! :D

tests/cross/src/test/scala/tests/hover/HoverErrorSuite.scala
tgodzik: Removed!

PR #3426, opened 2021-12-31
Show more information in the doctor

tgodzik:
We haven't revamped the doctor in a while and I feel with the upcoming Java support this is as good a chance as any.

What was changed:
- instead of scala version we now have sbt version writen under `Type` in case of sbt targets
- Java targets were moved to another column instead of being a separate entity
- merged definition and completions into one column called `Interactive` to indicate that Scala version is supported
- renamed `references` to semanticdb since there is currently a huge number of features that actually depend on it
- added explanations for different statuses underneath
- added a deubgging column and marked it disable explicitely for sbt targets

I would love to hear your opinions especially:
- will it work with other editors
- what else could we add
- do the new names and explanations make sense

Also related to https://github.com/scalameta/metals/issues/3398#issuecomment-1001158421

Example:

image

I was thinking of alternatively showing these things while hovering, but that can be easily missed.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
@@ -321,39 +321,115 @@ final class Doctor(
             )
         )
     } else {
+      val allTargetsInfo = targetIds.flatMap(extractTargetInfo)
       html
         .element("table")(
           _.element("thead")(
             _.element("tr")(
               _.element("th")(_.text("Build target"))
-                .element("th")(_.text("Scala"))
+                .element("th")(_.text("Type"))
                 .element("th")(_.text("Diagnostics"))
-                .element("th")(_.text("Goto definition"))
-                .element("th")(_.text("Completions"))
-                .element("th")(_.text("Find references"))
+                .element("th")(_.text("Interactive"))
+                .element("th")(_.text("Semanticdb"))
+                .element("th")(_.text("Debugging"))
+                .element("th")(_.text("Java support"))
                 .element("th")(_.text("Recommendation"))
             )
-          ).element("tbody")(html => buildTargetRows(html, targetIds))
+          ).element("tbody")(html => buildTargetRows(html, allTargetsInfo))
         )
+
+      // Additional explanations
+      explanation(
+        html,
+        title = "Diagnostics:",
+        correctMessage =
+          s"${Icons.unicode.check} - diagnostics correctly being reported by the build server",
+        incorrectMessage =
+          s"${Icons.unicode.alert} - only syntactic errors being reported",
+        show = allTargetsInfo.exists(_.diagnosticsStatus.isCorrect == false)
+      )
+
+      explanation(
+        html,
+        title = "Interactive features (completions, hover):",
+        correctMessage = s"${Icons.unicode.check} - supported Scala version",
+        incorrectMessage =
+          s"${Icons.unicode.error} - unsupported Scala version",
+        show = allTargetsInfo.exists(_.interactiveStatus.isCorrect == false)
+      )
+
+      explanation(
+        html,
+        title =
+          "Semanticdb features (references, renames, go to implementation):",
+        correctMessage =
+          s"${Icons.unicode.check} - build tool automatically creating needed semanticdb files",
+        incorrectMessage =
+          s"${Icons.unicode.error} - semanticdb not being produced",
+        show = allTargetsInfo.exists(_.indexesStatus.isCorrect == false)
+      )
+
+      explanation(
+        html,
+        title = "Debugging (run/test, breakpoints, evaluation):",
+        correctMessage =
+          s"${Icons.unicode.check} - users can run or test their code with debugging capabilities",
+        incorrectMessage =
+          s"${Icons.unicode.error} - the tool does not support debugging in this target",
+        show = allTargetsInfo.exists(_.debuggingStatus.isCorrect == false)
+      )
+
+      explanation(
+        html,
+        title = "Java Support:",
+        correctMessage =
+          s"${Icons.unicode.check} - working non-interactive features (references, rename etc.)",
+        incorrectMessage =
+          s"${Icons.unicode.error} - missing semanticdb plugin, might not be added automatically by the build server",
+        show = allTargetsInfo.exists(_.javaStatus.isCorrect == false)
+      )
     }
+  }
+
+  private def explanation(
+      html: HtmlBuilder,
+      title: String,
+      correctMessage: String,
+      incorrectMessage: String,
+      show: Boolean
+  ) = {
+    html.element("div")(
+      _.element("p")(_.text(title)).element("ul") { ul =>
+        ul.element("li")(_.text(correctMessage))
+        if (show)
+          ul.element("li")(_.text(incorrectMessage))
+      }
+    )
+
   }
 
   private def buildTargetRows(
       html: HtmlBuilder,
-      targetIds: Seq[BuildTargetIdentifier]
+      targetIds: Seq[DoctorTargetInfo]
   ): Unit = {
     targetIds
-      .flatMap(extractTargetInfo)
       .sortBy(f => (f.baseDirectory, f.name, f.dataKind))
       .foreach { targetInfo =>
         val center = "style='text-align: center'"
         html.element("tr")(
           _.element("td")(_.text(targetInfo.name))
-            .element("td")(_.text(targetInfo.scalaVersion))
-            .element("td", center)(_.text(Icons.unicode.check))
-            .element("td", center)(_.text(targetInfo.definitionStatus))
-            .element("td", center)(_.text(targetInfo.completionsStatus))
-            .element("td", center)(_.text(targetInfo.referencesStatus))
+            .element("td")(_.text(targetInfo.targetType))
+            .element("td", center)(
+              _.text(targetInfo.diagnosticsStatus.explanation)
+            )
+            .element("td", center)(
+              _.text(targetInfo.interactiveStatus.explanation)
+            )
+            .element("td", center)(_.text(targetInfo.indexesStatus.explanation))
+            .element("td", center)(
+              _.text(targetInfo.debuggingStatus.explanation)
+            )
+            .element("td", center)(_.text(targetInfo.javaStatus.explanation))
             .element("td")(_.raw(targetInfo.recommenedFix))
         )
       }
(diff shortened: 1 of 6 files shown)


tgodzik:
Bigger example here: https://imgur.com/a/tuJEqVC

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -326,16 +326,107 @@ final class Doctor(
...
+    // Additional explanations
+    html.element("div")(
+      _.element("p")(
+        _.text(
+          "Diagnostics:"
+        )
+      ).element("ul")(
+        _.element("li")(
+          _.text(
+            s"${Icons.unicode.check} - connected build server showing diagnostics"
```
ckipp01, line 350:
```suggestion
            s"${Icons.unicode.check} - diagnostics correctly being reported by the build server"
```
🤔 thinking about this a bit, we could also include the build server and version here and no long need the extra section on top? Although it'd make it less up and front.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -326,16 +326,107 @@ final class Doctor(
...
+          )
+        )
+          .element("li")(
+            _.text(s"${Icons.unicode.alert} - only syntactic errors")
```
ckipp01, line 354:
```suggestion
            _.text(s"${Icons.unicode.alert} - only syntactic errors being reported")
```

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -326,16 +326,107 @@ final class Doctor(
...
+          "Interactive features (completions, hover):"
+        )
+      ).element("ul")(
+        _.element("li")(
+          _.text(
+            s"${Icons.unicode.check} - supported Scala version"
+          )
+        )
+          .element("li")(
+            _.text(s"${Icons.unicode.error} - unsupported Scala version")
```
ckipp01, line 371: What do you think of only showing the icons that appear up above? Would that be difficult? It might be nice to for example not show this part at all unless a build target is actually using an unsupported Scala version.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -326,16 +326,107 @@ final class Doctor(
...
+
+    html.element("div")(
+      _.element("p")(
+        _.text(
+          "Semanticdb features (references, renames, go to implementation):"
+        )
+      ).element("ul")(
+        _.element("li")(
+          _.text(
+            s"${Icons.unicode.check} - build tool creates needed semanticdb files automatically"
```
ckipp01, line 384:
```suggestion
            s"${Icons.unicode.check} - build tool automatically creating needed semanticdb files"
```

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -326,16 +326,107 @@ final class Doctor(
...
+          )
+        )
+          .element("li")(
+            _.text(
+              s"${Icons.unicode.error} - no needed semanticdb indexes are produced"
```
ckipp01, line 389:
```suggestion
              s"${Icons.unicode.error} - semanticdb not being produced"
```

This is maybe another one that I'd just fully leave out unless there is an issue, wdyt?

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -326,16 +326,107 @@ final class Doctor(
...
+          "Java Support:"
+        )
+      ).element("ul")(
+        _.element("li")(
+          _.text(
+            s"${Icons.unicode.check} - working non-interactive features (references, rename etc.)"
+          )
+        )
+          .element("li")(
+            _.text(s"${Icons.unicode.error} - missing semanticdb plugin")
```
ckipp01, line 426: Same as the above suggestions, I think it might be better to not display these unless they are up in the table. Another thought, we should try to make them as actionable as possible. I know for example when using `scalal-cli` you get the warning that there is no semanticdb for java targets, but it sort of leaves the user with the question "well what do I need to do to get it? How do I fix this?" In this situation will there also be a recommendation attatched?

ckipp01 (changes requested):
Overall I like this approach a lot better. I like that it explains a bit about what the columns actually mean. The biggest though I think I have is that maybe the keys down below should only show the errors/warnings if there is an error or a warning up above. I don't think it's helpful to show that `X means unsupported Scala versions` if they aren't using an unsupported Scala version.

In terms of support for other editors, this does break the Doctor, but the changes are minor, so no issues on my end for `nvim-metals`.

Other than that jus some small wording change recommendations.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: I think it's good to show it up top in a prominent location and I think it would be a bit akward to include it in the `legend`

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: The recommendation should show up on the right, I can add some better info there.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: Should we also show the explanation for a correct status if everything is correct in the column?

ckipp01: I _think_ that still makes sense. Because for example semanticdb may not be super well known, so having a ✅ with information about what it means might help people understand it better.

tgodzik: Something along these lines?

image

ayoub-benali:
Maybe the JSON should have a version number, instead of the clients relying on the Metals version.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: image

Ok, this time showing all the existing statuses.

tgodzik: Updated the PR with the current version.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: Added a bit more info that it might be automatically added by the build server. I will work on sbt to add this automatically.

tgodzik:
> Maybe the JSON should have a version number, instead of the clients relying on the Metals version.

Having just a separate version field or make client send the server its expected version and get a response with that exact version?

Alternatively, we could adapt the json format to describe the columns, such as it would correspond to what is being shown in VS Code and can be renedered by the client as they will?

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -362,63 +438,77 @@ final class Doctor(
...
       targetId: BuildTargetIdentifier
   ): List[DoctorTargetInfo] = {
-    val scalaDoctorInfo =
-      buildTargets.scalaTarget(targetId).map(extractScalaTargetInfo).toList
-    val javaDoctorInfo =
-      buildTargets.javaTarget(targetId).map(extractJavaTargetInfo).toList
-    scalaDoctorInfo ::: javaDoctorInfo
+    val javaTarget = buildTargets.javaTarget(targetId)
+    buildTargets
+      .scalaTarget(targetId)
```
Arthurm1, line 443: I don't think this will return anything if the JavaTarget exists and the ScalaTarget doesn't which is possible in Bloop/Gradle projects.

tgodzik: So in that case we should have a separate Java target I guess, good catch!

ckipp01:
> > Maybe the JSON should have a version number, instead of the clients relying on the Metals version.
> 
> Having just a separate version field or make client send the server its expected version and get a response with that exact version?

We could just add a `version` to the `DoctorResults` case class. Then when we get it we can do like no version, then do the old format, if we get version 1, then new format, and then just increment it if we change it. It's probably not super often we'll change headings and keys like this.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -362,63 +438,77 @@ final class Doctor(
...
+  private def extractScalaTargetInfo(
+      target: ScalaTarget,
+      javaTarget: Option[JavaTarget]
+  ) = {
     val scalaVersion = target.scalaVersion
-    val definition: String =
+    val interactive =
       if (mtagsResolver.isSupportedScalaVersion(scalaVersion)) {
-        Icons.unicode.check
+        DoctorStatus(Icons.unicode.check, isCorrect = true)
```
dos65, line 455: It seems that there is only 3 instances of statuses so they might be extracted into companion object
```
object DoctorStatus {
   val check = DoctorStatus(Icons.unicode.check, true)
   val alert = DockerStatus(Icons.unicode.alert, false)
   val error = DocktorStatus(Icons.unicode.error, false)
}
```

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -362,63 +438,77 @@ final class Doctor(
...
+      case Some(target) =>
+        (
+          DoctorStatus(Icons.unicode.alert, isCorrect = false),
+          problemResolver.recommendation(target)
+        )
+      case None => (DoctorStatus(Icons.unicode.alert, isCorrect = false), None)
+    }
+
+    val canRun = target.info.getCapabilities().getCanRun()
+    val canTest = target.info.getCapabilities().getCanTest()
```
dos65, line 489: @kpodsiad I remember that there were issue with these capabilities values in bloop and sbt. Are they reliable now?

dos65 (commented):
Good job! It makes more sense than the old one.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
kpodsiad: run and test are reliable, only debug was problematic

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: It seems that we can't add it yet to sbt BSP: https://github.com/sourcegraph/lsif-java/issues/390
I added the explanation that the only supported tool is Bloop

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
tgodzik: Good idea! Refactored and it looks much nicer.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -321,39 +321,115 @@ final class Doctor(
...
+          ul.element("li")(_.text(incorrectMessage))
+      }
+    )
+
+  }
+
   private def buildTargetRows(
       html: HtmlBuilder,
-      targetIds: Seq[BuildTargetIdentifier]
+      targetIds: Seq[DoctorTargetInfo]
```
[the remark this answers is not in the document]
tgodzik: Fixed

metals/src/main/scala/scala/meta/internal/metals/DoctorResults.scala
```diff
@@ -8,10 +8,12 @@ final case class DoctorResults(
     messages: Option[List[DoctorMessage]],
     targets: Option[Seq[DoctorTargetInfo]]
 ) {
+  private val version = 1
   def toJson: Obj = {
     val json = ujson.Obj(
       "title" -> title,
-      "headerText" -> headerText
+      "headerText" -> headerText,
+      "version" -> version
```
tgodzik, line 16: @ckipp01 @ayoub-benali I added a version field starting at `1`

Arthurm1:
I think it would be better to have `Semanticdb` ticked on a pure Java project.  It can be checked with `javaTarget.isSemanticdbEnabled`

Currently it's red which doesn't match what the legend says

image

image

tgodzik:
> I think it would be better to have `Semanticdb` ticked on a pure Java project. It can be checked with `javaTarget.isSemanticdbEnabled`
> 
> Currently it's red which doesn't match what the legend says

You're right, it should be fixed now! Thanks for taking a look. :smiley:

kpodsiad:
Screenshot 2022-01-11 at 14 48 32

Screenshot 2022-01-11 at 14 47 47

I think we should change the description of interactive features for Java targets. `Unsupported Scala version` doesn't take into account java targets.

tgodzik:
> I think we should change the description of interactive features for Java targets. `Unsupported Scala version` doesn't take into account java only targets.

Changed to `interactive features are unsupported for Java and older Scala versions`, thanks!

kpodsiad (approved):
Awesome!

dos65 (approved):
LGTM!

tgodzik:
@ckipp01 is the version parameter ok?

Arthurm1 (approved):
Works great for me

ckipp01:
> 

Yup, that works perfectly for me!

ckipp01:
So I'm working on implementing this right now and I can look into it, but at first look, was the extra explanations meant to be left out of the json payload? Right now for example it only returns:
```json
{
  "title": "Metals Doctor",
  "headerText": "Build server currently being used is Bloop v1.4.11.nnThese are the installed build targets for this workspace. One build target corresponds to one classpath. For example, normally one sbt project maps to two build targets: main and test.",
  "version": 1,
  "targets": [
    {
      "buildTarget": "Sanity",
      "targetType": "Scala 2.12.15",
      "diagnostics": "✅ ",
      "interactive": "✅ ",
      "semanticdb": "✅ ",
      "debugging": "✅ ",
      "java": "✅ ",
      "recommendation": ""
    },
    {
      "buildTarget": "Sanity.test",
      "targetType": "Scala 2.12.15",
      "diagnostics": "✅ ",
      "interactive": "✅ ",
      "semanticdb": "✅ ",
      "debugging": "✅ ",
      "java": "✅ ",
      "recommendation": ""
    }
  ]
}
```

tgodzik:
I didn't include them yes, totally forgot to take a look at it. Sorry about that!

If you could take a look at it it would be awesome, I am trying to work on the release notes.

ckipp01:
> I didn't include them yes, totally forgot to take a look at it. Sorry about that!
> 
> If you could take a look at it it would be awesome, I am trying to work on the release notes.

👍🏼 yup on it.

PR #3453, opened 2022-01-01
Scala Steward - ignore org.eclipse updates

dos65:
We pinned eclipse deps versions and we aren't goint to update as long as we keep jdk8 support


ckipp01 (approved):
👍🏼

PR #3459, opened 2022-01-03
fix: skip extra search for main methods

### LINKED ISSUE #3458: Scala 3 @main can't be discovered using debug discovery
**Describe the bug**

When using the debug discovery with ~`run` or~ `runOrTestFile` on a Scala 3 main defined with `@main` it won't work. However when running with the code lens, it does work.

**To Reproduce**

1. `g8 scala/scala3.g8`
2. Import and navigate to `Main.scala`
3. Run the file with the code lens, all works
4. Try to trigger a run file, fails.

You see this in the logs:

```
SEVERE: Internal error: scala.meta.internal.metals.debug.NoRunOptionException$: There is nothing to run or test in the current file.
java.util.concurrent.CompletionException: scala.meta.internal.metals.debug.NoRunOptionException$: There is nothing to run or test in the current file.
```

**Expected behavior**

I'd expect this command to find the main method just like it can in Scala 2.

**Installation:**

- Operating system: macOS
- Editor: nvim-metals
- Metals version: 0.10.9+273-2dcf620b-SNAPSHOT

ckipp01:
Since we already have the mainClasses we can pass them right into
`verifyMain` just like we do when we use `run`. So this skips going
through the document symbols again and trying to get the main class. I'm
not 100% sure why this works for Scala 2 but not for Scala 2 using
`@main`, but this fixes it for me when using `@main`.

Closes #3458

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
@@ -305,13 +305,8 @@ class DebugProvider(
           symbol = symbolInfo.symbol
           testClass <- testClasses.get(symbol)
         } yield testClass
-        val mains = for {
-          symbolInfo <- textDocument.symbols
-          symbol = symbolInfo.symbol
-          mainClass <- classes.get(symbol)
-        } yield mainClass
-        if (mains.nonEmpty) {
-          verifyMain(buildTarget, mains.toList, params)
+        if (classes.nonEmpty) {
+          verifyMain(buildTarget, classes.values.toList, params)
         } else if (tests.nonEmpty) {
           Future(
             new b.DebugSessionParams(


metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -305,13 +305,8 @@ class DebugProvider(
...
         } yield testClass
-        val mains = for {
-          symbolInfo <- textDocument.symbols
-          symbol = symbolInfo.symbol
-          mainClass <- classes.get(symbol)
-        } yield mainClass
-        if (mains.nonEmpty) {
-          verifyMain(buildTarget, mains.toList, params)
+        if (classes.nonEmpty) {
+          verifyMain(buildTarget, classes.values.toList, params)
```
tgodzik, line 309: Doesn't this mean we will get all the main classes in the project? I think this is why I added the extra step, but it seems that the symbol is maybe wrongly added to `classes` ?

ckipp01: Ahhh duh, of course. I totally blanked on that. Yes, this does mean you get everything for the entire target since that's how `run` works.  🤔 I'll keep digging to see what's going on then.

ckipp01: Ahhh the symbol is `_empty_/Main$package.hello().` whereas the main classes have `_empty_/hello.

Issue #3454, opened 2022-01-02
Insert type annotation in Scala 3 breaks on `Either`


ckipp01:
**Describe the bug**

When trying to use the "Insert type annotation" feature in Scala 3 when the inferred type is an `Either` somehow the `import scala.{util => ju}` auto import is getting brought in and `ju.Either` is getting slammed into the variable name.

**To Reproduce**

1. Have some code like this

```scala
  def returnEither(value: String) =
    if value == "left" then Left("a") else Right("b")
```
2. Try to use the "Insert type annotation" code action on `returnEither`
3. See the following

```
import scala.{util => ju}

def returnEitju.Eitherher(value: String): Either[String, String] =
  if value == "left" then Left("a") else Right("b")
```
**Expected behavior**

Just inserts the expected type without the `ju.Either stuff`

**Screenshots**

![2022-01-02 09 50 45](https://user-images.githubusercontent.com/13974112/147870871-c3fc9659-1b3d-4113-a271-7d914b41f9ca.gif)

**Installation:**

- Operating system: macOS
- Editor: `nvim-metals`
- Metals version: v0.10.9+271-a8bb69f6-SNAPSHOT

PR #3460, opened 2022-01-03
[Scala 3] Fix insert type in case of type aliases

tgodzik:
Previously, we would treat Either type alias from Predef as a separate entity than scala.util.Either. Now, we also check type aliases for conflicting imports.

Fixes https://github.com/scalameta/metals/issues/3454

### DIFF
--- a/tests/cross/src/test/scala/tests/pc/InsertInferredTypeSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/InsertInferredTypeSuite.scala
@@ -454,6 +454,30 @@ class InsertInferredTypeSuite extends BaseCodeActionSuite {
        |""".stripMargin
   )
 
+  checkEdit(
+    "either",
+    """|object O{
+       |  def <<returnEither>>(value: String) = {
+       |    if (value == "left") Left("a") else Right("b")
+       |  }
+       |}""".stripMargin,
+    """|object O{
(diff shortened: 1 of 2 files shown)


tests/cross/src/test/scala/tests/pc/InsertInferredTypeSuite.scala
```diff
@@ -454,6 +454,30 @@ class InsertInferredTypeSuite extends BaseCodeActionSuite {
        |""".stripMargin
   )
 
+  checkEdit(
+    "either",
+    """|object O{
+       |  def <<returnEither>>(value: String) = {
```
tgodzik, line 460: I can't seem to reproduce the other problems from the issue (wrong rename and position) with the proposed fix in place, so it seems it might only be related to the basic Scala type aliases.

dos65 (approved):
Looks good but I'm worried that #3454 spots more issues.
I can't explain why method name rewriting happens in attached gif.

tgodzik:
> Looks good but I'm worried that #3454 spots more issues. I can't explain why method name rewriting happens in attached gif.

I haven't been able to reproduce it with the fix here, so might be specific issue with the Predef object.

ckipp01 (approved):
I did just try this out btw and everything seems to be working fine now in the example I gave before. 👍🏼 LGTM

PR #3467, opened 2022-01-04
refactor: refactor bug_report.md to be a yaml template

ckipp01:
You can see an example of what this looks like here: https://github.com/ckipp01/metals-minimal-sbt/issues/new/choose

I sort of like how the yml templates are displayed nicer. This also gives us a bit more control over marking this as mandatory. Wdyt?


tgodzik (approved):
Looks cool! Thanks!

dos65:
@ckipp01 I've noticed that having almost all fields required might be inconvenient for raising a bit abstract issues. :smile: 

Don't you mind if I drop `required: true` from everything except of description?
Or maybe we need a different type of issue for [this case](https://github.com/scalameta/metals/issues/3473)

ckipp01:
> Don't you mind if I drop required: true from everything except of description?

🤔 I think for most issues that people create a lot of stuff like version etc should be required, but I totally get what you mean. I can imagine the one you just created in https://github.com/scalameta/metals/issues/3473 didn't fit the template well at all 😆 .

> Or maybe we need a different type of issue for this case

Yea this could be an option as well, but it might be confusing for users.

Maybe let's relax it like you said, mark them not as required and we'll just see how it goes. That's probably easiest for now.

PR #3468, opened 2022-01-04
Add folding range for Java using Scanner

tgodzik:
Turns out there is also a Scanner we can use for Java files, which is handy for figuring out braces and comments in the code.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/parsing/JavaFoldingRangeExtractor.scala
+++ b/metals/src/main/scala/scala/meta/internal/parsing/JavaFoldingRangeExtractor.scala
@@ -89,39 +89,47 @@ final class JavaFoldingRangeExtractor(
       if (token != ITerminalSymbols.TokenNameEOF) {
         val addLine = scanner.getRawTokenSource().count(_ == '\n')
         def next = scanner.getNextToken()
-        def startCharacter = {
-          if (foldOnlyLines) 0
-          else
-            scanner.getCurrentTokenStartPosition() - scanner.getLineStart(
-              line
-            ) - 1
-        }
-        def endCharacter = {
-          if (foldOnlyLines) 0
-          else
-            scanner.getCurrentTokenEndPosition() -
-              scanner.getLineStart(line + addLine) - 1
-        }
+        def currentCharacterStart =
+          scanner.getCurrentTokenStartPosition() - scanner.getLineStart(line)
+        def currentCharacterEnd =
+          scanner.getCurrentTokenEndPosition() - scanner.getLineStart(
+            line + addLine
+          )
+
         token match {
           case ITerminalSymbols.TokenNameLBRACE =>
-            val ch = startCharacter
-            gather(next, line, LBrace(line, ch) :: acc)
+            val startCharacter =
+              if (foldOnlyLines) 0 else currentCharacterStart - 1
+            gather(next, line, LBrace(line, startCharacter) :: acc)
           case ITerminalSymbols.TokenNameRBRACE =>
-            val ch = startCharacter
-            gather(next, line, RBrace(line, ch) :: acc)
+            val (endLine, endChracter) =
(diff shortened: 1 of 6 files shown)


ckipp01 (approved):
Very cool, LGTM!

Arthurm1:
Works great locally but it's different from the Scala one.

In Scala you see the opening **and** closing bracket/comment when folded.  e.g.
```scala
  /*...
  */

  case class Foo(bar: Int) {...
  }
```
In Java you see only the opening one...
```Java
  /*...

  public class Foo {...
```

Not sure which is correct as they seem to both perform functionally the same.

Folding imports works the same.

Arthurm1:
From the way VSCode folds JSON, it looks like the Scala folding is correct.

The Scala `FoldingRangeProvider` has this...
```scala
  def add(range: FoldingRange, adjust: Boolean): Unit = {
    if (isNotCollapsed(range)) {
      if (adjust && foldOnlyLines) {
        range.setEndLine(
          range.getEndLine - 1
        ) // we want to preserve the last line containing e.g. '}'
      }
```

tgodzik:
> From the way VSCode folds JSON, it looks like the Scala folding is correct.
> 
> The Scala `FoldingRangeProvider` has this...
> ```scala
>   def add(range: FoldingRange, adjust: Boolean): Unit = {
>     if (isNotCollapsed(range)) {
>       if (adjust && foldOnlyLines) {
>         range.setEndLine(
>           range.getEndLine - 1
>         ) // we want to preserve the last line containing e.g. '}'
>       }
> ```

Wasn't sure about the exact desired behaviour, thanks for checking! I should be able to do it the same as in Scala code

tgodzik:
@Arthurm1 it should work correctly now with the last line also shown in case of comments and braces.

metals/src/main/scala/scala/meta/internal/parsing/JavaFoldingRangeExtractor.scala
```diff
@@ -0,0 +1,213 @@
...
+        line + addLine
+      }
+    }
+
+    @tailrec
+    def gatherImportLines(
+        token: Int,
+        line: Int,
+        lastImportLine: Int
+    ): (Int, Int, Int, Int) = {
```
kpodsiad, line 65: wdyt about using case class instead of tuple? I know that's it's used only once, but four Ints looks a little bit obscure

kpodsiad (approved):

Arthurm1:
@tgodzik The closing brace now shows - which is great.

There is still a difference between Scala and Java but I'm not sure it matters - I've never actually looked at folding this closely before.

In Scala you can fold anything with more than 1 line between the braces.  So here `Foo` will fold but `Bar` won't...
```scala
class Foo {
  // 1
  // 2
}

class Bar {
  // 1
}
```
In Java `Foo` will fold but `Bar` won't but there's additional lines in both...
```java
class Foo {
  // 1
  // 2
  // 3
}

class Bar {
  // 1
  // 2
}
```

I guess `JavaFoldingRangeExtractor#createFoldingRange` should have `if (endLine - startLine >= spanThreshold)` instead of `if (endLine - startLine > spanThreshold)`?

tgodzik:
> I guess `JavaFoldingRangeExtractor#createFoldingRange` should have `if (endLine - startLine >= spanThreshold)` instead of `if (endLine - startLine > spanThreshold)`?

Changed! Should be ok now, I also fixed issues with Windows, which have shown an issue where I was  using 0-based lines instead of 1-base, which is how the scanner works.

metals/src/main/scala/scala/meta/internal/parsing/JavaFoldingRangeExtractor.scala
tgodzik: Done!

Arthurm1:
LGTM

Arthurm1 (approved):

PR #3469, opened 2022-01-04
docs: migrate gitter references to Discord

ckipp01:
Following the official announcement that Discord is now where the
official Scala chat is located, I think it makes sense to focus on
support efforts on Discord. Lately I've also noticed that our Gitter has
been much quieter with more people coming to Discord anyways. We don't
have to sunset Gitter or anything, but instead just remove the
references to it in various places since most seem to be heading to
Discord anyways.

What do you all think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Metals
 
-<a href="https://gitter.im/scalameta/metals">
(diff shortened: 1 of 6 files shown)


dos65 (approved):
:+1:  from me
Honestly, I haven't opened Gitter for about a several months :smile_cat:

tgodzik (approved):
Good idea, I think we should discourage from gitter, I would prefer less communication channels to follow :sweat_smile:

PR #3470, opened 2022-01-05
[Java] Support go-to-definition for `*.java` dependency sources

dos65:
Implemented using sourcegraph semanticdb compiler plugin.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
@@ -261,6 +261,14 @@ object Embedded {
       semanticdbScalacDependency(scalaVersion),
       Some(scalaVersion)
     )
+
+  def downloadSemanticdbJavac: List[Path] = {
+    downloadDependency(
+      Dependency.of("com.sourcegraph", "semanticdb-javac", "0.7.3"),
+      None
+    )
+  }
+
   def downloadMtags(scalaVersion: String): List[Path] =
     downloadDependency(mtagsDependency(scalaVersion), Some(scalaVersion))
 
--- a/metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
@@ -40,7 +40,8 @@ final class InteractiveSemanticdbs(
     statusBar: StatusBar,
     compilers: () => Compilers,
     clientConfig: ClientConfiguration,
-    semanticdbIndexer: () => SemanticdbIndexer
+    semanticdbIndexer: () => SemanticdbIndexer,
+    javaInteractiveSemanticdb: Option[JavaInteractiveSemanticdb]
 ) extends Cancelable
     with Semanticdbs {
 
@@ -77,8 +78,8 @@ final class InteractiveSemanticdbs(
       )
     }
 
-    // anything aside from `*.scala`, `*.sbt` and `*.sc` file
-    def isExcludedFile = !source.isScalaFilename
+    // anything aside from `*.scala`, `*.sbt`, `*.sc`, `*.java` file
+    def isExcludedFile = !source.isScalaFilename && !source.isJavaFilename
 
     if (isExcludedFile || !shouldTryCalculateInteractiveSemanticdb) {
       TextDocumentLookup.NotFound(source)
@@ -160,6 +161,17 @@ final class InteractiveSemanticdbs(
   }
 
   private def compile(source: AbsolutePath, text: String): s.TextDocument = {
+    if (source.isJavaFilename)
+      javaInteractiveSemanticdb.fold(s.TextDocument())(
+        _.textDocument(source, text)
+      )
+    else scalaCompile(source, text)
+  }
+
+  private def scalaCompile(
+      source: AbsolutePath,
+      text: String
+  ): s.TextDocument = {
     def worksheetCompiler =
       if (source.isWorksheet) compilers().loadWorksheetCompiler(source)
       else None
--- a/metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
@@ -1,0 +1,250 @@
+package scala.meta.internal.metals
+
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.Path
+
+import scala.concurrent.Await
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+import scala.util.Properties
+import scala.util.Try
+import scala.util.control.NonFatal
+
+import scala.meta.internal.io.FileIO
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.mtags.MD5
+import scala.meta.internal.process.SystemProcess
+import scala.meta.internal.{semanticdb => s}
+import scala.meta.io.AbsolutePath
+
+import com.typesafe.config.ConfigFactory
+import com.typesafe.config.ConfigParseOptions
+import com.typesafe.config.ConfigSyntax
+
+class JavaInteractiveSemanticdb(
+    javac: AbsolutePath,
+    jdkVersion: JavaInteractiveSemanticdb.JdkVersion,
+    pluginJars: List[Path],
+    workspace: AbsolutePath,
+    buildTargets: BuildTargets
+) {
+
+  private val readonly = workspace.resolve(Directories.readonly)
+
+  def textDocument(source: AbsolutePath, text: String): s.TextDocument = {
+    val sourceRoot = source.parent
+    val workDir = AbsolutePath(
+      Files.createTempDirectory("metals-javac-semanticdb")
+    )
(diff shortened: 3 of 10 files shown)


metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
```diff
@@ -0,0 +1,227 @@
...
+      outLine => stdout += outLine,
+      errLine => stdout += errLine
+    )
+
+    val future = ps.complete.recover { case NonFatal(e) =>
+      scribe.error("Running javac-semanticdb failed", e)
+      1
+    }
+
+    val exitCode = Await.result(future, 10.seconds)
```
tgodzik, line 73: There would be no need to wait if we returned future from the method.

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
```diff
@@ -0,0 +1,227 @@
...
+    // that is declared in some existing module.
+    // It fails with: `error: package exists in another module: $packageName`
+    // but it might be fixed by passing `--patch-module $moduleName=$sourceRoot` option.
+    //
+    // Currently there is no infrastucture to detect if package belong to jigsaw module or not
+    // so this case is covered only for JDK sources.
+    if (jdkVersion.hasJigsaw) {
+      source.toRelativeInside(readonly) match {
+        case Some(rel) =>
+          val names = rel.toNIO.iterator().asScala.toList.map(_.filename)
```
tgodzik, line 116: This might not work with the changes in the Virtual docs PR. We can for sure fix it there afterwards, but maybe it would be possible to prepare for it here?

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
```diff
@@ -0,0 +1,227 @@
...
+        )
+        None
+    }
+  }
+
+  private def getJavaVersionFromJavaHome(
+      javaHome: AbsolutePath
+  ): Option[JdkVersion] = {
+
+    def fromReleaseFile: Option[JdkVersion] = {
```
tgodzik, line 176: Wouldn't this be the same version as the currently running one?

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
dos65: If we aren't going to include Virtual docs to the next release let's keep as it is.
I don't know how to write tests on virtual docs without actual virtual docs impl.

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
dos65: Hmm.. I'm bit lost in java homes.
Can `userConfig.javaHome` be different from process one?

One additional thing I've noticed is that BSP might have different jdk settings (`JvmBuildTarget.javaHome`) but we currently don't handle it.

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
dos65: yep. but it's will trigger much-much more changes. `InteractiveSemanticdb` does the same thing for PC because of `Semandticdb` interface signature.
I can turn it into an issue

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
tgodzik: > Can userConfig.javaHome be different from process one?

`userConfig.javaHome` can be different if the user changed the setting while running, otherwise it should be the same.

> One additional thing I've noticed is that BSP might have different jdk settings (JvmBuildTarget.javaHome) but we currently don't handle it.

Yeah, we could get the java home from BSP I think and maybe use that instead for javaHome in case of definitions?

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
tgodzik: It's fine, we anyway want to limit the time we spend on the compilation.

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
dos65: > Yeah, we could get the java home from BSP I think and maybe use that instead for javaHome in case of definitions?

I would turn it into a seperate PR. This should work well for existing JdkSources mechanic that is taken from `userConfig.javaHome || process java home`. It's needed to switch them on jdk from BSP first. Otherwise there will be jdk version mismatch between javac in this class and the actual source that we pass to it.

dos65: @tgodzik 
reported it here - #3479

Could you we merge as it is?
I think it will be ok to fix in the next release.

metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
```diff
@@ -160,6 +161,17 @@ final class InteractiveSemanticdbs(
   }
 
   private def compile(source: AbsolutePath, text: String): s.TextDocument = {
+    if (source.isJavaFilename)
+      javaInteractiveSemanticdb.fold(s.TextDocument())(
+        _.textDocument(source, text)
+      )
```
kpodsiad, line 167: Do you have a strong preference for `fold`?
```suggestion
      javaInteractiveSemanticdb.map(_.textDocument(source, text)).getOrElse(s.TextDocument())
```

tests/unit/src/test/scala/tests/JavaDefinitionSuite.scala
```diff
@@ -0,0 +1,158 @@
...
+    )
+  )
+
+  def check(
+      name: TestOptions,
+      depSymbol: String,
+      input: String,
+      expected: String,
+      dependencies: List[String] = Nil
+  ): Unit = {
```
kpodsiad, line 70:
```suggestion
  )(implicit loc: Location): Unit = {
```

metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
dos65: No, but what do you have against of it? :detective:

kpodsiad: I just prefer a more explicit way - pattern matching, `gerOrElse`. Fold seems a little bit obscure

tgodzik (approved):
LGTM! Just a small comment, but we can do it in a follow up.

metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
```diff
@@ -261,6 +261,14 @@ object Embedded {
       semanticdbScalacDependency(scalaVersion),
       Some(scalaVersion)
     )
+
+  def downloadSemanticdbJavac: List[Path] = {
```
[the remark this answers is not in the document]
dos65: Done

metals/src/main/scala/scala/meta/internal/metals/InteractiveSemanticdbs.scala
dos65: @kpodsiad Don't you mind if I'll ignore this? :smiling_face_with_tear: 

Fold is a legal function in our codebase.
For me `fold` here is a kind of signal that it's a bit different from usual plain data operation.

Arthurm1:
@tgodzik Does this compile?  Not sure if it was rebased

tgodzik:
> @tgodzik Does this compile? Not sure if it was rebased

My bad, will fix it. Thanks for noticing!

Arthurm1:
@dos65 Are the `--patch-module` options needed here?  I'm playing around with virtual docs + `JavaInteractiveSemantidb` and I can't get it to fail on any example I use after removing these options

dos65:
@Arthurm1 Need to check it one more time. Originally I was just testing it directly using `javac` and noticed that without it you can't compile jdk sources without errors.
If you try to generate it for `java/base/String.java` there will be a lot of compilation errors that I assume will lead that not all parts of this source will have correct occurrences.

Arthurm1:
@dos65 Could you give it a go again when you have a chance and see if you can produce a concrete example.  I've tried with `temurin17` on various JDK classes and it seems to work fine without the `--patch-module` setting.

There doesn't appear to be too many examples out there on why you'd use `--patch-module`, but I was wondering if the idea of it is to allow classes to depend on a new version of a class. i.e. I want to change `String` class and have everything that depends on it compile with my new class instead of the jdk version.  But here we're only compiling one java file with no dependencies so we shouldn't need it.  But I'm not entirely sure

dos65:
@Arthurm1 I've checked that. There are cases when it's needed - I create a pr with such case #3538

PR #3474, opened 2022-01-05
[Issue Template] Relax requirements

dos65:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -27,15 +22,14 @@ body:
...
-    validations:
-      required: true
 
   - type: dropdown
     id: os
     attributes:
       label: Operating system
       description: What OS are you using?
       options:
+        - Any
```
ckipp01, line 32: If we are going to add `Any` in here we might want to change the description. I think maybe we should just remove Any and make it optional. Wdyt?

dos65: turned it into optional

ckipp01 (approved):

PR #3475, opened 2022-01-05
[WIP] Migrate to 2.13

ckipp01:
When I saw https://github.com/scalacenter/bloop/pull/1646 I knew I needed to try this out. I was curious how much work would be needed to migrate to 2.13. This is at least publishable locally now. There are still a lot of warnings, some tests to figure out etc, but I figured I'd throw this up here to give you an idea of the diff to migrate to 2.13.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/NewProjectProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/NewProjectProvider.scala
@@ -4,6 +4,7 @@ import java.io.File
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+import scala.jdk.CollectionConverters._
 import scala.util.Try
 import scala.util.matching.Regex
 
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -17,8 +17,6 @@ import java.util.concurrent.TimeUnit
 import java.util.concurrent.TimeoutException
 
 import scala.annotation.tailrec
-import scala.collection.convert.DecorateAsJava
-import scala.collection.convert.DecorateAsScala
 import scala.collection.mutable
 import scala.compat.java8.FutureConverters
 import scala.concurrent.Await
@@ -26,6 +24,7 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.Promise
 import scala.concurrent.duration.FiniteDuration
+import scala.jdk.CollectionConverters._
 import scala.util.Properties
 import scala.util.Try
 import scala.util.control.NonFatal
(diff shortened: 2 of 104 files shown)


tests/input/src/main/scala/example/NamedArguments.scala
```diff
@@ -21,14 +21,14 @@ object NamedArguments {
     )
 
   // anonymous classes
-  @deprecated(
-    message = "a",
-    since = susan
-  ) def b = 1
+  //@deprecated(
+  //  message = "a",
+  //  since = susan
```
kpodsiad, line 26: I took a quick look, `final val susan = "Susan"` should help.

kpodsiad (commented):
Wow, can't wait to have 2.13 in Metals.

metals/src/main/scala/scala/meta/internal/builds/NewProjectProvider.scala
```diff
@@ -4,6 +4,7 @@ import java.io.File
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+import scala.jdk.CollectionConverters._
```
tgodzik, line 7: This was usually added via `MetalsEnrichments` and the `DecorateAsScala`/`DecorateAsJava` traits,  I think we can just switch to `AsJavaExtensions` and `AsScalaExtensions` instead.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -57,19 +56,11 @@ import org.eclipse.{lsp4j => l}
...
- *  import scala.compat.java8.FutureConverters._
- *  import scala.meta.internal.jdk.CollectionConverters._
- * }}}
- *
  * If this doesn't scale because we have too many unrelated extension methods
  * then we can split this up, but for now it's really convenient to have to
  * remember only one import.
  */
-object MetalsEnrichments
-    extends DecorateAsJava
```
tgodzik, line 70: > This was usually added via MetalsEnrichments and the DecorateAsScala/DecorateAsJava traits, I think we can just switch to AsJavaExtensions and AsScalaExtensions instead.

Actually here, we could just replace those I think.

https://www.scala-lang.org/api/current/scala/collection/convert/AsJavaExtensions.html

metals/src/main/scala/scala/meta/internal/builds/NewProjectProvider.scala
ckipp01: So I started out with that, but then I realized there was a lot of places that only brought in `MetalsEnrichments` just to use those instead of just using them directly. Do you still think we should do this?

ckipp01:
Again, until Bloop gets merged/publish I wouldn't spend a lot of time reviewing this. It was more of just an experiment to see what it'd take to get it to compile. So there is still quite a bit to do.

kpodsiad:
There are fresh artifacts on https://repo1.maven.org/maven2/ch/epfl/scala/bloop-launcher_2.13/ ;)

ckipp01:
Superseded by https://github.com/scalameta/metals/pull/3631

PR #3480, opened 2022-01-07
Add edit distance for Java files

tgodzik:
~~Previously, we would not try to adjust positions in java files in case of modifications. Now, we use the Java tokenizer for that purpose.~~

~~I translate the Java tokens to scalameta Token, which can alter be used in EditDistance class. There is no 100% compatibility of every possible token, but for token distance it's not that important. Even if we treat some keywords as identifiers they will still need to match on the exact string inside the identifiers.~~

~~This popped up while working on the rename for Java files.~~

Previously, we would not try to adjust positions in java files in case of modifications. Now, we use the Java tokenizer for that purpose.

I translate the Java and Scala tokens to custom BaseToken class, which can be used to find the proper edit distance.

Another issue that this cause is that it seems that MD5 string produced by the semanticdb for Java is different than the one we create in Metals. I have no idea why, but to accomodate it I copied over the Java implementation from sourcegraph/lsif-java and use it for Java files. That seems to have solved the issues.

This popped up while working on the rename for Java files.


metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
```diff
@@ -191,8 +191,7 @@ final class Diagnostics(
     val edit = TokenEditDistance(
       snapshot,
       current,
-      trees,
-      doNothingWhenUnchanged = false
```
tgodzik, line 195: No idea why would this be needed, the only thing it would do is to copy the range.

metals/src/main/scala/scala/meta/internal/parsing/TokenEditDistance.scala
```diff
@@ -323,30 +323,36 @@ object TokenEditDistance {
...
+      }
+      result.getOrElse(noMatch)
     } else {
       val result = for {
-        revised <- trees
-          .tokenized(revisedInput)
-          .toOption
-        original <- {
-          if (originalInput == revisedInput) Some(revised)
-          else Trees.defaultTokenizerDialect(originalInput).tokenize.toOption
```
tgodzik, line 345: This might have been a bug I haven't noticed before :fearful:

metals/src/main/scala/scala/meta/internal/parsing/JavaTokens.scala
```diff
@@ -0,0 +1,159 @@
...
+      buffer += create(
+        source,
+        dialects.Scala212,
+        start,
+        end,
+        translate(scanner.getRawTokenSource().mkString)
+      )
+    }
+
+    i match {
```
kpodsiad, line 78:
```suggestion
    import scala.annotation.switch
    (i: @switch) match {
```

metals/src/main/scala/scala/meta/internal/parsing/JavaTokens.scala
```diff
@@ -0,0 +1,159 @@
...
+    val array = buffer.toArray
+    Tokens(array, 0, array.size)
+  }
+
+  private def mapToken(
+      i: Int,
+      scanner: IScanner,
+      source: Input,
+      buffer: ArrayBuffer[Token]
+  ): Unit = {
```
kpodsiad, line 49: `getToken` should return token which will be added to the buffer in `loop`. This will simplify implementation of `token`, `keyword`, `constant` since they will be doing just one thing -- creating token.
```suggestion
  private def getToken(
      i: Int,
      scanner: IScanner,
      source: Input
  ): Token = {
```

metals/src/main/scala/scala/meta/internal/parsing/JavaTokens.scala
```diff
@@ -0,0 +1,159 @@
...
+    i match {
+      case ITerminalSymbols.TokenNameCOMMENT_BLOCK => token(Token.Comment.apply)
+      case ITerminalSymbols.TokenNameWHITESPACE =>
+        scanner.getCurrentTokenSource().foreach {
+          case '\n' => keyword(Token.LF.apply)
+          case '\r' => keyword(Token.CR.apply)
+          case '\f' => keyword(Token.FF.apply)
+          case '\t' => keyword(Token.Tab.apply)
+          case ' ' => keyword(Token.Space.apply)
+          case _ =>
```
kpodsiad, line 87: Can `ITerminalSymbols` be imported to shorten names? Also, what do you think about `// format: off` before pattern match? I find it almost unreadable if code blocks after `=>` are sometimes at the same line, sometimes at the newline. Having all of them at the same line as `=>` looks far better.

kpodsiad (approved):

dos65:
@tgodzik I'm wondering wouldn't it be easier to make some `JavaToken(idx: Int, text: String)` and implement `diff.myers.Equializer` for that rather than converting them into scalameta tokens?

Arthurm1:
I don't know what this is supposed to do so it's hard to review 😆 

Would not having this have caused VSCode to scroll down 1 line for every chararacter I typed in a Java file?  I got this the other day but couldn't reproduce.

tgodzik:
> I don't know what this is supposed to do so it's hard to review laughing
> 
> Would not having this have caused VSCode to scroll down 1 line for every chararacter I typed in a Java file? I got this the other day but couldn't reproduce.

Any time that we modify the Java file the semanticdb becomes obsolete, so all the possitions might be wrong. Edit distance use used to modify the semanticdb positions to work with the modified file in the buffers.

tgodzik:
> @tgodzik I'm wondering wouldn't it be easier to make some `JavaToken(idx: Int, text: String)` and implement `diff.myers.Equializer` for that rather than converting them into scalameta tokens?

We would need to rework the whole EditDistance class to take a generic Tokens trait, which is a bit more effort.  This is why I decided to just reuse the scalameta tokens as they would fit the purpose well enough and I wouldn't need to create a whole new hierarchy.

You proposal is probably better though :thinking: but it will require more effort.

Arthurm1:
> Any time that we modify the Java file the semanticdb becomes obsolete, so all the possitions might be wrong. Edit distance use used to modify the semanticdb positions to work with the modified file in the buffers.

Ahh - makes sense.  I'll try it out tomorrow

tgodzik:
@dos65 I reworked it to use a separate BaseToken, let me know what you think!

metals/src/main/scala/scala/meta/internal/parsing/JavaTokens.scala
```diff
@@ -0,0 +1,89 @@
...
+      i,
+      scanner.getCurrentTokenSource().mkString,
+      start,
+      end,
+      source,
+      isLF = false
+    )
+    i match {
+      case ITerminalSymbols.TokenNameWHITESPACE =>
+        scanner.getCurrentTokenSource().foreach {
```
tgodzik, line 76: Still decide to split these just in case.

mtags/src/main/scala/scala/meta/internal/mtags/MD5Java.java
```diff
@@ -0,0 +1,39 @@
...
+import java.nio.CharBuffer;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+/** Java semanticdb plugin returns a different md5 hashsum than the Scala one
+  * Copied over from:
+  * https://github.com/sourcegraph/lsif-java
+  */
+public final class MD5Java {
```
tgodzik, line 13: This is a fun one, so it seems the MD5 produced by the Java implementation and Scala one is different! Why? I cannot figure that out :scream:

metals/src/main/scala/scala/meta/internal/parsing/JavaTokens.scala
```diff
@@ -0,0 +1,89 @@
...
+      None
+    case Success(value) => Some(value)
+  }
+
+  private def mapToken(
+      i: Int,
+      scanner: IScanner,
+      source: Input,
+      buffer: ArrayBuffer[JavaToken]
+  ): Unit = {
```
kpodsiad, line 62: Could `mapToken` return a token instead of adding it to the buffer? It's always better to have fewer side effects.

metals/src/main/scala/scala/meta/internal/parsing/BaseToken.scala
```diff
@@ -0,0 +1,50 @@
...
+  def input: Input
+  def structure: String
+  def start: Int
+  def end: Int
+  def isLF: Boolean
+}
+
+case class ScalaToken(token: Token) extends BaseToken {
+
+  override def pos: Position = token.pos
```
kpodsiad, line 18: You could use `val` instead of def in overrides

mtags/src/main/scala/scala/meta/internal/mtags/MD5Java.java
kpodsiad: Could you check which implementation is correct using `md5sum` in the terminal?

kpodsiad (approved):

dos65 (commented):
@tgodzik Looks better!

The only I'm worried about is performance.
While I was investigating memory leaks I noticed that `TokenEditDistance` was the hottest execution path in Metals. Typing in editor constantly triggers a lot of lsp calls that leads to `TokenEditDistance` calculation.

I've mentioned one moment with Equalizer.

Also, to be completely safe I would introduce a typeclass to work with different type of token here that would allow to avoid mapping scalameta-token to a special one.
But I'm not sure about this suggestion. It's hard to say if this mapping affects significantly or not. We don't have benchmarks on this thing to compare.

metals/src/main/scala/scala/meta/internal/parsing/TokenEditDistance.scala
```diff
@@ -355,10 +363,16 @@ object TokenEditDistance {
...
+  private object TokenEqualizer extends Equalizer[BaseToken] {
+    override def equals(original: BaseToken, revised: BaseToken): Boolean =
+      (original, revised) match {
+        case (ScalaToken(org), ScalaToken(rev)) =>
+          org.productPrefix == rev.productPrefix &&
+            org.pos.text == rev.pos.text
+        case (org: JavaToken, rev: JavaToken) =>
+          org.id == rev.id && org.text == rev.text
+        case _ => false
+      }
```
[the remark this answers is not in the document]
tgodzik: I tried, but the Equalizer is invariant, which is causing  it not to compile :/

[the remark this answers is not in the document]
dos65: As I see it used only in `fromTokens`. 
I mean smth like that:
```scala
private def fromTokens[A <: BaseToken](
  original: Seq[A],
  revised: Seq[A],
  equalizer: TokenEqualizer[A]
): TokenEditDistance =
  ...
  
def apply(
      originalInput: Input.VirtualFile,
      revisedInput: Input.VirtualFile,
      trees: Trees,
      doNothingWhenUnchanged: Boolean = true
  ): TokenEditDistance = {
     .. 
     
     if (isJava) fromTokens(..., ..., JavaTokenEqualizer)
     else  fromTokens(..., ..., ScalaTokenEqualizer)
 }

private object ScalaTokenEqualizer extends Equalizer[Token] {
  ...
} 
private object JavaTokenEqualizer extends Equalizer[JavaToken] {
  ...
}
 
```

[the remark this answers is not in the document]
tgodzik: That works! But we still need to have the additional ScalaToken though :thinking:  So we need to wrap all the scalameta tokens. We could probably use structural types, but that feels a bit much :sweat_smile:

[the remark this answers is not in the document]
tgodzik: Or maybe it's not that bad? Take a look at the current state (unless I broke some more tests now :scream: )

metals/src/main/scala/scala/meta/internal/parsing/JavaTokens.scala
tgodzik: I wanted to make it more effective, since this is invoked a lot of the times.

mtags/src/main/scala/scala/meta/internal/mtags/MD5Java.java
tgodzik: Good idea, it seems it's the Scala one that returns the same MD5 string.

tgodzik: Reported upstream https://github.com/sourcegraph/lsif-java/issues/391

metals/src/main/scala/scala/meta/internal/parsing/JavaTokens.scala
tgodzik: Inlined this anyway so no more side effects from that function!

metals/src/main/scala/scala/meta/internal/parsing/BaseToken.scala
tgodzik: This is no longer needed, so removed it, but good idea. I used it in the JavaToken

Arthurm1:
This works so I can navigate from a scala reference to a java file definition even when the java file changes without saving.  Should it work with `textDocument/documentHighlight`?

This java file...
image

Goes to this on a newline without saving...
image

which looks like the semantic db info hasn't changed before saving.  Foo is incorrectly highlighted.  I wasn't sure how far this PR is supposed to help with that kind of thing.

metals/src/main/scala/scala/meta/internal/parsing/TokenEditDistance.scala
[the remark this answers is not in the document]
dos65: I never checked that by myself but as I know structural types are using reflection that adds overhead.

I think the best would be to introduce a type class for working with any type of token. But it would require more refactoring in `TokenEditDistance` but that way should keep the performance as it is.
Smth like:
```scala
trait TokenOps[A] {
  def pos(a: A): Position
  def input(a: A): Input
  def start(a: A): Int
  def end(a: A): Int
}

object TokenOps {

  implicit val scalaToken: TokenOps[Token] =
    new TokenOps[Token] {
      def pos(a: Token): Position = a.pos
      def input(a: Token): Input = a.input
      def start(a: Token): Int = a.start
      def end(a: Token): Int = a.end
    }

  implicit val javaToken: TokenOps[JavaToken] =
    new TokenOps[JavaToken] {
      def pos(a: JavaToken): Position = a.pos
      def input(a: JavaToken): Input = a.input
      def start(a: JavaToken): Int = a.start
      def end(a: JavaToken): Int = a.end
    }
}
```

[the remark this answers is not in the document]
tgodzik: Switched to type classes, though now I need to return `TokenDistance[Any]` since we can't have `TokenDistance[Token | JavaToken]` :smiling_face_with_tear:

tgodzik:
> This works so I can navigate from a scala reference to a java file definition even when the java file changes without saving. Should it work with `textDocument/documentHighlight`?
> 
> which looks like the semantic db info hasn't changed before saving. Foo is incorrectly highlighted. I wasn't sure how far this PR is supposed to help with that kind of thing.

That was a bug, I was using start position both for start and end. Fixed now.

kpodsiad (commented):
I had trouble with understanding the current implementation so I tried to do it on my own.
@tgodzik, @dos65 could you look into https://github.com/kpodsiad/metals/commit/7dbcb3a4fac66c8062cc020847f784ac77cf289d?
Maybe naming could be better, but the big picture should look like:
- `DistanceToken` provide a translation from arbitrary T to `EditDistanceToken`
- `EditDistanceToken` contains all necessary information from the token - thanks to that there is only one translation from T to `EditDistanceToken` - in `MatchingToken`.

dos65 (approved):
I think now it's ready to merge!

Arthurm1 (approved):
Works well for me thanks!

PR #3481, opened 2022-01-08
Cache build targets

Arthurm1:
Currently `CommonTarget`, `ScalacOptionsItem` and `JavacOptionsItem` are cached in `BuildTargets`

Nearly everywhere they are used they are converted to `ScalaTarget` and `JavaTarget` on the fly.  For `ScalaTarget` this calls `decodeJson` every time which feels expensive (during indexing it's called targetCount * sourceFileCount number of times and it's called all the time on UI changes e.g. hover etc.)

So instead `BuildTargets` now caches `BuildTarget`, `ScalaTarget` and `JavaTarget`.  I've got rid of `CommonTarget` as it's not needed.  This simplifies `BuildTargets` a bit.

My only worry is that is there ever a case where a target doesn't have a `ScalacOptionsItem` or a `JavacOptionsItem`?  I can't think of one since Metals requires SemanticDB plugin and that's got to be specified in the options or Metals won't know the `targetroot`.  I assume all BSP servers return `ScalacOptionsItem` but don't know for sure if there's some target type I'm not aware of.


Arthurm1:
This came to light when I was seeing why Kafka takes so long to index.  About 80s on Windows and 50s on WSL2 on my 10 year old machine.  Probably because indexing is nearly always (or always?) done on a cold JVM.  It's not a huge amount of time but it'd be nice to improve it.

I was wondering if there could be some optimisations here...
1) Bloop uses [JSoniter](https://github.com/plokhotnyuk/jsoniter-scala) for JSON - isn't that faster than [Gson](https://github.com/google/gson)?  Gson seems to use reflection vs Jsoniter's macros.
2) Could `MetalsLanguageServer#indexWorkspaceSources` be parallelized?  It seems to just update TrieMaps which are thread safe so an easy win.
3) `FileWatcher#initFileTreeRepository` also looks expensive (as it's running MurmerHash3 serially on every file) and could be run in parallel - although I don't know if the library allows that.  As an aside - does a file have to be hashed to check for differences - isn't the filesystems last updated date good enough?

tgodzik:
> My only worry is that is there ever a case where a target doesn't have a ScalacOptionsItem or a JavacOptionsItem? I can't think of one since Metals requires SemanticDB plugin and that's got to be specified in the options or Metals won't know the targetroot. I assume all BSP servers return ScalacOptionsItem but don't know for sure if there's some target type I'm not aware of.

I think all of them return at least ScalacOptionsItem.

> Bloop uses JSoniter for JSON - isn't that faster than Gson? Gson seems to use reflection vs Jsoniter's macros.

We use GSON because it's the default for LSP java API and we wanted to avoid additional dependencies. I don't think this is a problem, but we could try to test it out.

> Could MetalsLanguageServer#indexWorkspaceSources be parallelized? It seems to just update TrieMaps which are thread safe so an easy win.

Pretty sure it should be doable especially for working on different jars at the same time, but I never thought about this :sweat: . I already paralelize some stuff in ImplementationProvider and that helps a ton. 

> FileWatcher#initFileTreeRepository also looks expensive (as it's running MurmerHash3 serially on every file) and could be run in parallel - although I don't know if the library allows that. As an aside - does a file have to be hashed to check for differences - isn't the filesystems last updated date good enough?

Totally not sure, maybe we could ask in swoval what is their recomendation?

Olaf did a lot of optimisations early on and you can find it here https://scalameta.org/metals/blog/2019/01/22/bloom-filters
He used flamegraphs for that and we could check it based on kafka indexing.

PS: I will take a look at the PR next week, thanks a lot!

kpodsiad (approved):
Looks good!

tgodzik (approved):
Looks awesome! LGTM!

Arthurm1:
What's the etiquette on this - can I merge this myself now I have 2 approvals?

tgodzik:
> What's the etiquette on this - can I merge this myself now I have 2 approvals?

Yes! Please feel free to merge it :tada:

PR #3484, opened 2022-01-09
[Scala3] Fix autoimport with prefixes

### LINKED ISSUE #3478: SymbolPrefixes are messing up other util packages
### Describe the bug

When you have code that is in another util package and there is a conflicting import, the symbol prefix for `java/util` screws up your import.

You can reproduce this with the following snippet:
```scala
// using scala 3.1.0
// using lib org.scala-lang::scala3-compiler:3.1.0

import dotty.tools.dotc.interfaces.SourceFile

@main def Main: Unit =
// <-- try to complete `SourceFile` here from the util package
```

You'll end up with the following:

```scala
// using scala 3.1.0
// using lib org.scala-lang::scala3-compiler:3.1.0

import dotty.tools.dotc.interfaces.SourceFile
import dotty.tools.dotc.{util => ju}

@main def Main: Unit =
  ju.SourceFile
```

### Expected behavior

I'd expect the fully qualified name just be inserted in place or something like `util.SourceFile`

```scala
// using scala 3.1.0
// using lib org.scala-lang::scala3-compiler:3.1.0

import dotty.tools.dotc.interfaces.SourceFile
import dotty.tools.dotc.{util => ju}

@main def Main: Unit =
  dotty.tools.dotc.util.SourceFile
```

### Operating system

_No response_

### Editor/Extension

_No response_

### Version of Metals

0.10.9+295-ad8c7778-SNAPSHOT

My guess is the issue is right here:

https://github.com/scalameta/metals/blob/ad8c7778e75a48267d3723efe36ef0eccf4c6a6b/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala#L61-L66

I think in this case `util` is the simple name and the keys here are simple names so it incorrectly thinks the `util -> ju` entry matching when it shouldn't.

dos65:
Fixes #3478

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
@@ -58,18 +58,21 @@ object AutoImports:
     import indexedContext.ctx
 
     val importPos = autoImportPosition(pos, text, tree)
-    val renameConfig: Map[SimpleName, String] =
-      config.symbolPrefixes.asScala.map { (from, to) =>
+    val renameConfig: Map[Symbol, String] =
+      config.symbolPrefixes.asScala.flatMap { (from, to) =>
         val fullName = from.stripSuffix("/").replace("/", ".")
         val pkg = requiredPackage(fullName)
-        (pkg.name.toSimpleName, to.stripSuffix(".").stripSuffix("#"))
+        val rename = to.stripSuffix(".").stripSuffix("#")
+        List(pkg, pkg.moduleClass)
+          .filter(_ != NoSymbol)
+          .map((_, rename))
       }.toMap
 
     val renames =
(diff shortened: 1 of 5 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -135,7 +138,7 @@ object AutoImports:
         case IndexedContext.Result.Conflict =>
           val owner = symbol.owner
           val simpleName = owner.name.toSimpleName
```
tgodzik, line 140:
```suggestion
```
seems to be unsued now

tgodzik (approved):
Great work! Just a small comment, but it is already mergable as is.

ckipp01 (approved):
Nice! Thanks for fixing this @dos65! LGTM!

PR #3485, opened 2022-01-09
parallelize source file indexing

Arthurm1:
Index source files in parallel.

This implementation may be a little naive but as far as I tell `indexSourceFile` and `buildTargets.linkSourceFile` are thread safe as they only update `TrieMaps`.

Anecdotally - this drops my machine's time spent indexing Kafka from 52s to 32s and Metals from 23s to 14s.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -2215,16 +2215,24 @@ class MetalsLanguageServer(
   }
 
   private def indexWorkspaceSources(): Unit = {
+    case class SourceToIndex(
+        source: AbsolutePath,
+        sourceItem: AbsolutePath,
+        targets: Iterable[b.BuildTargetIdentifier]
+    )
+    val sourcesToIndex = mutable.ListBuffer.empty[SourceToIndex]
     for {
       (sourceItem, targets) <- buildTargets.sourceItemsToBuildTargets
       source <- sourceItem.listRecursive
       if source.isScalaOrJava
     } {
-      targets.asScala.foreach { target =>
-        buildTargets.linkSourceFile(target, source)
-      }
-      indexSourceFile(source, Some(sourceItem), targets.asScala.headOption)
+      targets.asScala.foreach(buildTargets.linkSourceFile(_, source))
+      sourcesToIndex += SourceToIndex(source, sourceItem, targets.asScala)
     }
+
+    sourcesToIndex.par.foreach(f =>
+      indexSourceFile(f.source, Some(f.sourceItem), f.targets.headOption)
+    )
   }
 
   private def reindexWorkspaceSources(
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -7,6 +7,7 @@ import java.nio.file._
 import java.util
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.Executors
+import java.util.concurrent.ForkJoinPool
 import java.util.concurrent.ScheduledExecutorService
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -2215,15 +2215,21 @@ class MetalsLanguageServer(
...
       if source.isScalaOrJava
-    } {
-      targets.asScala.foreach { target =>
-        buildTargets.linkSourceFile(target, source)
-      }
-      indexSourceFile(source, Some(sourceItem), targets.asScala.headOption)
+    } { sourcesToIndex += SourceToIndex(source, sourceItem, targets.asScala) }
+
+    sourcesToIndex.par.foreach { f =>
+      f.targets.foreach(buildTargets.linkSourceFile(_, f.source))
```
kpodsiad, line 2231: If `buildTargets.linkSourceFile` updates `ConcurrentHashSet` then doing it in parallel will have any effect? Threads will be blocked and there won't be any benefits from `par`.

I wonder what results will look like if we extract `f.targets.foreach(buildTargets.linkSourceFile(_, f.source))` to separate `foreach`.

When you mentioned Kafka, did you refer to the original repo? I want to do some experiments on my own.

kpodsiad (commented):
Thanks @Arthurm1 for thinking about performance!
It looks good, I left a small comment.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
Arthurm1: Good point - I should probably move that back to the first for loop.
I'm using kafka from [here](https://github.com/apache/kafka).  The new "Start Metals" button is very handy for repeatedly checking out indexing speed 😄

kpodsiad: Got the results from my machine (MBP 2019).
before -> from this branch -> after moving back `buildTargets.linkSourceFile` to the first loop
Kafka: 25s -> 13s -> 12s
Metals: 15s -> 12s -> 11s
Metals (second attempt): 12s -> 8s -> 8s

kpodsiad (approved):
Works well!

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
Arthurm1: Looks like I need a new machine

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -2215,16 +2215,24 @@ class MetalsLanguageServer(
...
     } {
-      targets.asScala.foreach { target =>
-        buildTargets.linkSourceFile(target, source)
-      }
-      indexSourceFile(source, Some(sourceItem), targets.asScala.headOption)
+      targets.asScala.foreach(buildTargets.linkSourceFile(_, source))
+      sourcesToIndex += SourceToIndex(source, sourceItem, targets.asScala)
     }
+
+    sourcesToIndex.par.foreach(f =>
```
tgodzik, line 2233: @kpodsiad mentioned that this way indexing is using up all the cores of the machine. I wonder whether this will be ok for the users? Is it possible to work on other things in the meantime? Or should we reduce paralelization to half the processors?

Arthurm1: I've limited it to cores/2 which I think should be ok as I think this should be mostly File IO limited.

@kpodsiad Could you check on your machine again if this is easier on cores without losing performance.  I find looking at core usage on WSL doesn't seem to work very well

kpodsiad: @Arthurm1 cores/2 case is much lighter in terms of CPU load (no surprise - half of the cores are free) and indexing duration isn't much longer.
1 core -> cores/2-> all cores
kafka: 22s -> 14s -> 12s

I'm ok with using all cores since indexing is taking only a few seconds, but for more conservative users cores/2 will be better.

Arthurm1: Thanks for testing

tgodzik (approved):
LGTM!

PR #3489, opened 2022-01-11
fix: override pprint in docs

ckipp01:
I have no idea how this has been working, since locally I couldn't run this for a while it seems, but it still seems to be working in ci 🤷🏼

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -808,7 +808,8 @@ lazy val docs = project
     sharedSettings,
     publish / skip := true,
     moduleName := "metals-docs",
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM! Does it work locally for you now?

ckipp01:
> LGTM! Does it work locally for you now?

Yes, it does. 👍🏼 thanks for the pointer

tgodzik:
I've spent a while trying to fix it in mdoc, but it didn't work with different versions in docs and mdoc-code :/

PR #3490, opened 2022-01-11
docs: update user config docs

ckipp01:
This updates the default fallbackScalaVersion which is Scala3, but we
are still showing Scala 2.12. This also updates the `testUserInface` to
only have the value since it's dynamically put together. Right now it
shows:
```json
{
  "metals": {
    "testUserInterface": { "testUserInterface" : "Test explorer" }
  }
}
```
So this fixes that.


tgodzik (approved):
:rocket:

PR #3498, opened 2022-01-12
Remove workaround for the wrong Java semanticdb md5 hash

tgodzik:


dos65 (approved):
LGTM!


Also, requires #3499 as I hardcoded it for interactive.
I didn't know that we send just a version of this plugin over BSP.
Does it mean that java support works only with bloop?

PR #3505, opened 2022-01-13
[JavaInteractiveSemanticdb] Fix Jdk version parsing

dos65:
The same thing as [in bloop](https://github.com/scalacenter/bloop/pull/1657)


tgodzik (approved):
Thanks for spotting this!

Issue #3501, opened 2022-01-13
Typing in VSCode always executes command to refresh test-exporer


dos65:
### Describe the bug

Reproduction step:
- Enable: `Files: Auto Save: afterDelay`
- Start watching traces: `tail -f ~/.cache/metals/lsp.trace.json`
- Open any file and start typing with a delay (one char per sec)
- Notice that CPU usage gets higher
- Notice  `window/executeCommand` with the following result constantly appears:
  ```
  [Trace - 00:45:09 PM] Sending response 'workspace/executeCommand - (268)'. Processing request took 1000ms
  Result: [
    {
    "targetName": "unit-test",
    "targetUri": "file:/home/dos65/projects/metals/metals/tests/unit/?id\u003dunit-test",
    "discovered": [
      {
        "prefix": "tests",
        "children": [
          {
            "fullyQualifiedName": "tests.AnalyzeStacktraceLspSuite",
            "className": "AnalyzeStacktraceLspSuite",
            "location": {
              "uri": "file:///home/dos65/projects/metals/metals/tests/unit/src/test/scala/tests/AnalyzeStacktraceLspSuite.scala",
              "range": {
                "start": {
                  "line": 2,
                  "character": 6
                },
                "end": {
                  "line": 2,
                  "character": 31
                }
              }
            },
            "kind": "suite"
          },

  ```

### Expected behavior

It seems that at least this command shouldn't be executed if testing tab isn't opened but I think there might be some additional optimizations.

### Operating system

Linux

### Editor/Extension

VS Code

### Version of Metals

v0.11.0

kpodsiad:
Is this happening with `Code Lenses` too? Refreshing tests via Text Explorer shouldn't be much more expensive since they both share the same mechanism - tests are refreshed after every compilation.

dos65:
@kpodsiad nope. I set explorer to `Code Lenses` and cpu spikes disappeared. I can look at it more detailed at it a bit later.

tgodzik:
Do you have autosave on ? I can't seem to reproduce it :thinking:

dos65:
@tgodzik 
> Do you have autosave on ?

Yes

vishalovercome:
Probably this is happening under both Code Lens as well as Test Explorer (explained in the thread above)

Issue #3509, opened 2022-01-14
Duplicated symbols in `BuildTargetClasses`'s test classes


kpodsiad:
### Describe the bug

Not sure if it's even a bug, but I'm creating an issue to not forget about it and I'll try to get back to it later.

For every test suite `a.MunitTestSuite` Metals will create 2 test entries, but only one of them has a location in the source code.
```
"a/MunitTestSuite., a.MunitTestSuite"
"a/MunitTestSuite#, a.MunitTestSuite"
```

Here code responsible for that:
https://github.com/scalameta/metals/blob/5898e3b5bd4d56a4e930e37c5cb3fba8366af055/metals/src/main/scala/scala/meta/internal/metals/debug/BuildTargetClasses.scala#L114-L115

### Expected behavior

`testClasses` in `BuildTargetClasses` shouldn't have duplicated entries

### Version of Metals

v0.11.0

kpodsiad:
It's not a bug. At this moment Metals don't know if the class name refers to the class (e.g. munit or junit) or object (e.g. utest).

PR #3510, opened 2022-01-14
Fix test explorer performance issues

kpodsiad:
fixes https://github.com/scalameta/metals/issues/3501

Previously, due to lack of caching and https://github.com/scalameta/metals/issues/3509 discovery request were able to take over 2 seconds in Metals repo.

Now, only the first request takes a significant amount of time (due to https://github.com/scalameta/metals/issues/3509 additional lookups for non-existing symbols may happen) and every subsequent request takes constant, close to zero time.

```shell
2022.01.14 14:21:18 INFO  Cache refresh took: 1752ms
2022.01.14 14:21:25 INFO  Cache refresh took: 1ms
2022.01.14 14:21:25 INFO  Cache refresh took: 1ms
```

From what I observed this [code](https://github.com/scalameta/metals/blob/5898e3b5bd4d56a4e930e37c5cb3fba8366af055/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala#L220-L223) was calling refresh a few times within less than a second at the start of Metals. In order to avoid problems with unnecessary computing, I used a single-thread executor. After first run which populates caches, subsequent ones don't have much work. The executor also ensures that requests from the client will always be responded with the newest data. Maybe someone has some ideas on how to improve it?

There's one corner case that comes into my mind - if someone will rename a file without renaming the test suite then the wrong location will be cached for that test suite. I'll try to do something in this scenario.


metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesFinder.scala
```diff
@@ -1,65 +1,137 @@
...
+
+    val time0 = System.currentTimeMillis()
+    for {
+      SymbolsPerTarget(buildTarget, testSymbols) <- symbolsPerTarget
+      (symbol, fullyQualifiedClassName) <- testSymbols.readOnlySnapshot().toList
+      // IMPORTANT this check is meant to check for class name, not a symbol
+      if !cachedTestSuites.contains(fullyQualifiedClassName)
+    } computeTestEntry(buildTarget, symbol, fullyQualifiedClassName)
+    val time1 = System.currentTimeMillis()
+    scribe.info(s"Cache refresh took: ${time1 - time0}ms")
```
kpodsiad, line 108: Left on purpose

tgodzik: That might happen a lot, no need to leave it in since the same can be accomplished by using trace.

kpodsiad: I meant it's left on purpose for debugging reasons

tgodzik: Ok, so the plan is to remove it before the merge?

kpodsiad: Yup

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -0,0 +1,167 @@
...
+   */
+  private val cachedTestSuites =
+    mutable.Map[FullyQualifiedClassName, TestEntry]()
+
+  /**
+   * Retrieves cached test suites.
+   */
+  def findTestSuites()
+      : CompletableFuture[java.util.List[TestSuiteDiscoveryResult]] =
+    CompletableFuture.supplyAsync(
```
tgodzik, line 55: Is this needed? Can we instead use TrieMap and handle this normally?

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -0,0 +1,167 @@
...
+    )
+
+  /**
+   * Find test suites for all build targets in current projects and update caches.
+   */
+  def refreshTestSuites(): Unit = {
+    if (clientConfig.isTestExplorerProvider()) refreshTestSuitesImpl()
+  }
+
+  def refreshTestSuitesImpl(): Unit = executor.execute { () =>
```
tgodzik, line 80:
```suggestion
  def refreshTestSuites(): Unit = if (clientConfig.isTestExplorerProvider()) { () =>
```
with TrieMap?

tgodzik (changes requested):
Looks good, but I would not use the executor here.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
kpodsiad: I used it primarily in order to avoid parallel updating of cache and it's not about synchronization, it's about duplicating work but `BatchedFunction` solves this. 
The secondary goal was to ensure that `refresh` will happen always after the end of the cache update. A situation like below could happen and test explorer won't have all of the test suites. However, I think it's not a huge issue since refresh happens very often and I already have a plan for switching approach from pull-based to push-based.
```shell
TestSuitesProvider.scala:96 "start": "start"
TestSuitesProvider.scala:56 "refresh": "refresh"
TestSuitesProvider.scala:105 "end": "end"
```

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -0,0 +1,165 @@
...
+    buildTargets: BuildTargets,
+    buildTargetClasses: BuildTargetClasses,
+    definitionProvider: DefinitionProvider,
+    clientConfig: ClientConfiguration
+)(implicit ec: ExecutionContext) {
+
+  /**
+   * Cached, already discovered test suites.
+   * Cache is updated via refreshTestSuites method called after compilation is finished.
+   * There is no need to have synchronization since all access is available via
```
tgodzik, line 36: This is no longer true though?

tgodzik (approved):
LGTM! Thanks for the quick fix!

We can fix the comments later.

PR #3512, opened 2022-01-14
Use ArraySeq instead of List for Diffutils

tgodzik:
I haven't noticed that `toList` would previously do it on `ArraySeq`, which meant we had an Array under the java.List interface. It seems it might be causing perfomance regression in large inputs.


kpodsiad (approved):

tgodzik:
We can confirm the fix with the users, but it seems to fix the issue for me. I was super confused by the use of List there, which I thought meant that we didn't worry that much about performance in DiffUtils :sweat:

PR #3523, opened 2022-01-15
Call javap with -private flag

durban:
In the "Metals Analyse Source" -> "Show decompiled with javap" and "... with javap verbose" currently `javap` is called without the `-private` option. This causes some missing members in the output. This PR adds the `-private` option to the `javap` calls.


kpodsiad (approved):
Thanks for contributing! Could you also add/modify existing testcase [here](https://github.com/scalameta/metals/blob/c77d9714b48c148d2813b8c4da3c05756869dd39/tests/slow/src/test/scala/tests/feature/FileDecoderProviderLspSuite.scala)?

durban:
Yeah, done. (I was looking for an existing test, but failed to find it. :-)

kpodsiad:
@tgodzik tests are failing on windows, your powers are needed to merge this ;)

PR #3529, opened 2022-01-17
Revert scribe update

tgodzik:


tgodzik:
It might be possible that flyway needs to reverted also.

PR #3538, opened 2022-01-17
[JavaInteractiveSemanticdb] Add test-case to cover --patch-module flag usage

dos65:
Just an additional test that doesn't work on jdk11+ without `--patch-module $module` flag


tgodzik (approved):
:rocket:

Arthurm1:
@dos65 Thanks - that was a great help.

PR #3540, opened 2022-01-17
fix: Open other window when using Analyze Stacktrace

tgodzik:
When reworking the links and adding support for Sublime I forgot to test out if the stacktrace opens files in another window and it turns out this got broken.

Because of that, I reworked the GotoLocation command to take in an additional parameter, which makes it a bit easier to handle and always send it tpo the client.

Needs https://github.com/scalameta/metals-vscode/pull/828


tgodzik:
@ayoub-benali
Can we update the command used by the links in Sublime and also https://github.com/scalameta/metals-sublime/blob/3105cf31820d73c9214c802353a7e036f8d9118e/core/handle_execute_client.py#L19 ? 

Will the additional parameter be a problem? 

Sorry I haven't done this before, but I should have done it when we worked on the links. I dropped the additional parameter by mistake.

kpodsiad (approved):

ckipp01 (approved):
👍🏼 LGTM

PR #3543, opened 2022-01-18
[Actions] Install coursier for nightly checks

dos65:
I missed that `./coursier` is in `.gitignore`

### DIFF
--- a/.github/workflows/check_scala3_nightly.yml
+++ b/.github/workflows/check_scala3_nightly.yml
@@ -7,10 +7,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: olafurpg/setup-scala@v13
+      - uses: coursier/setup-action@v1
+        with:
+          apps: ""
       - name: "Query maven-central"
         run: |
-          TODAY_NIGHTLY=$(./coursier complete org.scala-lang:scala3-compiler_3: | grep $(date '+%Y%m%d'))
+          TODAY_NIGHTLY=$(cs complete org.scala-lang:scala3-compiler_3: | grep $(date '+%Y%m%d'))
           if [ ! -z $TODAY_NIGHTLY ]; then
             gh workflow run mtags-auto-release.yml -f scala_version=$TODAY_NIGHTLY
           fi


.github/workflows/check_scala3_nightly.yml
```diff
@@ -10,7 +10,9 @@ jobs:
       - uses: olafurpg/setup-scala@v13
       - name: "Query maven-central"
         run: |
-          TODAY_NIGHTLY=$(./coursier complete org.scala-lang:scala3-compiler_3: | grep $(date '+%Y%m%d'))
+          curl -fLo cs https://git.io/coursier-cli-"$(uname | tr LD ld)"
```
kpodsiad, line 13: I thought that `coursier` comes `olafurpg/setup-scala@v13`. If not, could you remove it (`- uses: olafurpg/setup-scala@v13`)?

ckipp01: Actually, instead of using setup-scala here and then manually installing coursier you could also just consider using https://github.com/coursier/setup-action, which will still ensure you have everything you need to use coursier, without any extra work.

dos65: switched it on `coursier/setup-action`

.github/workflows/check_scala3_nightly.yml
```diff
@@ -7,10 +7,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: olafurpg/setup-scala@v13
+      - uses: coursier/setup-action@v1
+        with:
+          apps: ""
```
kpodsiad, line 12: Is this needed?

kpodsiad (approved):
LGTM!

.github/workflows/check_scala3_nightly.yml
ckipp01: I don't think so. I use this here as well https://github.com/scalameta/nvim-metals/blob/d5864d9395152d4820bcb8a66f1235c5970e46a5/.github/workflows/ci.yml#L63-L65 and don't need to have that.

dos65: the default value of apps is `sbtn` but there is no need to install it as for this job only jvm and cs is required

ckipp01: Ahh cool! I didn't know that.

PR #3544, opened 2022-01-18
refactor: check lowercase test-user-interface

ckipp01:
I've started to play around with this in nvim-metals and just hit
on this and realized in the docs we say Test explorer as an example
but then we check on an exact match of Test Explorer. This just
calls .toLowerCase on the value and then we check that.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
@@ -252,7 +252,7 @@ object UserConfiguration {
       UserConfigurationOption(
         "test-user-interface",
         "Code Lenses",
-        "Test explorer",
+        "test explorer",
         "Test UI used for tests and test suites",
         "Default way of handling tests and test suites."
       ),
@@ -463,9 +463,9 @@ object UserConfiguration {
     val defaultScalaVersion =
       getStringKey("fallback-scala-version").filter(_ != "automatic")
     val disableTestCodeLenses = {
-      val isTextExplorerEnabled = clientConfiguration.isTestExplorerProvider()
-      getStringKey("test-user-interface") match {
-        case Some("Test Explorer") if isTextExplorerEnabled =>
+      val isTestExplorerEnabled = clientConfiguration.isTestExplorerProvider()
+      getStringKey("test-user-interface").map(_.toLowerCase()) match {
+        case Some("test explorer") if isTestExplorerEnabled =>
           TestUserInterfaceKind.TestExplorer
         case _ =>
           TestUserInterfaceKind.CodeLenses
--- a/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
@@ -39,7 +39,7 @@ import org.eclipse.{lsp4j => l}
  * @param slowTaskProvider if the client implements `metals/slowTask`.
  * @param statusBarProvider if the client implements `metals/status`.
  * @param treeViewProvider if the client implements the Metals Tree View Protocol.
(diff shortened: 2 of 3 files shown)


kpodsiad:
@ckipp01 FYI, I'm trying to refactor Test Explorer to a more push-based approach and I don't know yet how many things it will break :/

After you get familiar with the current state, it'll be great to hear what this looks like from the other editor perspective and how you want to tackle this.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -464,8 +464,8 @@ object UserConfiguration {
       getStringKey("fallback-scala-version").filter(_ != "automatic")
     val disableTestCodeLenses = {
       val isTextExplorerEnabled = clientConfiguration.isTestExplorerProvider()
-      getStringKey("test-user-interface") match {
-        case Some("Test Explorer") if isTextExplorerEnabled =>
+      getStringKey("test-user-interface").map(_.toLowerCase()) match {
+        case Some("test explorer") if isTextExplorerEnabled =>
```
kpodsiad, line 468:
```suggestion
        case Some("test explorer") if isTestExplorerEnabled =>
```
I committed a lot of `text` typos :c

kpodsiad (approved):

ckipp01:
> @ckipp01 FYI, I'm trying to refactor Test Explorer to a more push-based approach and I don't know yet how many things it will break :/
> 
> After you get familiar with the current state, it'll be great to hear what this looks like from the other editor perspective and how you want to tackle this.

ahhh, so I'm assuming that means breaking changes to how it works then? With the current state the flow actually seems pretty nice for what I'm playing around with in `nvim-metals`. There are two different things that I'm playing around with but both will pretty much follow the same pattern. I have the details outlined here: https://github.com/scalameta/nvim-metals/issues/324.

> I'm trying to refactor Test Explorer to a more push-based approach

I'm assuming this is due to the performance issues that people have been talking about since VS Code is firing for this all the time? Either approach would _probably_ work fine, and I think only updating the tests if something actually changes would be a good thing. It'd be great to cache the suites that we've sent and then only push them again to the client if the suites have changed instead of just spamming them all the time.

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
ckipp01: fixed.

kpodsiad:
> ahhh, so I'm assuming that means breaking changes to how it works then? With the current state the flow actually seems pretty nice for what I'm playing around with in nvim-metals. 

Not necessary, I guess the current API could stay if it's convenient for nvim.
My biggest pain is the fact that test refresh is done after compilation is finished and a few first requests from the client can retrieve the partial results. 

> I'm assuming this is due to the performance issues that people have been talking about since VS Code is firing for this all the time? Either approach would probably work fine, and I think only updating the tests if something actually changes would be a good thing. It'd be great to cache the suites that we've sent and then only push them again to the client if the suites have changed instead of just spamming them all the time.

Spamming them all the time is actually very cheap and easy, mostly easy :D Push based approach is handy when we want to switch with results from coarsed (suites) to fine-grained (suites + test cases). There can be a lot of test cases and they have to be discovered lazily, for that push is needed.

ckipp01:
> Not necessary, I guess the current API could stay if it's convenient for nvim.

Meh, I'd say let's try to favor correctness and speed over convenience for nvim 😆 

> Spamming them all the time is actually very cheap and easy, mostly easy :D Push based approach is handy when we want to switch with results from coarsed (suites) to fine-grained (suites + test cases). There can be a lot of test cases and they have to be discovered lazily, for that push is needed.

Sure, I still think though that spamming the same thing over and over isn't really necessary. It just produces a ton of noise.

Issue #3539, opened 2022-01-17
Using Ammonite constantly logs out "could not load snapshot"


ckipp01:
### Describe the bug

Create a single `test.sc` script in a directory. Open it up with Metals and tell it to import it as an Ammonite script. Start moving around. You'll see this in your logs.

```
2022.01.17 17:39:51 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:51 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:51 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:52 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:52 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:52 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:53 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:53 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:53 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:53 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:54 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:54 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:54 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:55 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
2022.01.17 17:39:55 WARN  Could not load snapshot text for /Users/ckipp/Documents/scala-workspace/amm-test/test.sc
```

### Expected behavior

[...]

PR #3545, opened 2022-01-18
Compare only uppercase MD5

tgodzik:
Previously, when semanticdb had md5 with lower case letters we would not find the snapshot text for semanticdb. Now, we convert it always to upper case.

Fixes https://github.com/scalameta/metals/issues/3539

PS: So the logging might be a bit excesive, but it did help to find an issue :man_dancing:


kpodsiad (approved):
because semanticDB stores md5 as an upper case :/
Awesome catch @tgodzik

ckipp01 (approved):
Amazing 😆 LGTM!

PR #3546, opened 2022-01-18
docs: Add docs for discover-test-suites

ckipp01:
There is currently no mention of `discover-test-suites` in the docs.


kpodsiad (approved):
I promise to remember about docs in the future!

PR #3549, opened 2022-01-19
docs: add a small note about decode not working on extra file types.

ckipp01:
This also improves the message that a user sees if they accidentally try
to use one of these while in a worksheet or something. I hit on this
today and tried using cfr quickly while in a worksheet and all I saw as
an error that said:

```
Invalid extension
```


tgodzik (approved):
LGTM!

PR #3550, opened 2022-01-19
Refactor Ammonite support…

alexarchambault:
…to prepare the ground for Scala CLI-specific features.

This PR makes some of the Ammonite-related features more generic, so that these can be more easily re-used by upcoming Scala CLI changes. This requires some refactoring, done in the first commits.

In more details, it:
- moves indexing-related methods to a new `Indexer` class
- manages bits of data kept in `BuildTargets` by BSP connection (rather than in a single place for all BSP connections at once), which later allows to pass the Ammonite BSP connection (and later the Scala CLI one) to `BuildTargets`, which in turn would allow Metals to find the main and test classes of Ammonite (which Ammonite doesn't provide yet, but Scala CLI does)
- manages sources that need to be wrapped (before being passed to the interactive compiler) in a dedicated class, `SourceMapper`

Then, it proceeds to make less classes tap directly into the `Ammonite` instance:
- the `ForwardingMetalsBuildClient` first,
- `BuildTargets` next,
- then `SourceMapper`.

Lastly, the last 2 commits:
- tweak minor things in the `Ammonite` class
- pave the ground for non-Ammonite scripts

About the `BuildTargets` refactoring, note that it would later on allow to make most indices per-BSP-connection (rather than globally shared), with some more work. This would allow to refresh only some specific connection indices, rather than refreshing indices for all connections when only one of them changed. As of now, because of this limitation, it's quite hard to work on Ammonite scripts (and soon Scala CLI sources) alongside a bigger sbt or Mill-based project, as adding a dependency in Ammonite or Scala CLI makes Metals re-index the sbt or Mill project too, which can be costly.


metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -642,4 +511,310 @@ object BuildTargets {
...
+
+    def allWorkspaceJars: Iterator[AbsolutePath]
+
+    def addSourceItem(
+        sourceItem: AbsolutePath,
+        buildTarget: BuildTargetIdentifier
+    ): Unit
+  }
+
+  final case class DataSeq(head: WritableData, tail: List[WritableData]) {
```
ckipp01, line 562: Maybe a silly question, but I'm assuming this is mainly for the utilities that you have in here like `fromIterators` and `fromOptions`, but why not just have `DataSeq` be a wrapper around `List[WritableData]`? Why have a head and the tail here?

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -379,17 +255,20 @@ final class BuildTargets(
...
-            allBuildTargetIds.find { id =>
+            allBuildTargetIdsInternal.find { case (_, id) =>
               targetJarClasspath(id).exists(_.contains(jar))
             }
           }
-        fromJar.foreach(addSourceItem(source, _))
-        fromJar
+        for ((data0, id) <- fromJar)
+          data0.addSourceItem(source, id)
+        fromJar.map(_._2)
```
ckipp01, line 266:
```suggestion
        fromJar.map { case (data, id) =>
          data.addSourceItem(source, id)
          id
        }
```
Unless you don't want to mix these, we could just iterate through them just once. Take it or leave it.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -642,4 +511,310 @@ object BuildTargets {
...
+
+  object Data {
+    def create(): WritableData =
+      new DataImpl
+  }
+
+  private final class DataImpl extends WritableData {
+
+    val sourceItemsToBuildTarget
+        : TrieMap[AbsolutePath, ConcurrentLinkedQueue[BuildTargetIdentifier]] =
```
ckipp01, line 599: Side grumble, I wish we could change our scalafmt config to not allow breaking on `:` like this.

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
```diff
@@ -46,8 +45,7 @@ final case class Indexer(
     timerProvider: TimerProvider,
     scalafixProvider: () => ScalafixProvider,
     indexingPromise: Promise[Unit],
-    ammonite: () => Ammonite,
-    lastImportedBuilds: () => List[ImportedBuild],
+    buildData: () => Seq[(String, BuildTargets.WritableData, ImportedBuild)],
```
ckipp01, line 48: This is pretty hard to grok at first sight, especially when you create the `Indexer` with
```scala
    () =>
      Seq(
        (
          "main",
          mainBuildTargetsData,
          ImportedBuild.fromList(lastImportedBuilds)
        ),
        ("ammonite", ammonite.buildTargetsData, ammonite.lastImportedBuild)
      ),
```

Would it be possible to instead having a tuple here to create some sort of case class or something for this?

ckipp01 (commented):
Thanks for all the work on this @alexarchambault! Mostly just questions from me at this point.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -167,7 +160,9 @@ class MetalsLanguageServer(
   val excludedPackageHandler: ExcludedPackagesHandler =
     new ExcludedPackagesHandler(userConfig.excludedPackages)
   var ammonite: Ammonite = _
-  val buildTargets: BuildTargets = BuildTargets.withAmmonite(() => ammonite)
+  private val mainBuildTargetsData = BuildTargets.Data.create()
```
tgodzik, line 163: Why do we need to create empty data here? Could we not create it when indexing?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -421,17 +423,30 @@ class MetalsLanguageServer(
...
-          () => treeView,
-          () => worksheetProvider,
-          () => ammonite
+          onBuildTargetDidChangeFunc = { params =>
+            val (ammoniteChanges, otherChanges) =
+              params.getChanges.asScala.partition(
+                _.getTarget.getUri.isAmmoniteScript
+              )
+            if (ammoniteChanges.nonEmpty)
+              ammonite.importBuild().onComplete {
```
tgodzik, line 436: Could we move it to quickConnectToBuildServer (or a new method that does it?)

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -771,6 +786,7 @@ class MetalsLanguageServer(
             scalaVersionSelector
           )
         )
+        buildTargets.addData(ammonite.buildTargetsData)
```
tgodzik, line 789: Same question here, could we add ammonite data when we make the connection to ammonite?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1341,7 +1357,7 @@ class MetalsLanguageServer(
     ) {
       event.eventType match {
         case EventType.Create =>
-          buildTargets.onCreate(path)
+          mainBuildTargetsData.onCreate(path)
```
tgodzik, line 1360: Could we instead of differentiating the different connections, just find targetData this path might belong to?

tgodzik (commented):
Thanks for the extensive PR! I left a bunch of comments, mostly me trying to figure out if we can make the code nicer to maintain.

I think there is a bunch of improvements we can make, probably most would not be a part of this PR, but hopefully we can improve this one a bit.

Do ask questions if something doesn't make sense :sweat_smile:

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
```diff
@@ -0,0 +1,453 @@
...
+        scribe.warn("No build session currently active to reload.")
+        Future.successful(BuildChange.None)
+      case Some(session) if forceRefresh => reloadAndIndex(session)
+      case Some(session) =>
+        workspaceReload().oldReloadResult(checksum) match {
+          case Some(status) =>
+            scribe.info(s"Skipping reload with status '${status.name}'")
+            Future.successful(BuildChange.None)
+          case None =>
+            synchronized {
```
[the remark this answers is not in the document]
alexarchambault: The `synchronized` wraps a `Future`, so now that I think about it, it doesn't really make sense to me either…

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
```diff
@@ -0,0 +1,453 @@
...
+import scala.meta.internal.metals.watcher.FileWatcher
+import scala.meta.internal.mtags.OnDemandSymbolIndex
+import scala.meta.internal.semanticdb.Scala._
+import scala.meta.internal.tvp.TreeViewProvider
+import scala.meta.internal.worksheets.WorksheetProvider
+import scala.meta.io.AbsolutePath
+
+import ch.epfl.scala.{bsp4j => b}
+
+final case class Indexer(
```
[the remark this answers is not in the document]
alexarchambault: Sure.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
alexarchambault: Seems I wanted to ensure it's non-empty, although I can't recall why… and it seems it's not needed.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
alexarchambault: Sure. I just saw it's just an `Option`, so it shouldn't change much things.

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
alexarchambault: Sure!

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -642,4 +498,341 @@ object BuildTargets {
       leaves: collection.Set[BuildTargetIdentifier]
   )
 
+  trait Data {
```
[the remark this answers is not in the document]
alexarchambault: Sure!

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -642,4 +498,341 @@ object BuildTargets {
...
+    def writableDataIterator: Iterator[WritableData] = (head :: tail).iterator
+    def iterable: Iterable[Data] = list.toIterable
+
+    def fromIterators[T](f: Data => Iterator[T]): Iterator[T] =
+      iterator.flatMap(f)
+    def fromOptions[T](f: Data => Option[T]): Option[T] =
+      fromIterators(f(_).iterator).take(1).toList.headOption
+  }
+
+  trait WritableData extends Data {
```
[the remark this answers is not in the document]
alexarchambault: Right, looking at it again, it seems it doesn't make sense anymore to separate `Data` and `WritableData`. I think I wanted to pass the read-only `Data` around when possible, but it seems `WritableData` actually needs to be passed around most of the time…

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -410,7 +291,8 @@ final class BuildTargets(
   ): Option[BuildTargetIdentifier] = {
     val targetMetaBuildDir =
       if (file.isSbt) file.parent.resolve("project") else file.parent
-    buildTargetInfo.values
+    data
+      .fromIterators(_.buildTargetInfo.valuesIterator)
```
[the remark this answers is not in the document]
alexarchambault: Do you mean, `data` should be public and users should call `buildTargets.data.sbtBuildScalaTarget` for example? Or that we should have methods on `BuildTargets` just calling the ones on `DataSeq`, like
```scala
def sbtBuildScalaTarget(file: AbsolutePath): Option[BuildTargetIdentifier] =
  data.sbtBuildScalaTarget(file)
```
That would be many such "proxy" methods.

metals/src/main/scala/scala/meta/internal/metals/ForwardingMetalsBuildClient.scala
```diff
@@ -46,12 +41,9 @@ final class ForwardingMetalsBuildClient(
...
     time: Time,
     didCompile: CompileReport => Unit,
-    onBuildChanged: Seq[b.BuildTargetEvent] => Unit,
-    treeViewProvider: () => TreeViewProvider,
-    worksheetProvider: () => WorksheetProvider,
-    ammonite: () => Ammonite
-)(implicit ec: ExecutionContext)
-    extends MetalsBuildClient
+    onBuildTargetDidCompile: BuildTargetIdentifier => Unit,
+    onBuildTargetDidChangeFunc: b.DidChangeBuildTarget => Boolean
```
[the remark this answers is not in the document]
alexarchambault: There's already a `ForwardingMetalsBuildClient.onBuildTargetDidChange`, hence the different name here.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alexarchambault: This sounds like "why make things immutable when we could make them mutable?" to me 😅

Creating it upfront allows to pass it around and "wire" things upfront.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alexarchambault: Do you mean moving the whole `onBuildTargetDidChangeFunc` value, or just this Ammonite thing? For the latter, if the `onComplete` handlers stay here, this won't save any space.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alexarchambault: These can be wired upfront, so there's no reason not to do it IMO.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alexarchambault: We should probably do something like this, yeah.

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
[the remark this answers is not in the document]
alexarchambault: Done

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
```diff
@@ -0,0 +1,500 @@
...
+      .rebuildIndex(targets)
+      .foreach { _ =>
+        testProvider().refreshTestSuites()
+        languageClient.refreshModel()
+      }
+  }
+
+  def indexWorkspaceSources(data: Seq[BuildTargets.WritableData]): Unit = {
+    for (data0 <- data.iterator)
+      indexWorkspaceSources(data)
```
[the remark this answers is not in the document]
alexarchambault: Should be fixed.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
alexarchambault: Ok, changed that to `final case class DataSeq(list: List[TargetData])`

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
alexarchambault: Done.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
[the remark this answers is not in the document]
alexarchambault: Done.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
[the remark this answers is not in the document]
tgodzik: I was thinking the second, but if it's too much work don't worry about it.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: I would say the whole `onBuildTargetDidChangeFunc`

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: Sure, I was just thinking that in some cases you don't actually need those targets. For example when there is only ammonite ones.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -515,68 +393,45 @@ final class BuildTargets(
...
   def buildServerOf(
       id: BuildTargetIdentifier
-  ): Option[BuildServerConnection] = {
-    ammoniteBuildServer(id).orElse(targetToConnection.get(id))
-  }
+  ): Option[BuildServerConnection] =
+    data.fromOptions(_.targetToConnection.get(id))
+
+  def addData(data: TargetData): Unit =
+    this.data = BuildTargets.DataSeq(data :: this.data.list)
```
tgodzik, line 430: This is more like `setData`, should it be synchronized?

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -649,4 +504,256 @@ object BuildTargets {
...
+    def fromIterators[T](f: TargetData => Iterator[T]): Iterator[T] =
+      iterator.flatMap(f)
+    def fromOptions[T](f: TargetData => Option[T]): Option[T] =
+      fromIterators(f(_).iterator).take(1).toList.headOption
+  }
+
+  def createData(): TargetData =
+    new DataImpl
+
+  private final class DataImpl extends TargetData {
```
tgodzik, line 521: Do we still need a separate `DataImpl` if it's just really a single class using this interface? I would just move everything to TargetData and make relevant fields private

alexarchambault: Should be fixed.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alexarchambault: Fixed.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
alexarchambault: Done.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
alexarchambault: It's only called when `MetalsLanguageServer` is instantiated (or during the initialize request), so there's no concurrency issue for now.

But I just synchronized it anyway, just in case this is called in a different way in the future.

alexarchambault:
This should be ready for another round of review, I think all previous review comments are addressed.

alexarchambault:
This should be ready for another round of review, I think all previous review comments are addressed.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -91,149 +59,78 @@ final class BuildTargets(
...
+  def sourceItems: Iterable[AbsolutePath] =
+    data.iterable.flatMap(_.sourceItemsToBuildTarget.keys)
   def sourceItemsToBuildTargets
       : Iterator[(AbsolutePath, JIterable[BuildTargetIdentifier])] =
-    sourceItemsToBuildTarget.iterator
+    data.fromIterators(_.sourceItemsToBuildTarget.iterator)
+  private def allBuildTargetIdsInternal
+      : Iterator[(TargetData, BuildTargetIdentifier)] =
+    data.fromIterators(d => d.allBuildTargetIds.iterator.map((d, _)))
+  def actualSource(path: AbsolutePath): Option[TargetData.MappedSource] =
```
tgodzik, line 70: Could we change the name a bit? Not sure `actualSource` means in this case.

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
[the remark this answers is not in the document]
tgodzik: Let's remove it then?

tgodzik (changes requested):
Thanks for working on this! I have some smaller leftover comments, otherwise looks good!

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -547,69 +393,48 @@ final class BuildTargets(
...
   def buildServerOf(
       id: BuildTargetIdentifier
-  ): Option[BuildServerConnection] = {
-    ammoniteBuildServer(id).orElse(targetToConnection.get(id))
-  }
+  ): Option[BuildServerConnection] =
+    data.fromOptions(_.targetToConnection.get(id))
+
+  def addData(data: TargetData): Unit =
+    dataLock.synchronized {
```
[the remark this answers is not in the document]
alexarchambault: I think it's safer with it. `data` is a var (it's the one we mutate), so I'd rather not synchronize on it.

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
alexarchambault: Done (it's `mappedTo` now).

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -425,17 +425,14 @@ class MetalsLanguageServer(
...
+          onBuildTargetDidCompile = { target =>
+            treeView.onBuildTargetDidCompile(target)
+            worksheetProvider.onBuildTargetDidCompile(target)
           },
-          () => treeView,
-          () => worksheetProvider,
-          () => ammonite
+          onBuildTargetDidChangeFunc = { params =>
+            maybeQuickConnectToBuildServer(params)
+            true
```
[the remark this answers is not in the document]
alexarchambault: There used to during my developments… Just dropped the boolean.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1366,7 +1364,13 @@ class MetalsLanguageServer(
     ) {
       event.eventType match {
         case EventType.Create =>
-          buildTargets.onCreate(path)
+          for {
+            buildTargetIds <- buildTargets.sourceBuildTargets(path)
+            buildTargetId <- buildTargetIds
+            targetData <- buildTargets.targetData(buildTargetId)
```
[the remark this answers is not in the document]
alexarchambault: Good point, fixed.

alexarchambault:
Seems `tests.FileWatcherLspSuite` failed twice in a row (third attempt ongoing), even though I can't reproduce that error locally or on an Ubuntu VM. CI is all ✅ apart from that, so I guess this is ready for another review or a merge…

tgodzik (approved):
LGTM!

`tests.FileWatcherLspSuite ` is flaky unfortunately, I need to investigate what is going on.

@ckipp01 what do you think?

ckipp01 (approved):

PR #3551, opened 2022-01-20
[Actions] Mtags release

dos65:
I missed that ci-release also does additional things.
Let's try this way for while.

I'm going to look at ci-release plugin to find put changes it might require for our case. 
Currently the only way to configure it is to use env-variables but it would require to parse tag names in bash (before running sbt) that I want to avoid.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -164,15 +164,21 @@ commands ++= Seq(
   Command.single("test-mtags-dyn") { (s, scalaV) =>
     crossTestDyn(s, scalaV)
   },
-  Command.args("publish-mtags-dyn", "<scala-version> <version>") {
-    (state, args) =>
-      val configured =
-        configureMtagsScalaVersionDynamically(state, args(0), Some(args(1)))
-      val (out, _) =
-        Project
-          .extract(configured)
-          .runTask(mtags / PgpKeys.publishSigned, configured)
-      out
+  Command.args("publish-mtags-dyn", "<scala-version> <version>") { (state, args) =>
+    val scalaV = args(0)
+    val metalsV = args(1)
+    CiReleasePlugin.setupGpg()
+    val reloadKeyFiles =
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -164,15 +164,21 @@ commands ++= Seq(
...
-  Command.args("publish-mtags-dyn", "<scala-version> <version>") {
-    (state, args) =>
-      val configured =
-        configureMtagsScalaVersionDynamically(state, args(0), Some(args(1)))
-      val (out, _) =
-        Project
-          .extract(configured)
-          .runTask(mtags / PgpKeys.publishSigned, configured)
-      out
+  Command.args("publish-mtags-dyn", "<scala-version> <version>") { (state, args) =>
```
tgodzik, line 167: Can't we just invoke only publishSigned?

In the docs:
```
(optional) CI_RELEASE: the command to publish all artifacts for stable releases. Defaults to +publishSigned if not provided.
```
we could set CI_RELEASE to "mtags/publishSigned"?

dos65: We also need override `version` + `scalaVersion`

tgodzik: We can run `sbt ++<scalaversion> ci-release` maybe? And set the version manually in the project

dos65: yes we can.
This would require to write in bash smth like:
```sh
if [ $GITHUB_REF == refs/tags*] && [ $GITHUB_REF_NAME == mtags_*]; then
  VERSION=$(echo $GITHUB_REF_NAME | cut -d "_" -f2 )
  SCALA_VERSION=$(echo $GITHUB_REF_NAME | cut -d "_" -f3 )
  # ... validate versions
  CI_RELEASE="mtags/publishSigned"
  sbt "set mtags / version :=\"$VERSION\"; set mtags/scalaVersion := \"$SCALA_VERSION\"; ci-release"
fi
```
But I don't think that it's the best option.

tgodzik (approved):
ok, let's try this one then!

PR #3553, opened 2022-01-20
[Actions] Mtags publishing - move all logic to ci-script

dos65:
Unfortunately, there is no way to make it work using on sbt side .

The problem is, that mtags release happens from latest release commit.
That means that there is no way to apply fixes for sbt-command.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,22 @@ jobs:
       - uses: coursier/cache-action@v6.3
       - name: Publish
         run: |
-          sbt gh-actions-release
+          COMMAND="ci-release"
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -15,7 +15,22 @@ jobs:
       - uses: coursier/cache-action@v6.3
       - name: Publish
         run: |
-          sbt gh-actions-release
+          COMMAND="ci-release"
```
tgodzik, line 18: We will be getting the `3.1.1` release soon, it will be perfect time to test it out!

dos65: Tested this code locally with `CI_RELEASE=mtags/publishLocal` and I hope there will be no issues with bash code.

I'm already testing it on `3.1.2-RC1-bin-20220119-2849aed-NIGHTLY`, so could you put an approve? :wink:

kpodsiad (approved):

PR #3554, opened 2022-01-20
feat, refactor: use notification to send test updates

kpodsiad:
Needs https://github.com/scalameta/metals-vscode/pull/836

What has been done in this PR:
- [x]  do not send all discovered suites to the clients, send only diffs
- [x] send diffs after compilatio is finished 
- [x] discover single tests for JUnit
- [x] refresh testcases for onChange/onDidOpen 

What won't be done here:
- add a new debug request with fine-grained tests selection

A client has to implement `metals-update-test-explorer` client's command.


metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,72 +1,118 @@
...
-  private val cachedTestSuites =
-    mutable.Map[FullyQualifiedClassName, TestEntry]()
+  private val index = new TestSuitesIndex
+  private val junitTestFinder = new JunitTestFinder
 
   val refreshTestSuites = new BatchedFunction[Unit, Unit](_ =>
     CancelableFuture(doRefreshTestSuites())
   )
 
+  def onChange(docs: TextDocuments, file: AbsolutePath): Unit = {
```
ckipp01, line 39: So thinking about the flow here a bit, I wonder if mimicking what we do with TVP might be a good idea. Mainly if a user doesn't have the test view open, then they are still going to be sent the `update-test-explorer` events. Do you think it'd make sense to keep track of the state of the explorer window like we do with TVP and if this is closed then don't even sent the update events? Then when they navigate back to it do a `metals.discover-test-suites`?

ckipp01: Thinking more about this, I'm not sure what would be best because in the scenario where we need to use `metals.discover-test-suites` right now we are sending everything, which is fine, but I'm glad we are taking the approach withe `update-test-explorer` to only send what is updated. That makes it 1000% time easier to manage on the client side imo. For example VS Code does some magic behind the scenes in TVP where it without any interruptions nicely updates the parts of the tree that are updated and only those parts. However in `nvim-metals` it isn't quite that smooth because we are just notified that a build target is updated, and then we replace the entire thing. It's not ideal, but if not I'd need to manually compare the two trees and only update the parts that have changed while retaining the collapse state etc. It's a lot of manual stuff that is free in VS Code.

All that to say, I'm glad we are doing it this way for the explorer and I'm just trying to think of a way to not fire a lot events to the client when it's not needed, but maybe it doesn't matter. wdyt?

metals/src/main/scala/scala/meta/internal/metals/testProvider/BuildTargetUpdate.scala
```diff
@@ -0,0 +1,39 @@
...
+  ) extends TestExplorerEvent("removeSuite")
+
+  final case class AddTestSuite(
+      fullyQualifiedClassName: String,
+      className: String,
+      symbol: String,
+      location: l.Location
+  ) extends TestExplorerEvent("addSuite")
+
+  final case class AddTestCases(
```
ckipp01, line 28: I haven't looked too closely at the VS Code implementation yet, but do we need to have more info here about the parent? I know we have the `fullyQualifiedName`, but if we have for example

```
targetA
  suiteA
  suiteB
target B
  suiteB
```
And then we get a `AddTestCases` event. We have the `targetName` that it belongs to (we'll say targetA), but not the `suite`. In this scenario wouldn't we also need to know the `suite` that it belongs to in order to correctly place it without having to parse the `fullyQualifiedName`?

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -84,12 +130,30 @@ final class TestSuitesProvider(
           )
         )
 
+      val deleted = mutable.ArrayBuffer[(BuildTarget, RemoveTestSuite)]()
```
ckipp01, line 133: Since this is a tuple all the accessing we do with it is a bit confusing with the `._1` and `._2` ceremony. For easiness of reading, is it possible to just store this as a `TestEntry` as well? I also think it's nicer to keep it in one format right until we send it out, then convert it to an `Event`. So internally we are only dealing with entries. wdyt?

ckipp01 (commented):
Cool, I like the direction this is headed @kpodsiad. A couple questions and thoughts.

ckipp01:
Also one other question I have is just about the general flow. My understanding of at least the part between client and server now looks like this:

```
     ,-----------.                                                                        ,------.
     |nvim-metals|                                                                        |Metals|
     `-----+-----'                                                                        `--+---'
           |   workspace/executeCommand (metals.discover-test-suites) opens the first time   |    
           |--------------------------------------------------------------------------------->    
           |                                                                                 |    
           |            workspace/executeCommand (response, BuildTargetUpdate[])             |    
           |<---------------------------------------------------------------------------------    
           |                                                                                 |    
           |workspace/executeClientCommand (metals.update-test-explorer, BuildTargetUpdate[])|    
           |<---------------------------------------------------------------------------------    
           |                                                                                 |    
           |workspace/executeClientCommand (metals.update-test-explorer, BuildTargetUpdate[])|    
           |<---------------------------------------------------------------------------------    
     ,-----+-----.                                                                        ,--+---.
     |nvim-metals|                                                                        |Metals|
     `-----------'                                                                        `------'
```
Is this also what you had in mind? the `discover-test-suites` happens at the start when the user would open it up for the first time? Or when would this happen again after that? Or does this not really happen at all with the new approach and it _can_ happen if the client wants a full refresh, but if not it's all `metals -> client` with `update-test-explorer` events? If the latter, then I guess you always have to send them no matter what even if the window isn't open, which then gets into my point about that.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
kpodsiad: From vscode's perspective, Metals have to fire events when:
- file which contains tests is opened
- TestExplorer view is opened
Otherwise, users can see stale information (gutter icons in file and entries is TE view). 
Sending events only for already "visited" test suites should be sufficient to keep a valid state. By "visited" I mean every test suite which has testcases listed in TE view.

A short GIF to show what I meant.

my-app-1642959912310

If I understood correctly you're concerned about unwanted events, yes? A few strategies like: `send all`, `send diffs` can be included in logic and one will be chosen depending on the client. Then vscode will get all events (because it needs them) and `nvim` can opt into `send all` and don't receive unnecessary events.

ckipp01: Actually thinking about it more since I wrote this, I don't think it's a valid worry. I was worried that it'd be like TVP where _every_ time anything in the target changes you get sent stuff that changes were made, which can then be really expensive if you need to recreate the state of that tree. However since we are only ever sending diffs, I don't think it's even a big deal. Just just ignore that concern.

metals/src/main/scala/scala/meta/internal/metals/testProvider/BuildTargetUpdate.scala
kpodsiad: Currently, `vscode.TestItem` (which is a basic block used in testing) has an id which consist of `build_target_id#fullyQualifiedName`. `TestItem`s form a tree data structure:
```
app
└── app#a
    ├── app#a.b.Suite1
    └── app#a.b
        ├── app#a.b.Suite2
        └── app#a.b.Suite3
```
So in my case, I have to parse `fullyQualifiedName` and at the same time it's sufficient to correctly place test cases. The only thing which can help will be to preprocess `fullyQualifiedName` into list of prefixes `[a, a.b, a.b.Suite1]` but I don't know if it will be useful for other clients.

kpodsiad:
I have to do some cleanup (pprints, unused stuff), but more or less final shape should be visible.

@ckipp01 Here's the current flow
```
     ,------.                                                  ,------.                    
     |Metals|                                                  |editor|                    
     `--+---'                                                  `--+---'                    
        |----.                                                                             
        |    | Build is imported and initial compilation has ended.                        
        |<---'  Send update to the client.                                                 
        |                                                                                  
        |                                                         |                        
        |   workspace/executeClientCommand                        |                        
        |    (metals.update-test-explorer, BuildTargetUpdate[])   |                        
        | -------------------------------------------------------->                        
        |                                                         |                        
        |----.                                                    |                        
        |    | File with test suite has changed.                  |                        
        |<---'  Send update to client.                            |                        
        |                                                         |                        
        |                                                         |                        
        |   workspace/executeClientCommand                        |                        
        |    (metals.update-test-explorer, BuildTargetUpdate[])   |                        
        | -------------------------------------------------------->                        
        |                                                         |                        
        |                                                         |----.                   
        |                                                         |    | Fetch test suites 
        |                                                         |<---' for workspace     
        |                                                         |                        
        |                                                         |                        
        |               workspace/executeCommand                  |                        
        |                (metals.discover-tests, {})              |                        
        | <--------------------------------------------------------                        
        |                                                         |                        
        |             workspace/executeCommand                    |                        
        |              (response, BuildTargetUpdate[])            |                        
        | -------------------------------------------------------->                        
        |                                                         |                        
        |                                                         |----.                   
        |                                                         |    | Fetch testcases   
        |                                                         |<---' for a given file  
        |                                                         |                        
        |                                                         |                        
        |             workspace/executeCommand                    |                        
        |              (metals.discover-tests, {uri})             |                        
        | <--------------------------------------------------------                        
        |                                                         |                        
        |             workspace/executeCommand                    |                        
        |              (response, BuildTargetUpdate[])            |                        
        | -------------------------------------------------------->                        
     ,--+---.                                                  ,--+---.                    
     |Metals|                                                  |editor|                    
     `------'                                                  `------'                    
```

ckipp01:
Cool! Thanks a lot for the diagram @kpodsiad, that's super helpful. For the `Fetch test suites for a given workspace`, is that because the user would want to refresh everything, or because they are first opening the panel? Basically on the editor side if the `update-test-explorer` events were already sent, should those be saved by the client before the window is open, which would mean there wouldn't need to be a fetch back to metals? Or what happens in that scenario?

Same question for `Fetch test cases for a given file`. Shouldn't the client already have them?

kpodsiad:
> For the `Fetch test suites for a given workspace`, is that because the user would want to refresh everything, or because they are first opening the panel? Basically on the editor side if the `update-test-explorer` events were already sent, should those be saved by the client before the window is open, which would mean there wouldn't need to be a fetch back to metals? Or what happens in that scenario?

Yes, those sent events should be somehow saved by the client and there shouldn't be any need to fetch everything from Metals. Metals send all events related to test suites on their own.

> Same question for `Fetch test cases for a given file`. Shouldn't the client already have them?

Testcases (currently only for junit4) are discovered lazily. Logic on the server-side should be sufficient for most common scenarios between editors:
- if a test suite is discovered and if its source file is currently opened (you created a test suite, added test case and press save. It's reasonable that you'll see both test suite and test case)
- if a change is made to a file that has test suites (one can add new testcase/delete existing - need to refresh them)
- if you open a file that has test suite (previously there only a whole test suite might be discovered)

`Fetch test cases for a given file` serves vs code purposes - in Test Explorer one can expand chosen test suite to see its children and this use case can't be handled o server-side.


I just realized that a flag `hasResolvedChildren` can help to reduce unnecessary computations for `onChange` - do it only for these tests which has `hasResolvedChildren` set to true.

@ckipp01 could you help me with wording? If a test suite/class is a class which contains tests, how would you call these tests? Just `tests` can be ambigous in this context since everything has `test` in its name. I used `testcases` but I'm not sure if it's a meaningful name.

ckipp01:
> @ckipp01 could you help me with wording? If a test suite/class is a class which contains tests, how would you call these tests? Just `tests` can be ambigous in this context since everything has `test` in its name. I used `testcases` but I'm not sure if it's a meaningful name.

I think sticking to the norm here is probably best, and I _think_ the norm would be that you have test suites, and those suites have tests???

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -170,6 +171,51 @@ object ClientCommands {
     """.stripMargin
   )
 
+  val UpdateTestExplorer = new ListParametrizedCommand[BuildTargetUpdate](
+    "metals-update-test-explorer",
+    "Update Test Explorer",
+    "Notifies the client that the test explorer model has to be updated",
+    """
+      |Sends to the client an Array of `BuildTargetUpdate`s.
```
tgodzik, line 179: Should it maybe be an object? It's easier to modify it later on.

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
```diff
@@ -170,6 +171,51 @@ object ClientCommands {
...
+      |export type TestExplorerEvent =
+      | RemoveTestSuite
+      | AddTestSuite
+      | AddTestCases;
+      |
+      |interface BaseTestExplorerEvent {
+      |  fullyQualifiedClassName: FullyQualifiedClassName;
+      |  className: ClassName;
+      |}
+      |export interface RemoveTestSuite extends BaseTestExplorerEvent {
```
tgodzik, line 196: We don't need extend data here I think? We can just explicitely write what each event provides.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -644,11 +644,27 @@ object MetalsEnrichments
...
       textDocument.occurrences
         .find(o => o.role.isDefinition && o.symbol == symbol)
         .map { occ => occ.toLocation(uri) }
     }
+
+    /**
+     * Convert given symbol into location in code only if symbol's occurrence
+     * has range defined.
+     */
+    def toLocation(uri: String, symbol: String): Option[l.Location] =
```
tgodzik, line 661: In reality the range should always exist, so no need to create a separate method.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -112,39 +112,65 @@ object ServerCommands {
...
+  final case class DiscoverTestParams(
+      @Nullable uri: String = null
+  )
+  val DiscoverTestSuites = new ParametrizedCommand[DiscoverTestParams](
+    "discover-tests",
+    "Discover tests",
+    """
+      |Discovers all test suites in project
+      |
+      |The response will be an Array of `BuildTargetUpdate`s.
```
tgodzik, line 125: This seems to be duplicated from the Client command, will we support communication both ways? For sure it's best to have it in one place.

It's a bit weird to havve this kind of mirror protocol, can't we make it one way only?

metals/src/main/scala/scala/meta/internal/metals/testProvider/BuildTargetUpdate.scala
```diff
@@ -0,0 +1,78 @@
...
+  def fullyQualifiedName: FullyQualifiedName
+  def clsName: ClassName
+}
+object TestExplorerEvent {
+
+  final case class FullyQualifiedName(value: String) extends AnyVal
+  final case class ClassName(value: String) extends AnyVal
+  final case class RemoveTestSuite(
+      @transient fullyQualifiedName: FullyQualifiedName,
+      @transient clsName: ClassName
```
tgodzik, line 40: Why do we create the class, mark them with transient and then do `.value` on them? Wouldn't it be simpler to avoid inheritance at all?

metals/src/main/scala/scala/meta/internal/metals/BatchedFunction.scala
```diff
@@ -11,6 +11,15 @@ import scala.util.control.NonFatal
 
 import scala.meta.internal.async.ConcurrentQueue
 
+final case class BatchedCallbackData[A, B](
```
tgodzik, line 14: Could we do the changes to BatchedFunction in another PR?

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesIndex.scala
```diff
@@ -0,0 +1,131 @@
...
+    target: BuildTarget,
+    testSymbols: TrieMap[
+      BuildTargetClasses.Symbol,
+      BuildTargetClasses.FullyQualifiedClassName
+    ]
+)
+
+private[testProvider] final case class TestFileMetadata(
+    entries: List[TestEntry],
+    textDocument: Option[TextDocument]
```
tgodzik, line 24: We can't keep all the text document in memory, that's why in References we use BloomFilter

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,77 +1,139 @@
...
+      doc <- metadata.textDocument.toSeq
+      metadataEntry <- metadata.entries
+      testEntry <- index.getTestEntry(
+        metadataEntry.buildTarget,
+        metadataEntry.suiteName
+      )
+      if symbol.forall(_ == testEntry.symbol)
+    } yield {
+      val testClass = testEntry.testClass
+      val testCases = junitTestFinder.findTests(doc, path, testEntry.symbol)
```
tgodzik, line 124: You should do this onChange and not save semanticdb docs in memory.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,77 +1,139 @@
...
+        BuildTargetUpdate(buildTarget, events)
+      }
+    buildTargetUpdateOpt.toSeq
+  }
+
+  private def refreshTestCases(path: AbsolutePath): Unit = {
+    val buildTargetUpdateOpt = getTestCases(path)
+    buildTargetUpdateOpt.foreach { update =>
+      client.metalsExecuteClientCommand(
+        ClientCommands.UpdateTestExplorer.toExecuteCommandParams(
```
tgodzik, line 99: Is it checked whether test explorer is supported?

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,77 +1,139 @@
...
+          update
+        )
+      )
+    }
+  }
+
+  /**
+   * Searches for test cases for a given path.
+   */
+  private def findTestCases(
```
tgodzik, line 109: Test cases are still not supported to be run individually. Would it make sense to make this PR first work with push, but exclude test cases? So that we don't put everything into one big PR?

tgodzik (changes requested):
Thanks for the great work! I think we could split it up a bit so that we don't merge everything at one:
1. BatchedFunction changes with an explanation why we need them
2. Change to push strategy, but let's not push single test cases yet.
3. Add single test cases for specific framerworks.

Also we would need to wait with 3 until BSP is able to handle single test cases.

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
kpodsiad: It's copypasted from typescript's file, I didn't too much attention to such details

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
kpodsiad: I never was a fan of `should always` and `should never happen` :/ In my use case, it's really ok to get the Option back, so I would leave it.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
kpodsiad: > will we support communication both ways?

Yes, it's needed for single test cases - https://github.com/scalameta/metals/pull/3554#issuecomment-1022034873

metals/src/main/scala/scala/meta/internal/metals/BatchedFunction.scala
kpodsiad: Of course - https://github.com/scalameta/metals/pull/3571

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
kpodsiad: Yes, they aren't yet supported. But it will be really helpful if it'll be merged now.

metals/src/main/scala/scala/meta/internal/metals/testProvider/BuildTargetUpdate.scala
kpodsiad: `FullyQualifiedName` and `ClassName` are value classes because working on bare strings is very errorprone (for the record yes, I made a few mistakes before these value classes).

`@transient` is a leftover, now it can be removed, thanks!

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
kpodsiad: >It's easier to modify it later on.

I'm curious, easier how?

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesIndex.scala
kpodsiad: Good catch, I removed it and replaced it with `hasResolvedChildren` to keep info if this file is "relevant"

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
kpodsiad: all private methods are called from public methods and they're called if and only if test explorer is enabled

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
kpodsiad: > not save semanticdb docs in memory.

done, good catch!

> You should do this onChange

I'm doing it, among others, in `onChange` but there are other scenarios where it's useful - for example when the file is currently opened and suite is being discovered.

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
tgodzik: You can just add an additional parameter when needed, which will be impossible in an array. Though that might never be needed.

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
tgodzik: Would be cool if we could specify it all explictely, but not super high priority.

metals/src/main/scala/scala/meta/internal/metals/BatchedFunction.scala
tgodzik: Thanks!

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
tgodzik: I realized this after I wrote the comment, cool!

ckipp01 (approved):
This looks great @kpodsiad. I haven't progressed much on the nvim side on this yet, but from what I can tell this should work fine there as well!

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
kpodsiad: I moved them to `ClientCommands` since it's more important than `ServerCommands`

metals/src/main/scala/scala/meta/internal/metals/ClientCommands.scala
kpodsiad: Specified explicitly in `ClientCommands`

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,165 +1,337 @@
...
+    val buildTargetUpdates =
+      (deletedSuites.keySet ++ addedSuites.keySet ++ addedTestCases.keySet)
+        .map { buildTarget =>
+          val deleted = deletedSuites.get(buildTarget).getOrElse(Seq.empty)
+          val added = addedSuites.get(buildTarget).getOrElse(Seq.empty)
+          val added0 = addedTestCases.get(buildTarget).getOrElse(Seq.empty)
+          BuildTargetUpdate(buildTarget, deleted ++ added ++ added0)
+        }
+        .filterNot(_.events.isEmpty())
+        .toSeq
```
dos65, line 285: Hmm, it works but looks unusual for me. Maybe smth like:
```scala
val aggregated = 
  (deletedSuites.toSeq ++ addedSuites.toSeq ++ addedTestCases.toSeq)
    .foldLeft(Map.empty[BuildTarget, List[TestExplorerEvent]]){
      case (acc, (target, events)) =>
         val prev = acc.getOrElse(target, Nil)
         acc.updated(target, events ::: prev)
    }

aggregated.map {case (target, events) => BuildTargetUpdate(target, events)}
```

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,165 +1,337 @@
...
+      symbol: mtags.Symbol,
+      fullyQualifiedName: FullyQualifiedName
+  ): Option[(TestEntry, TextDocument)] = {
+    // fullyQualifiedClassName always contains at least one element - class name
+    val fullyQualifiedClassName = fullyQualifiedName.value.split('.').toSeq
+    val entryWithDocumentOpt =
+      for {
+        (definition, doc) <- getSemanticDb(symbol)
+        location <- doc.toLocation(definition.path.toURI.toString, symbol.value)
+        className <- fullyQualifiedClassName.takeRight(1).headOption
```
dos65, line 309: `last` should be safe as it can't be empty after `split('.')`
```suggestion
    val className = fullyQualifiedName.value.split('.').last
    val entryWithDocumentOpt =
      for {
        (definition, doc) <- getSemanticDb(symbol)
        location <- doc.toLocation(definition.path.toURI.toString, symbol.value)
```

dos65 (changes requested):
Nice work! :tada: 

I left some suggestions but looks good in general.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesIndex.scala
```diff
@@ -0,0 +1,133 @@
...
+   *
+   * For every test suite a.TestSuiteName Metals returns 2 symbols,
+   * one for object and one for class.
+   * However, only one of them has location in source code and it's useful.
+   * That's why FullyQualifiedClassName is used as a key.
+   * "a/TestSuiteName., a.TestSuiteName"
+   * "a/TestSuiteName#, a.TestSuiteName"
+   */
+  private val cachedTestSuites =
+    mutable.Map[
```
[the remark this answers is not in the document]
kpodsiad: I think it shouldn't be necessary since all updates are done through the `BatchedFunction` and most of the updates are also done automatically in `BatchedFunction`. The only concern is a situation when someone queries Metals exactly during `refresh` functions. 

I'll check on Metals codebase if switching implementation causes significant difference in needed time.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesIndex.scala
```diff
@@ -0,0 +1,133 @@
...
+      BuildTarget,
+      mutable.Map[FullyQualifiedName, TestEntry]
+    ]()
+  private val fileToMetadata = mutable.Map[AbsolutePath, TestFileMetadata]()
+
+  def suites: Iterable[(BuildTarget, Iterable[TestEntry])] =
```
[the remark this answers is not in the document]
kpodsiad: Yeah, good advice!

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
kpodsiad: > Did you faced such case?

I think it was caused by semanticDB with legacy schema, but if I no longer use them these safety cautions can be removed.

metals/src/main/scala/scala/meta/internal/metals/SemanticdbIndexer.scala
```diff
@@ -89,4 +87,11 @@ class SemanticdbIndexer(
       }
     }
   }
+
+  private def onChange(path: AbsolutePath, docs: TextDocuments): Unit = {
+    referenceProvider.onChange(docs, path)
+    implementationProvider.onChange(docs, path)
+    implicitDecorator.onChange(docs, path)
+    testSuitesProvider.onChange(docs, path)
```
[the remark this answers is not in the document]
kpodsiad: > Shouldn't it also be called in onDelete method?

I can add that to be consistent.

> btw. I think it would be good to refactor make some command trait for providers.

I thought of something similar, but Tomek already sent me "`+942 −353  😞`" after I asked him for review hah :D

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
kpodsiad: Nice suggestion, thanks!

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesIndex.scala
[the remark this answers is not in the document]
kpodsiad: Zero difference, I'll switch

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,165 +1,360 @@
...
-          .readOnlySnapshot()
-          .toList
-        // IMPORTANT this check is meant to check for class name, not a symbol
-        if !cachedTestSuites.contains(fullyQualifiedClassName)
-      } computeTestEntry(buildTarget, symbol, fullyQualifiedClassName)
+      val currentlyOpened = openedFiles().toSet
+      val addedTestCases = addedEntries.mapValues {
+        _.flatMap { case (entry, doc) =>
+          if (currentlyOpened.contains(entry.path))
+            findTestCases(entry.path, Some(doc))
```
tgodzik, line 215: Since we never do anything about non opened files, there is no need to load semanticdb into memory for them.

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,165 +1,360 @@
...
+      )
+      if symbol.forall(_ == testEntry.symbol)
+      // if text document isn't defined try to fetch it from semanticdbs
+      doc <- textDocument.orElse(getSemanticDb(testEntry.symbol).map(_._2))
+    } yield {
+      val testClass = testEntry.testClass
+      val testCases = junitTestFinder.findTests(doc, path, testEntry.symbol)
+
+      if (testCases.nonEmpty) {
+        index.setHasResolvedChildren(path)
```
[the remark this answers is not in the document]
kpodsiad: Maybe, I've picked `HasResolvedChildren` only because of vscode option's name

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,165 +1,360 @@
...
+      }
+
+      AddTestCases(
+        testClass.fullyQualifiedClassName,
+        testClass.className,
+        testCases.asJava
```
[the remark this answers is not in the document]
kpodsiad: At this moment there is no need for this.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
kpodsiad: The semanticDB has been already loaded for the test suite in `computeTestEntry`

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,165 +1,360 @@
...
-          fullyQualifiedClassName,
-          fullyQualifiedName.takeRight(1).head,
-          location
+      symbol: mtags.Symbol,
+      fullyQualifiedName: FullyQualifiedName
+  ): Option[(TestEntry, TextDocument)] = {
+    val className = fullyQualifiedName.value.split('.').last
+    val entryWithDocumentOpt =
+      for {
+        (definition, doc) <- getSemanticDb(symbol)
```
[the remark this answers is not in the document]
kpodsiad: it's used for computing location, do you have a better idea of how to determine class position?

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,165 +1,360 @@
...
+    }.toSeq
+  }
+
+  /**
+   * Retrieve test cases for a given file. Even an empty list
+   * because it can mean that all testcases were deleted.
+   */
+  private def getTestCases(
+      path: AbsolutePath,
+      doc: Option[TextDocument] = None
```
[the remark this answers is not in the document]
kpodsiad: it's `Some` in `onChange`

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,165 +1,360 @@
...
+      )
+    }
   }
 
+  /**
+   * Searches for test cases for a given path.
+   */
+  private def findTestCases(
+      path: AbsolutePath,
+      textDocument: Option[TextDocument],
```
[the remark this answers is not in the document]
kpodsiad: it's `Some` in `onChange`

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,165 +1,360 @@
...
+        refreshTestCases(file, Some(doc))
+      }
+    }
+  }
+
+  def onDelete(file: Path): Unit = {
+    val removed = index.remove(AbsolutePath(file))
+    removed match {
+      case Nil => ()
+      case head :: _ =>
```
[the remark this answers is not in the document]
kpodsiad: Yes, we can. I should fix this.

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,165 +1,360 @@
...
+   * because it can mean that all testcases were deleted.
+   */
+  private def getTestCases(
+      path: AbsolutePath,
+      doc: Option[TextDocument] = None
+  ): Option[BuildTargetUpdate] = {
+    val buildTargetUpdateOpt =
+      for {
+        metadata <- index.getMetadata(path)
+        buildTarget <- metadata.entries.headOption.map(_.buildTarget)
```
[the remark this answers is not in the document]
kpodsiad: It's another place which needs to take multiple build targets into account

tgodzik (changes requested):
There are some tests still failing and I also left one comment. Othwerwise we should be ok to merge. :rocket:

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -1,165 +1,381 @@
...
+      location <- {
+        definition.range
+          .map { range =>
+            new l.Location(
+              definition.path.toURI.toString,
+              range.toLSP
+            )
+          }
+          .orElse {
+            semanticdbs
```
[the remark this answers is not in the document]
kpodsiad: I use (0,0,0,0) as a fallback and this cause a whole chain reduction of simplifies ;)

kpodsiad:
I applied your suggestion @tgodzik about default range fallback and this allowed me to simplify the logic. I also (I hope) clarified a little bit what functions discover test cases. There are two of them currently: `getTestCasesForPath` and `getTestCasesForSuites`. `getTestCasesForPath` takes into account that file (test suite) can belong to many build targets and returns updates for all of them. `getTestCasesForSuites` concerns only about finding test cases for provided path and test suites, it doesn't know anything about build targets. 
Thanks to this separation there is no more ambiguous function or code duplication.

tgodzik (approved):

dos65 (approved):
LTGM :rocket:

PR #3560, opened 2022-01-23
feature: [Scala 3] Add keyword completions

tgodzik:
This also tweaks a couple of things with Scala 2:
- allow import without package (since we can have those)
- allow more keywords in ammonite files

We also add new keywords for Scala 3 except the `then` keyword which is complicated to add and would require some by hand parsing.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/CompletionValue.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/CompletionValue.scala
@@ -2,25 +2,52 @@ package scala.meta.internal.pc
 
 import scala.meta.internal.pc.CompletionValue.Kind
 
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.Flags.*
+import dotty.tools.dotc.core.Symbols.NoSymbol
 import dotty.tools.dotc.core.Symbols.Symbol
 import dotty.tools.dotc.interactive.Completion
+import org.eclipse.lsp4j.CompletionItemKind
 
 case class CompletionValue(
     label: String,
     symbol: Symbol,
-    kind: Kind
-)
+    kind: Kind,
+    isCustom: Boolean = false,
+    insertText: Option[String] = None
+):
+
+  def completionItemKind(using ctx: Context): CompletionItemKind =
+    if kind == CompletionValue.Kind.Keyword then CompletionItemKind.Keyword
+    else if symbol.is(Package) || symbol.is(Module) then
+      CompletionItemKind.Module // No CompletionItemKind.Package (https://github.com/Microsoft/language-server-protocol/issues/155)
+    else if symbol.isConstructor then CompletionItemKind.Constructor
+    else if symbol.isClass then CompletionItemKind.Class
+    else if symbol.is(Mutable) then CompletionItemKind.Variable
+    else if symbol.is(Method) then CompletionItemKind.Method
+    else CompletionItemKind.Field
+
+end CompletionValue
 
 object CompletionValue:
 
   enum Kind:
-    case NamedArg, Workspace, Compiler, Scope
+    case Keyword, NamedArg, Workspace, Compiler, Scope
 
   def fromCompiler(completion: Completion): List[CompletionValue] =
     completion.symbols.map(CompletionValue(completion.label, _, Kind.Compiler))
 
   def namedArg(label: String, sym: Symbol): CompletionValue =
-    CompletionValue(label, sym, Kind.NamedArg)
+    CompletionValue(label, sym, Kind.NamedArg, isCustom = true)
+
+  def keyword(label: String, insertText: String): CompletionValue =
+    CompletionValue(
+      label,
+      NoSymbol,
+      Kind.Keyword,
+      isCustom = true,
+      insertText = Some(insertText)
+    )
 
   def workspace(label: String, sym: Symbol): CompletionValue =
     CompletionValue(label, sym, Kind.Workspace)
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/KeywordsCompletions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/KeywordsCompletions.scala
@@ -1,5 +1,7 @@
 package scala.meta.internal.pc.completions
 
+import scala.meta.internal.mtags.MtagsEnrichments.given
+import scala.meta.internal.pc.CompletionPos
 import scala.meta.internal.pc.CompletionValue
 import scala.meta.internal.pc.Keyword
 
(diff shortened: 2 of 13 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/CompletionValue.scala
```diff
@@ -2,25 +2,52 @@ package scala.meta.internal.pc
...
+import org.eclipse.lsp4j.CompletionItemKind
 
 case class CompletionValue(
     label: String,
     symbol: Symbol,
-    kind: Kind
-)
+    kind: Kind,
+    isCustom: Boolean = false,
+    insertText: Option[String] = None
```
dos65, line 17: Maybe switch it sealed trait/enum to avoid  dealing with `NoSymbol`?

dos65 (approved):
Looks good! :tada: 

I left some suggestions but they are optional

kpodsiad (commented):
There isn't much I can say about this code and PR. However, this got me thinking - it'll be awesome have _some_ documentation about how some stuff in Metals is done. 

There are a few great articles (like [this one](https://scalameta.org/metals/blog/2018/12/12/fast-goto-definition)) but it isn't enough. I feel like there is room for improvement when it comes to knowledge propagation

kpodsiad (changes requested):
Can you write some documentation in `CompletionKeywordSuite` or `BaseCompletionSuite`, in which you'll shed some light on how code works? 
Also, why there is strange `(commit: '", " ", "')` in some places? 

Thanks!

mtags/src/main/scala-3/scala/meta/internal/pc/CompletionValue.scala
tgodzik: I was thinking about it, but I would rather do it as a separate refactoring, I think this might be even more useful with some additional custom completions. Will do it as a next step.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/KeywordsCompletions.scala
```diff
@@ -0,0 +1,159 @@
...
+      enclosing: List[Tree],
+      pos: SourcePosition
+  ): Option[String] =
+    enclosing match
+      case Ident(name) :: _ => Some(name.toString())
+      case (vd: ValDef) :: _ => Some(vd.name.toString())
+      case _ =>
+        getLeadingIdentifierText(pos)
+
+  private def getLeadingIdentifierText(pos: SourcePosition): Option[String] =
```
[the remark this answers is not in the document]
tgodzik: Changed to use `CompletionPos`, thanks!

mtags/src/main/scala-3/scala/meta/internal/pc/completions/KeywordsCompletions.scala
```diff
@@ -0,0 +1,159 @@
...
+              ) && notInComment =>
+            CompletionValue.keyword(kw.name, kw.insertText)
+        }
+    end match
+  end contribute
+
+  private def checkIfNotInComment(
+      pos: SourcePosition,
+      path: List[Tree]
+  ): Boolean =
```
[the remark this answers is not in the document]
tgodzik: Extracted as much as possible to CommonMtagsEnrichments.

tgodzik:
> Can you write some documentation in `CompletionKeywordSuite` or `BaseCompletionSuite`, in which you'll shed some light on how code works? Also, why there is strange `(commit: '", " ", "')` in some places?
> 
> Thanks!

I added the docs to all the methods in BaseCompletionSuite.

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -1385,10 +1385,11 @@ class CompletionSuite extends BaseCompletionSuite {
     "select-ignores-next-line",
```
tgodzik, line 1385: Checking `CompletionsBaseSuite` made me realize that this test was never checked and didn't work on Scala 2

tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
```diff
@@ -157,7 +220,6 @@ abstract class BaseCompletionSuite extends BasePCSuite {
       compat: Map[String, String] = Map.empty,
       postProcessObtained: String => String = identity,
       stableOrder: Boolean = true,
-      postAssert: () => Unit = () => (),
```
tgodzik, line 160: This one was unused.

tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
```diff
@@ -54,14 +54,36 @@ abstract class BaseCompletionSuite extends BasePCSuite {
...
+   *
+   * @param name name of the test
+   * @param original snippet to test with `@@` indicating the cursor position
+   * @param fn function used to run assertions on completion items
+   */
   def checkItems(
       name: TestOptions,
       original: String,
-      fn: Seq[CompletionItem] => Unit
+      fn: Seq[CompletionItem] => Boolean
```
tgodzik, line 67: Checked the method here, since it was misleading and people might have assumed that it runs assert on fn returned code..

tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
```diff
@@ -148,6 +191,26 @@ abstract class BaseCompletionSuite extends BasePCSuite {
...
+   * @param includeDocs whether to include documentation in the completion description
+   * @param includeCommitCharacter  show commit characters, which when typed will
+   *    indicate that completion is accepted.
+   * @param compat additional compatibility map for different Scala versions
+   * @param postProcessObtained function used to modify resulting completion list
+   * @param stableOrder we should not sort completions if set to true
+   * @param postAssert additional assertions to make on the results
+   * @param topLines a number of completions to include, by default all
+   * @param filterText filter returned completions according to text
+   * @param includeDetail include the copmpletion detail in the results, true by default
```
dos65, line 209:
```suggestion
   * @param includeDetail include the completion detail in the results, true by default
```

tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
```diff
@@ -124,6 +159,14 @@ abstract class BaseCompletionSuite extends BasePCSuite {
     }
   }
 
+  /**
+   * Check snippet syntax returned in the completions. Snippets show the editor where the cursor should end up ($0).
+   *
+   * @param name name of the test
+   * @param original snippet to test with `@@` indicating the cursor position
+   * @param expected string with a lsit of obtained completions
```
dos65, line 167:
```suggestion
   * @param expected string with a list of obtained completions
```

tests/cross/src/main/scala/tests/BaseCompletionSuite.scala
```diff
@@ -88,6 +110,19 @@ abstract class BaseCompletionSuite extends BasePCSuite {
     )
   }
 
+  /**
+   * Check the results of applying the first completion suggested at a cursor position
+   *     inicated by `@@`.
```
dos65, line 115:
```suggestion
   *     indicated by `@@`.
```

dos65 (approved):
great! 
mentioned few typos only.

kpodsiad (approved):
Awesome, thanks for docstring @tgodzik !

PR #3564, opened 2022-01-24
minor: remove non-relevant things

dos65:


tgodzik (approved):
:rocket:

metals/src/main/scala/scala/meta/internal/builds/BloopInstall.scala
```diff
@@ -64,11 +63,7 @@ final class BloopInstall(
         args,
         workspace,
         buildTool.redirectErrorOutput,
-        Map(
-          "COURSIER_PROGRESS" -> "disable",
-          "METALS_ENABLED" -> "true",
```
ckipp01, line 69: I'm not sure we should remove this.  In the last few months I've actually recommended users to look for this as a workaround for stuff related to Bloop. Removing this will remove the env variable right?

dos65: >  In the last few months I've actually recommended users to look for this as a workaround for stuff related to Bloop. 

Do you have any example when they might be useful?
I'm only noticed that `bloopInstall` doesn't use these variables

ckipp01: yea there have been times where they want a scalacOption for their build, but not for Metals development. So they could detect this options to conditionally set an option.

Here are a couple places where it came up:

- https://discord.com/channels/632150470000902164/884881696300859442/909901395132579860
- https://discord.com/channels/632150470000902164/884881696300859442/923167135025483806
- https://discord.com/channels/632642981228314653/632841743435563028/894525255819862066
- https://discord.com/channels/632642981228314653/632652693013528589/887409566596358195

dos65: @ckipp01 thx for examples!

I added a comment to these variables to prevent their removal in future :smile:

ckipp01 (approved):
👍🏼 good idea to add the comment. LGTM!

Issue #3566, opened 2022-01-25
Scala 3 mouseover type hints crush singleton types


tpolecat:
### Describe the bug

Singleton types In Scala 3 are widened in mouseover hints. 

![image](https://user-images.githubusercontent.com/1200131/151000354-54ab97e3-be5f-44ef-abbf-c7dbea6d73c2.png)

### Expected behavior

The hint should read `val one: 1`.

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.1

PR #3567, opened 2022-01-25
fix: [Scala 3] Show exact literal type

tgodzik:
Fixes https://github.com/scalameta/metals/issues/3566

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/Signatures.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/Signatures.scala
@@ -163,7 +163,6 @@ def shortType(longType: Type, history: ShortenedNames)(using
           ),
           loop(restpe, None)
         )
-      case ConstantType(value) => value.tpe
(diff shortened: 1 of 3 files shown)


dos65 (commented):
Could you also add a test on union of constant types like it was reported in discord?
```scala
val x: 1 | 2 = ???
```

dos65:
@tgodzik 

> Getting (1: Int | 2: Int) which is the default way, but I wonder if we shouldn't maybe keep it like that?

Looks good.
I'm thinking that maybe we should render `val x: 1` in the same way as `1: Int` .

tgodzik:
> @tgodzik
> 
> > Getting (1: Int | 2: Int) which is the default way, but I wonder if we shouldn't maybe keep it like that?
> 
> Looks good. I'm thinking that maybe we should render `val x: 1` in the same way as `1: Int` .

That's actually easier :D I tried to work around it to be the same as Scala 2, but maybe it's not needed.

dos65 (approved):
:+1:

PR #3568, opened 2022-01-25
build: add date to SNAPSHOT version

dos65:
Snapshot version will be:  `0.11.1+22-bc2fa7a6-${yyyyMMdd}-SNAPSHOT`.
This thing is need for scalameta/metals-languageclient#407

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,5 @@
+import java.time.{Instant, ZoneId}
+import java.time.format.DateTimeFormatter
 import scala.collection.mutable
 import scala.sys.process._
 import Developers._
@@ -42,9 +44,18 @@ usefulTasks := Welcome.tasks
 
 inThisBuild(
   List(
-    version ~= { dynVer =>
-      if (isCI) dynVer
-      else localSnapshotVersion // only for local publishing
+    version := {
+      val dynVer = version.value
+      val dynVerOut = dynverGitDescribeOutput.value
+      if (isCI) {
+        // We need to have date in version for published SNAPSHOT artifacts
+        // Do it manually as there is no way to configure dynver got this case
+        if (!dynVerOut.isDirty && dynVerOut.isSnapshot) {
+          val now = Instant.now().atZone(ZoneId.of("UTC"))
+          val date = now.format(DateTimeFormatter.ofPattern("yyyyMMdd"))
+          dynVer.replace("-SNAPSHOT", s"-$date-SNAPSHOT")
+        } else dynVer
+      } else localSnapshotVersion // only for local publishing
     },
     // note bucket created by @tgodzik
     scalaVersion := V.scala212,


tgodzik (changes requested):
Can't we use the same mechanism as https://github.com/scalameta/metals/blob/main/metals-docs/src/main/scala/docs/Snapshot.scala ?

dos65:
@tgodzik I use the same mechanic it but it's needed to have an easy option to check the date of build to not query snapshots page every time metals starts.

tgodzik:
I think switching this to doing normal SNAPSHOT release would probably be  easier to maintain.

dos65:
> I think switching this to doing normal SNAPSHOT release would probably be easier to maintain.

I was thinking about that but it think the more reliable way would be not to use it.

With mutable SNAPSHOT it would be hard to say on which exact version you are now + it will be under coursier TTL (24h  by default).
Also, it will be clashing with local SNAPSHOT that will require us to change local SNAPSHOT name.

Also there is hypothetical use case that won't work:
 - you are using remote snapshot
 - it has some issue that spoils the usage experience
 - the issue was fixed quickly but to switch on this version with fix you need: 
    - either somehow force cousier to update SNAPSHOT before TTL exceeded
   -  or switch on non-mutable specific snapshot version and then not to forget to update version on mutable SNAPSHOT back

Having all this things, I think it would be better not to rely on mutable SNAPSHOT.
At least the cost on proposed solution for switching between versions isn't big enough to worry about it.

ckipp01:
So not to rain on the parade, but honestly are we over thinking this? The amount of users that are using the snapshot releases and want them automatically updated is probably incredibly small. Is it worth adding complexity to the build definition, changing the snapshots names etc just to accomplish this? Also, maven artifacts are tagged by time when they are uploaded, so why can't we just rely on the listing time instead of baking this into the actual snapshot name?

gabro:
> With mutable SNAPSHOT it would be hard to say on which exact version you are now + it will be under coursier TTL (24h by default). Also, it will be clashing with local SNAPSHOT that will require us to change local SNAPSHOT name.

About knowning which version you're on, I think this can be solved by printing it in the logs (or surfacing it somewhere like in Doctor or the tree view)

About coursier TTL: do we care about updating that often?

About the snapshot name I think changing the local snapshot name to `local-SNAPSHOT` is easy enough, the current name was chosen without any particular thought.

Alternatively one could publish the latest version as `latest-SNAPSHOT` or some other name. I don't think the names matter that much, as long as it's clear what you're "subscribing" to.

> 
> Also there is hypothetical use case that won't work:
> 
> * you are using remote snapshot
> * it has some issue that spoils the usage experience
> * the issue was fixed quickly but to switch on this version with fix you need:
>   
>   * either somehow force cousier to update SNAPSHOT before TTL exceeded
>   * or switch on non-mutable specific snapshot version and then not to forget to update version on mutable SNAPSHOT back

I don't see this as terribly problematic. This feature is going to be used by power users anyway, who hopefully know the risks of the bleeding edge. As long as there's manual command to execute to update the snapshot in case you need to, I think it's acceptable.


I sort of agree with @ckipp01 that this seems a lot of effort, not much in absolute terms, but in relation to how many users are going to actually use this feature. For example the changes here https://github.com/scalameta/metals-languageclient/pull/407 are not minuscule, and they would need to be replicated for nvim, whereas relying on a mutable snapshot version would work out of the box for any editor supporting the `scalaVersion` setting (which I assume is all of them).


Anyway, I'm just voicing a concern about the effort/value ratio and the impact of these changes across the different projects, don't consider my opinion blocking in any way, and feel free to proceed with the solution you think is best.

tgodzik:
> Also, it will be clashing with local SNAPSHOT that will require us to change local SNAPSHOT name.

We could do `0.11.2-latest-SNAPSHOT` and you can always change for the latest to a specific snapshot.

I agree that the changes in scalameta/metals-languageclient#407  are a bit much and I would rather want to avoid it, since we would need to maintain it later.

ckipp01:
One other point, there are a few `nvim-metals` users that I know of that are cutting edge, using snapshots all the time. They are the minority, even in Neovim. I actually don't version `nvim-metals`, and merge in changes as long as they aren't breaking to the current stable, so they are running "nightly" all the time. I know beginners with VS Code and someone jumping in Neovim are probably a bit different of an audience, but even in `nvim-metals` the amount of people I have always on nightly is small. Plus we have a chunk of nix users that a feature like this is simply not going to work. I really think the use-case here is quite small. I update my version every morning (I just have a small script to grab the new one) and it's really not a huge deal to me, at least not to warrant a full overhaul imo.

Regarding mutable snapshots, my vote would be to avoid anything that will lead to users reporting "0.11.2-latest-SNAPSHOT" as their current version when opening issues. You then either lose the ability to know what they are using exactly, or it requires an extra round of "can you check this" before we can help, also causing the user to take that step.

Same as @gabro, don't let my opinions block this, but I'd just caution the benefit to cost ratio of this feature. Because as I understand it, the vast majority of this work is so that someone on the nightly VS Code extension doesn't have to just grab the snapshot of Metals themselves and update it.

_This is partly ironic that I'm saying all of this because I wanted a way to automatically install the latest snapshot locally, but there are ways to do this only the editor extension. So if I was to pursue this, imo it's something that the client should do seeing it's so niche._

dos65:
@ckipp01 yep, target audience of mentioned feature isn't big - maintainers plus some group of active users.

The reason for having it is to provide automatic way for testing latest snapshots and spot issues before release.
Personally, I rarely use snapshot and even more rarely remember to upgrade it.

So, I want to have an automatic way to switch on snapshots. Also I want to ask active users to use thing too and it would be nice it they can enable it just by a setting.

Mutable snapshot is quite the same thing but I had a feeling, that with described case in https://github.com/scalameta/metals/pull/3568#issuecomment-1022089111 some people will forget (at least I will definitely do it) to switch back on `-SNAPSHOT` that will spoil the effect.

What about the changes in scalameta/metals-languageclient#407 the dominant part of them is about version parsing and I wouldn't say that it would require a lot of support.

What about version name change, even ignoring the original motivation I think it would be better to have a date. Sometimes when I see snapshot version of Metals I have no idea how far it from the latest one. For example, scala3-nightly versions have it and I found it useful.

dos65:
@ckipp01 The fun thing that I at some point during implementing this thing in metals-languageclient I got angry on typescript and started thinking that it would take much less code if it would be written in Scala on server. 

What you all think if I move this on server and there will be a special client-command like `MetalsQuickPick` and the only what will be required by client is to implement config update?

ckipp01:
> What you all think if I move this on server and there will be a special client-command like `MetalsQuickPick` and the only what will be required by client is to implement config update?

For `nvim-metals` this would be next to impossible to implement. There is no "settings file" that a user has. It could be being set from a table in some random file on the packpath, or set as a global lua variable, etc. The idea of a `settings.json` file just doesn't exist in some editors. I believe Emacs would have the same problem.

dos65:
@ckipp01 I mean just somehow update metals version not the config file in general.

ckipp01:
> @ckipp01 I mean just somehow update metals version not the config file in general.

If I understand what you mean, that gets messy really quickly. Because would this be allowed if a user does have a `serverVersion` already set? What if they do? In `nvim-metals` and in Emacs users could be using the Metals that they installed with coursier, they installed locally, or that they let the editor install.

To give an example in `nvim-metals`, let's say I have `serverVersion` set to 1.11.1-1-SNAPSHOT and then I trigger this client command to update, it would then get the version from Metals, and then run the installation for the new version. What then happens when the user closes Metals and restarts? They are then stuck with a Metals that is hardcoded in their settings for one version but have a different version locally. This is partly unique to nvim-metals because it doesn't even look at the value of `serverVersion` unless an install or update command is triggered. I know for `nvim-metals` this would just lead to confusion.

Again, I want to reiterate, don't let this block the current approach if you truly think it's best. It just seems like a lot of small moving pieces to me.

dos65:
@ckipp01 

> Because would this be allowed if a user does have a serverVersion already set?

We should have the proper version in BuildInfo or we can add it.
So I see it would work in the following way:
- Metals server start, receives user configuration
- if there is a special setting enabled, it looks for the newest snapshots comparing them with it's own version from BuildInfo
- if it finds a version to updrade it notifies the client with newest version value

What about incompleted installation your case, I don't know the nvim details but I believe there should be some option to split it on steps and replace some values on the installed version after it will complete. 
Wouldn't be there the same issue if you run installation as usual and then cancel it before completion?

PR #3571, opened 2022-01-26
fix: add possibility to define callback for `BatchedFunction`

kpodsiad:
`BatchedFunction` batches asynchronous requests but it hasn't offered any mechanism to execute callbacks after request is completed. This lead to executing `foreach` method on returned futures, which led to race conditions in `Compilations.scala` Side effects have been executing in parallel which was visible in the Test Explorer GUI.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BatchedFunction.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BatchedFunction.scala
@@ -11,6 +11,15 @@ import scala.util.control.NonFatal
 
 import scala.meta.internal.async.ConcurrentQueue
 
+final case class BatchedCallbackData[A, B](
+    arguments: Seq[A],
+    result: B
+)
+
+object BatchedCallbackData {
+  def noop[A, B]: BatchedCallbackData[A, B] => Unit = _ => ()
+}
+
 /**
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/BatchedFunction.scala
```diff
@@ -90,9 +113,17 @@ final class BatchedFunction[A, B](
...
       if (requests.nonEmpty) {
         val args = requests.flatMap(_.arguments)
+        val callbacks = requests.map(_.callback)
         val result = fn(args)
         this.current.set(result)
-        result.future.onComplete { response =>
+        val resultF = for {
+          result <- result.future
+          _ <- Future {
+            val callbackData = BatchedCallbackData(args, result)
```
tgodzik, line 122: Do we need `BatchedCallbackData` at all? This doesn't seem to be used

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/metals/BatchedFunction.scala
kpodsiad: I wanted to create a proper callback, may be I should have used `Unit => Unit`

tgodzik (approved):
LGTM!

PR #3572, opened 2022-01-27
cleanup: remove unused stuff from CompilerInterfaces

ckipp01:
Was digging around in here for something today and realized none of this was used. Can we just remove it all?


dos65 (approved):
:+1: 
I don't remember how it was used but I'm sure there now there is no need for it!

ckipp01:
> 👍 I don't remember how it was used but I'm sure there now there is no need for it!

Ha seeing that the diagnostic source in one of those methods is "dotty" I'm assuming this was just copied over from the Dotty language server and then never taken out.

tgodzik (approved):
:rocket:

PR #3574, opened 2022-01-28
Pass -Xsource:3 to scalafix organize-imports action if present

hugo-vrijswijk:
When using organize-imports on Scala 2 with `-Xsource:3` a crash would occur when using `as` import renaming:

```
2022.01.28 11:08:13 ERROR /..../ConfigReader.scala:8: error: } expected but identifier found
import pureconfig.{ConfigReader as PureConfigReader, ConfigSource}
```

This is because Scalafix tries to parse the file as Scala 2. Scalafix has support for `-Xsource:3` since https://github.com/scalacenter/scalafix/pull/1409 but it needs to be passed in the scalacOptions. With this fix `-Xsource:3` is passed to Scalafix if used, fixing the crash and letting Scalafix properly parse and fix the file.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
@@ -3,7 +3,7 @@ package scala.meta.internal.metals
 import java.net.URLClassLoader
 import java.nio.file.Files
 import java.nio.file.Path
-import java.util.Collections
+import java.{util => ju}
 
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
Thanks for fixing this! Let's see if the CI passes, then we should be good to merge!

hugo-vrijswijk:
Thanks! This has been a minor annoyance since using `-Xsource:3` and was a surprisingly easy fix!

PR #3577, opened 2022-01-28
Don't focus diagnotics if compialtion returned with error

tgodzik:
Previously, we would focus diagnotics every time a compilation finished with an error, which means we could steal focus from the user multiple times. Now, the only palce we focus diagnostics is when running/debugging and the debug session failed to run due to compilation errors.

This has surfaced due to fixing an issue in the VS Code client.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
@@ -513,9 +513,6 @@ class DebugProvider(
       languageClient.metalsStatus(
         Messages.DebugErrorsPresent(clientConfig.icons())
       )
-      languageClient.metalsExecuteClientCommand(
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -513,9 +513,6 @@ class DebugProvider(
       languageClient.metalsStatus(
         Messages.DebugErrorsPresent(clientConfig.icons())
       )
-      languageClient.metalsExecuteClientCommand(
```
tgodzik, line 516: This is already done by using `metalsStatus`

ckipp01 (approved):
Fine by me. I actually never figured out a way to use the commands yet that are tied to the status messages due tot he way they are displayed in nvim, so on my end I never actually use this.

LGTM! 👍🏼

PR #3578, opened 2022-01-28
Make sure we only analyze stacktrace lines that do look the part

tgodzik:


kpodsiad (commented):
Could yu alter also https://github.com/scalameta/metals/blob/47187096527be9765270aa640a10f82d6ccb345d/metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala#L159-L176 if you inlined condition?

ckipp01 (approved):
Looks good 👍🏼

tgodzik:
> Could yu alter also
> 
> https://github.com/scalameta/metals/blob/47187096527be9765270aa640a10f82d6ccb345d/metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala#L159-L176
> 
> if you inlined condition?

Good catch, I moved the condition into `fileLocationFromLine`

kpodsiad (approved):

Issue #2595, opened 2021-03-11
Invalid syntax error in ammonite multi-stage scripts


usommerl:
**Describe the bug**
Ammonite has a feature that allows you to divide a script into separate compilation units (See [multi-stage scripts](https://ammonite.io/#Multi-stageScripts)). This is sometimes required, for instance, when you need to add a dependency resolver. Within an Ammonite script, you can use the `@` symbol to indicate the beginning of a separate compilation unit. However, Metals flags the following line as an error because it expects an identifier after the `@` symbol.

**To Reproduce**

Use the following ammonite script and import it with metals:

```
// Ammonite 2.3.8, scala 2.13.4

interp.repositories() ++= Seq(coursierapi.MavenRepository.of("https://jitpack.io"))
@
import $ivy.`io.circe::circe-json-schema:0.1.0`

import io.circe.schema.Schema

val schema = Schema.loadFromString("{}")
println(schema.isSuccess)
```

**Expected behavior**

No errors are reported.

**Screenshots**

![issue_metals_2](https://user-images.githubusercontent.com/4050763/110795202-e1254500-8276-11eb-8ecf-2c202713ed43.png)

**Installation:**

- Operating system: Linux
- Editor: Vim
- Metals version: v0.10.0

**Search terms**

ammonite
multi-stage scripts

tgodzik:
Thanks for reporting! This is related to https://github.com/scalameta/scalameta/issues/2155 and needs to be fixed in scalameta first.

kitbellew:
> Thanks for reporting! This is related to [scalameta/scalameta#2155](https://github.com/scalameta/scalameta/issues/2155) and needs to be fixed in scalameta first.

not sure it should be part of scalameta. after all, it's not part of standard scala. another tool could feasibly come up with a different way to split portions of scala code. scalafmt handles this for `.sc` files, by splitting on `@`.

tgodzik:
> > Thanks for reporting! This is related to [scalameta/scalameta#2155](https://github.com/scalameta/scalameta/issues/2155) and needs to be fixed in scalameta first.
> 
> not sure it should be part of scalameta. after all, it's not part of standard scala. another tool could feasibly come up with a different way to split portions of scala code. scalafmt handles this for `.sc` files, by splitting on `@`.

[...]

PR #3579, opened 2022-01-29
docs(ammonite): add note about ammonite support

ckipp01:
This comes from the comment in #2595. I add a small note about our
Ammonite not supporting multi-stage scripts and also not currrently
supporting Scala 3.

I also noticed we had a small "Unsupported features" section which was
out of date, but also very minimal. So I don't really think it's useful
unless we have a more exhaustive list of what is not supported.


tgodzik (approved):
Thanks! Lgtm

PR #3580, opened 2022-01-30
Fix goto definition for local synthetics

tgodzik:
Previously, we would GotoPosition, which is a server command and because of that got removed from the client. Now, we use GotoLocation instead, which serves the same purpose.


kpodsiad (commented):
It seems like `otherWindow`, when false, is not being present in the command. Will it be a problem?

tgodzik:
> It seems like `otherWindow`, when false, is not being present in the command. Will it be a problem?

Not sure I understand, the `otherWindow` param is sent to the server as false (which is why tests failed :sweat_smile: )

kpodsiad (approved):

PR #3582, opened 2022-01-30
Show implicit decorations in worksheets

tgodzik:
Previously, we would not show implicit decorations in worksheet files, because they could get overriden by the worksheet decorations. Now, we send two different types of decorations, inline and not inline, which makes sure that one will not override the other.

This got tested in VS Code, but we would also need a confirmation on how to make it work in Sublime text.


metals/src/main/scala/scala/meta/internal/worksheets/DecorationWorksheetPublisher.scala
```diff
@@ -65,7 +65,11 @@ class DecorationWorksheetPublisher() extends WorksheetPublisher {
       decorations: Array[DecorationOptions]
   ): Unit = {
     val params =
-      new PublishDecorationsParams(path.toURI.toString(), decorations)
+      new PublishDecorationsParams(
+        path.toURI.toString(),
+        decorations,
+        isInline = false
```
ckipp01, line 71: In the `SyntheticsDecorationProvider` we do a check to see if the client is a `inlineDecorationProvider`, should we do that here as well? For example right now in `nvim-metals` in worksheets I'm getting `isInline: false` being sent here even though it doesn't support inline decorations.

docs/integrations/decoration-protocol.md
```diff
@@ -109,11 +109,20 @@ export interface PublishDecorationsParams {
    * Use empty list to clear all decorations.
    */
   options: DecorationOptions[];
+
+  /**
+   * Used to distinguish if inline decorations from the
+   * line ones as otherwise they will get replaced. It's
+   * not sent if client doesn't support inline decorations.
```
ckipp01, line 116: Final nit, maybe make it super clear here what the behavior is. Meaning these will never be sent in general since a client that doesn't support inline decorations will never be sent these decorations? or they will be sent these decorations, but the `inline` value won't be there? It's a bit ambiguous right now if you don't have the full context. Probably best to _over_ communicate here ha.

ckipp01 (approved):
One final nit that isn't a blocker, but LGTM!

docs/integrations/decoration-protocol.md
ckipp01: **Perfect**

tgodzik: :rocket:

metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
```diff
@@ -157,6 +144,47 @@ final class SyntheticsDecorationProvider(
...
+      textDocument match {
+        case Some(doc) =>
+          decorations(path, doc)
+        case _ =>
+          Nil
+      }
+
+    } else {
+      Nil
+    }
```
kpodsiad, line 171:
```suggestion
    currentDocument(path) match {
      case Some(doc) if isFocusedDocument =>
        decorations(path, doc)
      case _ =>
        Nil
    }
```
Mo Brackets, Mo Problems :D

mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
```diff
@@ -48,7 +49,9 @@ class SemanticdbTextDocumentProvider(
     } else {
       None
     }
-    val document = unit.toTextDocument(explicitDialect)
+    // we recalculate md5, since there seems to be issue with newlines sometimes
```
kpodsiad, line 52: What problems were caused by this newlines problem?

tests/unit/src/main/scala/tests/TestingClient.scala
```diff
@@ -349,7 +348,18 @@ class TestingClient(workspace: AbsolutePath, val buffers: Buffers)
...
     val path = params.uri.toAbsolutePath
-    decorations.put(path, params.options)
+    decorations.compute(
+      path,
+      {
+        case (_, decorationTypes) => {
+          if (decorationTypes == null) {
+            Set(params)
+          } else {
+            decorationTypes.filter(p => p.isInline != params.isInline) + params
```
kpodsiad, line 358: Why are you filtering decorations?

mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
tgodzik: Different md5 results produced by default when no newline was at the end. Not sure exactly why.

tests/unit/src/main/scala/tests/TestingClient.scala
tgodzik: To replace the particular inline type. If `isInline=true` we will only replace decorations with inline=true and vice versa

mtags/src/main/scala-2/scala/meta/internal/pc/SemanticdbTextDocumentProvider.scala
kpodsiad: It makes sense that inserting a character changes the document hash, but I still don't see when the problem starts. When is this newline inserted? `unit.toTextDocument` has `md5 = unit.source.toMD5`, so it indicates that file content is changed after compilation unit is created. Is this vscode problematic behavior?

tgodzik: > When is this newline inserted?

I don't know :D That is the reason, and I am not even sure if it's a newline. Just that on the same text we were calculating two different md5 results.

kpodsiad: Do you have steps for reproduction? It bothers me a lot :c

tgodzik: When I create worksheet with last statement without a newline this started popping up. It's not strickly related tot his PR so I can revert the change

kpodsiad: It's okay, there is no need for reverting. I just want to investigate it ;)

kpodsiad (approved):

kubukoz:
image

Pretty nice :D I understand there's no hover yet?

tgodzik:
> image
> 
> Pretty nice :D I understand there's no hover yet?

There should be, though it also shouldn't yet work :D We also need the VS Code PR

ayoub-benali:
@tgodzik How is the editor supposed to treat the new `inline` argument ?

tgodzik:
> @tgodzik How is the editor supposed to treat the new `inline` argument ?

It's just an information for the editor so that we don't override worksheet decorations with synthetic ones and vice versa. They are being sent at a different time.

PR #3595, opened 2022-02-01
Migrate to the newest H2 version

tgodzik:
This version has removed support for all PageStore format, so a totally new file will get create. I also needed to add `unique` to id, which is used as a foreign key.

### DIFF
--- a/metals/src/main/resources/db/migration/V3__Jar_symbols.sql
+++ b/metals/src/main/resources/db/migration/V3__Jar_symbols.sql
@@ -1,6 +1,6 @@
 -- Indexed jars, the MD5 digest of path, modified time and size as key
 create table indexed_jar(
-  id int auto_increment,
+  id int auto_increment unique,
   md5 varchar primary key
 );
 -- Top Level Symbols per jar, allow for multiple jars with same symbols and paths
--- a/metals/src/main/scala/scala/meta/internal/metals/Tables.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Tables.scala
@@ -7,6 +7,7 @@ import java.sql.DriverManager
 import scala.util.control.NonFatal
 
 import scala.meta.internal.builds.Digests
+import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.pc.InterruptException
 import scala.meta.io.AbsolutePath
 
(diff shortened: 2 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/Tables.scala
```diff
@@ -96,12 +97,17 @@ final class Tables(
       if (isAutoServer) ";AUTO_SERVER=TRUE"
       else ""
     val dbfile = workspace.resolve(".metals").resolve("metals")
+    val oldDbfile = workspace.resolve(".metals").resolve("metals.h2.db")
```
ckipp01, line 100: Won't this be found on ever re-launch of metals? So then it will delete and recreate every restart?

kpodsiad: After h2 update db file will have a new name - `metals.mv.db`

tgodzik: Actually, due to a change of database the name also changes, wo I am only deleting the old database.

ckipp01: ahhhh ok, that makes sense then.

metals/src/main/resources/db/migration/V3__Jar_symbols.sql
```diff
@@ -1,6 +1,6 @@
 -- Indexed jars, the MD5 digest of path, modified time and size as key
 create table indexed_jar(
-  id int auto_increment,
+  id int auto_increment unique,
   md5 varchar primary key
 );
```
kpodsiad, line 5: If the database has to be created from scratch can we also merge all migrations into one? What do you think?

metals/src/main/scala/scala/meta/internal/metals/Tables.scala
kpodsiad: maybe the new database name is worth mentioning in the comment?

metals/src/main/resources/db/migration/V3__Jar_symbols.sql
tgodzik: I think it's better to leave it as is, since we will get the proper git history (why was this added and when). I don't see any clear benefit to migrating them to one file.

metals/src/main/scala/scala/meta/internal/metals/Tables.scala
tgodzik: added the comment!

ckipp01 (approved):
LGTM!

PR #3600, opened 2022-02-01
Do not insert inferred type on variable bound to another

tgodzik:
Previously, we would special case `case a @ b =>` to allow inserting type on b. This however was causing flakines in a tests (and possibly in real usage), it wasn't working properly and this edge case is unlikely (we would have a double pattern var). Now, we disable inferring type when parent of pattern variable is pattern bind.


kpodsiad (approved):

ckipp01:
btw ignore the sourcegraph upload issue. we are hitting on this https://github.com/coursier/coursier/issues/2351

dos65 (approved):

PR #3601, opened 2022-02-01
Revert to Mdoc 2.2.24 for Scala 2.11

tgodzik:
With mdoc 2.3.0 we dropped the support for Scala 2.11, so we need to hardcode last correct version.


ckipp01 (approved):
LGTM

PR #3603, opened 2022-02-02
chore(ci): update setup-action

ckipp01:
This will fix the error we are seeing that was reported here:

https://github.com/coursier/coursier/issues/2351


tgodzik (approved):
:rocket:

PR #3604, opened 2022-02-02
Update Algolia keys

tgodzik:
@olafurpg  Is the apiKey and appId the same for all indexes? Or should it be different for Metals?


tgodzik:
Not sure how to proceed here, the searches seems to be working correctly, while Algolia didn't still add me to the organization. @olafurpg did you get any more comunication from them?

PR #3605, opened 2022-02-03
feat: add range to symbol definition

kpodsiad:
Adding range to symbol definition will allow tests provider to not load semanticDB per each test suite. Instead, the position can be obtained from the aforementioned symbol definition.

credits: @tgodzik

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/mtags/GlobalSymbolIndex.scala
+++ b/mtags/src/main/scala/scala/meta/internal/mtags/GlobalSymbolIndex.scala
@@ -2,6 +2,7 @@ package scala.meta.internal.mtags
 
 import scala.meta.Dialect
 import scala.meta.internal.mtags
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala/scala/meta/internal/mtags/GlobalSymbolIndex.scala
```diff
@@ -2,6 +2,7 @@ package scala.meta.internal.mtags
 
 import scala.meta.Dialect
 import scala.meta.internal.mtags
+import scala.meta.internal.semanticdb
```
tgodzik, line 5:
```suggestion
import scala.meta.internal.{semanticdb => s}
```

tgodzik (approved):
Just a super tiny nitpick, which we can follow up in the other PR. LGTM

Issue #2960, opened 2021-07-14
IdentifierComparator will fail with long numbers


tgodzik:
**Describe the bug**

It seems that with big projects some of the nodes in three view might have ids that are not representable by int. We should switch IdentifierComparator to use Long or use a separate comparator for TreeView

**To Reproduce**

1. Open a really large project
2. Try clicking around on the tree view

This was reported on discord, though I totally forgot to raise it here.

**Expected behavior**

No exception.

**Installation:**

- Operating system: macOS/Windows/Linux
- Editor: Visual Studio Code/Atom/Vim/Sublime/Emacs
- Metals version: v0.10.5

**Search terms**

long tree view IdentifierComparator

Z1kkurat:
A bit of clarification: as pointed out in [discord channel](https://discord.com/channels/632642981228314653/632652693013528589/869549202877841459), the problem is in the sorting of the list, namely in `IdentifierComparator` which sorts entries so that `Function10` goes after `Function2` in the list.  
To do that it takes that number substring and tries to convert it to Int which fails. But I don't think that using Long instead will fix the problem, it will just minimize the impact a bit, because in my example it failed when trying to convert a number substring in  `javax.servlet.jsp.jstl-1.2.0.v201105211821.jar` and failed for `201105211821`. There's no prevention from having even longer numeric substring in jar file name which will overflow Long

Z1kkurat:
Here's the illustration of the error for this exact library

![image](https://user-images.githubusercontent.com/7479593/127151702-22092ae9-8d15-4ba8-b2a0-f3aa94b3ceb4.png)

PR #3607, opened 2022-02-03
fix: compare long numbers in TVP

kpodsiad:
fixes https://github.com/scalameta/metals/issues/2960

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/IdentifierComparator.scala
+++ b/mtags/src/main/scala/scala/meta/internal/IdentifierComparator.scala
@@ -2,6 +2,8 @@ package scala.meta.internal.pc
 
 import java.util.Comparator
 
+import scala.annotation.tailrec
+
 /**
  * A comparator for identifier like "Predef" or "Function10".
  *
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala/scala/meta/internal/IdentifierComparator.scala
```diff
@@ -14,35 +16,47 @@ import java.util.Comparator
...
+    val first = toDigit(s1, idx)
+    val second = toDigit(s2, idx)
+    first.compare(second)
   }
-  private def toDigit(cs: CharSequence, i: Int): Int = {
+
+  private def toDigit(cs: CharSequence, i: Int): BigInt = {
     val digit = cs.subSequence(i, seekNonDigit(cs, i))
-    Integer.parseUnsignedInt(digit.toString)
+    BigInt(digit.toString())
```
tgodzik, line 60: Will BigInt not cause a bit of performance regression? This is used in completions, so I would love to avoid it.

There was some discusssion on the topic here: https://contributors.scala-lang.org/t/optimizing-bigint-for-small-integers/3412/3

However it seems this would only be available for Scala 2.13.

What if instead, we first compared the lengths from `seekNonDigit(cs, i) - i`and then if it's the same compare digits until one is bigger?

tgodzik (changes requested):
Thanks for the fix! I am a bit worried about using BigInt here.

mtags/src/main/scala/scala/meta/internal/IdentifierComparator.scala
kpodsiad: >  This is used in completions

Ahhh, I forgot about that. I was worried about performance but then I thought: it's just TVP, it's not an issue.

I was thinking about a similar approach, I'll do that.

mtags/src/main/scala/scala/meta/internal/IdentifierComparator.scala
```diff
@@ -10,39 +12,82 @@ import java.util.Comparator
...
+        isFirstLonger
+      } else {
+        val a = first.charAt(index1)
+        val b = second.charAt(index2)
+        val byDigit = Character.compare(a, b)
+        // if digits are the same keep comparing
+        // if they differ and they're last digits pick greater one
+        // if they differ but they aren't last pick longer sequence
+        if (byDigit == 0) compareLoop(index1 + 1, index2 + 1)
+        else if (index1 == firstMaxIndex && index2 == secondMaxIndex) byDigit
```
tgodzik, line 74: Why do we need this condition?  `2111` and `1111` will return 0, no ?

tgodzik (changes requested):

tgodzik (approved):
LGTM! Thanks!

PR #3610, opened 2022-02-04
chore(fmt): bump scalafmt to 3.4.0

ckipp01:
I mainly wanted to bump the scalafmt version that by default gets included in a new project if you don't have a .scalafmt file, since the 3.0.5 version has some issues with significant whitespace that I keep hitting on, so I always update it anyways.

Just to align them, I also then bumped the version that we are using for our project as well. There is only one place I really don't like the new formatting, but it does align with other formatting in the same file. I'll comment there.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
@@ -553,11 +553,9 @@ object FormattingProvider {
 
   }
 
-  def newScalafmt(respectVersion: Boolean = true): Scalafmt =
+  def newScalafmt(): Scalafmt =
     Scalafmt
       .create(this.getClass.getClassLoader)
       .withReporter(EmptyScalafmtReporter)
-      .withDefaultVersion(BuildInfo.scalafmtVersion)
-      .withRespectVersion(respectVersion)
 
 }
--- a/tests/unit/src/test/scala/tests/testProvider/TestSuitesProviderSuite.scala
+++ b/tests/unit/src/test/scala/tests/testProvider/TestSuitesProviderSuite.scala
@@ -45,7 +45,9 @@ class TestSuitesProviderSuite extends BaseLspSuite("testSuitesFinderSuite") {
           |        "fullyQualifiedName": "NoPackage",
           |        "className": "NoPackage",
           |        "location": {
-          |          "uri": "${classUriFor("app/src/main/scala/NoPackage.scala")}",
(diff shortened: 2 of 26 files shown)


tests/unit/src/test/scala/tests/testProvider/TestSuitesProviderSuite.scala
```diff
@@ -45,7 +45,9 @@ class TestSuitesProviderSuite extends BaseLspSuite("testSuitesFinderSuite") {
           |        "fullyQualifiedName": "NoPackage",
           |        "className": "NoPackage",
           |        "location": {
-          |          "uri": "${classUriFor("app/src/main/scala/NoPackage.scala")}",
+          |          "uri": "${classUriFor(
+        "app/src/main/scala/NoPackage.scala"
+      )}",
```
ckipp01, line 50: These are the cases that I really don't like. I wish it didn't split here. However, I see further down in the file it does the same thing.

dos65: could you try to add the following to config?
```
newlines.inInterpolation = avoid
```
It might help with this case

ckipp01: huh, I didn't even see `avoid` in the docs, thanks. I ended up not adding this since it won't "force" all the other places this is done, so we'd need to manually go through and change those which would create an even bigger diff than these two new cases. So because the other ones are already doing this I just left it.

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
```diff
@@ -553,11 +553,9 @@ object FormattingProvider {
 
   }
 
-  def newScalafmt(respectVersion: Boolean = true): Scalafmt =
+  def newScalafmt(): Scalafmt =
     Scalafmt
       .create(this.getClass.getClassLoader)
       .withReporter(EmptyScalafmtReporter)
-      .withDefaultVersion(BuildInfo.scalafmtVersion)
-      .withRespectVersion(respectVersion)
```
ckipp01, line 561: So these are both deprecated now, so I instead went the route where in `DownloadDependencies` I write a basic config file so that it's still able to pull down scalafmt and run once. When testing locally with a config that doesn't have a scalafmt version the behavior is still the same of prompting the user for a version and then addin git.

dos65 (approved):

PR #3611, opened 2022-02-04
Change hashing to lastModified

tgodzik:


tgodzik:
Based on the idea from @gabro

ritschwumm:
this is just using mtime instead of hashes, right? not sure it's relevant here, but maybe have a look at https://apenwarr.ca/log/20181113 before you decide to merge this.

tgodzik:
> this is just using mtime instead of hashes, right? not sure it's relevant here, but maybe have a look at https://apenwarr.ca/log/20181113 before you decide to merge this.

Looks like the topic is more complicated than I first thought. I will need to look exactly into swoval to see what it does.

tgodzik:
> > this is just using mtime instead of hashes, right? not sure it's relevant here, but maybe have a look at https://apenwarr.ca/log/20181113 before you decide to merge this.
> 
> Looks like the topic is more complicated than I first thought. I will need to look exactly into swoval to see what it does.

Took a look at the article, thanks! I think we are safe to merge it though. We are fine as long as mtime changes at all, we don't depend on any specific properties of mtime.

PR #3620, opened 2022-02-09
fix: handle the various enum hovers with more detail.

ckipp01:
This pr makes a couple small changes to the way the enum hover works for
Scala 3:

1. In the past when you'd hover on an actual enum we were returning the
   owner's full type string meaning that instead of showing the enum
   type, it'd show the containing class type. The first change changes
   this so it returns a shortened name of the actual enum.
2. The second change is handling enum case class instead of just vals.
   Previously they were displayed differently returning the same thing
   that they actual enum class was returning. Now this is aligned with
   the other cases and `case A(a: B)` will return the expected `case A:
   A` hover just like `case A, B, C` would.

### DIFF
--- a/tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
+++ b/tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
@@ -63,7 +63,7 @@ class HoverScala3TypeSuite extends BaseHoverSuite {
        |   case Red, Green, Blue
        |
        |""".stripMargin,
-    """|enum Color: enums2.SimpleEnum
+    """|enum Color: Color
        |""".stripMargin.hover
   )
 
@@ -76,7 +76,7 @@ class HoverScala3TypeSuite extends BaseHoverSuite {
        |  val color = <<Col@@or>>.Red
        |
        |""".stripMargin,
-    """|enum Color: enums3.SimpleEnum
+    """|enum Color: Color
        |""".stripMargin.hover
   )
 
@@ -92,6 +92,17 @@ class HoverScala3TypeSuite extends BaseHoverSuite {
        |
        |""".stripMargin,
     """|case Green: Color
+       |""".stripMargin.hover
+  )
+
+  check(
+    "enum-class",
+    """|
+       |object SimpleEnum:
+       |  enum Foo:
+       |    case <<Ba@@r(a: Int)>> extends Foo
+       |""".stripMargin,
+    """|case Bar: Bar
        |""".stripMargin.hover
   )
 
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
@@ -98,6 +98,7 @@ object HoverProvider:
                     expressionType
                   ) && !symbol.isType && !symbol.flags.isAllOf(EnumCase)
                 )
+
               val content = HoverMarkup(
                 expressionType,
                 hoverString,
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/SymbolPrinter.scala
@@ -64,11 +64,13 @@ class SymbolPrinter(using ctx: Context) extends RefinedPrinter(ctx):
     sym match
       case p if p.is(Flags.Package) =>
         s"package ${p.fullNameBackticked}"
-      case c if c.is(Flags.EnumVal) =>
+      // simple enum cases -> case A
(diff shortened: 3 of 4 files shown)


ckipp01:
I think this improves stuff, but it still isn't exactly what I'd like. For example

```scala
enum Foo:
  case A, B, C
```
Will return `A: A` if you hover on A whereas

```scala
enum Foo:
  case A extends Foo
  case B(a: Int) extends Foo
  case C extends Foo
```
Will return `A: Foo` if you hover on A. That's the one case that isn't what I'd like it to be. Even though hovering on `B` here now will return `B: B`. There seems to be a lot going on with Enums.

tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
```diff
@@ -63,7 +63,7 @@ class HoverScala3TypeSuite extends BaseHoverSuite {
        |   case Red, Green, Blue
        |
        |""".stripMargin,
-    """|enum Color: enums2.SimpleEnum
```
tgodzik, line 66: `enum Color: Color` seems less informative than the previous version. I would opt to leave this one as is.

ckipp01: My reasoning for this was that for example if we have `case class Foo()` and hover on `Foo()` we'll get `Foo: Foo`, which this sort of aligns with. Morever, `enums2.SimpleEnum` is actually incorrectly here, that's not the type. It could be `enums2.SimpleEnum.Color`, which would be fully qualified and correct, but very verbose. So mainly I just think` enums2.SimpleEnum` is incorrect, since it's showing you the object, but not the actual enum type. wdyt?

tgodzik: This is actually how it's handled in other cases like traits, so I guess it's better to do it this way :+1:

tgodzik: It seems that the `HoverTermSuite` is failing though so there might be some more cases to fix.

ckipp01: Yea just noticed that and actually is shows a nice example

```scala
FileVisi@@tResult.CONTINUE
```
Used to return `FileVisitResult: java.nio.file` before, which I think is confusing, so I like it better now that it just shows `enum FileVisitResult: FileVisitResult`.

tgodzik: I kind of liked that :sweat_smile:

tgodzik (approved):
LGTM

tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
ckipp01: haha I don't have strong feelings on this, and this was sort of just me poking around, so by all means push back.

For me I guess I'm just trying to understand _what_ should actually be shown here. Like what is the desired type to show for `FileVisitResult` for example. Like it is an enum, so is showing the package name for example correct as the type? My gut says no, which is why I made the change, but I might totally be missing something.

tgodzik: I think actually there is no need to show the type in this case, but we can change that separately at some point alongside traits and classes.

ckipp01:
Actually now when you hover on `FileVisitResult.CON@@TINUE` it's showing

```scala
enum CONTINUE: FileVisitResult
```
Which is worse than what it was showing before
```scala
case CONTINUE: FileVisitResult
```

ugh, I'm not satisfied with this yet. Let me dig in some more with this.

dos65 (commented):
I think we need to align this case with how it's done in Scala2 which I think is better than current solution.

For example, for object we print full name and I think we should do the same for enums: https://github.com/scalameta/metals/blob/a3e35922e4d7469569136cd8e3b3d3f588fe73ec/tests/cross/src/test/scala/tests/hover/HoverTermSuite.scala#L338-L359

ckipp01:
> I think we need to align this case with how it's done in Scala2 which I think is better than current solution.
> 
> For example, for object we print full name and I think we should do the same for enums:
> 
> https://github.com/scalameta/metals/blob/a3e35922e4d7469569136cd8e3b3d3f588fe73ec/tests/cross/src/test/scala/tests/hover/HoverTermSuite.scala#L338-L359

Yea agree, I'll try to align this. The interesting this here is that we do the same thing in 3. We return `app.Outer`, but that's the enclsoing type, not the actual type. I wouldn't expect this, I would hope that this would be similiar to the 2 hover showing `object Outer.Foo` instead of `object Foo:  Outer`.

PR #3627, opened 2022-02-11
Fix issues with Ammonite multistage scripts

tgodzik:
Fixes https://github.com/scalameta/metals/issues/2595


metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -188,8 +188,7 @@ object Messages {
       s"${icons.info} Partial navigation",
       tooltip = "This external library source has compile errors. " +
         "To fix this problem, update your build settings to use the same compiler plugins and compiler settings as " +
-        "the external library.",
-      command = ClientCommands.FocusDiagnostics.id
```
tgodzik, line 192: This is unrelated, but found an issue with focus diagnostics when there were errors in dependencies.

kpodsiad (approved):

PR #3632, opened 2022-02-14
Fix showing virtual docs in jars with paths that need escaping

tgodzik:
There is an issue within ZipFileSytemProvider that will use the decoded path of the jar, which will not work in case of any special characters.

There is stack overflow question related to that:

https://stackoverflow.com/questions/9873845/java-7-zip-file-system-provider-doesnt-seem-to-accept-spaces-in-uri

### DIFF
--- a/tests/slow/src/test/scala/tests/feature/FileDecoderProviderLspSuite.scala
+++ b/tests/slow/src/test/scala/tests/feature/FileDecoderProviderLspSuite.scala
@@ -26,6 +26,37 @@ class FileDecoderProviderLspSuite extends BaseLspSuite("fileDecoderProvider") {
     Right(FileDecoderProviderLspSuite.tastySingle)
   )
 
+  val userHome: String = System.getProperty("user.home")
+  check(
+    "decode-jar",
+    s"""
+       |/metals.json
+       |{
+       |  "a": {
+       |    "scalaVersion": "${scala.meta.internal.metals.BuildInfo.scala212}",
+       |    "libraryDependencies": [
+       |      "ch.epfl.scala:com-microsoft-java-debug-core:0.21.0+1-7f1080f1"
+       |    ]
+       |  }
+       |}
+       |/a/src/main/scala/a/Main.scala
+       |package a
+       |import com.microsoft.java.debug.core.protocol.Events._
+       |
+       |object Main {
+       |  val a : ExitedEvent = null
+       |  println(a)
+       |}
+       |""".stripMargin,
+    "a/src/main/scala/a/Main.scala",
+    None,
+    "file-decode",
+    Right(FileDecoderProviderLspSuite.EventsJarFile),
+    customUri = Some(
+      s"jar:file://${userHome}/.cache/coursier/v1/https/repo1.maven.org/maven2/ch/epfl/scala/com-microsoft-java-debug-core/0.21.0%2B1-7f1080f1/com-microsoft-java-debug-core-0.21.0%2B1-7f1080f1-sources.jar!/com/microsoft/java/debug/core/protocol/Events.java"
+    )
+  )
+
   check(
     "tasty-multiple",
     s"""|/metals.json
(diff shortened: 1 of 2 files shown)


Arthurm1:
I actually think this is wrong in Coursier and that the current amount of decoding is correct...

https://github.com/coursier/coursier/issues/2186

tgodzik:
Not sure if this is related to coursier. The file system doesn't seem to recognize the jar path without escaping, which is showing up here.

Arthurm1:
I don't know huge amounts about this but then isn't that URI invalidly encoded?

Shouldn't a path of...
_/com-microsoft-java-debug-core/0.21.0 **%2B** 1-7f1080f1/com-microsoft-java-debug-core-0.21.0 **%2B** 1-7f1080f1-sources.jar_

become a URI of...
_file:///com-microsoft-java-debug-core/0.21.0 **%252B** 1-7f1080f1/com-microsoft-java-debug-core-0.21.0 **%252B** 1-7f1080f1-sources.jar_

In fact in the `buildTarget/dependencySources` section of the bsp trace when opening Bloop in Metals I can see...
_file:///com-microsoft-java-debug-core/0.21.0 **%252B** 1-7f1080f1/com-microsoft-java-debug-core-0.21.0 **%252B** 1-7f1080f1-sources.jar_

which appears to be properly encoded.

To me - the URI you're decoding in your example has already been decoded but it shouldn't have been.

I'm wondering if there's a missing encode or a double-decode somewhere before this.

Arthurm1:
I think this is down to me [here](https://github.com/scalameta/metals-vscode/blob/c52d310a6ff82c543b26e12e3ab5c2a2d435bedf/src/metalsContentProvider.ts#L34).  I'm guessing you're using VSCode

I don't think this is true anymore.  Over encoding is dealt with in the `MetalsEnrichments.XtensionString#toAbsolutePath` so I should probably change to `uri.toString(false)`

I'd like to play with it a bit first though.  Good job there's a pre-release.

Arthurm1:
Well that's not it.

It might take me a while to work out why this is happening.

Feel free to merge if it helps with current encoding issues.

dos65 (approved):
LGTM!

kpodsiad (approved):
ZipFileSystemProvider is weird.

tgodzik:
@Arthurm1 the other option is to encode `%`, which would mean that when it gets decode it decodes into `%2B`. Not sure if that is better though.

It's really an issue in ZipFileProvider, which is not dealing well with escaping - it should actually escape things themselves.

I will merge this one and we can improve if we find a better way in the future.

tests/slow/src/test/scala/tests/feature/FileDecoderProviderLspSuite.scala
```diff
@@ -26,6 +26,37 @@ class FileDecoderProviderLspSuite extends BaseLspSuite("fileDecoderProvider") {
...
+       |  val a : ExitedEvent = null
+       |  println(a)
+       |}
+       |""".stripMargin,
+    "a/src/main/scala/a/Main.scala",
+    None,
+    "file-decode",
+    Right(FileDecoderProviderLspSuite.EventsJarFile),
+    customUri = Some(
+      s"jar:file://${userHome}/.cache/coursier/v1/https/repo1.maven.org/maven2/ch/epfl/scala/com-microsoft-java-debug-core/0.21.0%2B1-7f1080f1/com-microsoft-java-debug-core-0.21.0%2B1-7f1080f1-sources.jar!/com/microsoft/java/debug/core/protocol/Events.java"
```
ckipp01, line 56: Maybe I'm missing something, but how does this work locally? For example if I run this test on my local machine, MacOS, this isn't where my coursier cache is located by default?

tgodzik: It's not in `userHome` ? I kind of assumed since it passed that my guess was right

ckipp01: No, on Mac it's odd, it doesn't go to `$HOME/.cache` but rather `$HOME/Library/Caches`

tgodzik: This shouldn't have passed on CI then :sweat_smile:  I am very much confused. 

I will try and fix that.

ckipp01: ha that was my thought as well, I don't get how it passes in CI because locally it won't for me.

PR #3639, opened 2022-02-15
Fix FileDecoderProviderLspSuite to work locally

tgodzik:


ckipp01 (approved):
Ha I still don't get how this was passing before in CI, but this seems to make sense.

PR #3640, opened 2022-02-15
fix: actions, discover new scala3 nigtly properly

dos65:
The previous logic was invalid as usually nightly release date in version
doesn't match the today.
For example: `3.1.3-RC1-bin-20220214-4105afb-NIGHTLY` was released at
`2022-02-15`.

Also there might be more thant nightly release per day.

Now, the release will be triggered for all nightly versions that comes
after the last nightly mtags release.

### DIFF
--- a/.github/workflows/check_scala3_nightly.yml
+++ b/.github/workflows/check_scala3_nightly.yml
@@ -13,9 +13,12 @@ jobs:
           apps: ""
       - name: "Query maven-central"
         run: |
-          TODAY_NIGHTLY=$(cs complete org.scala-lang:scala3-compiler_3: | { grep $(date '+%Y%m%d') || true; })
-          if [ ! -z $TODAY_NIGHTLY ]; then
-            gh workflow run mtags-auto-release.yml -f scala_version=$TODAY_NIGHTLY
-          fi
+          cs complete org.scala-lang:scala3-compiler_3: | grep NIGHTLY | sort > all_nightly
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
I would love to have it in ammonite or scalacli script :sweat_smile:

kpodsiad:
> I would love to have it in ammonite or scalacli script 😅

If you give ack, can you explain to me what code is doing? :P

tgodzik:
> > I would love to have it in ammonite or scalacli script sweat_smile
> 
> If you give ack, can you explain to me what code is doing? :P

It's finding all the unreleased nightly versions since the last one that we released mtags for and then releasing for those.

kpodsiad:
> > > I would love to have it in ammonite or scalacli script sweat_smile
> > 
> > 
> > If you give ack, can you explain to me what code is doing? :P
> 
> It's finding all the unreleased nightly versions since the last one that we released mtags for and then releasing for those.

Clever answer :P

I was joking but this one is serious. How while loop works? `while IFS= read -r version; do`? It's looping over nightlies and at the same time it's changing separator and reading and help :c @dos65

dos65:
@tgodzik 
> I would love to have it in ammonite or scalacli script sweat_smile

yeah. I initially thought to start writing this in scala but then realized that it still might be expressed in bash :smile: 
I promise to do than next time if there will be more changes 

@kpodsiad 
> I was joking but this one is serious. How while loop works?

I don't know. I always search how to iterate trough lines in bash and it's the most poplar answer. I have never used `while` in bash except of this case.

PR #3644, opened 2022-02-17
fix: release workflow - prevent parallel publishing to sonatype

dos65:
Running this job in parallel leads to issues with publishing process.
Only one job was completed from the bunch of tags that was created for scala3-nightly: [#1004-#1015 runs](https://github.com/scalameta/metals/actions/workflows/release.yml)

Unfortunately, there is no way to configure this using std settings.
There is `job.concurrent` setting but it automatically [cancels all pending jobs](https://github.com/github/feedback/discussions/5435) from the same group.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,6 +15,19 @@ jobs:
       - uses: coursier/cache-action@v6.3
       - name: Publish
         run: |
+          LOCKED=true
+          while [ $LOCKED = true ]; do
+            # if there are several release jobs at the same time
+            # allow to continue the one which id is less than others
+            FIRST_IN_PROGRESS=$(gh api /repos/scalameta/metals/actions/workflows/release.yml/runs?status=in_progress -q ".workflow_runs[].id" | sort | head -n 1)
+            if [ $FIRST_IN_PROGRESS == $GITHUB_RUN_ID ]; then
+              LOCKED=false
+            else
+              echo "Waiting the completion of other release job..."
+              sleep 20
+            fi
+          done
+
           COMMAND="ci-release"
           UPDATE_DOCS=true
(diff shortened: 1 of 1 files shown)


tgodzik:
> There is job.concurrent setting but it automatically https://github.com/github/feedback/discussions/5435 from the same group.

But from what I understand it run the latest job, which would be perfect here, no? We are mostly interested in having the latest deployed.

tgodzik:
You would have `nightly-release-0.11.0-<NIGHTLY-VERSION>`, so essentially take the name from the tag or main.

dos65:
> Can't we name each release job differently, so that they don't cancel each other out?

Then there won't be concurrency limit and there will be issues with parallel uploading to sonatype.

So, the thing is that we need some kind of lock to guarantee that only one release job can publish to sonatype at one point in time. Others jobs should wait the completion and they shouldn't be cancelled.

Unfortunately, `job.concurrent` doesn't work for this case because it cancels other jobs.

tgodzik:
> > Can't we name each release job differently, so that they don't cancel each other out?
> 
> Then there won't be concurrency limit and there will be issues with parallel uploading to sonatype.
> 
> So, the thing is that we need some kind of lock to guarantee that only one release job can publish to sonatype at one point in time. Others jobs should wait the completion and they shouldn't be cancelled.
> 
> Unfortunately, `job.concurrent` doesn't work for this case because it cancels other jobs.

Is there a problem even if those are different artifacts? I am wondering why it hasn't been an issue when we did release in different repositories at the same time, or did we never catch it?

dos65:
@tgodzik 
> Is there a problem even if those are different artifacts?

yep

> I am wondering why it hasn't been an issue when we did release in different repositories at the same time, or did we never catch it?

I believe we never caught it

PR #3646, opened 2022-02-17
Cancel PRs automatically if new commits are pushed

tgodzik:


.github/workflows/ci.yml
```diff
@@ -4,6 +4,14 @@ on:
     branches:
       - main
   pull_request:
+
+concurrency:
+  # Taken from gradle/gradle
+  # On main, we don't want any jobs cancelled so the sha is used to name the group
+  # On PR branches, we cancel the job if new commits are pushed
+  group: ${{ (github.ref == 'refs/heads/main' ||  startsWith(github.ref, 'refs/tags') ) && format('contributor-pr-base-{0}', github.sha) || format('contributor-pr-{0}', github.ref) }}
```
dos65, line 12: Just curious - have found an explanation for this kind of `if/else`?
I've tried but without success

dos65 (approved):
Lets' try! :rocket:

.github/workflows/ci.yml
tgodzik: It's just the typescript thing with. You can join `Boolean` and other value so `true && something` will be something and similar with `||`

ckipp01:
This wasn't working as expected for me last night. I sort of assumed this mean you push a commit, and then you push another and it stops the first and starts the second. However,  in. my case it was the other way around. I second was halted _until_ the first was finished. Just to clarify, was is the exact intention here?

tgodzik:
> This wasn't working as expected for me last night. I sort of assumed this mean you push a commit, and then you push another and it stops the first and starts the second. However, in. my case it was the other way around. I second was halted _until_ the first was finished. Just to clarify, was is the exact intention here?

Damn, it was supposed to stop the first one and run the second. It might be waiting for the other job to cancel though, but when I was testing it, everything was working as expected :/

tgodzik:
It seems to have worked in https://github.com/scalameta/metals/actions/runs/1863084519 :thinking:

ckipp01:
🤔 I have rebased but maybe it was a funky timing thing. I'll check again and see.

tgodzik:
> Ahhh yea that was my bad. I was impatient laughing. When you first push up you see
> 
> And the little clock1 appears until it's finishing cancelling your other stuff. I didn't realize that's what it was doing and thought it was waiting for it to finish instead.

Ufff :sweat_smile:

PR #3649, opened 2022-02-18
Change displayBuildTarget to listBuildTargets

Arthurm1:
This removes the `target-info-display` command and adds the `list-build-targets` command

`list-build-targets` will return a list of build target names to the client.

It's then up to the client to prompt the user with those targets and send the `file-decode` command to Metals with the appropriate uri (e.g. `metalsDecode:file:///workspacePath/buildTargetName.metals-buildtarget`)


Arthurm1:
@ckipp01 Is this more what you had in mind?  See the linked VSCode PR for implementation on the client side.

Note it also has to deal with transforming the project URIs in the MetalsTreeView - not sure if VIM implements a TreeView?
They're of the format...
`projects:file:/root/metals/.mtags/?id=mtags3!/_root_/,`
and need to be in...
`metalsDecode:file:///root/metals/mtags3.metals-buildtarget`
The build target name is in the `id=` section

ckipp01:
> @ckipp01 Is this more what you had in mind? See the linked VSCode PR for implementation on the client side.

Works great for me @Arthurm1, thanks a lot for going back and address this.



> Note it also has to deal with transforming the project URIs in the MetalsTreeView - not sure if VIM implements a TreeView?

No the tree view that is in `nvim-metals` is very minimal so it doesn't implement this at all.

Arthurm1:
I don't think the failing tests are relevant...

```
tests.pc.InsertInferredTypeSuite
tests.hover.HoverScala3TypeSuite
```

tgodzik (approved):

tgodzik:
Thanks again for changing it! :tada:

PR #3650, opened 2022-02-21
Fix issues in tests for Scala 3 Nightlies

tgodzik:
This is related to a couple of fixes I did in the compiler itself.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
@@ -79,8 +79,8 @@ final class InferredTypeProvider(
         printer.typeString(short)
     /*
      * Get the exact position in ValDef pattern for val (a, b) = (1, 2)
-     * Suprisingly, val ((a, c), b) = ((1, 3), 2) will be covered by Bind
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -79,8 +79,8 @@ final class InferredTypeProvider(
         printer.typeString(short)
     /*
      * Get the exact position in ValDef pattern for val (a, b) = (1, 2)
-     * Suprisingly, val ((a, c), b) = ((1, 3), 2) will be covered by Bind
-     * https://github.com/lampepfl/dotty/issues/12627
+     * val ((a, c), b) = ((1, 3), 2) will be covered by Bind
+     * This is needed to Scala versions pre 3.1.2.
```
dos65, line 83: :thinking: I'm wondering how this function call is avoided for versions greater than 3.1.2.

tgodzik: I don't think this will get invoked for versions 3.1.2 and later since this should not go to that branch. Besides, this isn't really an issue, since the code is only invoked by the user in a code action.

dos65 (approved):

PR #3651, opened 2022-02-21
Scala3 PC refactorings

dos65:
Just a set of moves and renames:
- completions were extracted out from `ScalaPresentationCompiler` into a separate package
- printer was merged with `ShortenedNames` as all methods in existing printer impl require `ShortenedNames`


tanishiking:
I'll take a look tomorrow :) (as I'm somewhat familiar with the Completions part of code).

mtags/src/main/scala-3/scala/meta/internal/pc/completions/NamedArgCompletions.scala
```diff
@@ -0,0 +1,113 @@
...
+import dotty.tools.dotc.core.Flags
+import dotty.tools.dotc.core.NameKinds.DefaultGetterName
+import dotty.tools.dotc.core.Names.Name
+import dotty.tools.dotc.core.Symbols.Symbol
+import dotty.tools.dotc.interactive.Completion
+import dotty.tools.dotc.util.SourcePosition
+
+object NamedArgCompletions:
+
+  def contribute(
```
tanishiking, line 24: I named this method `contribute` just because I followed what Scala2 version of `CompletionPosition` abstract class does: https://github.com/tanishiking/metals/blob/0868c44b5bec2ff69e3cda2e51fa9ad8d8b779ba/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala#L348-L388 

As Scala3 version of completions don't have a common base class, we can rename `contribute` to `complete` or something that make more sense than `contribute`.
(It's OK to stay `contribute` in this PR!)

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImportsProvider.scala
```diff
@@ -8,6 +8,7 @@ import scala.jdk.CollectionConverters.*
 
 import scala.meta.internal.mtags.MtagsEnrichments.*
 import scala.meta.internal.pc.AutoImports.*
+import scala.meta.internal.pc.completions.CompletionPos
```
tanishiking, line 11: Unused import?

mtags/src/main/scala-3/scala/meta/internal/pc/printer/DotcPrinter.scala
```diff
@@ -0,0 +1,69 @@
...
+                else tpe.termSymbol
+
+              val text = super.toTextPrefix(tp)
+              if sym.is(ModuleClass) && indexed.toplevelClashes(sym) then
+                Str("_root_.") ~ text
+              else text
+            case _ => super.toTextPrefix(tp)
+        case _ => super.toTextPrefix(tp)
+    }
+  end ForInferredType
```
tanishiking, line 68: Can we remove `ScopeAwareTypePrinter.scala` now?

tanishiking (approved):
I left some nitpick comments, but LGTM from my side!
Now files around completions logic in Scala3 are more organized ✨ Great job!

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImportsProvider.scala
dos65: Actually it's used here because `autoImport` instead of just adding import might prepend owner prefix. Like: `List` -> `ju.List`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/NamedArgCompletions.scala
dos65: Yep, I think it fine leave it as `contribute` to keep the similarity with scala2 impl.

mtags/src/main/scala-3/scala/meta/internal/pc/printer/DotcPrinter.scala
dos65: done, thx!

PR #3652, opened 2022-02-21
Bump scalameta to 4.5.0

tgodzik:
This also allows us to use AnnonymousFunction instead of searching for placeholders to determine if we are in a function.


dos65 (approved):
:+1:

PR #3658, opened 2022-02-22
fix: support Scala 2.11 PC on jdk higher than 8

dos65:
There was an error during PC <init>: ` java.lang.NoClassDefFoundError: javax/tools/DiagnosticListener`.


tgodzik (approved):
LGTM! I will look at the issues with nightlies.

PR #3661, opened 2022-02-24
fix: do not show test explorer related error for client which don't implement it

### LINKED ISSUE #3660: Test Explorer won't work... message shouldn't be shown in clients that don't implement the test explorer
### Describe the bug

When using a client that doesn't implement the test explorer and Metals detects something that might now work right you get a warning that pops up. For example when using `nvim-metals` I hit on this today:

```
Test Explorer won't work due to mis-configuration. Select 'More information' to learn how to fix this problem..

Request Actions:
1. More information
2. Don't show again
Type number and <Enter> or click with the mouse (q or empty cancels):
```

### Expected behavior

`nvim-metals` doesn't implement the test explorer so any issues related to it shouldn't surface to the client. This is probably due to the problem resolver in metals looking for things like outdated junit interfaces even in situations where the explorer won't be used. There should probably be extra checks there.

### Operating system

macOS

### Editor/Extension

Nvim (nvim-metals)

### Version of Metals

0.11.1+161-ff365f7c-SNAPSHOT

_No response_

kpodsiad:
fixes #3660

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
@@ -45,7 +45,8 @@ final class Doctor(
       workspace,
       mtagsResolver,
       currentBuildServer,
-      javaHome
+      javaHome,
+      () => clientConfig.isCompletionItemResolve()
     )
 
   /**
--- a/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
@@ -46,7 +46,7 @@ final class Doctor(
       mtagsResolver,
       currentBuildServer,
       javaHome,
-      () => clientConfig.isCompletionItemResolve()
+      () => clientConfig.isTestExplorerProvider()
     )
 
   /**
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
```diff
@@ -19,7 +19,8 @@ class ProblemResolver(
     workspace: AbsolutePath,
     mtagsResolver: MtagsResolver,
     currentBuildServer: () => Option[BspSession],
-    javaHome: () => Option[String]
+    javaHome: () => Option[String],
+    isTestExplorerProvider: () => Boolean
```
kpodsiad, line 23: I'm not a fan of passing boolean flags, but it was easier than passing whole `clientConfig`

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -45,7 +45,8 @@ final class Doctor(
       workspace,
       mtagsResolver,
       currentBuildServer,
-      javaHome
+      javaHome,
+      () => clientConfig.isCompletionItemResolve()
```
tgodzik, line 49: This seems like the wrong config, no?

kpodsiad: wow, I don't have any idea why I did this 😳

ckipp01 (approved):
👍🏼 thanks for fixing this!

tgodzik (approved):
LGTM!

kpodsiad:
One test on windows is failing but it's not related to my changes so I'm merging.

PR #3665, opened 2022-02-25
Don't watch entire workspace on MacOS

Duhemm:
Previously, on MacOS, the root of the workspace would be registered for
file watching. With this patch, up to 32 of the longest prefixes will be
registered.

This change improves the performance in large monorepos, where the
workspace is set to the root of the monorepo, but usually few
subdirectories are imported in the IDE.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
@@ -39,6 +39,7 @@ import scala.meta.pc.PresentationCompilerConfig.OverrideDefFormat
  *                                       be turned off. By default this is on, but Metals only
  *                                       supports a small subset of this, so it may be problematic
  *                                       for certain clients.
+ * @param macOsMaxWatchRoots The maximum number of root directories to watch on MacOS.
  */
 final case class MetalsServerConfig(
     globSyntax: GlobSyntaxConfig = GlobSyntaxConfig.default,
@@ -89,7 +90,12 @@ final case class MetalsServerConfig(
     ),
     bloopPort: Option[Int] = Option(System.getProperty("metals.bloop-port"))
       .filter(_.forall(Character.isDigit(_)))
-      .map(_.toInt)
+      .map(_.toInt),
+    macOsMaxWatchRoots: Int =
+      Option(System.getProperty("metals.macos-max-watch-roots"))
+        .filter(_.forall(Character.isDigit(_)))
+        .map(_.toInt)
+        .getOrElse(32)
 ) {
   override def toString: String =
     List[String](
@@ -105,7 +111,8 @@ final case class MetalsServerConfig(
       s"ask-to-reconnect=$askToReconnect",
       s"icons=$icons",
       s"statistics=$statistics",
-      s"bloop-port=${bloopPort.map(_.toString()).getOrElse("default")}"
+      s"bloop-port=${bloopPort.map(_.toString()).getOrElse("default")}",
+      s"macos-max-watch-roots=${macOsMaxWatchRoots}"
     ).mkString("MetalsServerConfig(\n  ", ",\n  ", "\n)")
 }
 object MetalsServerConfig {
--- a/metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
@@ -12,6 +12,7 @@ import scala.meta.internal.metals.BuildTargets
 import scala.meta.internal.metals.Cancelable
 import scala.meta.internal.metals.Directories
 import scala.meta.internal.metals.MetalsEnrichments._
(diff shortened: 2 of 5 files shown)


Swoorup:
Out of curiousity how much perf improvement does this yield?

Duhemm:
> Out of curiousity how much perf improvement does this yield?

For the projects I tried in our monorepo, the indexing time went from 10+ minutes to about 30 seconds. I don't expect to see any improvement or degradation in repositories where the entire workspace is imported.

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
```diff
@@ -118,21 +118,31 @@ object FileWatcher {
...
       val trie = PathTrie(
         filesToWatch.sourceFiles ++ filesToWatch.sourceDirectories ++ filesToWatch.semanticdDirectories
       )
       val isWatched = trie.containsPrefixOf _
 
+      // Select up to `maxRoots` longest prefixes of all files in the trie for
+      // watching. Watching the root of the workspace may have bad performance
+      // implications if it contains many other projects that we don't need to
+      // watch (eg. in a monorepo)
+      val watchRoots = trie.longestPrefixes(workspace.getRoot(), maxRoots = 32)
```
Duhemm, line 136: The `32` here is completely arbitrary and worked well for us in the projects where I tried this change. What's the policy around adding configuration knobs? Are system properties the preferred option?

Swoorup:
> > Out of curiousity how much perf improvement does this yield?
> 
> For the projects I tried in our monorepo, the indexing time went from 10+ minutes to about 30 seconds. I don't expect to see any improvement or degradation in repositories where the entire workspace is imported.

Does that mean it only improves the startup time or throughout the IDE run session?

Duhemm:
This should only improve the startup time.

tgodzik (approved):
Looks good from my side! Not sure if we need to include a configuration option unless there are plans to experiment with it somewhere. This might be too in depth for the users to ever configure.

Duhemm:
@tgodzik Thanks for the review! I just updated the PR to add a system property `metals.macos-max-watch-roots`, which defaults to 32.

I don't know what's the actual maximum number of FSEvent streams on MacOS. I was able to create a project with about 1000 roots, and didn't encounter any issues with Metals. I think that 32 roots should be safe.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
```diff
@@ -89,7 +90,12 @@ final case class MetalsServerConfig(
     ),
     bloopPort: Option[Int] = Option(System.getProperty("metals.bloop-port"))
       .filter(_.forall(Character.isDigit(_)))
-      .map(_.toInt)
+      .map(_.toInt),
+    macOsMaxWatchRoots: Int =
+      Option(System.getProperty("metals.macos-max-watch-roots"))
+        .filter(_.forall(Character.isDigit(_)))
+        .map(_.toInt)
+        .getOrElse(32)
```
kpodsiad, line 98: Could we default to 1? I find it more adequate assuming that most of the people work in small to medium-sized projects (pet projects, work ones) and only a few percent works with big monorepos. 
With 32 as a default, a lot of these ordinary people (not power-users) will see the message `INFO  Registering root for file watching: ...` being printed multiple times in Metals output and they may wonder why this suddenly changed. Sometimes it can be very misleading e.g. registering file watchers for sbt's aggregate projects will register watchers for phantom `src` and `test` in projects root (since it doesn't have sources or tests, it just aggregates other projects).

kpodsiad (approved):
I created artificial monorepo and I got an over 10x import speedup! That's awesome for monorepos.

I have only one suggestion regarding the default value of the newly created property.

metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala
Duhemm: For projects where the semantic db and class files are stored outside of the workspace root, this would result in a performance degradation when using the default configuration (their common parent would be watched). This can be solved by configuring Metals, but I feel like since 32 should be a safe value for everyone, this is an unnecessary step. But I'm perfectly open to changing it to 1 if you feel strongly about it.

Would you prefer the log message is moved to debug?

tgodzik: Maybe let's change the message if `32` is safe. :+1:

Duhemm: I set it to debug. Thank you!

kpodsiad: I was mostly worried about confusing messages, currently with the debug it's fine. Thanks!

objmagic:
Thanks for working on this! Can we make a release that includes this change?

Issue #1691, opened 2020-05-05
Weird format of scala.Left on hover


bishabosha:
**Describe the bug**
The tooltip on hover on `Left` gives the type of its argument as its type

**To Reproduce**
Open a fresh file foo.scala in a `0.23.0` project

```scala
def foo = Left("") // tooltip: class Left: String
//        ^ cursor here
```

**Expected behavior**
A tooltip describing Left.apply

**Installation:**
- Operating system: macOS
- Editor: Visual Studio Code
- Metals version: v0.9.0

**Search terms**
scala 3, dotty, hover

tgodzik:
Thanks for reporting! I do agree it looks weird, this should be just a fix of how we present the generic classes.

ltrojanowski:
couldn't reproduce this with
![Screencast-2020-06-05-16_11_06](https://user-images.githubusercontent.com/3951626/83886504-707d5500-a747-11ea-9486-96ec92b97162.gif)

bishabosha:
@ltrojanowski i copied what you did and get the same result, you must put the mouse cursor over "Left" and leave the caret on a different location of the source code

tgodzik:
I think this is due to the fact that the `apply` there is synthetic. This might be a bit more complex.

PR #3666, opened 2022-02-28
fix: Add test case when using Left class and Scala 3

tgodzik:
Fixes https://github.com/scalameta/metals/issues/1691

### DIFF
--- a/tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
+++ b/tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
@@ -173,4 +173,12 @@ class HoverScala3TypeSuite extends BaseHoverSuite {
        |def apply[T](a: T)(using Int): T
        |""".stripMargin.hover
   )
+
+  check(
+    "toplevel-left",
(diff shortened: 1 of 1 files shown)


tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
```diff
@@ -173,4 +173,13 @@ class HoverScala3TypeSuite extends BaseHoverSuite {
        |def apply[T](a: T)(using Int): T
        |""".stripMargin.hover
   )
+
+  check(
+    "toplevel-left",
+    """
+      |def foo = <<L@@eft>>("")
+      |""".stripMargin,
```
kpodsiad, line 181: If you add pipe to the first line then both strings will be aligned :D

kpodsiad (approved):

tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
tgodzik: Fixed!

PR #3673, opened 2022-03-02
chore: fix flaky test provider suite

kpodsiad:
Sometimes test provider suites was failing with: obtained `"kind": "addTestCases"` expected `"kind": "addSuite"`. Tests checks only first event send to the client and if this event is `addTestCases` then something isn't right, it should be `addSuite` followed by the `addSuite`.

I think it was due to the following:
- when searching for the test suite, if one was found then it was added to cached test suites
- for junit suites test provider load semanticDB
- loading semanticDB was calling `onChange` which was sending `addTestCases` event

tldr; side effects, it's always side effect :p

I delayed the moment when suite is added to cached suites.


tgodzik (approved):
LGTM

Issue #1687, opened 2020-05-04
No signature help inside pattern match for Scala 3


tgodzik:
**Describe the bug**
```
case Some(@@) =>
```

With the cursor at @@ there will be no signature help coming from the unapply of Option or any other class.

**To Reproduce**
1. Create pattern match for a class with an unapply method
2. Try to use signature help inside the pattern match
3. No signature help available

**Expected behavior**
Signature help shows the signature of an unapply method. 

**Installation:**
- Operating system: macOS/Windows/Linux
- Editor: Visual Studio Code/Atom/Vim/Sublime/Emacs
- Metals version: v0.8.4

**Additional context**
This might need to be fixed inside the Dotty compiler - might require some help from the compiler team.

**Search terms**
scala3 unapply signature help

bishabosha:
in `0.24.0-RC1` i can hover on the argument of the `Some` pattern fine:

```scala
def foo(o: Option[String]) = o match
  case Some(s) => s + s
  case _ => ""
```

tgodzik:
@bishabosha Hover works, that's true. Signature help on the other hand should be shown while you are writting the arguments or via `show signature help` shortcut.

tgodzik:
Looks like has been fixed in the compiler.

PR #3675, opened 2022-03-02
fix: [Scala 3] Add support for signature help in unapply method

tgodzik:
Fixes https://github.com/scalameta/metals/issues/1687

Related fix in Dotty: https://github.com/lampepfl/dotty/pull/14611

Most of the code is copied over from either the compiler or ScalaPResentationCompiler class.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
@@ -12,8 +12,8 @@ import dotty.tools.dotc.core.Constants.Constant
 import dotty.tools.dotc.core.Contexts.Context
 import dotty.tools.dotc.core.Denotations.SingleDenotation
 import dotty.tools.dotc.core.Symbols
+import dotty.tools.dotc.core.Symbols.*
 import dotty.tools.dotc.core.Types.ErrorType
-import dotty.tools.dotc.core.Types.MethodType
 import dotty.tools.dotc.interactive.Interactive
 import dotty.tools.dotc.interactive.InteractiveDriver
 import dotty.tools.dotc.reporting.AmbiguousOverload
@@ -24,6 +24,12 @@ import dotty.tools.dotc.util.Spans.Span
 import org.eclipse.lsp4j.ParameterInformation
 import org.eclipse.lsp4j.SignatureHelp
 import org.eclipse.lsp4j.SignatureInformation
+import dotty.tools.dotc.core.Types.PolyType
+import dotty.tools.dotc.core.Types.Type
+import dotty.tools.dotc.core.Types.MethodType
+import scala.annotation.tailrec
+import scala.collection.immutable.LazyList.cons
+import dotty.tools.dotc.core.Flags
 
 object SignatureHelpProvider:
 
@@ -33,7 +39,6 @@ object SignatureHelpProvider:
     driver.run(uri, sourceFile)
 
     given ctx: Context = driver.currentCtx
-    // given locatedCtx: Context = driver.localContext(params)
     val pos = driver.sourcePosition(params)
     val trees = driver.openedTrees(uri)
     val path =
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
```diff
@@ -0,0 +1,163 @@
...
+      callableN,
+      paramN
+    )
+  end contribute
+
+  private def callInfo(path: List[Tree], span: Span)(using
+      ctx: Context
+  ): (Int, Int, List[SingleDenotation]) =
+    path match
+      case (tpd.UnApply(fun, implicits, patterns)) :: _ =>
```
tgodzik, line 67: This is the main fix, we apply almost the same logic as for Apply in `Signatures`

mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
```diff
@@ -0,0 +1,163 @@
...
+    driver.run(uri, sourceFile)
+
+    given ctx: Context = driver.currentCtx
+    // given locatedCtx: Context = driver.localContext(params)
+    val pos = driver.sourcePosition(params)
+    val trees = driver.openedTrees(uri)
+    val path =
+      Interactive
+        .pathTo(trees, pos)(using ctx)
+        .dropWhile {
```
tgodzik, line 42: This is another place that actually got changes.

mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
```diff
@@ -0,0 +1,163 @@
...
+          case _: tpd.Apply => false
+          case unapply: tpd.UnApply =>
+            // Don't show tuple unapply method
```
tanishiking, line 45: What happens if we show the tuple unapply method?

mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
tanishiking: Are we going to remove this pattern match once https://github.com/lampepfl/dotty/pull/14611 merged? In this case, can we wait for the next release that contains https://github.com/lampepfl/dotty/pull/14611 ?


Or leave this method for older Scala3 compilers? In this case, can we add a comment on why there's a kind of duplicated logic in metals and dotty?

tanishiking (commented):
Metals part (`SignatureHelpProvider.contribute`) looks good to me :)
But I'm wondering maybe we should wait for the next release of Scala3 compiler 🤔

mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
tgodzik: I think we should leave it as a fallback, but I still need to work a bit on this.

mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
kpodsiad: +1

tgodzik: It's just not really useful, since when writing tuples we are not really interested in signature help.

tests/cross/src/test/scala/tests/pc/SignatureHelpPatternSuite.scala
```diff
@@ -54,13 +56,19 @@ class SignatureHelpPatternSuite extends BaseSignatureHelpSuite {
...
       |}
       |""".stripMargin,
     """|unapply(a: T, b: T): Two[T]
-       |        ^^^^
-       |""".stripMargin
+       |              ^^^^
+       |""".stripMargin,
+    compat = Map(
+      "3" ->
+        """|Two[T](a: T, b: T)
```
tgodzik, line 68: I reworked the unapply signature for case classes to show the actual case classes. It's a bit different than in case of Scala 2, but I think this is a bit better since `def unapply(a: T, b: T): Two[T]` does not exist. The alternative would be to show the parameters alone.

tgodzik:
The failures are related to my tinkering in the compiler.

mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
```diff
@@ -0,0 +1,162 @@
...
+              if isExplicitUnapply then implicitsBefore
+              else defaultParamIndex
+
+            val signatureSymbol =
+              if isSomeMatch then
+                funSymbol.owner.companionClass.primaryConstructor
+              else if isExplicitUnapply then funSymbol
+              else
+                // if unapply doesn't return option it means it's a case class
+                retType.typeSymbol.primaryConstructor
```
tanishiking, line 97: [Note] Here we find the symbol of Constructor (e.g. `Two` for `Two.unapply`) 👍

mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
```diff
@@ -0,0 +1,162 @@
...
+                case n => n + implicitsBefore
+
+            val funSymbol = fun.symbol
+            val retType = funSymbol.info.finalResultType
+            val isSomeMatch = funSymbol.owner.companionClass == defn.SomeClass
+            val isExplicitUnapply =
+              !isSomeMatch && retType.typeSymbol == defn.OptionClass
+            val paramIndex =
+              if isExplicitUnapply then implicitsBefore
+              else defaultParamIndex
```
tanishiking, line 89: Why we use `implicitsBefore` for `Foo.unapply(@@)` ?

tanishiking (approved):
Diff seems good! And showing the actual case classes for unapply signature of case classes is a great idea 👍 

I just left some questions (as it seems a bit tricky for me), it could be helpful if you can leave some code comments answering my questions. (those comments may help future us to comprehend the program around here :)

mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
```diff
@@ -0,0 +1,162 @@
...
+            val implicitsBefore = countParams(fun)
+
+            def defaultParamIndex =
+              patterns.indexWhere(_.span.contains(span)) match
+                case -1 => (patterns.length - 1 max 0) + implicitsBefore
+                case n => n + implicitsBefore
+
+            val funSymbol = fun.symbol
+            val retType = funSymbol.info.finalResultType
+            val isSomeMatch = funSymbol.owner.companionClass == defn.SomeClass
```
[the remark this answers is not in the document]
tgodzik: Yeas, that's the special case I wanted to handle.

mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
tgodzik: You can have in Scala 3 `def unapply(using A)(value: Value) = ???` so we need to highlight the first parameter after the using params.

mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
```diff
@@ -0,0 +1,162 @@
...
+            val isExplicitUnapply =
+              !isSomeMatch && retType.typeSymbol == defn.OptionClass
```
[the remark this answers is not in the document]
tgodzik: >isExplicitUnapply will be true when we explicitly call unapply method like Foo.unapply(@@)

It might be, but we will be at an Apply node, so the code will not be invoked.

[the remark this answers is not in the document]
tanishiking: Auch, I got it! `isExplicitUnapply = true` means `unapply` method is **explicitly defined**, not **explicitly called**

mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
tanishiking: I got it! Okay, so `implicitsBefore` means (the number of) "**implicit** parameter**s before**" (actual parameter(?) of the constructor) :+1:

tanishiking (approved):
Again it's LGTM. This one is what I'd been longing for in Scala3. I look forward to using this feature!

dos65 (approved):
Cool! LGTM!

tests/cross/src/test/scala/tests/pc/SignatureHelpPatternSuite.scala
```diff
@@ -95,4 +109,24 @@ class SignatureHelpPatternSuite extends BaseSignatureHelpSuite {
...
+      |}
+      |object Main {
+      |  given b: Boolean = true
+      |  given str: String = ""
+      |  (null: Any) match {
+      |    case HKT(1, @@) =>
+      |  }
+      |}
+      |""".stripMargin,
+    """|unapply(implicit x$1: String)(implicit x$2: Boolean)(a: Int): Option[(Int, Int)]
```
kpodsiad, line 128: For scala 3 shouldn't `implicit` be called `using`?

tanishiking: can be fixed with https://github.com/scalameta/metals/pull/3707 :)

tgodzik:
Closing this for now, we might get help from the compiler team for this :tada:

PR #3678, opened 2022-03-03
feat: implement test selection request for sbt

kpodsiad:
Previously only bloop was capable of handling test selection suites request. Now, with sbt capable of handling it (https://github.com/scalacenter/scala-debug-adapter/pull/198), we have to change a few details in implementation. The most important ones are changes in "bsp" model which now are more aligned to the bsp spec and sbt's contraband capabilities (for example contraband cannot express `Map[String, String]` so env is passed as `List[String]`).

blocked on: https://github.com/scalameta/metals/pull/3706
needs: https://github.com/scalameta/metals-vscode/pull/917


tgodzik (approved):
LGTM!

PR #3683, opened 2022-03-05
Go to definition of synthetic symbols.

olafurpg:
Previously, Metals didn't go the definition of symbols that the
presentation compiler defines as synthetic. These symbols include
auto-generated case class `apply` methods and regular class
constructors. Now, we ignore the `isSynthetic` bit on the definition
symbol so that "Go to definition" works as expected for cases where it
didn't before.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
@@ -136,7 +136,6 @@ class PcDefinitionProvider(
 
       case (head @ Select(target, name)) :: _
           if head.symbol.is(Synthetic) && name == StdNames.nme.apply =>
-        println(head)
         val sym = target.symbol
         if sym.is(Synthetic) && sym.is(Module) then List(sym.companionClass)
         else List(target.symbol)
--- a/tests/cross/src/test/scala/tests/pc/PcDefinitionSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/PcDefinitionSuite.scala
@@ -30,13 +30,25 @@ class PcDefinitionSuite extends BasePcDefinitionSuite {
     """|
        |object Main {
        |  for {
-       |    <<x>> <- List(1)
+       |    <<>>x <- List(1)
        |    y <- 1.to(x)
        |    z = y + x
        |    if y < @@x
        |  } yield y
        |}
-       |""".stripMargin
+       |""".stripMargin,
+    compat = Map(
+      "3" ->
+        """|object Main {
+           |  for {
+           |    <<x>> <- List(1)
+           |    y <- 1.to(x)
+           |    z = y + x
+           |    if y < x
+           |  } yield y
+           |}
+           |""".stripMargin
+    )
   )
 
   check(
@@ -225,7 +237,7 @@ class PcDefinitionSuite extends BasePcDefinitionSuite {
     "named-arg-local",
     """|
        |object Main {
-       |  <<def foo(arg: Int): Unit = ()>>
+       |  def <<>>foo(arg: Int): Unit = ()
        |
        |  foo(a@@rg = 42)
        |}
@@ -295,7 +307,15 @@ class PcDefinitionSuite extends BasePcDefinitionSuite {
        |object Main {
        |  val n = ma@@th.max(1, 2)
        |}
-       |""".stripMargin
+       |""".stripMargin,
+    compat = Map(
+      "3" ->
+        """|
+           |object Main {
+           |  val n = ma@@th.max(1, 2)
+           |}
+           |""".stripMargin
+    )
   )
 
   check(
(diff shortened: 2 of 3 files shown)


olafurpg:
I tried fixing the failing Scala 3 tests but I'm not familiar with that part of the codebase. I'd appreciate any help to get the Scala 3 tests green

tanishiking (commented):
Scala2 part looks good 👍 
In terms of Scala3, it looks like [enclosingSymbols](https://github.com/scalameta/metals/blob/7871d17c7ea19d9b54c5af6025d242d7e5f5a0e1/mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala#L137-L147) seems the root cause, as all the inputs (of failing tests) are swallowed by [this match case](https://github.com/tanishiking/metals/blob/7871d17c7ea19d9b54c5af6025d242d7e5f5a0e1/mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala#L137-L139).
If I remove that case, I got the following test result for `eta-2`, which is still different from the expected result 🤔 


```diff
=> Obtained
    """|object Main {
       |  List(1).foldLeft(0)(_ + /*scala/Int#`+`(+4). Int.scala*/_)
       |}
       |""".stripMargin
=> Diff (- obtained, + expected)
 object Main {
-  List(1).foldLeft(0)(_ + /*scala/Int#`+`(+4). Int.scala*/_)
+  List(1).foldLeft(0)(_ + <<>>_)
 }
    at munit.FunSuite.assertNoDiff(FunSuite.scala:11)
    at tests.pc.BasePcDefinitionSuite.check$$anonfun$1(BasePcDefinitionSuite.scala:68)
```

tanishiking:
I partly reviewed the PR, but I don't have time to take a look for a few days. If someone can delve into the problem with Scala3, please go for it :)

tests/cross/src/test/scala/tests/pc/PcDefinitionSuite.scala
```diff
@@ -351,4 +360,26 @@ class PcDefinitionSuite extends BasePcDefinitionSuite {
...
+       |  hello(Us@@er())
+       |}
+       |""".stripMargin
+  )
+
+  check(
+    "synthetic-definition-class-constructor",
+    """|
+       |class Main {
+       |  <<class User(name: String, age: Int)>>
```
dos65, line 379: position isn't perfect. 
It would be nice if it would point at least at the beginning of `User`

olafurpg: Agreed, done

olafurpg:
@tanishiking Thank you for the review! I added `compat = Map("3" -> ..)` assertions to make the Scala 3 tests pass.

dos65:
@olafurpg I made a pr onto your branch with fixes for scala3 - https://github.com/olafurpg/metals/pull/4

olafurpg:
Thank you @dos65 ! I merged your branch and updated this PR

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
```diff
@@ -130,13 +130,16 @@ class PcDefinitionProvider(
...
-        List(select.symbol)
-
       // wildcard param
       case head :: _ if (head.symbol.is(Param) && head.symbol.is(Synthetic)) =>
-        Nil
+        List(head.symbol)
+
+      case (head @ Select(target, name)) :: _
+          if head.symbol.is(Synthetic) && name == StdNames.nme.apply =>
+        println(head)
```
tanishiking, line 139: print-debug

tanishiking (approved):
I checked @dos65 's PR: https://github.com/olafurpg/metals/pull/4 and now it's LGTM!
(except one nitpick comment)

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
olafurpg: Good catch! Fixed

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
```diff
@@ -130,13 +130,15 @@ class PcDefinitionProvider(
...
-        List(select.symbol)
-
       // wildcard param
       case head :: _ if (head.symbol.is(Param) && head.symbol.is(Synthetic)) =>
-        Nil
+        List(head.symbol)
+
+      case (head @ Select(target, name)) :: _
+          if head.symbol.is(Synthetic) && name == StdNames.nme.apply =>
+        val sym = target.symbol
```
kpodsiad, line 139: Could you add a comment which case is handled by this match for future us?

kpodsiad (approved):
Awesome!

dos65 (approved):
:+1:

mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala
dos65: it's for synthetic apply method in case-class

Issue #2400, opened 2021-01-18
Autocompletion for named method arguments not working for args contains $


tanishiking:
Original issue https://github.com/scalameta/metals/issues/2348 reported by @nkgm

Auto-completion for named args not working when the name contains `$`

**To Reproduce**

- Define a method whose arg start with `$`
  - like `def test($foo: String, $bar: String): Unit`
- Try to autocomplete its args
  - `test($@@`
- No args are completed

**Expected behavior**
The completions `$foo =` and `$bar = ` should appear as completion candidates.

**Screenshots**

**Installation:**

- Operating system: macOS
- Editor: Visual Studio 
- Metals version: v0.9.8

**Additional context**

- args contain `$` are excluded from arg completion here https://github.com/scalameta/metals/blob/15c0a47418ef6312d0913ded846b011d8d7cd2da/mtags/src/main/scala-2/scala/meta/internal/pc/completions/ArgCompletions.scala#L42 to exclude the synthesized params like `evidence$`.
- We have to modify the way to exclude the synthesized params, it might be better to use the `SYNTHESIZED` flag attached to their symbols

**Search terms**

`autocomplete` `named arguments` `default parameter values`

PR #3686, opened 2022-03-08
Fix named arg completion when parameter contains $

tanishiking:
Fix https://github.com/scalameta/metals/issues/2400

dollar-arg-completion

- Check the parameter is synthetic by `.isSynthetic` instead of $
- Escape `$` from text edit. Otherwise, `$foo will be treated as a parameter in snippet mode.
  - https://code.visualstudio.com/docs/editor/userdefinedsnippets#_snippet-syntax

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
@@ -238,7 +238,7 @@ class CompletionsProvider(
                     mkItem(ident, ident.backticked)
                   case _ => mkWorkspaceItem(ident, sym.fullNameBackticked)
       case CompletionValue.NamedArg(label, _) =>
-        mkItem(ident, ident.replace("$", "\\$")) // escape $ for snippet
+        mkItem(ident, ident.replace("$", "$$")) // escape $ for snippet
       case CompletionValue.Keyword(label, text) => mkItem(label, text)
       case _ => mkItem(ident, ident.backticked)
     end match
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
@@ -237,7 +237,7 @@ class CompletionsProvider(
                   case IndexedContext.Result.InScope =>
                     mkItem(ident, ident.backticked)
                   case _ => mkWorkspaceItem(ident, sym.fullNameBackticked)
-      case CompletionValue.NamedArg(label, _) => mkItem(ident, ident)
+      case CompletionValue.NamedArg(label, _) => mkItem(ident, ident.replace("$", "\\$")) // escape $ for snippet
       case CompletionValue.Keyword(label, text) => mkItem(label, text)
       case _ => mkItem(ident, ident.backticked)
     end match
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
@@ -237,7 +237,8 @@ class CompletionsProvider(
                   case IndexedContext.Result.InScope =>
                     mkItem(ident, ident.backticked)
                   case _ => mkWorkspaceItem(ident, sym.fullNameBackticked)
-      case CompletionValue.NamedArg(label, _) => mkItem(ident, ident.replace("$", "\\$")) // escape $ for snippet
+      case CompletionValue.NamedArg(label, _) =>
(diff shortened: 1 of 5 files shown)


tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -231,6 +231,33 @@ class CompletionArgSuite extends BaseCompletionSuite {
...
+  checkSnippet( // known issue: the second parameter with $ become | (returned from compiler)
+    "explicit-dollar-autofill".tag(IgnoreScala3),
+    """
+      |object Main {
+      |  def test($foo: Int, $bar: Int): Int = ???
+      |  test($f@@)
+      |}
+      |""".stripMargin,
+    """|\$foo = 
+       |\$foo = ${1:???}, | = ${2:???}
```
tanishiking, line 256: This might require the compiler fix because we get a symbol `|` instead of `$bar` here :thinking: https://github.com/scalameta/metals/blob/72ef7297e6b4527a3305d67d8f858431b028f92e/mtags/src/main/scala-2/scala/meta/internal/pc/completions/ArgCompletions.scala#L23-L28

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -231,6 +231,33 @@ class CompletionArgSuite extends BaseCompletionSuite {
...
 
+  checkSnippet( // see: https://github.com/scalameta/metals/issues/2400
+    "explicit-dollar",
+    """
+      |object Main {
+      |  def test($foo: Int, $bar: Int): Int = ???
+      |  test($f@@)
+      |}
+      |""".stripMargin,
+    """|\$foo =
```
tanishiking, line 242: Checking `$` is escaped (otherwise, $foo will be interpreted as a value in vscode's snippet)

tanishiking:
test should be fixed by https://github.com/scalameta/metals/pull/3687

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
```diff
@@ -237,7 +237,8 @@ class CompletionsProvider(
                   case IndexedContext.Result.InScope =>
                     mkItem(ident, ident.backticked)
                   case _ => mkWorkspaceItem(ident, sym.fullNameBackticked)
-      case CompletionValue.NamedArg(label, _) => mkItem(ident, ident)
+      case CompletionValue.NamedArg(label, _) =>
+        mkItem(ident, ident.replace("$", "\\$")) // escape $ for snippet
```
dos65, line 241: I'm wondering if it might be worth escaping `$` for all completion items.
`$` char might come not only from NamedArg

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
kpodsiad: I think I have already seen escaping `$` by `$$`. Could we stick to double dollars?

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
kpodsiad: Is there an issue for that? If not, could you create it and leave a link above the test?

kpodsiad (approved):

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
tanishiking: There's no ticket yet, I'll open an issue and link to that :+1:

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
tanishiking: done https://github.com/scalameta/metals/pull/3686/commits/52903b9480df042f0b99988fc2fae7636848bdd8

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
tanishiking: https://github.com/scalameta/metals/pull/3686/commits/5fcdc95917b3cfa0202d8732952d5d347e3b4d85

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
tanishiking: That's fair enough, but we have to carefully escape `$` because we can't escape in [snippet syntax of LSP](https://github.com/microsoft/language-server-protocol/blob/main/snippetSyntax.md).
I couldn't figure out what is the best way to escape `$` from whole completion items without breaking the snippet syntax. Therefore, let me limit to NamedArg here :bow:

dos65 (approved):

PR #3687, opened 2022-03-08
fix: adjust tests for 3.2.0-NIGHTLY

dos65:
Also, exnteded the `compat` inn tests cases.
Now it's possible to specify matching version not only as starts with
but as greater or equal.

For example:
```
check(
  "input ... ",
  "default expected ...",
  compat = Map(
   ">=3.1.0" -> "expected for version gte than 3.1.0",
   "3" -> "any otherversion that starts from 3"
  )
)
```


mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
```diff
@@ -72,7 +72,9 @@ object CompletionValue:
   case class Keyword(label: String, insertText: String) extends CompletionValue
 
   def fromCompiler(completion: Completion): List[CompletionValue] =
-    completion.symbols.map(Compiler(completion.label, _))
+    def undoBacktick(label: String): String =
+      label.stripPrefix("`").stripSuffix("`")
```
ckipp01, line 76: I'm assuming these are due to my changes in https://github.com/lampepfl/dotty/pull/14594. Is this how we want to handle this though? In these cases aren't we stripping them here and then adding them back later? Would it be better to instead check when we are about to add them and if they exist already, then don't add them?

dos65: @ckipp01 I agree that it isn't perfect but they anyway should be removed from completion label

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
```diff
@@ -1,79 +1,66 @@
...
+      val Array(major, minor, patch) = parts.take(3).map(_.toInt)
+      val (rc, milestone) = parts
+        .lift(3)
+        .map { v =>
+          if (v.startsWith("RC")) (Some(v.stripPrefix("RC").toInt), None)
+          else (None, Some(v.stripPrefix("M").toInt))
+        }
+        .getOrElse((None, None))
+      val date = parts.lift(5).map(_.toInt)
+      Version(major, minor, patch, rc, milestone, date)
```
tanishiking, line 61: Basically same with https://github.com/tanishiking/metals/blob/72ef7297e6b4527a3305d67d8f858431b028f92e/project/Scala3NightlyVersions.scala#L73-L87

tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
```diff
@@ -81,7 +81,7 @@ class HoverScala3TypeSuite extends BaseHoverSuite {
        |""".stripMargin,
     "",
     compat = Map(
-      "3.1.3-RC1" ->
+      ">=3.1.3-RC1-bin-20220301-fae7c09-NIGHTLY" ->
```
tanishiking, line 84: Can we say `>3.1.3-RC1` here? (this is not a strong opinion. If you prefer this way, go ahead 👍  )

-  If we write `>3.1.3-RC1`, the nightly releases of `3.1.3-RC1` before `20220301` will fail, but do we test against older Scala3 nightly release?  
- I feel it's too specific
- If we can write `>3.1.3-RC1`,  we can omit `nightlyDate` field from `Version` case class, and keep it simple :)

tanishiking (commented):
Except for @ckipp01 's comment, it's LGTM from my side (I left one comment).

tests/cross/src/test/scala/tests/hover/HoverScala3TypeSuite.scala
dos65: @tanishiking Initially I wanted to use just `>` but then realized that the actual usage of this might be gte only. Some change happens in specific version so you add a new compat case that should be used from this version and next ones. 
I agree that `>=` is more verbose than `>` but I think at least it's more clear.

> If we write >3.1.3-RC1, the nightly releases of 3.1.3-RC1 before 20220301 will fail, but do we test against older Scala3 nightly release?

We have a ci workflow that tests and backpublishes support of new scala3-nightly versions for the latest Metals version.
Being too specific with versions will help to stabilize this workflow as currently it often fails.

tanishiking: Thank you for the detailed explanation! 

> We have a ci workflow that tests and backpublishes support of new scala3-nightly versions for the latest Metals version.
> Being too specific with versions will help to stabilize this workflow as currently it often fails.

I'm convinced that being specific with versions would be better now.
Now, it's LGTM from my side.

tests/mtest/src/main/scala/tests/Compat.scala
```diff
@@ -0,0 +1,39 @@
...
+      scalaVersion: String,
+      cases: Map[String, A]
+  ): Option[A] = {
+    val (startsWith, gt) =
+      cases.partition({ case (v, _) => !v.startsWith(">=") })
+
+    val fromStartWith = startsWith.collectFirst({
+      case (v, a) if scalaVersion.startsWith(v) => a
+    })
+    matchesGte(scalaVersion, gt) orElse fromStartWith
```
kpodsiad, line 22:
```suggestion
    val (startsWith, gt) =
      cases.partition { case (v, _) => !v.startsWith(">=") }

    val fromStartWith = startsWith.collectFirst {
      case (v, a) if scalaVersion.startsWith(v) => a
    }

    matchesGte(scalaVersion, gt).orElse(fromStartWith)
```
few braces less ;)

tests/mtest/src/main/scala/tests/Compat.scala
```diff
@@ -0,0 +1,39 @@
...
+  private def matchesGte[A](
+      version: String,
+      gtCases: Map[String, A]
+  ): Option[A] = {
+    val parsed = SemVer.Version.fromString(version)
+    val less = gtCases
+      .map { case (v, a) =>
+        (SemVer.Version.fromString(v.stripPrefix(">=")), a)
+      }
+      .filter({ case (v, _) => parsed >= v })
```
kpodsiad, line 34:
```suggestion
      .filter { case (v, _) => parsed >= v }
```

tests/mtest/src/main/scala/tests/Compat.scala
```diff
@@ -0,0 +1,39 @@
+package tests
+
+import scala.meta.internal.semver.SemVer
+
+object Compat {
+
+  /**
+   * Cases might be described as:
+   *  - Starts swith: "3.0.1" -> value | "3.0" -> value | "3"
```
kpodsiad, line 9:
```suggestion
   *  - Starts with: "3.0.1" -> value OR "3.0" -> value OR "3" -> value
```

kpodsiad (approved):

tests/mtest/src/main/scala/tests/Compat.scala
dos65: thx for suggestions! applied all of them

Issue #3689, opened 2022-03-08
Flaky printing in `dependencies` WorkspaceSymbolLspSuite`.


ckipp01:
### Describe the bug

The `dependencies` test is currently marked flaky in `WorkspaceSymbolLspSuite` since the references that are returned on the same line seem to keep flopping when you run it locally or in CI causing some non-deterministic behavior. You can witness this with the results that are returned for lines 524 and 550.

### Expected behavior

I'd expect for the returned values to always be in the same order.

PR #3691, opened 2022-03-09
Fix flaky WorkspaceSymbolLspSuite test

tgodzik:
Fixes https://github.com/scalameta/metals/issues/3689

### DIFF
--- a/tests/unit/src/main/scala/tests/TestingServer.scala
+++ b/tests/unit/src/main/scala/tests/TestingServer.scala
@@ -1352,6 +1352,10 @@ final class TestingServer(
     )
     server.references(params).asScala.map { r =>
       r.asScala
+        .sortBy { l =>
+          val start = l.getRange().getStart()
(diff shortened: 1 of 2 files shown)


tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -1352,6 +1352,10 @@ final class TestingServer(
     )
     server.references(params).asScala.map { r =>
       r.asScala
+        .sortBy { l =>
```
ckipp01, line 1355: Ahh, ha much easier than I thought. Awesome. Out of curiosity, how come this comes back in different orders? I thought this always came in the correct order?

tgodzik: I think it might be sorted just by the lines somewhere, which is normally not an issue since we just highlight the correct place.

kpodsiad (approved):
😎

Issue #3688, opened 2022-03-08
Ignored DefinitionCrossLspSuite test fails in CI but not locally.


ckipp01:
### Describe the bug

The `underscore` test in `DefinitionCrossLspSuite` fails in CI but not locally.

### Expected behavior

I'd expect this to pas.

NOTE: I'll include some more details on this, but just creating this for now to get a ticket number to put in https://github.com/scalameta/metals/pull/3631

PR #3692, opened 2022-03-09
 Fix DefinitionCrossLspSuite

tgodzik:
Previously, we were not waiting for the compilation to finish which was causing the semanticdb not to be available. Now we properly use the for comprehension.

Fixes https://github.com/scalameta/metals/issues/3688

### DIFF
--- a/tests/slow/src/test/scala/tests/feature/DefinitionCrossLspSuite.scala
+++ b/tests/slow/src/test/scala/tests/feature/DefinitionCrossLspSuite.scala
@@ -29,10 +29,7 @@ class DefinitionCrossLspSuite
     }
   }
 
-  // TODO MIGRATE I can't reproduce this locally. In CI this fails every time,
-  // but I it always passes locally
-  // https://github.com/scalameta/metals/issues/3688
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):
Nice job chasing this down! LGTM

tests/slow/src/test/scala/tests/feature/DefinitionCrossLspSuite.scala
```diff
@@ -59,8 +56,9 @@ class DefinitionCrossLspSuite
            |}
            |""".stripMargin
       )
-      _ = server.didOpen("a/src/main/scala/a/Main.scala")
-      _ = server.didOpen("a/src/main/scala/a/Test.scala")
+      _ <- server.didOpen("a/src/main/scala/a/Main.scala")
+      _ <- server.didOpen("a/src/main/scala/a/Test.scala")
+      _ = assertNoDiagnostics()
```
kpodsiad, line 61: That's one of the reasons why I'm always complaining about `_ = ...` in for comprehensions :)

Issue #2183, opened 2020-11-02
Avoid continuous compilation of generated sources when using sbt BSP


ckipp01:
**Describe the bug**

If you're using sbt BSP and you have a plugin `sbt-jmh` for example that generates sources, this can potentially cause continuous compilations happening. This is a follow-up of the original issue created here in [sbt](https://github.com/sbt/sbt/issues/6037). The core of the issues pointed out by @adpi2 seems to be the following:

1. Metals sends a compileRequest on bench-jmh
2. sbt replies that bench-jmh is compiled successfully
3. Metals sends a scalaMainClasses request on bench-jmh
4. sbt runs discoveredMainClasses that generates new sources
5. sbt replies that bench-jmh contains no Main classes
6. Metals detects the new generated sources and goes to step 1

This results in this:

![2020-10-30 21 53 28](https://user-images.githubusercontent.com/13974112/97756292-ee793880-1afa-11eb-914f-5c2051b93bbf.gif)

**Expected behavior**

I'd expect for these sources to be generated once, and then stop.

**Additional context**

The possible solution mentioned there by @tgodzik 

> The alternative as we discussed with @adpi2 is that because:

> The buildTarget/sources response tells if a file is generated or not. we could try to not run compilation on changes to those files.

**Search terms**

continuous compilation, generated sources, sbt BSP

tanishiking:
Let me delve into this, as I'm unable to use sbt + BSP in my private project because of this issue.

tanishiking:
So

- We can potentially fix this issue by https://github.com/sbt/sbt/issues/6037#issuecomment-720384726 👍 
- There's a concern about the re-compilation fired by program generation https://github.com/sbt/sbt/issues/6037#issuecomment-720426998 and I also agree with @tgodzik https://github.com/sbt/sbt/issues/6037#issuecomment-720439642 
- Looks like there's already `generated` flag in BSP https://build-server-protocol.github.io/docs/specification.html#build-target-sources-request
- TODOs
  - Checking sbt returns `generated` field or not
  - Making metals not to compile for `generated` source files.

PR #3694, opened 2022-03-10
Exclude generated source items from FileWatcher (sbt BSP)

tanishiking:
Fix https://github.com/scalameta/metals/issues/2183
Previously, when a project generate sources at compile-time, for example, if `bench-jmh` project generate source files at compile time

- 1. Metals sends a compileRequest on bench-jmh
- 2. sbt replies that bench-jmh is compiled successfully
- 3. Metals sends a scalaMainClasses request on bench-jmh
- 4. sbt runs discoveredMainClasses that generates new sources
- 5. sbt replies that bench-jmh contains no Main classes
- 6. Metals detects the new generated sources and goes to step 1

This PR prevents metals from going to step 6, by stopping watching the file events fired from generated sources.

- We can tell the file is generated or not via BSP request `buildTarget/sources` which provides us whether the file/directory is generated or not. https://build-server-protocol.github.io/docs/specification.html#build-target-sources-request
- And sbt BSP has an implementation of `generated` field
https://github.com/sbt/sbt/blob/fedf6af60d86ed39ae9c94161f13a79ffc4494e8/main/src/main/scala/sbt/internal/server/BuildServerProtocol.scala#L125-L129

---

This PR doesn't contain any unit/integration tests because I was not sure where and what to test... any advice?

---

Not related to the original issue, but I witness some wired behavior with sbt BSP

- Metals sends a compile request every time I open a new file (with sbt BSP)
- sbt compiles all projects on save (for example in metals, when I modify a file in `mtags` project, compilation against `intefaces` will be fired) 🤔

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -52,8 +52,8 @@ final class BuildTargets(
     TrieMap.empty[BuildTargetIdentifier, util.Set[AbsolutePath]]
   private val inverseDependencySources =
     TrieMap.empty[AbsolutePath, Set[BuildTargetIdentifier]]
-  private val buildTargetGeneratedDirs: mutable.Set[AbsolutePath] =
-    mutable.Set.empty
+  private val buildTargetGeneratedDirs: TrieMap[AbsolutePath, Unit] =
+    TrieMap.empty[AbsolutePath, Unit]
   private val sourceJarNameToJarFile = TrieMap.empty[String, AbsolutePath]
   private val isSourceRoot =
     ConcurrentHashSet.empty[AbsolutePath]
@@ -206,7 +206,7 @@ final class BuildTargets(
       sourceItem.getKind() == SourceItemKind.DIRECTORY &&
       sourceItem.getGenerated()
     ) {
-      buildTargetGeneratedDirs += sourceItemPath
+      buildTargetGeneratedDirs(sourceItemPath) = ()
     }
     addSourceItem(sourceItemPath, buildTarget)
   }
@@ -276,7 +276,7 @@ final class BuildTargets(
   }
 
   def checkIfGeneratedSource(source: Path): Boolean = {
-    buildTargetGeneratedDirs.exists(generatedDir =>
+    buildTargetGeneratedDirs.keys.exists(generatedDir =>
       source.startsWith(generatedDir.toNIO)
     )
   }
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -1,7 +1,10 @@
 package scala.meta.internal.metals
 
 import java.lang.{Iterable => JIterable}
+import java.net.URI
 import java.net.URLClassLoader
+import java.nio.file.Path
+import java.nio.file.Paths
 import java.util
 import java.util.concurrent.ConcurrentLinkedQueue
 import java.{util => ju}
@@ -22,6 +25,8 @@ import ch.epfl.scala.bsp4j.BuildTarget
 import ch.epfl.scala.bsp4j.BuildTargetIdentifier
 import ch.epfl.scala.bsp4j.JavacOptionsResult
 import ch.epfl.scala.bsp4j.ScalacOptionsResult
+import ch.epfl.scala.bsp4j.SourceItemKind
+import ch.epfl.scala.bsp4j.SourcesResult
 import ch.epfl.scala.bsp4j.WorkspaceBuildTargetsResult
 
 /**
@@ -49,6 +54,7 @@ final class BuildTargets(
     TrieMap.empty[BuildTargetIdentifier, util.Set[AbsolutePath]]
   private val inverseDependencySources =
     TrieMap.empty[AbsolutePath, Set[BuildTargetIdentifier]]
+  private var buildTargetGeneratedDirs: List[AbsolutePath] = Nil
   private val sourceJarNameToJarFile = TrieMap.empty[String, AbsolutePath]
   private val isSourceRoot =
     ConcurrentHashSet.empty[AbsolutePath]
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -255,6 +261,32 @@ final class BuildTargets(
...
+            case item
+                if (item.getKind() == SourceItemKind.DIRECTORY && item
+                  .getGenerated()) =>
+              AbsolutePath(Paths.get(new URI(item.getUri())))
+          }
+          .toList
+      }
+      .toList
+    buildTargetGeneratedDirs = generatedDirs
+  }
```
dos65, line 282: There is `addSourceItem` and `sourceItemToBuildTarget` in `BuildTargets` that already stores the info about sources.
It would be better to extend it's data  and adapt other methods that uses this field:
```scala

case class SourceItemInfo(buildTargets: List[BuildTarget], isGenerated: Boolean)

private val sourceItemsToBuildTarget = // need to rename
  TrieMap.empty[AbsolutePath, SourceItemInfo]
```

dos65 (commented):
@tanishiking thx for looking at it! I think it's done in the right direction. I have one suggestion.

Arthurm1:
Might be nice to see these dirs listed in a separate section in [BuildTargetInfo](https://github.com/scalameta/metals/blob/864f16a5dbc3bf24f7c94e7d0f5c6f42d0aeba3f/metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala#L105)

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
tanishiking: - :+1: to register `SourceItem` to `BuildTargets` via `addSourceItem` instead of `addWorkspaceBuildSources`
  - I didn't realize there's a method to add `sourceItems`, nice catch!
- Yeah, it would be nice to store `isGenerated` in `SourceItemInfo` :+1:
  - because as @Arthurm1 mentioned, it's nice to see the `isGenerated` information in `SourceDirectories` section of `BuildTargetInfo` (of FileDecodeProvider).
  - https://github.com/scalameta/metals/blob/6c7d33f05eae64dd36f8c161dc2115f3a4c23c6f/metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala#L221-L227
- However, I don't think we should **lookup** `isGenerated` information in that way because:
  - We should be able to prefix match against generated source dirs (and `TrieMap[AbsolutePath, SourceItemInfo]` can't) because, at the point of `import build` time, we might not have generated **files** yet.
    - Therefore, we register the list of directories where generated files will be stored, and exclude files that belong to those directories. (by prefix match)
  - Also, I would like to minimize the size of `buildTargetGeneratedDirs` as the prefix match algorithm is now `O(N)`, and it would be better to reduce the N.
    - Though we can optimize the prefix match by Trie or aho-corasick or something, I guess it's still worth reducing the size of the data structure.

tanishiking:
@Arthurm1 Can I work on this, in a separate issue? I'll open a ticket :)
> Might be nice to see these dirs listed in a separate section in [BuildTargetInfo](https://github.com/scalameta/metals/blob/864f16a5dbc3bf24f7c94e7d0f5c6f42d0aeba3f/metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala#L105)

tanishiking:
Oops, I'll fix the failing test

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -49,6 +52,8 @@ final class BuildTargets(
     TrieMap.empty[BuildTargetIdentifier, util.Set[AbsolutePath]]
   private val inverseDependencySources =
     TrieMap.empty[AbsolutePath, Set[BuildTargetIdentifier]]
+  private val buildTargetGeneratedDirs: mutable.Set[AbsolutePath] =
+    mutable.Set.empty
```
dos65, line 56: `mutable.Set` isn't thread-safe.
I don't if there is a thread-safe set collection for scala but it might be replaced at least by `TrieMap.empty[AbsolutePath, ()]`

dos65 (commented):
@tanishiking I got your point to my comment. I agree that it it fine to not merge it with existing structure.

I have only one comment

tanishiking:
I'm trying to test FileWatcher doesn't fire an event on `FileWrites.write` under `src_managed` directory with `SbtServerInitializer`, and I'm wondering how can we verify file write **didn't** generate a file event? 🤔 

I see some tests that file write to `.scala` generate an event, and `.c` file doesn't generate an event, but I'm not sure how these tests verify there was no file event. Do you have any ideas? @tgodzik @Arthurm1 ? (who I can see from git history on `FileWatcherLspSuite`)
https://github.com/olafurpg/metals/blob/924694f98d2c9774a436d4382b845185c2094293/tests/unit/src/test/scala/tests/FileWatcherLspSuite.scala#L73-L96

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
tanishiking: Maybe we should use `ConcurrentLinkedQueue` or something?
like: https://github.com/scalameta/metals/blob/924694f98d2c9774a436d4382b845185c2094293/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala#L42

tgodzik:
> I'm trying to test FileWatcher doesn't fire an event on `FileWrites.write` under `src_managed` directory with `SbtServerInitializer`, and I'm wondering how can we verify file write **didn't** generate a file event? thinking
> 
> I see some tests that file write to `.scala` generate an event, and `.c` file doesn't generate an event, but I'm not sure how these tests verify there was no file event. Do you have any ideas? @tgodzik @Arthurm1 ? (who I can see from git history on `FileWatcherLspSuite`) https://github.com/olafurpg/metals/blob/924694f98d2c9774a436d4382b845185c2094293/tests/unit/src/test/scala/tests/FileWatcherLspSuite.scala#L73-L96

It looks like those tests might not work at all :thinking: I don't see a great way to test it aside from invoking a special method in test server. Maybe we don't need to test it actually? there is not much logic that operates on changed events

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
tanishiking: fixed https://github.com/scalameta/metals/pull/3694/commits/a76f120e762fd4bab4a82ec861eb34f9af7502d4

tanishiking:
Thank you @tgodzik !
Ok, I don't test anything in this PR for now. (maybe, at least, we can test we receive `generated` information from sbt BSP in https://github.com/scalameta/metals/issues/3718 (because it's easy to test `FileDecoderProvider`:+1: )

dos65 (approved):
LGTM!

PR #3699, opened 2022-03-10
Fix issues with Bloop snapshot versions

tgodzik:


kpodsiad (approved):
I wonder if using https://github.com/vdurmont/semver4j wouldn't be simpler :P

mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala
```diff
@@ -31,7 +31,7 @@ object SemVer {
           .getOrElse(List(2, 0))
 
       List(major, minor, patch) ++ rcMilestonePart ++
-        List(nightlyDate.getOrElse(Int.MaxValue))
+        List(nightlyDate.getOrElse(0))
```
dos65, line 34: For scala versions it should be `Int.MaxValue`.
There is a difference with bloop versions: 
- for scala nightly `3.0.0` is greater than `3.0.0-bin-20220310-NIGHTLY`
- for bloop  `1.4.13` is less than `1.4.13-$n-$sha-20220310`

To make it work correctly we either need to introduce a different version impl for bloop or change bloop versioning scheme.

tgodzik: Maybe we should just detect if it ends with nightly?

dos65: @tgodzik maybe the right option would be to revert SemVer.Version back and introduce a specific ScalaVersion?
I just realized that scala-nightly version doesn't follow semver.

dos65: > Maybe we should just detect if it ends with nightly?

The initial motivation of this change was to able to compare NIGHTLY version between each other too.

tgodzik: It should work correctly now, I added a check for `NIGHTLY`, which we recongnize as being only the scala nightly

dos65 (approved):
Looks good! Thx for the fix!

PR #3702, opened 2022-03-11
fix: scala3 completion - fix isses with code indented using tabs

### LINKED ISSUE #3698: Autocomplete not working properly for tab-indented Scala 3 code
**Describe the bug**

Intellisense/autocomplete does not work as expected in Scala 3 projects on the lines that are indented with tabs. 

**To Reproduce** Steps to reproduce the behavior:

1. Open a Scala 3 project
2. Indent a line of code with one or more tabs
3. Anywhere within the line, add a dot after an identifier, and press Ctrl+Space
4. "No suggestions" pops up instead of the expected list.

**Expected behavior**

Expected to work like for space-indented lines.

**Interesting workaround**

Adding a number of spaces after the dot, equal to the number of indentation tabs, helps (see the second screenshot below).

**Screenshots**

Not working:

scala_vscode_autocomp_not_working

Working:

scala_vscode_autocomp_working

**Installation:**

- Operating system: Linux
- VSCode version: 1.65.1
- VSCode extension version: 1.13.0
- Metals version: 0.11.2

**Additional context**

There are those who prefer indenting their code with tabs (at the beginning of the lines), aligning with spaces... Please, support us as well!

**Search terms**

VSCode, Metals, Scala 3, tabs, indentation, intellisense, autocomplete, not working

dos65:
Fixes #3698

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionPos.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionPos.scala
@@ -25,12 +25,13 @@ case class CompletionPos(
 
   def toEditRange: l.Range =
     def toPos(offset: Int): l.Position =
-      new l.Position(
-        cursorPos.source.offsetToLine(offset),
-        cursorPos.source.column(offset)
-      )
+      val lineStartOffest = cursorPos.source.startOfLine(offset)
+      val line = cursorPos.source.offsetToLine(lineStartOffest)
+      val column = offset - lineStartOffest
+      new l.Position(line, column)
 
     new l.Range(toPos(start), toPos(end))
+  end toEditRange
 end CompletionPos
 
 object CompletionPos:
--- a/tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
@@ -1456,4 +1456,21 @@ class CompletionSuite extends BaseCompletionSuite {
     topLines = Some(3)
   )
 
+  checkEdit(
+    "tab-indented",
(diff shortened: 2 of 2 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionPos.scala
```diff
@@ -25,12 +25,12 @@ case class CompletionPos(
 
   def toEditRange: l.Range =
     def toPos(offset: Int): l.Position =
-      new l.Position(
-        cursorPos.source.offsetToLine(offset),
-        cursorPos.source.column(offset)
-      )
+      val line = cursorPos.source.offsetToLine(offset)
+      val column = offset - line
+      new l.Position(line, column)
```
kpodsiad, line 30: Could you explain what was wrong? I cannot comprehend this by myself :D

tgodzik: This looks a bit weird. Why would be subtracting line number from file offset :thinking:

tgodzik (approved):
It looks good now! Thanks!

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionPos.scala
dos65: oops. updated

@kpodsiad  the actual issue is that dotty's `SourceFile.column` method treats tabs incorrectly. If line starts with tabs they just don't count as symbols that results to having a wrong editRange in completionItems.

I will look how to fix that in dotty but we anyway need to fix in Metals for existing scala3 versions.

PR #3703, opened 2022-03-11
Warn users properly when wrong release flag was used

tgodzik:
Reported by @romanowski

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
+++ b/metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
@@ -7,6 +7,7 @@ import scala.meta.internal.metals.BloopServers
 import scala.meta.internal.metals.BuildInfo
 import scala.meta.internal.metals.JavaTarget
 import scala.meta.internal.metals.JdkSources
+import scala.meta.internal.metals.JdkVersion
 import scala.meta.internal.metals.Messages
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.MtagsResolver
@@ -36,8 +37,11 @@ class ProblemResolver(
     }
   }
 
-  def recommendation(java: JavaTarget): Option[String] = {
-    findProblem(java).map(_.message)
+  def recommendation(
+      java: JavaTarget,
+      scalaTarget: Option[ScalaTarget]
+  ): Option[String] = {
+    findProblem(java, scalaTarget).map(_.message)
   }
 
   def recommendation(scala: ScalaTarget): Option[String] = {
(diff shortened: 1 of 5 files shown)


metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
```diff
@@ -252,6 +261,32 @@ class ProblemResolver(
...
+
+  private def isWrongJavaRelease(
+      javaTarget: JavaTarget,
+      scalaTarget: Option[ScalaTarget]
+  ): Option[JavaProblem] = {
+    def buildJavaVersion =
+      for {
+        target <- scalaTarget
+        javaHome <- target.jvmHome
+        path = javaHome.toAbsolutePath
```
kpodsiad, line 275: I would inline path

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
```diff
@@ -252,6 +261,32 @@ class ProblemResolver(
...
+        version <- JdkVersion.getJavaVersionFromJavaHome(path)
+      } yield version
+
+    val releaseVersion = javaTarget.releaseVersion.flatMap(JdkVersion.parse)
+    releaseVersion.zip(buildJavaVersion) match {
+      case Some((releaseVersion, jvmHomeVersion))
+          if jvmHomeVersion.major < releaseVersion.major || jvmHomeVersion.minor < releaseVersion.minor =>
```
kpodsiad, line 282: I'm not convicted about `|| jvmHomeVersion.minor < releaseVersion.minor` Shouldn't minor version be compatible?

kpodsiad (approved):
Looks good. However, I have one question

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
tgodzik: minor in this case is if we get `1.8` instead of `8`. This might not be needed at all really, usually we only set major.

But maybe I should switch the condition a bit.

dos65: @tgodzik I think we can align versions to a single format in `JdkVersion`.

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
tgodzik: Done!

metals/src/main/scala/scala/meta/internal/troubleshoot/ProblemResolver.scala
tgodzik: I switched to just using major. We don't actually need the minor, so it should be ok now.

dos65: @tgodzik thx!

Issue #3625, opened 2022-02-10
Autocomplete at end of line with whitespace


adampauls:
### Describe the bug

With this prefix:

```scala
object Test:
  case class Widget(name: String, other: Int)
  val otxxx: Int = 1
  Widget(name = "foo", 
```

autocomplete will (correctly) suggest both `otxxx` and `other` if either (a) the cursor is immediately after the comma, or (b) there is another line (like `val x = 5`) after the Widget line. 

I think this might be a bug in the underlying Scala parser, but I am not sure. 

### Expected behavior

Autocomplete should behave the same regardless of whitespace in this case. 

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v1.12.0

tgodzik:
Thanks for reporting! Suprisingly, if you start writing `o` it will also suggest the correct completions. I think this is one of the cases that we might need to add the artificial `CURSOR` name so that the parser doesn't fail. It's something we haven't yet ported from Scala 2.

adampauls:
Hmm, actually, now I'm confused. Look at what I believe are [dotty's completion unit tests](https://github.com/lampepfl/dotty/blob/main/language-server/test/dotty/tools/languageserver/CompletionTest.scala), there are no unit tests for filling in arguments to an `Apply`, i.e. nothing like `foo(${m1}`. Does the functionality for filling in argument names and values by type live only in Metals? If so, can I have a good pointer? If not, do you know where it dotty it is tested?

adampauls:
Ah, I found [this](https://github.com/scalameta/metals/blob/1ad113dace0375e8d7f1748d613fdae33862201e/mtags/src/main/scala-3/scala/meta/internal/pc/Completions.scala), so it seems that the Scala compiler only provides some of the completions functionality.

tgodzik:
We do a number of custom completions that are built upon the information from the Dotty presentation compiler. We try to port some of the improvements to the compiler itself, but since that can be used by a number of different tools we need to be careful what we include upstream.

PR #3704, opened 2022-03-11
fix: add test case for #3625

tgodzik:
The issue has been fixed in the compiler.

Fixes https://github.com/scalameta/metals/issues/3625


tanishiking (approved):
LGTM

Issue #3381, opened 2021-12-17
Incorrect (but correct in some sense) completions in expression evaluation


ckipp01:
**Describe the bug**

I very randomly hit on this today and thought it was a fun little issue. I can see exactly where the issue is and I have a fix for it, but _cannot_ get the tests to pass and don't get why.

The issue is when you are trying to use the expression evaluation completions. Let's say you have the following code:

```scala
val x = 3

println(x) <-- set your breakpoint here
```

Then when you start your expression evaluation trigger a completion with `.`.  You'll then see the completions for Int since it thinks you're trying to complete a member of `3`.

The issue is right here https://github.com/scalameta/metals/blob/3fbe9bd873ebf329a424c54ad005849e4a887e30/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala#L285-L287

Since we try to insert the expression at the breakpoint and shift everything down, technically `.toD` would correctly complete with `toDouble` since it thinks we are completing on `3`. I thought I could do the following to fix it:

```diff
    val modified =
+      s"${oldText.substring(0, lineStart)};${expression
-       s"${oldText.substring(0, lineStart)}${expression
          .getText()}\n${" " * indentation}${oldText.substring(lineStart)}"
```

Forcing the termination of whatever came before, and then starting with a fresh expression.  So this would give you:

```scala
val x = 3

;. // <-- because of the ; no completions on 3 given
println(x)
```

Is there something I'm missing with why this seems to work fine locally but blows up in the tests and it returns a whole bunch of irrelevant completions.

**Expected behavior**

I'd expect no completions on `.` in the above situation.

**Screenshots**

<img width="418" alt="Screenshot 2021-12-17 at 16 06 06" src="https://user-images.githubusercontent.com/13974112/146565000-ebc075ea-7d77-4af4-a43f-cd69e3cfe172.png">

**Installation:**

- Operating system: macOS
- Editor: `nvim-metals`
- Metals version: `0.10.9+176-3fbe9bd8-SNAPSHOT`

**Additional context**

This probably isn't a _huge_ issue, but could potentially confuse someone. I only noticed because in the dap repl for `nvim-dap` you start some of the built in things with a `.` like `.scopes` which provides you all the scopes etc.

tgodzik:
An alternative would be to do `stripPrefix(".")` maybe? Though adding `;` should work :thinking:

PR #3705, opened 2022-03-11
Fix completions on dot in expression evaluator

tgodzik:
Previously, we would show completions for the previous statement. Now, we block it by adding `;`

Fixes https://github.com/scalameta/metals/issues/3381


ckipp01 (approved):
Simple fixes are the best fixes. 👍🏼 LGTM

PR #3707, opened 2022-03-11
refactor: use "using" instead of "implicit" in Scala 3 signature help.

ckipp01:
This just aligns the usage of "using" since currently if you do a hover
on a method you'll see the implicit parameter using the "using" keyword
but then if you do a signature help on the same method you'll see
"implicit". This just ensure both are "using".


tanishiking (approved):
Nice catch! :+1:

PR #3708, opened 2022-03-12
Move gabro to "past maintainers" in README

gabro:
It's been a while since I've actively contributed to the project, let alone some reviews on the VS Code extension here and there.

I've also not been working with Scala at work for a while now, and I probably won't in the near future (I've been having lots of fun with design systems and UI architectures, in case anyone's curious 😅), so I haven't been able to justify much time spent on Scala OSS projects.

That said, nothing changes in practice (I'll be around Discord as always), but I think it's only fair to clarify my Metals maintainer status 🙂

Do absolutely feel free to keep asking for my review for the VS Code extension PRs if you feel it's valuable, I'll keep up with those as best as I can!


ckipp01 (approved):
I think I speak for many when I say thanks for all your hard work, insight, and reviews over the years. ❤️

PR #3709, opened 2022-03-14
support for creating companion objects

### LINKED ISSUE #3736: code action - generate companion object
### Is your feature request related to a problem? Please describe.

When developing in vscode, I'd like to generate a companion object snippet.

### Describe the solution you'd like

Add a "generate companion object" code action. See comparable feature in intelliJ: 
image
Result:
image

### Describe alternatives you've considered

Haven't found anything comparable in vscode.

### Additional contex

_No response_

### Search terms

companion object snippet generate

zmerr:
resolves #3736  ! I need help with code insertion as well as code navigation to the location of the companion object!
demo:

https://user-images.githubusercontent.com/93259201/158370343-c285624c-a3d9-41b8-b523-da90e3414938.mov


kpodsiad:
Hey @zmerr! Thanks for your contribution.
If you are opening a PR that is not intended to be reviewed immediately you could open it as a draft or add `[WIP]` prefix to the title.

First of all, you can create tests for these new features, in order to do that you can take a look at [already existing](https://github.com/scalameta/metals/tree/main/tests/unit/src/test/scala/tests/codeactions) code actions suites. Thanks to tests, you'll be able to test your changes very quickly and the feedback loop will shorten significantly. You will be able to run your tests via sbt's `unit/testOnly tests.codeactions.<suite_name>`

I think you could use `go to position` command https://github.com/scalameta/metals/blob/536ff4b13a4ac8d5140913b42cefbb34a9e88501/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala#L319 in order to implement code navigation.

What are your problems with code insertion?

zmerr:
Hey @kpodsiad Thank you very much for the tips!
> What are your problems with code insertion?
I think my implementation overwrites the contents of the file at the location it creates the companion object! I want to insert instead, without writing over the existing content, erasing them! 

I have seen this sample in the code, but that is through `ServerCommands`

```
codeAction.setCommand(
  ServerCommands.InsertInferredType.toLSP(
    new l.TextDocumentPositionParams(
      params.getTextDocument(),
      range
    )
  )
)
```

Any help would be appreciated!

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
```diff
@@ -0,0 +1,134 @@
...
+      ec: ExecutionContext
+  ): Future[Seq[l.CodeAction]] = Future {
+    pprint.log("contribute")
+    val uri = params.getTextDocument().getUri()
+    val path = uri.toAbsolutePath
+    val range = params.getRange()
+    val applyTree =
+      if (range.getStart == range.getEnd)
+        trees
+          .findLastEnclosingAt[Defn.Class](
```
tgodzik, line 36: We might want to do the same for trait also, in that case we might look for `Template` and check the parent.

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
```diff
@@ -0,0 +1,134 @@
...
+      path: AbsolutePath,
+      classDefinition: Defn.Class,
+      uri: String
+  ): l.CodeAction = {
+    val codeAction = new l.CodeAction()
+    codeAction.setTitle(CreateCompanionObjectCodeAction.companionObjectInfo)
+    codeAction.setKind(this.kind)
+    val rangeStart = classDefinition.pos.toLSP.getEnd
+    val rangeEnd = classDefinition.pos.toLSP.getEnd
+    rangeEnd.setLine(rangeEnd.getLine + 5)
```
tgodzik, line 66: This might break the code if there are other classes afterwards. Instead of adding the line we can prepend `companionObjectFirstLine` with newlines

tests/unit/src/test/scala/tests/codeactions/CompanionObjectSuite.scala
```diff
@@ -0,0 +1,32 @@
+package tests.codeactions
+
+import scala.meta.internal.metals.codeactions.CreateCompanionObjectCodeAction
+
+class CompanionObjectSuite extends BaseCodeActionLspSuite("companionObject") {
+
+  check(
+    "insert-companion-object",
```
tgodzik, line 8: I would also test some inner classes like:
```
class A {
  class B {
   <cursor here>
  }
}
```

tgodzik (commented):
Looks good! I left some initial comments, but I think this is on a right track!

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
zmerr: Yes! I would add that! Thanks for the tip!

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
zmerr: Sorry, I am not sure if I understand this! How do I specify the range end then? How do I prepend new lines?

tests/unit/src/test/scala/tests/codeactions/CompanionObjectSuite.scala
zmerr: Sure! I would add a very comprehensive test for you!

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
zmerr: Great! Let me try!

zmerr:
What is missing in this PR is a permanent way of testing where the cursor is in the output! Because the cursor position should get navigated to the beginning of the existing or newly created companion object!

ckipp01:
Excited for this! Just a drive by comment, can you make sure to squash your commits down to get rid of all the "draft" and "fixed" commits just to make the final commit/s a bit cleaner.

tests/unit/src/test/scala/tests/codeactions/CompanionObjectSuite.scala
```diff
@@ -0,0 +1,478 @@
...
+       |  object Baz{
+       |
+       |  }
+       |
+       |
+       |""".stripMargin,
+    selectedActionIndex = 1
+  )
+
+  check(
```
zmerr, line 444: @ckipp01 yes, sure I would squash things! But there is a huuuuuge problem with the current state of the code: it doesn’t seem to parse the braceless syntax of the test! Can you please take a look?

ckipp01: Honestly my personal opinions is to just get the feature working and not worry about braceless for now. Most of the time scalafmt can just take care of this, and it seems like the vast majority of the complexity in the pr is around braceless syntax. Don't be afraid to push the feature without support for braceless, and then we can iterate and try to improve or add it later. It's up to you though.

zmerr: @ckipp01  thank you very much for the input! 
But the biggest problem with the braceless is that it does not get parsed in the first place to get to the point of the execution of the code action! 
I think this issue demands some more attention!

zmerr: Even copying and pasting from [Scastie](https://scastie.scala-lang.org/w7MswamuT4eMEBFSq2XkLA) into the test, makes the whole tree to be `None` still! nothing gets parsed!

zmerr: Ok, now actually in the live demo, the braceless syntax works perfectly with the current code! I wonder what is wrong with the test case! I think I should just remove the test for braceless, then???
@tgodzik

zmerr: I think most probably the test method is messing up with white space, behind the scenes, somewhere🧐

zmerr: huh, got it! The whole problem is coming from here in the `check` method of `BaseCodeActionLspSuite`
 initialize(s"""/metals.json
                           |{"a":{$scalacOptionsJson "scalaVersion" : "$scalaVersion"}}
                           |$scalafixConf
                           |/$path
                           |$fileContent"""
          .stripMargin
        )

when it strips the margin, it becomes wellformed json with the braceless syntax somehow; but when it does not, due to the `:` notation of the bracelsss syntax, it somehow gets mistaken as malformed json!

tests/unit/src/test/scala/tests/codeactions/CompanionObjectSuite.scala
```diff
@@ -0,0 +1,471 @@
...
+      |  object Baz{
+      |
+      |  }
+      |
+      |}
+      |""".stripMargin
+  )
+
+  check(
+    "don't create existing companion object of trait inside a parent package",
```
tgodzik, line 407: I wonder if we could reduce a bit the number of tests? For example we already test to not create a companion object if there is already one for classes, so it seems we are testing the same branches a number of times. 

I would test:
- create companion object for class
- create companion object for trait
- don't create companion object for class if it exists
- create companion object for a trait inside another object
- test creating companion object with Scala 3 and optional braces (this one needs to be done in Scala3CodeActionLspSuite since unit tests are able to only test the 2.13.8 version)

The reason here is that we always initialize the entire workspace in these tests, which makes our CI run a long while. If you want you could add a test suite that just tests the specific Code action (this is more akin to actual unit tests). I wanted to reduce our CI times for reworking the RenameSuite to do a similar thing.

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
```diff
@@ -0,0 +1,212 @@
...
+      name: String
+  ): Option[Defn.Object] =
+    tree.parent.flatMap(_.children.collectFirst {
+      case potentialCompanionObject: Defn.Object
+          if (potentialCompanionObject.name.value == name) =>
+        potentialCompanionObject
+    })
+
+  private def applyWithSingleFunction: Tree => Boolean = {
+    case _: Defn.Class =>
```
tgodzik, line 198:
```suggestion
    case _: Defn.Class | _: Defn.Trait | _: Defn.Enum => true
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
```diff
@@ -0,0 +1,212 @@
...
+      treePos: Position,
+      document: String
+  ) = {
+    val scanner = new Scanner(document)
+
+    (0 to treePos.startLine)
+      .map(_ => scanner.nextLine())
+      .last
+      .takeWhile(_.isWhitespace)
+  }
```
tgodzik, line 84:
```suggestion
  private def getIndentationForPositionInDocument(
      treePos: Position,
      document: String
  ) = {
    document.substring(treePos.start - treePos.endColumn, treePos.start).takeWhile(_.isWhitespace)
  }
```

this way we don't have iterate over the entire document.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -631,6 +647,42 @@ object MetalsEnrichments
 
     def lineAtIndex(index: Int): Int =
       indexToLspPosition(index).getLine
+//
+//    /**
+//     * @param pos the position at which non-blank text starts
+//     * @return the length of indentation at the beginning of this string
+//     */
+//    def indentationLength(pos: Position): Int = {
```
tgodzik, line 655: Why won't we use this one instead of the Scanner actually? It's similar to what I suggested beneath.

tgodzik (changes requested):
Great work! I have some comments about the PR, do let me know if you have any questions!

tests/unit/src/test/scala/tests/codeactions/CompanionObjectSuite.scala
```diff
@@ -0,0 +1,471 @@
...
+       |
+       |object Foo {
+       |
+       |}
+       |""".stripMargin,
+    selectedActionIndex = 1
+  )
+
+  check(
+    "don't create existing companion object inside a parent object",
```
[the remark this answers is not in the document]
zmerr: great!

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
```diff
@@ -0,0 +1,155 @@
...
+
+  private def applyWithSingleFunction: Defn.Class => Boolean = {
+    // exclude case when body has more than one line (is a Block) because it cannot be rewritten to parens
+    case _ => true
+  }
+}
+
+object CreateCompanionObjectCodeAction {
+  val companionObjectCreation = "Create class companion object"
+  val companionObjectInfo = "Show class companion object"
```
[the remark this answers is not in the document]
zmerr: Yes! Navigation can help in quickly finding an existing companion object! Let’s keep it!

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
```diff
@@ -0,0 +1,212 @@
...
+  private def hasBraces(tree: Tree, document: String): Boolean = {
+    tree.children
+      .collectFirst { case template: Template =>
+        document(template.pos.start) == '{'
+      }
+      .getOrElse(false)
+    //  document(tree.pos.end-1) == '}'
+  }
+
+  private def getName(tree: Tree): String = {
```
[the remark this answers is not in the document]
zmerr: This is a private function and in this context, Tree can only be of type enum, class or trait. Therefore, it always had a name! So keeping it non optional would make the code cleaner!

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
zmerr: great!

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
zmerr: Great! Thank you!

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
zmerr: That code is buggy! Kamil made a bug free version! But it is also much longer and more difficult to maintain!

tests/unit/src/test/scala/tests/codeactions/CompanionObjectSuite.scala
zmerr: according to @tgodzik 

> test creating companion object with Scala 3 and optional braces (this one needs to be done in Scala3CodeActionLspSuite since unit tests are able to only test the 2.13.8 version)

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
```diff
@@ -0,0 +1,187 @@
...
+        trees
+          .findLastEnclosingAt[Tree](
+            path,
+            range.getStart(),
+            applyWithSingleFunction
+          )
+      else
+        None
+
+    applyTree.flatMap { tree =>
```
tgodzik, line 52: If we used Option for `getNameOfClassTraitOrEnumTree` we can rewrite this all to:
```scala
val companionObjectAction = for {
  tree <- applyTree
  name <- getNameOfClassTraitOrEnumTree(tree)
  if !hasCompanionObject(tree, name)
  document <- buffers.get(path).getOrElse("")
} yield buildCreatingCompanionObjectCodeAction(
            path,
            tree,
            uri,
            indentationString,
            name,
            hasBraces(tree, document)
          )
companionObjectAction.toSeq
```

We often use for comprehensions in the codebase as it makes it a bit clearer when we have Options

tgodzik (changes requested):
Great work! Some final comments, mostly about removing comments and one suggestion for the style

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
```diff
@@ -0,0 +1,181 @@
...
+      document: String
+  ): String =
+    document
+      .substring(treePos.start - treePos.endColumn + 1, treePos.start)
+      .takeWhile(_.isWhitespace)
+
+  private def hasBraces(tree: Tree, document: String): Boolean = {
+    tree.children
+      .collectFirst { case template: Template =>
+        document(template.pos.start) == '{'
```
tgodzik, line 81: Actually, we can also have `case class Location(city: String)` and it doesn't have braces, but we can't create a companion object `object Location:` if it's not Scala 3.

We might want to use `canUseBracelessSyntax` method from `ExtractValueCodeAction` so that we are sure we adhere to the overall style of the file.

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
```diff
@@ -0,0 +1,181 @@
...
+    maybeCompanionObject.toSeq
+
+  }
+
+  private def getIndentationForPositionInDocument(
+      treePos: Position,
+      document: String
+  ): String =
+    document
+      .substring(treePos.start - treePos.endColumn + 1, treePos.start)
```
tgodzik, line 75: This seems to work for me:
```suggestion
      .substring(treePos.start - treePos.startColumn, treePos.start)
```

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
zmerr: On progress...

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -111,8 +115,8 @@ object MetalsEnrichments
   }
 
   implicit class XtensionCompileResult(result: b.CompileResult) {
-    def isOK: Boolean = result.getStatusCode.isOK
-    def isError: Boolean = result.getStatusCode.isError
+    def isOK: Boolean = result.getStatusCode == b.StatusCode.OK
```
tgodzik, line 118: I think this might have changed due to resolution of merge conflicts? Could you revert it?

tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
```diff
@@ -131,10 +132,9 @@ class Scala3CodeActionLspSuite
     """|object Main {
        |  def method2(i: Int) = ???
        |  def main = {
-       |    def inner(i : Int) = {
```
tgodzik, line 134: I think this one should stay the same since main has braces, we also want inner to have braces.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
```diff
@@ -976,6 +1005,49 @@ object MetalsEnrichments
...
+    def canUseBracelessSyntax(source: String): Boolean = {
+
+      def allowBracelessSyntax(tree: Tree) = tree.origin match {
+        case p: Parsed => p.dialect.allowSignificantIndentation
+        case _ => false
+      }
+
+      def isNotInBraces(t: Tree): Boolean = {
+        t match {
+          case _: Template | _: Term.Block => source(t.pos.start - 1) != '{'
```
tgodzik, line 1023: This should be:
```suggestion
          case _: Template | _: Term.Block if  t.pos.start < source.length => source(t.pos.start) != '{'
```
that's why the tests changed.

tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
zmerr: hmm… so by changing the code the way you suggested it above, the test would work?
I would try...

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
zmerr: @tgodzik I rebased onto main for resolving the conflict and I guess I picked the wrong resolution! I should have picked the main one…
I didn’t pay attention and thought that the difference was just the white space!

tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
tgodzik: It should work, I compared the changes in the methods.

metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
tgodzik: No worries, that's why it's usually a good idea to take a look at the PR itself again to double check. I usually do that especially after rebasing with conflicts.

zmerr: solved! And yes, I need to take a look at the PR after rebase...

tgodzik (approved):
LGTM! Great work!

PR #3710, opened 2022-03-14
feat: add compilation status to the doctor

kpodsiad:
This PR adds `Compilation status` column to the Metals doctor

my-app-1647246849605

I have to add some tests (to test both vscode and other clients) but I can't find any doctor-related test suite.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
@@ -30,6 +30,7 @@ import ch.epfl.scala.bsp4j.BuildTargetIdentifier
 final class Doctor(
     workspace: AbsolutePath,
     buildTargets: BuildTargets,
+    diagnostics: Diagnostics,
     languageClient: MetalsLanguageClient,
     currentBuildServer: () => Option[BspSession],
     calculateNewBuildServer: () => BspResolvedResult,
@@ -256,6 +257,7 @@ final class Doctor(
         .sortBy(f => (f.baseDirectory, f.name, f.dataKind))
 
       val explanations = List(
+        DoctorExplanation.CompilationStatus.toJson(allTargetsInfo),
         DoctorExplanation.Diagnostics.toJson(allTargetsInfo),
         DoctorExplanation.Interactive.toJson(allTargetsInfo),
         DoctorExplanation.SemanticDB.toJson(allTargetsInfo),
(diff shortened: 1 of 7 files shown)


Arthurm1:
That's handy - I've often thought it would be useful to have more info on the compilation status - both "compiling" and "failed"

Is it possible to add this info to the Metals Tree View - i.e. have the project name go red on an error (yellow on warnings) - maybe even drill down to package or class level?  And have an hourglass next to the project name if it's compiling?

Also - is it possible to highlight the Metals icon on the left hand side - like when the VS Code file explorer shows a blue "1" when you have one file unsaved?  Could it have a red "3" for 3 targets failing to compile - and a blue "3" if they are in the middle of compiling so you could see the compiling status more clearly without having to flip to the Metals page.

kpodsiad:
> Is it possible to add this info to the Metals Tree View - i.e. have the project name go red on an error (yellow on warnings) - maybe even drill down to package or class level? And have an hourglass next to the project name if it's compiling?
> 
> Also - is it possible to highlight the Metals icon on the left hand side - like when the VS Code file explorer shows a blue "1" when you have one file unsaved? Could it have a red "3" for 3 targets failing to compile - and a blue "3" if they are in the middle of compiling so you could see the compiling status more clearly without having to flip to the Metals page.

Wow, these are brilliant ideas @Arthurm1! I'll check if it's possible to implement them.

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -407,10 +414,20 @@ final class Doctor(
     }
   }
 
+  private def extractCompilationStatus(
+      targetId: BuildTargetIdentifier
+  ): DoctorStatus = {
+    val errors = diagnostics.getCompilationErrors(targetId)
```
tgodzik, line 420: Did you check what happens (if any errors are shown) in case the compilation broke? Like with the release flag?

tgodzik (commented):
Looks great! I also like the proposals from @Arthurm1 :tada:

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
kpodsiad: now doctor should show `❌` for targets where compilation was aborted or ended with error status code

tgodzik: Awesome, great! Now the doctor will finally not ignore people's issues :S

tgodzik (approved):
LGTM! Could we put @Arthurm1 suggestions into an issue somewhere? Maybe VS Code?

PR #3711, opened 2022-03-14
Fallback to current file contents always

tgodzik:
Previously, it was possible that for some files we would not load the current file contents especially with the file was not opened. Now, we will always try fallback to current text.

This might have caused previously a flaky warning related to snapshot text.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MutableMd5Fingerprints.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MutableMd5Fingerprints.scala
@@ -1,6 +1,7 @@
 package scala.meta.internal.metals
 
 import java.nio.charset.Charset
+import java.nio.charset.StandardCharsets
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentLinkedQueue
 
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/MutableMd5Fingerprints.scala
```diff
@@ -14,28 +15,42 @@ final class MutableMd5Fingerprints extends Md5Fingerprints {
...
 
+    currentLookup match {
+      case None =>
+        val text = FileIO.slurp(path, StandardCharsets.UTF_8)
+        val currentMD5 = MD5.compute(text)
+        if (md5 == currentMD5) {
+          add(path, text, Some(md5))
+          Some(text)
+        } else None
+      case found => found
```
kpodsiad, line 51: What do you think about `.orElse { ... }` instead of pattern matching?

kpodsiad (approved):

metals/src/main/scala/scala/meta/internal/metals/MutableMd5Fingerprints.scala
tgodzik: Done!

dos65 (approved):
:+1:

PR #3721, opened 2022-03-15
fix: Ensure the "no run or test" message is shown to user.

ckipp01:
This just ensures that when a user triggers a `runOrTest` for debug
discovery that if nothing is found the error is actually shown to the
user, which now it just quietly fails without notifying the user.


tgodzik (approved):
LGTM

ckipp01:
Test failures all seem to be random and unrelated.

PR #3723, opened 2022-03-16
fix: ensure user gets warned of workspace error when running with lens

ckipp01:
Currently if the current workspace has an error but there is still a
code lens to run the project, we just forward the params and try to
start the debug server. This then fails since there are errors in the
workspace and doesn't really show anything to the user. This is just a
slight change that adds a check before that to see if there are
workspace errors and if so, show the warning.

#### Previous behavior

_Notice the user really has no feedback that something is wrong_.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1808,7 +1808,7 @@ class MetalsLanguageServer(
         val debugSessionParams: Future[b.DebugSessionParams] = args match {
           case Seq(debugSessionParamsParser.Jsonized(params))
               if params.getData != null =>
-            Future.successful(params)
+            debugProvider.ensureNoWorkspaceErrors(params)
 
           case Seq(mainClassParamsParser.Jsonized(params))
               if params.mainClass != null =>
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1808,7 +1808,7 @@ class MetalsLanguageServer(
         val debugSessionParams: Future[b.DebugSessionParams] = args match {
           case Seq(debugSessionParamsParser.Jsonized(params))
               if params.getData != null =>
-            Future.successful(params)
+            debugProvider.ensureNoWorkspaceErrors(params)
```
kpodsiad, line 1811: Nice, I'll have to add similar construct to the TestExplorer request.

kpodsiad (approved):
I'll check this with vscode later (because I don't see how vscode will handle this failed message) but feel free to merge it instead of waiting.

ckipp01:
> I'll check this with vscode later (because I don't see how vscode will handle this failed message) but feel free to merge it instead of waiting.

It should handle like it does in other situations where it just displays the popup. All the other cases that it can fall into here handle them this same way, but this one was the one that differed since we just returned a `Future.successful(params)` before. So it should handle if fine.

PR #3725, opened 2022-03-16
fix(doctor): ensure compilationStatus is in the json payload

ckipp01:


metals/src/main/scala/scala/meta/internal/metals/DoctorResults.scala
```diff
@@ -56,6 +56,7 @@ final case class DoctorTargetInfo(
   def toJson: Obj =
     ujson.Obj(
       "buildTarget" -> name,
+      "compilationStatus" -> compilationStatus.explanation,
```
tgodzik, line 59: Shouldn't we bump the version?

ckipp01: Ahhh good catch, I totally forgot about the version. Yes I'll bump this and update the docs.

tgodzik (approved):
LGTM!

PR #3728, opened 2022-03-16
Prefer Scala 2.13 version over older versions

tgodzik:
Related to https://github.com/scalameta/metals-feature-requests/issues/258


ckipp01 (approved):
Make sense LGTM!

PR #3738, opened 2022-03-18
Show better unapply signature help

tgodzik:
Previously, we would not show correct parameters for explicitely written unapply instead we would only show the unapply method signature. Now, we properly use unapply result type to show what types we can match on.

Additionally, we no longer show `unapply` method name and return type on unapply, since they are not needed.


tests/cross/src/test/scala/tests/pc/SignatureHelpPatternSuite.scala
```diff
@@ -91,8 +85,113 @@ class SignatureHelpPatternSuite extends BaseSignatureHelpSuite {
...
+    "pat4",
+    """
+      |object & {
+      |  def unapply[A](a: A): Some[(A, A)] = Some((a, a))
+      |}
+      |object a {
+      |  "" match {
+      |    case "" & s@@
+      |  }
+      |}
```
kpodsiad, line 186: This looks strange

mtags/src/main/scala-2/scala/meta/internal/pc/SignatureHelpProvider.scala
```diff
@@ -89,16 +89,23 @@ class SignatureHelpProvider(val compiler: MetalsGlobal) {
         case tpe => tpe
       }
     }
+    val isUnapplyMethod: Boolean = tree match {
+      case _: UnApply => true
+      // in some case we will get an Apply tree, but in that case position indicates unapply
+      case _ =>
```
kpodsiad, line 95: In what cases we will get Apply tree?

tgodzik: No idea, that's what is returned from the compiler.

tests/cross/src/test/scala/tests/pc/SignatureHelpPatternSuite.scala
tgodzik: The case was added by Olaf a while ago, I haven't yet fixed it

mtags/src/main/scala-2/scala/meta/internal/pc/SignatureHelpProvider.scala
dos65: @tgodzik in which case? I replaced this case by returning false and cross/test pass for 3.1.1

tgodzik: The test is not enabled at all for Scala 3. I don't remember exactly which of the cases, I think `generic` test case?

dos65: :facepalm: I missed that it's about scala2 :smiling_face_with_tear:

tgodzik: I got back to Scala 2 to figure out what we should expect from the functionality in Scala 3 as well.

dos65 (approved):
LGTM!

kpodsiad (approved):
👍🏻

PR #3739, opened 2022-03-18
fix: return an error when trying to decode tasty in Scala 2

ckipp01:
Since this isn't supported for now we should correctly return an error
for this. Right now you can still trigger this in Scala 2 and it just
returns an empty string not really giving the user much information
about why it didn't work. This way clients can now handle this situation
better.


tgodzik (approved):
LGTM!

PR #3743, opened 2022-03-21
fix: [Scala 3] Do not show a definition and hover on whitespaces and comments

### LINKED ISSUE #1707: Go to definition / hover revert to enclosing tree if cursor is not on a symbol
**Describe the bug**

I Scala 3 in code like:
```scala
class A {
  
  def foo() = {
    // this is a comment
  }
}
```

**To Reproduce**
1. use go to definition on the comment
2. Metals will go to foo.

**Expected behavior**
Nothing happens.

**Installation:**
- Metals version: v0.9.0

**Additional context**
This is most likely due to compiler behavior when faced with a non-existing symbol position.

**Search terms**
scala3 go to definition

tdudzik:
fixes #1707

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
@@ -121,4 +121,10 @@ object MtagsEnrichments extends CommonMtagsEnrichments:
     def backticked: String =
       KeywordWrapper.Scala3.backtickWrap(s)
 
+  extension (params: OffsetParams)
+    def isWhitespace: Boolean =
+      params.offset() < 0 ||
+        params.offset() >= params.text().length ||
+        params.text().charAt(params.offset()).isWhitespace
+
 end MtagsEnrichments
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
@@ -2,11 +2,13 @@ package scala.meta.internal.pc
 
 import java.{util as ju}
 
+import scala.jdk.javaapi.OptionConverters.toJava
 import scala.util.control.NonFatal
 
 import scala.meta.internal.mtags.MtagsEnrichments.*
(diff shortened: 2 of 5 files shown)


tanishiking:
@tdudzik Thank you for working on this issue!

Now hover on whitespace doesn't show the tooltip (for enclosing symbol) anymore, awesome :tada:
Screen Shot 2022-03-23 at 18 00 40

---

I'm concerned about this strategy (checking if the parameter offset is white space or inside comments) may not resolve the problems like the following cases.


Screen Shot 2022-03-23 at 17 47 56
Screen Shot 2022-03-23 at 17 47 13
(from https://github.com/scalameta/metals/blob/934ac115a3469f5585f284023da804f04bfa2877/mtags/src/main/scala-3/scala/meta/internal/pc/SelectionRangeProvider.scala)

Hover on the no-symbol part of the program (e.g. `then`) shows a tooltip for the enclosing symbol.

Maybe we can fix this issue by comparing the `enclosingSymbol`s position and parameter position around here? https://github.com/scalameta/metals/blob/934ac115a3469f5585f284023da804f04bfa2877/mtags/src/main/scala-3/scala/meta/internal/pc/PcDefinitionProvider.scala#L76-L80

tgodzik:
@tdudzik do you have time to finish it?

Otherwise, @tanishiking could you help out here and finish it?

tanishiking:
Hi @tdudzik 
I guess we can use [SourceTree#namePos](https://github.com/lampepfl/dotty/blob/a26a2c3eedae851249ec4ad21264f9e5bb6996c5/compiler/src/dotty/tools/dotc/interactive/SourceTree.scala#L21-L22) which returns the position of name in the `NameTree` (for example, `<def foo = { ... }>.namePos` returns the position of `foo` instead of the position of `def foo = { ... }`).
I did a bit of prototyping here https://github.com/scalameta/metals/commit/1e34f795ac73f1b0e94b536135ab6eeb9dc069b8 for `HoverProvider`, and I assume we can do the same thing for `DefinitionProvider`. Hope this prototype help you :)
If you need any help, please let me know.

tgodzik:
Work on this continues in different PRs, thanks @tdudzik for starting the work on this!

PR #3745, opened 2022-03-21
fix: create `new scala file` with a single NL

dos65:
Previously there `New Scala File` -> `Empty File` were created with two
newlines between package and cursor instead of one

### DIFF
--- a/tests/unit/src/test/scala/tests/NewFileLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/NewFileLspSuite.scala
@@ -530,7 +530,9 @@ class NewFileLspSuite extends BaseLspSuite("new-file") {
       existingFiles: String = "",
       expectedException: List[Class[_]] = Nil,
       javaMinVersion: Option[String] = None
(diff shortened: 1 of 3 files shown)


tests/unit/src/test/scala/tests/NewFileLspSuite.scala
```diff
@@ -132,8 +132,7 @@ class NewFileLspSuite extends BaseLspSuite("new-file") {
     expectedFilePath = "a/src/main/scala/foo/Foo.scala",
     expectedContent = """|package foo
                          |
-                         |final case class Foo()
-                         |""".stripMargin
+                         |final case class Foo()""".stripMargin
```
tgodzik, line 135: Why did this test change actually? Shouldn't we have a newline after the definition?

dos65: fixed it too. there were no newline

tgodzik (approved):
LGTM!

PR #3746, opened 2022-03-22
bloop jvm settings

### LINKED ISSUE #2697: Make it easier for users to change build server's memory requirements
When Metals runs it actually starts 2 processes. One is the server itself and the other is the build server (either Bloop or sbt). This is not always clear to the users and they might be surprised by a rouge process taking up a lot of their memory.

I propose that we expose `metals.buildServerProperties` similar to `metals.serverProperties` that would modify the proper json file (`.bloop/bloop.json` or `.bsp/sbt.json`) and then show a message to restart the server. This way users will not need to dig in the Bloop docs to find what they need to do.

Changing automatically someones `bloop.json` files is not perfect, but honestly I don't see any other solution here. And we can easily append to existing options.

Also, we could use the Metals `javaHome` for Bloop if it's not already set there.

We can add a warning to the setting, that it will modify a file on disk and give them a link to it?

**Additional context**

https://github.com/scalameta/metals-vscode/issues/204

**Search terms**

bloop bsp memory sbt server

zmerr:
It resolves #2697 
It is for getting bloop JVM settings from user config and then writing it in the global bloop json file.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
@@ -1,19 +1,15 @@
 package scala.meta.internal.metals
 
-import java.io.BufferedWriter
-import java.io.ByteArrayInputStream
-import java.io.FileWriter
-import java.io.OutputStream
-import java.io.PrintStream
+import java.io._
 import java.nio.channels.Channels
 import java.nio.channels.Pipe
 import java.nio.charset.StandardCharsets
-import java.nio.file.Path
 import java.nio.file.Paths
 
 import scala.concurrent.ExecutionContextExecutorService
 import scala.concurrent.Future
 import scala.concurrent.Promise
+import scala.util.Try
 
 import scala.meta.internal.bsp.BuildChange
 import scala.meta.internal.metals.MetalsEnrichments._
@@ -88,12 +84,12 @@ final class BloopServers(
    *
    * @param expectedVersion desired version that the user has passed in. This
    *                        could either be a newly passed in version from the
-   *                            user or the default Bloop version.
-   * @param runningVersion the current running version of Bloop.
-   * @param userDefinedNew whether or not the user has defined a new version.
-   * @param userDefinedOld whether or not the user has the version running
-   *                       defined or if they are just running the default.
-   * @param reconnect      function to connect back to the build server.
+   *                        user or the default Bloop version.
+   * @param runningVersion  the current running version of Bloop.
+   * @param userDefinedNew  whether or not the user has defined a new version.
+   * @param userDefinedOld  whether or not the user has the version running
+   *                        defined or if they are just running the default.
+   * @param reconnect       function to connect back to the build server.
    */
   def ensureDesiredVersion(
       expectedVersion: String,
@@ -125,38 +121,117 @@ final class BloopServers(
     }
   }
 
+  private def writeJVMPropertiesToBloopGlobalJsonFile(
+      bloopGlobalJsonFile: File,
+      requestedBloopJvmProperties: List[String],
+      javaHome: String
+  ): Unit = {
+    val bw = new BufferedWriter(
+      new FileWriter(bloopGlobalJsonFile)
+    )
+    val text = new Gson().toJson(
+      List("javaOptions" -> requestedBloopJvmProperties, "javaHome" -> javaHome)
+    )
+    bw.write(text)
+    bw.close()
+  }
+
   def ensureDesiredJvmSettings(
-      requestedBloopJvmProperties: Option[Map[String, String]],
-      runningBloopJvmProperties: Option[Map[String, String]],
+      maybeRequestedBloopJvmProperties: Option[List[String]],
+      maybeRunningBloopJvmProperties: Option[List[String]],
+      maybeJavaHome: Option[String],
       reconnect: () => Future[BuildChange]
   ): Future[Unit] = {
-
-    if (requestedBloopJvmProperties != runningBloopJvmProperties) {
-
-      val bloopGlobalJsonPath: Path = Paths
-        .get(System.getProperty("user.home"))
-        .resolve(".bloop/bloop.json")
-      val bw = new BufferedWriter(
-        new FileWriter(bloopGlobalJsonPath.toFile)
+    val lastModifiedByMetalsTime: Long = Try {
+      val source = scala.io.Source.fromFile(
+        Paths
+          .get(System.getProperty("user.home"))
+          .resolve(".bloop/created_by_metals.lock")
+          .toFile
       )
-      val text = new Gson().toJson(requestedBloopJvmProperties.asJava)
-      bw.write(text)
-      bw.close()
-      languageClient
-        .showMessageRequest(
-          Messages.BloopVersionChange.params()
-        )
-        .asScala
-        .flatMap {
-          case item if item == Messages.BloopJvmPropertiesChange.reconnect =>
-            shutdownServer()
-            reconnect().ignoreValue
-          case _ =>
-            Future.successful(())
-        }
-    } else {
-      Future.successful(())
+      val value: Long = source.mkString.toLong
+      source.close()
+      value
     }
+      .getOrElse(0)
+
+    val result = for {
+      requestedBloopJvmProperties <- maybeRequestedBloopJvmProperties
+      javaHome <- maybeJavaHome
+      bloopGlobalJsonFile <- Try {
+        Paths
+          .get(System.getProperty("user.home"))
+          .resolve(".bloop/bloop.json")
+          .toFile
+      }.toOption
+      bloopGlobalJsonLastModifiedTime <- Try(
+        bloopGlobalJsonFile.lastModified()
+      ).toOption
+      bloopGlobalJsonFileExists <- Try(bloopGlobalJsonFile.exists()).toOption
+    } yield {
+      if (maybeRequestedBloopJvmProperties != maybeRunningBloopJvmProperties) {
+
+        if (
+          bloopGlobalJsonFileExists
+          && bloopGlobalJsonLastModifiedTime > lastModifiedByMetalsTime
+        ) {
+          languageClient
+            .showMessageRequest(
+              Messages.BloopGlobalJsonFilePremodified.params()
+            )
+            .asScala
+            .flatMap {
+              case item
+                  if item == Messages.BloopGlobalJsonFilePremodified.applyAndRestart =>
+                writeJVMPropertiesToBloopGlobalJsonFile(
+                  bloopGlobalJsonFile,
+                  requestedBloopJvmProperties,
+                  javaHome
+                )
+                shutdownServer()
+                reconnect().ignoreValue
+
+              case item
+                  if item == Messages.BloopGlobalJsonFilePremodified.saveButNotRestart =>
+                writeJVMPropertiesToBloopGlobalJsonFile(
+                  bloopGlobalJsonFile,
+                  requestedBloopJvmProperties,
+                  javaHome
+                )
+                Future.successful(())
+              case item
+                  if item == Messages.BloopGlobalJsonFilePremodified.openGlobalJsonFile =>
+                Future.successful(())
+              case item
+                  if item == Messages.BloopGlobalJsonFilePremodified.useGlobalFile =>
+                Future.successful(())
+            }
+        } else {
+          writeJVMPropertiesToBloopGlobalJsonFile(
+            bloopGlobalJsonFile,
+            requestedBloopJvmProperties,
+            javaHome
+          )
+          languageClient
+            .showMessageRequest(
+              Messages.BloopJvmPropertiesChange.params()
+            )
+            .asScala
+            .flatMap {
+              case item
+                  if item == Messages.BloopJvmPropertiesChange.reconnect =>
+                shutdownServer()
+                reconnect().ignoreValue
+              case _ =>
+                Future.successful(())
+            }
+        }
+
+      } else {
+        Future.successful(())
+      }
+    }
+    result.getOrElse(Future.successful())
   }
 
   private def connectToLauncher(
@@ -180,9 +255,11 @@ final class BloopServers(
     val serverStarted = Promise[Unit]()
     val bloopLogs = new OutputStream {
       private lazy val b = new StringBuilder
+
       override def write(byte: Int): Unit = byte.toChar match {
         case c => b.append(c)
       }
+
       def logs = b.result.linesIterator
     }
 
--- a/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
@@ -9,6 +9,8 @@ import java.nio.file.Paths
 import scala.concurrent.ExecutionContextExecutorService
 import scala.concurrent.Future
 import scala.concurrent.Promise
+import scala.util.Failure
+import scala.util.Success
 import scala.util.Try
 
 import scala.meta.internal.bsp.BuildChange
@@ -19,6 +21,7 @@ import bloop.bloopgun.BloopgunCli
 import bloop.bloopgun.core.Shell
 import bloop.launcher.LauncherMain
 import com.google.gson.Gson
+import org.eclipse.lsp4j.MessageActionItem
 import org.eclipse.lsp4j.services.LanguageClient
 
 /**
@@ -79,7 +82,7 @@ final class BloopServers(
   }
 
   /**
-   * Ensure Bloop is running the inteded version that the user has passed
+   * Ensure Bloop is running the intended version that the user has passed
    * in via UserConfiguration. If not, shut Bloop down and reconnect to it.
    *
    * @param expectedVersion desired version that the user has passed in. This
@@ -98,6 +101,7 @@ final class BloopServers(
       userDefinedOld: Boolean,
       reconnect: () => Future[BuildChange]
   ): Future[Unit] = {
+    // TODO the collection and display of errors in case something goes wrong
     val correctVersionRunning = expectedVersion == runningVersion
     val changedToNoVersion = userDefinedOld && !userDefinedNew
     val versionChanged = userDefinedNew && !correctVersionRunning
@@ -125,7 +129,7 @@ final class BloopServers(
       bloopGlobalJsonFile: File,
       requestedBloopJvmProperties: List[String],
       javaHome: String
-  ): Unit = {
+  ): Try[Unit] = Try {
     val bw = new BufferedWriter(
       new FileWriter(bloopGlobalJsonFile)
     )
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1262,6 +1262,21 @@ class MetalsLanguageServer(
           workspaceSymbols.indexClasspath()
         }
 
+//        val source = scala.io.Source.fromFile(bloopGlobalJsonPath.toUri)
+//        val bloobGlobalJsonContent = try source.mkString finally source.close()
+//        bloobGlobalJsonContent.parseJson.getAsJsonObject
+        if (userConfig.bloopJvmProperties != old.bloopJvmProperties) { // TODO is the comparison meaningful
+          val bloopGlobalJsonPath: Path = Paths
```
ckipp01, line 1269: So the logic for this can probably actually go somewhere else. Like we can check if the `bloopJvmProperties` are changed and then if so, maybe call into `BloopServers` or something and have a method in the companion object there do this sort of stuff.

ckipp01 (commented):
Thanks for starting to dig into this @zmerr! Keep in mind that one other thing we'll need to think about is how to handle this when Bloop is already running. We'll probably want to write the new settings, then restart Bloop to ensure it's taken into account.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
zmerr: thank you so much for the input…
So I would be adding the code for bloop restart as you mentioned
Now for displaying the parameters of the bloop global json file to the user in the settings menu, we would have to load it on client side right?

zmerr: @dos65 Hey Vadim, can you please provide a tad bit of input on whether this should be solved on the client side, and if so how?

zmerr: `def shutdownServer()` and `def newServer()` are now in the class of `BloopServers`, do we want them in the companion object, or do we want to find some way to access an existing created instance of `BloopServers` and call the methods on the instance?

zmerr: Because it seems that we already have access to an instance of `BloopServers` though the `bloopServers` variable of `MetalsLanguageServers`
So I guess I would go the direction of calling them on the instance…

zmerr: We have the method `ensureDesiredVersion` in `BloopServers` which I think should be blended with some `ensureDesiredJvmSettings` functionality to achieve the purpose...

zmerr:
@ckipp01 Now everything on the server side is implemented. Can you please take a second look? 
The problem with the current implementation I think is that, if both the Bloop version and the Bloop JVM properties get modified, the user receives two dialogues asking them to restart. And if the Jvm properties one is not approved, then the bloop’s global json file would not be overwritten with the newly requested jvm memory settings.
So is there currently a standard way to blend in those dialogues in one, or should I do it my own preferred way?

zmerr:
ah, one more thing: in the metals standup today, it was agreed upon that I put a checkbox on metals settings asking the user whether they want to manually enter the Jvm properties in the Metal extension’s settings, or they just want the contents of their bloop global json file to be used for that…
So that is yet to be implemented...

tgodzik:
> ah, one more thing: in the metals standup today, it was agreed upon that I put a checkbox on metals settings asking the user whether they want to manually enter the Jvm properties in the Metal extension’s settings, or they just want the contents of their bloop global json file to be used for that… So that is yet to be implemented...

I was thinking about something a bit different:
- if user doesn't have the global bloop json create then we just modify it and add an additional field to notify that it's Metals generated
- next time user modifies the settings we could check if that field exists, if it does just replace everything, otherwise offer message request with option to override or just opening the file (invoke goto lcoation).

zmerr:
> add an additional field to notify that it's Metals generated
@tgodzik then can’t it potentially cause some kind of error later in Bloop or another tool reading the file, due to an unrecognized key?
Is there a good way of adding it as some kind of comment or somewhere else that wouldn’t cause this?

tgodzik:
> > add an additional field to notify that it's Metals generated
> 
> @tgodzik then can’t it potentially cause some kind of error later in Bloop or another tool reading the file, due to an unrecognized key? Is there a good way of adding it as some kind of comment or somewhere else that wouldn’t cause this?

Ach, you are right, I hoped it would just get ignored. We could add an additional file (kind of a lock file) like `.bloop/metals.lock` maybe?

zmerr:
> We could add an additional file (kind of a lock file) like .bloop/metals.lock maybe
@tgodzik Ok, then I would go with that approach… I hope in the updates or reinstallations of Bloop, it won’t get erased…
Also, what if the user later modifies the json file after the `metals.lock` is created? I would say, let’s just persist the time at which we are modifying the bloop json file by reading its "last modified time" value after we create or write to it in the metals server internal files, then in the next times, we compare the modification date to our persisted one, to see if it was us who modified it...

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -1,15 +1,17 @@
 package scala.meta.internal.metals
 
-import java.io.ByteArrayInputStream
-import java.io.OutputStream
-import java.io.PrintStream
+import java.io._
```
tgodzik, line 3: Could we revert this? We usually explicitely specify all the imports and avoid wildcard ones.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -121,6 +127,182 @@ final class BloopServers(
     }
   }
 
+  private def writeJVMPropertiesToBloopGlobalJsonFile(
+      bloopGlobalJsonFile: File,
```
tgodzik, line 131:
```suggestion
      bloopGlobalJsonFile: AbsolutePath,
```

this would allow to use `.writeText` extension method without the additional BufferedWriter etc.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -121,6 +127,182 @@ final class BloopServers(
...
+    val source = scala.io.Source.fromFile(
+      Paths
+        .get(System.getProperty("user.home"))
+        .resolve(".bloop/created_by_metals.lock")
+        .toFile
+    )
+    val value: Long = source.mkString.toLong
+    source.close()
+    value
+  }.getOrElse(0)
```
tgodzik, line 155:
```suggestion
  private def getBloopGlobalJsonLastModifiedByMetalsTime: Option[Long] = {
    val lockFilePath = Paths
        .get(System.getProperty("user.home"))
        .resolve(".bloop/created_by_metals.lock")
    val absolutePath = AbsolutePath(lockFilePath)
    if (absolutePath.exists())
       Try(text.toLong).toOption
    else
      None
  }
```

We can then check if it's defined to see if it was MEtals that modified the Bloop file.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -121,6 +127,182 @@ final class BloopServers(
...
+
+  }
+
+  private def getBloopGlobalJsonFile(): Option[File] =
+    Try {
+      Paths
+        .get(System.getProperty("user.home"))
+        .resolve(".bloop/bloop.json")
+        .toFile
+    }.toOption
```
tgodzik, line 237:
```suggestion
  private def getBloopGlobalJsonFile(): Option[AbsolutePath] =
     sys.props.get("user.home").map{
       home =>
         AbsolutePath(Paths.get(home).resolve(".bloop/bloop.json"))
     }
```
it will be much easier to use since it already contains a bunch of utility methods.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -121,6 +127,182 @@ final class BloopServers(
...
+  ): Future[Unit] = {
+
+    val result = for {
+      requestedBloopJvmProperties <- maybeRequestedBloopJvmProperties
+      javaHome <- maybeJavaHome
+      bloopGlobalJsonFile <- getBloopGlobalJsonFile()
+      bloopGlobalJsonLastModifiedTime <- Try(
+        bloopGlobalJsonFile.lastModified()
+      ).toOption
+      bloopGlobalJsonFileExists <- Try(bloopGlobalJsonFile.exists()).toOption
```
tgodzik, line 268:
```suggestion
      bloopGlobalJsonFileExists = bloopGlobalJsonFile.exists()
```

should be enough. Exists doesn;t throw exceptions

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -296,6 +296,55 @@ object Messages {
     }
   }
 
+  object BloopGlobalJsonFilePremodified {
+    def applyAndRestart: MessageActionItem =
+      new MessageActionItem("Apply and Restart Bloop")
+    def saveButNotRestart: MessageActionItem =
```
tgodzik, line 302: I think we don't need this option. If some one already modifies the jvm settings their intention is to restart it as well.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -121,6 +127,182 @@ final class BloopServers(
...
+    } yield
+      if (
+        maybeRequestedBloopJvmProperties != maybeRunningBloopJvmProperties
```
tgodzik, line 271:
```suggestion
        maybeRequestedBloopJvmProperties != maybeRunningBloopJvmProperties && maybeRequestedBloopJvmProperties.isDefined
```
?

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -121,6 +127,182 @@ final class BloopServers(
...
+      ) // the properties are updated
+        if (
+          bloopGlobalJsonFileExists
+          && bloopGlobalJsonLastModifiedTime > getBloopGlobalJsonLastModifiedByMetalsTime
```
tgodzik, line 275: with one of the above suggestion we can do
```suggestion
          && getBloopGlobalJsonLastModifiedByMetalsTime.extsts( bloopGlobalJsonLastModifiedTime > _)
```

tgodzik (changes requested):
Great work! The approach looks solid, I left some overall comments, but nothing major.

Did you manage to make it work with the VS Code extension?

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -98,6 +103,7 @@ final class BloopServers(
       userDefinedOld: Boolean,
       reconnect: () => Future[BuildChange]
   ): Future[Unit] = {
+    // TODO the collection and display of errors in case something goes wrong
```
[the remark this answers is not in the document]
zmerr: The ones related to not having permission to read or write to the file.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: It is done automatically be either Scalafix or Scalafmt
So I’m afraid after reverting it, it would again turn back to the previous state, perhaps? any suggestion?

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: Good hint! Thanks, I would use it...

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: oops, just realized why the bug existed for not working in the extension...
The for comprehension conditions the execution. of logic to when the bloopGlobalJson file preexisted. So I need to revise that code.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: this part should be rewritten fundamentally.
The logic of the for comprehension is buggy

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
zmerr: ok I would remove it

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: makes sense...

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: I need to look at this, because as I mentioned, the logic of the for comprehension is buggy, so first it should get revised.

zmerr:
@ckipp01 do you think this should be done for other editors as well?

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -121,6 +125,207 @@ final class BloopServers(
...
+    }
+
+  }
+
+  private def getBloopGlobalJsonFilePath(): Option[AbsolutePath] =
+    Try {
+      AbsolutePath(
+        Paths
+          .get(System.getProperty("user.home"))
+          .resolve(".bloop/bloop.json")
```
ckipp01, line 237: We check for `.bloop/bloop.json`, but what about users that are utilizing the `.bloop/.jvmopts` file? For example that's what I was using locally to set my Bloop jvm opts, and now I have both that file and a newly created `.bloop/bloop.json` file.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -121,6 +125,207 @@ final class BloopServers(
...
+      )
+    }.toOption
+
+  private def getBloopCreatedByMetalsFilePath(): Option[AbsolutePath] =
```
ckipp01, line 241: So this method and the method above `getBloopGlobalJsonFilePath` are the exact same except for the property and the filename they are trying to grab. Let's refactor this to a single method that takes in a property and a filename and returns an `Option[AbsolutePath]`.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -121,6 +125,207 @@ final class BloopServers(
...
+    val result =
+      for { // if the requestedBloopJvmProperties and bloopGlobalJsonFilePath are defined
+        requestedBloopJvmProperties <- maybeRequestedBloopJvmProperties
+        bloopGlobalJsonFilePath <- getBloopGlobalJsonFilePath()
+        bloopCreatedByMetalsFilePath <- getBloopCreatedByMetalsFilePath()
+      } yield
+        if (
+          maybeRequestedBloopJvmProperties != maybeRunningBloopJvmProperties && requestedBloopJvmProperties.nonEmpty
+        ) // the properties are updated
+          if (
```
ckipp01, line 288: Since we have some nested `ifs` here let's wrap them in `{}` for easier reading

ckipp01 (commented):
Great job on this so far @zmerr! I've left a few suggestions, questions, and nits.

> @ckipp01 do you think this should be done for other editors as well?

So since this is using existing LSP methods already this should already work as is for other editors. For example in `nvim-metals` I just needed to add the newly created `bloopJvmProperties` as a possible configuration value and everything else _just worked_ ™️

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -121,6 +125,207 @@ final class BloopServers(
...
+
+  private def getBloopGlobalJsonLastModifiedTime(
+      bloopGlobalJsonFilePath: AbsolutePath
+  ): Long =
+    Try {
+      bloopGlobalJsonFilePath.toNIO.toFile.lastModified()
+    }.getOrElse(0)
+
+  /**
+   * First we check if the user requested to update the Bloop JVM properties through the extension.
```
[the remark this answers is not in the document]
zmerr: Then why didn’t scalafmt do it automatically this time? I am sure I ran it…
Can you please provide a tad bit of more hints?

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -121,6 +125,207 @@ final class BloopServers(
...
+   * @param maybeRequestedBloopJvmProperties Bloop JVM Properties requested through the Metals Extension settings
+   * @param maybeRunningBloopJvmProperties
+   * @param maybeJavaHome                    Metals' `javaHome`, which Bloop should also preferebly use
+   * @param reconnect                        function to connect back to the build server.
+   * @return `Future.successful` if the purpose is achieved or `Future.failure` if a problem occured such as lacking enough permissions to open or write to files
+   */
+  def ensureDesiredJvmSettings(
+      maybeRequestedBloopJvmProperties: Option[List[String]],
+      maybeRunningBloopJvmProperties: Option[List[String]],
+      maybeJavaHome: Option[String],
```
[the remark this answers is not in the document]
zmerr: Wait a minute…
Why are there two `BloopJvmProperties` variables in here? I think I might have added it twice by mistake.
Also, frankly adding `maybe` really helps myself in writing code… So can it stay?

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -177,6 +178,15 @@ object UserConfiguration {
         """|This version will be used for the Bloop build tool plugin, for any supported build tool,
            |while importing in Metals as well as for running the embedded server""".stripMargin
       ),
+      UserConfigurationOption(
+        "bloop-jvm-properties",
+        """["-Xmx1G"].""",
+        """["-Xmx1G"]""",
+        "Bloop JVM Properties",
+        """|Optional list of JVM properties to pass along to the Bloop server.
```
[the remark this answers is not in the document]
zmerr: Sure, I am going to add that “It replaces the existing Bloop JVM properties file and overwrites it."

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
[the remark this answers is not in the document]
ckipp01: > Also, frankly adding maybe really helps myself in writing code… So can it stay?

Of course :) it was just a suggestion. It comes down to personal preference, and I see in some other places we do the same, so no worries. Just one tip is to not rely too much on it, because sooner or later you'll hit on something called `maybe*` and it won't be an `Option`.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -296,6 +296,52 @@ object Messages {
...
+
+  object BloopJvmPropertiesChange {
+    def reconnect: MessageActionItem =
+      new MessageActionItem("Restart Bloop")
+    def notNow: MessageActionItem =
+      new MessageActionItem("Not now")
+    def params(): ShowMessageRequestParams = {
+      val params = new ShowMessageRequestParams()
+      params.setMessage(
+        s"Bloop JVM properties were updated; do you want to restart the running Bloop server?"
```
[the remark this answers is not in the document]
zmerr: Perfectly fluent! I just think the second sentence can be changed to 
"Do you want to restart Bloop in order for them to take effect?"

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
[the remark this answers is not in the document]
ckipp01: Sure, so with no settings Scalafmt will wrap your comments for you, but part of that is dropping them down so:

```scala
/** my docs
 * @param foo
 *      it drops the explanation of foo down here instead of leaving. it above
 */
```
So because it drops it down and (I personally) find that super annoying since it wastes a line, we decided to put the following setting in our `.scalafmt.conf` file to avoid it:

```
docstrings.wrap = no
```
But there is a trade off, because now even when our comments are super long, it won't wrap them.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: ah, thank you very much for providing this extra information. This can definitely cause problems. And is the contents of the `.bloop/.jvmopts` file supposed to be the same as the `.bloop/bloop.json` one?
What are your suggestions for the logic of resolving this when the `.jvmopts` file is present?

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: Perfect! Thanks!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: Very sensible!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: @ckipp01  I need help with this!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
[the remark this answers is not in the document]
zmerr: So `docstrings.wrap = yes` ?

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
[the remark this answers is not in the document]
zmerr: Thanks for the tip!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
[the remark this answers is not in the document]
ckipp01: No no, don't change the config. Just manually wrap your lines if possible so they aren't super super wide. Like

```
* here is my super super long thing that is super long long and if I now go wider than my screen might allow so I might want to wrap
```
become
```
* here is my super super long thing that is super long long and if I
* now go wider than my screen might allow so I might want to wrap
```

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: Looks like it was actually done by Intellij. Just replace it back with
```
import java.io.ByteArrayInputStream
import java.io.OutputStream
import java.io.PrintStream
```

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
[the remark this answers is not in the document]
ckipp01: I think this comment might be tied to the wrong place, but you should be able to just change your wild card import to
```
import java.io.ByteArrayInputStream
import java.io.OutputStream
import java.io.PrintStream
```
I don't think scalafix will change it.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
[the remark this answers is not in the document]
zmerr: LOL

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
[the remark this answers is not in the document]
ckipp01: I think actually just leaving it as a statement would be best:

```New Bloop JVM properties detected. Bloop will need to be restarted in order for them to take effect.```

File: metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala (hunk 11 above)
Comment (ckipp01) [reply metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala:237]: Well I'm not sure actually. @tgodzik is it not recommend with Bloop to use the `.bloop/.jvmopts` file? It might just be this is legacy and I shouldn't be doing it.

File: metals/src/main/scala/scala/meta/internal/metals/Messages.scala (hunk 17 above)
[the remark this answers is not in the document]
Comment (zmerr) [reply metals/src/main/scala/scala/meta/internal/metals/Messages.scala:332]: perfect, thanks

File: metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala (hunk 2 above)
Comment (ckipp01) [reply metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala:3]: You should be able to just change it to the code snippet I have above. Scalafix won't replace it.

File: metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala (hunk 11 above)
Comment (tgodzik) [reply metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala:237]: I think the best option would be to follow similar thing as for an existing `bloop.json` but the user decides we should remove `.jvmopts` file and otherwise suggest they migrate to `bloop.json` themselves

`.jvmopts` format is just a list of options, it's not a json file at all.

File: metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala (hunk 2 above)
Comment (ckipp01) [reply metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala:3]: > Looks like it was actually done by Intellij. 😱 Hacking on Metals with IntelliJ should be a crime 🚨 .

_Joking of course_

File: metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala [hunk 18]
```diff
@@ -121,6 +125,207 @@ final class BloopServers(
...
+          if item == Messages.BloopGlobalJsonFilePremodified.useGlobalFile =>
+        Future.successful(())
+    }
+  }
+
+  private def updateBloopGlobalJsonFileThenRestart(
+      bloopGlobalJsonFilePath: AbsolutePath,
+      bloopCreatedByMetalsFilePath: AbsolutePath,
+      requestedBloopJvmProperties: List[String],
+      maybeJavaHome: Option[String],
```
tgodzik, line 204: I wonder also if we should update when just java home is set in case Metals and Bloop runs a different Java version. This could make the logic a bit more complicated, but I would treat `javaHome` as kind of additional property.

I've seen a bunch of issues related to that recently. We can follow up in a different PR for sure once this is merge or we can iterate on this one.

Opinions?

tgodzik (changes requested):
Great work so far! I left some more ideas, mostly because doing this PR right would help a lot of people.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -121,6 +125,207 @@ final class BloopServers(
...
+   * @param reconnect                        function to connect back to the build server.
+   * @return `Future.successful` if the purpose is achieved or `Future.failure` if a problem occured such as lacking enough permissions to open or write to files
+   */
+  def ensureDesiredJvmSettings(
+      maybeRequestedBloopJvmProperties: Option[List[String]],
+      maybeRunningBloopJvmProperties: Option[List[String]],
+      maybeJavaHome: Option[String],
+      reconnect: () => Future[BuildChange]
+  ): Future[Unit] = {
+    // TODO the collection and display of errors in case something goes wrong
```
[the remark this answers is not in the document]
zmerr: I’m gonna finish it in this PR if I know how to report errors! Through the `languageClient`?

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: https://github.com/scalacenter/bloop/pull/1707

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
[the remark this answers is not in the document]
tgodzik: If it's jsut a message that shows up rarely in case of problems it's probably ok to use `showMessage` method

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
ckipp01: Yup, perfect 👍🏼

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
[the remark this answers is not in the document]
zmerr: done

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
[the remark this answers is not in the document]
zmerr: done

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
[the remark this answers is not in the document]
zmerr: done

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: I would do it in another PR.

tgodzik: Ok cool! Let's do it in a different PR

tgodzik (changes requested):
Just tested it locally and everything looks great!

Could you rebase your PR on the newest changes? We can merge afterwards, thanks!

tgodzik (approved):
Let's merge this!

PR #3748, opened 2022-03-22
Enclosing brace completion

zmerr:
This is a rough draft and the location of insertion is not correct: immediately after the opening brace!

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/formatting/BraceEnclosingFormatter.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/formatting/BraceEnclosingFormatter.scala
@@ -1,0 +1,48 @@
+package scala.meta.internal.metals.formatting
+
+import org.eclipse.lsp4j.{Range, TextEdit}
+
+import scala.meta.internal.metals.UserConfiguration
+import scala.meta.internal.mtags.MtagsEnrichments._
+import scala.meta.{Term, Tree}
+
+case class BraceEnclosingFormatter(userConfig: () => UserConfiguration)
+  extends OnTypeFormatter
+    with RangeFormatter {
+
+  private val openingBrace = '{'
+  private val closingBrace = "}"
+
+  override def contribute(
+                           params: OnTypeFormatterParams
+                         ): Option[List[TextEdit]] = {
+    val position = params.position
+    val triggerChar = params.triggerChar
+    if (triggerChar == openingBrace) {
+      val maybeApplyTree =
+        if (params.range.getStart == params.range.getEnd)
+          params.trees
+            .findLastEnclosingAt[Tree](
+              params.path,
+              position,
+              applyWithSingleFunction
+            )
+        else None
+      maybeApplyTree.map { applyTree =>
+        val textEdit = new TextEdit()
+        textEdit.setNewText(closingBrace)
+        applyTree.pos.toLSP.getEnd
+        textEdit.setRange(new Range(applyTree.pos.toLSP.getEnd, applyTree.pos.toLSP.getEnd))
+        List(textEdit)
+      }
+
+    }
+    else None
+  }
+
+  private def applyWithSingleFunction: Tree => Boolean = {
+    case _: Term.Apply => true
+    case _ => false
+  }
+
+}
--- a/metals/src/main/scala/scala/meta/internal/metals/formatting/BraceEnclosingFormatter.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/formatting/BraceEnclosingFormatter.scala
@@ -4,7 +4,7 @@ import org.eclipse.lsp4j.{Range, TextEdit}
 
 import scala.meta.internal.metals.UserConfiguration
 import scala.meta.internal.mtags.MtagsEnrichments._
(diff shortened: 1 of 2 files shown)


ckipp01:
Thanks for creating this @zmerr! Could you maybe include a bit more information on what the inspiration for this was? Is this fixing a problem that you have been hitting on or that has been reported somewhere? I typically consider the insertion of a closing brace as something the vast majority of editors will automatically take care of and is mostly language agnostic. So I'm not 100% sure this is something we'd want as an `onTypeFormatting` feature unless there is a good reason for it.

dos65:
I agree with @ckipp01. Most editors have such feature.
If it's an issue only with vscode, we can try to fix on client side. Currently we have [some rules](https://github.com/scalameta/metals-vscode/blob/main/src/extension.ts#L1196) in extension

zmerr:
@ckipp01 thanks for the comment. This was for solving my own problem. The insertion of enclosing brace in this PR is **not** meant to be immediately after the opening one, as the existing functionality is providing.
It is meant for to be inserted at the end of the tree at whose start position the opening brace is inserted. This removes the confusion of programmer when editing code, and creates a smoother coding experience. 
Particularly, in Scala 3 when there can be multiple lines of braceless code...

zmerr:
@ckipp01 I mean not necessarily at the end of the tree, but it can also take indentation into account, so each time you type an opening brace, anything indented under it can be included in the enclosing brace….

ckipp01:
I think the general flow would typically be that you create a `{` and then a `}` is automatically created, and then it's up to your editor rules to ensure that the `}` is placed in the right place. I agree with what @dos65 mentioned and think this should probably be handled on the client side.

zmerr:
@ckipp01 thanks for pointing me to the right direction. So is there any particular part of the code on the extension that you think the changes should get applied?

ckipp01:
> @ckipp01 thanks for pointing me to the right direction. So is there any particular part of the code on the extension that you think the changes should get applied?

😅 I'm not a VS Code user so I'm not super sure, but @dos65 will probably have a better idea.

zmerr:
@dos65 eagerly waiting for you input on this ….

tgodzik:
I suggested that we might want to use onTypeFormatting since we don't have any information about trees in VS Code or any other client. We could try with indentation, but I am not sure even if we can modify two lines that could be far apart?

ckipp01:
> I suggested that we might want to use onTypeFormatting since we don't have any information about trees in VS Code or any other client. We could try with indentation, but I am not sure even if we can modify two lines that could be far apart?

Gotcha. I think I'm still a bit confused on the actual issue. Looking at the code currently this is inserting a closing brace when there isn't one correct? I've never actually encountered this issue where I don't have a closing brace. I'm also a bit concerned about the timing on this because I feel like you'll have a user that triggers a `{` and then what happens if metals gets the request and responds with a `}` and then my editor also created a `}`. Or am I misunderstanding the use case for this? It just seems like a lot to use LSP communication to insert a `}`.

zmerr:
> you'll have a user that triggers a `{` and then what happens if metals gets the request and responds with a `}` and then my editor also created a `}`
@ckipp01 half of the problem is detecting if the editor has created an enclosing brace or not and at what stage, because if it is not at the right location, it should get removed by LSP

tgodzik:
The main intention is similar to what Intellij does with indented code, it would try to enclose the whole indented block in braces. Might be particulary useful if we have a single statement function and we want to add more statements. Another option would be to do it via code action if we don't find a satisfactory solution here.

ckipp01:
> @ckipp01 half of the problem is detecting if the editor has created an enclosing brace or not and at what stage, 
I guess I disagree, in what scenario does your editor not create an enclosing brace when you create an opening brace?

> because if it is not at the right location, it should get removed by LSP

I actually _really_ don't think that metals should remove a closing brace from your code and then create another one in a different place. I think that's just going down a rabbit hole of trying to get the placement _exactly_ right for a problem that I really don't think exists. Does this happen often for you? In other places in metals we've sort of went with "formatting should be done by scalafmt", and that's something I think we should follow here as well.

zmerr:
> The main intention is similar to what Intellij does with indented code, it would try to enclose the whole indented block in braces.

Yes, this is the main thing that I had in mind… I think the only thing getting in the way is the enclosing brace that the editor is putting immediately after it. So I am wondering if in the onTypeFormatter, alongside checking for the opening brace, we can also check for the enclosing one, provided that the character typed right before it is an opening brace. I think most probably the server would not be able to detect if an enclosing brace is typed by the user or the editor though…
It would most probably consider both as just “typing"

zmerr:
I would leave this untouched for some while to see how the discussion would be concluding @ckipp01 @tgodzik @dos65

ckipp01:
> Btw. @ckipp01 not sure if we explained the exact use case, so just in case for example for:

ahhhhh, that is super helpful. That's why I asked. This actually is a good use case. I didn't realize it before. For sure worth giving it a try then.

tgodzik:
closing for now since the entire topic is a bit complicated currently. We can reopen once we start working on this again.

PR #500, opened 2019-01-25
Trigger didFocus when current compile may affect focused buffer.

olafurpg:
Fixes #497. Previously, we did not trigger didFocus compiles if there
was an ongoing compilation that might affect the focused buffer. This
could results in missing compilation when fixing an error in main and
then instantly focus on a file in tests before the main compilation
finished.

PR #3752, opened 2022-03-23
Introduce uses of buildTarget/inverseSources

Duhemm:
Previously, Metals would only use the result of `buildTarget/sources` to
figure which target owns a given source file. Because
`buildTarget/sources` is issued as part of build import, the list of
sources (directories and files) returned by the build server could be
stored and re-used to immediately determine the owning target, given a
source file.

It turns out, however, that for some targets this knowledge can become
outdated. For instance, Bloop supports defining the source files as a
set of globs [1]. In this case, Bloop will respond to the
`buildTarget/sources` request issued during build import with the list
of files that match at this point in time. If the developer create a new
file which would be matched by Bloop's globs, Metals will be unable to
associate it with a target.

Determining which targets own a given source file is supported by the
BSP protocol via the `buildTarget/inverseSources` request [2].

This commit introduces a new method `BuildTargets#inverseSourcesBsp`,
which finds the target owning the given source file, just like the
existing `BuildTargets#inverseSources` does, but using BSP if the
original method couldn't find a match. Because this new method is
asynchronous (it communicates with the BSP server), migrating all
call-sites to use it would require many changes to Metals.

When used with Bloop, requires scalacenter/bloop#1704.

[1]: https://github.com/scalacenter/bloop/blob/a9b3ea59631ca300c06394effbdf4b2a8e88dbf1/config/src/main/scala/bloop/config/Config.scala#L231-L236
[2]: https://build-server-protocol.github.io/docs/specification.html#inverse-sources-request

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -233,6 +233,10 @@ final class BuildTargets() {
         }
         Future.sequence(queries).map { results =>
           val target = results.flatten.maxByOption(buildTargetsOrder)
+          for {
+            tgt <- target
+            data <- targetData(tgt)
+          } data.addSourceItem(source, tgt)
           target
         }
       case some =>
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -7,6 +7,8 @@ import java.{util => ju}
 
 import scala.annotation.tailrec
 import scala.collection.mutable
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
 import scala.util.control.NonFatal
 
 import scala.meta.internal.io.PathIO
@@ -16,6 +18,8 @@ import scala.meta.io.AbsolutePath
 
 import ch.epfl.scala.bsp4j.BuildTarget
 import ch.epfl.scala.bsp4j.BuildTargetIdentifier
+import ch.epfl.scala.bsp4j.InverseSourcesParams
+import ch.epfl.scala.bsp4j.TextDocumentIdentifier
 
 /**
  * In-memory cache for looking up build server metadata.
@@ -208,6 +212,31 @@ final class BuildTargets() {
         .orElse(inferBuildTarget(source))
     } else {
       Some(orSbtBuildTarget.maxBy(buildTargetsOrder))
+    }
+  }
+
+  def inverseSourcesBsp(
+      source: AbsolutePath
+  )(implicit ec: ExecutionContext): Future[Option[BuildTargetIdentifier]] = {
+    inverseSources(source) match {
+      case None =>
+        val identifier = new TextDocumentIdentifier(
+          source.toTextDocumentIdentifier.getUri()
+        )
+        val params = new InverseSourcesParams(identifier)
+        val connections =
+          data.fromIterators(_.targetToConnection.values.toIterator).distinct
+        val queries = connections.map { connection =>
+          connection
+            .buildTargetInverseSources(params)
+            .map(_.getTargets.asScala.toList)
+        }
+        Future.sequence(queries).map { results =>
+          val target = results.flatten.maxByOption(buildTargetsOrder)
+          target
+        }
+      case some =>
+        Future.successful(some)
     }
   }
 
--- a/build.sbt
+++ b/build.sbt
@@ -176,7 +176,7 @@ lazy val V = new {
   val ammonite213Version = "2.13.7"
 
   val ammonite = "2.5.2"
-  val bloop = "1.4.13-48-a9b3ea59"
+  val bloop = "1.4.13-50-8332ee9c"
   val bloopNightly = bloop
   val bsp = "2.0.0-M15"
   val coursier = "2.1.0-M5"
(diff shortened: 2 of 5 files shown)


metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -211,6 +215,31 @@ final class BuildTargets() {
...
+        val connections =
+          data.fromIterators(_.targetToConnection.values.toIterator).distinct
+        val queries = connections.map { connection =>
+          connection
+            .buildTargetInverseSources(params)
+            .map(_.getTargets.asScala.toList)
+        }
+        Future.sequence(queries).map { results =>
+          val target = results.flatten.maxByOption(buildTargetsOrder)
+          target
```
tanishiking, line 236: Can we register the result to https://github.com/scalameta/metals/blob/934ac115a3469f5585f284023da804f04bfa2877/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala#L41-L42 to reuse the result as cache in the next request?

tanishiking (commented):
Thank you so much for working on this, @Duhemm !
One minor comment, but this will be definitely a great improvement :)

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
Duhemm: Thanks for the suggestion! I made the change. Is that what you had in mind?

Duhemm:
Thank you for the initial feedback. How do you feel about merging this and opening an issue about slowly migrating calls to `BuildTargets#inverseSources` over to `BuildTargets#inverseSourcesBsp`?

I think the failures are caused by flaky tests?

kpodsiad (approved):
LGTM!

> How do you feel about merging this and opening an issue about slowly migrating calls to BuildTargets#inverseSources over to BuildTargets#inverseSourcesBsp?

It's a good idea to keep track of this migration.

tanishiking (approved):
LGTM!

> merging this and opening an issue about slowly migrating calls to BuildTargets#inverseSources over to BuildTargets#inverseSourcesBsp?

Sounds good :+1:

Duhemm:
Thanks!

There's one test that consistently fails on CI, but consistently passes on my machine... I'm not sure what's going on, and it's unclear to me how this patch could affect `tests.DidFocusLspSuite#497`. `buildTarget/inverseSources` doesn't appear in the BSP trace for that test 🤔

tgodzik:
> There's one test that consistently fails on CI, but consistently passes on my machine... I'm not sure what's going on, and it's unclear to me how this patch could affect `tests.DidFocusLspSuite#497`. `buildTarget/inverseSources` doesn't appear in the BSP trace for that test thinking

I wonder why that assert fails, it seems that we are returning something else than `didCompile == Compiled` <- Maybe it's `didCompile == AlreadyCompiled` ? I haven't touched anything around there, but maybe invoking inver soruces might trigger compilation and then we we try again it's already compiled?

tanishiking:
> it seems that we are returning something else than didCompile == Compiled <- Maybe it's didCompile == AlreadyCompiled ?

I did a bit of print debug [here](https://github.com/tanishiking/metals/pull/10) as I couldn't reproduce it in my local environment and looks like yes. The result returns `AlreadyCompiled` instead of `Compiled`.
https://github.com/tanishiking/metals/runs/5707840749?check_suite_focus=true#step:5:4287

### Analysis
I assume this test is inherently flaky:

- Looking back at the [original issue](https://github.com/scalameta/metals/pull/500) what we try to test here is `didFocus` invoke the compile (which means it returns `Compiled`), if the target file is affected by the **current compilation**.
- So, we have to make sure we call `didFocus(.../B.scala)` when a compilation is going on (`by `didSave(.../A.scala)`)
  - [isCompiling for project a will be true during compilation](https://github.com/Duhemm/metals/blob/76df7f8ed35a656693c8606fdaab81cfe4416212/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala#L219-L224), and [isAffectedByCurrentCompilation](https://github.com/tanishiking/metals/blob/76df7f8ed35a656693c8606fdaab81cfe4416212/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1138-L1144) should be true
- However, `isAffectedByCurrentCompilation` is false in the test because `compilations.currentlyCompiling` is empty in the test. That's why we receive `AlreadyCompiled` here.
  - because we're trying to `didFocus` during the compilation by calling `didFocus` immediately after the `didSave` [here](https://github.com/tanishiking/metals/blob/76df7f8ed35a656693c8606fdaab81cfe4416212/tests/unit/src/test/scala/tests/DidFocusLspSuite.scala#L121-L127) but it doesn't make sure didFocus occurs during compilation.

I'm not sure why this PR triggers this to fail, but I guess introducing `inverseSources` somehow modified the timing of compilation? 🤔 

### How can we stabilize the test
I guess we should be able to add a delay or something in the test environment around here https://github.com/Duhemm/metals/blob/76df7f8ed35a656693c8606fdaab81cfe4416212/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala#L219-L224 to make sure we call `didFocus` during the compilation is running.

Anyway, it's not your fault @Duhemm sorry to bother you :)

tanishiking:
This change could fix the flaky test (maybe?) https://github.com/tanishiking/metals/pull/10/commits/2ed68da0f694a00fb657e6ad134a317d8121f3c1

Duhemm:
Thanks a lot for the investigation @tanishiking ! Please feel free to push your fix to this branch (I believe you should be able to), otherwise I'll cherry-pick your commit later 🤞

tanishiking:
Ok, let's see my change actually fix the test :eyes: Then I will push the change to this PR, and I want someone others to check my workaround :)

tgodzik (approved):
LGTM let's merge this!

PR #3753, opened 2022-03-24
dep: bump release candidate version to 3.1.2-RC3

ckipp01:


build.sbt
```diff
@@ -170,7 +170,7 @@ lazy val V = new {
   val scala212 = "2.12.15"
   val scala213 = "2.13.8"
   val scala3 = "3.1.1"
-  val nextScala3RC = "3.1.2-RC2"
+  val nextScala3RC = "3.1.2-RC3"
```
tgodzik, line 173: I am thinking that maybe we should put the old version to deprecated versions so that people on edge don't have their Metals stop working

ckipp01: Ah good call. I'll let CI finish just to ensure nothing breaks, and then add RC2 in `deprecatedScala3Versions`.

tgodzik: Awesome, thanks!

tgodzik: It seems to be working, I am being asked the compiler people for this so I could maybe finish up the PR if you don't have time?

tgodzik (approved):
LGTM!

PR #3701, opened 2022-03-10
fix: do not decode every URI when invoking `toAbsolutePath`

kpodsiad:
Yesterday on discord it was reported that an internal server error happens when Metals have to deal with a directory that contains spaces. The source of the problem was fact that Mtags and Metals xtension method for string differentiated in implementation.

PR #3756, opened 2022-03-25
Fix issues with non alpha numeric build target names

tgodzik:
This issues shows up for example in Mill projects which have the `[2.13.8]` added to indicate the version of the target.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala
@@ -100,7 +100,10 @@ class BuildTargetInfo(buildTargets: BuildTargets) {
       )
     })
     commonInfo.foreach(info => {
-      output ++= getSection("Base Directory", List(info.baseDirectory))
+      output ++= getSection(
+        "Base Directory",
+        List(URIEncoderDecoder.decode(info.baseDirectory))
+      )
       output ++= getSection("Source Directories", getSources(info))
     })
 
--- a/metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
@@ -144,7 +144,7 @@ final class FileDecoderProvider(
    * metalsDecode:file:///workspacePath/buildTargetName.metals-buildtarget
    */
   def decodedFileContents(uriAsStr: String): Future[DecoderResponse] = {
-    Try(URI.create(uriAsStr)) match {
+    Try(URI.create(URIEncoderDecoder.encode(uriAsStr))) match {
       case Success(uri) =>
         uri.getScheme() match {
           case "jar" => Future { decodeJar(uriAsStr) }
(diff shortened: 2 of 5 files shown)


metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
```diff
@@ -615,3 +619,16 @@ final class FileDecoderProvider(
         )
     }
 }
+
+object URIEncoder {
```
tgodzik, line 623: Finally decided to do our own encoder for things not accepted by URI.create. I can't find anything in the standard library that does it

kpodsiad: Now there is unnecessary code in `CommandHTMLFormat`

kpodsiad: Or maybe not, encoding logic is slightly different :/

metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala
```diff
@@ -100,7 +101,10 @@ class BuildTargetInfo(buildTargets: BuildTargets) {
       )
     })
     commonInfo.foreach(info => {
-      output ++= getSection("Base Directory", List(info.baseDirectory))
+      output ++= getSection(
+        "Base Directory",
+        List(URLDecoder.decode(info.baseDirectory, "UTF-8"))
```
kpodsiad, line 106: Calling `URLDecoder.decode` on strings that may not be encoded is dangerous. Here, `baseDirectory` can contain spaces and it will cause `URLDecoder` to throw an exception.

Probably we should write our own `URIDecoder`

kpodsiad (changes requested):

metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala
kpodsiad: Same case as in https://github.com/scalameta/metals/pull/3701

Arthurm1 (approved):
Makes sense.

It's probably a good idea to write a custom encoder/decoder as I imagine this will come up again on another code path.

metals/src/main/scala/scala/meta/internal/metals/FileDecoderProvider.scala
tgodzik: Ok, I could merge them both, but it would make the logic a bit more complicated, so I decided against, since we don't need a decoder for CommandHTMLFormat

metals/src/main/scala/scala/meta/internal/metals/BuildTargetInfo.scala
tgodzik: I created a separate decode and encoder for URIs, it's probably not perfect, but seems to work fine for the current cases.

tests/unit/src/test/scala/tests/UriEncoderDecoderSuite.scala
```diff
@@ -0,0 +1,42 @@
...
+      "metalsDecode:file:///home/tgodzik/Documents/metals/tests/slow/target/e2e/fileDecoderProvider/buildtarget/a[2.13.8].metals-buildtarget"
+    val expected =
+      "metalsDecode:file:///home/tgodzik/Documents/metals/tests/slow/target/e2e/fileDecoderProvider/buildtarget/a%5b2.13.8%5d.metals-buildtarget"
+
+    val obtained = URIEncoderDecoder.encode(entry)
+    assertEquals(
+      obtained,
+      expected
+    )
+  }
```
kpodsiad, line 18: I would add a newlines between tests

tgodzik: Added

Issue #3020, opened 2021-08-03
.bloop folder grows without boundaries (bloop-bsp-clients-classes)


arbitrary-dev:
**Describe the bug**

Each new run of vim + coc-metals results in new `bloop-bsp-clients-classes` entries which are identical to the previous ones:

```
$ du -sh .bloop/core/bloop-bsp-clients-classes/* 
712K	.bloop/core/bloop-bsp-clients-classes/classes-Metals-5TfFX5RASPeyHEOh-O7kGQ==
712K	.bloop/core/bloop-bsp-clients-classes/classes-Metals-VnEP2HueTwu_JQsGeGv7pg==
712K	.bloop/core/bloop-bsp-clients-classes/classes-Metals-ZU6uaWGiQ-eReOpykI5IaQ==
424K	.bloop/core/bloop-bsp-clients-classes/test-classes-Metals-5TfFX5RASPeyHEOh-O7kGQ==
424K	.bloop/core/bloop-bsp-clients-classes/test-classes-Metals-VnEP2HueTwu_JQsGeGv7pg==
424K	.bloop/core/bloop-bsp-clients-classes/test-classes-Metals-ZU6uaWGiQ-eReOpykI5IaQ==
```

**To Reproduce**

1. Open project in `vim`
2. Close project
3. Repeat 1-2
4. Observe `.bloop/module/bloop-bsp-clients-classes`

**Expected behavior**

Classes reused and no compilation initiated on repeated project opens

**Installation**

- Operating system: Linux
- Vim version: 8.2 8020814
- coc.nvim version: 0.0.80-9fd90ad761
- [coc-metals](https://github.com/scalameta/coc-metals/issues/410) extension version: 1.0.7
- Metals version: 0.10.5

arbitrary-dev:
The same seems to be with `.metals/metal.h2.db`

tgodzik:
Thanks for reporting! Bloop directories are removed as soon as the Metals client disconnects, I can't reproduce any issues with that myself. Would it possible that client still runs after restart? It's an issue we've seen in VS Code. @ckipp01 are you able to reproduce with coc-metals ?

As for `.metals/metals.h2.db` we use it to save some information about the workspaces, but it should not grow too much beyond the initial size, does it seem to be growing too much?

ckipp01:
[...]

PR #3758, opened 2022-03-25
Reduce file watcher memory usage

pvid:
This PR replaces `FileTreeRepository` with `PathWatcher`. The former keeps a cached version of watched file trees while the latter does not. Since the implementation no longer hashes files, but looks only on modification time, we can only watch modifications without keeping any state.

This makes a huge difference especially on macOS, where in large projects the root of the project has to be watched. My motivation for this change is that from time to time, metals stopped working and threw OOM exceptions (I think I experienced something akin to https://github.com/scalameta/metals/issues/3020) and I had to delete the `.bloop` folder and restart metals. In a particular case, the cached paths took more than half of 4GB heap dedicated to metals.

Shoutout and huge thanks to @eatkins that fixed an issue https://github.com/swoval/swoval/issues/155 that I experienced while testing the change on Linux.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/TargetData.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/TargetData.scala
@@ -19,7 +19,8 @@ import ch.epfl.scala.bsp4j.BuildTargetIdentifier
 import ch.epfl.scala.bsp4j.JavacOptionsResult
 import ch.epfl.scala.bsp4j.ScalacOptionsResult
 import ch.epfl.scala.bsp4j.SourceItem
-import ch.epfl.scala.bsp4j.SourceItemKind
+import ch.epfl.scala.bsp4j.SourceItemKind.DIRECTORY
+import ch.epfl.scala.bsp4j.SourceItemKind.FILE
 import ch.epfl.scala.bsp4j.WorkspaceBuildTargetsResult
 import org.eclipse.{lsp4j => l}
 
@@ -49,6 +50,8 @@ final class TargetData {
     ConcurrentHashSet.empty[AbsolutePath]
   // if workspace contains symlinks, original source items are kept here and source items dealiased
   val originalSourceItems: util.Set[AbsolutePath] =
+    ConcurrentHashSet.empty[AbsolutePath]
+  val sourceItemFiles: util.Set[AbsolutePath] =
     ConcurrentHashSet.empty[AbsolutePath]
 
   val targetToConnection: MMap[BuildTargetIdentifier, BuildServerConnection] =
@@ -176,11 +179,16 @@ final class TargetData {
       buildTarget: BuildTargetIdentifier
   ): Unit = {
     val sourceItemPath = sourceItem.getUri.toAbsolutePath(followSymlink = false)
-    if (
-      sourceItem.getKind() == SourceItemKind.DIRECTORY &&
-      sourceItem.getGenerated()
-    ) {
-      buildTargetGeneratedDirs(sourceItemPath) = ()
+
+    sourceItem.getKind() match {
+      case DIRECTORY => {
+        if (sourceItem.getGenerated()) {
+          buildTargetGeneratedDirs(sourceItemPath) = ()
+        }
+      }
+      case FILE => {
+        sourceItemFiles.add(sourceItemPath)
+      }
     }
     addSourceItem(sourceItemPath, buildTarget)
   }
@@ -214,6 +222,10 @@ final class TargetData {
         buf += target.getId
       }
     }
+  }
+
+  def isSourceFile(source: AbsolutePath): Boolean = {
+    sourceItemFiles.contains(source)
   }
 
   def checkIfGeneratedSource(source: Path): Boolean = {
--- a/metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
@@ -1,8 +1,11 @@
 package scala.meta.internal.metals.watcher
 
-import java.io.IOException
 import java.nio.file.Path
+import java.util.concurrent.BlockingQueue
+import java.util.concurrent.LinkedBlockingQueue
+import java.util.concurrent.atomic.AtomicBoolean
 
+import scala.annotation.tailrec
 import scala.collection.mutable
 
 import scala.meta.internal.metals.BuildTargets
@@ -12,11 +15,11 @@ import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.MetalsServerConfig
 import scala.meta.io.AbsolutePath
 
-import com.swoval.files.FileTreeDataViews.CacheObserver
-import com.swoval.files.FileTreeDataViews.Converter
-import com.swoval.files.FileTreeDataViews.Entry
-import com.swoval.files.FileTreeRepositories
-import com.swoval.files.FileTreeRepository
+import com.swoval.files.FileTreeViews
+import com.swoval.files.FileTreeViews.Observer
+import com.swoval.files.PathWatcher
+import com.swoval.files.PathWatchers
+import com.swoval.files.PathWatchers.Event.Kind
 
 /**
  * Watch selected files and execute  a callback on file events.
@@ -42,30 +45,24 @@ final class FileWatcher(
   import FileWatcher._
 
   @volatile
-  private var disposeAction: Option[() => Unit] = None
+  private var stopWatcher: () => Unit = () => ()
 
   override def cancel(): Unit = {
-    disposeAction.map(_.apply())
-    disposeAction = None
+    stopWatcher()
   }
 
-  def restart(): Unit = {
-    disposeAction.map(_.apply())
-
-    val newDispose = startWatch(
+  def start(): Unit = {
+    stopWatcher = startWatch(
       config,
       workspaceDeferred().toNIO,
       collectFilesToWatch(buildTargets),
       onFileWatchEvent,
       watchFilter
     )
-    disposeAction = Some(newDispose)
   }
 }
 
 object FileWatcher {
-  type Hash = Long
-
   private case class FilesToWatch(
       sourceFiles: Set[Path],
       sourceDirectories: Set[Path],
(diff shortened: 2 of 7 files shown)


metals/src/main/scala/scala/meta/internal/metals/TargetData.scala
```diff
@@ -50,6 +51,8 @@ final class TargetData {
   // if workspace contains symlinks, original source items are kept here and source items dealiased
   val originalSourceItems: util.Set[AbsolutePath] =
     ConcurrentHashSet.empty[AbsolutePath]
+  val sourceItemFiles: util.Set[AbsolutePath] =
```
pvid, line 54: not too happy with creating another random set here. I was thinking of keeping some more information about source items, maybe in its original BSP form as `Map[AbsolutePath, SourceItem]` that could also be used for distinguishing generated dirs



This version has the advantage of being very straight forward and since the vast majority of source items are directories, the set is very small

what do you think?

eatkins:
@pvid I am wondering if I pushed an update to swoval that allowed the FileTreeRepository to filter out the paths you don't care about if that would be a simpler fix? If you want to try a branch that should handle the use case, I pushed it here: https://github.com/swoval/swoval/pull/158. I think the fix with the swoval update might be as simple as:
```diff
- val repo = FileTreeRepositories.get(converter, /*follow symlinks*/ true)
+ val repo = FileTreeRepositories.get(converter, tp => watchFilter(tp.getPath), /*follow symlinks*/ true)
```
but I haven't tried it myself.

pvid:
> @pvid I am wondering if I pushed an update to swoval that allowed the FileTreeRepository to filter out the paths you don't care about if that would be a simpler fix? If you want to try a branch that should handle the use case, I pushed it here: [swoval/swoval#158](https://github.com/swoval/swoval/pull/158). I think the fix with the swoval update might be as simple as:
> 
> ```diff
> - val repo = FileTreeRepositories.get(converter, /*follow symlinks*/ true)
> + val repo = FileTreeRepositories.get(converter, tp => watchFilter(tp.getPath), /*follow symlinks*/ true)
> ```
> 
> but I haven't tried it myself.

that would certainly help with the problem I described, but conceptually, there is still no reason to keep a cached representation given metals uses just the PathWatcher functionality

eatkins:
Fair enough. One drawback to PathWatcher is that it isn't very reliable in terms of correctly identifying what kind of event happened. For instance, it isn't possible to distinguish a file creation from a file modification on macOS. In order to get the event types right, it is necessary to cache the state of every file so the additional overhead of storing say a last modified time or hash is small. But it sounds like getting the event types right may not really matter for metals in which case PathWatcher is fine.

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
```diff
@@ -117,89 +117,140 @@ object FileWatcher {
...
+    // Upstream callbacks implicitly assume that events for a single file
+    // arrive and are processed serially.
+    // This consumption could be parallelized, but in case in indexing,
+    // compilation is usually the dominating bottleneck
+    repeatUntilStopped(
+      stopWatchingSignal, {
+        Future {
+          blocking {
+            // do not block indefinitely to allow the loop to be stopped by stop signal
+            Option(watchEventQueue.poll(200, TimeUnit.MILLISECONDS))
```
eatkins, line 182: It isn't maybe idiomatic scala, but the way I like to do something like this is:
```scala
val thread = new Thread("metals-watch-callback-thread") {
  @tailrec def loop(): Unit = {
    try callback(watchEventQueue.take)
    catch { case _: InterruptedException =>  }
    if (!stopWatchingSignal.get) loop()
  }
  override def run(): Unit = loop()
}
...
() => {
  watcher.close()
  stopWatchingSignal.set(true)
  thread.interrupt()
}
```

It's a minor thing but with this technique the watch thread stays idle until there really is work to do and it will also shutdown essentially instantaneously instead of on average taking 100ms.

pvid:
> Fair enough. One drawback to PathWatcher is that it isn't very reliable in terms of correctly identifying what kind of event happened. For instance, it isn't possible to distinguish a file creation from a file modification on macOS. In order to get the event types right, it is necessary to cache the state of every file so the additional overhead of storing say a last modified time or hash is small. But it sounds like getting the event types right may not really matter for metals in which case PathWatcher is fine.

that is a good point with the type of events. Since metals does the same thing for CREATE and MODIFY, I will coalesce them into a single event type here with a comment about the shortcomings of `PathWatcher` and that `FileTreeRepositories` should be used when that distinction is important.

I will leave it to the maintainers to decide if they want to stick with `FileTreeRepositories` or not. Thank you for your input! I really does better illuminate the pros and cons of both

eatkins:
Happy to help. I just want to add for any reviewers that internally a FileTreeRepository delegates to a PathWatcher so this should be a safe change provided that the distinction between creation and modification events is not important.

metals/src/main/scala/scala/meta/internal/metals/TargetData.scala
tanishiking: > I was thinking of keeping some more information about source items, maybe in its original BSP form as Map[AbsolutePath, SourceItem] that could also be used for distinguishing generated dirs

I was considering adding those additional data (like `SourceItem`) to `buildTargetSources` :)
https://github.com/scalameta/metals/blob/9fe87ba4bd2abaf41a352e4cb7146df45535b78d/metals/src/main/scala/scala/meta/internal/metals/TargetData.scala#L41-L42

While creating another random set here wouldn't be a big problem, can we stay `path.isScalaOrJava` here https://github.com/scalameta/metals/pull/3758/files#diff-314e34520d936536cd836f653612edc262cf2afb106ae5072025978256a1fa8cL83 ? As the argument of the `collect` method is from `buildTarget.sourceItems`, they are obviously a source item of the build target.

Or can we use `buildTargetSources`? (By somehow giving a pair of `AbsolutePath` and `BuildTargetIdentifier` to the `collect` method).

tanishiking (approved):
I haven't yet tested this branch locally, but the change looks on the right track, great work! I left some comments.

> provided that the distinction between creation and modification events is not important

In terms of a distinction between creation and modification, it should be safe in Metals as we invoke the same callback method for both creation and modification.
https://github.com/scalameta/metals/blob/9fe87ba4bd2abaf41a352e4cb7146df45535b78d/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1375-L1378

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
tanishiking: I prefer this version :) `watchEventQueue.take` will block until there's an event in the queue and the watch thread stays idle 😍

tanishiking (commented):
Sorry, I approved by mistake (while it looks good:+1: I want to test it locally), I was just commenting

kpodsiad:
> This makes a huge difference especially on macOS, where in large projects the root of the project has to be watched. [...] In a particular case, the cached paths took more than half of 4GB heap dedicated to metals.

@pvid is this >2GB of cached paths after https://github.com/scalameta/metals/pull/3665 or before?

I'm good with using `PathWatcher`. If at some point of the time it would be necessary to distinguish between event types we could always rollback to the `FileTreeRepository`.

pvid:
> > This makes a huge difference especially on macOS, where in large projects the root of the project has to be watched. [...] In a particular case, the cached paths took more than half of 4GB heap dedicated to metals.
> 
> @pvid is this >2GB of cached paths after #3665 or before?
> 
> I'm good with using `PathWatcher`. If at some point of the time it would be necessary to distinguish between event types we could always rollback to the `FileTreeRepository`.

that PR is useful for monorepos with just a few scala modules (see default limit of 32 roots). The project I am working on has > 100 modulesand it would require more roots than macOS file watching is able to handle.

metals/src/main/scala/scala/meta/internal/metals/TargetData.scala
pvid: The reason for going from `path.isScalaOrJava` to getting info from BSP is that some tests were failing, because of a `*.sbt` source file file that failed that check. However, it still needs to be watched as a file.

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
pvid: changed in a fixup commit

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
```diff
@@ -117,89 +117,140 @@ object FileWatcher {
...
+          pathsToWatch.files ++ pathsToWatch.directories
+        )
+        val isWatched = trie.containsPrefixOf _
+
+        // Select up to `maxRoots` longest prefixes of all files in the trie for
+        // watching. Watching the root of the workspace may have bad performance
+        // implications if it contains many other projects that we don't need to
+        // watch (eg. in a monorepo)
+        val watchRoots =
+          trie.longestPrefixes(workspace.getRoot(), 1)
```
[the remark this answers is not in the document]
pvid: my mistake, I changed the value for some testing. Fixed in a fixup commit

tanishiking (approved):
LGTM from my side, do you have any additional comments? @kpodsiad

pvid:
I have squashed the fixup commits resulting from the review

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
```diff
@@ -42,62 +44,56 @@ final class FileWatcher(
...
+    val files = mutable.Set.empty[Path]
 
     def collect(path: AbsolutePath): Unit = {
       if (buildTargets.isInsideSourceRoot(path)) {
         () // Do nothing, already covered by a source root
       } else if (!buildTargets.checkIfGeneratedSource(path.toNIO)) {
-        if (path.isScalaOrJava) {
-          sourceFilesToWatch.add(path.toNIO)
+        if (buildTargets.isSourceFile(path)) {
+          files.add(path.toNIO)
```
kpodsiad, line 79: I have some proposition of refactor for `collect`
```scala
def collect(path: AbsolutePath): Unit = {
  val shouldBeWatched =
    // Do nothing, already covered by a source root
    !buildTargets.isInsideSourceRoot(path) &&
      !buildTargets.checkIfGeneratedSource(path.toNIO)

  if (shouldBeWatched) {
    if (buildTargets.isSourceFile(path)) files.add(path.toNIO)
    else directories.add(path.toNIO)
  }
}
```

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
```diff
@@ -117,89 +113,126 @@ object FileWatcher {
...
-      override def onDelete(entry: Entry[Hash]): Unit = {
-        val path = entry.getTypedPath().getPath()
-        if (watchFilter(path)) callback(FileWatcherEvent.delete(path))
+      queue: BlockingQueue[FileWatcherEvent]
+  ): PathWatcher[PathWatchers.Event] = {
+    val watcher = PathWatchers.get( /*follow symlinks*/ true)
+
+    watcher.addObserver(new Observer[PathWatchers.Event] {
+      override def onError(t: Throwable): Unit = {
+        scribe.error(s"Error encountered during file watching: ${t}")
```
kpodsiad, line 209: We should also print the stack trace of the encountered error.

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcherEvent.scala
```diff
@@ -22,4 +23,6 @@ object FileWatcherEvent {
     FileWatcherEvent(EventType.Modify, path)
   def delete(path: Path): FileWatcherEvent =
     FileWatcherEvent(EventType.Delete, path)
+  def overflow(path: Path): FileWatcherEvent =
+    FileWatcherEvent(EventType.Overflow, path)
```
kpodsiad, line 27: What exactly indicates `EventType.Overflow`? I think it'll be worth adding some comment next to the `case object Overflow extends EventType`

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
```diff
@@ -117,89 +113,126 @@ object FileWatcher {
...
-        if (previous.getValue != current.getValue && watchFilter(path)) {
-          callback(FileWatcherEvent.modify(path))
+
+      override def onNext(event: PathWatchers.Event): Unit = {
+        val path = event.getTypedPath.getPath
+        if (watchFilter(path)) {
+          event.getKind match {
+            case Kind.Create => {
+              queue.add(FileWatcherEvent.create(path))
+            }
```
kpodsiad, line 218: Could you remove the brackets surrounding the cases' bodies?

kpodsiad (commented):
Looks good, I have only a few suggestions. 
Also, I'm going to test this change manually today.

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcherEvent.scala
```diff
@@ -14,6 +14,7 @@ object FileWatcherEvent {
     case object Create extends EventType
```
pvid, line 14: TODO merge Create and Modify events, see https://github.com/scalameta/metals/pull/3758#issuecomment-1080005296

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcherEvent.scala
pvid: Overflow indicates that the file watcher was overwhelmed and may have lost some events . I will add a comment

metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala
pvid: > Looks good, I have only a few suggestions. Also, I'm going to test this change manually today.

Thanks! I will address your comments this evening. I also noticed that I forgot to address one issue, see https://github.com/scalameta/metals/pull/3758#discussion_r841469303

kpodsiad:
@pvid Will it be a problem if this PR got merged after a Metals release (which may happen this week)? 
This PR touches a Metals core so it would be safer to merge this after the stable version is released. Thanks to that we will have more time to battle-test this until the next stable release. 
So, will it be ok for you and your colleagues at work to use the snapshot version with these changes?

tgodzik:
I need to still look into some possible issues with Bloop before the next release, so I would rather postpone it if possible.

kpodsiad:
Don't worry @tgodzik, with @dos65 we are aware of some of these issues and there won't be any release without resolving them first. Enjoy the rest of your holidays, I'm not cutting a release without you knowing about it :D

pvid:
@kpodsiad sure, no problem

pvid:
I have rebased the changes onto current main and addressed all the comments and suggestions from review. It should be ready for merge after the next release as noted above.

I will be using a SNAPSHOT version of metals with these changes to make sure that some unforeseen issues do not arise in day-to-day use

tanishiking:
Now the latest version is out, it's time to merge 🚀

PR #3759, opened 2022-03-26
chore: reduce size of build.sbt

kpodsiad:
I find large `build.sbt` files terryfying.


tgodzik (changes requested):
I wouldn't change it, it's much harder to find things sbt definition is split into multiple files and this is not really that large.

kpodsiad:
Ok, `crossSetting` may stay in `build.sbt` but

> it's much harder to find things sbt definition

can't agree with versions because it's a dedicated file and one don't have to scroll through >700loc file to change the bloop version.

project/Helpers.scala
```diff
@@ -0,0 +1,36 @@
+import sbt._
+import scala.sys.process._
+import Developers._
+import Tests._
+import Helpers._
```
dos65, line 5: seem like it's unused

dos65 (approved):
I don't mind of moving these part into separate file.
Extracting versions/dependencies into a separate file is a common approach that is used in sbt.
Also I see that some unused settings were removed.

I wouldn't say that I had any problems with current build but I totally sure that these changes are positive ones.

kpodsiad:
@tgodzik @ckipp01 what do you think about meeting halfway? I'll extract versions and leave the rest in the `build.sbt`.

tgodzik:
> @tgodzik @ckipp01 what do you think about meeting halfway? I'll extract versions and leave the rest in the `build.sbt`.

I actually prefer the versions in build.sbt because the go to definition doesn't seem to work perfectly between files, but that might be a metals problem. Also, are we sure that Scala Steward will manage to update the versions there?

I would be ok with extracting the helper methods

tgodzik (commented):
If you feel strongly about it @kpodsiad let's go the middle ground.

tgodzik (approved):

PR #3760, opened 2022-03-26
chore: enable `-Wunused`

kpodsiad:
Turned out we had a lot of unused warnings.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,6 +3,8 @@ import scala.sys.process._
 import Developers._
 import Tests._
 
+Global / onChangedBuildSource := ReloadOnSourceChanges
+
 def localSnapshotVersion = "0.11.3-SNAPSHOT"
 def isCI = System.getenv("CI") != null
 
@@ -15,24 +17,17 @@ def isScala3(v: Option[(Long, Long)]): Boolean = v.exists(_._1 == 3)
 def crossSetting[A](
     scalaVersion: String,
     if211: List[A] = Nil,
-    if2: List[A] = Nil,
-    ifLaterThan211: List[A] = Nil,
-    if3: List[A] = Nil
+    if213: List[A] = Nil,
+    if3: List[A] = Nil,
+    if2: List[A] = Nil
 ): List[A] =
   CrossVersion.partialVersion(scalaVersion) match {
     case partialVersion if isScala211(partialVersion) => if211 ::: if2
-    case partialVersion if isScala212(partialVersion) => ifLaterThan211 ::: if2
-    case partialVersion if isScala213(partialVersion) => ifLaterThan211 ::: if2
+    case partialVersion if isScala212(partialVersion) => if2
+    case partialVersion if isScala213(partialVersion) => if2 ::: if213
     case partialVersion if isScala3(partialVersion) => if3
     case _ => Nil
   }
-
-// -Xlint is unusable because of
-// https://github.com/scala/bug/issues/10448
-val scala212CompilerOptions = List(
-  "-Ywarn-unused:imports", "-Ywarn-unused:privates", "-Ywarn-unused:locals",
-  "-Ywarn-unused:patvars", "-Ywarn-unused:implicits"
-)
 
 logo := Welcome.logo
 usefulTasks := Welcome.tasks
@@ -50,7 +45,7 @@ inThisBuild(
     scalacOptions ++= List(
       "-target:jvm-1.8",
       "-Yrangepos"
-    ) ::: scala212CompilerOptions,
+    ),
     scalafixDependencies += "com.github.liancheng" %% "organize-imports" % V.organizeImportRule,
     organization := "org.scalameta",
     licenses := Seq(
@@ -276,6 +271,24 @@ lazy val V = new {
     ).toList
 }
 
+// -Xlint is unusable because of
+// https://github.com/scala/bug/issues/10448
+def lintingOptions(scalaVersion: String) = {
+  val unused213 = "-Wunused"
+  val unused3 = "-Wunused:all"
+  val common = List(
+    // desugaring of for yield caused pattern var to complain
+    // https://github.com/scala/bug/issues/10287
+    "-Wconf:msg=parameter value .+ in anonymous function:silent"
+  )
+  crossSetting(
+    scalaVersion,
+    if213 = unused213 :: common,
+    if3 = unused3 :: common,
+    if211 = List("-Ywarn-unused-import")
+  )
+}
+
 val sharedSettings = List(
   libraryDependencies ++= crossSetting(
     scalaVersion.value,
@@ -289,16 +302,15 @@ val sharedSettings = List(
     scalaVersion.value,
     if3 = List(
       "-language:implicitConversions",
-      "-Xtarget:8",
-      "-Xsemanticdb"
+      "-Xtarget:8"
     ),
-    if211 = List("-Xexperimental", "-Ywarn-unused-import")
+    if211 = List("-Xexperimental")
   ),
   scalacOptions --= crossSetting(
     scalaVersion.value,
-    if3 = "-Yrangepos" :: "-target:jvm-1.8" :: scala212CompilerOptions,
-    if211 = scala212CompilerOptions
-  )
+    if3 = List("-Yrangepos", "-target:jvm-1.8")
+  ),
+  scalacOptions ++= lintingOptions(scalaVersion.value)
 )
 
 publish / skip := true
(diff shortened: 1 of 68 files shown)


build.sbt
```diff
@@ -276,6 +271,35 @@ lazy val V = new {
...
+    // desugaring of for yield caused pattern var to complain
+    // https://github.com/scala/bug/issues/10287
+    "-Wconf:msg=parameter value .+ in anonymous function:silent",
+    // silence unused parameters in mtags
+    "-Wconf:src=*.ScaladocParser.scala&msg=in method reportError:silent",
+    "-Wconf:src=*.Completions.scala&msg=in method (isCandidate|isPrioritized):silent",
+    "-Wconf:src=*.JavaMtags.scala&msg=in method (visitConstructor|visitMethod):silent",
+    "-Wconf:src=*.MtagsIndexer.scala&msg=in method visitOccurrence:silent",
+    // silence "The outer reference in this type test cannot be checked at run time." in mtags
+    "-Wconf:src=.*(CompletionProvider|ArgCompletions|Completions|Keywords).scala&msg=The outer reference:silent"
```
kpodsiad, line 289: I need to dive into this warning deeper, but for now, I silenced it to have 0 warnings in the problems tab.

build.sbt
```diff
@@ -276,6 +271,35 @@ lazy val V = new {
     ).toList
 }
 
+// -Xlint is unusable because of
+// https://github.com/scala/bug/issues/10448
+def lintingOptions(scalaVersion: String) = {
+  val unused213 = "-Wunused"
+  val unused3 = "-Wunused:all"
```
kpodsiad, line 278: I think it's better to enable all of them and them eventually filter out noisy ones via `Wconf`

build.sbt
```diff
@@ -276,6 +271,35 @@ lazy val V = new {
...
+  val common = List(
+    // desugaring of for yield caused pattern var to complain
+    // https://github.com/scala/bug/issues/10287
+    "-Wconf:msg=parameter value .+ in anonymous function:silent",
+    // silence unused parameters in mtags
```
ckipp01, line 283: Maybe a silly question, but are they false positives? Or what not remove the params if they are unused instead of silencing them?

build.sbt
```diff
@@ -276,6 +271,35 @@ lazy val V = new {
...
+    "-Wconf:src=*.ScaladocParser.scala&msg=in method reportError:silent",
```
ckipp01, line 284: This seems pretty broad, or is it just me? Like are there more messages that the target one that include `in method` that we might be silencing that we don't want to?

metals/src/main/scala/scala/meta/internal/metals/codelenses/CodeLens.scala
```diff
@@ -1,13 +1,19 @@
...
+
 import scala.meta.internal.implementation.TextDocumentWithPath
 import scala.meta.io.AbsolutePath
 
 import org.eclipse.{lsp4j => l}
 
 trait CodeLens {
   def isEnabled: Boolean
+
+  @nowarn
```
ckipp01, line 13: Can we specify what we are silencing with these?

metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
```diff
@@ -146,6 +147,7 @@ case class IndentOnPaste(userConfig: () => UserConfiguration)
       case (line, _) =>
         PastedLine.plainOrEmpty(normalizeSpacesAndTabs(line, opts))
     }
+    @nowarn("msg=The outer reference")
```
ckipp01, line 150: So aren't we silencing this same warning in other places in the `build.sbt`, but specifically here as well? I find it a bit confusing if we silence it in some packages in the `build.sbt` and then in this file here. Could we had this place in there as well?

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
```diff
@@ -103,6 +101,9 @@ final class Trees(
           case Parsed.Success(tree) =>
             trees(path) = tree
             List()
+          case _ =>
+            scribe.info("Received unexpected parser result")
```
ckipp01, line 105: Will this happen often? Should this be info or warn?

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -469,6 +467,7 @@ object WorksheetProvider {
   }
   final case class MdocRef(scalaVersion: String, value: Mdoc)
 
+  // TODO
```
ckipp01, line 470: Leftover TODO?

tests/cross/src/main/scala/tests/BasePCSuite.scala
```diff
@@ -107,9 +108,12 @@ abstract class BasePCSuite extends BaseSuite {
       snippetAutoIndent = false
     )
 
-  protected def extraDependencies(scalaVersion: String): Seq[Dependency] = Nil
+  @nowarn
```
ckipp01, line 111: Can we specify what we are silencing with these two in here?

ckipp01 (commented):
Nice to see all the cleanup! Good job on this! Couple questions/suggestions.

build.sbt
kpodsiad: Yes, they're false positives. It's because:
```scala
  for {
    value <- opt
    if value
    result <- computeSth()
  } yield result
```
is being desugared more or less to the:
```scala
  opt
    .withFilter(value => value)
    .flatMap(value =>
      computeSth()
        .map(result => result)
    )
```
and it is true that `value` in flatmap is unused.

I could rewrite them but it's an inadequate amount of work compared to the value it brings.

ckipp01: Gotcha, thanks for the explanation.

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
kpodsiad: Oh, I forgot about it.
Yeah, it's for @tgodzik because he introduced these methods :D

metals/src/main/scala/scala/meta/internal/metals/formatting/IndentOnPaste.scala
kpodsiad: Good catch!
I started with `@nowarns` in files but it yielded a compilation error in mtags since `nowarn` was introduced in some specific version of scala and this one slipped through. 
I'll move it to the `build.sbt` as others.

build.sbt
kpodsiad: Sorry, could you rephrase the second sentence?

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
kpodsiad: I changed severity to the `error` and I logged the class's simple name. Maybe it's too cautious?

build.sbt
ckipp01: Ha, yea sorry. I mean this will ignore any message that has `in method` in it right? If so, are there potentially multiple warnings that might have `in method` in it that we are ignoring when we specifically only want to ignore one? Just trying to make sure this isn't too broad.

kpodsiad: IIRC every message that comes from `ScaladocParser.scala` file and has `in method reportError` so it's a bit moe narrow. But I think I can change it to the `parameter (name1|name2) in method reportError` and this should be very precise.

metals/src/main/scala/scala/meta/internal/parsing/Trees.scala
```diff
@@ -102,9 +100,14 @@ final class Trees(
             )
           case Parsed.Success(tree) =>
             trees(path) = tree
-            List()
+            List.empty
+          case other =>
+            scribe.error(
```
tgodzik, line 105: I don;t think we need to log it since it's not possible currently to get any other result.

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
tgodzik: Wait, but what would I need to do?

kpodsiad: what is the difference between both `worksheetScala3Adjustments` if they both ignore the second parameter? Why one is adding +1 to the position?

tgodzik: The second parameter looks like it can be removed. The second method has the additiona Position mapping. It's +1 because we wrap worksheet input in an object, since scripts do not yet work for Scala 3

kpodsiad:
@ckipp01 @dos65 suggestions applied, CI is green, the number of warnings is equal to zero.

ckipp01 (approved):
🚢 it.

kpodsiad:
Fatal warnings for `2.13.8` on the CI, will it be useful or troublesome?

PR #3763, opened 2022-03-27
add navigation from doctor to build target

Arthurm1:
image

Adds html links to build target info from the Doctor as in image above.

Does the command to navigate to a uri have to include location?  Or is there another command instead of `metals-goto-location`?  Ideally it would just open the URI.

@ckipp01 Is the addition of `gotoCommand` to the Doctor JSON the best way to handle non vscode clients?

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
@@ -21,6 +21,7 @@ import scala.meta.internal.troubleshoot.ProblemResolver
 import scala.meta.io.AbsolutePath
 
 import ch.epfl.scala.bsp4j.BuildTargetIdentifier
+import org.eclipse.{lsp4j => l}
 
 /**
  * Helps the user figure out what is mis-configured in the build through the "Run doctor" command.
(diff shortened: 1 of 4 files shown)


kpodsiad (approved):
Looks good from vscode's perspective. 

About your question - I don't think that such a command exists, I've never met it and now I didn't found any.

ckipp01:
> @ckipp01 Is the addition of gotoCommand to the Doctor JSON the best way to handle non vscode clients?

Yea that should be fine as I'm not sure what else is better. For now, I'd just leave it as is and when I implement it I'll see if there is anything else that might need to change, but at first glance it should work. Thanks for checking!

Arthurm1:
@kpodsiad What's the etiquette on merge - can I merge this now with 1 approval?  Or do I leave it for one of you to merge?

kpodsiad:
@Arthurm1 I'll say that my approval weights 1/2 of approvals of @tgodzik or @ckipp01 ones :D

All things considered, this is a small PR so my approval should be ok :D And because you're a member of the scalameta feel free to merge it!

ckipp01:
> @kpodsiad What's the etiquette on merge - can I merge this now with 1 approval? Or do I leave it for one of you to merge?

It's sort of at your discretion. If it's something smaller like this, then feel free to merge after a ✅ . If it's something you'd like an extra set of eyes on, then feel free to wait. Although if there is already a ✅ and you want another, best to request a review from specific people on it.

tgodzik:
> @Arthurm1 I'll say that my approval weights 1/2 of approvals of @tgodzik or @ckipp01 ones :D

I would say it's totally not true :wink: You do great reviews and have already a lot of experience with the codebase.

PR #3765, opened 2022-03-28
Try to fix flaky test by adding a delay in compilation

tanishiking:
Follow up for https://github.com/scalameta/metals/pull/3752

We encountered the flaky test result of `DidFocusLspSuite` in the PR, because
we expect the test suite invokes `didFocus("b/.../B.scala")`
**during the compilation by `server.didSave("a/.../A.scala")`**, and
didFocus triggers a compilation.

However, we didn't have a way to control the timing of compilation
start, end, and when we invoke `didFocus`. That's why the test was flaky.
see more details: https://github.com/scalameta/metals/pull/3752#issuecomment-1079878023

This PR ensure that we call `server.didFocus("b/.../B.scala")` during
the compilation by `server.didSave("a/.../A.scala")` by

- Adding 5 seconds of delay during the compilation
  - to make sure we invoke `didFocus` before the compilation completed.
- Wait for 2.5 seconds before invoking `didFocus`
  - to make sure we invoke `didFocus` after the compilation started.

### DIFF
--- a/tests/unit/src/test/scala/tests/DidFocusLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/DidFocusLspSuite.scala
@@ -89,7 +89,9 @@ class DidFocusWhileCompilingLspSuite
     super.beforeEach(context)
   }
 
-  test("497") {
+  test(
+    "Trigger compilation by didFocus when current compile may affect focused buffer"
+  ) {
     cleanWorkspace()
     for {
       _ <- initialize(
--- a/tests/unit/src/test/scala/tests/DidFocusLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/DidFocusLspSuite.scala
@@ -1,7 +1,5 @@
 package tests
 
-import scala.concurrent.Future
-
 import scala.meta.internal.metals.DidFocusResult._
 import scala.meta.internal.metals.Time
 
@@ -79,8 +77,8 @@ class DidFocusWhileCompilingLspSuite
   val compileDelayMillis = 5000
   override def time: Time = fakeTime
 
-  // sleep 10s during the compilation, so that we can make sure
-  // invoking `didFocus` during compilation.
+  // sleep 5s during the compilation, so that we can make sure
+  // calling `didFocus` during compilation.
   override def beforeEach(context: BeforeEach): Unit = {
     fakeTime = new FakeTime()
     onStartCompilation = () => {
@@ -145,7 +143,7 @@ class DidFocusWhileCompilingLspSuite
         _.replace("Int", "String")
       )
       // Wait until compilation against project a is started (before we invoke didFocus on project b)
-      _ <- Future { Thread.sleep(compileDelayMillis / 2) }
+      _ <- server.waitFor(compileDelayMillis / 2)
       // Focus before compilation of A.scala is complete.
       // And make sure didFocus during the compilation causes compilation against project b.
       didCompile <- server.didFocus("b/src/main/scala/b/B.scala")
@@ -165,5 +163,4 @@ class DidFocusWhileCompilingLspSuite
       )
     } yield ()
   }
-
 }
--- a/tests/unit/src/test/scala/tests/DidFocusLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/DidFocusLspSuite.scala
@@ -1,5 +1,7 @@
 package tests
 
+import scala.concurrent.Future
+
 import scala.meta.internal.metals.DidFocusResult._
 import scala.meta.internal.metals.Time
 
@@ -10,6 +12,7 @@ class DidFocusLspSuite extends BaseLspSuite("did-focus") {
     fakeTime = new FakeTime()
     super.beforeEach(context)
   }
+
   test("is-compiled") {
     cleanWorkspace()
     for {
@@ -67,9 +70,26 @@ class DidFocusLspSuite extends BaseLspSuite("did-focus") {
       )
     } yield ()
   }
+}
 
-  // Ignore flaky test, see the details: https://github.com/scalameta/metals/pull/3752#issuecomment-1079878023
-  test("497".ignore) {
+// https://github.com/scalameta/metals/issues/497
+class DidFocusWhileCompilingLspSuite
+    extends BaseLspSuite("did-focus-while-compiling") {
+  var fakeTime: FakeTime = _
+  val compileDelayMillis = 5000
+  override def time: Time = fakeTime
+
+  // sleep 10s during the compilation, so that we can make sure
+  // invoking `didFocus` during compilation.
+  override def beforeEach(context: BeforeEach): Unit = {
+    fakeTime = new FakeTime()
+    onStartCompilation = () => {
+      Thread.sleep(compileDelayMillis)
+    }
+    super.beforeEach(context)
+  }
+
+  test("497") {
     cleanWorkspace()
     for {
       _ <- initialize(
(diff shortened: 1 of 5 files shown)


tests/unit/src/test/scala/tests/DidFocusLspSuite.scala
```diff
@@ -122,10 +144,16 @@ class DidFocusLspSuite extends BaseLspSuite("did-focus") {
       didSaveA = server.didSave("a/src/main/scala/a/A.scala")(
         _.replace("Int", "String")
       )
+      // Wait until compilation against project a is started (before we invoke didFocus on project b)
+      _ <- Future { Thread.sleep(compileDelayMillis / 2) }
```
ckipp01, line 148: I believe as an alternative we could also use
```
      _ <- server.waitFor(TimeUnit.SECONDS.toMillis(10)) // whatever time you want
```

Would just adding something like that in be enough so we don't have to pollute `Compilations` with `onStartCompilation`? It's sort of a bummer that we need to add that justs for tests.

kpodsiad: After seeing `TimeUnit.SECONDS.toMillis(10)` I wondered like ~5seconds if it's 10s transformed into millis or 10millis transformed to seconds :P
I suggest changing signature to `def waitFor(duration: FiniteDuration): Future[Unit] = Future { Thread.sleep(duration.toMillis) }`

tests/unit/src/test/scala/tests/DidFocusLspSuite.scala
```diff
@@ -67,9 +70,28 @@ class DidFocusLspSuite extends BaseLspSuite("did-focus") {
...
+    onStartCompilation = () => {
+      Thread.sleep(compileDelayMillis)
+    }
+    super.beforeEach(context)
+  }
 
-  // Ignore flaky test, see the details: https://github.com/scalameta/metals/pull/3752#issuecomment-1079878023
-  test("497".ignore) {
+  test(
+    "Trigger compilation by didFocus when current compile may affect focused buffer"
```
kpodsiad, line 93: I think we have the naming convention to use a shorter name written in kebab-case. Current name could be extracted as a comment I think

kpodsiad (approved):
Maybe sleeping it's not great but I have no better idea so 👍🏻 👍🏻  from me.

tests/unit/src/test/scala/tests/DidFocusLspSuite.scala
tanishiking: > Would just adding something like that be enough so we don't have to pollute Compilations with onStartCompilation? It's sort of a bummer that we need to add that just for tests.

Yeah, it's a bummer... but I don't think just adding a `waitFor` in the test fixes the problem.

### (1) Expected scenario
The diagram below is the expected test scenario:
- `didSave(A.scala)` start compilation
- and then `didSave(B.scala)` 
- and then compilation by `didSave(A.scala)` ends

```mermaid
sequenceDiagram
  participant Client
  participant Server

  Client->>Server: didSave(A.scala)
  rect rgb(191, 223, 255)
    Server->>Server: Start compile (by didSave)
    rect rgb(200, 150, 255)
      Client->>Server: didFocus
      Server->>Server: Compile by didFocus
    end
    Server->>Server: Finish compile (by didSave)
  end
```

### (2) didFocus too early

This (↓) is what we were:

- We invoke `didSave(A.scala)`
- but we call `didFocus(B.scala)` before the compilation (by didSave) starts
- then compilation by didSave(A.scala) starts and end.

```mermaid
sequenceDiagram
  participant Client
  participant Server

  Client->>Server: didSave(A.scala)
  rect rgb(200, 150, 255)
    Client->>Server: didFocus
    Server->>Server: Compile by didFocus
  end
  rect rgb(191, 223, 255)
    Server->>Server: Start compile (by didSave)

    Server->>Server: Finish compile (by didSave)
  end
```

We can solve this problem by adding a sleep before calling `didFocus` in the test.

### (3) didFocus too late

However, if we delay only before the `didFocus`, another issue occurs:

- didSave(A.scala)
- we sleep for a sec (before calling `didSave`)
- compilation by didSave starts and ends
- Then we call `didFocus`

```mermaid
sequenceDiagram
  participant Client
  participant Server

  Client->>Server: didSave(A.scala)
  rect rgb(191, 223, 255)
    Server->>Server: Start compile (by didSave)
    Server->>Server: Finish compile (by didSave)
  end
    rect rgb(200, 150, 255)
      Client->>Server: didFocus
      Server->>Server: Compile by didFocus
    end
```

In this scenario, the test passes :+1:. However, this is not what this test suite wants to test: according to the original PR https://github.com/scalameta/metals/pull/500, we try to test here: Metals triggers compilation by `didFocus` when **current compile** is going on (and it affects the focused buffer).
That's why we need a delay in compilation to make sure we call `didFocus` **during** ongoing compilation.

---

Maybe we can compromise not to test the "(1) expected scenario" and test (3). (Because I agree it's a bummer to add a hook in Compilation just for test).
On the other hand, maybe we can use this hook for other purposes in the future (not a test thing, I'm not sure... 🤔 )

ckipp01: 😆 thank you for the incredibly detailed examples. How do you like mermaid? Do you also have experience with plantuml? Any thoughts on the differences if you do?

Regarding the actual change, no worries. I think what you have is minimal enough it's not an issue.

dos65 (commented):
@tanishiking after reading [your explation](https://github.com/scalameta/metals/pull/3752#issuecomment-1079878023) I'm wondering maybe there is an issue with logic.

> However, isAffectedByCurrentCompilation is false in the test because compilations.currentlyCompiling is empty in the test. That's why we receive AlreadyCompiled here. 

What about another check - `isAffectedByLastCompilation`. Shouldn't it return `true`?
Technically we tried to compile `B.scala` by previous `didSave`.

tests/unit/src/test/scala/tests/DidFocusLspSuite.scala
tanishiking: Yeah, I've been a big fan of plantuml, and quite new to mermaid :)
I couldn't find a big difference between plantuml and mermaid in terms of their functionalities, but I prefer mermaid rendered diagrams (just because their shape and color), and I found we can freely customize the styles by CSS and adding colored blocks by `rgb` is quite useful. 
On top of that,  being able to draw a diagram with "mermaid```...```" is pretty handy :+1:

> Regarding the actual change, no worries. I think what you have is minimal enough it's not an issue.

Thanks!

tanishiking:
@dos65 

> What about another check - isAffectedByLastCompilation. Shouldn't it return true?
> Technically we tried to compile B.scala by previous didSave.

No, it (`isAffectedbyLastCompilation`) will be `false` here, because it is defined as follows

https://github.com/scalameta/metals/blob/d526fa5af7fc2d9f0850c18f61c871b802522747/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1136-L1141

and `!compilations.wasPreviouslyCompiled(target)` will be `false` because `project b` was compiled in the previous compilation.
Hm, right, something is strange indeed 🤔  So `isAffectedByLastCompilation` means more like "is affected by last compilation but it's not yet compiled somehow" I guess.

---

For example, if we have projects

```scala
lazy val proj1 = (project in file("proj1"))
lazy val proj2 = (project in file("proj2")).dependsOn(proj1)
```

- compile `proj1`, and `proj2` wasn't compiled somehow (I'm not sure how it happens 🤔 )
- then `didFocus` on a source in `proj2`
  - `!compilations.wasPreviouslyCompiled(target)` (target = proj2) will be `true` because it wasn't compiled.
  - `isInverseDependency(proj2, List(proj1))` will be `true`
- In this case, yes, `proj2` should be compiled. Because it might be affected by the change in `proj1` (previous compile) but proje2 is not yet compiled.

kpodsiad:
> In this case, yes, proj2 should be compiled. Because it might be affected by the change in proj1 (previous compile) but proje2 is not yet compiled.

@tanishiking This is intentional behavior, in large projects, such compilation can trigger multiple compilations of dependent modules (like 20 or 50 even). Because of this "feature" `Cascade compile` command exists.
At least this is what I remember from what Tomek said to me.

kpodsiad:
I wonder if we could add a switch for that behavior. In smaller projects (like Metals for example) it's not a big deal and sometimes it leads to situations where you fix errors in module A but module B still has the errors because it wasn't compiled after A was fixed.

tanishiking:
I'll merge this PR after CI passes (later!) as the latest CIs started failing.

Issue #917, opened 2019-09-12
Don't autocomplete symbols that have been renamed/removed


jvican:
**Describe the bug**

Metals PC shows references to symbols that were renamed.

**To Reproduce**
1. Go to a file `Foo.scala` defining `Foo`
2. Rename `Foo` to `Foo2`
3. Go to file `UseSite.scala` and try to complete `Foo`.
4. `Foo` will be completed while it shouldn't.

**Expected behavior**

Metals PC should detect when symbols have been renamed and remove them from completions. A way to do this would be to modify the PC to check that the class file associated with a symbol does indeed exist. If it doesn't exist, it means it was removed and should be filtered out.

**Screenshots**

**Installation:**
- Operating system: macOS/Windows/Linux
- Editor: Visual Studio Code/Atom/Vim/Sublime/Emacs
- Metals version: v0.7.5

**Additional context**

**Search terms**
rename,symbol,pc,invalid

olafurpg:
Thank you for reporting! This is a known limitation since we currently rely on the classfiles from disk to provide the state of public signatures in the workspace. I see two possible approaches that could work to address this issue:

* keep up-to-date symbol table in-memory that we feed to the presentation compiler, this could maybe be done with the new -Youtline compiler option in 2.12.10
* use "sourcepath" in the presentation compiler to let the compiler find the signatures itself. I'm not too familiar with how this could work, however.

I personally believe the first approach is better since it works nicely across different projects.

martingd:
Hi @olafurpg 

Is it correct to assume that the issue is caused by Bloop not updating the class files until everything compiles?

If that's the case, is there any way to have Bloop update class files for source files that compile without errors so the class files are in sync with the sources (i.e., all sources that compile without errors)?

If yes, I assume Metals will start suggesting the correct names.

tgodzik:
Bloop might be copying the compiled artifacts only after compiling the whole build target. It should be possible to be changed, but not exactly sure about the consequences. This might go even deeper to Zinc, which is used underneath. I think that going the way explained by @olafurpg might be better.

tgodzik:
There is a related PR here: https://github.com/scalacenter/bloop/pull/1348/files where the old artifacts are copied for any failed compilation, so we can always have something working when the workspace is started again.

tgodzik:
After thinking about it for a bit let's say we have files A, B, C, D that depend on each other: A <- B, C <- D.

[...]

PR #3767, opened 2022-03-28
scala3 - exclude 3.2.0-NIGHTLY and align tests cases

dos65:
The existing `3.2.0` nightlies were published with incorrect version.
Current versions are published with `3.1.3-RC1*` prefix. So, to fix issues with ordering (these two looks like they are latest) I had to exclude them.

### DIFF
--- a/tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
@@ -304,7 +304,7 @@ class CompletionArgSuite extends BaseCompletionSuite {
     s"""|case class A(argument: Int)
         |object Main {
         |  def foo(argument: Int): A =
-        |    A(arg@@)
+        |    A(argu@@)
         |}
         |""".stripMargin,
     """|argument: Int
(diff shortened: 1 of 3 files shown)


tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -304,7 +304,7 @@ class CompletionArgSuite extends BaseCompletionSuite {
     s"""|case class A(argument: Int)
         |object Main {
         |  def foo(argument: Int): A =
-        |    A(arg@@)
+        |    A(argu@@)
```
dos65, line 307: I had to change these cases.

After [this fix](https://github.com/lampepfl/dotty/commit/f47c8c64af8382d1d41261d8b53fb18624243175) in dotty, it started returning package names in completions.
In this suite we have cases that enters packages with `arg` prefix ("arg1", "arg2") into compiler state and for latest version completions returns them too.

Unfortunately, I haven't found a way to cleanup them. Added some info about this problem in [this comment](https://github.com/scalameta/metals/issues/917#issuecomment-1080835816)

kpodsiad (approved):

Issue #3755, opened 2022-03-25
Calls to `doctor.check` on each build target compilation significantly degrade performance


pvid:
### Describe the bug

Using recent SNAPSHOT versions, I encountered a significant delay between saving a file and editor receiving diagnostics. It went from nearly imperceptible to > 10 seconds.

It is happening in a module that transitively depends on multiple other modules in a large sbt project. I managed to track the issue down to`doctor.check()` being called on each build target compilation.

https://github.com/scalameta/metals/blob/c2d8baa140b675bf900b1783daad2fa9eceeb424/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L427)

It is triggered even on no-op compilation, so the number of calls scale with the size of module dependencies.

Using a locally published metals version with that call removed fixes my issue, but I did not examine what it means for the new compilation status feature of the doctor. The `Doctor#check` method seems to go through all build targets on every call. I would image that it should just update the status of build targets whose compilation status changed. 

### Expected behavior

Do not run a heavy operations on no-op compilations

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

0.11.2+49-ee7372d7-SNAPSHOT

tgodzik:
Thanks for reporting! @kpodsiad it's probably related to your changes?

kpodsiad:
@pvid could you try https://github.com/kpodsiad/metals/tree/fix/doctor-check out? Personally, I don't have access to any codebase big enough to see what the difference looks like.

pvid:
@kpodsiad thank you for the quick turnaround! That version seems to fix the issue 👍

kpodsiad:
Glad to hear that it helps, but unfortunately it's rather quick & dirty than a rightful fix.

The problem is that Metals server doesn't know anything about vscode's Doctor panel, if it is open or closed. Without knowing that fact server doesn't know if reloading doctor is necessary or not. 

One possible solution is to keep track of its state on the server side. However, AFAIK clients like nvim just display doctor in json and don't have such dedicated view which can stay opened.

Although there is `metals-doctor-reload` command already, it takes an already computed html or json. In case when doctor window is closed then the result is discarded - `Reload the HTML contents of an open Doctor window, if any. Should be ignored if there is no open doctor window.`
I think introducing a new command `metals-refresh-doctor` is justified and makes sense. It will just notify the client that there might be an doctor client, but it'll up to the client what to do with this information.

[...]

PR #3768, opened 2022-03-29
fix: do not call `doctor.check()` on each build target compilation end

kpodsiad:
closes https://github.com/scalameta/metals/issues/3755
needs https://github.com/scalameta/metals-vscode/pull/930

Because `Compilations.scala` uses `BatchedFunction` to batch multiple requests it's very tempting to add there the callback that will execute doctor refresh. But I didn't want to couple doctor and compilations so I created `Subject`, should I just use `() => doctor.executeRefreshDoctor` as a `Compilations's` constructor argument?


metals/src/main/scala/scala/meta/internal/metals/doctor/Doctor.scala
```diff
@@ -115,7 +120,10 @@ final class Doctor(
         case DoctorFormat.Html => buildTargetsHtml()
       }
       val params = clientCommand.toExecuteCommandParams(output)
-      languageClient.metalsExecuteClientCommand(params)
+      val shouldDisplayForVsCode = (clientConfig.isVscode && isVisible.get())
+      if (shouldDisplayForVsCode || !clientConfig.isVscode) {
```
kpodsiad, line 124: Check visible status only for vscode - this gives other editors some time to adjust.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -720,6 +724,8 @@ class MetalsLanguageServer(
           mtagsResolver,
           () => userConfig.javaHome
         )
+        compilationCallbacks.addObserver(_ => doctor.executeRefreshDoctor())
```
kpodsiad, line 727: I think I should add check `if client == vscode` here.

kpodsiad (commented):
Actually, creating a new client's provider option should solve all of the above-mentioned suggestions. `doctorVisibilityProvider`?

metals/src/main/scala/scala/meta/internal/metals/InitializationOptions.scala
```diff
@@ -48,6 +48,7 @@ import org.eclipse.{lsp4j => l}
  *                                    copy results to the local buffer.
  * @param disableColorOutput in the situation where your DAP client may not handle color codes in
  *                            the output, you can enable this to strip them.
+ * @param doctorVisibilityProvider if the clients implements `metals/doctorVisibilityDidChange`
```
ckipp01, line 51: Can we also add this to `new-editor.md` with a note about how it's supposed to be used by a client.

metals/src/main/scala/scala/meta/internal/metals/Subject.scala
```diff
@@ -0,0 +1,7 @@
+package scala.meta.internal.metals
+
+trait Subject[S] {
```
ckipp01, line 3: Is `Subject` a common name for this? I normally think of this as an `Observable`?

kpodsiad: I think it heavily depends on language and people. My first encounter with reactive programming was rxjs and this feature seems like a https://rxjs.dev/guide/subject for me. 
I think the correct hierarchy looks like this
```scala
trait Observable[T] {
  def subscribe(fn: T => Unit): Unit
}

trait Subject[T] extends Observable[T] {
  def push(value: T): Unit
}
```

ckipp01: Ah ok, cool. Thanks for the explanation.

kpodsiad: I'll check Monix and ZIO to see how they name such things. It's better to stick with Scala conventions.

metals/src/main/scala/scala/meta/internal/metals/Subject.scala
```diff
@@ -0,0 +1,12 @@
+package scala.meta.internal.metals
+
+trait Observable[T] {
+  def subscribe(fn: T => Unit): Unit
+}
+
+final class Subject[T] extends Observable[T] {
+  private var subscribers: List[T => Unit] = Nil
```
dos65, line 8: it isn't thread-safe.
I would prefer to use just a function callback as long a there is no need in having multiply mutable subscribers

dos65 (changes requested):
Overall looks good.
I have only one comment about Subject

metals/src/main/scala/scala/meta/internal/metals/Subject.scala
kpodsiad: > I would prefer to use just a function callback

I wanted to avoid that because this requires moving `private var doctor: Doctor = _` much earlier than it currently is. At some point of the time, we will have to get rid of the early initializations, but let future us worry about this :D

kpodsiad:
@dos65 I removed Subject. I also encourage you to do reviews in metals-vscode and language-client ;)

dos65 (approved):
:+1:

PR #3770, opened 2022-03-30
fix: munit test discovery when there is no package

kpodsiad:


metals/src/main/scala/scala/meta/internal/metals/testProvider/frameworks/MunitTestFinder.scala
```diff
@@ -37,13 +37,16 @@ class MunitTestFinder(trees: Trees) {
     /**
      * Class definition is valid when package + class name is equal to one we are looking for
      */
-    def isValid(cls: Defn.Class, currentPackage: String): Boolean =
-      s"$currentPackage.${cls.name.value}" == suiteName.value
+    def isValid(cls: Defn.Class, currentPackage: Vector[String]): Boolean = {
+      val fullyQualifiedName =
+        if (currentPackage.isEmpty) cls.name.value
+        else s"${currentPackage.mkString(".")}.${cls.name.value}"
```
dos65, line 43:
```suggestion
      val fullyQualifiedName =
        (currentPackage :+ cls.name.value).mkString(".")
```

kpodsiad: I'll go with `currentPackage.appended(cls.name.value)`, I'm not a fan of collection operators different than `::` :D

dos65 (approved):

Issue #3479, opened 2022-01-07
Jdk sources selected incorrectly


dos65:
### Describe the bug

Currently, in case of several JDK are available its sources might be selected incorrectly.

Metals might take jdkHome from `userConfiguration.javaHome` or the default one that corresponds to JDK in which Metals process was started.

However, project might use a different one for compilation.
BSP has [`JvmBuldTarget`](https://build-server-protocol.github.io/docs/extensions/jvm.html#jvm-build-target) structure that is contained in `ScalaBuildTarget` and `JavaBuildTarget` and Metals should use this one to avoid showing wrong sources + probably it might cause some issues during debugging.

### Version of Metals

v0.10.9

PR #3772, opened 2022-03-30
feat: add java version and Metals version to doctor

kpodsiad:
Zero cost implementation of https://github.com/scalameta/metals-feature-requests/issues/264

However, I would love to have more detailed information about the server version and JDK (is it the default one, is it configured in the workspace or in user settings) so I'm not sure if the feature request should be closed.

Screenshot 2022-03-30 at 15 27 39


metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
```diff
@@ -223,9 +223,17 @@ final class Doctor(
       .render
   }
 
+  private def getJdkInfo(): Option[String] = {
+    for {
+      home <- javaHome().orElse(JdkSources.defaultJavaHome)
```
kpodsiad, line 228: Will `.orElse(JdkSources.defaultJavaHome)` be tricky? I'm not sure if our default java detection mechanism at client side is the same as at the server-side :/

ckipp01:
😁 Please don't forget to also add it to the json output.

kpodsiad (commented):
@ckipp01 Would you like separate fields in JSON or adding this info to the `heading` is ok?

metals/src/main/scala/scala/meta/internal/metals/Doctor.scala
ckipp01: 😆 ha, I should have waited. I think for now we can put it in the header text right under the build server server. However since we keep adding it, maybe we should change the format for the json header, but let's not do that here. I can think about it a bit more and then try to figure out what is best since I'm the main person that actually uses the json ha.

Arthurm1: Is `ScalaTarget#jvmHome` relevant here?  Doesn't it allow a jvm home per build target or have I misunderstood its purpose

dos65: Summarizing all aspects about java homes in Metals:
- `JdkSources.defaultJavaHome` - this java home is correct value for Metals process. 
- jdkHome from UserConfiguration. It may be different from the jdk in which the process is running and used for navigation into jdk sources.
- `ScalaTarget#jvmHome` this jvm home which bsp uses for compilation. Currently we don't use it Metals but there is an issue - #3479

kpodsiad: @dos65 are you sure that `UserConfiguration.javaHome` is taken into account as a second one?

dos65: @kpodsiad in case of the current state of VSCode extension, `UserConfiguration.javaHome` and `JdkSources.defaultJavaHome` should be equal at the moment of server start. However, it's possible to start metals and then change `javaHome` in UserConfiguration and they will be different.

We anyway should have at least two fields for jdk because it will be needed after #3479.
One is current server jdk, second is that is used for sources and compilation.

kpodsiad: I could rename java version to `Metals Java version` to not create confusion for users but honestly, now I don't know what I should put under that field :c Maybe I should just run `System.getProperty("java.version")` + `java.vendor` + `java.version.date` without a path?

kpodsiad:
@ckipp01 I took the liberty of merging this PR, hope you don't mind that I didn't wait for your ack.

ckipp01:
> @ckipp01 I took the liberty of merging this PR, hope you don't mind that I didn't wait for your ack.

😄 no problem at all.

PR #3773, opened 2022-03-31
fix: nightly script

dos65:


kpodsiad (approved):
Ok, I trust your shell scripting skills 😆

PR #3787, opened 2022-04-05
Support MatchType in SemanticdbTreePrinter

tanishiking:
- scalameta 4.5.1 supports [MatchType](https://dotty.epfl.ch/docs/reference/new-types/match-types.html) in SemanticDB  https://github.com/scalameta/scalameta/releases/tag/v4.5.1 
- The next version of Scala3 [will support MatchType](https://github.com/lampepfl/dotty/pull/14608) in SemanticDB

This PR should enable metals to show inferred `MatchType` with the latest Scala3 compiler.

I tried to test it locally, but I guess this change won't affect the behavior at this moment because
- `SemanticdbTreePrinter` is used only for showing the inferred type (in Scala3 at least, at this moment).
- However, I'm not sure about the situation where MatchType is inferred.

So this PR is safe to merge :)

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
+++ b/metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
@@ -86,8 +86,8 @@ class SemanticdbTreePrinter(
         } else {
           s"${printType(tpe)} {${printScope(scope)}}"
         }
-      case s.MatchType(scrutinee, _) =>
-        s"${printType(scrutinee)} match { ... }"
+      case s.MatchType(scrutinee, cases) =>
+        s"${printType(scrutinee)} match { ${cases.size} cases }"
     }
 
   def printScope(scope: Option[s.Scope]): String = {
--- a/metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
+++ b/metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
@@ -86,6 +86,13 @@ class SemanticdbTreePrinter(
         } else {
           s"${printType(tpe)} {${printScope(scope)}}"
         }
+      case s.MatchType(scrutinee, cases) =>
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM from my side, but I wonder what would be a place we would want to print it in and if so does it make sense to print everything. Maybe we could replace the cases by `...` ?

tanishiking:
You mean `s"${printType(scrutinee)} match { ... }"` instead of `s"${printType(scrutinee)} match { ${casesStr} }"`? I guess that's a good idea, print everything makes screen clatter especially for MatchType which usually have several cases.

metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala
```diff
@@ -86,6 +86,8 @@ class SemanticdbTreePrinter(
         } else {
           s"${printType(tpe)} {${printScope(scope)}}"
         }
+      case s.MatchType(scrutinee, _) =>
+        s"${printType(scrutinee)} match { ... }"
```
kpodsiad, line 90: Maybe
```suggestion
      case s.MatchType(scrutinee, cases) =>
        s"${printType(scrutinee)} match { ${cases.size} cases }"
```
or even some heuristic like if cases.size > 2 then print only size, otherwise print all cases (2 of them)?

kpodsiad (approved):
Had one suggestion, but it looks good already!

tgodzik:
> You mean `s"${printType(scrutinee)} match { ... }"` instead of `s"${printType(scrutinee)} match { ${casesStr} }"`? I guess that's a good idea, print everything makes screen clatter especially for MatchType which usually have several cases.

Thanks! I am also fine with what @kpodsiad is suggesting.

tanishiking:
Thanks! @kpodsiad @tgodzik 
Let's go with `s"${printType(scrutinee)} match { ${cases.size} cases }"`

tgodzik (approved):
LGTM!

PR #3788, opened 2022-04-05
chore: deprecate running Metals on Java 8

kpodsiad:
As discussed in https://github.com/scalameta/metals/discussions/3784

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/doctor/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/doctor/Doctor.scala
@@ -264,7 +264,7 @@ final class Doctor(
     versionString <- Option(System.getProperty("java.version"))
     version <- JdkVersion.parse(versionString)
     if version.major < 11
-  } yield "Running Metals on Java 8 is deprecated. In the future Metals will require Java 11"
+  } yield s"Running Metals on Java ${version.major} is deprecated. In the future Metals will require Java 11 or higher"
 
   private def getJdkInfo(): Option[String] =
     for {
--- a/metals/src/main/scala/scala/meta/internal/metals/doctor/Doctor.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/doctor/Doctor.scala
@@ -22,6 +22,7 @@ import scala.meta.internal.metals.FileDecoderProvider
 import scala.meta.internal.metals.HtmlBuilder
 import scala.meta.internal.metals.Icons
 import scala.meta.internal.metals.JavaTarget
+import scala.meta.internal.metals.JdkVersion
 import scala.meta.internal.metals.Messages.CheckDoctor
 import scala.meta.internal.metals.MetalsEnrichments._
 import scala.meta.internal.metals.MetalsHttpServer
@@ -162,7 +163,9 @@ final class Doctor(
   def check(): Unit = {
     val scalaTargets = buildTargets.allScala.toList
     val javaTargets = buildTargets.allJava.toList
-    val summary = problemResolver.problemMessage(scalaTargets, javaTargets)
+    val summary = problemResolver
+      .problemMessage(scalaTargets, javaTargets)
+      .orElse(deprecatedJavaVersion())
     executeReloadDoctor(summary)
     summary match {
       case Some(problem) =>
@@ -174,6 +177,9 @@ final class Doctor(
           hasProblems.set(true)
           languageClient.showMessageRequest(params).asScala.foreach { item =>
             if (item == CheckDoctor.moreInformation) {
+              if (clientConfig.isDoctorVisibilityProvider()) {
+                isVisible.set(true)
+              }
               executeRunDoctor()
             } else if (item == CheckDoctor.dismissForever) {
               notification.dismissForever()
@@ -251,12 +257,26 @@ final class Doctor(
       .render
   }
 
+  /**
+   * Java 8 is deprecated and in the future Metals will require Java >=11
+   */
+  private def deprecatedJavaVersion(): Option[String] = for {
+    versionString <- Option(System.getProperty("java.version"))
+    version <- JdkVersion.parse(versionString)
+    if version.major < 11
+  } yield "Running Metals on Java 8 is deprecated. In the future Metals will require Java 11"
+
   private def getJdkInfo(): Option[String] =
     for {
       version <- Option(System.getProperty("java.version"))
       vendor <- Option(System.getProperty("java.vendor"))
-      home <- Option(System.getProperty("java.home"))
-    } yield s"$version from $vendor located at $home"
+      // java.home points to the installation directory for Java Runtime Environment
+      // https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html
+      jreHome <- Option(System.getProperty("java.home"))
+    } yield {
+      val home = jreHome.stripSuffix(s"/jre")
+      s"$version from $vendor located at $home"
+    }
 
   private def buildTargetsJson(): String = {
     val targetIds = allTargetIds()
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/doctor/Doctor.scala
```diff
@@ -162,7 +163,9 @@ final class Doctor(
   def check(): Unit = {
     val scalaTargets = buildTargets.allScala.toList
     val javaTargets = buildTargets.allJava.toList
-    val summary = problemResolver.problemMessage(scalaTargets, javaTargets)
+    val summary = problemResolver
+      .problemMessage(scalaTargets, javaTargets)
+      .orElse(deprecatedJavaVersion())
```
kpodsiad, line 168: build targets' problems first, then jdk deprecation

metals/src/main/scala/scala/meta/internal/metals/doctor/Doctor.scala
```diff
@@ -174,6 +177,9 @@ final class Doctor(
           hasProblems.set(true)
           languageClient.showMessageRequest(params).asScala.foreach { item =>
             if (item == CheckDoctor.moreInformation) {
+              if (clientConfig.isDoctorVisibilityProvider()) {
+                isVisible.set(true)
+              }
```
kpodsiad, line 182: When there is some problem there is also `More Information` button, without this snippet, it won't work.

metals/src/main/scala/scala/meta/internal/metals/doctor/Doctor.scala
```diff
@@ -251,12 +257,26 @@ final class Doctor(
...
   }
 
+  /**
+   * Java 8 is deprecated and in the future Metals will require Java >=11
+   */
+  private def deprecatedJavaVersion(): Option[String] = for {
+    versionString <- Option(System.getProperty("java.version"))
+    version <- JdkVersion.parse(versionString)
+    if version.major < 11
+  } yield "Running Metals on Java 8 is deprecated. In the future Metals will require Java 11"
```
kpodsiad, line 267: Maybe `on Java 8` -> `on Java lesser than 11`?

metals/src/main/scala/scala/meta/internal/metals/doctor/Doctor.scala
kpodsiad: I broke this in https://github.com/scalameta/metals/pull/3768

metals/src/main/scala/scala/meta/internal/metals/doctor/Doctor.scala
```diff
@@ -251,12 +257,26 @@ final class Doctor(
...
   }
 
+  /**
+   * Java 8 is deprecated and in the future Metals will require Java >=11
+   */
+  private def deprecatedJavaVersion(): Option[String] = for {
+    versionString <- Option(System.getProperty("java.version"))
+    version <- JdkVersion.parse(versionString)
+    if version.major < 11
+  } yield s"Running Metals on Java ${version.major} is deprecated. In the future Metals will require Java 11"
```
tanishiking, line 267: Can we say "... will require Java 11 or higher"? So that users won't misunderstand Metals requires exactly Java 11.

kpodsiad: Good call!

metals/src/main/scala/scala/meta/internal/metals/doctor/Doctor.scala
```diff
@@ -390,6 +411,17 @@ final class Doctor(
       }
     }
 
+    deprecatedJavaVersion().foreach { deprecatedJdk =>
```
tgodzik, line 414: How does it look in the doctor?

tgodzik (approved):
Looks good, just a question.

tgodzik (changes requested):
Actually, I wonder about a case where you want to use JDK 8 to build Java files, it currently defaults to whatever you used for import. If you are unable to use it then you might not be able to compile your code. `javaHome` should in that case be used for Bloop + running sbt. 

We should probably only issue this warnings once we switch to Metals downloading a specific JDK version, so that this would only pop up for users of other editors, who would still be able to run Metals with different versions. (maybe we could make a requirement in coursier? OR build Metals using JDK17 ?)

dos65:
@tgodzik 

I think these cases are rare and might be solved by having several jdk installed.
For example, you might have jdk8 as a default but the jdk selection rules in metals-vscode can select a higher non-default one `>=11` for Metals process.

tgodzik:
> @tgodzik
> 
> I think these cases are rare and might be solved by having several jdk installed. For example, you might have jdk8 as a default but the jdk selection rules in metals-vscode can select a higher non-default one `>=11` for Metals process.

Yeah, that's why I think we should separate Java used to run Metals (even download our own) from the one used for compilation etc.

dos65:
It seems that we stuck here.
What about adding adding an additional default jdk as parameter to initialize options?
Then we would be able running Metals on higher jdk version and keep using jdk8 for bloop if it's a default one.

tgodzik:
> It seems that we stuck here. What about adding adding an additional default jdk as parameter to initialize options? Then we would be able running Metals on higher jdk version and keep using jdk8 for bloop if it's a default one.

I think we should do a couple of separate step:
- download our own binary for Metals JDK 17 and decouple javaHome (maybe rename the description)
- change the plugins for generating Bloop configuration to add the proper release flags to make sure it's run with a proper version
- automatically set javaHome in Bloop to be the same JDK 17 as Metals uses

There is not much benefit in deprecating it here without those steps I feel.

kpodsiad:
> I think we should do a couple of separate step:
> 
> * download our own binary for Metals JDK 17 and decouple javaHome (maybe rename the description)
> * change the plugins for generating Bloop configuration to add the proper release flags to make sure it's run with a proper version
> * automatically set javaHome in Bloop to be the same JDK 17 as Metals uses
> 
> There is not much benefit in deprecating it here without those steps I feel.

Sounds very reasonable.

dos65:
@tgodzik 
> change the plugins for generating Bloop configuration to add the proper release flags to make sure it's run with a proper version

What do you mean here by `release flags`? javac and scalac?

tgodzik:
> @tgodzik
> 
> > change the plugins for generating Bloop configuration to add the proper release flags to make sure it's run with a proper version
> 
> What do you mean here by `release flags`? javac and scalac?

Yeah, similar way that ScalaCLI does it. Currently, it's not working too great since we just run with whatever Bloop is started with.

dos65:
> > @tgodzik
> > > change the plugins for generating Bloop configuration to add the proper release flags to make sure it's run with a proper version
> > 
> > 
> > What do you mean here by `release flags`? javac and scalac?
> 
> Yeah, similar way that ScalaCLI does it. Currently, it's not working too great since we just run with whatever Bloop is started with.

I wouldn't go this way. Even if we tune options with proper release flags some user might face the issues during the `run`/`debug` because the runtime is different from the expected one.
Also, there will be difference between bsp servers in this moment. Auto-jdk17 with bloop and system-default for sbt/mill/...

dos65:
@tgodzik ok, sounds good!
Looks like this one can be closed.

kpodsiad:
> I think we should do a couple of separate step:
> 
> * download our own binary for Metals JDK 17 and decouple javaHome (maybe rename the description)
> * change the plugins for generating Bloop configuration to add the proper release flags to make sure it's run with a proper version
> * automatically set javaHome in Bloop to be the same JDK 17 as Metals uses
> 
> There is not much benefit in deprecating it here without those steps I feel.

Could you put this (in a bit more verbose form) into an issue?

tgodzik:
Done! https://github.com/scalameta/metals/issues/3923

PR #3789, opened 2022-04-05
Log BSP server stderr

alexarchambault:
This PR redirects the stderr of BSP servers to the logs of Metals. It can be helpful when anything goes wrong with a BSP server, giving Metals users any message that the BSP server printed.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/bsp/BspServers.scala
+++ b/metals/src/main/scala/scala/meta/internal/bsp/BspServers.scala
@@ -21,6 +21,7 @@ import scala.meta.internal.metals.SocketConnection
 import scala.meta.internal.metals.Tables
 import scala.meta.internal.metals.clients.language.MetalsLanguageClient
 import scala.meta.internal.mtags.MD5
+import scala.meta.internal.process.SystemProcess
 import scala.meta.io.AbsolutePath
 
 import ch.epfl.scala.bsp4j.BspConnectionDetails
@@ -65,23 +66,31 @@ final class BspServers(
   ): Future[BuildServerConnection] = {
 
     def newConnection(): Future[SocketConnection] = {
-      val process = new ProcessBuilder(details.getArgv)
-        .directory(projectDirectory.toFile)
-        .start()
+      scribe.info(s"Running BSP server ${details.getArgv}")
+
+      val proc = SystemProcess.run(
+        details.getArgv.asScala.toList,
+        projectDirectory,
+        redirectErrorOutput = false,
+        Map(),
+        processOut = None,
+        processErr = Some(l => scribe.info("BSP server: " + l)),
+        discardInput = false,
+        threadNamePrefix = s"bsp-${details.getName}"
+      )
 
       val output = new ClosableOutputStream(
-        process.getOutputStream,
+        proc.outputStream,
         s"${details.getName} output stream"
       )
       val input = new QuietInputStream(
-        process.getInputStream,
+        proc.inputStream,
         s"${details.getName} input stream"
       )
 
       val finished = Promise[Unit]()
-      Future {
-        process.waitFor()
-        finished.success(())
+      proc.complete.ignoreValue.onComplete { res =>
+        finished.tryComplete(res)
       }
 
       Future.successful {
@@ -90,7 +99,7 @@ final class BspServers(
           output,
           input,
           List(
-            Cancelable(() => process.destroy())
+            Cancelable(() => proc.cancel)
           ),
           finished
         )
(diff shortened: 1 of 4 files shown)


dos65 (commented):
@alexarchambault Thx for contribution!

The only thing is that we already have a wrapper to java process api inside metals - `scala.meta.interal.process.SystemProcess.run`. It allows the same things that you added - process output handling and cancellation.
Would you like to switch this part on it?

dos65:
@alexarchambault update - I found that branch. Actually, I started doing that but haven't completed and dropped these changes without committing :smile:

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
```diff
@@ -112,6 +125,16 @@ object SystemProcess {
...
       override def complete: Future[Int] = Future.successful(1)
+      override def inputStream: InputStream =
+        new InputStream {
+          def read() = -1
+          override def read(b: Array[Byte], off: Int, len: Int) = -1
+        }
+      override def outputStream: OutputStream =
+        new OutputStream {
+          def write(value: Int) = {}
+          override def write(b: Array[Byte], off: Int, len: Int): Unit = {}
```
kpodsiad, line 136:
```suggestion
          override def write(value: Int) = ()
          override def write(b: Array[Byte], off: Int, len: Int): Unit = ()
```

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
```diff
@@ -112,6 +125,16 @@ object SystemProcess {
   val Failed: SystemProcess =
     new SystemProcess {
       override def complete: Future[Int] = Future.successful(1)
+      override def inputStream: InputStream =
+        new InputStream {
+          def read() = -1
+          override def read(b: Array[Byte], off: Int, len: Int) = -1
```
kpodsiad, line 131:
```suggestion
          override def read() = -1
          override def read(b: Array[Byte], off: Int, len: Int) = -1
```

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
```diff
@@ -53,8 +57,9 @@ object SystemProcess {
...
       ps: Process,
       redirectErrorOutput: Boolean,
-      processOut: String => Unit,
-      processErr: String => Unit
+      processOut: Option[String => Unit],
+      processErr: Option[String => Unit],
+      discardInput: Boolean
   ): SystemProcess = {
     def readOutput(stream: InputStream, f: String => Unit): Thread = {
       val filter = AnsiFilter()
```
kpodsiad, line 65: I really liked that you gave a thread a meaningful name before, could we do this now too?

metals/src/main/scala/scala/meta/internal/bsp/BspServers.scala
```diff
@@ -65,23 +66,30 @@ final class BspServers(
...
-        .start()
+      scribe.info(s"Running BSP server ${details.getArgv}")
+
+      val proc = SystemProcess.run(
+        details.getArgv.asScala.toList,
+        projectDirectory,
+        redirectErrorOutput = false,
+        Map(),
+        processOut = None,
+        processErr = Some(l => scribe.info("BSP server: " + l)),
```
kpodsiad, line 77: Shouldn't it be `scribe.error`?

kpodsiad (approved):
Looks good 👍🏻

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
alexarchambault: Should be fixed.

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
alexarchambault: Should be fixed too.

metals/src/main/scala/scala/meta/internal/process/SystemProcess.scala
alexarchambault: Good idea, yeah. Should be fixed now.

metals/src/main/scala/scala/meta/internal/bsp/BspServers.scala
alexarchambault: I'm not sure. It's stderr, because it's the only place a BSP server can print in the console, but it doesn't mean there's an actual error here. It's whatever the BSP server wants to print.

tgodzik (approved):
LGTM!

PR #3791, opened 2022-04-05
fix: handle backticked names in inffered-type

### LINKED ISSUE #3785: Misplaced type annotation on backticked definitions
### Describe the bug

Invoking `Insert type annotation` code action on a backticked definition adds the annotation in a wrong position.
For a `def`:
```scala
object Foo {
  def `bar` = 1
}
```
becomes
```scala
object Foo {
  def `ba: Intr` = 1
}
```
in scala 2.13.8 but works correctly in scala 3.1.1

For a `val`:
```scala
object Foo {
  val `bar` = 1
}
```
gives
```scala
object Foo {
  val `ba: Intr` = 1
}
```
in scala 2.13.8 and
```scala
object Foo {
  val `bar: Int` = 1
}
```
in scala 3.1.1

### Expected behavior

```scala
object Foo {
  def `bar`: Int = 1
}
```

```scala
object Foo {
  val `bar`: Int = 1
}
```

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

0.11.2+123-b44dc8c0-SNAPSHOT

_No response_

dos65:
Fixes #3785

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
@@ -17,6 +17,8 @@ import scala.meta.tokens.{Token as T}
 import dotty.tools.dotc.ast.Trees.*
 import dotty.tools.dotc.ast.untpd
 import dotty.tools.dotc.core.Contexts.*
+import dotty.tools.dotc.core.NameOps.*
+import dotty.tools.dotc.core.Names.*
 import dotty.tools.dotc.core.Symbols.*
 import dotty.tools.dotc.core.Types.*
 import dotty.tools.dotc.interactive.Interactive
@@ -24,6 +26,7 @@ import dotty.tools.dotc.interactive.InteractiveDriver
 import dotty.tools.dotc.util.SourceFile
 import dotty.tools.dotc.util.SourcePosition
 import dotty.tools.dotc.util.Spans
+import dotty.tools.dotc.util.Spans.Span
 import org.eclipse.lsp4j.TextEdit
 
 /**
@@ -121,9 +124,10 @@ final class InferredTypeProvider(
        */
       case Some(vl @ ValDef(sym, tpt, _)) =>
         val isParam = path.tail.headOption.exists(_.symbol.isAnonymousFunction)
-        def baseEdit(withParens: Boolean) =
+        def baseEdit(withParens: Boolean): TextEdit =
+          val endPos = findNamePos(params.text, vl).endPos
           new TextEdit(
-            vl.namePos.endPos.toLSP,
+            endPos.toLSP,
             ": " + printType(tpt.tpe) + {
               if withParens then ")" else ""
             }
@@ -134,7 +138,7 @@ final class InferredTypeProvider(
             toCheckFor: Char,
             blockStartPos: SourcePosition
         ) =
-          val text = params.text.toString()
+          val text = params.text
           val isParensFunction: Boolean = text(applyEndingPos) == toCheckFor
 
           val alreadyHasParens =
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -251,4 +255,32 @@ final class InferredTypeProvider(
...
+        case Some((start, nextMatch :: left)) =>
+          if text.charAt(idx) == nextMatch then
+            lookup(idx + 1, Some((start, left)), withBacktick)
+          else lookup(idx + 1, None, withBacktick = false)
+        case Some((start, Nil)) =>
+          val end = if withBacktick then idx + 1 else idx
+          tree.source.atSpan(Span(start, end, start))
+        case None =>
+          val ch = text.charAt(idx)
+          if ch == realName.head then
```
tanishiking, line 278: Maybe it's too cautious, but can we use `headOption` to avoid `NoSuchElementException` here?

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -251,4 +255,32 @@ final class InferredTypeProvider(
...
   end inferredTypeEdits
 
+  private def namePos(text: String, tree: untpd.NameTree)(using
+      Context
+  ): SourcePosition =
+    val realName = tree.name.stripModuleClassSuffix.lastPart.toString.toList
+
+    def lookup(
+        idx: Int,
+        start: Option[(Int, List[Char])],
```
tanishiking, line 265: [note]
A pair of next position and trailing realName

tanishiking (approved):
LGTM!
Maybe we can backport this `namePos` strategy (in Scala3) to `Interactive#namePos` https://github.com/lampepfl/dotty/blob/73cda0c9ad9a23132dabc8f8a7a47663da3ee3fa/compiler/src/dotty/tools/dotc/interactive/SourceTree.scala#L34-L36 ?

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
kpodsiad: Is it possible to get zero length name?

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
kpodsiad: Maybe forge this note into code?
```scala
case class Lookup(position: Int, tralingName: List[Char])
```

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -251,4 +255,32 @@ final class InferredTypeProvider(
     end match
   end inferredTypeEdits
 
+  private def namePos(text: String, tree: untpd.NameTree)(using
+      Context
+  ): SourcePosition =
+    val realName = tree.name.stripModuleClassSuffix.lastPart.toString.toList
```
kpodsiad, line 261: Could you provide some example how `tree.name.stripModuleClassSuffix` and `tree.name.stripModuleClassSuffix.lastPart` differs from each other?

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -251,4 +255,32 @@ final class InferredTypeProvider(
...
+      start match
+        case Some((start, nextMatch :: left)) =>
+          if text.charAt(idx) == nextMatch then
+            lookup(idx + 1, Some((start, left)), withBacktick)
+          else lookup(idx + 1, None, withBacktick = false)
+        case Some((start, Nil)) =>
+          val end = if withBacktick then idx + 1 else idx
+          tree.source.atSpan(Span(start, end, start))
+        case None =>
+          val ch = text.charAt(idx)
```
kpodsiad, line 277: Should we guard `idx` with, for example, `tree.sourcePos.end`?

kpodsiad (approved):
Looks good.

dos65:
@tanishiking 
> Maybe we can backport this namePos strategy (in Scala3) to Interactive#namePos https://github.com/lampepfl/dotty/blob/73cda0c9ad9a23132dabc8f8a7a47663da3ee3fa/compiler/src/dotty/tools/dotc/interactive/SourceTree.scala#L34-L36 

I thought about that. Honestly, I don't like the approach in this pr but there is no other ways to fix it. 
On compiler side it should be handled in a more elegant way.

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
dos65: @kpodsiad  I don't know. I just copied this line from from compiler impl for `namePos`

kpodsiad: Ok, adding info that `this line from from compiler impl for namePos` will also be helpful for future us.

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
tgodzik: I don't think it's possible, we could add a require with a proper message just in case

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
tgodzik: `case None if  idx < text.length` should probably suffice

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
dos65: ok, just removed `lastPart` call. Tests pass without it.

mtags/src/main/scala-3/scala/meta/internal/pc/InferredTypeProvider.scala
dos65: added a check for that

PR #3792, opened 2022-04-05
[Scala3] Don't show a tooltip on hover if cursor is not on a symbol

tanishiking:
Partially fix https://github.com/scalameta/metals/issues/1707 (hover part of the issue), taking over from https://github.com/scalameta/metals/pull/3743

FYI @tdudzik I opened a PR to fix the issue on the Hover part, if you're interested in working on go-to-definition part, please let me know :) 

This PR fixes Scala3 HoverProvider not to show a tooltip on hover if the cursor is not on a symbol by checking the `namePos` (for example, `namePos` of `def foo(x: Int) = { ... }` is a position of `def <foo>(...) = { ... }` part, not a entire `<def foo(x: Int) = { ... }>`) of the enclosing tree actually contains the cursor position.

Calculate the `namePos` using [SourceTree#namePos](https://github.com/lampepfl/dotty/blob/d8e50f249ce0206faae35d5d3d28f45f1e0f58d1/compiler/src/dotty/tools/dotc/interactive/SourceTree.scala#L21-L22).

no-hover-on-non-symbol

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
@@ -85,10 +85,8 @@ object HoverProvider:
     end isHoveringOnName
 
     if tp.isError || tpw == NoType || tpw.isError || path.isEmpty ||
-      (pos.start == pos.end && !isHoveringOnName(
-        enclosing,
-        pos
-      )) // don't check isHoveringOnName for RangeHover
+      (pos.isPoint // don't check isHoveringOnName for RangeHover
+        && !isHoveringOnName(enclosing, pos))
     then ju.Optional.empty()
     else
       Interactive.enclosingSourceSymbols(enclosing, pos) match
@@ -146,6 +144,9 @@ object HoverProvider:
           end match
     end if
   end hover
+
+  extension (pos: SourcePosition)
+    private def isPoint: Boolean = pos.start == pos.end
 
   private def qual(tree: Tree): Tree =
     tree match
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
@@ -20,6 +20,7 @@ import dotty.tools.dotc.core.Symbols.*
 import dotty.tools.dotc.core.Types.*
 import dotty.tools.dotc.interactive.Interactive
 import dotty.tools.dotc.interactive.InteractiveDriver
+import dotty.tools.dotc.interactive.SourceTree
 import dotty.tools.dotc.util.ParsedComment
 import dotty.tools.dotc.util.SourcePosition
 import org.eclipse.lsp4j.Hover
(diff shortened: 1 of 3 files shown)


dos65 (approved):
LGTM!

mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
```diff
@@ -49,8 +51,45 @@ object HoverProvider:
...
+        .headOption
+        .getOrElse(EmptyTree)
+      contains(enclosing)
+    end isHoveringOnName
+
+    if tp.isError || tpw == NoType || tpw.isError || path.isEmpty ||
+      (pos.start == pos.end && !isHoveringOnName(
+        enclosing,
+        pos
+      )) // don't check isHoveringOnName for RangeHover
```
kpodsiad, line 91: Could you extract this part of the condition to `def`?

kpodsiad (approved):

mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
tanishiking: Refactored a bit in https://github.com/scalameta/metals/pull/3792/commits/cf020525fb33f675ac665bd090553bdee7db99c4

mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
```diff
@@ -49,8 +51,43 @@ object HoverProvider:
...
+     * ```
+     * @param path - path to the position given by `Interactive.pathTo`
+     */
+    def isHoveringOnName(path: List[Tree], sourcePos: SourcePosition): Boolean =
+      def contains(tree: Tree): Boolean = tree match
+        case tree: NameTree =>
+          source.fold(false) { s =>
+            SourceTree(tree, s).namePos.contains(sourcePos)
+          }
+        // TODO: check the positions for NamedArg and Import
```
tgodzik, line 73: @tanishiking should it have been addressed in this PR before merging?

tests/cross/src/test/scala/tests/hover/HoverTermSuite.scala
```diff
@@ -112,7 +112,7 @@ class HoverTermSuite extends BaseHoverSuite {
   )
 
   check(
-    "interpolator-apply",
+    "interpolator-apply".tag(IgnoreScala3),
```
tgodzik, line 115: This seems to have worked before, why was it ignored?

mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
tanishiking: The drawback of not checking these trees is: we show a hover tooltip if the cursor is on

- ` = ` part of named argument
- ` as ` part of import statement

which aren't big problems, so I believe we can address these in another PR (maybe we should open another issue)

Screen Shot 2022-04-08 at 12 39 08

Screen Shot 2022-04-08 at 12 39 00

tests/cross/src/test/scala/tests/hover/HoverTermSuite.scala
tanishiking: Oh, I guess I overlooked it. It should be fixed!

mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
tgodzik: For sure this is not a big issue :+1:

Issue #3794, opened 2022-04-05
Requests fail when editing file with path that includes spaces


brdoney:
### Describe the bug

Whenever I edit a file that has a path which includes spaces, for example `~/Library/Mobile Documents/com~apple~CloudDocs/Documents/testing.sc` (note the space between `Mobile` and `Documents`), requests for folding ranges, code lenses, completion, and more seem to fail, even though the features are still provided as I'm actually editing the code.

I looked at the error log in VSCode and found that for nearly every request - `textDocument/foldingRange`, `textDocument/documentHighlight`, and `textDocument/codeLens`, to name a few - I get the same `URISyntaxException` at the root of the stack trace triggered by the failing request:

<details>
	<summary>Stack trace for when the textDocument/codeLens requests fail</summary>

```rawtext
java.util.concurrent.CompletionException: java.lang.IllegalArgumentException: Illegal character in path at index 42: file:///Users/brdoney/Library/Mobile Documents/com~apple~CloudDocs/Documents/testing.sc
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:331)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:346)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:704)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088)
	at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1(CancelTokens.scala:40)
	at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1$adapted(CancelTokens.scala:38)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)

Caused by: java.lang.IllegalArgumentException: Illegal character in path at index 42: file:///Users/brdoney/Library/Mobile Documents/com~apple~CloudDocs/Documents/testing.sc
	at java.base/java.net.URI.create(URI.java:883)
[...]

PR #3795, opened 2022-04-05
fix: uri encoding

kpodsiad:
closes https://github.com/scalameta/metals/issues/3794

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/mtags/MtagsEnrichments.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/mtags/MtagsEnrichments.scala
@@ -133,8 +133,13 @@ trait MtagsEnrichments extends CommonMtagsEnrichments {
         URLDecoder.decode(value, "UTF-8").toAbsolutePath(followSymlink)
       else if (value.toUpperCase.startsWith("JAR:FILE%3A"))
         URLDecoder.decode(value, "UTF-8").toAbsolutePath(followSymlink)
-      else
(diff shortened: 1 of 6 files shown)


mtags/src/main/scala-2/scala/meta/internal/mtags/MtagsEnrichments.scala
```diff
@@ -133,8 +133,13 @@ trait MtagsEnrichments extends CommonMtagsEnrichments {
         URLDecoder.decode(value, "UTF-8").toAbsolutePath(followSymlink)
       else if (value.toUpperCase.startsWith("JAR:FILE%3A"))
         URLDecoder.decode(value, "UTF-8").toAbsolutePath(followSymlink)
-      else
-        URI.create(value.stripPrefix("metals:")).toAbsolutePath(followSymlink)
+      else if (value.toUpperCase.startsWith("JAR"))
+        URI.create(value).toAbsolutePath(followSymlink)
```
kpodsiad, line 137: We do not want to overcode JARs, right @Arthurm1?

Arthurm1: Good question - I don't think so.
I'm trying to move away from `jar` URIs in https://github.com/scalameta/metals/pull/3750 so that only Metals will deal with `jar` URIs and only in a few specific classes and the client won't encounter them at all so won't mess with the encoded value.

tgodzik (approved):
LGTM!

PR #3799, opened 2022-04-06
docs: Neovim and Plug clarification

tacerus:
Signed-off-by: Georg Pfuetzenreuter <mail@georg-pfuetzenreuter.net>

Hi,

I figured this bit needed a little bit of clarification for people not that familiar with Neovim and/or vim-plug, as the configuration file differs, and needs a preceding Plug-call.

Best,
Georg

### DIFF
--- a/docs/editors/vim.md
+++ b/docs/editors/vim.md
@@ -67,7 +67,6 @@ plugin manager like vundle works too, but requires slightly different configurat
 following settings.
 
 ```vim
-call plug#begin()
 Plug 'neoclide/coc.nvim', {'branch': 'release'}
 ```
 
--- a/docs/editors/vim.md
+++ b/docs/editors/vim.md
@@ -63,10 +63,11 @@ Language Server Protocol support to Vim/Nvim  to communicate with the Metals
 language server.
 
 Assuming [`vim-plug`](https://github.com/junegunn/vim-plug) is used (another
-plugin manager like vundle works too), update your `~/.vimrc` to include the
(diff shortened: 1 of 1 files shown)


docs/editors/vim.md
```diff
@@ -63,10 +63,11 @@ Language Server Protocol support to Vim/Nvim  to communicate with the Metals
 language server.
 
 Assuming [`vim-plug`](https://github.com/junegunn/vim-plug) is used (another
-plugin manager like vundle works too), update your `~/.vimrc` to include the
+plugin manager like vundle works too, but requires slightly different configuration), update your `~/.vimrc` (or, in case of Neovim, `~/.config/nvim/init.vim`) to include the
 following settings.
 
 ```vim
+call plug#begin()
```
ckipp01, line 70: I think I'd leave this out to potentially avoid people copy pasting this in and having multiple calls to `plug#begin()`. There is a bit of assumed knowledge here about how a plugin manager works. The reason I only include the `Plug` line is that most plugins that list installation instructions do it this way, so it should be pretty standard.

ckipp01 (commented):
Thanks for the pr @tacerus!

I'd say 👍🏼 to the top change, but I'd recommend removing the `plug#begin()`.

docs/editors/vim.md
tacerus: That makes sense. I was thinking of someone using vim-plug for the first time, encountering an error after copy pasting only the Plug line - but it can of course be expected from people to read the vim-plug README first.

ckipp01 (approved):
Awesome, thanks @tacerus!

tacerus:
Thanks as well! :-)

Issue #3133, opened 2021-09-16
[Scala 3] Rename extension parameter does not work


tgodzik:
**Describe the bug**

```scala
  extension (sbd: String)
    def double = sbd + sbd
    def double2 = sbd + sbd
  end extension
```

Renaming will `sbd` only rename the ones in one of the method.

**To Reproduce**

Use the above code snippet

**Expected behavior**

All occurrences of `sbd` are renamed.

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.10.6

**Additional context**

This is caused by the fact that extension methods generate separate method with different parameters for each method in the extension group. This is also related to https://github.com/scalameta/scalameta/issues/2443

For now we could try searching for all occurences of symbols at the definition of the main parameter, since the main parameter will have a separate occurrence generated for each method.

**Search terms**

scala 3 extension group

tanishiking:
In the discussion https://github.com/scalameta/scalameta/issues/2443 we decided to deal with this problem on Metals side, I'll take a look if it's feasible option :)

PR #3800, opened 2022-04-07
fix: Rename extension parameter in Scala3

tanishiking:
Fix https://github.com/scalameta/metals/issues/3133

rename-extension

This PR enables Metals to rename all extension parameters.

Previously, when metals rename `sbd`  in `double` method.

```scala
  extension (sbd: String)
    def double = sbd + sbd
    def double2 = sbd + sbd
  end extension
```

Metals couldn't rename `sbd` in `double2` method because `sbd` in `double` and `sbd` in `double2` have different (semanticdb) symbols.

This PR fixes the issue by

- Make `ReferenceProvider` return multiple `ReferenceResult` because `find-references` may find references for multiple symbols.
  - For example, `sbd` in `extension (sbd: String)` is attached to two different symbols. see: https://github.com/scalameta/scalameta/issues/2443
  - Now `reference` returns all references for all symbols.
  - Therefore, this PR also fixes "find-references" for the extension parameter :+1:
- In RenameProvider, find all symbol occurrences (a reference to the symbol) from the definition site.
  - For example, while `sbd` in `double` has only one symbol,
  - `sbd` in `extension (sbd: String)` has two symbols, and we can find all references both in `double` and `double2`.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
@@ -196,6 +196,38 @@ final class DefinitionProvider(
     ResolvedSymbolOccurrence(sourceDistance, occurrence)
   }
 
+  def positionOccurrences(
+      source: AbsolutePath,
+      dirtyPosition: Position,
+      snapshot: TextDocument
+  ): List[ResolvedSymbolOccurrence] = {
+    // Convert dirty buffer position to snapshot position in "source"
+    val sourceDistance = buffers.tokenEditDistance(source, snapshot.text, trees)
+    val snapshotPosition = sourceDistance.toOriginal(
+      dirtyPosition.getLine,
+      dirtyPosition.getCharacter
+    )
+
+    // Find matching symbol occurrence in SemanticDB snapshot
+    val occurrence = for {
+      queryPosition <- snapshotPosition.toPosition(dirtyPosition)
+    } yield {
+      val occs = snapshot.occurrences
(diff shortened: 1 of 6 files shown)


tanishiking:
I realized document highlighting also doesn't work correctly for extension parameters because of the same reason as `RenameProvider`.
https://github.com/scalameta/metals/blob/e355f0fefbe07142fd42ad05d2dc8fba4dd70beb/metals/src/main/scala/scala/meta/internal/metals/DocumentHighlightProvider.scala#L30-L34

However, this approach requires us to invoke `go-to-definition` internally, and I'm not sure it's worth running go-to-definition (which requires additional CPU resources) every time we hover a cursor on a symbol only for the sake of extension parameters 🤔

tgodzik:
> I realized document highlighting also doesn't work correctly for extension parameters because of the same reason as `RenameProvider`.
> 
> https://github.com/scalameta/metals/blob/e355f0fefbe07142fd42ad05d2dc8fba4dd70beb/metals/src/main/scala/scala/meta/internal/metals/DocumentHighlightProvider.scala#L30-L34
> 
> However, this approach requires us to invoke `go-to-definition` internally, and I'm not sure it's worth running go-to-definition (which requires additional CPU resources) every time we hover a cursor on a symbol only for the sake of extension parameters thinking

I've been meaning to swap DocumentHighlight to use the presentation compiler, which would make it more reliable. We can address the issue there.

tgodzik (approved):
Finally got to review the PR. Looks great!

PR #3805, opened 2022-04-07
Remove broken nightly version

tgodzik:
It seems that the version is indeed listed on the maven central page, but nothing ever got released.


dos65 (approved):
Thx!

Issue #3796, opened 2022-04-05
Installation documentation is very terse


juhp:
I am still a newcomer to scala and metals, and an Emacs user.
I find the website documentation about installation is really too terse.
I think a note should be added somewhere about how to install metals from the command line.

kpodsiad:
Hi @juhp, thanks for reporting this!
Do you manage to get the Metals working?

ckipp01:
Thanks for the feedback @juhp. You can see some instructions on how to bootstrap metals from the command line [here](https://scalameta.org/metals/docs/editors/emacs#eglot) if you scroll down a bit. However, it's even easier than that as you can also do a `cs install metals` if you have [coursier](https://get-coursier.io/) installed.

Most of the editor-specific docs are maintained by users of that editor, so please do feel free to report back improvements or make changes as you see mistakes or opportunities. Doc updates are always very welcome. You can see how to do those [here](https://scalameta.org/metals/docs/contributors/updating-website).

juhp:
Thanks for the helpful reply.

Would it make sense to mention `coursier install metals` as an option in a small section of the top text editors overview page?

ckipp01:
> Would it make sense to mention coursier install metals as an option in a small section of the top text editors overview page?

Honestly probably not on the overview page. Mainly because the vast majority of users will be using a client where it automatically will install Metals for you. So I'd rather not cause confusion by putting the `coursier install metals` right out in front.

PR #3811, opened 2022-04-10
docs overview: mention editor pages for installation (#3796)

juhp:


docs/editors/overview.md
```diff
@@ -234,6 +234,8 @@ out-of-the-box.
 **Few steps**: installing Metals requires a few custom steps and minimal
 configuration to work.
 
+Go to the page for your editor to find out how to install.
```
ckipp01, line 237:
```suggestion
_You can find instructions on how to install Metals for you editor on your editor-specific page._
```

What do you think about something like this instead? These sections are just meant to serve as an explanation to the table above, but I suppose it's a good idea to clarify where to get installation instructions like this if it's causing confusion.

ckipp01 (changes requested):
Thanks for the pr @juhp! Just a small suggestion.

docs/editors/overview.md
juhp: If it were me I would further even add links to each editor page in the text.

I think you have a small typo: "for _your_ editor".

ckipp01: > I think you have a small typo: "for your editor".

Correct, good catch.  As for the link to each editor, I think a note here and then the side bar should be enough to point people in the right direction.

ckipp01 (approved):
Thanks @juhp! LGTM!

PR #3812, opened 2022-04-10
Remove -release option from the Scala 3 PC

tgodzik:
Previously, in a number of scenarios this option would break the presentation compiler for Scala 3. Now we filter it out since it's not important for the frontend phases.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
@@ -49,6 +49,8 @@ case class ScalaPresentationCompiler(
   val scalaVersion = BuildInfo.scalaCompilerVersion
 
   private val forbiddenOptions = Set("-print-lines", "-print-tasty")
+  private val forbiddenDoubleOptions = Set("-release")
+
   val compilerAccess: CompilerAccess[StoreReporter, InteractiveDriver] =
     Scala3CompilerAccess(
       config,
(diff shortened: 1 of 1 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -58,10 +60,19 @@ case class ScalaPresentationCompiler(
       using ec
     )
 
+  private def removeDoubleOptions(options: List[String]): List[String] =
+    options match
+      case head :: _ :: tail if forbiddenDoubleOptions(head) =>
+        removeDoubleOptions(tail)
+      case head :: tail => head :: removeDoubleOptions(tail)
+      case Nil => options
```
tanishiking, line 68: Why not

```suggestion
      options.filterNot(forbiddenDoubleOptions)
```

?

tanishiking: Ahh, nevermind, `removeDoubleOptions` transforms `List("-release", "8", ...)` to `List(...)`

tanishiking (approved):
LGTM

```
-release              Compile code with classes specific to the given version
                      of the Java platform available on the classpath and emit
                      bytecode for this version.
```

Indeed it's not important for `frontend` phases :+1:

kpodsiad (approved):
Great!

mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
tgodzik: I had to get a bit more creative here :sweat_smile:

Issue #2627, opened 2021-03-22
[Scala3] No definition for extension parameter symbol


dos65:
In #2600 ScalaMtags got initial support for extension methods but it doesn't cover parameters -https://github.com/scalameta/metals/blob/8f78cf55fc46e8b6fcdb7061b57b8dc4900a94df/mtags/src/main/scala-2/scala/meta/internal/mtags/ScalaMtags.scala#L248-L250

This issue is just a reminder to complete `extension methods` after [this issue in Scala3](https://github.com/lampepfl/dotty/issues/11690) will be fixed.

tanishiking:
Gonna work simultaneously with https://github.com/scalameta/metals/issues/3133 as the issue requires a fix also in mtags .

PR #3814, opened 2022-04-11
fix: Register extension parameter symbols by mtags

tanishiking:
fix: https://github.com/scalameta/metals/issues/2627

Register extension parameter symbols for each extension method. For example, `s` has two symbols
- .`..package.asInt().(s)` and
- `...package.double().(s)`

```scala
package example
extension (s: String) {
  def asInt: Int = s.toInt
  def double: String = s * 2
end extension
```

see:
- https://github.com/scalameta/scalameta/issues/2443
- https://github.com/lampepfl/dotty/issues/11690

### Testing
Also, this PR introduces `MtagsScala3Suite` to check how mtags for Scala3 work.
`MtagsScala3Suite` doesn't compare the mtags-generated symbols and scalac-generated symbols (which should be done in the future) at this moment, and it doesn't check the `unkownSymbols`, because we still have a lot to do in mtags for scala3.
It should be done in another PR, I will create another issue.

At least for the Extension parameter, I could confirm this change resolves the issue by checking: https://github.com/scalameta/metals/pull/3814/files#diff-a09cd855c6e8396ab0ad0ec025d94a68729943b623cbb3f519110e8e248c8a06R6

---

Out of curiosity, does anybody know how this change on `mtags` improves Metas? (which feature depends on `Mtags.index` instead of `Mtags.toplevels` ?)
I tested locally by the following change, and Metals was still able to find references of extension parameters. 

```diff
diff --git a/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala b/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
index 8fae54fbf..d570a3571 100644
--- a/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
@@ -186,11 +186,7 @@ final class DefinitionProvider(
     // Find matching symbol occurrence in SemanticDB snapshot
     val occurrence = for {
       queryPosition <- snapshotPosition.toPosition(dirtyPosition)
-      occurrence <-
-        snapshot.occurrences
-          .find(_.encloses(queryPosition, true))
-          // In case of macros we might need to get the postion from the presentation compiler
-          .orElse(fromMtags(source, queryPosition))
+      occurrence <- fromMtags(source, queryPosition)
     } yield occurrence
 
     ResolvedSymbolOccurrence(sourceDistance, occurrence)
@@ -229,7 +225,7 @@ final class DefinitionProvider(
 
   private def fromMtags(source: AbsolutePath, dirtyPos: Position) = {
     Mtags
-      .allToplevels(source.toInput, scalaVersionSelector.getDialect(source))
+      .index(source.toInput, scalaVersionSelector.getDialect(source))
       .occurrences
       .find(_.encloses(dirtyPos))
   }
```


mtags/src/main/scala/scala/meta/internal/mtags/ScalaMtags.scala
```diff
@@ -283,37 +285,62 @@ class ScalaMtags(val input: Input.VirtualFile, dialect: Dialect)
...
-          // need to wait for https://github.com/lampepfl/dotty/issues/11690
+          val extensionParamss = t.paramss.flatten.map(p => p.name)
           val (owner, overloads) =
             if (isPackageOwner)
               topleveSourceData
             else
               (currentOwner, new OverloadDisambiguator())
 
-          def addDefnDef(t: Defn.Def): Unit =
+          def addDefnDef(t: Defn.Def): Unit = {
```
dos65, line 295: I'm wondering if it might be fixed with concatenation of def.params with exntesions params.
Smth like:
```scala
case  Defn.ExtensionGroup =>
   val extensionParamss = t.paramss.flatten.map(p => p.name)
   
   def addDefnDef(t: Defn.Def): Unit = {
      disambiguatedMethod(
        t,
        t.name,
        Nil,
        t.params ++ extensionParamss, // register both def and extension params
        Kind.CONSTRUCTOR,
        overloads
      )
  }
  ...
```

tanishiking:
Realized there's MtagsSuite, I'll add a test there.

mtags/src/main/scala/scala/meta/internal/mtags/ScalaMtags.scala
tanishiking: Works well :+1: https://github.com/scalameta/metals/pull/3814/commits/efd3b5952ea72032e9292decb24fae8d23631ad2

tests/unit/src/test/scala/tests/MtagsSuite.scala
```diff
@@ -13,35 +14,34 @@ import scala.meta.internal.semanticdb.Scala._
...
             case occ if symtab.info(occ.symbol).isEmpty =>
               val pos = input.toPosition(occ)
               pos.formatMessage("error", s"unknown symbol: ${occ.symbol}")
           }
-          if (unknownSymbols.nonEmpty) {
+          if (!ignoreUnknwonSymbols && unknownSymbols.nonEmpty) {
             fail(unknownSymbols.mkString("\n"))
           }
-          if (file.isScala && !hasSemanticdbBug(file)) {
+          if (file.isScala && !ignoreAssertion && !exclude(file)) {
```
dos65, line 44: `ignoreAssertion` leads to that MtagsScala3Suite doesn't perform any check.

dos65 (changes requested):

tests/unit/src/test/scala/tests/MtagsSuite.scala
tanishiking: Yes, if `ignoreAssertion = true` MtagsSuite doesn't do any checks, just generates a result into the designated directory.

tests/unit/src/test/scala/tests/MtagsSuite.scala
```diff
@@ -63,3 +63,33 @@ class MtagsSuite extends DirectoryExpectSuite("mtags") {
...
 }
+object MtagsSuite {
+  def hasScala2SemanticdbBug(file: InputFile): Boolean = {
+    // don't assert fidelity where semanticdb-scalac has known bugs and mtags is correct.
+    List(
+      "ImplicitClasses",
+      "PatternMatching",
+      "ImplicitConversions",
+      "MacroAnnotation"
+    ).exists { name => file.file.toNIO.endsWith(s"$name.scala") }
```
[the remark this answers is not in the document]
tanishiking: :+1: done https://github.com/scalameta/metals/pull/3814/commits/a1110c811c2ec73f473f9ffd2dd780fd21cabc30

tests/unit/src/test/scala/tests/MtagsSuite.scala
tanishiking: > I mean, if there are no checks what is the purpose of this test?
> I think it would be better tot turn it into an issue and introduce MtagsScala3Suite later.


I will turn it into an issue, and make `MtagsScala3Suite` actually works later (in another PR).
Indeed, `MtagsScala3Suite` doesn't have any checks, and looks like useless, I believe generating expected files will be helpful for us to check how mtags (for Scala3) generate `SymbolOccurrence`s.
With these generated files, we can see how they're different from scalac-semanticdb generated semanticdb, and tell what's the issue.

dos65: Ok, now I see the point. Here we declare how it works now it be able to spot issue if we break ScalaMtags  in future but skip the one-to-one comparison with semanticdb.

So, then maybe simplify it a bit.
Instead of introducing `ignoreAssertion` parameter we can use `exclude`:
```scala
class MtagsScala3Suite
    extends MtagsSuite(
      InputProperties.scala3(),
      "mtags-scala3",
      dialects.Scala3,
      _ => true, // exclude all
      // Do not assert unknown symbols and compare to semanticdb,
      // There're still lot to improve in mtags for Scala3
      ignoreUnknownSymbols = true,
    )
```
I would also rename `exclude` to some other name (`compareWithSemanticdb` ?) to express that it isn't fully excluded but just doesn't equal to semanticdb.

dos65 (approved):
LGTM!

PR #3818, opened 2022-04-12
bloop - bump version

dos65:


tgodzik (approved):
I will check my reproduction scenario and report back!

tgodzik:
Works! Great work @dos65 !!!

ckipp01:
> I will check my reproduction scenario and report back!

Just a curiosity, it sounds like this bump was fixing a specific issue? If so, what issue, and in the future can we include that info in the commit? It might be helpful for people that come across this later on or are trying to debug where something was introduced 😅 .

dos65:
@ckipp01 it fixes an issue with changes detection but only for recent versions of bloop which comes with original zinc instead of fork. Metals wasn't released with this version so this fix fall into category: `upgraded bloop which doesn't use specific zinc fork anymore`

PR #3820, opened 2022-04-12
Draft for Braceless bracefull switch code action

zmerr:


tanishiking:
Thanks for working on this, @zmerr ! Adding/removing braces seems pretty useful!

I guess this comes from the discussion in https://github.com/scalameta/metals/pull/3748 
According to the original PR, it seems like `onTypeFormatting` is a good option for this feature (and I agree with it). https://github.com/scalameta/metals/pull/3748#issuecomment-1076577517
Did you consider working on it with `onTypeFormatting` or found any difficulties/flakiness? It would be really awesome if it's available with `onTypeFormatting` IMO 😄
(Maybe removing braces is hard to achieve with `onTypeFormatting` though).

BTW, I would recommend you to open an issue on `scalameta/metals` or `scalameta/metals-feature-request` before working on coding, to make many users on the same page about what we want to achieve and how can we resolve the issue. (Of course, making a draft PR is also a great way to kick off the discussion :+1:) It will help us avoid some reworks :)

tgodzik:
> Thanks for working on this, @zmerr ! Adding/removing braces seems pretty useful!
> 
> I guess this comes from the discussion in #3748 According to the original PR, it seems like `onTypeFormatting` is a good option for this feature (and I agree with it). [#3748 (comment)](https://github.com/scalameta/metals/pull/3748#issuecomment-1076577517) Did you consider working on it with `onTypeFormatting` or found any difficulties/flakiness? It would be really awesome if it's available with `onTypeFormatting` IMO smile (Maybe removing braces is hard to achieve with `onTypeFormatting` though).
> 

I think both code action and using onTypeFormatting might be useful for users. With code action you would only need to invoke it while in a single line definition, while typing `{` would require going to the front. I am also not sure if we will end up with a good enough solution for `onTypeFormatting` (the other PR still needs more work)

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulFormMap.scala
```diff
@@ -0,0 +1,36 @@
+package scala.meta.internal.metals.codeactions
```
tgodzik, line 1: Looks like these enums are not needed?

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
```diff
@@ -0,0 +1,241 @@
...
+        val startTextEdit = new TextEdit(
+          new l.Range(braceableBranchStart, braceableBranchStart),
+          bracelessStart
+        )
+        val endTextEdit = new TextEdit(
+          new l.Range(braceableBranchEnd, braceableBranchEnd),
+          bracelessEnd
+        )
+        val codeAction = new l.CodeAction()
+        codeAction.setTitle(BracelessBracefulSwitchCodeAction.goBraceless)
```
tgodzik, line 47: We should probably make sure that the indentation is correct in this case,

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
```diff
@@ -0,0 +1,241 @@
...
+
+    val result =
+      for {
+        tree <- maybeTree
+        document <- buffers.get(path)
+      } yield {
+        tree match {
+          case _: Pkg => None
+          case classDefn: Defn.Class =>
+            createCodeActionForBraceableTree(
```
tgodzik, line 114: We could probably reduce a number of cases here, but we can wait until it's out of draft.

tgodzik (commented):
Looks good! I added some intial comments, for sure we will need a number of tests. At least for:
- adding braces in class with `:`
- removing braces for class with `{` to `:`
- not removing braces for Scala 2
- adding braces for def
- removing braces for single statement in def

It will probably be easier to create coupld of LSP tests, but cover everything else with mock `trees` and `buffers` similar to https://github.com/scalameta/metals/blob/main/tests/unit/src/main/scala/tests/BaseClassFinderSuite.scala

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulFormMap.scala
zmerr: No. I need to remove them

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: Yes, the problem is to get a hold of the value name to pass it to the  function…

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: This PR absolutely needs the auto-indentation functionality for getting it right!

zmerr:
@dos65 @kpodsiad Can you please lend a pair of eyes in here? This code action is inserting instead of overwriting when putting the braces or removing them! I can’t find what is causing it to behave in this way!

kpodsiad:
> This code action is inserting instead of overwriting when putting the braces or removing them! I can’t find what is causing it to behave in this way!

@zmerr without looking at the code I can tell you that this is related to the ranges of `TextEdit`. The behavior of edit (insert vs replace) depends on the range - via the javadoc: `The range of the text document to be manipulated. To insert text into a document create a range where start === end.`

I'll try to check what's causing the issue but I wanted to share with you this hint before I deep dive into your changes.

tests/slow/src/test/scala/tests/feature/BracingSwitchCodeActionLspSuite.scala
```diff
@@ -0,0 +1,108 @@
...
+  check(
+    "braceful-to-braceless-object",
+    """|object Ma<<>>in {
+       |  def method2(i: Int) = ???
+       |}
+       |
+       |class A
+       |""".stripMargin,
+    s"""|${ExtractRenameMember.title("object", "Main")}
+        |${BracelessBracefulSwitchCodeAction.goBraceless}""".stripMargin,
```
kpodsiad, line 24: Because `BracelessBracefulSwitchCodeAction` is listed as a second one you have to specify `selectedActionIndex` for the `check` method. In your case it is `selectedActionIndex = 1`

zmerr: Thank you so much :)

zmerr: It was caused by not copying `l.Position` through new instance creating and assigning the right character and line numbers. I was instead just assigning the old Position to a new val, which was pass by reference. So when I was editing the character of the end position for example, that of the start position was also getting modified as a side effect.

zmerr:
TODO: refactoring to remove repetitive code.
TODO: Tests

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
```diff
@@ -0,0 +1,558 @@
...
+          case defDefn: Defn.Def =>
+            createCodeActionForPotentialBlockHolder(
+              defDefn,
+              path,
+              document,
+              defDefn.body,
+              "def definition"
+            ).toSeq
+          case termTry: Term.Try =>
+            Seq(
```
tgodzik, line 111: Could we only show one action based on the current position to avoid clutter? IF we are at a Try in the block or keyword then show the actions for Try, the same for catch and finally

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
```diff
@@ -0,0 +1,558 @@
...
+                  finallyp,
+                  path,
+                  document,
+                  finallyp,
+                  "finally expression"
+                )
+              )
+            ).flatten
+          case caseTree: Case =>
+            createCodeActionForCasesListWrapper(caseTree, path).toSeq
```
tgodzik, line 136: We should only show dropping braces in this case. Though it would be awesome to show it based on some linting information about redundant braces. This could be done via scalafix.

Anyone, for this I would only show an action to drop braces, but not to add them

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
```diff
@@ -0,0 +1,558 @@
...
+          case termIf: Term.If =>
+            Seq(
+              createCodeActionForPotentialBlockHolder(
```
tgodzik, line 139: Same as with Try, could we only show the most relevant action for the current position?

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
```diff
@@ -0,0 +1,558 @@
...
+      else None
+
+    val result =
+      for {
+        tree <- maybeTree
+        document <- buffers.get(path)
+      } yield {
+        tree match {
+          case _: Pkg => None.toSeq
+          case classDefn: Defn.Class =>
```
tgodzik, line 52: Can't we use Template for this and the ones below?

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
```diff
@@ -0,0 +1,558 @@
...
+              "block"
+            ).toSeq
+        }
+      }
+    result.toSeq.flatten
+  }
+
+  private def hasBraces(template: Template): Boolean = {
+    util
+      .Try(template.stats.maxBy(_.pos.end).pos.end)
```
tgodzik, line 203: Could we just check text(template.pos.end) == '}'?

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
```diff
@@ -0,0 +1,558 @@
...
+      .getOrElse(-1) != template.pos.end && util
+      .Try(template.tokens.maxBy(_.pos.end).text)
+      .getOrElse("") == "}"
+  }
+
+  private def hasBraces(termBlock: Term.Block): Boolean = {
+    util
```
tgodzik, line 210: `text(termBlock.pos.start)` should always point at '{'

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
```diff
@@ -0,0 +1,558 @@
...
+   *         argument has braces
+   */
+  private def isBlockEmbraceableBraced(blockEmbraceable: Term): Boolean = {
+    blockEmbraceable match {
+      case termBlock: Term.Block => hasBraces(termBlock)
+      case _ => false
+    }
+  }
+
+  private def createCodeActionForCatchP(
```
tgodzik, line 230: I wonder if it would be easier to add the code actions incrementally. Start with Template and Block, work on other trees in next PRs. It would be easier to review and much more focused.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
```diff
@@ -0,0 +1,558 @@
...
+            ).toSeq
+          case valDefn: Defn.Val =>
+            createCodeActionForPotentialBlockHolder(
+              valDefn,
+              path,
+              document,
+              valDefn.rhs,
+              "val definition"
+            ).toSeq
+          case varDefn: Defn.Var =>
```
tgodzik, line 92: We should be able to detect if we are at a Block instead of having a separate case for each.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
```diff
@@ -0,0 +1,558 @@
...
+
+  override def contribute(params: CodeActionParams, token: CancelToken)(implicit
+      ec: ExecutionContext
+  ): Future[Seq[l.CodeAction]] = Future {
+    val path = params.getTextDocument().getUri().toAbsolutePath
+    val range = params.getRange()
+    val maybeTree =
+      if (range.getStart == range.getEnd)
+        trees
+          .findLastEnclosingAt[Tree](
```
tgodzik, line 38:
```suggestion
          .findLastEnclosingAt[Template](
```

and in the next step:

```
```suggestion
          .findLastEnclosingAt[Block](
```

so this way we will no need to do all the cases for val/var etc.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
```diff
@@ -0,0 +1,558 @@
...
+      getIndentationForPositionInDocument(templateHolder.pos, document)
+
+    util.Try(templ.inits.maxBy(_.pos.end).pos.end).toOption
+    val expectedBraceStartPos = util
+      .Try {
+        val lastInit = templ.inits.maxBy(init => init.pos.end)
+        lastInit.pos.end
+      }
+      .getOrElse(templ.pos.start)
+    pprint.log("templ.inits: \n" + templ.inits)
```
tgodzik, line 376: pprint is not available on CI, so this might break the compilation

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: No. only class/enum/trait/object are template holders. Also, the code action should include the type it is removing/adding braces to/from in the title. Otherwise, the braces would be added to or removed from the wrong subject. I have another idea for refactoring the redundant code, which I would do at the end. It is based on just passing the right subtree arguments and title.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: I have tried this. It creates bugs. Because not at all times the potential block holders have a block at the position of their relevant subtree.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: I tried that. Unfortunately, there is not a straightforward way for knowing which keyword of try, catch, or finally we are at! It just captures the whole try-catch-finally tree. You can inspect this on the ast-explorer

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: Oh, this one is a TODO

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: there is not a straightforward way to know whether we are at then/else branches. It only captures the whole if-else tree. Any suggestions for capturing where we are at the tree are welcome

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: No. This creates a bug. Let’s say you have a braceless Template with a braced method at the end. The last token would be `}` but the Template is braceless.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: Oh, I have already covered most of them. Now the common problem in all of them is the insertion point of the open brace, and also the auto indentation before going braceless, which is super important. No one of them can be merged in isolation because they all miss the same things.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: Sure. Just for debugging.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: aha, No this would create another bug in Scala 3. Because the open brace might be for one of the stats rather Than the whole block. So always it should be checked that the open brace or the closing brace is not part of one of the stats.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
tgodzik: If we look for Blocks themselves this would not be a problem.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
tgodzik: We can check the parent of the template for the title and have a function for that.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
tgodzik: I wanted to let you know, since the compilation issue is not obvious :sweat_smile:

zmerr: Ok, thank you so much!

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: Yes! Good idea actually! But then no code action would appear when the cursor is on the class name for example. That makes it super counter intuitive!

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: Then no code action would appear when the cursor is on the variable name.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: So just for making the code shorter, this would break a functionality! Not good!

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: I think the code action for adding braces to a block or removing them, should be removed. Because in some cases such as the block of code passed as the argument of `flatMap` the braces are mandatory! Or perhaps, just in those cases when the parent is a collection method, the braces should stay.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
tgodzik: Actually, I think the last char in that case is `\n` from what I checked:

```
import scala.meta._

val code = """|object A extends B:
              |  def b = {
              |
              |  }
              |
              |""".stripMargin
implicit val dl = scala.meta.dialects.Scala3

val parsed = code.parse[Source].get

val tmpl = parsed.stats.head.asInstanceOf[Defn.Object].templ

code(tmpl.pos.end)
```

zmerr: Told you! it’s a mine field! No one else should touch this PR! haha

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
tgodzik: I would just check if we are closer to thenp or elsep statement.

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
tgodzik: for flatMap it's not a problem since it's in Apply not in Block.

> Sometimes you want to add braces even when we do not have a block! for example when there is a very long multiple line chain of map/flatMap functions and such

For that case yeah, it's not 100% sure where to show the code action. And in some places there will be no Block :thinking:  Might be good to leave as is for those case, but maybe we should only show the action if the code is a bit longer than a couple of charaters? Some kind of threshold so not to always show it?

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
tgodzik: So maybe we should check if the last is `}` and if that doesn't belong to the last statement?

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
tgodzik: Some as above then. We should also check if the `{` doesn't belong to the first stat

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: What I’ve written does exactly that!

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: Well, the user does not have to click on it every time! There are many other things that show up but the users just don’t click on!

zmerr: And if somebody clicks on it, the whole hassle is just an undo!

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: Maybe!

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: Could be! I would check! Thank you!

metals/src/main/scala/scala/meta/internal/metals/codeactions/BracelessBracefulSwitchCodeAction.scala
zmerr: I was thinking that we can check the number of lines threshold for deciding whether to put an `EndMarker` in place when going braceless

tgodzik:
> @tgodzik is the threshold of 20 lines good for inserting an end marker when going braceless?

Maybe let's leave it up to scalafmt, since people will have different ideas about it and scalafmt already has a proper config option.

zmerr:
As conclusion to discussion with @tgodzik this is going to be done for  auto-indenting the block before going braceless, for preventing a change in semantics in the resulted braceless code:

- calling the `.format` function of `scalafmt` on the file for getting the formatted `String` of the file, in the return!
- calling `.parse` on the `String` for obtaining the tree structure of the file
- searching for the relevant block, removing the braces, and calling `.toString` on it for getting the final result!
- using the resulted string for constructing a TextEdit and, ultimately` a code action using that!
- the title of the code action should be: “auto indent the block and remove braces"

zmerr:
TODO: refactor the bracing logic!
TODO: use default scalafmt config for auto indentation when a config file not present

zmerr:
Since the removal of braces from incorrectly indented code causes semantic change to it, this is the summary of the latest points discussed and considered :

- The current code scissors the tree from which the braces are to be removed, or its parent, when it is present. Then formats it to Scalafmt. Then parses the formatting result, removes braces from it, and patches the result back in the position of the scissored code.
- this causes the patched code, to be indented with zero white space, which in most cases is wrong

The possible solutions are:

- as proposed by @tgodzik to check the indentation of the code we are removing braces from, and only remove them when the indentation is correct. When completed, for which Tomasz wrote a draft code, it can also be used in cases when the Scalafmt config file is absent. 
- format the whole file with scalafmt and then find the formatted version of the tree we want to remove braces from in the hierarchy. The challenge is that formatting can sometimes change the parse tree by the addition or removal of end markers or the addition or removal of braces. There is also uncertainty, in this direction, in figuring out all the possible ways Scalafmt can affect the parse tree.
- format the scissored part of code containing the braces and then append the original indentation of the first line in the unformatted code, to each subsequent line in the formatted patch.

tgodzik:
> * as proposed by @tgodzik to check the indentation of the code we are removing braces from, and only remove them when the indentation is correct. When completed, for which Tomasz wrote a draft code, it can also be used in cases when the Scalafmt config file is absent.

I think this one carries least risk, but if any of the other solution end up working better I am also ok with that.

zmerr:
Summary of the technical conclusions in todays standup:

- Since the execution of Scalafmt each time the cursor moves in the file for the production of the Code Action Text Edits, can potentially slow down metals, it was decided that the formatting and the calculation of the edits get transferred to the code action command, instead.
- In executing the code action command for removing barces, Metals should 

1.  format the whole file,
2. then acquire the desired tree by re-invocation of `findlastClosingAt` according to the cursor,
3. remove the braces and add the edits to the codeAction command, then apply it on the formatted file.

zmerr:
The latest demo:

https://user-images.githubusercontent.com/93259201/169501650-3baa7daa-7ef1-4370-a822-a0ef26edd566.mov

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
```diff
@@ -133,8 +140,53 @@ final class FormattingProvider(
...
+   * if the configuration of Scalafmt for the project is properly done,
+   * this method is going to produce a List of a single TextEdit for the
+   * formatted string of the whole file, regardless of whether the file is
+   * already well formatted or not.
+   * @param path the path to the file we want to get the formatting [[TextEdit]]s for
+   */
+  def programmaticallyFormat(
+      path: AbsolutePath,
+      token: CancelChecker
+  ): Future[util.List[l.TextEdit]] = {
```
tgodzik, line 153: Why would you need to return something always irrelevant of whether the file was formatter?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1932,6 +1939,83 @@ class MetalsLanguageServer(
...
+      case ServerCommands.FormatThenRemoveBraces(textDocumentPositionParams) =>
+        CancelTokens.future { token =>
+          val uri = textDocumentPositionParams.getTextDocument().getUri()
+          val path = uri.toAbsolutePath
+          val initialFileCode = path.toInputFromBuffers(buffers).text
+
+          formattingProvider
+            .programmaticallyFormat(path, token)
+            .map(formattingEditsList =>
+              if (!formattingEditsList.isEmpty) {
```
tgodzik, line 1951: It's fine if it's empty.

metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala
zmerr: This way,  only when the Scalafmt config file is not present, we would receive an empty result.
Also, if the file is not changed, still, we need the result to feed into the rest of the processing for removing the braces.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
zmerr: the method `programmaticallyFormat` returns empty if and only if scalafmt is not properly configured.

tgodzik: Why just not check if the file exists

zmerr: I thought that checking whether scalafmt successfully finishes execution on the file, is a more certain way for making sure it is properly configured. And even if the file does not have any formatting problem, we still need the Scalafmt result for being able to conclude that..

zmerr: The logic works this way: we run scalafmt on the file, then continue the processing with that result regardless of whether it is different to the input or not.
If the scalafmt cannot produce a result, we do not continue the processing.

zmerr:
The latest update: The functionality of the PR is working correctly in manual demo tests. But, all the braceless tests are failing, without any edits being applied.
If anyone has time to review the PR, that would be appreciated. 
Otherwise, this PR is going to get dropped and in the future reintroduced in smaller PRs for getting a better review opportunity.

tgodzik:
Closing for now since we need some smaller steps like https://github.com/scalameta/metals/pull/3963

PR #3825, opened 2022-04-13
fix: unblock release workflow

dos65:
Currently, github actions have some issues with filtering `runs`.
`gh api/repos/scalameta/metals/actions/workflows/release.yml/runs?status=in_progress` doesn't return any results.
There are several issues about that in github community forum - https://github.community/t/bug-we-are-having-problems-searching-workflow-runs-the-results-may-not-be-complete/244650

The only option for now is disable this lock logic.


tanishiking (approved):
LGTM :cry:

PR #3828, opened 2022-04-14
fix: show test explorer misconfigurations

kpodsiad:
Take scala version suffix in munit.
Do not show warning for sbt targets.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
@@ -227,7 +227,7 @@ class ProblemResolver(
 
     // 1.0.0-M3 or higher is valid
     def outdatedMunitInterface =
-      if (!isTestExplorerProvider()) None
+      if (!isTestExplorerProvider() || scalaTarget.isSbt) None
       else {
         def isInvalid(
             major: Int,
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
```diff
@@ -227,7 +227,7 @@ class ProblemResolver(
 
     // 1.0.0-M3 or higher is valid
     def outdatedMunitInterface =
-      if (!isTestExplorerProvider()) None
+      if (!isTestExplorerProvider() || scalaTarget.isSbt) None
```
ckipp01, line 230: This is a great change. I've actually had a few people ask about this since they think something is wrong when they see warnings about not being able to debug targets when it's just an sbt target anyways.

ckipp01 (approved):
LGTM!

metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
kpodsiad: I think I've only seen people with JUnit problem but I was too lazy to check what was the cause and I took the easier path to add this to the condition. It still bothers me, but not that much to spend time on it. it's better to do sth more useful :D

PR #3829, opened 2022-04-14
Backtick types when neccessary

tgodzik:
This is backported from https://github.com/scalacenter/scalafix/pull/1578/files

Scalafix rule for explicit result types started of as a copy of Metals one.

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
@@ -178,6 +178,35 @@ class MetalsGlobal(
       this(string + ".", string)
   }
 
+  private def backtickify(
+      tpe: Type,
+      loop: (Type, Option[ShortName]) => Type,
+      withPrefix: Boolean
+  ): Type =
+    tpe match {
+      case TypeRef(pre, sym, args)
+          if Identifier.needsBacktick(sym.decodedName) &&
+            sym != definitions.ByNameParamClass && // `=> T` is OK
+            !sym.isPackageObject &&
+            !definitions.isRepeated(sym) &&
+            !sym.isAbstractType &&
+            sym.owner != definitions.ScalaPackageClass =>
+        val rawPrettyName = Identifier.backtickWrap(sym.decodedName)
+        val prefix = if (withPrefix) pre.prefixString else ""
+        if (args.isEmpty) {
+          new PrettyType(prefix + rawPrettyName)
+        } else {
+          val typeArgsStr =
+            args.map(arg => loop(arg, None)).map(_.toString()).mkString(", ")
+          new PrettyType(prefix + s"$rawPrettyName[$typeArgsStr]")
+        }
+      case TypeRef(p, sym, args) =>
+        val pre = if (withPrefix) p else NoPrefix
+        TypeRef(pre, sym, args.map(arg => loop(arg, None)))
+      case _ =>
+        tpe
+    }
+
   /**
    * Shortens fully qualified package prefixes to make type signatures easier to read.
    *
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala
```diff
@@ -178,6 +178,35 @@ class MetalsGlobal(
...
+      loop: (Type, Option[ShortName]) => Type,
+      withPrefix: Boolean
+  ): Type =
+    tpe match {
+      case TypeRef(pre, sym, args)
+          if Identifier.needsBacktick(sym.decodedName) &&
+            sym != definitions.ByNameParamClass && // `=> T` is OK
+            !sym.isPackageObject &&
+            !definitions.isRepeated(sym) &&
+            !sym.isAbstractType &&
```
dos65, line 192: It seems that this approach isn't reliable.
`sym.decodedName` is different from the string representation of type.

Instead of checking the whole value we need to check each part of type. But I understand that it might be complicated.

Maybe as alternative to this, we can use scalameta parser?
I mean, we can turn type into string but instead passing it directly into textEdit use scalameta parser and then reconstruct it and applying backticks?

tests/cross/src/test/scala/tests/pc/InsertInferredTypeSuite.scala
```diff
@@ -511,6 +511,53 @@ class InsertInferredTypeSuite extends BaseCodeActionSuite {
...
+       |  val <<foo>> = A.`Foo-Foo`(1)
+       |}""".stripMargin,
+    """|object A{
+       |  case class `Foo-Foo`[A](i: A)
+       |}
+       |object O{
+       |  val foo: A.`Foo-Foo`[Int] = A.`Foo-Foo`(1)
+       |}
+       |""".stripMargin
+  )
```
dos65, line 559: test case that doesn't work with current approach
```scala
  checkEdit(
    "backticks-7".tag(IgnoreScala3).only,
    """|object A{
       |  class `x-x`()
       |  case class Foo[A](i: A)
       |}
       |object O{
       |  val <<foo>> = A.Foo(new A.`x-x`())
       |}""".stripMargin,
    """|object A{
       |  class `x-x`
       |  case class Foo[A](i: A)
       |}
       |object O{
       |  val foo: A.Foo[`x-x`] = A.Foo(new A.`x-x`)
       |}
       |""".stripMargin
  )
```

ckipp01:
Marking this as a draft for now just to separate it from the active prs.

tgodzik:
If anyone else have time and wants to take a look, let me know. I added this to the issues board (could be transformed into an issue if needed.)

tests/cross/src/test/scala/tests/pc/InsertInferredTypeSuite.scala
kasiaMarek: I started looking at this PR and it's lost on what is wrong with the current approach. I played around with this a bit and seems to work fine for me.
This test also passes:
```
  checkEdit(
    "backticks-7".tag(IgnoreScala3).only,
    """|object A{
       |  class `x-x`
       |  case class Foo[A](i: A)
       |}
       |object O{
       |  val <<foo>> = A.Foo(new A.`x-x`)
       |}""".stripMargin,
    """|object A{
       |  class `x-x`
       |  case class Foo[A](i: A)
       |}
       |object O{
       |  val foo: A.Foo[A.`x-x`] = A.Foo(new A.`x-x`)
       |}
       |""".stripMargin
  )
  ```
Do you have any case when `sym.decodedName` is different than the string?
What type was inferred in this test for you?

dos65: hmm. Honestly, the only I remember is that I got curious by these clauses - https://github.com/scalameta/metals/blob/c1d7984b5e8de705177743182bf5c714a01abbb2/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala#L190-L193

Maybe tests might spot some issues in approach if you remove them

If you won't find anything, feel free to merge this PR

PR #3843, opened 2022-04-17
Add semantic db info display on jar files

Arthurm1:
Displays semantic db info on `java` and `scala` jar files using the internal interactive db info.  Jar files don't have a semanticdb file.

I've found this helpful when debugging goto/ref information on jar files.


dos65 (commented):
@Arthurm1 Looks good!
The only thing I'm not sure is new extensions `semanticdb-internal-*`.
Why we can use existing `semanticdb-compact` and others for jars?

Arthurm1:
@dos65 Good point.  In fact I could just always check the interactive ones first and then fallback onto the file ones if interactive doesn't exist.  The only time both file and interactive seem to exist is when there is an unsaved source file and we can't currently show the interactive ones for that because it involves passing the content as well as the uri.
I'll change it to do that.

Arthurm1:
Should be good now

tgodzik (approved):
LGTM

PR #3846, opened 2022-04-19
Fix breakpoints in virtual docs

kpodsiad:


metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathAdapter.scala
```diff
@@ -17,12 +18,26 @@ private[debug] final class SourcePathAdapter(
...
+      val jarPath = workspace.toNIO.relativize(sourcePath.toNIO)
+      val x = jarPath.toString
+        .stripPrefix("jar:")
+        .replace("%3A", ":")
+        .replace("%21", "")
+      pprint.log(x)
+      val uri = new URI(x)
+
+      pprint.log(uri)
+      Some(uri)
```
kpodsiad, line 39: I tried here multiple approaches but none of them worked. I'm still trying to figure out how Metals should map virtual doc's URI to the DAP one.
If I'm correct then [here](https://github.com/microsoft/java-debug/blob/625fe84cef4825a4b6b476d04383f5168e81708d/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java#L284) `ISourceLookUpProvider` is used to convert given uri into a fqcn. But I cannot infer what kind of URI [`SourceLookUpProvider`](https://github.com/scalacenter/scala-debug-adapter/blob/53a062a71c280da15ccaf1cc1d365a03be8bdaa9/core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceLookUpProvider.scala#L42) expects. Can I ask for some clue @adpi2 ?

kpodsiad: Previously it was easier because Metals were copying source files from jars to the `.metals/dependecies` directory but now we don't do that anymore (if the editor supports virtual docs) so here we should create a URI which will point to the file inside a JAR.

kpodsiad: Actually nevermind, I found what I have to do.

Arthurm1: @kpodsiad I've updated the `URIEncoderDecoder` [here](https://github.com/Arthurm1/metals/tree/uri_decoding)
You should be able to grab that and run `URIEncoderDecoder.decode` on the `jar:file%3A` uri rather than picking out specific chars.
It's VSCode that's encoding the uri this way as it will encode any `:` after the scheme `:`.
Java should be able to open a jar using a uri of the format `jar:file:///somepath/somejar.jar!/somepackage/somefile.java` which (hopefully) is what you will get from that method.  So I'd expect/hope that the debugger will cope with that uri.
Of course the debugger might not understand `jar` uris at all because before virtual docs, all `jar` files were saved in `readonly/jarName/package/class` and I don't know where the code that handles that is.

kpodsiad: > I've updated the URIEncoderDecoder [here](https://github.com/Arthurm1/metals/tree/uri_decoding)
You should be able to grab that and run URIEncoderDecoder.decode on the jar:file%3A uri rather than picking out specific chars.

Thanks, that might be helpful!

> Of course the debugger might not understand jar uris at all because before virtual docs, all jar files were saved in readonly/jarName/package/class and I don't know where the code that handles that is.

Funny thing, the code is right below my "changes". I switched back to the old Metals server version and it turns out that debugger understands only `jar` uris. But now I can see what should I do so this PR should be relatively easy to finish.

adpi2: > it turns out that debugger understands only jar uris.

Yep, that's correct :slightly_smiling_face:

metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathAdapter.scala
```diff
@@ -17,12 +19,36 @@ private[debug] final class SourcePathAdapter(
...
+        URIEncoderDecoder.decode(sourceFileJarPath).toAbsolutePath
+      val parts = decodedPath.toNIO.iterator().asScala.map(_.toString).toVector
+      val jarPartIndex = parts.indexWhere(path => path.endsWith(".jar!"))
+      val jarPath = AbsolutePath(
+        parts
+          .take(jarPartIndex + 1)
+          .mkString(File.separator, File.separator, "")
+          .stripSuffix("!")
+      )
+      val relative = parts.drop(jarPartIndex + 1).mkString(File.separator)
```
kpodsiad, line 45: Honestly, this code is disgusting, but I don't have any other idea how to turn path like
```scala
AbsolutePath(
  toNIO = /Users/kpodsiadlo/vl/my-app/jar:file%3A/Users/kpodsiadlo/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.8/scala-library-2.13.8-sources.jar%21/scala/collection/immutable/List.scala
)
```
into base jar path + relative counterpart

Arthurm1: I'm hoping to get rid of the jar encoding when I finally get the filesystem api working so I wouldn't spend much time on making it pretty.

tgodzik: As long as it works it's fine, it's not that bad :sweat_smile:

kpodsiad: Ok, so if it's not that bad and it's a huge chance that it's a temporary workaround I can leave it as it is now.

mtags/src/main/scala-2/scala/meta/internal/metals/URIEncoderDecoder.scala
```diff
@@ -2,14 +2,16 @@ package scala.meta.internal.mtags
...
   private val toEscape: Map[Char, String] =
-    Set('"', '<', '>', '&', '\'', '[', ']', '{', '}', ' ')
+    Set('"', '<', '>', '&', '\'', '[', ']', '{', '}', ' ', '+', '!')
       .map(char => char -> ("%" + char.toHexString))
       .toMap
 
   private val toDecode: Map[String, Char] = toEscape.map { case (k, v) =>
     v -> k
-  }
+  } + ("%3a" -> ':')
```
Arthurm1, line 14: @kpodsiad didn't you say `:` results in issues with over/under encoding?

Arthurm1:
This works for me when testing breakpoints in jars locally

metals/src/main/scala/scala/meta/internal/metals/debug/SourcePathAdapter.scala
Arthurm1: I think the bizarre relative path format is down to VSCode only supporting `pathFormat` of `path`, not `uri`.  So all non `file:///` uris are just prepended with the workspace path.  There's an issue [here](https://github.com/microsoft/vscode/issues/68808)

mtags/src/main/scala-2/scala/meta/internal/metals/URIEncoderDecoder.scala
kpodsiad: That was `%` and it was, I believe, in encoding where `%25` was changed into `%2125` or sth similar.

tgodzik (approved):
LGTM! Hopefully we can drop this in the future

kpodsiad:
Better to have this than nothing ;)

PR #3849, opened 2022-04-19
fix: Make signature help work properly with a non-tuple result

tgodzik:
Previously, we would always assume unapply returns a tuple, which would result in wrong signature help for single parameter unapply. Now, we take into account situations where the returned value is not an Option of tuple.

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/SignatureHelpProvider.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/SignatureHelpProvider.scala
@@ -489,12 +489,16 @@ class SignatureHelpProvider(val compiler: MetalsGlobal) {
       case TypeRef(
             _,
             cls,
-            TypeRef(_, _, args) :: Nil
+            TypeRef(_, symbol, args) :: Nil
           )
           if isUnapplyMethod &&
-            (cls == definitions.OptionClass || cls == definitions.SomeClass) &&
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/SignatureHelpProvider.scala
```diff
@@ -489,12 +489,16 @@ class SignatureHelpProvider(val compiler: MetalsGlobal) {
...
-            TypeRef(_, _, args) :: Nil
+            TypeRef(_, symbol, args) :: Nil
           )
           if isUnapplyMethod &&
-            (cls == definitions.OptionClass || cls == definitions.SomeClass) &&
-            args.nonEmpty =>
-        List(args.map(_.typeSymbol))
+            (cls == definitions.OptionClass || cls == definitions.SomeClass) =>
+        // tuple unapply results
+        if (args.nonEmpty && definitions.isTupleSymbol(symbol))
```
tanishiking, line 497: [nitpick] Do we need `args.nonEmpty` ? I believe there's no tuple with no arguments.

```suggestion
        if (definitions.isTupleSymbol(symbol))
```

tgodzik: probably not, I wanted to be too safe :sweat_smile:

PR #3859, opened 2022-04-22
putting the companion object after the end marker

### LINKED ISSUE #3855: the companion object is created before the end marker in the code action
### Describe the bug

https://user-images.githubusercontent.com/93259201/164269839-283eff87-b577-4ead-83f7-c09663c55b7c.mov

### Expected behavior

Putting it after the end marker

### Operating system

_No response_

### Editor/Extension

_No response_

### Version of Metals

v0.11.3-SNAPSHOT

_No response_

zmerr:
resolves #3855

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
@@ -1,10 +1,13 @@
 package scala.meta.internal.metals.codeactions
 
+import scala.collection.mutable.ListBuffer
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 
 import scala.meta.Defn
+import scala.meta.Member
 import scala.meta.Template
+import scala.meta.Term
 import scala.meta.Tree
 import scala.meta.inputs.Position
 import scala.meta.internal.metals.Buffers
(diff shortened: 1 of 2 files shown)


dos65 (commented):
Thanks for the fix!

Looks like it should work but there is a bit easier approach to detect if there is a EndMarker or not. Scalameta Tree has a nice feature, you can take a parent of a tree. Using it you can simplify the logic for searching endMarker. Smth like this should work:

```scala
val tree: Tree = _
val endMarker: Option[Term.EndMarker] =
    t.parent.flatMap { p =>
       val idx = p.children.indexOf(t)
       // checks is the next child in parent after given tree is `EndMarker`
       if (idx + 1 < p.children.length) {
          p.children(idx + 1) match {
             case marker: Term.EndMarker => Some(marker)
             case _ => None
          }
       } else
         None
     }
```

tanishiking:
I was thinking the same thing as @dos65 😄 
`Term.EndMarker` always comes right after the corresponding tree,   https://github.com/scalameta/scalameta/blob/1a9ec6f9aa06e20bc786b5c7ef6d7aa500e59eaf/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/EndMarkerSuite.scala we can simplify the way of looking up `EndMarker` by checking the adjacent tree node :+1:

kpodsiad:
I'll throw my 2 coins here, don't forget about test case for this scenario ;)

zmerr:
> Thanks for the fix!
> 
> Looks like it should work but there is a bit easier approach to detect if there is a EndMarker or not. Scalameta Tree has a nice feature, you can take a parent of a tree. Using it you can simplify the logic for searching endMarker. Smth like this should work:
> 
> ```scala
> val tree: Tree = _
> val endMarker: Option[Term.EndMarker] =
>     t.parent.flatMap { p =>
>        val idx = p.children.indexOf(t)
>        // checks is the next child in parent after given tree is `EndMarker`
>        if (idx + 1 < p.children.length) {
>           p.children(idx + 1) match {
>              case marker: Term.EndMarker => Some(marker)
>              case _ => None
>           }
>        } else
>          None
>      }
> ```

Great approach! thank you so much for that!

zmerr:
@dos65 that was a beautifully clean approach which I tried! Unfortunately, the `indexOf` function does not seem to be workin as expected, producing index out of bounds exception. So I guess we would be stuck with the existing approach :(

dos65:
@zmerr Could you share in which case the proposed solution doesn't work?
I've made [a branch with it](https://github.com/dos65/metals/tree/endmarker_fix) where `tests.codeactions.CompanionObjectSuite` and `tests.feature.Scala3CodeActionLspSuite` work well.

PR #3861, opened 2022-04-22
feature: sbt-BSP - enable java-semanticdb plugin

dos65:


sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
```diff
@@ -44,4 +71,89 @@ object MetalsPlugin extends AutoPlugin {
...
+            s"-Xplugin:semanticdb -sourceroot:${sourceRoot} -targetroot:${targetRoot} -build-tool:sbt"
+          pluginOption :: exportsFlags(version)
+      }
+    }
+  }
+
+  case class JdkVersion(major: Int) {
+
+    def hasJigsaw: Boolean = major >= 9
+  }
```
dos65, line 106: copypasted it from `JavaIteractiveSemanticdb`.
Technically it's possible to extract into a module and share it for `metals` and `sbt-metals` but I'm not sure if it's worth it.

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
```diff
@@ -1,17 +1,29 @@
...
+import sbt._
 import sbt.internal.inc.ScalaInstance
 import sbt.plugins.SemanticdbPlugin
 
 object MetalsPlugin extends AutoPlugin {
   override def requires = SemanticdbPlugin
   override def trigger = allRequirements
 
+  lazy val javaSemanticdbEnabled =
+    settingKey[Boolean]("Enables SemanticDB Javac plugin")
```
adpi2, line 25: Should we move this to the `autoimport` object? So that the user can deactivate it in its `build.sbt`.

adpi2 (commented):
Some suggestions

tgodzik (approved):
Looks good! Thanks! Just one question

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
```diff
@@ -31,6 +43,21 @@ object MetalsPlugin extends AutoPlugin {
...
+    javacOptions ++= {
+      if (bspEnabled.value && javaSemanticdbEnabled.value)
+        javaSemanticdbOptions.value
+      else
+        Nil
+    },
+    allDependencies ++= {
+      if (bspEnabled.value && javaSemanticdbEnabled.value)
+        List(
+          "com.sourcegraph" % "semanticdb-javac" % BuildInfo.javaSemanticdbVersion
```
adpi2, line 57: This should be added to `libraryDependencies` instead of `allDependencies`.

Also, we probably don't want this library to end up in the runtime class path of the application. So we should not put it in the `Compile` scope.

I don't know if the `CompilerPlugin` scope works with javac, probably not but you can try it with:
```scala
"com.sourcegraph" % "semanticdb-javac" % BuildInfo.javaSemanticdbVersion % Configurations.CompilerPlugin
```

Otherwise there is a notion of `CompileInternal` in sbt that should work:
```scala
"com.sourcegraph" % "semanticdb-javac" % BuildInfo.javaSemanticdbVersion % Configurations.CompileInternal
```

tgodzik: Don't we actually need the compiler plugin on the normal classpath in case of Java?

adpi2: We need it in the compile class path but we don't need it in the runtime class path.

dos65: @adpi2 ` Configurations.CompileInternal` works! thx!

metals/src/main/scala/scala/meta/internal/metals/FileSystemSemanticdbs.scala
```diff
@@ -32,9 +32,13 @@ final class FileSystemSemanticdbs(
       val paths = for {
         buildTarget <- buildTargets.inverseSources(file)
         workspace <- buildTargets.workspaceDirectory(buildTarget)
-        targetroot <-
-          if (file.toLanguage.isScala) buildTargets.scalaTargetRoot(buildTarget)
-          else buildTargets.javaTargetRoot(buildTarget)
+        targetroot <- {
+          val nonScala =
+            if (!file.toLanguage.isScala)
```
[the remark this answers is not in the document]
dos65: :smile: 
I changed name.

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
```diff
@@ -31,6 +46,21 @@ object MetalsPlugin extends AutoPlugin {
...
         )
+    },
+    javaSemanticdbEnabled := bspEnabled.value,
+    javacOptions ++= {
+      if (javaSemanticdbEnabled.value)
+        javaSemanticdbOptions.value
+      else
+        Nil
+    },
+    allDependencies ++= {
```
adpi2, line 57: This should go to `libraryDependencies` even if it is not a library. In sbt all the compiler plugins are added to `libraryDependencies` (see [here](https://github.com/sbt/sbt/blob/f12a199a1aa58496ade139adcf8a7a8eba94b968/main/src/main/scala/sbt/Defaults.scala#L4323-L4324))

```suggestion
    libraryDependencies ++= {
```

adpi2 (approved):
I have one more suggestion but otherwise it looks good.

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
dos65: Looks like it works fine with allDependencies.
I took it from [`SemanticdbPlugin`](https://github.com/sbt/sbt/blob/f12a199a1aa58496ade139adcf8a7a8eba94b968/main/src/main/scala/sbt/plugins/SemanticdbPlugin.scala#L37)

adpi2: Well maybe it's done on purpose to "hide" the dependency. So it's fine.

PR #3862, opened 2022-04-22
Properly infer types for unapply signature help

tgodzik:
The only regression seems to be that a couple of test cases changed to `Any`, however this is a proper type and having `T` or `C[T]` probably looks a bit better, but doesn't actually show the real type.


dos65 (approved):
LGTM!

tanishiking (approved):
LGTM:+1:

PR #3863, opened 2022-04-22
handles the removal of jvm properties case

zmerr:
Previously, when a Bloop JVM property was getting removed, if the new properties list were becoming empty, the Bloop settings were not getting updated!
This PR fixes that!


tgodzik (approved):
LGTM! Thanks for the quick fix!

PR #3864, opened 2022-04-25
Use bloop.json instead of .jvmopts for Bloop memory properties

tgodzik:
.jvmopts is now deprecated and cannot be used

also bump Bloop to 1.5.0

### DIFF
--- a/bin/bloop.json
+++ b/bin/bloop.json
@@ -1,0 +1,3 @@
+{
+  "javaOptions": ["-Xmx1G", "-Xss16m"]
+}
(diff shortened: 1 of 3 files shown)


kpodsiad (approved):
> also bump Bloop to 1.4.5

1.5.0 ;)

tgodzik:
I would also like to mention here that I am very happy that I added the warning about `.jvmopts`, because without it I would dig a bunch to figure out why our tests stopped working :sweat_smile:

PR #3866, opened 2022-04-25
fix: create comption object after end marker

### LINKED ISSUE #3855: the companion object is created before the end marker in the code action

dos65:
Fixes #3855

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
@@ -5,6 +5,7 @@ import scala.concurrent.Future
 
 import scala.meta.Defn
 import scala.meta.Template
+import scala.meta.Term
 import scala.meta.Tree
 import scala.meta.inputs.Position
 import scala.meta.internal.metals.Buffers
@@ -56,7 +57,7 @@ class CreateCompanionObjectCodeAction(
       document <- buffers.get(path)
     } yield buildCreatingCompanionObjectCodeAction(
       path,
-      tree,
+      endPosOfTree(tree),
       uri,
       getIndentationForPositionInDocument(tree.pos, document),
       name,
@@ -66,6 +67,19 @@ class CreateCompanionObjectCodeAction(
 
     maybeCompanionObject.toSeq
 
+  }
+
+  private def endPosOfTree(t: Tree): l.Position = {
+    val endMarkerPos: Option[Position] =
+      t.parent.flatMap { p =>
+        // check if tree just after t is its end marker
+        val possibleEndMarker = p.children.dropWhile(_ != t).tail.headOption
+        possibleEndMarker match {
+          case Some(marker: Term.EndMarker) => Some(marker.pos)
+          case _ => None
+        }
+      }
+    endMarkerPos.getOrElse(t.pos).toLSP.getEnd
   }
 
   private def getIndentationForPositionInDocument(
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
```diff
@@ -68,6 +69,22 @@ class CreateCompanionObjectCodeAction(
 
   }
 
+  def endPosOfTree(t: Tree): Position = {
```
zmerr, line 72: I think it would be more readable if this returns `l.Position` to truly reflect the end position of the tree.

zmerr (commented):
Thanks for the PR!

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
dos65: In this case this is `scala.meta.inputs.Position`.
We use `l` prefix(`import org.eclipse.{lsp4j => l}`) for lsp structures.

tgodzik (approved):
LGTM!

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
dos65: oh, right! changed

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
```diff
@@ -68,6 +69,22 @@ class CreateCompanionObjectCodeAction(
...
+        // checks if the next child in parent after given tree is `EndMarker`
+        if (idx + 1 < p.children.length) {
+          p.children(idx + 1) match {
+            case marker: Term.EndMarker => Some(marker)
+            case _ => None
+          }
+        } else
+          None
+      }
+    endMarker.map(_.pos).getOrElse(t.pos).toLSP.getEnd
```
kpodsiad, line 85: Because `p.children` is a list I would rather avoid traversing it 3 times (`indexOf`, `p.children.length` and `p.children(idx + 1)`).
```suggestion
    val endMarkerPos: Option[Position] =
      t.parent.flatMap { p =>
        // check if tree just after t is its end marker
        val possibleEndMarker = p.children.dropWhile(_ != t).tail.headOption
        possibleEndMarker match {
          case Some(marker: Term.EndMarker) => Some(marker.pos)
          case _ => None
        }
      }
    endMarkerPos.getOrElse(t.pos).toLSP.getEnd
```

kpodsiad (commented):
One minor suggestion

metals/src/main/scala/scala/meta/internal/metals/codeactions/CreateCompanionObjectCodeAction.scala
dos65: thx!

kpodsiad (approved):
LGTM!

PR #3867, opened 2022-04-26
traces: do not log `build/taskProgress` notifications

kpodsiad:
I'm curious what's yours opinion about that.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Trace.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Trace.scala
@@ -63,7 +63,14 @@ object Trace {
             StandardOpenOption.CREATE,
             StandardOpenOption.TRUNCATE_EXISTING // don't append infinitely to existing file
           )
-          Some(new PrintWriter(fos))
+          // logs are flooded with build/taskProgress logs which are pretty useless
+          // don't log them
+          val printWriter = new PrintWriter(fos) {
+            override def print(str: String): Unit =
+              if (str.contains("build/taskProgress")) ()
+              else super.print(str)
+          }
+          Some(printWriter)
         case _ :: tail =>
           setupPrintWriter(tail)
         case Nil =>


metals/src/main/scala/scala/meta/internal/metals/Trace.scala
```diff
@@ -63,7 +63,14 @@ object Trace {
             StandardOpenOption.CREATE,
             StandardOpenOption.TRUNCATE_EXISTING // don't append infinitely to existing file
           )
-          Some(new PrintWriter(fos))
+          // logs are flooded with build/taskProgress logs which are pretty useless
+          // don't log them
+          val printWriter = new PrintWriter(fos) {
+            override def print(str: String): Unit =
+              if (str.contains("build/taskProgress")) ()
```
tgodzik, line 70: I would avoid removing some things, since we want to have a full view in the logs and at some point if someone looks into the trace they might wander why there is no logs about taskProgress.

You can always do a grep, or create a smaller script that can parse and include/exclude some logs. That shouldn't be too hard to add and offers much more fexibility.

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/metals/Trace.scala
kpodsiad: What if this could be something a user (by user I mean me 😅) can opt-in by setting some property? Will this be acceptable?

tgodzik: I would rather see the script :P An option would be for this sole purpose, while the script is forever.

Something like this: https://gist.github.com/tgodzik/58fdea2e11112d545d9182bff6c6f857

kpodsiad: You've never told that you have your own script :P

tgodzik: I just wrote it :P

PR #3871, opened 2022-04-26
Update Bloop Java Home to that of Metals

ghost:
If Metals Java Home gets updated, the Bloop java home gets also updated with that of the metals, through the Bloop global json file.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
@@ -131,13 +131,13 @@ final class BloopServers(
   private def writeJVMPropertiesToBloopGlobalJsonFile(
       bloopGlobalJsonFilePath: AbsolutePath,
       bloopCreatedByMetalsFilePath: AbsolutePath,
-      maybeRequestedBloopJvmProperties: Option[List[String]],
+      maybeBloopJvmProperties: Option[List[String]],
       maybeJavaHome: Option[String]
   ): Try[Unit] = Try {
 
-    val javaOptionsString = maybeRequestedBloopJvmProperties
-      .map { requestedBloopJvmProperties =>
-        s"\"javaOptions\": [${requestedBloopJvmProperties.map(property => s"\"$property\"").mkString(", ")}]"
+    val javaOptionsString = maybeBloopJvmProperties
+      .map { bloopJvmProperties =>
+        s"\"javaOptions\": [${bloopJvmProperties.map(property => s"\"$property\"").mkString(", ")}]"
       }
       .getOrElse("")
 
@@ -168,7 +168,7 @@ final class BloopServers(
       messageActionItem: MessageActionItem,
       bloopGlobalJsonFilePath: AbsolutePath,
       bloopCreatedByMetalsFilePath: AbsolutePath,
-      maybeRequestedBloopJvmProperties: Option[List[String]],
+      maybeBloopJvmProperties: Option[List[String]],
       maybeJavaHome: Option[String],
       reconnect: () => Future[BuildChange]
   ): Future[Unit] = {
@@ -179,7 +179,7 @@ final class BloopServers(
         writeJVMPropertiesToBloopGlobalJsonFile(
           bloopGlobalJsonFilePath,
           bloopCreatedByMetalsFilePath,
-          maybeRequestedBloopJvmProperties,
+          maybeBloopJvmProperties,
           maybeJavaHome
         ) match {
           case Failure(exception) => Future.failed(exception)
@@ -209,7 +209,7 @@ final class BloopServers(
   private def updateBloopGlobalJsonFileThenRestart(
       bloopGlobalJsonFilePath: AbsolutePath,
       bloopCreatedByMetalsFilePath: AbsolutePath,
-      maybeRequestedBloopJvmProperties: Option[List[String]],
+      maybeBloopJvmProperties: Option[List[String]],
       maybeJavaHome: Option[String],
       bloopJsonUpdateCause: BloopJsonUpdateCause,
       reconnect: () => Future[BuildChange]
@@ -217,7 +217,7 @@ final class BloopServers(
     writeJVMPropertiesToBloopGlobalJsonFile(
       bloopGlobalJsonFilePath,
       bloopCreatedByMetalsFilePath,
-      maybeRequestedBloopJvmProperties,
+      maybeBloopJvmProperties,
       maybeJavaHome
     ) match {
       case Failure(exception) => Future.failed(exception)
@@ -317,12 +317,16 @@ final class BloopServers(
         bloopJsonUpdateCause <-
           if (
             maybeRequestedBloopJvmProperties != maybeBloopGlobalJsonJvmProperties
+            && maybeRequestedBloopJvmProperties.nonEmpty
           ) Some(BloopJsonUpdateCause.JVM_OPTS)
           else if (maybeRequestedMetalsJavaHome != maybeBloopGlobalJsonJavaHome)
             Some(BloopJsonUpdateCause.JAVA_HOME)
           else None
+        maybeBloopJvmProperties = maybeRequestedBloopJvmProperties.orElse(
+          maybeBloopGlobalJsonJvmProperties
+        )
       } yield updateBloopJvmProperties(
-        maybeRequestedBloopJvmProperties,
+        maybeBloopJvmProperties,
         bloopGlobalJsonFilePath,
         bloopCreatedByMetalsFilePath,
         maybeRequestedMetalsJavaHome,
@@ -336,7 +340,7 @@ final class BloopServers(
   }
 
   private def updateBloopJvmProperties(
-      maybeRequestedBloopJvmProperties: Option[List[String]],
+      maybeBloopJvmProperties: Option[List[String]],
       bloopGlobalJsonFilePath: AbsolutePath,
       bloopCreatedByMetalsFilePath: AbsolutePath,
       maybeJavaHome: Option[String],
@@ -363,7 +367,7 @@ final class BloopServers(
             _,
             bloopGlobalJsonFilePath,
             bloopCreatedByMetalsFilePath,
-            maybeRequestedBloopJvmProperties,
+            maybeBloopJvmProperties,
             maybeJavaHome,
             reconnect
           ) andThen {
@@ -382,7 +386,7 @@ final class BloopServers(
       updateBloopGlobalJsonFileThenRestart(
         bloopGlobalJsonFilePath,
         bloopCreatedByMetalsFilePath,
-        maybeRequestedBloopJvmProperties,
+        maybeBloopJvmProperties,
         maybeJavaHome,
         bloopJsonUpdateCause,
         reconnect


metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
```diff
@@ -39,6 +39,7 @@ case class UserConfiguration(
     bloopSbtAlreadyInstalled: Boolean = false,
     bloopVersion: Option[String] = None,
     bloopJvmProperties: Option[List[String]] = None,
+    bloopJavaHome: Option[String] = None,
```
tgodzik, line 42: I would just use the same java Home as the Metals one it will be a bit confusing for the users otherwise.

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/metals/UserConfiguration.scala
zmerr: Ok, so I would remove it!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -276,76 +273,89 @@ final class BloopServers(
...
-                )
-              case Success(_) => Future.successful()
-            }
-          }
-        } else Future.successful(())
 
-    result.getOrElse(Future.successful(()))
+        requestedBloopJvmProperties = maybeRequestedBloopJvmProperties
+          .getOrElse(List.empty)
+        if maybeRequestedBloopJvmProperties != maybeRunningBloopJvmProperties ||
```
tgodzik, line 289: Could we rework `maybeRunningBloopJvmProperties` to read from the `bloop.json` file so that even if we have `None` then we will not show the message about restarting if it's not needed? Currently this shows up every time we start the server.

The same with `maybeRunningMetalsJavaHome`, it will be None in the default value and updated by the client each time. We should check if it's already set as Java home. I am thinking we might need an additional workflow for this actually.

We could set the `bloop.json` at the start of the server before starting Bloop (you can take a look at the BloopServers class) so that in most cases users don't need to do anything and not even ask about anything unless there is a `bloop.json` file already with a different home. In case bloop.json exists with a different home we can show a message to change the javaHome to avoid issues.

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: @tgodzik not sure if I understood what should be done for javaHome! But with the changes I just pushed, Metals does not propose to restart Bloop anymore, if the user requested setting matches the contents of the global json file! You can take a look!

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -276,76 +294,95 @@ final class BloopServers(
   def ensureDesiredJvmSettings(
       maybeRequestedBloopJvmProperties: Option[List[String]],
       maybeRunningBloopJvmProperties: Option[List[String]],
-      maybeJavaHome: Option[String],
+      maybeRequestedMetalsJavaHome: Option[String],
+      maybeRunningMetalsJavaHome: Option[String],
```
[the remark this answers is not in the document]
zmerr: Well, the way it works is this: it checks the running version from memory, then if it is None, reads from the json file! By removing them we would have to read from the json file each time by default! Is this the functionality we are looking for?

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -241,6 +241,27 @@ final class BloopServers(
     }.toOption
   }
 
+  private def maybeLoadBloopGlobalJsonFile(
+      bloopGlobalJsonFilePath: AbsolutePath
+  ): (Option[String], Option[List[String]]) = {
+    val maybeData = util.Try {
```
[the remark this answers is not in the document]
zmerr: I remember last time I tried that, there were syntax highlighting issues! I would try again!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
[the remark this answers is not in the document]
tgodzik: > Well, the way it works is this: it checks the running version from memory, then if it is None, reads from the json file

I don't see where that happens, this condition:
```
 (maybeRequestedBloopJvmProperties != maybeRunningBloopJvmProperties
          && maybeRequestedBloopJvmProperties != maybeBloopGlobalJsonJvmProperties)
```
will be true at the start when both running java home and properties are None and the requested ones are Some. So we will ask to restart even if Bloop is running properly already.

It also seems we need to change both at the same time to get the pop up about restarting Bloop. It might be a bit less complex to use the global setting only instead of the one in the user configuration.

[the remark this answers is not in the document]
zmerr: At the start, `maybeRequestedBloopJvmProperties != maybeRunningBloopJvmProperties` will be true! Then `maybeRequestedBloopJvmProperties != maybeBloopGlobalJsonJvmProperties` will be false! This means that 
`(maybeRequestedBloopJvmProperties != maybeRunningBloopJvmProperties
          && maybeRequestedBloopJvmProperties != maybeBloopGlobalJsonJvmProperties)` will be false! 
Therefor, at the start we are not going to ask to restart Bloop!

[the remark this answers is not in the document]
zmerr: > It also seems we need to change both at the same time to get the pop up about restarting Bloop
  No, how did you deduct that?
according to this condition:

 if (maybeRequestedBloopJvmProperties != maybeRunningBloopJvmProperties
           && maybeRequestedBloopJvmProperties != maybeBloopGlobalJsonJvmProperties) ||
           (maybeRequestedMetalsJavaHome != maybeRunningMetalsJavaHome
             && maybeRequestedMetalsJavaHome != maybeBloopGlobalJsonJavaHome)
changing just one of either `bloopJvmProperties` or `javaHome` triggers the update! Because they are connected by `||`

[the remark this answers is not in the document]
zmerr: > Well, the way it works is this: it checks the running version from memory, then if it is None, reads from the json file!

I guess the problem is that we are always reading from the bloop json file anyway! So I think I should just remove the `running` method arguments or change the semantics to the quoted logic?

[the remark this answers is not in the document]
zmerr: I can right it as just 
```
 if maybeRequestedBloopJvmProperties != maybeBloopGlobalJsonJvmProperties ||
            maybeRequestedMetalsJavaHome != maybeBloopGlobalJsonJavaHome
```
for simplicity!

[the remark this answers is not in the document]
tgodzik: Ach, you are right I misread the conditions, which is a perfect example of it being a bit complicated :sweat: 

Maybe we don't need the running since each time we change to requested that gets saved to global config, which should be a good source of truth.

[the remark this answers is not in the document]
tgodzik: Yep, I think it should be much simpler this way.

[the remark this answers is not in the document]
zmerr: Ok, no problem!

tgodzik:
If anyone else could also take a look and double check the approach here I would be super thankful!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -275,77 +293,92 @@ final class BloopServers(
...
+          maybeRequestedMetalsJavaHome != maybeBloopGlobalJsonJavaHome
+      } yield updateBloopJvmProperties(
+        requestedBloopJvmProperties,
+        bloopGlobalJsonFilePath,
+        bloopCreatedByMetalsFilePath,
+        maybeRequestedMetalsJavaHome,
+        reconnect
+      )
+    result.getOrElse {
+      Future.successful()
```
kpodsiad, line 320:
```suggestion
      Future.unit
```

kpodsiad (commented):
Could you remind me what are the benefits from coupling Metals' java and Bloop's java?

dos65 (commented):
I don't mind of having this approach.

The only I'm worried is that javaHome changes silently lead to a notification `bloop properties was changed: Restart/Now now` that might be not clear for a user.

tgodzik:
> I don't mind of having this approach.
> 
> The only I'm worried is that javaHome changes silently lead to a notification `bloop properties was changed: Restart/Now now` that might be not clear for a user.

@zmerr could we make sure that when java Home changes we get a different messages than when properties change?

zmerr:
> > I don't mind of having this approach.
> > The only I'm worried is that javaHome changes silently lead to a notification `bloop properties was changed: Restart/Now now` that might be not clear for a user.
> 
> @zmerr could we make sure that when java Home changes we get a different messages than when properties change?

Sure! so as to avoid surprises for the user!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: All applied!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -241,6 +243,27 @@ final class BloopServers(
...
+      ujson.read(jsonString)
+    }
+    val javaHome = maybeData.flatMap { data =>
+      Try(data("javaHome").str)
+    }.toOption
+    val javaOptions = maybeData.flatMap { data =>
+      Try {
+        data("javaOptions").arr.toList.map(_.str)
+      }
+    }
```
tgodzik, line 263:
```suggestion
     val javaHome = for {
        obj <- maybeData.objOpt
        strValue <- obj.get("javaHome")
        str <- strValue.strOpt
      } yield str
    val javaOptions = for {
      obj <- maybeData.objOpt
      value <- obj.get("javaOptions")
      arr <- value.arrOpt
      str = arr.flatMap(_.strOpt)
   } yield str
```

No need for the Try this way and it's better to handle things explicitely.

tgodzik (changes requested):
Looks almost ready to merge. Just a few minor comments and you can change it from the draft PR to a normal one.

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -299,14 +323,20 @@ object Messages {
...
-    def params(): ShowMessageRequestParams = {
+
+    def params(
+        bloopJsonUpdateCause: BloopJsonUpdateCause
+    ): ShowMessageRequestParams = {
       val params = new ShowMessageRequestParams()
       params.setMessage(
-        s"""|You have previously modified the JVM settings of Bloop in its global json file,
+        s"""|Setting $bloopJsonUpdateCause will result in updating Bloop's global Json file, by Metals.
+            |But, you have previously modified this file with other values!
```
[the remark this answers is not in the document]
zmerr: Are there any cases in which Bloop global json file’s modification could be automated, other than Metals?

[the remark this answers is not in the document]
zmerr: applied. Thanks!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: applied. Thanks!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -241,6 +243,27 @@ final class BloopServers(
...
+  private def maybeLoadBloopGlobalJsonFile(
+      bloopGlobalJsonFilePath: AbsolutePath
+  ): (Option[String], Option[List[String]]) = {
+    val maybeData = Try {
+      val source = scala.io.Source.fromFile(bloopGlobalJsonFilePath.toURI)
+      val jsonString =
+        try source.mkString
+        finally source.close()
+      ujson.read(jsonString)
+    }
```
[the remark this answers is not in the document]
zmerr: applied. Thanks!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -275,77 +295,101 @@ final class BloopServers(
    */
   def ensureDesiredJvmSettings(
       maybeRequestedBloopJvmProperties: Option[List[String]],
```
tgodzik, line 297: I think the issue with test is coming from the fact, that we are sending `None` here and we already have something set for CI. None should be treated as `use whatever is set currently` while `Some(Nil)` would be `remove all the options`

zmerr: Thank you for this. I think the guard 
```
bloopJsonUpdateCause <-
          if (
            maybeRequestedBloopJvmProperties != maybeBloopGlobalJsonJvmProperties
          else if (maybeRequestedMetalsJavaHome != maybeBloopGlobalJsonJavaHome)
            Some(BloopJsonUpdateCause.JAVA_HOME)
          else None

```
was enforcing that! Still, as an extra test, I checked that the `maybeRequestedBloopJvmProperties` is never `None`
It can only be `List.empty` when the user has not set anything!

tgodzik: I think it would be enought to add and if:
`if maybeRequestedBloopJvmProperties.nonEmpty`
somewhere in the for comprehension

zmerr: Then what about the cases when it’s empty, while javaHome is set?

zmerr: I think it might have to do with how the language client is called.

zmerr: specifically in cases when the bloop global json is manually premodified. The errors do not show for the other case.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -275,77 +301,100 @@ final class BloopServers(
...
-              case Failure(exception) =>
-                languageClient.showMessage(
-                  MessageType.Error,
-                  exception.getMessage
-                )
-              case Success(_) => Future.successful()
-            }
-          }
-        } else Future.successful(())
+            maybeRequestedBloopJvmProperties != maybeBloopGlobalJsonJvmProperties
```
tgodzik, line 319:
```suggestion
     maybeRequestedBloopJvmProperties.nonEmpty &&   maybeRequestedBloopJvmProperties != maybeBloopGlobalJsonJvmProperties
```

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -120,29 +121,35 @@ final class BloopServers(
...
       bloopGlobalJsonFilePath: AbsolutePath,
       bloopCreatedByMetalsFilePath: AbsolutePath,
-      requestedBloopJvmProperties: List[String],
+      maybeRequestedBloopJvmProperties: Option[List[String]],
       maybeJavaHome: Option[String]
   ): Try[Unit] = Try {
 
-    val javaOptionsString =
-      s"\"javaOptions\": [${requestedBloopJvmProperties.map(property => s"\"$property\"").mkString(", ")}]"
+    val javaOptionsString = maybeRequestedBloopJvmProperties
```
tgodzik, line 138: Btw. if `maybeRequestedBloopJvmProperties` is empty we might want to use the existing options that are in the file.

tgodzik (approved):
LGTM!

PR #3872, opened 2022-04-27
upgrade bsp4j version to a stable one

dos65:
Fixes the issue with `cs install` - https://github.com/coursier/coursier/issues/2406


tanishiking (approved):
LGTM!
Should we release a new version that includes this change?

dos65:
@tanishiking 
> Should we release a new version that includes this change?

Yes, we should. Will do preparations for that.

kpodsiad:
Thanks @dos65, I forgot that about the snapshot version and I've realized about it too late.

PR #3875, opened 2022-04-27
fix: fix link to example configuration

ckipp01:
Looks like if you break in the middle of the `[]()` it won't render the
link.


tanishiking:
The link is indeed broken, can we re-open the PR and merge it? @ckipp01 
Screen Shot 2022-04-28 at 16 33 50
https://scalameta.org/metals/docs/editors/vim

ckipp01:
> The link is indeed broken, can we re-open the PR and merge it? @ckipp01 Screen Shot 2022-04-28 at 16 33 50 https://scalameta.org/metals/docs/editors/vim

yea I noticed it was broken however the fix I tried didn't work. I wasn't able to quickly fix it and I'm on holiday currently. I will look closer when I get back.

PR #3876, opened 2022-04-27
fix: sbt-metals + java semanticdb issue on jdk17

dos65:
In case of jdk17, semanticdb-plugin requires `-J*` flags otherwise
compilation fails. These flags can be passed only at the JVM start.

Her, we explictly define `javaHome` for jdk >= 17.
It forces sbt to use `ForkJavaCompiler` that accepts `-J` flags instead
of `LocalJavaCompiler` which ignores them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,11 @@ jobs:
             command: bin/test.sh 'sbt-metals/scripted; slow/testOnly -- tests.sbt.*'
             name: Sbt integration
             os: ubuntu-latest
+          - type: sbt-metals jdk17
+            command: bin/test.sh sbt-metals/scripted
+            name: Sbt-metals/scripted jdk17
+            os: ubuntu-latest
+            java: "17"
           - type: maven
             command: bin/test.sh 'slow/testOnly -- tests.maven.*'
             name: Maven integration
--- a/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
+++ b/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
@@ -48,6 +48,22 @@ object MetalsPlugin extends AutoPlugin {
         )
     },
     javaSemanticdbEnabled := bspEnabled.value,
+    javaHome := {
+      javaHome.value match {
+        case None =>
+          // In case if jdk is >= 17 we need to set `javaHome` explicitly
+          // to force sbt creating ForkJavaCompiler.
+          // Otherwise it creates LocalJavaCompiler that ignores `-J` flags
+          // and semanticdb plugin fails compilation
+          // See sbt.internal.inc.javac.JavaCompiler.directOrFork
+          // https://github.com/sbt/zinc/blob/dd1f3596494b4c4d7a19256bc921e765d3dc12c8/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavaCompiler.scala#L41-L67
+          JdkVersion.getJavaVersionFromJavaHome(processJavaHome) match {
+            case Some(v) if v.major >= 17 => Some(processJavaHome)
+            case _ => None
+          }
+        case defined @ Some(_) => defined
+      }
+    },
     javacOptions ++= {
       if (javaSemanticdbEnabled.value)
         javaSemanticdbOptions.value
@@ -63,6 +79,8 @@ object MetalsPlugin extends AutoPlugin {
         Nil
     }
   )
+
+  private def processJavaHome: File = file(System.getProperty("java.home"))
 
   def requiresSemanticdb: Def.Initialize[Boolean] = Def.setting {
     bspEnabled.value &&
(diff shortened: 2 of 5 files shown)


sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
```diff
@@ -89,9 +105,8 @@ object MetalsPlugin extends AutoPlugin {
       } else Nil
     }
     Def.setting {
-      val defined = javaHome.value
-      val value = defined.getOrElse(file(System.getProperty("java.home")))
-      JdkVersion.getJavaVersionFromJavaHome(value) match {
+      val explicit = javaHome.value
+      JdkVersion.inferJavaVersion(explicit) match {
```
tanishiking, line 109: [note] refactor

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
```diff
@@ -48,6 +48,20 @@ object MetalsPlugin extends AutoPlugin {
         )
     },
     javaSemanticdbEnabled := bspEnabled.value,
+    javaHome := {
+      javaHome.value match {
+        case None =>
+          // In case if jdk is >= 17 we need to set `javaHome` explicitly
+          // to force sbt creating ForkJavaCompiler.
+          // Otherwise it creates LocalJavaCompiler that ignores `-J` flags
+          // and semanticdb plugin fails compilation
```
tanishiking, line 57: Can you leave a link to

- https://github.com/sbt/zinc/blob/dd1f3596494b4c4d7a19256bc921e765d3dc12c8/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavaCompiler.scala#L41-L67
- and https://github.com/sbt/zinc/blob/dd1f3596494b4c4d7a19256bc921e765d3dc12c8/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/LocalJava.scala#L299-L304 ?

At first, I couldn't understand who is `ForkJavaCompiler` and `LocalJavaCompiler`.

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
```diff
@@ -48,6 +48,20 @@ object MetalsPlugin extends AutoPlugin {
         )
     },
     javaSemanticdbEnabled := bspEnabled.value,
+    javaHome := {
+      javaHome.value match {
+        case None =>
+          // In case if jdk is >= 17 we need to set `javaHome` explicitly
```
tanishiking, line 54: Why this is the case only when `jdk is >= 17`?
I took a quick look at `sbt/zinc` and it seems like the case for all java versions 🤔

kpodsiad: starting from jdk17 a lot of modules were closed and has to be explicitly opened, otherwise you'll get 
```shell
java.lang.IllegalAccessError: class com.sourcegraph.semanticdb_javac.SemanticdbPlugin 
(in unnamed module @0x29e5de87) cannot access class com.sun.tools.javac.api.BasicJavacTask 
(in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.api 
to unnamed module @0x29e5de87
```
like I got

.github/workflows/ci.yml
```diff
@@ -83,6 +83,11 @@ jobs:
             command: bin/test.sh 'sbt-metals/scripted; slow/testOnly -- tests.sbt.*'
             name: Sbt integration
             os: ubuntu-latest
+          - type: sbt-metals jdk17
+            command: bin/test.sh sbt-metals/scripted
```
kpodsiad, line 87: Are our current tests covering this issue?

kpodsiad (approved):

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
tanishiking: Ok, I got it!

- So, from JDK16, we have to pass flags like `add-exports` to access JDK internals as https://openjdk.java.net/jeps/396
- Metals pass `add-exports` flags via   https://github.com/scalameta/metals/blob/d4bf310e34eeb67539475a248d3a453549417413/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala#L80-L88
  - As documented in https://sourcegraph.github.io/lsif-java/docs/getting-started.html#java
- But if we use `LocalJavaCompiler`, it doesn't accept those flags, and we hit `IllegalAccessError`

👍

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
dos65: added a link

.github/workflows/ci.yml
dos65: Ouch. Thx for pointing at this.
Now it checks

tanishiking (approved):
LGTM!

Issue #3878, opened 2022-04-28
Don't show quickfix if it's not available


tanishiking:
### Describe the bug

Currently, in Scala3, several code actions are not yet available.
However, Metals show quick help as if they're available, and when we click it, nothing happens.

It isn't very clear for those who don't know which code actions are implemented in Scala3 or not (meaning almost all the users).

### Expected behavior

Metals don't show a quick help if the codeaction is not available.

### Version of Metals

v0.11.4

### Extra context or search terms

codeaction

tanishiking:
Let's do our best implement all code actions for Scala3 before next release https://github.com/scalameta/metals/pull/3880#issuecomment-1112156453

Issue #3877, opened 2022-04-28
VSCode Metals quickhelp doesnt implement methods


yadavan88:
### Describe the bug

The quickhelp suggestions in vscode with metals doesn't work. 

Steps to reproduce:
1. Create a small object extending IOApp.Simple in cats effect
2. It will show compilation error since run() is not implemented
3. Click on the bulb icon and then click on "implement all members"
4. Nothing happens

### Expected behavior

It should implement the _run()_ method in this case

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.4

### Extra context or search terms

Scala Version: 3.1.2
Bloop : v1.5.0
Java: AdoptOpenJDK 8
Metals Server version: 0.11.4

Screenshot Below:
<img width="390" alt="image" src="https://user-images.githubusercontent.com/9131956/165690570-acb3899f-3a1d-4f64-a96d-21ccbbe51ed1.png">

tanishiking:
@yadavan88 Thank you for reporting! Unfortunately, this feature is not yet available in Scala3 https://github.com/scalameta/metals/issues/1692 (Maybe we shouldn't show a CodeAction in Scala3 because it's confusing).
However, we are working on this feature, stay tuned!

tanishiking:
I opened an issue not to show a quick help if the code action is not available https://github.com/scalameta/metals/issues/3877

yadavan88:
@tanishiking Thank you very much for such a quick response.

tanishiking:
Let me close this issue for https://github.com/scalameta/metals/issues/3878 and https://github.com/scalameta/metals/issues/1692
Again, thank you very much for reporting!

Issue #1692, opened 2020-05-05
Action “Implement all members” does not work on Scala 3


julienrf:
**Describe the bug**

The action “implement all members” does nothing on Scala 3.

**To Reproduce**

1. ~~~ scala
    trait Foo {
      def bar(): Unit
    }
    
    object Qux extends Foo {
    
    }
    ~~~

2. Compilation error, click on “Quick-fix”, then “Implement all members”
3. Nothing happens

**Expected behavior**

The file should be updated with an implementation stub like:

~~~ scala
object Qux extends Foo {
  def bar(): Unit = ???
}
~~~

**Installation:**
- Operating system: Linux
- Editor: Visual Studio Code/
- Metals version: v1.9.0
- Scala version: 0.24.0-RC1

**Search terms**
Dotty

tgodzik:
Thanks for finding this! I can confirm it doesn't work. Will need to take a closer look.

bramvdbogaerde:
@tgodzik Can you give an update on this? It still doesn't work for me, running Scala 3.0.2 with Metals 0.10.6.

dos65:
@bramvdbogaerde I still isn't implemented

tanishiking:
Now `AutoImports` for Scala3 is implemented, I'm going to work on this feature (as this is my favorite one 😄 )

tanishiking:
I was trying to develop a prototype of this feature https://github.com/scalameta/metals/commit/29753fb6094c111d610a8db89baedb112d03b53b
and I stuck around listing all the members of the class.

In Scala2, we get all the members (including abstract members) by `typed.tpe.members`, and then filter methods that should be implemented.
https://github.com/scalameta/metals/blob/50c60add4e0f250e9352ab36af21cf6a6ff4d917/mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala#L285-L289

In Scala3, we should be able to list all the members by [Type#allMembers](https://github.com/lampepfl/dotty/blob/7ee8986699dab695af34d4cf9671e6e840e154ad/compiler/src/dotty/tools/dotc/core/Types.scala#L1017-L1021), it works if the concrete class and the parent (abstract) classes are in the same file.

[...]

PR #3880, opened 2022-04-28
feature: don't show unsupported quickfix in Scala3

tanishiking:
fix https://github.com/scalameta/metals/issues/3878

As reported in https://github.com/scalameta/metals/issues/3877 and https://github.com/scalameta/metals/issues/1692, in Scala3, Metals show a quickfix command "implement all members" but when we click it, nothing happens (because it's not yet available in Scala3).

This is confusing behavior for users: they don't have any clues whether this is expected behavior, this is a bug, or it's not yet an available feature.

This PR disable Metals to show quickfix command if it's not yet available in Scala3:

Screen Shot 2022-04-28 at 18 13 36

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/CodeAction.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/CodeAction.scala
@@ -16,6 +16,8 @@ trait CodeAction {
    */
   def kind: String
 
+  def supportScala3: Boolean
+
   def contribute(
       params: l.CodeActionParams,
       token: CancelToken
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
@@ -85,6 +85,8 @@ class SourceOrganizeImports(
     ) {
 
   override val kind: String = SourceOrganizeImports.kind
+
+  override def supportScala3: Boolean = false
   override protected val title: String = SourceOrganizeImports.title
 
   override protected def isCallAllowed(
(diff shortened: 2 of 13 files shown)


metals/src/main/scala/scala/meta/internal/metals/codeactions/OrganizeImports.scala
```diff
@@ -132,6 +134,8 @@ class OrganizeImportsQuickFix(
     ) {
 
   override val kind: String = OrganizeImportsQuickFix.kind
+
+  override def supportScala3: Boolean = false
```
tanishiking, line 138: I suppose this is not yet available in Scala3. Am I right?

metals/src/main/scala/scala/meta/internal/metals/CodeAction.scala
```diff
@@ -16,6 +16,8 @@ trait CodeAction {
    */
   def kind: String
 
+  def supportScala3: Boolean
```
kpodsiad, line 19: Maybe we should set the default value to true and override only in relevant actions?

dos65:
@tanishiking I'm wondering if this will useful at the time of the next release. 
I think I mean, the chances of finally implementing this action for scala3 are probably high that will lead to the situation that all action supports scala3  :smile:

tanishiking:
> I think I mean, the chances of finally implementing this action for scala3 are probably high that will lead to the situation that all action supports scala3 😄

Right, only a few actions are not available in Scala3.
Ok, closing this PR for now, and let's see the progress when the next release is around the corner :)
Thank you for your comments @dos65 @kpodsiad

PR #3865, opened 2022-04-25
[WIP] Add scaladocs for Scala 3

tanishiking:
I'll write some PR description, refactoring, and testing, but it works in my local environment :tada:

https://github.com/scalameta/metals-feature-requests/issues/109

PR #3881, opened 2022-04-28
Show parent scaladoc if implementation is returning empty

tgodzik:
Previously, we would only return docstring if the current method has them defined, which wouldn't work well in case of a lot of standard library methods. Now, we check parents for symbols so that we can print a proper scaladoc

Prompted by @kpodsiad , we can merge after https://github.com/scalameta/metals/pull/3865

### DIFF
--- a/metals-bench/src/main/scala/bench/ClasspathOnlySymbolSearch.scala
+++ b/metals-bench/src/main/scala/bench/ClasspathOnlySymbolSearch.scala
@@ -6,6 +6,7 @@ import java.{util => ju}
 
 import scala.meta.internal.metals.ClasspathSearch
 import scala.meta.internal.metals.WorkspaceSymbolQuery
+import scala.meta.pc.ParentSymbols
 import scala.meta.pc.SymbolDocumentation
 import scala.meta.pc.SymbolSearch
 import scala.meta.pc.SymbolSearchVisitor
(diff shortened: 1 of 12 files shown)


mtags/src/main/scala-2/scala/meta/internal/metals/Docstrings.scala
```diff
@@ -30,18 +33,49 @@ class Docstrings(index: GlobalSymbolIndex) {
...
       case None =>
         indexSymbol(symbol)
         val result = cache.get(symbol)
-        if (result.isEmpty) {
-          cache(symbol) = EmptySymbolDocumentation
+        val resultWithDocs = result match {
+          case None =>
+            cache(symbol) = EmptySymbolDocumentation
+            result
+          /* Fall back to parent scaladocs is nothing is specified for the current symbol
```
tanishiking, line 51:
```suggestion
          /* Fall back to parent scaladocs if nothing is specified for the current symbol
```

tanishiking (approved):
One type, otherwise LGTM :tada:

mtags/src/main/scala-2/scala/meta/internal/metals/Docstrings.scala
tgodzik: Fixed!

PR #3886, opened 2022-04-29
docs: fix latest versions

dos65:


tgodzik (approved):
ooooch, that's right. Thanks for figuring it out!

PR #3887, opened 2022-04-29
Don't create the Presentation Compiler for Java files

tgodzik:
Previously, we would try to create a presentation compiler even if currently working on a Java file.
Now, we will return default values in case of Java files.

This showed up the mtags-interfaces, which show up as ScalaTargets (might be an issue in Bloop), but it doesn't contain Scala library on the classpath.


dos65 (approved):
LGTM!

PR #3896, opened 2022-05-02
Handle modules in stacktrace analyzer

Arthurm1:
Currently the stacktrace analyzer will not show links for jdk classes when the jdk has modules

So this would get a hyperlink...
```java
at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
```

But this would not...
```java
at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
```

Now the module is removed before the symbol is searched for and the jdk classes get hyperlinks.


tgodzik (approved):
LGTM!

PR #3899, opened 2022-05-06
Fix nightly tests failures

tgodzik:
It seems we are missing two methods:

image

### DIFF
--- a/tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
@@ -269,6 +269,47 @@ class CompletionSuite extends BaseCompletionSuite {
            |synchronized[T0](x$1: T0): T0
            |toString(): String
            |""".stripMargin,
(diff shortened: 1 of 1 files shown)


tgodzik:
This is being checked on the compiler side.

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -269,6 +269,47 @@ class CompletionSuite extends BaseCompletionSuite {
            |synchronized[T0](x$1: T0): T0
            |toString(): String
            |""".stripMargin,
+      "3.2" ->
```
dos65, line 272:
```suggestion
      ">=3.2.0-RC1-bin-20220504-26f9e77-NIGHTLY" ->
```

tgodzik: Actually, we might fix it in the compiler, so I will hold off for now.

PR #3900, opened 2022-05-07
feat: handle cats-effect async stacktrace in stacktrace analyzer

kpodsiad:
Handles following:
```scala
//> using scala "2.13.8"
//> using lib "org.typelevel::cats-effect:3.3.11"
package a

import cats.effect.IOApp
import cats.effect.IO

object Main extends IOApp.Simple {
  val io = IO(5)
  override def run: IO[Unit] = io
      .map { _ => 
        throw new Exception
        ()
      }.void
}

```

```
java.lang.Exception
        at a.Main$.$anonfun$run$1(Stacktraces.scala:12)
        at apply @ a.Main$.<clinit>(Stacktraces.scala:9)
        at map @ a.Main$.run(Stacktraces.scala:11)
        at void @ a.Main$.run(Stacktraces.scala:11)
        at run$ @ a.Main$.run(Stacktraces.scala:8)
```

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/StacktraceAnalyzer.scala
@@ -3,6 +3,7 @@ package scala.meta.internal.metals
 import java.io.FileWriter
 
 import scala.util.Try
+import scala.util.matching.Regex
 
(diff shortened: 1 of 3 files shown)


armanbilge:
Ah, so that's why you're poking at this :) fantastic!

tgodzik (approved):
LGTM!

Issue #3903, opened 2022-05-09
`plug-in not found: semanticdb` `javac returned non-zero exit code`


armanbilge:
### Describe the bug

https://github.com/armanbilge/schrodinger/commit/bffe56896175c2b2668c220dfda873882563b26e

sbt bsp server

```
$ java -version
openjdk version "11.0.13" 2021-10-19 LTS
OpenJDK Runtime Environment Zulu11.52+13-CA (build 11.0.13+8-LTS)
OpenJDK 64-Bit Server VM Zulu11.52+13-CA (build 11.0.13+8-LTS, mixed mode)
```

```
sbt:root> testsJVM/Test/compileJava
[info] compiling 11 Scala sources and 1 Java source to /workspace/schrodinger/tests/jvm/target/scala-3.1.2/test-classes ...
[error] plug-in not found: semanticdb
[error] (testsJVM / Test / compileIncremental) javac returned non-zero exit code
```

```
sbt:root> show testsJVM/Test/javacOptions
[info] * -Xplugin:semanticdb -sourceroot:/workspace/schrodinger -targetroot:/workspace/schrodinger/tests/jvm/target/scala-3.1.2/meta -build-tool:sbt
[info] * -encoding
[info] * utf8
[info] * -Xlint:all
```

```
sbt:root> show testsJVM/javaSemanticdbEnabled
[info] true
```

```
sbt:root> show testsJVM/allDependencies
[info] * org.scala-lang:scala3-compiler:3.1.2:scala-tool->default(compile)
[info] * org.scala-lang:scaladoc:3.1.2:scala-doc-tool->default(compile)
...
[info] * com.sourcegraph:semanticdb-javac:0.7.4:compile-internal
```

### Expected behavior

Either it works out-of-the-box, or if the Java SemanticDB is unsupported with the sbt bsp then it should be disabled.

Edit: I see, it should be supported as of
- https://github.com/scalameta/metals/pull/3861

### Operating system

Linux

### Editor/Extension

VS Code

### Version of Metals

v0.11.5

dos65:
Thanks for reporting!
I will looks what is going wrong.

As a temporal workaround you can add `ThisBuild/javaSemanticdbEnabled := false` to your build.sbt

armanbilge:
Reproduced the problem in an scripted test in https://github.com/scalameta/metals/pull/3907.

PR #3907, opened 2022-05-09
Add semanticdb-javac to `TestInternal`

armanbilge:
So far, added a test to reproduce the problem. Now, let's see if I can fix it too 🙃 

Edit: done. Fixes https://github.com/scalameta/metals/issues/3903

### DIFF
--- a/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
+++ b/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
@@ -71,18 +71,17 @@ object MetalsPlugin extends AutoPlugin {
         Nil
     },
     allDependencies ++= {
-      if (javaSemanticdbEnabled.value)
-      val existing = ivyConfigurations.value.map(_.name).toSet
-      val configurations = Seq(
-        Configurations.CompileInternal,
-        Configurations.TestInternal,
-        Configurations.IntegrationTestInternal
-      ).map(_.name).filter(existing.contains).mkString(",")
-       List(
+      if (javaSemanticdbEnabled.value) {
+        val existing = ivyConfigurations.value.map(_.name).toSet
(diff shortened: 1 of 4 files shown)


sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
```diff
@@ -72,8 +72,8 @@ object MetalsPlugin extends AutoPlugin {
     },
     allDependencies ++= {
       if (javaSemanticdbEnabled.value)
-        List(
-          "com.sourcegraph" % "semanticdb-javac" % BuildInfo.javaSemanticdbVersion % Configurations.CompileInternal
+        List(Configurations.CompileInternal, Configurations.TestInternal).map(
+          "com.sourcegraph" % "semanticdb-javac" % BuildInfo.javaSemanticdbVersion % _
```
dos65, line 76: Tried it locally and it's need also to cover Integrations tests.
```suggestion
      val existing = ivyConfigurations.value.map(_.name).toSet
      val configurations = Seq(
        Configurations.CompileInternal,
        Configurations.TestInternal,
        Configurations.IntegrationTestInternal
      ).map(_.name).filter(existing.contains).mkString(",")
       List(
          "com.sourcegraph" % "semanticdb-javac" % BuildInfo.javaSemanticdbVersion % configurations
        )
```

dos65 (commented):
Thanks for the fix!

sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala
armanbilge: Thanks, good point!

dos65 (approved):

tgodzik (approved):
LGTM!

PR #3790, opened 2022-04-05
Add better Scala CLI support

alexarchambault:
This PR makes Scala CLI support more reliable in two ways:
- it makes scripts (`.sc` files) work better in Scala CLI projects, by having Metals be aware of the way Scala CLI "wraps" those before passing them to scalac -  this is achieved by using a custom BSP endpoint, and relying on past work for Ammonite, in the commit `Handle pre-processed sources via BSP` (currently, https://github.com/scalameta/metals/pull/3790/commits/42253f62e19683ccb1a71aba311fc65fb7ec96a2)
- it adds non-BSP discovery based support for Scala CLI (that works alongside the existing 100% BSP-based one), in commit `Add non BSP-based Scala CLI project support` (currently, https://github.com/scalameta/metals/pull/3790/commits/6352c6d1027f2a909c43915a34cf5a6148bbf3d8)

Before these two commits, https://github.com/scalameta/metals/pull/3790/commits/dd657de7b0aaed564df09642f4d32fc382e8d318, https://github.com/scalameta/metals/pull/3790/commits/c4d7bde92ddebf6f3a25d1f1590c4219cf50f91d, and https://github.com/scalameta/metals/pull/3790/commits/98678b1eaefb98c3ed3f82c36dd07b08e0e131aa, are small refactorings that prepare the ground for the two others.

The non-BSP discovery based support allows to load Scala CLI projects *alongside* existing sbt or Mill projects (in the same VSCode window for example), just like it can be done with Ammonite scripts. This support loads whole directories at a time in Scala CLI, but could be made more atomic in the future when addressing https://github.com/VirtusLab/scala-cli/issues/1146 in Scala CLI (with it, loading a single file from Metals could automatically pull some other files it needs).

PR #3908, opened 2022-05-10
Don't assume locations are modified in place

alexarchambault:
That seems to be implied by the `AdjustLspData` API, and by the main `adjustLocations` implementation, that filters out some locations. Yet it seems that makes some tests fail, `tests.feature.Worksheet3LspSuite.definition` without the changes proposed here.


alexarchambault:
Closing, will do that as part of https://github.com/scalameta/metals/pull/3790 I think.

PR #3910, opened 2022-05-11
Update Bloop and Scala Debug adapter

tgodzik:
This adds support for Scala 3 expression evaluator


kpodsiad:
> I will go ahead and merge since it's just an update and I wanted to demo this today :P

I hope that today's demo was better than your's usual demo 😄

tgodzik:
The only thing that broke was the HDMI cable connection :sweat_smile:

PR #3913, opened 2022-05-13
[WIP] Show parameter name decorations in methods

tanishiking:
https://github.com/scalameta/metals-feature-requests/issues/184

This PR enables Metals to show parameter name decorations, as [IntelliJ does](https://www.jetbrains.com/help/idea/inlay-hints.html).

Screen Shot 2022-05-13 at 14 30 18

### TODOs
- [x] Support Scala3
- [ ] Support Scala2
- [ ] Add an server/client options to on/off this feature
- [ ] Should we show the decorations only for "simple" type?
  - How can we tell the type is "simple"?
- [ ] Refactoring (maybe we shouldn't add the logic into `SyntheticsDecorationProvider`, because it's not from SemanticDBs synthetics.
- [ ] Testing
- [ ] Cache the result, and try to reduce the number of calculation/profiling
- [ ] (Hopefully) migrate to Inlay Hint (but LSP4j doesn't support it until LSP4J 0.13.0), and support https://github.com/scalameta/metals/issues/4059

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
@@ -47,7 +47,8 @@ final class SyntheticsDecorationProvider(
     focusedDocument: () => Option[AbsolutePath],
     clientConfig: ClientConfiguration,
     userConfig: () => UserConfiguration,
-    trees: Trees
+    trees: Trees,
+    compilers: m.internal.metals.Compilers
 )(implicit ec: ExecutionContext)
     extends SemanticdbFeatureProvider {
   private object Document {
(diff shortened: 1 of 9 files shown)


kpodsiad:
This will be awesome @tanishiking! 🚀

tgodzik:
> Refactoring (maybe we shouldn't add the logic into SyntheticsDecorationProvider, because it's not from SemanticDBs synthetics.

Semanticdb decorations are quite fast, so I wouldn't add it to that provider. We could do a separate CompilerDecorationsProvider, but we would need to republish everything together :thinking: 

>  Cache the result, and try to reduce the number of calculation/profiling

One of the main advantages of using the presentation compiler is that we can make it work in a non compiling code, but on each change we would need to recalculate and I am not sure how fast that would be. Maybe we could only republish on save?

Would be cool to try and benchmark this. Also you can stress test on the ScalametaParser file in scalameta/scalameta

>  Should we show the decorations only for "simple" type?

How does it work in Intellij? We could define simple types as those having their declaration in the apply maybe?

tanishiking:
It's already pretty old, let me close this and create another fresh one :)

PR #3917, opened 2022-05-13
fix: symbolSearch - support encoded names from classpath

dos65:
Classpath search is based on filenames inside jars which might be
encoded. Like: `<:<` -> `$less$colon$less`.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/metals/PackageIndex.scala
+++ b/mtags/src/main/scala/scala/meta/internal/metals/PackageIndex.scala
@@ -13,6 +13,7 @@ import java.util.jar.JarFile
 import java.util.logging.Level
 import java.util.logging.Logger
 
+import scala.reflect.NameTransformer
 import scala.util.Properties
 import scala.util.control.NonFatal
 
@@ -56,7 +57,7 @@ class PackageIndex() {
   def addMember(pkg: String, member: String): Unit = {
     if (!member.contains("module-info.class")) {
       val members = packages.computeIfAbsent(pkg, enterPackage)
-      members.add(member)
+      members.add(NameTransformer.decode(member))
     }
   }
 
--- a/tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala
@@ -360,4 +360,24 @@ class WorkspaceSymbolLspSuite extends BaseLspSuite("workspace-symbol") {
       )
     } yield ()
   }
+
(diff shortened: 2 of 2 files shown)


kpodsiad (commented):
On discord Olaf wrote `metals needs to encode the query to remove non-alphanumeric characters`. Here, you went with yet another approach, namely you're decoding classfiles from classpath. Could you elaborate about both approaches?

mtags/src/main/scala/scala/meta/internal/metals/BinaryNameDecoder.scala
```diff
@@ -0,0 +1,60 @@
+package scala.meta.internal.metals
+
+import scala.annotation.tailrec
+
+object BinaryNameDecoder {
```
tanishiking, line 5: Can we use compiler defined `NameTransfoer`'s decode methods?
- scala2: https://github.com/scala/scala/blob/7952071b88a85e2502e9ecd4e8bd57040d9448f5/src/library/scala/reflect/NameTransformer.scala#L98-L103
- scala3: https://github.com/lampepfl/dotty/blob/31be2b0843d6bd9317c0ef9f2af9d8ec93271a55/compiler/src/dotty/tools/dotc/util/NameTransformer.scala#L86-L89

dos65: Nice catch!
I forgot that server has a compiler dependency

dos65:
@kpodsiad 
> On discord Olaf wrote metals needs to encode the query to remove non-alphanumeric characters. Here, you went with yet another approach, namely you're decoding classfiles from classpath. Could you elaborate about both approaches?

The opposite approach is hard to implement because [Fuzzy treats `$` as a delimiter](https://github.com/scalameta/metals/blob/0087d07f72a4a865a5dbd72813807c931d385b77/mtags/src/main/scala/scala/meta/internal/metals/Fuzzy.scala#L301) and it's required to decode name before indexing

@tanishiking Thx for suggestion, switched it on a decoder from scala.reflect

mtags/src/main/scala/scala/meta/internal/metals/PackageIndex.scala
```diff
@@ -55,8 +55,9 @@ class PackageIndex() {
 
   def addMember(pkg: String, member: String): Unit = {
     if (!member.contains("module-info.class")) {
+      val decoded = BinaryNameDecoder.decode(member)
```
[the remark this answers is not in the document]
dos65: nope. I've checked that this thing is used only in symbol search.

kpodsiad (approved):

PR #3918, opened 2022-05-14
chore, docs: simplify cats-effect stacktrace regex, add docstring

kpodsiad:
I forgot to add some docstring and during this process I realized that regex is overcomplicated without reason so I simplified it a bit.


tanishiking (approved):
LGTM! Now it's better understandable 👍

PR #3919, opened 2022-05-14
chore: update scala3 version in welcome msg

kpodsiad:
Use current scala3 version instead of hardcoded `3.0.0`

Screenshot 2022-05-14 at 10 47 05


tanishiking (approved):
Nice catch!

PR #3924, opened 2022-05-17
chore(docs): ensure metals_2.12 is replaced with metals_2.13

ckipp01:
Since Metals is now published for 2.13, this replaces the last remaining
places in the docs where it mentions 2.12 in any instructions.


tgodzik (approved):
Good catch! Thanks!

PR #3925, opened 2022-05-18
chore: speedup lsp tests

dos65:
`ClasspathSearch.fromPackages` call with an empty clasppath takes
about a second on my machine.
In theory, replacing it by `empty` should save 4-7 minutes on CI.


tgodzik (approved):

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -1688,4 +1691,37 @@ object TestingServer {
       treeViewProvider = Some(true),
       slowTaskProvider = Some(true)
     )
+
+  // In most cases lsp unit tests have the same classpath: one jvm + scala-library without any external deps
+  // Than means that in all these tests ClasspathSearch should be the same and we can calculate it once per run
+  // instead of once per test
+  val testingClasspathSearchIndexer: ClasspathSearch.Indexer =
```
dos65, line 1698: Added an additional improvement by reducing the time for ClasspathSearch indexing

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -172,8 +172,10 @@ class Compilers(
   }
 
   def load(paths: Seq[AbsolutePath]): Future[Unit] =
-    if (Testing.isEnabled) Future.successful(())
-    else {
+    if (Testing.isEnabled) {
+      scribe.info(s"LOAD $paths")
```
tgodzik, line 176: Is this log needed? Could we maybe adjust the message to show what we are printing?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -163,8 +165,9 @@ class MetalsLanguageServer(
   private val recentlyFocusedFiles = new ActiveFiles(time)
   private val languageClient = new DelegatingLanguageClient(NoopLanguageClient)
   var userConfig: UserConfiguration = UserConfiguration()
-  val excludedPackageHandler: ExcludedPackagesHandler =
-    new ExcludedPackagesHandler(userConfig.excludedPackages)
+  var excludedPackageHandler: ExcludedPackagesHandler =
```
tgodzik, line 168: Why do we need to have it as a var ? Can't we keep the state inside ExcludedPackagesHandler?

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1049,10 +1053,12 @@ class MetalsLanguageServer(
     // Update in-memory buffer contents from LSP client
     buffers.put(path, params.getTextDocument.getText)
 
-    packageProvider
-      .workspaceEdit(path)
-      .map(new ApplyWorkspaceEditParams(_))
-      .foreach(languageClient.applyEdit)
+    timerProvider.timedThunk("package", true) {
```
tgodzik, line 1056: Is this needed still? Second parameter is anyway true
```suggestion
    timerProvider.timedThunk("package") {
```

tests/unit/src/main/scala/tests/TestingServer.scala
tgodzik: Aren't there any unit tests that have dependencies and the classpath search would be influenced by this?

tgodzik (changes requested):
Great to see these improvements! :tada: 

I added a few questions about the approach.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
dos65: oops, just leftovers from investigations. removed

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
dos65: We can. I did it in the same manner as other things in MetalsLanguageServer

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
dos65: removed

tests/unit/src/main/scala/tests/TestingServer.scala
dos65: There are and they work fine with this change.
Cache is done by a key: dependency jars + excludedPackages conf + bucket size.
Unit test with a unique dependencies creates a new index but most of them reuses the default.

tanishiking: Can we add your comment to the code comment?

> There are and they work fine with this change.
> Cache is done by a key: dependency jars + excludedPackages conf + bucket size.
> Unit test with a unique dependencies creates a new index but most of them reuses the default.

tgodzik (approved):
LGTM!

tests/unit/src/main/scala/tests/TestingServer.scala
dos65: updated this comment

tanishiking (approved):
LGTM :+1:

PR #3930, opened 2022-05-19
Include proper names when using implement all completion and code action

tgodzik:
Previously, we would use x$0 etc for Java parameter names, which was caused by completions not resolving documentation by design. Now, we resolve the names for 'implement all' completion and always resolve docs for 'implement all' code action.

This turned out much harder to fix than I estimated.

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
@@ -24,6 +24,21 @@ trait OverrideCompletions { this: MetalsGlobal =>
       val detail: String
   ) extends ScopeMember(sym, NoType, true, EmptyTree)
 
+  class ImplementAllMember(
+      filterText: String,
+      edit: l.TextEdit,
+      detail: String,
+      additionalTextEdits: List[l.TextEdit],
+      val additionalSymbols: List[Symbol]
+  ) extends OverrideDefMember(
+        label = "Implement all members",
+        edit,
+        filterText,
+        sym = completionsSymbol("implement"),
+        additionalTextEdits,
+        detail
+      )
+
   /**
    * An `override def` completion to implement methods from the supertype.
    *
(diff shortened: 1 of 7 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
```diff
@@ -24,6 +24,21 @@ trait OverrideCompletions { this: MetalsGlobal =>
       val detail: String
   ) extends ScopeMember(sym, NoType, true, EmptyTree)
 
+  class OverrideAllMember(
```
tanishiking, line 27: Can we rename it to `ImplementAllMember`? I feel like there's a subtle difference between all in "OverrideAll" and "ImplementAll"

- "all" in `OverrideAll` means all the members we can override.
- "all" in `ImplementAll` means all the **abstract** members we can implement (override).

mtags/src/main/scala/scala/meta/internal/pc/CompletionItemData.scala
```diff
@@ -25,4 +32,6 @@ object CompletionItemData {
   def empty: CompletionItemData = CompletionItemData("", "")
   // This is an `override def` completion item.
   val OverrideKind: java.lang.Integer = 1
+  // This is a completion implementing all abstract members
+  val OverrideAllKind: java.lang.Integer = 2
```
tanishiking, line 36:
```suggestion
  val ImplementAllKind: java.lang.Integer = 2
```
?

tanishiking (commented):
I left minor comments on naming, but otherwise LGTM!

mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
tgodzik: Good idea! Renamed

mtags/src/main/scala/scala/meta/internal/pc/CompletionItemData.scala
tgodzik: Done!

dos65 (approved):
:+1:

Issue #3929, opened 2022-05-19
Hover on selection of transparent inline method application


prolativ:
### Describe the bug

With scala 3.1.2:

```scala
trait Foo
class Bar extends Foo

transparent inline def foo(i: Int): Foo = new Bar
val bar = foo(1)
```

Hovering on selected `foo(1)` shows

```scala
Expression type:
Foo

Symbol signature:
def foo(i: Int): Foo
```

### Expected behavior

It should show
```scala
Expression type:
Bar

Symbol signature:
transparent inline def foo(i: Int): Foo
```

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.5

PR #3931, opened 2022-05-19
Fix expression type for inlined methods and show all flags for methods

tgodzik:
Previously, when an expression was inlined and was transparent we would not show the proper type as indicated by transparent. Besides that no flags were shown for methods. Now, we also check for Inlined tree node when expanding and also show the proper flag string.

Fixes https://github.com/scalameta/metals/issues/3929


tanishiking (approved):
LGTM!

PR #3932, opened 2022-05-20
Fix nightlies tests and uncomment SignaturePat suite tests

tgodzik:
Two main changes here:
- updates to the signature help and Scala 3 (a lot of fixes and a bug reported upstream)
- fixed the failing test for CompletionSuite (it will probably fail once again when the proper fix is introduced into Dotty, but I hoped the fix would be made much sooner)
- added testing for Scala 3 unapply signature help with some more bugs reported upstream


dos65 (approved):
LGTM

I agree about the fix with for completion test.
Because of it we missed that signature tests started failing

PR #3934, opened 2022-05-20
feature: Show scaladocs for signature help

tgodzik:
Previously, we wouldn't show any documentation for signature help. Now, we properly look for it the same way as in hover and completions

I also extracted signature help from the ScalaPresentationCompiler class.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/SemanticdbSymbols.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/SemanticdbSymbols.scala
@@ -103,7 +103,11 @@ object SemanticdbSymbols:
           decls0 ++ sym.owner.companionClass.info.decls.lookupAll(sym.name)
         else decls0
       end decls
-      val alts = decls.filter(_.isOneOf(Method | Mutable)).toList.reverse
+      // private java constructors do not have a symbol created
+      val alts = decls
+        .filter(d => d.isOneOf(Method | Mutable) && !d.is(Private))
+        .toList
+        .reverse
       def find(filter: Symbol => Boolean) = alts match
         case notSym :: rest if !filter(notSym) =>
           val idx = rest.indexWhere(filter)
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
@@ -1,0 +1,154 @@
+package scala.meta.internal.pc
+
+import scala.collection.JavaConverters.*
+
+import scala.meta.internal.mtags.BuildInfo
+import scala.meta.internal.mtags.MtagsEnrichments.*
+import scala.meta.internal.semver.SemVer
+import scala.meta.pc.OffsetParams
+import scala.meta.pc.SymbolDocumentation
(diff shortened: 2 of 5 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/SignatureHelpProvider.scala
```diff
@@ -0,0 +1,144 @@
...
+
+      }
+    new l.SignatureHelp(
+      signatureInfos.map(signatureToSignatureInformation).asJava,
+      callableN,
+      paramN
+    )
+  end signatureHelp
+
+  private def withDocumentation(
```
tgodzik, line 67: This is the only new method added along with the above:
```scala
val signatureInfos =
      alternatives.flatMap { denot =>
        val doc = search.symbolDocumentation(denot.symbol)
        (doc, Signatures.toSignature(denot)) match
          case (Some(info), Some(signature)) =>
            withDocumentation(info, signature)
          case (_, sig) => sig
```

mtags/src/main/scala-3/scala/meta/internal/pc/SemanticdbSymbols.scala
```diff
@@ -103,7 +103,11 @@ object SemanticdbSymbols:
           decls0 ++ sym.owner.companionClass.info.decls.lookupAll(sym.name)
         else decls0
       end decls
-      val alts = decls.filter(_.isOneOf(Method | Mutable)).toList.reverse
+      // private java constructors do not have a symbol created
+      val alts = decls
+        .filter(d => d.isOneOf(Method | Mutable) && !d.is(Private))
```
tgodzik, line 108: This is how it's done in the Javadoc indexer and in semanticdb for Scala 2.

tests/cross/src/test/scala/tests/pc/SignatureHelpDocSuite.scala
```diff
@@ -125,6 +131,16 @@ class SignatureHelpDocSuite extends BaseSignatureHelpSuite {
             |        ^^^^^^^^^^^^^
             |  @param ifEmpty String the expression to evaluate if empty.
             |  @param f the function to apply if nonempty.
+            |""".stripMargin,
+      "3" ->
+        s"""|$foldLatestDocs
+            |**Parameters**
+            |- `f`: the function to apply if nonempty.
+            |- `ifEmpty`: the expression to evaluate if empty.
+            |fold[B](ifEmpty: => B)(f: A => B): B
```
dos65, line 140: It seems that we need  `seenFrom` as in [HoverProvider](https://github.com/scalameta/metals/blob/806786093d6f3cf8f9e4aa490dbe3a94662f1230/mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala#L140) but we can do it later

dos65 (approved):
LGTM

tests/cross/src/test/scala/tests/pc/SignatureHelpDocSuite.scala
tgodzik: I will do an update with that!

tgodzik: Follow up here: https://github.com/scalameta/metals/pull/3941

Issue #3937, opened 2022-05-22
Out of bounds cursor position given when creating new empty scala file


ckipp01:
### Describe the bug

When a user attempts to create a new scala file using the `scala-file` id (Empty file), the `metals-goto-location` that is returned via the `metals/executeClientCommand` has a position that is out of bounds in the file that was actually created. This throws errors in some clients since the position is not valid. Here is an example of what a `nvim-metals` user sees:

```
Error executing vim.schedule lua callback: ...vim/HEAD-95580f3/share/nvim/runtime/lua/vim/lsp/util.lua:1047: Cursor position outside buffer
stack traceback:
        [C]: in function 'nvim_win_set_cursor'
        ...vim/HEAD-95580f3/share/nvim/runtime/lua/vim/lsp/util.lua:1047: in function 'jump_to_location'
        ...te/pack/packer/start/nvim-metals/lua/metals/handlers.lua:45: in function 'handler'
        ...r/neovim/HEAD-95580f3/share/nvim/runtime/lua/vim/lsp.lua:824: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
```

Here is the full trace:
```
[Trace - 10:55:54 AM] Received request 'workspace/executeCommand - (5)'
Params: {
  "command": "metals.new-scala-file",
  "arguments": [
    null,
    null,
    null
  ]
}


[Trace - 10:55:54 AM] Sending request 'metals/quickPick - (1)'
Params: {
  "items": [
    {
      "id": "scala-file",
      "label": "Empty file"
    },
    {
      "id": "scala-class",
      "label": "Class"
    },
    {
      "id": "scala-case-class",
      "label": "Case Class"
    },
    {
      "id": "scala-object",
      "label": "Object"
    },
    {
      "id": "scala-trait",
      "label": "Trait"
    },
    {
      "id": "scala-package-object",
      "label": "Package Object"
    },
    {
      "id": "scala-worksheet",
      "label": "Worksheet"
    },
    {
      "id": "ammonite-script",
      "label": "Ammonite Script"
    },
    {
      "id": "scala-enum",
      "label": "Enum"
    }
  ],
  "placeHolder": "Select the kind of file to create"
}


[Trace - 10:55:55 AM] Received response 'metals/quickPick - (1)' in 1200ms
Result: {
  "itemId": "scala-file"
}
Error: null


[Trace - 10:55:55 AM] Sending request 'metals/inputBox - (2)'
Params: {
[...]

PR #3942, opened 2022-05-24
fix: avoid out of bounds for editors that treat \n as a line end

ckipp01:
So this isn't really a bug, but sort of an interesting difference in how
different editors handle \n -- whether they treat it as a line
terminator or an actual new blank line when at the end of a file.

Currently when you use the new scala file provider the text you end up
getting is just the package followed by multiple new lines. Some editors
will interpret this as the package line and then two new blank lines,
while others (like vim) will treat this as a package line, one blank
line, and then a line ending. Then when the Metals calculates the
position of the cursor it will be out of bounds for vim. This small
change just ensures the `@@` for the new blank file template is not on
the last line meaning it won't be out of bounds. There will be a slight
difference in VS Code though as there will be an extra new line, but the
cursor won't be on that new line. I figured this was a fair compromise
to ensure it still worked on all editors. Is this alright?

refs: #3937

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/newScalaFile/NewFileProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/newScalaFile/NewFileProvider.scala
@@ -209,7 +209,8 @@ class NewFileProvider(
       .packageStatement(path)
       .map(_.fileContent)
       .getOrElse("")
-    val template = s"$pkg@@"
+    val template = s"""|$pkg@@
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
Looks sensible! The additional new line might actually be useful even in VS Code

dos65 (commented):
Let me check how it works in vscode.
I remember I touched this part because it was producing two \n and it was inconvenient.

PR #3946, opened 2022-05-24
Fix SBT version check in Doctor

iantabolt:
Previously, this sbt version error in Metals Doctor was wrongly suggesting I upgrade sbt to 1.3.2:

image

This is also paired with a startup message saying "Multiple problems detected in your build."

~It looks like there was a bug where the sbt version check was doing the same thing as the Scala version check (checking `ScalaVersions.isDeprecatedScalaVersion(scalaVersion)`) . I updated this to check the `sbtVersion` against `1.3.2` as the error message implies.~

~**Note** I fixed the check for `UnsupportedSbtVersion` but removed the checks for `DeprecatedSbtVersion` and `FutureSbtVersion` since they were also broken and it isn't clear what those version bounds should be... I can add them back if those bounds exist~

---

Update: Turns out the message was just out of date and everything was working as expected 😃. I moved the minimum sbt version to a constant alongside the deprecated Scala 2 versions, so hopefully it will be more obvious to update in the future.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
@@ -59,8 +59,6 @@ class ProblemResolver(
     var misconfiguredProjects = 0
     var misconfiguredTestFrameworks = 0
     var unsupportedSbt = false
-    var deprecatedSbt = false
-    var futureSbt = false
 
     for {
       target <- scalaTargets
@@ -72,9 +70,7 @@ class ProblemResolver(
         case FutureScalaVersion(version) => futureVersions += version
         case _: SemanticDBDisabled => misconfiguredProjects += 1
         case _: MissingSourceRoot => misconfiguredProjects += 1
-        case UnsupportedSbtVersion => unsupportedSbt = true
-        case DeprecatedSbtVersion => deprecatedSbt = true
-        case FutureSbtVersion => futureSbt = true
+        case _: UnsupportedSbtVersion => unsupportedSbt = true
         case MissingJdkSources(_) => misconfiguredProjects += 1
         case OutdatedJunitInterfaceVersion => misconfiguredTestFrameworks += 1
         case OutdatedMunitInterfaceVersion => misconfiguredTestFrameworks += 1
@@ -112,12 +108,8 @@ class ProblemResolver(
       None
     }
 
-    val deprecatedSbtMessage =
-      if (deprecatedSbt) Some(Messages.DeprecatedSbtVersion.message) else None
     val unsupportedSbtMessage =
-      if (deprecatedSbt) Some(Messages.UnsupportedSbtVersion.message) else None
-    val futureSbtMessage =
-      if (deprecatedSbt) Some(Messages.FutureSbtVersion.message) else None
+      if (unsupportedSbt) Some(Messages.UnsupportedSbtVersion.message) else None
 
     val semanticdbMessage =
       if (
@@ -153,9 +145,7 @@ class ProblemResolver(
       deprecatedMessage,
       unsupportedMessage,
       futureMessage,
-      deprecatedSbtMessage,
       unsupportedSbtMessage,
-      futureSbtMessage,
       semanticdbMessage,
       testFrameworks
     ).flatten
@@ -188,11 +178,12 @@ class ProblemResolver(
       mtagsResolver.isSupportedScalaVersion(version)
 
     val scalaVersionProblem = scalaTarget.scalaVersion match {
-      case version if !isSupportedScalaVersion(version) && scalaTarget.isSbt =>
-        if (ScalaVersions.isFutureVersion(version))
-          Some(FutureSbtVersion)
-        else
-          Some(UnsupportedSbtVersion)
+      case _ if scalaTarget.isSbt =>
+        scalaTarget.sbtVersion match {
+          case Some(version) if !SemVer.isCompatibleVersion("1.3.2", version) =>
+            Some(UnsupportedSbtVersion(version))
+          case _ => None
+        }
       case version if !isSupportedScalaVersion(version) =>
         if (ScalaVersions.isFutureVersion(version))
           Some(FutureScalaVersion(version))
@@ -210,11 +201,6 @@ class ProblemResolver(
           if !scalaTarget.isSourcerootDeclared && !ScalaVersions
             .isScala3Version(scalaTarget.scalaVersion) =>
         Some(MissingSourceRoot(workspace.scalaSourcerootOption))
-      case version
-          if ScalaVersions.isDeprecatedScalaVersion(
-            version
-          ) && scalaTarget.isSbt =>
-        Some(DeprecatedSbtVersion)
       case version if ScalaVersions.isDeprecatedScalaVersion(version) =>
         Some(DeprecatedScalaVersion(version))
       case _ => None
--- a/metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
@@ -59,6 +59,8 @@ class ProblemResolver(
     var misconfiguredProjects = 0
     var misconfiguredTestFrameworks = 0
     var unsupportedSbt = false
+    var deprecatedSbt = false
+    var futureSbt = false
 
     for {
       target <- scalaTargets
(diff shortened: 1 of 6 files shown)


metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
```diff
@@ -188,11 +178,12 @@ class ProblemResolver(
...
-        if (ScalaVersions.isFutureVersion(version))
-          Some(FutureSbtVersion)
-        else
-          Some(UnsupportedSbtVersion)
+      case _ if scalaTarget.isSbt =>
+        scalaTarget.sbtVersion match {
+          case Some(version) if !SemVer.isCompatibleVersion("1.3.2", version) =>
+            Some(UnsupportedSbtVersion(version))
+          case _ => None
+        }
```
tanishiking, line 186: Looks like this case makes the following cases (like `case version if !isSupportedScalaVersion(version) =>` unreachable for sbt project.

I guess we need to move the `scalaTarget.sbtVersion` check into `case` like 

```scala
case _ if scalaTarget.sbtVersion match { ... } => 
```

tanishiking: Or maybe `findProblem` doesn't need to return `Option[ScalaProblem]`, but it can return something like `List[ScalaProblem]` 🤔 , because sbt  deprecation problem and unsupported scalaversion problems are independent problems.
What do you think about it? @tgodzik @kpodsiad (someone who might be familiar with Metals Doctor)

Anyway, you don't need to work on it in this PR 👍

metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
```diff
@@ -210,11 +201,6 @@ class ProblemResolver(
           if !scalaTarget.isSourcerootDeclared && !ScalaVersions
             .isScala3Version(scalaTarget.scalaVersion) =>
         Some(MissingSourceRoot(workspace.scalaSourcerootOption))
-      case version
-          if ScalaVersions.isDeprecatedScalaVersion(
-            version
-          ) && scalaTarget.isSbt =>
-        Some(DeprecatedSbtVersion)
```
tanishiking, line 217: :+1:
Hmm, I'm not sure why did we have this case.

metals/src/main/scala/scala/meta/internal/metals/doctor/ScalaProblem.scala
```diff
@@ -72,17 +72,9 @@ case class MissingSourceRoot(sourcerootOption: String) extends ScalaProblem {
...
-    "Code navigation is not supported for this sbt version, please upgrade to at least 1.3.2 and reimport the build"
-}
-case object DeprecatedSbtVersion extends ScalaProblem {
-  override def message: String =
-    "Code navigation might not be supported in the future for this sbt version, please upgrade to at least 1.3.2 and reimport the build"
-}
-case object FutureSbtVersion extends ScalaProblem {
-  override def message: String =
-    "Code navigation for this sbt version is not yet supported"
+    s"Code navigation might not be supported sbt $version, please upgrade to at least 1.3.2 and reimport the build"
```
tanishiking, line 77: 👍

metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
```diff
@@ -188,11 +178,12 @@ class ProblemResolver(
       mtagsResolver.isSupportedScalaVersion(version)
 
     val scalaVersionProblem = scalaTarget.scalaVersion match {
-      case version if !isSupportedScalaVersion(version) && scalaTarget.isSbt =>
-        if (ScalaVersions.isFutureVersion(version))
-          Some(FutureSbtVersion)
-        else
-          Some(UnsupportedSbtVersion)
```
tanishiking, line 195: I think it's fine to drop this case as we have `case version if !isSupportedScalaVersion(version) =>` later anyway 👍

tests/unit/src/test/scala/tests/troubleshoot/ProblemResolverSuite.scala
```diff
@@ -60,30 +58,16 @@ class ProblemResolverSuite extends FunSuite {
...
-    FutureSbtVersion.message,
-    sbtVersion = Some("1.6.0")
-  )
-
   checkRecommendation(
     "ok-sbt-version",
     scalaVersion = BuildInfo.scala212,
     "",
-    sbtVersion = Some("1.6.0")
+    sbtVersion = Some("1.3.12")
```
tanishiking, line 70: Do you mean `1.3.2`?

tests/unit/src/test/scala/tests/troubleshoot/ProblemResolverSuite.scala
```diff
@@ -60,30 +58,16 @@ class ProblemResolverSuite extends FunSuite {
...
-    DeprecatedSbtVersion.message,
-    sbtVersion = Some("1.3.0")
-  )
-
-  checkRecommendation(
-    "future-sbt-version",
-    scalaVersion = "2.12.51",
-    FutureSbtVersion.message,
-    sbtVersion = Some("1.6.0")
-  )
```
tanishiking, line 80: can we leave this test to check `2.12.51` is reported as `UnsupportedScalaVersion`?

tests/unit/src/test/scala/tests/troubleshoot/ProblemResolverSuite.scala
```diff
@@ -60,30 +58,16 @@ class ProblemResolverSuite extends FunSuite {
...
+    UnsupportedSbtVersion("1.2.0").message,
     sbtVersion = Some("1.2.0")
   )
 
-  checkRecommendation(
-    "deprecated-sbt-version",
-    scalaVersion = "2.12.8",
-    DeprecatedSbtVersion.message,
-    sbtVersion = Some("1.3.0")
-  )
```
tanishiking, line 73: Can we leave this test to check `1.3.0` is unsupported sbt version?

tanishiking (changes requested):
Thank you for the PR, I understand the point and the main change looks great!
I just left some comments.

metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
```diff
@@ -188,11 +178,12 @@ class ProblemResolver(
       mtagsResolver.isSupportedScalaVersion(version)
 
     val scalaVersionProblem = scalaTarget.scalaVersion match {
-      case version if !isSupportedScalaVersion(version) && scalaTarget.isSbt =>
```
tgodzik, line 191: So this was actually done intentionally to check for Scala version. Metals knows which Scala versions it supports, but it doesn't know anything about sbt versions.

However, if sbt version uses a Scala version that Metals supports then  we also support that sbt version. The same with everything else.

tgodzik (changes requested):
Thanks for the contribution! This was actually all done intentionally since it's hard to keep the list of supported sbt versions. It's not easy to check the actual Scala version it uses unless we get that information from BSP.

Since we deprecated a new Scala 2 version, we should also change the recommendation to the first sbt version that that uses a non deprecated Scala 2 version.

We should recommend 1.4.0 version now.
`val recommended = "1.4.0"`

It's probably good to add a message that when we deprecate Scala 2 version, we should update the recommendation

iantabolt:
That explains it! I'll revert this and change the message instead, and meanwhile upgrade my sbt to 1.4.0 😀

metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
iantabolt: Ok so if I understand correctly, when a target has `isSbt` the `scalaVersion` is not the project's Scala version but actually the Scala version that sbt was built on. In my case, my project uses Scala 2.12.14 (not deprecated by Metals) but sbt 1.3.12 uses Scala 2.12.10 (deprecated by Metals). So when checking the sbt Scala version, we want to surface a message to the user that says "Please upgrade sbt" even though the underlying problem is sbt's Scala version.

> maybe findProblem doesn't need to return Option[ScalaProblem], but it can return something like List[ScalaProblem]  

I think it is an `Option` because each target can have one recommendation. So you could have a Scala target that says "Please upgrade Scala" and the sbt target that says "Please upgrade sbt" but each target can only say one thing.

metals/src/main/scala/scala/meta/internal/metals/doctor/ProblemResolver.scala
```diff
@@ -115,9 +115,9 @@ class ProblemResolver(
     val deprecatedSbtMessage =
       if (deprecatedSbt) Some(Messages.DeprecatedSbtVersion.message) else None
     val unsupportedSbtMessage =
-      if (deprecatedSbt) Some(Messages.UnsupportedSbtVersion.message) else None
+      if (unsupportedSbt) Some(Messages.UnsupportedSbtVersion.message) else None
     val futureSbtMessage =
-      if (deprecatedSbt) Some(Messages.FutureSbtVersion.message) else None
+      if (futureSbt) Some(Messages.FutureSbtVersion.message) else None
```
iantabolt, line 120: This must be a bug, right? (Flags `unsupportedSbt` and `futureSbt` were set but never used)

Issue #3938, opened 2022-05-22
Implicit parameter is not visible in yield clause


kpodsiad:
Following code
```scala
//> using scala "2.13.8"

class Evidence

final case class DataType(number: Int) {
  def next(implicit evidence: Evidence): Int = number + 1
}

object Example extends App {

  implicit val evidence: Evidence = ???

  for {
    number <- Option(5)
  } yield DataType(number).next

  Option(5).map(DataType(_).next)

}
```
yields this

<img width="504" alt="Screenshot 2022-05-22 at 18 38 37" src="https://user-images.githubusercontent.com/37124721/169706115-62f9d4c0-a2f5-4b7e-b9ca-10df381f35fe.png">

`evidence` is visible in the ordinary map, but there is no implicit parameter in yield clause.

PR #3948, opened 2022-05-25
Fix issues with synthetics in for comprehension's yield

tgodzik:
Previously, we would not take into account sythetics inside for comprehension trees, which could lead to missing decorations inside the yield statement. Now, we recursively go into those trees to get the yield statement.

Fixes https://github.com/scalameta/metals/issues/3938

I recommend reviewing without the whitespaces

### DIFF
--- a/tests/unit/src/test/scala/tests/decorations/SyntheticDecorationsLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/decorations/SyntheticDecorationsLspSuite.scala
@@ -696,4 +696,65 @@ class SyntheticDecorationsLspSuite extends BaseLspSuite("implicits") {
       )
     } yield ()
   }
+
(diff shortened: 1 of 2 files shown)


tests/unit/src/test/scala/tests/decorations/SyntheticDecorationsLspSuite.scala
```diff
@@ -696,4 +696,64 @@ class SyntheticDecorationsLspSuite extends BaseLspSuite("implicits") {
       )
     } yield ()
   }
+
+  test("issue-3938") {
```
kpodsiad, line 700: Suggestion: maybe paste link to issue next to test and give it human-readable name like `for-yield`?

kpodsiad (approved):

Issue #3945, opened 2022-05-24
Methods with empty parens argument lists are suggested incorrectly


keynmol:
### Describe the bug

Scala 3.1.2

When a method is defined with empty parens, the suggestion from autocomplete is incorrect:
![2022-05-24 15 53 31](https://user-images.githubusercontent.com/1052965/170066571-6cf4cf0e-a681-4c7b-9492-97d557bda320.gif)

### Expected behavior

The method in question is suggested as `now()` and inserted as such.

### Operating system

macOS

### Editor/Extension

Nvim (nvim-metals)

### Version of Metals

v0.11.5+93-e93e94dd-SNAPSHOT

### Extra context or search terms

scala3, autocomplete, parens, argument, method

PR #3959, opened 2022-05-26
feature: [Scala 3] Add snippet completions

tgodzik:
Previously, when completing a function we would only write the name even if the method needs parameters. Now if the client supports snippets we additionally add parenthesis where aplicable.

I also added a completion for apply in case of companion objects.

Fixes https://github.com/scalameta/metals/issues/3945

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -3,9 +3,11 @@ package completions
 
 import scala.collection.mutable
 
+import scala.meta.internal.mtags.BuildInfo
 import scala.meta.internal.mtags.MtagsEnrichments.*
 import scala.meta.internal.pc.IdentifierComparator
 import scala.meta.internal.pc.completions.KeywordsCompletions
+import scala.meta.internal.semver.SemVer
 import scala.meta.pc.*
 
 import dotty.tools.dotc.ast.tpd.*
@@ -41,6 +43,10 @@ class Completions(
 ):
 
   implicit val context: Context = ctx
+
+  // versions prior to 3.1.0 sometimes didn't manage to detect properly Java objects
+  val canDetectJavaObjectsCorrectly =
+    SemVer.isLaterVersion("3.1.0", BuildInfo.scalaCompilerVersion)
 
   private lazy val shouldAddSnippet = path match
     /* In case of `method@@()` we should not add snippets and the path
@@ -113,9 +119,16 @@ class Completions(
       label: String,
       toCompletionValue: (String, Symbol, Option[String]) => CompletionValue
   ): List[CompletionValue] =
+    // workaround for earlier versions that force correctly detecting Java flags
+    def isJavaDefined = if canDetectJavaObjectsCorrectly then
+      sym.is(Flags.JavaDefined)
+    else
+      sym.info
+      sym.is(Flags.JavaDefined)
+
     // find the apply completion that would need a snippet
     val methodSymbols =
-      if shouldAddSnippet && sym.is(Flags.Module) && !sym.is(Flags.JavaDefined)
+      if shouldAddSnippet && sym.is(Flags.Module) && !isJavaDefined
       then
         val applSymbols = sym.info.member(nme.apply).allSymbols
         sym :: applSymbols
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -12,12 +12,14 @@ import dotty.tools.dotc.ast.tpd.*
 import dotty.tools.dotc.ast.untpd
 import dotty.tools.dotc.core.Constants.Constant
 import dotty.tools.dotc.core.Contexts.*
+import dotty.tools.dotc.core.Denotations.*
 import dotty.tools.dotc.core.Flags
 import dotty.tools.dotc.core.Flags.*
 import dotty.tools.dotc.core.NameOps.*
 import dotty.tools.dotc.core.Names.*
 import dotty.tools.dotc.core.StdNames
 import dotty.tools.dotc.core.StdNames.*
+import dotty.tools.dotc.core.SymDenotations.NoDenotation
 import dotty.tools.dotc.core.Symbols.*
 import dotty.tools.dotc.core.Types.*
 import dotty.tools.dotc.interactive.Completion
@@ -55,7 +57,7 @@ class Completions(
           case _ =>
             val (_, compilerCompletions) = Completion.completions(pos)
             val (compiler, result) = compilerCompletions
-              .flatMap(CompletionValue.fromCompiler)
+              .flatMap(toCompletionValues)
               .filterInteresting()
             (advanced ++ compiler, result)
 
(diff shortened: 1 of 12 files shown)


tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -312,47 +312,6 @@ class CompletionSuite extends BaseCompletionSuite {
            |wait(): Unit
            |wait(x$0: Long): Unit
            |wait(x$0: Long, x$1: Int): Unit
-           |""".stripMargin,
-      "3.2" ->
```
tgodzik, line 316: Fixed in https://github.com/lampepfl/dotty/pull/15283

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -65,6 +67,77 @@ class Completions(
...
+        case Nil => None
+        case List(Nil) => Some("()")
+        case _ =>
+          val onlyImplicitOrTypeParams = paramss.forall(
+            _.exists { sym =>
+              sym.isType || sym.is(Implicit) || sym.is(Given)
+            }
+          )
+          if onlyImplicitOrTypeParams then None
+          else Some("($0)")
```
tanishiking, line 116: I'm wondering can we make something like this 

```suggestion
          else if shouldAddSnippet && config.isCompletionSnippetsEnabled then Some("($0)")
          else Some("()")
```

and the top if predicate makes 

```diff
-    if shouldAddSnippet &&
-      config.isCompletionSnippetsEnabled &&
-      methodSymbol.is(Flags.Method)
+      methodSymbol.is(Flags.Method)
```

so that Metals can add `()` for methods defined with empty parens even if clients support snippet.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -65,6 +67,77 @@ class Completions(
...
+
+    completion.symbols.flatMap(
+      completionsWithSuffix(_, completion.label, shouldAddSnippet)
+    )
+  end toCompletionValues
+
+  inline private def undoBacktick(label: String): String =
+    label.stripPrefix("`").stripSuffix("`")
+
+  private def findApplySymbols(denot: Denotation): List[Symbol] =
```
tanishiking, line 89:
```suggestion
  private def allOverloadedSymbols(denot: Denotation): List[Symbol] =
```

or something?
Indeed, this method is used for collecting all overloaded `apply` methods, but this method itself is more generic.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -65,6 +67,77 @@ class Completions(
     (values, result)
   end completions
 
+  private def toCompletionValues(
+      completion: Completion
+  ): List[CompletionValue.Compiler] =
+    val shouldAddSnippet = path match
+      // if the parent is an apply then we already have ()
```
tanishiking, line 74: Can you elaborate on it? I couldn't catch it.

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
```diff
@@ -52,6 +52,21 @@ class CompletionSnippetSuite extends BaseCompletionSuite {
        |""".stripMargin
   )
 
+  checkSnippet(
+    "nullary2",
```
tanishiking, line 56: I think this case is more like `nilary`, instead of `nullary`
https://stackoverflow.com/questions/7600910/difference-between-function-with-parentheses-and-without

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
```diff
@@ -275,4 +373,47 @@ class CompletionSnippetSuite extends BaseCompletionSuite {
...
+        |}
+        |object Main {
+        |  Wi@@
+        |}
+        |""".stripMargin,
+    "Widget -  example",
+    compat = Map(
+      "3" ->
+        """|Widget -  example
+           |Widget($0) - (name: String): Widget
```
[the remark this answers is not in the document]
tanishiking: I personally think showing all the possible signatures is useful 👍

tanishiking (changes requested):
Great work @tgodzik !
I left some request changes, but the main architecture looks good 👍

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -65,6 +67,77 @@ class Completions(
...
+      case (fun) :: (appl: GenericApply) :: _ if appl.fun == fun =>
+        false
+      case (_: Import) :: _ => false
+      case _ :: (_: Import) :: _ => false
+      case _ => true
+
+    completion.symbols.flatMap(
+      completionsWithSuffix(_, completion.label, shouldAddSnippet)
+    )
```
dos65, line 83: Maybe put the check on `shouldAddSnippet` here and avoid passing this param to other suffix related functions?
```scala
if shouldAddSnippet && config.isCompletionSnippetsEnabled then
     completion.symbols.flatMap(completionsWithSuffix(_, completion.label))
else
    completion.symbols.flatMap(/*create completionValue*/)
   
```

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -65,6 +67,77 @@ class Completions(
...
+    else None
+
+  private def completionsWithSuffix(
+      sym: Symbol,
+      label: String,
+      shouldAddSnippet: Boolean
+  ) =
+    // find the apply completion that would need a snippet
+    val methodSymbols = if shouldAddSnippet && sym.is(Flags.Module) then
+      val appl = sym.info.member(nme.apply)
```
dos65, line 126: It's needed to refactor a part with symbolSearch completions a bit to have apply completions for both compiler and workspace found symbols.
Module symbols might come from `enrichWithSymbolSearch` which is called in `filterInteresting` while here we do these checks only for symbols returned by compiler.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: So if the path to the current position contains an apply it means it's something like `metho@@()` instead of just `metho@@`. I will fix the explanation

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: Do you have an example in mind maybe?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: Changed! Good idea! That's actually in line with the Scala 2 behaviour.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: Changed and moved to MetalsEnrichments

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: Added some more comments

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
tgodzik: Ok cool! Changed the name

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -65,6 +67,77 @@ class Completions(
...
+  inline private def undoBacktick(label: String): String =
+    label.stripPrefix("`").stripSuffix("`")
+
+  private def findApplySymbols(denot: Denotation): List[Symbol] =
+    denot match
+      case MultiDenotation(denot1, denot2) =>
+        List(findApplySymbols(denot1), findApplySymbols(denot2)).flatten
+      case NoDenotation => Nil
+      case _ =>
+        List(denot.symbol)
```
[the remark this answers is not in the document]
tgodzik: Extracted!

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: Ok, I found an example and fixed!

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: I extracted it anyway and it's also now used for Workspace completions

tanishiking (approved):
LGTM from my side :)

dos65 (approved):
:+1:

dos65:
@tgodzik some tests are failing. looks like at least some of them requires the changes in expectations

tgodzik:
> @tgodzik some tests are failing. looks like at least some of them requires the changes in expectations

And I thought I managed to fix it all :sweat_smile: 

@tanishiking Let's merge it after the release?

tanishiking:
Yup, let's merge after release :)

tgodzik:
Ok, it seems I needed to do a workaround for Scala 3.0.2 in https://github.com/scalameta/metals/pull/3959/commits/126f5a527aabc1832fc855050c0bf706f31f7523

PR #3963, opened 2022-05-27
Adding braces

zmerr:


metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
```diff
@@ -0,0 +1,641 @@
...
+            termBlock.parent.flatMap { blockHolder =>
+              createCodeActionForPotentialBlockHolder(
+                blockHolder,
+                path,
+                document,
+                termBlock,
+                "block"
+              )
+            }
+          case termMatch: Term.Match =>
```
tgodzik, line 221: Could we drop Match and Catch for now and focus on templates and blocks?

metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
```diff
@@ -0,0 +1,641 @@
...
+
+  private def hasBraces(template: Template): Boolean = {
+    util
+      .Try(template.stats.maxBy(_.pos.end).pos.end)
+      .getOrElse(-1) != template.pos.end && util
+      .Try(template.tokens.maxBy(_.pos.end).text)
+      .getOrElse("") == "}"
+  }
+
+  private def hasBraces(termBlock: Term.Block): Boolean = {
```
tgodzik, line 242:
```suggestion
  private def hasBraces(term: Tree, stats: Seq[Tree]): Boolean = {
```
we could replace the two functions with this since they are identical.

metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
```diff
@@ -0,0 +1,641 @@
...
+      document = document,
+      codeActionSubjectTitle = codeActionSubjectTitle,
+      maybeEndMarkerPos = maybeGetEndMarkerPos(templateHolder)
+    )
+  }
+
+  private def createCodeActionForGoingBraceful(
+      path: AbsolutePath,
+      indentation: String,
+      document: String,
```
tgodzik, line 559: Seems unused

metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
```diff
@@ -0,0 +1,641 @@
...
+      ec: ExecutionContext
+  ): Future[Seq[l.CodeAction]] = Future {
+    val textDocumentIdentifier = params.getTextDocument
+    val path = textDocumentIdentifier.getUri.toAbsolutePath
+    val cursorRange = params.getRange
+    val cursorPosition = cursorRange.getStart()
+    val maybeTree =
+      if (cursorRange.getStart == cursorRange.getEnd)
+        trees
+          .findLastEnclosingAt[Tree](
```
tgodzik, line 41:
```suggestion
          .findLastEnclosingAt[Block](
```
and then invoke:
```
createCodeActionForPotentialBlockHolder(
              tree.parent,
              path,
              document,
              tree,
              "val definition"
            ).toSeq
```

tgodzik (changes requested):
Could we try and reduce the duplication? Seems we have a lot of duplicated parts of code (with small differences). Maybe also reduce the scope to make it work with Blocks and Templates for now. It will be easier to see how to add more if you work on this iteratively.

metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
zmerr: Yes, sure.

metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
zmerr: This is the common part of code between all functions which add braces. So it is refactored here. It is called in 
`createCodeActionToTakePotentialBlockHolderBraceful`
and `createCodeActionToTakeTemplateHolderBraceful`

metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
zmerr: I’m afraid that would exclude `Template`

tgodzik: We can do that next in the same way.

tgodzik: Or we can add a new method that will try and match the first out of two

tgodzik: Something like:

```
  def findLastEnclosingAt[T1 <: Tree: ClassTag, T2 <: Tree: ClassTag](
      source: AbsolutePath,
      lspPos: l.Position
  ): Option[Tree] = {

    def loop(t: Tree, pos: Position): Option[Tree] = {
      t match {
        case _: T1 | _: T2=>
          enclosedChildren(t.children, pos)
            .flatMap(loop(_, pos))
            .orElse(Some(t))
        case other =>
          enclosedChildren(other.children, pos).flatMap(loop(_, pos))
      }
    }
    get(source).flatMap { tree =>
      val pos = lspPos.toMeta(tree.pos.input)
      loop(tree, pos)
    }

  }
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
tgodzik: But it's not needed, is it? I don't see it used in this function.

Also I think you can shorter the names a bit like:
`blockBracefulAction` and `templateBracelessAction`

otherwise it's hard to read.

metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
tgodzik: it's a private function so no need for that. It's more important when designing a library API, but here others can stilll change anything and we will be able to catch that on a review.

metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
zmerr: oh, ooops!
Just noticed that we are showing code actions on not just blocks themselves but also the BlockHolder trees which are not blocks. I mean, val, var, defn, ...

metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
```diff
@@ -0,0 +1,478 @@
...
+            .toOption
+        )
+      bracePose =
+        if (
+          bracelessStartToken.text == bracelessStart && bracelessStart.length > 0
+        )
+          bracelessStartToken.pos.toLSP.getStart
+        else bracelessStartToken.pos.toLSP.getEnd
+
+    } yield createCodeActionForGoingBraceful(
```
zmerr, line 279: the function `createCodeActionForGoingBraceful` is called here.

metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
```diff
@@ -0,0 +1,478 @@
...
+      indentation: String
+  ): Option[l.CodeAction] = {
+
+    for {
+      colonPose <- templ.tokens
+        .find(token =>
+          token.text == ":" && token.pos.start >= expectedBraceStartPos
+        )
+        .map(_.pos.toLSP.getStart)
+    } yield createCodeActionForGoingBraceful(
```
zmerr, line 380: and also the `createCodeActionForGoingBraceful` is called here. If it is removed, the code of it should be duplicated in these two places.

metals/src/main/scala/scala/meta/internal/metals/codeactions/AddingBracesCodeAction.scala
zmerr: hmmm… I see. no problem.

zmerr: oh, oops! the `.stats` member of `termBlock` and `template` are used in the body of this method. But this member is unavailable on `Tree` type.

zmerr: I am going to replace the method with `isStatsHolderBraced(tree: Tree, stats: List[Tree])`

zmerr: I think the way for reducing the number of calls to `codeActionForBlockHolder` is to extract the `(blockHolder, blockEmbraceable)` pair of various blockHolders and then pass the result to a single call to `codeActionForBlockHolder`.

zmerr:
@tgodzik it seems that this PR is now in a good condition for merging :)  Can you please take a look? I fixed the tests that weren’t passing.

zmerr:
@tgodzik @dos65 @tanishiking this PR is rebased :) I would say it can also get merged, perhaps? Could you please give it a last round of reviews?

tgodzik:
> @tgodzik @dos65 @tanishiking this PR is rebased :) I would say it can also get merged, perhaps? Could you please give it a last round of reviews?

I will try and take a look later this week, but I want to concentrate on bugfixes for this week and do the release.

tgodzik:
Closing this for now as it's a bit too complex for what it does. The best workaround currently is to use selection range to select the needed block and type { which should enclose the selection in {}

PR #3965, opened 2022-05-28
feat: add function name to rewrite parens/braces code action

kpodsiad:
Sometimes it's hard to tell which function will be rewritten, I hope this will help (me).
It ain't much, but it's honest work.


kpodsiad:
BTW, it's interesting that inferred type for map `[Int]` dissapears when it's written with braces, is it a feature @tgodzik? ;)

tgodzik:
> BTW, it's interesting that inferred type for map `[Int]` dissapears when it's written with braces, is it a feature @tgodzik? ;)
> EDIT: it is visible after saving file, case closed ;)

It's because the token changes that the decoration is attached to I think.

tanishiking (approved):
LGTM 👍 good idea!

PR #3966, opened 2022-05-30
refactor: Fix testing library breakpoints and simplify metals adapter logic

tgodzik:
Working on https://github.com/scalameta/metals/pull/3962 I realised that we don't actually test breakpoints in libraries and also we never seems to test the current virtual doc approach. I refactored the code and made sure that the tests work well with the current VS Code approach while also making sure that relevant tests are still checked without virtual docs.

This also led to simplifying the logic responsible for translating URIs from Metals to the ones for DAP.


tests/unit/src/main/scala/scala/meta/internal/metals/debug/DebugFileLayout.scala
```diff
@@ -1,27 +0,0 @@
-package scala.meta.internal.metals.debug
```
tgodzik, line 1: Renamed and modified in FileBreakpoints

tests/unit/src/main/scala/scala/meta/internal/metals/debug/Debugger.scala
```diff
@@ -40,7 +40,7 @@ final class Debugger(server: RemoteServer)(implicit ec: ExecutionContext) {
     val arguments = new InitializeRequestArguments
     arguments.setAdapterID("test-adapter")
     arguments.setLinesStartAt1(true)
-    arguments.setPathFormat(InitializeRequestArgumentsPathFormat.URI)
+    arguments.setPathFormat(InitializeRequestArgumentsPathFormat.PATH)
```
tgodzik, line 43: This is actually sent in VS Code.

tests/unit/src/main/scala/scala/meta/internal/metals/debug/TestDebugger.scala
```diff
@@ -57,10 +57,15 @@ final class TestDebugger(
 
   def setBreakpoints(
       path: AbsolutePath,
-      positions: List[Position]
```
tgodzik, line 60: No need for position since only line is used anyway.

tests/unit/src/main/scala/tests/DapTestEnrichments.scala
```diff
@@ -1,27 +1,20 @@
...
 import scala.meta.io.AbsolutePath
 
 import org.eclipse.lsp4j.debug.Source
-import org.eclipse.lsp4j.debug.SourceBreakpoint
 
 object DapTestEnrichments {
   implicit class DapXtensionAbsolutePath(path: AbsolutePath) {
     def toDAP: Source = {
+      val adaptedPath =
+        if (path.isJarFileSystem) path.toURI.toString
```
tgodzik, line 11: Simulate how VS Code does things. Doesn't change anything if virtual docs are not used.

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -511,8 +505,14 @@ final class TestingServer(
     require(server.bspSession.isDefined, "Build server did not initialize")
   }
 
-  def toPath(filename: String): AbsolutePath =
+  def toPath(filename: String): AbsolutePath = {
     TestingServer.toPath(workspace, filename, virtualDocSources)
+  }
+
+  def toPathFromSymbol(symbol: String, filename: String): AbsolutePath = {
+    workspaceSymbol(symbol)
```
tgodzik, line 513: We need to first find the particular symbol in order to cache the location of the jar. Could try and simplify it further, but I don't think it's neccessary currently.

tgodzik:
This should be ok to merge, anyone wants to take a look?

dos65 (approved):
LGTM! :tada:

PR #3967, opened 2022-05-30
docs: remove status-bar as option for slowTaskProvider

ckipp01:
This is documented, but looking at the implementation there is no
`status-bar` option anymore. This just removes this from the docs to not
confuse anyone.


tanishiking (approved):
Indeed there isn't, nice catch!

Issue #3968, opened 2022-05-31
Content of global Bloop JSON is not escaped


mbovel:
### Describe the bug

Discovered with a student that had set a path containing `\` characters. This resulted in an invalid JSON file.

https://github.com/scalameta/metals/blob/9378da0ffdc7e53d50abccbb84500ce40f3b2a87/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala#L151

### Expected behavior

Should escape content of strings.

Could we use a real JSON writer for more robustness?

### Version of Metals

v0.11.5

mbovel:
cc @adpi2

dos65:
@mbovel Thanks for reporting!

We can switch this thing on ujson

PR #3957, opened 2022-05-26
docs: release note for Metals 0.11.6

tanishiking:
If it's all good, I'm planning to release 0.11.6 next Tuesday (2022-05-31) (as I don't have much time until then).

```
❯ git show "v0.11.6"                                                      
tag v0.11.6
Tagger: Rikito Taniguchi <rikiriki1238@gmail.com>
Date:   Fri Jun 3 16:01:18 2022 +0900

v0.11.6

commit e716a55ac6a3aed8561c0d8e9308ab0eff74de38 (tag: v0.11.6, metals/main, main)
Merge: 56063859e1 6ed9732806
Author: Tomasz Godzik <tgodzik@users.noreply.github.com>
Date:   Thu Jun 2 16:46:45 2022 +0200

    Merge pull request #3969 from tgodzik/fix-windows-escaping
    
    bugfix: Escape java home path on windows

```

- before merging this
  - [x] `s/2022-05-27/2022-06-03/`
- after merging this
  - [x] `git push upstream --tags` to run the release workflow
    - [ ] wait for release github action https://github.com/scalameta/metals/actions/runs/2434129281
  - [ ] merge this PR
  - [ ] Verify it is released by https://scalameta.org/metals/docs/contributors/releasing#before-official-release
  - [ ] Bump versions like https://github.com/scalameta/metals/pull/3883
  - [ ] do release on GitHub
  - [ ] Announce from `@scalameta` twitter account (I don't have access)

Issue #3542, opened 2022-01-18
Failed to connect to build server (Metals 0.11.1, M1 macOS)


Meowcolm024:
### Describe the bug

The server can successfully start yesterday but, it does not work today.
When checking the log, the Metals version used yesterday is 0.11.0, but today is 0.11.1.

It keeps `Waiting for the bsp connection to come up...`

### Expected behavior

Server should start.

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.1

### Extra context or search terms

Log:
<details>

``` txt
2022.01.18 11:31:17 WARN  no build target for: /Users/yatchunglaw/VSCodeProjects/mlscript/shared/src/main/scala/mlscript/ConstraintSolver.scala
2022.01.18 11:31:21 INFO  no build target found for /Users/yatchunglaw/VSCodeProjects/mlscript/shared/src/main/scala/mlscript/ConstraintSolver.scala. Using presentation compiler with project's scala-library version: 3.1.0
2022.01.18 11:31:33 INFO  shutting down Metals
2022.01.18 11:31:38 INFO  tracing is disabled for protocol LSP, to enable tracing of incoming and outgoing JSON messages create an empty file at /Users/yatchunglaw/VSCodeProjects/mlscript/.metals/lsp.trace.json or /Users/yatchunglaw/Library/Caches/org.scalameta.metals/lsp.trace.json
2022.01.18 11:31:38 INFO  logging to file /Users/yatchunglaw/VSCodeProjects/mlscript/.metals/metals.log
2022.01.18 11:31:38 INFO  Started: Metals version 0.11.1 in workspace '/Users/yatchunglaw/VSCodeProjects/mlscript' for client Visual Studio Code 1.63.2.
2022.01.18 11:31:38 INFO  Parse release: 17.0.1
2022.01.18 11:31:39 INFO  time: initialize in 0.95s
2022.01.18 11:31:40 INFO  Attempting to connect to the build server...
2022.01.18 11:31:39 INFO  skipping build import with status 'Installed'
1月 18, 2022 11:31:49 上午 org.eclipse.lsp4j.jsonrpc.RemoteEndpoint handleCancellation
警告: Unmatched cancel notification for request id 5
2022.01.18 11:31:50 INFO  no build target found for /Users/yatchunglaw/VSCodeProjects/mlscript/shared/src/main/scala/mlscript/TypeSimplifier.scala. Using presentation compiler with project's scala-library version: 3.1.0
[...]

PR #3969, opened 2022-05-31
bugfix: Escape java home path on windows

tgodzik:
This was reported by a student on EPFL. JSON doesn't allow for `\`

@tanishiking let's merge this before the release

Fixes https://github.com/scalameta/metals/issues/3968

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
@@ -49,6 +49,15 @@ final class BloopServers(
 
   import BloopServers._
 
+  private val bloopJsonPath: Option[AbsolutePath] =
+    getBloopFilePath(fileName = "bloop.json")
+  private val bloopLockFile: Option[AbsolutePath] =
+    getBloopFilePath(fileName = "created_by_metals.lock")
+
+  private def bloopLastModifiedTime: Long = bloopJsonPath
+    .flatMap(path => Try(path.toNIO.toFile().lastModified()).toOption)
+    .getOrElse(0L)
+
   def shutdownServer(): Boolean = {
     val dummyIn = new ByteArrayInputStream(new Array(0))
     val cli = new BloopgunCli(
@@ -78,7 +87,8 @@ final class BloopServers(
         workspace,
         client,
         languageClient,
-        () => connectToLauncher(bloopVersion, config.bloopPort),
+        () =>
+          connectToLauncher(bloopVersion, config.bloopPort, userConfiguration),
         tables.dismissedNotifications.ReconnectBsp,
         config,
         name
@@ -129,56 +139,41 @@ final class BloopServers(
   }
 
   private def writeJVMPropertiesToBloopGlobalJsonFile(
-      bloopGlobalJsonFilePath: AbsolutePath,
-      bloopCreatedByMetalsFilePath: AbsolutePath,
-      maybeBloopJvmProperties: Option[List[String]],
+      maybeBloopJvmProperties: List[String],
       maybeJavaHome: Option[String]
   ): Try[Unit] = Try {
+    if (maybeJavaHome.isDefined || maybeBloopJvmProperties.nonEmpty) {
+      val javaOptionsField =
+        if (maybeBloopJvmProperties.nonEmpty)
+          Some(
+            "javaOptions" -> ujson.Arr(
+              maybeBloopJvmProperties.map(opt => ujson.Str(opt.trim())): _*
+            )
+          )
+        else None
+      val fields: List[(String, ujson.Value)] =
+        List(
+          maybeJavaHome.map(v => "javaHome" -> ujson.Str(v.trim())),
+          javaOptionsField
+        ).flatten
+      val obj = ujson.Obj.from(fields)
+      val jvmPropertiesString = ujson.write(obj)
 
-    val javaOptionsString = maybeBloopJvmProperties
-      .map { bloopJvmProperties =>
-        s"\"javaOptions\": [${bloopJvmProperties.map(property => s"\"$property\"").mkString(", ")}]"
-      }
-      .getOrElse("")
-
-    val jvmPropertiesString = maybeJavaHome
-      .map { javaHome =>
-        if (javaOptionsString.isEmpty)
-          s"{ \"javaHome\": \"$javaHome\"}"
-        else
-          s"""|{
-              | $javaOptionsString,
-              | \"javaHome\": \"$javaHome\"
-              |}""".stripMargin
-      }
-      .getOrElse(s"{$javaOptionsString}")
-    bloopGlobalJsonFilePath.writeText(jvmPropertiesString)
-    bloopCreatedByMetalsFilePath.writeText(
-      bloopGlobalJsonFilePath.toNIO.toFile.lastModified().toString
-    )
+      bloopJsonPath.foreach(_.writeText(jvmPropertiesString))
+      bloopLockFile.foreach(_.writeText(bloopLastModifiedTime.toString()))
+    }
   }
-
-  private def getBloopGlobalJsonLastModifiedByMetalsTime(
-      bloopCreatedByMetalsFilePath: AbsolutePath
-  ): Long = Try {
-    bloopCreatedByMetalsFilePath.readText.toLong
-  }.getOrElse(0)
 
   private def processUserPreferenceForBloopJvmProperties(
       messageActionItem: MessageActionItem,
-      bloopGlobalJsonFilePath: AbsolutePath,
-      bloopCreatedByMetalsFilePath: AbsolutePath,
-      maybeBloopJvmProperties: Option[List[String]],
+      maybeBloopJvmProperties: List[String],
       maybeJavaHome: Option[String],
       reconnect: () => Future[BuildChange]
   ): Future[Unit] = {
-    messageActionItem match {
-
-      case item
+    (messageActionItem, bloopJsonPath) match {
+      case (item, _)
           if item == Messages.BloopGlobalJsonFilePremodified.applyAndRestart =>
         writeJVMPropertiesToBloopGlobalJsonFile(
-          bloopGlobalJsonFilePath,
-          bloopCreatedByMetalsFilePath,
           maybeBloopJvmProperties,
           maybeJavaHome
         ) match {
@@ -188,70 +183,65 @@ final class BloopServers(
             reconnect().ignoreValue
         }
 
-      case item
+      case (item, Some(bloopPath))
           if item == Messages.BloopGlobalJsonFilePremodified.openGlobalJsonFile =>
         val position = new Position(0, 0)
         val range = new org.eclipse.lsp4j.Range(position, position)
         val command = ClientCommands.GotoLocation.toExecuteCommandParams(
           ClientCommands.WindowLocation(
-            bloopGlobalJsonFilePath.toURI.toString,
+            bloopPath.toURI.toString,
             range
           )
         )
         Future.successful(languageClient.metalsExecuteClientCommand(command))
 
-      case item
-          if item == Messages.BloopGlobalJsonFilePremodified.useGlobalFile =>
-        Future.unit
+      case _ => Future.unit
+
     }
   }
 
   private def updateBloopGlobalJsonFileThenRestart(
-      bloopGlobalJsonFilePath: AbsolutePath,
-      bloopCreatedByMetalsFilePath: AbsolutePath,
-      maybeBloopJvmProperties: Option[List[String]],
+      maybeBloopJvmProperties: List[String],
       maybeJavaHome: Option[String],
-      bloopJsonUpdateCause: BloopJsonUpdateCause,
-      reconnect: () => Future[BuildChange]
+      reconnect: () => Future[BuildChange],
+      bloopJsonUpdateCause: BloopJsonUpdateCause
   ): Future[Unit] = {
-    writeJVMPropertiesToBloopGlobalJsonFile(
-      bloopGlobalJsonFilePath,
-      bloopCreatedByMetalsFilePath,
-      maybeBloopJvmProperties,
-      maybeJavaHome
-    ) match {
-      case Failure(exception) => Future.failed(exception)
-      case Success(_) =>
-        languageClient
-          .showMessageRequest(
-            Messages.BloopJvmPropertiesChange.params(bloopJsonUpdateCause)
-          )
-          .asScala
-          .flatMap {
-            case messageActionItem
-                if messageActionItem == Messages.BloopJvmPropertiesChange.reconnect =>
+    languageClient
+      .showMessageRequest(
+        Messages.BloopJvmPropertiesChange.params(bloopJsonUpdateCause)
+      )
+      .asScala
+      .flatMap {
+        case messageActionItem
+            if messageActionItem == Messages.BloopJvmPropertiesChange.reconnect =>
+          writeJVMPropertiesToBloopGlobalJsonFile(
+            maybeBloopJvmProperties,
+            maybeJavaHome
+          ) match {
+            case Failure(exception) => Future.failed(exception)
+            case Success(_) =>
               shutdownServer()
               reconnect().ignoreValue
-            case _ =>
-              Future.unit
           }
-    }
+        case _ =>
+          Future.unit
+      }
 
   }
 
   private def getBloopFilePath(fileName: String): Option[AbsolutePath] = {
-    Try {
+    sys.props.get("user.home").map { home =>
       AbsolutePath(
         Paths
-          .get(System.getProperty("user.home"))
+          .get(home)
           .resolve(s".bloop/$fileName")
       )
-    }.toOption
+    }
   }
 
   private def maybeLoadBloopGlobalJsonFile(
       bloopGlobalJsonFilePath: AbsolutePath
-  ): (Option[String], Option[List[String]]) = {
+  ): (Option[String], List[String]) = {
 
     val maybeLinkedHashMap =
       bloopGlobalJsonFilePath.readTextOpt.map(ujson.read(_)).flatMap(_.objOpt)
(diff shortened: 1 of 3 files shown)


tanishiking:
Looks good, but I'll test it locally at least it works on macOS just in case.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1306,20 +1306,24 @@ class MetalsLanguageServer(
...
+              bloopServers
+                .ensureDesiredVersion(
+                  userConfig.currentBloopVersion,
+                  session.version,
+                  userConfig.bloopVersion.nonEmpty,
+                  old.bloopVersion.isDefined,
+                  () => autoConnectToBuildServer
+                )
+                .flatMap { alreadyRestarted =>
+                  if (!alreadyRestarted)
```
tanishiking: [note]
If `ensureDesiredVersion` reconnect to the bloop server, we don't need to `ensureDesiredJvmSettings` because it's already restarted and jvm settings should be already desired state.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -222,19 +229,11 @@ final class BloopServers(
...
-          .asScala
-          .flatMap {
-            case messageActionItem
-                if messageActionItem == Messages.BloopJvmPropertiesChange.reconnect =>
-              shutdownServer()
-              reconnect().ignoreValue
-            case _ =>
-              Future.unit
-          }
+        scribe.info(
```
tgodzik, line 232: I decided that we don't need to ask in case the json file is managed by Bloop, since that will confuse new users. Especially that we set the javaHome at the start.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1306,20 +1306,21 @@ class MetalsLanguageServer(
...
+              bloopServers
+                .ensureDesiredVersion(
+                  userConfig.currentBloopVersion,
+                  session.version,
+                  userConfig.bloopVersion.nonEmpty,
+                  old.bloopVersion.isDefined,
+                  () => autoConnectToBuildServer
+                )
+                .flatMap { _ =>
+                  bloopServers.ensureDesiredJvmSettings(
```
tgodzik, line 1318: Let's not show both messages at the same time, anyway usually it should be just one thing.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -134,28 +134,36 @@ final class BloopServers(
       maybeBloopJvmProperties: Option[List[String]],
       maybeJavaHome: Option[String]
   ): Try[Unit] = Try {
+    if (maybeJavaHome.isDefined || maybeBloopJvmProperties.isDefined) {
+      val jvmPropertiesString =
+        (maybeBloopJvmProperties, maybeJavaHome) match {
+          case (None, Some(javaHome)) =>
+            ujson.write(
+              ujson.Js.Obj(
```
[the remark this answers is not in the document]
tgodzik: Ach, missed it!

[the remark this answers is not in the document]
tgodzik: And should be fixed.

tanishiking (approved):
I left one nitpick, otherwise it's LGTM!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -134,28 +134,36 @@ final class BloopServers(
...
+                "javaHome" -> javaHome,
+                "javaOptions" -> javaOptions.map(ujson.Str(_))
+              )
+            )
+          case (Some(javaOptions), None) =>
+            ujson.write(
+              ujson.Obj(
+                "javaOptions" -> javaOptions.map(ujson.Str(_))
+              )
+            )
```
dos65, line 158: There is a way to simlify duplication inside pattern matching
```scala
      val fields: List[(String, ujson.Value)] =
        List(
          maybeJavaHome.map(v => "javaHome" -> ujson.Str(v)),
          maybeBloopJvmProperties.map(opts =>
            "javaOptions" -> ujson.Arr(opts.map(ujson.Str(_)))
          )
        ).flatten
      val obj = ujson.Obj.from(fields)
      val propertiesString = ujson.write(obj)
```

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -134,28 +134,36 @@ final class BloopServers(
...
+          case _ => ""
```
dos65, line 159: Is it ok to leave it empty?
it looks like like previously the empty was written as an object `{}`

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: I was thinking about it, but it seemed a bit clearer this way with everything written explicitely.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: This can't really happen since we don't write anything if both are None.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -211,7 +219,6 @@ final class BloopServers(
       bloopCreatedByMetalsFilePath: AbsolutePath,
       maybeBloopJvmProperties: Option[List[String]],
       maybeJavaHome: Option[String],
-      bloopJsonUpdateCause: BloopJsonUpdateCause,
```
zmerr, line 214: why is this removed? This was used for telling the user why bloop was to be restarted, especially when it was caused by changing Metals java home.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: It seems to me that when the restart is done due to update to Metals Java home, the users need a way to know this is going to affect their Bloop java home too. Don’t they?

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: Ok sure, not a big deal.

zmerr: I’m afraid the format ujson might use for producing the json might not be exactly compatible with the form in Bloop docs. I remember that’s manual serialization was used in the code in the first place for producing the json. It’s good to double check the produced result and check also if it can be properly consumed both during reading and by Bloop.

tgodzik: Changed!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -388,7 +387,6 @@ final class BloopServers(
         bloopCreatedByMetalsFilePath,
         maybeBloopJvmProperties,
         maybeJavaHome,
-        bloopJsonUpdateCause,
```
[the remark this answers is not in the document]
tgodzik: I decided that since we are changing things in the settings it just adds additional work to user to then click to restart Bloop server. Especially, that a lot of users might not understand it fully.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: Same as above.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: We can't anyway use manual serialization (as that leads to a lot of issues) and since this a valid json it should all work correctly. I double checked it. Also, Bloop actually uses a json library for reading, so this should be ok.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: The main problem is that we will also do this at the start and we would always show it to the user even if it's a new user that doesn't know anything about the Java home etc. So I decide that it's better to hide it currently instead. Otherwise I expect a lot of questions from confused users.

dos65 (approved):

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: Please, let me pull the branch and double check if it can find the expected value.

zmerr: Sorry Tomasz, in local testing of the branch, Metals cannot detect that the value written in the Bloop json 
```
{"javaOptions": [-Xmx1G]}
```
is the same as the one assigned to Bloop Jvm Properties as `-Xmx1G` and shows the popup!
To reproduce, first manually change the file to the content shown above. then go to settings and set bloop jvm properties to `Xmx1G`
The expected behavior is to do nothing. Instead a popup dialogue is shown.

zmerr: Oh, it is written automatically with quotation marks, but when I put the quotes there:

```
{"javaOptions": ["-Xmx1G", "Xms100M”]}
```
it still shows the dialogue even if the requested value is equal to the content of the file.
Not sure why.

zmerr: I think it is unrelated to json. Now I am trying to find the root of the issue, since this was supposed to be solved in the previous PR.

tanishiking:
I'll wait the [release](https://github.com/scalameta/metals/pull/3957) until this is merged.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: Ok then we will merge this one and if you could explain the other problem in an issue or create a PR that would be awesome!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -134,28 +134,24 @@ final class BloopServers(
       maybeBloopJvmProperties: Option[List[String]],
       maybeJavaHome: Option[String]
   ): Try[Unit] = Try {
+    if (maybeJavaHome.isDefined || maybeBloopJvmProperties.isDefined) {
+      val fields: List[(String, ujson.Value)] =
+        List(
+          maybeJavaHome.map(v => "javaHome" -> ujson.Str(v.trim())),
+          maybeBloopJvmProperties.map(opts =>
+            "javaOptions" -> ujson.Arr(
+              opts.map(opt => ujson.Str(opt.trim())): _*
```
tgodzik, line 143: I just added the trim to make sure it all works even when whitespaces are added

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -352,7 +352,7 @@ object Messages {
 
   object BloopJvmPropertiesChange {
     def reconnect: MessageActionItem =
-      new MessageActionItem("Restart Bloop")
+      new MessageActionItem("Apply and restart Bloop")
```
tgodzik, line 355: I reverted back to asking the user because otherwise when multiple metals servers were running this would result in race conditions

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -403,8 +371,21 @@ final class BloopServers(
 
   private def connectToLauncher(
       bloopVersion: String,
-      bloopPort: Option[Int]
+      bloopPort: Option[Int],
+      userConfiguration: UserConfiguration
   ): Future[SocketConnection] = {
+    // we should set up Java before running Bloop in order to not restart it
+    bloopJsonPath match {
```
tgodzik, line 378: This is to make sure that we don't restart Bloop after setting bloop.json for the first time.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -188,70 +181,65 @@ final class BloopServers(
...
-                if messageActionItem == Messages.BloopJvmPropertiesChange.reconnect =>
+    languageClient
+      .showMessageRequest(
+        Messages.BloopJvmPropertiesChange.params(bloopJsonUpdateCause)
+      )
+      .asScala
+      .flatMap {
+        case messageActionItem
+            if messageActionItem == Messages.BloopJvmPropertiesChange.reconnect =>
+          writeJVMPropertiesToBloopGlobalJsonFile(
```
tgodzik, line 215: Changed the order here to first ask then update to avoid race conditions which might be what @zmerr was seeing

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: I think I found the issue, mentioned it in the other comment.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -304,31 +287,26 @@ final class BloopServers(
...
+              .exists(requested =>
+                requested != maybeBloopGlobalJsonJvmProperties
+              )
           ) Some(BloopJsonUpdateCause.JVM_OPTS)
           else if (maybeRequestedMetalsJavaHome != maybeBloopGlobalJsonJavaHome)
             Some(BloopJsonUpdateCause.JAVA_HOME)
           else None
-        maybeBloopJvmProperties = maybeRequestedBloopJvmProperties.orElse(
+        maybeBloopJvmProperties = maybeRequestedBloopJvmProperties.getOrElse(
           maybeBloopGlobalJsonJvmProperties
```
zmerr, line 306: so `maybeBloopGlobalJsonJvmProperties` is always defined?

tgodzik: Yes, if it's an empty list it results in no options set for Bloop

zmerr: So it covers also the scenario when we want to remove the properties, right?

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
zmerr: this should only happen when the javaHome that Bloop was using was already the same as that of Metals, or as the one we set for Metals in the setting.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -267,16 +255,9 @@ final class BloopServers(
       javaOptionsValue <- linkedHashMap.get("javaOptions")
       javaOptionsValueArray <- javaOptionsValue.arrOpt
     } yield javaOptionsValueArray.flatMap(_.strOpt).toList
-    (maybeJavaHome, maybeJavaOptions)
+    (maybeJavaHome, maybeJavaOptions.getOrElse(Nil))
```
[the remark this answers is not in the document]
tgodzik: IT might be a None if it doesn't exist in the json file, but we treat it the same as an empty list.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: Yes, that works

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: This will always happen at the start, when no bloop.json was set up previously.

tanishiking: So now, if

- there's no `$HOME/.bloop/bloop.json` and
- Whether there's `metals.bloopJvmProperties` defined or not

Metals will always write the global bloop setting (when connect to bloop launcher) file to `$HOME/.bloop/bloop.json` where `javaHome` will be inferred java home location by `locate-java-home`.

---

I'm a bit concerned about the following scenario:
- Metals write the global bloop file to `$HOME/.bloop/bloop.json` where `javaHome` is javaHome of currently using java (inferred by `locate-java-home`)
- User upgrade java version (and now `locate-java-home` locate the new java's javahome.
- Metals won't update the global bloop setting file, because the user setting of Metals didn't change (and `workspace/didChangeConfiguration` doesn't fire).
  - Or the user declines Metals to write bloop global file.
- While Metals uses new java, bloop still has old javaHome in the global bloop setting file, and bloop will continue using old Java.

It might be better to
- (1) every time Metals server starts, write JVM properties to bloop global json file even if there exists a file (but I'm not sure how should we handle the existing file :thinking:)
- (2) don't write java home if a user didn't specify java home explicitly via user settings.

tanishiking (changes requested):
I commented about one concern

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tanishiking: @tgodzik 
Oh does workspace/didChangeConfigration will  be fired (at least by VSCode) on server starts, even if the setting has changed before Metals server starts?
In that case, it's all good :+1:

 
> The only issue will be on other editors if Metals was started and the setting was later updated manually

I'm not sure, but other clients also should fire. Anyway, it would be better to implement (1) as you mentioned, if it's handy. If it's complexed, we can go as it is, at least for now

tgodzik: > Oh does workspace/didChangeConfigration will be fired (at least by VSCode) on server starts, even if the setting has changed before Metals server starts?
> In that case, it's all good

Yeah, it should be sent correctly.

> I'm not sure, but other clients also should fire. Anyway, it would be better to implement (1) as you mentioned, if it's handy. If it's complexed, we can go as it is, at least for now

So another thing is that if set javaHome before starting Bloop, we would have to restart it in case it's running, which is a bit more complex and would be outside the user's influence. For now I would leave at is and maybe improve the behaviour later on?

tgodzik: Coming to think of it you currently need to connect to Bloop to restart it, so we would need to invoke Bloop CLI for that (which could be a part of a different issue though) https://github.com/scalameta/metals/issues/3542

tanishiking (approved):

tanishiking:
Well, I realized that when I manually set `javaHome` of the workspace setting (of VSCode), Metals also write the workspace javaHome setting to `$HOME/.bloop/bloop.json`.
However, workspace settings should be limited to the workspace, and it shouldn't change the global settings I guess. Is this an expected behavior?

tanishiking:
Anyway @tgodzik great work :tada:

PR #3971, opened 2022-06-01
Add ConvertToNamedArguments code action

camgraff:
@ maintainers thanks for all the work you do! This is my first PR, I appreciate any feedback :slightly_smiling_face: 

Fixes https://github.com/scalameta/metals-feature-requests/issues/229

Adds a code action which converts all positional arguments in the enclosing `apply` method to named args.
metals-convert-to-named-args

TODOS/open questions:
[ ] - Scala 3 support
[ ] - tests. I started by trying to extend `BasicCodeActionLspSuite`, but ran into this error with the presentation compiler. I'd appreciate some direction here :)
```
Caused by: java.lang.AbstractMethodError: Missing implementation of resolved method 'abstract java.util.concurrent.CompletableFuture convertToNamedArguments(scala.meta.pc.OffsetParams, int)' of abstract class scala.meta.pc.PresentationCompiler.
```
[ ] - Currently, if there are no args in the enclosing `apply` the code action is not available which means if my cursor were on `params.text()` in the gif above, I'd see no code action. Maybe we'd want to consider the next enclosing `apply` (ie `addCompilationUnit`) as a candidate?
[ ] - There's probably some edge cases that I haven't considered. One is when the `apply` is inside a block which can't have named args.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -1,5 +1,6 @@
 package scala.meta.internal.metals
 
+import java.{util => ju}
 import javax.annotation.Nullable
 
 import scala.meta.internal.metals.newScalaFile.NewFileTypes
@@ -8,7 +9,6 @@ import ch.epfl.scala.{bsp4j => b}
 import org.eclipse.lsp4j.Location
 import org.eclipse.lsp4j.TextDocumentIdentifier
 import org.eclipse.lsp4j.TextDocumentPositionParams
-import java.{util => ju}
 
 /**
  * LSP commands supported by the Metals language server.
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -478,7 +478,7 @@ object ServerCommands {
       """|Whenever a user chooses code action to convert to named arguments, this command is later ran to 
          |determine the parameter names of all unnamed arguments and insert names at the correct locations.
          |""".stripMargin,
-      """|This command should be sent in with the LSP [`TextDocumentPositionParams`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams)
+      """|Object with `position` of type [TextDocumentPositionParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams) and `numUnnamedArgs` (int)
          |""".stripMargin
     )
 
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -1,6 +1,5 @@
 package scala.meta.internal.metals
 
-import java.{util => ju}
 import javax.annotation.Nullable
 
 import scala.meta.internal.metals.newScalaFile.NewFileTypes
@@ -468,7 +467,7 @@ object ServerCommands {
        |""".stripMargin
   )
 
-  final case class ConvertToNamedArgsRequest(position: TextDocumentPositionParams, argIndices: ju.List[Integer])
+  final case class ConvertToNamedArgsRequest(position: TextDocumentPositionParams, numUnnamedArgs: Int)
   val ConvertToNamedArguments = new ParametrizedCommand[ConvertToNamedArgsRequest](
     "convert-to-named-arguments",
     "Convert positional arguments to named ones",
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -467,16 +467,20 @@ object ServerCommands {
        |""".stripMargin
   )
 
-  final case class ConvertToNamedArgsRequest(position: TextDocumentPositionParams, numUnnamedArgs: Int)
-  val ConvertToNamedArguments = new ParametrizedCommand[ConvertToNamedArgsRequest](
-    "convert-to-named-arguments",
-    "Convert positional arguments to named ones",
-    """|Whenever a user chooses code action to convert to named arguments, this command is later ran to 
-       |determine the parameter names of all unnamed arguments and insert names at the correct locations.
-       |""".stripMargin,
-    """|This command should be sent in with the LSP [`TextDocumentPositionParams`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams)
-       |""".stripMargin
+  final case class ConvertToNamedArgsRequest(
+      position: TextDocumentPositionParams,
+      numUnnamedArgs: Int
   )
+  val ConvertToNamedArguments =
+    new ParametrizedCommand[ConvertToNamedArgsRequest](
+      "convert-to-named-arguments",
+      "Convert positional arguments to named ones",
+      """|Whenever a user chooses code action to convert to named arguments, this command is later ran to 
+         |determine the parameter names of all unnamed arguments and insert names at the correct locations.
+         |""".stripMargin,
+      """|This command should be sent in with the LSP [`TextDocumentPositionParams`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams)
+         |""".stripMargin
+    )
 
   val GotoLog = new Command(
     "goto-log",
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -8,6 +8,7 @@ import ch.epfl.scala.{bsp4j => b}
 import org.eclipse.lsp4j.Location
 import org.eclipse.lsp4j.TextDocumentIdentifier
 import org.eclipse.lsp4j.TextDocumentPositionParams
+import java.{util => ju}
 
 /**
  * LSP commands supported by the Metals language server.
@@ -462,6 +463,17 @@ object ServerCommands {
     "Insert inferred type of a value",
     """|Whenever a user chooses code action to insert the inferred type this command is later ran to 
        |calculate the type and insert it in the correct location.
+       |""".stripMargin,
+    """|This command should be sent in with the LSP [`TextDocumentPositionParams`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams)
+       |""".stripMargin
+  )
+
+  final case class ConvertToNamedArgsRequest(position: TextDocumentPositionParams, argIndices: ju.List[Integer])
+  val ConvertToNamedArguments = new ParametrizedCommand[ConvertToNamedArgsRequest](
+    "convert-to-named-arguments",
+    "Convert positional arguments to named ones",
+    """|Whenever a user chooses code action to convert to named arguments, this command is later ran to 
+       |determine the parameter names of all unnamed arguments and insert names at the correct locations.
        |""".stripMargin,
     """|This command should be sent in with the LSP [`TextDocumentPositionParams`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams)
        |""".stripMargin
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -475,10 +475,10 @@ object ServerCommands {
     new ParametrizedCommand[ConvertToNamedArgsRequest](
       "convert-to-named-arguments",
       "Convert positional arguments to named ones",
-      """|Whenever a user chooses code action to convert to named arguments, this command is later ran to 
+      """|Whenever a user chooses code action to convert to named arguments, this command is later run to 
          |determine the parameter names of all unnamed arguments and insert names at the correct locations.
          |""".stripMargin,
-      """|Object with `position` of type [TextDocumentPositionParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams) and `numUnnamedArgs` (int)
+      """|Object with [TextDocumentPositionParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams) of the target Apply and `numUnnamedArgs` (int)
          |""".stripMargin
     )
 
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -8,6 +8,7 @@ import ch.epfl.scala.{bsp4j => b}
 import org.eclipse.lsp4j.Location
 import org.eclipse.lsp4j.TextDocumentIdentifier
 import org.eclipse.lsp4j.TextDocumentPositionParams
+import java.{util => ju}
 
 /**
  * LSP commands supported by the Metals language server.
@@ -469,7 +470,7 @@ object ServerCommands {
 
   final case class ConvertToNamedArgsRequest(
       position: TextDocumentPositionParams,
-      numUnnamedArgs: Int
+      argIndices: ju.List[Integer]
   )
   val ConvertToNamedArguments =
     new ParametrizedCommand[ConvertToNamedArgsRequest](
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
@@ -29,13 +29,10 @@ class ConvertToNamedArguments(trees: Trees)
       .findLastEnclosingAt[Term.Apply](path, range.getStart())
       .map { apply =>
         {
-          // TODO: Skip block args
           // TODO: Go to parent apply, if current has no candidates
-          pprint.log(apply)
-          val argIndices = apply.args.zipWithIndex.filterNot { case (arg, _) => arg.isInstanceOf[Term.Assign]}.map(_._2)
-          pprint.log(argIndices)
-          //val numUnnamedArgs =
-            //apply.args.takeWhile(!_.isInstanceOf[Term.Assign]).length
+          val argIndices = apply.args.zipWithIndex.filterNot { case (arg, _) => 
+            arg.isInstanceOf[Term.Assign] || arg.isInstanceOf[Term.Block]
+          }.map(_._2)
           if (argIndices.isEmpty) Future.successful(Nil)
           else {
             val codeAction = new l.CodeAction(title)
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
@@ -1,0 +1,55 @@
+package scala.meta.internal.metals.codeactions
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+import scala.meta.Term
+import scala.meta.internal.metals.CodeAction
+import scala.meta.internal.metals.Compilers
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.parsing.Trees
+import scala.meta.pc.CancelToken
+
+import org.eclipse.{lsp4j => l}
+
+class ConvertToNamedArguments(trees: Trees, compilers: Compilers)
+    extends CodeAction {
+  override val kind: String = l.CodeActionKind.RefactorRewrite
+
+  override def contribute(params: l.CodeActionParams, token: CancelToken)(
+      implicit ec: ExecutionContext
+  ): Future[Seq[l.CodeAction]] = {
+
+    val path = params.getTextDocument().getUri().toAbsolutePath
+    val range = params.getRange()
+
+    trees
+      .findLastEnclosingAt[Term.Apply](path, range.getStart())
+      .map { apply =>
+        {
+          val numUnnamedArgs =
+            apply.args.takeWhile(!_.isInstanceOf[Term.Assign]).length
+          if (numUnnamedArgs == 0) Future.successful(Nil)
+          else {
+            val textDocumentPositionParams = new l.TextDocumentPositionParams(
+              params.getTextDocument(),
+              new l.Position(apply.pos.endLine, apply.pos.endColumn)
+            )
+            val edits = compilers.convertToNamedArguments(
+              textDocumentPositionParams,
+              numUnnamedArgs,
+              token
+            )
+            edits.map(e => {
+              val codeAction = new l.CodeAction("Convert to named arguments")
+              codeAction.setEdit(
+                new l.WorkspaceEdit(Map(path.toURI.toString -> e).asJava)
+              )
+              Seq(codeAction)
+            })
+          }
+        }
+      }
+      .getOrElse(Future.successful(Nil))
+  }
+}
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
@@ -20,14 +20,14 @@ class ConvertToNamedArguments(trees: Trees) extends CodeAction {
 
   def firstApplyWithUnnamedArgs(
       term: Option[Tree]
-  ): Option[ApplyTermWithNumArgs] = {
+  ): Option[ApplyTermWithArgIndices] = {
     term match {
       case Some(apply: Term.Apply) =>
-        val numUnnamedArgs = apply.args.takeWhile { arg =>
-          !arg.isInstanceOf[Term.Assign] && !arg.isInstanceOf[Term.Block]
-        }.size
-        if (numUnnamedArgs == 0) firstApplyWithUnnamedArgs(apply.parent)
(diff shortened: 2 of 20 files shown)


tanishiking:
@camgraff This is really great!!! Thank you very much for working on this.

###  > tests. I started by trying to extend BasicCodeActionLspSuite, but ran into this error with the presentation compiler. I'd appreciate some direction here :)

I think this is because there's no implementation in [scala3/....../ScalaPresentationCompiler.scala](https://github.com/scalameta/metals/blob/main/mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala), you have to add something like this

```scala
  override def convertToNamedArguments(
      params: OffsetParams,
      numUnnamedArgs: Int
  ): CompletableFuture[ju.List[TextEdit]] = ???
```

---

### > Scala3 Support
- For Scala3 support, my WIP implementation of [show-parameter-name decoration PR](https://github.com/scalameta/metals/pull/3913) may help :)
  - BTW, I'm planning to implement this feature based on [Inlay Hint](https://github.com/microsoft/language-server-protocol/issues/956) that enables us to insert parameter names to the code easily. However, this CodeAction is still pretty useful because
    - This CodeAction can insert names for **all** arguments, while we have to insert one by one with `inlay hint`.
    - Inlay hint might not be available for some editors, in that case, we have to provide this kind of features as CodeAction anyway.
 

---

### Other improvement
Also, there's something to improve:

`ConvertToNamedArguments#contribute` will run every time users open / edit a file. Every time users open / edit a file, Metals calculate the code actions for all the `Apply`s in the file. This may slow down the overall performance of the Metals server.

We should separate showing a code action tooltip and constructing the TextEdit part. For example [InsertInferredType code action](https://github.com/scalameta/metals/blob/d0cf15ced536460bca4d8df5e16042a8595fd4d5/metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala#L23-L42) do the thing:

- `InsertInferredType#contribute` construct codeActions for a file, but it doesn't contain `TextEdit`, it has `Command` instead.
- When users run the codeAction, Metals will receive the command and construct the `TextEdit`
  - [Here](https://github.com/scalameta/metals/blob/d0cf15ced536460bca4d8df5e16042a8595fd4d5/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1956-L1967) is the endpoint for the command.


--- 

### > There's probably some edge cases that I haven't considered. One is when the apply is inside a block which can't have named args.
I think you don't have to worry too much about these edge cases :) If users add parameter names and compile failed, users can immediately notice it doesn't work and undo the code action.

There might be other edge cases, but we can improve this feature iteratively :+1:

---

### > Maybe we'd want to consider the next enclosing apply (ie, addCompilationUnit) as a candidate?

Sounds good! However, I'm concerned that it may introduce complexities.

In the scenario of this gif, it looks useful, but if the no-parameter method call is surrounded by another method call far away from the inner method call, falling back to the surrounding method call might be confusing for users.

I would go without this feature (next enclosing apply), at least in this PR, see the reactions from users. If many people agree it would help, consider falling back to the next enclosing apply.

mtags/src/main/scala-2/scala/meta/internal/pc/ConvertToNamedArgumentsProvider.scala
```diff
@@ -0,0 +1,33 @@
...
+) {
+  import compiler._
+  def convertToNamedArguments: List[l.TextEdit] = {
+    val unit = addCompilationUnit(
+      code = params.text(),
+      filename = params.uri().toString(),
+      cursor = None
+    )
+
+    val typedTree = typedTreeAt(unit.position(params.offset))
```
tanishiking, line 20: > Maybe we'd want to consider the next enclosing apply (ie, addCompilationUnit) as a candidate?

You can access `lastVisitedParentTrees` to see the enclosing trees (make sure you access `lastVisitedParentTrees` after calling `typedTreeAt(pos)` or `locateTree(pos)` (as long as I remember)).
https://github.com/scalameta/metals/blob/d0cf15ced536460bca4d8df5e16042a8595fd4d5/mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala#L92-L97

tanishiking:
Also, I think it's ok to work on Scala3 support in another PR if you want :)

camgraff:
Thanks @tanishiking for the pointers! I've made the code action delegate to a `ServerCommand` to return the edits. I've also excluded `Term.Block` args from the action and added logic to consider the next enclosing `Term.Apply` if the current one has no unnamed args.

camgraff:
Looks like many of the LSP code action tests are failing due to this action now being returned by the server. Is the best way to fix this just going through and updating [`expectedActions`](https://github.com/camgraff/metals/blob/8883d35357dddf4e0319314856dd57bac2f49b11/tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala#L87) for each failing test?

tanishiking:
@camgraff Thank you for working on this! I'll take a look today

tanishiking:
@camgraff 
> Is the best way to fix this just going through and updating [expectedActions](https://github.com/camgraff/metals/blob/8883d35357dddf4e0319314856dd57bac2f49b11/tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala#L87) for each failing test?

Good point!

### option1
Yes, basically we can fix those test failures by updating `expectedActions` as follows. And, as you may notice `check` method has `selectedActionIndex` parameter to select a code action to apply from the received list of code actions 👍 

However, I personally don't think this is not the best way to fix failing tests (even though we'd been doing that).
In this way, we end up testing we receive `convert to named arg code action` from the server **in ExtractValueLspSuite**, which should focus only on extract value code action.
I believe tests should assert only things they're really interested in (in this case, `ExtractValueLspSuite` should exclusively verify server returns `extract value code action`) whether the server returns other actions or not).

```diff
diff --git a/tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala b/tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala
index fc5373cb2e..0868d10afd 100644
--- a/tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala
@@ -1,6 +1,7 @@
 package tests.codeactions
 
 import scala.meta.internal.metals.codeactions.ExtractValueCodeAction
+import scala.meta.internal.metals.codeactions.ConvertToNamedArguments
 
 class ExtractValueLspSuite
     extends BaseCodeActionLspSuite("extractValueRewrite") {
@@ -15,7 +16,8 @@ class ExtractValueLspSuite
        |
        |}
        |""".stripMargin,
-    ExtractValueCodeAction.title,
+    s"""|${ExtractValueCodeAction.title}
+        |${ConvertToNamedArguments.title}""".stripMargin,
     """|object Main {
        |  def method2(i: Int) = ???
        |  def method1(s: String): Unit = {
```

---

### option2
Instead of updating `expectedActions`, I prefer to filter code actions so that tests can focus on things they're interested in.
Like this:

```diff
diff --git a/tests/unit/src/main/scala/tests/TestingServer.scala b/tests/unit/src/main/scala/tests/TestingServer.scala
index 41c22befc1..7553009606 100644
--- a/tests/unit/src/main/scala/tests/TestingServer.scala
+++ b/tests/unit/src/main/scala/tests/TestingServer.scala
@@ -1131,10 +1131,17 @@ final class TestingServer(
       query: String,
       expected: String,
       kind: List[String],
-      root: AbsolutePath = workspace
+      root: AbsolutePath = workspace,
+      filterAction: l.CodeAction => Boolean = _ => true
   )(implicit loc: munit.Location): Future[List[l.CodeAction]] =
     for {
-      (codeActions, codeActionString) <- codeAction(filename, query, root, kind)
+      (codeActions, codeActionString) <- codeAction(
+        filename,
+        query,
+        root,
+        kind,
+        filterAction
+      )
     } yield {
       Assertions.assertNoDiff(codeActionString, expected)
       codeActions
@@ -1175,7 +1182,8 @@ final class TestingServer(
       filename: String,
       query: String,
       root: AbsolutePath,
-      kind: List[String]
+      kind: List[String],
+      filterAction: l.CodeAction => Boolean = _ => true
   ): Future[(List[l.CodeAction], String)] =
     for {
       (_, params) <- codeActionParams(
@@ -1188,10 +1196,13 @@ final class TestingServer(
         )
       )
       codeActions <- server.codeAction(params).asScala
-    } yield (
-      codeActions.asScala.toList,
-      codeActions.map(_.getTitle()).asScala.mkString("\n")
-    )
+    } yield {
+      val actions = codeActions.asScala.filter(filterAction)
+      (
+        actions.toList,
+        actions.map(_.getTitle()).mkString("\n")
+      )
+    }
 
   def assertHighlight(
       filename: String,
diff --git a/tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala b/tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
index ccd3c9c8e7..522d3b30a7 100644
--- a/tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
+++ b/tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
@@ -7,6 +7,7 @@ import scala.meta.internal.metals.{BuildInfo => V}
 
 import munit.Location
 import munit.TestOptions
+import org.eclipse.{lsp4j => l}
 import tests.BaseLspSuite
 
 abstract class BaseCodeActionLspSuite(suiteName: String)
@@ -97,7 +98,8 @@ abstract class BaseCodeActionLspSuite(suiteName: String)
       extraOperations: => Unit = (),
       fileName: String = "A.scala",
       changeFile: String => String = identity,
-      expectError: Boolean = false
+      expectError: Boolean = false,
+      filterAction: l.CodeAction => Boolean = _ => true
   )(implicit loc: Location): Unit = {
     val scalacOptionsJson =
       if (scalacOptions.nonEmpty)
@@ -133,7 +135,8 @@ abstract class BaseCodeActionLspSuite(suiteName: String)
               path,
               changeFile(input),
               expectedActions,
-              kind
+              kind,
+              filterAction = filterAction
             )
             .recover {
               case _: Throwable if expectError => Nil
diff --git a/tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala b/tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala
index fc5373cb2e..e3942ee4f1 100644
--- a/tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala
@@ -24,7 +24,8 @@ class ExtractValueLspSuite
        |  }
        |
        |}
-       |""".stripMargin
+       |""".stripMargin,
+    filterAction = action => action.getTitle() == ExtractValueCodeAction.title
   )
 
   check(
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
```diff
@@ -0,0 +1,73 @@
...
+  import ConvertToNamedArguments._
+  override val kind: String = l.CodeActionKind.RefactorRewrite
+
+  def firstApplyWithUnnamedArgs(
+      term: Option[Tree]
+  ): Option[ApplyTermWithNumArgs] = {
+    term match {
+      case Some(apply: Term.Apply) =>
+        val numUnnamedArgs = apply.args.takeWhile { arg =>
+          !arg.isInstanceOf[Term.Assign] && !arg.isInstanceOf[Term.Block]
```
tanishiking, line 27: Why do we need to exclude `Term.Block` here?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
```diff
@@ -0,0 +1,73 @@
...
+
+    val maybeApply = for {
+      term <- trees.findLastEnclosingAt[Term.Apply](path, range.getStart())
+      apply <- firstApplyWithUnnamedArgs(Some(term))
+    } yield apply
+
+    maybeApply
+      .map { apply =>
+        {
+          val codeAction = new l.CodeAction(title)
```
tanishiking, line 51: It would be useful to add function name to the code action's title like https://github.com/scalameta/metals/pull/3965

metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
```diff
@@ -0,0 +1,73 @@
...
+          )
+          Future.successful(Seq(codeAction))
+        }
+      }
+      .getOrElse(Future.successful(Nil))
+  }
+}
+
+object ConvertToNamedArguments {
+  case class ApplyTermWithNumArgs(term: Term.Apply, numUnnamedArgs: Int)
```
tanishiking, line 71: [nit]
```suggestion
  case class ApplyTermWithNumArgs(app: Term.Apply, numUnnamedArgs: Int)
```

?

mtags/src/main/scala-2/scala/meta/internal/pc/ConvertToNamedArgumentsProvider.scala
tanishiking: Now, `ConvertToNamedArgumentsProvider` receive the exact position of the target Apply, and no need to traverse trees

mtags/src/main/scala-2/scala/meta/internal/pc/ConvertToNamedArgumentsProvider.scala
```diff
@@ -0,0 +1,37 @@
...
+    typedTree match {
+      case Apply(fun, args) =>
+        args
+          .take(numUnnamedArgs)
```
tanishiking, line 25: What if we have a named argument in the middle like

```scala
def foo(x: Int, y: Int, z: Int)
foo(1, y = 2, 3)
```

?

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -455,6 +455,21 @@ object ServerCommands {
...
 
+  final case class ConvertToNamedArgsRequest(
+      position: TextDocumentPositionParams,
+      numUnnamedArgs: Int
+  )
+  val ConvertToNamedArguments =
+    new ParametrizedCommand[ConvertToNamedArgsRequest](
+      "convert-to-named-arguments",
+      "Convert positional arguments to named ones",
+      """|Whenever a user chooses code action to convert to named arguments, this command is later ran to
```
tanishiking, line 466:
```suggestion
      """|Whenever a user chooses code action to convert to named arguments, this command is later run to 
```

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -455,6 +455,21 @@ object ServerCommands {
...
+      numUnnamedArgs: Int
+  )
+  val ConvertToNamedArguments =
+    new ParametrizedCommand[ConvertToNamedArgsRequest](
+      "convert-to-named-arguments",
+      "Convert positional arguments to named ones",
+      """|Whenever a user chooses code action to convert to named arguments, this command is later ran to 
+         |determine the parameter names of all unnamed arguments and insert names at the correct locations.
+         |""".stripMargin,
+      """|Object with `position` of type [TextDocumentPositionParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams) and `numUnnamedArgs` (int)
```
tanishiking, line 469:
```suggestion
      """|Object with [TextDocumentPositionParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentPositionParams) of the target Apply and `numUnnamedArgs` (int)
```

?

tests/unit/src/test/scala/tests/codeactions/ConvertToNamedArgumentsSuite.scala
```diff
@@ -0,0 +1,55 @@
+package tests.codeactions
+
+import scala.meta.internal.metals.codeactions.ConvertToNamedArguments
+import scala.meta.internal.metals.codeactions.ExtractValueCodeAction
+
+class ConvertToNamedArgumentsSuite
+    extends BaseCodeActionLspSuite("convertToNamedArguments") {
+
+  check(
+    "convert-to-named-args-basic",
```
tanishiking, line 10:
```suggestion
    "basic",
```

Don't need `convert-to-named-args` prefix, as it's obvious from the test suite name.
(and we can run only this test by `testOnly tests.codeactions.ConvertToNamedArgumentSuite -- *basic*`)

metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
tanishiking: Ah, nevermind, `convert-to-named-args-block` tells why 👍

tests/unit/src/test/scala/tests/codeactions/ConvertToNamedArgumentsSuite.scala
```diff
@@ -0,0 +1,55 @@
...
+       |}""".stripMargin
+  )
+
+  checkActionMissing(
+    "convert-to-named-args-block",
+    """|object Something {
+       |  def f(x: Seq[Int]) = x.map <<{>> _.toLong }
+       |}""".stripMargin,
+    ConvertToNamedArguments.title
+  )
```
tanishiking, line 38: Can't we use `checkNoAction` instead? (maybe we need `filterAction` thing?

tests/unit/src/test/scala/tests/codeactions/ConvertToNamedArgumentsSuite.scala
```diff
@@ -0,0 +1,55 @@
...
+       |}""".stripMargin,
+    s"""|${ConvertToNamedArguments.title}
+        |""".stripMargin,
+    """|object Something {
+       |  case class Foo(param1: Int, param2: Int, param3: Int)
+       |  Foo(param1 = 1, param2 = 2, param3 = 3)
+       |}""".stripMargin,
+    selectedActionIndex = 0
+  )
+
```
tanishiking, line 23: We need more test cases for example:

- `def` case (in addition to the case class's apply method)
- `Foo<<(>>1, param2 = 2, 3)` and `Foo<<(>>param1 = 1, 2, 3)`
- if there're multiple paramss like `foo(x: Int, y: Int)(z: Int)`

tanishiking: In addition, it would be awesome if we have a test using external libraries

you can create a test using scala standard libraries by 

```scala
class ConvertToNamedArgumentsSuite extends BaseCodeActionSuite {
   override def requiresScalaLibrarySources: Boolean = true
```

in `tests.pc`, refer 
- https://github.com/scalameta/metals/blob/0f6851d77d673368382019be6db9842b11f8b643/tests/cross/src/test/scala/tests/pc/AutoImplementAbstractMembersSuite.scala 
- https://github.com/scalameta/metals/blob/0f6851d77d673368382019be6db9842b11f8b643/tests/cross/src/test/scala/tests/pc/InsertInferredTypeSuite.scala

tanishiking (changes requested):
I left several comments on the test, and some nitpick comments, but overall it looks good to me!

tests/unit/src/test/scala/tests/codeactions/ConvertToNamedArgumentsSuite.scala
camgraff: yep!

mtags/src/main/scala-2/scala/meta/internal/pc/ConvertToNamedArgumentsProvider.scala
camgraff: I guess we can handle this case even though it's not valid syntax, but it's not clear what the behavior should be if the existing named arg is not in the same order as the parameter declaration e.g.
```scala
def foo(x: Int, y: Int, z: Int)
foo(1, z = 2, 3)
```

tests/unit/src/test/scala/tests/codeactions/ConvertToNamedArgumentsSuite.scala
camgraff: Thanks! I added more tests and a [new suite](https://github.com/camgraff/metals/blob/feat/convert_to_named_args/tests/cross/src/test/scala/tests/pc/ConvertToNamedArgumentsSuite.scala) with a single test for the scala standard lib. Should I add more tests here as well? It seems like any additional tests here would be covering the same thing as the `LspSuite` tests.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  override def contribute(params: l.CodeActionParams, token: CancelToken)(
+      implicit ec: ExecutionContext
+  ): Future[Seq[l.CodeAction]] = {
+
+    val path = params.getTextDocument().getUri().toAbsolutePath
+    val range = params.getRange()
+
+    val maybeApply = for {
+      term <- trees.findLastEnclosingAt[Term.Apply](path, range.getStart())
```
tgodzik, line 47:
```suggestion
      term <- trees.findLastEnclosingAt[Term.Apply](path, range.getStart())
      if !term.fun.encloses(range)
```
So that we don't get the code action when at for example `F<<>>uture.succesfull(1)`

tgodzik (changes requested):
This looks great!  There are few tests failing, but I think it's fine to add the code action in most of them. I wouldn't filter it out everywhere (maybe just in case of `noAction`), since there might be situations that we show the code action in too many places

tests/unit/src/test/scala/tests/codeactions/ConvertToNamedArgumentsSuite.scala
tanishiking: Looks good!
You don't need to add the same test to `tests.pc.ConvertToNamedArgumentsSuite` :+1:

---

The difference between `tests.pc.ConvertToNamedArgumentsSuite` and `LspSuite` is

- While `LspSuite` checks if the action is available or not, `tests.pc....Suite` doesn't care if the action is available or not, PC test just invokes the code action anyway and sees the result.
  - So, basically PC suites are more like unit tests, and `LspSuite`s are integration tests
  - If you don't need to check the action's availability, and what you care is the behavior of the action, it should go to PC suites (as it's lighter weight)

mtags/src/main/scala-2/scala/meta/internal/pc/ConvertToNamedArgumentsProvider.scala
```diff
@@ -0,0 +1,36 @@
...
+      cursor = None
+    )
+
+    val typedTree = typedTreeAt(unit.position(params.offset))
+    typedTree match {
+      case Apply(fun, args) =>
+        args.zipWithIndex
+          .zip(fun.tpe.params)
+          .collect {
+            case ((arg, index), param) if argIndices.contains(index) => {
```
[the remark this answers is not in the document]
camgraff: I had tried this originally, but it seems like the compiler doesn't pick up on `NamedArg`. e.g. when I trigger the code action on 
```scala
addCompilationUnit(
      params.text(),
      filename = params.uri().toString(),
      cursor  = None
    )
```
and add some logging, it shows the args as:
```
ConvertToNamedArgumentsProvider.scala:25 a: Apply(
  fun = Select(
    qualifier = Select(qualifier = This(qual = ConvertToNamedArgumentsProvider), name = params),
    name = text
  ),
  args = List()
)
ConvertToNamedArgumentsProvider.scala:26 a.isInstanceOf[NamedArg]: false
ConvertToNamedArgumentsProvider.scala:25 a: Apply(
  fun = Select(
    qualifier = Apply(
      fun = Select(
        qualifier = Select(qualifier = This(qual = ConvertToNamedArgumentsProvider), name = params),
        name = uri
      ),
      args = List()
    ),
    name = toString
  ),
  args = List()
)
ConvertToNamedArgumentsProvider.scala:26 a.isInstanceOf[NamedArg]: false
ConvertToNamedArgumentsProvider.scala:25 a: Select(qualifier = Ident(name = scala), name = None)
ConvertToNamedArgumentsProvider.scala:26 a.isInstanceOf[NamedArg]: false

```

tests/unit/src/test/scala/tests/codeactions/InsertInferredTypeLspSuite.scala
```diff
@@ -4,7 +4,12 @@ import scala.meta.internal.metals.codeactions.InsertInferredType
 import scala.meta.internal.metals.codeactions.RewriteBracesParensCodeAction
 
 class InsertInferredTypeLspSuite
-    extends BaseCodeActionLspSuite("insertInferredType") {
+    extends BaseCodeActionLspSuite(
+      "insertInferredType",
+      filterAction = act =>
```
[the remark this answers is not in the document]
camgraff: sounds good, I will move it to `check()`

tests/unit/src/test/scala/tests/codeactions/ConvertToNamedArgumentsSuite.scala
camgraff: Got it, thanks for the clarification

tanishiking (approved):
LGTM! great job!

tanishiking:
Let's wait for @tgodzik review, BTW, it would be awesome if you can squash commits into more meaningful granularities, at least squashing "scalafmt", "scalafix", and "add/fix tests" commits would be helpful.

tgodzik:
And of course I forgot to squash :man_facepalming: Maybe we should set it as a default

PR #3972, opened 2022-06-01
fix: lsif-java -> scip-java

ckipp01:
The project was recently renamed to scip-java and lsif-java has been
[deprecated](https://github.com/coursier/apps/pull/153). I see we are
having an issue where no main class is being found, which was attempted
to be fixed [here](https://github.com/coursier/apps/pull/155), but I'm
still unable to run it locally. This just ensures that we are using the
most recent setup to ensure that our sourcegraph uploads are still
working.

### DIFF
--- a/.github/workflows/sourcegraph.yml
+++ b/.github/workflows/sourcegraph.yml
@@ -19,8 +19,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: coursier/setup-action@v1.2.0-M2
-      - run: cs install --contrib lsif-java
-      - run: lsif-java index
+      - run: cs launch --contrib scip-java -- index
       - name: src lsif upload
         run: |
           mkdir -p bin
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,6 @@ website/.cache-loader
 
 project/metals.sbt
 coursier
+
+# Sourcegraph specific
+index.scip


ckipp01:
Just a heads up @olafurpg I see you shot in a pr to apps to fix the no main method issue, but locally and in CI I'm still getting it with the latest apps release.

.github/workflows/sourcegraph.yml
```diff
@@ -19,8 +19,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: coursier/setup-action@v1.2.0-M2
-      - run: cs install --contrib lsif-java
-      - run: lsif-java index
+      - run: cs install --contrib scip-java
+      - run: scip-java index
```
tanishiking, line 23:
```suggestion
      - run: cs launch --contrib scip-java -- index
```

Not sure why, but I could run `scip-java index` by `cs launch` instead of `cs install` 🤔

ckipp01: Yea I noticed that as well before. I suppose it's maybe just a good idea to do that now so the build stops being red, and then I'll change it later when the install works correctly.

ckipp01:
Looks like it's hitting on another issue. I'll dig into it.

tanishiking:
Looks like we need to use `libraryDependencies += "com.sourcegraph" %% "scip-java" % "0.8.0"` instead

ckipp01:
> Looks like we need to use `libraryDependencies += "com.sourcegraph" %% "scip-java" % "0.8.0"` instead

No I don't believe so. I think https://github.com/sourcegraph/sbt-sourcegraph just needs to be updated (which I'm working on right now). Because `scip-java` calls `sbt sourcegraphScip`, but the sbt plugin only has a `sbt sourcegraphLsif`. I just don't think the sbt plugin has been updated yet.

tanishiking:
Ah, right I misunderstood

ckipp01:
Alrighty, so I manually re-enabled the Sourcegraph job. Everything seems to be working now as you can see here: https://github.com/scalameta/metals/runs/7038835149?check_suite_focus=true.

tanishiking (approved):
LGTM!
So, this is now fixed because of https://github.com/coursier/apps/pull/156 ?

ckipp01:
> LGTM! So, this is now fixed because of [coursier/apps#156](https://github.com/coursier/apps/pull/156) ?

Yup, exactly.

PR #3976, opened 2022-06-02
bugfix: Remove Wconf from scalac options in worksheets

tgodzik:
It seems to be breaking currently in worksheets and I don't think it's neccessary to fix that currently.


dos65 (approved):
:+1:

PR #3977, opened 2022-06-03
chore(ci): add Scala Steward into CI

ckipp01:


.github/workflows/scala-steward.yml
```diff
@@ -0,0 +1,18 @@
+on:
+  schedule:
+    - cron: '0 0 1,15 *'
```
ckipp01, line 3: Since we have a cron here, I removed it from the `.scala-steward.conf` file.

.github/workflows/scala-steward.yml
```diff
@@ -0,0 +1,18 @@
...
+  scala-steward:
+    runs-on: ubuntu-latest
+    name: Launch Scala Steward
+    steps:
+      - name: Launch Scala Steward
+        uses: scala-steward-org/scala-steward-action@v2
+        with:
+          github-token: ${{ secrets.SCALAMETA_BOT_GH_TOKEN }}
+          author-email: scalameta@gmail.com
+          author-name: Scalameta Bot
```
ckipp01, line 18: I'm not 100% sure this will correctly tie it to the scalameta-bot account, but I _think_ it might.

ckipp01:
If VL ends up spinning up a public instance as mentioned [here](https://contributors.scala-lang.org/t/can-scala-maintain-a-public-scala-steward-instance/5754/3), then we can just remove it later.

.github/workflows/scala-steward.yml
ckipp01: I think this needs to match the public email and name of the bot, but I don't think the bot has either.

tgodzik (approved):
Let's try this out!

tgodzik:
Seems we are getting:
`invalid `cron` attribute "0 0 1,15 *"`

Should be soemthing along the lines of ` '0 5 * * 0,6'`

ckipp01:
Yup, I opened #3980 to address it.

PR #3978, opened 2022-06-03
chore: update version for v0.11.6

tanishiking:
Bump the versions for v0.11.6, following the release guide https://scalameta.org/metals/docs/contributors/releasing
Let's merge after we confirm that v0.11.6 is out on Maven central.

### DIFF
--- a/.github/workflows/mtags-auto-release.yml
+++ b/.github/workflows/mtags-auto-release.yml
@@ -8,11 +8,11 @@ on:
       metals_version:
         description: "Metals Version"
         required: true
-        default: "v0.11.5"
+        default: "v0.11.6"
       metals_ref:
         description: "Tag/branch-name from which run release"
         required: true
-        default: "v0.11.5_mtags_release"
+        default: "v0.11.6"
 jobs:
   test_and_release:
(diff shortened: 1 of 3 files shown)


.github/workflows/mtags-auto-release.yml
```diff
@@ -8,11 +8,11 @@ on:
       metals_version:
         description: "Metals Version"
         required: true
-        default: "v0.11.5"
+        default: "v0.11.6"
       metals_ref:
         description: "Tag/branch-name from which run release"
         required: true
-        default: "v0.11.5_mtags_release"
+        default: "v0.11.6_mtags_release"
```
dos65, line 15: you need either to put `v0.11.6` or create a branch `v0.11.6_mtags_release` from v0.11.6 tag.
I would prefer to update it to `v0.11.6` after release.
If there will be a need to port some changes for mtags release we could switch it on branch at any time

tanishiking: Ach, right, fixed to just `v0.11.6`, thank you for pointing out!

dos65 (approved):

PR #3981, opened 2022-06-03
chore: Update tests to work with newest nightlies

tgodzik:


tests/cross/src/test/scala/tests/pc/CompletionKeywordSuite.scala
```diff
@@ -379,7 +379,9 @@ class CompletionKeywordSuite extends BaseCompletionSuite {
       |  val x: Map[Int, new@@]
       |}
     """.stripMargin,
-    ""
+    "",
+    // to avoid newMain annotation
+    filter = str => !str.contains("newMain")
```
ckipp01, line 384: Should the annotation even be offered here? Is that a bug in Dotty?

tgodzik: It's a type, so I think it's valid in most places. The same works currently with for example `tailrec`. Maybe it can be improved for cases though, but it should automatically do `_:` instead I think

PR #3995, opened 2022-06-05
chore(deps): add in other scalameta repos to steward job

ckipp01:
So this again is sort of temporary with the hopes that a public instance
gets added, but since we already have this up and running for Metals,
it's quite easy to just add the other scalameta repos into this job as
well. I copied the list of scalameta projects from
`scala-steward-org/repos`.


tgodzik (approved):
Thanks! I need to setup something similar for Bloop finally

PR #3996, opened 2022-06-05
feature: Allow to run all scalafix rules on a file

tgodzik:
Previously, it would not be possible to run sclafix inside metals aside from organize imports. Now, it's possible to run a list of predefined rules and ones added in settings.

Fixes https://github.com/scalameta/metals/issues/4005

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
@@ -95,6 +95,12 @@ object Messages {
     s"Metals is unable to start ${buildTool}. Please try to connect after starting it manually."
   )
 
+  def unknownScalafixRules(unknownRules: Set[String]) =
+    new MessageParams(
+      MessageType.Warning,
+      s"Metals is unable to run ${unknownRules.mkString(", ")}. Please add the rule dependency to `metals.scalafixRulesDependencies`."
+    )
+
   val ImportProjectFailed = new MessageParams(
     MessageType.Error,
     "Import project failed, no functionality will work. See the logs for more details"
--- a/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Messages.scala
@@ -95,11 +95,18 @@ object Messages {
     s"Metals is unable to start ${buildTool}. Please try to connect after starting it manually."
   )
 
-  def unknownScalafixRules(unknownRules: Set[String]) =
+  def unknownScalafixRules(unknownRuleMessage: String): MessageParams = {
+    // To match: "Rule not found 'ARuleName'"
+    val regex = ".*'(.*)'.*".r
+    val rule = unknownRuleMessage match {
+      case regex(rule) => rule
+      case _ => "a rule"
+    }
     new MessageParams(
       MessageType.Warning,
-      s"Metals is unable to run ${unknownRules.mkString(", ")}. Please add the rule dependency to `metals.scalafixRulesDependencies`."
+      s"Metals is unable to run ${rule}. Please add the rule dependency to `metals.scalafixRulesDependencies`."
     )
+  }
 
   val ImportProjectFailed = new MessageParams(
     MessageType.Error,
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
@@ -20,6 +20,8 @@ import scala.meta.internal.mtags.SemanticdbClasspath
 import scala.meta.internal.semanticdb.TextDocuments
 import scala.meta.io.AbsolutePath
 
+import com.typesafe.config.ConfigFactory
+import coursierapi.Dependency
 import org.eclipse.lsp4j.MessageParams
 import org.eclipse.lsp4j.MessageType
 import org.eclipse.{lsp4j => l}
@@ -40,8 +42,8 @@ case class ScalafixProvider(
 )(implicit ec: ExecutionContext) {
   import ScalafixProvider._
   private val scalafixCache = TrieMap.empty[ScalaBinaryVersion, Scalafix]
-  private val organizeImportRuleCache =
-    TrieMap.empty[ScalaBinaryVersion, URLClassLoader]
+  private val rulesClassloaderCache =
+    TrieMap.empty[ScalafixRulesClasspathKey, URLClassLoader]
 
   // Warms up the Scalafix instance so that the first organize imports request responds faster.
   def load(): Unit = {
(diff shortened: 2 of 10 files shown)


metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -158,6 +158,17 @@ object ServerCommands {
        |""".stripMargin
   )
 
+  val RunScalafix = new ParametrizedCommand[TextDocumentPositionParams](
+    "scalafix-run",
+    "Run all Scalafix Rules",
+    """|Run all the supported scalafix rules in your codebase.
+       |
+       |If the rules are missing please add them to user configuration `metals.scalafixRulesDependencies`.
```
kubukoz, line 166: this is in coursier format, right?

tgodzik: That's what I am thinking. It would be better to get that info from sbt, but it's not currently exposed. I am going to add a bunch of docs for that

tgodzik:
I think the approach is as reasonable as it can get, though I still wonder if we should only run available rules instead of failing in that case :thinking: it might be actually more work for the user to add it a more complex setting. The best road would be to get the scalafix deps from the build tool, but we currently don't have that and we would need to change that everywhere that scalafix can be used (which might be a bunch of places)

I will try add more defaults based on the scalafix contributed rules pages.

nolondil:
Hey, great to see that you're adding a way to run all scalafix rules!

With that, would it be easier for me to wire the diagnostics for https://github.com/scalameta/metals-feature-requests/issues/230?

I was a bit stuck with wiring everything and downloading the dependency when I tried to tackle my feature request

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -240,6 +284,24 @@ case class ScalafixProvider(
     }
   }
 
+  private def rulesFromScalafixConf(): Set[String] = {
+    scalafixConf(false) match {
```
ckipp01, line 288:
```suggestion
    scalafixConf(isScala3 = false) match {
```

ckipp01 (commented):
Very cool. This was just a cursorary glance from me. I want to play around with it a bit. I agree, I totally we could just pull these from the build tool since now if you are using a custom rule of sort you'll need to remember to not only include it in your build to use via sbt, but also in your Metals configuration.

>  only run available rules instead of failing

I think if this is possible, then I'd favor this. Because let's say you have 3 built-in rules and one custom one, it'd be nice to still be able to trigger and run those 3 built in ones and just leave out the custom once instead of failing the whole thing. Then we'd need to still figure out how to nicely show the user that info, something like "hey! We were able to run scalafix, but we couldn't find X rule. You'll need to add it to run it"

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -364,20 +433,25 @@ case class ScalafixProvider(
...
+      scalfixRulesKey: ScalafixRulesClasspathKey,
       scalafixClassLoader: ClassLoader
   ): Try[URLClassLoader] = {
-    organizeImportRuleCache.get(scalaBinaryVersion) match {
+    rulesClassloaderCache.get(scalfixRulesKey) match {
       case Some(value) => Success(value)
       case None =>
-        statusBar.trackBlockingTask("Downloading organize import rule") {
+        statusBar.trackBlockingTask(
+          "Downloading scaladix rules' dependencies"
```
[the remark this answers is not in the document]
kubukoz: lmao

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
```diff
@@ -95,6 +95,18 @@ object Messages {
     s"Metals is unable to start ${buildTool}. Please try to connect after starting it manually."
   )
 
+  def unknownScalafixRules(unknownRuleMessage: String): MessageParams = {
+    val regex = ".*'(.*)'.*".r
+    val rule = unknownRuleMessage match {
+      case regex(rule) => rule
+      case _ => "a rule"
+    }
```
kpodsiad, line 103: Can you add an example what will be matched by regex and why as a comment?

kubukoz:
noob question, how can I trigger this command on a local server snapshot?

dos65 (approved):
Great job!

I'm only thinking that it's not fully IDE integrated without highlighting.

I'm wondering if we can try to turn Patches not only into textEdits but also into code actions?
I mean, if you have an `ExplicitResultType` rule, you won't notice that you need to run scalafix.

dos65:
@kubukoz you also need to run extension from [this branch](https://github.com/scalameta/metals-vscode/pull/1013#pullrequestreview-998532277)

ckipp01 (approved):
Have been playing around with a bit, and didn't hit on anything funky yet! LGTM!

tgodzik:
> Very cool. This was just a cursorary glance from me. I want to play around with it a bit. I agree, I totally we could just pull these from the build tool since now if you are using a custom rule of sort you'll need to remember to not only include it in your build to use via sbt, but also in your Metals configuration.

The fastest way to support it would be making scalafix take the dependencies in the configuration file, which was already suggested by @olafurpg

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -400,8 +474,52 @@ object ScalafixProvider {
   type ScalaBinaryVersion = String
   type ScalaVersion = String
 
+  case class ScalafixRulesClasspathKey(
+      scalaBinaryVersion: ScalaBinaryVersion,
+      usedRulesWithClasspath: Set[Dependency]
+  )
+
+  def scalafixRulesClasspathKey(
```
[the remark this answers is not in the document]
tgodzik: Added a companion object and an apply

metals/src/main/scala/scala/meta/internal/metals/Messages.scala
tgodzik: Added an example message

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -400,8 +474,52 @@ object ScalafixProvider {
...
+      userConfig: UserConfiguration
+  ): ScalafixRulesClasspathKey = {
+    val rulesClasspath =
+      rulesDependencies(scalaVersion, scalaBinaryVersion, userConfig)
+    ScalafixRulesClasspathKey(scalaBinaryVersion, rulesClasspath.toSet)
+  }
   case class ScalafixRunException(msg: String) extends Exception(msg)
 
   val organizeImportRuleName = "OrganizeImports"
+  val explicitResultTypesRuleName = "ExplicitResultTypes"
```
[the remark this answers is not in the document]
tgodzik: Removed

Issue #4001, opened 2022-06-07
Constant prompts to restart Bloop


ckipp01:
### Describe the bug

I've had some [reports of this](https://github.com/scalameta/nvim-metals/discussions/411) in `nvim-metals` and I'm also hitting on it this morning so thought I'd open this up although it's not 100% clear what's causing it yet. My guess is that Bloop just isn't being correctly shut down.

However, the issue is that when you open a project you're prompted like so:

```
Setting Metals Java Home will result in updating Bloop's global Json file, by Metals.
Bloop will need to be restarted in order for these changes to take effect.

Request Actions:
1. Apply and restart Bloop
2. Not now
Type number and <Enter> or click with the mouse (q or empty cancels):
```

Then when you select `1.` , I end up getting prompted again for the exact same thing:

```
Setting Metals Java Home will result in updating Bloop's global Json file, by Metals.
Bloop will need to be restarted in order for these changes to take effect.

Request Actions:
1. Apply and restart Bloop
2. Not now
Type number and <Enter> or click with the mouse (q or empty cancels):
```

It doesn't seem to happen a third time, but when I look in the logs I see:

```
2022.06.07 07:18:46 INFO  Connected to Build server: Bloop v1.5.0-22-91111c15
2022.06.07 07:18:49 INFO  time: indexed workspace in 2.58s
2022.06.07 07:19:14 INFO  Disconnecting from Bloop session...
2022.06.07 07:19:16 ERROR timeout: build server 'bloop' during shutdown
2022.06.07 07:19:16 ERROR timeout: build server 'bloop' during shutdown
2022.06.07 07:19:16 INFO  Attempting to connect to the build server...
2022.06.07 07:19:18 INFO  tracing is enabled: /Users/ckipp/Library/Caches/org.scalameta.metals/bsp.trace.json
2022.06.07 07:19:18 INFO  Attempting to connect to the build server...
2022.06.07 07:19:18 INFO  tracing is enabled: /Users/ckipp/Library/Caches/org.scalameta.metals/bsp.trace.json
2022.06.07 07:19:19 INFO  time: Connected to build server in 2.49s
2022.06.07 07:19:19 INFO  Connected to Build server: Bloop v1.5.0-22-91111c15
2022.06.07 07:19:19 INFO  time: Imported build in 0.11s
2022.06.07 07:19:20 INFO  time: indexed workspace in 1.24s
2022.06.07 07:20:06 INFO  Disconnecting from Bloop session...
[...]

PR #4002, opened 2022-06-07
fix: only compare Bloop JavaHome with Metals JavaHome if set

ckipp01:
Currently even if a user hasn't set the `javaHome` setting, but the
`.bloop/bloop.json` has a `javaHome` setting we try to compare them
which will always be false, since we're comparing `None !=
Some(javaHome)`. This changes the logic to compare the contained values
_only_ if they are set.

This is part of what is going on in #4001.


metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -299,7 +299,11 @@ final class BloopServers(
                 requested != maybeBloopGlobalJsonJvmProperties
               )
           ) Some(BloopJsonUpdateCause.JVM_OPTS)
-          else if (maybeRequestedMetalsJavaHome != maybeBloopGlobalJsonJavaHome)
+          else if (
+            maybeRequestedMetalsJavaHome.exists { metalsJavaHome =>
+              maybeBloopGlobalJsonJavaHome.exists(_ != metalsJavaHome)
+            }
+          )
```
tanishiking, line 306: We might need to update `.bloop/bloop.json` if `.bloop/bloop.json` doesn't have `javaHome` and Metals' `bloop java home` has a value 🤔 

Summary:

|     |      maybeGlobalJsonJavaHome = None      |  maybeGlobalJsonJavaHome=Some(value) |
|----------|:-------------:|------:|
| maybeRequestedMetalsJavaHome=None |  nothing happens | nothing happens |
| maybeRequestedMetalsJavaHome=Some(value) |    should write value to json |  compare version and if they're different, write value to json  |

ckipp01: Ah, good point. I'll update this.

tanishiking: So, I misunderstood `maybeRequestedMetalsJavaHome` is always `Some(value)` in https://github.com/scalameta/metals/pull/3969#discussion_r887650732 but it is the case only with `metals-vscode`...

tanishiking: FYI @tgodzik

ckipp01:
Alright, updated. I added some comment so it should be clear.

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
ckipp01: Ahhh, I totally missed that it's always set with VS Code. Yea, that won't happen in other clients.

tanishiking (approved):
LGTM! It is much more understandable, nice catch 👍

tgodzik (approved):
Thanks, I noticed that I haven't thought of this scenario. Should do a Bugfix release with this?

Issue #4003, opened 2022-06-07
Server Experimental Capabilities are no longer not passing through


ninth-dev:
### Describe the bug

Server Experimental Capabilities are not passing through. 

1. Enable Logging 
2. Start Metals 
3. Check Logs
4. Check the initialize response
5. Check the value under `.capabilities.experimental`
6. Observe `experimental: {} ... `

### Expected behavior

Expect the server experimental to have `{'rangeHoverProvider': True}`

See [MetalsExperimental.scala](https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala#L8-L110) and [MetalsLanguageServer.scala]( 
 https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L894)

### Operating system

macOS

### Editor/Extension

Sublime

### Version of Metals

v0.11.6

### Extra context or search terms

Sublime relies on the experimental flag to enable the `Type annotations on code selection` feature.

Last known working version `v0.11.2`.

PR #4007, opened 2022-06-07
bugfix: Properly send back experimental capabilities

tgodzik:
Fixes https://github.com/scalameta/metals/issues/4003

This seems to work, while the object approach looked to not be working before.

@ayoub-benali could you cofirm the fix?

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
@@ -7,5 +7,5 @@ package scala.meta.internal.metals
 
 case class MetalsExperimental(
     // A workaround for https://github.com/microsoft/language-server-protocol/issues/377 until it is resolved
-    val rangeHoverProvider: java.lang.Boolean = true
+    val rangeHoverProvider: Boolean = true
 )
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
@@ -5,7 +5,7 @@ package scala.meta.internal.metals
  * capabilities" in the initialize response.
  */
 
-object MetalsExperimental {
-  // A workaround for https://github.com/microsoft/language-server-protocol/issues/377 until it is resolved
-  val rangeHoverProvider: java.lang.Boolean = true
-}
+case class MetalsExperimental(
+    // A workaround for https://github.com/microsoft/language-server-protocol/issues/377 until it is resolved
+    val rangeHoverProvider: java.lang.Boolean = true
+)
--- a/tests/unit/src/test/scala/tests/QuickBuildSuite.scala
+++ b/tests/unit/src/test/scala/tests/QuickBuildSuite.scala
@@ -10,7 +10,7 @@ class QuickBuildSuite extends BaseLspSuite(s"quick-build") {
   test("basic", withoutVirtualDocs = true) {
     cleanCompileCache("b")
     for {
-      _ <- initialize(
+      initResults <- initialize(
         """
           |/metals.json
           |{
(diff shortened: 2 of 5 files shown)


ayoub-benali:
@tgodzik Problem fixed thanks !
Just out of curiosity why did it break ? a change in the API for library or LSP ?

Also, is it possible to add a test case to prevent future regressions ?

tgodzik:
> @tgodzik Problem fixed thanks ! Just out of curiosity why did it break ? a change in the API for library or LSP ?
> 
> Also, is it possible to add a test case to prevent future regressions ?

I will try to add something, though I have a suspicion that it might have never worked.

ayoub-benali:
As posted in the original issue, the feature used to work. The regression appeared after v0.11.2

tests/unit/src/test/scala/tests/QuickBuildSuite.scala
```diff
@@ -45,6 +45,11 @@ class QuickBuildSuite extends BaseLspSuite(s"quick-build") {
           |}
         """.stripMargin
       )
+      // metals should always return experimental capabilities
+      _ = assertNoDiff(
```
tgodzik, line 49: Didn't want to create a while new suite for this, but it should be fine anyway.

metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
```diff
@@ -5,7 +5,7 @@ package scala.meta.internal.metals
  * capabilities" in the initialize response.
  */
 
-object MetalsExperimental {
-  // A workaround for https://github.com/microsoft/language-server-protocol/issues/377 until it is resolved
-  val rangeHoverProvider: java.lang.Boolean = true
-}
+case class MetalsExperimental(
```
ckipp01, line 8: Weird, so was this just an issue with how the object was getting serialized?

ckipp01 (approved):
Just a quick question, but otherwise LGTM!

metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
kpodsiad: I checked and yes, Gson is emitting empty json object for scala object
```scala
//> using scala "2.13.8"
//> using lib "com.lihaoyi::pprint:0.7.3"
//> using lib "com.google.code.gson:gson:2.9.0"


import com.google.gson.Gson

object GsonExample extends App {
  val gson = new Gson()

  case object Data {
    final val boolean: java.lang.Boolean = true
  }

  case class Data_2(boolean: Boolean)

  pprint.log(gson.toJson(Data))
  pprint.log(gson.toJson(Data_2(true)))
}
```
```
Gson.scala:17 gson.toJson(Data): "{}"
Gson.scala:18 gson.toJson(Data_2(true)): "{\"boolean\":true}"
```

metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
```diff
@@ -5,7 +5,7 @@ package scala.meta.internal.metals
...
+    // A workaround for https://github.com/microsoft/language-server-protocol/issues/377 until it is resolved
+    val rangeHoverProvider: java.lang.Boolean = true
```
kpodsiad, line 10:
```suggestion
    val rangeHoverProvider: Boolean = true
```

metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
ckipp01: `scala-cli`, providing easy reproductions since 2022. ™️

metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
tgodzik: Interesting, I always though gson didn't manage to make it work properly with scala Boolean. Or is it just an alias?

metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
tgodzik: I tested it and it never seemed to have worked with an object. No idea, why it worked before.

metals/src/main/scala/scala/meta/internal/metals/MetalsExperimental.scala
tgodzik: Seems to work ok, thanks!

ayoub-benali:
Thanks @tgodzik for the fix

PR #4008, opened 2022-06-08
docs: Add architecture.md

tanishiking:
close https://github.com/scalameta/metals-feature-requests/issues/276

- > Do name important files, modules, and types. Do not directly link them (links go stale). Instead, encourage the reader to use symbol search to find the mentioned entities by name. 
- > the shorter it is, the less likely it will be invalidated by some future change.

I didn't include components for each feature like `CompletionProvider`, `ReferenceProvider`, and so on, because they are easy to follow from the `MetalsLanguageServer`. I would like to focus on the things that are hard to follow by code navigation without the architectural knowledge of Metals (for example, how Metals communicate with semanticdb, presentation compilers, and build servers).

I want to update this document iteratively, and hear what we should add / drop from both recurring contributors and occasional contributors.


zmerr:
Super glad that this document is getting produced 💯  Thank you so much @tanishiking

architecture.md
```diff
@@ -0,0 +1,56 @@
...
+@JsonRequest("textDocument/completion")
+def completion(...) = ...
+```
+
+In addition to the LSP endpoints, this file contains many components, for example `private var semanticdbs: Semanticdbs = _`. Those components will be initialized later at some points like on LSP's `initialize` request.
+
+## Presentation compiler
+Metals features are powered by presentation compilers, if you hit `Compilers.scala` it is the client of compilers.
+
+`ScalaPresentationCompiler.scala` under `scala-2` and `scala-3` works as a wrapper of Scala compilers.
```
tgodzik, line 23: MAybe let's mention why they are separate and how mtags work?

architecture.md
```diff
@@ -0,0 +1,56 @@
...
+
+For more details
+
+- [An introduction to the Scala presentation compiler](https://www.chris-kipp.io/blog/an-intro-to-the-scala-presentation-compiler)
+- Presentation compilers' endpoint
```
tgodzik, line 28: There are endpoint from the compiler, so maybe explain that metals contacts those endpoint via those classes?

architecture.md
```diff
@@ -0,0 +1,56 @@
...
+  - Scala3 `InteractiveDriver.scala` in [lampepfl/dotty](https://github.com/lampepfl/dotty)
+  - Scala2: `interactive/Global.scala` in [scala/scala](https://github.com/scala/scala)
+
+## SemanticDB
+Metals uses [semanticdb](https://scalameta.org/docs/semanticdb/guide.html) for many features (Providers that receive the instance of `AggreagateSemanticdbs`). Metals consumes SemanticDBs in two ways:
```
tgodzik, line 33:
```suggestion
Metals uses [semanticdb](https://scalameta.org/docs/semanticdb/guide.html) for many features (Providers that receive the instance of `Semanticdbs`). Metals consumes SemanticDBs in two ways:
```

architecture.md
```diff
@@ -0,0 +1,56 @@
...
+
+- `FileSystemsSemanticdbs` consumes SemanticDBs on disks
+  - Semanticdb will be generated by the build servers to disk, see https://scalameta.org/metals/docs/integrations/new-build-tool#enable-semanticdb
+- `InteractiveSemanticdbs` will generate SemanticDB on the fly, in case `FileSystemsSemanticdbs` failed.
```
tgodzik, line 37:
```suggestion
- `InteractiveSemanticdbs` will generate SemanticDB on the fly, in case `FileSystemsSemanticdbs` failed or when the build tool doesn't compiler specific files such as dependency sources or sbt files
```

tgodzik (commented):
Thanks for looking into that ! Ping me if you need help with any of the sctions

architecture.md
tanishiking: I added some words about `mtags` 👍

architecture.md
```diff
@@ -0,0 +1,72 @@
...
+
+Metals uses [semanticdb](https://scalameta.org/docs/semanticdb/guide.html) for many features (Providers that receive the instance of `Semanticdbs`). Metals consumes SemanticDBs in two ways:
+
+- `FileSystemsSemanticdbs` consumes SemanticDBs on disks
+  - Semanticdb will be generated by the build servers to disk, see https://scalameta.org/metals/docs/integrations/new-build-tool#enable-semanticdb
+- `InteractiveSemanticdbs` will generate SemanticDB on the fly using the presentation compiler via `mtags`, in case `FileSystemsSemanticdbs` failed or when the build tool doesn't compiler specific files such as dependency sources or sbt files.
+
+In addition, classes extend `SemanticdbFeatureProvider` index the semanticdb symbols, that will be updated by `SemanticdbIndexer`.
+
+## Build Server
```
tanishiking, line 47: I'm not sure this is enough, but `architecture.md` is basically for pointing out the endpoint of the topic, so I won't go further and let readers navigate around source code.

architecture.md
```diff
@@ -0,0 +1,72 @@
...
+## Formatting
+
+`FormattingProvider.scala` takes care of how Metals handles `textDocument/formatting`. It uses [scalafmt](https://github.com/scalameta/scalafmt) as a code formatter downloading dynamically using `scalafmt-dynamic`.
+We don't embed a specific version of scalafmt into Metals so that users can switch scalafmt's version using `.scalafmt.conf`.
+
+Note that `FormattingProvider` doesn't handle `textDocument/rangeFormatting`.
+
+## Scalafix
+
+TBD
```
tanishiking, line 68: @tgodzik would be the best person to write this section 😄 ?

tgodzik (changes requested):
Sorry for the veeeery late review, but finally I managed to add some paragraphs.

tanishiking:
Hey @tgodzik  thank you for adding those sections! I guess it's ready to merge :)

tgodzik (approved):
LGTM!

Issue #4000, opened 2022-06-06
"Show decompiled with CFR" doesn't show the decompiled sources


armanbilge:
### Describe the bug

https://github.com/armanbilge/sandbox/tree/cfr-bug

1. Open https://gitpod.io/#https://github.com/armanbilge/sandbox/tree/cfr-bug
2. `sdk install scalacli`
3. `scala-cli setup-ide bug.scala`
4. Import build to Metals.
5. Metals: Show decompiled with CFR.

Output:
```
Dependency(Module(org.benf, cfr), 0.151)
org.benf.cfr.reader.Main
/workspace/sandbox/.scala-build/project_183d125c5c-45290ff20c/classes/main
List(--extraclasspath, /workspace/sandbox/.scala-build/project_183d125c5c-45290ff20c/classes/main:/workspace/sandbox/.scala-build/project_183d125c5c-45290ff20c/classes/main:/home/gitpod/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-library_3/3.1.2/scala3-library_3-3.1.2.jar:/home/gitpod/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.8/scala-library-2.13.8.jar:/home/gitpod/.cache/scalacli/local-repo/v0.1.7/org.virtuslab.scala-cli/stubs/0.1.7/jars/stubs.jar:/home/gitpod/.cache/coursier/v1/https/repo1.maven.org/maven2/com/sourcegraph/semanticdb-javac/0.7.4/semanticdb-javac-0.7.4.jar, --elidescala, true, --analyseas, CLASS, bug$package.class)
Picked up JAVA_TOOL_OPTIONS:  -Xmx3435m
```

### Expected behavior

To see decompiled sources :)

### Operating system

Linux

### Editor/Extension

VS Code

### Version of Metals

v0.11.6

tgodzik:
Thanks for reporting! It seems that if we have `JAVA_TOOL_OPTIONS` set it will get printed to the error stream and because of that we assume it;s broken (but it's actually not)

armanbilge:
Aha, thanks. `JAVA_TOOL_OPTIONS` is just one of those annoying GitPod things.

tgodzik:
Trying to fix it in https://github.com/scalameta/metals/pull/4012 but not sure how to test it without modifying everything to run with the tools option

PR #4012, opened 2022-06-08
bugfix: Always print out CFR output if there is any available

tgodzik:
fixes https://github.com/scalameta/metals/issues/4000

To test it we would need to add the env flag, but I am not sure how to add it to the single test. And the change is rather straightforward


dos65 (approved):

armanbilge:
Where is the relevant test suite for this?

kpodsiad (approved):

Issue #3997, opened 2022-06-06
Synthetic's link in for yield doesn't work for `<-`


kpodsiad:
### Describe the bug

EDIT: by `No synthetic for evidence >>` I meant no link in synthetic ;)

```scala
//> using scala "2.13.8"

object Example extends App {
  class Evidence
  implicit val evidence: Evidence = ???

  def printOpt[T](t: T)(implicit ev: Evidence): Option[Unit] = Option(())
  def print[T](t: T)(implicit ev: Evidence) = ()

  for {
    number <- Option(5)
    _ <- printOpt("No synthetic for evidence >>")
    _ = print("here it is >>")
  } yield ()
}
```
Both `print` and `printOpt` have synthetics for `String` and `evidence`. However, `printOpt`'s evidence doesn't have link to used implicit argument.

To reproduce:
1. Hover over `print`'s `evidence` - there is link.
2. Hover over `printOpt`'s `evidence` - there is no link.

<img width="619" alt="Screenshot 2022-06-06 at 10 56 08" src="https://user-images.githubusercontent.com/37124721/172129584-dbd33985-4e57-4935-b64f-3b0a376a1b04.png">

### Expected behavior

`printOpt`'s `evidence` has link to the implicit argument either.

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v.0.11.6

### Extra context or search terms

synthetic decoration

PR #4014, opened 2022-06-09
bugfix: Properly show links to implicit parameters in decoration hovers

tgodzik:
Previously, no links would be shown for implicit parameter decorations in for comprehension. Now it's properly showed for the user.

The issue was that we were filtering synthatic decorations for a line based on the toplevel synthetic range, which in case of for comprehensions would be the full for comprehensions tree.

Fixes https://github.com/scalameta/metals/issues/3997


kpodsiad:
`tests.decorations.SyntheticHoverLspSuite.hovers` fails

tgodzik:
> `tests.decorations.SyntheticHoverLspSuite.hovers` fails

It should work now! `SyntheticHoverLspSuite` should always show the line even if synthetic is not enclosing the cursor.

kpodsiad (approved):

PR #4015, opened 2022-06-09
chore: Fix errors for kind projector

tgodzik:
Previously, we were testing kind projector with `?` and since Scala 2.12.16 allows to use instead of `_`, the test stopped working. Now, the test case is updated to work with `*` instead and kind projector is updated to newest. I also the compiler plugins in tests to update automatically.


tanishiking (approved):
LGTM!

Issue #4006, opened 2022-06-07
popup: Test Explorer won't work due to mis-configuration


jimka2001:
### Describe the bug

I'm running vscode/metals within gitpod, so I don't know whether this is a metals issue, vscode issue, or gitpod issue.   But I often see the  popup message shown in the image here.  When I click on the "More Information" button nothing happens.   Any suggestions about what this means, or what I should be doing about it? 

<img width="469" alt="Screenshot 2022-06-06 at 18 25 15" src="https://user-images.githubusercontent.com/6414129/172371470-9f1b07cb-eab8-4187-9790-425d4c0fd9a0.png">

### Expected behavior

I'm not 100% sure what to expect.  Perhaps that the popup tell me more information about what I'm doing wrong?  

### Version of Metals

v0.9.10

### Extra context or search terms

The bug report form asks the OS.  Since I'm running in gitpod, I don't really know the OS.
I'm running gitpod in my web browser on MacOS, but I doubt the remote machine is macOS.

tgodzik:
Thanks for reporting! What does happen after clicking `More information` ? What testing library are you using? @kpodsiad maybe we could easy of with the warning and only show it in the doctor?

kpodsiad:
@tgodzik nothing' haooens after clicking on it. I spotted this behavior but I didn't have enought time to investigate what's causing the problem.

However, I know why nothing happens. It's because we keep track doctor opened/closed state and this action probably doesn't change that state.

jimka2001:
nothing seems to happen when I tap More information.  It seems to have no effect.  I forgot whether that makes the popup go away or not.

PR #4019, opened 2022-06-11
fix: run doctor when `more information` button was clicked

kpodsiad:
closes https://github.com/scalameta/metals/issues/4006

https://github.com/scalameta/metals/pull/3768 introduced doctor visibility provider which prevents from calling metals doctor when it isn't necessary. Doctor status should be set to visible if user want to see more information (see doctor view).


tgodzik (approved):
LGTM!

Issue #4023, opened 2022-06-13
vscode-metals doesn't work for  Apache Spark


jievince:
### Describe the bug

`Go to definition` doesn't work for https://github.com/apache/spark
![image](https://user-images.githubusercontent.com/38901892/173292239-de1af070-de77-48e9-9189-cf9459488892.png)

![image](https://user-images.githubusercontent.com/38901892/173292282-2c5b08ac-4fdf-4ff2-9d0d-1d662d65a533.png)

The log: 
```
Java home: /usr/lib/jvm/java-11-openjdk-amd64
Metals version: 0.11.6
2022.06.13 14:11:02 INFO  logging to file /home/jie.wang/spark/.metals/metals.log
2022.06.13 14:11:02 INFO  Started: Metals version 0.11.6 in workspace '/home/jie.wang/spark' for client Visual Studio Code 1.68.0.
2022.06.13 14:11:02 INFO  time: initialize in 0.55s
Initializing Scala Debugger
2022.06.13 14:11:02 INFO  Flyway Community Edition 8.5.10 by Redgate
2022.06.13 14:11:02 INFO  See what's new here: https://flywaydb.org/documentation/learnmore/releaseNotes#8.5.10
2022.06.13 14:11:02 INFO  
2022.06.13 14:11:02 INFO  Database: jdbc:h2:file:/home/jie.wang/spark/.metals/metals (H2 2.1)
2022.06.13 14:11:02 WARN  Flyway upgrade recommended: H2 2.1.212 is newer than this version of Flyway and support has not been tested. The latest supported version of H2 is 2.1.210.
2022.06.13 14:11:02 INFO  Successfully validated 3 migrations (execution time 00:00.021s)
2022.06.13 14:11:02 INFO  Current version of schema "PUBLIC": 3
2022.06.13 14:11:02 INFO  Schema "PUBLIC" is up to date. No migration necessary.
2022.06.13 14:11:02 INFO  Attempting to connect to the build server...
2022.06.13 14:11:02 INFO  tracing is disabled for protocol BSP, to enable tracing of incoming and outgoing JSON messages create an empty file at /home/jie.wang/spark/.metals/bsp.trace.json or /home/jie.wang/.cache/metals/bsp.trace.json
2022.06.13 14:11:02 INFO  Attempting to connect to the build server...
2022.06.13 14:11:02 INFO  Attempting to connect to the build server...
[...]

PR #4024, opened 2022-06-13
chore(maven): update the maven wrapper

ckipp01:
Looks like this can potentially cause issues with Maven builds that are
requiring a more modern version of Maven.

refs #4023


tgodzik:
Seems we are getting:
` no main manifest attribute,`
but we are running via `-jar` so maybe something changed and we need to add the main class?

ckipp01:
> Seems we are getting: ` no main manifest attribute,` but we are running via `-jar` so maybe something changed and we need to add the main class?

Yup, I'll look into it a bit tonight. Also the wrapper we are using is actually deprecated and maven publishes their own wrapper now, so I'm looking at using that instead.

tgodzik:
> Maybe you remember from before @tgodzik but the old `maven-wrapper.jar` that we have does have the `Main-class` specified in the manifest, however when I pull that same version down and look, it's not there:
> 
> ```
> cp $(cs fetch io.takari:maven-wrapper:0.5.5) test.jar
> ```
> 
> And then when I look at the manifest, there is no main. Do you remember how you got this jar? I am able to get it to work by instead using `-cp <jar> <main>` which I _think_ should be fine, especially since we are testing this, but it is a bit different.

I think I just downloaded it from github. Or maybe I also modified it to have the main class in the manifest? That was a long time ago and I am not super sure about it. Are we able to migrate to the official wrapper?

ckipp01:
> Are we able to migrate to the official wrapper?

Yup, tests are now ✅ .

.jvmopts
```diff
@@ -1,6 +1,5 @@
 -Xss4m
--Xms1G
--Xmx2G
+-Xmx1G
```
tgodzik, line 2: Do we need to change `.jvmopts`? Did you find that these were not neccessary? I think I changed these for the CI :thinking:

tgodzik (approved):
LGTM! Just one question otherwise

.jvmopts
ckipp01: Ach, sorry, I was running `.bin/test.sh` which changes these automatically. I will revert this.

tgodzik: Ach, so `-Xmx1G` is actually used for the CI, totally forgot

PR #4026, opened 2022-06-14
Adjust to signature help changes in scala3 compiler

dos65:
Fixes tests failures in latest nightly
@tgodzik I rebased your branch

### DIFF
--- a/mtags/src/main/scala-3.2/scala/meta/internal/pc/MetalsSignatures.scala
+++ b/mtags/src/main/scala-3.2/scala/meta/internal/pc/MetalsSignatures.scala
@@ -1,0 +1,30 @@
+package scala.meta.internal.pc
+
+import dotty.tools.dotc.ast.tpd
+import dotty.tools.dotc.util.SourcePosition
+import dotty.tools.dotc.core.Contexts.*
+import dotty.tools.dotc.util.Signatures
+import dotty.tools.dotc.util.Signatures.Signature
+import scala.meta.internal.mtags.MtagsEnrichments.*
+import scala.meta.pc.SymbolSearch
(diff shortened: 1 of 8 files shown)


dos65:
The new method `Signatures.signatureHelp` works only from 2022-06-10 nightly while `test download dependencies` tries to compile 5 last nighty releases(which includes 2022-06-09 that doesn't have this method yet)

I think it's easier to merge it as it is. This issue with test will disappear after one more nightly version will be released.

mtags/src/main/scala-3.2/scala/meta/internal/pc/MetalsSignatures.scala
```diff
@@ -0,0 +1,30 @@
...
+
+object MetalsSignatures:
+
+  def signatures(
+      search: SymbolSearch,
+      path: List[tpd.Tree],
+      pos: SourcePosition
+  )(using ctx: Context): (Int, Int, List[(Signature, Denotation)]) =
+    val (paramN, callableN, alternatives) =
+      Signatures.signatureHelp(path, pos.span)
```
tanishiking, line 21: [note]
`callInfo` is deprecated in favor of `signatureHelp`
https://github.com/lampepfl/dotty/pull/15278

tanishiking (approved):
The change itself is LGTM:tada: (except cross/test looks failing).
Also, can you add comments in code or commit message that why do we have this change pointing to the issue https://github.com/lampepfl/dotty/issues/15244 ?

tanishiking:
:+1: to commit message

dos65:
hmm. latest nightly cross test fails again :disappointed:

Issue #3928, opened 2022-05-19
[Scala3] Better OverrideCompletions


tanishiking:
### Describe the bug

Follow-up of https://github.com/scalameta/metals/pull/3897

## [ ] OverrideCompletion should complete already overridden symbol
see: `OverrideCompletionSuite#overriden-twice`

```scala
trait A {
  def close: Unit
}
trait B extends A{
  override def close : Unit = {}
}
class C extends B{
  clos@@
}
```

In this case, Metals doesn't offer override completion for `close` method.

### Expected behavior

Metals should complete `override def close: Unit = ${0:???}`

---

## [ ] should complete more accurate signature on error
See `CompletionOverrideSuite#error`.

In Scala3, we print `Iterator[Any]` for `Iterator[Unknown]`

### Expected behavior

It should print `Iterator[A]` where `A` is the type parameter name of `Iterator`.

---

## [x] should accept `PresentaitonCompilerConfig#defaultSymbolPrefixes`
https://github.com/scalameta/metals/pull/4037
See: CompletionOverrideSuite#mutable / jutil

---

## [x] should accept rename config to `MetalsPrinter`
https://github.com/scalameta/metals/pull/4037

so we can (for example) shorten `import java.util` to `import java.{util as ju}`

---

## [x] Remove TermRef prefix from signature when OverrideCompletion completes an underlying class
see https://github.com/scalameta/metals/pull/3897#discussion_r871513361

see: `CompletionOverrideSuite#path-dependent`

https://github.com/scalameta/metals/pull/4046

### Version of Metals

v0.11.6

PR #4037, opened 2022-06-16
feat: OverrideCompletion can handle symbolPrefix

tanishiking:
https://github.com/scalameta/metals/issues/3928

This PR enables OverrideCompletions and implement-all abstract members' code action for Scala3 to import using preset renames like `java.util => ju`.

```scala
abstract class JUtil {
  def foo: java.util.List[Int]
}
class Main extends JUtil {
  def foo@@
}
```

completes

```scala
import java.{util => ju}
abstract class JUtil {
  def foo: java.util.List[Int]
}
class Main extends JUtil {
  def foo: ju.List[Int] = ${0:???}
}
```

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
@@ -71,8 +71,6 @@ object AutoImports:
         Context
     ): Map[Symbol, String] =
       config.symbolPrefixes.asScala.flatMap { (from, to) =>
-        val fullName = from.stripSuffix("/").replace("/", ".")
-        // val pkg = requiredPackage(fullName)
         val pkg = SemanticdbSymbols.inverseSemanticdbSymbol(from)
         val rename = to.stripSuffix(".").stripSuffix("#")
         List(pkg, pkg.map(_.moduleClass)).flatten
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
@@ -30,7 +30,7 @@ object AutoImports:
 
     /**
      * Rename symbol owner and add renamed prefix to tpe symbol.
-     * For example, `Reanmed(sym = <java.util.Map>, ownerRename = "ju")` represents
+     * For example, `Renamed(sym = <java.util.Map>, ownerRename = "ju")` represents
      * complete `ju.Map` while auto import `import java.{util => ju}`.
      * `toEdits` method convert `Renamed` to
      * AutoImportEdits(
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
@@ -29,14 +29,27 @@ object AutoImports:
     case Simple(sym: Symbol)
 
     /**
-     * Rename symbol owner and add renamed prefix to tpe symbol
-     * `Map -> (ju.Map, import java.{util => ju})`
+     * Rename symbol owner and add renamed prefix to tpe symbol.
+     * For example, `Reanmed(sym = <java.util.Map>, ownerRename = "ju")` represents
+     * complete `ju.Map` while auto import `import java.{util => ju}`.
+     * `toEdits` method convert `Renamed` to
+     * AutoImportEdits(
+     *  nameEdit = "ju.Map",
+     *  importEdit = "import java.{util => ju})"
+     * )
      */
     case Renamed(sym: Symbol, ownerRename: String)
 
     /**
      * Rename symbol itself, import only.
-     * `Map -> ("", import java.{util => ju})`
+     * For example, `SelfRenamed(sym = <java.util>, name = "ju")` represents
+     * auto import `import java.{util => ju}` without completing anything, unlike `Renamed`.
+     *
+     * `toEdits` method convert `SelfRenamed` to
+     * AutoImportEdits(
+     *  nameEdit = None,
+     *  importEdit = "import java.{util => ju})"
+     * )
      */
     case SelfRenamed(sym: Symbol, name: String)
 
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
@@ -5,6 +5,7 @@ import scala.jdk.CollectionConverters.*
 
 import scala.meta.internal.mtags.MtagsEnrichments.*
 import scala.meta.internal.pc.AutoImports.AutoImportEdits
+import scala.meta.internal.pc.printer.ShortenedNames.ShortName
 import scala.meta.pc.PresentationCompilerConfig
 
 import dotty.tools.dotc.ast.tpd.*
@@ -34,6 +35,12 @@ object AutoImports:
     case Renamed(sym: Symbol, ownerRename: String)
 
     /**
+     * Rename symbol itself, import only.
+     * `Map -> ("", import java.{util => ju})`
+     */
+    case SelfRenamed(sym: Symbol, name: String)
+
+    /**
      *  Import owner and add prefix to tpe symbol
      * `Map -> (mutable.Map, import scala.collection.mutable)`
      */
@@ -46,6 +53,20 @@ object AutoImports:
     ): AutoImport =
       if sym.owner.showName == ownerRename then SpecifiedOwner(sym)
       else Renamed(sym, ownerRename)
+
+    def renameConfigMap(config: PresentationCompilerConfig)(using
+        Context
+    ): Map[Symbol, String] =
+      config.symbolPrefixes.asScala.flatMap { (from, to) =>
+        val fullName = from.stripSuffix("/").replace("/", ".")
+        // val pkg = requiredPackage(fullName)
+        val pkg = SemanticdbSymbols.inverseSemanticdbSymbol(from)
+        val rename = to.stripSuffix(".").stripSuffix("#")
+        List(pkg, pkg.map(_.moduleClass)).flatten
+          .filter(_ != NoSymbol)
+          .map((_, rename))
+      }.toMap
+  end AutoImport
 
   /**
    * Returns AutoImportsGenerator
@@ -67,15 +88,7 @@ object AutoImports:
     import indexedContext.ctx
 
     val importPos = autoImportPosition(pos, text, tree)
-    val renameConfig: Map[Symbol, String] =
-      config.symbolPrefixes.asScala.flatMap { (from, to) =>
-        val fullName = from.stripSuffix("/").replace("/", ".")
-        val pkg = requiredPackage(fullName)
-        val rename = to.stripSuffix(".").stripSuffix("#")
-        List(pkg, pkg.moduleClass)
-          .filter(_ != NoSymbol)
-          .map((_, rename))
-      }.toMap
+    val renameConfig: Map[Symbol, String] = AutoImport.renameConfigMap(config)
 
     val renames =
       (sym: Symbol) =>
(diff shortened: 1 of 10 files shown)


tanishiking:
I could filter out by checking `span` exists

tanishiking:
> vadim chelyhov — 今日 19:54
Which test fails?

> there is a thing that every operation in PC enters symbols from passed sources and they will be available until the restart.
in cross tests that  means that if you run two tests:

```scala
  // enters example.A
  check(
    "first",
    """package example
       class A
    """
  )
  // enters example.B
  // but if look at example.members you should be able to find example.A too
  check(
    "first",
    """package example
       class B
    """
  )
```

https://discord.com/channels/632642981228314653/718047292749381662/986946861061533706

---

@dos65 Thank you for answering the question! The failing tests are `OverrideCompletionSuite`'s `jutil` and `jutil-conflict`
https://github.com/scalameta/metals/blob/1db3cf9dc3f1e2388eb5572f9de5e70fd4cf245f/tests/cross/src/test/scala/tests/pc/CompletionOverrideSuite.scala#L400-L464

In `jtuil`, `import java.{util => ju}` will introduce `ju` into the root package. And then `jutil-conflict` will fail because it doesn't expect to see `ju` in the scope. I tried to change the package of `jtuil` test, but it doesn't work 🤔

tanishiking:
Ahh, right!
I switched from `PrettyType` to crafting fake types because `PrettyType` doesn't seem to work in some test cases, but `PrettyType` works better overall :+1: 

`PrettyType` doesn't work for example in https://github.com/scalameta/metals/pull/4037/files#diff-0e5f05439309c52c64903477e432b9ac53c154f396d0e9c77bfc121d5221c720R77 because `RefinedPrinter` doesn't take into account `PrettyType` when it appears as a type constructor of AppliedType.

- [AppliedType will be converted to text by `appliedText`](https://github.com/lampepfl/dotty/blob/140693d5ab028979a8803a19c0848eaed7bcd433/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala#L233-L235)
- [`tyconName` convert type constructor to Name](https://github.com/lampepfl/dotty/blob/140693d5ab028979a8803a19c0848eaed7bcd433/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala#L226)
- [`tyconName` uses typeSymbol.name](https://github.com/lampepfl/dotty/blob/140693d5ab028979a8803a19c0848eaed7bcd433/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala#L196)
- `PrettyType.typeSymbol` is always `NoSymbol` because https://github.com/lampepfl/dotty/blob/140693d5ab028979a8803a19c0848eaed7bcd433/compiler/src/dotty/tools/dotc/core/Types.scala#L484-L491

We can't change the behavior of the printer, because `tyconNmae` is defined inside of the `toText` function, and `typeSymbol` is final def. Maybe we need to implement our own pretty printer for Scala3 types 🤔

tanishiking:
It's a kind of monkey patch, but I could somehow override `toText` and fix the problem :+1: Thank you so much for your input @dos65

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -47,6 +54,20 @@ object AutoImports:
       if sym.owner.showName == ownerRename then SpecifiedOwner(sym)
       else Renamed(sym, ownerRename)
 
+    def renameConfigMap(config: PresentationCompilerConfig)(using
+        Context
+    ): Map[Symbol, String] =
+      config.symbolPrefixes.asScala.flatMap { (from, to) =>
+        val fullName = from.stripSuffix("/").replace("/", ".")
+        // val pkg = requiredPackage(fullName)
+        val pkg = SemanticdbSymbols.inverseSemanticdbSymbol(from)
```
tanishiking, line 63: Using `inverseSemanticdbSymbol` as I found `requiredPackage` doesn't create proper symbols.

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -128,30 +141,44 @@ object AutoImports:
...
+     */
+    def forShortName(shortName: ShortName): Option[List[l.TextEdit]] =
+      if shortName.isRename then Some(toEdits(shortName.asImport).edits)
+      else forSymbol(shortName.symbol)
+
     /**
      * @param symbol A missing symbol to auto-import
      */
     def editsForSymbol(symbol: Symbol): Option[AutoImportEdits] =
-      inferAutoImport(symbol).map { ai =>
```
tanishiking, line 135: Just separate the logic into `toEdits` method.

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -33,6 +34,12 @@ object AutoImports:
      */
     case Renamed(sym: Symbol, ownerRename: String)
 
+    /**
+     * Rename symbol itself, import only.
+     * `Map -> ("", import java.{util => ju})`
+     */
+    case SelfRenamed(sym: Symbol, name: String)
```
tgodzik, line 41: Why do we need a separate case to `Renamed`?

mtags/src/main/scala-3/scala/meta/internal/pc/printer/DotcPrinter.scala
```diff
@@ -41,6 +42,22 @@ object DotcPrinter:
...
+      // `toText` in RefinedPrinter doesn't pretty print AppliedType
+      // if tycon is instance of PrettyType.
+      // For example, if we don't override `toText`,
+      // CompletionoverrideConfigSuite's `package` test will fail with
+      // completing `def function: Int <none> String = ${0:???}`
+      // instead of `def function: f.Function[Int, String] = ${0:???}`
+      tp match
+        case tp: AppliedType =>
+          tp.tycon match
+            case p: PrettyType =>
```
tgodzik, line 57: Overriding it only for PrettyType wouldn't work?

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
tanishiking: Because `Renamed(<java.util.Optional>, "ju")` is for when we complete `ju.Optional` auto importing `java.{util => ju}`. On the other hand, `SelfRenamed(<java.util>, "ju")` represents, import `java.{util => ju}` without completing anything.

`Renamed` and `SelfRenamed` expect to have different symbols, maybe we should add a comment explaining this.

mtags/src/main/scala-3/scala/meta/internal/pc/printer/DotcPrinter.scala
tanishiking: No, if we have `AppliedType(PrettyType("Foo"), Seq(Int, String))`, `DocPrinter` will handle `AppliedType` using [RefinedPrinter's toText](https://github.com/lampepfl/dotty/blob/9e3953d53c815a61f85311e4863f0431690063d8/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala#L140-L300) and it never call this method for `PrettyType` recursively, we will end up having `Int <none> String` instead of `Foo[Int, String]`.

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -33,6 +34,12 @@ object AutoImports:
      */
     case Renamed(sym: Symbol, ownerRename: String)
 
+    /**
+     * Rename symbol itself, import only.
+     * `Map -> ("", import java.{util => ju})`
```
dos65, line 39:
```suggestion
     * `Map -> ("ju.Map", no-import)`
```

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -183,6 +210,8 @@ object AutoImports:
           case AutoImport.SpecifiedOwner(sym) => importName(sym.owner)
           case AutoImport.Renamed(sym, rename) =>
             s"${importName(sym.owner.owner)}.{${sym.owner.nameBackticked} => $rename}"
+          case AutoImport.SelfRenamed(sym, rename) =>
+            s"${importName(sym.owner)}.{${sym.nameBackticked} => $rename}"
```
dos65, line 214: Shouldn't it be unreachable?

dos65 (approved):
LGTM

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
tanishiking: No, `SelfRenamed` is for `import` only, and it won't complete anything (like `ju.Map`)

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
tanishiking: I don't think so, `SelfRenamed` will be caught here https://github.com/tanishiking/metals/blob/9cef97a78e0811aa7ff217a3d5ff1d8bbdd44f79/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala#L178-L179 and then we will hit here in `importEdit`

tests/cross/src/test/scala/tests/pc/CompletionIssueSuite.scala
```diff
@@ -272,11 +272,11 @@ class CompletionIssueSuite extends BaseCompletionSuite {
     filter = (str) => str.contains("createProgress"),
     compat = Map(
       "3" -> """|import org.eclipse.lsp4j.services.LanguageClient
-                |import java.util.concurrent.CompletableFuture
                 |import org.eclipse.lsp4j.WorkDoneProgressCreateParams
+                |import java.{util => ju}
```
tgodzik, line 276: Interesting, should we make it work the same for Scala 2 ?

tgodzik (approved):
LGTM!

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -28,11 +29,30 @@ object AutoImports:
     case Simple(sym: Symbol)
 
     /**
-     * Rename symbol owner and add renamed prefix to tpe symbol
-     * `Map -> (ju.Map, import java.{util => ju})`
+     * Rename symbol owner and add renamed prefix to tpe symbol.
+     * For example, `Reanmed(sym = <java.util.Map>, ownerRename = "ju")` represents
```
dos65, line 33:
```suggestion
     * For example, `Renamed(sym = <java.util.Map>, ownerRename = "ju")` represents
```

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
dos65: oh. now I see that SelfRenamed is a bit different from what I expected))
It's kind of auto import for `import` :thinking:

tests/cross/src/test/scala/tests/pc/CompletionIssueSuite.scala
tanishiking: fixed https://github.com/scalameta/metals/pull/4037/commits/68892369439a232ba9e6874b0a5a8a70a19d2896
I realized Scala2 renames only if the direct owner symbol (of the symbol we're completing) is the `java.util`.

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
tanishiking: :+1: https://github.com/scalameta/metals/pull/4037/commits/68892369439a232ba9e6874b0a5a8a70a19d2896

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -47,6 +54,20 @@ object AutoImports:
       if sym.owner.showName == ownerRename then SpecifiedOwner(sym)
       else Renamed(sym, ownerRename)
 
+    def renameConfigMap(config: PresentationCompilerConfig)(using
+        Context
+    ): Map[Symbol, String] =
+      config.symbolPrefixes.asScala.flatMap { (from, to) =>
+        val fullName = from.stripSuffix("/").replace("/", ".")
```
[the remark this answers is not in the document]
tanishiking: :+1: https://github.com/scalameta/metals/pull/4037/commits/152944f028639fb67be9f33300fe5bf0c3dfdec6

tgodzik (approved):
Ship it! :ship:

PR #4038, opened 2022-06-17
fix(worksheets): filter out NonUnitStatements wartremover in worksheets

ckipp01:
I also hit on this during work and it annoys me. If you have a build
that is using wartremover and you are using the default set of
`Warts.unsafe` it includes the `NonUnitStatements` rule. When this rule
is part of your `scalacOptions` when you attempt to create a worksheet
next to your sources it will never work due to the binder in mdoc.
You'll get the following error at the top of your file and nothing will
evaluate:

```scala
val x = 3 // example piece of code in your worksheet
```
```
Diagnostics:
scala.worksheet.sc:9 (mdoc generated code) [wartremover:NonUnitStatements] Statements must return Unit
val x = 3; $doc.binder(x, 1, 4, 1, 5)
```

This change just strips out that rule so it doesn't get passed in during
evaluation so that your worksheet can actually evaluate.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
@@ -434,9 +434,13 @@ class WorksheetProvider(
         }
         if isSupported
       } yield {
+        // We filter out NonUnitStatements from wartremover or you'll get an
+        // error about $doc.binder returning Unit from mdoc, which causes the
+        // worksheet not to be evaluated.
         val scalacOptions = info.scalac.getOptions.asScala
           .filterNot(_.contains("semanticdb"))
           .filterNot(_.contains("-Wconf"))
+          .filterNot(_.contains("org.wartremover.warts.NonUnitStatements"))
           .asJava
         val mdoc = embedded
           .mdoc(


tanishiking (approved):
LGTM!

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
```diff
@@ -434,9 +434,13 @@ class WorksheetProvider(
         }
         if isSupported
       } yield {
+        // We filter out NonUnitStatements from wartremover or you'll get an
+        // error about $doc.binder returning Unit from mdoc, which causes the
+        // worksheet not to be evaluated.
         val scalacOptions = info.scalac.getOptions.asScala
           .filterNot(_.contains("semanticdb"))
           .filterNot(_.contains("-Wconf"))
+          .filterNot(_.contains("org.wartremover.warts.NonUnitStatements"))
```
dos65, line 443: maybe drop all that contains `wartremover` ?

dos65 (approved):
LGTM

metals/src/main/scala/scala/meta/internal/worksheets/WorksheetProvider.scala
ckipp01: I was thinking that, but I thought this would maybe cause the least amount of confusions. I'm still not 100% sure tbh.

dos65: I don't know. I took a quick look to check how options is passed for wartremover.
I'm not sure, but in documentation it looks like a single option where rules are separated with `:`. Am I right?
In this case it should be equal to removing all wartremover configuration if `NonUnitStatements` is enabled.

ckipp01: No, it ends up being split up in the `scalacOptions` like this:

```
-Xplugin:target/compiler_plugins/wartremover_2.13.8-3.0.4.jar, -P:wartremover:traverser:org.wartremover.warts.Any, -P:wartremover:traverser:org.wartremover.warts.AsInstanceOf, -P:wartremover:traverser:org.wartremover.warts.DefaultArguments, -P:wartremover:traverser:org.wartremover.warts.EitherProjectionPartial, -P:wartremover:traverser:org.wartremover.warts.IsInstanceOf, -P:wartremover:traverser:org.wartremover.warts.IterableOps, -P:wartremover:traverser:org.wartremover.warts.NonUnitStatements, -P:wartremover:traverser:org.wartremover.warts.Null, -P:wartremover:traverser:org.wartremover.warts.OptionPartial, -P:wartremover:traverser:org.wartremover.warts.Product, -P:wartremover:traverser:org.wartremover.warts.Return, -P:wartremover:traverser:org.wartremover.warts.Serializable, -P:wartremover:traverser:org.wartremover.warts.StringPlusAny, -P:wartremover:traverser:org.wartremover.warts.Throw, -P:wartremover:traverser:org.wartremover.warts.TryPartial, -P:wartremover:traverser:org.wartremover.warts.Var```
So every rule is passed in as a flag.

dos65: ok, thanks for explanation!
sorry got the wrong assumption

PR #4039, opened 2022-06-17
feat(mill): also check default version in mill file

ckipp01:
I just hit on this with https://github.com/com-lihaoyi/requests-scala
since they have a `mill` file which contains the default mill version
being used, but we only check for the version in the `.mill-version`
file. While most repos will have a `.mill-version` this adds an extra
fallback to check in the `mill` file to also grab it before we use the
default one we have. It can be the case that they are using an older
version of `mill` and our default version won't work to do the build
import. This guards against that.


ckipp01:
Actually hold on, I realized in the new `millw` script it has `DEFAULT_MILL_VERSION` in the comment so that won't work. I need to fix that.

tanishiking (approved):
I confirmed it works with the `mill` script that specifies the version with `DEFAULT_MILL_VERSION` :+1: LGTM!

Issue #4027, opened 2022-06-14
Support tab-indentation for implement-all abstract members code action


tanishiking:
### Describe the bug

follow-up from https://github.com/scalameta/metals/pull/3960#discussion_r895719962

```scala
trait Foo:
  def foo(x: Int): String
  def bar(x: Int): Option[String]

class FooImpl extends Foo:
	override def bar(x: Int): Option[String] = None // tab indent
```

run `implement members` code action on `FooImpl`, we get

```scala
class FooImple extends Foo:

override def bar: Int = ???

	override def bar(x: Int): Option[String] = None // tab indent
```

### Expected behavior

`implement-all` code action should work with tab indent

```scala
class FooImple extends Foo:

	override def bar: Int = ???

	override def bar(x: Int): Option[String] = None // tab indent
```

These code examples are written in Scala3, but Scala2 should have the same problem.

Maybe we can fix this following: https://github.com/scalameta/metals/pull/3702

### Operating system

Any OSes

### Editor/Extension

Any editors

### Version of Metals

v0.11.6

### Extra context or search terms

implement-all / code action

PR #4042, opened 2022-06-17
fix: support tab indent in implement-all code action

tanishiking:
fix https://github.com/scalameta/metals/issues/4027

Before this commit, the code action always add indent with spaces, now
Metals support tab indent in implement-all code action.

If the class / object to implement is indented with tabs,
Metals will use tabs for indents instead of spaces.

As a limitation, Metals cannot infer the indent from the whole file:
for example

```scala
package foo

abstract class Foo:
\tdef foo: Int

class FooImpl extends Foo
```

we should implement `FooImpl` with tabs, but we can't
because Metals infers the indent from the line of `FooImpl`.
In this case, there's no indent, and fallback to space indent.


tgodzik (approved):
LGTM!

tgodzik:
@tanishiking you just need to run scalafix on the PR

Issue #4043, opened 2022-06-17
Override completion for default java method includes <defaultMethod>/


ckipp01:
### Describe the bug

When trying to completed a java method that is a `default` it seems that it actually includes the `<defaultMethod>` inside of the `newText` for the completion.

For example, go to the `DelegatingLanguageClient.scala` in the metals codebase and attempt to complete `def createProgress`. Accept the override completion and you'll end up with:

````scala
override <defaultmethod> def createProgress(params: WorkDoneProgressCreateParams): CompletableFuture[Void] = ???
```


### Expected behavior

The `<defaultmethod>` shouldn't be there.

### Operating system

macOS

### Editor/Extension

Nvim (nvim-metals)

### Version of Metals

0.11.6+105-859dac6c-SNAPSHOT

### Extra context or search terms

_No response_

PR #4045, opened 2022-06-20
bugfix: Fix wrong flag for default methods

tgodzik:
Fixes https://github.com/scalameta/metals/issues/4043

I also unignored the CompletionIssueSuite for Scala 3 and adjusted to the current output. I will work on the ignored ones a bit later.

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
@@ -2,6 +2,7 @@ package scala.meta.internal.pc.completions
 
 import scala.collection.immutable.Nil
 import scala.collection.mutable
+import scala.reflect.internal.Flags
 
 import scala.meta.internal.pc.AutoImportPosition
 import scala.meta.internal.pc.CompletionFuzzy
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/completions/OverrideCompletions.scala
```diff
@@ -234,11 +235,15 @@ trait OverrideCompletions { this: MetalsGlobal =>
...
-        sym.flagString.replace(
-          sym.privateWithin.toString(),
-          sym.privateWithin.name.toString()
-        )
+        sym
+          .flagString(mask)
+          .replace(
+            sym.privateWithin.toString(),
+            sym.privateWithin.name.toString()
+          )
```
dos65, line 246: There is no more need in this `replace`

tgodzik: Och, how did you find that?

tgodzik: It seems that the tests fail if I remove the replace since private within is doing `package a` on `toString`:
`override protected[package a] def f(): Unit = ???`

dos65: oops, it's a different thing))
I thought that it was a prev approach to same thing and run only CompletionIssueSuite

Issue #4034, opened 2022-06-15
Imprecise return type of expression


prolativ:
### Describe the bug

```scala
//> using scala "3.1.2"

trait A
object A1 extends A

trait Foo:
  type Out
  def out: Out

def fooOut(f: Foo): f.Out = f.out

object FooA1 extends Foo:
  type Out = A1.type
  def out = A1

val x = fooOut(FooA1)
```
The expressions type of selection `fooOut(FooA1)` is shown to be `Out`

### Expected behavior

`Out` should be replaced with the actual type of the expression, which is `A1.type` in this case

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.6+94-4fdaace3-SNAPSHOT

PR #4046, opened 2022-06-20
bugfix: Fix wrong type shown when type alias is used

tgodzik:
Aside from dealiasing it turns out we never shortened ThisType, which also showed that the Dotty pritner would never properly print lang packages even if needed.

Fixes https://github.com/scalameta/metals/issues/4034

Related to https://github.com/scalameta/metals/pull/4037 but it seems that we tackle complementary bugs?


mtags/src/main/scala-3/scala/meta/internal/pc/printer/ShortenedNames.scala
```diff
@@ -136,7 +136,7 @@ class ShortenedNames(val indexedContext: IndexedContext):
               designator.asInstanceOf[Symbol]
             else tpe.termSymbol
           val short = ShortName(sym)
-          if tryShortenName(name) then NoPrefix
+          if tryShortenName(short) then TermRef(NoPrefix, designator)
```
tanishiking, line 139:
```suggestion
          if tryShortenName(short) then TermRef(NoPrefix, sym)
```

Can we do this?
if `designator` is not a Symbol, we will fail at runtime because of this assertion https://github.com/lampepfl/dotty/blob/75d8eea1e943bbd2c605d8411c2d52d69974d6f3/compiler/src/dotty/tools/dotc/core/Types.scala#L2733

tests/cross/src/test/scala/tests/pc/CompletionOverrideSuite.scala
```diff
@@ -1103,20 +1081,7 @@ class CompletionOverrideSuite extends BaseCompletionSuite {
...
-           |  object Outer {
-           |    class Inner
-           |  }
-           |  def inner: Outer.Inner
-           |}
-           |class Under extends Over {
-           |  def inner: Under#Outer.Inner = ${0:???}
-           |}
-           |""".stripMargin
-    )
```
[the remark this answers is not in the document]
tanishiking: Is this fixed because of the change in `toTextPrefix`? or change in `tryShortenName` ?

mtags/src/main/scala-3/scala/meta/internal/pc/printer/DotcPrinter.scala
```diff
@@ -39,13 +39,33 @@ object DotcPrinter:
     def tpe(t: Type): String =
       toText(t).mkString(defaultWidth, false)
 
+    /* Overriden method because the RefinedPrinter will omit
+     * `lang` even if it's needed in case of conflicts
+     */
```
tanishiking, line 44: Can you add a link to https://github.com/lampepfl/dotty/blob/75d8eea1e943bbd2c605d8411c2d52d69974d6f3/compiler/src/dotty/tools/dotc/printing/RefinedPrinter.scala#L107-L123 so we can

- tell which revision of `toTextPrefix` we borrowed
- what is the difference from the original implementation

tanishiking (changes requested):
Good catch!
I left some minor comments, but otherwise LGTM :)

tests/cross/src/test/scala/tests/pc/CompletionOverrideSuite.scala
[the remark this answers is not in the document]
tgodzik: it's because of the change in tryShortenName, but it didn't yet fully work for all the cases, so I needed to update `toTextPrefix`

mtags/src/main/scala-3/scala/meta/internal/pc/printer/ShortenedNames.scala
tgodzik: Changed! That works correctly, thanks!

mtags/src/main/scala-3/scala/meta/internal/pc/printer/DotcPrinter.scala
```diff
@@ -39,13 +39,33 @@ object DotcPrinter:
...
+      tp match
+        case tp @ TermRef(pre, _) =>
+          val sym = tp.symbol
+          if sym.isPackageObject && !homogenizedView && !printDebug then
+            toTextPrefix(pre)
+          else printPrefix(tp)
+        case _ => super.toTextPrefix(tp)
+    }
+
+    /* Taken from PlainPrinter in order not to use super `toTextPrefix` method
```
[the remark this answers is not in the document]
tgodzik: Done!

mtags/src/main/scala-3/scala/meta/internal/pc/printer/DotcPrinter.scala
tgodzik: Done!

tanishiking (approved):
LGTM!

PR #4047, opened 2022-06-20
fix: remove improper javaOptions initialization

dos65:
All options were concatenated into a single string that might lead to
an incorrect values in important properties (`user.dir`, ...)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -60,16 +60,6 @@ inThisBuild(
     // faster publishLocal:
     packageDoc / publishArtifact := sys.env.contains("CI"),
     packageSrc / publishArtifact := sys.env.contains("CI"),
-    // forking options
-    javaOptions += {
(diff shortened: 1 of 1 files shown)


kpodsiad:
Did you encounter such error? I also wonder why this code was needed before.

tgodzik (approved):
Looks good! I don't think this is needed anymore.
It was introduced in https://github.com/scalameta/metals/commit/f847ba66f54722270c720c6e0b980b9e42b36137
as a way of forwarding java options to forked tests maybe? I don't think it's needed.

dos65:
@kpodsiad yep, it breaks a lot if it starts from user.dir : `-Duser.dir=$dir -Danother.option=...`.  Then it's processed as single option flag and `sys.props("user.dir")` is equal to: `$dir -Danother.option=...`

PR #4049, opened 2022-06-21
feature: Don't update readonly bloop.json and allow to dismiss notifications

tgodzik:
Previously, metals would suggest to update bloop.json file even if it was readonly and there would be no way to dismiss the notification forever. Now, we will not suggest to update the file if it's readonly and users can ignore the notification if they want to. It's also possible to later reset all the notification so that everything can pop up again.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
@@ -195,6 +195,10 @@ final class BloopServers(
         )
         Future.successful(languageClient.metalsExecuteClientCommand(command))
 
+      case (item, _)
+          if item == Messages.BloopGlobalJsonFilePremodified.useGlobalFile =>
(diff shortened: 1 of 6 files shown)


dos65 (approved):
:+1:

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -195,6 +195,10 @@ final class BloopServers(
         )
         Future.successful(languageClient.metalsExecuteClientCommand(command))
 
+      case (item, _)
+          if item == Messages.BloopGlobalJsonFilePremodified.useGlobalFile =>
```
zmerr, line 199:
```scala
case (Messages.BloopGlobalJsonFilePremodified.useGlobalFile, _) =>
```

tgodzik: It would need to be a val unfortunately, so this doesn't work.

Issue #4052, opened 2022-06-15
Adding stub implementations for abstract given instances


prolativ:
### Is your feature request related to a problem? Please describe.

Given the following scala 3 snippet
```scala
trait Foo:
  def bar: Int
  def baz: String
  
given Foo
```

### Describe the solution you'd like

`given Foo` should provide a code action for implementing its missing members resulting in
```scala
trait Foo:
  def bar: Int
  def baz: String
  
given Foo with
  def bar: Int = ???
  def baz: String = ???
```

### Describe alternatives you've considered

Currently one has to write
```scala
given Foo with {}
```
for the code action to be invocable

### Search terms

Dotty

PR #4055, opened 2022-06-22
feat: Adding stub implementations for abstract given instances 

tanishiking:
fix: https://github.com/scalameta/metals/issues/4052

Now, implement all abstract members code action is
available for given instance:
When we have

```scala
trait Foo:
  def bar: Int
  def baz: String

given Foo
```

the code action will implement the given instance like

```scala

given Foo with

  override def bar: Int = ???

  override def baz: String = ???
```


tgodzik:
@tanishiking is this ready to review actually?

tanishiking:
Yes, it's ready for review :)

dos65 (approved):
LGTM!

tgodzik (approved):
LGTM! You just need to rebase on master

Issue #3953, opened 2021-01-21
[Scala 3] Add interpolator completions 


tgodzik:
**Is your feature request related to a problem? Please describe.**
In Scala 2 it's possible to have proper completions on trying to complete inside of an interpolated and non interpolated strings. It would be good to have the same for Scala 3

**Describe the solution you'd like**
The potential solutions should take a look at `InterpolatorCompletions.scala` and related tests to see what needs to be supported but in general we would need:
- add interpolator when completing `$` + value inside a normal string
- wrap `{}` when needed
- wrap any `$` that already exist in the string

**Describe alternatives you've considered**

Write it manually.

**Additional context**
This is part of the effort to create proper tooling for Scala 3.

**Search terms**:
inteprolator completions scala3

tgodzik:
Looking into this one!

PR #4061, opened 2022-06-23
feature: [Scala 3] Add interpolation completions

tgodzik:
Previously, we Scala 3 didn't have interpolation completions. Now, we copied it over and adapted from Scala 2.

There are two edge cases that I need to follow up on, but it's still some good progress.

Fixes https://github.com/scalameta/metals/issues/3953


mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
```diff
@@ -15,18 +15,22 @@ import dotty.tools.dotc.util.ParsedComment
...
 import org.eclipse.lsp4j.CompletionItemTag
 import org.eclipse.lsp4j.Range
+import org.eclipse.lsp4j.TextEdit
 
 sealed trait CompletionValue:
   def label: String
+  def insertText: Option[String] = None
   def snippetSuffix: Option[String] = None
+  def additionalEdits: List[TextEdit] = Nil
+  def range: Option[Range] = None
```
tanishiking, line 25: I'm fine with adding these fields to the abstract `CompletionValue`. What do you think about this? @dos65 

Kinda off-topic: I'm wondering maybe we should always pre-compute the `additionalEdits` before creating `CompletionValue`s, because while `CompletionValue.Workspace` generate auto imports (as `additionalEdits` in `CompletionsProvider` after `CompletionValue`, some completions pre-compute auto imports before `CompletionValue`.

As a result, I am always confused about when and where should I generate the `additionalEdits`, and maybe it makes it harder to comprehend the program.

tanishiking (commented):
I left some comments, but overall it looks good to me 👍

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -15,6 +16,7 @@ import scala.meta.pc.*
 
 import dotty.tools.dotc.ast.tpd.*
 import dotty.tools.dotc.ast.untpd
+import dotty.tools.dotc.core.Constants
```
kpodsiad, line 19: Are `scala.collection.immutable.LazyList.cons` and `dotty.tools.dotc.core.Constants` needed?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -237,7 +239,18 @@ class Completions(
           if Fuzzy.matches(td.symbol.name.decoded, filename) =>
         val values = FilenameCompletions.contribute(filename, td)
         (values, true)
-
+      case (lit @ Literal(Constant(str: String))) :: parent :: _ =>
```
kpodsiad, line 242: Could you add comment with example what will be matched by this case?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/InterpolatorCompletions.scala
```diff
@@ -0,0 +1,251 @@
...
+    then
+      if cursor < 0 then None
+      else if s.charAt(1).isUnicodeIdentifierStart then
+        val isValidIdentifier =
+          2.until(cursor).forall(i => s.charAt(i).isUnicodeIdentifierPart)
+        if isValidIdentifier then Some(s.substring(1, cursor))
+        else None
+      else Some("")
+    else None
+  end interpolatorMemberSelect
```
dos65, line 73: Hm. I'm wondering if it might replaced by `CompletionPos.query`. 
What the query is in case of interpolator completions?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
dos65: I'm also fine with adding them at least for a while.
It feels like the completion processing is just happens in different places. For interpolater everything is calculated while others are processed at the moment of convertation them into lsp structure.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/InterpolatorCompletions.scala
tgodzik: it's the name after `.` when we don't have a braces

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
tgodzik: It might be hard with the more custom completions to compute everything later on :thinking:

mtags/src/main/scala/scala/meta/internal/pc/InterpolationSplice.scala
```diff
@@ -0,0 +1,51 @@
+package scala.meta.internal.pc
+
+case class InterpolationSplice(dollar: Int, name: String, needsBraces: Boolean)
+
+object InterpolationSplice {
+
+  def apply(
+      offset: Int,
+      chars: Array[Char],
+      originalText: String
```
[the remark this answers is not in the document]
tgodzik: chars might contain `CURSOR` while originalText doesn't. That's why I renamed it.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/InterpolatorCompletions.scala
tgodzik: Actually `CompletionPos.query` works! Changed!

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: Added!

mtags/src/main/scala-3/scala/meta/internal/pc/completions/InterpolatorCompletions.scala
```diff
@@ -0,0 +1,251 @@
...
+  ): String =
+    suffix match
+      case Some(suffix) if suffix == "()" =>
+        suffix + "$0"
+      case Some(suffix) => suffix
+      case None if areSnippetsSupported => "$0"
+      case _ => ""
+
+  /**
+   * A provides completions to convert a string literal into a string literal,
```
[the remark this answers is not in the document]
tgodzik: Changed the comment a bit, I mostly copy pasted it, which might have been a bad choice.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/InterpolatorCompletions.scala
```diff
@@ -0,0 +1,251 @@
...
+    }.flatten
+    memberCompletions.flatten
+  end contributeMember
+
+  private def suffixEnding(
+      suffix: Option[String],
+      areSnippetsSupported: Boolean
+  ): String =
+    suffix match
+      case Some(suffix) if suffix == "()" =>
```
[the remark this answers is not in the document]
tgodzik: Good catch, added!

mtags/src/main/scala/scala/meta/internal/pc/InterpolationSplice.scala
```diff
@@ -0,0 +1,51 @@
+package scala.meta.internal.pc
+
+case class InterpolationSplice(dollar: Int, name: String, needsBraces: Boolean)
```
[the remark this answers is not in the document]
tgodzik: Added!

tanishiking (approved):
LGTM!

tgodzik:
I will hold off until the release with this merge.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/InterpolatorCompletions.scala
```diff
@@ -0,0 +1,249 @@
...
+          label,
+          (name, sym, suffix) =>
+            CompletionValue.Interpolator(
+              sym,
+              label,
+              Some(newText(name, suffix, identOrSelect)),
+              Nil,
+              Some(cursor.withStart(identOrSelect.span.start).toLSP),
+              // Needed for VS Code which will not show the completion otherwise
+              Some(identOrSelect.show + "." + label)
```
tgodzik, line 148: I changed one controversial thing here, which is needed for VS Code to show the completion. Otherwise it will skip it unfortunately.

PR #4067, opened 2022-06-26
java version through shell

### LINKED ISSUE #3536: JavaInteractiveSemanticDb - jdk version inferring issue
### Describe the bug

It was reported [here](https://github.com/scalameta/metals/issues/3500#issuecomment-1013722332).

Currently, we detect jdk version by reading `$JVM_HOME/release` and `version` field value from it.
We expected that it should works for Jdk11+ version but it seems there is no guarantee that `release` file will be presented in jdk-distributive.

As a solution if there is no `release` file we can try to obtain it from the running `java --version`.

### Expected behavior

_No response_

### Operating system

_No response_

### Editor/Extension

_No response_

### Version of Metals

v0.11.1

_No response_

zmerr:
resolves #3536

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
@@ -1,5 +1,7 @@
 package scala.meta.internal.builds
 
+import java.nio.charset.StandardCharsets
+
 import scala.concurrent.Future
 import scala.concurrent.Promise
 import scala.util.Properties
@@ -30,6 +32,7 @@ class ShellRunner(
 ) extends Cancelable {
 
   private val cancelables = new MutableCancelable()
+
   override def cancel(): Unit = {
     cancelables.cancel()
   }
@@ -129,4 +132,82 @@ class ShellRunner(
     result.future
   }
 
+}
+
+object ShellRunner {
+
+  /**
+   * Note: if returning a [[Future]] is more suitable, all you
+   * need to do is to uncomment the commented code.
+   */
+  def quickRunWithShortStringResult(
+      //        commandRun: String,
+      args: List[String],
+      directory: AbsolutePath,
+      redirectErrorOutput: Boolean,
+      additionalEnv: Map[String, String] = Map.empty,
+      processOut: String => Unit = scribe.info(_),
+      processErr: String => Unit = scribe.error(_),
+      propagateError: Boolean = false,
+      maybeJavaHome: Option[String]
+
+      //       logInfo: Boolean = true
+  ):
+  // Future[
+  Option[String]
+  //  ]
+  = {
+    //   val elapsed = new Timer(time)
+
+    val env = additionalEnv ++ maybeJavaHome.map("JAVA_HOME" -> _).toMap
+    val ps = SystemProcess.run(
+      args,
+      directory,
+      redirectErrorOutput,
+      env,
+      Some(processOut),
+      Some(processErr),
+      propagateError
+    )
+    //    // NOTE(olafur): older versions of VS Code don't respect cancellation of
+    //    // window/showMessageRequest, meaning the "cancel build import" button
+    //    // stays forever in view even after successful build import. In newer
+    //    // VS Code versions the message is hidden after a delay.
+    //    val taskResponse =
+    //      languageClient.metalsSlowTask(
+    //        new MetalsSlowTaskParams(commandRun)
+    //      )
+
+    //    val result = Promise[Option[String]]
+    //    taskResponse.asScala.foreach { item =>
+    //      if (item.cancel) {
+    //        if (logInfo)
+    //          scribe.info(s"user cancelled $commandRun")
+    //        result.trySuccess(None)
+    //        ps.cancel
+    //      }
+    //    }
+
+    //   val processFuture = ps.complete
+
+    //    cancelables
+    //      .add(() => ps.cancel)
+    //      .add(() => taskResponse.cancel(false))
+
+    //    processFuture.map { code =>
+    //      taskResponse.cancel(false)
+    //      if (logInfo)
+    //        scribe.info(s"time: ran '$commandRun' in $elapsed")
+    //     result.trySuccess(code match {
+    //       case ExitCodes.Success =>
+    util
+      .Try(
+        new String(ps.inputStream.readAllBytes(), StandardCharsets.UTF_8)
+      )
+      .toOption
+    //        case _ => None
+    //      })
+    //    }
+    //    result.future
+  }
 }
--- a/metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
@@ -1,9 +1,10 @@
 package scala.meta.internal.builds
 
-import java.util.Scanner
-
+import scala.concurrent.Await
 import scala.concurrent.Future
 import scala.concurrent.Promise
+import scala.concurrent.duration.DurationInt
+import scala.language.postfixOps
 import scala.util.Properties
 
 import scala.meta.internal.metals.Cancelable
@@ -141,26 +142,30 @@ object ShellRunner {
       directory: AbsolutePath,
       redirectErrorOutput: Boolean,
       additionalEnv: Map[String, String] = Map.empty,
-      processOut: String => Unit = scribe.info(_),
       processErr: String => Unit = scribe.error(_),
       propagateError: Boolean = false,
       maybeJavaHome: Option[String]
   ): Option[String] = {
+
+    val sbOut = new StringBuilder()
     val env = additionalEnv ++ maybeJavaHome.map("JAVA_HOME" -> _).toMap
     val ps = SystemProcess.run(
       args,
       directory,
       redirectErrorOutput,
       env,
-      Some(processOut),
+      Some(s => {
+        sbOut.append(s)
+        sbOut.append(Properties.lineSeparator)
+      }),
       Some(processErr),
       propagateError
     )
 
-    val scanner = new Scanner(ps.inputStream).useDelimiter("\\A")
-    if (scanner.hasNext)
-      Some(scanner.next)
-    else None
+    val exit = Await.result(ps.complete, 10 second)
 
+    if (exit == 0) {
+      Some(sbOut.toString())
+    } else None
   }
 }
--- a/metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
@@ -1,6 +1,6 @@
 package scala.meta.internal.builds
 
-import java.nio.charset.StandardCharsets
+import java.util.Scanner
 
 import scala.concurrent.Future
 import scala.concurrent.Promise
@@ -200,11 +200,10 @@ object ShellRunner {
     //        scribe.info(s"time: ran '$commandRun' in $elapsed")
     //     result.trySuccess(code match {
     //       case ExitCodes.Success =>
-    util
-      .Try(
-        new String(ps.inputStream.readAllBytes(), StandardCharsets.UTF_8)
-      )
-      .toOption
+
+    val scanner = new Scanner(ps.inputStream).useDelimiter("\\A")
+    if (scanner.hasNext) Some(scanner.next) else None
+
     //        case _ => None
     //      })
     //    }
--- a/metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
@@ -136,12 +136,7 @@ class ShellRunner(
 
 object ShellRunner {
 
-  /**
-   * Note: if returning a [[Future]] is more suitable, all you
-   * need to do is to uncomment the commented code.
-   */
-  def quickRunWithShortStringResult(
-      //        commandRun: String,
+  def runSync(
       args: List[String],
       directory: AbsolutePath,
       redirectErrorOutput: Boolean,
@@ -150,15 +145,7 @@ object ShellRunner {
       processErr: String => Unit = scribe.error(_),
       propagateError: Boolean = false,
       maybeJavaHome: Option[String]
-
-      //       logInfo: Boolean = true
-  ):
-  // Future[
-  Option[String]
-  //  ]
-  = {
-    //   val elapsed = new Timer(time)
-
+  ): Option[String] = {
     val env = additionalEnv ++ maybeJavaHome.map("JAVA_HOME" -> _).toMap
     val ps = SystemProcess.run(
       args,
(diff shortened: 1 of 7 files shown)


metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
```diff
@@ -130,3 +133,80 @@ class ShellRunner(
...
+
+      //       logInfo: Boolean = true
+  ):
+  // Future[
+  Option[String]
+  //  ]
+  = {
+    //   val elapsed = new Timer(time)
+
+    val env = additionalEnv ++ maybeJavaHome.map("JAVA_HOME" -> _).toMap
```
tgodzik, line 162: IF we already have `"JAVA_HOME" ` we should use that. I don't think there is a need for additional env.

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
```diff
@@ -130,3 +133,80 @@ class ShellRunner(
...
 
 }
+
+object ShellRunner {
+
+  /**
+   * Note: if returning a [[Future]] is more suitable, all you
+   * need to do is to uncomment the commented code.
+   */
+  def quickRunWithShortStringResult(
```
tgodzik, line 143: Why not add it along the other methods? We could cache the Java home in that case inside the class.

```suggestion
  def runSync(
```

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
```diff
@@ -130,3 +133,80 @@ class ShellRunner(
...
+      redirectErrorOutput: Boolean,
+      additionalEnv: Map[String, String] = Map.empty,
+      processOut: String => Unit = scribe.info(_),
+      processErr: String => Unit = scribe.error(_),
+      propagateError: Boolean = false,
+      maybeJavaHome: Option[String]
+
+      //       logInfo: Boolean = true
+  ):
+  // Future[
```
tgodzik, line 156: Let's remove these if not needed. I think it should be fine to run it synchronously since it should be quick. Just let's name the method explicitely.

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
```diff
@@ -130,3 +133,80 @@ class ShellRunner(
...
+    //      .add(() => taskResponse.cancel(false))
+
+    //    processFuture.map { code =>
+    //      taskResponse.cancel(false)
+    //      if (logInfo)
+    //        scribe.info(s"time: ran '$commandRun' in $elapsed")
+    //     result.trySuccess(code match {
+    //       case ExitCodes.Success =>
+
+    val scanner = new Scanner(ps.inputStream).useDelimiter("\\A")
```
tgodzik, line 204: Could you maybe move the reading of the string to `JdkVersion` and test it with unit tests?

I don't think it will work with this sample output:
```
openjdk version "17.0.2" 2022-01-18
OpenJDK Runtime Environment Temurin-17.0.2+8 (build 17.0.2+8)
OpenJDK 64-Bit Server VM Temurin-17.0.2+8 (build 17.0.2+8, mixed mode, sharing)
```
did you test it?

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
```diff
@@ -227,38 +227,38 @@ case class JdkVersion(
 
 object JdkVersion {
 
-  def getJavaVersionFromJavaHome(
-      javaHome: AbsolutePath
+  def maybeJdkVersionFromJavaHome(
+      maybeJavaHome: Option[AbsolutePath]
   ): Option[JdkVersion] = {
-
-    def fromReleaseFile: Option[JdkVersion] = {
```
tgodzik, line 234: This should not be removed. We should only use `quickRunWithShortStringResult` as a last resort.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -389,13 +389,21 @@ class MetalsLanguageServer(
           workspace,
           fingerprints
         )
+
+        val optJavaHome =
+          (userConfig.javaHome orElse JdkSources.defaultJavaHome)
+            .map(AbsolutePath(_))
+        val maybeJdkVersion: Option[JdkVersion] =
+          JdkVersion.maybeJdkVersionFromJavaHome(optJavaHome)
```
tgodzik, line 397: If we already have javaHome thre is no need to run it.

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
zmerr: I remember on the linked issue, you concluded the matter in a different way. Could you please take a look [here](https://github.com/scalameta/metals/issues/3536#issuecomment-1110088096) to see what best can be done about it?

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
tgodzik: Do you access it in the tests? I can't find it.

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
zmerr: Ok, no problem.

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
zmerr: tests/unit/src/test/scala/tests/troubleshoot/ProblemResolverSuite.scala is calling this static method:

`JdkVersion.maybeJdkVersionFromJavaHome(javaHome.map(AbsolutePath(_)))`

which in turn is calling `quickRunWithShortStringResult` as a static method.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: I just meant we should do `optJavaHome.getOrElse(JdkVersion.maybeJdkVersionFromJavaHome())`

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
tgodzik: Ok, I guess we could leave it as is, but please try and cache the value maybe in an atomic inside the object?.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: It's javaHome :facepalm: It's fine then :+1:

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
zmerr: You are right. from pprint:

> the shell runner obtained java version as OpenJDK 64-Bit Server VM Homebrew (build 18.0.1.1+0, mixed mode, sharing)

zmerr: Hey, instead of going the `ShellRunner` path, do you think it would have been a better idea to simply do `System.getProperty("java.version”)` according to [this](https://stackoverflow.com/questions/5103121/how-to-find-the-jvm-version-from-a-program) ?

In my case, it is produces just  `18.0.1.1`

which does not require any extraction :)

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
```diff
@@ -227,38 +231,56 @@ case class JdkVersion(
...
+      val properties = ConfigFactory.parseFile(
+        releaseFile.toFile,
+        ConfigParseOptions.defaults().setSyntax(ConfigSyntax.PROPERTIES)
+      )
+      try {
+        val version = properties
+          .getString("JAVA_VERSION")
+          .stripPrefix("\"")
+          .stripSuffix("\"")
+        //   pprint.log("version from releaseFile is "+ version)
```
tgodzik, line 267:
```suggestion
```

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
```diff
@@ -227,38 +231,56 @@ case class JdkVersion(
...
+          .runSync(
+            List("java", "-version"),
+            javaHome,
+            redirectErrorOutput = true,
+            maybeJavaHome = maybeJavaHome.map(_.toString())
+          )
+          .flatMap { javaVersionResponse =>
+            "\\d+\\.\\d+\\.\\d+".r
+              .findFirstIn(javaVersionResponse)
+              .flatMap(JdkVersion.parse)
```
tgodzik, line 249:
```suggestion
        val sbOut = new StringBuilder()
        val exit = ShellRunner
          .runSync(
            List("java", "-version"),
            javaHome,
            redirectErrorOutput = true,
            processOut = s => {
              sbOut.append(s)
              sbOut.append(Properties.lineSeparator)
           },
            maybeJavaHome = maybeJavaHome.map(_.toString())
          )
           if (exit == 0){
             val javaVersionResponse = sbOut.toString
             "\\d+\\.\\d+\\.\\d+".r
              .findFirstIn(javaVersionResponse)
              .flatMap(JdkVersion.parse)
           } else {
             None
           }
```

Also, if you move it to a separate function we could write proper unit tests for it, which would be safer.

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
```diff
@@ -130,3 +133,34 @@ class ShellRunner(
...
+      env,
+      Some(processOut),
+      Some(processErr),
+      propagateError
+    )
+
+    val scanner = new Scanner(ps.inputStream).useDelimiter("\\A")
+    if (scanner.hasNext)
+      Some(scanner.next)
+    else None
```
tgodzik, line 163:
```suggestion
   Await.result(ps.complete, ...)
```
I don't think it's working yet, is it?

tgodzik (changes requested):

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
zmerr: it is.

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
zmerr: `runSync` is returning an `Option[String]`
So it is reading the `inputStream` of the `SystemProcess`
So if there is anything to be read from it, it returns that, otherwise, it returns `None`.
Then if the version regex cannot be detected in the String read from that, the output would again be None in the `fromShellRunner`. So this is delivering the same logic. Just added `JdkVersionSuite` to test and make sure.

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
zmerr: added automated test in JdkVersionSuite.

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
zmerr: If using             

```scala
processOut = s => {
              sbOut.append(s)
              sbOut.append(Properties.lineSeparator)
           }, 

```
is the correct way of getting the process output, I can change the runSync to reflect that. But then, it would be side-effecting, wouldn’t it?
perhapse, I can move the `StringBuilder` in the runSync itself, then return a tuple of (exitCode, maybeOutputString) from it.

zmerr: It’s just that the test is pretty flaky. Sometimes it passes and sometimes it succeeds. Is it due to the incorrect way of getting the process output? If so, I can use the previous logic in runSync which was returning a future. Then the name also perhaps can change.

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
zmerr: the test is flaky. So I would return to the previous logic when it was returning the Future and awaiting its result.

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
tgodzik: > But then, it would be side-effecting, wouldn’t it?

We are running a process, so this is side effecting anyway and also in tooling I wouldn't worry too much about it. We should keep the function consistent with the other ones.

metals/src/main/scala/scala/meta/internal/builds/ShellRunner.scala
tgodzik: I just wonder what invocation is blocking in the function

zmerr: perhaps it sometimes reads the inputStream before the process completes and hence, there is nothing to be read from it.

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
```diff
@@ -180,7 +181,8 @@ object JavaInteractiveSemanticdb {
   def create(
       javaHome: AbsolutePath,
       workspace: AbsolutePath,
-      buildTargets: BuildTargets
+      buildTargets: BuildTargets,
+      maybeJDKVersion: Option[JdkVersion]
```
dos65, line 185:
```suggestion
      jdkVersion: JdkVersion
```
I think this shouldn't be an option.
In `MetalsLanguageServer` we infer a version - it's it's none then there is no point in having `JavaInteractiveSemanticdb`

I noticed that only here there is a call of `JdkVersion.getJavaVersionFromJDK8FallBack` but probably it should be included into a method that infers version in `MetalsLanguageServer`

zmerr: is the fallback to `JdkVersion.getJavaVersionFromJDK8FallBack` useful also in `ProblemSolver`? because that is the other place in MetalsLanguageServer in whicch `maybeJdkVersion` is used.

dos65: > is the fallback to JdkVersion.getJavaVersionFromJDK8FallBack useful also in ProblemSolver

I've just realized that the right thing would be to remove `getJavaVersionFromJDK8FallBack`.
Anyway it won't be called because the version should be obtained from shell.

tests/unit/src/test/scala/tests/troubleshoot/ProblemResolverSuite.scala
```diff
@@ -221,7 +222,8 @@ class ProblemResolverSuite extends FunSuite {
         new TestMtagsResolver,
         () => None,
         () => javaHome,
-        () => isTestExplorerProvider
+        () => isTestExplorerProvider,
+        JdkVersion.maybeJdkVersionFromJavaHome(Some(AbsolutePath(".")))
```
tgodzik, line 226:
```suggestion
        JdkVersion.maybeJdkVersionFromJavaHome(Some(AbsolutePath(System.getProperty("java.home"))))
```

tgodzik (changes requested):
Some small comments from me, otherwise looks good!

metals/src/main/scala/scala/meta/internal/metals/JavaInteractiveSemanticdb.scala
```diff
@@ -227,40 +228,61 @@ case class JdkVersion(
...
   }
 
+//  def getJavaVersionFromJDK8FallBack(
+//      javaHome: AbsolutePath
+//  ): Option[JdkVersion] = {
+//    val rtJar = javaHome.resolve("jre").resolve("lib").resolve("rt.jar")
+//    if (rtJar.exists) Some(JdkVersion(8))
+//    else None
+//  }
+
```
tgodzik, line 285:
```suggestion
```
If it's no longer needed

tgodzik (changes requested):
One last change and we should be able to merge it.

PR #4073, opened 2022-06-28
feat: update test suite location

kpodsiad:
needs https://github.com/scalameta/metals-vscode/pull/1043
VS Code updates range of test suite using their own heuristic but sometimes it yields wrong results. This should be handy also for potential non-vscode clients.

Todo: 
- [x] docs
- [ ] test
- [x] ~now I realized that VS Code doesn't need location, range is enough~ Actually, it doesn't matter that much.

I want to do some cleanup in Test Explorer, get rid of ballast code but this will happen on subsequent PRs.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
@@ -71,25 +71,22 @@ final class TestSuitesProvider(
    * 1. Test Explorer view can be opened and tree view is visible
    * 2. test suite's file can be opened and test cases are visible
    */
-  override def onChange(docs: TextDocuments, file: AbsolutePath): Unit = {
-    if (isEnabled && docs.documents.nonEmpty) {
-      val doc = docs.documents.headOption
-
-      val suiteLocationChanged =
-        if (index.contains(file))
-          doc.toList.flatMap { doc =>
+  override def onChange(docs: TextDocuments, file: AbsolutePath): Unit =
+    docs.documents.headOption.foreach {
+      case doc if isEnabled =>
+        val suiteLocationChanged =
+          if (index.contains(file))
             getTestSuitesLocationUpdates(file, doc)
-          }
-        else Nil
+          else Nil
 
-      val buildTargetUpdates =
-        if (index.hasTestCasesGranularity(file))
-          getTestCasesForPath(file, doc)
-        else Nil
+        val buildTargetUpdates =
+          if (index.hasTestCasesGranularity(file))
+            getTestCasesForPath(file, Some(doc))
+          else Nil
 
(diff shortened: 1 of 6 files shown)


zmerr:
Oh, you’re back ❤️👻

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -124,6 +137,34 @@ final class TestSuitesProvider(
...
+      if loc != entry.testClass.location
+    } yield {
+      val event = UpdateSuiteLocation(
+        entry.suiteInfo.fullyQualifiedName.value,
+        entry.suiteInfo.className.value,
+        loc
+      )
+      (entry.buildTarget, event)
+    }
+    events.groupBy(_._1).toList.map { case (buildTarget, events) =>
```
tgodzik, line 160:
```suggestion
    events.groupBy{case (target, _) }.toList.map { case (buildTarget, events) =>
```

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -124,6 +137,34 @@ final class TestSuitesProvider(
...
+      BuildTargetUpdate(
+        buildTarget,
+        events.map(_._2)
```
tgodzik, line 163:
```suggestion
        events.map{case (_, event) => event}
```

metals/src/main/scala/scala/meta/internal/metals/testProvider/TestSuitesProvider.scala
```diff
@@ -62,19 +63,31 @@ final class TestSuitesProvider(
...
+            getTestSuitesLocationUpdates(file, doc)
+          }
+        else Nil
+
+      val buildTargetUpdates =
+        if (index.hasTestCasesGranularity(file))
+          getTestCasesForPath(file, doc)
+        else Nil
+
+      updateClientIfNonEmpty(suiteLocationChanged ::: buildTargetUpdates)
```
tgodzik, line 90:
```suggestion
  docs.documents.headOption.foreach {
      case doc if isEnabled =>

      val suiteLocationChanged =
        if (index.contains(file))
          doc.toList.flatMap { doc =>
            getTestSuitesLocationUpdates(file, doc)
          }
        else Nil

      val buildTargetUpdates =
        if (index.hasTestCasesGranularity(file))
          getTestCasesForPath(file, doc)
        else Nil

      updateClientIfNonEmpty(suiteLocationChanged ::: buildTargetUpdates)
      case _ =>
```

tgodzik (commented):
Some minor comments, but looks good!

tgodzik (approved):
LGTM!

PR #4074, opened 2022-06-28
bugfix: Make sure that the Scala 3 compiler always retries properly

tgodzik:
Previously, if the compiler was stuck in an infinite loop we would not be able to invoke any methods on the compiler. For Scala 2 we had a separate presentation compiler thread, which would get properly interupted stopped. Now, ~~we interrupt the thread the compiler is running~~ we need to reset the job queue and in the local testing this has helped. Unfortunately, I don't know how to reproduce the infinite loop and this should no longer be an issue for later Scala 3 versions

I also added handling of compiler exceptions, since there are currently none that I know of that should just we thrown and hnadled specifically.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/CompilerAccess.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/CompilerAccess.scala
@@ -23,7 +23,8 @@ import scala.meta.pc.PresentationCompilerConfig
 abstract class CompilerAccess[Reporter, Compiler](
     config: PresentationCompilerConfig,
     sh: Option[ScheduledExecutorService],
-    newCompiler: () => CompilerWrapper[Reporter, Compiler]
+    newCompiler: () => CompilerWrapper[Reporter, Compiler],
+    shouldResetJobQueue: Boolean
 )(implicit ec: ExecutionContextExecutor) {
   private val logger: Logger =
     Logger.getLogger(classOf[CompilerAccess[_, _]].getName)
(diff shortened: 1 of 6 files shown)


mtags/src/main/scala/scala/meta/internal/pc/CompilerAccess.scala
```diff
@@ -214,8 +214,14 @@ abstract class CompilerAccess[Reporter, Compiler](
...
           if (!result.isDone()) {
             try {
+              val runsOnSameThread =
+                _compiler.presentationCompilerThread.isEmpty
               result.cancel(false)
               shutdownCurrentCompiler()
+              /* If running inside the executor, we need to reset the job queue
+               * Otherwise it will block indefinetely in case of infinite loops.
+               */
+              if (runsOnSameThread) jobs.reset()
```
tgodzik, line 224: I changed the logic since, running interrupt work in an simple infinite loop I had, while more complex scenarios it broke. I will try and add a test later.

mtags/src/main/scala/scala/meta/internal/pc/CompilerJobQueue.scala
```diff
@@ -85,6 +85,18 @@ class CompilerJobQueue(newExecutor: () => ThreadPoolExecutor) {
     }
   }
 
+  def reset(): Unit = {
+    state.get() match {
+      case curr @ State.Initialized(v) =>
+        if (state.compareAndSet(curr, State.Empty)) {
+          v.shutdown()
+        } else {
+          delay()
```
dos65, line 94:
```suggestion
          delay()
          reset()
```

dos65 (approved):

mtags/src/main/scala/scala/meta/internal/pc/CompilerJobQueue.scala
tgodzik: Done!

tests/cross/src/test/scala/tests/pc/CancelCompletionSuite.scala
```diff
@@ -93,4 +102,54 @@ class CancelCompletionSuite extends BaseCompletionSuite {
...
+    override def checkCanceled(): Unit = {
+      var hello = true
+      var i = 0
+      while (hello) i += 1
+      hello = false
+    }
+
+  }
+
+  test("break-compilation".tag(IgnoreScala2)) {
```
tgodzik, line 121: I managed to simulate the frozen compilation here :tada:

tanishiking (approved):
LGTM!

Issue #4053, opened 2022-06-22
Metals caches path to java


kurnevsky:
### Describe the bug

Metacls caches path to java in `~/.bloop/bloop.json` and then uses it to start bloop. It's not always correct thing to do. For instance, in NixOS I get the following value of this file: `{"javaHome":"/nix/store/mir2hanra9zc3cmn8h78insschpxdg0i-openjdk-17.0.3+7/lib/openjdk"}`, i.e. it cached path to nix store which might change after NixOS update. This leads to the error `java.io.IOException: Cannot run program "/nix/store/h6f224sdr2iqmcscrbywylsr9q2qhj7m-openjdk-17.0.3+7/lib/openjdk/bin/java"` when starting metals.

### Operating system

Linux

### Editor/Extension

Emacs (lsp-metals)

### Version of Metals

0.11.6

### Extra context or search terms

nixos java bloop

kurnevsky:
Probably metals just followed the symlink `/run/current-system/sw/bin/java -> /nix/store/mir2hanra9zc3cmn8h78insschpxdg0i-openjdk-17.0.3+7/bin/java`?

tgodzik:
> @tgodzik I'm thinking that instead of writing into a config we can extends `bloopgun` and pass the javaHome and options as arguments.

Then the users running Bloop from CLI could have different JVM. Config is the safest place currently, though it would probably be best if bloopgun downloaded the Java instance itself and defaulted to latest LTS + release flags.

dos65:
> Config is the safest place currently, though it would probably be best if bloopgun downloaded the Java instance itself and defaulted to latest LTS + release flags.

@tgodzik if the plan is to prepare to switch to jdk17 we anyway need to download it before starting Metals(run both metals-server and bloop on jdk17+). In this case, the approach with passing jvmHome would work well.

The only non-compat case might came from bloop-cli users but I think we can detect it on Metals side. 

Config thing isn't reliable enough. The same problem with bloop-cli might happen not only in case of NixOS. For example, you have a jdk11, then you decide to install another version and delete the previous one.

kurnevsky:
> though it would probably be best if bloopgun downloaded the Java instance itself and defaulted to latest LTS + release flags.

NixOS is not FHS system, so downloaded binaries don't work here in most cases, without additional patching or environment setup. So if this happens can we have an option to use system java? :)

tgodzik:
[...]

PR #4075, opened 2022-06-29
bugfix: Don't cache Java for nix and additionally use JAVA_HOME

tgodzik:
Fixes https://github.com/scalameta/metals/issues/4053

There is no scenario in which we should cache "/nix/store" path.


kurnevsky:
What about `guix`? :)

tgodzik:
> What about `guix`? :) As I can see they have it in `/gnu/store/`

Added!

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -412,7 +412,13 @@ final class BloopServers(
       case Some(bloopPath) if !bloopPath.exists =>
         // we want to use the same java version as Metals, so it's ok to use java.home
         val metalsJavaHome =
-          userConfiguration.javaHome.orElse(sys.props.get("java.home"))
+          userConfiguration.javaHome
+            .orElse(sys.env.get("JAVA_HOME"))
+            .orElse(sys.props.get("java.home"))
+            // We can't cache java for nix/guix, since the path might change
+            .filterNot(path =>
+              path.contains("/nix/store") || path.contains("/gnu/store/")
```
kurnevsky, line 420: Maybe better to use `startsWith`? Also one is with `/` and another without :)

tgodzik: Sure, changed!

tgodzik:
> I' afraid that this won't solve issues for windows users like [this one](https://discord.com/channels/632642981228314653/632841682211045396/991387711556554772).
> 
> Apart from that, looks good!

Damn, that should have asked the user about updating Bloop json, since it was different than java home.

kurnevsky:
This case doesn't see much different from nixos. Should metals ask to update java path if it differs from JAVA_HOME?

tgodzik:
> This case doesn't see much different from nixos. Should metals ask to update java path if it differs from JAVA_HOME?

It should and that case should be already covered.

tgodzik:
> Hm, I didn't get anything from metals but that exception. Does it ask if JAVA_HOME isn't defined? Maybe metals should check if path defined in bloop config exists?

Hmm.... that's a good idea, at the start we set Metals Java Home, but later we don't check if it still works.

dos65:
@tgodzik hmm. I just remembered that I did a related to `JAVA_HOME` fix for bloop-launcher before we introduced bloop config updates - https://github.com/scalacenter/bloop/pull/1671. Maybe we should just not to write javaHome into it?

There will be 3 cases:
 - no `javaHome` in `bloop.json` - do not write it
 - there is `javaHome` and it was updated by Metals - remove it
 - there is `javaHome` and it was added by a user - make a warning if it's different with Metals one

tgodzik:
> Oh, Update - the fix I mentioned was an only for explicit `JAVA_HOME` env if it's defined before starting Metlas server. However, I still think that passing it thought launcher is the easiest option.

It would if only Metals used Bloop, but since you can connect to it via CLI I didn't want to do it that way.

>what if java is updated? metals will be launched with new java but bloop with old cached java?

It can yeah, but in that case might also suggest updating. I will change that later on so that if the approach backfire then we don't break things :sweat_smile:

dos65:
> It would if only Metals used Bloop, but since you can connect to it via CLI I didn't want to do it that way.

If we want keep bloop-cli we should consider adding an bsp method to it that would return an info about jvm.
This approach would work for all cases and won't affect CLI users.

Tho possible cases in establishing connection to bloop will be:
- no bloop-server - > run it using launcher and specify the same jvm as in Metals
- there is a server -> check that it was run on the same jvm as Metals server or suggest to restart

In this case, we avoid having a wrong jvmHome at all except of the case when it was set by a user explicitly.

tgodzik:
> > It would if only Metals used Bloop, but since you can connect to it via CLI I didn't want to do it that way.
> 
> If we want keep bloop-cli we should consider adding an bsp method to it that would return an info about jvm. This approach would work for all cases and won't affect CLI users.
> 
> Tho possible cases in establishing connection to bloop will be:
> 
> * no bloop-server - > run it using launcher and specify the same jvm as in Metals
> * there is a server -> check that it was run on the same jvm as Metals server or suggest to restart
> 
> In this case, we avoid having a wrong jvmHome at all except of the case when it was set by a user explicitly.

That would be ok for sure and would work with other build servers, but I think the current approach with the fix should work most of the time. If this causes more issues we can switch to your approach, but that would require changes to BSP and I would like fix the current user's problems as best as we can. Especially that the new approach would also take a while to iron out.

dos65:
@tgodzik Honestly, I wondering how often the issue with different jvm happens. In my experience I had it only few times when I was switching default jvm on testing purpose.
So, from side it looks like: there were a rare issue -> now it's fixed but we have another rare issue :smile:

tgodzik:
> @tgodzik Honestly, I wondering how often the issue with different jvm happens. In my experience I had it only few times when I was switching default jvm on testing purpose. So, from side it looks like: there were a rare issue -> now it's fixed but we have another rare issue smile

Ech maybe, but I think a couple of users had issues with it and I just wanted to make it easier to update bloop.json. I might have gone too far though with the automation. Maybe a button in the settings would have been enough :/ 

But I feel that reverting will cause another bunch of issues right now :/

Issue #3955, opened 2021-01-21
[Scala 3] Add scaladoc completion


tgodzik:
**Is your feature request related to a problem? Please describe.**
In Scala 2 it's possible to add a completion that will create a scaladoc stub when `/*` is completed before a method/class . It would be good to have the same for Scala 3.

**Describe the solution you'd like**
The potential solutions should take a look at `ScaladocCompletions.scala` and related tests to see what needs to be supported but in general we would need:
- completion that will create s scaladoc stub should appear when the cursor is after `/*` and before a method, class, trait etc.
- it should also be possible to complete simple multiline comment when nothing exists after `/*`
- 
**Describe alternatives you've considered**
Write it manually.

**Additional context**
This is part of the effort to create proper tooling for Scala 3.

**Search terms**:
scaladoc comments completions completions scala3

tanishiking:
I'm on it :)

PR #4076, opened 2022-06-29
feat: ScaladocCompletion for Scala3

tanishiking:
close https://github.com/scalameta/metals/issues/3955

scaladoc

Overview, basically the same with Scala2 implementation:

- if the cursor follows `/**`, insert `*/` on the cursor and compile it using `InteractiveDriver`
  - We need to close the multiline comment, otherwise, the following code will be commented out, and we can't retrieve the tree of the code under `/**`.
- Find the class/method definition we're documenting
  - the class/method that defined just below the cursor
    - (I simplified the algorithm than Scala2 one. In Scala2, we try to find the class/method that is defined closest to the cursor)
- Extract the method or class constructor parameters, and construct the text.

### DIFF
--- a/mtags/src/main/scala-3.2/scala/meta/internal/pc/MetalsSignatures.scala
+++ b/mtags/src/main/scala-3.2/scala/meta/internal/pc/MetalsSignatures.scala
@@ -1,14 +1,15 @@
 package scala.meta.internal.pc
 
-import dotty.tools.dotc.ast.tpd
-import dotty.tools.dotc.util.SourcePosition
-import dotty.tools.dotc.core.Contexts.*
-import dotty.tools.dotc.util.Signatures
-import dotty.tools.dotc.util.Signatures.Signature
 import scala.meta.internal.mtags.MtagsEnrichments.*
 import scala.meta.pc.SymbolSearch
-import dotty.tools.dotc.core.Flags
+
+import dotty.tools.dotc.ast.tpd
+import dotty.tools.dotc.core.Contexts.*
 import dotty.tools.dotc.core.Denotations.*
+import dotty.tools.dotc.core.Flags
+import dotty.tools.dotc.util.Signatures
+import dotty.tools.dotc.util.Signatures.Signature
+import dotty.tools.dotc.util.SourcePosition
 
 object MetalsSignatures:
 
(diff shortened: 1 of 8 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/completions/ScaladocCompletions.scala
```diff
@@ -0,0 +1,130 @@
...
+   * @return
+   */
+  def isScaladocCompletion(pos: SourcePosition, text: String): Boolean =
+    pos.point >= 3 &&
+      text.charAt(pos.point - 3) == '/' &&
+      text.charAt(pos.point - 2) == '*' &&
+      text.charAt(pos.point - 1) == '*'
+
+  class AssociatedMemberDefFinder(pos: SourcePosition) extends TreeTraverser:
+    private var associatedDef: Option[MemberDef] = None
```
tgodzik, line 88: Does it need to be a var? Maybe we could return it from traverse?

tgodzik (commented):
Looks already great!

mtags/src/main/scala-3.2/scala/meta/internal/pc/MetalsSignatures.scala
```diff
@@ -1,14 +1,15 @@
...
-import dotty.tools.dotc.util.SourcePosition
 import dotty.tools.dotc.core.Contexts.*
+import dotty.tools.dotc.core.Denotations.*
+import dotty.tools.dotc.core.Flags
 import dotty.tools.dotc.util.Signatures
 import dotty.tools.dotc.util.Signatures.Signature
-import scala.meta.internal.mtags.MtagsEnrichments.*
-import scala.meta.pc.SymbolSearch
-import dotty.tools.dotc.core.Flags
-import dotty.tools.dotc.core.Denotations.*
```
tanishiking, line 11: Fixed by `scalafixAll`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
```diff
@@ -23,6 +23,7 @@ sealed trait CompletionValue:
   final def completionItemKind(using Context): CompletionItemKind =
     this match
       case _: CompletionValue.Keyword => CompletionItemKind.Keyword
+      case _: CompletionValue.Document => CompletionItemKind.Snippet
```
tanishiking, line 26: Not sure which `CompletionItemKind` is the most appropriate for doc string 🤔

tanishiking:
Let's hold off until 0.11.7 is released

PR #4077, opened 2022-06-29
feature: Add known scalafix rules' dependencies

tgodzik:
Previously, if a user used an unknown rule we would fail and make them add the dependency manually. Now, we added all known hygiene rules from https://scalacenter.github.io/scalafix/docs/rules/community-rules.html by default.

I was also considering code generation rules, but we can add it later if needed.

This is a stop gap measure until it's possible to define dependencies in `.scalafix.conf` https://github.com/scalacenter/scalafix/issues/1625

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -595,6 +595,21 @@ def isInTestShard(name: String, logger: Logger): Boolean = {
   }
 }
 
+lazy val metalsDependencies = project
+  .in(file("target/.dependencies"))
+  .settings(
+    publish / skip := true,
+    libraryDependencies ++= List(
+      // The dependencies listed below are only listed so Scala Steward
+      // will pick them up and update them. They aren't actually used.
+      "com.lihaoyi" %% "ammonite-util" % V.ammonite,
+      "org.typelevel" % "kind-projector" % V.kindProjector cross CrossVersion.full,
+      "com.olegpy" %% "better-monadic-for" % V.betterMonadicFor,
+      "com.lihaoyi" % "mill-contrib-testng" % V.mill
+    )
+  )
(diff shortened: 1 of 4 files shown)


ckipp01 (approved):
Nice, this is probably a good idea until we get the ability to actually define in the `.scalafix.conf` file.

build.sbt
```diff
@@ -595,6 +600,31 @@ def isInTestShard(name: String, logger: Logger): Boolean = {
...
+  .in(file("target/.dependencies"))
+  .settings(
+    publish / skip := true,
+    excludeDependencies ++= Seq(
+      // we are not interested in downloading scalafix here
+      ExclusionRule("ch.epfl.scala", "scalafix-core_2.13")
+    ),
+    libraryDependencies ++= List(
+      // The dependencies listed below are only listed so Scala Steward
+      // will pick them up and update them. They aren't actually used.
```
dos65, line 613: I'm wondering if we can just use `latest.release` for scalafix rules in Embedded instead of exact build-in version

dos65 (approved):

build.sbt
tgodzik: Good idea, let me check that. Totally forgot about it, we wouldn't need to rely on scala steward

tgodzik: That works, thanks!

build.sbt
```diff
@@ -595,6 +595,21 @@ def isInTestShard(name: String, logger: Logger): Boolean = {
   }
 }
 
+lazy val metalsDependencies = project
```
tgodzik, line 598: I will still keep the dependencies in a separate project, since intransitive prints warnings for each dependency.

PR #4079, opened 2022-06-29
bugfix: Change Java home on startup if broken

tgodzik:
@kpodsiad this should also fix the issue on windows, what do you think?

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
@@ -402,29 +402,46 @@ final class BloopServers(
     }
   }
 
+  private def updateBloopJavaHomeBeforeLaunch(
+      userConfiguration: UserConfiguration
+  ) = {
+    def metalsJavaHome =
+      userConfiguration.javaHome
+        .orElse(sys.env.get("JAVA_HOME"))
+        .orElse(sys.props.get("java.home"))
+    // we should set up Java before running Bloop in order to not restart it
+    bloopJsonPath match {
+      case Some(bloopPath) if !bloopPath.exists =>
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
```diff
@@ -402,29 +402,46 @@ final class BloopServers(
...
+        maybeLoadBloopGlobalJsonFile(bloopPath) match {
+          case (Some(javaHome), opts) =>
+            Try {
+              val homePath = AbsolutePath(Paths.get(javaHome))
+              // fix java home in case it changed
+              if (!homePath.exists)
+                writeJVMPropertiesToBloopGlobalJsonFile(
+                  opts,
+                  metalsJavaHome
+                )
```
kpodsiad, line 430: I would add here two `scribe.info`s here. First one just after if - `detected non existing java path in <path-to-bloop-config> file. The second one after overwriting it `overwrited it with <metalsJava>`

kpodsiad (approved):
Both idea and code looks code. Even though I don't have time to check this myself if it works just as it looks like I'll trust it works. LGTM

ckipp01 (approved):

metals/src/main/scala/scala/meta/internal/metals/BloopServers.scala
tgodzik: Added!

tgodzik: I also added info about which Java home and options is used currently for Bloop and where it is defined, so the users can potentially debug things easier themselves.

And also the first set up should be now logged.

Issue #4062, opened 2022-06-24
Scala 3 "implement all members" adds them above the class


armanbilge:
### Describe the bug

You can see the output of running "implement all members" action on `InMemoryNodeDb` right above it! 😀 

https://github.com/armanbilge/cheshire/blob/069da8e64aab69894330b7bec24cf852b5f531b8/treedb/src/main/scala/cheshire/treedb/InMemoryNodeDb.scala#L33

Should be easy to reproduce by deleting these lines and running the action yourself.
https://github.com/armanbilge/cheshire/blob/069da8e64aab69894330b7bec24cf852b5f531b8/treedb/src/main/scala/cheshire/treedb/InMemoryNodeDb.scala#L19-L30

### Operating system

Linux

### Editor/Extension

VS Code

### Version of Metals

0.11.6+141-a71fe7ab-SNAPSHOT

tanishiking:
Thank you for reporting!
I could reproduce it with

```scala
trait NodeDb[F[_]]:
  type N

  extension (node: N)
    def parent: F[Option[N]]
    def leftChild: F[Option[N]]
    def rightChild: F[Option[N]]

private final class InMemoryNodeDb[F[_]] extends NodeDb[F]
```

and we got

```scala
trait NodeDb[F[_]]:
  type N

  extension (node: N)
    def parent: F[Option[N]]
    def leftChild: F[Option[N]]
    def rightChild: F[Option[N]]


override extension (node: NodeDb.this.N) def parent: Any = ???

override extension (node: NodeDb.this.N) def leftChild: Any = ???

override extension (node: NodeDb.this.N) def rightChild: Any = ???


private final class InMemoryNodeDb[F[_]] extends NodeDb[F]
```

tgodzik:
I will take a quick look here.

tgodzik:
It seems we could group the completions by extensionParams if it exists and create proper extensions. Another issue seems to be that we don't offer overrides for abstract extension methods.

I will leave it to you @tanishiking

tanishiking:
@armanbilge The insert position is fixed, but the code action still works wrong with your code example 😭 
 will follow-up in https://github.com/scalameta/metals/issues/4082

PR #4081, opened 2022-06-30
fix: (scala3) insert correct position for implement-all code action on class with type parameters

tanishiking:
fix https://github.com/scalameta/metals/issues/4062

Implement-all code action for Scala3 infers the insert position based on the existing members if there are no braces.

### fix1
Previously, Metals infer the insert position based on the class's constructor parameters and type parameters.
For example,

```scala
class Concrete[T] extends Base
```

Metals had been inferring the position based on `T` and we end up
inserting stub implementation in the wrong positions.

This commit excludes type params from the "existing declaration".

### fix2
Extension methods shouldn't have an `override` keyword. For example, the following code will be invalid.

```scala
class Concrete extends Base:
  override extension (x: Int) def foo: Int = ???
```

This commit excludes the `override` keyword if the completing method is an extension method.
It's ideal for adding `override` before `def` like 

```scala
class Concrete extends Base:
  extension (x: Int) override def foo: Int = ???
```

But it's a bit tricky to insert the keyword there because the `extension (x: Int) def ...` part is generated by MetalsPrinter.

---

The code action in Scala3 still has a problem with higher kind types, for example

```scala
trait NodeDb[F[_]]:
  type N

  extension (node: N)
    def parent: F[Option[N]]

private final class InMemoryNodeDb[F[_]] extends NodeDb[F]
```

will be 

```scala
trait NodeDb[F[_]]:
  type N

  extension (node: N)
    def parent: F[Option[N]]

private final class InMemoryNodeDb[F[_]] extends NodeDb[F] {
  extension (node: NodeDb.this.N) def parent: Any = ???
}
```

that doesn't compile. But let's work on this issue in another PR :)

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/OverrideCompletions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/OverrideCompletions.scala
@@ -128,7 +128,7 @@ object OverrideCompletions:
       config: PresentationCompilerConfig
   ): ju.List[l.TextEdit] =
     object FindTypeDef:
-      def unapply(path: List[Tree]): Option[TypeDef] = path match
+      def unapply(path: List[Tree])(using Context): Option[TypeDef] = path match
         case (td: TypeDef) :: _ => Some(td)
         // new Iterable[Int] {}
         case (_: Ident) :: _ :: (_: Template) :: (td: TypeDef) :: _ =>
@@ -138,7 +138,11 @@ object OverrideCompletions:
           Some(td)
         case (_: Ident) :: (_: New) :: (_: Select) :: (_: Apply) :: (_: Template) :: (td: TypeDef) :: _ =>
           Some(td)
+        case (dd: DefDef) :: (_: Template) :: (td: TypeDef) :: _
+            if dd.symbol.isConstructor =>
+          Some(td)
         case _ => None
+    end FindTypeDef
 
     val uri = params.uri
     val ctx = driver.currentCtx
@@ -158,14 +162,14 @@ object OverrideCompletions:
     )
     import indexedContext.ctx
 
-    val autoImportsGen = AutoImports.generator(
+    lazy val autoImportsGen = AutoImports.generator(
       pos,
       params.text,
       unit.tpdTree,
       indexedContext,
       config
     )
-    val implementAll = implementAllFor(
+    lazy val implementAll = implementAllFor(
       indexedContext,
       params.text,
       search,
@@ -262,6 +266,7 @@ object OverrideCompletions:
       val decls = td.tpe.decls.toList
         .filter(sym =>
           !sym.isPrimaryConstructor &&
+            !sym.isTypeParam &&
             !sym.is(ParamAccessor) && // `num` of `class Foo(num: int)`
             sym.span.exists
         )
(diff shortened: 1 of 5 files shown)


tanishiking:
Also, it might be ideal to complete like 


```scala
class Concrete extends Base:
  extension (x: Int)
    def foo: Int = ???
    def bar: Int = ???
```

instead of 

```scala
class Concrete extends Base:
  extension (x: Int) def foo: Int = ???
  extension (x: Int) def bar: Int = ???
```

but this PR doesn't do that because

- It requires complete re-implementation of a building text part of the code-action
- Not sure there's a demand

Let's do this if someone raise an issue in another PR

tests/cross/src/test/scala/tests/pc/AutoImplementAbstractMembersSuite.scala
```diff
@@ -935,6 +965,62 @@ class AutoImplementAbstractMembersSuite extends BaseCodeActionSuite {
...
+    """|package a
+       |
+       |trait Base:
+       |  extension (x: Int)
+       |    def foo: Int
+       |    def bar: String
+       |
+       |class Concrete extends Base {
+       |
+       |  exntesion (x: T) def foo: Int = ???
```
tgodzik, line 988:
```suggestion
       |  extension (x: Int) def foo: Int = ???
```

tests/cross/src/test/scala/tests/pc/AutoImplementAbstractMembersSuite.scala
```diff
@@ -935,6 +965,62 @@ class AutoImplementAbstractMembersSuite extends BaseCodeActionSuite {
...
+       |
+       |  extension (x: T) def bar: Int = ???
```
tgodzik, line 990:
```suggestion
       |  extension (x: Int) def bar: Int = ???
```

tanishiking: Auch, fixed! https://github.com/scalameta/metals/pull/4081/commits/724f72652552ddec7ab5d1b62568ba13df774623

tanishiking:
> Maybe we should make sure that MetalsPrinter inserts that correctly?

Added `override` keyword for extension methods

https://github.com/scalameta/metals/pull/4081/commits/724f72652552ddec7ab5d1b62568ba13df774623

mtags/src/main/scala-3/scala/meta/internal/pc/completions/OverrideCompletions.scala
```diff
@@ -353,18 +358,14 @@ object OverrideCompletions:
...
 
-    val asciOverrideDef: String =
-      if sym.is(Abstract) then ""
-      else overrideKeyword
-
-    val overrideDef: String = config.overrideDefFormat() match
+    val overrideDefLabel: String = config.overrideDefFormat() match
       case OverrideDefFormat.Unicode =>
-        if sym.is(Abstract) then "🔼 "
+        if sym.is(Deferred) then "🔼 "
```
tanishiking, line 366: This is irrelevant to the main change, but this change makes `OverrideConfigSuite`'s `unitcode` test closer to Scala2.

PR #4090, opened 2022-07-01
fix: ConvertToNamedArgument code action: register the server command / hide in Scala3 projects

tanishiking:
- I realized that `ConvertToNamedArgument` is not registered as a ServerCommands, and if we try to run the code action, we get `command 'convert-to-named-arguments'` error, and the code action doesn't work
  - fixed: https://github.com/scalameta/metals/commit/60ddbddab3cdbae598ac4cfebcccf3d758f6ef5b
- Hide the `ConvertToNamedArgument` code action in Scala3 project https://github.com/scalameta/metals/commit/716af5938cc91fb46c15b3126ba35914c27add21
  - Though this code action is not yet available in Scala3 project, we can see the code action as if it's available in Scala3.
  - If we try to run the code action in Scala3, we receive the error from the server and nothing happens. This behavior will be confusing for end users.
  - So this commit hide the code action in Scala3 project


metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
```diff
@@ -41,39 +44,47 @@ class ConvertToNamedArguments(trees: Trees) extends CodeAction {
...
-            new l.Position(apply.app.pos.endLine, apply.app.pos.endColumn)
-          )
-          codeAction.setCommand(
-            ServerCommands.ConvertToNamedArguments.toLSP(
-              ServerCommands
-                .ConvertToNamedArgsRequest(
-                  position,
-                  apply.argIndices.map(new Integer(_)).asJava
-                )
+      val maybeApply = for {
```
tanishiking, line 56: see the diff with `w=1`, it's just a indent

tanishiking:
This PR should be merged before 0.11.7

tanishiking:
oops LSP test is failing, I'll take a look

tgodzik (approved):
LGTM!

PR #4091, opened 2022-07-01
Add sbt-dependency-graph workflow

adpi2:
Could be useful to submit the dependency graph of Metals to Github to find vulnerabilities.
But also I would like to test the sbt-dependency-graph-action on a project with many cross Scala versions.


adpi2:
The sbt-dependency-graph workflow will only run on pushes to the main branch because it is (often) not possible to submit a dependency snapshot from a PR and it won't be analysed by Github.

tgodzik (approved):
Ach ok, let's test it out then!

PR #4092, opened 2022-07-01
Fix sbt-dependency-graph workflow

adpi2:
Exclude the test projects to avoid resolving the snapshot version of mtags


tgodzik (approved):

.github/workflows/sbt-dependency-graph.yml
```diff
@@ -16,3 +16,4 @@ jobs:
       - uses: scalacenter/sbt-dependency-graph-action@v1
         with:
           scala-versions: 2.11.12 2.12.16 2.13.8 3.1.3
+          projects: mtags-interfaces mtags mtags3 metals
```
ckipp01, line 19: dang, cool. It even has a projects settings. Nice.

PR #4095, opened 2022-07-04
Fix sbt-dependengy-graph workflow again

adpi2:


.github/workflows/sbt-dependency-graph.yml
```diff
@@ -15,5 +15,5 @@ jobs:
       - uses: actions/checkout@v3
       - uses: scalacenter/sbt-dependency-graph-action@v1
         with:
-          scala-versions: 2.11.12 2.12.16 2.13.8 3.1.3
+          scala-versions: 2.13.8
```
ckipp01, line 18: Are you sure switching to only 2.13.8 will fix the issue? Looks like the issue was actually related to 2.13.

```
[error] sbt.librarymanagement.ResolveException: Error downloading org.scalameta:mtags_2.13.8:0.0.0+1-98d9d6b6-SNAPSHOT```

ckipp01: ahhh, gotcha. 👍🏼

ckipp01 (approved):
Thanks for the fix!

ckipp01:
Noice, looks like we have a dep list! Thanks @adpi2! https://github.com/scalameta/metals/network/dependencies

PR #4101, opened 2022-07-05
docs: Fix wording about the expression evaluation

tgodzik:
I realzied that the wording here is quite poor and I haven't really explained that there might be some bugs


ckipp01 (approved):
LGTM

PR #4102, opened 2022-07-05
Source jar index

### LINKED ISSUE #3815: Files in sources jars with extra parent directories are not found
## Issue description

When working with Scala in VS Code, Metals will download sources jars and support dereferencing classes from them in code files by the "Go to definition" context menu option. However if said source jars contain extra parent directories, e.g. source files are under `src/main/java/com/...` in the jar instead of `com/...` then this silently fails and you get "No definition found". The "Language Support for Java by Red Hat" on the other hand correctly finds the source files and allows dereferencing. The attached project demonstrates this for the [bindings-rxjava](https://search.maven.org/artifact/com.daml/bindings-rxjava/2.0.0/jar) and [bindings-java](https://search.maven.org/artifact/com.daml/bindings-java/2.0.0/jar) jars which respectively have single and multiple buried roots.

## Reproduction steps

1. Set JAVA_HOME to a JDK-11 installation.
2. Install Scala (Metals) and Language Support for Java by Red Hat
3. Unpack the attached `src-jar-test.zip` and load the resulting `src-jar-test` directory into VS Code
4. Wait for the project to import
5. Open the "JarTest.java" file.
6. Note that hovering or Ctrl/Cmd hovering over the classes `DamlLedgerClient` (single-buried-root) and `TransactionFilter` (multiple roots) results in Javadoc and/or source previews, and Ctrl/Cmd-clicking goes to the source.
7. Open the "JarTest.scala" file.
8. Note that Ctrl/Cmd hovering / clicking does NOT work in the Scala file, though it WILL work for the slf4j classes.

[src-jar-test.zip](https://github.com/scalameta/metals/files/8464190/src-jar-test.zip)

### Expected behavior

I have not found anything in the [Jar file specification](https://docs.oracle.com/en/java/javase/11/docs/specs/jar/jar.html) specifically addressing this point, but in practice the Java plugins for VS Code and IntelliJ both find the sources despite the extra parent directories, and I expected Metals to do so as well.

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.2

_No response_

zmerr:
it resolves #3815

### DIFF
--- a/metals-bench/src/main/scala/bench/MetalsBench.scala
+++ b/metals-bench/src/main/scala/bench/MetalsBench.scala
@@ -77,8 +77,22 @@ class MetalsBench {
 
   @Benchmark
   @BenchmarkMode(Array(Mode.SingleShotTime))
+  def mtagsJavaIndex(): Unit = {
+    javaDependencySources.inputs.foreach { input =>
+      Mtags.index(input, dialects.Scala213)
+    }
+  }
+
+  @Benchmark
+  @BenchmarkMode(Array(Mode.SingleShotTime))
   def toplevelsScalaIndex(): Unit = {
     scalaDependencySources.inputs.foreach { input => Mtags.toplevels(input) }
+  }
+
+  @Benchmark
+  @BenchmarkMode(Array(Mode.SingleShotTime))
+  def toplevelsJavaIndex(): Unit = {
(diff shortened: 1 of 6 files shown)


tgodzik:
We will need to go back to this later on as it's a bit more complex and only needed in a niche scenario

Issue #4035, opened 2022-06-15
.sc file gets error in VScode 


ZongHan13:
### Describe the bug
When I open a project in vscode which is  builded by mill then metals will detect Ammonite script, it will ask you to import `.sc` file or not.
![截圖 2022-06-15 20 23 10](https://user-images.githubusercontent.com/63963971/173829615-3d08af00-a564-4dcf-984e-79a0cca93bf1.png)
---

If choose import or Automatically all the `.sc` file will get error like this.
![截圖 2022-06-15 20 45 03](https://user-images.githubusercontent.com/63963971/173830017-98fe36bb-2167-4207-af17-f4c8083fc805.png)
![截圖 2022-06-15 20 45 47](https://user-images.githubusercontent.com/63963971/173830238-48ef246f-da0b-4f5a-ab48-ed3dc3aedbef.png)
---
When import ammonite script metals output 
![截圖 2022-06-15 21 50 21](https://user-images.githubusercontent.com/63963971/173844547-3324e5db-8df9-4d94-aa7f-e20d49416f09.png)
---

Metals Run doctor 
![截圖 2022-06-15 21 57 08](https://user-images.githubusercontent.com/63963971/173845672-cf516ecd-5485-4428-886e-3ad315c7969b.png)
### And the error part 
![截圖 2022-06-15 21 58 07](https://user-images.githubusercontent.com/63963971/173845700-636bf962-8182-4590-8a31-f25a09917002.png)

But if choose dismiss metals  will working  fine.
![截圖 2022-06-15 20 48 58](https://user-images.githubusercontent.com/63963971/173830776-311693b7-20cb-4635-9d08-8ca4b6911e3e.png)
----
Reproduce step:
1. Create mill project.
2. Open project will VScode
3. When metals detect ammonite script import script
---

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.06

### Extra context or search terms

Ammonite, VScode, metals

tanishiking:
Hmm, Metals should show `New Mill workspace detected, would you like to import the build?` popup instead of detecting `build.sc` as an Ammonite script.
I tried to reproduce the issue, but I couldn't get `Ammonite script detected` for `build.sc` 🤔 
Does anybody have any ideas who know more about mill/ammonite support?

ckipp01:
> Does anybody have any ideas who know more about mill/ammonite support?

[...]

PR #4106, opened 2022-07-06
bugfix: Build.sc was seen as Ammonite script

jkciesluk:
Previously, metals saw mill build.sc as ammonite script, now we filter it out in isAmmoniteScript so that it's fixed

fixes https://github.com/scalameta/metals/issues/4035

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/mtags/CommonMtagsEnrichments.scala
+++ b/mtags/src/main/scala/scala/meta/internal/mtags/CommonMtagsEnrichments.scala
@@ -355,7 +355,7 @@ trait CommonMtagsEnrichments {
     def isAmmoniteGeneratedFile: Boolean =
       doc.endsWith(".amm.sc.scala")
     def isAmmoniteScript: Boolean =
-      isScalaScript && !isWorksheet
+      isScalaScript && !isWorksheet && !doc.endsWith("build.sc")
     def asSymbol: Symbol = Symbol(doc)
     def endsWithAt(value: String, offset: Int): Boolean = {
       val start = offset - value.length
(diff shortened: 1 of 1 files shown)


mtags/src/main/scala/scala/meta/internal/mtags/CommonMtagsEnrichments.scala
```diff
@@ -355,7 +355,7 @@ trait CommonMtagsEnrichments {
     def isAmmoniteGeneratedFile: Boolean =
       doc.endsWith(".amm.sc.scala")
     def isAmmoniteScript: Boolean =
-      isScalaScript && !isWorksheet
+      isScalaScript && !isWorksheet && !doc.endsWith("build.sc")
```
tgodzik, line 358: We should accept scripts like `test-build.sc` etc.
```suggestion
      isScalaScript && !isWorksheet && !doc.endsWith("/build.sc")
```

mtags/src/main/scala/scala/meta/internal/mtags/CommonMtagsEnrichments.scala
```diff
@@ -500,7 +500,7 @@ trait CommonMtagsEnrichments {
       filename.endsWith(".sc")
     }
     def isAmmoniteScript: Boolean =
-      isScalaScript && !isWorksheet
+      isScalaScript && !isWorksheet && !filename.endsWith("build.sc")
```
tgodzik, line 503:
```suggestion
      isScalaScript && !isWorksheet && filename != "build.sc"
```

tgodzik (changes requested):
Thanks for digging into this! I realized we still have an edge case to cover.

tanishiking (commented):
Just FYI, this would be a stretch goal, but for more accuracy, it would be great if we can check the build tool is mill by https://github.com/scalameta/metals/blob/d2587f0ef7765180bf1afb783a182ccb03909842/metals/src/main/scala/scala/meta/internal/metals/ammonite/Ammonite.scala#L209-L213
However, it requires a `BuildTool` instance, and it takes more refactoring effort to use it.

Current implementation won't see `build.sc` as ammonite script even if it's not under the mill project.
However, nevermind, it's just a corner case, you don't need to care it :+1:

tgodzik (approved):
LGTM!

PR #4109, opened 2022-07-06
bugfix: Occurence highlight did not work for local vars

jkciesluk:
Previously occurrences of `var` in methods as reference and assignment were different, now we have separate logic for them, so that it highlights all occurrences properly

fixes https://github.com/scalameta/metals/issues/4093

### DIFF
--- a/tests/unit/src/test/scala/tests/DocumentHighlightLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/DocumentHighlightLspSuite.scala
@@ -99,6 +99,47 @@ class DocumentHighlightLspSuite extends BaseLspSuite("documentHighlight") {
       |}""".stripMargin
   )
 
+  check(
+    "local-var",
+    """
+      |object Test {
+      |  def met() = {
+      |    class T1(var <<abc>>: Int) {
+      |      <<ab@@c>> = 4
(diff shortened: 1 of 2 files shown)


kpodsiad (approved):
Hi @jkciesluk! 
I would add a few sentences of scaladoc for both `findLocalsInScope` and `localVarAlternatives` for future us ;) Otherwise LGTM 💪

tests/unit/src/test/scala/tests/DocumentHighlightLspSuite.scala
```diff
@@ -99,6 +99,47 @@ class DocumentHighlightLspSuite extends BaseLspSuite("documentHighlight") {
...
+      |    }
+      |  }
+      |}""".stripMargin
+  )
+  check(
+    "local-assign",
+    """
+      |object Test {
+      |  def met() = {
+      |    class T1(var <<abc>>: Int) {
```
tgodzik, line 119: Could you add a second class T2 so that we can see if we don't highlight too much.

tgodzik (changes requested):
Great work! Thanks for looking into this!

You need to run `sbt scalafixAll` for all the CI checks to pass. It should sort the imports.

tgodzik (approved):
LGTM!

PR #4111, opened 2022-07-06
bugfix: Allow completions in multiline expressions when debugging

tgodzik:
Previously, I assumed that only single line is allowed when using expression evaluator, which turned out to be false. Now, we should provide proper completions even if the expression is multiline.

Unfortunately the logic is not as clear as I would like it to be, but making it work was not fun. The API for those completions is much worse than the normal one.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
@@ -324,7 +324,7 @@ class Compilers(
         val offsetParams = CompilerOffsetParams(
           path.toURI,
           modified,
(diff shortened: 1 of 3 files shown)


adpi2:
That's great but I will take some time to test it locally before approving.

tgodzik:
@adpi2 thanks for checking this out! I had 2 off by one errors which were negating themselves :sweat: One was in the test.

You can see that in https://github.com/scalameta/metals/pull/4111/commits/8dbe12f9721acb099de116c6eaa13848030a219e

adpi2 (approved):
LGTM

PR #4114, opened 2022-07-07
Add LSP jar file system

Arthurm1:
This is yet another way of implementing the FileSystemAPI.

I found with https://github.com/scalameta/metals/pull/3750 that it was getting too complicated and there were code paths in the MetalsURLConnection that I was unsure of.  There were also performance issues and I think it would have taken a lot of investigation to get the custom URLConnection and custom FileSystem on par with the built in Java ones.

So instead this PR simply maps `jar:` <-> `metalsfs:` uris whenever Metals communicates via LSP or via Debug Adaptor.  Bugs in this area should be easy to track down/fix as it's probably down to missing mapper code in one of the messages.

This PR shouldn't change any client that doesn't support `jar:` scheme in virtual documents.  As far as I know - that's only VSCode.  If there are others then they will need to change from handling `jar:` to handling `metalsfs:` scheme.  If they want they can also implement the LSP FileSystemAPI but they don't have to as virtual docs is separate from FileSystemAPI support.

By default - on VSCode - the user won't switch to a multi-workspace and display `Metals libraries` in the File Explorer but they have a command `Show libraries folder in file explorer` to do that.

I've tested this PR with navigation, debugging, searching jars, Metals Tree View and all seems ok.

This PR also auto-decompiles class files when they are missing source jars and allows navigation within that decompiled class file.  It also allows decoding of class and tasty files from within jars.  Currently it only does this from the file explorer. You can't run "goto definition" on a symbol that has no source jar.  Ideally this would jump to an auto-decompiled class file of the workspace jar.  I'm just not sure where this should be done.  It doesn't find the non-source jar because `SymbolIndexBucket#query0` only calls `loadFromSourceJars` - it doesn't try a `loadFromWorkspaceJars` equivalent.  Is that where the code should be changed?  Or should it be in `MetalsLanguageServer#definitionOrReferences` I don't want to cause issues (or performance issues) in other codepaths.

I've changed the MetalsDebugAdapter interfaces to use URI (as a String) instead of AbsolutePath.  Is this an issue - is it a public interface?  I see that there are 2 versions of the interface - do I need to create a 3rd?
The issue is that `metalsfs:` is not a valid path (which AbsolutePath wraps) and passing it as a URI fixes this.
I wonder if the uri mapping will also help VSCode notebook cells which I think have the scheme `vscode-notebook-cell:`

I've tried to test this without virtual docs enabled to mimic the behaviour of non-VSCode clients but it would be handy if someone double checked I've not caused any issues for those clients.

I've done nothing with handling `CompletionRequest ` in `DebugProxy` line:120->146.  I'm not sure if this should handle jar files - what does it actually do?

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/LSPFileSystemProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/LSPFileSystemProvider.scala
@@ -1,0 +1,249 @@
+package scala.meta.internal.metals
+
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.Path
+import java.util.stream.Collectors
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.util.control.NonFatal
+
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.clients.language.MetalsLanguageClient
+
+final case class FSReadDirectoryResponse(
+    name: String,
+    isFile: Boolean,
+)
+final case class FSReadDirectoriesResponse(
+    name: String,
+    directories: Array[FSReadDirectoryResponse],
+    error: String,
+)
+
+final case class FSReadFileResponse(
+    name: String,
+    value: String,
+    error: String,
+)
+
+final case class FSStatResponse(
+    name: String,
+    isFile: Boolean,
+    error: String,
+)
+
+final class LSPFileSystemProvider(
+    languageClient: MetalsLanguageClient,
+    uriMapper: URIMapper,
+    fileDecoderProvider: FileDecoderProvider,
+    clientConfig: ClientConfiguration,
+)(implicit ec: ExecutionContext) {
+
+  // cache the last few decompiled class files as they're repeatedly accessed by VSCode once opened or referenced
+  private val cache = LRUCache[Path, String](10)
+
+  def sendLibraryFileSystemReady(): Unit = {
+    if (clientConfig.isLibraryFileSystemSupported()) {
+      val params =
+        ClientCommands.LibraryFileSystemReady.toExecuteCommandParams()
+      languageClient.metalsExecuteClientCommand(params)
+    }
+  }
+
+  private def getLocalPathDirectory(
+      uri: String,
+      prefix: String,
+      mappingFunction: String => FileSystemInfo,
+  ): Array[FSReadDirectoryResponse] = {
+    val (name, remaining) = URIMapper.getURIParts(uri, prefix)
+    val fs = mappingFunction(name).fs
+    // assume all jar paths only have `/` as root
+    val path = fs.getPath(remaining.getOrElse("/"))
+    Files
+      .list(path)
+      .collect(Collectors.toList())
+      .asScala
+      .map(path =>
+        FSReadDirectoryResponse(
+          path.getFileName.toString,
+          Files.isRegularFile(path),
+        )
+      )
+      .toArray
+  }
+
+  private def getLocalSystemStat(
+      uri: String,
+      prefix: String,
+      mappingFunction: String => FileSystemInfo,
+  ): FSStatResponse = {
+    val (name, remaining) = URIMapper.getURIParts(uri, prefix)
+    try {
+      val fs = mappingFunction(name).fs
+      // assume all jar paths only have `/` as root
+      val path = fs.getPath(remaining.getOrElse("/"))
+      FSStatResponse(uri, Files.isRegularFile(path), null)
+    } catch {
+      case NonFatal(e) =>
+        scribe.error(s"getLocalSystemStat $uri $name $remaining", e)
+        throw e
+    }
+  }
+
+  private def getLocalFileContents(
+      uri: String,
+      prefix: String,
+      mappingFunction: String => FileSystemInfo,
+  ): FSReadFileResponse = {
+    val (name, remaining) = URIMapper.getURIParts(uri, prefix)
+    val fs = mappingFunction(name).fs
+    val path = fs.getPath(remaining.getOrElse("/"))
+    val contents =
+      if (path.isClassfile) {
+        val cacheResult = cache.get(path)
+        cacheResult.getOrElse({
+          val response =
+            fileDecoderProvider.decodeCFRAndWait(path.toUri.toAbsolutePath)
+          val success = Option(response.value)
+          success.foreach(content => cache += path -> content)
+          success.getOrElse(response.error)
+        })
+      } else
+        new String(Files.readAllBytes(path), StandardCharsets.UTF_8)
+    FSReadFileResponse(uri, contents, null)
+  }
+
+  def readDirectory(
+      uri: String
+  ): Future[FSReadDirectoriesResponse] = Future {
+    try {
+      val subPaths = uri match {
+        case URIMapper.parentURI =>
+          Array(
+            FSReadDirectoryResponse(URIMapper.jdkDir, false),
+            FSReadDirectoryResponse(URIMapper.workspaceJarDir, false),
+            FSReadDirectoryResponse(URIMapper.sourceJarDir, false),
+          )
+        case URIMapper.jdkURI =>
+          uriMapper.getJDKs
+            .map(jdk => FSReadDirectoryResponse(jdk, false))
+            .toArray
+        case URIMapper.workspaceJarURI =>
+          uriMapper.getWorkspaceJars
+            .map(jar => FSReadDirectoryResponse(jar, false))
+            .toArray
+        case URIMapper.sourceJarURI =>
+          uriMapper.getSourceJars
+            .map(jar => FSReadDirectoryResponse(jar, false))
+            .toArray
+        case jdk if jdk.startsWith(URIMapper.jdkURI) =>
+          getLocalPathDirectory(
+            jdk,
+            URIMapper.jdkURI,
+            uriMapper.getJDKFileSystem,
+          )
+        case workspaceJar
+            if workspaceJar.startsWith(URIMapper.workspaceJarURI) =>
+          getLocalPathDirectory(
+            workspaceJar,
+            URIMapper.workspaceJarURI,
+            uriMapper.getWorkspaceJarFileSystem,
+          )
+        case sourceJar if sourceJar.startsWith(URIMapper.sourceJarURI) =>
+          getLocalPathDirectory(
+            sourceJar,
+            URIMapper.sourceJarURI,
+            uriMapper.getSourceJarFileSystem,
+          )
+      }
+      FSReadDirectoriesResponse(uri, subPaths, null)
+    } catch {
+      case NonFatal(e) => FSReadDirectoriesResponse(uri, null, e.toString)
+    }
+  }
+
+  def readFile(uri: String): Future[FSReadFileResponse] = Future {
+    try {
+      uri match {
+        case jdk if jdk.startsWith(URIMapper.jdkURI) =>
+          getLocalFileContents(
+            jdk,
+            URIMapper.jdkURI,
+            uriMapper.getJDKFileSystem,
+          )
+        case workspaceJar
+            if workspaceJar.startsWith(URIMapper.workspaceJarURI) =>
+          getLocalFileContents(
+            workspaceJar,
+            URIMapper.workspaceJarURI,
+            uriMapper.getWorkspaceJarFileSystem,
+          )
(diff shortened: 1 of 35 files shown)


kpodsiad:
(I haven't looked at this PR yet so this comment may be absurd)
I was experimenting with Java & vscode and I found that they show dependencies in a separate tree view. It's similar to `packages` in Metals view). I didn't check how they handle with missing sources but this might a a good lead.

<details closed>
<summary>Screenshot</summary>
Screenshot 2022-07-07 at 15 07 05
</details>

tgodzik:
> @tgodzik This is ready for review whenever you have time.
> 
> I think the `FileWatcherLspSuite` test failure on the Mac is spurious as it happens on virtual docs and the readonly. But I can't test locally as I don't have mac.

Awesome! I will take a look after the weekend.

tgodzik:
> I've done nothing with handling CompletionRequest in DebugProxy line:120->146. I'm not sure if this should handle jar files - what does it actually do?

It handles completions in the debug console, so most likely it doesn't need to be handled here. Taking a look at the PR!

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1058,15 +1070,20 @@ class MetalsLanguageServer(
...
     // and we would lose the notion of the focused document
     focusedDocument.foreach(recentlyFocusedFiles.add)
     focusedDocument = Some(path)
     recentlyOpenedFiles.add(path)
 
     // Update md5 fingerprint from file contents on disk
-    fingerprints.add(path, FileIO.slurp(path, charset))
+    fingerprints.add(
+      path,
+      if (path.isJarFileSystem) params.getTextDocument.getText
```
tgodzik, line 1084: We shouldn't need to add fingerprints for jarFileSystem. There are really only useful if the file can change,

metals/src/main/scala/scala/meta/internal/metals/LSPFileSystemProvider.scala
```diff
@@ -0,0 +1,249 @@
...
+    error: String,
+)
+
+final case class FSStatResponse(
+    name: String,
+    isFile: Boolean,
+    error: String,
+)
+
+final class LSPFileSystemProvider(
```
tgodzik, line 37: I wonder if creating out own FileSystemProvider wouldn't actually cause less work (less changes and no need for a mapper), we could handle those URIs like normal ones and paths should be fine also. We would only mos likely need to map them for DAP

A lot of the methods in https://docs.oracle.com/javase/8/docs/api/java/nio/file/spi/FileSystemProvider.html can be left blank.

I can try and take a look at it if you don't have the time (unless you think it's not a great idea)

metals/src/main/scala/scala/meta/internal/metals/LSPFileSystemProvider.scala
```diff
@@ -0,0 +1,249 @@
...
+    languageClient: MetalsLanguageClient,
+    uriMapper: URIMapper,
+    fileDecoderProvider: FileDecoderProvider,
+    clientConfig: ClientConfiguration,
+)(implicit ec: ExecutionContext) {
+
+  // cache the last few decompiled class files as they're repeatedly accessed by VSCode once opened or referenced
+  private val cache = LRUCache[Path, String](10)
```
tgodzik, line 45: Do we need it? Most likely the user will have the class open until it's not needed.

metals/src/main/scala/scala/meta/internal/metals/LSPFileSystemProvider.scala
```diff
@@ -0,0 +1,249 @@
...
+    } catch {
+      case NonFatal(e) =>
+        scribe.error(s"getSystemStat $uri", e)
+        FSStatResponse(uri, false, e.toString())
+    }
+  }
+}
+
+import scala.collection.mutable
+// TODO this is not thread safe - do we care?
```
tgodzik, line 236: I think we can drop the cache, I don't really see much benefit to it.

tgodzik (changes requested):
Thanks for this PR! That was a lot of great work. I left a couple of comments and lose thoughts.

Does the PR need some more changes in VS Code extension? I seem to be getting:

image

metals/src/main/scala/scala/meta/internal/metals/LSPFileSystemProvider.scala
Arthurm1: @tgodzik I have created a working FileSystemProvider in #3750.  It's pretty much complete - I just never pushed the last version because it was slow.  I guess I could spend some time optimising.  My other issue was the URL translation for the custom URLConnection class needed - it's an area that I don't know much about so I'm not sure how many types of URL have to be dealt with for the URLClassLoader to work with the custom FileSystem - I ended up going round in circles a bit there.  You're right in that it's less invasive to the rest of the codebase - no mapping required except for the debugger.  If you want, I can push the latest version of #3750 and you can decide for yourself if you think it's a better approach.

metals/src/main/scala/scala/meta/internal/metals/LSPFileSystemProvider.scala
Arthurm1: The problem here is that there are a lot of LSP messages flying around when using a document and the message handlers nearly always require the text of the document.
The `metals/didFocusTextDocument` message - which appears to be a Metals specific message? - doesn't pass the text of the document (unlike the standard LSP messages: `textDocument/didOpen`, `textDocument/didChange`, `textDocument/didClose`) and so every time there's a change of focus involving a decompiled class file, Metals has to decompile it again as it needs the text.  This takes a few seconds, especially as it's an external process and we don't use CFR as a library which would be much faster.
Ideally we'd change `metals/didFocusTextDocument` to include the text and pass that text into `interactiveSemanticdbs.didFocus(path)`, `syntheticsDecorator.publishSynthetics(path)`, `worksheetProvider.onDidFocus(path)` which all appear to reload the text from disk.  There's a lot of fileIO going on for all of these.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
Arthurm1: I agree except I think fingerprints are used to get the text of the file in `SyntheticsDecorationProvider#enrichWithText` but I would like that to change so the text is passed in from the LSP message.

Arthurm1:
@tgodzik There is possibly a third way of implementing this.

The LSP filesystem could present the exact file system info the jar has.
e.g. `jar:file:///home/me/coursier/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.12.16/scala-library-2.12.16.jar`

This would mean no special `metalfs:` uri scheme or java file system needing to be created.  Or any mapping of URIs within message handlers.  Or any mapping for the debug adapter.

The downside is that the file explorer would not be flat so the jars would be in deep dirs. Breadcrumbs would work within a jar - which is likely where they are most used.
To solve the file explorer issue - we could add a additional nodes to the `Packages` Tree view (under `Projects` and `Libraries`) which would allow the user to browse the jdk, source jars and workspace jars.

I think this would add a chunk of functionality to the user (breadcrumb navigation and jar browsing) that could be built on by Metals later.  Then we could either go down the java file system root or down the mapping root but that could be decided at a later date.

Ideally something like https://github.com/microsoft/vscode/pull/124903 would be implemented so that a flat structure could be represented by the LSPFileSystemAPI but the original `jar:` URIs would be used for file management

metals/src/main/scala/scala/meta/internal/metals/LSPFileSystemProvider.scala
tgodzik: If the issue with `URLClassLoader` is complicated then maybe the current approach would work better. Let me think about it for a bit.

metals/src/main/scala/scala/meta/internal/metals/LSPFileSystemProvider.scala
tgodzik: Ok sure, makes sense to use it then.

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
tgodzik: So do you save it in the fingerprints to avoid the complexity of reading it from the file system?

Arthurm1: When I came to look at auto-decompiling class files using CFR and realised the decompile took a few seconds for every time that file is read from disk, I started tracing how often Metals uses `FileIO.slurp` and it seems to use it a lot and, to me, that seems unnecessary since all the LSP messages send us the file contents.  I see that fingerprinting is used to go back to the last successfully compiled version but I don't get why it's used in `SyntheticsDecorationProvider#enrichWithText` instead of just using the contents passed by LSP and I don't get why it's used in many other places e.g. InteractiveSemanticDB, when the text is readily available.  I doubt this has been an issue with any files present on the file system but when you have to decompile the `class` file then that has to be wired in and it still takes time so why do it at all?
But yes - mostly because of the additional decompile phase needed when slurping a class file.

tgodzik: >  I see that fingerprinting is used to go back to the last successfully compiled version but I don't get why it's used in

We only have semanticdb for the last compiled state, so whenever we use it we try to check for the differences between the current version and the compiled one to calculate adjustments to positions. 

> But yes - mostly because of the additional decompile phase needed when slurping a class file.

I think that should be fine then, but for that case ideally we would use buffers, which contain all the currently opened files. Fingerprints are also saved currently in between sessions and there is no need to save them if they don't really change.

Arthurm1:
> There is possibly a third way of implementing this.
> 
> The LSP filesystem could present the exact file system info the jar has.
> e.g. jar:file:///home/me/coursier/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.12.16/scala-library-2.12.16.jar
> 

Just remembered we can't do that as VSCode doesn't like "jar:file:" uris under the FileSystemAPI, only under virtual documents.

tgodzik:
> > There is possibly a third way of implementing this.
> > The LSP filesystem could present the exact file system info the jar has.
> > e.g. jar:file:///home/me/coursier/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.12.16/scala-library-2.12.16.jar
> 
> Just remembered we can't do that as VSCode doesn't like "jar:file:" uris under the FileSystemAPI, only under virtual documents.

Ach ok, I am wondering though if we could do a proxy filesystem that would exactly the same mapping we do here and would work automatically with the new URIs. Would that be possible? I haven't yet looked into it, but I plan to see if we can improve this PR a bit in terms of maintainability

Arthurm1:
@tgodzik 
> I am wondering though if we could do a proxy filesystem that would exactly the same mapping we do here and would work automatically with the new URIs. Would that be possible? I haven't yet looked into it, but I plan to see if we can improve this PR a bit in terms of maintainability

#3750 is pretty much a proxy filesystem.  It fakes the initial path section: `metalsfs:/root/source/somejar.jar` and then delegates any further path to the `jar:` filesystem.  I can look into why it's slow if you're not keen on this mapping PR.

tgodzik:
> @tgodzik
> 
> > I am wondering though if we could do a proxy filesystem that would exactly the same mapping we do here and would work automatically with the new URIs. Would that be possible? I haven't yet looked into it, but I plan to see if we can improve this PR a bit in terms of maintainability
> 
> #3750 is pretty much a proxy filesystem. It fakes the initial path section: `metalsfs:/root/source/somejar.jar` and then delegates any further path to the `jar:` filesystem. I can look into why it's slow if you're not keen on this mapping PR.

Sorry for talking so long to look into this! I am just still trying to figure out which is the best way here. I would probably prefer the filesystem as the mapper adds the requirement that we always need to remember to map the URI, which is a bit less maintainable I think. On the other hand if using the file system is slow, I can try and look into that. In which cases was it slow?

I haven't forgotten about your PR and would love to have it merged, just want to minimize the amount of potential issues.

ckipp01:
Going to mark this as a draft for now. If you dive back into this, please do mark it back as ready to review!

tgodzik:
> Going to mark this as a draft for now. If you dive back into this, please do mark it back as ready to review!

It's really stuck on me, since I was trying to think on how to make it a bit less complex and have so far failed to figure it out :/

tgodzik:
https://github.com/scalameta/metals-vscode/pull/2073 might be a better approach

Issue #4044, opened 2021-02-10
Support Call Hierarchy Requests


ckipp01:
In LSP 3.16 support for Call Hierarchy requests were added. You can find information about them [here in the spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_prepareCallHierarchy).

You can see an example of this from Rust, which will give you an idea of what it looks like:

![](https://user-images.githubusercontent.com/6868531/71634596-6d57c900-2beb-11ea-9957-d1bb1aa92d3b.png)

https://github.com/rust-analyzer/rust-analyzer/pull/2698
**Search terms**:
LSP, call hierarchy

PR #4115, opened 2022-07-07
Feature/call hierarchy

riiswa:
Here is the PR related to the call hierarchy feature #4044.

Here is the list of features that I'm working on and their status , feel free to suggest others :

| Feature                                  | Implemented | Tested | Example                                       |
|------------------------------------------|:-----------:|--------|-----------------------------------------------|
| Def function IC                          |      X      |        | `def f = ???`                                 |
| Def function OC                          |      X      |        | `def f = ???`                                 |
| Named anonymous function IC              |      X      |        | `val f = () => ???`                           |
| Named anonymous function OC              |      X      |        | `val f = () => ???`                           |
| Named anonymous function with unnaply IC |             |        | `val (f1, f2) = ((() => ???), (() => ???))`   |
| Named anonymous function with unnaply OC |             |        | `val (f1, f2) = (( () => ???), (() => ???))`  |
| Primary constructor IC                   |      X      |        | `class Foo(x: Int)`                           |
| Primary constructor OC                   |      X     |        | `class Foo(x: Int)`                           |
| Secondary constructor IC                 |      X      |        | `class Foo(x: Int) { def this(x: Int) = ???}` |
| Secondary constructor OC                 |      X      |        | `class Foo(x: Int) { def this(x: Int) = ???}` |
| Case class constructor IC                |      X      |        | `case class Foo(x: Int)`                      |
| Case class constructor OC                |             |        | `case class Foo(x: Int)`                      |
| Workspace support                        |       X      |        |                                               |

(IC = Incoming calls, OC = Outgoing calls)


metals/src/main/scala/scala/meta/internal/metals/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,376 @@
...
+) extends SemanticdbFeatureProvider {
+
+  private var referencedPackages: BloomFilter[CharSequence] =
+    BloomFilters.create(10000)
+
+  case class IndexEntry(
+      id: BuildTargetIdentifier,
+      bloom: BloomFilter[CharSequence]
+  )
+  val index: TrieMap[Path, IndexEntry] = TrieMap.empty
```
tgodzik, line 60: Index doesn't seem to be used, no?

metals/src/main/scala/scala/meta/internal/metals/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,376 @@
...
+      .map { case (k, v) => (k, v.head._3, v.map(_._2)) }
+      .toList
+  }
+
+  def incomingCalls(
+      params: CallHierarchyIncomingCallsParams
+  ): List[CallHierarchyIncomingCall] = {
+    val source = params.getItem.getUri.toAbsolutePath
+
+    val info = params.getItem.getData.asInstanceOf[JsonObject]
```
tgodzik, line 269: You can convert this to a case class with `JsonParser` and `as` method

metals/src/main/scala/scala/meta/internal/metals/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,376 @@
...
+    ): List[(SymbolOccurrence, l.Range, l.Range)] = tree match {
+      case name: Name
+          if (parent.isDefined && definition
+            .positionOccurrences(source, name.pos.toLSP.getEnd, doc)
+            .flatMap(_.occurrence)
+            .exists(occ =>
+              occ.symbol == info.symbol && occ.role == Role.REFERENCE
+            )) ||
+            (name.parent.map(_.is[Term.Apply]).getOrElse(false) && definition
+              .positionOccurrences(source, name.pos.toLSP.getEnd, doc)
```
tgodzik, line 215: You could find all symbol references first and create a set out of it to jsut check here.

metals/src/main/scala/scala/meta/internal/metals/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,376 @@
...
+  }
+
+  def incomingCalls(
+      params: CallHierarchyIncomingCallsParams
+  ): List[CallHierarchyIncomingCall] = {
+    val source = params.getItem.getUri.toAbsolutePath
+
+    val info = params.getItem.getData.asInstanceOf[JsonObject]
+
+    semanticdbs.textDocument(source).documentIncludingStale match {
```
tgodzik, line 271: You might need to use ReferenceProvider to find all the documents that contain the symbol we are interested in. And then just iterate over them.

tgodzik (commented):
This is going pretty well! I left some initial comments, ignore me if I missed something,

metals/src/main/scala/scala/meta/internal/metals/CallHierarchyProvider.scala
riiswa: Do you think I should use the public `references method` and get all URIs ?

tgodzik: We could, or we could add a new method if that would be easier to handle. That method could just give the potential files based on Bloom filters

kpodsiad:
First of all, nice job @riiswa pushing this that far ;)
I see that there are no tests for call hierarchy and because of that I want to ask, how does yours workflow look like, how do you test this PR? [Unit tests](https://scalameta.org/metals/docs/contributors/getting-started#unit-tests) together with `pprint.log()` might reduce feedback loop significantly as they're much quicker than manual tests.

riiswa:
> First of all, nice job @riiswa pushing this that far ;) I see that there are no tests for call hierarchy and because of that I want to ask, how does yours workflow look like, how do you test this PR? [Unit tests](https://scalameta.org/metals/docs/contributors/getting-started#unit-tests) together with `pprint.log()` might reduce feedback loop significantly as they're much quicker than manual tests.

Indeed, there are no unit tests yet, I'm doing manual tests for the moment, but I'm starting to look at how to write a test suite, I might need some help/advice

ckipp01:
> Indeed, there are no unit tests yet, I'm doing manual tests for the moment, but I'm starting to look at how to write a test suite, I might need some help/advice

If you're in the office tomorrow I can sit with you for a bit and look at this if you want.

tgodzik (commented):
I commented a bit on the tests, do let me know if I misunderstood something.

Overall, in tests with put in usually what the user might see at the end, so I would modify it a bit, since this will make writing new tests easier.

We could additionally try to have some similar tests to the definition expect tests. Though that might require more work and the input projects might not contain enough examples to test out :thinking:

tests/unit/src/test/scala/tests/CallHierarchyLspSuite.scala
```diff
@@ -0,0 +1,133 @@
...
+       |package a
+       |
+       |object Main {
+       |  def a() {}
+       |  def <<b>>() { <<a>>() }
+       |  def c() { b() }
+       |  def d() { a() }
+       |}
+       |
+       |""".stripMargin,
```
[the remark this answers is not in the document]
riiswa: Good idea, but is it important to check the link between an invoker and where the function is invoked ? for example the function b call the first a() and not the second one.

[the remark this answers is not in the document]
tgodzik: I only see a single a definition, both d and b seems to be invoking it?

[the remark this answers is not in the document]
riiswa: Yes but for exemple for this case the response of the request will be something like :

```scala
List(
 CallHierarchyIncomingCall(from = definition of the function b, fromRanges= List(lsp.Range of the first a() call)),
   CallHierarchyIncomingCall(from = definition of the function d, fromRanges= List(lsp.Range of the second a() call))
)
```

I mean, should I check the link between the fields `from` and `fromRanges` of the CallHierarchyIncomingCall objects.

riiswa:
Hey @tgodzik @ckipp01 can you take a look, the tests are better constructed now :D

tests/unit/src/main/scala/tests/BaseCallHierarchySuite.scala
```diff
@@ -0,0 +1,181 @@
...
+  def assertPrepareOutgoingCalls(
+      input: String,
+      scalaVersion: Option[String] = None,
+  )(implicit loc: Location) = assertCallHierarchy(
+    outgoingCallGetter,
+    input,
+    scalaVersion,
+  )
+
+  def assertIncomingCalls(
```
tgodzik, line 160: Isn't this the same as `assertPrepareIncomingCalls`? `item` seems to be unused.

tests/unit/src/test/scala/tests/CallHierarchyLspSuite.scala
```diff
@@ -0,0 +1,71 @@
...
+           |
+           |""".stripMargin,
+        result("1"),
+      )
+    } yield ()
+  }
+
+  test("def-outgoing-call") {
+    for {
+      result <- assertPrepareOutgoingCalls(
```
tgodzik, line 38: Looks good! Though I think we can just used `assertOutgoingCalls` in both places. The same in `assertIncomingCalls`

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -255,6 +255,25 @@ final class ReferenceProvider(
     }
   }
 
+  private[metals] def pathsMightContainSymbol(
+      source: AbsolutePath,
+      isSymbol: Set[String],
+  ): Set[AbsolutePath] = {
+    buildTargets.inverseSources(source) match {
```
kpodsiad, line 262: Can we use `buildTargets.inverseSourcesBsp`? It's has been recently added because in some cases plain `inverseSources` might not be enough. It's more awkward  to handle since it brings future to the scope but I see that it's already present in `CallHierarchyProvider` so issue shouldn't be too big. 
`inverseSourcesBsp` was added in https://github.com/scalameta/metals/pull/3752, but I can't see followup issue with call migration, is there a chance that there isn't one? cc: @tanishiking

metals/src/main/scala/scala/meta/internal/metals/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,653 @@
...
+import scala.meta.Term
+import scala.meta.pc.CancelToken
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.meta.Decl
+import scala.meta.Type
+import scala.meta.Init
+import scala.meta.Template
+
+final case class CallHierarchyProvider(
```
kpodsiad, line 38: It's a huge file with a lot of stuff going on. What do think about partitioning it a bit? 
- File itself can be moved to separate package ( `scala.meta.internal.metals.callHierarchy` maybe? Camel case isn't the best one for packages but in this case IMO it's better than `call.hierarchy` or similar)
- `CanBuildCallHierarchyItem` can be extracted to file and made package private to not pollute namespace
- there are a few tuples (e.g. return type of `findOutgoingCallsSynthetics` method) - can we have a case classes for them with some scaladoc? It'll make methods like `groupIncomingCalls` more readable

What do you think?

metals/src/main/scala/scala/meta/internal/metals/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,653 @@
...
+        occurence
+      )
+    }
+
+  private implicit val selectTree2CanBuildCallHierarchyItem =
+    new CanBuildCallHierarchyItemBuilder[s.SelectTree] {
+      def apply(
+          selectTree: s.SelectTree
+      ): CanBuildCallHierarchyItem[s.SelectTree] = SelectTreeCHI(selectTree)
+    }
```
kpodsiad, line 73: It's kind of rule of thumb to:
- insert explicit types for implicits
- define implicits in companion object of typeclass (trait) or datatype for which you want to provide them. In this case they can be moved to the companion object of `CanBuildCallHierarchyItemBuilder`


Moreover, traits and abstract classes with only one abstract method are called Single Abstract Types (SAM Types) and can be implemented as it was a function

Using that knowledge, you can do:
```scala
  object CanBuildCallHierarchyItemBuilder {
    implicit val symbolOccurence2CanBuildCallHierarchyItem: CanBuildCallHierarchyItemBuilder[SymbolOccurrence] =
      occurence => SymbolOccurenceCHI(occurence)

    implicit val selectTree2CanBuildCallHierarchyItem: CanBuildCallHierarchyItemBuilder[s.SelectTree] =
      selectTree => SelectTreeCHI(selectTree)
  }
```

metals/src/main/scala/scala/meta/internal/metals/CallHierarchyProvider.scala
riiswa: Thanks for the advice, I finish reorganize the code, and adding some pieces of scaladoc.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+    )
+
+  /**
+   * Resolve incoming calls for a given call hierarchy item.
+   */
+  def incomingCalls(
+      params: CallHierarchyIncomingCallsParams,
+      token: CancelToken,
+  )(implicit ec: ExecutionContext): Future[List[CallHierarchyIncomingCall]] = {
+    val info = params.getItem.getData
```
ckipp01, line 292: We spoke about this in person, but maybe add a comment both here and where we set `itemSetData` and explain this so it's not confusing why we have to cast here.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+      .asInstanceOf[JsonElement]
+      .as[CallHierarchyItemInfo]
+      .get
+
+    for {
+      paths <-
+        if (!info.isLocal)
```
ckipp01, line 299: Add some comments in here.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+          references
+            .pathsMightContainSymbol(
+              params.getItem.getUri.toAbsolutePath,
+              info.symbols.toSet,
+            )
+        else Future.successful(Set.empty)
```
ckipp01, line 305:
```suggestion
        else Future.successful(Set.empty[AbsolutePath])
```
Nit but just so the type of `paths` isn't funky.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+
+  override def onDelete(file: AbsolutePath): Unit = ()
+
+  override def onChange(docs: TextDocuments, file: AbsolutePath): Unit = ()
+
+  /**
+   * Data that is preserved between a call hierarchy prepare and incoming calls or outgoing calls request.
+   *
+   * @param symbols The set of symbols concerned by the request.
+   * @param visited Symbols that are already visited to handle recursive cases.
```
ckipp01, line 60: Make sure scaladocs match the signature.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+        if (!info.isLocal)
+          references
+            .pathsMightContainSymbol(
+              params.getItem.getUri.toAbsolutePath,
+              info.symbols.toSet,
+            )
+        else Future.successful(Set.empty)
+      localPath = if (info.searchLocal) Set(workspace) else Set.empty
+      result <- Future.sequence(
+        (paths ++ localPath).toList
```
ckipp01, line 308: If possible let's try to abstract some of this out into function since we have a ton of nestsed logic in here and even some nested `Future.sequence`s.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+    for {
+      paths <-
+        if (!info.isLocal)
+          references
+            .pathsMightContainSymbol(
+              params.getItem.getUri.toAbsolutePath,
+              info.symbols.toSet,
+            )
+        else Future.successful(Set.empty)
+      localPath = if (info.searchLocal) Set(workspace) else Set.empty
```
ckipp01, line 306:
```suggestion
      localPath = if (info.searchLocal) Set(workspace) else Set.empty[AbsolutePath]
```

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+      result <- Future.sequence(
```
ckipp01, line 307: Also we should refactor this nested `Future.sequence`s. We end up with a `List[List[...` which is not what we want here.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+              )
+              callsSynthetics <- findOutgoingCallsSynthetics(
+                source,
+                doc,
+                root,
+              )
+            } yield calls ++ callsSynthetics
+          )
+          .getOrElse(Future.successful(Nil))
+          .flatMap(results =>
```
ckipp01, line 588: Same thing, let's extract some of this out and clean it up to read better.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+              .map(info => {
+                val item = new CallHierarchyItem(
+                  info.displayName,
+                  info.kind.toLSP,
+                  source.toURI.toString,
+                  range,
+                  itemRange.toLSP,
+                )
+
+                val symbols = (Set(info.symbol) union (
```
ckipp01, line 100: Alright same thing here, reading through this we have a ton of deeply nested stuff. Let's try to extract some stuff out to improve on the readability and therefore maintainability.

ckipp01, line 100:
```suggestion
                val symbols = (Set(info.symbol).union(
```

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+
+      tree match {
+        case name: Name
+            if !isTypeDeclaration(name) && definition
+              .positionOccurrences(source, name.pos.toLSP.getEnd, doc)
+              .exists(
+                _.occurrence.exists(occ =>
+                  occ.role.isReference && info.symbols.contains(occ.symbol)
+                )
+              ) =>
```
ckipp01, line 210: Same thing here with some of the guards even if they could be extracted out into a function this whole thing would read easier.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+    FindIncomingCallsResult.group(search(root, None, None))
+  }
+
+  private def findIncomingCallsSynthetics(
+      source: AbsolutePath,
+      doc: TextDocument,
+      info: CallHierarchyItemInfo,
+  ): List[FindIncomingCallsResult] =
+    FindIncomingCallsResult.group(
+      doc.synthetics
```
ckipp01, line 261: Let's extract this value out and then give it to `group` after.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+  )(implicit
+      ec: ExecutionContext
+  ): Future[Option[(SymbolOccurrence, AbsolutePath, TextDocument)]] =
+    (if (!symbol.isLocal)
+       references
+         .pathsMightContainSymbol(
+           source,
+           Set(symbol),
+         )
+     else Future.successful(Set(source)))
```
ckipp01, line 372: Just a nit, my preference is typically to do the operations you need, and then at the end if it needs to be returned in a `Future`, then do this at the end.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+                    rso.occurrence
+                      .map(occ => findDefinitionOccurence(source, occ.symbol))
+                  )
+              )
+              .map(
+                _.flatten
+                  .sortBy(
+                    _._1.symbol.length * -1
+                  )
+                  .headOption // The most specific occurrence is the longest
```
ckipp01, line 419: I find this pretty hard to grok without spending time and digging into it. Maybe refactoring this either into a for comprehension or something else would be easier to read.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyProvider.scala
```diff
@@ -0,0 +1,619 @@
...
+                        .encloses(range) && getSpecifiedOrFindDefinition(
+                        trees.findLastEnclosingAt(source, range.toLSP.getStart),
+                        Some(definition),
+                      ).exists(_._1 == definition) =>
+                    lazy val mightBeCaseClassConstructor =
+                      """\.apply\(\)\.$""".r
+
+                    for {
+                      definitionInfo <- findDefinitionOccurence(source, symbol)
+                        .flatMap {
```
ckipp01, line 501: Same thing as pretty much all the other comments, let's extract some of this stuff out. We have a _ton_ going on in here which makes it hard to read and follow.

ckipp01 (commented):
Great job so far, it's really cool to see this working. Now that everything seems to be working pretty well, as we talked about in person, let's start focusing on refactoring and making this all a bit easier to read by breaking things out. This will really help with the long-term readability and maintainability. 🚀

metals/src/main/scala/scala/meta/internal/metals/ReferenceProvider.scala
```diff
@@ -255,6 +257,29 @@ final class ReferenceProvider(
...
+            val result = for {
+              (path, entry) <- index.iterator
+              if allowedBuildTargets.contains(entry.id) &&
+                isSymbol.exists(entry.bloom.mightContain)
+              sourcePath = AbsolutePath(path)
+              if sourcePath.exists
+            } yield sourcePath
+            result.toSet
+        }
+      )
```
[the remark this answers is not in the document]
ckipp01: I think you missed this one.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/RealRoot.scala
```diff
@@ -0,0 +1,23 @@
...
+import scala.meta.Name
+
+/**
+ * Where to start looking for occurrences.
+ * When using the findLastEnclosingAt method, it often happens that you get Name, not the root of the definition,
+ * this trait allows to store the root and the potential name of a definition.
+ */
+private[callHierarchy] sealed trait RealRoot {
+  def root: Tree
+  def mayName: Option[Name]
```
ckipp01, line 13: I find `mayName` a bit confusing. What if instead we just made this:

```
name: Option[Name]
```
And then change `NamedRealRoot` to

```
private[callHierarchy] case class NamedRealRoot(root: Tree, private val _name: Name)
    extends RealRoot {
  val name = Some(name)
}
```

So then you'd just always try to access `name` which is either there or not.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyHelpers.scala
```diff
@@ -0,0 +1,179 @@
...
+      case _ =>
+        getSpecifiedOrFindDefinition(
+          tree.parent,
+          specified,
+          Some(tree),
+        )
+    }
+
+    (specified
+      .collect { case tree if from.contains(tree) => UnamedRealRoot(tree) })
```
ckipp01, line 90:
```suggestion
    specified
      .collect { case tree if from.contains(tree) => UnamedRealRoot(tree) }
```

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/FindCallsResult.scala
```diff
@@ -0,0 +1,158 @@
...
+            definitionNameRange,
+            _,
+          ) :: _ =>
+        FindIncomingCallsResult(
+          occurence,
+          definitionNameRange,
+          result.flatMap(_.fromRanges),
+        )
+      case _ =>
+        throw new IllegalArgumentException("Can't apply group on a empty list.")
```
ckipp01, line 96: I don't think we should throw here. If we hit this maybe log a warning if you want and then return an empty list.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/FindCallsResult.scala
```diff
@@ -0,0 +1,158 @@
...
+          ) :: _ =>
+        FindOutgoingCallsResult(
+          occurence,
+          definitionNameRange,
+          result.flatMap(_.fromRanges),
+          source,
+          doc,
+        )
+      case _ =>
+        throw new IllegalArgumentException("Can't apply group on a empty list.")
```
ckipp01, line 156: Same as above, let's not throw.

ckipp01 (commented):
Great job refactoring this @riiswa, it's way way easier to read now. For me it's just a few nits left at this point, but let's also get some input from the others.

tgodzik:
> Great job refactoring this @riiswa, it's way way easier to read now. For me it's just a few nits left at this point, but let's also get some input from the others.

I will take a look next Monday :rocket: but looks like great work!

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
```diff
@@ -0,0 +1,148 @@
...
+          range,
+          selectionRange.toLSP,
+          getDetail(
+            getSignatureFromHover(hover).getOrElse(""),
+            visited,
+            symbol,
+          ),
+          CallHierarchyItemInfo(
+            // When we search the call hierarchy of a class, we also need to search the symbol of the primary constructor
+            (Set(symbol) ++ getClassContructorSymbol(info)).toArray,
```
kpodsiad, line 106: This might be extracted to function (probably incorporated into `getClassContructorSymbol` + rename will be necessary).

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
```diff
@@ -0,0 +1,148 @@
...
+      info.signature
+        .asInstanceOf[s.ClassSignature]
+        .declarations
+        .flatMap(
+          _.symlinks.find(_.endsWith("`<init>`()."))
+        )
+        .toList ++ Option.when(info.isCase)(
+        info.symbol.replace("#", ".apply().")
+      ) // handling case class constructors
+    else Nil
```
kpodsiad, line 54: Like I mentioned earlier, this function might return array which doesn't need any modifications after. Also, I'm not a fan of `asInstanceOf` casts. Function fro suggestion needs better name.
```suggestion
// When we search the call hierarchy of a class, we also need to search the symbol of the primary constructor
  private def getClassContructorSymbol(
      info: s.SymbolInformation
  ): Array[String] =
    info.signature match {
      case ClassSignature(_, _, _, declarations) =>
        val constructor = declarations
          .flatMap(
            _.symlinks.find(_.endsWith("`<init>`()."))
          )

        // for case class apply is like a constructor
        val applySymbol = Option.when(info.isCase)(
          info.symbol.replace("#", ".apply().")
        )
        Set(Some(info.symbol), constructor, applySymbol).flatten.toArray
      case _ =>
        Array(info.symbol)
    }
```

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
```diff
@@ -0,0 +1,148 @@
...
+      )
+    else info.displayName
+
+  private def getDetail(
+      signature: String,
+      visited: Array[String],
+      symbol: String,
+  ): String =
+    (if (visited.dropRight(1).contains(symbol)) icons.sync + " "
+     else "") + signature
```
kpodsiad, line 71: Can you add comment which will explain `visited.dropRight(1).contains`?

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
```diff
@@ -0,0 +1,148 @@
...
+
+  private def getSignatureFromHover(hover: Option[l.Hover]): Option[String] =
+    (for {
+      hover <- hover
+      hoverContent <- hover.getContents().asScala.toOption
+      `match` <- """Symbol signature\*\*:\n```scala\n(.*)\n```""".r
+        .findFirstMatchIn(hoverContent.getValue)
+    } yield `match`.group(1))
```
kpodsiad, line 79: Parenthesis around for aren't needed. I wonder, call to `group` might return null, but we are certain that `match` will have that group defined. It can be wrapped into an option but it seems like a redundant "should never happen" case 🤔 

Anyway, I think `""Symbol signature\*\*:\n```scala\n(.*)\n```""".r` can be extracted and enhanced with a good comment.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemInfo.scala
```diff
@@ -0,0 +1,16 @@
+package scala.meta.internal.metals.callHierarchy
+
+/**
+ * Data that is preserved between a call hierarchy prepare and incoming calls or outgoing calls request.
+ *
+ * @param symbols The set of symbols concerned by the request.
+ * @param visited Symbols that are already visited to handle recursive cases.
+ * @param isLocal Do we have to search for symbols in the whole workspace.
+ * @param searchLocal Do we have to search for symbols locally.
```
kpodsiad, line 9: `isLocal` and `searchLocal`, if I understand correctly models 3 cases? 
- search only local scope
- search the whole workspace
- search the whole workspace except local scope

Am I correct?

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
```diff
@@ -0,0 +1,148 @@
...
+            symbol,
+          ),
+          CallHierarchyItemInfo(
+            // When we search the call hierarchy of a class, we also need to search the symbol of the primary constructor
+            (Set(symbol) ++ getClassContructorSymbol(info)).toArray,
+            visited,
+            symbol.isLocal,
+            symbol.isLocal || source.isDependencySource(
+              workspace
+            ) || buildTargets.inverseSources(source).isEmpty,
```
kpodsiad, line 111: Can you explain why do we need to search in local scope when the last 2 conditions are met?

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/FindCallsResult.scala
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /** Aggregate results of incoming calls finder by grouping the fromRanges. */
+  def group[T <: FindCallsResult[_, T]](
+      results: List[T]
+  ): List[T] =
+    results.groupBy(_.occurence).values.toList.collect {
+      case result @ first :: _ =>
+        first.aggregate(clearRanges(result.flatMap(_.fromRanges)))
+    }
```
kpodsiad, line 151: This needs more explanation what's going on. If you answer and explain to me why this is needed in this thread, we can come up with some scaladoc together ;)

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/FindCallsResult.scala
```diff
@@ -0,0 +1,152 @@
...
+ * @param source AbsolutePath where the item that is called is defined.
+ * @param doc TextDocument where the item that is called is defined.
+ */
+private[callHierarchy] case class FindOutgoingCallsResult(
+    occurence: SymbolOccurrence,
+    definitionNameRange: lsp4j.Range,
+    fromRanges: List[lsp4j.Range],
+    source: AbsolutePath,
+    doc: TextDocument,
+)(implicit ec: ExecutionContext)
```
kpodsiad, line 109: Execution context should be rather argument of `toLSP`. BTW, don't we have `toLsp` rather than `toLSP` convention?

kpodsiad (commented):
There is a lot of stuff happening in this PR so I'm trying to do review incrementally day by day ;)

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemInfo.scala
riiswa: Yes, this is inspired from code in ReferenceProvider, so the associated comment is : 

```scala
/* search local in the following cases:
     * - it's local symbol
     * - it's a dependency source.
     *   We can't search references inside dependencies so at least show them in a source file.
     * - it's a standalone file that doesn't belong to any build target
     */
```

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
riiswa: Same as my previous comment, this is inspired from ReferenceProvider

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/FindCallsResult.scala
riiswa: If you search for occurrences of `toLSP`, we have both. I can do an another PR to unify if necessary

riiswa: #4330

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
```diff
@@ -0,0 +1,161 @@
...
+  ): Array[String] =
+    info.signature match {
+      case s.ClassSignature(_, _, _, declarations) =>
+        val constructor = declarations
+          .flatMap(
+            _.symlinks.find(_.endsWith("`<init>`()."))
+          )
+        // for case class apply is like a constructor
+        val applySymbol = Option.when(info.isCase)(
+          info.symbol.replace("#", ".apply().")
```
tgodzik, line 53: What about user defined apply? The symbol might be different, so you would need to search for all apply :thinking:

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+  /**
+   * Get the display name from the symbol information.
+   * For constructor, strip the symbol to the class name,
+   * otherwise return symbols's display name.
+   */
+  private def getName(info: s.SymbolInformation): String =
+    if (info.isConstructor && info.displayName == "<init>")
+      info.symbol.slice(
```
tgodzik, line 67: You might be able to use the parser similar to `SymbolAlternatives` in ReferenceProvider

tgodzik, line 67: You might be able to use the parser similar to `SymbolAlternatives` in ReferenceProvider

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
```diff
@@ -0,0 +1,161 @@
...
+        info.symbol.lastIndexOf("/") + 1,
+        info.symbol.lastIndexOf("#"),
+      )
+    else info.displayName
+
+  // Regular expression to capture the group inside hovers that indicate the symbol signature.
+  private val symbolSignatureExtractor =
+    """Symbol signature\*\*:\n```scala\n(.*)\n```""".r
```
tgodzik, line 75: We could potentially have an additional method or just boolean option for presentation compiler to just get the signature, which would not require matching here.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemInfo.scala
```diff
@@ -0,0 +1,16 @@
...
+ */
+private[callHierarchy] final case class CallHierarchyItemInfo(
+    symbols: Array[String],
+    visited: Array[String],
+    isLocal: Boolean,
+    searchLocal: Boolean,
```
tgodzik, line 15: Isn't this two conditions related? Maybe a sealed trait with object would work here,

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/FindCallsResult.scala
riiswa: Basically, when we search for calls, we needs to find the ranges of the calls (the `fromRanges` attribute), and after we need to group the result that have the same CallHierarchyItem.

Example: 

```scala
def foo  = {
  a() //1
  a() //2
}
```
If we compute the outgoings call of `foo` We will have as results , something like : 
```
[
{to: CallHierarchyItem that represent the definition of a(), fromRanges: [range of the first a() call] }, 
{to: CallHierarchyItem that represent the definition of a(), fromRanges: [range of the second a() call] },
]
```
So we need to group them to have:
```
[{to: CallHierarchyItem that represent the definition of a(), fromRanges: [range of the first a() call, range of the second a() call]}]
```

Finally the goal of clearRanges is to avoid the potential duplication of calls caused by the synthetics symbols by just make sure to not have a range that encloses an other.

It that clear ?

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
riiswa: Yeah I will try, this was not my first choice because the method that handle that is almost 100 lines of codes and a little bit hard to understand, but I will dive on it

tgodzik: If the problem is the additional docs + expression type then it should be simple to disable with a flag

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyHelpers.scala
```diff
@@ -0,0 +1,178 @@
...
+        case d: Defn.Def =>
+          d.decltpe.contains(tree) || d.tparams.contains(tree)
+        case _: Type.Bounds => true
+        case t @ (_: Type | _: Name | _: Init) => isTypeDeclaration(t)
+        case _ => false
+      }
+
+  /**
+   * Go up in the tree to find a specified tree or find any definition.
+   * If the specified item is founded this function will return the `UnamedRealRoot(specified)`,
```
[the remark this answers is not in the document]
riiswa: I fixed it, I seperate the two purpose of this function :) It's better now

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemInfo.scala
kpodsiad: related comment https://github.com/scalameta/metals/pull/4115/files#r958979482

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
riiswa: The apply method of case class is special, because when we click on the name of the class, we want to take in consideration the constructor. For example:

```scala
case class Fo@@o(x: Int)
...
<<Foo>>(0)
```
`.apply().` is a special symbol in this case that appear in synthetics. The feature work well with user defined feature, I just add a test for it.

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
riiswa: Can I have more information about what do you mean by use the parser, I read the code inside `SymbolAlternatives`, but I think that I missing smth.

tgodzik: Ach, sorry. By doing:
```
Symbol(info.symbol) 
```
you will get an AST of the semanticdb name such as:
```
Symbols.Global(
            Symbols.Global(owner.value, Descriptor.Type(obj)),
            Descriptor.Term(name),
          )
```
you can use that to check if something is a constructor and just create  new name without <init>

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyItemBuilder.scala
tgodzik: ok cool!

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/OutgoingCallsFinder.scala
```diff
@@ -0,0 +1,252 @@
...
+        FindOutgoingCallsResult(
+          occurence,
+          tree.pos.toLsp,
+          List(fromRange),
+          source,
+          doc,
+        )
+      )
+
+    def isSameDefinition(
```
tgodzik, line 64: Why is this needed? Could you add a sample in a comment? Is it to match the tree with a symbol occurrence?

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/OutgoingCallsFinder.scala
```diff
@@ -0,0 +1,252 @@
...
+          pos: Position
+      ): Future[Option[DefinitionInformation]] = {
+        val potentialsDefinition = Future.sequence(
+          definition.positionOccurrences(source, pos, doc).map {
+            case ResolvedSymbolOccurrence(_, Some(occurence)) =>
+              findDefinitionOccurence(source, occurence.symbol)
+            case _ => Future.successful(None)
+          }
+        )
+        potentialsDefinition.map(_.flatten.maxByOption(_.symbolLength))
```
tgodzik, line 137: Why sort it by length?

tgodzik (approved):
I left a couple of comments, but nothing blocking from my side. It's not on any critical path, so should be save to merge. I will leave it up to @ckipp01 if that is ok?

tgodzik:
Also, forgot to mention... Great work @riiswa !! This was a tough one from what I am seeing

metals/src/main/scala/scala/meta/internal/metals/callHierarchy/CallHierarchyHelpers.scala
```diff
@@ -0,0 +1,177 @@
...
+    else
+      extractNameFromMember(from) match {
+        case result @ Some(_) =>
+          result
+        case None =>
+          advancedFindDefinition(from)
+      }
+  }
+
+  def containsDuplicates[T](visited: Seq[T]): Boolean =
```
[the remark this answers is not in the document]
riiswa: I use this method in the case of recursive call, to avoid infinite loop in call hierarchy request. I will add a comment to explain it.

tgodzik:
@riiswa @ckipp01 I think we should be able to merge, the leftover comments are not that important, but let's fix the scalafix job, maybe?

riiswa:
Hey @tgodzik 😄 , let's merge it tomorrow, I will add some comments/scaladoc in the afternoon 😉

tgodzik:
> Hey @tgodzik smile , let's merge it tomorrow, I will add some comments/scaladoc in the afternoon wink

Great!

ckipp01:
> @riiswa @ckipp01 I think we should be able to merge, the leftover comments are not that important, but let's fix the scalafix job, maybe?

Sounds good to me! I'm on holiday atm and haven't been spending a lot of time on the computer, so feel free to merge. Great job on getting this across the board @riiswa!

riiswa:
Ready for merging 😄

tgodzik (approved):
Let's do it!

PR #4116, opened 2022-07-07
fix: do not start debug session for test explorer if projects contains errors

kpodsiad:
Previously, when project has errors and user clicked run test button they was spinning due to debug session not started. Now errors are checked and in case of any errors debug session is not started.
Plays well with https://github.com/scalameta/metals-vscode/pull/1053

Demo:

sbt-playground-1657200325163

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
@@ -357,18 +357,15 @@ class DebugProvider(
    * session to not actually work, but fail silently.
    */
   def ensureNoWorkspaceErrors(
-      params: DebugSessionParams
-  )(implicit ec: ExecutionContext): Future[b.DebugSessionParams] = {
+      buildTargets: Seq[BuildTargetIdentifier]
+  )(implicit ec: ExecutionContext): Future[Unit] = {
+    val hasErrors = buildTargets.exists { target =>
+      buildClient.buildHasErrors(target)
+    }
     val result =
-      if (
-        params.getTargets().asScala.toList.exists { target =>
-          buildClient.buildHasErrors(target)
-        }
-      ) {
-        Future.failed(WorkspaceErrorsException)
-      } else {
-        Future.successful(params)
-      }
+      if (hasErrors) Future.failed(WorkspaceErrorsException)
+      else Future.unit
+
     result.failed.foreach(reportErrors)
     result
   }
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -554,26 +551,35 @@ class DebugProvider(
...
+          Future.successful(debugSession)
+        case None =>
+          val error = BuildTargetNotFoundException(request.target.getUri)
+          reportErrors(error)
+          Future.failed(error)
+      }
+
+    for {
+      _ <- compilations.cascadeCompile(List(request.target))
+      _ <- ensureNoWorkspaceErrors(List(request.target))
```
kpodsiad, line 580: I think some error message will be nice, now only problem stab is being silently focused. `ensureNoWorkspaceErrors` calls `reportErrors`, maybe it should also show message>

kpodsiad: Oh, sorry. If you'll very careful, there is status displayed on status bar between `Scala` and `GoLive` `Errors in workspace` but it's quickly covered by information about compilation.

dos65: I'm wondering if this might be fixed only on client side.
I remember there where a similar problem with ctrl + F5 - https://github.com/scalameta/metals-vscode/pull/904

kpodsiad: > However, after the second look, I see that here we trigger cascadeComplie but only for testSelection.
> How is it different from running main class/ other test requests?

Good question. I used `cascadeComplie` because it seemed to have more sense. If I'm about to run class from particular projects it (class) may depends on another class from another module. Now I'm wondering if `cascadeCompile` makes sense. It sends request for dependency leaves if I'm correct 🤔

kpodsiad: `compile` works so I'd stick to it

tgodzik (approved):
LGTM!

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
dos65: I meant a different thing :smiling_face_with_tear:  not compile vs cascadeCompile.

I meant - here we have a bunch of debugProvider calls:
https://github.com/kpodsiad/metals/blob/72a87ae81b7d6a12c5a9ebd6bdfeff3fd2dd5c8f/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1863-L1884

For me it looks suspicious that in this PR we trigger compilation on server side only in case of `resolveTestSelectionParams`.
I think we should either trigger compilation for all cases on server side or always save file (which triggers compilation) on client side. That's is what I was about :smile_cat:

kpodsiad: I think you meant both `compile vs cascadeCompile` and the latter thing. I didn't want to touch something that is more or less working now and I focused on particular issue. However, I should touch this code in the near future so i'll try to address this issue then

PR #4118, opened 2022-07-07
add scala 3.2.0-RC2

dos65:


tgodzik (changes requested):
You might need to also ignore the member2 test in CompletionInterpolatorSuite

The fix will land in 3.2.1

tgodzik (approved):
LGTM!

PR #4124, opened 2022-07-08
feature: Add a way to turn on debug logging and fix scalafix warmup

tgodzik:
Seems the main idea behind scribe is to make it easy to define in the code, so I decided to add an new metals property.

One issue is that Flyway debug logs still seem to end up in whichever handler is defined first, but this should not be an issue in most cases, since those logs are not super interesting.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
@@ -48,31 +48,38 @@ case class ScalafixProvider(
   // Warms up the Scalafix instance so that the first organize imports request responds faster.
   def load(): Unit = {
     if (!Testing.isEnabled) {
+      val tmp = workspace
+        .resolve(Directories.tmp)
+        .resolve(s"Main${Random.nextLong()}.scala")
       try {
         val targets =
           buildTargets.allScala.toList.groupBy(_.scalaVersion).flatMap {
             case (_, targets) => targets.headOption
           }
-        val tmp = workspace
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
```diff
@@ -48,14 +48,14 @@ case class ScalafixProvider(
   // Warms up the Scalafix instance so that the first organize imports request responds faster.
   def load(): Unit = {
     if (!Testing.isEnabled) {
+      val tmp = workspace
```
ckipp01, line 51: What was actually happening here? How come moving it out of the `try` fixes it?

ckipp01 (approved):
Ah nice, I always wondered about debug for scribe.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
tgodzik: Not 100 percent sure 😅 it seems there was a race, but how I have no idea 😑

darkfrog26:
Let me know if there's something I could add to Scribe that would assist in making this easier.

metals/src/main/scala/scala/meta/internal/metals/ScalafixProvider.scala
tgodzik: I just added additional checks to see if evaluation indeed finishes successfully and it seems it does. There might be some magic behind the finally block.

PR #4125, opened 2022-07-08
support for partial function

zmerr:


metals/src/main/scala/scala/meta/internal/metals/codeactions/FlatMapToForComprehensionCodeAction.scala
```diff
@@ -448,6 +491,57 @@ class FlatMapToForComprehensionCodeAction(
...
+      case _: Pat.Wildcard => None // since it does not make sense to put
+      // Term.PlaceHolder on the right side of an assignment
+      // though perhaps we need to generate a name for the wildcard
+      case Pat.Xml(parts, args) =>
+        val termArgs = args.map(turnPatToTerm)
+        if (termArgs.forall(_.isDefined))
+          Some(Term.Xml(parts, termArgs.flatten))
+        else None
+      case _ => None
+    }
```
dos65, line 542: Similar to the previous comment:
```scala
pat match {
      case Pat.Bind(lhs, _) => turnPatToTerm(lhs)
      case Pat.Tuple(pats) =>
        val terms = pats.map(turnPatToTerm)
        if (terms.forall(_.isDefined))
          Some(Term.Tuple(terms.flatten))
        else None
      case Pat.Typed(lhs, _) => turnPatToTerm(lhs)
      case Pat.Var(name) => Some(name)
      case name: Term.Name => Some(name)
      case _ => None
}
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/FlatMapToForComprehensionCodeAction.scala
```diff
@@ -246,29 +248,66 @@ class FlatMapToForComprehensionCodeAction(
...
+      case _: Pat.Repeated => false
+      case _: Term.Select => false
+      case _: Pat.SeqWildcard => false
+      case Pat.Tuple(pats) => pats.forall(isSimple)
+      case _: Pat.Typed => true
+      case _: Pat.Var => true
+      case _: Pat.Wildcard => false
+      case Pat.Xml(_, args) => args.forall(isSimple)
+      case _ => false
+    }
```
dos65, line 277: There is no need to explicitly cover all possible pats(for example xml) - only these that occurs in tests.
I reduced it locally to and tests pass:
```scala
    pat match {
      case Pat.Tuple(pats) => pats.forall(isSimple)
      case _: Pat.Var => true
      case _: Term.Name => true
      case _: Pat.Typed => true
      case _ => false
    }
```
I'm not sure if all that I left is needed. Could you double check that?

dos65 (commented):
Nice work!
I left a few comments to simplify the code

metals/src/main/scala/scala/meta/internal/metals/codeactions/FlatMapToForComprehensionCodeAction.scala
vzmerr: Hi @dos65 Thank you so much for the review. I have not included the extra cases which are covered in the tests, but they were working on scastie. I mean like XML. Do I add them to the tests?

For [example](https://scastie.scala-lang.org/vzmerr/tPvC34B4RM2EIywhJlcLRg/10) :
```
val a = <p> here </p>
val b = <p> over there </p>

val c = for {
  <p>{value}</p> <- List(a, b)
} yield value

println(c)
```

prints `List( here ,  over there )`

dos65: oh! nice.
So, then I think it would better to cover the cases that are valid and supported by tests.

And anyway these blocks requires some refactoring. Put all valid option at the top and turn all options that returns false   into a single last wildcard: `case _ => false`

tgodzik (approved):
LGTM! We might find some more bugs along the line when users try to use it in some hard to imagine scenarios :sweat_smile:

PR #4127, opened 2022-07-08
bugfix: Save fingerprints between restarts

tgodzik:
Previously, if a user reloaded a broken workspace they would keep getting the dreaded "Could not load snapshot text for". Now, we save the fingerprints between restarts to improve the probability of finding the correct text, which corresponds to a semanicdb document.

It shouldn't save too much, since fingerprints are only updated for open files and we clean it after each reload.

### DIFF
--- a/metals/src/main/resources/db/migration/V4__Fingerprints.sql
+++ b/metals/src/main/resources/db/migration/V4__Fingerprints.sql
@@ -1,0 +1,7 @@
+-- Fingerprints saved between invocations
+create table fingerprints(
+  path varchar not null,
+  text varchar not null,
+  md5 varchar not null,
+  id int auto_increment unique
+);
--- a/metals/src/main/scala/scala/meta/internal/metals/Fingerprints.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Fingerprints.scala
@@ -1,0 +1,71 @@
+package scala.meta.internal.metals
+
+import java.nio.file.Paths
+import java.sql.Connection
+import java.sql.PreparedStatement
+
+import scala.meta.internal.metals.Fingerprint
+import scala.meta.internal.metals.JdbcEnrichments._
+import scala.meta.io.AbsolutePath
+
+/**
+ * Wrapper around the fingerprints sql table.
+ *
+ * It's needed to save the last known state of the workspace files to
+ * calculate differences between current state and semanticdb view of the file.
+ *
+ * We need to save it between restarts in case user reloaded a workspace with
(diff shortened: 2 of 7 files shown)


metals/src/main/resources/db/migration/V4__Fingerprints.sql
```diff
@@ -0,0 +1,7 @@
+-- Fingerprints saved between invocations
+create table fingerprints(
+  path varchar not null,
+  text varchar not null,
+  md5 varchar not null,
```
tanishiking, line 5: Why do we need both base64 and md5 encoded content? (if I understand correctly)
I guess there's no point encrypting the content with md5, since now we have base64 encoded one?

tanishiking: Maybe this is because we are using Md5Fingerprint class? Maybe we should refactor in another PR not to use md5 🤔

tanishiking (approved):
Overall, LGTM! but I left some quetions around content hashing.

metals/src/main/resources/db/migration/V4__Fingerprints.sql
tgodzik: Maybe base64 is no needed in reality, just felt like it's a bit cleaner to save it this way. I can revert that.

As for MD5, we still need it since that is much better suited for a hash, base64 can be quite long.

kpodsiad: Base64 is just an encoding which size is probably much longer than original text, where md5 is a hash function which converts content to always (I think) 16 bytes.

PS. Base64 is not an encryption :D

tgodzik: > Base64 is just an encoding which size is probably much longer than original text
Somehow I thought that it's shorter, only after thinking on this a bit a realised that was not a great idea 😅

kpodsiad: About length - https://stackoverflow.com/a/13378842

tanishiking: Ah, if we don't need to decode to the original text, we can keep md5 :)
I was mentioning md5 as encrypt, not base64

tgodzik: Removed the usage of Base64. It was needed at all indeed.

tanishiking (approved):
LGTM!

PR #4129, opened 2022-07-10
chore: refactor update test cases

kpodsiad:
* check if md5 checksum of file changed
* use batched function to prevent race conditions
* additionally, remove unused code


tgodzik (approved):
LGTM!

PR #4130, opened 2022-07-11
[docs] Update Maven integration launcher to 2.13

keynmol:
I was confused why the command from the docs wasn't working at all

What complicated things is that `maven-bloop` is somehow impossible to find on https://mvnrepository.com, even though it's indexed on https://index.scala-lang.org/scalacenter/bloop/artifacts/maven-bloop/1.5.2


tgodzik (approved):
Thanks! I totally forgot to update this one

Issue #4059, opened 2021-08-24
Append parameter names in method for simple types for Scala3


dgalichet:
### Is your feature request related to a problem? Please describe.

This feature is related to https://github.com/scalameta/metals-feature-requests/issues/184 that proposes to show parameter names for simple types

It could be nice to have a code action to append these parameters directly on the source code (according to @tgodzik ) it should be pretty simple to do)

### Describe the solution you'd like

```scala
def easyMethod(boolParam: Boolean) = ???
easyMethod(true)
```

After invoking code action, the code should be:
```scala
easyMethod(boolParam = true)
```

### Describe alternatives you've considered

the alternative is to show parameter only, without inserting it as described in https://github.com/scalameta/metals-feature-requests/issues/184

### Search terms

named parameters decorations inferred information

tanishiking:
Are you interested in working on Scala3 support @camgraff ? If you're interested in it, let me assign it to you (you don't have to be in rush :)

PR #4050, opened 2022-06-21
chore: change formatting of trailing commas

kpodsiad:
Just a proposition to reduce comma noise in PR's. (for example https://github.com/scalameta/metals/pull/4049/files)

I decided to create PR instead of creating discussion so everyone can look at changes and decide if this is cool or not. I think [multiple](https://scalameta.org/scalafmt/docs/configuration.html#trailing-commas-multiple) is the most sensible option.

PR #4131, opened 2022-07-11
Add convertToNamedParameters support for Scala 3

jkciesluk:
Closes https://github.com/scalameta/metals/issues/4059
Add Scala 3 support for https://github.com/scalameta/metals/pull/3971

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
@@ -5,18 +5,15 @@ import scala.concurrent.Future
 
 import scala.meta.Term
 import scala.meta.Tree
-import scala.meta.internal.metals.BuildTargets
 import scala.meta.internal.metals.CodeAction
 import scala.meta.internal.metals.MetalsEnrichments._
-import scala.meta.internal.metals.ScalaVersions
 import scala.meta.internal.metals.ServerCommands
 import scala.meta.internal.parsing.Trees
 import scala.meta.pc.CancelToken
 
 import org.eclipse.{lsp4j => l}
 
-class ConvertToNamedArguments(trees: Trees, buildTargets: BuildTargets)
-    extends CodeAction {
+class ConvertToNamedArguments(trees: Trees) extends CodeAction {
 
   import ConvertToNamedArguments._
   override val kind: String = l.CodeActionKind.RefactorRewrite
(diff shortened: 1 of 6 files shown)


metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
```diff
@@ -44,47 +41,40 @@ class ConvertToNamedArguments(trees: Trees, buildTargets: BuildTargets)
   ): Future[Seq[l.CodeAction]] = {
 
     val path = params.getTextDocument().getUri().toAbsolutePath
-    val isScala3 = buildTargets
-      .scalaVersion(path)
-      .map(ScalaVersions.isScala3Version)
-      .getOrElse(false)
-    if (isScala3) Future.successful(Nil)
-    else {
```
tanishiking, line 52: :+1:

tanishiking (changes requested):
Thank you for working on this @jkciesluk !

This implementation converts all Apply to named arguments instead of selected Apply, we have to convert only the selected one (in the following case, we should convert only the arguments of `foo` method).

named-convert

---

Also, we would need to copy some test suite from https://github.com/scalameta/metals/blob/ec89eac187c5c9e1358a8229ebac85b64a71a8a9/tests/unit/src/test/scala/tests/codeactions/ConvertToNamedArgumentsLspSuite.scala to https://github.com/scalameta/metals/blob/ec89eac187c5c9e1358a8229ebac85b64a71a8a9/tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala (I'm wondering we can cross run those LSP test 🤔  but not this time)

mtags/src/main/scala-3/scala/meta/internal/pc/ConvertToNamedArgumentsProvider.scala
```diff
@@ -0,0 +1,51 @@
...
+                case ((arg, index), param) if argIndices.contains(index) => {
+                  val position = arg.sourcePos.toLSP
+                  position.setEnd(position.getStart())
+                  new l.TextEdit(position, s"${param.name.show} = ")
+                }
+              }
+              .foreach(result.addOne(_))
+          case t =>
+            traverseChildren(t)
+    traverser.traverse(tree)(using newctx)
```
tanishiking, line 48: Instead of traversing the whole tree, it would be useful to check the enclosing tree (of the given position (via param)) using `Interactive.pathTo` or `Interactive.enclosingTree`

jkciesluk: Hi, thank you for help! 
Using `pathTo` works. I've also copied tests to https://github.com/scalameta/metals/blob/ec89eac187c5c9e1358a8229ebac85b64a71a8a9/tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala and extended them by testing multiple Apply.

tanishiking (approved):
LGTM!

tanishiking:
You may have to rebase on `main` and run scalafmt for https://github.com/scalameta/metals/pull/4050

tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
```diff
@@ -328,6 +338,264 @@ class Scala3CodeActionLspSuite
     expectNoDiagnostics = false,
   )
 
+  check(
```
tgodzik, line 341: We can really just add one test (first one looks good), since we only want to test the branch when we have Scala 3 dialect.  All the other cases are already handled in the ConvertToNamedArgumentsLspSuite.

Anything specific to Scala 3 we can test in cross suite.

The reason is for this comment is that these tests each can last up to a couple of seconds and it's harsh on out CI.

tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
```diff
@@ -12,12 +13,16 @@ import scala.meta.internal.mtags.MtagsEnrichments.XtensionAbsolutePath
...
 import munit.Location
 import munit.TestOptions
+import org.eclipse.lsp4j.CodeAction
 import tests.codeactions.BaseCodeActionLspSuite
 
 class Scala3CodeActionLspSuite
     extends BaseCodeActionLspSuite("cross-code-actions") {
 
   override protected val scalaVersion: String = BuildInfo.scala3
+  val filterAction: CodeAction => Boolean = { act: CodeAction =>
```
tgodzik, line 23: Why is this needed? I see that we anyway `ConvertToNamedArguments` action showing up below?

tgodzik (changes requested):
Looks good! Just very minor remarks about the tests

tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
tanishiking: Ah, right.

> All the other cases are already handled in the ConvertToNamedArgumentsLspSuite.
`ConvertToNamedArgumentsLspSuite` test against Scala 2.13, it doesn't verify anything against Scala 3.

Actually, those tests that verify the behavior (rather than the first one) should be in PC tests instead of `xxxLSPSuite`.
Sorry, we should did this in https://github.com/scalameta/metals/pull/3971

For example, Implementing all abstract member code action 
 have two (three?) test suite 
- cross test in pc (presentation compiler) 
  - https://github.com/scalameta/metals/blob/53249aa4047ee304275a818d3995a4b56a0b42b2/tests/cross/src/test/scala/tests/pc/AutoImplementAbstractMembersSuite.scala#L14
  - These tests verify the behavior of code action, but it doesn't check the code action is available or not. Those test anyway invoke the code action even if the user won't see it's available.
  - These tests run against all Scala versions
  - You can run those tests with the specific version by `++3.1.3 cross/testOnly tests.pc.AutoImplementAbstractMemberSuite`.
- ImplementAbstractMemberLspSuite
  - https://github.com/scalameta/metals/blob/53249aa4047ee304275a818d3995a4b56a0b42b2/tests/unit/src/test/scala/tests/codeactions/ImplementAbstractMembersLspSuite.scala#L5-L6
  - These test verify the code actions are available for users when we have a certain error (missing implementation error)
  - This test runs against only Scala 2.13
- Scala3CodeActionLspSuite
  - https://github.com/scalameta/metals/blob/53249aa4047ee304275a818d3995a4b56a0b42b2/tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala#L332-L358
  - This test check the code action is available in Scala3. We test this because if the Scala3 compiler has different error message than Scala2, we might need to change the regex to capture the error message for missing implementation error.
  - This test runs against only Scala3.

---

So, we should migrate most of `ConvertToNamedArgumentsLspSuite` into pc suite (that we haven't yet implemented). but it can be done in another PR :)

tgodzik (approved):
LGTM!

tgodzik (approved):
LGTM!

tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
tgodzik: Scala3CodeActionLspSuite will run the entire server, while cross tests will only invoke the compiler, which should be much faster.

tgodzik: It's probably fine to test if the workflow works for Scala 3, which confirms one new path for testing, while we can test anything specific to Scala 3 compiler in the cross tests.

Issue #4133, opened 2022-07-11
Syntax highlighting doesn't respect -Xsource


kubukoz:
### Describe the bug

1. Enable the `-Xsource:3.0` scalacOption

2. Write this
```scala
true
&& false
```

3. See error: `; expected but false found`

### Expected behavior

No redness

### Operating system

macOS

### Editor/Extension

vscode

### Version of Metals

0.11.7+7-67ce8743-SNAPSHOT

### Extra context or search terms

scalameta, scalafmt

tanishiking:
Hmm, looks like the dialect for scalameta parser used for emitting diagnostics inside Metals doesn't take `-X:source3` into account 🤔

tanishiking:
Ah, gotcha, it has to be `-Xsource:3` instead of `-Xsource:3.0`
https://github.com/scalameta/metals/blob/ec89eac187c5c9e1358a8229ebac85b64a71a8a9/metals/src/main/scala/scala/meta/internal/metals/ScalaTarget.scala#L72

tanishiking:
As it is valid to specify `-Xsource:3.0` or `-Xsource:3.x.x`, we should recognise `-Xsource:3.x.x` as well 👍 
https://github.com/scala/scala/blob/d23424cd3aa17f7ad95b81018c70ceed2566962b/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala#L105

PR #4134, opened 2022-07-12
fix: Dialect should be scala21xSource3 for `-Xsource:3.x.x`

tanishiking:
fix https://github.com/scalameta/metals/issues/4133

Previously, Metals recognise only `-Xsource:3`, but it doesn't for
`-Xsource:3.0` (`-Xsource:3.x.x`) even though it is valid in Scala2
compiler.
see: https://github.com/scala/scala/blob/d23424cd3aa17f7ad95b81018c70ceed2566962b/src/compiler/scala/tools/nsc/settings/ScalaSettings.scala#L105

This commit change Metals to understand `-Xsource:3.x.x` in addition to
`-Xsource:3` scalacOption.

### DIFF
--- a/tests/slow/src/test/scala/tests/feature/FoldingCrossLspSuite.scala
+++ b/tests/slow/src/test/scala/tests/feature/FoldingCrossLspSuite.scala
@@ -48,7 +48,7 @@ class FoldingCrossLspSuite extends BaseLspSuite("foldingRange-cross") {
         s"""|
             |/metals.json
             |{
-            |  "a": { "scalaVersion" : "${V.scala213}", "scalacOptions" : ["-Xsource:3"] }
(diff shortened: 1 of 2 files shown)


tests/slow/src/test/scala/tests/feature/FoldingCrossLspSuite.scala
```diff
@@ -48,7 +48,7 @@ class FoldingCrossLspSuite extends BaseLspSuite("foldingRange-cross") {
         s"""|
             |/metals.json
             |{
-            |  "a": { "scalaVersion" : "${V.scala213}", "scalacOptions" : ["-Xsource:3"] }
+            |  "a": { "scalaVersion" : "${V.scala213}", "scalacOptions" : ["-Xsource:3.0"] }
```
tanishiking, line 51: Maybe we should have a unit test for `ScalaTarget` 🤔

dos65 (approved):
LGTM

PR #4139, opened 2022-07-12
Extend extract value with new cases

jkciesluk:
Connected to https://github.com/scalameta/metals/pull/3297
Now we can extract value not only from `apply` but also from
- if true/false expression
- if/while/do-while condition
- tuple
- throw
- return
- match expression
- interpolation


tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
```diff
@@ -259,6 +259,127 @@ class Scala3CodeActionLspSuite
...
+
+  check(
+    "extract-match",
+    """|object Main:
+       |  def main(i : Int) =
+       |    1 + <<2>> + 3 match {
+       |      case _ => 6
+       |    }
+       |
+       |""".stripMargin,
```
kpodsiad, line 321: I wonder how it will behave for `1 + <<(1 + 2)>> + 3` (I haven't look at the code yet)

kpodsiad: I wonder, wouldn't extracting first valid thing which encloses current range be a better behavior? Moreover, it would be awesome to have information what is being extracted in code action message - like `extract 1 + 2 + 3 to value`

tgodzik: > I wonder, wouldn't extracting first valid thing which encloses current range be a better behavior? 

Maybe only if a range is specified? So if no range is specified then the last enclosing thing, but when range is specified then just the last thing that fits? I would do that in a separate PR though. 

Adding a title might be a good idea, but let's truncate it if it's too long

tgodzik (commented):
Looks good, I will have further look tomorrow, but I think including the title of the action would improve things really nicely

tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
kpodsiad: > Adding a title might be a good idea, but let's truncate it if it's too long

Yeah, truncating to e.g. 10 character and then showing ellipsis (`...`) should be enough to show what will be extracted

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -33,14 +33,19 @@ class ExtractValueCodeAction(
...
+            loop(tr) :+ tr
+          case Some(tr) => loop(tr)
+          case None => Nil
+        }
+      }
+      trees.get(path).map(loop(_)).getOrElse(Nil)
+    }
+    val textEdits: Seq[(Seq[l.TextEdit], String)] = for {
+      term <- allTrees
       stats <- lastEnclosingStatsList(term)
```
jkciesluk, line 49: I'm also thinking if we should put this and name generation above the for comprehension, because the result will be the same for every extract option

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -33,55 +33,131 @@ class ExtractValueCodeAction(
...
+          val replacementText =
+            term match {
+              case apply: Term.Apply
+                  if argument.is[Term.Block] && !applyHasParens(apply) =>
+                s"($name)"
+              case _ => name
+            }
+          val replacedArgument =
+            new l.TextEdit(argument.pos.toLSP, replacementText)
+          (valueTextWithBraces :+ replacedArgument, argument.toString())
```
kpodsiad, line 79: Similarly here, `:+` -> `::`. 
Speaking about List and Seq, I think `withBraces` can have explicit `List` type and `Seq` constructors can be replaced by `List` one.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -33,55 +33,131 @@ class ExtractValueCodeAction(
...
+
     textEdits match {
-      case Some(edits) =>
-        val codeAction = new l.CodeAction()
-        codeAction.setTitle(ExtractValueCodeAction.title)
-        codeAction.setKind(this.kind)
-        codeAction.setEdit(
-          new l.WorkspaceEdit(
-            Map(path.toURI.toString -> edits.asJava).asJava
+      case Some(actions) =>
```
kpodsiad, line 85: I think `textEdits.getOrElse(Nil).map { case (edits, title) => ???}` will look better than match, it's always lesser amount of braces.

kpodsiad (commented):
Few small suggestions, but the rest looks good.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
kpodsiad: Or we can add `.getOrElse(Nil)` directly to the `textEdits`, I mean sth like that
```scala
    val texEdits = {
    ...
    }.getOrElse(Nil)

    textEdits.map { case (edits, title) =>
```

tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala
```diff
@@ -142,7 +142,7 @@ class ExtractValueLspSuite
        |  }
        |}
        |""".stripMargin,
-    s"""|${ExtractValueCodeAction.title}
+    s"""|${ExtractValueCodeAction.title("1 + 2")}
```
kpodsiad, line 145: you have to update expected code actions in this suite

kpodsiad (approved):
LGTM! Checked out and works correctly for me. Good job @jkciesluk!

BTW, food for thought - https://speakerdeck.com/rafalsumislawski/we-should-stop-using-list-everywhere

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -33,55 +33,131 @@ class ExtractValueCodeAction(
...
-      argumentOpt <- apply.args.find { arg => arg.pos.encloses(range) }
-      argument = argumentOpt match {
-        // named paramaeter
-        case Term.Assign(_, rhs) => rhs
-        case other => other
+    val allTrees: Seq[Term] = {
+      def loop(t: Tree): Seq[Term] = {
+        t.children.find(_.pos.encloses(range)) match {
+          case Some(tr: Term) if existsRangeEnclosing(tr, range) =>
+            loop(tr) :+ tr
```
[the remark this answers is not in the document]
jkciesluk: If we do `tr :: loop(tr)`, should we also reverse the list after to get the resulting extract options from smallest range?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -266,5 +342,8 @@ class ExtractValueCodeAction(
 }
 
 object ExtractValueCodeAction {
-  val title = "Extract value"
+  def title(expr: String): String = {
+    if (expr.length <= 10) s"Extract value $expr"
+    else s"Extract value ${expr.take(10)}(...)"
```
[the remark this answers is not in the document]
kpodsiad: I like `as value` suffix idea

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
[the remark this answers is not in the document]
kpodsiad: Yeah, sure.

> `:+` on List (List is default Seq implementation) appends alement to the end of collection. Each time whole collection needs to be traversed so it has linear complexity. `::` on the other hand prepends element and has O(1) complexity.

Obviously in this case this doesn't make any difference since here in 99% there will be only one element in that List but it's better to do it because in general it won't be necessary similar case.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -33,53 +33,126 @@ class ExtractValueCodeAction(
...
-        val codeAction = new l.CodeAction()
-        codeAction.setTitle(ExtractValueCodeAction.title)
-        codeAction.setKind(this.kind)
-        codeAction.setEdit(
-          new l.WorkspaceEdit(
-            Map(path.toURI.toString -> edits.asJava).asJava
-          )
+    val textEdits = {
+      for {
+        head <- allTrees.headOption
```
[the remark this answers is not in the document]
jkciesluk: Yes, I've just found a case where it's not working.

jkciesluk:
I've reversed the changes, now we run the whole code every time. Previously it was not working for cases like
`if(cond) { a(a(<<2>>)) }`
If we'd choose the option to extract the whole block, the scope found would be this same block.

tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
```diff
@@ -260,6 +260,163 @@ class Scala3CodeActionLspSuite
...
+    s"""|${ExtractValueCodeAction.title("2 > 3")}
+        |""".stripMargin,
+    """|object Main:
+       |  def main(i : Int) =
+       |    val newValue = 2 > 3
+       |    if(newValue) 5 else 4
+       |
+       |""".stripMargin,
+  )
```
kpodsiad, line 278: General question about those tests: shouldn't we add just 1 test for extract value here to see if it works with Scala3 and move the rest to the `unit` ones?
cc: @tgodzik

kpodsiad: [Project structure](https://scalameta.org/metals/docs/contributors/getting-started#project-structure) section in docs isn't helpful about slow tests

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -33,53 +33,121 @@ class ExtractValueCodeAction(
...
-        Seq(codeAction)
-      case None => Nil
+      )
+      codeAction
+    }
+
+  }
+  private def applyArgument(argument: Term): Term =
+    argument match {
+      // named paramaeter
```
tgodzik, line 94:
```suggestion
      // named parameter
```

tgodzik (changes requested):
Just last minor changes, otherwise looks pretty amazing!

tgodzik (approved):
LGTM!

Issue #2105, opened 2020-09-30
No auto import completions inside a string


tgodzik:
**Describe the bug**

Completing:

```scala
"this is an interesting $File@@"
```

does not give us a completion that would automatically add an import.

**To Reproduce**

Just try it the above snippet.

**Expected behavior**

`java.nio.Files` is suggested

**Installation:**

- Operating system: Linux
- Editor: Visual Studio Code
- Metals version: v0.9.4

**Additional context**

It does work for:
```scala
s"this is an interesting $File@@"
```

**Search terms**

completion string interpolation

eilite:
When checking the logs, I see that the metals server does not receive a `textDocument/completion` request in this case : `"this is an interesting $File"` but in that case `s"this is an interesting $File"` it does.
Could that be the cause of this issue ?

ckipp01:
So I just quickly looked at this, and it does seem that when I try out the text I get a completion request:

<img width="780" alt="Screenshot 2020-10-12 at 19 14 16" src="https://user-images.githubusercontent.com/13974112/95773663-36a4f800-0cbf-11eb-83b7-6546c05ee201.png">

```
[Trace - 07:13:14 PM] Received request 'textDocument/completion - (123)'
Params: {
  "context": {
    "triggerKind": 1
  },
  "textDocument": {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/tester/src/main/scala/example/Main.scala"
  },
  "position": {
    "line": 4,
    "character": 28
  }
}
```
So it does look like the request is there, but more that the correct completion is coming back. I'd dive in [here](https://github.com/scalameta/metals/blob/main/mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala) to start and try to follow where you'd imagine the completion should come from and narrow it down to where the issue may be.

PR #4140, opened 2022-07-12
bugfix: [Scala 2] Automatically import types in string interpolations

tgodzik:
Previously, when trying to provide inside a string we would not get completions from the workspace. Now, we properly add them.

Related to https://github.com/scalameta/metals/issues/2105 but it still needs a fix for Scala 3

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
@@ -82,6 +82,13 @@ class CompletionProvider(
         case t: TextEditMember if t.detail.isDefined => t.detail.get
         case _ => detailString(member, history)
       }
+
+      def labelWithSig =
+        if (member.sym.isMethod || member.sym.isValue) {
+          ident + detail
+        } else {
+          ident
+        }
       val label = member match {
         case _: NamedArgMember =>
           val escaped = if (isSnippet) ident.replace("$", "$$") else ident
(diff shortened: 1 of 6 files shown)


tanishiking (approved):
LGTM!

Just a few questions, but they won't be blockers for merging this PR :)

- Is there an option or something to invoke interpolation completion without triggering completions manually?
  - I trigger interpolation completion by hand using "^+Space" (for me), but it would be useful if interpolation completion will be triggered automatically when we write `$xxx` inside of `""`.

interpolator

---

Also, is it possible to extend this fix to normal completion in`s""`?

interpolator-s

I usually write `s""` and start writing string interpolation, so it would be great if this fix is available in `s""` :)

tgodzik:
> LGTM!
> 
> Just a few questions, but they won't be blockers for merging this PR :)
> 
> * Is there an option or something to invoke interpolation completion without triggering completions manually?
>   
>   * I trigger interpolation completion by hand using "^+Space" (for me), but it would be useful if interpolation completion will be triggered automatically when we write `$xxx` inside of `""`.

That's controlled by `editor.quickSuggestions`  in VS Code, you can change it for the comments

> Also, is it possible to extend this fix to normal completion in`s""`?
> I usually write `s""` and start writing string interpolation, so it would be great if this fix is available in `s""` :)

Fixed!

I also changed the label to include the signature of the method for InterpolationMemberCompletions, which makes it look like all the other methods. No need to make it special.

See the two added commits for detals.

dos65 (approved):
:+1:

PR #4141, opened 2022-07-12
feat: Import missing extension method

tanishiking:
refer https://github.com/scalameta/metals-feature-requests/issues/141

This PR enables Metals to auto-import missing extension method **from the workspace (it doesn't support import from third-party libraries).**

extension-import

This is done by

- Index extension methods by `ScalaToplevelMtags`
- Show AutoImport code action for `value xxx is not a member of yyy`.

## What I didn't in this PR
- Auto import given instances
- Auto import implicit classes
- completion for extension method and auto-import
  - I'll open an feature-request-issue

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/mtags/MtagsIndexer.scala
+++ b/mtags/src/main/scala/scala/meta/internal/mtags/MtagsIndexer.scala
@@ -64,18 +64,18 @@ trait MtagsIndexer {
         )
     }
   }
-  def term(name: String, pos: m.Position, kind: Kind, properties: Int): Unit =
+  def term(name: String, pos: m.Position, kind: Kind, properties: Int): String =
     addSignature(Descriptor.Term(name), pos, kind, properties)
-  def term(name: Term.Name, kind: Kind, properties: Int): Unit =
+  def term(name: Term.Name, kind: Kind, properties: Int): String =
     addSignature(Descriptor.Term(name.value), name.pos, kind, properties)
-  def tparam(name: Name, kind: Kind, properties: Int): Unit =
+  def tparam(name: Name, kind: Kind, properties: Int): String =
     addSignature(
       Descriptor.TypeParameter(name.value),
       name.pos,
       kind,
       properties
     )
-  def param(name: Name, kind: Kind, properties: Int): Unit =
+  def param(name: Name, kind: Kind, properties: Int): String =
     addSignature(
       Descriptor.Parameter(name.value),
       name.pos,
@@ -86,7 +86,7 @@ trait MtagsIndexer {
       disambiguator: String,
       pos: m.Position,
       properties: Int
-  ): Unit =
+  ): String =
     addSignature(
       Descriptor.Method(Names.Constructor.value, disambiguator),
       pos,
@@ -98,7 +98,7 @@ trait MtagsIndexer {
       disambiguator: String,
       pos: m.Position,
       properties: Int
-  ): Unit =
+  ): String =
     addSignature(
       Descriptor.Method(name, disambiguator),
       pos,
@@ -110,7 +110,7 @@ trait MtagsIndexer {
       disambiguator: String,
       kind: Kind,
       properties: Int
-  ): Unit = {
+  ): String = {
     val methodName = name match {
       case Name.Anonymous() => Names.Constructor.value
       case _ => name.value
@@ -122,11 +122,11 @@ trait MtagsIndexer {
       properties
     )
   }
-  def tpe(name: String, pos: m.Position, kind: Kind, properties: Int): Unit =
+  def tpe(name: String, pos: m.Position, kind: Kind, properties: Int): String =
     addSignature(Descriptor.Type(name), pos, kind, properties)
-  def tpe(name: Name, kind: Kind, properties: Int): Unit =
+  def tpe(name: Name, kind: Kind, properties: Int): String =
     addSignature(Descriptor.Type(name.value), name.pos, kind, properties)
-  def pkg(name: String, pos: m.Position): Unit = {
+  def pkg(name: String, pos: m.Position): String = {
     addSignature(Descriptor.Package(name), pos, Kind.PACKAGE, 0)
   }
   def pkg(ref: Term): Unit =
@@ -142,7 +142,7 @@ trait MtagsIndexer {
       definition: m.Position,
       kind: s.SymbolInformation.Kind,
       properties: Int
-  ): Unit = {
+  ): String = {
     val previousOwner = currentOwner
     currentOwner = symbol(signature)
     myLastCurrentOwner = currentOwner
@@ -163,6 +163,7 @@ trait MtagsIndexer {
       displayName = signature.name.value
     )
     visitOccurrence(occ, info, previousOwner)
+    syntax
   }
   def symbol(signature: Descriptor): String =
     if (currentOwner.eq(Symbols.EmptyPackage) && signature.isPackage)
--- a/mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
+++ b/mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
@@ -78,8 +78,8 @@ class ScalaToplevelMtags(
       Some(ExpectTemplate(indent, currentOwner, false, false))
     def newExpectPkgTemplate: Some[ExpectTemplate] =
       Some(ExpectTemplate(indent, currentOwner, true, false))
-    def newExpectExtensionTemplate: Some[ExpectTemplate] =
-      Some(ExpectTemplate(indent, currentOwner, false, true))
+    def newExpectExtensionTemplate(owner: String): Some[ExpectTemplate] =
+      Some(ExpectTemplate(indent, owner, false, true))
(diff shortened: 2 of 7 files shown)


tanishiking:
Realized it doesn't work with toplevel extensions,

mtags/src/main/scala/scala/meta/internal/mtags/MtagsIndexer.scala
```diff
@@ -142,7 +142,7 @@ trait MtagsIndexer {
       definition: m.Position,
       kind: s.SymbolInformation.Kind,
       properties: Int
-  ): Unit = {
+  ): String = {
```
tanishiking, line 145: I changed `addSignature` to return the added symbol so we can use those symbols as a next owner.

tanishiking:
Now, ready for review :+1:

tests/slow/src/test/scala/tests/feature/ImportMissingSymbolCrossLspSuite.scala
```diff
@@ -41,4 +42,85 @@ class ImportMissingSymbolCrossLspSuite
...
+            |def main =
+            |  println(1.incr)
+            |""".stripMargin
+      )
+      _ <- server.didOpen(path)
+      _ <- server.assertCodeAction(
+        path,
+        s"""|package x
+            |def main =
+            |  println(1.<<incr>>)
```
tgodzik, line 74: Could we maybe add it to the code actions Scala 3 suite? Or at least it would be nice to see the code action applied to test if it works correctly

mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
```diff
@@ -114,10 +118,39 @@ class ScalaToplevelMtags(
             )
           } else
             loop(indent, false, currRegion, newExpectTemplate)
+        case IDENTIFIER
```
tgodzik, line 121: Should this not show up in `ToplevelSuite` ? It would be cool to test it in the expect suites

tgodzik (changes requested):
Looks good! TopLevels should also be run on dependency sources, so we should suggest extension methods from dependencies also not only from the workspace.

mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
dos65: There is also `ScalaTopLevelSuite`.

Overall looks good. I only noticed that with `includeInnerClasses = true` it might return inner methods.
I added the following test in ScalaTopLevelSuite with fails:
```scala
  check(
    "toplevel-extesion".only,
    """|package a
       |
       |extension (s: String)
       |  def foo: Int = ???
       |  def bar: String =
       |    def hmm: Int = ???  // <- shouldn't be returned
       |    ???
       |
       |""".stripMargin,
    List(
      "a/",
      "a/Test$package.",
      "a/Test$package.bar.",
      "a/Test$package.foo.",
    ),
    all = true,
  )
```
Output:
```
- tests.ScalaToplevelSuite:
  * tests.ScalaToplevelSuite.toplevel-extesion - munit.ComparisonFailException: /home/dos65/projects/metals/samples/inc_error/metals/tests/unit/src/test/scala/tests/ScalaToplevelSuite.scala:250
249:
250:  check(
251:    "toplevel-extesion".only,
diff assertion failed
=> Obtained
    """|a/
       |a/Test$package.
       |a/Test$package.bar.
       |a/Test$package.foo.
       |a/Test$package.hmm.
       |""".stripMargin
``` 


However, I can't remember when `includeInnerClasses=true` is used.
Also, unfortunately, I don't remember what was the purpose in returning toplevel DEF/VAL/EXTENSION and others properly. :smiling_face_with_tear: 
In theory, the only toplevel that is required for symbols search is `${packageName}/${sourceName}$package.`
I'm wondering, if `Import missing` action would work if `extension` kw would  just register that there is a toplevel owner without parsing inner Def :thinking:

tanishiking: > I'm wondering, if Import missing action would work if extension kw would just register that there is a toplevel owner without parsing inner Def 🤔

It would be great, but I think it's impossible because

- extension method callee doesn't know what import will make the extension method call possible
  - the only clue is the name of the extension method
  - if we expand this feature into extension method auto-completion, we need to know (index) the method name

tanishiking: I added tests and fix the issue @dos65 mentioned in https://github.com/scalameta/metals/pull/4141/commits/91a3903c791aaffa2a975f21a7f0f8ad964ca609 👍

> Should this not show up in ToplevelSuite ? It would be cool to test it in the expect suites

I don't think it appears in ToplevelSuite, as `ToplevelSuite` generate by `Mtags.toplevels` that never parse the inner classes or extension methods, not `Mtags.allToplevels`.
Should we add expect suite for allToplevels ?

tanishiking:
> TopLevels should also be run on dependency sources, so we should suggest extension methods from dependencies also not only from the workspace.

@tgodzik 
Year, it would be awesome to have! Could you give me some pointers on where to code, and where we're testing them? I'm unfamiliar with those areas. FYI @dos65

dos65:
@tanishiking I took a look at `TopLevelSuite`. It requires some work to make it work for Scala3 .At least, I see that `BaseExpectSuite` should be parametrized by an input - currently it uses only scala2 input classpath.
I think, it would be ok not to touch this suite in this PR.

tgodzik:
> > TopLevels should also be run on dependency sources, so we should suggest extension methods from dependencies also not only from the workspace.
> 
> @tgodzik Year, it would be awesome to have! Could you give me some pointers on where to code, and where we're testing them? I'm unfamiliar with those areas. FYI @dos65

Wait, don't we already do it? We should be caching toplevels in ScalaToplevelMtags.scala, so esentially extension methods should currently work from dependencies in this PR, or did I totally forget something?

tanishiking:
> Wait, we are talking just about tests, aren't we? 😅

Ahhh, I just mistook `Toplevels` for something like `JarToplevelMtags` (whatever), but you meant `TopLevelSuite`.

So, you mean `TopLevelSuite` (or maybe another suite that extends `BaseExpectSuite`) should generate expect files by indexing the symbols using `PackageIndex` against third-party libraries such as `scala-library` to verify there's no regression in dependencies symbol indexing?

---

While @dos65 talks about TopLevelSuite to work on Scala3.
However, it looks like `TopLevelSuite` already works against Scala3.

In `TopSevelSuite`, 
`forInput(..., Scala213) runs `Mtags.toplevels` against `tests/input/src/main/scala/example/*.scala`  and `forInput(..., Scala3) runs `Mtags.toplevels` against `tests/input2/src/main/scala/example/*.scala`.
ToplevelSuite concat the result of both Scala213 and Scala3 and then write it to `toplevel.expect` file.
 https://github.com/scalameta/metals/blob/a163e221cf1c5316f9a01f7b0a6c8d7e2a1d7ba9/tests/unit/src/test/scala/tests/ToplevelSuite.scala#L39-L40

tests/unit/src/test/scala/tests/ScalaToplevelSuite.scala
```diff
@@ -249,6 +249,49 @@ class ScalaToplevelSuite extends BaseSuite {
     dialect = dialects.Scala213,
   )
 
+  check(
+    "extesion-methods",
```
tgodzik, line 253:
```suggestion
    "extension-methods",
```

dos65:
> While @dos65 talks about TopLevelSuite to work on Scala3.
However, it looks like TopLevelSuite already works against Scala3.

There is suspicious clause and comment but I'm not sure how it affects this test.
https://github.com/scalameta/metals/blob/f9660d5f9780dda91b26b3bed0225a92a7abfd99/tests/unit/src/test/scala/tests/ToplevelSuite.scala#L30-L31

tests/unit/src/test/scala/tests/ScalaToplevelSuite.scala
```diff
@@ -249,6 +249,49 @@ class ScalaToplevelSuite extends BaseSuite {
...
+      "a/A.",
+      "a/A.bar.",
+      "a/A.foo.",
+    ),
+    all = true,
+    dialect = dialects.Scala3,
+  )
+
+  check(
+    "toplevel-extesion",
```
tgodzik, line 275:
```suggestion
    "toplevel-extension",
```

mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
tgodzik: Ach, ok I looked again into the test suites and realised that we have the additional `innerClasses` boolean option. No need to tests in ToplevelSuite

tgodzik (approved):
Added super minor comments, which we can fix later on. LGTM! Great feature to have, we can comment on the issue to document what is added and what still needs to be done.

tanishiking:
Hmm, right I'll take a look in another PR :eyes:

PR #4144, opened 2022-07-12
fix: avoid dupl;icate migration appllication

dos65:
Use `AtomicReference` for connection instead of `var` to prevent
concurrent migration application.

In my case, I have two `Flyway` migrations at server start and once I got broken db state (no new table and constant migration application without creating a new table)


metals/src/main/scala/scala/meta/internal/metals/Tables.scala
```diff
@@ -35,20 +38,41 @@ final class Tables(
...
+
+  def connect(): Connection = {
+    ref.get() match {
+      case empty @ ConnectionState.Empty =>
+        if (ref.compareAndSet(empty, ConnectionState.InProgress)) {
+          val conn = tryAutoServer()
+          ref.set(ConnectionState.Connected(conn))
+          conn
+        } else
+          connect()
```
kpodsiad, line 52: Isn't it the same case as `InProgress` and we should wait a bit too?

kpodsiad (approved):

metals/src/main/scala/scala/meta/internal/metals/Tables.scala
tgodzik: We will end up at Tables.ConnectionState.InProgress  right away, so it should be ok

metals/src/main/scala/scala/meta/internal/metals/Tables.scala
```diff
@@ -35,20 +38,41 @@ final class Tables(
   val fingerprints =
     new Fingerprints(() => connection)
 
-  def connect(): Unit = {
-    this._connection =
-      if (clientConfig.initialConfig.isAutoServer) tryAutoServer()
```
tgodzik, line 40: Nice catch! :sweat_smile:

tgodzik (approved):
LGTM!

PR #4150, opened 2022-07-14
refactor: Show more debug messages

tgodzik:
Added some more debug logs for:
- RenameProvider, since sometimes it fails without any messages and it's super hard to reproduce
- ConvertToNamedArguments command instead of Future.filter, which would throw an exception.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -213,6 +213,11 @@ final class RenameProvider(
             renamedImportOccurrences(source, symbolOccurrence)
           else allReferences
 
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
```diff
@@ -213,6 +213,13 @@ final class RenameProvider(
             renamedImportOccurrences(source, symbolOccurrence)
           else allReferences
 
+        if (fallbackOccurences.isEmpty) {
+          if (symbolOccurrence.isEmpty)
+            scribe.debug(s"Symbol occurence was $symbolOccurrence")
```
kpodsiad, line 218: won't it be just `Symbol occurence was None`?

PR #4151, opened 2022-07-14
fix triple quoted new line on type formatting

### LINKED ISSUE #3817: `formatOnType` misbehaves with interpolated triple-quoted multiline strings

vzmerr:
fixes #3817

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
@@ -65,7 +65,7 @@ case class MultilineString(userConfig: () => UserConfiguration)
     space * index
   }
 
-  private def getIndexOfLastQuote(line: String): Option[(Int, Boolean)] = {
+  private def getIndexOfLastOpenQuote(line: String): Option[(Int, Boolean)] = {
 
     var lastQuote = -1
     var escaped = false
@@ -88,12 +88,13 @@ case class MultilineString(userConfig: () => UserConfiguration)
     if (lastQuote != -1) Some((lastQuote, quoteClosed)) else None
   }
 
-  private def getIndexOfLastTripleQuote(
-      line: String
+  private def getIndexOfLastOpenTripleQuote(
+      closedFromPreviousLines: Boolean,
+      line: String,
   ): Option[(Int, Boolean)] = {
     var lastTripleQuote = -1
     var escaped = false
-    var tripleQuoteClosed = true
+    var tripleQuoteClosed = closedFromPreviousLines
     var quoteNum = 0
     for (i <- 0 until line.size) {
       val char = line(i)
@@ -134,9 +135,9 @@ case class MultilineString(userConfig: () => UserConfiguration)
   }
 
   private def hasNQuotes(start: Int, text: String, n: Int): Boolean =
-    (start until start + n).forall { i =>
-      if (i < 0 || i > text.length) false else text(i) == quote
-    }
+    (start until start + n).forall(i =>
+      if (i < 0 || i >= text.length) false else text(i) == quote
+    )
 
   private def indentWhenNoStripMargin(
       expr: StringLiteralExpr,
@@ -261,7 +262,7 @@ case class MultilineString(userConfig: () => UserConfiguration)
       position: Position,
   ): Boolean = {
     val lineBefore = splitLines(position.getLine - 1)
-    getIndexOfLastQuote(lineBefore).exists { case (_, quoteClosed) =>
+    getIndexOfLastOpenQuote(lineBefore).exists { case (_, quoteClosed) =>
       !quoteClosed
     }
   }
@@ -270,10 +271,19 @@ case class MultilineString(userConfig: () => UserConfiguration)
       splitLines: Array[String],
       position: Position,
   ): Boolean = {
-    val currentLine = splitLines(position.getLine - 1)
-    getIndexOfLastTripleQuote(currentLine).exists { case (_, quoteClosed) =>
-      !quoteClosed
+    var closedFromPreviousLines = true
+    var existed = false
+    for (i <- 0 until position.getLine()) {
+      val currentLine = splitLines(i)
+      getIndexOfLastOpenTripleQuote(closedFromPreviousLines, currentLine)
+        .foreach { case (_, quoteClosed) =>
+          closedFromPreviousLines = quoteClosed
+          existed = true
+        }
     }
+    if (existed)
+      !closedFromPreviousLines
+    else false
   }
 
   private def fixStringNewline(
@@ -298,7 +308,7 @@ case class MultilineString(userConfig: () => UserConfiguration)
       )
         defaultIndent + 2
       else defaultIndent
-    val interpolationString = getIndexOfLastQuote(previousLine)
+    val interpolationString = getIndexOfLastOpenQuote(previousLine)
       .map { case (lastQuoteIndex, _) =>
         if (lastQuoteIndex > 0 && previousLine(lastQuoteIndex - 1) == 's') "s"
         else ""
@@ -321,9 +331,8 @@ case class MultilineString(userConfig: () => UserConfiguration)
   }
 
   private def addTripleQuote(pos: Position): List[TextEdit] = {
-    val pos1 = new Position(pos.getLine, pos.getCharacter)
-    val pos2 = new Position(pos.getLine, pos.getCharacter + 1)
-    List(new TextEdit(new Range(pos1, pos2), "\"\"\""))
+    val endPos = new Position(pos.getLine, pos.getCharacter + 1)
+    List(new TextEdit(new Range(pos, endPos), "\"\"\""))
   }
 
   private def formatPipeLine(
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
```diff
@@ -232,6 +266,16 @@ case class MultilineString(userConfig: () => UserConfiguration)
...
   }
 
+  private def wasTripleQuoted(
+      splitLines: Array[String],
+      position: Position,
+  ): Boolean = {
+    val currentLine = splitLines(position.getLine - 1)
+    getIndexOfLastTripleQuote(currentLine).exists { case (_, quoteClosed) =>
+      !quoteClosed
+    }
```
tanishiking, line 276:
```suggestion
    val prevLine = splitLines(position.getLine - 1)
    val currLine = splitLines(position.getLine)
    prevLine.endsWith("\"\"\"") && !currLine.startsWith("\"\"\"")
```

Can we simplify it like this?
The idea is, that what we need to know are:
- Did the user enter a newline after `"""`?
- Is the triple quote closed (that means the current line of cursor starts with `"""`)? (maybe there's some whitespace for indent)

metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
```diff
@@ -100,9 +134,9 @@ case class MultilineString(userConfig: () => UserConfiguration)
   }
 
   private def hasNQuotes(start: Int, text: String, n: Int): Boolean =
-    (start until start + n).forall(i =>
+    (start until start + n).forall { i =>
       if (i < 0 || i > text.length) false else text(i) == quote
```
tanishiking, line 138: Not related to this change, but I encounter some `IndexOutofBoundsException` when I write quote in the end of a file, maybe we should check the length of text before accessing `text(I)` :thinking:

metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
```diff
@@ -276,6 +320,12 @@ case class MultilineString(userConfig: () => UserConfiguration)
     List(new TextEdit(new Range(pos1, pos2), "\"\"\"\"\"\""))
   }
 
+  private def addTripleQuote(pos: Position): List[TextEdit] = {
+    val pos1 = new Position(pos.getLine, pos.getCharacter)
+    val pos2 = new Position(pos.getLine, pos.getCharacter + 1)
+    List(new TextEdit(new Range(pos1, pos2), "\"\"\""))
```
tanishiking, line 326: [nit]
```suggestion
    List(new TextEdit(new Range(pos, pos), "\"\"\""))
```

It's ok with point

metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
vzmerr: Thank you so much for reviewing this Rikito :)

Well, frankly this was copied from `doubleQuoteNotClosed` whose logic was previously tested in the code.

As an example of the issue's trickiness, the `getIndexOfLastTripleQuote`, which was copied from `getIndexOfLastQuote` does not just get the index of last triple quote, but also It gets the index of the the last triple quote, which was not closed. 

So basically this prevents us from putting an extra triple quote if the previous line had two triple quotes: one begin, and one close.

Another tricky matter is that, this code action does itself generate a text edit to put the closing triple quotes in the `curLine`which is the line after the newly inserted newline character. 

So there won't be any triple quotes in it.

metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
vzmerr: hmmm, it seems that it is checking `i > text.length`

metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
vzmerr: I am just afraid about removing the `pos.getCharacter + 1` for the end of range, because this was copied from existing logic and the indices in TextEdits are so tricky.

metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
tanishiking: Ah, I got it, if i == text.length it will fail

tanishiking: it should be i >= text.length

vzmerr: I wonder why it is if `(i < 0 || i > text.length)` in the first place, it is not checking the `0` index either.

metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
tanishiking: Nevermind, @vzmerr pointed me out that it doesn't work for `"""foo|"""` and enter newline.

vzmerr: I just noticed that the suggestions might lead to unexpected results in a few cases, for example
`val a = """" this is an open triple quoted string not ending with triple quotes`

Also, the function `getIndexOfLastTripleQuote` does more than just escaping the escaped ones. It also keeps opening and closing the triple quotes of the previous line, as it traverses through the characters to make sure that the last triple quote in the line is still open.
`a = """ first one """ + """ second one """ + """ third one which is not closed`

Also in case of `sql"""` the closing triple quotes do not get inserted automatically, so they are not present in the first place.

metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
```diff
@@ -87,10 +88,44 @@ case class MultilineString(userConfig: () => UserConfiguration)
...
+            lastTripleQuote = i
+            tripleQuoteClosed = !tripleQuoteClosed
+            quoteNum = 0
+          }
+        } else {
+          escaped = !escaped
+        }
+      } else if (char == '\\') {
+        quoteNum = 0
+        escaped = !escaped
```
tanishiking, line 114: I still don't think escape handling is not needed because of https://github.com/scalameta/metals/pull/4151#discussion_r921969587

tanishiking (commented):
I left one comment, but otherwise LGTM!

metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
vzmerr: we are traversing character by character. What if the escape is at the beginning?

vzmerr: `\"""`

vzmerr: ` val a = """ This is a sample \""" quoted string in which escaping matters

tanishiking: Ah, right

metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
```diff
@@ -227,11 +262,30 @@ case class MultilineString(userConfig: () => UserConfiguration)
...
     }
   }
 
+  private def wasTripleQuoted(
+      splitLines: Array[String],
+      position: Position,
+  ): Boolean = {
+    var closedFromPreviousLines = true
+    var existed = false
+    for (i <- 0 until position.getLine()) {
```
tgodzik, line 276: It's a pity we need to go over every line, but I don't see a way around it :thinking:

tgodzik (approved):
LGTM!

tanishiking (commented):
LGTM!
but can you add those cases to test? it's ok in another PR :)

vzmerr:
> LGTM!
> but can you add those cases to test? it's ok in another PR :)
> [#4151 (comment)](https://github.com/scalameta/metals/pull/4151#discussion_r921987100)
> [#4151 (comment)](https://github.com/scalameta/metals/pull/4151#discussion_r923176489)

good idea 👍

metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
vzmerr: turns out that you were right @tanishiking 

In multiline strings, one cannot escape `"` using `\` character.

vzmerr: https://scastie.scala-lang.org/vzmerr/tPvC34B4RM2EIywhJlcLRg/46

vzmerr: https://scastie.scala-lang.org/vzmerr/tPvC34B4RM2EIywhJlcLRg/47

tanishiking: Thank you for investigating that!

PR #4152, opened 2022-07-14
bugfix: Reenable RenameLspSuite

tgodzik:
It seems that at some point we added abstract flga, but never actually created an implementation of that :O


dos65 (approved):
Good catch! :+1:

Issue #4138, opened 2022-07-12
Metals worksheet broken?


Swoorup:
### Describe the bug

It appears that VS worksheet appears to be broken for 3.2.0-RC2 Scala 

for a simple worksheet containing `println("hello")`

I get the following error

```shell
test.worksheet.sc:2 (mdoc generated code) 
 package scala.compiletime does not have a member method erasedValue
object MdocSession extends _root_.mdoc.internal.document.DocumentBuilder {
```

Other times I get 

```shell
2022.07.12 21:20:03 INFO  time: evaluated worksheet 'circe.worksheet.sc' in 0.67s
2022.07.12 21:20:07 ERROR worksheet: /Users/swoorup/circe.worksheet.scjava.lang.ClassCastException: class dotty.tools.dotc.core.Types$NoType$ cannot be cast to class dotty.tools.dotc.typer.Implicits$SearchFailureType (dotty.tools.dotc.core.Types$NoType$ and dotty.tools.dotc.typer.Implicits$SearchFailureType are in unnamed module of loader java.net.URLClassLoader @65991992)
	at dotty.tools.dotc.typer.Implicits$SearchFailure.reason(Implicits.scala:412)
	at dotty.tools.dotc.typer.ErrorReporting$Errors.selectErrorAddendum$$anonfun$1$$anonfun$1(ErrorReporting.scala:182)
	at scala.collection.immutable.List.foreach(List.scala:333)
	at dotty.tools.dotc.typer.ErrorReporting$Errors.selectErrorAddendum$$anonfun$1(ErrorReporting.scala:186)
	at scala.runtime.function.JProcedure1.apply(JProcedure1.java:15)
	at scala.runtime.function.JProcedure1.apply(JProcedure1.java:10)
	at scala.Option.foreach(Option.scala:437)
	at dotty.tools.dotc.typer.ErrorReporting$Errors.selectErrorAddendum(ErrorReporting.scala:186)
	at dotty.tools.dotc.typer.TypeAssigner.addendum$1(TypeAssigner.scala:170)
	at dotty.tools.dotc.typer.TypeAssigner.$anonfun$4(TypeAssigner.scala:173)
	at dotty.tools.dotc.reporting.NotAMember.msg(messages.scala:349)
	at dotty.tools.dotc.reporting.Message.message(Message.scala:87)
	at dotty.tools.dotc.reporting.Diagnostic.message(Diagnostic.scala:97)
[...]

PR #4153, opened 2022-07-14
bughack: Force specific Scala 3 compiler for worksheets

tgodzik:
Mdoc is currently being released for Scala 3.1.x, which means that if we use mdoc for worksheets it will not be able to run with a classpath containing any Scala 3.2.0 code. However, since mdoc should be comptable with any future Scala 3 version, we can force a later compiler version to run with it.

This efectively fixes the worksheet issue for later Scala 3 versions. That is until the interfaces change, but hopefully we use a very limited api, so this should never be an issue.

Fixes https://github.com/scalameta/metals/issues/4138

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Embedded.scala
@@ -39,11 +39,17 @@ final class Embedded(
     mdocs.clear()
   }
 
-  def mdoc(scalaBinaryVersion: String): Mdoc = {
+  def mdoc(scalaVersion: String): Mdoc = {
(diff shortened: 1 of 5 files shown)


dos65 (commented):
It looks like a valid workaround for metals. :thinking: 

However, I'm wondering what to do with mdoc. I suppose, projects that will switch on 3.2.x won't be able to work with mdoc if they have some custom modifiers.

tgodzik:
Already got it working with a similar workaround in https://github.com/scalameta/mdoc/pull/669

dos65 (approved):
Ok, lets try this way

tgodzik:
> Ok, lets try this way

The only alternative is releasing for each binary version, which I would like to avoid :sweat_smile:

PR #4157, opened 2022-07-15
Code Actions doc page

### LINKED ISSUE #2296: Add a section to the docs that show all code actions
There isn't currently a great way to know what code actions exist in Metals apart from looking in the source. I think it'd be a great idea to have a small section in the docs highlighting the code actions that Metals supports. As we continue to add in more refactoring related actions, it may be wise to have a centralized place to show those. It seems to be a somewhat common questions asked of Metals, _What refactoring support does Metals have?_

vzmerr:
It resolves #2296


docs/codeactions/code-actions.md
```diff
@@ -0,0 +1,79 @@
...
+Here is a list of the code actions supported by Metals. Some of these code actions support much more complex use cases, which you would eventually discover through your code, by clicking on the magic 💡, as it appears, to get your surprise!
+
+## Convert To Named Arguments
+
+This code action adds the name to parameters passed to a method call or a class instant.
+
+[Convert To Named Arguments Code Action](https://github.com/scalameta/metals/docs/codeactions/gifs/ConvertToNamedArguments.gif)
+
+
+## Create Companion Object
```
tgodzik, line 26: Let's add a short explanation even if it seems obvious

docs/codeactions/code-actions.md
```diff
@@ -0,0 +1,96 @@
...
+Whenever you are stuck or unsure, please open an issue or [ask us on
+Discord](https://discord.gg/DwTc8xbNDd) or [on our Matrix
+bridge](https://matrix.to/#/#scalameta:metals-contributors). This project
+follows [Scalameta's contribution
+guidelines](https://github.com/scalameta/scalameta/blob/master/CONTRIBUTING.md)
+and the [Scala CoC](https://scala-lang.org/conduct/).
+
+Code Actions are there to provide refactoring or rewriting of your current code in order to enhance its readability or save you some time.
+
+The common way of using them, is to click on the yellow light bulb appearing on your code, near the cursor and then, to pick the desired action to peform on the code.
```
tgodzik, line 15:
```suggestion
In Visual Studio Code, to use them you can click on the yellow light bulb appearing on your code, near the cursor and then, to pick the desired action to perform on the code. Alternatively, you can use the `editor.action.quickFix` shortcut.

For other editors this should also be available via specified shortcut.
```

docs/codeactions/code-actions.md
```diff
@@ -0,0 +1,96 @@
...
+## Convert To Named Arguments
+
+This code action adds the name to parameters passed to a method call or a class instant.
+
+[Convert To Named Arguments Code Action](https://github.com/scalameta/metals/docs/codeactions/gifs/ConvertToNamedArguments.gif)
+
+
+## Create Companion Object
+
+If you have a `trait` or a `class` definition, you can invoke this code action for creating its companion object.
```
tgodzik, line 28:
```suggestion
If you have a `trait` or a `class` definition, you can invoke this code action to create its companion object.
```

docs/codeactions/code-actions.md
```diff
@@ -0,0 +1,96 @@
...
+
+### map, flatMap, withFilter to For Yield
+
+It converts a chain of `map`, `flatMap`, `filter` and `filterNot` methods into a for comprehension
+
+[To For Comprehension](https://github.com/scalameta/metals/docs/codeactions/gifs/FlatMapToForComprehension.gif)
+
+## Implement Abstract Members of the Parent Type
+
+Upon inherting from a type, you also have to implement its abstract members. But manuallylooking them all up and copying their signature is time consuming, isn't it? You can just use this code action instead.
```
tgodzik, line 63:
```suggestion
Upon inherting from a type, you also have to implement its abstract members. But manually looking them all up and copying their signature is time consuming, isn't it? You can just use this code action instead.
```

docs/codeactions/code-actions.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Insert Inferred Type
+
+This one, tries to infer the type of variables according to the context, and insert them for you.
+
+[Insert Inferred Type](https://github.com/scalameta/metals/docs/codeactions/gifs/InsertInferredType.gif)
+
+## Organize Imports
+
+Yes, just have all the clutter, as well as the unnecessary ones removed and tidied up.
```
tgodzik, line 81: This is a bit unclear, it would also be good to mention that we need a special compiler option for removing unused imports and that in some editors it's not available on the "lgiht bulb"

docs/codeactions/code-actions.md
vzmerr: The informal descriptions are nowadays very popular in technical manuals and blog posts. They entertain and engage the user as they go through the article. So they increase the chance of the article for getting read to the end.

vzmerr: But people are different. However you write it, there will still be people who won't be ok with it, won't there?

docs/codeactions/code-actions.md
```diff
@@ -0,0 +1,92 @@
...
+
+For other editors this should also be available via specified shortcut.
+
+Here is a list of the code actions supported by Metals. Some of these code actions support much more complex use cases, which you would eventually discover through your code, by clicking on the magic 💡, as it appears, to get your surprise!
+
+## Convert To Named Arguments
+
+This code action adds the name to parameters passed to a method call or a class constructor.
+
+[Convert To Named Arguments Code Action](https://github.com/scalameta/metals/docs/codeactions/gifs/ConvertToNamedArguments.gif)
```
tanishiking, line 19:
```suggestion
![convert to named arguments code action](./gifs/converttonamedarguments.gif)
```

to show the gif in the webpage instead of just a link to gif

tanishiking (commented):
The document itself looks good, but you need to add `"codeactions/codeactions"` to somewhere in `website/sidebar.json` 👍

docs/codeactions/code-actions.md
```diff
@@ -0,0 +1,92 @@
...
+## Rename File
+
+Renames a file to match that of the prominent type defined in it.
+
+![Rename File](./gifs/RenameFile.gif)
+
+### Extract Value
+
+If you think your code would look nicer, if the calcuation of a value is extracted and assined to a new `val`; this code action would come to your aid.
+![Extract Value](./gifs/ExtractValue.gif)
```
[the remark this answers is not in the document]
vzmerr: ooops, yeah. Thank you so much for taking the time to try this 👍

docs/codeactions/code-actions.md
```diff
@@ -0,0 +1,93 @@
...
+
+![Extract Member Code Action](./gifs/ExtractMember.gif)
+
+## Rename File
+
+Renames a file to match that of the prominent type defined in it.
+
+![Rename File](./gifs/RenameFile.gif)
+
+### Extract Value
```
tanishiking, line 46:
```suggestion
## Extract Value
```

docs/codeactions/code-actions.md
```diff
@@ -0,0 +1,93 @@
...
+
+If you think your code would look nicer, if the calcuation of a value is extracted and assined to a new `val`; this code action would come to your aid.
+
+![Extract Value](./gifs/ExtractValue.gif)
+
+### map, flatMap, withFilter to For Yield
```
tanishiking, line 52:
```suggestion
## map, flatMap, withFilter to For Yield
```

docs/codeactions/code-actions.md
```diff
@@ -0,0 +1,93 @@
...
+
+![To For Comprehension](./gifs/FlatMapToForComprehension.gif)
+
+## Implement Abstract Members of the Parent Type
+
+Upon inherting from a type, you also have to implement its abstract members. But manually looking them all up and copying their signature is time consuming, isn't it? You can just use this code action instead.
+
+![Implement Abstract Members](./gifs/ImplementAbstractMembers.gif)
+
+### Import Missing Symbol
```
tanishiking, line 64:
```suggestion
## Import Missing Symbol 
```

tanishiking (approved):
I left nitpicks but LGTM!
BTW, I'm wondering maybe we should have a section (in `sidebar.json`) something like `Guide` and code actions could be under that 🤔 but nevermind, we don't have to work on it in this PR :)

tanishiking:
The `yarn build` is failing because

```
These sidebar document ids do not exist:
- codeactions/code-actions
```

instead we should say `codeactions/codeactions`

as the error mentions

```
Available document ids are:
- build-tools/bloop
...
- codeactions/codeactions
```

it doesn't have `hyphen` between `code` and `actions` because you set `id: codeactions` in https://github.com/scalameta/metals/pull/4157/files#diff-0e21d372368917c3f98bf0a8e612f99861c109b89be879f8b1a7bc90fbd9cd35R2

docs/codeactions/code-actions.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Import Missing Symbol 
+
+Because, of course, you often do not remember the package where it is located, without using this code action. right?
+
+![Import Missing Symbol](./gifs/ImportMissingSymbol.gif)
+
+## Insert Inferred Type
+
+This code action tries to infer the type of variables according to the context, and insert them for you.
```
dos65, line 72:
```suggestion
This code action infers the type of variables according to the context, and insert them for you.
```

docs/codeactions/code-actions.md
```diff
@@ -0,0 +1,93 @@
...
+---
+
+
+Code Actions are used to provide automatic refactoring or rewriting of your current code in order to enhance its readability or save users' time.
+
+In Visual Studio Code, to use them you can click on the yellow light bulb appearing on your code, near the cursor and then, to pick the desired action to perform on the code. Alternatively, you can use the `editor.action.quickFix` shortcut.
+
+For other editors this should also be available via specified shortcut.
+
+Here is a list of the code actions supported by Metals. Some of these code actions support much more complex use cases, which you would eventually discover through your code, by clicking on the magic 💡, as it appears, to get your surprise!
```
dos65, line 13:
```suggestion
Here is a list of the code actions supported by Metals.
```
I would omit or rephrase the last sentence here :smile_cat:
This part adds some ambiguity: what means `much more complex use cases`, `you would eventually discover` - when, where?

dos65 (commented):
Nice work! :tada:

docs/codeactions/code-actions.md
vzmerr: well, the current gifs are only covering the simplest cases for getting the message across. So let's say the use case of covering partial functions in map/flatMap is not shown, so the user would eventually discover this more complex one through clicking the light bulb.

dos65: > So let's say the use case of covering partial functions in map/flatMap is not shown

From user viewpoint there is no difference. I think, people would be more surprised if it doesn't work in some cases rather than it support some particular tree inside an expression.

> so the user would eventually discover 

Yep, sometimes actions are discovered by a user but I think it's a rare occasion.

I think, in general, interaction with actions comes not from light-bulb suggestion.
User know which actions Metals supports and calls them to avoid manual rewriting.
So, you want to specify a type of your val - you know that there is an action. The same with for-comp conversion and others. Usually you know in advance that some particular code should have a rewrite action.

vzmerr: ok, no problem.

tgodzik (approved):
LGTM! I just moved the gifs to the `static` folder next to other imgs

tgodzik:
> LGTM! I just moved the gifs to the `static` folder next to other imgs

Well, that totally failed :| It seems it was working locally, but due to some caching. Reverted the change

vzmerr:
> > LGTM! I just moved the gifs to the `static` folder next to other imgs
> 
> Well, that totally failed :| It seems it was working locally, but due to some caching. Reverted the change

perhaps the gifs in the static folder were not added to git, because they do not show up in the corresponding commit

PR #4161, opened 2022-07-18
bugfix: Include method signature in the label

tgodzik:
Previously, we would be getting a lot of messages such as:

```
Caused by: java.lang.IllegalArgumentException: 76 is not a valid line number, allowed [0..72]
    at scala.meta.internal.inputs.InternalInput.lineToOffset(InternalInput.scala:37)
    at scala.meta.internal.inputs.InternalInput.lineToOffset$(InternalInput.scala:32)
```
which most likely are cause by the fact that input is not the same as the one used for position (most likely file changed and wasn't reparsed again)

Now, we explicitely handle any issues when the position is not available.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
@@ -301,62 +301,67 @@ class Compilers(
     loadCompiler(path)
       .map { compiler =>
         val input = path.toInputFromBuffers(buffers)
-        val metaPos = breakpointPosition.toMeta(input)
-        val oldText = metaPos.input.text
-        val lineStart = oldText.indexWhere(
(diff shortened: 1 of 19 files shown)


tgodzik:
Btw. I used the new code action in a couple of places, great work @vzmerr !

tanishiking (commented):
The change looks good, but the commit message and PR title don't align with the change (?)

tanishiking (approved):
LGTM!
So, now we can handle the exception by ourselves, do we still need to investigate the root cause of the exceptions?

tgodzik:
> LGTM! So, now we can handle the exception by ourselves, do we still need to investigate the root cause of the exceptions?

If the position cannot be found in the document, scalameta throws an exception, so that's expected really, but we should handle it explicitely.

PR #4162, opened 2022-07-18
refactor: Print better debug infor when InferredType command failed

tgodzik:
Rather than using Futre.filter, which will cause a difficult to diagnose

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1999,7 +1999,11 @@ class MetalsLanguageServer(
           val uri = textDocumentParams.getTextDocument().getUri()
           for {
             edits <- compilers.insertInferredType(textDocumentParams, token)
-            if (!edits.isEmpty())
+            _ = if (edits.isEmpty())
+              scribe.error(
+                s"Could not infer type at ${textDocumentParams.getPosition()} in file\n${buffers
+                    .get(uri.toAbsolutePath)}"
+              )
             workspaceEdit = new l.WorkspaceEdit(Map(uri -> edits).asJava)
             _ <- languageClient
               .applyEdit(new ApplyWorkspaceEditParams(workspaceEdit))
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1999,7 +1999,11 @@ class MetalsLanguageServer(
           val uri = textDocumentParams.getTextDocument().getUri()
           for {
             edits <- compilers.insertInferredType(textDocumentParams, token)
-            if (!edits.isEmpty())
+            _ = if (edits.isEmpty())
+              scribe.error(
+                s"Could not infer type at ${textDocumentParams.getPosition()} in file\n${buffers
+                    .get(uri.toAbsolutePath)}"
+              )
```
kpodsiad, line 2006: What do you think about creating wrapper on `log if condition is met`
```scala
package object internal {
  private def scribeLogWhen(level: Level, condition: Boolean, msg: String)(implicit
      pkg: sourcecode.Pkg,
      fileName: sourcecode.FileName,
      name: sourcecode.Name,
      line: sourcecode.Line,
      mdc: MDC,
  ): Unit =
    if (condition)
      scribe.log(level, mdc, msg)

  def scribeLogError(condition: Boolean)(msg: String)(implicit
      pkg: sourcecode.Pkg,
      fileName: sourcecode.FileName,
      name: sourcecode.Name,
      line: sourcecode.Line,
      mdc: MDC,
  ): Unit = scribeLogWhen(Level.Error, condition, msg)
}
```
Then we could save on all those ifs :D

metals/src/main/scala/scala/meta/internal/metals/logging/package.scala
```diff
@@ -0,0 +1,32 @@
+package scala.meta.internal.metals
```
kpodsiad, line 1: if this package object is toplevel then it'll be possible to use just `logging` not matter where it will be used 🤔

tanishiking (approved):
LGTM from myside!

metals/src/main/scala/scala/meta/internal/metals/logging/package.scala
tgodzik: Not sure, what do you mean?

dos65 (approved):

PR #4164, opened 2022-07-19
Add Extract Method code action

jkciesluk:
Connected to https://github.com/scalameta/metals/issues/4165

Implement new code action that allows extracting blocks of code as methods and changes local variables in them to parameters.

image
image


kpodsiad:
If I may suggest something, it being able to specify scope to which method/value is about to be extracted will be awesome.

<details open>
<summary>Example from typescript language server</summary>
Screenshot 2022-07-14 at 09 34 45
</details>

tgodzik:
> If I may suggest something, being able to specify scope to which method/value is about to be extracted will be awesome.
> 

Actually the idea is that you can extract any range you select. In typescipt (which is actually not a language server :sweat_smile: ) it's an extension of extract value from what I see, which we could support functions for sure.

kpodsiad:
> Actually the idea is that you can extract any range you select. In typescipt (which is actually not a language server 😅 ) it's an extension of extract value from what I see, which we could support functions for sure.

I didn't meant **_what_** to extract but rather **_where_** to extract it ;)

jkciesluk:
> If I may suggest something, being able to specify scope to which method/value is about to be extracted will be awesome.
> 
> Example from typescript language server

I've tried this approach, but I'm not sure about it, because the amount of options grows very quickly, so it might be unreadable 
image

jkciesluk:
> I didn't meant **_what_** to extract but rather **_where_** to extract it ;)

Oh, good idea. I will work on it

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -409,6 +409,21 @@ class Compilers(
     }
   }.getOrElse(Future.successful(Nil.asJava))
 
+  def extractMethod(
+      params: TextDocumentPositionParams,
+      applRange: Integer,
+      lv: Integer,
```
tgodzik, line 415: We could send in the position to insert it into maybe? Wouldn't require making sure the lvl is the same in the compiler.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -476,6 +476,20 @@ object ServerCommands {
     """|This command should be sent in with the LSP [`TextDocumentPositionParams`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams)
        |""".stripMargin,
   )
+  final case class ExtractMethodParams(
+      position: TextDocumentPositionParams,
+      applRange: Integer,
+      lv: Integer,
+  )
```
tgodzik, line 483:
```suggestion
  final case class ExtractMethodParams(
      textDocument: TextDocumentIdentifier,
      range: Range, // from LSP
      extractPosition: Position,
  )
```
maybe this would make it easier to handle? And also `applRange` is a bit unclear, it's better to be explicit and send in the entire range

tests/unit/src/test/scala/tests/codeactions/ConvertToNamedArgumentsLspSuite.scala
```diff
@@ -24,6 +24,7 @@ class ConvertToNamedArgumentsLspSuite
        |  case class Foo(param1: Int, param2: Int, param3: Int)
        |  Foo(param1 = 1, param2 = 2, param3 = 3)
        |}""".stripMargin,
+    filterAction = filterAction,
```
tgodzik, line 27: Could you add the possible actions instead? I think it would be good to see all the possible actions to make sure that they don't pop up in unexpected places.

tests/unit/src/test/scala/tests/codeactions/ExtractMethodLspSuite.scala
```diff
@@ -0,0 +1,163 @@
...
+       |      newMethod(d)
+       |    }
+       |  }
+       |}""".stripMargin,
+    selectedActionIndex = 1,
+  )
+
+  check(
+    "simple-block",
+    """|object A{
```
tgodzik, line 134: Let's add a test that extracts multiple statements for example:
```
...
<<val a = 123
   val b = 123
   a + b >>
```

mtags/src/main/scala-3/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,156 @@
...
+import org.eclipse.lsp4j.Position
+import org.eclipse.lsp4j.TextEdit
+import org.eclipse.{lsp4j as l}
+
+final class ExtractMethodProvider(
+    params: OffsetParams,
+    applRange: Int,
+    lv: Int,
+    driver: InteractiveDriver,
+    config: PresentationCompilerConfig,
```
tgodzik, line 36: Doesn't seem to be needed

tests/unit/src/test/scala/tests/codeactions/RewriteBracesParensLspSuite.scala
```diff
@@ -52,6 +53,7 @@ class RewriteBracesParensLspSuite
        |}
        |""".stripMargin,
     s"""|${RewriteBracesParensCodeAction.toBraces("foreach")}
+        |${ExtractMethodCodeAction.title("object `Main`")}
```
tgodzik, line 56: Shouldn't we show the new code action only if a range is selected? Otherwise it's hard to figure out what to extract.

tests/slow/src/test/scala/tests/feature/Scala3CodeActionLspSuite.scala
```diff
@@ -405,6 +408,166 @@ class Scala3CodeActionLspSuite
...
+        |  val a = { 
+        |    val c = 5
+        |    newMethod(c)
+        |  }
+        |}""".stripMargin,
+    selectedActionIndex = 1,
+  )
+
+  check(
+    "higher-scope",
```
tgodzik, line 538: It's probably fine to just leave one or two tests and move everything else to cross tests,

tests/cross/src/test/scala/tests/pc/ExtractMethodSuite.scala
```diff
@@ -0,0 +1,89 @@
...
+import scala.meta.internal.metals.TextEdits
+
+import munit.Location
+import munit.TestOptions
+import org.eclipse.{lsp4j => l}
+import tests.BaseCodeActionSuite
+
+class ExtractMethodSuite extends BaseCodeActionSuite {
+
+  checkEdit(
```
tgodzik, line 16: Let's add more tests here so that we can test both Scala 2 and Scala 3 with compat options.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -409,6 +410,21 @@ class Compilers(
...
   }.getOrElse(Future.successful(Nil.asJava))
 
+  def extractMethod(
+      params: TextDocumentIdentifier,
+      range: LspRange,
+      defnPos: LspRange,
+      token: CancelToken,
+  ): Future[ju.List[TextEdit]] = {
+    withPCAndAdjustLsp(params, range) { (pc, pos, adjust) =>
+      pc.extractMethod(CompilerRangeParams.fromPos(pos, token), range, defnPos)
```
tgodzik, line 420:
```suggestion
      pc.extractMethod(CompilerRangeParams.fromPos(pos, token), range, adjust.adjustPos(defnPos))
```
might be needed in ammonite files

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -2012,6 +2012,20 @@ class MetalsLanguageServer(
...
           } yield ().asInstanceOf[Object]
         }
+      case ServerCommands.ExtractMethod(
+            ServerCommands.ExtractMethodParams(doc, range, defnPos)
+          ) =>
+        CancelTokens.future { token =>
+          val uri = doc.getUri()
+          for {
+            edits <- compilers.extractMethod(doc, range, defnPos, token)
+            if (!edits.isEmpty())
```
tgodzik, line 2022: Let's use  `logging.logErrorWhen`, otherwise we will get an unclear error with filter.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -476,6 +477,20 @@ object ServerCommands {
...
+      range: lsp4j.Range, // from LSP
+      extractPosition: lsp4j.Range,
+  )
+  val ExtractMethod = new ParametrizedCommand[ExtractMethodParams](
+    "extract-method",
+    "Extract method from range",
+    """|Whenever a user chooses code action to extract method, this command is later ran to
+       |calculate parameters for the new method with types.
+       |""".stripMargin,
+    """|This command should be sent in with the LSP [`TextDocumentPositionParams`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams) and `applRange` Integer
```
tgodzik, line 491: We should include information with `ExtractMethodParams` showed as a json like for example `DiscoverTestSuites`

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractMethodCodeAction.scala
```diff
@@ -0,0 +1,161 @@
...
+      val tree: Option[Tree] = trees.get(path)
+      def loop(appl: Tree): Option[Tree] = {
+        appl.children.find(_.pos.encloses(range)) match {
+          case Some(child) =>
+            loop(child)
+          case None =>
+            Some(appl)
+        }
+      }
+      val enclosing = tree.flatMap(loop(_))
```
tgodzik, line 40:
```suggestion
      val enclosing =trees.findLastEnclosingAt[Tree](
            path,
            range.getStart())
```
wouldn't work?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractMethodCodeAction.scala
```diff
@@ -0,0 +1,161 @@
...
+        case Term.Block(stats) =>
+          stats.filter((s: Tree) => range.encloses(s.pos.toLSP))
+        case Template(_, _, _, stats) =>
+          stats.filter((s: Tree) => range.encloses(s.pos.toLSP))
+        case ap if returnsValue(ap) => List(ap)
+        case _ => Nil
+      })
+    }
+
+    val edits = {
```
tgodzik, line 51:
```suggestion
    val function = {
```

mtags/src/main/scala-2/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,182 @@
...
+    val history = new ShortenedNames(
+      lookupSymbol = name =>
+        context.lookupSymbol(name, sym => !sym.isStale) :: Nil,
+      config = renameConfig,
+      renames = re
+    )
+    def prettyType(tpe: Type) =
+      metalsToLongString(tpe.widen.finalResultType, history)
+
+    def pathTo(tree: Tree, range: l.Range): List[Tree] = {
```
tgodzik, line 36: compiler.lastVisitedParentTrees wouldn't work?

mtags/src/main/scala-2/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,182 @@
...
+          case ValDef(_, name, tpt, _) if tpt.tpe != null =>
+            List((name, prettyType(tpt.tpe)))
+          case _ => Nil
+        }
+      )
+    }
+
+    def genName(path: List[Tree]): String = {
+      def defsOnPath(ts: List[Tree]): Set[String] = {
+        ts.flatMap(t =>
```
tgodzik, line 71: Could we do `ts.traverse` and only match on names? Might be simpler to do

mtags/src/main/scala-2/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,182 @@
...
+        case Select(qualifier, name) =>
+          if (!defns(name.toTermName)) names += name.toTermName
+          traverse(qualifier)
+        case ValDef(_, name, _, rhs) =>
+          defns.add(name)
+          traverse(rhs)
+        case _ => tree.children.foreach(traverse(_))
+      }
+
+      ts.map(traverse(_))
```
tgodzik, line 108: traverse on Tree already exists

mtags/src/main/scala-2/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,182 @@
...
+      while ((line(i) == ' ' || line(i) == '\t') && i < oldIndent) {
+        i += 1
+      }
+      newIndent + additional + line.drop(i)
+    }
+
+    val path = pathTo(unit.body, range)
+    val edits =
+      for {
+        enclosing <- path.find(_.pos.toLSP.encloses(range))
```
tgodzik, line 129:
```suggestion
        enclosing <- path.headOption
```
since we already check for enclosing?

mtags/src/main/scala-2/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,182 @@
...
+        extracted = extractFromBlock(enclosing)
+        head <- extracted.headOption
+        appl <- extracted.lastOption
+        shortenedPath =
+          path.takeWhile(src => defnPos.encloses(src.pos.toLSP))
```
tgodzik, line 134:
```suggestion
          path.takeWhile(src => defnRange.encloses(src.pos.toLSP))
```

mtags/src/main/scala-2/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,182 @@
...
+        val typs = withType
+          .map { case (name, tpe) => s"$name: $tpe" }
+          .mkString(", ")
+        val newAppl = typedTreeAt(appl.pos)
+        val applType =
+          if (newAppl.tpe != null) s": ${prettyType(newAppl.tpe.widen)}" else ""
+        val applParams = withType.map(_._1).mkString(", ")
+        val name = genName(path)
+        val text = params.text()
+        val indent = stat.pos.column - (stat.pos.point - stat.pos.start) - 1
```
tgodzik, line 152: Could we extract the indentation calculation to use in both Scala 2 and Scala 3?

mtags/src/main/scala-3/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,163 @@
...
+          temp.body.filter(stat => range.encloses(toLSP(stat.sourcePos)))
+        case other => List(other)
+
+    def localRefs(ts: List[tpd.Tree]): Set[TermName] =
+      val names = Set.newBuilder[TermName]
+      val valDefs = mutable.Set[TermName]()
+      class NamesCollector extends TreeTraverser:
+        override def traverse(t: tpd.Tree)(using Context): Unit =
+          t match
+            case Select(qualifier, name) =>
```
tgodzik, line 77: Can we match on Name?

mtags/src/main/scala-3/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,163 @@
...
+            case t @ Template(_, _, _, _) =>
+              defsOnPath(t.body)
+            case _ => Nil
+        ).toSet
+      val usedNames = defsOnPath(path)
+      if !usedNames("newMethod") then "newMethod"
+      else
+        var i = 0
+        while usedNames(s"newMethod$i") do i += 1
+        s"newMethod$i"
```
tgodzik, line 106: This could be extracted to use for both Scala 2 and 3 similar to how we reuse MEtalsNAmes

vzmerr:
Thank you for the amazing work! 
Since the PR is really big, I guess it would be better to stash the commits that address the same problem into a single one, and add descriptive commit messages, which would enable incremental review of the changes.

tgodzik:
> Thank you for the amazing work! Since the PR is really big, I guess it would be better to stash the commits that address the same problem into a single one, and add descriptive commit messages, which would enable incremental review of the changes.

I think that might be really hard to do right now unfortunately, since the work here is on a single new feature and that had some different iterations and fixes.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
tgodzik: Could you explain the parameters of `ExtractMethodParams` in the docs?

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
```diff
@@ -95,6 +96,11 @@ public abstract class PresentationCompiler {
      */
     public abstract CompletableFuture<List<TextEdit>> insertInferredType(OffsetParams params);
 
+    /**
+     * Extract method in selected range
+     */
+    public abstract CompletableFuture<List<TextEdit>> extractMethod(OffsetParams params, Range range, Range defnRange);
```
tgodzik, line 102: This could make the tests a bit easier and no need to forward LSP ranges.
```suggestion
    public abstract CompletableFuture<List<TextEdit>> extractMethod(OffsetParams params, RangeParams range, RangeParams defnRange);
```
Could you also add information

mtags-interfaces/src/main/java/scala/meta/pc/PresentationCompiler.java
```diff
@@ -95,6 +96,11 @@ public abstract class PresentationCompiler {
      */
     public abstract CompletableFuture<List<TextEdit>> insertInferredType(OffsetParams params);
 
+    /**
+     * Extract method in selected range
```
tgodzik, line 100: Could you add explanations about params?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractMethodCodeAction.scala
```diff
@@ -0,0 +1,165 @@
...
+        }
+        val enclosing = tree.flatMap(loop(_))
+        enclosing.map(_ match {
+          case Term.Block(stats) =>
+            stats.filter((s: Tree) => range.encloses(s.pos.toLSP))
+          case Template(_, _, _, stats) =>
+            stats.filter((s: Tree) => range.encloses(s.pos.toLSP))
+          case ap if returnsValue(ap) => List(ap)
+          case _ => Nil
+        })
```
tgodzik, line 51:
```suggestion
        enclosing.map{
          case Term.Block(stats) =>
            stats.filter((s: Tree) => range.encloses(s.pos.toLSP))
          case Template(_, _, _, stats) =>
            stats.filter((s: Tree) => range.encloses(s.pos.toLSP))
          case ap if returnsValue(ap) => List(ap)
          case _ => Nil
        }
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractMethodCodeAction.scala
```diff
@@ -0,0 +1,165 @@
...
+      }
+    }
+
+    val edits = {
+      for {
+        exprs <- toExtract
+        last <- exprs.lastOption
+        head <- exprs.headOption if returnsValue(last)
+        scopeOptions = enclosingList(head).flatMap(enclosingDef(_))
```
tgodzik, line 60:
```suggestion
        scopeOptions = enclosingList(head).flatMap(enclosingDef)
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractMethodCodeAction.scala
```diff
@@ -0,0 +1,165 @@
...
+      } yield {
+        (scopeOptions, last, head)
+      }
+    }
+    edits
+      .map { case (scopes: List[(Tree, Tree)], apply: Tree, head: Tree) =>
```
tgodzik, line 66:
```suggestion
      .map { case (scopes: List[(Tree, Tree)], expr: Tree, head: Tree) =>
```
?

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -476,6 +477,20 @@ object ServerCommands {
     """|This command should be sent in with the LSP [`TextDocumentPositionParams`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentPositionParams)
        |""".stripMargin,
   )
+  final case class ExtractMethodParams(
+      param: TextDocumentIdentifier,
+      range: lsp4j.Range, // from LSP
+      extractPosition: lsp4j.Range,
```
tgodzik, line 483:
```suggestion
      extractPosition: lsp4j.Position,
```
it seems we only need a position?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractMethodCodeAction.scala
```diff
@@ -0,0 +1,165 @@
...
+      case vr: Defn.Var =>
+        vr.pats.head match {
+          case Pat.Var(name) => s"var `${name}`"
+          case _ => "var"
+        }
+      case df: Defn.Def => s"method `${df.name}`"
+      case cl: Defn.Class => s"class `${cl.name}`"
+      case en: Defn.Enum => s"enum ${en.name}`"
+      case ob: Defn.Object => s"object `${ob.name}`"
+      case _ => "block"
```
tgodzik, line 112: Maybe if we have multiple blocks we should differentiate it with the first line?

tests/cross/src/test/scala/tests/pc/ExtractMethodSuite.scala
```diff
@@ -0,0 +1,381 @@
...
+    "simple-expr",
+    s"""|object A{
+        |  val b = 4
+        |  def method(i: Int) = i + 1
+        |  <@val a = <<123 + method(b)>>@>
+        |}""".stripMargin,
```
tgodzik, line 22:
```suggestion
    s"""|object A{
        |  val b = 4
        |  def method(i: Int) = i + 1
        |  @@val a = <<123 + method(b)>>
        |}""".stripMargin,
```
Does `<@` symbolises where to extract the range? Maybe we could just use a position with `@@`

mtags/src/main/scala-2/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,126 @@
...
+        val noLongerAvailable = scopeSymbols
+          .filter(sym =>
+            defnRange.encloses(sym.pos.toLSP) && !range.encloses(sym.pos.toLSP)
+          )
+        val names = localRefs(extracted)
+        val paramsToExtract = noLongerAvailable
+          .filter(sym => names.contains(sym.name))
+          .map(sym => (sym.name, sym.info))
+          .sortBy(_._1.decoded)
+        val newAppl = typedTreeAt(appl.pos)
```
tgodzik, line 79: Wouldn't it be the same tree as appl?

tgodzik (changes requested):
Looks really good! I left a few comments that could simplify a couple of things.

metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
```diff
@@ -655,6 +676,32 @@ class Compilers(
...
+          params,
+          compiler.scalaVersion(),
+        )
+      for {
+        metaRange <- adjust.adjustRange(range).toMeta(input)
+        metaExtractionPos <- adjust.adjustPos(extractionPos).toMeta(input)
+      } yield fn(compiler, metaRange, metaExtractionPos, adjust)
+    }
+  }
+
```
tgodzik, line 704: You should use`adjustRequest` for input

```suggestion
  private def withPCAndAdjustLsp[T](
      uri: String,
      range: LspRange,
      extractionPos: LspPosition,
  )(
      fn: (
          PresentationCompiler,
          Position,
          Position,
          AdjustLspData,
      ) => T
  ): Option[T] = {
    val path = uri.toAbsolutePath
    loadCompiler(path).flatMap { compiler =>
      val (input, adjustRequest, adjustResponse)=
        sourceAdjustments(
          uri,
          compiler.scalaVersion(),
        )
      for {
        metaRange <- adjustRequest.adjustRange(range).toMeta(input)
        metaExtractionPos <- adjustRequest.adjustPos(extractionPos).toMeta(input)
      } yield fn(compiler, metaRange, metaExtractionPos, adjustResponse)
    }
  }

```

mtags/src/main/scala-2/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,138 @@
...
+    val history = new ShortenedNames(
+      lookupSymbol = name =>
+        context.lookupSymbol(name, sym => !sym.isStale) :: Nil,
+      config = renameConfig,
+      renames = re
+    )
+    def prettyType(tpe: Type) =
+      metalsToLongString(tpe.widen.finalResultType, history)
+
+    val rangeSourcePos = Position.range(
```
tgodzik, line 39: Do you need to convert it here? We could just change the `encloses` method to work with RangeParams

mtags/src/main/scala-2/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,138 @@
...
+
+final class ExtractMethodProvider(
+    val compiler: MetalsGlobal,
+    range: RangeParams,
+    extractionPos: OffsetParams
+) extends ExtractMethodUtils {
+  import compiler._
+  def extractMethod: List[l.TextEdit] = {
+    val text = range.text()
+    val params = range.toOffset
```
tgodzik, line 20: `range` should also have `offset` method. Do we need to convert it and have the additional convertion method?

mtags/src/main/scala/scala/meta/internal/mtags/CommonMtagsEnrichments.scala
```diff
@@ -134,6 +134,14 @@ trait CommonMtagsEnrichments {
           CompilerRangeParams(params.uri, params.text, start, end, params.token)
       }
     }
+
+    def toOffset: OffsetParams =
```
tgodzik, line 138: This isn't needed. `RangeParams` has the same methods available and inherit from RangeParams

tgodzik (changes requested):
Some more suggestion, but it's getting ready to ship!

tgodzik (changes requested):
Just some minor comments, but I think we are close to having this mergable!

tgodzik (approved):
LGTM from my side! Great work!

I will merge tomorrow, unless anyone else still wants to take a look?

mtags/src/main/scala-2/scala/meta/internal/pc/ExtractMethodProvider.scala
```diff
@@ -0,0 +1,164 @@
...
+            Set(id.symbol).filter(s =>
+              (s.isTerm || s.isTypeParameterOrSkolem) && !s.isMethod
+            )
+        }
+        sym.filter(nonAvailable(_))
+      }
+
+      def traverse(symbols: Set[Symbol], tree: Tree): Set[Symbol] =
+        tree match {
+          case id @ Ident(_) =>
```
tgodzik, line 82:
```suggestion
          case id : Ident =>
```

tgodzik (changes requested):
There are some remaining issues in ExtractMethodCodeAction.scala, otherwise this looks ok.

@jkciesluk do we want to merge?

tgodzik (approved):
LGTM! Anyone has any comments to add?

Great work @jkciesluk !

PR #4168, opened 2022-07-19
Multiline string enhance

vzmerr:
optimise the call to wasTripleQuoted, so that it would only get called once

remove escaping of " for multilines

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
@@ -93,28 +93,19 @@ case class MultilineString(userConfig: () => UserConfiguration)
       line: String,
   ): Option[(Int, Boolean)] = {
     var lastTripleQuote = -1
-    var escaped = false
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/formatting/MultilineString.scala
```diff
@@ -93,28 +93,19 @@ case class MultilineString(userConfig: () => UserConfiguration)
...
-            tripleQuoteClosed = !tripleQuoteClosed
-            quoteNum = 0
-          }
-        } else {
-          escaped = !escaped
+        quoteNum = quoteNum + 1
+        if (quoteNum == 3) {
+          lastTripleQuote = i
+          tripleQuoteClosed = !tripleQuoteClosed
+          quoteNum = 0
```
tanishiking, line 105: [note] context https://github.com/scalameta/metals/pull/4151#discussion_r924487813

tanishiking (approved):
nice catch! LGTM :tada:

PR #4174, opened 2022-07-20
Scala 3 type completion

### LINKED ISSUE #3954: [Scala 3] Add new completions
In Scala 2 it's possible to have completions after `new CompletedName@@`. It would be good to have the same for Scala 3.

The potential solutions should take a look at `NewCompletions.scala` and related tests to see what needs to be supported but in general we would need:
- completions when the cursor is in the name after `new` keyword

Write it manually.

**Additional context**
This is part of the effort to create proper tooling for Scala 3.

**Search terms**:
new completions completions scala3

### LINKED ISSUE #3956: [Scala 3] Add type completions
In Scala 2 it's possible to have completions  completion inside a type position, example `val x: Map[Int, Strin@@]`. It would be good to have the same for Scala 3.

The potential solutions should take a look at `TypeCompletions.scala` and related tests to see what needs to be supported but in general we would need:
- complete types when at type position

Write it manually.

**Additional context**
This is part of the effort to create proper tooling for Scala 3.

**Search terms**:
type parameters completions scala3

vzmerr:
it **partially** resolves #3954 
it **partially** resolves #3956
-  **filtering out invalid trait, object, class, import, or apply method completions** according to position for scala 3.
-  adding **square bracket suffix** where relevant for type, new, and instantiation completions, for scala 3.
-  adding **curly brace suffix** in new completions of trait, abstract, java interface types for scala 3.
-  applying **filtering** without enriching with symbols on the **interpolate completions**

**Not covered in the PR:**

https://github.com/scalameta/metals-feature-requests/issues/292

- member completion for Scala 3 as in:
 ```
object A {
   new scala.Iterable@@
}
```
Covering it requires covering the case for `CompletionKind.Members` in `enrichWithSymbolSearch` and **filtering out the non-members,** in a similar way to the `postProcess` method in `CompletionApplication`.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -94,8 +94,8 @@ class Completions(
     def isObjectValidForPos = (!isTypePosition && !isNewPosition)
 
     /**
-     * classes abd traits are type symbols. They are not suitable for
-     *        instantiation or method call positions. Only apply methods or objects
+     * classes and traits are type symbols. They are not suitable for
+     *        instantiation or method call positions. Only apply methods and objects
      *        can be used in such positions.
      */
     def isClassOrTraitValidForPos = !isInstantiationOrMethodCallPos
@@ -119,10 +119,9 @@ class Completions(
   ): CursorPositionCondition =
     path match
       case (head: (Select | Ident)) :: tail =>
-        // this method of calculating hasNoSquareBracket instead of using the
-        // path for the purpose is due to the issue that trees after typer
-        // lose information in dotty:
         // https://github.com/lampepfl/dotty/issues/15750
+        // due to this issue in dotty, because of which trees after typer lose information,
+        // we have to calculate hasNoSquareBracket manually:
         val hasNoSquareBracket =
           val span: Span = head.srcPos.span
           if span.exists then
@@ -210,10 +209,31 @@ class Completions(
   private def getParams(symbol: Symbol) =
     lazy val extensionParam = symbol.extensionParam
     if symbol.is(Flags.Extension) then
-      symbol.paramSymss.filter(
-        !_.contains(extensionParam)
+      symbol.paramSymss.filterNot(
+        _.contains(extensionParam)
       )
     else symbol.paramSymss
+
+  private def isAbstractType(symbol: Symbol) =
+    (symbol.toString
+      .startsWith( // just in case the abstract or trait flags in dotty api
+        // are not correctly set
+        "trait"
+      ) // trait A{ def doSomething: Int}
+    // object B{ new A@@}
+    // Note: I realised that the value of Flag.Trait is flaky and
+    // leads to the failure of one of the DocSuite tests
+      || symbol.info.typeSymbol.isAllOf(
+        Flags.JavaInterface // in Java:  interface A {}
+        // in Scala 3: object B { new A@@}
+      ) || symbol.info.typeSymbol.isAllOf(
+        Flags.PureInterface // in Java: abstract class Shape { abstract void draw();}
+        // Shape has only abstract members, so can be represented by a Java interface
+        // in Scala 3: object B{ new Shap@@ }
+      ) || symbol.info.typeSymbol.is(Flags.Abstract)
+      // abstract class A(i: Int){ def doSomething: Int}
+      // object B{ new A@@}
+      || symbol.info.typeSymbol.isOneOf(Flags.AbstractOrTrait))
 
   private def findSuffix(symbol: Symbol): Option[String] =
 
@@ -252,25 +272,7 @@ class Completions(
 
     val templateSuffix =
       if shouldAddSnippet && cursorPositionCondition.isCurlyBracesValidForPos
-        // just in case the abstract or trait flags in dotty api
-        // are not correctly set
-        && (symbol.toString.startsWith(
-          "trait"
-        ) // trait A{ def doSomething: Int}
-        // object B{ new A@@}
-        // Note: I realised that the value of Flag.Trait is flaky and
-        // leads to the failure of one of the DocSuite tests
-          || symbol.info.typeSymbol.isAllOf(
-            Flags.JavaInterface // in Java:  interface A {}
-            // in Scala 3: object B { new A@@}
-          ) || symbol.info.typeSymbol.isAllOf(
-            Flags.PureInterface // in Java: abstract class Shape { abstract void draw();}
-            // Shape has only abstract members, so can be represented by a Java interface
-            // in Scala 3: object B{ new Shap@@ }
-          ) || symbol.info.typeSymbol.is(Flags.Abstract)
-          // abstract class A(i: Int){ def doSomething: Int}
-          // object B{ new A@@}
-          || symbol.info.typeSymbol.isOneOf(Flags.AbstractOrTrait))
+        && isAbstractType(symbol)
       then
         if bracketSuffix.nonEmpty || bracesSuffix.contains("$0") then " {}"
         else " {$0}"
@@ -516,17 +518,9 @@ class Completions(
               val include =
                 !isUninterestingSymbol(sym) &&
                   isNotLocalForwardReference(sym)
-                  && (!sym.info.typeSymbol.is(
-                    Flags.Module
-                  ) || cursorPositionCondition.isObjectValidForPos)
-                  && (cursorPositionCondition.isMethodValidForPos || !sym.is(
-                    Flags.Method
-                  ))
-                  // !sym.info.typeSymbol.is(Flags.Method) does not detect Java methods
-                  && (cursorPositionCondition.isClassOrTraitValidForPos || (!sym.isClass && !sym.toString
-                    .startsWith(
-                      "trait "
-                    )))
+                  && isNotAModuleOrModuleIsValidForPos(sym)
+                  && isNotAMethodOrMethodIsValidForPos(sym)
+                  && isNotClassOrTraitOrTheyAreValidForPos(sym)
 
               (id, include)
             case kw: CompletionValue.Keyword => (kw.label, true)
@@ -548,6 +542,25 @@ class Completions(
           enrichWithSymbolSearch(visit, qualType).getOrElse(SymbolSearch.Result.COMPLETE)
         (buf.result, searchResult)
       else (buf.result, SymbolSearch.Result.COMPLETE)
+
+    end filterInteresting
+
+    private def isNotAModuleOrModuleIsValidForPos(sym: Symbol) =
+      !sym.info.typeSymbol.is(
+        Flags.Module
+      ) || cursorPositionCondition.isObjectValidForPos
+
+    private def isNotAMethodOrMethodIsValidForPos(sym: Symbol) =
+      (cursorPositionCondition.isMethodValidForPos || !sym.is(
+        Flags.Method
+      )) // !sym.info.typeSymbol.is(Flags.Method) does not detect Java methods
+
+    private def isNotClassOrTraitOrTheyAreValidForPos(sym: Symbol) =
+      (cursorPositionCondition.isClassOrTraitValidForPos || (!sym.isClass && !sym.toString
+        .startsWith(
+          "trait "
+        )))
+  end extension
 
   private lazy val isUninterestingSymbol: Set[Symbol] = Set[Symbol](
     defn.Any_==,
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -167,6 +167,17 @@ class Completions(
     else symbol.paramSymss
 
   private def findSuffix(symbol: Symbol): Option[String] =
+
+    val bracketSuffix =
+      if shouldAddSnippet && noSquareBracketExists
+        && (isTypePosition || isNewPosition || isInstantiationOrMethodCallPos)
+        && (symbol.info.typeParams.nonEmpty
+          || (symbol.isAllOf(
+            Flags.JavaModule
+          ) && symbol.companionClass.typeParams.nonEmpty))
+      then "[$0]"
+      else ""
+
     val bracesSuffix =
       if shouldAddSnippet && symbol.is(
           Flags.Method
@@ -185,19 +196,10 @@ class Completions(
             )
             if onlyParameterless then "()" * paramss.length
             else if onlyImplicitOrTypeParams then ""
+            else if bracketSuffix == "[$0]" then "()"
             else "($0)"
           case _ => ""
         end match
-      else ""
-
-    val bracketSuffix =
-      if shouldAddSnippet && noSquareBracketExists
-        && (isTypePosition || isNewPosition || isInstantiationOrMethodCallPos)
-        && (symbol.info.typeParams.nonEmpty
-          || (symbol.isAllOf(
-            Flags.JavaModule
-          ) && symbol.companionClass.typeParams.nonEmpty))
-      then "[$0]"
       else ""
 
     val templateSuffix =
@@ -209,7 +211,9 @@ class Completions(
           Flags.PureInterface
         ) || symbol.info.typeSymbol.is(Flags.Abstract)
           || symbol.info.typeSymbol.isOneOf(Flags.AbstractOrTrait))
-      then " {}"
+      then
+        if bracketSuffix.nonEmpty || bracesSuffix.contains("$0") then " {}"
+        else " {$0}"
       else ""
 
     val concludedSuffix = bracketSuffix + bracesSuffix + templateSuffix
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -64,17 +64,59 @@ class Completions(
     case (_: Ident) :: (_: SeqLiteral) :: _ => false
     case _ => true
 
-  private lazy val (
-    isTypePosition,
-    isNewPosition,
-    isInstantiationOrMethodCallPos,
-    noSquareBracketExists,
-  ) = calculateTypeInstanceAndNewPositions(path)
+  /**
+   * It shows the conditions of the code in the cursor position, with
+   * respect to assessing
+   *  1. whether it makes semantic sense to complete
+   *  the position with symbols that are `class`, `trait`,
+   *  `abstract`, `object`, or method
+   *  1. whether it would make syntactic sense to add `[]` `()` or `{}`
+   *  suffixes to the symbols suggested for completion.
+   *
+   *  Note: In the comments `@@` represents the cursor position.
+   *
+   * @param isTypePosition if this is the position for defining a type such as
+   *                       `val a: In@@` or `def i: scala.IndexedSe@@`
+   * @param isNewPosition if it is the position for instantiation after
+   *                      the new keword, as in `new Fil@@`
+   * @param isInstantiationOrMethodCallPos if it is the position for method call or
+   *                                       instantiation as in `object A{ MyCaseClas@@ }`
+   * @param noSquareBracketExists if there is already a square bracket in the position as in
+   *                              `val a: IndexedMa@@[]`
+   */
+  private case class CursorPositionCondition(
+      isTypePosition: Boolean,
+      isNewPosition: Boolean,
+      isInstantiationOrMethodCallPos: Boolean,
+      noSquareBracketExists: Boolean,
+  ):
+
+    def isObjectValidForPos = (!isTypePosition && !isNewPosition)
+
+    /**
+     * classes abd traits are type symbols. They are not suitable for
+     *        instantiation or method call positions. Only apply methods or objects
+     *        can be used in such positions.
+     */
+    def isClassOrTraitValidForPos = !isInstantiationOrMethodCallPos
+    def isMethodValidForPos = !isTypePosition
+
+    def isParanethesisValidForPos = !isTypePosition
+    def isSquareBracketsValidForPos = noSquareBracketExists
+      && (isTypePosition || isNewPosition || isInstantiationOrMethodCallPos)
+
+    /**
+     * as in ```new MyTrait@@```
+     */
+    def isCurlyBracesValidForPos = isNewPosition
+  end CursorPositionCondition
+
+  private lazy val cursorPositionCondition =
+    calculateTypeInstanceAndNewPositions(path)
 
   private def calculateTypeInstanceAndNewPositions(
       path: List[Tree]
-  ): (Boolean, Boolean, Boolean, Boolean) =
-
+  ): CursorPositionCondition =
     path match
       case (head: (Select | Ident)) :: tail =>
         val hasNoSquareBracket =
@@ -89,26 +131,29 @@ class Completions(
         tail match
           case (v: ValOrDefDef) :: _ =>
             if v.tpt.sourcePos.contains(pos) then
-              (true, false, false, hasNoSquareBracket)
-            else (false, false, false, hasNoSquareBracket)
+              CursorPositionCondition(true, false, false, hasNoSquareBracket)
+            else
+              CursorPositionCondition(false, false, false, hasNoSquareBracket)
           case New(selectOrIdent: (Select | Ident)) :: _ =>
             if selectOrIdent.sourcePos.contains(pos) then
-              (false, true, false, hasNoSquareBracket)
-            else (false, false, false, hasNoSquareBracket)
+              CursorPositionCondition(false, true, false, hasNoSquareBracket)
+            else
+              CursorPositionCondition(false, false, false, hasNoSquareBracket)
           case (a @ AppliedTypeTree(_, args)) :: _ =>
             if args.exists(_.sourcePos.contains(pos)) then
-              (true, false, false, hasNoSquareBracket)
-            else (false, false, false, false)
-          case (_: Import) :: _ => (false, false, false, hasNoSquareBracket)
+              CursorPositionCondition(true, false, false, hasNoSquareBracket)
+            else CursorPositionCondition(false, false, false, false)
+          case (_: Import) :: _ =>
+            CursorPositionCondition(false, false, false, hasNoSquareBracket)
           case _ =>
-            (false, false, true, hasNoSquareBracket)
+            CursorPositionCondition(false, false, true, hasNoSquareBracket)
         end match
 
       case (_: TypeTree) :: TypeApply(Select(newQualifier: New, _), _) :: _
           if newQualifier.sourcePos.contains(pos) =>
-        (false, true, false, true)
+        CursorPositionCondition(false, true, false, true)
 
-      case _ => (false, false, false, true)
+      case _ => CursorPositionCondition(false, false, false, true)
     end match
   end calculateTypeInstanceAndNewPositions
 
@@ -169,8 +214,7 @@ class Completions(
   private def findSuffix(symbol: Symbol): Option[String] =
 
     val bracketSuffix =
-      if shouldAddSnippet && noSquareBracketExists
-        && (isTypePosition || isNewPosition || isInstantiationOrMethodCallPos)
+      if shouldAddSnippet && cursorPositionCondition.isSquareBracketsValidForPos
         && (symbol.info.typeParams.nonEmpty
           || (symbol.isAllOf(
             Flags.JavaModule
@@ -181,7 +225,7 @@ class Completions(
     val bracesSuffix =
       if shouldAddSnippet && symbol.is(
           Flags.Method
-        ) && !isTypePosition
+        ) && cursorPositionCondition.isParanethesisValidForPos
       then
         val paramss = getParams(symbol)
         paramss match
@@ -203,7 +247,9 @@ class Completions(
       else ""
 
     val templateSuffix =
-      if shouldAddSnippet && isNewPosition
+      if shouldAddSnippet && cursorPositionCondition.isCurlyBracesValidForPos
+        // just in case the abstract or trait flags in dotty api
+        // are not correctly set
         && (symbol.toString.startsWith(
           "trait"
         ) // trait A{ def doSomething: Int}
@@ -221,8 +267,6 @@ class Completions(
           // abstract class A(i: Int){ def doSomething: Int}
           // object B{ new A@@}
           || symbol.info.typeSymbol.isOneOf(Flags.AbstractOrTrait))
-        // just in case the abstract or trait flags in dotty api
-        // are not correctly set
       then
         if bracketSuffix.nonEmpty || bracesSuffix.contains("$0") then " {}"
         else " {$0}"
@@ -470,10 +514,12 @@ class Completions(
                   isNotLocalForwardReference(sym)
                   && (!sym.info.typeSymbol.is(
                     Flags.Module
-                  ) || (!isTypePosition && !isNewPosition))
-                  && (!isTypePosition || !sym.is(Flags.Method))
+                  ) || cursorPositionCondition.isObjectValidForPos)
+                  && (cursorPositionCondition.isMethodValidForPos || !sym.is(
+                    Flags.Method
+                  ))
                   // !sym.info.typeSymbol.is(Flags.Method) does not detect Java methods
-                  && (!isInstantiationOrMethodCallPos || (!sym.isClass && !sym.toString
+                  && (cursorPositionCondition.isClassOrTraitValidForPos || (!sym.isClass && !sym.toString
                     .startsWith(
                       "trait "
                     )))
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -64,7 +64,7 @@ class Completions(
     case (_: Ident) :: (_: SeqLiteral) :: _ => false
     case _ => true
 
-  private def calculateTypeInstanceAndNewPositions =
+  private def calculateTypeInstanceAndNewPositions() =
 
     path match
       case head :: tail
@@ -439,7 +439,6 @@ class Completions(
                 if sym.isClass || sym.is(Module) then
                   // drop #|. at the end to avoid duplication
                   name.substring(0, name.length - 1)
-                  // + sym.toString
                 else name
 
               val include =
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -405,7 +405,7 @@ class Completions(
             this,
             config.isCompletionSnippetsEnabled(),
           )
-          .filterInteresting(false)
+          .filterInteresting(enrich = false)
           ._1
         (completions, true)
       // From Scala 3.1.3-RC3 (as far as I know), path contains
@@ -537,7 +537,9 @@ class Completions(
 
       if enrich then
         val searchResult =
-          enrichWithSymbolSearch(visit, qualType).getOrElse(SymbolSearch.Result.COMPLETE)
+          enrichWithSymbolSearch(visit, qualType).getOrElse(
+            SymbolSearch.Result.COMPLETE
+          )
         (buf.result, searchResult)
       else (buf.result, SymbolSearch.Result.COMPLETE)
 
@@ -554,7 +556,9 @@ class Completions(
       ) // !sym.info.typeSymbol.is(Flags.Method) does not detect Java methods
 
     private def isNotClassOrTraitOrTheyAreValidForPos(sym: Symbol) =
-      cursorPositionCondition.isClassOrTraitValidForPos || (!sym.isClass && !sym.info.typeSymbol
+      cursorPositionCondition.isClassOrTraitValidForPos || sym.is(
+        Flags.Method
+      ) || (!sym.isClass && !sym.info.typeSymbol
         .is(Trait))
 
   end extension
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -63,6 +63,45 @@ class Completions(
     case (_: Ident) :: (_: SeqLiteral) :: _ => false
     case _ => true
 
+  private def calculateTypeInstanceAndNewPositions =
+
+    path match
+      case head :: tail
+          if head.isInstanceOf[Select] || head.isInstanceOf[Ident] =>
+        tail match
+          case (v: ValOrDefDef) :: _ =>
+            if v.tpt.sourcePos.contains(pos) then isTypePosition = true
+          //    pprint.log(path)
+          case New(selectOrIdent) :: _
+              if selectOrIdent.isInstanceOf[Select] || selectOrIdent
+                .isInstanceOf[Ident] =>
+            if selectOrIdent.sourcePos.contains(pos) then
+              //     pprint.log(selectOrIdent)
+              isNewPosition = true
+          case (a @ AppliedTypeTree(_, args)) :: _ =>
+            if args.exists(_.sourcePos.contains(pos)) then
+              //     pprint.log(a)
+              isTypePosition = true
+          case (_: Import) :: _ =>
+          case _ =>
+            //  pprint.log(tail)
+            isInstantiationOrMethodCallPos = true
+        //     pprint.log(tail)
+        end match
+
+      case (_: TypeTree) :: TypeApply(Select(newQualifier: New, _), _) :: _
+          if newQualifier.sourcePos.contains(pos) =>
+        //    pprint.log(newQualifier)
+        isNewPosition = true
+
+      case _ =>
+    end match
+  end calculateTypeInstanceAndNewPositions
+
+  private var isTypePosition = false
+  private var isNewPosition = false
+  private var isInstantiationOrMethodCallPos = false
+
   def completions(): (List[CompletionValue], SymbolSearch.Result) =
     val (advanced, exclusive) = advancedCompletions(path, pos, completionPos)
     val (all, result) =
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala [hunk 1]
```diff
@@ -258,7 +258,70 @@ class Completions(
       case Literal(Constant(null)) :: tl =>
         advancedCompletions(tl, pos, completionPos)
 
+      case (select: Select) :: tail =>
```
tgodzik, line 261: We could instead of having it here have:
```
lazy val isTypePosition = path match
  ... // all the cases you found
 case _ =>
   false
```

then when in filterInteresting you can add condition:
`!isTypePosition || isTypePosition && sym.isType`
which should filter out all the not needed types.

Then in `completionsWithSuffix` method you can change:
`shouldAddSnippet && sym.is(Flags.Module) && !isJavaDefined`

to
`shouldAddSnippet && sym.is(Flags.Module) && !isJavaDefined && !isTypePosition`

Although probably the change in filterInteresting would filter this out also.

Lastly, findSuffix can have an additional else condition:
 if shouldAddSnippet && methodSymbol.is(Flags.Method) then
  ...
 else if isTypePosition &&  methodSymbol.typeParams.nonEmpty then
   if shouldAddSnippet then Some("[$0]")
   else Some("[]")
 else 
....
```

Comment (tgodzik) [reply mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala:261]: The main issue is finding all the cases for `isTypePosition` which you already did (but that might require some tuning still)

Comment (vzmerr) [reply mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala:261]: Thank you so much for taking the time to add the tips 👍

**Comment by vzmerr** [comment]
help needed on detecting the existing type parameter in the following code:
`def foo ArrayBuffer@@[Int] = ???`
I have tried printing different members of the `symbol`, including its `paramSymss`; as well as  printing the `Tree` `path`. But I could not get to detecting the existing type parameter.

File: mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala [hunk 2]
```diff
@@ -102,33 +141,87 @@ class Completions(
...
+        end match
+      else ""
+
+    // symbol.info.typeParams.foreach(param =>
+    //   pprint.log(param.paramInfo)
+    // )
+
+    val bracketSuffix =
+      if shouldAddSnippet
+        && (isTypePosition || isNewPosition || isInstantiationOrMethodCallPos)
```
tgodzik, line 182:
```suggestion
     
    val shouldAddSquareBracket = path match {
      case (_ : Ident) :: (_ : AppliedTypeTree ) :: _ => false
      case (i : Ident):: _ =>
        val span = i.srcPos.span
        pprint.log(text(span.end))
        if (span.exists){ // TODO check spaces/tabs
          text(span.end) != '['
        } else false
      case _ => false
    }

    val bracketSuffix =
      if shouldAddSnippet && shouldAddSquareBracket
        && (isTypePosition || isNewPosition || isInstantiationOrMethodCallPos)
```

seems the ArrayBuffer[Int] is not seens as TypeApply, which is why I think we need to do a workaround. Something like the above

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
```diff
@@ -230,9 +225,8 @@ class CompletionSnippetSuite extends BaseCompletionSuite {
            |""".stripMargin,
       "3" ->
         // scala 3 type completions not implemented, so no way to distinguish if we are at a type position
```
tgodzik, line 227: You can remove the comment above and the other places that we now properly support type completions.

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
```diff
@@ -373,12 +366,9 @@ class CompletionSnippetSuite extends BaseCompletionSuite {
...
     // additional completion when apply method is present
     compat = Map(
-      "3" ->
+      "3" -> // TODO: please double check
         """|Try
            |Try($0)
-           |TryBlock
-           |TryModule
-           |TryMethods
            |TryMethods
```
tgodzik, line 372: Shouldn't it be as in Scala 2 ?
```suggestion
           |TryBlock
```

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -353,13 +353,26 @@ class CompletionSuite extends BaseCompletionSuite {
        |ProcessBuilder - scala.sys.process
        |CertPathBuilder - java.security.cert
        |CertPathBuilderSpi - java.security.cert
-       |ProcessBuilderImpl - scala.sys.process
```
tgodzik, line 356: Interesting, seems it's a private trait, so it shouldn't show up for Scala 2 :thinking: 

Not important in this PR though.

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -910,9 +922,8 @@ class CompletionSuite extends BaseCompletionSuite {
        |""".stripMargin,
     compat = Map(
```
tgodzik, line 923: You can remove the entire compat section

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -926,8 +937,7 @@ class CompletionSuite extends BaseCompletionSuite {
...
       "3" ->
-        """|ListBuffer[A](elems: A*): CC[A]
-           |ListBuffer - scala.collection.mutable
+        """|ListBuffer - scala.collection.mutable
```
tgodzik, line 940: Also here, compat is no longer needed.

tgodzik (changes requested):
Thanks for working on this! I left a lot of comments, but almost all of them are minor. I think the PR improves things a lot.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -63,6 +64,48 @@ class Completions(
...
+          case New(selectOrIdent) :: _
+              if selectOrIdent.isInstanceOf[Select] || selectOrIdent
+                .isInstanceOf[Ident] =>
+            if selectOrIdent.sourcePos.contains(pos) then isNewPosition = true
+          case (a @ AppliedTypeTree(_, args)) :: _ =>
+            if args.exists(_.sourcePos.contains(pos)) then isTypePosition = true
+            else shouldAddSquareBracket = false
+          case (_: Import) :: _ =>
+          case _ =>
+            isInstantiationOrMethodCallPos = true
```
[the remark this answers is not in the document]
vzmerr: yes, there are cases such as 
```
object A {
  Str@@
}
```
or
```
object A{
   Try@@
}
```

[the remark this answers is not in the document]
tgodzik: But we don't want to add `[]` in those cases, no?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -332,15 +408,25 @@ class Completions(
...
                   name.substring(0, name.length - 1)
+                  // + sym.toString
                 else name
 
               val include =
                 !isUninterestingSymbol(sym) &&
                   isNotLocalForwardReference(sym)
+                  && (!sym.toString.startsWith(
+                    "object "
+                  ) || (!isTypePosition && !isNewPosition))
```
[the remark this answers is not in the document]
vzmerr: Thanks for suggesting this. It is definitely cleaner, but sometimes the flags on symbol are not correctly set. I would try them on the `.info` field and see.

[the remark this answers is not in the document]
vzmerr: oh, I remembered, there was no specific flag for `object`

tests/cross/src/test/scala/tests/pc/CompletionInterpolatorSuite.scala
```diff
@@ -368,7 +373,12 @@ class CompletionInterpolatorSuite extends BaseCompletionSuite {
       "2.12" ->
         """|empty[A]: List[A]
            |equals(x$1: Any): Boolean
-           |""".stripMargin
+           |""".stripMargin,
+      "3" ->
+        """|empty[A]: List[A]
+           |eq(x$0: Object): Boolean
```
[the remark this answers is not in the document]
vzmerr: right 🧐 let me check

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
vzmerr: Scala 2 is not filtering some things correctly. In this specific case, I remember `TryBlock` was either a `trait` or a `class`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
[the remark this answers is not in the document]
tgodzik: It's actually `Module`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -63,6 +64,48 @@ class Completions(
...
+
+            if (i < text.length()) then text(i) != '['
+            else true
+          else false
+        tail match
+          case (v: ValOrDefDef) :: _ =>
+            if v.tpt.sourcePos.contains(pos) then isTypePosition = true
+          case New(selectOrIdent) :: _
+              if selectOrIdent.isInstanceOf[Select] || selectOrIdent
+                .isInstanceOf[Ident] =>
```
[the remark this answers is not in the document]
vzmerr: oh, unfortunately, a condition is getting checked there... so we need the variable name. I mean I could not find a way to extend the variable name to also apply on `Ident`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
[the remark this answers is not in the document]
vzmerr: sorry I forgot to mention  this extra case:
```
case class Blue[T](model: T)
object A{
   Bl@@
}
```

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -102,33 +145,64 @@ class Completions(
...
+              }
+            )
+            if onlyParameterless then "()" * paramss.length
+            else if onlyImplicitOrTypeParams then ""
+            else "($0)"
+          case _ => ""
+        end match
+      else ""
+
+    val bracketSuffix =
```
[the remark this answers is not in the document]
vzmerr: interestingly, the `$0` is only inserted in the square bracket in such cases.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -102,33 +145,64 @@ class Completions(
...
+          || (symbol.isAllOf(
+            Flags.JavaModule
+          ) && symbol.companionClass.typeParams.nonEmpty))
+      then "[$0]"
+      else ""
+
+    val templateSuffix =
+      if shouldAddSnippet && isNewPosition && (symbol.info.typeSymbol.is(
+          Flags.Trait
+        ) || symbol.info.typeSymbol.isAllOf(
```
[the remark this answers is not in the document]
vzmerr: Yes, sure.

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
```diff
@@ -93,7 +93,8 @@ class CompletionSnippetSuite extends BaseCompletionSuite {
 
   checkSnippet(
     // scala 3 type completions not implemented, so no way to distinguish if we are at a type position
-    "type-empty".tag(IgnoreScala3),
+    "type-empty"
+      .tag(IgnoreScala3),
```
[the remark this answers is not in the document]
vzmerr: oh, unfortunately, this was the fuzzy match which was not implemented in Dotty for Scala 3.

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
```diff
@@ -106,7 +107,8 @@ class CompletionSnippetSuite extends BaseCompletionSuite {
 
   checkSnippet(
     // scala 3 type completions not implemented, so no way to distinguish if we are at a type position
-    "type-new-empty".tag(IgnoreScala3),
+    "type-new-empty"
```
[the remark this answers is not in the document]
vzmerr: unfortunately, this was the fuzzy match, which dotty does not support yet :(

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
```diff
@@ -132,14 +134,14 @@ class CompletionSnippetSuite extends BaseCompletionSuite {
...
-        """|IndexedSeq
-           |IndexedSeq
+        """|IndexedSeq[$0]
            |""".stripMargin
     ),
   )
 
   checkSnippet(
-    "type2".tag(IgnoreScala3),
+    "type2"
```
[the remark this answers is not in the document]
vzmerr: It should actually. Let me check that.

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
vzmerr: yes, Metals should not suggest private members. Room for investigation, later.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -63,6 +64,48 @@ class Completions(
     case (_: Ident) :: (_: SeqLiteral) :: _ => false
     case _ => true
 
+  private def calculateTypeInstanceAndNewPositions =
+
+    path match
+      case head :: tail
+          if head.isInstanceOf[Select] || head.isInstanceOf[Ident] =>
```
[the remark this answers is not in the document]
vzmerr: hmmm.... it seems that we are using the `head` in later getting `span`. Is there a clean way to both have access to `head` and check for both `Ident` and `Select`?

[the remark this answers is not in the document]
tgodzik: Maybe union type?

`(selOrIdent : (Select | Ident) )`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -63,6 +64,48 @@ class Completions(
...
+        isNewPosition = true
+
+      case _ =>
+    end match
+  end calculateTypeInstanceAndNewPositions
+
+  private var isTypePosition = false
+  private var isNewPosition = false
+  private var isInstantiationOrMethodCallPos = false
+  private var shouldAddSquareBracket = true
```
tanishiking, line 107: Can we make all of them return values of `calculateTypeInstanceAndNewPositions`? (making them immutable?)
we shouldn't use mutable fields as much as possible if it's not needed because it makes it hard to debug (in the future) / understand the code in general.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -341,6 +418,14 @@ class Completions(
               val include =
                 !isUninterestingSymbol(sym) &&
                   isNotLocalForwardReference(sym)
+                  && (!sym.toString.startsWith(
+                    "object "
+                  ) || (!isTypePosition && !isNewPosition))
+                  && (!isTypePosition || !sym.toString.startsWith("method "))
+                  && (!isInstantiationOrMethodCallPos || (!sym.toString
+                    .startsWith("class ") && !sym.toString.startsWith(
+                    "trait "
```
tanishiking, line 427: Same here, use a more robust way like flags. You can find what makes those kind strings in `Symbol.toString` here https://github.com/lampepfl/dotty/blob/b1afa225353100fc47092edf923af2b946ff88cd/compiler/src/dotty/tools/dotc/core/SymDenotations.scala#L1491-L1500
(you can navigate to here with a few hops by find definitions on `Symbol.toString`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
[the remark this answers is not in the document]
tanishiking: I think it's still needed. **Currently** we never put multiple suffixes with `$0`  at the same time.

However, it's extremely difficult for me to understand there won't be multiple `$0` because each of them is not separated by `if/else if/else`.  We (Metals developers in the future) have to read all of the code around here to check all those conditions won't be all true in any cases, to prevent Metals from putting multiple braces. That leads to the code being error-prone.

tanishiking (changes requested):

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -63,6 +64,48 @@ class Completions(
...
+  private def calculateTypeInstanceAndNewPositions =
+
+    path match
+      case head :: tail
+          if head.isInstanceOf[Select] || head.isInstanceOf[Ident] =>
+        shouldAddSquareBracket =
+          val span: Span = head.srcPos.span
+          if span.exists then
+            var i = span.end
+            while i < (text.length() - 1) && text(i).isWhitespace do i = i + 1
```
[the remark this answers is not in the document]
vzmerr: This is related to this issue in Dotty https://github.com/lampepfl/dotty/issues/15750
because of which we had to implement this workaround to detect the existing type parameters to avoid inserting a duplicate extra bracket in Type Completion.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: Good idea 👍 I would try and see. I remember once when I was calling `calculateTypeInstanceAndNewPositions` for doing that though, and surprisingly, the values were not calculated correctly. I need to see this time.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: Thank you 🙏 super helpful

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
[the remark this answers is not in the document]
vzmerr: It requires implementing the case for `CompletionKind.Members` in `enrichWithSymbols`, since nothing is returned from the compiler.
So it can get done in this PR or maybe handled in a new one instead?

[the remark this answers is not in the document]
vzmerr: update: I'm already working on it. It turns out that it is very much similar to handling `CompletionKind.Scope`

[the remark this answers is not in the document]
vzmerr: update: It requires separate substantial work in filtering non-member suggestions, which also cause failure in the currently passing tests. Better be delegated to a followup PR, IMHO, because the code for it should go in `enrichWithSymbols` which is not currently touched in this PR.

tanishiking:
Is this ready for review? Please let me know when it's ready by `re-request review` :)

tests/cross/src/test/scala/tests/pc/CompletionInterpolatorSuite.scala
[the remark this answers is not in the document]
vzmerr: This is `exclusive`ly coming from `InterpolateCompletions`
Currently, the `exclusive` `advanced` completions do not go through `filterInteresting`
I can apply filtering on the results of `InterpolateCompletions`, though.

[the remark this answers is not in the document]
vzmerr: albeit with `enrichWithSymbolSearch`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -102,33 +151,80 @@ class Completions(
...
+          case List(Nil) => "()"
+          case _ if config.isCompletionSnippetsEnabled =>
+            val onlyParameterless = paramss.forall(_.isEmpty)
+            lazy val onlyImplicitOrTypeParams = paramss.forall(
+              _.exists { sym =>
+                sym.isType || sym.is(Implicit) || sym.is(Given)
+              }
+            )
+            if onlyParameterless then "()" * paramss.length
+            else if onlyImplicitOrTypeParams then ""
```
tanishiking, line 191: what if we have `foo()(implicit x: Int) = ???`?
I guess both `onlyParameterless` and `onlyImplicitOrTypeParams` will be `false`, but `onlyImplicitOrTypeParams` should be `true`?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -102,33 +151,80 @@ class Completions(
...
+            else "($0)"
+          case _ => ""
+        end match
+      else ""
+
+    val templateSuffix =
+      if shouldAddSnippet && isNewPosition
+        && (symbol.toString.startsWith(
+          "trait"
+        ) // trait A{ def doSomething: Int}
```
tanishiking, line 202:
```suggestion
        && (symbol.is(Trait)) // trait A{ def doSomething: Int}
```

tanishiking: can you explain which test will fail? I think it shouldn't be flaky (if it is, it's a compiler bug and we should report it)

tanishiking: What is the meaning of `// trait A{ def doSomething: Int}` ?
do you mean we will add `{}` suffix when we type-complete `trait A`? (in that case, `def doSomething: Int` part is redundant, I guess.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -102,33 +151,80 @@ class Completions(
...
+        // object B{ new A@@}
```
tanishiking, line 203: what the meaning of `@@`, I imagine this represents cursor position (based on how we write completionSuite in Metals), but it doesn't make sense.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -244,15 +340,18 @@ class Completions(
...
+          .contribute(
+            pos,
+            completionPos,
+            indexedContext,
+            lit,
+            path,
+            this,
+            config.isCompletionSnippetsEnabled(),
+          )
+          .filterInteresting(false)
```
tanishiking, line 353: Why do we need to `filterInteresting` here? Metals will call `filterInteresting` later, right?
`filterInteresting` is a bit heavy operation even if we don't invoke `enrich`, so we shouldn't call it multiple times.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -341,6 +443,15 @@ class Completions(
...
                   isNotLocalForwardReference(sym)
+                  && (!sym.info.typeSymbol.is(
+                    Flags.Module
+                  ) || (!isTypePosition && !isNewPosition))
+                  && (!isTypePosition || !sym.is(Flags.Method))
+                  // !sym.info.typeSymbol.is(Flags.Method) does not detect Java methods
+                  && (!isInstantiationOrMethodCallPos || (!sym.isClass && !sym.toString
+                    .startsWith(
+                      "trait "
+                    )))
```
tanishiking, line 454: can we move this filtering logic into `findSuffix` or somewhere related to suffix thing?
I think `filterInteresting` should filter symbols purely by those symbols properties, and suffix specific logic shouldn't be here.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: Thank you so much for taking the time to look into this. 

Actually, this was the preexisting parenthesis suffix code. I only changed the return type from Option[String] to String. I did not touch the logic otherwise, just in case this would break something.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: When putting Traits and Abstracts after `new`, we have to put `{}` for implementing the abstract members.
the sample is added in the following line.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: Why? 
```
object {
    new A{
      override def doSomething: Int = 2
    }
}

```

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: Unfortunately, `filterInteresting` is not applied on any result from the advanced completions. So in case of interpolate completion, this was generating results like `.eq(object: Object)` which should have been filtered.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: well, indeed the `isInstantiationOrMethodCallPos` `isNewPosition` and `isTypePosition` have double use:

- for knowing whether to add the suffix or not,
- for knowing whether to keep a result or filter it out, for example, in the case that you've highlighted, we should not allow this code, regardless of what the suffix is, because `A` is a `trait` and it would be semantically invalid to put a `trait` in instantiation or method call positions.

```
trait A {}

object B{
  @@
}

```

as another case, we should not be putting `object`s in the `TypePosition`s. So for example, the following code is semantically invalid regardless of the suffix, because only `class` or `trait` can be types.

```
object A {
   def apply(): A = ???
 }

object B{
    val a: A@@ = null
}

```
Similarly, the apply method of `A` is a wrong completion for the type position.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tanishiking: okay, but it should be fixed anyway. You can fix it in this PR or open an issue to track the issue :)

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: I wanted to apply filterInteresting on the results of `advancedCompletions` too but did not do that just in case there was a good reason to not write it like that, initially. Do you think it would be a good idea to explore that avenue?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -63,6 +64,54 @@ class Completions(
...
+  private lazy val (
+    isTypePosition,
+    isNewPosition,
+    isInstantiationOrMethodCallPos,
+    noSquareBracketExists,
+  ) = calculateTypeInstanceAndNewPositions(path)
+
+  private def calculateTypeInstanceAndNewPositions(
+      path: List[Tree]
+  ): (Boolean, Boolean, Boolean, Boolean) =
```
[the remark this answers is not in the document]
vzmerr: That's such an excellent idea for structuring the code indeed 👍

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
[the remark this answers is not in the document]
tanishiking: can you add a comment with a link to the issue?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tanishiking: I don't think so, because

- `foo()(implicit x: int)` will have `paramss = List(Nil, List(implicit x))`.
-  `Nil.exists(pred)` will be false for any pred
- therefore, `paramss.forall(...)` will be false

no?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tanishiking: Ah, I understand, thanks!

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tanishiking: I misunderstood in a same way.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tanishiking: Yes, I think we should do that (actually Scala2 does), but we can do this in another PR :+1:

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tanishiking: Ah, I realized this PR is doing filter out symbols based on positions, on top of the type-completion :+1:

Can you extract these lines into a method whose name is something we can tell it filters out invalid trait ... etc completions according to position?

vzmerr: Precisely. I was actually doing it in the new case class you suggested, which was inspired by putting the suffix related methods there 👍

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tanishiking, line 202:
```suggestion
        && (symbol.is(Trait)) // trait A{ def doSomething: Int}
```

can you explain which test will fail? I think it shouldn't be flaky (if it is, it's a compiler bug and we should report it)

tanishiking: (sorry for duplicated comment, but I want to discuss this in another thread)

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tanishiking: Yeah, let's check how it works.
even if it works weirdly, you don't need to fix that in this PR :+1:

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: oh, I would add the test case. Generally the flags are flaky. I have it on my todo list to put the issue on dotty and add the link in here. As an example, checking the `is(Trait)` flag on some cases was retuning `false`, while calling it on `.info.typeSymbol.` was returning `true`. We discussed it with Vadim on Slack, but certainly it has to get formally documented somewhere.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: Let me check and if no test breaks as a result, then I would do it here. but if the tests behave weirdly, I would postpone it to the next PR.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr:
```
checkSnippet(
    "empty-params-with-implicit".only,
    s"""|object Main {
        |  def doSomething()(implicit x: Int) = x
        |  val bar = doSomethi@@
        |}
        |""".stripMargin,
      "doSomething($0)",
  )
```
passing 👍

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: I filtered all the advanced completions and no test failed as a result. So do I keep it this way for consistency with what we have for Scala 2?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: this was the type5 test that used to fail before but now passes. So I reverted to using flags 👍

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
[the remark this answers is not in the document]
tgodzik: Do we have an issue about it? If not, could you raise it?

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
tgodzik: och makes sens. And TryMethods is a given, so looks correct.

tgodzik (approved):
LGTM from my side! Just one comment.

tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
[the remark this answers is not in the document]
vzmerr: https://github.com/scalameta/metals-feature-requests/issues/292 I wanted to self-assign but not having permission

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: Update: the reason the demo was failing was the use of `Flags.Trait` instead of `toString.startsWith("trait ")` 
Now that I reverted to using toString, the demo works correctly. I am now scratching my head.

tgodzik: Maybe it's not a trait, but a Java interface or an abstract class?

vzmerr: I was thinking about that. But it is actually coming from `scala.collection.mutable`

vzmerr: I wonder how I can add a test that fails with `Flags.Trait` and why `type5` is now passing regardless.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: it turned out that filtering the advanced completions was a huge performance toll. So I guess that was not a good idea.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: It turned out that it was Intellisense that was messing up with things.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -244,15 +385,17 @@ class Completions(
...
+        val completions = InterpolatorCompletions
+          .contribute(
+            pos,
+            completionPos,
+            indexedContext,
+            lit,
+            path,
+            this,
+            config.isCompletionSnippetsEnabled(),
+          )
```
kpodsiad, line 397: As far I can see there are no changes here, only formatting, can you revert it? (less modifications is better)

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -341,6 +487,15 @@ class Completions(
...
                   isNotLocalForwardReference(sym)
+                  && (!sym.info.typeSymbol.is(
+                    Flags.Module
+                  ) || cursorPositionCondition.isObjectValidForPos)
+                  && (cursorPositionCondition.isMethodValidForPos || !sym.is(
+                    Flags.Method
+                  ))
+                  // !sym.info.typeSymbol.is(Flags.Method) does not detect Java methods
+                  && (cursorPositionCondition.isClassOrTraitValidForPos || (!sym.isClass && !sym
+                    .is(Flags.Trait)))
```
kpodsiad, line 498: Same here, can you extract something to `def` to end up only with `&&` operators?

kpodsiad (commented):
I wonder, can code from `Completions.scala` can be extracted to object `TypeCompletions.scala`?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -63,23 +64,121 @@ class Completions(
...
+
+  private def calculateTypeInstanceAndNewPositions(
+      path: List[Tree]
+  ): CursorPositionCondition =
+    path match
+      case (head: (Select | Ident)) :: tail =>
+        // this method of calculating hasNoSquareBracket instead of using the
+        // path for the purpose is due to the issue that trees after typer
+        // lose information in dotty:
+        // https://github.com/lampepfl/dotty/issues/15750
```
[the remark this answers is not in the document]
vzmerr: sorry for the confusion, I think I forgot the punctuation. Does it read better with the comma:

```
 // this method of calculating hasNoSquareBracket, instead of using the
 // path for this purpose, is due to the issue that trees after typer
 // lose information in dotty:
```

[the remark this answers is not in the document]
kpodsiad: Yeah, now it's better. Although I'd just leave sth like that
`due to the issue in dotty, because of which trees after typer lose information, we have to calculate hasNoSquareBracket manually
`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -63,23 +64,121 @@ class Completions(
...
+          if newQualifier.sourcePos.contains(pos) =>
+        CursorPositionCondition(false, true, false, true)
+
+      case _ => CursorPositionCondition(false, false, false, true)
+    end match
+  end calculateTypeInstanceAndNewPositions
+
   def completions(): (List[CompletionValue], SymbolSearch.Result) =
     val (advanced, exclusive) = advancedCompletions(path, pos, completionPos)
+    val filteredAdvanced = advanced.filterInteresting(false)._1
```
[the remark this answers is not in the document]
vzmerr: oh, indeed, I realised that the filtration of advanced completions takes a huge performance toll. So I would be removing it, and only apply it on interpolate completions where it is absolutely necessary.

vzmerr:
**Update**

There seems to be a Dotty bug which causes the `requiredPackage` in `toSymbols` method of `CompilerSearchVisitor` to drop `java` from `java.io` and turn it into just `object io`
But types such as `java.io.StringBufferInputStream` must be specified with their package, or they won't be recognised by the compiler. 
I am not familiar with the exact purpose of the `toSymbols` method` though. Help would be appreciated.

This bug is causing the `StringBufferInputStream` to be suggested as a completion to `StringBuffe@@` without any clue for the user about in which package it can be found.

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
vzmerr: I put it in an issue to later work on: https://github.com/scalameta/metals/issues/4206

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: I created this Issue in Dotty to explain the problem https://github.com/lampepfl/dotty/issues/15790

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tanishiking: I want to filter advanced completion not because I want to make it consistent with Scala2, but more like consistent across all advanced completions (I feel it's wired if we filter only StringInterpolation and not others).
I don't think it makes so much performance deterioration 🤔 but it's ok, we can take a look later :+1:

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tanishiking: Thank you for taking a look!
Let's go with the current implementation, and we can refactor later 👍

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: well, basically because of the interruption exception issue, it is very difficult to have a confident opinion on the performance impact. I was as well surprised that this was causing a performance hog, considering the very small size of the advanced completions. But let's solve it later.

tanishiking (approved):
Looks like unit tests are failing, but it's LGTM:+1: thank you so much for working on this!

> Also, technically speaking, I do not understand the exact purpose of the toSymbols method there. We are feeding it java.io which is the correct package. and also StringBufferInputStream. Then the method just messes it up and erases the package information in producing the symbol. What's the purpose of that?
> I can fix it in this PR or I can put it in a followup?

Thank you for finding the issue!  You can create an issue and work on that in a follow-up 👍 

---


For what it's worth, in my opinion, if I were you, I will make a PR as small as possible and try to do one thing in one PR.
For example

- You're doing 4 things in this PR (as you explained in the PR description)
  - Instead, you might be able to create a PR for each thing.
  - (Of course, you can do more than one thing if it's comfortable for you, or if the changes are an atomic changes).
  - If you do multiple things in one PR
    - PR will be bigger, if PR is bigger, it takes more time to merge the PR in general 😭 
    - it makes it harder for reviews to review the diff because reviews can't tell which parts of diff are for which purpose.
      - If you want to do more than one thing, it would be good if you can explicitly separate the commits, https://github.com/scalameta/metals/pull/3790 is a good example. This PR is doing two things, but he did those things in a different commit. Therefore we could (relatively) easily review the change (even though it's gigantic).
- `toSymbols` issue is somehow related to this change, but is it necessary to fix in this PR?
  - if it's trivial, we can fix it immediately, but if it may take some efforts (or requires some investigations), I would recommend  you to do it in a separate follow-up PR :+1:
  - I mentioned something around `Flags.Trait`, but I didn't mean "you have to fix the problem right now, otherwise you can't merge it" (sorry if you felt pressured :bow:). You can open an issue (as you did 👍 ) and follow-up it later :)

vzmerr:
> Looks like unit tests are failing, but it's LGTM👍 thank you so much for working on this!
> 
> > Also, technically speaking, I do not understand the exact purpose of the toSymbols method there. We are feeding it java.io which is the correct package. and also StringBufferInputStream. Then the method just messes it up and erases the package information in producing the symbol. What's the purpose of that?
> > I can fix it in this PR or I can put it in a followup?
> 
> Thank you for finding the issue! You can create an issue and work on that in a follow-up 👍
> 
> For what it's worth, in my opinion, if I were you, I will make a PR as small as possible and try to do one thing in one PR. For example
> 
> * You're doing 4 things in this PR (as you explained in the PR description)
>   
>   * Instead, you might be able to create a PR for each thing.
>   * (Of course, you can do more than one thing if it's comfortable for you, or if the changes are an atomic changes).
>   * If you do multiple things in one PR
>     
>     * PR will be bigger, if PR is bigger, it takes more time to merge the PR in general 😭
>     * it makes it harder for reviews to review the diff because reviews can't tell which parts of diff are for which purpose.
>       
>       * If you want to do more than one thing, it would be good if you can explicitly separate the commits, #3790 is a good example. This PR is doing two things, but he did those things in a different commit. Therefore we could (relatively) easily review the change (even though it's gigantic).
> * `toSymbols` issue is somehow related to this change, but is it necessary to fix in this PR?
>   
>   * if it's trivial, we can fix it immediately, but if it may take some efforts (or requires some investigations), I would recommend  you to do it in a separate follow-up PR 👍
>   * I mentioned something around `Flags.Trait`, but I didn't mean "you have to fix the problem right now, otherwise you can't merge it" (sorry if you felt pressured 🙇). You can open an issue (as you did 👍 ) and follow-up it later :)

Thank you so much for taking the time to write this detailed feedback 👍 
I would just be fixing the tests and pushing.

Issue #4166, opened 2022-07-19
Remember choice of "Don't show this again" for reloading builds


asflierl:
Whenever I change something in my SBT (which I use as BSP server) build, a small notification pops up in VSCode that I have to reload my build. I usually click on "don't show this again". But as soon as I reload my build - I usually do that on the console with sbtn reload - and then reconnect to SBT again with the "connect to build server" command, VSCode forgets what I have chosen before and will show that popup again after the first build.sbt change.

@kpodsiad found in code that this setting is being reset after the "connect to build server" command
@tgodzik suggested this could be changed for when SBT is the build server

tgodzik:
Thanks for reporting! Moving it to the main repo, since I think this is a bug really.

PR #4175, opened 2022-07-20
fix: Remember choice `Don't show this again` for sbt as build server

jkciesluk:
fixes https://github.com/scalameta/metals/issues/4166

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -2348,10 +2348,12 @@ class MetalsLanguageServer(
         case None =>
           Future.successful(BuildChange.None)
       }
-      _ = {
-        tables.dismissedNotifications.ImportChanges.reset()
-        treeView.init()
-      }
+      _ = maybeSession.map(session =>
+        if (!session.main.isSbt)
+          tables.dismissedNotifications.ImportChanges.reset()
+        else {}
+      )
+      _ = treeView.init()
     } yield result)
       .recover { case NonFatal(e) =>
         disconnectOldBuildServer()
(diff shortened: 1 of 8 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -2348,10 +2348,12 @@ class MetalsLanguageServer(
...
           Future.successful(BuildChange.None)
       }
-      _ = {
-        tables.dismissedNotifications.ImportChanges.reset()
-        treeView.init()
-      }
+      _ = maybeSession.map(session =>
+        if (!session.main.isSbt)
+          tables.dismissedNotifications.ImportChanges.reset()
+        else {}
```
kpodsiad, line 2354:
```suggestion
        else ()
```

kpodsiad (approved):

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -2348,10 +2348,12 @@ class MetalsLanguageServer(
         case None =>
           Future.successful(BuildChange.None)
       }
-      _ = {
-        tables.dismissedNotifications.ImportChanges.reset()
-        treeView.init()
-      }
+      _ = maybeSession.map(session =>
+        if (!session.main.isSbt)
+          tables.dismissedNotifications.ImportChanges.reset()
```
tgodzik, line 2353: Actually this seems to have been added for https://github.com/scalameta/metals/commit/6b748f8db10e6fcdfc6c1f63b9ba190cccbb589a
and this might have been done to fix a test, so we can probably just remove it altogether and just check if the test passes still. If the test doesn't pass we might need to investigate why:
`~slow/testOnly tests.sbt.SbtServerSuite -- *reload*`

should run it.

tgodzik (changes requested):
Thanks for looking into this! I think we might need to still fix a test.

tgodzik (approved):
LGTM!

PR #4176, opened 2022-07-20
fix: return all inversed dependencies in `inverseDependenciesAll`

kpodsiad:
Previously, when project contained module which was aggregating all other modules Metals might omit some build targets when computing `inverseDependenciesAll`.
In the following example, Metals were omitting build targets `D` and `E` (maybe not always as this might be non deterministic).
(note: this graph shows dependency relation between projects)
[](https://mermaid.live/edit#pako:eNpNjkEOwiAQRa9C_rq9AAsTW3oCXRk2kzJaokCDsDBN7y6KJsxqXt5LZjbMwTAkbpHWRZyV9qLMUfT9QQwVhi9cWjNWGFtTQbUwVVA1QwfH0ZE15d72URppYccasqyG4l1D-710eTWUeDI2hQh5pceTO1BO4fTyM2SKmf-RslR-d79qfwO_1jrn)

Now, `inverseDependenciesAll` return all projects.

---

This bug was affecting all features which were using `inverseDependenciesAll`, in my case it was `find definitions` and `find all references`

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -556,8 +556,8 @@ object BuildTargets {
                 // in Bloop to compile only `B` than `A+B`.
                 leaves += head
             }
-            loop(tail)
           }
+          loop(tail)
       }
     loop(root)
     InverseDependencies(isVisited, leaves)
--- a/tests/unit/src/test/scala/tests/InverseDependenciesSuite.scala
+++ b/tests/unit/src/test/scala/tests/InverseDependenciesSuite.scala
@@ -7,25 +7,32 @@ import scala.meta.internal.metals.BuildTargets
 
 import ch.epfl.scala.bsp4j.BuildTargetIdentifier
 import munit.Location
+import munit.TestOptions
 
 class InverseDependenciesSuite extends BaseSuite {
+
   class Graph(val root: String) {
     val inverseDependencies: mutable.Map[String, ListBuffer[String]] =
-      mutable.Map.empty[String, ListBuffer[String]]
-    def dependsOn(project: String, dependency: String): this.type = {
+      mutable.Map.empty
+
+    def dependsOn(project: String, dependency: String): Graph = {
       val dependencies =
         inverseDependencies.getOrElseUpdate(dependency, ListBuffer.empty)
       dependencies += project
       this
     }
   }
+
   def root(x: String): Graph = new Graph(x)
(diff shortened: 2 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -556,8 +556,8 @@ object BuildTargets {
                 // in Bloop to compile only `B` than `A+B`.
                 leaves += head
             }
-            loop(tail)
           }
+          loop(tail)
```
kpodsiad, line 560: 4h of debugging and narrowing bug source only to change 1 line :D

tanishiking: Wow, wow, great catch!
So, previously, we stopped looping when the `head` is already visited. This is a problem.

Consider the following dependencies (`B -> A` represents "B depends on A").

```mermaid
graph TD
    A --> B
    A --> C
    B --> C
    B --> D
```

When we `inverseDependencies(A, graph)`, it **sometimes** ignored `D` (depends on the search order).

For example, it ignore `D` in the following scenario

- `loop([A]), isVisited = {}`
- `loop([C,B]), isVisited = {A}`
- `loop([B]), isVisited = {A, C}`
- `loop([C,D]), isVisited={A,C,B}` stop looping here because `C` is already visited! even though we haven't visited D yet!!!

but it visits all members in the following scenario

- `loop([A]), isVisited = {}`
- `loop([B, C]), isVisited = {A}` note that we start from B
  - `loop([C,D]), isVisited = {A, B}`
  - `loop([D]), isVisited = {A, B, C}`
  - `loop([]), isVisited = {A, B, C, D}`
- `loop([D]), isVisited = {A, B, C, D}` // we stop here but we already visited all members.

---

Now, we call `loop(tail)` **outside of if(isVisited(head)) block**, we keep searching in the first scenario 👍 👍 👍

tanishiking (approved):
LGTM!!!!! :tada::tada: Really great work!

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
kpodsiad: > So, previously, we stopped looping when the `head` is already visited. This is a problem.

Exactly, we're omitting `tail` even though it might contain unseen build targets. Thanks for putting detailed explanation @tanishiking

dos65 (approved):
:+1:

PR #4177, opened 2022-07-21
[WIP] Import missing given instance

tanishiking:
refer to https://github.com/scalameta/metals-feature-requests/issues/141

This PR enables us to import missing given instance

import-given

## TODO
- parse top-level given
- testing

## There're some limitations:

- we don't filter missing instances by type, so if there's `stringOrd: Ord[String]`, we will receive suggestions both for `intOrd` and `stringOrd`, I'm wondering can we filter it with a reasonable computation cost 🤔 
- if there's no explicit method definition under the given instances 
  - for example, we can't find `lt` and `gt` because they're not under given instances.

```scala
trait Ord[T]:
  def compare(x: T, y: T): Int
  extension (x: T) def lt (y: T): Boolean = compare(x, y) > 0
  extension (x: T) def gt (y: T): Boolean = compare(x, y) < 0

given intOrd: Ord[Int] with
  def compare(x: Int, y: Int): Int = if x < y then 1 else -1
```

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ImportMissingSymbol.scala
@@ -50,7 +50,10 @@ class ImportMissingSymbol(compilers: Compilers) extends CodeAction {
 
             val codeAction = new l.CodeAction()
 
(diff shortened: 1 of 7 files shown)


tanishiking:
I think this PR needs some rework, closing for now

PR #4179, opened 2022-07-21
bugfix: [Scala 3] Show correct param names in java methods

tgodzik:
Previously, we would show params in Java methods as $x0 etc. Now, we show the correct name based on the information cached in Metals.


mtags/src/main/scala-3/scala/meta/internal/pc/printer/MetalsPrinter.scala
```diff
@@ -157,12 +158,10 @@ class MetalsPrinter(
...
       )
 
-    lazy val defaultValues =
+    lazy val paramsDocs =
       symbolSearch.symbolDocumentation(gsym) match
         case Some(info) =>
-          (info.parameters.asScala ++ info.typeParameters.asScala)
-            .map(_.defaultValue)
-            .toSeq
+          (info.typeParameters.asScala ++ info.parameters.asScala).toSeq
```
tgodzik, line 164: The actual order was wrong :sweat_smile:

dos65 (approved):

PR #4180, opened 2022-07-21
fix: request configuration before connecting to build server

dos65:
Use `workspace/configuration` call to the client to sync userConfiguration before starting an actual server
initialization.

`didChangeConfiguration` notification call happens after `initialized`
that leads that server uses a default config during connecting to the
build server and doesn't use the actual value of `bloopVersion` setting

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -93,6 +93,7 @@ import ch.epfl.scala.bsp4j.CompileReport
 import ch.epfl.scala.{bsp4j => b}
 import com.google.gson.Gson
 import com.google.gson.JsonElement
+import com.google.gson.JsonObject
 import com.google.gson.JsonPrimitive
 import io.undertow.server.HttpServerExchange
 import org.eclipse.lsp4j.ExecuteCommandParams
@@ -167,6 +168,8 @@ class MetalsLanguageServer(
   private val recentlyOpenedFiles = new ActiveFiles(time)
   private val recentlyFocusedFiles = new ActiveFiles(time)
   private val languageClient = new DelegatingLanguageClient(NoopLanguageClient)
+
+  @volatile
   var userConfig: UserConfiguration = UserConfiguration()
   var excludedPackageHandler: ExcludedPackagesHandler =
     ExcludedPackagesHandler.default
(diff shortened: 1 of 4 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -991,17 +994,19 @@ class MetalsLanguageServer(
...
       registerNiceToHaveFilePatterns()
-      val result = Future
-        .sequence(
-          List[Future[Unit]](
-            quickConnectToBuildServer().ignoreValue,
-            slowConnectToBuildServer(forceImport = false).ignoreValue,
-            Future(workspaceSymbols.indexClasspath()),
-            Future(startHttpServer()),
-            Future(formattingProvider.load()),
+      val result = syncUserconfiguration().flatMap(_ =>
```
tanishiking, line 997: [note]
> server is not allowed to send any requests or notifications to the client until it has responded with an [InitializeResult](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeResult)
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/

but the server is allowed to send requests to a client after the server send `InitializeResult` :+1:

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -167,6 +168,8 @@ class MetalsLanguageServer(
   private val recentlyOpenedFiles = new ActiveFiles(time)
   private val recentlyFocusedFiles = new ActiveFiles(time)
   private val languageClient = new DelegatingLanguageClient(NoopLanguageClient)
+
+  @volatile
```
tanishiking, line 172: Adding `@volatile` to `var userConfig` looks good 👍 

Just out of curiosity, did you find any troubles if we don't have `@volatile`? Or you added just in case?

tanishiking (commented):
The change looks good, but not sure why tests are failing 🤔

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
dos65: Yep - just in case)

tanishiking (approved):
LGTM!

PR #4183, opened 2022-07-22
feat: Auto complete (missing) extension methods

tanishiking:
https://github.com/scalameta/metals-feature-requests/issues/282

auto-complete-extension2

<details>
<summary>outdated screen record</summary>

auto-complete-extension

</details>

### Index extension methods
Thanks to https://github.com/scalameta/metals/pull/4141, now extension methods are indexed by `MetalsSymbolSearch.`
This PR enables Metals to enrich completions by `enrichWithSymbolSearch,` which includes the extension methods found by symbol search.

Note that, we index extension methods in `val inWorkspaceMethods: TrieMap[Path, Seq[WorkspaceSymbolInformation]]` which is newly created in this PR.

see more details: https://github.com/scalameta/metals/pull/4183#issuecomment-1197614724

### enrichWithSymbolSearch
When `completion.kind == Method`, now Metals search for the extension methods from by `SymbolSearch#searchMethods`.

In `MetalsSymbolSearch`, it implements `searchMethods` to search for extension methods from `inWorkspaceMethods` by `Fuzzy.matches`.

### Filter found extension methods by qualifier's type 

We filter out the extension methods whose extension parameter type doesn't match with the qualifier on the cursor. 

For example

```scala
// A.scala
extension (num: Int)
  def incr = num + 1

// B.scala
def main = "b".inc@@ // we shouldn't suggest `incr` because the types don't match
```

## Tricky part2
~As you can see from the screen record above, Metals can't suggest the extension method `offsetToPos` until we type `off`  or `offsetTo` (it depends... on what...?)~
~This is probably because we index the workspace symbols by trigram in bloom filters, and therefore we can't match the extension method until we type at least `off` 🤔 ~

~Actually, by commenting out 
https://github.com/scalameta/metals/blob/a6299f3a2845d4a751f46d76c90c54b712d66f05/metals/src/main/scala/scala/meta/internal/metals/WorkspaceSymbolProvider.scala#L116 (which we shouldn't do this actually) we could find those symbols by typing `o` or `of`.~

~I believe we can go with the current change as it's still an improvement, but I'm wondering how we can find the extension methods by typing one or two prefixes of the methods without causing the performance deterioration (such as removing the bloom filter check), in the future.~

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
@@ -30,6 +30,7 @@ import scala.meta.internal.metals.doctor.Doctor
 import scala.meta.internal.metals.watcher.FileWatcher
 import scala.meta.internal.mtags.OnDemandSymbolIndex
 import scala.meta.internal.semanticdb.Scala._
+import scala.meta.internal.semanticdb.SymbolInformation.Kind
 import scala.meta.internal.tvp.TreeViewProvider
 import scala.meta.internal.worksheets.WorksheetProvider
 import scala.meta.io.AbsolutePath
@@ -430,11 +431,34 @@ final case class Indexer(
         val reluri = source.toIdeallyRelativeURI(sourceItem)
         val input = sourceToIndex0.toInput
         val symbols = ArrayBuffer.empty[WorkspaceSymbolInformation]
+        val methodSymbols = ArrayBuffer.empty[WorkspaceSymbolInformation]
         SemanticdbDefinition.foreach(input, dialect) {
           case SemanticdbDefinition(info, occ, owner) =>
-            if (WorkspaceSymbolProvider.isRelevantKind(info.kind)) {
+            // TODO: Do not index (extension) METHOD, they will be indexed later
+            // we index methods for auto-import missing extension methods feature for now
+            // but those feature should use methodSymbols
+            if (
+              WorkspaceSymbolProvider.isRelevantKind(
+                info.kind
+              ) || info.kind == Kind.METHOD
+            ) {
               occ.range.foreach { range =>
                 symbols += WorkspaceSymbolInformation(
+                  info.symbol,
+                  info.kind,
+                  range.toLSP,
+                )
+              }
+            }
+            // what we really want to index are "extension" methods
+            // However, we filter by `Kind.Method` because semanticdb doesn't have properties
+            // that represents "extension".
+            // Knowing `SemanticdbDefinition.foreach` uses `ScalaToplevelMtags` and it puts
+            // `Kind.Method` only to extension methods, we can safely filter extension methods
+            // by `info.kind == Kind.Method`. (TODO: add exntension properties to semanticdb schema).
+            if (info.kind == Kind.METHOD) {
+              occ.range.foreach { range =>
+                methodSymbols += WorkspaceSymbolInformation(
                   info.symbol,
                   info.kind,
                   range.toLSP,
@@ -454,7 +478,7 @@ final case class Indexer(
               )
             }
         }
-        workspaceSymbols().didChange(source, symbols.toSeq)
+        workspaceSymbols().didChange(source, symbols.toSeq, methodSymbols.toSeq)
 
         // Since the `symbols` here are toplevel symbols,
         // we cannot use `symbols` for expiring the cache for all symbols in the source.
--- a/metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
@@ -437,6 +437,7 @@ final case class Indexer(
             // TODO: Do not index (extension) METHOD, they will be indexed later
             // we index methods for auto-import missing extension methods feature for now
             // but those feature should use methodSymbols
+            // see: https://github.com/scalameta/metals/issues/4212
             if (
               WorkspaceSymbolProvider.isRelevantKind(
                 info.kind
--- a/metals/src/main/scala/scala/meta/internal/metals/WorkspaceSymbolProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/WorkspaceSymbolProvider.scala
@@ -32,6 +32,10 @@ final class WorkspaceSymbolProvider(
 ) {
   val inWorkspace: TrieMap[Path, WorkspaceSymbolsIndex] =
     TrieMap.empty[Path, WorkspaceSymbolsIndex]
+
+  // symbols for extension methods
+  val inWorkspaceMethods: TrieMap[Path, Seq[WorkspaceSymbolInformation]] =
+    TrieMap.empty[Path, Seq[WorkspaceSymbolInformation]]
   var inDependencies: ClasspathSearch =
     ClasspathSearch.empty
 
@@ -58,6 +62,16 @@ final class WorkspaceSymbolProvider(
     inDependencies.search(query, visitor)
   }
 
+  def searchMethods(
+      query: String,
+      visitor: SymbolSearchVisitor,
+      target: Option[BuildTargetIdentifier],
+  ): SymbolSearch.Result = {
+    workspaceMethodSearch(query, visitor, target)
+    if (query.isEmpty()) SymbolSearch.Result.INCOMPLETE
+    else SymbolSearch.Result.COMPLETE
+  }
+
   def indexClasspath(): Unit = {
     try {
       indexClasspathUnsafe()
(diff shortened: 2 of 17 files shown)


tanishiking:
Ah, wait, I think we should do some cosmetics on label and description for extension methods

tests/slow/src/test/scala/tests/feature/CompletionCrossLspSuite.scala
```diff
@@ -57,4 +57,52 @@ class CompletionCrossLspSuite
...
+           |import b.plus
+           |
+           |object A {
+           |  1.plus($0)
+           |}
+           |""".stripMargin,
+        filter = _.contains("plus"),
+      )
+      _ <- assertCompletion(
+        "1.plu@@",
```
tanishiking, line 96: If it's `1.pl@@`, test fails because of the reason I described in the `Tricky part2`.

dos65 (commented):
I'm wondering if it's already good to merge or it would be better to do changes in index to be able to query extension methods without a limitation on a min-3-chars-query.

It seems that extension methods should be stored in a another index. 
Smth like `Map[String, List[String]]` where the key is semanticdb name for a qual symbol and value is a list of extension method names.
Then, instead of doing a symbol search, we can query this map by all parents of current qual and filter them by `startsWith`. This also would work for autoImport.

tanishiking:
The problem is, while we index the Mtags symbols into the bloom filter like:

https://github.com/scalameta/metals/blob/81b014d4deb62f55fdef99f266e2aedf02efcc90/mtags/src/main/scala/scala/meta/internal/metals/Fuzzy.scala#L281-L287

but when the symbol is an extension method, we don't index the all prefixes of the extension methods' name somehow. I will take a look a bit more.

@dos65 Even if we can find the symbol by its prefix, do you still think we should index extension methods in a separate index?

dos65:
@tanishiking I'm bit worried about performance. 
In this approach we tests every matching symbol if it's an extension Method which adds some overhead because most them (for a small query) will be toplevel symbols

tanishiking:
> In this approach we tests every matching symbol if it's an extension Method which adds some overhead because most them (for a small query) will be toplevel symbols

Yeah, that's understandable. With this approach, when we type `foo.pl@@`, Metals will search for symbols, and it receives a lot of non-extension method symbols that are not interested in that context, then we filter out those symbols.
It would be better performant if we can filter extension methods symbols before matching irrelevant (top-level) symbols :+1:
That's why I also think it might be a good idea to have a separate index.

Let me consider is there another way to make it performant / I'll do some profiling

tanishiking:
I realized vscode doesn't request `textDocument/completion` every time we type something after `xxx.pl@`. Instead

- when we type `xxx.@@` vscode will request `textDocument/completion` to server and receive all the completion members.
- when we type something `xxx.p@@`, the client doesn't send a request to the server, the client filter out completion items on client-side.

tanishiking:
https://github.com/scalameta/metals/pull/4183/commits/d5a5c3e58bbb1373a2854c7bfba80b586c3aca3d I decided to index extension methods in a separate index `val inWorkspaceMethods: TrieMap[Path, Seq[WorkspaceSymbolInformation]]` instead of `val inWorkspace: TrieMap[Path, WorkspaceSymbolsIndex]`, and search for extension methods from `inWorkspaceMethods` for two reasons:

- as @dos65 mentioned, if we search for extension methods from `inWorkspace` (which indexes tons of top-level symbols) when we type like `xxx.f@@` Metals search for the extension methods with query `f` from `inWorkspace` index.
  - However, it's very ineffective to look up from `inWorkspace` because most of the indexed symbols in `inWorkspace` are not extension methods.
- Realized bloomFilter doesn't index each prefix for the lowercase methods, for example, `incr` will be indexed as `incr` as is and trigrams. So, if we wanna match by prefix, we can't use bloom filter (or we have to index each prefix) anyway.

tanishiking:
Oops, WorkspaceSymbolLspSuite fails

mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
```diff
@@ -159,7 +159,15 @@ class ScalaToplevelMtags(
...
           withOwner(currRegion.owner) {
-            term(name.name, name.pos, Kind.OBJECT, 0)
+            term(
+              name.name,
+              name.pos,
+              // hack: (exclusively) making symbol kind of extension methods to Kind.Method
+              // so that we can tell it's an extension method in Indexer.scala
+              // TODO: add properties that represents "extension" method to semanticdb schema
+              Kind.METHOD,
+              0
```
tanishiking, line 169: set extension methods symbol kind to `METHOD` to distinguish between extension methods and others.

It would be better if we can attach some properties that represents this is "extension" method, but there's no properties for it in SemanticDB https://scalameta.org/docs/semanticdb/specification.html#symbolinformation
Maybe we should add `extension` property to SemanticDB ?🤔  

what do you think about it @tgodzik ?

metals/src/main/scala/scala/meta/internal/metals/WorkspaceSymbolProvider.scala
```diff
@@ -58,6 +62,16 @@ final class WorkspaceSymbolProvider(
     inDependencies.search(query, visitor)
   }
 
+  def searchMethods(
+      query: String,
+      visitor: SymbolSearchVisitor,
+      target: Option[BuildTargetIdentifier],
+  ): SymbolSearch.Result = {
+    workspaceMethodSearch(query, visitor, target)
+    if (query.isEmpty()) SymbolSearch.Result.INCOMPLETE
```
tanishiking, line 71: if it's `SymbolSearch.Result.COMPLETE` when `query.isEmpty`, once the client requests completion on `foo.` with an empty query, the client won't request completions when we type `foo.x`

as LSP specification says

```
	/**
	 * This list is not complete. Further typing should result in recomputing
	 * this list.
	 *
	 * Recomputed lists have all their items replaced (not appended) in the
	 * incomplete completion sessions.
	 */
	isIncomplete: boolean;
```
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
```diff
@@ -430,9 +431,17 @@ final case class Indexer(
...
         SemanticdbDefinition.foreach(input, dialect) {
           case SemanticdbDefinition(info, occ, owner) =>
-            if (WorkspaceSymbolProvider.isRelevantKind(info.kind)) {
+            // TODO: Do not index (extension) METHOD, they will be indexed later
+            // we index methods for auto-import missing extension methods feature for now
+            // but those feature should use methodSymbols
+            if (
+              WorkspaceSymbolProvider.isRelevantKind(
+                info.kind
+              ) || info.kind == Kind.METHOD
```
tgodzik, line 443:
```suggestion
              )
```
do we want to search for methods? I don't think we normally index it.

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
```diff
@@ -441,6 +450,21 @@ final case class Indexer(
                 )
               }
             }
+            // what we really want to index are "extension" methods
+            // However, we filter by `Kind.Method` because semanticdb doesn't have properties
+            // that represents "extension".
+            // Knowing `SemanticdbDefinition.foreach` uses `ScalaToplevelMtags` and it puts
+            // `Kind.Method` only to extension methods, we can safely filter extension methods
+            // by `info.kind == Kind.Method`. (TODO: add exntension properties to semanticdb schema).
```
tgodzik, line 458: Is there an issue in scalameta already? If not could you raise it?

tgodzik (commented):
Looks pretty great! I just have some minor questions.

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
tanishiking: Nice catch! However, we have to index methods for now because

- now [ScalaToplevelMtags emit extension methods with Kind.Method](https://github.com/scalameta/metals/blob/57b05f3f8a32d7ce109a2f1011ae54f1660b3956/mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala#L158-L164)
- We have to store them in `inWorkspace` because auto-import missing extension feature still uses `inWorkspace` index.
  - we can migrate auto-import missing extension feature to use `inWorkspaceMethod` (that is defined in this PR) later, then we can remove this
  - I'll open an issue :)

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
tanishiking: Not yet, I'll open an issue and link from comment 👍

tgodzik (approved):
Thanks, otherwise everything  LGTM

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
tanishiking: https://github.com/scalameta/scalameta/issues/2799

tanishiking: oops this one https://github.com/scalameta/metals/issues/4212

metals/src/main/scala/scala/meta/internal/metals/Indexer.scala
tanishiking: https://github.com/scalameta/scalameta/issues/2799

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -298,7 +306,23 @@ class Completions(
                 ).forall(visit),
         )
         Some(search.search(query, buildTargetIdentifier, visitor))
-      case _ => None
+      case CompletionKind.Members if query.nonEmpty =>
```
dos65, line 309: Nice rework in general:+1: 

It would be ideal if it were allowed to search even with an empty query.
Then, user will be able do discover extension method by autocomplete

tanishiking: Thank you for your comment!
Right, it would be really helpful, I remember I guarded by `query.nonEmpty` because something has failed if a query is empty, but we can fix it and make them available with empty query 👍

Issue #4182, opened 2022-07-21
[Scala3] Add keyword completion for match


kpodsiad:
In Scala 3 I would like to have keyword completion for `match`:
```scala
val x: Option[Int] = ???
x ma@@
```

Scala 2 for the reference:
<img width="288" alt="Screenshot 2022-07-21 at 21 13 55" src="https://user-images.githubusercontent.com/37124721/180299495-856002d1-67de-44d4-8e47-d07b28516c4e.png">

PR #4185, opened 2022-07-22
Add match keyword completion

jkciesluk:
Solves https://github.com/scalameta/metals/issues/4182

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/KeywordsCompletions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/KeywordsCompletions.scala
@@ -23,6 +23,9 @@ object KeywordsCompletions:
           case kw if (kw.isPackage || kw.isTemplate) && notInComment =>
             CompletionValue.keyword(kw.name, kw.insertText)
         }
+      case Select(qual, name) :: (appl: Apply) :: _
+          if "match".startsWith(name.toString()) =>
+        List(CompletionValue.keyword("match", "match {\n\tcase $0\n}"))
       case _ =>
         val isExpression = this.isExpression(path)
         val isBlock = this.isBlock(path)
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/KeywordsCompletions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/KeywordsCompletions.scala
@@ -25,7 +25,7 @@ object KeywordsCompletions:
         }
       case Select(qual, name) :: (appl: Apply) :: _
           if "match".startsWith(name.toString()) =>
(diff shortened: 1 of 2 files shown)


tests/cross/src/test/scala/tests/pc/CompletionKeywordSuite.scala
```diff
@@ -186,6 +186,21 @@ class CompletionKeywordSuite extends BaseCompletionSuite {
        |""".stripMargin,
   )
 
+  check(
```
tgodzik, line 189: Could you also add a `checkEdit` test ? We could make sure how the file looks afterwards exactly.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/KeywordsCompletions.scala
```diff
@@ -23,6 +23,9 @@ object KeywordsCompletions:
           case kw if (kw.isPackage || kw.isTemplate) && notInComment =>
             CompletionValue.keyword(kw.name, kw.insertText)
         }
+      case Select(qual, name) :: (appl: Apply) :: _
+          if "match".startsWith(name.toString()) =>
+        List(CompletionValue.keyword("match", "match {\n\tcase $0\n}"))
```
tgodzik, line 28: I think we should suggest optional braces if possible, since match case is already very much in style of Scala 3
```suggestion
        List(CompletionValue.keyword("match", "match {\n\tcase $0\n}"))
```

tgodzik (changes requested):
Awesome work! I left a few thoughts about it in the comments.

tgodzik (approved):
LGTM!

tgodzik (changes requested):
It seems that `CompletionKeywordSuite` is failing now, could you look into it?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/KeywordsCompletions.scala
```diff
@@ -23,6 +22,8 @@ object KeywordsCompletions:
           case kw if (kw.isPackage || kw.isTemplate) && notInComment =>
             CompletionValue.keyword(kw.name, kw.insertText)
         }
+      case Select(qual, name) :: _ if "match".startsWith(name.toString()) =>
+        List(CompletionValue.keyword("match", "match\n\tcase $0\n"))
```
kpodsiad, line 26: Will `\t` be valid for different indentation sizes and styles (spaces vs tabs)?

tgodzik: That's actually an indication for the editor to increase the indentation. It already works this way in other places.

tgodzik (approved):
LGTM!

Issue #4186, opened 2022-07-25
Incorrect inferred type for type alias in for comprehension


kpodsiad:
### Describe the bug

```scala
type Foo = (Int, String)
def foo: Option[Foo] = ???
for {
  a <- foo
} yield a
```
<img width="245" alt="Screenshot 2022-07-25 at 08 37 46" src="https://user-images.githubusercontent.com/37124721/180713804-a5d05fec-9dea-4e48-8d41-ae989b6eb314.png">

### Expected behavior

`a` has inferred type `Foo`

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

0.11.7+59-4322516c-SNAPSHOT

kpodsiad:
Thanks @tgodzik!

PR #4188, opened 2022-07-25
bugfix: Print local type aliases properly

tgodzik:
Previously, we would not print local type names properly and instead just print "_". Now we search for the display name of that local type alias and display it instead.

Fixes https://github.com/scalameta/metals/issues/4186

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
@@ -386,7 +386,7 @@ final class SyntheticsDecorationProvider(
       textDoc.symbols
         .find(_.symbol == symbol)
         .map(_.displayName)
-        .getOrElse(symbol)
+        .getOrElse("_")
     else symbol.desc.name.value
   }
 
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/decorations/SyntheticsDecorationProvider.scala
```diff
@@ -559,6 +559,19 @@ final class SyntheticsDecorationProvider(
         case m: MethodSignature =>
           val decoration = ": " + decorationPrinter.printType(m.returnType)
           Some(decorationOptions(lspRange, decoration))
+        // local type aliases
+        case ValueSignature(tp: s.TypeRef) if tp.symbol.isLocal =>
+          textDocument.symbols.collectFirst {
```
kpodsiad, line 564: what if type alias is defined in other file? It isn't local then.

tgodzik: Then it should work normally, the problem is with printing local type aliases as far as I was able to check. Or do you have a counter example?

tgodzik: Changed the test to include a toplevel object.

kpodsiad: Right, non local type was working, my bad.

kpodsiad (approved):

dos65 (commented):
@tgodzik Looks good but I'm wondering why we have this line in printer ?
https://github.com/scalameta/metals/blob/869e08a6a5e38dc6eaf146856d3ab1f73ba283aa/metals/src/main/scala/scala/meta/internal/decorations/SemanticdbTreePrinter.scala#L35

I removed it locally and without it the new test works well without changes in SyntheticDecorationProvider

dos65 (approved):
:+1:

Issue #4190, opened 2022-07-26
Wrong code folding in Scala3


Meowcolm024:
**Describe the bug**

Code blocks do not fold well with the new syntax introduced by Scala 3

**Expected behavior**

Should fold like code block with braces

**Screenshots**

**Using indent and colon only**

<img width="386" alt="螢幕截圖 2022-07-26 上午11 36 19" src="https://user-images.githubusercontent.com/23007687/180917777-bcec9c28-b357-4bf5-abe0-00993d7ad4a8.png">
<img width="351" alt="螢幕截圖 2022-07-26 上午11 36 27" src="https://user-images.githubusercontent.com/23007687/180917780-e6dea3f0-8fb9-4ce3-9933-c99ca675da27.png">

**Using the old way**

<img width="385" alt="螢幕截圖 2022-07-26 上午11 37 01" src="https://user-images.githubusercontent.com/23007687/180917785-691ed6a1-2ce6-49a3-8db1-1c9c1388eb97.png">
<img width="384" alt="螢幕截圖 2022-07-26 上午11 37 09" src="https://user-images.githubusercontent.com/23007687/180917787-fe067010-0efb-49ce-a77d-dfacc1c59bc0.png">

**Installation:**

- Operating system: macOS
- VSCode version: 1.69.2 (Universal)
- VSCode extension version: v1.18.0
- Metals version: 0.11.7

tanishiking:
@Meowcolm024 Thank you for reporting! I transferred this issue to `scalameta/metals` as it seems a bug in the Language Server implementation, not vscode specific.

Indeed `yield()` should be folded even in Scala3 :+1:

tanishiking:
Realized the root cause is https://github.com/scalameta/metals/blob/81b014d4deb62f55fdef99f266e2aedf02efcc90/metals/src/main/scala/scala/meta/internal/parsing/FoldingRangeProvider.scala#L86-L91
we `-1` from the end line of folding range to preserve end of the block `}`.

However, in braceless syntax, we don't need this adjustment.

PR #4191, opened 2022-07-26
fix: Fold the end line of template / block if it's braceless

tanishiking:
fix: https://github.com/scalameta/metals/issues/4190

FoldRangeProider for Scala3 with braceless syntax had a problem:

If clientCapabilities set `lineFoldingOnly: true`

```ts
	/**
	 * If set, the client signals that it only supports folding complete lines.
	 * If set, client will ignore specified `startCharacter` and `endCharacter`
	 * properties in a FoldingRange.
	 */
	lineFoldingOnly?: boolean;
```

Metals minus 1 from the end line of foldRange response
https://github.com/scalameta/metals/blob/81b014d4deb62f55fdef99f266e2aedf02efcc90/metals/src/main/scala/scala/meta/internal/parsing/FoldingRangeProvider.scala#L86-L91
so that Metals can preserve the last line containing `}`.

However, in Scala3, braceless syntax doesn't have `}` in the last line for no reason.
As a result, for example, if we fold the following code,

```scala
def foo =
  ???
  ???
  ???
```

we will see

```scala
def foo = ...
  ???
```

(screen-shot from the issue)
螢幕截圖 2022-07-26 上午11 36 19
螢幕截圖 2022-07-26 上午11 36 27

---

This commit fixes this behavior by not adjusting the last line if it's braceless syntax.


metals/src/main/scala/scala/meta/internal/parsing/FoldingRangeExtractor.scala
```diff
@@ -148,21 +148,21 @@ final class FoldingRangeExtractor(
   }
 
   private object Foldable {
-    def unapply(tree: Tree): Option[Position] =
+    def unapply(tree: Tree): Option[(Position, Boolean)] =
```
tanishiking, line 151: Returns a flag whether FoldingProvider should adjust the last line or not

tests/unit/src/test/resources/foldingRange-scala3-foldLineOnly/input/Givens.scala
```diff
@@ -0,0 +1,28 @@
...
+      "o" +
+      "o"
+
+given Foo[String] with 
+  def a: String = 
+    "f" +
+      "o" +
+      "o"
+
+given x: AnyRef with
```
tgodzik, line 16: Could you add one test with end marker?

tgodzik (changes requested):

tests/unit/src/test/resources/foldingRange-scala3-foldLineOnly/input/Givens.scala
tanishiking: Added https://github.com/scalameta/metals/pull/4191/commits/81e7f54d7aa5705a66caba4a677dc6f64a41b103 :+1:

tgodzik (approved):
LGTM!

PR #68, opened 2017-11-27
Invite Shane to the team!

olafurpg:

PR #4193, opened 2022-07-26
`extends` keyword completion

vzmerr:
Support for the completion of `extends` keyword.
It resolves [#68](https://github.com/scalameta/metals-feature-requests/issues/68)

### DIFF
--- a/tests/cross/src/test/scala/tests/pc/CompletionKeywordSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/CompletionKeywordSuite.scala
@@ -486,11 +486,28 @@ class CompletionKeywordSuite extends BaseCompletionSuite {
   )
 
   check(
+    "extends",
+    """
+      |package foo
+      |
+      |class Foo ext@@{
+      |}
+    """.stripMargin,
+    """|extension
+       |extends
+       |""".stripMargin,
+    compat = Map(
+      "2" -> "extends"
+    ),
+  )
+
+  check(
     "topLevel",
     "@@",
     """|abstract class
        |case class
        |class
+       |extends
        |final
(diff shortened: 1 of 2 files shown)


tests/cross/src/test/scala/tests/pc/CompletionKeywordSuite.scala
```diff
@@ -485,12 +485,29 @@ class CompletionKeywordSuite extends BaseCompletionSuite {
...
+    ),
+  )
+
   check(
     "topLevel",
     "@@",
     """|abstract class
        |case class
        |class
+       |extends
```
tgodzik, line 510: That's the biggest issue here actually. We don't want to show extends if it's not after an identifier. `extend` cannot live alone.

vzmerr: But the `path` is `List()` :(( 
How can this be checked?

vzmerr: I can try to add a flag like `isAfterIdentifier`?

tgodzik: I think one suggestion was to use diagnostics and see if there is an error, which has the current position and we can use.

vzmerr: The other one as mentioned by you in the metals standup was using tokenisation to see where we are in the file.

tanishiking:
Closing for https://github.com/scalameta/metals/pull/4416 thank you for kicking off this feature!

PR #4199, opened 2022-07-28
fix: index inline extension methods in ScalaToplevelMtags

tanishiking:
Previously, ScalaToplevelMtags didn't parse inline extension methods

```scala
extension (num: Int) def incr = ???
```

As a result, we couldn't auto-import those extension methods that are
written in inline style.

This commit fixes to parse those extension methods, and also, added some
pc tests for auto-import extension methods.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
+++ b/mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
@@ -134,15 +134,22 @@ class ScalaToplevelMtags(
           val nextOwner =
             if (
               dialect.allowToplevelStatements &&
-              needEmitFileOwner(currRegion)
+              currRegion.produceSourceToplevel
             ) {
-              sourceTopLevelAdded = true
-              val pos = newPosition
               val srcName = input.filename.stripSuffix(".scala")
               val name = s"$srcName$$package"
-              withOwner(currRegion.owner) {
-                term(name, pos, Kind.OBJECT, 0)
+              val owner = withOwner(currRegion.owner) {
+                symbol(Descriptor.Term(name))
               }
+
+              if (needEmitFileOwner(currRegion)) {
+                sourceTopLevelAdded = true
+                val pos = newPosition
+                withOwner(currRegion.owner) {
+                  term(name, pos, Kind.OBJECT, 0)
+                }
+              }
+              owner
             } else currentOwner
           scanner.nextToken()
           loop(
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
```diff
@@ -155,13 +162,29 @@ class ScalaToplevelMtags(
...
           acceptTrivia()
           val name = newIdentifier
           withOwner(currRegion.owner) {
             term(name.name, name.pos, Kind.OBJECT, 0)
           }
           loop(indent, isAfterNewline = false, region, newExpectIgnoreBody)
+        // inline extension method `extension (...) def foo = ...`
+        case DEF if expectTemplate.map(needToParseExtension).getOrElse(false) =>
+          expectTemplate match {
+            case None => fail("never happens")
```
tgodzik, line 177: LEt's add a better message if that happens due to some changes in the future?

mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
```diff
@@ -155,13 +162,29 @@ class ScalaToplevelMtags(
...
+            case Some(expect) =>
+              val next =
+                expect.startIndentedRegion(currRegion, expect.isExtension)
```
tgodzik, line 180: Why do we need to start indented region if the PR is about  inline extensions?

tgodzik (changes requested):
Thanks for noticing this one!

mtags/src/main/scala/scala/meta/internal/mtags/ScalaToplevelMtags.scala
tanishiking: Right, we didn't need it :+1: https://github.com/scalameta/metals/pull/4199/commits/5f699f0b1e5c65cf998a9f5fb8bde9c0237d89ba

tests/cross/src/test/scala/tests/pc/AutoImportExtensionMethodsSuite.scala
```diff
@@ -0,0 +1,34 @@
+package tests.pc
+
+import tests.BaseAutoImportsSuite
+
+class AutoImportExtensionMethodsSuite extends BaseAutoImportsSuite {
+
+  override def ignoreScalaVersion = Some(IgnoreScala2)
```
tgodzik, line 7:
```suggestion
  override def ignoreScalaVersion: Option[IgnoreScalaVersion] = Some(IgnoreScala2)
```

tgodzik (changes requested):
Looks good, the only thing left is running `sbt scalafixAll`

tanishiking:
Oops, thanks! applied scalafix.

tgodzik (approved):
LGTM!

kpodsiad:
@tanishiking I saw PR's title and I'm confused, what is inline extension method and how it differs from "just" extension method? Is this just a naming convention or am I missing sth?
```scala
  extension (i: Int)
    inline def inlineDouble: Int = i * 2
    def double: Int = i * 2
```

tanishiking:
I mention

```scala
extension (num: Int) def incr = ???
```

as "inline extension" (sorry for confusion, I'm not sure how should I call it), see the PR description.

kpodsiad:
Ok, I didn't that placing `def` on the same line or line below matters ;)

Are completions for "extension group" supported? Just trying to get on the same page ;)
```scala
extension (num: int)
  def incr = ???
```

Issue #4196, opened 2022-07-27
Interrupting presentation compiler thread closes jar filesystem


tgodzik:
### Describe the bug

I managed to reproduce it often by trying to get completions for `for@@` in weird places in ScalaPrecentationCompiler for Scala 3
```
2022.07.27 19:17:07 ERROR searching for `scala/Equals#canEqual().` failedjava.nio.channels.ClosedByInterruptException
	at java.nio.channels.spi.AbstractInterruptibleChannel.end(AbstractInterruptibleChannel.java:199)
	at sun.nio.ch.FileChannelImpl.endBlocking(FileChannelImpl.java:171)
	at sun.nio.ch.FileChannelImpl.readInternal(FileChannelImpl.java:844)
	at sun.nio.ch.FileChannelImpl.read(FileChannelImpl.java:824)
	at jdk.nio.zipfs.ZipFileSystem.readFullyAt(ZipFileSystem.java:1241)
	at jdk.nio.zipfs.ZipFileSystem.readFullyAt(ZipFileSystem.java:1236)
	at jdk.nio.zipfs.ZipFileSystem$EntryInputStream.initDataPos(ZipFileSystem.java:2417)
	at jdk.nio.zipfs.ZipFileSystem$EntryInputStream.read(ZipFileSystem.java:2359)
	at jdk.nio.zipfs.ZipFileSystem$2.fill(ZipFileSystem.java:2309)
	at java.util.zip.InflaterInputStream.read(InflaterInputStream.java:158)
	at java.io.InputStream.readNBytes(InputStream.java:409)
	at java.io.InputStream.readAllBytes(InputStream.java:346)
	at jdk.nio.zipfs.ZipFileSystem.newByteChannel(ZipFileSystem.java:977)
	at jdk.nio.zipfs.ZipPath.newByteChannel(ZipPath.java:864)
	at jdk.nio.zipfs.ZipFileSystemProvider.newByteChannel(ZipFileSystemProvider.java:238)
	at java.nio.file.Files.newByteChannel(Files.java:380)
	at java.nio.file.Files.newByteChannel(Files.java:432)
	at java.nio.file.Files.readAllBytes(Files.java:3288)
	at scala.meta.internal.io.PlatformFileIO$.slurp(PlatformFileIO.scala:45)
	at scala.meta.internal.io.FileIO$.slurp(FileIO.scala:24)
[...]

PR #4200, opened 2022-07-28
bugfix: Don't use interrupt for the Scala 3 compiler

tgodzik:
Scala 3 compiler doesn't operate on a separate thread the same way as Scala 2 does, so interrupt will cause it to close all the channels while not actually stopping any computation by the compiler.

Fixes https://github.com/scalameta/metals/issues/4196

I tried reopening, but that's not really possible, so best is to not interrupt the thread.

### DIFF
--- a/mtags/src/main/scala/scala/meta/internal/pc/CompilerAccess.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/CompilerAccess.scala
@@ -99,12 +99,11 @@ abstract class CompilerAccess[Reporter, Compiler](
             isFinished.compareAndSet(false, true) &&
             isDefined
           ) {
-            _compiler.presentationCompilerThread.foreach(_.interrupt())
-            if (
-              _compiler.presentationCompilerThread.isEmpty || !_compiler.presentationCompilerThread
-                .contains(thread)
-            ) {
-              thread.interrupt()
+            _compiler.presentationCompilerThread match {
+              case None => // don't interrupt if we don't have separate thread
+              case Some(pcThread) =>
+                pcThread.interrupt()
+                if (thread != pcThread) thread.interrupt()
             }
           }
         }


tanishiking (commented):
The change seems reasonable, but let me clarify the problem as I'm not familiar with this area.

- While Scala2 presentation compiler runs on [MetalsGlobalThread](https://github.com/scalameta/metals/blob/cab71832a091dbd7295495fe111aa16ae09cf7a4/mtags/src/main/scala-2/scala/tools/nsc/interactive/GlobalProxy.scala#L54), Scala3 presentation compiler runs on just a Future, so we can't tell on which thread the presentation compiler is running. (and that's why `thread.interrupt` doesn't necessarily interrupt the thread the compiler running).
- The problem is, `compiler.presentationCompilerThread` in Scala3 is always `None` as we can see [here](https://github.com/scalameta/metals/blob/cab71832a091dbd7295495fe111aa16ae09cf7a4/mtags/src/main/scala-3/scala/meta/internal/pc/Scala3CompilerWrapper.scala#L29), as a result, we always run `thread.interrupt()` on `$/cancelRequest` which interrupt the thread that compiler job queue is running. (and doing other things?)
- Now in Scala3, basically `withInterruptableCompiler` doesn't do anything on `$/cancelRequest`.
  - Where `$/cancelRequest` notification will be sent from client if we run `completions` or something consecutively I didn't know VSCode https://github.com/microsoft/vscode-languageserver-node/issues/522 (I didn't know that)

---

I'm wondering how `CompilerAccess` in Scala2 works. `onCancel`,

- `_compiler.presentationCompilerThread.foreach(...)` will interrupt the presentation compiler thread.
- `_compiler.presentationCompilerThread.isEmpty` is always `false` and
- `!_compiler.presentationCompilerThread.contains(thread)` is always `true`, am I right?
  - because `presentationCompilerThread` is a newly spawned thread, while `thread` is `Thread.currentThread` which is the thread the job queue is running?
- in that case, we always run `thread.interrupt()` and we should encounter the same situation in Scala2
  - but it doesn't, am I missing something?

tgodzik:
> The change seems reasonable, but let me clarify the problem as I'm not familiar with this area.
> 
> * While Scala2 presentation compiler runs on [MetalsGlobalThread](https://github.com/scalameta/metals/blob/cab71832a091dbd7295495fe111aa16ae09cf7a4/mtags/src/main/scala-2/scala/tools/nsc/interactive/GlobalProxy.scala#L54), Scala3 presentation compiler runs on just a Future, so we can't tell on which thread the presentation compiler is running. (and that's why `thread.interrupt` doesn't necessarily interrupt the thread the compiler running).

The Scala 2 compiler actually checks for interrupts in some places, so it's able to cancel some computation.

> * The problem is, `compiler.presentationCompilerThread` in Scala3 is always `None` as we can see [here](https://github.com/scalameta/metals/blob/cab71832a091dbd7295495fe111aa16ae09cf7a4/mtags/src/main/scala-3/scala/meta/internal/pc/Scala3CompilerWrapper.scala#L29), as a result, we always run `thread.interrupt()` on `$/cancelRequest` which interrupt the thread that compiler job queue is running. (and doing other things?)
> * Now in Scala3, basically `withInterruptableCompiler` doesn't do anything on `$/cancelRequest`.

Yeah, unfortunately it doesn't. I haven't figured out how to do it currently. Scala3CompilerWrapper was an effort to make Scala 2 and 3 share more code, but that might be doomed to fail at least here.

>   * Where `$/cancelRequest` notification will be sent from client if we run `completions` or something consecutively I didn't know VSCode [microsoft/vscode-languageserver-node#522](https://github.com/microsoft/vscode-languageserver-node/issues/522) (I didn't know that)
> 
> I'm wondering how `CompilerAccess` in Scala2 works. `onCancel`,

The compiler should check for interrupted and stop computation then.

> * `_compiler.presentationCompilerThread.foreach(...)` will interrupt the presentation compiler thread.
> * `_compiler.presentationCompilerThread.isEmpty` is always `false` and
> * `!_compiler.presentationCompilerThread.contains(thread)` is always `true`, am I right?
>   

I think so, might have been different at some point.

>   * because `presentationCompilerThread` is a newly spawned thread, while `thread` is `Thread.currentThread` which is the thread the job queue is running?
> * in that case, we always run `thread.interrupt()` and we should encounter the same situation in Scala2
>   
>   * but it doesn't, am I missing something?

presentationCompilerThread is interrupted instead of currentThread I think, which is why this doesn''t happen I think.

tanishiking (approved):
Thank you!
I still have to figure out how it works, but change itself good to me :+1:

dos65 (commented):
It seems that in case of scala3 PC we can interrupt `queueThread` here.
Have you tried that?

Also, it's suspicious that `tests.pc.CancelCompletionSuite` passes

tgodzik:
> It seems that in case of scala3 PC we can interrupt `queueThread` here. Have you tried that?

I think that is exactly what we are doing actually and that breaks things. https://github.com/scalameta/metals/pull/4200/files#diff-fc8e04e48f9ccd4f2d013b9f320038e3f12c6c993cfe3c0d8aedc240a18f0098R96

> Also, it's suspicious that `tests.pc.CancelCompletionSuite` passes

It only checks if we cancel and it does that before running anything, so that works.

dos65 (approved):
> I think that is exactly what we are doing actually and that breaks things.

@tgodzik  Oh - yep you right. I'm starting getting this a bit.
So, in case of scala2 we don't interrupt the thread which serves PC request and we shouldn't do that for scala3.

Then, I got interested `what exactly we cancel/interrupt for scala2`. 
I'm not sure but it looks like we do not perform anything on compler thread or scheduler.
If we look at https://github.com/scala/scala/blob/cbc012e573346dc685c478eec5fcbb56d22ea884/src/interactive/scala/tools/nsc/interactive/CompilerControl.scala#L37-L39 there is a limited set of functions that are invoked on this thread and I don't think that we use them.

tgodzik:
Isn't it only the typer stuff that we can cancel this way then? So nothing on the Metals side would get cancelled and that's ok since there is nothing particularly computation extensive. The most expensive is typechecking.

Issue #4192, opened 2022-07-26
Wrong hover info for structural members in scala 3


prolativ:
### Describe the bug

The expression type shown when hovering on a structural member selection in scala 3 is displayed incorrectly as the declared return type of `selectDynamic` or `applyDynamic`, e.g. for the following snippet
```scala
trait Sel extends Selectable:
  def foo1: Int = ???
  def bar1(x: Int): Int = ???
  def selectDynamic(name: String): Any = ???
  def applyDynamic(name: String)(args: Any*): Any = ???
val sel = (new Sel {}).asInstanceOf[Sel { def foo2: Int; def bar2(x: Int): Int }]
val foo1 = sel.foo1
val bar1 = sel.bar1
val foo2 = sel.foo2
val bar2 = sel.bar2
```
- hover on `foo2` and `bar2` after a dot (lines 9 & 10) without a selection doesn't show hover popups at all
- hover on `.foo2` (when at least one character is selected) shows expression type `Any` and symbol signature `def selectDynamic(name: String): Any`
- hover on `.bar2` (when at least one character is selected) shows expression type `Any` and symbol signature `def applyDynamic(name: String)(args: Any*): Any`

### Expected behavior

The behaviour of hover for structural members should be the same as for nonstructural members, i.e. in the example above hover popups on `.foo2` and `.bar2` should match (modulo def names) the popups on `.foo1` and `.bar1` respectively ():

- `.foo2`: expression type `Int` and symbol signature `def foo2: Int`
- `.bar2`: expression type `Int` and symbol signature `def bar2(x: Int): Int`

and selecting a part of the identifier should not be necessary to make a popup appear.

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.7+65-81b014d4-SNAPSHOT

PR #4202, opened 2022-07-28
bugfix: Print correct method signature for Selectable

tgodzik:
Previously, we would not show anything when selectDynamic or applyDynamic methods were used. Now, we look up the refinement to get the proper signatures.

Fixes https://github.com/scalameta/metals/issues/4192

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
@@ -12,11 +12,13 @@ import scala.meta.pc.SymbolDocumentation
 import scala.meta.pc.SymbolSearch
 
 import dotty.tools.dotc.ast.tpd.*
+import dotty.tools.dotc.core.Constants.*
 import dotty.tools.dotc.core.Contexts.*
 import dotty.tools.dotc.core.Flags.*
 import dotty.tools.dotc.core.NameKinds.*
 import dotty.tools.dotc.core.NameOps.*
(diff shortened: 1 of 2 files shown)


prolativ (approved):

tanishiking (approved):
LGTM 🥳

mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
```diff
@@ -129,6 +132,35 @@ object HoverProvider:
...
     private def isPoint: Boolean = pos.start == pos.end
 
+  private def fallbackToDynamics(
+      path: List[Tree],
+      printer: MetalsPrinter,
+  )(using Context): ju.Optional[Hover] = path match
+    case (head @ Apply(
+          Select(sel, n),
+          List(Literal(Constant(name: String))),
+        )) :: _ if n == nme.selectDynamic || n == nme.applyDynamic =>
```
dos65, line 142:
```suggestion
    case Apply(
          Select(sel, n),
          List(Literal(Constant(name: String))),
        ) :: _ if n == nme.selectDynamic || n == nme.applyDynamic =>
```

dos65 (approved):
LGTM!

mtags/src/main/scala-3/scala/meta/internal/pc/HoverProvider.scala
tgodzik: Fixed! Good catch!

PR #4203, opened 2022-07-29
Update debug adapter to 2.2.0 stable (dependency update)

arixmkii:
Bump dependency to latest stable (to not always override it in SBT dependencies).

Also, I see that metals is missing from the list of projects covered by new public instance of Scala Steward: https://github.com/VirtusLab/scala-steward-repos/blob/main/repos-github.md Should I create a PR there for it to be added. I can see steward config in the repo, and I guess it was using old instance (hosted by the creator of scala steward), which is now decommissioned.

### DIFF
--- a/project/V.scala
+++ b/project/V.scala
@@ -19,7 +19,7 @@ object V {
   val bsp = "2.1.0-M1"
   val coursier = "2.1.0-M6"
   val coursierInterfaces = "1.0.8"
-  val debugAdapter = "2.2.0-M1"
+  val debugAdapter = "2.2.0"
   val genyVersion = "0.7.1"
   val gradleBloop = bloop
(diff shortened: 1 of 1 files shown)


kpodsiad:
> Also, I see that metals is missing from the list of projects covered by new public instance of Scala Steward:

Scalameta has its own scala steward instance running - see https://github.com/VirtusLab/scala-steward-repos/pull/25

arixmkii:
> Scalameta has its own scala steward instance 

Then everything is cool. And it just has some slower schedule bumping versions. Noticed this specific version trying to debug debug adapter 😅 and though it might be somehow missed by steward. Feel free to close the MR if irrelevant for the project right now.

tgodzik (approved):
LGTM! Thanks!

Issue #4217, opened 2022-07-29
Adjust type if mismatch


BlackAnubis7:
### Is your feature request related to a problem? Please describe.

When I define a value/variable of a wrong type (i.e. `val a: Int = "abc"`), I have no way to automatically fix it

### Describe the solution you'd like

A context action which will change the variable type. Example:
`val a: Int = Some("abc")` -> `val a: Option[String] = Some("abc")`

### Describe alternatives you've considered

Manually changing the type. In extreme cases, these can be very long:
`List[(Int, Int, Map[String, List[Either[Option[Int], (Int, Boolean)]]])]`

### Search terms

type, mismatch, context

PR #4218, opened 2022-08-02
feature: Adjust type if mismatch

tgodzik:
Previously, if the type was wrong user would need to remove the type and invoke insert type code action again or just do it all manuall. Now, we suggest fixing the type whenever a type mismatch error is encountered.

Fixes https://github.com/scalameta/metals/issues/4217

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala
@@ -78,7 +78,7 @@ class InsertInferredType(trees: Trees) extends CodeAction {
       title <- inferTypeTitle(name)
     } yield insertInferTypeAction(title)
 
-    def adjustTypeTree(tree: Defn): Option[l.Range] = tree match {
+    def typedDefnTreePos(tree: Defn): Option[l.Range] = tree match {
       case Defn.Def(_, name, _, _, tpe, _) if tpe.isDefined =>
         Some(name.pos.toLSP)
       case Defn.GivenAlias(_, name, _, _, _, _) =>
@@ -110,7 +110,7 @@ class InsertInferredType(trees: Trees) extends CodeAction {
       (mismatchedType, diag) <- typeMismatch
       defn <-
         trees.findLastEnclosingAt[Defn](path, diag.getRange().getStart())
-      pos <- adjustTypeTree(defn)
+      pos <- typedDefnTreePos(defn)
     } yield adjustTypeAction(mismatchedType, pos)
 
     List(actions, adjustType).flatten
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala
@@ -9,6 +9,7 @@ import scala.meta.Pat
 import scala.meta.Term
 import scala.meta.internal.metals.CodeAction
 import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals.ScalacDiagnostic
 import scala.meta.internal.metals.ServerCommands
 import scala.meta.internal.parsing.Trees
 import scala.meta.pc.CancelToken
@@ -24,6 +25,16 @@ class InsertInferredType(trees: Trees) extends CodeAction {
       params: l.CodeActionParams,
       token: CancelToken,
   )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = Future {
+
+    def typeMismatch: Option[(String, l.Diagnostic)] = {
+      params
+        .getContext()
+        .getDiagnostics()
+        .asScala
+        .collectFirst { case ScalacDiagnostic.TypeMismatch(typ, diag) =>
+          (typ, diag)
+        }
+    }
 
     def insertInferTypeAction(title: String): l.CodeAction = {
       val codeAction = new l.CodeAction()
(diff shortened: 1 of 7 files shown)


metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala
```diff
@@ -67,11 +78,47 @@ class InsertInferredType(trees: Trees) extends CodeAction {
       title <- inferTypeTitle(name)
     } yield insertInferTypeAction(title)
 
-    actions.toList
+    def adjustTypeTree(tree: Defn): Option[l.Range] = tree match {
```
tanishiking, line 81: [nit]
```suggestion
    def defnTreePos(tree: Defn): Option[l.Range] = tree match {
```

or something? `adjustTypeTree` sounds like this method adjust the type tree.

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -67,16 +72,30 @@ final class InferredTypeProvider(
...
       /* `val a = 1` or `var b = 2`
        *     turns into
        * `val a: Int = 1` or `var b: Int = 2`
        */
-      case vl @ ValDef(_, name, tpt, _) if !vl.symbol.isParameter =>
+      case vl @ ValDef(_, name, tpt, rhs) if !vl.symbol.isParameter =>
         val nameEnd = findNameEnd(tpt.pos.start, name)
         val nameEndPos = tpt.pos.withEnd(nameEnd).withStart(nameEnd).toLSP
+        adjustedEndPos.foreach(nameEndPos.setEnd(_))
         val typeNameEdit = new TextEdit(nameEndPos, ": " + prettyType(tpt.tpe))
```
tanishiking, line 90:
```suggestion
        lazy val typeNameEdit = new TextEdit(nameEndPos, ": " + prettyType(tpt.tpe))
```

or calculate this in `else` branch

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
```diff
@@ -28,10 +29,14 @@ final class InferredTypeProvider(
 ) {
   import compiler._
 
-  def inferredTypeEdits(): List[TextEdit] = {
+  def inferredTypeEdits(
+      text: Option[String] = None,
+      adjustedEndPos: Option[l.Position] = None,
+      retryType: Boolean = true
```
tanishiking, line 35: Can we group these parameters in a case class like

```scala
case class AdjustTypeOpts(
  text: String,
  adjustedEndPos: l.Position
)

def inferredTypeEdits(
  adjustOpt: Option[AdjustTypeOpts] = None
) = {
  val retryType = adjustOpt.isEmpty
}
```

?

Those parameters are always

- `None, None, true` for the first try, or
- `Some(...), Some(...), false` when it's retry

if we group them up, we can easily know the parameters like `Some(...), None, true` won't be happen.

tanishiking: BTW, removing the `: Type` part and retrying to insert inferred type is really good idea! Great job :+1:

tests/cross/src/test/scala/tests/pc/InsertInferredTypeSuite.scala
```diff
@@ -31,6 +31,77 @@ class InsertInferredTypeSuite extends BaseCodeActionSuite {
        |}""".stripMargin,
   )
 
+  checkEdit(
+    "wrong-val",
+    """|object A{
+       |  val <<alpha>> :  String = 123
```
tanishiking, line 37:
```suggestion
       |  val <<alpha>>: String = 123
```
I'm curious does this work?

tanishiking (commented):
I left some comments, but it's almost LGTM :rocket:

metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala
```diff
@@ -67,11 +78,47 @@ class InsertInferredType(trees: Trees) extends CodeAction {
...
+      case Defn.Def(_, name, _, _, tpe, _) if tpe.isDefined =>
+        Some(name.pos.toLSP)
```
dos65, line 83: If there are parameters it should point ad the last `)`. 
Otherwise the action removes parameters,

tgodzik: Not sure what you mean here, do you maybe have a code example that this breaks on?

dos65: this one fails if you add it
```scala
  checkEdit(
    "wrong-def-params".only,
    """|object A{
       |  def <<alpha>>(a: Int, b: String): String = 123
       |}""".stripMargin,
    """|object A{
       |  def alpha(a: Int, b: String): Int = 123
       |}""".stripMargin,
  )
```

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
tgodzik: > BTW, removing the `: Type` part and retrying to insert inferred type is really good idea! Great job +1

I actually tried getting it from RHS, but the type was just error.

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
tgodzik: Fixed!

metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala
tgodzik: Renamed!

mtags/src/main/scala-2/scala/meta/internal/pc/InferredTypeProvider.scala
tgodzik: I applied the refactoring, thanks for the suggestion!

tests/cross/src/test/scala/tests/pc/InsertInferredTypeSuite.scala
tgodzik: It does, actually I also tested that in the LSP tests. I also changed the example here.

metals/src/main/scala/scala/meta/internal/metals/codeactions/InsertInferredType.scala
tgodzik: Thanks! I totally forgot about this edge case. It's fixed now.

dos65 (approved):
LGTM!

tgodzik:
Let's merge it after release?

PR #4219, opened 2022-08-02
chore: Switch to JDK 17 for most tests

tgodzik:
It should be enough to mostly test on JDK 17 with a couple tests also running on JDK 8.


ckipp01 (approved):
👍🏼 LGTM

dos65 (approved):
:+1:

PR #4220, opened 2022-08-03
adding AmmoniteFileCompletions for scala 3

### LINKED ISSUE #4201: [Scala 3] File completions for Ammonite
Ammonite allows to import symbols from other file using `import $file` syntax, for example:
```scala
import $file.dir.otherFile
```

This is already supported for Scala 2 in https://github.com/scalameta/metals/blob/main/mtags/src/main/scala-2/scala/meta/internal/pc/completions/AmmoniteCompletions.scala

And tests are available in:
https://github.com/scalameta/metals/blob/main/tests/unit/src/main/scala/tests/BaseAmmoniteSuite.scala#L224

and are currently ignored for Scala 3.

We should extract some of the logic to a separate trait from AmmoniteCompletions and reused the same logic for Scala 3.

vzmerr:
It resolves #4201 

- It takes the workspace path of the `ScalaPresentationCompiler` and provides it to` CompletionsProvider` and ultimately the `AmmoniteFileCompletions`

- It utilises the same logic and code as in Scala 2 AmmoniteCompletions with the goal of making the tests for `Ammonite3Suite` pass

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/AmmoniteFileCompletions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/AmmoniteFileCompletions.scala
@@ -13,6 +13,8 @@ import scala.meta.io.AbsolutePath
 import dotty.tools.dotc.ast.tpd.Tree
 import dotty.tools.dotc.ast.tpd.*
 import dotty.tools.dotc.ast.untpd.ImportSelector
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.StdNames.*
 import org.eclipse.{lsp4j as l}
 
 object AmmoniteFileCompletions:
@@ -22,7 +24,7 @@ object AmmoniteFileCompletions:
       case Select(qual, name) =>
         val pathPart = name.toString()
         translateImportToPath(qual) + "/" + {
-          if pathPart == "$up" then ".."
+          if pathPart == "^" then ".."
           else pathPart
         }
       case Ident(_) =>
@@ -32,11 +34,11 @@ object AmmoniteFileCompletions:
   def contribute(
       select: Tree,
       selector: List[ImportSelector],
-      editRange: l.Range,
+      posRange: l.Range,
       rawPath: String,
       workspace: Option[Path],
       rawFileName: String,
-  ): List[CompletionValue] =
+  )(using Context): List[CompletionValue] =
 
     val fileName = rawFileName
       .split("/")
@@ -47,18 +49,22 @@ object AmmoniteFileCompletions:
       .split("\\$file")
       .toList
 
+    val editRange = selector.headOption.map { sel =>
+      if sel.sourcePos.span.isZeroExtent then posRange
+      else sel.imported.sourcePos.toLSP
+    }
     val query = selector.collectFirst { case sel: ImportSelector =>
-      sel.name.toString.replace(Cursor.value, "")
-
+      if sel.name.isEmpty || sel.name == nme.ERROR then ""
+      else sel.name.toString.replace(Cursor.value, "")
     }
 
     def parent =
       val name = "^"
 
       CompletionValue.FileSystemMember(
-        isDirectory = true,
         name,
         editRange,
+        isDirectory = true,
       )
 
     (split, workspace) match
@@ -70,24 +76,20 @@ object AmmoniteFileCompletions:
           current.getParent.resolve(importPath)
         )
         val parentTextEdit =
-          if query
-              .exists(
-                _.isEmpty()
-              ) && currentPath.parentOpt.isDefined && currentPath.isDirectory
+          if query.exists(_.isEmpty()) &&
+            currentPath.parentOpt.isDefined && currentPath.isDirectory
           then List(parent)
           else Nil
         currentPath.list.toList
           .filter(_.filename.stripSuffix(".sc") != fileName)
           .collect {
             case file
-                if (file.isDirectory || file.isAmmoniteScript) && query
-                  .exists(
-                    CompletionFuzzy.matches(_, file.filename)
-                  ) =>
+                if (file.isDirectory || file.isAmmoniteScript) &&
+                  query.exists(CompletionFuzzy.matches(_, file.filename)) =>
               CompletionValue.FileSystemMember(
-                isDirectory = file.isDirectory,
                 file.filename,
                 editRange,
+                isDirectory = file.isDirectory,
               )
           } ++ parentTextEdit
       case _ =>
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/AmmoniteFileCompletions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/AmmoniteFileCompletions.scala
@@ -1,0 +1,97 @@
+package scala.meta.internal.pc
+package completions
+
+import java.nio.file.Path
+import java.{util as ju}
+
+import scala.collection.JavaConverters.*
+
+import scala.meta.internal.mtags.MtagsEnrichments.*
+import scala.meta.internal.pc.AutoImports.AutoImport
+import scala.meta.io.AbsolutePath
+
+import dotty.tools.dotc.ast.tpd.Tree
+import dotty.tools.dotc.ast.tpd.*
+import dotty.tools.dotc.ast.untpd.ImportSelector
+import org.eclipse.{lsp4j as l}
+
+object AmmoniteFileCompletions:
+
+  private def translateImportToPath(tree: Tree): String =
+    tree match
+      case Select(qual, name) =>
+        val pathPart = name.toString()
+        translateImportToPath(qual) + "/" + {
+          if pathPart == "$up" then ".."
+          else pathPart
+        }
+      case Ident(_) =>
+        ""
+      case _ => ""
+
+  def contribute(
+      select: Tree,
+      selector: List[ImportSelector],
+      editRange: l.Range,
+      rawPath: String,
+      workspace: Option[Path],
+      rawFileName: String,
+  ): List[CompletionValue] =
+
+    val fileName = rawFileName
+      .split("/")
+      .last
+      .stripSuffix(".amm.sc.scala")
+
+    val split = rawPath
+      .split("\\$file")
+      .toList
+
+    val query = selector.collectFirst { case sel: ImportSelector =>
+      sel.name.toString.replace(Cursor.value, "")
+
+    }
+
+    def parent =
+      val name = "^"
+
+      CompletionValue.FileSystemMember(
+        isDirectory = true,
+        name,
+        editRange,
+      )
+
+    (split, workspace) match
+      case (_ :: script :: Nil, Some(workspace)) =>
+        // drop / or \
+        val current = workspace.resolve(script.drop(1))
+        val importPath = translateImportToPath(select).drop(1)
+        val currentPath = AbsolutePath(
+          current.getParent.resolve(importPath)
+        )
+        val parentTextEdit =
+          if query
+              .exists(
+                _.isEmpty()
+              ) && currentPath.parentOpt.isDefined && currentPath.isDirectory
+          then List(parent)
+          else Nil
+        currentPath.list.toList
+          .filter(_.filename.stripSuffix(".sc") != fileName)
+          .collect {
+            case file
+                if (file.isDirectory || file.isAmmoniteScript) && query
+                  .exists(
+                    CompletionFuzzy.matches(_, file.filename)
+                  ) =>
+              CompletionValue.FileSystemMember(
+                isDirectory = file.isDirectory,
+                file.filename,
+                editRange,
+              )
+          } ++ parentTextEdit
+      case _ =>
+        Nil
+    end match
+  end contribute
+end AmmoniteFileCompletions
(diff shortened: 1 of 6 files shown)


tests/unit/src/main/scala/tests/BaseAmmoniteSuite.scala
```diff
@@ -220,97 +220,95 @@ abstract class BaseAmmoniteSuite(scalaVersion: String)
     } yield ()
   }
 
-  if (scalaVersion != V.ammonite3)
```
vzmerr, line 223: Note for the reviewers: the body of the test is not changed at all. I only removed the condition for excluding Scala 3

tests/unit/src/main/scala/tests/BaseAmmoniteSuite.scala
```diff
@@ -220,97 +220,95 @@ abstract class BaseAmmoniteSuite(scalaVersion: String)
...
-                                    |others""".stripMargin
-        completionList <- server.completion(
-          "b/otherMain.sc",
-          "import $file.other@@",
-        )
-        _ = assertNoDiff(completionList, expectedCompletionList)
-
-      } yield ()
-    }
+  test("file-completion") {
```
vzmerr, line 223: Note for the reviewers: the body of the test is not changed at all. I only removed the condition for excluding Scala 3

tests/unit/src/main/scala/tests/BaseAmmoniteSuite.scala
```diff
@@ -220,97 +220,95 @@ abstract class BaseAmmoniteSuite(scalaVersion: String)
...
+      expectedCompletionList = """|other.sc
+                                  |otherScript.sc
+                                  |others""".stripMargin
+      completionList <- server.completion(
+        "b/otherMain.sc",
+        "import $file.other@@",
+      )
+      _ = assertNoDiff(completionList, expectedCompletionList)
 
-  if (scalaVersion != V.ammonite3)
```
vzmerr, line 279: Note for the reviewers: the body of the test is not changed at all. I only removed the condition for excluding Scala 3

tests/unit/src/main/scala/tests/BaseAmmoniteSuite.scala
```diff
@@ -220,97 +220,95 @@ abstract class BaseAmmoniteSuite(scalaVersion: String)
...
-          "import $file.foos.Script@@",
-        )
-        _ = assertNoDiff(completionList, expectedCompletionList)
-
-      } yield ()
-    }
+    } yield ()
+  }
+
+  test("file-completion-path") {
```
vzmerr, line 278: Note for the reviewers: the body of the test is not changed at all. I only removed the condition for excluding Scala 3

mtags/src/main/scala-3/scala/meta/internal/pc/completions/AmmoniteFileCompletions.scala
```diff
@@ -0,0 +1,127 @@
...
+import dotty.tools.dotc.core.Types.Type
+import dotty.tools.dotc.interactive.Interactive
+import dotty.tools.dotc.interactive.InteractiveDriver
+import dotty.tools.dotc.util.SourceFile
+import dotty.tools.dotc.util.SourcePosition
+import dotty.tools.dotc.util.Spans.Span
+import org.eclipse.{lsp4j as l}
+
+object AmmoniteFileCompletions:
+  val CURSOR = "_CURSOR_"
```
tgodzik, line 42:
```suggestion
  val CURSOR = "CURSOR"
```
It's adifferent name in Scala 3. Probably best to move it to a separate object. It's defined in CompletionsProvider:145

mtags/src/main/scala-3/scala/meta/internal/pc/completions/AmmoniteFileCompletions.scala
```diff
@@ -0,0 +1,127 @@
...
+      val split = rawPath
+        .split("\\$file")
+        .toList
+
+      val query = selector.collectFirst {
+        case sel: ImportSelector if sel.name.toString().contains(CURSOR) =>
+          sel.name.toString.replace(CURSOR, "")
+
+        case sel: ImportSelector =>
+          sel.name.toString
```
tgodzik, line 79:
```suggestion
        case sel: ImportSelector =>
          sel.name.toString.replace(CURSOR, "")

```
The result should be the same.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/AmmoniteFileCompletions.scala
```diff
@@ -0,0 +1,127 @@
...
+                CompletionValue.FileSystemMember(
+                  isDirectory = file.isDirectory,
+                  file.filename,
+                  editRange,
+                )
+            } ++ parentTextEdit
+        case _ =>
+          Nil
+      end match
+    catch
```
tgodzik, line 123: No need to catch it here, it should be caught in the CompilerAccess and shown properly.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
```diff
@@ -318,7 +322,7 @@ class CompletionsProvider(
         mkItem(label, text.getOrElse(label))
       case CompletionValue.Document(label, doc, desc) =>
         mkItem(label, doc, filterText = Some(desc))
-      case _ =>
+      case _ => // TODO check if CompletionValue.FileSystemMember should be covered separately
```
tgodzik, line 325:
```suggestion
      case _ =>
```
I think it should be fine as long as it works.

tgodzik (changes requested):
Great work! I left a couple of comments, but nothing major.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
```diff
@@ -64,6 +65,8 @@ sealed trait CompletionValue:
         printer.tpe(tpe)
       case CompletionValue.Document(_, _, desc) => desc
       case _: CompletionValue.Keyword => ""
+      case fileSysMem @ CompletionValue.FileSystemMember(one, two, three) =>
```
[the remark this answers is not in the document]
vzmerr: I remember the tests were failing until I added this, because the result of the test was the whole `FileSystemMember` with all the fields instead of just the label.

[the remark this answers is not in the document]
tgodzik:
```suggestion
      case _: CompletionValue.FileSystemMember => ""
```
maybe then

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
```diff
@@ -330,3 +336,6 @@ class CompletionsProvider(
     end match
   end completionItems
 end CompletionsProvider
+
+case object Scala3CursorString:
+  val value = "CURSOR"
```
vzmerr, line 341: cursor came here

tgodzik: Maybe we could put it into Completions companion object under `CURSOR` value?

vzmerr: It won't be a big improvement, would it? Since it is used only in `CompletionProvider` and `AmmoniteFileCompletion`

tgodzik: I don't think we need the additional `Scala3CursorString` object. 

Alternatively, you could rename it to `Cursor`  Scala3 part is redundant since we are in Scala 3 sources and String is not really true since it's an object

vzmerr: done

mtags/src/main/scala-3/scala/meta/internal/pc/completions/AmmoniteFileCompletions.scala
```diff
@@ -0,0 +1,120 @@
...
+import dotty.tools.dotc.core.StdNames
+import dotty.tools.dotc.core.SymDenotations.SymDenotation
+import dotty.tools.dotc.core.Symbols.NoSymbol
+import dotty.tools.dotc.core.Symbols.Symbol
+import dotty.tools.dotc.core.Types.Type
+import dotty.tools.dotc.interactive.Interactive
+import dotty.tools.dotc.interactive.InteractiveDriver
+import dotty.tools.dotc.util.SourceFile
+import dotty.tools.dotc.util.SourcePosition
+import dotty.tools.dotc.util.Spans.Span
```
tanishiking, line 38: Looks like those are unused imports?
```suggestion
```

mtags/src/main/scala-3/scala/meta/internal/pc/completions/AmmoniteFileCompletions.scala
```diff
@@ -0,0 +1,120 @@
...
+
+import scala.meta.internal.mtags.MtagsEnrichments.*
+import scala.meta.internal.pc.AutoImports.AutoImport
+import scala.meta.internal.pc.AutoImports.AutoImportsGenerator
+import scala.meta.internal.pc.printer.MetalsPrinter
+import scala.meta.io.AbsolutePath
+import scala.meta.pc.OffsetParams
+import scala.meta.pc.PresentationCompilerConfig
+import scala.meta.pc.PresentationCompilerConfig.OverrideDefFormat
+import scala.meta.pc.SymbolSearch
```
tanishiking, line 17: unused (I wish Scala3 compiler could have a flag to detect unused imports in the near future 😭 )

```suggestion
import scala.meta.io.AbsolutePath
```

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
tanishiking: Not a strong opinion, but I'm wondering why we wrap the cursor by `case object Cursor` instead of just defining the constant value `CURSOR`? I don't think there would have other members under `Cursor` case object other than `value`.

tanishiking (commented):
I left some nitpick comments, but otherwise looks good to me :tada:

mtags/src/main/scala-3/scala/meta/internal/pc/completions/AmmoniteFileCompletions.scala
vzmerr: Oh, I wonder why Scalafix has not removed them, already🧐
Thanks for pointing it out!

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
vzmerr: hmm, currently it is accessed by both `CompletionProvider` and `AmmoniteFileCompletion`, so if defined as a constant, it had to be passed all along the way as an argument from `CompletionProvider` through `Completions` to reach `AmmoniteFileCompletion` as a value. Or, it could be wrapped in an object to be statically available where it is needed.

vzmerr:
@tgodzik I wonder if it would be a good idea to have a "releasable" main branch in addition to the current main. That way, we would be able to merge and test features by a week or two sooner?

tgodzik:
> @tgodzik I wonder if it would be a good idea to have a "releasable" main branch in addition to the current main. That way, we would be able to merge and test features by a week or two sooner?

Maybe we could do that next time? But I feel like it's also fine to just postpone merging for a while.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/AmmoniteFileCompletions.scala
```diff
@@ -0,0 +1,99 @@
...
+    val fileName = rawFileName
+      .split("/")
+      .last
+      .stripSuffix(".amm.sc.scala")
+
+    val split = rawPath
+      .split("\\$file")
+      .toList
+
+    val editRange = selector.headOption.map { sel =>
```
tgodzik, line 52: I adjusted the ranges since the completions seemed to not be inserted correctly and only appended to the end.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/AmmoniteFileCompletions.scala
```diff
@@ -0,0 +1,99 @@
...
+import org.eclipse.{lsp4j as l}
+
+object AmmoniteFileCompletions:
+
+  private def translateImportToPath(tree: Tree): String =
+    tree match
+      case Select(qual, name) =>
+        val pathPart = name.toString()
+        translateImportToPath(qual) + "/" + {
+          if pathPart == "^" then ".."
```
tgodzik, line 27: Fixed this one since Dotty doesn't change `^` into `$up`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
```diff
@@ -120,6 +121,14 @@ object CompletionValue:
   case class Keyword(label: String, override val insertText: Option[String])
       extends CompletionValue
 
+  case class FileSystemMember(
+      filename: String,
```
tgodzik, line 125: I simplified FileSystemMember a bit

tgodzik (approved):
I managed to test it and there were two smaller issues I fixed, so it looks to be working correctly now.

PR #4221, opened 2022-08-03
chore(ci): migrate to actions/java to use built-in sbt caching

ckipp01:
I'd like to see if this works. There were a couple coversations a while back about
using setup-scala vs setup-java and it seems that many repos have migrated to just
using setup-java since it does offer sbt support, has a ton of users, and also
built-in sbt caching which simplifies a few things in our actions.

If you're rather just leave it as is since it is working fine, no
worries, I just really like that the setup-java offers built-in caching
for sbt.

### DIFF
--- a/tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala
+++ b/tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala
@@ -322,21 +322,38 @@ class WorkspaceSymbolLspSuite extends BaseLspSuite("workspace-symbol") {
       _ <- server.didOpen("a/src/main/scala/a/Before.scala")
       _ = assertNoDiff(
         server.workspaceSymbol("Future"),
-        """|scala.concurrent.Future
-           |scala.concurrent.Future
-           |java.util.concurrent.Future
-           |scala.sys.process.ProcessImpl#Future
-           |scala.jdk.FutureConverters.FutureOps
-           |java.util.concurrent.FutureTask
-           |java.io.ObjectStreamClass#EntryFuture
-           |java.util.concurrent.RunnableFuture
-           |java.util.concurrent.ExecutorCompletionService#QueueingFuture
-           |java.util.concurrent.ScheduledFuture
-           |scala.jdk.FutureConverters
-           |scala.jdk.javaapi.FutureConverters
-           |java.util.concurrent.CompletableFuture
-           |java.util.concurrent.ScheduledThreadPoolExecutor#ScheduledFutureTask
-           |""".stripMargin,
+        if (isJava17) {
+          """|scala.concurrent.Future
+             |scala.concurrent.Future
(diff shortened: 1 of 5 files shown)


tgodzik (approved):
I think it's good to use the setup-java action instead of Olaf's one, since we don't really want to maintain it.

If it works, let's merge it!

ckipp01:
Weird. So much more flaky.

tgodzik:
> Weird. So much more flaky.

It doesn't look flaky actually, but rather uses a different JDK than we did previously.

ckipp01:
> > Weird. So much more flaky.
> 
> It doesn't look flaky actually, but rather uses a different JDK than we did previously.

O duh, I moved to temurin because setup-java didn't list adopt as an option for 17. Lemme dig in to see if they actually do or what's the best approach.

tgodzik:
Temurin isn't the new adopt? I was rather thinking that maybe jabba had a very old version of JDK 17 and the sources changed a bit.

ckipp01:
> Temurin isn't the new adopt? I was rather thinking that maybe jabba had a very old version of JDK 17 and the sources changed a bit.

Yea from my understanding adopt is adoptium and releasing under temurin. 🤔 I'll need to dig in a bit more because I can change these to pass on 17, but they will then fail on 8.

tgodzik:
> > Temurin isn't the new adopt? I was rather thinking that maybe jabba had a very old version of JDK 17 and the sources changed a bit.
> 
> Yea from my understanding adopt is adoptium and releasing under temurin. thinking I'll need to dig in a bit more because I can change these to pass on 17, but they will then fail on 8.

Don't we have already some different expected for Java 8 and 17 in most places?

ckipp01:
> Don't we have already some different expected for Java 8 and 17 in most places?

Yup, that's what I'll use.

ckipp01:
Alright ready to rock and roll if you want to give it another look over.

tests/unit/src/test/scala/tests/WorkspaceSymbolLspSuite.scala
```diff
@@ -348,15 +365,25 @@ class WorkspaceSymbolLspSuite extends BaseLspSuite("workspace-symbol") {
...
+             |scala.jdk.FutureConverters
+             |scala.jdk.javaapi.FutureConverters
+             |scala.concurrent.impl.FutureConvertersImpl
+             |""".stripMargin
+        else
+          """|scala.concurrent.Future
+             |scala.concurrent.Future
+             |scala.sys.process.ProcessImpl#Future
+             |scala.jdk.FutureConverters.FutureOps
+             |java.io.ObjectStreamClass#EntryFuture
```
tgodzik, line 382: Weirdly enough this seems to be available still in JDK 17 :thinking:

tgodzik (approved):
Looks good, let merge it after the release.

PR #4223, opened 2022-08-03
bugfix: Show package completions

tgodzik:
Previously we would not show package completions in a number of places, which is useful when using for example pprint. Now we properly show them in probably all the situations.

Looks like we didn't have tests for these.

Scala 2 tests have an additional completion, but that seems to be ignored by VS Code, though I wonder if this is an issue for other editors :thinking: 

@tanishiking Let's merge it before the release.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -516,6 +516,8 @@ class Completions(
               val include =
                 !isUninterestingSymbol(sym) &&
                   isNotLocalForwardReference(sym)
+                  // to not duplicate with package itself
+                  && isNotPackageObject(sym)
                   && isNotAModuleOrModuleIsValidForPos(sym)
                   && isNotAMethodOrMethodIsValidForPos(sym)
                   && isNotClassOrTraitOrTheyAreValidForPos(sym)
@@ -545,6 +547,9 @@ class Completions(
 
     end filterInteresting
 
+    private def isNotPackageObject(sym: Symbol) =
+      !sym.isPackageObject
+
     private def isNotAModuleOrModuleIsValidForPos(sym: Symbol) =
       !sym.info.typeSymbol.is(
         Flags.Module
@@ -556,10 +561,9 @@ class Completions(
       ) // !sym.info.typeSymbol.is(Flags.Method) does not detect Java methods
 
     private def isNotClassOrTraitOrTheyAreValidForPos(sym: Symbol) =
-      cursorPositionCondition.isClassOrTraitValidForPos || sym.is(
-        Flags.Method
-      ) || (!sym.isClass && !sym.info.typeSymbol
-        .is(Trait))
+      cursorPositionCondition.isClassOrTraitValidForPos ||
+        sym.is(Flags.Method) || sym.is(Package) ||
+        !sym.isClass && !sym.info.typeSymbol.is(Trait)
 
   end extension
 
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -515,13 +515,14 @@ class Completions(
 
               val include =
                 !isUninterestingSymbol(sym) &&
-                  isNotLocalForwardReference(sym)
-                  // to not duplicate with package itself
-                  && isNotPackageObject(sym)
-                  && isNotAModuleOrModuleIsValidForPos(sym)
-                  && isNotAMethodOrMethodIsValidForPos(sym)
-                  && isNotClassOrTraitOrTheyAreValidForPos(sym)
-
+                  isNotLocalForwardReference(sym) && (
+                    // to not duplicate with package itself
+                    sym.is(Package) ||
+                      isNotPackageObject(sym)
+                      && isNotAModuleOrModuleIsValidForPos(sym)
+                      && isNotAMethodOrMethodIsValidForPos(sym)
+                      && isNotClassOrTraitOrTheyAreValidForPos(sym)
+                  )
               (id, include)
             case kw: CompletionValue.Keyword => (kw.label, true)
             case namedArg: CompletionValue.NamedArg =>
@@ -562,7 +563,7 @@ class Completions(
 
     private def isNotClassOrTraitOrTheyAreValidForPos(sym: Symbol) =
       cursorPositionCondition.isClassOrTraitValidForPos ||
-        sym.is(Flags.Method) || sym.is(Package) ||
+        sym.is(Flags.Method) ||
         !sym.isClass && !sym.info.typeSymbol.is(Trait)
 
   end extension
--- a/tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
@@ -1479,4 +1479,29 @@ class CompletionSuite extends BaseCompletionSuite {
     "",
   )
 
+  check(
+    "pkg",
+    s"""|object Foo {
+        |  scala.coll@@
+        |}
+        |""".stripMargin,
+    "collection scala",
+  )
+
+  check(
+    "pkg-scala",
+    s"""|object Foo {
+        |  scala@@
+        |}
+        |""".stripMargin,
+    """|scala <root>
+       |""".stripMargin,
+    compat = Map(
+      "2" ->
+        """|scala _root_
+           |`package` - scala
+           |""".stripMargin
+    ),
+  )
+
 }
--- a/tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
+++ b/tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
@@ -1480,7 +1480,7 @@ class CompletionSuite extends BaseCompletionSuite {
   )
 
   check(
-    "pkg",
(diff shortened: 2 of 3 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -556,10 +561,9 @@ class Completions(
       ) // !sym.info.typeSymbol.is(Flags.Method) does not detect Java methods
 
     private def isNotClassOrTraitOrTheyAreValidForPos(sym: Symbol) =
-      cursorPositionCondition.isClassOrTraitValidForPos || sym.is(
-        Flags.Method
-      ) || (!sym.isClass && !sym.info.typeSymbol
-        .is(Trait))
+      cursorPositionCondition.isClassOrTraitValidForPos ||
```
vzmerr, line 564:
```suggestion
      cursorPositionCondition.isClassTraitOrPackageValidForPos ||
```

tgodzik: Maybe `isNotValidObject` ?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -556,10 +561,9 @@ class Completions(
...
+        sym.is(Flags.Method) || sym.is(Package) ||
```
vzmerr, line 565: I would say it is better to put `|| sym.is(Package)` in `include` because package is really valid for all positions of type completion I can think of:

- `val a : scala.colle@@`
- `new scala.collec@@`
- `scala.collec@@`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -516,6 +516,8 @@ class Completions(
               val include =
                 !isUninterestingSymbol(sym) &&
                   isNotLocalForwardReference(sym)
+                  // to not duplicate with package itself
+                  && isNotPackageObject(sym)
                   && isNotAModuleOrModuleIsValidForPos(sym)
                   && isNotAMethodOrMethodIsValidForPos(sym)
                   && isNotClassOrTraitOrTheyAreValidForPos(sym)
```
vzmerr, line 523:
```suggestion
                  && isNotPackageObject(sys)
                 && ( (  isNotAModuleOrModuleIsValidForPos(sym)
                  && isNotAMethodOrMethodIsValidForPos(sym)
                  && isNotClassOrTraitOrTheyAreValidForPos(sym))  
                  || sym.is(Package))
```

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -1479,4 +1479,29 @@ class CompletionSuite extends BaseCompletionSuite {
     "",
   )
 
+  check(
+    "pkg",
+    s"""|object Foo {
+        |  scala.coll@@
```
vzmerr, line 1485: perhaps, it would be better if the test cases cover the cases listed [here](https://github.com/scalameta/metals/pull/4223/files#r936944963)

vzmerr: only the last one is covered now

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: `isObjectInvalidForPos`

vzmerr: but that would be cryptic a little bit for later maintenance

tgodzik: But long names are difficult to differentiate when looking at the code.

vzmerr: then, perhaps the longer explanation can go in scaladoc

tgodzik: Ok moved it out of the method, so I will leave it as is.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: Good catch, I added the test case and fixed the issues that popped up.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: Done!

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
tgodzik: They should cover all of them now

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -515,11 +515,14 @@ class Completions(
...
                 !isUninterestingSymbol(sym) &&
-                  isNotLocalForwardReference(sym)
-                  && isNotAModuleOrModuleIsValidForPos(sym)
-                  && isNotAMethodOrMethodIsValidForPos(sym)
-                  && isNotClassOrTraitOrTheyAreValidForPos(sym)
-
+                  isNotLocalForwardReference(sym) && (
+                    // to not duplicate with package itself
+                    sym.is(Package) ||
+                      isNotPackageObject(sym)
```
vzmerr, line 521: so if it is a package and is a package object, include should be true?

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -1479,4 +1479,47 @@ class CompletionSuite extends BaseCompletionSuite {
...
+    "pkg".tag(IgnoreScalaVersion.for3LessThan("3.1.3")),
+    s"""|object Foo {
+        |  val a : scala.coll@@
+        |}
+        |""".stripMargin,
+    "collection scala",
+  )
+
+  check(
+    "pkg".tag(IgnoreScalaVersion.for3LessThan("3.1.3")),
```
vzmerr, line 1501:
```suggestion
    "pkg-new".tag(IgnoreScalaVersion.for3LessThan("3.1.3")),
```

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -1479,4 +1479,47 @@ class CompletionSuite extends BaseCompletionSuite {
...
+    s"""|object Foo {
+        |  scala.coll@@
+        |}
+        |""".stripMargin,
+    "collection scala",
+  )
+
+  check(
+    "pkg".tag(IgnoreScalaVersion.for3LessThan("3.1.3")),
```
vzmerr, line 1492:
```suggestion
    "pkg-type".tag(IgnoreScalaVersion.for3LessThan("3.1.3")),
```

vzmerr: so that the three tests would not have the same name

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: It appears to me that the conditions need to be like this https://github.com/scalameta/metals/pull/4223#discussion_r936947521

vzmerr: with `isNotPackageObject` being out and separately "And"ed with all other conditions..

vzmerr:
As a last comment, imho, in all the package completions, we should suffix the package with `.` to hint towards selecting a member afterwards.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: I think the result is the same actually, so I will leave it as is.

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
tgodzik: Fixed!

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
tgodzik: done

tgodzik:
> As a last comment, imho, in all the package completions, we should suffix the package with `.` to hint towards selecting a member afterwards.

Maybe, but it's not really a big issue to type `.` manually. We can think about it later.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: with the way it is written now, if 

- `sym.is(Package)` is `true` and
-  `isNotPackageObject(sym)` is` false`

then `include` would be `true`.

With the suggested way, if  `isNotPackageObject(sym)` is` false`, then `include` is always going to be `false`.

tgodzik: Actually this doesn't happen. Both are either true or false.

tgodzik:
I merged it since this is needed for the next release.

Issue #4213, opened 2022-08-01
Inconsistent and incomplete completion results for Scala 2 and 3


vzmerr:
### Describe the bug

If you put this in `CompletionSuite`, you would see that there are differences between the results returned for scala 3 and 2 

- some of the results are members of java._ so they should not differ between Scala 2 and 3.
- some of the results of Scala 2 such as `String=` seem to be incorrect for the position.
-  a valid result such as `StringIndexOutOfBoundsException`, is omitted from Scala 3 result, although it is part of java api
  
```Scala
check(
    "simple-short".only,
    """
      |object A {
      |  val a: Strin@@
      |}""".stripMargin,
    """|String java.lang
       |StringBuffer java.lang
       |StringContext scala
       |StringFormat scala.Predef
       |StringOps - scala.collection
       |StringView - scala.collection
       |FromString - scala.util.CommandLineParser
       |StringJoiner - java.util
       |StringReader - java.io
       |StringWriter - java.io
       |StringBuilder - java.lang
       |StringParsers - scala.collection
       |StringToExpr - scala.quoted.ToExpr
       |""".stripMargin,
    compat = Map(
      "2" -> """|StringBuilder scala.collection.mutable
                |String java.lang
                |String= String
                |StringBuffer java.lang
                |StringBuilder java.lang
                |StringBuilder= StringBuilder
                |StringContext scala
                |StringIndexOutOfBoundsException java.lang
                |StringIndexOutOfBoundsException= StringIndexOutOfBoundsException
                |StringFormat scala.Predef
                |String - scala.math.Equiv
                |String - scala.math.Ordering
                |StringOps - scala.collection
                |StringProp - scala.sys.Prop
                |StringView - scala.collection
                |StringFormat - scala.Predef: [A](self: A): StringFormat[A] <and> StringFormat.type
                |StringJoiner - java.util
                |StringReader - java.io
                |StringWriter - java.io
                |StringParsers - scala.collection
                |""".stripMargin
    ),
  )
```

Also please note that `StringBufferInputStream` in the below test is generated, while omitted from the results of the above tests for both Scala 2 and 3.

```Scala
 check(
    "simple",
    """
      |object A {
      |  val a: StringBuffe@@
      |}""".stripMargin,
    """|StringBuffer java.lang
       |StringBufferInputStream - java.io
       |""".stripMargin,
  )

```

### Expected behavior

Consistent and complete results for scala 2 and 3 except when the Scala API itself differs.

### Version of Metals

[...]

PR #4224, opened 2022-08-04
objects completion in New and Type positions

vzmerr:
It partially addresses #4213 
✅ allowing Objects for type completion in New and Type positions.

TODO(Out of Scope)
- filtering out objects that are private members and inaccessible in scope. The catch is that even the java objects with only private and protected constructors can have a member that can be instantiated, so we should still include them in `new` positions.
- filtering out objects with no member that can be instantiated in `new` position.
- filtering out java objects, aka classes, with private and protected constructors in `new` and type positions; because they cannot be instantiated in Scala code. The catch is that  java objects with protected constructors can be obtained as a member or return type of a method in another class or object located in the same package as them, such as the so called "builder" ones.

**Update**

When the class/trait does not take any type parameters, it is useless to offer both the companion object and class/trait in the type position. Because both are going to generate the same edit.
lets say 

```
case class Foo()
object Foo:
  ???

object Main:
  val a: Foo@@

```

**Update**

The PR now introduces four additional changes:

✅**fixing the symName in SemanticdbSymbols, by adding # to the descriptor of classes and traits to mark them as types**
✅**filtering objects only when type parameter is present**
✅ **allowing dot in the index id of object symbols for differentiation from class**
✅ **covering the case for extends in detecting type position**

**Update**

TODO (**Out of the Scope of the PR**):
- **We need to filter out all the case classes, and classes and traits with `final` identifier from type completions after `extends` and `with`**
- **We need to filter out all the classes and traits with `sealed` identifier not defined in the same file, from type completions after `extends` and `with`**

Update:

Done: 

✅ **allowing JavaModule to be suggested where objects are not allowed, which is where type parameters are present, beacuse Java objects are actually classes**

**Done:** 

✅ **not allowing classes in new position. This fix, although at first glance is unrelated to the main topic of the PR, is done for getting the tests to pass.**
✅ **allowing duplicate class and companion object where the class takes type parameters** 
✅**allowing duplicate trait and companion object where the trait takes type parameters or curly braces.**

TODO(Out of Scope):
- fixing the printed label of classes/traits to include type parameters

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -89,7 +89,6 @@ class Completions(
       isNewPosition: Boolean,
       isInstantiationOrMethodCallPos: Boolean,
       noSquareBracketExists: Boolean,
-      isImportPosition: Boolean,
   ):
 
     /**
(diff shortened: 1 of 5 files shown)


tgodzik:
> filtering out objects that are private members and inaccessible in scope

That could be done separately in another PR.

> filtering out objects with no member that can be instantiated in new position.
I don't think we should do that, it would most likely be inefficient.

> detecting java objects, which are actually classes, and removing the . suffix from them.
We should add `.` suffix to them in the first place i think.

> filtering out java objects, aka classes, with private and protected constructors in new and type positions; because they cannot be instantiated in Scala code.

This could also be another PR

vzmerr:
> We should add . suffix to them in the first place i think.

should or should not?

tgodzik:
> > We should add . suffix to them in the first place i think.
> 
> should or should not?

Shouldn not :sweat_smile:  I updated the comment

dos65 (approved):
Let's include it into release.
@vzmerr there is conflict - could you rebase it on the main?

vzmerr:
> Let's include it into release.
> @vzmerr there is conflict - could you rebase it on the main?

@dos65 
sure, there is also an additional problem: currently the index used in `filterInteresting` is using `symbolName` as the key, which causes the objects which are now not filtered out to shade and block the classes and traits with the same key from entering the index :(
I am trying to come up with a better key for the index!
Any idea?

dos65:
@vzmerr 

> sure, there is also an additional problem: currently the index used in filterInteresting is using symbolName as the key, which causes the objects which are now not filtered out to shade and block the classes and traits with the same key from entering the index :(

yep. there is a such thing - we needed it to avoid duplication in completions items as at that time we didn't have a suffix (`[]` | `()`) support for scala3. Now we can just remove this hack from `filterInteresting` and there is no need to touch `SemanticdbSymbols`

vzmerr:
@dos65 Hey Vadim, please feel free to take on this PR and change it as you please to fit in the frame of the new rework. Please feel free to let me know if the explanations are insufficient or you need help ;)

tanishiking:
closing for now as it would be easier to re-work based on https://github.com/scalameta/metals/pull/4234, but anyone wants to work from this, feel free to re-open :)

PR #4225, opened 2022-08-05
Don't run InteractiveDriver.run if the content didn't "change"

tanishiking:
If the content didn't change since the last `driver.run`, do not compile
because the compilation unit should have the up to date typed tree.
Skip compilation saves up much of CPU usage when users are reading
source code (and hover on symbols) without modifying the source code.

---

What happens if we change in a file affect another file's type
information?

For example, when we have the following source files, if we hover on
`B.foo` in `A.scala`, Metals shows the symbol's type is `Int`.

```scala
// A.scala
def main = println(B.foo)

// B.scala
object B:
  def foo: Int = ???
```

When we modify `B.scala` to `def foo: String = ???`, what happens to
`A.scala`?
You may think that, Metals keep using the old compilation
unit that shows `B.foo` as `Int` instead of `String` because Metals
invalidates the compilation cache based on its content hash.

That's not true, `alreadyCompiled("A.scala", <content hash>)` will be
`false`, and Metals will run `driver.run` on `A.scala` and show `String`
for `B.foo` in `A.scala`.

Why? Because Metals will recreate a new `MetalsDriver` and `lastCompiled`
cache will be cleared up.

- When a file has changed, Metals call [Compilers.restart](https://github.com/scalameta/metals/blob/c6e0f30febeff67e970e463d79862e27e45ada3f/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala#L247).
- The restart method calls `shutdownCurrentCompiler` that [set `_compiler` to `null`.](https://github.com/scalameta/metals/blob/c6e0f30febeff67e970e463d79862e27e45ada3f/mtags/src/main/scala/scala/meta/internal/pc/CompilerAccess.scala#L57-L74).

This is basically the same way to cache compilation as Scala2
See my explanation on how Scala2 + Metals compilation cache works
https://contributors.scala-lang.org/t/how-does-scala2-compiler-cache-invalidate-the-typechecking-result-for-the-given-tree/5858/4?u=tanishiking

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
@@ -23,7 +23,7 @@ import dotty.tools.dotc.util.SourceFile
  * the typed tree under the situation like developers
  * sequentially hover on the symbols in the same file without any changes.
  *
- * Note: we decided to cache only if the target URI only if the same as the previous run
+ * Note: we decided to cache only if the target URI is the same as in the previous run
  * because of `InteractiveDriver.currentCtx` that should return the context that
  * refers to the last compiled source file.
  * It would be ideal if we could update currentCtx even when we skip the compilation,
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
@@ -1,0 +1,35 @@
+package scala.meta.internal.pc
+
+import java.net.URI
+import java.{util as ju}
+
+import scala.collection.concurrent.TrieMap
+
+import scala.meta.internal.mtags.MD5
+
+import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.interactive.InteractiveDriver
+import dotty.tools.dotc.interactive.SourceTree
+import dotty.tools.dotc.reporting.Diagnostic
+import dotty.tools.dotc.util.SourceFile
+
+class MetalsDriver(
+    override val settings: List[String]
+) extends InteractiveDriver(settings):
+
+  private def alreadyCompiled(uri: URI, content: Array[Char]): Boolean =
+    compilationUnits.get(uri) match
+      case Some(unit) if ju.Arrays.equals(unit.source.content(), content) =>
+        true
+      case _ => false
+
+  override def run(uri: URI, source: SourceFile): List[Diagnostic] =
+    if alreadyCompiled(uri, source.content) then Nil
+    else super.run(uri, source)
+
+  override def run(uri: URI, sourceCode: String): List[Diagnostic] =
+    val contentHash = MD5.compute(sourceCode)
+    if alreadyCompiled(uri, sourceCode.toCharArray()) then Nil
+    else super.run(uri, sourceCode)
+
+end MetalsDriver
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
@@ -13,23 +13,50 @@ import dotty.tools.dotc.interactive.SourceTree
 import dotty.tools.dotc.reporting.Diagnostic
 import dotty.tools.dotc.util.SourceFile
 
+/**
+ * MetalsDriver is a wrapper class that provides a compilation cache for InteractiveDriver.
+ * MetalsDriver skips running compilation if
+ * - the target URI of `run` is the same as the previous target URI
+ * - the content didn't change since the last compilation.
+ *
+ * This compilation cache enables Metals to skip compilation and re-use
+ * the typed tree under the situation like developers
+ * sequentially hover on the symbols in the same file without any changes.
+ *
+ * Note: we decided to cache only if the target URI only if the same as the previous run
+ * because of `InteractiveDriver.currentCtx` that should return the context that
+ * refers to the last compiled source file.
+ * It would be ideal if we could update currentCtx even when we skip the compilation,
+ * but we struggled to do that. See the discussion https://github.com/scalameta/metals/pull/4225#discussion_r941138403
+ * To avoid the complexity related to currentCtx,
+ * we decided to cache only when the target URI only if the same as the previous run.
+ */
 class MetalsDriver(
     override val settings: List[String]
 ) extends InteractiveDriver(settings):
 
+  @volatile private var lastCompiledURI: URI = _
+
   private def alreadyCompiled(uri: URI, content: Array[Char]): Boolean =
     compilationUnits.get(uri) match
-      case Some(unit) if ju.Arrays.equals(unit.source.content(), content) =>
+      case Some(unit)
+          if lastCompiledURI == uri &&
+            ju.Arrays.equals(unit.source.content(), content) =>
         true
       case _ => false
 
   override def run(uri: URI, source: SourceFile): List[Diagnostic] =
-    if alreadyCompiled(uri, source.content) then Nil
-    else super.run(uri, source)
+    val diags =
+      if alreadyCompiled(uri, source.content) then Nil
+      else super.run(uri, source)
+    lastCompiledURI = uri
+    diags
 
   override def run(uri: URI, sourceCode: String): List[Diagnostic] =
-    val contentHash = MD5.compute(sourceCode)
-    if alreadyCompiled(uri, sourceCode.toCharArray()) then Nil
-    else super.run(uri, sourceCode)
+    val diags =
+      if alreadyCompiled(uri, sourceCode.toCharArray()) then Nil
+      else super.run(uri, sourceCode)
+    lastCompiledURI = uri
+    diags
 
 end MetalsDriver
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
@@ -5,11 +5,8 @@ import java.{util as ju}
 
 import scala.collection.concurrent.TrieMap
 
-import scala.meta.internal.mtags.MD5
-
 import dotty.tools.dotc.core.Contexts.Context
 import dotty.tools.dotc.interactive.InteractiveDriver
-import dotty.tools.dotc.interactive.SourceTree
 import dotty.tools.dotc.reporting.Diagnostic
 import dotty.tools.dotc.util.SourceFile
 
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/Scala3CompilerAccess.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/Scala3CompilerAccess.scala
@@ -24,7 +24,7 @@ class Scala3CompilerAccess(
     sh: Option[ScheduledExecutorService],
     newCompiler: () => Scala3CompilerWrapper,
 )(using ec: ExecutionContextExecutor)
-    extends CompilerAccess[StoreReporter, InteractiveDriver](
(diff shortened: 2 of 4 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
```diff
@@ -0,0 +1,42 @@
...
+import dotty.tools.dotc.interactive.InteractiveDriver
+import dotty.tools.dotc.interactive.SourceTree
+import dotty.tools.dotc.reporting.Diagnostic
+import dotty.tools.dotc.util.SourceFile
+
+class MetalsDriver(
+    override val settings: List[String]
+) extends InteractiveDriver(settings):
+
+  private val lastCompiled: TrieMap[URI, String] = TrieMap.empty
```
tanishiking, line 19: I'm not sure which is better
- Storing content as `Array[Char]` and compare them by `Array.equals(...)` vs
- Hash content by md5 (or something) and compare
  - md5 consumes much less memory, but it requires encryption each time
  - but `lastCompiled` will be short-lived ones (as it will be expired when users modify a file), and memory footprint may not matter.

tgodzik: Could we rely on didChange/didOpen notifications and force recompilation if the method gets invoked? We wouldn't need to compute anything in that case. so should be faster.

tanishiking: > Could we rely on didChange/didOpen notifications and force recompilation if the method gets invoked?

Yeah, I think we can do that!
However, I prefer not to rely on those `didChange/didOpen` events because

- hash comparison (or `Array.equals`) is already pretty fast
  - I commented about `MD5` vs `Array.equals`, but it's about micro-optimization. TBH, either is practically fine (I was just curious) 😅 
- Cache invalidation by those events makes the source code more complicated, and even harder to troubleshoot.

tgodzik: ok sure, I was just worried that maybe the hash comparison was problematic :sweat_smile:

kpodsiad: @tanishiking you could use `TimerProvider` to measure how much time computing hash takes (but maybe log in on debug channel?)

dos65: I think it would be better at least to attach `didClose` notifications to cleanup cache.
Hypothetically, if you have a project in broken state for a long time it might took some memory ) all units will be in memory until compiler reload, which happens only if bsp compilation pass

mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
```diff
@@ -0,0 +1,42 @@
...
+      .map(cache => cache == hash)
+      .getOrElse(false)
+
+  override def run(uri: URI, source: SourceFile): List[Diagnostic] =
+    val content = source.content.mkString
+    val contentHash = MD5.compute(content)
+    if alreadyCompiled(uri, contentHash) then Nil
+    else
+      lastCompiled.put(uri, contentHash)
+      super.run(uri, source)
```
dos65, line 33: Shouldn't we also override `def currectCtx`?

mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
tanishiking: End up comparing content `Array[Char]` by `ju.Arrays.equals` as Scala2 does. Now `MetalsDriver` compare with the previous one by

```scala
  private def alreadyCompiled(uri: URI, content: Array[Char]): Boolean =
    compilationUnits.get(uri) match
      case Some(unit) if ju.Arrays.equals(unit.source.content(), content) =>
        true
      case _ => false
```

I think this is better because, now we can force re-compile if the driver somehow deleted the compilation unit. (The previous implementation may break the consistency between cache (in `MetalsDriver` and `compilationUnits`).

---

@dos65 
> happens only if bsp compilation pass

I don't think so, presentation compiler instance will be recreated even if BSP compilation fails https://github.com/scalameta/metals/blob/3ca41c31074b91fc4b0c7af0f98a593fa626cd74/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala#L238-L240

mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
tanishiking: I think we can use the `parent`'s `currentCtx` implementation (so I think we don't need to override `currentCtx`), what do you think about it?

dos65: I think it will work incorrectly in  the following case:
- call pc for `A.scala` (compilation happens) - `driver.currentCtx` refers to the ctx from `A.scala`
- call pc for `B.scala` (compilation happens) - `driver.currentCtx` refers to the ctx from `B.scala`
- switch back on `A.scala` without changes, call pc - `driver.currentCtx` refers to the ctx from `B.scala`

tanishiking: Hm, yeah, that should be a problem :+1: (maybe `run` should update the `currentCtx` anyway even MetalsDriver doesn't run compile).

I tried to produce some problems, but couldn't. Did you find any problem with the lack of `currentCtx` implementation?
I realized `driver.currentCtx.source` always returns `<no file>` and `driver.currentCtx.source.tpdTree` returns `EmptyTree`. Also, I skimmed through the `InteractiveDriver` and `Run`, but it looks like no one assigns the given source to the `currentCtx`. Maybe `currentCtx` doesn't refer to any specific files?

tgodzik: It's being used in `InferredTypeProvider` maybe you can check inserting type if it works properly?

dos65: @tanishiking I think it would be better to override it. I've disabled `auto save`  in vscode and got the situation at autoimports here: https://github.com/scalameta/metals/blob/15b8787c0f08520ef3bf54071b6ed8d318d9a703/mtags/src/main/scala-3/scala/meta/internal/pc/AutoImportsProvider.scala#L37-L43

where the request uri wasn't equal to `driver.currentCtx.run.units.head`

tanishiking: @dos65 Great! I could reproduce it :+1:

tanishiking: Tried to override `currentCtx` so that it returns the context connected to the last `run`ed compilation unit
https://github.com/scalameta/metals/pull/4225/commits/553e2b33ecc473849466db0afc4b51ee9162d9f5

This implementation solved the wrong context issue, but something funky started happening, for example when I run code completion. It looks like, under the cached context (whose `runId = 3`) the symbol will be invalid in the new run (where `runId = 4`).

I'm considering, maybe we should override `currentCtx` like

```scala
override def currentCtx =
  val ctx0 = super.currentCtx
  ctx0.setCompilationUnit(<last compiled compilation unit for the file>)
```


```
SEVERE: assertion failed: denotation module class C$ invalid in run 3. ValidFor: Period(1..5, run = 4)
java.lang.AssertionError: assertion failed: denotation module class C$ invalid in run 3. ValidFor: Period(1..5, run = 4)
        at scala.runtime.Scala3RunTime$.assertFailed(Scala3RunTime.scala:8)
        at dotty.tools.dotc.core.Denotations$SingleDenotation.updateValidity(Denotations.scala:722)
        at dotty.tools.dotc.core.Denotations$SingleDenotation.bringForward(Denotations.scala:747)
        at dotty.tools.dotc.core.Denotations$SingleDenotation.toNewRun$1(Denotations.scala:806)
        at dotty.tools.dotc.core.Denotations$SingleDenotation.current(Denotations.scala:880)
        at dotty.tools.dotc.core.Symbols$Symbol.recomputeDenot(Symbols.scala:120)
        at dotty.tools.dotc.core.Symbols$Symbol.computeDenot(Symbols.scala:114)
        at dotty.tools.dotc.core.Symbols$Symbol.denot(Symbols.scala:107)
        at dotty.tools.dotc.core.Symbols$.toDenot(Symbols.scala:497)
        at dotty.tools.dotc.core.Denotations$SingleDenotation.updateValidity(Denotations.scala:721)
        at dotty.tools.dotc.core.Denotations$SingleDenotation.bringForward(Denotations.scala:747)
        at dotty.tools.dotc.core.Denotations$SingleDenotation.toNewRun$1(Denotations.scala:806)
        at dotty.tools.dotc.core.Denotations$SingleDenotation.current(Denotations.scala:880)
        at dotty.tools.dotc.core.Types$NamedType.computeDenot(Types.scala:2256)
        at dotty.tools.dotc.core.Types$NamedType.denot(Types.scala:2216)
        at dotty.tools.dotc.ast.Trees$DenotingTree.denot(Trees.scala:256)
        at dotty.tools.dotc.ast.Trees$Tree.symbol(Trees.scala:145)
        at scala.meta.internal.pc.MetalsInteractive$.contextOfPath(MetalsInteractive.scala:56)
        at scala.meta.internal.pc.MetalsInteractive$.contextOfPath(MetalsInteractive.scala:34)
        at scala.meta.internal.pc.MetalsInteractive$.contextOfPath(MetalsInteractive.scala:34)
        at scala.meta.internal.pc.completions.CompletionsProvider.completions(CompletionsProvider.scala:59)
```

tanishiking: > I'm considering, maybe we should override currentCtx

Realized it doesn't work, we have to set compilation units to the `ctx.run` but it's not accessible outside of `Run`.
Hmmm, maybe we should do this in compiler as Scala2 does 🤔

tanishiking: > if there will be a sequential hover calls for the same source without changes

Ohhh, that's a good idea :+1:

tanishiking: @dos65 I implemented like
> if there will be a sequential hover calls for the same source without changes
and looks like the problem has gone :tada:

tanishiking:
I think it's now ready for review 👍

mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
```diff
@@ -0,0 +1,62 @@
...
+ * MetalsDriver is a wrapper class that provides a compilation cache for InteractiveDriver.
+ * MetalsDriver skips running compilation if
+ * - the target URI of `run` is the same as the previous target URI
+ * - the content didn't change since the last compilation.
+ *
+ * This compilation cache enables Metals to skip compilation and re-use
+ * the typed tree under the situation like developers
+ * sequentially hover on the symbols in the same file without any changes.
+ *
+ * Note: we decided to cache only if the target URI only if the same as the previous run
```
dos65, line 26:
```suggestion
 * Note: we decided to cache only if the target URI is the same as in the previous run
```

mtags/src/main/scala-3/scala/meta/internal/pc/MetalsDriver.scala
```diff
@@ -0,0 +1,62 @@
+package scala.meta.internal.pc
+
+import java.net.URI
+import java.{util as ju}
+
+import scala.collection.concurrent.TrieMap
+
+import scala.meta.internal.mtags.MD5
```
dos65, line 8: unused imports

dos65 (approved):
LGTM!

PR #4229, opened 2022-08-05
[WIP] Add completions for case and match for scala 3

jkciesluk:


tgodzik:
@tanishiking it's still not fully ready, but if you could find the time to look at it that would be awesome.

I was overeager with requesting reviews :sweat_smile:

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -408,10 +408,225 @@ class Completions(
...
+            m.selector,
+            completionPos,
+            path.last,
+            indexedContext,
+            config,
+            parent,
+          ),
+          true,
+        )
+      case (lt @ Literal(
```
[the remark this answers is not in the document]
tanishiking: I think this is something related to https://github.com/scalameta/metals/pull/4229/files#diff-19ea0dcdc913ee6276f02a95699e11ea68dc89a605ecfb2d1e875684b89b1298R625-R626 but why this new case is needed?

mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
```diff
@@ -3,6 +3,7 @@ package scala.meta.internal.mtags
 import java.net.URI
 
 import scala.annotation.tailrec
+import scala.collection.mutable
```
tanishiking, line 6: [nit]
unused ?

tanishiking (commented):
In the middle of reviewing (I haven't yet checking `MatchCompletions.scala`)

tests/cross/src/test/scala/tests/pc/CompletionCaseSuite.scala
```diff
@@ -402,7 +406,7 @@ class CompletionCaseSuite extends BaseCompletionSuite {
   )
 
   check(
-    "brace-negative",
+    "brace-negative".tag(IgnoreScala3),
```
tgodzik, line 409: Could you write what is the issue here? Maybe todo plus create an issue in the repo?

tgodzik (changes requested):

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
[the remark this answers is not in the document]
jkciesluk: This case is needed for 
image
Without this case if we select a completion we would get
image

That's why we have to change completion range

tests/cross/src/test/scala/tests/pc/CompletionCaseSuite.scala
jkciesluk: The issue there was that this was not a case completion but a named argument one and was probably here to check if not suggest case completions. Maybe we can just add compat for scala3 instead of ignoring it?

tests/cross/src/test/scala/tests/pc/CompletionKeywordSuite.scala
```diff
@@ -186,50 +186,6 @@ class CompletionKeywordSuite extends BaseCompletionSuite {
...
-       |  val abc: Option[Int] = ???
-       |  val a = {
-       |    abc match
-       |\tcase $$0
-       |
-       |  }
-       |}
-       |""".stripMargin,
-    filter = _ == "match",
-  )
```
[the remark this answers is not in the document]
jkciesluk: These tests were added with match case keyword completion in https://github.com/scalameta/metals/pull/4185/files
Now match isn't a keyword completion anymore, so thought we can remove these tests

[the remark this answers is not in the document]
tgodzik: Should we keep it as a completion still though? Some people might now want to use the exhaustive check.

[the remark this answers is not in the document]
jkciesluk: For scala2 both exhaustive match and basic match aren't keyword completions, the same here. We can maybe keep basic match as a keyword completion and here have only the exhaustive one, but I wouldn't do it since it was already a special case in keyword completions

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,418 @@
...
+        _.sealedStrictDescendants.foreach(sym =>
+          if !(sym.is(Sealed) && (sym.is(Abstract) || sym.is(Trait))) then
+            subclasses += sym
+        )
+      }
+      val subclassesResult = subclasses.toList
+      subclassesResult
+    end subclassesForType
+
+    val sortedSubclasses = subclassesForType(tpe)
```
tanishiking, line 229: Can't we just get all directSubclasses by [sym.children](https://github.com/lampepfl/dotty/blob/e560c2d6970c864ceea5607597d74cf49c443965/compiler/src/dotty/tools/dotc/core/SymDenotations.scala#L1601-L1632) instead of traverse all the subclasses by `sealedStrictDescendents`?

I think what we're interested in here (exhaustive pattern match) is only it's direct subclasses 🤔

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,418 @@
...
+      val subclasses = ListBuffer.empty[Symbol]
+      val parents = ListBuffer.empty[Symbol]
+      // Get parent types from refined type
+      def loop(tpe: Type): Unit =
+        tpe match
+          case AndType(tp1, tp2) =>
+            loop(tp1)
+            loop(tp2)
+          case t => parents += tpe.typeSymbol
+      loop(tpe.widen.bounds.hi)
```
tanishiking, line 218: What are we doing here? Is there any test that reflects these lines?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -408,10 +408,225 @@ class Completions(
           .filterInteresting(enrich = false)
           ._1
         (completions, true)
+
+      // Match completions
+      case (sel @ Select(qualifier, name)) :: _
+          if "match".startsWith(name.toString()) && text.charAt(
```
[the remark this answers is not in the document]
jkciesluk: For scala2 we are giving this completion for "match" prefixes, shouldn't we do the same here?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,418 @@
...
+   * @param parent the parent tree node of the pattern match, for example `Apply(_, _)` when in
+   *               `List(1).foreach { cas@@ }`, used as fallback to compute the type of the selector when
+   *               it's `EmptyTree`.
+   * @param patternOnly `true` if we want completions without `case` keyword
+   * @param hasBind `true` when `case _: @@ =>`, if hasBind we don't need unapply completions
+   */
+  def contribute(
+      selector: Tree,
+      completionPos: CompletionPos,
+      typedTree: Tree,
```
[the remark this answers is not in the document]
jkciesluk: Thanks, it also fixes an issue with imports positioning

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,418 @@
...
+          )
+        case None =>
+          result += completionGenerator.toCompletionValue(
+            sym.info.dealias.typeSymbol,
+            sym.decodedName,
+            Nil,
+          )
+    }
+
+    val basicMatch = CompletionValue.MatchCompletion(
```
tanishiking, line 254: 👍

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,418 @@
...
+        result += completionGenerator.toCompletionValue(
+          sym,
+          name,
+          autoImports,
+        )
+      end if
+    end visit
+    val selectorSym = parents.selector.typeSymbol
+
+    // Step 1: walk through scope members.
```
tanishiking, line 124: Why we're walking through scope members? I thought it doesn't related to case completions?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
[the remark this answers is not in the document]
jkciesluk: Also, in scala 3 it's allowed to write `xxx.match`
and typed tree here looks the same as for `xxx match`.
Currently we do not show exhaustive match in `xxx.ma@@`, but we do in `xxx.match@@`. I think it's a correct approach , but just want to be sure

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,418 @@
...
+    indexedContext.scopeSymbols
+      .foreach(s => visit(s.info.dealias.typeSymbol, s.decodedName, Nil))
+
+    // Step 2: walk through known direct subclasses of sealed types.
```
tanishiking, line 128:
```suggestion
    // Step 2: walk through known subclasses of sealed types.
```

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,418 @@
...
+      completionPos,
+      clientSupportsSnippets,
+    )
+    val result = ListBuffer.empty[CompletionValue]
+    val tpe = selector.tpe.widen.bounds.hi match
+      case tr @ TypeRef(_, _) => tr.underlying
+      case t => t
+
+    def subclassesForType(tpe: Type)(using Context): List[Symbol] =
+      tpe.typeSymbol.info
```
tanishiking, line 208:
```suggestion
```

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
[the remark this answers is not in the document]
tgodzik: > For scala2 we are giving this completion for "match" prefixes, shouldn't we do the same here?

Actually what case does it cover? So `abc matchThis` would also be covered? Should it? Or `value.matches`?

> Currently we do not show exhaustive match in xxx.ma@@, but we do in xxx.match@@. I think it's a correct approach , but just want to be sure

Yeah, we can change it if anyone reports it as an issue.

tests/cross/src/test/scala/tests/pc/CompletionCaseSuite.scala
tgodzik: It's better to add compat so that reviewers can see easier what is happening.

tanishiking (changes requested):
Great job! I haven't yet checked all the code, but I left several comments and questions.

tanishiking:
One general tip, it would be much easier for reviewers to code-review if you could split this task into smaller pieces and create multiple small PRs.
It looks like this PR tries to implement all the features that Scala2 has, all at once (that is incrementally developed), but we don't need to implement everything in one pull request. We can improve the feature incrementally 👍 

This blog post might be interesting [Anatomy of a perfect pull request | Opensource.com](https://opensource.com/article/18/6/anatomy-perfect-pull-request)


However, overall, this PR is really awesome. Great job!

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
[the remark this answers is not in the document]
jkciesluk: While writing `m`, `ma`, ... `matc` we have `Select`, for `match` we have `Match`, with `.` or without. Match completions are not exclusive, so `value.match@@` can give both exhaustive/basic match and `matches` completions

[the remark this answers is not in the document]
jkciesluk: Edit: Actually not  😅 
image
It's a bit tricky, I will work on it

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,418 @@
...
+        command = config.parameterHintsCommand().asScala,
+      )
+    end if
+
+    val res = result.result()
+    res
+  end contribute
+
+  /**
+   * A `match` keyword completion to generate an exhaustive pattern match for sealed types.
```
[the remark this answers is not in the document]
jkciesluk: Exhaustive match works for scala `enum`, but not for java
Case completions don't work for enum yet

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
jkciesluk: For things like in this test 
https://github.com/scalameta/metals/blob/0443f440379712314c85a2b77a6707177cb326a0/tests/cross/src/test/scala/tests/pc/CompletionCaseSuite.scala#L42-L69

But if we remove walking through scope members, it would enable `enum` completions without the need of any additional changes (they are seen as sealedStringDescendants, but are not a subclass co `visit` filters them out). The only issue then is to change the way we name patterns (remove unnecessary things like `scala.util.None` instead of `None`)

jkciesluk:
Thanks for the tip, I will surely do that for my future PRs. I have some ideas here to remove searching through scope members and use `shortened names`, so should I do that it this PR, or here add only extractor objects and do the rest in a follow-up?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
jkciesluk: `sealedStrictDescendants` is for sealed traits and sealed abstract classes, using it we can be more specific. Maybe we can use `children` and if a child is a `sealed` trait/abstract class do a loop or just use it without "splitting" it more?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
jkciesluk: Yes, this one https://github.com/scalameta/metals/blob/0443f440379712314c85a2b77a6707177cb326a0/tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala#L274-L275
I will add comments explaining it

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -32,8 +32,9 @@ class CompletionMatchSuite extends BaseCompletionSuite {
        |""".stripMargin,
   )
 
+  // In Scala3 it's allowed to write xxx.match
```
tgodzik, line 35: That's a good opportunity to make sure that we properly add {}

tanishiking:
Is this ready for review? If it is, feel free to ping me by request me to review ;)

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
tanishiking: Ah, so this is for non-sealed classes, right?

- First and foremost, I don't think we don't need to provide an exhaustive match for non-sealed classes like this.
  - Because there's nothing like "exhaustiveness" in non-sealed classes and traits.
- Does this work if the class definitions are different files? (I don't think so)
  - I think `children` would provide the direct subclasses (maybe?)

So my suggestion is, to remove this walking through scope symbols for non-sealed types, and let's enable to complete `enum`s :+1:

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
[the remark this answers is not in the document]
tanishiking: That's awesome! Can you add test cases for enum completions?

Not sure how difficult java case completions support is, but we can work on that in another PR :+1: Let's open a follow-up ticket

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,416 @@
...
+      case tr @ TypeRef(_, _) => tr.underlying
+      case t => t
+
+    def subclassesForType(tpe: Type)(using Context): List[Symbol] =
+      val subclasses = ListBuffer.empty[Symbol]
+      val parents = ListBuffer.empty[Symbol]
+      // Get parent types from refined type
+      def loop(tpe: Type): Unit =
+        tpe match
+          case AndType(tp1, tp2) => // for refined type like A with B with C
```
tanishiking, line 212: Why this is needed? Is there any test case that related to this?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
tanishiking: Ahh, never mind, I misunderstood the behavior of `sealedStrictDescendants`.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
jkciesluk: > First and foremost, I don't think we don't need to provide an exhaustive match for non-sealed classes like this. 
We are not, it's only for single `case` completions.

jkciesluk: > Does this work if the class definitions are different files? (I don't think so)

It works if the class is imported or in the same package, so is visible in scope

jkciesluk: > I think children would provide the direct subclasses (maybe?)

Yes, but only for sealed types. For exhaustive match we can use `children` instead of `sealedStrictDescendants`, but when a child is a sealed trait should we split it even more?

jkciesluk: > Why this is needed? Is there any test case that related to this?

Yes, this one https://github.com/scalameta/metals/blob/15b8787c0f08520ef3bf54071b6ed8d318d9a703/tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala#L274-L277

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -6,8 +6,8 @@ class CompletionMatchSuite extends BaseCompletionSuite {
 
   override def requiresScalaLibrarySources: Boolean = true
 
-  override def ignoreScalaVersion: Option[IgnoreScalaVersion] =
-    Some(IgnoreScala3)
+  // override def ignoreScalaVersion: Option[IgnoreScalaVersion] =
+  //   Some(IgnoreScala3)
```
tanishiking, line 10:
```suggestion
```

tests/cross/src/test/scala/tests/pc/CompletionCaseSuite.scala
```diff
@@ -248,9 +248,9 @@ class CompletionCaseSuite extends BaseCompletionSuite {
        |case Some(value) => scala
        |""".stripMargin,
   )
-
+  // for scala 3.2.1 we see subclasses like `Tuple2$mcCC$sp`. Might be an issue in Dotty
```
tanishiking, line 251: Can you open an issue in dotty and link to the issue from the comment?

tests/cross/src/test/scala/tests/pc/CompletionCaseSuite.scala
```diff
@@ -7,8 +7,8 @@ import tests.BaseCompletionSuite
 
 class CompletionCaseSuite extends BaseCompletionSuite {
 
-  override def ignoreScalaVersion: Option[IgnoreScalaVersion] =
-    Some(IgnoreScala3)
+  // override def ignoreScalaVersion: Option[IgnoreScalaVersion] =
+  //   Some(IgnoreScala3)
```
tanishiking, line 11:
```suggestion
```

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
tanishiking: Oops, nevermind, this is for case completion, not for match exhaustive one :bow:

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,416 @@
...
+        Some(
+          if !patternOnly then
+            if config.isCompletionSnippetsEnabled() then "case ($0) =>"
+            else "case () =>"
+          else if config.isCompletionSnippetsEnabled() then "($0)"
+          else "()"
+        ),
+        Nil,
+        range = Some(completionPos.toEditRange),
+        command = config.parameterHintsCommand().asScala,
```
tanishiking, line 159: What is this for?
It looks like this setting triggers parameter hint command https://scalameta.org/metals/docs/integrations/new-editor/#compileroptionsparameterhintscommand but, as far as I know, case/match completion moves the cursor to the position where parameter hint is not available (e.g. c`ase Some(value) => @@`)

Are there any examples where this command is useful?

tanishiking (commented):
I left some comments, but it's almost LGTM!



> I have some ideas here to remove searching through scope members and use shortened names, so should I do that it this PR, or here add only extractor objects and do the rest in a follow-up?

It would be really helpful if you can do that in another PR :+1: Can you open a follow-up issue?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
jkciesluk: > What is this for?
It looks like this setting triggers parameter hint command https://scalameta.org/metals/docs/integrations/new-editor/#compileroptionsparameterhintscommand but, as far as I know, case/match completion moves the cursor to the position where parameter hint is not available (e.g. case Some(value) => @@)

In this case we move cursor to `case (@@) =>`

>Are there any examples where this command is useful?

Yes, situations like `List(1 -> 2).map{ cas@@ }` (tests `brace` and `brace-function2` in `CompletionCaseSuite`

tests/cross/src/test/scala/tests/pc/CompletionCaseSuite.scala
jkciesluk: It was easier to fix on our side, by not going through subclasses and scope symbols for tuple types

tanishiking (commented):
I left a comment about `sealedStrictDesc`, but otherwise looks good to me!

mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
```diff
@@ -156,6 +159,71 @@ object MtagsEnrichments extends CommonMtagsEnrichments:
           )
         else false
       }
+    // Copied from dotty.tools.dotc.core.SymDenotations because it was not available for 3.0.0
```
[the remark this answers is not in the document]
jkciesluk: Great, I also think that's better than maintaining `sealedStrictDesc`. So should we just ignore tests for scala 3.0 or "adjust" the result to only scope symbols?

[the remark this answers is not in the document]
tanishiking: It would be nice to adjust the tests so that we can easily understand Metals doesn't completely support case completion by example, but it's optional, if it's too complicated, we can just ignore for Scala3 :)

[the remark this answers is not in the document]
jkciesluk: Or maybe, for 3.0.0 we can use `sym.children`? This would also provide ehxaustive match completions, just with more general cases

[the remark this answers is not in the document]
tgodzik: We already ignore the tests for Scala 3.0, we should only make sure it compiles.

I think it makes sense to do it as @tanishiking suggests, let's just do Nil for Scala 3.0

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -865,4 +934,114 @@ class Completions(
...
+          Some((m.selector, parent))
+        case _ => None
+
+  end CasePatternExtractor
+
+  object TypedCasePatternExtractor:
+    def unapply(path: List[Tree]) =
+      path match
+        // case _: Som@@ =>
+        case (ident @ Ident(name)) :: Typed(
```
tgodzik, line 1025:
```suggestion
        case (ident @ Ident(name)) :: _ ::(c:CaseDef)::(m:Match)::parent::_
```
Could we actually accept anything like this?  Is Bind and Typed the only options? Because I think we could quite reduce the number of cases this way

In that case we could merge the 4 cases into two:
```
        case (ident @ Ident(name)) :: (_ : Typed | _: Bind) ::(c:CaseDef)::(m:Match)::parent::_
```
I think

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -865,4 +934,114 @@ class Completions(
...
+        // case _: @@ =>
+        case Typed(_, _) :: (c: CaseDef) :: (m: Match) :: parent :: _ =>
+          Some((m.selector, parent))
+        // case ab: @@ =>
+        case Bind(
+              _,
+              Typed(_, _),
+            ) :: (c: CaseDef) :: (m: Match) :: parent :: _ =>
+          Some((m.selector, parent))
+        // case ab: Som@@ =>
```
tgodzik, line 1039: This is a weird case, why would we want to support this?  Is it for infix types? Are we able sensible show suggestions in that case?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
```diff
@@ -369,6 +381,30 @@ class CompletionsProvider(
       case CompletionValue.Document(label, doc, desc) =>
         mkItem(label, doc, filterText = Some(desc))
       case CompletionValue.Autofill(value) => mkItem(ident, value)
+      case CompletionValue.CaseKeyword(
+            _,
+            label,
+            text,
+            additionalEdit,
+            range,
+            // filterText,
```
tgodzik, line 390:
```suggestion
```

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
```diff
@@ -369,6 +381,30 @@ class CompletionsProvider(
...
+            command,
+          ) =>
+        mkItem(
+          label,
+          text.getOrElse(label),
+          additionalEdits = additionalEdit,
+          range = range,
+          // filterText = filterText,
```
tgodzik, line 398:
```suggestion
```

tests/cross/src/test/scala/tests/pc/CompletionCaseSuite.scala
```diff
@@ -172,6 +170,12 @@ class CompletionCaseSuite extends BaseCompletionSuite {
     """|case scala.util.Left(value) =>
        |case Right(value) => scala.util
        |""".stripMargin,
+    compat = Map(
+      "3" ->
+        """|case Left(value) => scala.util
```
tgodzik, line 175:
```suggestion
        """|case scala.util.Left((value) => scala.util
```
Left has a conflicting name in scope, we can fix it later, but let's create an issue in that case

tests/cross/src/test/scala/tests/pc/CompletionKeywordSuite.scala
```diff
@@ -427,7 +383,7 @@ class CompletionKeywordSuite extends BaseCompletionSuite {
     // to avoid newMain annotation
     filter = str => !str.contains("newMain"),
   )
-
+  // couldnt' reproduce
```
tgodzik, line 386:
```suggestion
```
We shouldn't show anything here since `new` is not a start for any of the completions. You can do a TODO issue to fix it later

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -55,7 +68,9 @@ class CompletionMatchSuite extends BaseCompletionSuite {
 
   // Assert that Workday/Weekend symbols from previous test don't appear in result.
   checkEdit(
-    "stale2",
+    "stale2".tag(
+      IgnoreScala3
+    ), // needs fix for scala3 (adding `isStale` filter)
```
tgodzik, line 73:
```suggestion
    "stale2"
```
this should work since we merged the other PR

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -958,20 +958,10 @@ class CompletionSuite extends BaseCompletionSuite {
...
       "3" ->
-        """|Some scala
-           |Some[A](value: A): Some[A]
-           |SomeToExpr - scala.quoted.ToExpr
-           |SomeToExpr[T: Type: ToExpr]: SomeToExpr[T]
-           |SomeFromExpr - scala.quoted.FromExpr
-           |SomeFromExpr[T](using Type[T], FromExpr[T]): SomeFromExpr[T]
-           |""".stripMargin,
+        """|Some(value) scala
+           |None scala
```
tgodzik, line 963:
```suggestion
```

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -1015,8 +1005,8 @@ class CompletionSuite extends BaseCompletionSuite {
     topLines = Some(2),
     compat = Map(
       "3" ->
-        """|NoManifest scala.reflect
-           |NoClassDefFoundError java.lang
+        """|None scala
+           |Some(value) scala
```
tgodzik, line 1009:
```suggestion
```
Some shouldn't be suggested since we already wrote `No`

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -1040,9 +1030,8 @@ class CompletionSuite extends BaseCompletionSuite {
            |Set scala.collection.immutable
            |""".stripMargin,
       "3" ->
-        """|Seq scala.collection.immutable
-           |Set scala.collection.immutable
-           |Stream scala.collection.immutable
+        """|Some(value) scala
+           |None scala
```
tgodzik, line 1034:
```suggestion
```

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -1066,9 +1055,8 @@ class CompletionSuite extends BaseCompletionSuite {
            |Set scala.collection.immutable
            |""".stripMargin,
       "3" ->
-        """|Seq scala.collection.immutable
-           |Set scala.collection.immutable
-           |Stream scala.collection.immutable
+        """|Some[_] scala
+           |None scala
```
tgodzik, line 1059:
```suggestion
```

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -1114,6 +1099,10 @@ class CompletionSuite extends BaseCompletionSuite {
        |NoManifest scala.reflect
        |""".stripMargin,
     topLines = Option(3),
+    compat = Map(
+      "3" ->
+        ""
```
tgodzik, line 1104: Shouldn't we get `Number` from compiler completions here? Maybe in some cases we should also allow compler completions?

tests/cross/src/test/scala/tests/pc/CompletionWorkspaceSuite.scala
```diff
@@ -334,7 +334,7 @@ class CompletionWorkspaceSuite extends BaseCompletionSuite {
   )
 
   checkEdit(
-    "match-typed",
+    "match-typed".tag(IgnoreScala3),
```
tgodzik, line 337:
```suggestion
    "match-typed",
```
this shouldn't be ignored without a reason.

tgodzik (changes requested):
Thanks for the great work! I still have suggestions and comments.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -865,4 +934,114 @@ class Completions(
             else compareByRelevance(o1, o2)
       end compare
 
+  object MatchExtractor:
+    def unapply(path: List[Tree]) =
+      path match
+        // foo mat@@
+        case (sel @ Select(qualifier, name)) :: _
+            if "match".startsWith(name.toString()) && text.charAt(
```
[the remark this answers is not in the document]
jkciesluk: It was resolved here https://github.com/scalameta/metals/pull/4229#discussion_r944207169
When we type `.match` the tree changes from `Select` to `Match` and we are not show the match completions for `xxx.ma@@` because I think most users would see it as a bug

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -865,4 +934,114 @@ class Completions(
...
+              completionPos.start - 1
+            ) == ' ' =>
+          Some(qualifier)
+        // foo match @@
```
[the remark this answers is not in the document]
jkciesluk: What would be the expected behaviour ?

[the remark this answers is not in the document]
jkciesluk: If we would have cases it would fall into next case 
```scala
// foo ma@tch (no cases)
        case (m @ Match(
              _,
              CaseDef(Literal(Constant(null)), _, _) :: Nil,
            )) :: _ =>
```
Should we provide exhaustive match for match with cases? What would be the expected behaviour ?

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
jkciesluk: Maybe we can allow compiler completions when we have typed bind i.e. `_: @@` or `abc`: xxx@@`?

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
jkciesluk: Right, it's because we don't filter completions by prefix while creating them and leave it to the editor. I have an idea how to fix it

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
jkciesluk: Also, should we move all the tests like `case xxx: So@@` from `CompletionSuite` to sth like `CasePatternSuite`, at least for Scala3, since we don't have this functionality for Scala2? I was thinking about creating new cross tests suite for case pattern completions both typed and untyped

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -865,4 +934,114 @@ class Completions(
             else compareByRelevance(o1, o2)
       end compare
 
+  object MatchExtractor:
```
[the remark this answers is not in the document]
jkciesluk: Did you mean `MatchCaseCompletions`? How can I then provide them `completionPos`, `text` etc?

tests/cross/src/test/scala/tests/pc/CompletionWorkspaceSuite.scala
jkciesluk: So shouldn't we just add a comment that there are no completions for null instead of removing tag?

tests/cross/src/test/scala/tests/pc/CompletionKeywordSuite.scala
jkciesluk: I will add a TODO, because it's a weird case, since `new` is not shown in `path` and the tree looks the same as for `case @@`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
jkciesluk: I am not sure what is weird in this case? When doing `match` on `Option` `case ab: Som@@ =>` will suggest `Some[_]` and `case ab: @@ =>` `Some[_]` and `None`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
jkciesluk: Then we could't distinguish `case abc @ So@@` and `case abc: So@@` when for the first one we would like `Some(value)` and for the second one `Some[_]`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,409 @@
...
+            sym.info.dealias.typeSymbol,
+            sym.decodedName,
+            value,
+          )
+        case None =>
+          result += completionGenerator.toCompletionValue(
+            sym.info.dealias.typeSymbol,
+            sym.decodedName,
+            Nil,
+          )
```
riiswa, line 237:
```suggestion
      result += completionGenerator.toCompletionValue(
            sym.info.dealias.typeSymbol,
            sym.decodedName,
            autoImport.getOrElse(Nil),
          )
```

riiswa (commented):
Just some suggestions :)

jkciesluk:
> Just some suggestions :)

Thanks, it looks better this way :)

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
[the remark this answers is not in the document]
jkciesluk: Or we can move `CaseExtractor`, `CasePatternExtractor` and `TypedCasePatternExtractor` and leave only `MatchExtractor`, because they don't use anything besides `path`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,481 @@
...
+        tpe match
+          case AndType(tp1, tp2) => // for refined type like A with B with C
+            loop(tp1)
+            loop(tp2)
+          case t => parents += tpe.typeSymbol
+      loop(tpe.widen.bounds.hi)
+      parents.toList.flatMap { parent =>
+        // There is an issue in Dotty, `sealedStrictDescendants` ends in an exception for java enums. https://github.com/lampepfl/dotty/issues/15908
+        if parent.isAllOf(JavaEnumTrait) then parent.children
+        else MetalsSealedDesc.strictDesc(parent)
```
tanishiking, line 218: Can we rename `strictDesc` to `sealedStrictDescendants`?

mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -450,7 +451,7 @@ trait Completions { this: MetalsGlobal =>
         ArgCompletion(ident, apply, pos, text, completions)
       }
     }
-
+    // pprint.pprintln(latestEnclosingArg.take(3))
```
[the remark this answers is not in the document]
tanishiking: ping :bow:

```suggestion
```

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
```diff
@@ -307,5 +422,26 @@ class CompletionMatchSuite extends BaseCompletionSuite {
...
+                |object Main {
+                |  def testExhaustive[T <: TestA with TestB](test: T): Boolean =
+                |    test match
+                |\tcase Foo => $$0
+                |\tcase Bar =>
+                |\tcase Baz =>
+                |\tcase Goo =>
+                |
+                |}""".stripMargin
+    ),
```
tanishiking, line 445: Can you add a test for Scala3 enum exhaustive match?

tests/cross/src/test/scala/tests/pc/CompletionCaseSuite.scala
```diff
@@ -372,7 +372,11 @@ class CompletionCaseSuite extends BaseCompletionSuite {
       "2.13" ->
         """|case head :: next => scala.collection.immutable
            |case Nil => scala.collection.immutable
-           |""".stripMargin
+           |""".stripMargin,
+      "3" ->
+        """|case head :: next => scala.collection.immutable
+           |case Nil => scala.collection.immutable
+           |""".stripMargin,
```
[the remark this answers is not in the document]
tanishiking: ping :bow:

[nit]
Making Scala 2.12 as special case makes it simple

```scala
      """|case head :: next => scala.collection.immutable
         |case Nil => scala.collection.immutable
         |""".stripMargin,
    compat = Map(
      "2.12" ->
      """|case head :: tl => scala.collection.immutable
         |case Nil => scala.collection.immutable
         |""".stripMargin,
    ),
```

tests/cross/src/test/scala/tests/pc/CompletionCaseSuite.scala
tanishiking: Let's create a follow-up ticket and fix it in another PR :+1:

tanishiking (commented):
I left several minor comments, but otherwise LGTM! Thank you very much for the great work :tada:

tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala
jkciesluk: Sure, I forgot about that. Also, currently we are importing all enum cases separately, maybe for enums we should do sth like `case Color.Red` instead of `import Color.Red`?

tanishiking: You meant to ` case Color` instead of import Color.Red` ?
Yes, it would be better with that :) However, we can do that in follow-up PR :+1:

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,479 @@
...
+          else "()"
+        ),
+        Nil,
+        range = Some(completionPos.toEditRange),
+        command = config.parameterHintsCommand().asScala,
+      )
+    else
+      // Step 1: walk through scope members.
+      indexedContext.scopeSymbols
+        .foreach(s => visit(s.info.dealias.typeSymbol, s.decodedName, Nil))
```
dos65, line 149: As I got it was added as there were issues with some particular cases like for `Nil$` and `Nil` in `indexedContext.findSymbol`.

I started looking if it's possible to fix `IndexedContext` to avoid checking all symbols in scope.

jkciesluk: There was already a discussion about going through scope symbols here https://github.com/scalameta/metals/pull/4229#discussion_r944422269
It's not because of a bug, but for not sealed traits. Actually I already have a prepared improvement, by moving the `isValid` filter from `visit` to something like `indexedContext.scopeSymbols.filter(isValid).foreach(..)`, but I wanted to do it in a follow up to not change anything more in this pr

jkciesluk: Or maybe we can use `indexedContext.scopeSymbols` as a fallback when `sealedStrictDescendants` is empty?

dos65: @jkciesluk thx for pointing at discussion. yep - let's solve it after merging this one

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -0,0 +1,479 @@
...
+    patternOnly match
+      case None => true
+      case Some(query) => CompletionFuzzy.matches(query, name)
+
+  def toCompletionValue(
+      sym: Symbol,
+      name: String,
+      autoImports: List[l.TextEdit],
+  )(using Context): CompletionValue.CaseKeyword =
+    sym.info
```
dos65, line 312: unused?

dos65 (approved):
Great work!
Even if we have some issues, I think they aren't affect other existing completions and we able to improve match-case completions in other PR's.


Let's merge this one!

tgodzik (approved):
Some last comments from me, but nothing blocking I think (might have forgot about some comments)

tanishiking (approved):
LGTM from my side! it would be great if @jkciesluk creates tickets for the follow-up stuffs (it's not necessary if you gonna work on them right after this PR :) )

mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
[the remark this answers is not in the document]
tgodzik: Ach, it's done. I had a very old comment :sweat_smile:

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: ach, makes sense

tests/cross/src/test/scala/tests/pc/CompletionKeywordSuite.scala
tgodzik: Might need the CURSOR trick

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
tgodzik: sure, makes sense to move out the tests

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -821,6 +891,16 @@ class Completions(
 
       override def compare(o1: CompletionValue, o2: CompletionValue): Int =
         (o1, o2) match
+          case (
+                sym1: CompletionValue.CaseKeyword,
```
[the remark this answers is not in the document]
jkciesluk: I think we came to the opposite conclusion 😅  Without compiler completions we won't ever see e.g. extractor objects
For now we show compiler completions only in pattern only completions

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
```diff
@@ -230,6 +230,10 @@ class CompletionsProvider(
       if config.isCompletionSnippetsEnabled then
         item.setInsertTextFormat(InsertTextFormat.Snippet)
 
+      command.foreach { command =>
+        item.setCommand(new Command("", command))
```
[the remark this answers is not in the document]
tgodzik: Ach, that's the same in Scala2.

tgodzik:
> LGTM from my side! it would be great if @jkciesluk creates tickets for the follow-up stuffs (it's not necessary if you gonna work on them right after this PR :) )

Yep, makes sense.

jkciesluk:
> LGTM from my side! it would be great if @jkciesluk creates tickets for the follow-up stuffs (it's not necessary if you gonna work on them right after this PR :) )

Thanks!
I already have few fixes ready and will create PRs right after merging this one. For the rest I will create issues

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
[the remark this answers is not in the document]
tgodzik: ach, that's ok then. Got lost a bit.

Issue #4228, opened 2022-08-05
Deleting worksheet with compiler error does not clear


matthughes:
### Describe the bug

Given a worksheet with a compiler error, delete the offending worksheet.  Metals/VsCode continue to report errors in this non-existent file.  The only work around I've found is to stop metals/blow away .bloop/.metals directory and reimport.

### Expected behavior

Deleted files should not report compiler errors.

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.7

### Extra context or search terms

worksheet error

PR #4233, opened 2022-08-08
bugfix: Remove worksheet diagnostics on close

tgodzik:
Fixes https://github.com/scalameta/metals/issues/4228

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
@@ -98,16 +98,20 @@ final class Diagnostics(
         syntaxError(path) = diagnostic
         publishDiagnostics(path)
       case _ =>
-        onNoSyntaxError(path)
+        onClose(path)
     }
   }
 
-  def onNoSyntaxError(path: AbsolutePath): Unit = {
-    syntaxError.remove(path) match {
-      case Some(_) =>
+  def onClose(path: AbsolutePath): Unit = {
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
```diff
@@ -98,16 +98,20 @@ final class Diagnostics(
...
-      case Some(_) =>
+  def onClose(path: AbsolutePath): Unit = {
+    val diags = if (path.isWorksheet) {
+      diagnostics.remove(path).toList.flatMap(_.asScala) ++
+        syntaxError.remove(path)
+    } else syntaxError.remove(path).toList
+    diags match {
+      case Nil =>
+        () // Do nothing, there was no previous error.
+      case _ =>
```
kpodsiad, line 113: `onClose` called from `onSyntaxError` seems a bit strange. What do you think about:
- adding some private method called e.g. `clearDiagostiscFor`
- `onClose` and `case _  =>` in `onSyntaxError` can call `clearDiagostiscFor`
- `clearDiagostiscFor` will have scaladoc which will explain where it's used

kpodsiad (approved):

dos65 (approved):

metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
tgodzik: Och, forgot about the comment. I will change that in another PR

PR #4234, opened 2022-08-08
fix: type completions additonal fixes

dos65:
- fixes completion for term symbol that aren't method
- allow object at type completion pos

 should include a fix for #4224

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -64,99 +64,105 @@ class Completions(
     case (_: Ident) :: (_: SeqLiteral) :: _ => false
     case _ => true
 
-  /**
-   * It shows the conditions of the code in the cursor position, with
-   * respect to assessing
-   *  1. whether it makes semantic sense to complete
-   *  the position with symbols that are `class`, `trait`,
-   *  `abstract`, `object`, or method
-   *  1. whether it would make syntactic sense to add `[]` `()` or `{}`
-   *  suffixes to the symbols suggested for completion.
-   *
-   *  Note: In the comments `@@` represents the cursor position.
-   *
-   * @param isTypePosition if this is the position for defining a type such as
-   *                       `val a: In@@` or `def i: scala.IndexedSe@@`
-   * @param isNewPosition if it is the position for instantiation after
-   *                      the new keword, as in `new Fil@@`
-   * @param isInstantiationOrMethodCallPos if it is the position for method call or
-   *                                       instantiation as in `object A{ MyCaseClas@@ }`
-   * @param noSquareBracketExists if there is already a square bracket in the position as in
-   *                              `val a: IndexedMa@@[]`
-   */
-  private case class CursorPositionCondition(
-      isTypePosition: Boolean,
-      isNewPosition: Boolean,
-      isInstantiationOrMethodCallPos: Boolean,
-      noSquareBracketExists: Boolean,
-  ):
+  enum CursorPos:
+    case Type(hasTypeParams: Boolean, hasNewKw: Boolean)
+    case Term
+    case Import
 
-    def isObjectValidForPos = (!isTypePosition && !isNewPosition)
+    def include(sym: Symbol)(using Context): Boolean =
+      val generalExclude =
+        isUninterestingSymbol(sym) ||
+          !isNotLocalForwardReference(sym) ||
+          sym.isPackageObject
 
-    /**
-     * classes and traits are type symbols. They are not suitable for
-     *        instantiation or method call positions. Only apply methods and objects
-     *        can be used in such positions.
-     */
-    def isClassOrTraitValidForPos = !isInstantiationOrMethodCallPos
-    def isMethodValidForPos = !isTypePosition
+      if generalExclude then false
+      else
+        this match
+          case Type(_, _) if sym.isType => true
+          case Type(_, _) if sym.isTerm =>
+            /* Type might be referenced by a path over an object:
+             * ```
+             * object sample:
+             *    class X
+             * val a: samp@@le.X = ???
+             * ```
+             * ignore objects that has companion class
+             *
+             * Also ignore objects that doesn't have any members.
+             * By some reason traits might have a fake companion object(example: scala.sys.process.ProcessBuilderImpl)
+             */
+            val allowModule =
+              sym.is(Module) &&
+                (sym.companionClass == NoSymbol && sym.info.allMembers.nonEmpty)
+            allowModule
+          case Term if sym.isTerm || sym.is(Package) => true
+          case Import => true
+          case _ => false
+      end if
+    end include
 
-    def isParanethesisValidForPos = !isTypePosition
-    def isSquareBracketsValidForPos = noSquareBracketExists
-      && (isTypePosition || isNewPosition || isInstantiationOrMethodCallPos)
+    def allowBracketSuffix: Boolean =
+      this match
+        case Type(hasTypeParams, _) => !hasTypeParams
+        case _ => false
 
-    /**
-     * as in ```new MyTrait@@```
-     */
-    def isCurlyBracesValidForPos = isNewPosition
-  end CursorPositionCondition
+    def allowTemplateSuffix: Boolean =
+      this match
+        case Type(_, hasNewKw) => hasNewKw
+        case _ => false
 
-  private lazy val cursorPositionCondition =
+  end CursorPos
+
+  private lazy val cursorPos =
     calculateTypeInstanceAndNewPositions(path)
 
   private def calculateTypeInstanceAndNewPositions(
       path: List[Tree]
-  ): CursorPositionCondition =
+  ): CursorPos =
     path match
+      case (_: Import) :: _ => CursorPos.Import
+      case _ :: (_: Import) :: _ => CursorPos.Import
       case (head: (Select | Ident)) :: tail =>
         // https://github.com/lampepfl/dotty/issues/15750
         // due to this issue in dotty, because of which trees after typer lose information,
         // we have to calculate hasNoSquareBracket manually:
-        val hasNoSquareBracket =
+        val hasSquareBracket =
           val span: Span = head.srcPos.span
           if span.exists then
             var i = span.end
             while i < (text.length() - 1) && text(i).isWhitespace do i = i + 1
 
-            if (i < text.length()) then text(i) != '['
-            else true
+            if (i < text.length()) then text(i) == '['
+            else false
           else false
+
+        def typePos = CursorPos.Type(hasSquareBracket, hasNewKw = false)
+        def newTypePos =
+          CursorPos.Type(hasSquareBracket, hasNewKw = true)
+
         tail match
-          case (v: ValOrDefDef) :: _ =>
-            if v.tpt.sourcePos.contains(pos) then
-              CursorPositionCondition(true, false, false, hasNoSquareBracket)
-            else
-              CursorPositionCondition(false, false, false, hasNoSquareBracket)
-          case New(selectOrIdent: (Select | Ident)) :: _ =>
-            if selectOrIdent.sourcePos.contains(pos) then
-              CursorPositionCondition(false, true, false, hasNoSquareBracket)
-            else
-              CursorPositionCondition(false, false, false, hasNoSquareBracket)
-          case (a @ AppliedTypeTree(_, args)) :: _ =>
-            if args.exists(_.sourcePos.contains(pos)) then
-              CursorPositionCondition(true, false, false, hasNoSquareBracket)
-            else CursorPositionCondition(false, false, false, false)
-          case (_: Import) :: _ =>
-            CursorPositionCondition(false, false, false, hasNoSquareBracket)
+          case (v: ValOrDefDef) :: _ if v.tpt.sourcePos.contains(pos) =>
+            typePos
+          case New(selectOrIdent: (Select | Ident)) :: _
+              if selectOrIdent.sourcePos.contains(pos) =>
+            newTypePos
+          case (a @ AppliedTypeTree(_, args)) :: _
+              if args.exists(_.sourcePos.contains(pos)) =>
+            typePos
+          case (Template(constr, parents, self, _)) :: _
+              if (constr :: self :: parents).exists(
+                _.sourcePos.contains(pos)
+              ) =>
+            typePos
           case _ =>
-            CursorPositionCondition(false, false, true, hasNoSquareBracket)
+            CursorPos.Term
         end match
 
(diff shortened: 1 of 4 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -64,99 +64,105 @@ class Completions(
...
-     * as in ```new MyTrait@@```
-     */
-    def isCurlyBracesValidForPos = isNewPosition
-  end CursorPositionCondition
-
-  private lazy val cursorPositionCondition =
+  enum CursorPos:
+    case Type(hasTypeParams: Boolean, hasNewKw: Boolean)
+    case Term
+    case Import
```
dos65, line 70: @vzmerr I'm sorry that I missed that thing in your initial PR.
In general this was correct but I reworked it a bit at to operate over a rule that we have 3 types of completions.
It's either a type or term or import.
So, having it we can safely filter out symbols: non type at TypePos, non Term at TermPos. For Import it's ok to reference Term and Type.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -547,25 +538,6 @@ class Completions(
...
-
-    private def isNotAMethodOrMethodIsValidForPos(sym: Symbol) =
-      cursorPositionCondition.isMethodValidForPos || !sym.is(
-        Flags.Method
-      ) // !sym.info.typeSymbol.is(Flags.Method) does not detect Java methods
-
-    private def isNotClassOrTraitOrTheyAreValidForPos(sym: Symbol) =
-      cursorPositionCondition.isClassOrTraitValidForPos ||
-        sym.is(Flags.Method) ||
-        !sym.isClass && !sym.info.typeSymbol.is(Trait)
```
dos65, line 567: This was not correct because symbol might be at least also a ValDef.
Also, my suggestion of checking if it's a trait over a `.info` was wrong :facepalm: 
```scala
def foo(param: java.util.List[A]): Unit =
  para@@ // <- here param is not a Method
        // <- also `sym.info.typeSymbol.is(Trait) == true` here because the of param is `java.util.List` 
```

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: Thanks for working on this. May I know the cases where we had the regressions?

tests/cross/src/test/scala/tests/pc/CompletionWorkspaceSuite.scala
```diff
@@ -344,21 +344,11 @@ class CompletionWorkspaceSuite extends BaseCompletionSuite {
     """|import java.util.ArrayDeque
        |object Main {
        |  def foo(): Unit = null match {
-       |    case x: ArrayDeque[$0] =>
```
vzmerr, line 347: ArrayDeque takes type parameters. doesn't it? So why is this removed?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: This method is only excluding class or trait where they are not valid for pos. It does not affect `ValDef`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
dos65: These two:
https://github.com/scalameta/metals/pull/4234/files#diff-9ab9c5f7b86393976b3619ea86ddef73737390ca43e2a3d8ce27d66aa1f576e7R1534-R1570

tests/cross/src/test/scala/tests/pc/CompletionWorkspaceSuite.scala
vzmerr: The current one doesn't compile. It should have type parameters.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
vzmerr: My current PR on object completion also brings substantial change, based on the existing code. Now, if we use this reworked code, how can we work that out?

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -1522,4 +1531,42 @@ class CompletionSuite extends BaseCompletionSuite {
...
+       |""".stripMargin,
+    compat = Map(
+      "2" ->
+        """|first: List[Int]
+           |""".stripMargin
+    ),
+  )
+
+  check(
+    "object-at-type-pos",
```
vzmerr, line 1555: this one passes on the other PR's branch

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
```diff
@@ -1522,4 +1531,42 @@ class CompletionSuite extends BaseCompletionSuite {
     ),
   )
 
+  check(
+    "class-members-trait-issue",
```
vzmerr, line 1535: this one is failing on the other PR's branch. Can we solve it in another way without a total rework, since now on this  branch we have the regression of not adding type parameter square brackets in here.

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
vzmerr: passing on the new PR I created for allowing objects at type pos.

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
vzmerr: perhaps, we can just remove the filter for `isNotClassOrTraitOrTheyAreValidForPos` for now to fix this regression.

vzmerr: by simply adding `|| sym.isTerm` to `isNotClassOrTraitOrTheyAreValidForPos` the regression would be solved.

tests/cross/src/test/scala/tests/pc/CompletionWorkspaceSuite.scala
dos65: I would turn into an issue after merge

I don't know why it's completed with `[]` in the main branch. 
There is no handler that marks this position by `isTypePosition` for the tree in main  or this case here: 
https://github.com/scalameta/metals/blob/c328f0925e9692b493aba395492ae03fd36da424/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala#L121-L152
While the tree is: `Ident(ArrayDeque), Typed(Ident(_),Ident(ArrayDeque)), Bind(x,Typed(Ident(_),Ident(ArrayDeque))), CaseDef(Bind(x,Typed(Ident(_),Ident(ArrayDeque)))`

vzmerr: Perhaps, this would lead to not adding square brackets also in other positions. Also, what about the other PR with substantial changes that would not be usable after this merge? Also, this is at the last minute, what if it introduces other problems? The regression can be solved by simply adding `|| sym.isTerm` in `isNotClassOrTraitOrTheyAreValidForPos` 
What if the changes of the other PR cannot be reworked on this one?

tanishiking:
Thank you very much @dos65 @vzmerr !
I'm considering maybe we should revert type completions PRs for now, and retry introducing the feature in the next release (0.11.9?).

- https://github.com/scalameta/metals/pull/4174
- https://github.com/scalameta/metals/pull/4174

As @vzmerr mentioned

> Also, this is at the last minute, what if it introduces other problems? The regression can be solved by simply adding || sym.isTerm in isNotClassOrTraitOrTheyAreValidForPos

I'm worried about introducing another problem by this PR (even if adding `|| sym.isTerm` in `isNotClassOrTraitOrTheyAreValidForPos` thing solves this issue) or maybe there's another issue related to type completion. I prefer to skip introducing type completion for 0.11.8 and rework for 0.11.9
What do you think about it? @dos65 @vzmerr FYI @tgodzik @ckipp01

tests/cross/src/test/scala/tests/pc/CompletionWorkspaceSuite.scala
dos65: yep, `isTerm` fixes this case but we also need to deal with objects.
I've checked that tests doesn't pass in your PR.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tgodzik: Looks like it's explained by @dos65  : `sym.info.typeSymbol.is(Trait) == true`

tests/cross/src/test/scala/tests/pc/CompletionSuite.scala
tgodzik: I think this refactoring is fine and might be more maintainable.

tgodzik (commented):
> Thank you very much @dos65 @vzmerr ! I'm considering maybe we should revert type completions PRs for now, and retry introducing the feature in the next release (0.11.9?).
> 
> * [Scala 3 type completion #4174](https://github.com/scalameta/metals/pull/4174)
> * [bugfix: Show package completions #4223](https://github.com/scalameta/metals/pull/4223)
> 
> As @vzmerr mentioned
> 
> > Also, this is at the last minute, what if it introduces other problems? The regression can be solved by simply adding || sym.isTerm in isNotClassOrTraitOrTheyAreValidForPos
> 
> I'm worried about introducing another problem by this PR (even if adding `|| sym.isTerm` in `isNotClassOrTraitOrTheyAreValidForPos` thing solves this issue) or maybe there's another issue related to type completion. I feel it would be safer if we skip introducing type completion for 0.11.8 and rework for 0.11.9 What do you think about it? @dos65 @vzmerr FYI @tgodzik @ckipp01

That would be safest to do probably right now. We could revert the 2 PRs then squash the commits from those two PRs with this one and continue working on this in a new PR that would also handle cases @vzmerr worked on.

@tanishiking could you create the revert and modify the release notes? We would go ahead with the release then to unblock all the waiting PRs.

dos65:
> I'm considering maybe we should revert type completions PRs for now, and retry introducing the feature in the next release (0.11.9?).

@tanishiking Yep, good idea. It seems that we need an additional time to cover all cases (like patterns in case def but I think there should be more) properly and test them before release.

tanishiking:
Thank you, everyone. Let's revert those PRs for now and go on the release.

vzmerr:
> Thank you very much @dos65 @vzmerr ! I'm considering maybe we should revert type completions PRs for now, and retry introducing the feature in the next release (0.11.9?).
> 
> * [Scala 3 type completion #4174](https://github.com/scalameta/metals/pull/4174)
> * [bugfix: Show package completions #4223](https://github.com/scalameta/metals/pull/4223)
> 
> As @vzmerr mentioned
> 
> > Also, this is at the last minute, what if it introduces other problems? The regression can be solved by simply adding || sym.isTerm in isNotClassOrTraitOrTheyAreValidForPos
> 
> I'm worried about introducing another problem by this PR (even if adding `|| sym.isTerm` in `isNotClassOrTraitOrTheyAreValidForPos` thing solves this issue) or maybe there's another issue related to type completion. I feel it would be safer if we skip introducing type completion for 0.11.8 and rework for 0.11.9 What do you think about it? @dos65 @vzmerr FYI @tgodzik @ckipp01

Yes, I think this is the best route. There will be enough time to see if also my new PR can be reworked on this one; and whether this new approach can catch and cover all the cases.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -64,99 +64,105 @@ class Completions(
...
+            CursorPos.Term
         end match
 
       case (_: TypeTree) :: TypeApply(Select(newQualifier: New, _), _) :: _
           if newQualifier.sourcePos.contains(pos) =>
-        CursorPositionCondition(false, true, false, true)
+        CursorPos.Type(hasTypeParams = false, hasNewKw = true)
 
-      case _ => CursorPositionCondition(false, false, false, true)
+      case _ => CursorPos.Term
```
vzmerr, line 165: why is this TermPos? It does not differentiate between `CursorPositionCondition(false, false, true, hasNoSquareBracket)` and this one

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -64,99 +64,105 @@ class Completions(
...
-              CursorPositionCondition(true, false, false, hasNoSquareBracket)
-            else
-              CursorPositionCondition(false, false, false, hasNoSquareBracket)
-          case New(selectOrIdent: (Select | Ident)) :: _ =>
-            if selectOrIdent.sourcePos.contains(pos) then
-              CursorPositionCondition(false, true, false, hasNoSquareBracket)
-            else
-              CursorPositionCondition(false, false, false, hasNoSquareBracket)
-          case (a @ AppliedTypeTree(_, args)) :: _ =>
-            if args.exists(_.sourcePos.contains(pos)) then
```
vzmerr, line 146: in here, it was intentional that the `if` guard was no put in the case. It creates a different logic when you add the guard to case. The `else` is different with the last `_` case that we have.

vzmerr: I think the tweaking of logic in here, is the cause of not inserting square bracket in the test that is not passing.
`pprints` would be showing that.

tanishiking (approved):
What is the status of this PR? I took a look and its LGTM 💯 

We talked with @dos65 in person, and we may want to re-work on type of completion stuff, but we can work on it in another PR (as this PR already fixed some regressions 👍 ).

### what kind of re-work we talked about?
Currently, type completion filters out "non-compilable completions" such as `Foo` (where `Foo` should have type parameters list `Foo[Int]`).

However, it would be nice if we provide both `Foo` (with lower priority) and `Foo[$0]` (with higher priority) instead of filtering out `Foo` (without brackets). Providing both of them makes Metals easier to introduce unexpected regressions.

---

(side note) My thoughts on completions in IDE

I believe completing just `Foo` is still fine as I don't expect completions to be compilable out of the box. Completions provide some help to developers, they don't need to be "semantically correct".

For example,

```scala
def foo(x: Int) = ???
val aaa: Int = 1
val abb: String = "1"

foo(a@@) // should we filter out `abb` from completion items
// because `abb`'s return type is String
// whereas `foo` should receive Int parameters?

// I don't think so, maybe users want to type `abb.toInt` after completing `abb`
```

Issue #4194, opened 2022-07-27
Go To Definition and Go To Implementation not working properly


vzmerr:
### Describe the bug

In the following code in Scala 3 Completions file, we have the following problem on the `toString` method of `Symbol.Symbol`

https://user-images.githubusercontent.com/109077949/181247893-630530b5-cd5a-422b-9e01-901cf76ddb93.mov

:

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

0.11.7+61-a6299f3a-SNAPSHOT

tgodzik:
Looks like it's the same issue as was fixed in https://github.com/lampepfl/dotty/pull/15536

We can confirm when we update metals to 3.2.0

vzmerr:
One extra case to check
![Go To def](https://user-images.githubusercontent.com/109077949/182612839-54d9d43d-ca39-4f0d-9ce2-75b034b93fff.gif)

tgodzik:
> One extra case to check

That is unrelated, but it does work for me. It needs the code to compile though.

vzmerr:
> > One extra case to check
> 
> That is unrelated, but it does work for me. It needs the code to compile though.

I ran the tests, so it is compiled. I wonder why it is not working on my side. 

By the way, I guess it would be cool if we add a trouble shoot tutorial in the docs.

vzmerr:
> Where did you run the tests though? 
sbt

tgodzik:
> > Where did you run the tests though?
> 
> sbt

Yeah, sbt compilation is unrelated unless you use sbt server. If you use Bloop it probably didn't compile. Maybe something is shown in the problems view?

vzmerr:
![Problem View Ammonite](https://user-images.githubusercontent.com/109077949/182635850-b5092149-2ca9-44f1-95d9-5074e03ff75e.gif)
 Yes, 66 problems

vzmerr:
the problem might be Bloop
<img width="749" alt="Screenshot 2022-08-03 at 16 42 15" src="https://user-images.githubusercontent.com/109077949/182637083-a47a856d-be2a-4bf9-b88f-be86d9b0ab17.png">

vzmerr:
> Try Clean compiler workspace command


after running `sbt clean` followed by `sbt compile`, the problems number  increased to 227.  

Same with `Clean Java Language Server Workspace` command.

dos65:
@vzmerr this clean command is located in Metals ta
[...]

PR #4235, opened 2022-08-09
docs: Move FAQ section to the toplevel to sidebar

tanishiking:
This PR just moved the link to `Frequently Asked Question` in the sidebar to the top level.
I believe the link to the FAQ shouldn't be hidden under the drop-down menu (`TroubleShooting`).

What do you think about it?

### before

Screen Shot 2022-08-09 at 17 39 25

### after

Screen Shot 2022-08-09 at 17 38 16

## Context
I saw @vzmerr think about writing a troubleshooting guide (I think it's a good idea 👍 )
https://github.com/scalameta/metals/issues/4194#issuecomment-1204146318

However, I wonder how many users notice a FAQ section in the doc because it's default hidden in the `TroubleShooting` section. (Not sure because we didn't do any access analysis using GA or something).

### DIFF
--- a/website/sidebars.json
+++ b/website/sidebars.json
@@ -1,43 +1,45 @@
 {
-  "docs": {
-    "Text Editors": [
-      "editors/overview",
-      "editors/vscode",
-      "editors/vim",
-      "editors/sublime",
-      "editors/emacs",
-      "editors/online-ides",
-      "editors/user-configuration"
-    ],
-    "Code Actions": [
-      "codeactions/codeactions"
-    ],
-    "Build Tools": [
-      "build-tools/overview",
-      "build-tools/bloop",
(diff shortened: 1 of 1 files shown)


tgodzik (changes requested):
Might be good to have it more visible, but in that case I would also move proxy and mirrors to toplevel.

tgodzik:
@tanishiking what do you want to do about this PR?

tanishiking:
@tgodzik Thanks for the ping!
Maybe what we need to do is categorizing `Frequently Asked Question` like `Proxy and Mirror`. (Because I'm not sure current `FAQ` is really frequently asked questions, they looks like just ways to trouble shoots).

I will open an issue with the proposal for re-structuring the document. For this PR, let me close for now.

Issue #4243, opened 2022-08-10
ArrayIndexOutOfBoundsException on request textDocument/foldingRange


szymon-rd:
### Describe the bug

The following error occurred when importing a small scala-cli script. I was unable to reproduce it with the same source code.
```
Error - 5:15:50 PM] Request textDocument/foldingRange failed.
  Message: Internal error.
  Code: -32603 
java.util.concurrent.CompletionException: java.lang.ArrayIndexOutOfBoundsException: 201
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:308)
	at java.util.concurrent.CompletableFuture.uniAccept(CompletableFuture.java:661)
	at java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:646)
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:488)
	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1990)
	at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1(CancelTokens.scala:40)
	at scala.meta.internal.metals.CancelTokens$.$anonfun$future$1$adapted(CancelTokens.scala:38)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:484)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
Caused by: java.lang.ArrayIndexOutOfBoundsException: 201
	at scala.meta.internal.parsing.FoldingRangeExtractor$Foldable$.isScala3BlockWithoutOptionalBraces(FoldingRangeExtractor.scala:248)
	at scala.meta.internal.parsing.FoldingRangeExtractor$Foldable$.unapply(FoldingRangeExtractor.scala:199)
	at scala.meta.internal.parsing.FoldingRangeExtractor.extractFrom(FoldingRangeExtractor.scala:36)
	at scala.meta.internal.parsing.FoldingRangeExtractor.$anonfun$extractFrom$2(FoldingRangeExtractor.scala:56)
[...]

PR #4245, opened 2022-08-10
bugfix: Don't throw exception if template ends at EOF

tgodzik:
Previously, if a template was ended by EOF we would throw IndexOutOfBounds exception, however with optional braces it can end at EOF token, which doesn't correspond to any char. Now, we added a size check for that case.

Fixes https://github.com/scalameta/metals/issues/4243

We could try and adjust end pos to previous character, but that's also not that great since `)` doesn't really end a block


tanishiking (approved):
LGTM, nice.catch!

Issue #4244, opened 2022-08-10
ArgCompletions not working for nested apply


jkciesluk:
### Describe the bug

ArgCompletions for scala 2 are not showing for nested apply. Example:

```scala
def method(a: Int, b: Int): Int = ???

method(method(@@), ) // named argument and autofill completions aren't showing
method(method(), @@) // completions showing correctly
```

### Expected behavior

See named argument completions in both cases.

---

I think the problem is with the `typedTreeAt` 
https://github.com/scalameta/metals/blob/46db7773f0d61c7bc4ea26292ec356209c124862/mtags/src/main/scala-2/scala/meta/internal/pc/completions/ArgCompletions.scala#L21-L24

It doesn't type inner apply, unless we fill more arguments (working correctly for `method(method(1, @@), 1)`.
When working correctly, typed tree looks like `Select(..., name = method)`, otherwise `Ident(name = method)`, which results in `method.tpe == null`.

Possible solution might be to search for `methodSym` in scope symbols using `name`, when `method.tpe == null`

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.8

### Extra context or search terms

completions namedArguments typedTreeAt

PR #4247, opened 2022-08-11
Fix ArgCompletions for nested apply

jkciesluk:
Sometimes typedTreeAt can't insert type correctly, now as a fallback we search for the symbol in metalsScopeMembers.
Fixes https://github.com/scalameta/metals/issues/4244

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/ArgCompletions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/ArgCompletions.scala
@@ -18,11 +18,23 @@ trait ArgCompletions { this: MetalsGlobal =>
   ) extends CompletionPosition {
     val editRange: l.Range =
       pos.withStart(ident.pos.start).withEnd(pos.start).toLSP
-    val method: Tree = typedTreeAt(apply.fun.pos)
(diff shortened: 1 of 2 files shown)


dos65 (commented):
@jkciesluk Thx for the fix!

There is also another method for locating tree: `MetalsGlobal.locateTree`.
Have you tried it?
There is a chance that it might work correct in this case. Or we can fix it because it's defined in metals codebase.

jkciesluk:
@dos65 I've tried it now and it doesn't work, even for not nested apply 😞 , which is weird, because it works the same as `locateTree` used in `typedTreeAt`. Maybe it's because of the `typeCheck(unit)` in `typedTreeAt`, which we don't run with `MetalsGlobal.locateTree` ?

dos65:
@jkciesluk I think in other places we do typecheck and then call locate

jkciesluk:
@dos65 Right, I've missed that. `MetalsGlobal.locateTree` works normally, but not for nested apply. It's probably an issue with typer, like @tgodzik said.

dos65 (approved):
Thx for double-checking!
:+1:

tgodzik (approved):
LGTM!

PR #4248, opened 2022-08-11
Add autofill named arguments completion for scala3

jkciesluk:
Adds autofill named arguments with default values completion and completions for single named arguments, the same way we have for scala2.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
@@ -62,7 +62,7 @@ sealed trait CompletionValue:
         s"${printer.completionSymbol(ext.symbol)} (extension)"
       case so: CompletionValue.Symbolic =>
         printer.completionSymbol(so.symbol)
-      case CompletionValue.NamedArg(label, tpe, _) =>
+      case CompletionValue.NamedArg(label, tpe) =>
         printer.tpe(tpe)
       case CompletionValue.Document(_, _, desc) => desc
       case _ => ""
@@ -121,9 +121,11 @@ object CompletionValue:
   case class NamedArg(
       label: String,
       tpe: Type,
-      override val insertText: Option[String],
   ) extends CompletionValue
-  case class Autofill (label: String, value: String) extends CompletionValue
+  case class Autofill(
+      value: String
+  ) extends CompletionValue:
+    override def label: String = "Autofill with default values"
   case class Keyword(label: String, override val insertText: Option[String])
       extends CompletionValue
 
@@ -147,10 +149,10 @@ object CompletionValue:
   case class Document(label: String, doc: String, description: String)
       extends CompletionValue
 
-  def namedArg(label: String, sym: Symbol, insertText: Option[String] = None)(
-      using Context
+  def namedArg(label: String, sym: Symbol)(using
+      Context
   ): CompletionValue =
-    NamedArg(label, sym.info.widenTermRefExpr, insertText)
+    NamedArg(label, sym.info.widenTermRefExpr)
 
   def keyword(label: String, insertText: String): CompletionValue =
     Keyword(label, Some(insertText))
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
@@ -32,6 +32,7 @@ sealed trait CompletionValue:
       case _: CompletionValue.NamedArg => CompletionItemKind.Field
       case _: CompletionValue.Override => CompletionItemKind.Method
       case _: CompletionValue.Extension => CompletionItemKind.Method
+      case _: CompletionValue.Autofill => CompletionItemKind.Enum
       case v: (CompletionValue.Compiler | CompletionValue.Workspace |
             CompletionValue.Scope | CompletionValue.Interpolator) =>
         val symbol = v.symbol
@@ -62,8 +63,7 @@ sealed trait CompletionValue:
       case so: CompletionValue.Symbolic =>
         printer.completionSymbol(so.symbol)
       case CompletionValue.NamedArg(label, tpe, _) =>
-        if label.startsWith("Autofill") then ""
-        else printer.tpe(tpe)
+        printer.tpe(tpe)
       case CompletionValue.Document(_, _, desc) => desc
       case _ => ""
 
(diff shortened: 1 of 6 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
```diff
@@ -61,8 +61,9 @@ sealed trait CompletionValue:
         s"${printer.completionSymbol(ext.symbol)} (extension)"
       case so: CompletionValue.Symbolic =>
         printer.completionSymbol(so.symbol)
-      case CompletionValue.NamedArg(_, tpe) =>
-        printer.tpe(tpe)
+      case CompletionValue.NamedArg(label, tpe, _) =>
+        if label.startsWith("Autofill") then ""
```
tanishiking, line 65: I prefer to have another `CompletionValue` subclass for the autofill-named-arguments feature.

This feature is kind of `NamedArg`, but the content is quite different from `NamedArg`, so I think there's no point representing this completion as `NamedArg` that could be error-prone in the future.

What do you think about this?

tests/cross/src/test/scala/tests/pc/CompletionArgSuite.scala
```diff
@@ -258,6 +258,11 @@ class CompletionArgSuite extends BaseCompletionSuite {
        |$$foo = ${1:???}, | = ${2:???}
        |""".stripMargin,
     topLines = Option(2),
+    compat = Map(
+      "3" -> """|$$foo = 
+                |$$foo = ${1:???}, $$bar = ${2:???}
```
tanishiking, line 263: :tada:

mtags/src/main/scala-3/scala/meta/internal/pc/completions/NamedArgCompletions.scala
```diff
@@ -104,12 +116,70 @@ object NamedArgCompletions:
     val params: List[Symbol] =
       allParams.filter(param => param.name.startsWith(prefix))
 
+    val completionSymbols = indexedContext.scopeSymbols
+    def matchingTypesInScope(paramType: Type): List[String] =
+      completionSymbols
+        .collect {
+          case sym
+              if sym.info <:< paramType && sym.isTerm && !sym.info.isNullType && !sym.info.isNothingType && !sym
+                .is(Flags.Method) =>
```
tanishiking, line 125: [nit] format
```suggestion
              if sym.info <:< paramType &&
                sym.isTerm &&
                !sym.info.isNullType &&
                !sym.info.isNothingType &&
                !sym.is(Flags.Method) =>
```

something like that

mtags/src/main/scala-3/scala/meta/internal/pc/completions/NamedArgCompletions.scala
```diff
@@ -104,12 +116,70 @@ object NamedArgCompletions:
...
+    def findPossibleDefaults(): List[CompletionValue] =
+      params.flatMap { param =>
+        val allMembers = matchingTypesInScope(param.info)
+        allMembers.map { memberName =>
+          val editText =
+            param.nameBackticked + " = " + memberName + " "
+          CompletionValue.namedArg(
+            label = editText,
+            param,
+            Some(editText),
```
tanishiking, line 172: What do you think about adding `defaultValue: Option[String]` to `CompletionValue.NamedArg` and construct the insertText on `CompletionsProvider` site? (Instead of adding `insertText: Option[String]` to `NamedArgument`, and construct it here).

Sorry for the mess, but I want to keep the `insertText` construction as a `CompletionsProvider`'s responsibility. (So that we can easily find where the `insertText` is constructed).


(TBH, I don't like having `insertText` field to `CompletionValue`, though it was needed for `InterpolationCompletion` but it should be limited to `InterpolationCompletion`)

tanishiking (changes requested):
@jkciesluk Thank you for working on this!
I left some comments about the architecture, but otherwise it looks really awesome 🚀

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
jkciesluk: That's a good idea, I will change it

mtags/src/main/scala-3/scala/meta/internal/pc/completions/NamedArgCompletions.scala
jkciesluk: Maybe we can remove `insertText` from here altogether, because it is the same as label?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
```diff
@@ -318,8 +318,11 @@ class CompletionsProvider(
           filterText,
           Some(completionPos.copy(start = start).toEditRange),
         )
-      case CompletionValue.NamedArg(label, _) =>
-        mkItem(ident, ident.replace("$", "$$")) // escape $ for snippet
+      case CompletionValue.NamedArg(label, _, insertText) =>
+        mkItem(
+          ident,
+          insertText.getOrElse(ident.replace("$", "$$")),
```
jkciesluk, line 324: If we remove insertText and use label instead, should we still replace "$" with "$$" here, because it might not be needed when we have completion with default argument. Can we do this `replace` in `NamedArgCompletions`?

tanishiking (approved):
LGTM!

tanishiking:
Ah, wait the cross test is failing, but otherwise LGTM :+1:

tanishiking (changes requested):
Ah, wait the cross test is failing, but otherwise LGTM :+1:

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionValue.scala
```diff
@@ -61,7 +62,7 @@ sealed trait CompletionValue:
         s"${printer.completionSymbol(ext.symbol)} (extension)"
       case so: CompletionValue.Symbolic =>
         printer.completionSymbol(so.symbol)
-      case CompletionValue.NamedArg(_, tpe) =>
+      case CompletionValue.NamedArg(label, tpe) =>
```
tgodzik, line 65:
```suggestion
      case CompletionValue.NamedArg(_, tpe) =>
```

tgodzik (approved):
Just a minor comment, but we can fix it in another PR.

Issue #4249, opened 2022-08-11
Incorrect `activeParameter`


ckipp01:
### Describe the bug

When executing a `textDocument/signatureHelp` at times I get the wrong `activeParameter`. You can see an example of this in the Metals codebase. Trigger signature help on top of `viewId` here:

https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/tvp/ClasspathTreeView.scala#L33

Here is the request and response:

```json
[Trace - 00:57:47 pm] Received request 'textDocument/signatureHelp - (131)'
Params: {
  "textDocument": {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/scalameta-org/metals/metals/src/main/scala/scala/meta/internal/tvp/ClasspathTreeView.scala"
  },
  "position": {
    "line": 32,
    "character": 10
  }
}


[Trace - 00:57:47 pm] Sending response 'textDocument/signatureHelp - (131)'. Processing request took 8ms
Result: {
  "signatures": [
    {
      "label": "apply(viewId: String, nodeUri: String, label: String, command: MetalsCommand \u003d ..., icon: String \u003d ..., tooltip: String \u003d ..., collapseState: String \u003d ...): TreeViewNode",
      "documentation": {
        "kind": "markdown",
        "value": ""
      },
      "parameters": [
        {
          "label": "viewId: String",
          "documentation": {
            "kind": "markdown",
            "value": ""
          }
        },
        {
          "label": "nodeUri: String",
          "documentation": {
            "kind": "markdown",
            "value": ""
          }
        },
        {
          "label": "label: String",
          "documentation": {
            "kind": "markdown",
            "value": ""
          }
        },
        {
          "label": "command: MetalsCommand \u003d ...",
          "documentation": {
            "kind": "markdown",
            "value": ""
          }
        },
        {
          "label": "icon: String \u003d ...",
          "documentation": {
            "kind": "markdown",
            "value": ""
          }
        },
        {
          "label": "tooltip: String \u003d ...",
          "documentation": {
            "kind": "markdown",
            "value": ""
          }
        },
        {
          "label": "collapseState: String \u003d ...",
          "documentation": {
            "kind": "markdown",
            "value": ""
          }
        }
      ]
    }
  ],
  "activeSignature": 0,
  "activeParameter": 1
}
```

This ends up showing the user that they're on the incorrect parameter:

[...]

PR #4251, opened 2022-08-11
bugfix: Fix signature help when named params are present

tgodzik:
Previously, we would show a wrong parameter, when named params were present, in two ways:
- use wrong position for checking current highlighted param - this is due to the fact that all the params are synthetic in such situation
- use wrong order of params for the parameters present in the correct place

Now, those two issue are handled properly by:
- first, correct position was already calculated, but not used
- second, by adjusting the sorting

Fixes https://github.com/scalameta/metals/issues/4249

I added more docs than code since the behaviour was a little unclear


ckipp01 (approved):
Good catch and nice explanation! LGTM!

PR #4252, opened 2022-08-11
chore: bump local snapshot version

kpodsiad:


tgodzik (commented):
Thanks for raising this, although the version is already bumped from what I see:

https://github.com/scalameta/metals/blob/75f6ac27550b4820dfa7bcd658768752a458d4a7/build.sbt#L8

wierdly the diff doesn't show it.

Issue #4254, opened 2022-08-12
Don't attempt to compile targets that have `canCompile` set to false


ckipp01:
### Describe the bug

I just hit on this in https://github.com/com-lihaoyi/mill/pull/1977 but Metals seems to send in compile requets for all targets no matter what. I dug in a bit since I thought it'd be a quick fix and thought I could just change this:

https://github.com/scalameta/metals/blob/ab314f8f23de51b84c4415aa0f8f355c893f5502/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala#L184-L196

to this:

```scala
    val targetsByBuildServer = targets
      .flatMap(target =>
        buildTargets.buildServerOf(target).map(_ -> target).toSeq
      )
      .groupBy { case (buildServer, _) =>
        buildServer
      }
      .map { case (buildServer, targets) =>
        val targets0 = targets.collect {
          case (_, target) if buildTargets.canCompile(target) =>
            target
        }
        (buildServer, targets0)
      }
```

And then add in a `canCompile` in `BuildTargets`:

```scala
  def canCompile(id: BuildTargetIdentifier): Boolean = {
    scribe.info(s"checking to see if we can compile ${id}")
    val result = data.iterable.exists(targetData => {
      pprint.log(targetData.buildTargetInfo)
      targetData.buildTargetInfo
        .get(id)
        .map[Boolean](_.getCapabilities().getCanCompile())
        .getOrElse(true)
    })
    result
  }
```
However when trying this out `builtTargetInfo` is always empty. There is a ton of stuff that recently changed here and it seems that here where this should be populated:

https://github.com/scalameta/metals/blob/ab314f8f23de51b84c4415aa0f8f355c893f5502/metals/src/main/scala/scala/meta/internal/metals/Indexer.scala#L196
`workspaceBuildTargets` this seems to add it, but I have a feeling the way we are trying to access this in `compilations` or something needs to be lazy.

### Expected behavior

I'd expect Metals to not send in a compile request to build targets that have `canCompile` set to false.

### Operating system

macOS

### Editor/Extension

Nvim (nvim-metals)

### Version of Metals

Latest snapshot.

tgodzik:
I'll take a look :eyes:

PR #4256, opened 2022-08-12
bugfix: Check if the target can be compiled

tgodzik:
Previously, we would not check if a build target can be compiled before sending a compile request. Now, we will not send the compile request if the build tool specifies a target as a one that can't be compiled.

Fixes https://github.com/scalameta/metals/issues/4254

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
@@ -202,6 +202,15 @@ final class BuildTargets() {
   ): Option[AbsolutePath] =
     buildServerOf(buildTarget).map(_.workspaceDirectory)
 
+  def canCompile(id: BuildTargetIdentifier): Boolean = {
+    targetData(id).exists { data =>
+      data.buildTargetInfo
+        .get(id)
+        .map[Boolean](_.getCapabilities().getCanCompile())
+        .getOrElse(true)
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
```diff
@@ -202,6 +202,15 @@ final class BuildTargets() {
   ): Option[AbsolutePath] =
     buildServerOf(buildTarget).map(_.workspaceDirectory)
 
+  def canCompile(id: BuildTargetIdentifier): Boolean = {
+    targetData(id).exists { data =>
```
tgodzik, line 206: @ckipp01 `data.iterable.exists` would return the first build data that was true and since there is an empty AmmoniteBuildData that would alywas be true. We need to first get the data for a specific target id.

ckipp01: O duh! Of course! 😅

ckipp01 (approved):
Thanks for the quick fix!! LGTM!

metals/src/main/scala/scala/meta/internal/metals/BuildTargets.scala
tgodzik: It's always an issue like this with `exists` and `forall` :sweat_smile: It's easier to figure out from the backseat

PR #4257, opened 2022-08-12
bugfix: Don't show preparing PC message

tgodzik:
Previously, we would show "Preparing presentation compiler" message, but since it seems to be on a critical path, it sometimes popped up and would not stop. Also the current place this is in is not actually downloading anything. Now, I decided after multiple iterations to just not show the message. If anything breaks we will show a proper message, but otherwise the message didn't add anything and might cause issues for the users.


dos65 (approved):
LGTM

The actual issue with this popup is somewhere on client side?

ckipp01 (approved):

tgodzik:
> LGTM
> 
> The actual issue with this popup is somewhere on client side?

I think it might be or a race. Anyway, the notification is being invoked many times and there is 5 second delay, so this is super prone to race conditions.

PR #4258, opened 2022-08-13
refactor: include the companion object name in code action title

ckipp01:
Currently when you're in a large file with possible nested classes and
you trigger a code action you just see the `Create companion object` as
the code action title, but you don't get an indication of what you'll be
creating the companion object for. We do have that information so this
small change just adds the name of the thing we'll be creating a
companion object for in the code action title.


tanishiking (approved):
noice, LGTM!

PR #4267, opened 2022-08-15
fix: specify  `--release 8` flags

dos65:
In #4221 we started releasing artifacts from jdk17


tgodzik (approved):
LGTM! Good catch! Shouldn't we also change The JDK to 8 for release?

dos65:
@tgodzik I think it's fine to keep it 17

PR #4270, opened 2022-08-15
bugfix: Properly restart worksheet presentation compiler

tgodzik:
Previously, we could get stale information in the worksheet, because we would not fully restart the worksheet presentation compilers. Now, we remove old compilers the same as with normal PC.


dos65 (approved):

tanishiking (approved):
:+1:

PR #4273, opened 2022-08-16
bugfix: [Scala 3] Auto import symbols in string interpolation

tgodzik:
Previously, when autocompleting inside a string, no workspace completions would be suggested. Now, they are added correctly.

It might be good to do some refactorings later on in the completions once we have everything in place.

Fixes https://github.com/scalameta/metals/issues/2105

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
@@ -17,7 +17,9 @@ import scala.meta.pc.SymbolSearch
 
 import dotty.tools.dotc.ast.tpd
 import dotty.tools.dotc.ast.tpd.*
+import dotty.tools.dotc.core.Constants.Constant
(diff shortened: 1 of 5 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
```diff
@@ -257,46 +259,97 @@ class CompletionsProvider(
...
+                      name,
+                    ),
+                    _,
+                  ),
+                  _,
+                ),
+                _,
+              ),
+              _,
+            ) :: _ =>
```
tanishiking, line 281: Can we accommodate these lines in a few lines? 😄

tanishiking (approved):
I left a nitpick about format, but otherwise LGTM!

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionsProvider.scala
tgodzik: Sure! Fixed.

Issue #4207, opened 2022-07-31
No effects from "Extract" code action


ckipp01:
### Describe the bug

Lately when triggering code actions I see a lot of messages that are unclear related to the "Extract ... as value" code action. For example I just opened Metals and triggered this in a few places and while I can't get it to happen all the time I often see code actions that look like this:

<img width="585" alt="Screenshot 2022-07-31 at 18 49 56" src="https://user-images.githubusercontent.com/13974112/182036872-689982f4-03a6-41a9-8ac5-855cf77f2e4f.png">

I'll try to get a more minimal reproduction, but I'm getting this often enough to just create this.

### Expected behavior

I'd never expect to see a code action like this. When selected it also jumps you to a random place in your file. There are times that it will offer me this, and then after a bit on the same place it won't offer it to me anymore.

Here is part of the response that is relevant
```
[Trace - 06:47:13 pm] Sending response 'textDocument/codeAction - (25)'. Processing request took 9ms
Result: [
  {
    "title": "Extract `{\n    val ` ... as value",
    "kind": "refactor.extract",
    "edit": {
      "changes": {
        "file:///Users/ckipp/Documents/scala-workspace/scalameta-org/metals/metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala": [
          {
            "range": {
              "start": {
                "line": 30,
                "character": 40
              },
              "end": {
                "line": 90,
                "character": 3
              }
            },
            "newText": "(newValue)"
          },
          {
            "range": {
              "start": {
                "line": 30,
                "character": 32
              },
              "end": {
                "line": 30,
                "character": 33
              }
            },
```

### Operating system

macOS

### Editor/Extension

Nvim (nvim-metals)

### Version of Metals

Latest snapshot

tgodzik:
Thanks for reporting! I think we have 3 different issues here (discussed it a bit offline with @jkciesluk ):
- the title - we should make it more meaningful, we shouldn't show empty lines and in case of longer expressions maybe a range? This might require some tinkering.
- Maybe we shouldn't extract a single apply with {}, since that is usually some kind of Future or Task and extracting value would change the logic. Instead Extract function should work here,
- Extracting in the case @ckipp01 found is not working at all, but it's only needed to fix if we don't change the second point.

kpodsiad:
Small reproduction for people without imagination (me 😄 )
```scala
    val x = Try {
      Some(new Exception)
    }
```
[...]

PR #4274, opened 2022-08-17
Better title for `extract value` code action

jkciesluk:
When extracting from block apply, like `Future { ...@@ }` we have to extract as method, not as a value, to not change the logic. Also improves titles, from `{ \n <<code>>` to `code`.
Connected to https://github.com/scalameta/metals/issues/4207

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
@@ -314,7 +314,8 @@ class ExtractValueCodeAction(
 
 object ExtractValueCodeAction {
   def title(expr: String): String = {
-    if (expr.length <= 10) s"Extract `$expr` as value"
-    else s"Extract `${expr.take(10)}` ... as value"
+    val trimmed = expr.trim.stripPrefix("{").stripSuffix("}").trim()
+    if (trimmed.length <= 10) s"Extract `$trimmed` as value"
+    else s"Extract `${trimmed.take(10)}` ... as value"
   }
 }


metals/src/main/scala/scala/meta/internal/metals/codeactions/ExtractValueCodeAction.scala
```diff
@@ -127,7 +137,7 @@ class ExtractValueCodeAction(
   private def existsRangeEnclosing(
       term: Tree,
       range: l.Range,
-  ): Boolean = {
+  ): (Boolean) = {
```
tanishiking, line 140: nit
```suggestion
  ): Boolean = {
```

tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala
```diff
@@ -164,12 +164,12 @@ class ExtractValueLspSuite
...
        |}
        |""".stripMargin,
-    ExtractValueCodeAction.title("{ 1 + 2 }"),
+    ExtractValueCodeAction.title("{ 1 + 2 }", "method"),
     """|object Main {
        |  def method2(i: Int) = ???
        |  val (newValue, newValue1) = (0, 1)
        |  def newValue0 = ???
-       |  val newValue2 = { 1 + 2 }
+       |  def newValue2 = { 1 + 2 }
```
tanishiking, line 172: Do we need to extract those blocks as a function? val newValue2 = { 1 + 2 } should aldo be valid

tanishiking (commented):
Badically,. LGTM, nice catch!
I'm just wondering can we keep the extracted block things as valur, instead of extracting them as a function?

tests/unit/src/test/scala/tests/codeactions/ExtractValueLspSuite.scala
jkciesluk: We can check if there is a `(` between method and `{ ... }` if that's what you mean

tanishiking: Thanks! but I meant to, we don't need to extract those values as function, not only in this case but all the cases.

> When extracting from block apply, like Future { ...@@ } we have to extract as method, not as a value

I don't think so. I think we can extract things as values, do you have an example where we need to extract code as a function, instead of values?

You tested against `Future { method(5) }` to `def newValue = { method(5) }`, but it can be `val newValue = method(5)`.
Even if the code trying to extract has multiple statements like

```scala
Future {
  println("foo")
  "foo"
}
```

we can extract them as value like `val newValue = { println("foo"); "foo" }`

tanishiking: I understand extracting as value may change the semantics (for example, if it's `IO`, extracting as value vs function change the semantics). However, most of the block applies (syntax) don't care if its function or value (for example, `Option { ... }`.

I think we don't necessarily need to preserve semantics by this command, it should be the users' responsibility.

jkciesluk: Yes, in most cases it doesn't matter if it's a value or function, but for `Future` don't we lose concurrency while extracting it as a value? How about a lazy val instead? 
Although you might be right that preserving semantics should be users responsibility, because there is (I think) no easy way  for us to distinguish where block apply extraction can change semantics

tanishiking: >  preserving semantics should be users responsibility, because there is (I think) no easy way for us to distinguish where block apply extraction can change semantics

Yeah, I think there's no way to preserve the semantics (we can treat `Future` or `IO` as a special case, but it's a slippery slope. Though I don't come up with another case off the top of my head, there could be other cases where this refactoring breaks the semantics, we can't cover everything).

So my opinion is that we shouldn't support such a smart refactoring feature, and this PR would fix the title of code action.

tanishiking (approved):
LGTM! 🚀

PR #4279, opened 2022-08-18
Fix mtags release

dos65:
Starting from 1.7.0 version sbt has more strict checks for crossVersions.
If a specific version isn't defined in `crossScalaVersion ` it can be passed only with force sign - `!`

### DIFF
--- a/.github/workflows/mtags-auto-release.yml
+++ b/.github/workflows/mtags-auto-release.yml
@@ -15,7 +15,7 @@ on:
         # If you update this line after release
         #   just put the tag name (`v*.*.*`) here as in `metals_version.value` above.
         # Don't be confused if this value contains `*.*.*_mtags_release`
-        default: "v0.11.8"
+        default: "0.11.8_mtags_release"
 jobs:
   test_and_release:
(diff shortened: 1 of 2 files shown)


.github/workflows/mtags-auto-release.yml
```diff
@@ -15,7 +15,7 @@ on:
         # If you update this line after release
         #   just put the tag name (`v*.*.*`) here as in `metals_version.value` above.
         # Don't be confused if this value contains `*.*.*_mtags_release`
-        default: "v0.11.8"
+        default: "0.11.8_mtags_release"
```
tanishiking, line 18:
```suggestion
        default: "v0.11.8_mtags_release"
```

Is this fine without `v` ?

tanishiking (approved):

.github/workflows/mtags-auto-release.yml
dos65: yep - I already created it without v because it's a branch - https://github.com/scalameta/metals/tree/0.11.8_mtags_release

Issue #4232, opened 2022-08-08
Implement All Members for trait instantiation with `new` not working


vzmerr:
### Describe the bug

Metals suggests t "Implement All Members" in the Quick Fix for the code below in Scala 3, but when it is selected, nothing happens
![ImplementAllMembersTrait](https://user-images.githubusercontent.com/109077949/183398215-9f54cd89-d319-40dc-9a44-d7b4712ffcfa.gif)

### Expected behavior

It should implement `def iterator: Iterator[Int] = ???` 

### Operating system

macOS

### Editor/Extension

VSCode

### Version of Metals

0.11.7+79-d2ab7248-SNAPSHOT

### Extra context or search terms

Scala 3, Implement all members, trait, new

tgodzik:
Thanks for finding this! @tanishiking could you take a look at it?

PR #4284, opened 2022-08-22
fix: fix implement-all members for anonymous class

tanishiking:
fix: https://github.com/scalameta/metals/issues/4232

We had a pc test
    https://github.com/scalameta/metals/blob/ae5c4650aefb6e6228ad77991c5813c299e860ed/tests/cross/src/test/scala/tests/pc/AutoImplementAbstractMembersSuite.scala#L220-L236

but the error position is different in Scala.
`<<new>> Foo` instead of `new <<Foo>>`

This PR fixes the issue, and added a LSP test.


dos65 (approved):
:+1:

tgodzik (approved):
LGTM!

PR #4289, opened 2022-08-22
bugfix: Allow to run/debug in ScalaCli when started from Metals

tgodzik:
Previously, if ScalaCLI was not started from `.bsp` it would not be possible to run or debug. Now, we also check if we can run using built in ScalaCLI


dos65 (approved):
LGTM!

PR #4291, opened 2022-08-23
fix: don't break using-directives by an auto-import for `.sc` in scala-cli

dos65:
Scala-cli wraps `.sc` files in a similar way as ammonite - puts it in
object so the resulting code in pc is:
```
object main {
/*script*///> using directive
//> using directive

...
}
```

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
@@ -1,6 +1,5 @@
 package scala.meta.internal.pc
 
-import scala.annotation.tailrec
 import scala.reflect.internal.FatalError
 
 import scala.meta.internal.mtags.MtagsEnrichments._
@@ -60,11 +59,30 @@ trait AutoImports { this: MetalsGlobal =>
             )
           }
 
-        def forScript =
-          for {
-            obj <- lastVisitedParentTrees.collectFirst { case mod: ModuleDef =>
-              mod
+        def forScript(isAmmonite: Boolean) = {
+          val startScriptOffest = {
+            if (isAmmonite)
+              ScriptFirstImportPosition.ammoniteScStartOffset(text)
+            else ScriptFirstImportPosition.scalaCliScStartOffset(text)
+          }
+
+          val scriptModuleDefAndPos =
+            startScriptOffest.flatMap { offset =>
+              val startPos = pos.withStart(offset).withEnd(offset)
+              lastVisitedParentTrees
+                .collectFirst {
+                  case mod: ModuleDef if mod.pos.overlaps(startPos) => mod
+                }
+                .map(mod => (mod, offset))
             }
+
+          val moduleDefAndPos = scriptModuleDefAndPos.orElse(
+            lastVisitedParentTrees
+              .collectFirst { case mod: ModuleDef => mod }
+              .map(mod => (mod, 0))
+          )
+          for {
+            (obj, firstImportOffset) <- moduleDefAndPos
           } yield {
             val lastImportOpt = obj.impl.body.iterator
               .dropWhile {
@@ -73,34 +91,33 @@ trait AutoImports { this: MetalsGlobal =>
               }
               .takeWhile(_.isInstanceOf[Import])
               .lastOption
-            val lastImportLine = lastImportOpt
+
+            val offset = lastImportOpt
               .map(_.pos.focusEnd.line)
-              .getOrElse(0) // if no previous import, add the new one at the top
+              .map(pos.source.lineToOffset)
+              .getOrElse(firstImportOffset)
             new AutoImportPosition(
-              pos.source.lineToOffset(lastImportLine),
+              offset,
               text,
               padTop = false
             )
           }
-
-        // Naive way to find the start discounting any first lines that may be
-        // scala-cli directives.
-        @tailrec
-        def findStart(text: String, index: Int): Int = {
-          if (text.startsWith("//")) {
-            val newline = text.indexOf("\n")
-            if (newline != -1)
-              findStart(text.drop(newline + 1), index + newline + 1)
-            else index + newline + 1
-          } else {
-            index
-          }
         }
 
         def fileStart =
-          AutoImportPosition(findStart(text, 0), 0, padTop = false)
+          AutoImportPosition(
+            ScriptFirstImportPosition.skipUsingDirectivesOffset(text),
+            0,
+            padTop = false
+          )
 
-        (if (pos.source.path.endsWith(".sc.scala")) forScript else None)
+        val path = pos.source.path
+        val scriptPos =
+          if (path.endsWith(".sc")) forScript(isAmmonite = false)
+          else if (path.endsWith(".amm.sc.scala")) forScript(isAmmonite = true)
+          else None
+
+        scriptPos
           .orElse(forScalaSource)
           .orElse(Some(fileStart))
     }
--- a/mtags/src/main/scala/scala/meta/internal/pc/ScriptFirstImportPosition.scala
+++ b/mtags/src/main/scala/scala/meta/internal/pc/ScriptFirstImportPosition.scala
@@ -1,0 +1,103 @@
+package scala.meta.internal.pc
+
+import scala.annotation.tailrec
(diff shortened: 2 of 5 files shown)


tests/cross/src/test/scala/tests/pc/AutoImportsSuite.scala
```diff
@@ -263,22 +298,10 @@ class AutoImportsSuite extends BaseAutoImportsSuite {
       """val p: <<Path>> = ???
         |""".stripMargin
     ),
-    // Import added *before* the wrapper here…
-    // This *seems* wrong, but the logic converting the scala file
-    // edits to sc file edits will simply add it at the beginning of
-    // the sc file, as expected.
```
dos65, line 269: There were a need to distinguish amm sc files from scala-cli so in addition I fixed first import for ammonite.

tanishiking (commented):
Hey @dos65, can you elaborate on what was the problem?

I tried to understand what the issue is, but I couldn't. Is this a user-facing issue? or some internal problem in Metals?

scalacli

dos65:
@tanishiking There is a difference how it works in scala2 and scala3. If you try to do the same for scala2 you will see that it puts first auto-import at the first line that breaks using-directives
```scala
//> using scala "2.13.8"

Try(...)
```

tanishiking:
Thanks! right, I could confirm that :+1:

mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -60,11 +60,40 @@ trait AutoImports { this: MetalsGlobal =>
...
+          // skip comments/ using directives
+          val additionalOffest =
+            text.drop(firstBodyLine).offsetAfterUsingDirectives
+
+          firstBodyLine + additionalOffest
+        }
+
+        def forScript = {
+          println(text)
+          println("~~~")
```
tanishiking, line 81:
```suggestion
```

mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -60,11 +60,40 @@ trait AutoImports { this: MetalsGlobal =>
...
+          val moduleDef =
+            if (text.startsWith("package ammonite")) {
+              // ammonite wraps sc code into `object test { ... }`
+              lastVisitedParentTrees.collect {
+                case mod: ModuleDef if mod.name.decoded == "test" => mod
```
tanishiking, line 86: This is not true, Ammonite wraps `object foo{` if the script name is `foo.sc` (backtick if it is required)

https://github.com/com-lihaoyi/Ammonite/blob/bec02ae120eca3002b19405ddaea71b1fbbcaed9/amm/compiler/src/main/scala/ammonite/compiler/DefaultCodeWrapper.scala#L24
(Ammonite script runner looks like provides the file name for it).

mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -73,34 +102,28 @@ trait AutoImports { this: MetalsGlobal =>
...
-          } else {
-            index
-          }
         }
 
-        def fileStart =
-          AutoImportPosition(findStart(text, 0), 0, padTop = false)
+        def fileStart = {
+          val x = text.offsetAfterUsingDirectives
+          println(s"WHAT? $x")
```
tanishiking, line 122:
```suggestion
```

tanishiking (changes requested):

mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -73,34 +102,28 @@ trait AutoImports { this: MetalsGlobal =>
...
               .takeWhile(_.isInstanceOf[Import])
               .lastOption
-            val lastImportLine = lastImportOpt
+
+            val offset = lastImportOpt
               .map(_.pos.focusEnd.line)
-              .getOrElse(0) // if no previous import, add the new one at the top
+              .map(pos.source.lineToOffset)
+              .getOrElse(
+                firstBodyLineOffset(obj.impl.pos)
```
tanishiking, line 110: It might be easier to get the position of  `/*<script>*/` that represents the start of `sc` content 🤔

mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
dos65: ouch. thx for finding this!

mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
tanishiking: Oops, that doesn't work because as you mentioned, scala-cli script will have something like

```scala
/*<script>*///> using scala "2.13.8"
println(1)
```

tanishiking: Maybe we can rely on `/*<script>*/` because skipping using directive is done by `offsetAfterUsingDirectives` ?

dos65: yep - might be an option
I've checked what exactly ammonite pass as a text and it has a similar marker -  `/*<start>*/`

tanishiking: Yeah it should be reliable :+1: 
https://github.com/com-lihaoyi/Ammonite/blob/bec02ae120eca3002b19405ddaea71b1fbbcaed9/amm/compiler/interface/src/main/scala/ammonite/compiler/iface/CodeWrapper.scala#L33

dos65: reworked it - now it's based on location of `/*<script>*/`

mtags/src/main/scala/scala/meta/internal/pc/ScriptFirstImportPosition.scala
```diff
@@ -0,0 +1,103 @@
...
+  def scalaCliScStartOffset(text: String): Option[Int] = {
+    val iterator = tokenize(text).iterator
+    startMarkerOffset(iterator, "/*<script>*/").map { startOffset =>
+      val offset =
+        skipUsingDirectivesOffset(iterator, None)
+          .getOrElse(startOffset)
+
+      offset + 1
+    }
+  }
```
tanishiking, line 44: Can we run `skipUsingDirectivesOffset` anyway for both the ammonite script and scala-cli script?
I think it's ok to insert an import under using directive for ammonite script (using directive doesn't mean anything in Ammonite script though).

tanishiking (commented):
it's almost looks good, but I left some minor comments

mtags/src/main/scala-2/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -60,11 +59,35 @@ trait AutoImports { this: MetalsGlobal =>
...
+              val modules = lastVisitedParentTrees
+                .collect { case mod: ModuleDef => mod }
+                .filter(mod => mod.pos.overlaps(startPos))
+
+              if (modules.nonEmpty) {
+                // select the smalles ModuleDef that contains start marker
+                val closestModule = modules
+                  .minBy(mod => mod.pos.end - mod.pos.start)
+                Some(closestModule, offset)
+              } else None
```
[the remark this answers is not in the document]
dos65: thx! updated!

mtags/src/main/scala/scala/meta/internal/pc/ScriptFirstImportPosition.scala
dos65: We can. But there a might be a case with 0.00...1% chance that you have an ammonite source that just has a comment starting from using comment :smile_cat: 
Like:
```scala
// using this script ....
// bla-bla
${code}
```
+ it isn't costly to use `skipUsingDirectives` only scala-cli sc

tanishiking (approved):
LGTM 👍 nice catch!

PR #4293, opened 2022-08-23
chore: improve scalatest style inferring

kpodsiad:
For every base class like `AnyFunSuite` there is also trait `AnyFunSuiteLike` which also can be used when writing tests. Previously, inferring styles wasn't working with those traits, this PR fixes it.


tanishiking (approved):
LGTM:+1:

PR #4295, opened 2022-08-24
feat: better mill-bsp semanticdb support

ckipp01:
In the last release of Mill they added better support for producing
SemanticDB for clients like Metals without needing any changes in your
build. However the way it's done is a bit different and unlike the other
build servers meaning that you can't detect it by looking at the
scalacOptions. This changes special-cases Mill for now when a user is
using mill-bsp in a version >= 0.10.6. Mainly this is so users don't see
a big red X mark next to semanticdb in the doctor. There still will be
one for Java support as currently there is no semanticdb being produced
for Java targets in Mill.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
+++ b/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala
@@ -129,6 +129,10 @@ case class MillBuildTool(userConfig: () => UserConfiguration)
 
 object MillBuildTool {
   val name: String = "mill-bsp"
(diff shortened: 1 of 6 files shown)


tanishiking (approved):
LGTM! I'm really happy to see mill-bsp natively support SemanticDB

lefou:
Which version first contains this PR?

ckipp01:
> Which version first contains this PR?

You can try this out with the latest snapshot: `0.11.8+72-6567f1a3-SNAPSHOT`.

Issue #1809, opened 2020-06-05
'find-definition' finds companion object instead of case class


chessman:
**Describe the bug**
```
case class A(b: String)
object A {}
A|("b")
```
The cursor is on the last line, invoke 'find-definition'.

**Expected behavior**
Jump to `case class A`

**Actual behavior**
Jump to `object A`

**Installation:**
- Operating system: Linux
- Editor: Emacs
- Metals version: v0.9.0

**Additional context**
If object defines `apply` method, then, of course, it should jump to this method.

tgodzik:
Thanks for reporting! I think what is happening here is that `A("b")` is a shorthand for `A.apply("b")`, which is a reference to the object and that is why we are jumping to definition.

We could maybe try in case of a synthetic apply try to jump to the class? However, I am not sure how difficult it would be currently.

tgodzik:
Also, I am not sure if it is exactly a bug :thinking: 

Your cursor is on an object reference and that's why it's jumping to the object. Any kind of workaround that we do here, might not sit well with everyone. Any opinions?

chessman:
I've tested more cases. If I create a case class only, without a companion object, then it's jumping to the case class. But if I have both and then delete the object, it says "No definitions found: A".

What does it mean that `A.apply("b")` is a reference to the object? The constructor is defined in the case class. Moreover, it jumps to the case class from `A.apply("b")`.

tgodzik:
> What does it mean that A.apply("b") is a reference to the object?

The way the case classes work is that they generate additional `apply` method inside of the companion object and apply methods can just be invoked using `()`. So A("b") is actually invoking the objects's apply method, which later on is invoking the actual constructor.

I think we could make it a bit more consistent for sure, but it would require some careful thought.

olafurpg:
Thank you for reporting! The current behavior is intentional but I agree it would be nice in some cases to jump to the class definition instead of the companion object. However, note that the apply method is technically defined inside the companion.

 IntelliJ handles this case by prompting the user to select which one to navigate, Metals could do the same by returning multiple definitions. If we go this route I think we would need to handle FunctionN.apply specially since in that case I think we always want to navigate to the functions symbol and not the apply method.

PR #4296, opened 2022-08-24
bugfix: Find correct definition of apply when using case classes

tgodzik:
Previously, DefinitionProvider would just look for the definition of object. Now, it also checks if the object is not a synthetic apply and in that case it tries to find that exact apply in addition to the object.

Fixes https://github.com/scalameta/metals/issues/1809


ckipp01 (approved):
Nice, LGTM!

PR #4297, opened 2022-08-24
Scala cli actionable diagnostic

lwronski:
This PR adds support for actionable diagnostic from `scala-cli`. These are diagnostics that contain information that allows users to fix some configs/directives by `Quick Fix` actions in metals.

For now, in `scala-cli` there is only implemented actionable diagnostics for updating lib in using directives, and it works in this way with metals:

https://user-images.githubusercontent.com/46607934/186489738-71ed74df-2ed0-4ce3-8466-b21d4674c2c0.mov

I added `didSave` flag to `check` method, because when actionable diagnostics were applied in unit tests and saved content to the file, the `scala-cli` reload bsp server and at the same time metals expect to get `Compile Result` from server but fail, because bsp server is down. I attach error logs:

```
2022.08.23 14:48:03 INFO  Disconnecting from scala-cli session...
2022.08.23 14:48:02 INFO  Shut down connection with build server.
2022.08.23 14:48:03 INFO  Attempting to connect to the build server...
2022.08.23 14:48:03 INFO  Running BSP server [/Users/lwronski/projects/scala-cli/out/cli/standaloneLauncher.dest/launcher, bsp, ., --actions, -S, 2.13.8]
2022.08.23 14:48:02 INFO  tracing is disabled for protocol BSP, to enable tracing of incoming and outgoing JSON messages create an empty file at /Users/lwronski/projects/metals/tests/unit/target/e2e/actionableDiagnostic/empty-string-multiline/.metals/bsp.trace.json or /Users/lwronski/Library/Caches/org.scalameta.metals/bsp.trace.json
tests.codeactions.ActionableDiagnosticsSuite:
==> X tests.codeactions.ActionableDiagnosticsSuite.empty-string-multiline  11.177s scala.meta.internal.metals.MetalsBspException: BSP connection failed in the attempt to get: CompileResult
Caused by: java.util.concurrent.ExecutionException: Boxed Exception
Caused by: java.lang.InterruptedException
```

To fix it, I just skip saving updating files to disk, and in this way, I avoiding to reload of bsp. I don't find any better solution to fix it in my test `ActionableDiagnosticsSuite`.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
@@ -1,8 +1,14 @@
 package scala.meta.internal.metals
 
+import scala.meta.internal.metals.MetalsEnrichments._
+
 import org.eclipse.{lsp4j => l}
 
 object ScalacDiagnostic {
+
+  object ScalaAction {
+    def unapply(d: l.Diagnostic): Option[l.TextEdit] = d.asTextEdit
+  }
 
   object NotAMember {
     private val regex = """(?s)value (.+) is not a member of.*""".r
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
@@ -1,0 +1,55 @@
+package scala.meta.internal.metals.codeactions
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+import scala.meta.internal.metals.MetalsEnrichments._
+import scala.meta.internal.metals._
+import scala.meta.pc.CancelToken
+
+import org.eclipse.{lsp4j => l}
+
+class ActionableDiagnostic() extends CodeAction {
+  override def kind: String = l.CodeActionKind.QuickFix
+
+  override def contribute(
+      params: l.CodeActionParams,
+      token: CancelToken,
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+
+    def createActionableDiagnostic(
+        diagnostic: l.Diagnostic,
+        textEdit: l.TextEdit,
+    ): l.CodeAction = {
+      val diagMessage = diagnostic.getMessage
+      val uri = params.getTextDocument().getUri()
+
+      CodeActionBuilder.build(
+        title =
+          s"Apply suggestion: ${diagMessage.linesIterator.headOption.getOrElse(diagMessage)}",
+        kind = l.CodeActionKind.QuickFix,
+        changes = List(uri.toAbsolutePath -> Seq(textEdit)),
+        diagnostics = List(diagnostic),
+      )
+    }
+
+    val codeActions = params
+      .getContext()
+      .getDiagnostics()
+      .asScala
+      .groupBy {
+        case ScalacDiagnostic.ScalaAction(textEdit) =>
+          Some(textEdit)
+        case _ => None
+      }
+      .collect {
+        case (Some(textEdit), diags)
+            if params.getRange().overlapsWith(diags.head.getRange()) =>
+          createActionableDiagnostic(diags.head, textEdit)
+      }
+      .toList
+      .sorted
+
+    Future.successful(codeActions)
+  }
+}
--- a/tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
+++ b/tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
@@ -9,6 +9,7 @@ import munit.Location
 import munit.TestOptions
 import org.eclipse.lsp4j.CodeAction
 import tests.BaseLspSuite
+import tests.BaseScalaCliSuite
 import tests.FileLayout
 
 abstract class BaseCodeActionLspSuite(
@@ -48,6 +49,7 @@ abstract class BaseCodeActionLspSuite(
       kind: List[String] = Nil,
       scalafixConf: String = "",
       scalacOptions: List[String] = Nil,
+      scalaCliOptions: List[String] = Nil,
       configuration: => Option[String] = None,
       scalaVersion: String = scalaVersion,
       renamePath: Option[String] = None,
@@ -56,6 +58,7 @@ abstract class BaseCodeActionLspSuite(
       changeFile: String => String = identity,
       expectError: Boolean = false,
       filterAction: CodeAction => Boolean = _ => true,
+      scalaCliLayout: Boolean = false,
   )(implicit loc: Location): Unit = {
     val scalacOptionsJson =
       if (scalacOptions.nonEmpty)
@@ -63,12 +66,21 @@ abstract class BaseCodeActionLspSuite(
       else ""
     val path = s"a/src/main/scala/a/$fileName"
 
-    val layout =
-      s"""/metals.json
-         |{"a":{$scalacOptionsJson "scalaVersion" : "$scalaVersion"}}
-         |$scalafixConf
-         |/$path
-         |$input""".stripMargin
+    val layout = {
+      if (scalaCliLayout)
+        s"""/.bsp/scala-cli.json
+           |${BaseScalaCliSuite.scalaCliBspJsonContent(scalaCliOptions)}
+           |/.scala-build/ide-inputs.json
+           |${BaseScalaCliSuite.scalaCliIdeInputJson(".")}
+           |/$path
+           |$input""".stripMargin
+      else
+        s"""/metals.json
+           |{"a":{$scalacOptionsJson "scalaVersion" : "$scalaVersion"}}
+           |$scalafixConf
+           |/$path
+           |$input""".stripMargin
+    }
 
     checkEdit(
       name,
--- a/tests/unit/src/test/scala/tests/codeactions/ActionableDiagnosticsSuite.scala
+++ b/tests/unit/src/test/scala/tests/codeactions/ActionableDiagnosticsSuite.scala
@@ -1,0 +1,36 @@
+package tests.codeactions
+
+import scala.meta.internal.mtags.CoursierComplete
+
+import coursier.version.Version
+
+class ActionableDiagnosticsSuite
+    extends BaseCodeActionLspSuite("actionableDiagnostic") {
+
+  val oldOsLibVersion: Version = Version("0.7.8")
+  val newestOsLib: String = CoursierComplete
+    .complete("com.lihaoyi::os-lib:")
+    .headOption
+    .getOrElse("0.8.1")
+
+  check(
+    "actionable-diagnostic-update",
+    s"""|//> <<>>using lib "com.lihaoyi::os-lib:${oldOsLibVersion.repr}"
+        |
+        |object Hello extends App {
+        |  println("Hello")
+        |}
+        |""".stripMargin,
+    s"Apply suggestion: com.lihaoyi::os-lib:0.7.8 is outdated, update to $newestOsLib",
+    s"""|//> using lib "com.lihaoyi::os-lib:$newestOsLib"
+        |
+        |object Hello extends App {
+        |  println("Hello")
+        |}
+        |""".stripMargin,
+    scalaCliOptions = List("--actions", "-S", scalaVersion),
+    expectNoDiagnostics = false,
+    scalaCliLayout = true,
+  )
+
+}
--- a/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Diagnostics.scala
@@ -261,12 +261,14 @@ final class Diagnostics(
       snapshot: Input,
   ): Option[Diagnostic] = {
     val result = edit.toRevised(d.getRange).map { range =>
-      new l.Diagnostic(
+      val ld = new l.Diagnostic(
         range,
         d.getMessage,
         d.getSeverity,
         d.getSource,
       )
+      ld.setData(d.getData)
+      ld
     }
     if (result.isEmpty) {
       d.getRange.toMeta(snapshot).foreach { pos =>
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsEnrichments.scala
@@ -677,6 +677,9 @@ object MetalsEnrichments
       val severity = d.getSeverity.toString.toLowerCase()
       s"$severity:$hint $uri:${d.getRange.getStart.getLine} ${d.getMessage}"
     }
+    def asTextEdit: Option[l.TextEdit] = {
+      decodeJson(d.getData, classOf[l.TextEdit])
+    }
   }
 
   implicit class XtensionSeverityBsp(sev: b.DiagnosticSeverity) {
@@ -752,8 +755,8 @@ object MetalsEnrichments
   }
 
   implicit class XtensionDiagnosticBsp(diag: b.Diagnostic) {
-    def toLsp: l.Diagnostic =
-      new l.Diagnostic(
+    def toLsp: l.Diagnostic = {
+      val ld = new l.Diagnostic(
         diag.getRange.toLsp,
         fansi.Str(diag.getMessage, ErrorMode.Strip).plainText,
         diag.getSeverity.toLsp,
@@ -761,6 +764,9 @@ object MetalsEnrichments
         // We omit diag.getCode since Bloop's BSP implementation uses 'code' with different semantics
         // than LSP. See https://github.com/scalacenter/bloop/issues/1134 for details
       )
+      ld.setData(diag.getData)
+      ld
+    }
   }
 
   implicit class XtensionHttpExchange(exchange: HttpServerExchange) {
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/CodeActionProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/CodeActionProvider.scala
@@ -29,6 +29,7 @@ final class CodeActionProvider(
     new ImplementAbstractMembers(compilers),
     new ImportMissingSymbol(compilers, buildTargets),
     new CreateNewSymbol(),
+    new ActionableDiagnostic(),
     new StringActions(buffers),
     extractMemberAction,
     new SourceOrganizeImports(
--- a/tests/unit/src/main/scala/tests/BaseScalaCliSuite.scala
+++ b/tests/unit/src/main/scala/tests/BaseScalaCliSuite.scala
@@ -101,35 +101,12 @@ abstract class BaseScalaCliSuite(scalaVersion: String)
        |}
        |
        |""".stripMargin
-
-  private def escape(s: String): String =
-    s.replace("\\", "\\\\")
   private def bspLayout =
     s"""/.bsp/scala-cli.json
-       |{
-       |  "name": "scala-cli",
-       |  "argv": [
-       |    "${escape(ScalaCli.javaCommand)}",
-       |    "-cp",
-       |    "${escape(ScalaCli.scalaCliClassPath().mkString(File.pathSeparator))}",
-       |    "${ScalaCli.scalaCliMainClass}",
-       |    "bsp",
-       |    "."
-       |  ],
-       |  "version": "${BuildInfo.scalaCliVersion}",
-       |  "bspVersion": "2.0.0",
-       |  "languages": [
-       |    "scala",
-       |    "java"
-       |  ]
-       |}
+       |${BaseScalaCliSuite.scalaCliBspJsonContent()}
        |
        |/.scala-build/ide-inputs.json
-       |{
-       |  "args": [
-       |    "."
-       |  ]
-       |}
+       |${BaseScalaCliSuite.scalaCliIdeInputJson(".")}
        |
        |""".stripMargin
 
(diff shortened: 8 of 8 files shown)


metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
```diff
@@ -1,9 +1,18 @@
...
+import scala.jdk.CollectionConverters._
+
 import org.eclipse.{lsp4j => l}
 
 object ScalacDiagnostic {
 
+  object ScalaAction {
+    def unapply(d: l.Diagnostic): Option[l.DiagnosticRelatedInformation] =
+      if (d.getSource == "scala-cli")
+        d.getRelatedInformation.asScala.toList.headOption
```
kpodsiad, line 12:
```suggestion
        d.getRelatedInformation.asScala.headOption
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
```diff
@@ -0,0 +1,66 @@
...
+
+class ActionableDiagnostic() extends CodeAction {
+  override def kind: String = l.CodeActionKind.QuickFix
+
+  override def contribute(
+      params: l.CodeActionParams,
+      token: CancelToken,
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+
+    lazy val parentUri =
```
kpodsiad, line 20:
```suggestion
    def parentUri =
```

metals/src/main/scala/scala/meta/internal/metals/scalacli/ScalaCli.scala
```diff
@@ -364,4 +364,34 @@ object ScalaCli {
 
   val name = "ScalaCli"
 
+  private def escape(s: String): String =
+    s.replace("\\", "\\\\")
+  def scalaCliBspJsonContent(args: List[String] = Nil): String = {
```
tgodzik, line 369: Is this needed only in the tests? Could you move it to a companion objects of BaseScalaCLISuite?

metals/src/main/scala/scala/meta/internal/metals/scalacli/ScalaCli.scala
```diff
@@ -364,4 +364,34 @@ object ScalaCli {
 
   val name = "ScalaCli"
 
+  private def escape(s: String): String =
```
tgodzik, line 367: You might use ujson for writing the Json and there will not be any need for escaping. Just a small suggestion, but I guess it was already like that before.

tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
```diff
@@ -54,19 +56,31 @@ abstract class BaseCodeActionLspSuite(
       changeFile: String => String = identity,
       expectError: Boolean = false,
       filterAction: CodeAction => Boolean = _ => true,
+      scalaCli: Boolean = false,
```
tgodzik, line 59: Could you maybe add layout as a parameter? Boolean `scalaCLi` is not clear on what it does.

tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
```diff
@@ -54,19 +56,31 @@ abstract class BaseCodeActionLspSuite(
...
+      skipDiff: Boolean = false,
+      didSave: Boolean = true,
```
tgodzik, line 61:
```suggestion
      saveResult: Boolean = true,
```

tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
```diff
@@ -54,19 +56,31 @@ abstract class BaseCodeActionLspSuite(
       changeFile: String => String = identity,
       expectError: Boolean = false,
       filterAction: CodeAction => Boolean = _ => true,
+      scalaCli: Boolean = false,
+      skipDiff: Boolean = false,
```
tgodzik, line 60:
```suggestion
      assertion:  Option[(Boolean, String)],
```
Maybe something like this? Have separate assertion to make sure `skipDiff` is not used by mistake?

metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
```diff
@@ -1,9 +1,18 @@
...
 
+import scala.jdk.CollectionConverters._
+
 import org.eclipse.{lsp4j => l}
 
 object ScalacDiagnostic {
 
+  object ScalaAction {
+    def unapply(d: l.Diagnostic): Option[l.DiagnosticRelatedInformation] =
+      if (d.getSource == "scala-cli")
```
tgodzik, line 11: I wonder if we might want to enable it for more build tools?

CC @ckipp01 This will work only for ScalaCLI for now, but I think this fits into the better diagnostics idea, what do you think about it? We could expand it to sbt/bloop/mill since they don't currently fill in that field.

```suggestion
      if (d.getRelatedInformation() != null && d.getRelatedInformation().nonEmpty)
```

tgodzik (changes requested):
Thanks for the great work! I left a bunch of comments, but mostly minor things.

metals/src/main/scala/scala/meta/internal/metals/scalacli/ScalaCli.scala
lwronski: Right, it is used only in the test, so I will move this functions to `BaseScalaCLISuite`.

tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
lwronski: I don't understand why you used Boolean there, so I changed the assertion type a bit to:
```
assertion:  Option[Unit => Unit]
```
which allows to override default assert in test

tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
lwronski: In `BaseCodeActionLspSuite` layout is created base on few parameters such as: `scalacOptionsJson`, `scalafixConf`, `path`, `input`. I'm not sure what you expect, do you want that users how to pass explicitly `layout: String`?

For now, I changed name from `scalaCli` to `scalaCliLayout`.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
```diff
@@ -0,0 +1,65 @@
...
+      .asScala
+      .groupBy {
+        case ScalacDiagnostic.ScalaAction(relatedInformation) =>
+          Some(relatedInformation)
+        case _ => None
+      }
+      .collect {
+        case (Some(relatedInformation), diags)
+            if params.getRange().overlapsWith(diags.head.getRange()) =>
+          createDiagnosticAction(diags.head, relatedInformation)
```
ckipp01, line 54: I think I'm a bit confused by this, but I might be missing it. Do you by chance have a trace that you could show from scala-cli about what this full diagnostic looks like with the related information?

If I've following this right, this assumes that related information is actionable right? But by definition, diagnostic related information isn't alway though right? Like in the situation where you have a diagnostic that also has a related info about another part of code that it relates to doesn't have any type of action tied to it, so this should't provide a code action in that case. Am I misunderstanding it, or wouldn't this provide a code action for every diagnostic that has a related information that has the same position as the diagnostic?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
```diff
@@ -0,0 +1,65 @@
...
+      }
+      .toList
+      .sorted
+
+    Future.successful(codeActions)
+  }
+}
+
+object ActionableDiagnostic {
+  def title = "Apply suggestion from build server"
```
ckipp01, line 64:
```suggestion
  def title = "Apply suggestion"
```
I don't think the user even needs to know that this is "from the build server".

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
tgodzik: It's kind of a hack really right now, that's why we first enabled it for ScalaCLI, since there is no way of defining what should be done with the related diagnostics. It would actually be great to be able get the `kind` of diagnostics :thinking:

tests/unit/src/main/scala/tests/codeactions/BaseCodeActionLspSuite.scala
tgodzik: I was thinking:
```
assertion.map{ case (condition, msg) => assert(condition, msg)}
```
This way we would steer with the types what needs to be done.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
ckipp01: Ahhh, I'd highly recommend not going against the spec for this then. This will sort of break assumptions then since the spec is pretty well defined here on how this should work. We haven't really used them at all, but if I'm understanding the flow correctly instead here we should be using `data`, which is part of the `Diagnostic` and can be persisted between a `textDocument/publishDiagnostics` and a `textDocument/codeAction`. This way when someone would trigger a code action on that diagnostic, the related information has nothing to do with the fix, since it's not meant to. We could then check against the data to know we can apply what is needed, and then apply it. If i'm not mistaken clang and tsserver do just this, they actual contain the fix right in the diagnostic data, so if it's triggered it can immediately be applied. I believe this is the approach we should follow instead of hacking this in this way.

tgodzik: Makes sense, but we would need to have the same thing for BSP's diagnostic, which is currently lacking. We only have RelatedInformation. We should add an object there then I think, though it might be ok for ScalaCLI use RelatedInformation for the time being :thinking:  Maybe for BSP we could actually add a specific `EditSuggestion`?

ckipp01: I think I'm still confused by the entire flow here. Do you have chance have a trace that I could look at going from diagnostic -> code action so I can see all the steps? Like where is the actual fix included? If it's just part of the related information, then I don't think that's where it should be as this is going against the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticRelatedInformation) and would need to be changed later the moment we start using this the correct way.

> Represents a related message and source code location for a diagnostic.
> This should be used to point to code locations that cause or are related to
> a diagnostics, e.g when duplicating a symbol in a scope.

For example this should be used just for extra information about other parts of your code, that's why by default it has a location so that location can be navigated to like you see [here](https://contributors.scala-lang.org/t/revisiting-dotty-diagnostics-for-tooling/5649/12?u=ckipp01).

tgodzik: Sure, that was the main hacky part, which is why I wanted to get someone to double check the decision :sweat_smile: 

I think we could use data, but despite `where Diagnostic is defined as it is in LSP.` in https://build-server-protocol.github.io/docs/specification.html#publish-diagnostics it's not actually true and we don't have an additional `data` object.

@lwronski would you be able to add that field to BSP, we could use it to be more correct here.

metals/src/main/scala/scala/meta/internal/metals/ScalacDiagnostic.scala
```diff
@@ -1,9 +1,21 @@
...
+
 import org.eclipse.{lsp4j => l}
 
 object ScalacDiagnostic {
 
+  object ScalaAction {
+    def unapply(d: l.Diagnostic): Option[l.DiagnosticRelatedInformation] =
+      if (
+        d.getRelatedInformation() != null &&
+        d.getRelatedInformation().asScala.nonEmpty
```
tanishiking, line 13: If Metals notifies diagnostics with `relatedInformation` (in addition to ActionalDiagnostics stuff) in the future, `ScalaAction` extractor may mis-extract those diagnostics as `ScalaAction`.

I think we can't rely only on the existence of `RelatedInformation` field. We need to check something like an error message or error code or something I guess.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
lwronski: @tgodzik @ckipp01 Sorry for the delay in answering, meanwhile I added a `data` field in `Diagnostic` in `bsp4j` - https://github.com/build-server-protocol/build-server-protocol/pull/363.

Instead of using the `DiagnosticRelatedInformation` field to send actionable actions, I use `data` field in `Diagnostic`.  In ScalaCLI I create `textEdit` with actionable action and then serialized object and save into `data` field.

ckipp01: Very cool @lwronski! I'll take a look. Do you have a trace of this that I could look at to see what the entire flow looks like? I'm assuming the latest release of scala-cli isn't sending these yet?

tests/unit/src/test/scala/tests/codeactions/ActionableDiagnosticsSuite.scala
```diff
@@ -0,0 +1,42 @@
...
+import coursier.version.Version
+
+class ActionableDiagnosticsSuite
+    extends BaseCodeActionLspSuite("actionableDiagnostic") {
+
+  val oldOsLibVersion: Version = Version("0.7.8")
+
+  check(
+    "actionable-diagnostic-update",
+    s"""|//> <<>>using lib "com.lihaoyi::os-lib:${oldOsLibVersion.repr}"
```
ckipp01, line 14: What's the actual diagnostic tied to this look like?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
```diff
@@ -0,0 +1,62 @@
...
+      }
+      .toList
+      .sorted
+
+    Future.successful(codeActions)
+  }
+}
+
+object ActionableDiagnostic {
+  def title = "Apply suggestion"
```
ckipp01, line 61: I commented on it in the tests, but the diagnostic message mention what the applied suggestion will be?

lwronski: For now no, so I updated it and use [AnnotatedTextEdit](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#annotatedTextEdit) to send some more additional message about the applied suggestion.  I will update this PR after release `scala-cli` to `0.1.17`.

For dependency actionable diagnostic - in field `AnnotatedTextEdit. annotationId` I set value to: `update dependency to 0.8.1`.

Screenshot 2022-10-17 at 11 55 41

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
lwronski: Scala CLI will send the following diagnostic, but yes, it will work after release to scala-cli `0.1.17`. In latest release is missing field `annotationId` in `TextEdit`.

```
"diagnostics": [
    {
      "range": {
        "start": {
          "line": 0,
          "character": 15
        },
        "end": {
          "line": 0,
          "character": 40
        }
      },
      "severity": 4,
      "source": "scala-cli",
      "message": "com.lihaoyi::os-lib:0.7.8 is outdated, update to 0.8.1\n     com.lihaoyi::os-lib:0.7.8 -\u003e com.lihaoyi::os-lib:0.8.1",
      "data": {
        "range": {
          "start": {
            "line": 0,
            "character": 15
          },
          "end": {
            "line": 0,
            "character": 40
          }
        },
        "newText": "com.lihaoyi::os-lib:0.8.1",
        "annotationId": "update dependency to 0.8.1"
      }
    }
  ]
```

For now, I disable actionable diagnostic by default because it is still in progress and you have to pass flag. But when it will be ready, I think we can enable it by default.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
tgodzik: Annotated text edit seems a bit more complex and might not be supported by all the clients, what about doing:

`Fix: <diagnostic message>` truncated to a sensible length?

tests/unit/src/test/scala/tests/codeactions/ActionableDiagnosticsSuite.scala
```diff
@@ -0,0 +1,42 @@
...
+        |object Hello extends App {
+        |  println("Hello")
+        |}
+        |""".stripMargin,
+    ActionableDiagnostic.title,
+    expectedCode = "",
+    scalaCliOptions = List("--actions", "-S", scalaVersion),
+    expectNoDiagnostics = false,
+    scalaCliLayout = true,
+    saveResult = false,
```
tgodzik, line 25: Why not save the results? That should be perfectly fine.

tgodzik (changes requested):

tests/unit/src/test/scala/tests/codeactions/ActionableDiagnosticsSuite.scala
lwronski: Yes, I checked it again and it works with saving results.

tests/unit/src/test/scala/tests/codeactions/ActionableDiagnosticsSuite.scala
lwronski:
```
[Trace - 02:46:40 PM] Received notification 'build/publishDiagnostics'
Params: {
  "textDocument": {
    "uri": "file:///private/tmp/scala-demo/Hello.scala"
  },
  "buildTarget": {
    "uri": "file:/private/tmp/scala-demo/.scala-build/?id\u003dproject_bd2c96d2de"
  },
  "diagnostics": [
    {
      "range": {
        "start": {
          "line": 0,
          "character": 15
        },
        "end": {
          "line": 0,
          "character": 40
        }
      },
      "severity": 4,
      "source": "scala-cli",
      "message": "com.lihaoyi::os-lib:0.7.8 is outdated, update to 0.8.1\n     com.lihaoyi::os-lib:0.7.8 -\u003e com.lihaoyi::os-lib:0.8.1",
      "data": {
        "range": {
          "start": {
            "line": 0,
            "character": 15
          },
          "end": {
            "line": 0,
            "character": 40
          }
        },
        "newText": "com.lihaoyi::os-lib:0.8.1"
      }
    }
  ],
  "reset": false
}
```

tgodzik (approved):
LGTM!

tests/unit/src/main/scala/tests/TestingServer.scala
```diff
@@ -1231,7 +1231,14 @@ final case class TestingServer(
         filterAction,
       )
     } yield {
-      Assertions.assertNoDiff(codeActionString, expected)
+      val isExpectedCodeAction = codeActionString.startsWith(expected)
```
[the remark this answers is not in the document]
lwronski: Good catch!, that simplified the test nicely

ckipp01 (approved):
Great work @lwronski. Excited to see more things like this.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  override def contribute(
+      params: l.CodeActionParams,
+      token: CancelToken,
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+
+    def path =
+      params.getTextDocument.getUri.toAbsolutePath.toURI.toString
+
+    def createDiagnosticAction(
```
kpodsiad, line 23:
```suggestion
    private def createActionableDiagnostic(
```

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
```diff
@@ -0,0 +1,61 @@
...
+class ActionableDiagnostic() extends CodeAction {
+  override def kind: String = l.CodeActionKind.QuickFix
+
+  override def contribute(
+      params: l.CodeActionParams,
+      token: CancelToken,
+  )(implicit ec: ExecutionContext): Future[Seq[l.CodeAction]] = {
+
+    def path =
+      params.getTextDocument.getUri.toAbsolutePath.toURI.toString
```
kpodsiad, line 21: is `params.getTextDocument.getUri.toAbsolutePath.toURI.toString` the same as `params.getTextDocument.getUri`?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
```diff
@@ -0,0 +1,61 @@
...
+        s"Apply suggestion: ${diagMessage.split(System.lineSeparator()).headOption.getOrElse(diagMessage)}"
+      )
+      codeAction.setDiagnostics(List(diagnostic).asJava)
+      codeAction.setEdit(
+        new l.WorkspaceEdit(
+          Map(path -> List(textEdit).asJava).asJava
+        )
+      )
+      codeAction
+    }
```
kpodsiad, line 40: there should code action builder, won't it work here?

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
lwronski: You can not add private modifier here.

metals/src/main/scala/scala/meta/internal/metals/codeactions/ActionableDiagnostic.scala
lwronski: Yes, it should be the same. I fixed it.

PR #4298, opened 2022-08-25
add isStale filter in IndexedContext

jkciesluk:
Sometimes `indexedContext.scopeSymbols` gave us symbols from an old version of the file, which was a problem in some tests, like here https://github.com/scalameta/metals/blob/cf2e8c0f977b6eb34aa95b9fd0cf3176be41ba17/tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala#L56-L59

Works the same as for scala2
https://github.com/scalameta/metals/blob/cf2e8c0f977b6eb34aa95b9fd0cf3176be41ba17/mtags/src/main/scala-2/scala/meta/internal/pc/MetalsGlobal.scala#L689-L692

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
@@ -144,6 +144,18 @@ object MtagsEnrichments extends CommonMtagsEnrichments:
       upd.rawParamss = paramsWithFlags
       upd
     end withUpdatedTpe
+
+    // Returns true if this symbol is locally defined from an old version of the source file.
+    def isStale: Boolean =
+      sym.sourcePos.span.exists && {
+        val source = ctx.source
+        if source ne sym.source then
+          !source.content.startsWith(
+            sym.decodedName.toString(),
+            sym.sourcePos.span.point,
+          )
+        else false
+      }
   end extension
 
   extension (name: Name)(using Context)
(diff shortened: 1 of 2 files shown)


mtags/src/main/scala-3/scala/meta/internal/mtags/MtagsEnrichments.scala
```diff
@@ -144,6 +144,18 @@ object MtagsEnrichments extends CommonMtagsEnrichments:
...
+
+    // Returns true if this symbol is locally defined from an old version of the source file.
+    def isStale: Boolean =
+      sym.sourcePos.span.exists && {
+        val source = ctx.source
+        if source ne sym.source then
+          !source.content.startsWith(
+            sym.decodedName.toString(),
+            sym.sourcePos.span.point,
+          )
```
dos65, line 156: This line looks suspicious :thinking: 
Wouldn't it be enough to check only `source ne ctx.source` ?

tgodzik: This is copied from Scala 2 implementation, maybe we should copy over the comment? Or maybe this is no longer neccessary in Scala 3?

jkciesluk: > Wouldn't it be enough to check only source ne ctx.source ?

Thanks, its enough, at least for match completions tests

dos65: hmm. I missed that it's a different `startsWith(.., offset)`. 
Then it makes sense for me, feel free to return it back if it's needed

I thought that it was usual `startsWith` at idx 0 and didn't get what is the purpose of this clause :facepalm:

jkciesluk: Turns out it actually is needed, otherwise it marks `not-stale` symbols as `stale` so I returned it.

dos65: Yep - I saw failed tests. Sorry for non-relevant suggestion

tgodzik (approved):

Issue #1539, opened 2020-03-24
Renaming class hierarchy inside a method produces incorrect results.


tigerinus:
**Describe the bug**

**To Reproduce** Steps to reproduce the behavior:

1. Have following code ready in some Scala function
```
    sealed trait Symbol
    case class Note(name: String, duration: String, octave: Int) extends Symbol
    case class Rest(duration: String) extends Symbol

    val symbol1: Symbol = Note("C", "Quarter", 3)
    val symbol2: Symbol = Rest("Whole")
```

2. Right click on `Symbol`, select _Rename Symbol_ and give it a new name, say `S`

Got

```
    sealed trait S
    case class S(name: String, duration: String, octave: Int) extends S
    case class S(duration: String) extends S

    val symbol1: S = Note("C", "Quarter", 3)
    val symbol2: S = Rest("Whole")
```

**Expected behavior**

```
    sealed trait S
    case class Note(name: String, duration: String, octave: Int) extends S
    case class Rest(duration: String) extends S

    val symbol1: S= Note("C", "Quarter", 3)
    val symbol2: S= Rest("Whole")
```

**Installation:**

Version: 1.43.1 (user setup)
Commit: fe22a9645b44368865c0ba92e2fb881ff1afce94
Date: 2020-03-18T07:01:20.184Z
Electron: 7.1.11
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Windows_NT x64 10.0.18363

- VSCode extension version: 1.8.4
- Metals version: 0.8.3

Fabszn:
Hi, 
I going to start working on this issue. It the first for me on this project.

tgodzik:
@Fabszn Thanks! I think the issue might be that we are looking for the implementations of the class, which most likely shouldn't be needed.

My guess is that the issue is [here](https://github.com/scalameta/metals/blob/37f4c238c0cc7d83f3362f7269cf63b2c7a4175d/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala#L127)

That condition might not be valid for local symbols. We most likely need to:
- if the symbol is local then we need to check `semanticDb` for the definition
- if the signature is ClassSignature then we don't look for the implementations

Please take a look at https://scalameta.org/metals/docs/contributors/getting-started.html

Best workflow would be to create a test case in `RenameLspSuite` and run it in the background via:
[...]

PR #4304, opened 2022-08-26
bugfix: Rename properly companion object in local scope

tgodzik:
Previously, we would not correctly rename companion objects in local contexts due to the fact that it's needed to eactually search for the symbol. Now, we try to correctly identify the companion object in local context based on the current scope.

One situation this might not work when we have the same named case class in an inner scope:

```scala
def method = {
  case class Instance()
  val ins = Instance()
  def inner = {
     case class Instance()
     val ins = Instance()
  }
}
```

however this seems like a rare enough scenario to not overcomplicate things.

We would need to find all the possible same named case classes and establish boundries, which might be quite difficult.

Alternatively, if we manage to make some local renames working with presentation compiler, this will work correctly and might actually be an easier approach.

Fixes https://github.com/scalameta/metals/issues/1539

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/rename/RenameProvider.scala
@@ -6,6 +6,7 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 
 import scala.meta.Importee
+import scala.meta.Term
 import scala.meta.Tree
 import scala.meta.internal.async.ConcurrentQueue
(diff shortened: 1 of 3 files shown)


tanishiking (commented):
It looks nice, great job! but I left some comments for refactoring.

tgodzik:
I am actually wondering if we can work around it somehow, since this still doesn't cover all the cases and is quite complicated :thinking: Let me ponder a bit about this PR

tgodzik:
We can now implement the local rename provider, which will use the presentation compiler. We'll work on it for the next release.

PR #4305, opened 2022-08-27
fix: Fix docs copy paste error.

antosha417:


kpodsiad (approved):
Thanks @antosha417 for spotting this

PR #4306, opened 2022-08-29
chore: mention Scala CLi in new file provider

dos65:
A script file might be either ammonite or scala-cli.
Changed name and label just to be more explicit +an additional promotion point for scala-cli


tgodzik (approved):
LGTM!

PR #4307, opened 2022-08-29
feature: Use PC for documentHighlight

tgodzik:
Previously, we would need semanticdb to show document highlight in a file, which would require full compilation. Now, we use presentation compiler, which will work in most cases without compilation.

Since we always operate on a single file document highlight is a perfect candidate to use presentation compiler.

This can help later on with local renames since the logic will be the same.


mtags/src/main/scala-3/scala/meta/internal/pc/ScalaPresentationCompiler.scala
```diff
@@ -117,6 +118,17 @@ case class ScalaPresentationCompiler(
...
 
+  def documentHighlight(
+      params: OffsetParams
+  ): CompletableFuture[ju.List[DocumentHighlight]] =
+    compilerAccess.withNonInterruptableCompiler(
+      List.empty[DocumentHighlight].asJava,
+      params.token,
+    ) { access =>
+      val driver = access.compiler()
+      PcDocumentHiglightProvider(driver).higlights(params).asJava
```
kpodsiad, line 129: I wonder about this pattern when for every request there is a new class instantiated, wouldn't it be a perfect opportunity to use context functions here? Afterall, those functions depends on some context which is passed by constructor arguments + using this feature is a good example of doogfooding.
```scala
object PcDocumentHiglightProvider {
type WithDriver[T] = InteractiveDriver ?=> T

  def higlights(
      params: OffsetParams
  ): WithDriver[List[DocumentHighlight]] = { driver ?=>
    ???
  }
}
```

tgodzik: It could also be an object, which might be less complicated. I don't think we need context functions in this case just yet.

tgodzik (commented):
Ok, I am actually pretty happy about this. Added much more tests than we had in the semanticdb version. I also managed to identify some corner cases that semanticdb doesn't support.

I anyone has any ideas about some more tests do let me know.

One caveat is that for large files it's slower, on MetalsLanguageServer the times went up from 5-10 ms to 20 - 60 ms plus more on the first compilation. But the numbers are similar to hover (even smaller most of the time), so I don't think it's a big deal. We get a much better correctness, which we can quite easily bugfix, which in case of semanticdb is a bit harder.

mtags/src/main/scala-3/scala/meta/internal/pc/PcDocumentHiglightProvider.scala
```diff
@@ -0,0 +1,213 @@
...
+
+        val traverser = new DeepFolder[Set[DocumentHighlight]](collectNames)
+        val all = traverser(Set.empty[DocumentHighlight], unit.tpdTree)
+        all.toList.distinct
+      case None => Nil
+    end match
+  end higlights
+
+  // @note (tgodzik) Not sure currently how to get rid of the warning, but looks to correctly
+  @nowarn
```
tgodzik, line 208: I was having some issue with getting things to work correctly with Positioned and untpd.Tree, I am not sure what the hierarchy is there, but this seems to work.

Anyone else had an issue like this?

dos65 (approved):
LGTM!

I added a few comment but looks like everything is fine in general.
Nice work!

tanishiking (approved):
LGTM except for @dos65's comments!

I'm wondering, maybe we can try to highlight using SemanticDB first, and fall back to the presentation compiler if SemanticDB is not available for the sake of CPU performance. WDYT? This is just an idea, let's go with this change and see how it goes :+1:

tgodzik:
> LGTM except for @dos65's comments!
> 
> I'm wondering, maybe we can try to highlight using SemanticDB first, and fall back to the presentation compiler if SemanticDB is not available for the sake of CPU performance. WDYT? This is just an idea, let's go with this change and see how it goes +1

I was thinking about it, but it seems that the compiler can give us a much more correct view of the file. I managed to fix some issues that were not working correctly with semanticdb

kpodsiad:
> I was thinking about it, but it seems that the compiler can give us a much more correct view of the file. I managed to fix some issues that were not working correctly with semanticdb

What issues have you fixed?

tgodzik:
> > I was thinking about it, but it seems that the compiler can give us a much more correct view of the file. I managed to fix some issues that were not working correctly with semanticdb
> 
> What issues have you fixed?

Named params in annotations, some more complex types with bounds. All the issues with local symbols, which a lot of them were not working.

tgodzik:
And another thing, this will work perfectly fine even if your workspace doesn't compiler, which isn't the case with semanticdb. I think we shouldn't have actually used semanticdb for this since we only ever operate on one file, which means it's a perfect candidate for implementing using presentation compiler.

tanishiking (approved):
LGTM :tada:

mtags/src/main/scala-2/scala/meta/internal/pc/PcDocumentHiglightProvider.scala
```diff
@@ -0,0 +1,277 @@
...
+import scala.meta.pc.OffsetParams
+
+import org.eclipse.lsp4j.DocumentHighlight
+import org.eclipse.lsp4j.DocumentHighlightKind
+
+/**
+ * @param compiler
+ */
+
+final class PcDocumentHiglightProvider(
```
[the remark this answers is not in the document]
tgodzik: Fixed!

tests/cross/src/test/scala/tests/highlight/DocumentHighlightSuite.scala
```diff
@@ -0,0 +1,503 @@
...
+    "same-name4",
+    """
+      |object Test {
+      |  def foo(<<name>>: String) = ???
+      |  def bar(name: String) = ???
+      |  foo(<<na@@me>> = "123")
+      |  bar(name = "123")
+      |}""".stripMargin,
+  )
+
```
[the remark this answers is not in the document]
tgodzik: Added some more tests and fixed the issues https://github.com/scalameta/metals/pull/4307/commits/3c77e92bf6ecdaf5e5206ec42f6837be08669882

dos65 (approved):
:+1:

PR #4308, opened 2022-08-29
fix: prevent scalafmt.conf changes suggestions for scala-cli targets

dos65:
The current workflow for checking `.scalafmt.conf` doesn't work perfect
with scala-cli. It sends build-change events by its self, so any change in
`using-directives` or even just importing an empty file triggers `.scalafmt.conf` validation which might be annoying.

Also, Scala-cli has it's own mechanic for providing `.scalafmt.conf` file - https://scala-cli.virtuslab.org/docs/commands/fmt/#how-scalafmtconf-file-is-generated


dos65:
@tgodzik you can reproduce it by creating an `.sc` file in metals repo and import it as scala-cli.
You will notice that it will be detected as scala3 target and there will be scalafmt config notification.

Also, if you try to edit - add a first line with using directive, then the same notification will be shown.
```scala
//> using scala "3.1.0"
```
So, in this situation, any change in using-directives leads to this notification.

This interaction is different from other build-tools where you do a bunch of changes and  then reload build.
It think it would be better just to ignore scala-cli targets in this check.

tgodzik:
> @tgodzik you can reproduce it by creating an `.sc` file in metals repo and import it as scala-cli. You will notice that it will be detected as scala3 target and there will be scalafmt config notification.

That is a valid notification though? We have a wrong scalafmt config. I don't think we should ignore that, Metals should be able to update configuration automatically in that case.

However, I see that we actually get scalaCLI started for every file in the workspace :fearful: And it seems that we might not be handling single files properly, which both should be fixed.

> Also, if you try to edit - add a first line with using directive, then the same notification will be shown.
> 
> ```scala
> //> using scala "3.1.0"
> ```
> 
> So, in this situation, any change in using-directives leads to this notification.
> 
> This interaction is different from other build-tools where you do a bunch of changes and then reload build. It think it would be better just to ignore scala-cli targets in this check.

If you change the build, the notification should show. You can also disable that notification without an issue though.

tgodzik:
> > That is a valid notification though? We have a wrong scalafmt config. I don't think we should ignore that, Metals should be able to update configuration automatically in that case.
> 
> The issue is that an empty file is processed as an scala3 target. From user side, you haven't specified the scala version but you already received that smth is wrong.

We could instead maybe show the warning when the user runs scalafmt? So this would show up in response to running scalafmt, which could potentially break your code.

tanishiking:
I'm actually failing to reproduce the issue 😅 

> @tgodzik you can reproduce it by creating an .sc file in metals repo and import it as scala-cli.

dos65:
@tanishiking do you use the latest snapshot?

tanishiking:
Oh, I see, I was not. I'll try with the latest snapshot

PR #4309, opened 2022-08-30
docs: Add a design doc for Language Support for notebook cells

tanishiking:
It would be easier to read with rendered one :) https://github.com/scalameta/metals/blob/ccd1bed3a6eabe0d67d401be18ba348d1c494f3a/docs/design/language-support-notebook.md

Added a design document for https://github.com/scalameta/metals-feature-requests/issues/236
This document is for
- design review the architecture before going further
- make sure we are on the same page about the goal, design, and requirements for this feature

---

This PR is just for the design review. I opened a PR because it's easier to review (maybe we could write a design doc on Google Docs or something).
Therefore, I don't need to merge this PR after people approved the architecture design, but this document could be useful as a log in the future.

### DIFF
--- a/docs/design/language-support-notebook.md
+++ b/docs/design/language-support-notebook.md
@@ -1,0 +1,303 @@
+Language support for VSCode notebook
+
+# Overview
+
+We can run Scala programs on Jupyter Notebooks using a Jupyter Kernel for Scala called [Almond](https://almond.sh/).
+
+Currently, when using notebook cells in VSCode, Metals cannot provide language support such as code completion (by Metals) and code navigations.
+
+This design document describes how Metals can support many language features for the notebook environment in VSCode
+
+# Scope
+
+The scope of this document is limited to **notebook in VSCode**. It doesn't cover LSP support for jupyter notebook in general (like integration with [jupyter-lsp](https://github.com/jupyter-lsp/jupyterlab-lsp)).
+
+However, supporting the notebooks in VSCode would be a good first step for integrations with other notebook environments.
+
+## Goal
+
+- When users running Almond Kernel in VSCode, users can enjoy language features that Metals provides.
+  - All the `textDocument/*` requests should be properly handled.
+  - (Maybe we can omit some features for notebook (like `textDocument/foldingRange` may be useless in notebook) though).
+
+## Non-Goal
+
+- Provide language support for other notebook environments other than VSCode notebook (such as `jupyter-lsp`).
+- Automatically set up Almond Kernel based on the project setting
+  - Install Almond Kernel when users open a `*.ipynb` file and no Scala Kernel is available.
+- Enable Almond to use project-defined components from notebooks.
+  - If we have `A.scala` adjacent to `B.ipynb`, `B.ipynb` should be able to `import` something from `A.scala`.
+
+# Context
+
+## LSP 3.17 NotebookCellTextDocumentFilter
+
+Starting from LSP 3.17, LSP clients can recognize notebook cells as documents and request `textDocument/*` to LSP servers.
+For example, the following change unlocks the `textDocument/*` requests for jupyter-notebook (of Scala cells) in VSCode.
+
+```diff
+diff --git a/src/extension.ts b/src/extension.ts
+index 13b49f6f..32a99c36 100644
+--- a/src/extension.ts
++++ b/src/extension.ts
+@@ -291,6 +291,10 @@ function launchMetals(
+       { scheme: "file", language: "java" },
+       { scheme: "jar", language: "scala" },
+       { scheme: "jar", language: "java" },
++      {
++        notebook: { scheme: "file", notebookType: "jupyter-notebook" },
++        language: "scala",
++      },
+     ],
+     synchronize: {
+       configurationSection: "metals",
+```
+
+see: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#notebookCellTextDocumentFilter
+
+Now, notebook on VSCode starts sending the `textDocument/*` request to LSP server.
+
+```json
+[Trace - 01:09:28 PM] Received request 'textDocument/completion - (140)'
+Params: {
+  "context": {
+    "triggerKind": 2,
+    "triggerCharacter": "."
+  },
+  "textDocument": {
+    "uri": "vscode-notebook-cell:/Users/tanishiking/src/github.com/tanishiking/scala3-playground/src/main/scala/Untitled-1.ipynb#W2sZmlsZQ%3D%3D"
+  },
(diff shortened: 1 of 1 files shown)


tanishiking:
Can you take a look, and give me some thoughts? @dos65 @tgodzik @alexarchambault (requesting review to someone who commented in the original feature-requests).

docs/design/language-support-notebook.md
```diff
@@ -0,0 +1,303 @@
...
+Instead of accepting `buildTarget/compile` and generate Scala sources on notebook changes, we can let Kernel compile Scala sources (enabling SemanticDB and writing out the generated Scala sources to the filesystem).
+
+It works only if Almond Kernel can detect it's running for Metals, but it looks like impossible as mentioned in the above section.
+
+# Milestones
+
+- [ ] Implement bare-minimum Almond BSP
+  - Create a sock file in a pre-defined location.
+  - Accept `build/initialize`
+  - `workspace/buildTargets` should returns the appropriate Scala version
```
dos65, line 286: I would also include a request to execute/interpret a cell.
I don't know if we would be able to reuse existing method or it requires introduce a new one for BSP protocol.

dos65 (commented):
@tanishiking Thx for the investigations! The plan  looks reasonable

I'm only not sure if there is need in adding it to metals docs.
Maybe it might be left in a issue comment to this feature request?

tgodzik (approved):
Looks like a good plan, though might be a long term solution and will require a lot of work if I am not mistaken :thinking:

docs/design/language-support-notebook.md
```diff
@@ -0,0 +1,303 @@
...
+  - Also, Metals have to convert the incoming request parameter (for cells) into the parameter for the (converted) compilable Scala source (back and forth).
+  - Converted sources have to be on the filesystem, because Metals expect sources to be on the filesystem.
+- Build Server that recognizes `*.ipynb` as a build target.
+  - Where `ScalaBuildTarget.scalaVersion` should match with the Scala version of the running Kernel.
+  - When users open a notebook with Scala Kernel, Metals should prompt running the Build Server for the notebook, like Metals does for Ammonite script.
+
+## Components
+
+- `Almond BSP`
+  - Run on the same process with `Almond Kernel`, but the I/O will be via unix domain socket at a pre-defined location (somewhere in `~/.local/share/almond/bsp/{kernel-name}.sock`). See: https://github.com/scalameta/metals/pull/4253#issuecomment-1222091839
```
alexarchambault, line 170: Thinking a bit further about this (that I proposed in https://github.com/scalameta/metals/pull/4253#issuecomment-1222091839), I'm very slightly worried about how Metals will be able to get the PID of the Almond process corresponding to a given notebook, or more generally how Metals can find a way to connect to the right Almond kernel.

I skimmed through the [notebook section of LSP 3.17](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#notebookCellTextDocumentFilter) (that you linked above), and couldn't find any way of letting servers know about the kernel processes that the LSP client starts.

To work around that, maybe we could have hashed the notebook path, and used that hash instead of the kernel PID. But the kernel [doesn't know about the path of the notebook it is asked to run](https://jupyter-client.readthedocs.io/en/stable/messaging.html), it only knows about individual code cells passed to it via [Jupyter message protocol](https://jupyter-client.readthedocs.io/en/stable/messaging.html) messages. So hashing the notebook path isn't an option either.

Another workaround could consist in having Almond leave its PID in the metadata of some [display data](https://jupyter-client.readthedocs.io/en/stable/messaging.html#display-data) or [execute_result](https://jupyter-client.readthedocs.io/en/stable/messaging.html#id6). That might be a bit flaky though, as for the PID to be written in the notebook file, this would require at least a cell to have run. Not to add too much overhead, we might want to return those metadata only for the first cell say, but then the metadata would go away if users delete that cell… (and the kernel doesn't know about cell deletion…) So it might need to be put in all execute_result metadata.

Also, those metadata, once written on disk, might be stale (they may correspond to the kernel that previously ran the notebook).

Don't know if asking people involved in LSP to update the protocol, so that LSP servers could know about Jupyter kernels started by LSP clients, could be an option here…

alexarchambault (approved):
This looks like a very good expanded explanation and wrap up of earlier discussions about this.

I added a comment about a point that worries me a bit (hoping others could have better ideas about it…)

docs/design/language-support-notebook.md
tanishiking: >  I'm very slightly worried about how Metals will be able to get the PID of the Almond process corresponding to a given notebook,

Yeah, I think it's difficult (or even impossible) to find a Kernel PID from Metals. That's why I'm thinking maybe we can use `kernel name` or something instead of PID (more details later)

---

> To work around that, maybe we could have hashed the notebook path, and used that hash instead of the kernel PID.
> Another workaround could consist in having Almond leave its PID in the metadata of some [display data](https://jupyter-client.readthedocs.io/en/stable/messaging.html#display-data) or [execute_result](https://jupyter-client.readthedocs.io/en/stable/messaging.html#id6) (AFAIK, these are persisted in the notebook).

I think it works :+1:

> this would require at least a cell to have run.

Not really, we should be able to get kernel's metadata like on `notebookDocument/didOpen`, and we'll have access to that information without running cells

```json
 "metadata": {
  "kernelspec": {
   "display_name": "Scala",
   "language": "scala",
   "name": "scala3"
  },
  "language_info": {
   "codemirror_mode": "text/x-scala",
   "file_extension": ".sc",
   "mimetype": "text/x-scala",
   "name": "scala",
   "nbconvert_exporter": "script",
   "version": "2.13.8"
  }
 },
```

On the other hand, what do you think about using the Kernel spec name instead of embedding PID information to the name or display name? 

One concern is: is the kernel name unique?; can we say there won't be multiple same kernels running at the same time?

---

> Don't know if asking people involved in LSP to update the protocol, so that LSP servers could know about Jupyter kernels started by LSP clients, could be an option here…

I think LSP's responsibility is just to provide some interfaces between the editor(notebook) and language servers and how to connect with Kernel is out of the scope of LSP, I guess.

(kinda off-topic: the notebook LSP spec seems like the notebook content can be evaluated purely with its content (by naively concatenating cells into a script), and assuming there's no need to connect with Kernel).

tanishiking:
> I'm only not sure if there is need in adding it to metals docs.
> Maybe it might be left in a issue comment to this feature request?

As @dos65 mentioned, we don't need to merge this PR (I just submit PR because it seems easier to review), I'll turn this into an issue and closing :)

tanishiking:
https://github.com/scalameta/metals/issues/4434

PR #4311, opened 2022-08-30
fix: check_scala3_nightly - switch from bash on Scala

dos65:
I decided to finally stop fixing the exiting logic written on a bash and
replaced it by a sbt command.

### DIFF
--- a/.github/workflows/check_scala3_nightly.yml
+++ b/.github/workflows/check_scala3_nightly.yml
@@ -8,26 +8,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: coursier/setup-action@v1
+      - uses: actions/setup-java@v3
         with:
-          apps: ""
+          distribution: "temurin"
+          java-version: ${{ matrix.java }}
+          cache: "sbt"
       - name: "Query maven-central"
         run: |
-          BROKEN=broken
-          echo "3.2.0-RC1-bin-20220307-6dc591a-NIGHTLY" > $BROKEN
-          echo "3.2.0-RC1-bin-20220308-29073f1-NIGHTLY" >> $BROKEN
-
-          cs complete-dep org.scala-lang:scala3-compiler_3: | grep NIGHTLY | grep -v -f $BROKEN | sort > all_nightly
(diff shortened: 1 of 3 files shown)


.github/workflows/check_scala3_nightly.yml
```diff
@@ -8,26 +8,16 @@ jobs:
...
+          java-version: ${{ matrix.java }}
+          cache: "sbt"
       - name: "Query maven-central"
         run: |
-          BROKEN=broken
-          echo "3.2.0-RC1-bin-20220307-6dc591a-NIGHTLY" > $BROKEN
-          echo "3.2.0-RC1-bin-20220308-29073f1-NIGHTLY" >> $BROKEN
-
-          cs complete-dep org.scala-lang:scala3-compiler_3: | grep NIGHTLY | grep -v -f $BROKEN | sort > all_nightly
-          LAST_NIGHTLY_MTAG=$(cs complete org.scalameta:mtags_3 | grep NIGHTLY | grep -v -f $BROKEN | sort | tail -n 1 | cut -d "_" -f2)
```
tanishiking, line 21: [note]
This is the root cause of the issue.

- `cs complete` is now deprecated https://github.com/coursier/coursier/pull/2476, so `LAST_NIGHTLY_MTAGS` will be something like `Usage: cs List(complete) format index ...args...` (error message from coursier).
- `LAST_NIGHTLY_POS` will be `""` because, of course, the error message can't be found in the nightly versions.
- Then `LAST_NIGHTLY_POS` will be 1
- And then, `NEW_NIGHTLIES` will be all of the nightly versions.

project/Scala3NightlyVersions.scala
```diff
@@ -46,6 +35,44 @@ object Scala3NightlyVersions {
     }
   }
 
+  def nonPublishedNightlyVersions(): List[DottyVersion] = {
+    val all = fetchScala3NightlyVersions()
+    lastPublishedMtagsForNightly() match {
+      case None =>
```
tanishiking, line 41: [note]
Now, if we couldn't find the mtags versions, we will fail 👍

tanishiking (approved):
So this is the fix for the  https://discord.com/channels/632642981228314653/718047292749381662/1011148810677919824 right?

LGTM!! Well typed!

Issue #4276, opened 2022-08-18
ConvertToNamedArguments doesn't work on constructor invocations


Jasper-M:
### Describe the bug

```
Foo(42, "foo", true)
new Foo(42, "foo", true)
```
On the second line Metals doesn't provide the option to convert to named arguments.

### Expected behavior

Constructors should be able to be converted to named arguments, just like regular methods or `apply` methods.

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.8

PR #4313, opened 2022-08-31
fix ConvertToNamedArg for constructor invocations

jkciesluk:
fixes https://github.com/scalameta/metals/issues/4276

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
@@ -38,7 +38,7 @@ class ConvertToNamedArguments(trees: Trees) extends CodeAction {
     term match {
       case Some(apply: Term.Apply) =>
         getTermWithArgs(apply, apply.args)
-      case Some(newAppl @ Term.New(init)) =>
+      case Some(newAppl @ Term.New(init)) if init.argss.length <= 1 =>
         getTermWithArgs(newAppl, init.argss.flatten)
       case Some(t) => firstApplyWithUnnamedArgs(t.parent)
       case _ => None
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
@@ -19,19 +19,27 @@ class ConvertToNamedArguments(trees: Trees) extends CodeAction {
   import ConvertToNamedArguments._
   override val kind: String = l.CodeActionKind.RefactorRewrite
 
+  def getTermWithArgs(
+      apply: Term,
+      args: List[Tree],
+  ): Option[ApplyTermWithArgIndices] = {
+    val argIndices = args.zipWithIndex.collect {
+      case (arg, index)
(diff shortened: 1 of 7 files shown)


metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
```diff
@@ -18,19 +18,27 @@ class ConvertToNamedArguments(trees: Trees) extends CodeAction {
...
-          case (arg, index)
-              if !arg.isInstanceOf[Term.Assign] && !arg
-                .isInstanceOf[Term.Block] =>
-            index
-        }
-        if (argIndices.isEmpty) firstApplyWithUnnamedArgs(apply.parent)
-        else Some(ApplyTermWithArgIndices(apply, argIndices))
+        getTermWithArgs(apply, apply.args)
+      case Some(newAppl @ Term.New(init)) =>
+        getTermWithArgs(newAppl, init.argss.flatten)
```
tanishiking, line 41: It doesn't work if the constructor has multiple parameters

```scala
class Foo(x: Int, y: Int)(z: Int)

new Foo(1, 2)(3)

// If we run the code action for `Foo(...)`
new Foo(1, 2)(z = 3)
```

Maybe we should treat differently for `Term.New`.

tanishiking (changes requested):
I found a corner case bug, maybe the code action should behave differently for `Term.New`.
but overall, looks nice :+1:

metals/src/main/scala/scala/meta/internal/metals/codeactions/ConvertToNamedArguments.scala
jkciesluk: I've encountered a lot of problems while working on case when the constructor has multiple parameter lists,
from finding which parameter list to extract, to getting correct type information (in scala I get things like `dotty.tools.dotc.core.Types$UnspecifiedErrorType$@6a25de0a`, in scala2 I get nothing :sweat_smile:), so for now this code actions will not be showing for `new` with multiple parameter lists

jkciesluk: There were some issues with finding which parameter list is selected (in cases like `Foo(1,2)<<(>>3)`) so, at least for now, we are converting all parameters for `Term.New`

tests/unit/src/test/scala/tests/codeactions/ConvertToNamedArgumentsLspSuite.scala
```diff
@@ -198,4 +236,12 @@ class ConvertToNamedArgumentsLspSuite
...
     filterAction = filterAction,
   )
+
+//   checkNoAction(
+//     "new-multiple-lists",
+//     """|object Something {
+//        |  class Foo(param1: Int)(param2: Int)
+//        |  val a = new Foo(<<1>>)(2)
+//        |}""".stripMargin,
+//   )
```
tanishiking, line 246:
```suggestion

```

tanishiking (approved):
I left one nitpick comment, but LGTM! good job :tada:

tgodzik:
@jkciesluk seems `ConvertToNamedArgumentsLspSuite` are failing, could you take a look?

tgodzik (approved):
LGTM!

PR #4314, opened 2022-08-31
fix extract value for `new` keyword

jkciesluk:
Previously we couldn’t extract value from
```scala
new Car(<<"red">>)
```
This pr enables it


tanishiking (approved):
LGTM, nice catch!

tgodzik (approved):

PR #4329, opened 2022-09-01
refactor: update scala-steward configs (run more frequently and limit number of PRs)

tanishiking:
This commit makes Scala-steward to

- Run scala-steward more frequently (twice a week)
- Create only 5 PRs to scalameta/metals at a time

Previously, scala-steward runs only 1st and 15th of a month,
and steward creates a bunch of PRs (around 10~PRs at a time).
However, scalameta/metals CIs are heavy (it has 20 workflows and
each takes around 20-30 minutes), and as a result, CI stucks every
scala-steward days.

This commit makes scala-steward to run more frequently, and limit
the number of PRs steward creates at a time.
This change would limit the number of PRs steward creates and makes
CIs work smoothly for library updates.

---

What do you think about this update?

### DIFF
--- a/.github/workflows/scala-steward.yml
+++ b/.github/workflows/scala-steward.yml
@@ -1,6 +1,6 @@
 on:
   schedule:
-    - cron: '0 0 1,15 * *'
+    - cron: "0 0 * * 1, 5" # every Monday and Friday
   workflow_dispatch:
 
 name: Launch Scala Steward
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -3,3 +3,4 @@ updates.ignore = [
   { groupId = "org.eclipse.platform" }
 ]
 commits.message = "build(deps): Update ${artifactName} from ${currentVersion} to ${nextVersion}"
+updates.limit = 5


.github/workflows/scala-steward.yml
```diff
@@ -1,6 +1,6 @@
 on:
   schedule:
-    - cron: '0 0 1,15 * *'
+    - cron: "0 0 * * 1, 5" # every Monday and Friday
```
tanishiking, line 3: I set Monday and Friday randomly (maybe we can run 3 times a week?)

.scala-steward.conf
```diff
@@ -3,3 +3,4 @@ updates.ignore = [
   { groupId = "org.eclipse.platform" }
 ]
 commits.message = "build(deps): Update ${artifactName} from ${currentVersion} to ${nextVersion}"
+updates.limit = 5
```
tanishiking, line 6: Probably, 5 is still large, but if we run scala-steward frequently, it would create small number of PRs at a time.

.github/workflows/scala-steward.yml
tgodzik: I think 2 is fine :+1:

tgodzik (approved):
I think it is a good idea! Thanks!

.scala-steward.conf
ckipp01: Yea I was thinking the same thing, the more often we run it, the less we really have to worry about this.

ckipp01 (approved):
👍🏼

.github/workflows/scala-steward.yml
tgodzik:
```suggestion
    - cron: "0 0 * * 1,5" # every Monday and Friday
```
I think there might an error and no space is allowed there?

PR #4332, opened 2022-09-02
bugfix: Always set changes or document changes

tgodzik:
Previously, we would set both and a language client will choose only one of them, so if they are set and empty it will apply empty changes. Now, only one of those is set.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/codeactions/CodeActionBuilder.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/codeactions/CodeActionBuilder.scala
@@ -23,7 +23,7 @@ object CodeActionBuilder {
 
     logging.logErrorWhen(
       !(changes.nonEmpty && documentChanges.nonEmpty),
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/codeactions/CodeActionBuilder.scala
```diff
@@ -20,16 +21,23 @@ object CodeActionBuilder {
...
-    workspaceEdits.setChanges(
-      changes
-        .map { case (path, edits) =>
-          path.toURI.toString -> edits.asJava
-        }
-        .toMap
-        .asJava
+    logging.logErrorWhen(
+      !(changes.nonEmpty && documentChanges.nonEmpty),
+      "Only document or documentChanges can be set in code action at the same time",
```
jkciesluk, line 26:
```suggestion
      "Only changes or documentChanges can be set in code action at the same time",
```

jkciesluk (approved):

tanishiking (approved):
👍

PR #4334, opened 2022-09-02
bugfix: Fix wrong condition for logging in CodeActionBuilder

tgodzik:
I realized it is the opposite condition.


tgodzik:
It's a simple follow up so I will go ahead and merge it.

PR #4336, opened 2022-09-02
fix: scala-steward cron

tanishiking:
https://github.com/scalameta/metals/actions/runs/2978475788

- Use single-quote instead of double quote
- Previously we run scala-steward every hour on Monday and Friday


tgodzik (approved):
Thanks! :rocket:

PR #4340, opened 2022-09-03
feature: Add Analyze Stacktrace command to metals menu

tgodzik:
I realized that it's hard for people to find it and this is the simples way to expose it.


dos65 (approved):
:+1:

PR #4349, opened 2022-09-06
Enable enum completions in pattern matching

jkciesluk:
By refactoring `visit` method we enable completions with `enum` cases.

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
@@ -147,9 +147,11 @@ object CaseKeywordCompletion:
         && (sym.is(Case) || sym.is(Flags.Module) || sym.isClass)
         && parents.isSubClass(sym, false)
       indexedContext.scopeSymbols
-        .map(_.info.dealias.typeSymbol)
-        .filter(isValid(_))
-        .foreach(s => visit(s, s.decodedName, Nil))
+        .foreach(s => {
+          val ts  = s.info.dealias.typeSymbol
+          if (isValid(ts))
+            visit(ts, ts.decodedName, Nil)
+         )}
 
       // Step 2: walk through known subclasses of sealed types.
       MetalsSealedDesc.sealedStrictDescendants(selectorSym).foreach { sym =>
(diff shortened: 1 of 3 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -145,17 +143,22 @@ object CaseKeywordCompletion:
...
+        .map(_.info.dealias.typeSymbol)
+        .filter(isValid(_))
+        .foreach(s => visit(s, s.decodedName, Nil))
 
       // Step 2: walk through known subclasses of sealed types.
       MetalsSealedDesc.sealedStrictDescendants(selectorSym).foreach { sym =>
         val autoImport = autoImportsGen.forSymbol(sym)
         autoImport match
           case Some(value) =>
-            visit(sym.info.dealias.typeSymbol, sym.decodedName, value)
```
jkciesluk, line 156: sealedStrictDescendants are mostly already types, so this is not needed. Also, in 
```scala
enum Color:
  case Red, Blue, Green
```
symbol `Red` is a value of type `Color` so we get incorrect completions (single `Color` completion instead of 3 completions for each color

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -145,17 +143,22 @@ object CaseKeywordCompletion:
...
     else
       // Step 1: walk through scope members.
+      def isValid(sym: Symbol) = !parents.isParent(sym)
+        && (sym.is(Case) || sym.is(Flags.Module) || sym.isClass)
+        && parents.isSubClass(sym, false)
       indexedContext.scopeSymbols
-        .foreach(s => visit(s.info.dealias.typeSymbol, s.decodedName, Nil))
+        .map(_.info.dealias.typeSymbol)
+        .filter(isValid(_))
+        .foreach(s => visit(s, s.decodedName, Nil))
```
riiswa, line 152: Personally, I prefer to avoid the usage of transformers, to end with a `foreach`, and you iterate over your collection 3 times.

```suggestion
        .foreach(s => {
          val ts  = s.info.dealias.typeSymbol
          if (isValid(ts))
            visit(ts, ts.decodedName, Nil)
         )}
```

jkciesluk: Thank, you are right, it's better this way

tgodzik (approved):
LGTM

PR #4351, opened 2022-09-06
Add case completions for single case classes in Scala3

jkciesluk:
This pr adds case completions for single case classes and not sealed traits.
It's mostly for 
```scala
case class Foo(a: Int, b: Int)
List(Foo(1,2)).map { case@@} 
```


tgodzik (approved):
LGTM

PR #4353, opened 2022-09-06
Change adding space after case completions

jkciesluk:
It moves the place where we add `space` at the end of completion.
It is mostly for pattern only completions
```scala
// while triggering completions here 
List(Option(1)).map{ case So@@ => }

//previously we got double space
List(Option(1)).map{ case Some(value) $0 => }

//now 
List(Option(1)).map{ case Some(value)$0 => }
```
It also adds `space` after every case in exhaustive pattern match, not only after the first one, although I'm not sure if its correct

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
@@ -353,10 +353,13 @@ class CompletionValueGenerator(
     val label =
       if patternOnly.isEmpty then s"case $pattern =>"
       else pattern
+    val cursorSuffix =
+      (if patternOnly.nonEmpty then "" else " ") +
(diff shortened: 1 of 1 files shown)


tanishiking (approved):
LGTM!

kpodsiad:
@jkciesluk previously there was no space at all and after your changes there's one? I'm a bit confused looking at the tests, but I didn't look through how exactly test methods works in this case.

jkciesluk:
> @jkciesluk previously there was no space at all and after your changes there's one? I'm a bit confused looking at the tests, but I didn't look through how exactly test methods works in this case.

There was a space, but it was not seen in label

kpodsiad:
> There was a space, but it was not seen in label.
> 
> Inserted text for `case Some(value) => ` is still the same, I just removed space in pattern only `Some(value)`

Ok, thanks for explanation :)

tgodzik (approved):
LGTM

Issue #4352, opened 2022-09-06
Scala 3 match completion works incorrectly for enums with parameters


keynmol:
_I just want to say how excited I am for this to land in future metals release_

### Describe the bug

Reproduction is very simple, any enum will do:

![2022-09-06 16 59 03](https://user-images.githubusercontent.com/1052965/188696771-72b0fe5c-b845-497b-a035-f647dbc04fb3.gif)

Notice how `Spoon` is generated without parameters

### Expected behavior

A match is generated correctly, like in Scala 2 case:

```scala
sealed trait Test 
object Test {
  case object Yo extends Test 
  case class YoYo(i: Int) extends Test 
}

sealed trait H 
object H {
  case object Bla extends H
  case object Tla extends H
  case class Spoon(t: Test, b: Some[String]) extends H
}

// generated
def hello(h: H) = h match
  case Bla => 
  case Tla =>
  case Spoon(t, b) =>
```

### Version of Metals

0.11.8+81-94ced16c-SNAPSHOT

### Extra context or search terms

scala 3, completion, match, enum, exhaustive, parameters

tanishiking:
Thanks for reporting! Could you take a look? @jkciesluk
(btw I'm also really excited that this feature will be available in the next Metals release :tada:)

jkciesluk:
Thanks for reporting! This pr should fix it https://github.com/scalameta/metals/pull/4354

PR #4354, opened 2022-09-06
Fix case completions for enum with params

jkciesluk:
fixes https://github.com/scalameta/metals/issues/4352

Also, I haven't realised before that test for scala enum in exhaustive match was broken and it is fixed now


dos65 (approved):

keynmol:
I can confirm it works :) Excellent turnaround @jkciesluk!

PR #4356, opened 2022-09-07
Add cross tests for pattern only case completions, fix for typed case completions

jkciesluk:
Adds a cross tests suite to check if we cover all cases like `case _: @@`, `case xxx @ So@@` etc.
Also fixes sorting completions when `CaseKeywordCompletions` are not exclusive. 

Edit: I moved all extractors to one place

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -359,8 +359,8 @@ class Completions(
             indexedContext,
             config,
             parent,
-            Some(identName),
-            true,
+            patternOnly = Some(identName),
+            hasBind = true,
           ),
           false,
         )
@@ -373,7 +373,7 @@ class Completions(
             indexedContext,
             config,
             parent,
-            Some(identName),
+            patternOnly = Some(identName),
           ),
           false,
         )
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -394,11 +394,18 @@ class Completions(
       // otherwise after accepting completion we would get `case case None =>`
       case (lt @ Literal(
             Constant(null)
-          )) :: (c: CaseDef) :: (m: Match) :: parent :: _ =>
-        advancedCompletions(
-          path.tail,
-          c.startPos,
-          CompletionPos.infer(c.startPos, text, path.tail),
+          )) :: (c: CaseDef) :: (m: Match) :: parent :: _
+          if text(lt.startPos.start) == ' ' =>
+        (
+          CaseKeywordCompletion.contribute(
+            m.selector,
+            completionPos,
+            indexedContext,
+            config,
+            parent,
+            Some(""),
+          ),
+          false,
         )
 
       // class FooImpl extends Foo:
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -896,53 +896,54 @@ class Completions(
         priority(o1) - priority(o2)
       end compareInApplyParams
 
+      def prioritizeCaseKeyword(
+          sym1: CompletionValue.Symbolic,
+          sym2: CompletionValue.Symbolic,
+      ): Boolean =
+        sym1.isInstanceOf[CompletionValue.CaseKeyword] && !sym2
+          .isInstanceOf[CompletionValue.CaseKeyword]
+
       override def compare(o1: CompletionValue, o2: CompletionValue): Int =
         (o1, o2) match
-          case (
-                sym1: CompletionValue.CaseKeyword,
-                sym2: CompletionValue.Compiler,
-              ) =>
-            0
-          case (
-                sym1: CompletionValue.Compiler,
-                sym2: CompletionValue.CaseKeyword,
-              ) =>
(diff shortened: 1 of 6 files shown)


mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -347,23 +347,25 @@ class Completions(
...
             selector,
             completionPos,
             indexedContext,
             config,
             parent,
+            Some(identName),
+            true,
           ),
-          true,
+          false,
```
tgodzik, line 365: Could you use named arguments for all the booleans?

tests/cross/src/test/scala/tests/pc/CompletionPatternSuite.scala
```diff
@@ -0,0 +1,127 @@
...
+  check(
+    "wildcard",
+    """
+      |object A {
+      |  Option(1) match {
+      |    case _: @@ =>
+      |  }
+      |}""".stripMargin,
+    """|None scala
+       |Some[_] scala
```
tgodzik, line 84:
```suggestion
       |Some[?] scala
```
`_` might get deprecated at some point

tests/cross/src/test/scala/tests/pc/CompletionPatternSuite.scala
```diff
@@ -0,0 +1,127 @@
...
+  )
+  check(
+    "wildcard-ident",
+    """
+      |object A {
+      |  Option(1) match {
+      |    case _: N@@ =>
+      |  }
+      |}""".stripMargin,
+    """|None scala
```
tgodzik, line 96: None is an object, so ideally we shouldn't suggest it

tgodzik (changes requested):
Thanks for adding the test cases! That highlighted some issues we might have

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -311,59 +315,64 @@ class CompletionValueGenerator(
...
+      case Some("") => true
       case Some(query) => CompletionFuzzy.matches(query, name)
 
   def toCompletionValue(
       sym: Symbol,
       name: String,
       autoImports: List[l.TextEdit],
-  )(using Context): CompletionValue.CaseKeyword =
+  )(using Context): Option[CompletionValue.CaseKeyword] =
     sym.info
```
dos65, line 326: this line is unused

tests/cross/src/test/scala/tests/pc/CompletionKeywordSuite.scala
```diff
@@ -409,8 +409,8 @@ class CompletionKeywordSuite extends BaseCompletionSuite {
     filter = str => !str.contains("newMain"),
     compat = Map(
       "3" ->
-        """|case head :: next => scala.collection.immutable
-           |case Nil => scala.collection.immutable""".stripMargin
+        """|head :: next scala.collection.immutable
+           |Nil scala.collection.immutable""".stripMargin
```
dos65, line 413: Shouldn't these result be filtered out by fuzzy-match?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -405,12 +414,37 @@ class CompletionValueGenerator(
...
+    completionPos: CompletionPos,
+):
+  object MatchExtractor:
+    def unapply(path: List[Tree]) =
+      path match
+        // foo mat@@
+        case (sel @ Select(qualifier, name)) :: _
+            if "match".startsWith(name.toString()) && text.charAt(
+              completionPos.start - 1
+            ) == ' ' =>
```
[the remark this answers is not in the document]
jkciesluk: In the first, main PR about match case completions for scala3 we came to conclusion to show match completions after `two.match` but not `two.m`, `two.ma` etc. because I think most users aren't using this syntax and having match completion there might be confusing (also, `match` completions are exclusive, maybe we should change that?)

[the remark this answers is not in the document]
jkciesluk: I think we can make match completions not exclusive, to enable things like `xxx.match@@` and completion `xxx.matches()`, but it's better to not show match completions in `xxx.mat@@` (only for `xxx.match@@`)

tests/cross/src/test/scala/tests/pc/CompletionKeywordSuite.scala
jkciesluk: There is a problem, because `new` is not seen on path (the tree looks the same as for `case@@`. I'm looking into it, if I won't find a solution soon I will create an issue

jkciesluk: fuzzy-match isn't working, because compiler doesn't see the Ident("new")

jkciesluk: I've created an issue https://github.com/scalameta/metals/issues/4367

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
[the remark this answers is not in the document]
dos65: Ok, thx for clarifying!

dos65 (approved):

tgodzik (approved):

PR #4359, opened 2022-09-08
fix: use proper semanticdb document for `.sc` in scala-cli

dos65:
Found this issue by running locally `tests.ScaCliSuite` - in my case a test with `.sc` file was flaky.

The problem is that scalaCli provides a semanticdb db for generated file from `.sc` that has a different positions (includes generated code) that then will be indexed by metals and break go-to-definition.

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/Compilers.scala
@@ -147,6 +147,7 @@ class Compilers(
                   trees,
                   buildTargets,
                   saveSymbolFileToDisk = !config.isVirtualDocumentSupported(),
+                  sourceMapper,
                 )
               ).getOrElse(EmptySymbolSearch),
               "default",
@@ -593,6 +594,7 @@ class Compilers(
             trees,
             buildTargets,
             saveSymbolFileToDisk = !config.isVirtualDocumentSupported(),
+            sourceMapper,
             workspaceFallback = Some(search),
           )
           newCompiler(
@@ -624,6 +626,7 @@ class Compilers(
               trees,
               buildTargets,
               saveSymbolFileToDisk = !config.isVirtualDocumentSupported(),
+              sourceMapper,
             ),
             path.toString(),
             path,
--- a/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala
@@ -59,6 +59,7 @@ final class DefinitionProvider(
     buildTargets: BuildTargets,
     scalaVersionSelector: ScalaVersionSelector,
     saveDefFileToDisk: Boolean,
+    sourceMapper: SourceMapper,
 )(implicit ec: ExecutionContext) {
 
   val destinationProvider = new DestinationProvider(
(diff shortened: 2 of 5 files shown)


tanishiking:
I remember the position should be adjusted by [mappedSource](https://github.com/scalameta/metals/blob/f7e47862deb39209d5fd24f619e706af081adec9/metals/src/main/scala/scala/meta/internal/metals/Indexer.scala#L213-L246) for ScalaCLI (and Ammonite) introduced by https://github.com/scalameta/metals/pull/3790

(Actually, I don't have problem in code navigation on ScalaCLI sources). Could you elaborate on why do you think this change is needed?

dos65:
@tanishiking I don't know exactly why `test.ScalaCliSuite` works well on CI but locally without this change [this test](https://github.com/scalameta/metals/blob/734fbf48ca66ea64406976c3fc45b2e0a7335493/tests/unit/src/main/scala/tests/BaseScalaCliSuite.scala#L159-L161) fails with 99% chance.

The issue is that once we connect to bsp, during source indexing we index a generated one - https://github.com/scalameta/metals/blob/f7e47862deb39209d5fd24f619e706af081adec9/metals/src/main/scala/scala/meta/internal/metals/Indexer.scala#L496-L497

For example:
```scala
/foo.sc  //<- original source
class Foo
/.scala-build/.../foo.scala  //<- indexed source
package scripts
object foo {
/*<script>*/class Foo {
  def value = 5
}
```
This leads that when we search definition using `OnDemandSymbolIndex`  for symbol `foo.Foo#` it returns a symbol that was indexed from generated file and has a wrong position. In `bestTextDocument` the additional adjustment happens if we are able to provide semanticdb  but we need search it for `mappedSource` in scalaCli case.

tanishiking (commented):
The change looks reasonable 👍 

> locally without this change [this test](https://github.com/scalameta/metals/blob/734fbf48ca66ea64406976c3fc45b2e0a7335493/tests/unit/src/main/scala/tests/BaseScalaCliSuite.scala#L159-L161) fails with 99% chance.

Hmm, can't reproduce it in my local environment 🤔 

---

by the way, go-to-definition works in your environment? I realized I get error something

```
2022.09.14 14:16:50 ERROR code navigation does not work for the file '/Users/tanishiking/src/github.com/tanishiking/scala3-playground/scripts/test.sc' because the SemanticDB file '/Users/tanishiking/src/github.com/tanishiking/scala3-playground/scripts/.scala-build/project_183d125c5c/classes/main/META-INF/semanticdb/scripts/test.sc.semanticdb' doesn't exist. There can be many reasons for this error.
```

when I have the following file structure and tried to jump from `test.sc` to another variable also defined in `test.sc`.

```
build.sbt
scripts
`-- test.sc
src
|-- main
|   `-- scala
|       `-- test2.scala
```

maybe we also need to map the path in

https://github.com/scalameta/metals/blob/734fbf48ca66ea64406976c3fc45b2e0a7335493/metals/src/main/scala/scala/meta/internal/metals/DefinitionProvider.scala#L80-L81

dos65:
> by the way, go-to-definition works in your environment? I realized I get error something

@tanishiking oh, that is another kind of issue. 
It happens because fiules are in `scripts` and we search semanticdb assuming that `.` is sourceroot while in scala-cli it's `./scrips/`

Let's merge this one, anyway there will be more fixes for scala-cli.
Actually, I was working on https://github.com/scalameta/metals-feature-requests/issues/299 and wanted to check that I don't break the existing tests and discovered this issue.

tanishiking (approved):
Working on the issue I mentioned in another PR makes sense :+1:
LGTM on this change!

Issue #4333, opened 2022-09-02
Enable mergify against scala-steward PRs


tanishiking:
### Describe the bug

Originally proposed by @keynmol at scalameta discord, mergify should only merge `semver-patche`s, to avoid unintended breakages.

Does anyone mind auto-merging scala-steward PRs?

TODOs

- [ ] scala-steward to add labels
  - https://github.com/scala-steward-org/scala-steward/pull/2523
  - This change enables scala-steward to add Github labels like `semver-spec-patch`. We can use those labels from `mergify`.
- [ ] Enable mergify for `scalameta/metals`
- [ ] Auto-approve PRs, see: https://docs.mergify.com/actions/review/
  - PRs by scala-steward and has the label `semver-spec-patch`.
  - Tests should be passed.
  - Not sure, mergify bot's approve unlock merging 🤔 
- [ ] Auto-merge PRs by mergify.

### Version of Metals

v0.11.8

tgodzik:
I approved mergify for metals :rocket:

tgodzik:
Looks like the cron is actually invalid https://github.com/scalameta/metals/actions/runs/2978475788 @tanishiking

tanishiking:
https://github.com/scalameta/metals/pull/4336 Just a hunch, maybe we should use single quote instead of double quote?

tgodzik:
> #4336 Just a hunch, maybe we should use single quote instead of double quote?

I think it was the space, since the job stopped failing.

tanishiking:
You mean `* * * * 1, 5`? I'm actually have no idea why it's invalid 🤔

tanishiking:
Ah, no we should `0 0 * * 1,5` https://crontab.guru/#0_0_*_*_1,5
`0 0 0 * 1,5` is invalid https://crontab.guru/#0_0_0_*_1,5 :bow:

tanishiking:
Hmm, auto-approve github action fails to approve PR created by scalameta-bot labeled with `semver-spec-batch`...

https://github.com/scalameta/metals/pull/4364
https://github.com/scalameta/metals/runs/8261788560?check_suite_focus=true

with this github action https://github.com/scalameta/metals/blob/46cf3548176501e1d5b0880f6d4de8d165f76630/.github/workflows/auto-approve.yml

It looks like scala-steward does

- create a pull request
- and then add labels to pull request
[...]

PR #4365, opened 2022-09-09
fix: Detect pull_request label event for auto-approve

tanishiking:
part of https://github.com/scalameta/metals/issues/4333
Sorry for messing the actions so many times 😓 

---

Previously, the `auto-approve` code action tried to automatically approve PRs that is created by `scalameta-bot` and labeled with `semver-spec-patch` (on PR is created).
However, it didn't work in the following PR and github action

https://github.com/scalameta/metals/pull/4364
https://github.com/scalameta/metals/runs/8261788560?check_suite_focus=true

Because it seems like, there's no label information available on PR is created by `scalameta-bot`

It looks like scala-steward does

- create a pull request
- and then add labels to pull request
  - because [create pull request doesn't support label](https://docs.github.com/ja/rest/pulls/pulls#create-a-pull-request) I guess.
- That's why `github.event.pull_request.labels[*].name` will be empty.
  - Though we could label the PR when we create a PR from web UI

https://github.com/scala-steward-org/scala-steward/blob/67a1b0a182a5f1bf06de0372090231ce049644a9/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala#L190-L202

---

Alternatively, this commit change the code action to detect the `labeled` event. And it will auto-approve it if the label contains `semver-spec-patch` and the user who labeled was `scalameta-bot`.
(I believe `label-author` should be `scalameta-bot`. Otherwise, all the people will be able to approve any pull request via attaching `semver-spec-patch` label to any PRs).

### DIFF
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -2,6 +2,7 @@ name: Auto approve patch updates
 
 on:
   pull_request_target:
+    types: [labeled]
 
 jobs:
   auto-approve:


dos65:
@tanishiking now the changes is that requirements for a trigger action were relaxed but at the moment when the step will be executed the required label will be available? Or you don't know it yet and you want to try?

tanishiking:
> at the moment when the step will be executed the required label will be available? 

what do you mean?

The label will be available on `pull_request.labeled` event, but I found that it's not yet available on `pull_request.created`.

dos65:
> what do you mean?
> The label will be available on pull_request.labeled event, but I found that it's not yet available on pull_request.created.

Oh, that a different type of event. Got it.

dos65 (approved):

tanishiking:
Thanks!

Issue #4358, opened 2022-09-08
[Scala3] Refactor type completions to provide both type symbol and object


tanishiking:
## TODOs
- [ ] Show both `List` and `List[$0]` as completions
- [ ] Prioritize `List[$0]` in the place where users should provide type parameters

### Describe the bug

Currently, type completion (in Scala3) filters out "non-compilable completions"

For example, for `val x: scala.IndexedSe@@`, Scala3 complete `IndexedSeq[$0]` (for `type IndexedSeq[T]`), but doesn't return `IndexedSeq` (for `object IndexedSeq`) as Scala2 does.

https://github.com/scalameta/metals/blob/46cf3548176501e1d5b0880f6d4de8d165f76630/tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala#L125-L141

Scala3 type completion is done this by filtering out those symbols based on the completion position in [CompletionPos.include](https://github.com/scalameta/metals/blob/46cf3548176501e1d5b0880f6d4de8d165f76630/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala#L74-L104)

However, we may want to stop filtering out those symbols for two reasons

- Filtering out based on completion position is error prone
  - All the filtering logic is based on our heuristics, and it's so hard to cover all the cases correctly.
  - Actually, original type completion (for Scala3) caused a regression (and fixed in the `main` branch) https://github.com/scalameta/metals/pull/4234
- I believe `IndexedSeq` completion is still valuable (while `IndexedSeq[$]` is also good)
  - I think completions provide just some help to developers, they don't need to be "semantically correct".

### Expected behavior

`def foo: scala.IndexedSe@@` completes both ``IndexedSeq[$0]` and `IndexedSeq`.

optionally, we may want to prioritize `IndexedSeq[$0]` based on the cursor position.

### Operating system

macOS

### Version of Metals

0.11.8+87-46cf3548-SNAPSHOT

### Extra context or search terms

type completion, scala3

tanishiking:
Also, I realized in Scala3, we filter out completions like `ListBuffer[A](elems: A*): CC[A]` are filtered based on completion positions.
However, I don't think they're never useful, and maybe we can filter out them based on symbols.

tanishiking:
kinda related https://github.com/scalameta/metals/issues/4250

PR #4369, opened 2022-09-09
[WIP] fix: [scala3] type complete both symbol with and without snippet suffix

tanishiking:
part of https://github.com/scalameta/metals/issues/4358

- This PR changes type completion to complete both `IndexedSeq[$0]` and `IndexedSeq` for `val x: scala.IndexedSeq@@` in Scala3.
  - What do you think about this behavior? Should we complete only `IndexedSeq[$0]`?
- If we complete both items, how should we differentiate them by labels?

Screen Shot 2022-09-09 at 21 36 07

↑ current implementation ↑
(former one: `ListBuffer`, and latter one is `ListBuffer[$0]`

---

Currently, type completion (in Scala3) filters out "non-compilable completions"

For example, for `val x: scala.IndexedSe@@`,
Scala3 complete `IndexedSeq[$0]` (for `type IndexedSeq[T]`), but doesn't return `IndexedSeq` (for `object IndexedSeq`).

https://github.com/scalameta/metals/blob/46cf3548176501e1d5b0880f6d4de8d165f76630/tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala#L125-L141

This commit enables Metals (in Scala3) to complete both `IndexedSeq[$0]` and `IndexedSeq`.

**why we want this change?**
Scala3 type completion is done this by filtering out those symbols based on the completion position in
[CompletionPos.include](https://github.com/scalameta/metals/blob/46cf3548176501e1d5b0880f6d4de8d165f76630/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala#L74-L104)

However, we may want to stop filtering out those symbols for two reasons

- Filtering out based on completion position is error prone
  - All the filtering logic is based on our heuristics, and it's hard to cover all the cases correctly.
  - Actually, original type completion (for Scala3) caused a regression (and fixed in the `main` branch) https://github.com/scalameta/metals/pull/4234
- I believe `IndexedSeq` completion is still valuable (while `IndexedSeq[$]` is also good)
  - I think completions provide just some help to developers. They don't need to be "semantically correct".

**concern and plan**
- (concern) Differentiate labels of the symbols with and without snippet suffix
  - Currently, both completion `IndexedSeq` and `IndexedSeq[$0]` has the same label (completion kinds are different, though). It would be ideal if we could show different labels for them.
- (plan) Sort those symbols prioritizing the one with snippet suffix (or based on the cursor position).

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
@@ -71,38 +71,6 @@ class Completions(
     case Term
     case Import
 
-    def include(sym: Symbol)(using Context): Boolean =
-      val generalExclude =
-        isUninterestingSymbol(sym) ||
-          !isNotLocalForwardReference(sym) ||
-          sym.isPackageObject
-
-      if generalExclude then false
-      else
-        this match
-          case Type(_, _) if sym.isType => true
-          case Type(_, _) if sym.isTerm =>
-            /* Type might be referenced by a path over an object:
-             * ```
-             * object sample:
-             *    class X
-             * val a: samp@@le.X = ???
-             * ```
-             * ignore objects that has companion class
-             *
-             * Also ignore objects that doesn't have any members.
-             * By some reason traits might have a fake companion object(example: scala.sys.process.ProcessBuilderImpl)
-             */
-            val allowModule =
-              sym.is(Module) &&
-                (sym.companionClass == NoSymbol && sym.info.allMembers.nonEmpty)
-            allowModule
-          case Term if sym.isTerm || sym.is(Package) => true
-          case Import => true
-          case _ => false
-      end if
-    end include
-
     def allowBracketSuffix: Boolean =
       this match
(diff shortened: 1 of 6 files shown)


kpodsiad:
> Differentiate labels of the symbols with and without snippet suffix
> * Currently, both completion `IndexedSeq` and `IndexedSeq[$0]` has the same label (completion kinds are different, though). It would be ideal if we could show different labels for them.

@tanishiking is it possible to show different labels for them? If yes, what about `IndexedSeq` and `IndexedSeq[T]` respectively?

tanishiking:
Yep, that's pssible.

> what about IndexedSeq and IndexedSeq[T] respectively?

looks good! I'll take a look how can we do this

dos65:
> maybe we should refactor `findSuffix` to return some kind of data structure instead of the string before working on this so that we can tell what kinds of suffix will be added to the edit.

@tanishiking I think that is the only way. Otherwise it will require to duplicate this logic to build a label with the suffix one more time

tanishiking:
Closing for now, I'm going to re-work based on https://github.com/scalameta/metals/pull/4412 :+1:

Issue #2547, opened 2021-02-24
Build import is not re-tried if VSCode is restarted while build import prompt was open


smarter:
- Open with vscode a simple sbt project (e.g. https://github.com/scala/scala3-example-project)
- Wait for the metals popup window about build import to appear
- Restart VSCode
- Metals doesn't prompt for build import anymore, in the logs I see:
```
2021.02.24 14:46:45 WARN  Build server is not auto-connectable.
2021.02.24 14:46:45 WARN  no build target for: /home/smarter/opt/cs206-martres/src/main/scala/scalashop/HorizontalBoxBlur.scala
2021.02.24 14:46:46 INFO  skipping build import with status 'Requested'
2021.02.24 14:46:49 INFO  no build target: using presentation compiler with only scala-library: 2.12.13
 ```

ckipp01:
Thanks for the report! So this is sort of expected behavior. If we prompt a user to import and they cancel the prompt, we mark it as "requested" and then don't prompt again right away. (we will eventually). When you restart while the prompt was never answered, it's basically the same thing as we know we requested, but never got a response. There probably isn't an easy way for us to know if the request was cancelled manually, or cancelled via the restart.

smarter:
IMO, you should only disable the prompt if the user explicitly clicked on the "don't ask me again" button, and in that case you can register that as a setting or something. In all other cases we should assume that the user wants his build to be imported, because metals will not work properly without that (in fact, even if they click on "don't ask me again", it would be nice to still have an indicator in the status bar that something wrong, such that clicking it would display the prompt again, but that's less important).

ckipp01:
So the main reason for doing in the way we have is so it doesn't spam the user if they are making continuous changes in the build definition as it's quite annoying if you make a change and save, you get a prompt, then another small change and save, another prompt. Since an import is expensive, most of the time the use wants to wait until they are finished, and then import.

The indication that an import is needed though is a good idea. I'm curious how we could achieve this is an editor agnostic way.

smarter:
> if you make a change and save, you get a prompt, then another small change and save, another prompt.

Hmm I see, but that shouldn't prevent the prompt from reappearing when restarting then, it's only about the current editing session.

PR #4373, opened 2022-09-12
Request import build after restart

jkciesluk:
Connected to https://github.com/scalameta/metals/issues/2547

Previously, reloading editor while import build prompt was open resulted in prompt not reappearing. Now the prompt reappears, but only if it was not answered. If user closes it or chooses any option it works the same as before (closing dismisses it for 2 minutes).


tgodzik (approved):
LGTM!

Issue #2151, opened 2020-10-17
Limit the concurrent running `import build`


hepin1989:
![image](https://user-images.githubusercontent.com/501740/96334724-7b58d680-10a5-11eb-8a26-1f7f262b501d.png)
After I click the `Import build` multi times.

tgodzik:
Thanks for reporting! I think we could register currently running process and not rerun in case it's still running.

ckipp01:
> Thanks for reporting! I think we could register currently running process and not rerun in case it's still running.

Do you think it'd be best to let the process finish and then run the second one, or simply cancel the first one on the second trigger, and then just runt he second one?

tgodzik:
I think it's better to let the current one finish. I assume if user would really want to rerun they would first cancel the current request.

pvid:
A colleague of mine encountered the same thing when trying out metals - it is confusing especially to newcomers. He was using gradle, the issue is not sbt specific.

By the way, clicking `Import build` multiple times should trigger some kind of popup or message with "Import already running. Please cancel the current import to run a new one" or something to that spirit. Otherwise the user could think that metals is just unresponsive

PR #4375, opened 2022-09-12
Limit running multiple import build

jkciesluk:
Connected to https://github.com/scalameta/metals/issues/2151

Now we can't force multiple build import using `Import Build` command

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -1756,10 +1756,13 @@ class MetalsLanguageServer(
         generateBspConfig().asJavaObject
       case ServerCommands.ImportBuild() => 
         if (!isImportInProcess) {
-          slowConnectToBuildServer(forceImport = true).asJavaObject
+          isImportInProcess = true
+          val buildChange = slowConnectToBuildServer(forceImport = true)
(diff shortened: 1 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -170,7 +170,8 @@ class MetalsLanguageServer(
   private val recentlyOpenedFiles = new ActiveFiles(time)
   private val recentlyFocusedFiles = new ActiveFiles(time)
   private val languageClient = new DelegatingLanguageClient(NoopLanguageClient)
-
+  @volatile
+  private var isImportInProcess: Boolean = false
```
kpodsiad, line 174: If we might end up in situation when two or more threads can modify the same variable, shouldn't `AtomicRef` (in this case `AtomicBoolean`) and CAS loops be used? At least I'm always going with Atomic* whenever I had to share some mutable value.

tgodzik: Ach yeah, we might need an AtomicBoolean or something. That's my bad, I suggested volotile

kpodsiad: Just an idea: maybe other usages of volatile should be rethinked?

tgodzik (approved):
LGTM!

PR #4376, opened 2022-09-13
Add Completions for ivy Imports (Ammonite)

LaurenceWarne:
Hi, I've been playing around with Ammonite ivy import completions in metals (heavily inspired by https://github.com/com-lihaoyi/Ammonite/pull/957) using the awesome coursier api, demo:

https://user-images.githubusercontent.com/17688577/189854045-ea525c33-29a3-40d3-935f-56857652366f.mp4

It works very similarly to the existing completions for `$file`s, essentially we just ask `coursierapi.Complete` for completions.  Additionally, when we're at `import $ivy.|` we return a snippet to handle backticks, and when we're at `group:|` we make sure to get scala verion-specific (e.g. `%%`) completions also (hopefully these are both clear from the video).

There are a few TODOS/stuff I'm not 100% about:
- Scala 2 only atm
- ~the Scala version used by the coursier api is currently hard-coded (I'm unsure the best way to wire through the value for where I need it?) as opposed to deducing from file comments like `Ammonite.command` does,~ along with the repositories
- Often the number of completion items returned is a lot (1k+)
- not using `coursierapi.Cache` (but performance appears ok without it)

If you're all happy with this, I'd prefer to work incrementally on the above so as not to make one huge PR :slightly_smiling_face: .  Anyway, let me know what you think, thanks!

### DIFF
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/AmmoniteCompletions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/AmmoniteCompletions.scala
@@ -1,10 +1,12 @@
 package scala.meta.internal.pc.completions
 
+import scala.collection.JavaConverters._
 import scala.util.control.NonFatal
 
 import scala.meta.internal.mtags.MtagsEnrichments._
 import scala.meta.internal.pc.CompletionFuzzy
 import scala.meta.internal.pc.MetalsGlobal
+import scala.meta.internal.tokenizers.Chars
 import scala.meta.io.AbsolutePath
 
 import org.eclipse.{lsp4j => l}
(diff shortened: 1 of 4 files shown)


mtags/src/main/scala-2/scala/meta/internal/pc/completions/AmmoniteCompletions.scala
```diff
@@ -106,6 +108,72 @@ trait AmmoniteCompletions { this: MetalsGlobal =>
...
+        val query = selector.collectFirst {
+          case sel: ImportSelector if sel.name.toString().contains(CURSOR) =>
+            sel.name.decode.replace(CURSOR, "")
+        }
+
+        query match {
+          case Some(imp) =>
+            val api = coursierapi.Complete
+              .create()
+              .withScalaVersion("2.13.8")
```
tgodzik, line 130:
```suggestion
              .withScalaVersion(BuildInfo.scalaCompilerVersion)
```

mtags/src/main/scala-2/scala/meta/internal/pc/completions/AmmoniteCompletions.scala
```diff
@@ -106,6 +108,72 @@ trait AmmoniteCompletions { this: MetalsGlobal =>
...
+              .withScalaBinaryVersion("2.13")
```
tgodzik, line 131:
```suggestion
```
is this needed? Can't coursier infer it itself?

We could also use `ScalaVersions.scalaBinaryVersionFromFullVersion` or just inline that method.

mtags/src/main/scala-2/scala/meta/internal/pc/completions/AmmoniteCompletions.scala
```diff
@@ -106,6 +108,72 @@ trait AmmoniteCompletions { this: MetalsGlobal =>
...
+
+            def completions(s: String): List[String] =
+              api.withInput(s).complete().getCompletions().asScala.toList
+            val javaCompletions = completions(imp)
+            val scalaCompletions =
+              if (imp.endsWith(":") && imp.count(_ == ':') == 1)
+                completions(imp + ":").map(":" + _)
+              else List.empty
```
tgodzik, line 139:
```suggestion
              if (imp.endsWith(":") && !imp.endsWith("::") )
                completions(imp + ":").map(":" + _)
              else List.empty
```

tgodzik (changes requested):
Great work! Some small comments, but otherwise this is almost good to merge.

We could potentially also support scala-cli in a very similar way.

mtags/src/main/scala-2/scala/meta/internal/pc/completions/AmmoniteCompletions.scala
LaurenceWarne: Yeah that makes sense, but I think it's necessary atm, see: https://github.com/coursier/interface/pull/274 so I've inlined it.

mtags/src/main/scala-2/scala/meta/internal/pc/completions/AmmoniteCompletions.scala
LaurenceWarne: Hmmm I think the suggestion will prompt an extra api call at for example 

```scala
import $ivy.`org.scalameta::metals:|`
```
?

tgodzik: ach ok then, I didn't realize it was also in that case. nvm!

mtags/src/main/scala-2/scala/meta/internal/pc/completions/AmmoniteCompletions.scala
```diff
@@ -106,6 +109,76 @@ trait AmmoniteCompletions { this: MetalsGlobal =>
...
+      try {
+
+        val query = selector.collectFirst {
+          case sel: ImportSelector if sel.name.toString().contains(CURSOR) =>
+            sel.name.decode.replace(CURSOR, "")
+        }
+
+        query match {
+          case Some(imp) =>
+            // TODO grab version from Ammonite file header if it exists
```
tgodzik, line 129: Actually, this will be the same version `BuildInfo.scalaCompilerVersion` as we get info about it from the compiler

LaurenceWarne: Ah nice!

tgodzik: From Ammonite - this is what I wanted to write :sweat_smile:

tgodzik (approved):
Great work!

LaurenceWarne:
Thanks!

Issue #4216, opened 2022-08-02
Delete files doesn't run compilation


tanishiking:
### Describe the bug

Create two files

```scala
// A.scala
package example
object A:
  def aaa = B.bbb

// B.scala
package example
object B:
  def bbb = ???
```

and then delete `B.scala`.

### Expected behavior

Run compilation and receive an error in `A.scala` (`Not found: B`).

---

I guess this is because we don't run compilation when Metals detect the file deletion https://github.com/scalameta/metals/blob/d2ab724848b0c8b5c066694e5cf431c809715d34/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1410-L1424 Is this an intentional behavior?

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.7

### Extra context or search terms

FileWatcher delete compile

tgodzik:
Thanks for finding this! I think we should actually invoke the compilation in such case

tgodzik:
Great to hear! We would probably need to add additional method:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didDeleteFiles

These are represented in MetalsLanguageServer as https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L1249

We would need to add a method also to [Compilations](https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/Compilations.scala) that would check which build target the file belonged to and then invoke the compileTarget method.

I think this should do it, let me know if this is clear enough :smiley:

tanishiking:
@PeuTit Thank you for your interest!

## some context
- Metals has a FileWatcher that observes file modification, creation, and deletion under the projects
  - [FileWatcher.scala](https://github.com/scalameta/metals/blob/c80f730d27092e6b1c51552c90bc9c7d5ec05421/metals/src/main/scala/scala/meta/internal/metals/watcher/FileWatcher.scala#L37)
- `MetalsLanguageServer` instantiate the file watcher, passing `didChangeWatchedFiles` as a call back function.
  - https://github.com/scalameta/metals/blob/c80f730d27092e6b1c51552c90bc9c7d5ec05421/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala#L215-L221
[...]

PR #4377, opened 2022-09-13
Compile project on file deletion

PeuTit:
- Update the branch that handle file deletion on the `didChangeWatchedFiles`
- Create a new `onDelete` function that compiles the project

#4216


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -1452,6 +1453,16 @@ class MetalsLanguageServer(
       .ignoreValue
   }
 
+  private def onDelete(paths: Seq[AbsolutePath]): Future[Unit] = {
+    Future
+      .sequence(
+        List(
+          compilations.compileFiles(paths)
```
tanishiking, line 1460: Maybe you can move `diagnostics.didDelete` to here :)

```suggestion
          compilations.compileFiles(paths),
          diagnostics.didDelete(path)
```

tanishiking (approved):
I tested it in my local environment, and it works really well, great job!
I left one minor comment, but it's LGTM :tada:

PR #4383, opened 2022-09-15
fix: Scala3 properly inverse semanticdb type symbol

dos65:
It was affecting woskpace completions: type symbol were converted into a term one and then filtered out.

### DIFF
--- a/tests/unit/src/main/scala/tests/BaseCompletionLspSuite.scala
+++ b/tests/unit/src/main/scala/tests/BaseCompletionLspSuite.scala
@@ -84,6 +84,11 @@ abstract class BaseCompletionLspSuite(name: String) extends BaseLspSuite(name) {
            |object A {
            |  // @@
            |}
+           |/a/src/main/scala/a/inner/FooSample.scala
+           |package a.sample
+           |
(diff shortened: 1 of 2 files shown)


tests/unit/src/main/scala/tests/BaseCompletionLspSuite.scala
```diff
@@ -172,13 +177,9 @@ abstract class BaseCompletionLspSuite(name: String) extends BaseLspSuite(name) {
       )
       _ <- assertCompletion(
         """
-          |locally {
```
tgodzik, line 175: Could you add another `assertCompletion` instead of changing this one?

tgodzik (changes requested):

tests/unit/src/main/scala/tests/BaseCompletionLspSuite.scala
dos65: oops, that how wasn't intended

dos65: fixed

PR #4384, opened 2022-09-15
bugfix: Always check if import is in progress

tgodzik:
Previously, we would only check if import was run manually and not check if it was run via message request. Now, we should not run import if the one from message request is already running.


dos65 (approved):
:+1:

PR #4385, opened 2022-09-15
chore: rename CompletionProvider

dos65:
Most of similar files in mtags2/3 have the same name. This additional `s` often annoys me when I do file search in VSCode


tanishiking (approved):
LGTM, nice catch! I was also aware of that, but I was too lazy to fix it :sweat: thank you!

PR #4391, opened 2022-09-16
bugfix: Fix tests using circe for completions

tgodzik:
PReviously, no circe native was avaialable. Now. circe being also being released for native, so we need to adjust the test.


tanishiking (approved):
LGTM :+1:

dos65 (approved):

Issue #4374, opened 2022-09-12
Env variables not propogated to tests


Quafadas:
### Describe the bug

After some chatter with @kpodsiad on discord; 

I had the hope that this launch file would start tests, with access to environment variables defined in the ".env-test" file.
```
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "scala",
            "request": "launch",
            "name": "Launch Test",
            "testClass": "example.EventSuite",            
            "envFile" : ".env-test"
        },
    ]
}
```
Which I do not believe it does. 

It is not clear, that this is a "bug" per se, as it is likely expected behaviour at the moment.

>  About envs, it seems like they're supported only for main classes

It was suggested to raise this as an issue though. 

### Expected behavior

Ideally, environment variables would be propogated to tests.

### Operating system

Windows

### Editor/Extension

VS Code

### Version of Metals

v0.11.8

### Extra context or search terms

This is commentary from Kamil, which I do not pretend to understand that well, 

I think it should be an issue. Some context: envs are supported only for main classes because BSP hasn't provide such capability for test classes until https://github.com/build-server-protocol/build-server-protocol/pull/304. 

But nobody hasn't switched to that version in Metals yet

https://github.com/scalameta/metals/blob/cfb33b0badce29bedaf08020154f38d97d125d77/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala#L497-L531 

this should be improved

PR #4393, opened 2022-09-16
bugfix: Allow to add jvmopts and env variables when running tests

tgodzik:
Previously, only the test suite name was sent via BSP. Now we also add env variables and jvmopts.

Fixes https://github.com/scalameta/metals/issues/4374

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
@@ -650,6 +650,9 @@ final case class ScalaTestSuiteSelection(
 case class DebugUnresolvedTestClassParams(
     testClass: String,
     @Nullable buildTarget: String = null,
+    @Nullable jvmOptions: java.util.List[String] = null,
+    @Nullable env: java.util.Map[String, String] = null,
+    @Nullable envFile: String = null,
 )
 
 case class DebugUnresolvedAttachRemoteParams(
--- a/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
@@ -231,6 +231,18 @@ class DebugProvider(
     buildTargets.buildServerOf(targetId).exists(_.supportsTestSelection)
   }
 
+  private def envFromFile(
+      envFile: Option[String]
+  )(implicit ec: ExecutionContext): Future[List[String]] =
+    envFile
+      .map { file =>
+        val path = AbsolutePath(file)(workspace)
+        DotEnvFileParser
+          .parse(path)
+          .map(_.map { case (key, value) => s"$key=$value" }.toList)
+      }
+      .getOrElse(Future.successful(List.empty))
+
   private def createMainParams(
       main: ScalaMainClass,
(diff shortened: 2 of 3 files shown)


metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -650,6 +650,9 @@ final case class ScalaTestSuiteSelection(
 case class DebugUnresolvedTestClassParams(
     testClass: String,
     @Nullable buildTarget: String = null,
+    @Nullable jvmOptions: java.util.List[String] = null,
+    @Nullable env: java.util.Map[String, String] = null,
+    @Nullable envFile: String = null,
```
dos65, line 655: Just a question.

Is it bloop only?
I haven't found these params in build-server-protocol

dos65 (approved):

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
kpodsiad: no, it's both bloop and sbt (all build servers which support debugging) - https://github.com/build-server-protocol/build-server-protocol/pull/304

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
```diff
@@ -515,13 +516,23 @@ class DebugProvider(
             "test",
           )
         }
-        Future.successful(
+        val env =
+          if (params.env != null) createEnvList(params.env) else Nil
```
kpodsiad, line 520:
```suggestion
        val env = Option(params.env).toList.flatMap(createEnvList)
```
maybe?

kpodsiad (approved):
LGTM
btw, it's not really a bugfix, more like a improvement ;)

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
dos65: thx for explanation!

metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala
tgodzik: Done, also in other places. Thanks!

kpodsiad:
Hmm, I wonder if this logic should also be applied to the requests from the Test Explorer. 
https://github.com/scalameta/metals/blob/42c5b16b38945a576aaad89e8f01f14f5dc2d30e/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala#L568-L592

tgodzik:
> Hmm, I wonder if this logic should also be applied to the requests from the Test Explorer.
> 
> https://github.com/scalameta/metals/blob/42c5b16b38945a576aaad89e8f01f14f5dc2d30e/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProvider.scala#L568-L592

Can you set addtional options for Test Explorer?

kpodsiad:
> Can you set addtional options for Test Explorer?

There is some mechanism for configuration/options but currently vscode doesn't use it. Sounds like an improvement.

PR #4394, opened 2022-09-16
bugfix: Fix document highlight for type projections

tgodzik:


mtags/src/main/scala-2/scala/meta/internal/pc/PcDocumentHighlightProvider.scala
```diff
@@ -283,6 +283,15 @@ final class PcDocumentHighlightProvider(
                 case (highlights, _) =>
                   highlights
               }
+            // catch all missed named trees
+            case name: NameTree if sought(name.symbol) && name.pos.isRange =>
```
tgodzik, line 287: Also added this to handle Bind, it should catch all the name trees that we forgot about.

tanishiking (approved):
LGTM! :tada:

PR #4396, opened 2022-09-16
fix: [Scala3] Don't compute docs on textDocument/completion

tanishiking:
Previously, Metals compute docstrings of the method using `symbolSearch.symbolDocumentation` from `defaultMethodSignatures`.

This is for two reasons:
- replace java parameters such as `x$1` with the one from the document
- to print default parameters

However, it will be covered by `documentItem/resolve` so we don't need to compute them on `textDocument/completion`.
Also, I profiled Metals and realized `symbolSearch.symbolDocumentation` takes much of the time of code completion. This change may significantly improve the response time of `textDocument/completion`, especially in the large repositories.

---

(sorry for the large screen shot of `async-profiler`)

Screen Shot 2022-09-19 at 10 06 40

paramsDocs part will disappear

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/printer/MetalsPrinter.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/printer/MetalsPrinter.scala
@@ -180,7 +180,7 @@ class MetalsPrinter(
                 param,
                 implicitEvidencesByTypeParam,
                 index,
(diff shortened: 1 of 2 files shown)


tanishiking:
Let me see how CI reacts to this change

tgodzik (approved):
LGTM!

PR #4397, opened 2022-09-16
fix: scala3 match-case completions - improve autoimports

### LINKED ISSUE #4368: Inconsistent case completions with conflicts
### Describe the bug

Case completions in scala 3 don't add prefix when conflict with another symbol occurs. Triggering completions in
```scala
object Hello extends App{
  val e: Either[Int, String] = ???
  type Left = String
  e match {
    case Le@@ =>
  }
}
```
results in
```scala
object Hello extends App{
  val e: Either[Int, String] = ???
  type Left = String
  e match {
    case Left(value) =>
  }
}
```

### Expected behavior

In scala 2 we get
```scala
object Hello extends App{
  val e: Either[Int, String] = ???
  type Left = String
  e match {
    case scala.util.Left(value) =>
  }
}
```

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

v0.11.8

_No response_

dos65:
Fixes #4368
Also, ports cases ordering logic from Scala2


mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -202,6 +220,41 @@ object AutoImports:
             case _ => None
         case IndexedContext.Result.InScope => None
 
+    def inferSymbolImport(symbol: Symbol): SymbolImport =
```
dos65, line 223: This PR adds some duplications. `SymbolImport` is an extended form of `AutoImport` that allows to customize import section edits logic.

Previously, we were adding only one import per action but for exhaustive match completions we need to add several at once.
I'm going to get rid of `AutoImport` in other places but I think it would be better to merge it as it is as this change blocks @jkciesluk

jkciesluk:
> Fixes https://github.com/scalameta/metals/issues/4368

I don't see any change there, it works the same as before

jkciesluk:
> Fixes https://github.com/scalameta/metals/issues/4368

I don't see any change there, it works the same

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -158,7 +153,7 @@ object CaseKeywordCompletion:
         case _ =>
           if !(selectorSym.is(Sealed) &&
               (selectorSym.is(Abstract) || selectorSym.is(Trait)))
-          then visit(selectorSym, selectorSym.decodedName, Nil)
+          then visit(SymbolImport.simple(selectorSym))
```
jkciesluk, line 156:
```suggestion
          then visit(autoImportsGen.inferSymbolImport(ts))
```

> >Fixes https://github.com/scalameta/metals/issues/4368

> It is fixed in exhaustive match, but in single case completions it works the same

This should fix it

jkciesluk (changes requested):

mtags/src/main/scala-3/scala/meta/internal/pc/completions/MatchCaseCompletions.scala
```diff
@@ -168,55 +163,63 @@ object CaseKeywordCompletion:
       indexedContext.scopeSymbols
         .foreach(s =>
           val ts = s.info.dealias.typeSymbol
-          if (isValid(ts)) then visit(ts, ts.decodedName, Nil)
+          if (isValid(ts)) then visit(SymbolImport.simple(ts))
```
jkciesluk, line 166:
```suggestion
          if (isValid(ts)) then visit(autoImportsGen.inferSymbolImport(selectorSym))
```
Maybe also here?

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -79,6 +79,24 @@ object AutoImports:
       }.toMap
   end AutoImport
 
+  sealed trait ImportSel:
+    def sym: Symbol
+
+  object ImportSel:
+    final case class Direct(sym: Symbol) extends ImportSel
```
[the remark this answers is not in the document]
dos65: Upd: removed the duplication. Now old methods are based on `inferSymbolImport`.

mtags/src/main/scala-3/scala/meta/internal/pc/IndexedContext.scala
```diff
@@ -12,6 +12,7 @@ import dotty.tools.dotc.core.Names.*
 import dotty.tools.dotc.core.Symbols.*
 import dotty.tools.dotc.core.Types.*
 import dotty.tools.dotc.typer.ImportInfo
+import dotty.tools.dotc.core.Flags
```
jkciesluk, line 15: Scalafix is failing because of it, I think we can remove this import since it's not used

jkciesluk (approved):
LGTM! Great work!

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
```diff
@@ -160,77 +160,118 @@ object AutoImports extends AutoImportsBackticks:
...
-      val editPos = pos.withSpan(Spans.Span(importPosition.offset)).toLsp
-      new l.TextEdit(editPos, formatted)
-    end importEdit
+              val importSel =
+                if rename != owner.showName then
+                  Some(ImportSel.Rename(owner, rename)).filter(_ =>
+                    !indexedContext.hasRename(owner, rename)
+                  )
+                else
+                  Some(ImportSel.Direct(owner)).filter(_ =>
```
tgodzik, line 229: What if owner also have conflict, shouldn't we go up the onwer chain in this case?

dos65: > What if owner also have conflict, shouldn't we go up the onwer chain in this case?

Yep, that's what I noticed too but when I started doing that properly it opened much more questions than I expected and it led to a bigger rework. At least right now it works as previously + as in Scala2

I have wip branch for that.

tgodzik: Ok sure, we can improve it later on. If you don't manage to create a PR, could you open an issue mentioning that?

tgodzik (approved):

mtags/src/main/scala-3/scala/meta/internal/pc/AutoImports.scala
dos65: I'm planning to turn it into PR during this week.
If it won't happen I will create an issue

PR #4346, opened 2022-09-05
Auto approve PR by scalameta-bot

tanishiking:
part of https://github.com/scalameta/metals/issues/4333

- Auto-approve the PRs that is created by `scalameta-bot` and has label `semver-spec-patch`.
- `scalameta-bot` will approve the PR

Added this auto-approve workflow because scalameta/metals
requires PRs to receive at least one approval from authorized users
(maintainers) to merge.
Therefore, until someone approve the PR, mergify cannot merge the PR.

This workflow will self-approve (by `scalameta-bot`) and enable mergify
to auto-merge the library update PRs.

---

- I didn't use mergify's auto-review functionality, because mergify
  cannot specify a user who approve the PR (with a free OSS plan)
  - https://docs.mergify.com/actions/review/
- This action doesn't depends on other workflows result, which means will auto-approve even if the CIs fail.
  - It's ideal approve only when CIs passed, but depending on other workflows' status in github actions are bit complicated (in comparison to mergify), let mergify check the CI statuses.
  - see: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow
- concerns
  - Can scalameta-bot self-approve?
  - Does scalameta-bot's approval unlock merging? (maybe we have to give
    scalameta-bot write right to this repo?)

PR #4404, opened 2022-09-19
ci(Mergify): configuration update

tanishiking:
This change has been made by @tanishiking from https://mergify.com config editor.

close https://github.com/scalameta/metals/issues/4333

This PR enables mergify in `scalameta/metals` and will auto-merge the PRs that is

- created by `scalameta-bot`
- approved
  - will be auto-approved by GitHub actions: see: https://github.com/scalameta/metals/pull/4346
- all tests passed (currently checking the number of `check-success` instead of listing all the checks)
- has labels (both of them should be labeled by scala-steward (scalameta-bot)) see https://github.com/scalameta/metals/pull/4337
  - `semver-spec-patch`
  - `library-update`

Some conditions are duplicated with the auto-approve code action, but I think it's safer to check those conditions again in mergify side.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,11 @@
+pull_request_rules:
+  - name: Automatic merge on approval
+    conditions:
+      - "#approved-reviews-by>=1"
+      - author=scalameta-bot
+      - "#check-success>=22"
+      - label=semver-spec-patch
+      - label=library-update
+    actions:
+      merge:
+        method: merge


.mergify.yml
```diff
@@ -0,0 +1,11 @@
+pull_request_rules:
+  - name: Automatic merge on approval
+    conditions:
+      - "#approved-reviews-by>=1"
+      - author=scalameta-bot
+      - "#check-success>=22"
```
tgodzik, line 6: Any chance to make it on all jobs? We might forget to modify it at some point

tgodzik (approved):

.mergify.yml
tanishiking: Maybe we can define another github action that checks all other (important) tests passing.

tgodzik: We can follow up later with that :+1:

PR #4405, opened 2022-09-19
Ivy Completions Dash Fix

LaurenceWarne:
Hi!  This is a small PR which fixes TextEdit ranges when `-` appears in a dependency used for ivy completions.  Previously for example, selecting `io.get-coursier` at `io.get-coursi|` would leave you with `io.get-io.get-coursier`.

I tried to find out if there were any other chars that potentially could appear in artefact ids, https://maven.apache.org/guides/mini/guide-naming-conventions.html is not overly specific: ` then you can choose whatever name you want with lowercase letters and no strange symbols` :laughing: 

Also snuck in: remove setting the Scala binary version when creating the coursier api which is no longer needed since https://github.com/scalameta/metals/pull/4390 (mentioned in previous PR).


tgodzik (approved):
LGTM! Thanks!

PR #4408, opened 2022-09-19
bugfix Try to fix flaky import tests

tgodzik:
It seems that we might have a race condition, which causes the build import not to run the second time if it hasn't yet finished


jkciesluk:
So this can be the same issue as with `CancelCompileLspSuite` (although there not only the test is flaky)

tgodzik:
> So this can be the same issue as with `CancelCompileLspSuite` (although there not only the test is flaky)

Yeah, could be, but let's merge this separately.

tanishiking (approved):
LGTM!
I agree with @jkciesluk but let's do that in separate PRs :+1:

PR #4410, opened 2022-09-19
fix: do not show ImportAlreadyRunning warning in case of build change

dos65:
In case if you already have popup about `Changes detected - Import Build` and you introduce another change to build or switch branch it show improper message because `slowConnectToBuildServer` is awaiting the response on the first popup

### DIFF
--- a/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
+++ b/metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
@@ -2291,7 +2291,8 @@ class MetalsLanguageServer(
   }
 
   private def slowConnectToBuildServer(
-      forceImport: Boolean
+      forceImport: Boolean,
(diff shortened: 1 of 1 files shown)


metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -2291,7 +2291,8 @@ class MetalsLanguageServer(
   }
 
   private def slowConnectToBuildServer(
-      forceImport: Boolean
+      forceImport: Boolean,
+      isBuildChangeNotification: Boolean = false,
```
kpodsiad, line 2295: could you add scalafoc for `isBuildChangeNotification`? In a few months it won't be clear why this parameter was added

tgodzik:
We reworked this in another PR, it should work ok now. Thanks @dos65 for raising this!

Issue #4004, opened 2022-06-07
Empty parentheses added to completed extension method


prolativ:
### Describe the bug

```scala
extension (s: String)
  def foo = 0
val bar = "abc".fo
```
Invoking code completion on `"abc".fo` results in `"abc".foo()` (which won't compile).

### Expected behavior

The code should be completed with just `"abc".foo` without following parentheses.

### Operating system

macOS

### Editor/Extension

VS Code

### Version of Metals

0.11.6+48-26045a93-SNAPSHOT

prolativ:
A probably related issue:
```scala
extension (s: String)
  def bar = 0
  def bax() = 0
  def baz()() = 0
val bar = "abc".ba
```
Here the suggested completions for `"abc".ba` are all displayed as parameterless methods returning an `Int`:
```scala
bar: Int
bax: Int
baz: Int
```
but they should be displayed as
```scala
bar: Int
bax(): Int
baz()(): Int
```

riiswa:
Hi, I think I have fixed the bug I am now trying to write a test. Is this the right file: `metals/tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala` ?

ckipp01:
> Hi, I think I have fixed the bug I am now trying to write a test. Is this the right file: `metals/tests/cross/src/test/scala/tests/pc/CompletionMatchSuite.scala` ?

Hey @riiswa, that specific suite is for the completions related to the `match` completion functionality. Check out the general `CompletionSuite.scala`, and you should be able to stick it in there. Also note how the `compat` param works as you might want different behavior with Scala 2 or Scala 3.

tgodzik:
It's probably bet to add it to `CompletionSnippetSuite` with a `tag(IgnoreScala2)` as the extension methods don't work for Scala 2.

PR #4412, opened 2022-09-20
refactor: [scala3] findSuffix to return a data structure so we can tell what kinds of suffix will be added to the edit.

tanishiking:
related to https://github.com/scalameta/metals/issues/4358

This commit refactors `findSuffix` to return `CompletionSuffix` that has information about what kind of completion suffix will be added: does compeltion suffix contains `()`, `[]`, or `{}`? where are we gonna move the cursor to?

This refactoring is required for improving the completion (with suffix)'s label.
see: https://github.com/scalameta/metals/pull/4369

### DIFF
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionSuffix.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionSuffix.scala
@@ -13,7 +13,7 @@ case class CompletionSuffix(
     snippet: SuffixKind,
 ):
   def hasSnippet = snippet != SuffixKind.NoSuffix
-  def map(copyFn: CompletionSuffix => CompletionSuffix) = copyFn(this)
+  def chain(copyFn: CompletionSuffix => CompletionSuffix) = copyFn(this)
   def toEdit: String =
     if !hasSuffix then ""
     else
--- a/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionSuffix.scala
+++ b/mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionSuffix.scala
@@ -1,0 +1,48 @@
+package scala.meta.internal.pc.completions
+
+/**
+ * @param brace should we add "()" suffix?
+ * @param bracket should we add "[]" suffix?
+ * @param template should we add "{}" suffix?
+ * @param snippet which suffix should we insert the snippet $0
+ */
+case class CompletionSuffix(
+    brace: Boolean,
+    bracket: Boolean,
+    template: Boolean,
+    snippet: SuffixKind,
+):
+  def hasSnippet = snippet != SuffixKind.NoSuffix
+  def map(copyFn: CompletionSuffix => CompletionSuffix) = copyFn(this)
+  def toEdit: String =
+    if !hasSuffix then ""
+    else
+      val braceSuffix =
+        if brace && snippet == SuffixKind.Brace then "($0)"
+        else if brace then "()"
+        else ""
+      val bracketSuffix =
+        if bracket && snippet == SuffixKind.Bracket then "[$0]"
+        else if bracket then "[]"
+        else ""
+      val templateSuffix =
+        if template && snippet == SuffixKind.Template then " {$0}"
+        else if template then " {}"
+        else ""
+      s"$bracketSuffix$braceSuffix$templateSuffix"
+  def toEditOpt: Option[String] =
+    val edit = toEdit
+    if edit.nonEmpty then Some(edit) else None
+  private def hasSuffix = brace || bracket || template
+end CompletionSuffix
+
+object CompletionSuffix:
+  val empty = CompletionSuffix(
+    brace = false,
+    bracket = false,
+    template = false,
+    snippet = SuffixKind.NoSuffix,
+  )
+
+enum SuffixKind:
+  case Brace, Bracket, Template, NoSuffix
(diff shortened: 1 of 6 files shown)


tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala
```diff
@@ -488,21 +488,4 @@ class CompletionSnippetSuite extends BaseCompletionSuite {
...
-        |    def bar()() = 0
-        |  val bar = "abc".ba@@
-    """.stripMargin,
-    s"""|package a
-        |object Foo:
-        |  extension (s: String)
-        |    def bar()() = 0
-        |  val bar = "abc".bar()()
-    """.stripMargin,
-  )
```
tanishiking, line 506: Now we can't add double `()` for something like `def bar()()` unfortunately. (originally included by https://github.com/scalameta/metals/issues/4004 )
However, I don't think it's useful in practice because I don't think people write like this signature 🤔

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -241,52 +241,51 @@ class Completions(
...
-      if shouldAddSnippet && cursorPos.allowTemplateSuffix
-        && isAbstractType(symbol)
-      then
-        if bracketSuffix.nonEmpty || bracesSuffix.contains("$0") then " {}"
-        else " {$0}"
-      else ""
-
-    val concludedSuffix = bracketSuffix + bracesSuffix + templateSuffix
-    if concludedSuffix.nonEmpty then Some(concludedSuffix) else None
+  private def findSuffix(symbol: Symbol): CompletionSuffix =
```
tanishiking, line 244: it would be easier to review with `?w=1`

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -241,52 +241,51 @@ class Completions(
...
+    CompletionSuffix.empty
+      .map { suffix => // for [] suffix
+        if shouldAddSnippet &&
+          cursorPos.allowBracketSuffix && symbol.info.typeParams.nonEmpty
+        then suffix.copy(bracket = true, snippet = SuffixKind.Bracket)
+        else suffix
+      }
+      .map { suffix => // for () suffix
```
tgodzik, line 252: Having map here is a bit weird, since I would expect some kind of collection to be used. Could we stay with the `if-else` staements?

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionSuffix.scala
```diff
@@ -0,0 +1,48 @@
...
+      val braceSuffix =
+        if brace && snippet == SuffixKind.Brace then "($0)"
+        else if brace then "()"
+        else ""
+      val bracketSuffix =
+        if bracket && snippet == SuffixKind.Bracket then "[$0]"
+        else if bracket then "[]"
+        else ""
+      val templateSuffix =
+        if template && snippet == SuffixKind.Template then " {$0}"
```
tgodzik, line 29: Is it expected to get all the suffixes at the same time like for example `Foo[](){}` when we have `trait Foo[T](a: Int)`?

tgodzik (changes requested):
Thanks for improving this! I am not sure about the map approach though.

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionSuffix.scala
tanishiking: Yes!

mtags/src/main/scala-3/scala/meta/internal/pc/completions/Completions.scala
tanishiking: what do you say if I rename `map` to something like `chain` or something?

---

I don't want to stay `if-else` because if we write in that way it's gonna be something like

```scala
val empty = CompletionSuffix.empty
val bracketSuffix = { if .. empty.copy(...) else ... }
val braceSuffix = { if ... bracketSuffix.copy(...) else ...}
val templateSuffix = { if ... braceSuffix.copy(...) else ... }
templateSuffix
```

and it's error-prone because we can access `bracketSuffix` from `templateSuffix = {...}` where it should only access `braceSuffix`.

tgodzik: ok sure, that's also fine.

tanishiking: renamed https://github.com/scalameta/metals/pull/4412/commits/ac8d8e90e2597084f30b28848019cbca3618324b

mtags/src/main/scala-3/scala/meta/internal/pc/completions/CompletionSuffix.scala
tgodzik: `Foo[$0]($0){$0} ` would not work though, we would need `Foo[$0]($1){$2} ` which would mean that users could jump with tab to next positions, we shouldn't have multiple `$0`

tanishiking: As `CompletionSuffix` has an enum for where to put the `$0`, there shouldn't be a case like `Foo[$0]($0){$0}`.

tgodzik: ach ok, it might be an improvement to do `$0,$1` etc.

PR #4413, opened 2022-09-20
refactor: Make it easier to define commands to run in code actions 

tgodzik:
This PR moves implementations of executing code actions commands to the actual code action, which makes it easier for users to implement new ones without searching multiple places. The now only need to implement one action and add it to the list, everything else is taken care of.


tanishiking (approved):
noice, LGTM!

metals/src/main/scala/scala/meta/internal/metals/MetalsLanguageServer.scala
```diff
@@ -2040,84 +2042,13 @@ class MetalsLanguageServer(
...
-            workspaceEdit = new l.WorkspaceEdit(Map(uri -> edits).asJava)
-            _ <- languageClient
-              .applyEdit(new ApplyWorkspaceEditParams(workspaceEdit))
-              .asScala
-          } yield ().asInstanceOf[Object]
-        }
-      case ServerCommands.ExtractMethod(
-            ServerCommands.ExtractMethodParams(doc, range, extractionPos)
+      case actionCommand
+          if codeActionProvider.allActionCommands.exists(
```
kpodsiad, line 2046: Wdyt od having Set with all code actions ids?

tgodzik: Addeded for both ServerCommands and code action commands.

metals/src/main/scala/scala/meta/internal/metals/ServerCommands.scala
```diff
@@ -620,6 +620,9 @@ object ServerCommands {
       StartScalaCliServer,
       StopScalaCliServer,
     )
+
+  def allIds: Set[String] = all.map(_.id).toSet
```
kpodsiad, line 624: why not `val` 🤔

tgodzik: Well, because I am quite blind and didn't realize what I wrote :sweat:

PR #4414, opened 2022-09-20
Scala 2.13.9

dos65:
Due to https://github.com/scala/scala/pull/9754 it was required to add additional compat method to filter out non relevant type completions.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -308,6 +308,17 @@ val mtagsSettings = List(
   buildInfoKeys := Seq[BuildInfoKey](
     "scalaCompilerVersion" -> scalaVersion.value
   ),
+  Compile / unmanagedSourceDirectories := {
+    val current = (Compile / unmanagedSourceDirectories).value
+    val base = (Compile / sourceDirectory).value
+    val regex = "(\\d+)\\.(\\d+)\\.(\\d+).*".r
+    // For scala 2.13.9 we need to have a special Compat.scala
+    // For this case filter out `scala-2.13` directory that comes by default
+    if (scalaVersion.value == "2.13.9")
+      current.filter(f => f.getName() != "scala-2.13")
+    else
+      current
+  },
 )
 
 lazy val mtags3 = project
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
@@ -328,7 +328,8 @@ class CompletionProvider(
         !isFileAmmoniteCompletion() &&
         completion.isCandidate(head) &&
         !head.sym.name.containsName(CURSOR) &&
-        isNotLocalForwardReference
+        isNotLocalForwardReference &&
+        !isAliasCompletion(head)
       ) {
         isSeen += id
         buf += head
--- a/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
+++ b/mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
@@ -85,11 +85,11 @@ trait Completions { this: MetalsGlobal =>
    */
   def relevancePenalty(m: Member): Int =
     m match {
-      case TypeMember(sym, _, true, isInherited, _) =>
+      case tm: TypeMember if tm.accessible =>
         computeRelevancePenalty(
-          sym,
+          tm.sym,
           m.implicitlyAdded,
-          isInherited
+          tm.inherited
         )
       case w: WorkspaceMember =>
         MemberOrdering.IsWorkspaceSymbol + w.sym.name.length()
@@ -101,9 +101,9 @@ trait Completions { this: MetalsGlobal =>
         ) >>> 15
         if (!w.sym.isAbstract) penalty |= MemberOrdering.IsNotAbstract
         penalty
-      case ScopeMember(sym, _, true, _) =>
+      case sm: ScopeMember if sm.accessible =>
         computeRelevancePenalty(
-          sym,
+          sm.sym,
           m.implicitlyAdded,
           isInherited = false
         )
--- a/tests/unit/src/test/resources/definition/example/Scalalib.scala
+++ b/tests/unit/src/test/resources/definition/example/Scalalib.scala
@@ -1,7 +1,7 @@
 package example
 
 class Scalalib/*Scalalib.scala*/ {
-  val nil/*Scalalib.scala*/ = List/*package.scala*/()
+  val nil/*Scalalib.scala*/ = List/*List.scala*/()
   val lst/*Scalalib.scala*/ = List/*package.scala*/[
     (
         Nothing,
--- a/tests/unit/src/test/resources/semanticdb/example/Scalalib.scala
+++ b/tests/unit/src/test/resources/semanticdb/example/Scalalib.scala
@@ -1,7 +1,7 @@
 package example
 
 class Scalalib/*example.Scalalib#*/ {
-  val nil/*example.Scalalib#nil.*/ = List/*scala.package.List.*/()
+  val nil/*example.Scalalib#nil.*/ = List/*scala.collection.immutable.Nil.*/()
   val lst/*example.Scalalib#lst.*/ = List/*scala.package.List.*/[
     (
         Nothing/*scala.Nothing#*/,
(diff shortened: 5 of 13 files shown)


build.sbt
```diff
@@ -308,6 +308,21 @@ val mtagsSettings = List(
   buildInfoKeys := Seq[BuildInfoKey](
     "scalaCompilerVersion" -> scalaVersion.value
   ),
+  Compile / unmanagedSourceDirectories := {
+    val current = (Compile / unmanagedSourceDirectories).value
+    val base = (Compile / sourceDirectory).value
+    val regex = "(\\d+)\\.(\\d+)\\.(\\d+).*".r
+    // Starting from scala 2.13.9 we need to specify
+    // Compat.scala per each patch version
```
kpodsiad, line 316: Why, what changed?

mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
```diff
@@ -328,7 +328,8 @@ class CompletionProvider(
         !isFileAmmoniteCompletion() &&
         completion.isCandidate(head) &&
         !head.sym.name.containsName(CURSOR) &&
-        isNotLocalForwardReference
+        isNotLocalForwardReference &&
+        !isAliasCompletion(head)
```
dos65, line 332: @kpodsiad @tanishiking 
I updated PR description.

This line was added because [this test](https://github.com/scalameta/metals/blob/c2c6b7077f46b9d30e9be3806691e792919bfa92/tests/cross/src/test/scala/tests/pc/CompletionSnippetSuite.scala#L94-L107) and similar to it started failing.

Having the following source:
```
object Main {
   type Bar = List[Int]
   val x: Ba@@
}
```
compiler started returning completion: List(`Bar`, `List`) which is incorrect and might be filtered out by checking over a new field `aliasInfo` in Type and Scope Member

mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -85,11 +85,11 @@ trait Completions { this: MetalsGlobal =>
    */
   def relevancePenalty(m: Member): Int =
     m match {
-      case TypeMember(sym, _, true, isInherited, _) =>
```
tgodzik, line 88: Are we also not matching on `true` ?

mtags/src/main/scala-2/scala/meta/internal/pc/CompletionProvider.scala
tgodzik: `List` should be actually removed due to fuzzy matching, the compiler can return `List` , but it should not be used by Metals

tgodzik: Ach I see, the compiler gives us the name `MT` instead of List, that seems like a bug really :thinking: 

I am looking if we can avoid adding new directory for Scala 2.13.9

tgodzik: Alternative I found is:
```
def isAliasType = head.getClass().getMethods().find(_.getName() == "aliasInfo") match {
          case Some(aliasType) => aliasType.invoke(head) match {
            case Some(_) => true
            case _ => false
          }
          case _ => false

        }
```

though that uses reflection :/

tgodzik (changes requested):
Should we maybe instead file a bug, change the test or ignore it for 2.13.9 and fix it in the compiler?

It seems this change https://github.com/scala/scala/pull/9754/files

dos65:
> Should we maybe instead file a bug, change the test or ignore it for 2.13.9 and fix it in the compiler?
> It seems this change https://github.com/scala/scala/pull/9754/files

@tgodzik Maybe. Need to check if it the only case . I'm not fully sure if don't have any additional cases when we are getting non-matching symbols from completions - https://github.com/scalameta/metals-vscode/pull/738

mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -85,11 +85,11 @@ trait Completions { this: MetalsGlobal =>
...
+      case tm: TypeMember =>
```
tgodzik, line 88:
```suggestion
      case tm: TypeMember  if tm.accessibile =>
```

mtags/src/main/scala-2/scala/meta/internal/pc/completions/Completions.scala
```diff
@@ -101,9 +101,9 @@ trait Completions { this: MetalsGlobal =>
         ) >>> 15
         if (!w.sym.isAbstract) penalty |= MemberOrdering.IsNotAbstract
         penalty
-      case ScopeMember(sym, _, true, _) =>
+      case sm: ScopeMember =>
```
tgodzik, line 104:
```suggestion
      case sm: ScopeMember if sm.accessibile =>
```

tgodzik (changes requested):
We can merge as is to unblock 2.13.9 and we can ponder on the topic later. I would just add the `if tm.accessabile` part before merging

tests/unit/src/test/resources/semanticdb/example/Scalalib.scala
```diff
@@ -1,7 +1,7 @@
 package example
 
 class Scalalib/*example.Scalalib#*/ {
-  val nil/*example.Scalalib#nil.*/ = List/*scala.package.List.*/()
+  val nil/*example.Scalalib#nil.*/ = List/*scala.collection.immutable.Nil.*/()
```
dos65, line 4: and this one

tests/unit/src/test/resources/definition/example/Scalalib.scala
```diff
@@ -1,7 +1,7 @@
 package example
 
 class Scalalib/*Scalalib.scala*/ {
-  val nil/*Scalalib.scala*/ = List/*package.scala*/()
+  val nil/*Scalalib.scala*/ = List/*List.scala*/()
```
dos65, line 4: @tgodzik this one looks like an issue. wdyt?

dos65:
@tgodzik I took a look and it seems that the thing with aliases isn't compiler issue.
we use `completionsAt`[https://github.com/scala/scala/blob/2.13.x/src/repl/scala/tools/nsc/interpreter/PresentationCompilation.scala#L122] method from Global while the REPL uses `completionCadidates`[https://github.com/scala/scala/blob/2.13.x/src/repl/scala/tools/nsc/interpreter/PresentationCompilation.scala#L191]. If you look at `completionCadidates` there is some nameFiltering logic.

tests/unit/src/test/resources/definition/example/Scalalib.scala
dos65: Ok, nevermind, it isn't an issue everything is fine

tgodzik:
> @tgodzik I took a look and it seems that the thing with aliases isn't compiler issue. we use [`completionsAt`](https://github.com/scala/scala/blob/2.13.x/src/repl/scala/tools/nsc/interpreter/PresentationCompilation.scala#L122) method from Global while the REPL uses [`completionCadidates`](https://github.com/scala/scala/blob/2.13.x/src/repl/scala/tools/nsc/interpreter/PresentationCompilation.scala#L191). If you look at `completionCandidates` there is some nameFiltering logic.

I think it's an issue with the compiler actually: https://github.com/scala/scala/pull/10148

Name matcher should not supply something that in reality has a different name. The API would be super misleading that way, I think the original fix was a bit of a hack and we should just dealias the type when needed.

dos65:
@tgodzik Let's keep this compat thing at least for a few Netals versions until there will be a new 2.13.x release. I don't think it will be hard to maintain. wdty?

build.sbt
```diff
@@ -308,6 +308,21 @@ val mtagsSettings = List(
...
+    // For this case filter out `scala-2.13` directory that comes by default
+    scalaVersion.value match {
```
tgodzik, line 318:
```suggestion
    if (scalaVersion.value == 2.13.9) {
```
so that we only do it for Scala 2.13.9 and it's fixed later on?

tgodzik (changes requested):
We can merge as is, but I would simplify it a bit since this should only happen for Scala 2.13.9
I raised the PR for upgrading Scala version for debug adapter.

tgodzik (approved):

